PR #6985, opened 2024-12-12
add job_id to the name of functional tests

ychebotarev:
## What changed?

Add job_id to the name of functional tests
## Why?

To be able to find a job_id of a failed test.


.github/workflows/run-tests.yml
```diff
@@ -492,7 +492,7 @@ jobs:
         uses: actions/upload-artifact@v4.4.3
         if: ${{ !cancelled() && !inputs.run_single_functional_test }}
         with:
-          name: junit-xml--${{github.run_id}}--${{github.run_attempt}}--${{matrix.runs-on}}--${{matrix.name}}--${{matrix.shard_index}}--functional-test
+          name: junit-xml--${{github.run_id}}--${{github.job_id}}--${{github.run_attempt}}--${{matrix.runs-on}}--${{matrix.name}}--${{matrix.shard_index}}--functional-test
```
dandavison, line 495: Looks like it should be `github.job`: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

dandavison: Actually, I'm realizing that this job ID is not the numeric identifier we are looking for. `github.job` is an identifier for the job _type_, whereas what we want is the numeric job ID used in the URL. For some reason GitHub doesn't make it easy to find that. However, on the plus side we do have a workflow that shows how to obtain this:

https://github.com/temporalio/saas-cicd/blob/11b3918efe21f4d4aafecb9631c428cd2d17b10c/.github/workflows/execute-oss-cicd-workflow.yaml#L126
https://github.com/temporalio/saas-cicd/blob/main/oss-cicd/internal/github/get-job-id-for-job-name

ychebotarev: so much magic...thanks Dan

.github/workflows/run-tests.yml
```diff
@@ -487,12 +488,24 @@ jobs:
...
+        id: get_job_id
+        run: |
+          set -e          
+          job_id=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{secrets.GITHUB_TOKEN}}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/"${GITHUB_REPOSITORY}"/actions/runs/"${GITHUB_RUN_ID}"/jobs \
+            | jq "first(.jobs[] | select(.name | endswith(\"$job_name\")) | .id)")
+          echo "JOB_ID=${job_id}" >> $GITHUB_OUTPUT
```
dnr, line 501: it's silly but this is what it's complaining about:
```suggestion
          echo "JOB_ID=${job_id}" >> "$GITHUB_OUTPUT"
```

.github/workflows/run-tests.yml
```diff
@@ -487,12 +488,24 @@ jobs:
...
 
+      - name: Get job ID from workflow run
+        id: get_job_id
+        run: |
+          set -e          
+          job_id=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{secrets.GITHUB_TOKEN}}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/"${GITHUB_REPOSITORY}"/actions/runs/"${GITHUB_RUN_ID}"/jobs \
```
dnr, line 499: I'd write this like
```suggestion
            "https://api.github.com/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/jobs" \
```
what you have is equivalent but the quotes on outside reads better

dnr (commented):
it's silly but that's what it's complaining about

.github/workflows/run-tests.yml
```diff
@@ -487,12 +488,24 @@ jobs:
           annotate_only: true
           skip_annotations: true
 
+      - name: Get job ID from workflow run
+        id: get_job_id
+        run: |
+          set -e
```
bergundy, line 494:
```suggestion
          set -euo pipefail
```

bergundy: I think this may be easier to obtain with the `gh` cli: https://stackoverflow.com/questions/71240338/obtain-job-id-from-a-workflow-run-using-contexts.

.github/workflows/run-tests.yml
```diff
@@ -487,12 +488,24 @@ jobs:
           annotate_only: true
           skip_annotations: true
 
+      - name: Get job ID from workflow run
+        id: get_job_id
+        run: |
+          set -e          
+          job_id=$(curl -L \
```
bergundy, line 495: I'd add a `--fail` too.

.github/workflows/run-tests.yml
```diff
@@ -487,12 +488,24 @@ jobs:
...
+            | jq "first(.jobs[] | select(.name | endswith(\"$job_name\")) | .id)")
+          echo "JOB_ID=${job_id}" >> "$GITHUB_OUTPUT"
+
       - name: Upload test results
         # Can't pin to major because the action linter doesn't recognize the include-hidden-files flag.
         uses: actions/upload-artifact@v4.4.3
         if: ${{ !cancelled() && !inputs.run_single_functional_test }}
         with:
-          name: junit-xml--${{github.run_id}}--${{github.run_attempt}}--${{matrix.runs-on}}--${{matrix.name}}--${{matrix.shard_index}}--functional-test
+          name: junit-xml--${{github.run_id}}--${{steps.get_job_id.outputs.JOB_ID}}--${{github.run_attempt}}--${{matrix.runs-on}}--${{matrix.name}}--${{matrix.shard_index}}--functional-test
```
bergundy, line 508: Is this only relevant to functional tests? What about unit tests and xdc?

.github/workflows/run-tests.yml
```diff
@@ -444,6 +444,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           ref: ${{ env.COMMIT }}
 
+
```
bergundy, line 447:
```suggestion
```

.github/workflows/run-tests.yml
ychebotarev: add to what?

.github/workflows/run-tests.yml
ychebotarev: add xdc and ndc func tests as well. The main purpose if functional tests. We can add unit tests later if needed

.github/workflows/run-tests.yml
ychebotarev: I add it, and then remove it. For some reason it start failing with the changes you proposed. It is not clear what exactly was causing it to fail, so I rollback to a working solution.

.github/workflows/run-tests.yml
```diff
@@ -487,12 +487,24 @@ jobs:
...
+        id: get_job_id
+        run: |
+          set -e
+          job_id=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{secrets.GITHUB_TOKEN}}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/jobs" \
+            | jq "first(.jobs[] | select(.name | endswith(\"$job_name\")) | .id)")
+          echo "JOB_ID=${job_id}" >> "$GITHUB_OUTPUT"
```
dandavison, line 500: Personally I'd probably store this as a shell script in the repo to avoid repeating it 3 times (and for potential re-use since this has now come up in `temporal` and `saas-cicd`). (But not blocking.)

dandavison (approved):

.github/workflows/run-tests.yml
```diff
@@ -487,12 +487,24 @@ jobs:
...
+      - name: Get job ID from workflow run
+        id: get_job_id
+        run: |
+          set -e
+          job_id=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{secrets.GITHUB_TOKEN}}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/jobs" \
+            | jq "first(.jobs[] | select(.name | endswith(\"$job_name\")) | .id)")
```
dandavison, line 499: The `job_name` variable is unset. I believe that's why we're getting the same job ID `34400904242` for every job:

```
Artifact junit-xml--12324062875--34400904242--1--ubuntu-20.04--postgres12--1--functional-test has been successfully uploaded!
```

(Could do `set -u` to catch unset bash variables)

.github/workflows/run-tests.yml
ychebotarev: you right, it is the same...sigh

.github/workflows/run-tests.yml
ychebotarev: yeah, going to try few more approaches.

dandavison:
Looks like the expected job ID shell variable name isn't set in this one (there might be others, I just happened to see this one)

```
Artifact junit-xml--12362620891----1--ubuntu-20.04--mysql8--2--functional-test has been successfully uploaded! Final size is 428562 bytes. Artifact ID is 2329024984
```

is it surprising that didn't cause a failure due to `set -eu`?

ychebotarev:
> Looks like the expected job ID shell variable name isn't set in this one (there might be others, I just happened to see this one)
> 
> ```
> Artifact junit-xml--12362620891----1--ubuntu-20.04--mysql8--2--functional-test has been successfully uploaded! Final size is 428562 bytes. Artifact ID is 2329024984
> ```
> 
> is it surprising that didn't cause a failure due to `set -eu`?

that job doesn't exist in:
https://api.github.com/repos/temporalio/temporal/actions/runs/12362620891/jobs
thus we can do nothing about it

.github/workflows/run-tests.yml
bergundy: Hmm.. I wouldn't expect this to be an issue. Can you check why `--fail` is failing? It's generally good practice to have this set since it ensures you get a 2xx response.

.github/workflows/run-tests.yml
```diff
@@ -580,6 +597,23 @@ jobs:
           path: ~/.cache/go-build
           key: go-${{ runner.os }}${{ runner.arch }}-build-${{ env.COMMIT }}
 
+      - name: Get job ID from workflow run
+        id: get_job_id
+        run: |
+          set -eu
```
bergundy, line 603:
```suggestion
          set -euo pipefail
```

In case the commands after the pipe fail.

.github/workflows/run-tests.yml
```diff
@@ -580,6 +597,23 @@ jobs:
           path: ~/.cache/go-build
           key: go-${{ runner.os }}${{ runner.arch }}-build-${{ env.COMMIT }}
 
+      - name: Get job ID from workflow run
+        id: get_job_id
+        run: |
+          set -euo pipefail
```
bergundy, line 603: This needs to be applied to all jobs.

ychebotarev: ok

alexshtin (approved):
I don't know much about GHA and bash but it looks reasonable good to me (assuming it works).

.github/workflows/run-tests.yml
bergundy: I see this PR was merged without addressing this comment.

bergundy:
I don't understand still why we're allowing the curl command to silently fail, in general this is bad practice. I would rather know if we have issues getting the status.

ychebotarev:
> I don't understand still why we're allowing the curl command to silently fail, in general this is bad practice. I would rather know if we have issues getting the job ID.

I already explained that. There is no benefit at failing job if that curl fail. There is no benefit at failing job if extracting job_id fail. Etc.

PR #6986, opened 2024-12-13
Delete Namespace: send delete executions heartbeats from separate go routine

alexshtin:
## What changed?

Delete Namespace: send delete executions heartbeats from separate go routine.

## Why?

Following best practices heartbeats should be sent outside of main activity function in case if operation takes longer than expected. Also `DeleteExecutionsActivity` doesn't send heartbeats on every deleted execution but sends accumulated result every 5 seconds.

## How did you test it?

Existing tests and manual run.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/service/worker/deletenamespace/deleteexecutions/activities.go
+++ b/service/worker/deletenamespace/deleteexecutions/activities.go
@@ -26,6 +26,7 @@ package deleteexecutions
 
 import (
 	"context"
+	"time"
 
 	"go.temporal.io/api/serviceerror"
 	workflowpb "go.temporal.io/api/workflow/v1"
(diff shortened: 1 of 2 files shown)


service/worker/deletenamespace/deleteexecutions/activities.go
```diff
@@ -126,10 +127,29 @@ func (a *LocalActivities) GetNextPageTokenActivity(ctx context.Context, params G
...
 
-	rateLimiter := quotas.NewRateLimiter(float64(params.RPS), params.RPS)
+	progressCh := make(chan DeleteExecutionsActivityResult, 1)
+	defer func() { close(progressCh) }()
+
+	go func() {
+		heartbeatTicker := time.NewTicker(deleteWorkflowExecutionsActivityOptions.HeartbeatTimeout / 2)
+		defer heartbeatTicker.Stop()
+
+		var lastKnownProgress DeleteExecutionsActivityResult
```
bergundy, line 137: This should be initialized to the previous heartbeat details to avoid losing progress.

service/worker/deletenamespace/deleteexecutions/activities.go
```diff
@@ -175,11 +196,14 @@ func (a *Activities) DeleteExecutionsActivity(ctx context.Context, params Delete
...
 		}
-		activity.RecordHeartbeat(ctx, result)
 		select {
+		case progressCh <- result:
+			// Send the current result to heartbeat go routine.
 		case <-ctx.Done():
+			// Stop deletion on cancellation.
 			return result, ctx.Err()
 		default:
+			// Don't block deletion if a progress channel is full.
```
bergundy, line 206: Are you sure that's desirable, probably better to ensure that you can keep heartbeating.
Maybe just give the channel a larger buffer size? Or even just use an atomic pointer instead of a channel to set the latest result.

You could create a background context `WithCancel` for the heartbeating goroutine and `defer cancel()` instead of using this channel.

service/worker/deletenamespace/deleteexecutions/activities.go
alexshtin: This is basically statistics and tool to monitor the progress in UI. It won't be 100% accurate and nothing really depends on it. But, I agree, `SuccessCount` should be carried over to be more accurate.

service/worker/deletenamespace/deleteexecutions/activities.go
alexshtin: I prefer to use channel here. And yes, I believe I don't want to block deletion, if for some reason heartbeat go routine is blocked. In fact, this only can be if `activity.RecordHeartbeat()` call took too long. Progress won't be lost as it is accumulated in the main goroutine.

bergundy (approved):
I would reconsider taking my suggestions but what you have is good enough AFAIC.

PR #6987, opened 2024-12-13
Chasm Interface

yycptt:
## What changed?

- Chasm interface

## Why?

- For the new CHASM framework


service/history/chasm/component.go
bergundy, line 1: I would put this in a top level `chasm` directory. There's likely going to be some chasm related code in other services.

service/history/chasm/components/activity/activity.go
```diff
@@ -0,0 +1,168 @@
...
+	Activity struct {
+		// ALL FIELDS MUST BE EXPORTED for reflection to work
+
+		// In V1, we will only support only one non-chasm.XXX field in the struct.
+		// and that field must be a proto.Message.
+		// TODO: define a serializer/deserializer interface?
+		//
+		// Framework will try to recognize the type and do serialization/deserialization
+		// proto.Message is recommended so the component get compatibility if state definition changes
+		State persistencepb.ActivityInfo // proto.Message
```
bergundy, line 23: Let's call this standard field `Data`, since `State` is usually an enum.

bergundy: Alternatively, we can let you embed a proto and treat that as the data field. That could be slightly nicer to use.

service/history/chasm/components/activity/activity.go
```diff
@@ -0,0 +1,168 @@
...
+
+		// At the end of a transition, framework will use reflection to understant the component
+		// tree structure.
+
+		// TODO: also support field name tag, so the fields can be renamed
+		Input  *chasm.Field[*common.Payload] `chasm:"lazy"`
+		Output *chasm.Field[*common.Payload] `chasm:"lazy"`
+
+		EventNotifier *chasm.Field[EventNotifier]
```
bergundy, line 32: I know this is just to show the pointer capabilities, but the real activity implementation will need to do more than just notify its parent, it'll need to be injected with a way to read the inputs and outputs. They won't be embedded payloads in the workflow case.

service/history/chasm/components/activity/activity.go
```diff
@@ -0,0 +1,168 @@
...
+func (i *Activity) RecordCompleted(
+	chasmContext chasm.MutableContext,
+	req *RecordCompletedRequest,
+) (*RecordCompletedResponse, error) {
+	// say we have a completedTime field in ActivityInfo
+	// i.State.CompletedTime = timestamppb.New(chasmContext.Now())
+	output := &common.Payload{
+		Data: req.Output,
+	}
+	i.Output = chasm.NewData(chasmContext, output)
```
bergundy, line 145: I'd probably call this method `NewDataField` since it constructs a `Field`.

service/history/chasm/components/activity/fx.go
```diff
@@ -0,0 +1,51 @@
...
+	matchingClient matchingservice.MatchingServiceClient
+}
+
+func (l Library) Name() string {
+	return "activity"
+}
+
+func (l Library) Components() []chasm.RegistrableComponent {
+	return []chasm.RegistrableComponent{
+		chasm.NewRegistrableComponent[*Activity](
```
bergundy, line 19: I'm kinda going back and forth between a `RegisterComponent` function and this approach where the library returns the list of components.

The main downside to this approach is that you have to wrap with `NewRegisterableComponent`.

If we do want this approach, we may want a `LibraryBase` struct that has empty implementations for all of the methods so library implementors only provide the items they need registered.

I like that with the concept of a library, you essentially get a namespace for all components and tasks.

service/history/chasm/components/activity/fx.go
bergundy, line 1: This is a library, and should be put in `chasm/lib/activity/` IMHO.

service/history/chasm/components/activity/handler.go
```diff
@@ -0,0 +1,160 @@
+package activity
+
+import (
+	"context"
+	"time"
+
+	"go.temporal.io/server/service/history/chasm"
+)
+
+// This will be nexus
```
bergundy, line 10: 👍 I'll work on a Nexus POC on top of this.

service/history/chasm/components/activity/tasks.go
```diff
@@ -0,0 +1,77 @@
...
+
+type TimeoutTask struct {
+	// similar to component state,
+	// use a proto message if possible
+	TimeoutType int
+}
+
+type TimeoutTaskHandler struct{}
+
+func (h *TimeoutTaskHandler) Validate(
```
bergundy, line 63: A little on the fence if `Validate` should be here or on the Component or the task but this works and has some advantages.

service/history/chasm/components/activity/tasks.go
```diff
@@ -0,0 +1,77 @@
...
+	chasmContext chasm.Context,
+	activity *Activity,
```
bergundy, line 65: I wonder if we reuse tasks for separate components and there will need to be a way to provide a `Validate` function for each of them.

service/history/chasm/components/activity/handler.go
```diff
@@ -0,0 +1,160 @@
...
+	request *RecordStartedRequest,
+) (*RecordStartedResponse, error) {
+	// resp := &RecordStartedResponse{}
+
+	ref, err := chasm.DeserializeComponentRef(request.RefToken)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, startedActivityRef, err := chasm.UpdateComponent(
```
bergundy, line 92: As discussed yesterday, I wish we could get the ref in the update function and keep the ref out of this signature. This is good to start with though.

service/history/chasm/components/activity/fx.go
bergundy: Library specific RPCs, configs, and protos should also be colocated in the lib directory.

service/history/chasm/components/activity/activity.go
```diff
@@ -0,0 +1,168 @@
...
+	if err := chasmContext.AddTask(
+		i,
+		chasm.TaskAttributes{}, // immediate task
+		DispatchTask{},
+	); err != nil {
+		return nil, err
+	}
+	if err := chasmContext.AddTask(
+		i,
+		chasm.TaskAttributes{
```
bergundy, line 85: I like that you've separated this out and not put it in the Task like I had it in my POC.

service/history/chasm/components/activity/handler.go
```diff
@@ -0,0 +1,160 @@
...
+		ctx,
+		ref,
+		func(a *Activity, ctx chasm.Context, _ *GetActivityResultRequest) bool {
+			return a.LifecycleState() == chasm.LifecycleStateCompleted
+		},
+		func(a *Activity, ctx chasm.MutableContext, _ *GetActivityResultRequest) (*GetActivityResultResponse, error) {
+			outputPayload, err := a.Output.Get(ctx)
+			resp.Output = outputPayload.Data
+			return resp, err
+		},
```
bergundy, line 153: I wonder if we'd be better off merging these two like so:

```suggestion
		func(a *Activity, ctx chasm.MutableContext, _ *GetActivityResultRequest) (*GetActivityResultResponse, error) {
			if a.LifecycleState() != chasm.LifecycleStateCompleted {
				return nil, chasm.ErrNoReady / *name TBD */
			}
			outputPayload, err := a.Output.Get(ctx)
			resp.Output = outputPayload.Data
			return resp, err
		},
```

service/history/chasm/engine.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// Not needed for V1
+// func WithEagerLoading(
+// 	paths []ComponentPath,
+// ) OperationOption {
+// 	panic("not implemented")
+// }
+
+func NewEntity[C Component, I any, O any](
```
bergundy, line 92: Alternatively, call this `CreateEntity` but what you have is fine too.

service/history/chasm/engine.go
```diff
@@ -0,0 +1,241 @@
...
+
+// this only applies to NewEntity and UpdateWithNewEntity
+func WithBusinessIDPolicy(
+	reusePolicy BusinessIDReusePolicy,
+	conflictPolicy BusinessIDConflictPolicy,
+) TransitionOption {
+	panic("not implemented")
+}
+
+// Not needed for V1
```
bergundy, line 85: True, this isn't needed but if we were to do it, there's a type safe way using a closure that gets an instance and returns field references.

service/history/chasm/components/activity/activity.go
Sushisource: Agree with naming change. Disagree with proto embed. I'd prefer for it to just be explicit.

service/history/chasm/components/activity/activity.go
```diff
@@ -0,0 +1,168 @@
...
+		},
+	); err != nil {
+		return nil, nil
+	}
+
+	return &ScheduleResponse{}, nil
+}
+
+func (i *Activity) GetDispatchInfo(
+	chasmContext chasm.MutableContext,
```
Sushisource, line 99: Should this be an immutable context?

service/history/chasm/components/activity/activity.go
Sushisource: I also had thought we wanted a way for the parent to be able to decide what notifications it cared about from its children, rather than the children needing to remember to notify the parent. I think that can still be implemented in terms of this a pointer just fine though.

Something as simple as the following I think would be nice, and isn't really even part of the framework, just a useful pattern. I think it keeps things a little more obvious for the activity implementer.

```go
func NewScheduledActivity(
	chasmContext chasm.MutableContext,
	params *NewActivityRequest,
	onComplete func(ActivityCompletedEvent)
) { ... }

// Then workflow constructs like
NewScheduledActivity(ctx, params, func(ce) { me.AsComponentPointer(ctx).OnCompletion(ce) })
```

service/history/chasm/components/activity/handler.go
Sushisource: I think I prefer them separate, because what happens if you mutate something and then say "not ready"? That would be some weird violation that shouldn't be possible, and separate contexts enforces that at the type level.

If you need context from one to the other, make the ready function return `(bool, T)` and `T` is passed into the mutator

service/history/chasm/context.go
```diff
@@ -0,0 +1,39 @@
+package chasm
+
+import (
+	"context"
+	"time"
+)
+
+type Context interface {
+	// Context is not binded to any component,
```
Sushisource, line 9:
```suggestion
	// Context is not bound to any component,
```

service/history/chasm/ref.go
```diff
@@ -0,0 +1,57 @@
...
+// then you can only interact with the top level entity.
+func NewComponentRef(
+	entityKey EntityKey,
+	rootComponentName string,
+) ComponentRef {
+	return ComponentRef{
+		EntityKey: entityKey,
+		// we probably don't even need this,
+		// can make the function generic and find the name from registry
+		rootComponentName: rootComponentName,
```
Sushisource, line 44: Agree. Taking in the type and having components define their name so you operate in terms of types rather than names is nice.

service/history/chasm/context.go
```diff
@@ -0,0 +1,39 @@
...
+	// e.g. HistoryEvent
+
+	// Get a Ref for the component
+	// This ref to the component state at the end of the transition
+	// Same as Ref(Component) method in Context,
+	// this only works for components that already exists at the start of the transition
+	//
+	// If we provide this method, then the method on the engine doesn't need to
+	// return a Ref
+	// NewRef(Component) (ComponentRef, bool)
```
Sushisource, line 38: You could have a `CreationContext` that simply doesn't include this, and that and `MutableContext` both embed an un-exported base interface.

Sushisource (commented):
Overall looking really good to me

service/history/chasm/context.go
yycptt: Problem is new components can be created in any transition not just the first one that creates the entity.

service/history/chasm/ref.go
yycptt: I realized later that if we pass in the component type, then we'll also need a way to access the registry. Either pass in the registry or the function impl reference a global variable... Neither feels like a good idea to me.

I need to better understand how the services will be exposed as nexus and then come back to this. We might solve the routing problem in a different way and can just store shardID in the ref.

service/history/chasm/component.go
yycptt: Synced offline. I've moved the package to top level. 
Though later during implementation, we will have to comment that some exported function are meant for internal implementation use and should not be used by lib/component author.

service/history/chasm/components/activity/handler.go
yycptt: I ended up making the ref type generic as well, and it can be `[]byte | ComponentRef`. 
For api, takes in `[]byte` makes sense.
For task processing, decoded ComponentRef makes more sense as the framework will construct ComponentRef in memory. Always takes in []byte means extra encoding/decoding. Framework also needs to put extra task validation fn in the ref as well.

service/history/chasm/components/activity/handler.go
yycptt: I changed the signature of the validation to return (T, bool, error), though that means the mutator is now taking 4 params...

bergundy: Not sure why there needs to be a mutator in poll or why you need both an error and a bool.

service/history/chasm/components/activity/tasks.go
bergundy: Let's try this out. Hard to tell without building some use cases on top of this.

service/history/chasm/ref.go
bergundy: Hmm... shard ID in the ref might work. I don't think any of this is blocking the initial work.

bergundy (approved):
Overall LGTM, had some followup questions.

I think we're in agreement but just want to double check that the structure will be:

```
chasm/ # framework code live here
chasm/lib # libraries live here
```

service/history/chasm/component.go
alexshtin: Top level, you mean `common/chasm`?

service/history/chasm/component.go
```diff
@@ -0,0 +1,71 @@
...
+
+// func (UnimplementedComponent) Terminate() {}
+
+func (UnimplementedComponent) mustEmbedUnimplementedComponent() {}
+
+// Shall it be named ComponentLifecycleState?
+type LifecycleState int
+
+const (
+	LifecycleStateUnspecified LifecycleState = 0
```
alexshtin, line 32:
```suggestion
	LifecycleStateUnspecified = LifecycleState(0)
```
and then it can be moved to the end of next `const` blocl.

yycptt:
> Overall LGTM, had some followup questions.
> 
> I think we're in agreement but just want to double check that the structure will be:
> 
> ```
> chasm/ # framework code live here
> chasm/lib # libraries live here
> ```

Yes. That will be the structure. I just realized I pushed my changes to my fork but this PR is created based on the branch in temporal repo... so the changes are not reflected in this PR. It's fixed now. Sorry about that. 

Not sure if you want to take another look. The changes are basically same as what I described in the comments.

service/history/chasm/component.go
bergundy: Just `chasm`.

chasm/ref.go
```diff
@@ -0,0 +1,61 @@
...
+	// From the component name, we can find the component struct definition
+	// use reflection to find sub-components and understand if those sub-components
+	// needs to be loaded or not. we only need to do this for sub-components of the component.
+	// path for parent/ancenstor component can be inferred from the path.
+	componentName      string
+	componentPath      componentPath
+	componentInitialVT *persistencespb.VersionedTransition // this identifies a component
+	entityLastUpdateVT *persistencespb.VersionedTransition // this is consistency token
+
+	validationFn func(Context, Component) error
```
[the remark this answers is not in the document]
yycptt: hmm this is a field in the ComponentRef struct, so `Fn` is right?

yycptt (commented):
Will remove anything under chasm/lib before merging the PR.

chasm/ref.go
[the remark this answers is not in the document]
alexshtin: my bad

alexshtin (approved):
I am pretty much sure we will do renames and move things around in future but this is something to start with.

chasm/engine.go
```diff
@@ -0,0 +1,293 @@
...
+// THE SOFTWARE.
+
+package chasm
+
+import (
+	"context"
+)
+
+// now that is not exported, a struct is good enough
+// instead of an interface
```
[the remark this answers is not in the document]
yycptt: oh my bad. Removed the comments. 

Long story... Originally this is an export Engine interface which does provide strongly typed experience to users. Later we moved to exported generic methods for creating & updating components. So the interface becomes private and I was thinking maybe we don't need the interface at all and can use just write an engineImpl. 
But now I think the interface def should still be useful when we write tests and also for the test framework which may use a different implementation.

chasm/engine.go
```diff
@@ -0,0 +1,296 @@
...
+const (
+	BusinessIDReusePolicyAllowDuplicate BusinessIDReusePolicy = iota
+	BusinessIDReusePolicyRejectDuplicate
+)
+
+type BusinessIDConflictPolicy int
+
+const (
+	BusinessIDConflictPolicyFail BusinessIDConflictPolicy = iota
+	BusinessIDConflictPolicyTermiateExisting
```
dandavison, line 79: Typo

PR #6988, opened 2024-12-13
Add attempt count to Nexus invocation and cancelation tasks

pdoerner:
## What changed?

Added attempt count to `nexusoperations.InvocationTask` and `nexusoperations.CancelationTask`
Exported `outbound_queue_factory.StateMachineTask` to make it accessible to executor wrappers


components/nexusoperations/tasks.go
```diff
@@ -187,11 +199,17 @@ func (CancelationTask) Validate(ref *persistencespb.StateMachineRef, node *hsm.N
 type CancelationTaskSerializer struct{}
 
 func (CancelationTaskSerializer) Deserialize(data []byte, attrs hsm.TaskAttributes) (hsm.Task, error) {
-	return CancelationTask{EndpointName: attrs.Destination}, nil
+	attempt, _ := binary.Varint(data)
```
bergundy, line 202: It's fine to not error here since we may have tasks without any data attached but you can either make it more explicit by checking if `len(data) == 0` or at least add a comment explaining why you'd ignore an error here.

bergundy (commented):
I would probably use a proto struct to have the ability to add more fields in the future, this approach is a bit limiting.

components/nexusoperations/tasks.go
pdoerner: `binary.Variant()` doesn't error, it returns the number of bytes read as the second value. So I think it is fine to leave it as the zero value. But irrelevant as I will replace with a struct

bergundy (approved):

PR #6992, opened 2024-12-14
Latest UI, Grafana and Prometheus images

stephanos:
## What changed?

Using the latest `grafana`, `temporalio/ui` and `prometheus` images for our local dev setup.

## Why?

The current versions are 6+ months old.

## How did you test it?

Started locally.

### DIFF
--- a/develop/docker-compose/docker-compose.yml
+++ b/develop/docker-compose/docker-compose.yml
@@ -59,7 +59,7 @@ services:
       - temporal-dev-network
   prometheus:
     # Use http://localhost:9090/ to access Prometheus.
-    image: prom/prometheus:v2.30.0
+    image: prom/prometheus:latest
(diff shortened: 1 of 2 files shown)


develop/docker-compose/docker-compose.yml
```diff
@@ -76,7 +76,7 @@ services:
     volumes:
       - ./grafana/provisioning/:/etc/grafana/provisioning/
   temporal-ui:
-    image: temporalio/ui:2.26.2
+    image: temporalio/ui:latest
```
stephanos, line 79: `latest` is usually not the best idea; but I think here it works great because we won't have to remember updating this - and the risk for breaking changes seems low to me.

alexshtin (approved):
Thanks for doing this!

PR #6993, opened 2024-12-14
Handle missing Update `acceptedRequest`

stephanos:
## What changed?

When applying an UpdateAccepted event to the MS, use the `protocolInstanceId` instead of the Update ID from the `acceptedRequest` attributes.

## Why?

See code comment.

## How did you test it?

Extended/modified existing tests.

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -4298,7 +4298,10 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionUpdateAcceptedEvent(
 	if ms.executionInfo.UpdateInfos == nil {
 		ms.executionInfo.UpdateInfos = make(map[string]*persistencespb.UpdateInfo, 1)
 	}
-	updateID := attrs.GetAcceptedRequest().GetMeta().GetUpdateId()
+	// NOTE: `attrs.GetAcceptedRequest().GetMeta().GetUpdateId()` was used here before, but that is problematic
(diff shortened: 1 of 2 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -4265,7 +4265,10 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionUpdateAcceptedEvent(
 	if ms.executionInfo.UpdateInfos == nil {
 		ms.executionInfo.UpdateInfos = make(map[string]*persistencespb.UpdateInfo, 1)
 	}
-	updateID := attrs.GetAcceptedRequest().GetMeta().GetUpdateId()
+	// NOTE: `attrs.GetAcceptedRequest().GetMeta().GetUpdateId()` was used here before, but that is problematic
+	// in a reset/conflict resolution scenario where there is no `acceptedRequest` since the previously written
+	// UpdateAdmitted event already contains the Update payload.
```
alexshtin, line 4270: I would rephrase it to:
1. not refer "previously" case,
2. support future durable admitted case.

```suggestion
	// NOTE: `attrs.GetAcceptedRequest().GetMeta().GetUpdateId()` can't be used here,
	// because `acceptedRequest` might be nil if UpdateAdmitted event with the Update payload
	// was previously written to the history. 
	// For example, in a reset/conflict resolution scenario.
```

alexshtin (approved):

PR #6994, opened 2024-12-14
Delete Namespace: block for replicated namespace

alexshtin:
## What changed?

Delete Namespace: block for replicated namespace.

## Why?

Disable namespace deletion if namespace is replicate because:
  - If namespace is passive in the current cluster, then WF executions will keep coming from
    the active cluster and namespace will never be deleted (`ReclaimResourcesWorkflow` will fail).
  - If namespace is active in the current cluster, then it technically can be deleted (and
    in this case it will be deleted from this cluster only because delete operation is not replicated),
    but this is confusing for the users, as they might expect that namespace is deleted from all clusters.

## How did you test it?

Added unit and functional tests.

## Potential risks

Replicated namespaces can't be deleted anymore and must be disconnected first.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/service/worker/deletenamespace/workflow.go
+++ b/service/worker/deletenamespace/workflow.go
@@ -26,6 +26,7 @@ package deletenamespace
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	enumspb "go.temporal.io/api/enums/v1"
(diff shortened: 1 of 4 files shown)


service/worker/deletenamespace/workflow.go
```diff
@@ -124,11 +125,28 @@ func DeleteNamespaceWorkflow(ctx workflow.Context, params DeleteNamespaceWorkflo
 	var namespaceInfo getNamespaceInfoResult
 	err := workflow.ExecuteLocalActivity(ctx1, la.GetNamespaceInfoActivity, params.NamespaceID, params.Namespace).Get(ctx, &namespaceInfo)
 	if err != nil {
-		return result, temporal.NewNonRetryableApplicationError(fmt.Sprintf("namespace %s is not found", params.Namespace), "", err)
+		ns := params.Namespace.String()
+		if ns == "" {
+			ns = params.NamespaceID.String()
+		}
+		return result, temporal.NewNonRetryableApplicationError(fmt.Sprintf("namespace %s is not found", ns), "", err)
```
hehaifengcn, line 132: how about printing both Namespace and NamespaceID?

hehaifengcn (approved):

yiminc (commented):
I think we should do that at the api handler as well, it will be better experience to reject the API request.

PR #7000, opened 2024-12-17
Fix a couple of issues in the slog adapter

bergundy:
## What changed?

- Always carry over the embedded `zapLogger`.
- Ensure `handler` is never `nil`.
- Allow extracting an underlying `slog.Logger`.

## Why?

Realized there are some issues using custom loggers, like the one used by the CLI.

### DIFF
--- a/common/log/slog.go
+++ b/common/log/slog.go
@@ -44,8 +44,17 @@ type handler struct {
 
 var _ slog.Handler = (*handler)(nil)
 
+// An interface that allows extracting an underlying slog logger.
+type SLogWrapper interface {
+	SLog() *slog.Logger
+}
+
 // NewSlogLogger creates an slog.Logger from a given logger.
 func NewSlogLogger(logger Logger) *slog.Logger {
+	// Try extracting and underlying slog logger (e.g. for Temporal CLI).
+	if sl, ok := logger.(SLogWrapper); ok {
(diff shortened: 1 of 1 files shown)


stephanos (approved):

pdoerner (approved):

common/log/slog.go
```diff
@@ -44,8 +44,17 @@ type handler struct {
 
 var _ slog.Handler = (*handler)(nil)
 
+// An interface that allows extracting an underlying slog logger.
+type SLogWrapper interface {
```
ychebotarev, line 48: I don't see any usage of this SLogWrapper.

ychebotarev (approved):
overall LGFM

common/log/slog.go
bergundy: Users that provide a custom logger may want to reference this interface.

I made a mistake and didn't use the Go docstring convention. Fixing.

common/log/slog.go
```diff
@@ -44,8 +44,17 @@ type handler struct {
 
 var _ slog.Handler = (*handler)(nil)
 
+// An interface that allows extracting an underlying slog logger.
```
bergundy, line 47:
```suggestion
// SLogWrapper allows extracting an underlying slog logger.
```

PR #7001, opened 2024-12-17
Add api at mutable state to track potential reapply events

xwduan:
## What changed?
Add api at mutable state to track potential reapply events, and reapply these events when needed.

## Why?
For state-based replication, we need extra fields to track events that need to reapply.

## How did you test it?
unit test.


service/history/workflow/mutable_state_impl.go
```diff
@@ -7387,3 +7394,19 @@ func (ms *MutableStateImpl) reschedulePendingActivities() error {
 
 	return nil
 }
+
+func (ms *MutableStateImpl) AddReapplyCandidateEvent(event *historypb.HistoryEvent) {
+	switch event.GetEventType() {
+	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED,
```
yux0, line 7400: I guess nexus handles in a different path?

xwduan: I don't see nexus handling it: https://github.com/temporalio/temporal/blob/b44aae88e0aa255b8dc4fc6ffb1735b8f7b3519b/service/history/ndc/hsm_state_replicator.go#L140 may be looking at wrong place.

yycptt: syncHSM task is only used for state-only changes and will confirm event version history is update to date before syncing so doesn't need to handle the reapply case.

Here we need to handle nexus events as well.

service/history/workflow/mutable_state_impl.go
```diff
@@ -218,6 +218,11 @@ type (
 		activityInfosUserDataUpdated map[int64]struct{}
 		timerInfosUserDataUpdated    map[string]struct{}
 
+		// in memory fields to track potential reapply events that needs to be reapplied during workflow update
+		// should only be used in the state based replication as state based replication does not have
+		// event inside history builder
+		reapplyEventsCandidate []*historypb.HistoryEvent
```
yycptt, line 224: nit: comment that this is only for x-run reapply (from zombie wf to current wf).

service/history/workflow/context.go
yycptt, line 1: we need to do the same for ConflictResolveWorkflowExecution I think.

service/history/ndc/workflow_state_replicator.go
yycptt, line 1: Is notFound case handled? I mean syncing state for a workflow that doesn't in target cluster yet (not the continueAsNew newRun case). I thought we need to update applySnapshotWhenWorkflowNotExist as well?

service/history/workflow/mutable_state_impl.go
hai719: updated according to offline sync.

service/history/workflow/mutable_state_impl.go
hai719: updated.

service/history/workflow/context.go
hai719: updated.

service/history/ndc/workflow_state_replicator.go
hai719: added.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -694,6 +694,9 @@ func (r *WorkflowStateReplicatorImpl) getNewRunMutableState(
 
 	if isStateBased {
 		newRunMutableState.InitTransitionHistory()
+		for _, event := range newRunHistory {
+			newRunMutableState.AddReapplyCandidateEvent(event)
+		}
```
yycptt, line 699: This new run (continue as new) is created by applying events so the history builder in newRunMutableState does contain those events I think.

service/history/workflow/context.go
```diff
@@ -612,15 +649,45 @@ func (c *ContextImpl) UpdateWorkflowExecutionWithNew(
...
-
-	if err := c.updateWorkflowExecutionEventReapply(
-		ctx,
-		shardContext,
-		updateMode,
-		updateWorkflowEventsSeq,
-		newWorkflowEventsSeq,
-	); err != nil {
-		return err
+	if len(updateWorkflowEventsSeq) != 0 || len(newWorkflowEventsSeq) != 0 {
```
yycptt, line 652: As commented above, I think newWorkflowEventsSeq is non-empty here. 
We need to check both xxxWorkflowEventsSeq and xxxEventReapplyCandidate and pick the non-empty one if existing for the reapply.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -976,6 +991,13 @@ func (r *WorkflowStateReplicatorImpl) applySnapshotWhenWorkflowNotExist(
 	if err != nil {
 		return err
 	}
+
+	if isStateBased {
+		for _, event := range events {
+			mutableState.AddReapplyCandidateEvent(event)
```
yycptt, line 997: also need to update the ReapplyEvents call in transaction_manager_new_workflow.go

service/history/workflow/context.go
hai719: updated.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -1150,7 +1170,17 @@ BackfillLoop:
...
+		}
+
+		if isStateBased {
+			// If backfill suceeds but later event reapply fails, during task's next retry,
+			// we still need to reapply events that have been stored in local DB.
+			events, err := r.historySerializer.DeserializeEvents(historyBlob.rawHistory)
+			if err != nil {
+				return common.EmptyEventTaskID, nil, err
+			}
+			eventsToReapply = append(eventsToReapply, events...)
```
yycptt, line 1183: nit: is it possible to add those events directly to ms here instead of buffering entire history in memory, return and then filter?

service/history/workflow/context.go
```diff
@@ -613,11 +647,23 @@ func (c *ContextImpl) UpdateWorkflowExecutionWithNew(
...
+		}
+	}
+
 	if err := c.updateWorkflowExecutionEventReapply(
 		ctx,
 		shardContext,
 		updateMode,
-		updateWorkflowEventsSeq,
+		eventsToReapply,
 		newWorkflowEventsSeq,
```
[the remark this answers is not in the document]
hai719: added

service/history/ndc/workflow_state_replicator.go
hai719: restructure the functions to add events directly to ms

service/history/workflow/context.go
```diff
@@ -426,15 +429,48 @@ func (c *ContextImpl) ConflictResolveWorkflowExecution(
...
-	if err := c.conflictResolveEventReapply(
-		ctx,
-		shardContext,
-		conflictResolveMode,
-		resetWorkflowEventsSeq,
-		newWorkflowEventsSeq,
-		// current workflow events will not participate in the events reapplication
-	); err != nil {
-		return err
+	if len(resetWorkflowEventsSeq) != 0 || len(newWorkflowEventsSeq) != 0 {
```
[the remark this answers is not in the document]
hai719: yes! updated as UpdateWorkflowExecutionWithNew

yycptt (approved):

PR #7002, opened 2024-12-17
fix(schema): check for extension before creating

ysdanielkim:
## What changed?
`CREATE EXTENSION` sql query is replaced with a pgsql script that checks for the existence of the extension programatically.

## Why?
`CREATE EXTENSION` can cause issues in environment where creating extensions are limited to certain users. (e.g. Azure)

## How did you test it?
I've ran the updated schema setup in both environments where btree_gin is installed and and not installed.

## Potential risks
Regression in visibility database setup.

## Documentation
Not needed.

## Is hotfix candidate?
No

### DIFF
--- a/common/persistence/query_util.go
+++ b/common/persistence/query_util.go
@@ -97,7 +97,7 @@ func LoadAndSplitQueryFromReaders(
 								return nil, errors.New("error reading contents: unmatched `END IF` keyword")
 							}
 							st = st[:len(st)-1]
-							j = after + len(sqlIfKeyword)
+							j = after + len(sqlIfKeyword) - 1
 						} else {
 							if len(st) == 0 || st[len(st)-1] != sqlBeginKeyword[0] {
 								return nil, fmt.Errorf("error reading contents: unmatched `END` keyword")
--- a/common/persistence/query_util.go
+++ b/common/persistence/query_util.go
@@ -15,6 +15,8 @@ const (
 
 	sqlLeftParenthesis  = '('
 	sqlRightParenthesis = ')'
+	sqlIfKeyword        = "if"
+	sqlEndIfKeyword     = "end if"
 	sqlBeginKeyword     = "begin"
 	sqlEndKeyword       = "end"
 	sqlLineComment      = "--"
@@ -76,6 +78,12 @@ func LoadAndSplitQueryFromReaders(
 					}
 					st = st[:len(st)-1]
 
+				case sqlIfKeyword[0]:
+					if hasWordAt(contentStr, sqlIfKeyword, j) {
+						st = append(st, sqlIfKeyword[0])
+						j += len(sqlIfKeyword) - 1
+					}
+
 				case sqlBeginKeyword[0]:
 					if hasWordAt(contentStr, sqlBeginKeyword, j) {
 						st = append(st, sqlBeginKeyword[0])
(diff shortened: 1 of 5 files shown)


jmbarzee:
@ysdanielkim Thank you for bringing this up and submitting a PR no less!

I'm reviewing the changes and running them through our internal tests. I'm additionally reviewing how this may have got through our tests which test upgrade scenarios.

Can you tell me what version of Postgress you are running?

rodrigozhou:
@ysdanielkim I don't really understand why the script would make any difference here, you're still calling `CREATE EXTENSION`. Can you provide more details or references for this change? Also, did you verify it works in PostgreSQL 12 (min version supported by Temporal)?

jmbarzee:
To add to @rodrigozhou 's question: Its unclear what situation this addresses. Can you clarify what situation this addresses? 

Does it just ensure that the user can check if the extension already exists even if they don't have permission to install extensions?

ysdanielkim:
Of course. This addresses an issue in Azure PostgreSQL where non-admin users cannot create extensions. The extensions in the database are usually set up by a separate admin user and the application is not given admin rights to the database.

This error is returned by Azure PostgreSQL if `CREATE EXTENSION` is executed by a non-admin user:
```
ERROR: Only members of "azure_pg_admin" are allowed to use CREATE EXTENSION (SQLSTATE 42501)
```
And [here is the related documentation](https://learn.microsoft.com/en-us/azure/postgresql/extensions/how-to-create-extensions) from Azure.

Adding this conditional ensures that `CREATE EXTENSION` command is not executed when the extension is already present in the database.

And yes, I've tested this on non-azure PG12 as well.

rodrigozhou:
@ysdanielkim This doesn't seem that will solve anything. The script is for creating necessary table for running Temporal, and we need the extension for that. So, whoever is setting up the DB needs the required permission for running the entire script. The script is meant to be executed only once with the required permissions.

ysdanielkim:
@rodrigozhou Thank you for rebasing. Looks like there's some issues with the build. Any idea what's going on?

rodrigozhou:
It seems it's failing in all PR's, some auth related issue. I'm looking into it.

rodrigozhou:
@ysdanielkim It seems that all tests with Postgres are failing. Please take a look at those

ysdanielkim:
@rodrigozhou this ready for another build.

common/persistence/query_util.go
```diff
@@ -76,13 +78,26 @@ func LoadAndSplitQueryFromReaders(
...
+					}
+
 				case sqlBeginKeyword[0]:
 					if hasWordAt(contentStr, sqlBeginKeyword, j) {
 						st = append(st, sqlBeginKeyword[0])
 						j += len(sqlBeginKeyword) - 1
 					}
 
 				case sqlEndKeyword[0]:
+					if hasWordAt(contentStr, sqlEndIfKeyword, j) {
```
rodrigozhou: I wonder if this would work if there were more than a single space (eg: `end   if`).

ysdanielkim: I don't think it'll work. Do you find the it important? I'd be happy to accommodate if you think so.

rodrigozhou: Yes, I think it would good to have it work properly if possible.

ysdanielkim: https://github.com/temporalio/temporal/pull/7002/commits/be320fa092c07d55cd481f105d6898379525f817

common/persistence/query_util.go
```diff
@@ -84,11 +92,19 @@ func LoadAndSplitQueryFromReaders(
...
 				case sqlEndKeyword[0]:
 					if hasWordAt(contentStr, sqlEndKeyword, j) {
-						if len(st) == 0 || st[len(st)-1] != sqlBeginKeyword[0] {
-							return nil, fmt.Errorf("error reading contents: unmatched `END` keyword")
+						if ok, after := hasWordAfter(contentStr, sqlIfKeyword, j+len(sqlEndKeyword)); ok {
+							if len(st) == 0 || st[len(st)-1] != sqlIfKeyword[0] {
+								return nil, errors.New("error reading contents: unmatched `END IF` keyword")
+							}
+							st = st[:len(st)-1]
+							j = after + len(sqlIfKeyword)
```
rodrigozhou: Need to subtract 1, right?
```suggestion
							j = after + len(sqlIfKeyword) - 1
```

rodrigozhou (changes requested):

common/persistence/query_util.go
ysdanielkim: Yes. Thanks.

ysdanielkim:
> Please fix the lint errors. Otherwise, LGTM

@rodrigozhou Fixed. Thanks for your help.

PR #7003, opened 2024-12-18
[Deployments] Sync workflow before activity tqs and batch updates

dnr:
## What changed?
- When the deployment workflow updates user data of task queues that are in it when the "current" state changes, update workflow before activity task queues.
- Call the sync user data activity in batches of 100, waiting for propagation between each one.

## Why?
- Updating workflow before activity is required for handling activities on unpinned workflows, to avoid workflows getting transitioned to a deployment by an activity start and then backwards by a workflow task start.
- Batching avoids too-large activity payloads and redoing too much work on retries.

## How did you test it?
existing tests

## Potential risks
the workflow logic change is guarded with GetVersion call


service/worker/deployment/deployment_workflow.go
```diff
@@ -300,34 +302,15 @@ func (d *DeploymentWorkflowRunner) handleSyncState(ctx workflow.Context, args *d
...
-		for tqName, byType := range d.State.TaskQueueFamilies {
-			for tqType, data := range byType.TaskQueues {
-				syncReq.Sync = append(syncReq.Sync, &deploymentspb.SyncUserDataRequest_SyncUserData{
-					Name: tqName,
-					Type: enumspb.TaskQueueType(tqType),
-					Data: d.dataWithTime(data),
-				})
-			}
+		if workflow.GetVersion(ctx, "syncToTaskQueues", workflow.DefaultVersion, 0) == workflow.DefaultVersion {
+			err = d.syncToTaskQueues1(ctx)
```
ShahabT, line 306: Nit: should we call these `syncToTaskQueues0` and `syncToTaskQueues1` to match the version?

service/worker/deployment/deployment_workflow.go
```diff
@@ -346,6 +329,89 @@ func (d *DeploymentWorkflowRunner) handleSyncState(ctx workflow.Context, args *d
...
+	}
+
+	activityCtx := workflow.WithActivityOptions(ctx, defaultActivityOptions)
+	tqNames := workflow.DeterministicKeys(d.State.TaskQueueFamilies)
+
+	for _, tqType := range []enumspb.TaskQueueType{
+		// Sync and wait for workflow task queues, then activity task queues, to ensure that
+		// workflows don't bounce between deployments due to activity/wft starts.
+		enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+		enumspb.TASK_QUEUE_TYPE_ACTIVITY,
```
ShahabT, line 376: We want to support Nexus tasks too. they should be synced with activities I believe.

service/worker/deployment/deployment_workflow.go
dnr: sure, makes sense

service/worker/deployment/deployment_workflow.go
dnr: ok

ShahabT:
It seems this wont land before we clone to entity workflows with new names. Let's close it but once the new wfs are added, add the batching part to the new wfs only. The workflow vs activity separation is not needed anymore.

dnr:
> It seems this wont land before we clone to entity workflows with new names. Let's close it but once the new wfs are added, add the batching part to the new wfs only. The workflow vs activity separation is not needed anymore.

Sounds good to me

PR #7005, opened 2024-12-18
increase the number of returned jobs

ychebotarev:
## What changed?

* Add page_num parameter to GH jobs API url.
* remove some debug lines
* remove pipefail
## Why?

The default number of returned jobs is 30, and we may have more then that.

With pipefail (as I understand it) we can be in a situation when job_id is set to the pipeline errors.
Without it it will be empty. Empty is better then a random error.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -473,11 +473,9 @@ jobs:
       - name: Get job ID from workflow run
         id: get_job_id
         run: |
-          set -eu
+          set -euo pipefail
           test_job_name="Functional test (${{matrix.runs-on}}, ${{matrix.shard_index}}, ${{matrix.name}})"
-          job_url="https://api.github.com/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/jobs"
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -600,11 +598,9 @@ jobs:
       - name: Get job ID from workflow run
         id: get_job_id
         run: |
-          set -euo pipefail
+          set -eu
```
bergundy, line 601: It's the other way around, you want `pipefail` to ensure errors aren't silently ignored.

bergundy: And I would add a `--fail` to the `curl` call unless you prove that these calls are expected to occasionally fail.

bergundy: Decided to not do this and instead explicitly call out that we are okay with this failing.

Also change to `JOB_ID=${job_id:-unknown}` so we at least have a sentinel value.

ychebotarev: done

bergundy (approved):

PR #7014, opened 2024-12-19
Add HSM Deletion Tombstone Tracking

justinp-tt:
## What changed?
- Extended tombstone tracking to capture HSM state machine deletions from Operation Log 
- Added `StateMachinePath` tombstone type for HSM deletions alongside existing activity/timer/etc. tombstones
- Maintained existing tombstone batch management and cleanup mechanisms

## Why?
Without this:
- HSM state machine deletions would not be properly replicated across clusters
- Could lead to state inconsistencies between active/passive clusters when using HSM
- No way to know if an HSM node was intentionally deleted or is missing due to an error

## How did you test it?
Draft PR. Tests not developed yet

## Potential risks
- Potential memory pressure from large HSM deletion batches
- New panic path if operation log access fails

Risk Mitigation:
- Operation log errors surface quickly via panic rather than silent failure

## Documentation
No changes required - this extends an internal implementation detail of how deletions are tracked.

## Is hotfix candidate?
No - this should follow normal deployment process.


service/history/workflow/mutable_state_impl.go
```diff
@@ -5906,7 +5907,70 @@ func (ms *MutableStateImpl) closeTransactionTrackTombstones(
...
+							RequestCancelInitiatedEventId: ms.parseID(key.ID),
+						},
+					}
+				case "Signal":
+					tombstone = &persistencespb.StateMachineTombstone{
+						StateMachineKey: &persistencespb.StateMachineTombstone_SignalExternalInitiatedEventId{
+							SignalExternalInitiatedEventId: ms.parseID(key.ID),
+						},
+					}
+				}
```
bergundy, line 5964: This is not necessary, you only need to track `StateMachinePath` changes here. Activities and other state machines pre-date the HSM framework and won't be reflected in the oplog.

justinp-tt: Changed! Thanks

service/history/workflow/mutable_state_impl.go
```diff
@@ -5907,6 +5907,39 @@ func (ms *MutableStateImpl) closeTransactionTrackTombstones(
 	}
 
 	var tombstones []*persistencespb.StateMachineTombstone
+	if ms.stateMachineNode != nil {
+		opLog, err := ms.stateMachineNode.Outputs()
+		if err != nil {
+			panic(fmt.Sprintf("Failed to get HSM operation log: %v", err))
```
bergundy, line 5913: Don't panic here, return an error instead and handle it in the caller.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5907,6 +5907,39 @@ func (ms *MutableStateImpl) closeTransactionTrackTombstones(
 	}
 
 	var tombstones []*persistencespb.StateMachineTombstone
+	if ms.stateMachineNode != nil {
+		opLog, err := ms.stateMachineNode.Outputs()
```
bergundy, line 5911: We can rename `Outputs` to `OpLog`.

service/history/workflow/mutable_state_impl.go
bergundy: I'd be okay panicking if the only reason this would happen only unexpectedly. Consider making the `Outputs()` function panic instead of returning an error.

bergundy (approved):
LGTM, will defer to @xwduan, @yycptt, or @hai719 who have more context here.

yycptt (approved):
LGTM. 

Applying those tombstones will be in a later PR I assume?

PR #7015, opened 2024-12-19
functional tests testing queries + nexus tasks with versioning-3

Shivs11:
## What changed?

- various functional tests (sticky, no-sticky) testing Query + Nexus tasks routing with versioning-3
- refactored `taskPoller` to now consider query and nexus tasks 

## Why?

- more testing

## How did you test it?

- no feature related line changes so N/A

## Potential risks

- none: shall reduce risks if not for anything

## Is hotfix candidate?

No


service/history/api/get_workflow_util.go
```diff
@@ -361,5 +361,6 @@ func MutableStateToGetResponse(
 		InheritedBuildId:             mutableState.GetInheritedBuildId(),
 		MostRecentWorkerVersionStamp: mostRecentWorkerVersionStamp,
 		TransitionHistory:            mutableState.GetExecutionInfo().TransitionHistory,
+		VersioningInfo:               mutableState.GetExecutionInfo().VersioningInfo,
```
Shivs11, line 364: I think we might have forgotten to add this while making the versioning-3 changes - this is required for query tasks

common/testing/taskpoller/taskpoller.go
```diff
@@ -239,9 +247,6 @@ func (p *workflowTaskPoller) pollTask(
 	}
 
 	events = history.Events
-	if len(events) == 0 {
```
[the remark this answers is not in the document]
ShahabT: cc @stephanos. Instead of removing it, should we skip the check for sticky queues only?

tests/versioning_3_test.go
```diff
@@ -205,13 +210,130 @@ func (s *Versioning3Suite) testWorkflowWithPinnedOverride(sticky bool) {
...
+	if sticky {
+		s.verifyWorkflowStickyQueue(we, tv.StickyTaskQueue())
+	}
+
+	s.pollAndQueryWorkflow(tv, sticky)
+
+	s.verifyWorkflowVersioning(tv, vbUnpinned, tv.Deployment(), nil, nil)
+	if sticky {
+		s.verifyWorkflowStickyQueue(we, tv.StickyTaskQueue())
+	}
```
ShahabT, line 315: Let's add a query redirect step to this test. After this line, we should change the current deployment, and query the wf and make sure the new deployment's poller receives the query.

common/testing/taskpoller/taskpoller.go
```diff
@@ -125,9 +146,123 @@ func (p *TaskPoller) PollAndHandleWorkflowTask(
 // If no task is available, it returns NoTaskAvailable.
 func (p *workflowTaskPoller) HandleTask(
```
ShahabT, line 147: Most of the time we do not work with queries, it seems a little to much to impose the nested TaskCompletedResponse on all usages. Why not keep all these methods untouched and add a `HandleQueryTask` instead:
```
func (p *workflowTaskPoller) HandleQueryTask(tv *testvars.TestVars,
	handler func(task *workflowservice.PollWorkflowTaskQueueResponse) (*workflowservice.RespondQueryTaskCompletedResponse, error)
```

common/testing/taskpoller/taskpoller.go
```diff
@@ -55,17 +58,29 @@ type (
...
-	optionFunc func(*options)
+	optionFunc           func(*options)
+	TaskCompletedRequest struct {
+		WorkflowTaskCompletedRequest *workflowservice.RespondWorkflowTaskCompletedRequest
+		QueryTaskCompletedRequest    *workflowservice.RespondQueryTaskCompletedRequest
+	}
+	TaskCompletedResponse struct {
+		WorkflowTaskCompletedResponse *workflowservice.RespondWorkflowTaskCompletedResponse
+		QueryTaskCompletedResponse    *workflowservice.RespondQueryTaskCompletedResponse
+	}
```
stephanos, line 77: Hm, I'm not sure this is the right direction. I understand that the current API doesn't work for these Queries, but 95% of the time, the task poller will probably only deal with WFTs. Right now, without thinking longer about it, I see three other directions:
1. Create a dedicated, exported function just for queries.
2. Change the return type to any and require type assertions.
3. Hide the fact that there's a "legacy" query (AFAIU) and use the "new" query API; but inside the taskpoller do the right thing auto-magically

common/testing/taskpoller/taskpoller.go
```diff
@@ -125,9 +146,123 @@ func (p *TaskPoller) PollAndHandleWorkflowTask(
...
+			},
+		})
+	return err
+}
+
+func (p *nexusTaskPoller) HandleTask(
+	tv *testvars.TestVars,
+	handler func(task *workflowservice.PollNexusTaskQueueResponse) (*workflowservice.RespondNexusTaskCompletedRequest, error),
+	opts ...optionFunc,
+) (*workflowservice.RespondNexusTaskCompletedResponse, error) {
```
stephanos, line 265: ^ Nexus handlers look good to me 👍 (having said that, I've never used Nexus APIs)

common/testing/taskpoller/taskpoller.go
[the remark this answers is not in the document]
stephanos: Yes, I think that would be better!

common/testing/taskpoller/taskpoller.go
stephanos: (1) would be the easiest and clearest; let me know if that's a feasible option. I haven't worked with this Query API before.

Shivs11: tagging you @ShahabT here as well since you had the exact concern - The main point seems to be to remove the need for `TaskCompletedResponse/Request` structures. 

In my earlier stab at this, I did not have any of these structures and had made one exported function inside `(*workflowTaskPoller).handleTask` and had changed the return type to `any` (all the places that required it). Not, that in this attempt I also intended on using `Versioning3Suite.pollWftAndHandle` since it served as a nice helper function for all my needs. My train of thought was: 

- That "having type assertions and changing the return type to `nil`" or having these two structures were effectively conveying the same purpose: The response can be one of two types - Query or Workflow
- Having understood this, if I were reading the codebase for the first time, I realized I would appreciate knowing that the handler deals with only Queries and Workflow related tasks (which is made clear by the structure in place since it has only those two fields vs `any` would make it a bit confusing)


Hence, I took the more tedious approach and went ahead with this option - I can certainly revert back if you both feel strongly of not keeping this the way it is but would be curious to hear your thoughts

tests/versioning_3_test.go
```diff
@@ -205,13 +210,129 @@ func (s *Versioning3Suite) testWorkflowWithPinnedOverride(sticky bool) {
...
+	override := makePinnedOverride(tv.Deployment())
+	we := s.startWorkflow(tv, override)
+
+	<-wftCompleted
+	s.verifyWorkflowVersioning(tv, vbUnpinned, tv.Deployment(), override, nil)
+	if sticky {
+		s.verifyWorkflowStickyQueue(we, tv.StickyTaskQueue())
+	}
+
+	s.pollAndQueryWorkflow(tv, sticky)
```
Shivs11, line 259: Note: Here (after we poll and process a query), I don't think we should be adding extra lines of code verifying the workflow execution's versioning. I think this is because query tasks don't add an extra event in history and neither does the worker conduct a `RespondWorkflowTaskCompleted`, in which the execution updates the deployment attribute. 

Thus, a point to be noted is that even though a query is always redirected to it's latest deployment, the `workflowexecution.versioning_info.deployment` attribute will not be updated until a workflow task ends up in adding a workflow task completed event.

Shivs11: [ my understanding might be a bit skewed above so feel free to correct me if I was wrong! ]

common/testing/taskpoller/taskpoller.go
stephanos: Your reasoning is totally reasonable; but I strongly believe it's not the right path because it's too verbose for the majority of use cases. That was already a concern before, and is exacerbated here.

> Why not keep all these methods untouched and add a HandleQueryTask instead:

To pick up Shahab's suggestion; is this feasible in your opinion? If so, it would be my preferred approach.

Shivs11: Will be attempting to have type assertions in place along with a separate `HandleQueryTask` method ⏳

tests/versioning_3_test.go
```diff
@@ -804,6 +1051,78 @@ func (s *Versioning3Suite) pollWftAndHandle(
 	return nil, nil
 }
 
+func (s *Versioning3Suite) pollWftAndHandleQueries(
```
Shivs11, line 1054: this was the giant guy I was trying to avoid all along which made me go down the rabbit hole of making those structures which in-turn stirred up the whole thread :( 

cc - @stephanos @ShahabT

tests/versioning_3_test.go
```diff
@@ -838,25 +1157,39 @@ func (s *Versioning3Suite) idlePollWorkflow(
...
 		},
-	).HandleTask(
-		tv,
-		func(task *workflowservice.PollWorkflowTaskQueueResponse) (*workflowservice.RespondWorkflowTaskCompletedRequest, error) {
-			s.Fail(unexpectedTaskMessage)
-			return nil, nil
-		},
-		taskpoller.WithTimeout(timeout),
 	)
+
```
Shivs11, line 1173: did not make a separate function for queries (something like idlePollQueryTasks) since queries are either dispatched via matching directly (so no new history events are being added) or on top of existing workflow tasks - thus, making a separate function for queries could give a reader the assumption that maybe we have a different queue altogether for queries which is not the case

tests/versioning_3_test.go
ShahabT: your understanding is correct.

tests/versioning_3_test.go
```diff
@@ -231,9 +354,10 @@ func (s *Versioning3Suite) TestUnpinnedWorkflow_NoSticky() {
 func (s *Versioning3Suite) testUnpinnedWorkflow(sticky bool) {
 	tv := testvars.New(s)
 	d := tv.Deployment()
+	handlingQueries := false
 
 	if sticky {
-		s.warmUpSticky(tv)
+		s.warmUpSticky(tv, handlingQueries)
```
ShahabT, line 360: just pass false

tests/versioning_3_test.go
```diff
@@ -906,20 +1266,28 @@ func (s *Versioning3Suite) verifyWorkflowStickyQueue(
 // create the sticky queue by polling it.
 func (s *Versioning3Suite) warmUpSticky(
 	tv *testvars.TestVars,
+	handlingQueries bool,
```
ShahabT, line 1269: this arg is not needed. warmup does need to be different for wft vs query. we just want to poll the wft queue to create it.

tests/versioning_3_test.go
ShahabT: because this function needs to fail in both cases, what's the point of passing the bool? Just the old implementation would fit all the needs, no?

common/testing/taskpoller/taskpoller.go
```diff
@@ -262,6 +389,51 @@ func (p *workflowTaskPoller) pollTask(
...
+func (p *workflowTaskPoller) handleQueries(
+	ctx context.Context,
+	opts *options,
+	task *workflowservice.PollWorkflowTaskQueueResponse,
+	handler func(task *workflowservice.PollWorkflowTaskQueueResponse) (*workflowservice.RespondQueryTaskCompletedRequest, error),
+) (*workflowservice.RespondQueryTaskCompletedResponse, error) {
+	p.t.Helper()
+	reply, err := handler(task)
+	if err != nil {
+		return nil, p.respondTaskFailed(ctx, opts, task.TaskToken, err)
```
stephanos, line 426: So from reading the [API docs](https://github.com/temporalio/api/blob/e8fc1a5c490987714081f2aa0d308601cbf1304b/temporal/api/workflowservice/v1/request_response.proto#L914), it seems to me that the correct response here would be to return a `RespondQueryTaskCompletedRequest` with `ErrorMessage` and `Failure` fields set appropriately.

common/testing/taskpoller/taskpoller.go
```diff
@@ -295,6 +467,31 @@ func (p *workflowTaskPoller) handleTask(
...
+	}
+	if reply == nil {
+		return nil, errors.New("missing RespondQueryTaskCompleted return")
+	}
+	if reply.Namespace == "" {
+		reply.Namespace = p.namespace
+	}
+	if reply.TaskToken == nil {
+		reply.TaskToken = task.TaskToken
+	}
```
stephanos, line 487: I think another important validation step would be to ensure `QueryResult` was set by the user. Since there was no error, it must be set, AFAICT.

tests/versioning_3_test.go
```diff
@@ -722,6 +906,52 @@ func respondCompleteWorkflow(
...
+func respondQueryTaskCompleted(
+	task *workflowservice.PollWorkflowTaskQueueResponse,
+	namespace string,
+) *workflowservice.RespondQueryTaskCompletedRequest {
+	request := &workflowservice.RespondQueryTaskCompletedRequest{
+		Namespace: namespace,
+		TaskToken: task.TaskToken,
+	}
+	return request
+}
```
stephanos, line 953: Do you need this helper? Shouldn't you just be able to return `&workflowservice.RespondQueryTaskCompletedRequest{...}` and the taskpoller fills in the two fields for you?

tests/versioning_3_test.go
```diff
@@ -212,6 +218,123 @@ func (s *Versioning3Suite) testWorkflowWithPinnedOverride(sticky bool) {
...
+
+func (s *Versioning3Suite) pollAndQueryWorkflow(
+	tv *testvars.TestVars,
+	sticky bool,
+) {
+	queryResultCh := make(chan interface{})
+	s.pollWftAndHandleQueries(tv, sticky, queryResultCh,
+		func(task *workflowservice.PollWorkflowTaskQueueResponse) (*workflowservice.RespondQueryTaskCompletedRequest, error) {
+			s.NotNil(task)
+			s.NotNil(task.Query)
```
stephanos, line 327: As mentioned elsewhere, I think the taskpoller should ensure that before invoking this handler, ie these lines should not be necessary.

common/testing/taskpoller/taskpoller.go
```diff
@@ -262,6 +389,51 @@ func (p *workflowTaskPoller) pollTask(
...
+) (*workflowservice.RespondQueryTaskCompletedResponse, error) {
+	p.t.Helper()
+	task, err := p.pollTask(ctx, opts)
+	if err != nil {
+		return nil, fmt.Errorf("failed to poll workflow task: %w", err)
+	}
+	return p.handleQueries(ctx, opts, task, handler)
+}
+
+func (p *workflowTaskPoller) handleQueries(
```
stephanos, line 417: Should this be called `handleQuery` since it's for the legacy `Query` field? (as opposed to the newer `Queries` field)

common/testing/taskpoller/taskpoller.go
```diff
@@ -262,6 +389,51 @@ func (p *workflowTaskPoller) pollTask(
 	return resp, err
 }
 
+func (p *workflowTaskPoller) HandleQueries(
```
stephanos, line 392: I think it would be best to signal that this is only for a _legacy_ query. Like `HandleLegacyQuery`. (same for the other methods below)

common/testing/taskpoller/taskpoller.go
```diff
@@ -295,6 +467,31 @@ func (p *workflowTaskPoller) handleTask(
...
+func (p *workflowTaskPoller) respondQueryTaskCompleted(
+	ctx context.Context,
+	task *workflowservice.PollWorkflowTaskQueueResponse,
+	reply *workflowservice.RespondQueryTaskCompletedRequest,
+) (*workflowservice.RespondQueryTaskCompletedResponse, error) {
+	p.t.Helper()
+	if task == nil {
+		return nil, errors.New("missing PollWorkflowTaskQueue")
+	}
+	if reply == nil {
```
stephanos, line 479: I think it would be best to check if there actually is a legacy query present. If there isn't, it seems unexpected for this to issue a `RespondQueryTaskCompleted` request? So I'd add a check for `task.Query != nil` and return an error if it isn't.

tests/versioning_3_test.go
Shivs11: Yeah that's right since inside both the if and the else condition, we poll to create the wft queue - however, on the odd event that we do get an error, (maybe a sticky workflow task or a query gets created), the way we handle each error is different right? That was the reasoning behind this

common/testing/taskpoller/taskpoller.go
Shivs11: missed this while writing this up - thank you Stephan!

common/testing/taskpoller/taskpoller.go
Shivs11: I like it - changed.

common/testing/taskpoller/taskpoller.go
Shivs11: good point - I fixed this by filling in this structure properly which resulted in not expecting an error too in `versioning3.pollAndQueryWorkflow` :)

Shivs11 (commented):
thank you for the feedback!

common/testing/taskpoller/taskpoller.go
```diff
@@ -295,6 +465,47 @@ func (p *workflowTaskPoller) handleTask(
...
+	}
+	if err != nil {
+		reply.ErrorMessage = err.Error()
+		reply.Failure = &failurepb.Failure{
+			Message: err.Error(),
+		}
+		reply.CompletedType = enumspb.QUERY_RESULT_TYPE_FAILED
+	}
+	reply.CompletedType = enumspb.QUERY_RESULT_TYPE_ANSWERED
+	reply.QueryResult = payloads.EncodeString("query-result")
```
stephanos, line 500: I think this should only be set if there is no error, according to the [API docs](https://github.com/temporalio/api/blob/e8fc1a5c490987714081f2aa0d308601cbf1304b/temporal/api/workflowservice/v1/request_response.proto#L926).

ie it should be in the `else` block for the error check.

tests/versioning_3_test.go
stephanos: Seeing the thumbs up, did you intend to inline this now?

common/testing/taskpoller/taskpoller.go
stephanos: Oh, and I think there should be an `if reply.QueryResult != nil` check, too. So the user can actually set their own result if they want to.

Shivs11: ah shoot, I had this but during some version checking with git the else blocked went away - fixed it!

I can add the `QueryResult`  check as well since someone in the future may want to add in a result.

stephanos (approved):
👍 LGTM from taskpoller perspective.

tests/versioning_3_test.go
```diff
@@ -212,6 +216,124 @@ func (s *Versioning3Suite) testWorkflowWithPinnedOverride(sticky bool) {
...
+	s.pollAndQueryWorkflow(tv, sticky)
+	<-pollerDone // wait for the idle poller to complete to not interfere with the next poller
+
+	// redirect query to new deployment
+	s.updateTaskQueueDeploymentData(tvB, 0, tqTypeWf, tqTypeAct)
+
+	go s.idlePollWorkflow(tv, true, ver3MinPollTime, "old deployment should not receive query")
+	// Since the current deployment has changed, task will move to the normal queue (thus, sticky=false)
+	s.pollAndQueryWorkflow(tvB, false)
+
```
Shivs11, line 318: I was able to catch some flakes happening because the previous poller, with deployment set to `tvB`, was still up and running when I tried to poll and process a new query. This is fixed by ensuring the first poller dies before we continue the test.

common/testing/taskpoller/taskpoller.go
ShahabT: Is there anything "legacy" about these query responses? (I see that they are called legacy in this [proto](https://github.com/temporalio/api/blob/33cbaa8657d1dc3d6634888f90e9da5f45642db4/temporal/api/workflowservice/v1/request_response.proto#L308))
But I don't think it's a good choice of word. Maybe something like standalone query fits better because these queries and `RespondQueryTaskCompleted` request/responses are normally used and they are not deprecated in any sense as one would think by seeing them described as "legacy". Unless I'm mistaken, if there is other work to do in the wf, the query will be stuffed into the anyways-needed wft. Otherwise it'll be sent as standalone query (still a wft but with less info), an in that case SDK will send a RespondQueryTaskCompletedRequest instead of a RespondWorkflowTaskCompletedRequest.

ShahabT (approved):
LGTM

common/testing/taskpoller/taskpoller.go
Shivs11: I agree, it felt a bit odd for me too to be calling something `legacy` but actively use it in some of my work here - maybe this is a better question for the SDK folks and I can take this up with them; 

I shall keep the function names with Legacy for now to respect which proto fields we are respecting but if we do see some changes in the proto side, I will alter the tests.

stephanos: I'm okay with a different name; it just seemed like this name stuck now as it's in the docs and used in our internal communication, too. I fear a different name - even if better/more accurate - only creates confusion.

PR #7020, opened 2024-12-19
Delete Namespace: ability to update sleep duration

alexshtin:
## What changed?

Delete Namespace: ability to update sleep duration.

## Why?

SRE experience improvement. Sometimes namespace delete delay is set to the large duration by mistake and there was no way to change it after WF is started.

Use:
```bash
temporal workflow update execute --namespace temporal-system --name update_namespace_delete_delay --workflow-id temporal-sys-reclaim-namespace-resources-workflow/default-deleted-93f5e --input '"10h"'
```
to update delay to the new value (10 hours, in the example above) or:
```bash
temporal workflow update execute --namespace temporal-system --name update_namespace_delete_delay --workflow-id temporal-sys-reclaim-namespace-resources-workflow/default-deleted-93f5e --input '"0"'
```
to remove delay completely.

Please note that new delay is started from the moment it is set, not from the moment when original timer was created (i.e. if WF already slept for 2 hours, and timer is updated to `10h`, it will sleep another 10 hours, not 8).

Access to `temporal-system` namespace is required. `WorkflowId` is built as `temporal-sys-reclaim-namespace-resources-workflow/<deleted_temp_name>`, where `<deleted_temp_name>` is `<original_name>-deleted-<first_5_ns_uuid_chars>`.

## How did you test it?

Added new unit tests + manual run.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/service/worker/deletenamespace/reclaimresources/workflow.go
+++ b/service/worker/deletenamespace/reclaimresources/workflow.go
@@ -135,22 +135,20 @@ func ReclaimResourcesWorkflow(ctx workflow.Context, params ReclaimResourcesParam
 
 	ctx = workflow.WithTaskQueue(ctx, primitives.DeleteNamespaceActivityTQ)
 
-	namespaceDeleteDelay := params.NamespaceDeleteDelay
-	var cancelDeleteDelay workflow.CancelFunc
+	var (
+		namespaceDeleteDelay = params.NamespaceDeleteDelay
+		cancelDeleteDelay    workflow.CancelFunc
+	)
 	err := workflow.SetUpdateHandlerWithOptions(ctx, "update_namespace_delete_delay", func(ctx workflow.Context, newNamespaceDeleteDelayStr string) (string, error) {
-		if newNamespaceDeleteDelayStr == "" {
-			logger.Info("Namespace delete delay is removed. Namespace will be deleted immediately after all workflow executions are deleted.",
-				"oldDuration", namespaceDeleteDelay.String())
-			namespaceDeleteDelay = 0
-		} else {
-			// This must succeed because Update validator already validated the input.
-			namespaceDeleteDelay, _ = time.ParseDuration(newNamespaceDeleteDelayStr)
-		}
+		// This must succeed because Update validator already validated the input.
+		namespaceDeleteDelay, _ = time.ParseDuration(newNamespaceDeleteDelayStr)
 
 		var updateResult string
 		if namespaceDeleteDelay == 0 {
+			logger.Info("Namespace delete delay is removed. Namespace will be deleted immediately after all workflow executions are deleted.")
 			updateResult = "Namespace delete delay is removed."
 		} else {
+			logger.Info("Namespace delete delay is updated.", "new-delete-delay", namespaceDeleteDelay)
 			updateResult = fmt.Sprintf("Namespace delete delay is updated to %s.", namespaceDeleteDelay)
 		}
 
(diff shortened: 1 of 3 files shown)


prathyushpv (approved):

service/worker/deletenamespace/reclaimresources/workflow.go
```diff
@@ -151,9 +198,15 @@ func ReclaimResourcesWorkflow(ctx workflow.Context, params ReclaimResourcesParam
 	}
 
 	// Step 2. Sleep before deleting namespace from a database.
-	err = workflow.Sleep(ctx, params.NamespaceDeleteDelay)
-	if err != nil {
-		return result, fmt.Errorf("%w: %v", errors.ErrUnableToSleep, err)
+	for namespaceDeleteDelay > 0 {
```
ychebotarev, line 201: why `for`, and not `if`?

service/worker/deletenamespace/reclaimresources/workflow.go
```diff
@@ -135,11 +135,58 @@ func ReclaimResourcesWorkflow(ctx workflow.Context, params ReclaimResourcesParam
 
 	ctx = workflow.WithTaskQueue(ctx, primitives.DeleteNamespaceActivityTQ)
 
+	namespaceDeleteDelay := params.NamespaceDeleteDelay
+	var cancelDeleteDelay workflow.CancelFunc
+	err := workflow.SetUpdateHandlerWithOptions(ctx, "update_namespace_delete_delay", func(ctx workflow.Context, newNamespaceDeleteDelayStr string) (string, error) {
```
ychebotarev, line 140: this is making the code very hard to read and understand.
Specifically closures on namespaceDeleteDelay and cancelDeleteDelay

I will recommend to do the same this that is done in for example DeploymentSeriesWorkflowRunner.
Create ReclameResourceRunner struct, make namespaceDeleteDelay and cancelDeleteDelay members of that struct.

Extract UpdateNamespaceDeleteDelay update handle into a separate function
Same for update validator.

then it will look like

err := workflow.SetUpdateHandlerWithOptions(ctx, "update_namespace_delete_delay", r.UpdateNamespaceDeleteDelay, 
workflow.UpdateHandlerOptions{
   Validator: r.UpdateHandlerOptionsValidator
}

service/worker/deletenamespace/reclaimresources/workflow.go
ychebotarev: I would also extract this into separate function.

service/worker/deletenamespace/reclaimresources/workflow.go
```diff
@@ -151,9 +198,15 @@ func ReclaimResourcesWorkflow(ctx workflow.Context, params ReclaimResourcesParam
...
-		return result, fmt.Errorf("%w: %v", errors.ErrUnableToSleep, err)
+	for namespaceDeleteDelay > 0 {
+		var cancelableCtx workflow.Context
+		cancelableCtx, cancelDeleteDelay = workflow.WithCancel(ctx)
+		logger.Info("Delaying namespace delete. Send 'update_namespace_delete_delay' update to change or clear the delay.",
+			"duration", namespaceDeleteDelay.String())
+
+		ndd := namespaceDeleteDelay
+		namespaceDeleteDelay = 0
+		_ = workflow.Sleep(cancelableCtx, ndd)
```
ychebotarev, line 209: it seems that intent is to be able to update namespaceDeleteDelay while sleep?

service/worker/deletenamespace/reclaimresources/workflow.go
alexshtin: We discussed it offline. I personally believe that closures are and inlined functions provide better readability and will leave it as is.

service/worker/deletenamespace/reclaimresources/workflow.go
alexshtin: Because new timer needs to be created if delay is updated.

service/worker/deletenamespace/reclaimresources/workflow.go
alexshtin: Entire PR is about this :-)

PR #7022, opened 2024-12-20
Delete Namespace: move Nexus endpoint validation to workflow

alexshtin:
## What changed?

Delete Namespace: move Nexus endpoint validation to workflow.

## Why?

Consolidate all delete namespace validation logic in one place.

## How did you test it?

Modified existing and new unit tests.

## Potential risks

This is a breaking change for `DeleteNamespace` workflow. But this workflow is very short lived and chances of being upgraded while running are very low.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/service/frontend/operator_handler.go
+++ b/service/frontend/operator_handler.go
@@ -68,6 +68,7 @@ import (
 	"google.golang.org/grpc/health"
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/durationpb"
 )
 
 var _ OperatorHandler = (*OperatorHandlerImpl)(nil)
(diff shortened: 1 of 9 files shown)


service/worker/deletenamespace/workflow.go
```diff
@@ -139,6 +139,16 @@ func validateNamespace(ctx workflow.Context, nsID namespace.ID, nsName namespace
 		return fmt.Errorf("%w: ValidateProtectedNamespacesActivity: %v", errors.ErrUnableToExecuteActivity, err)
 	}
 
+	ctx2 := workflow.WithLocalActivityOptions(ctx, localActivityOptions)
```
bergundy, line 142: `GetVersion()`?

service/worker/deletenamespace/activities.go
```diff
@@ -99,6 +109,35 @@ func (a *localActivities) ValidateProtectedNamespacesActivity(_ context.Context,
...
+				PageSize:              a.nexusEndpointListDefaultPageSize(),
+			})
+			if err != nil {
+				a.logger.Error("Unable to list Nexus endpoints from persistence.", tag.WorkflowNamespace(nsName.String()), tag.Error(err))
+				return fmt.Errorf("unable to list Nexus endpoints for namespace %s: %w", nsName, err)
+			}
+
+			for _, entry := range resp.Entries {
+				if endpointNsID := entry.GetEndpoint().GetSpec().GetTarget().GetWorker().GetNamespaceId(); endpointNsID == nsID.String() {
+					return temporal.NewNonRetryableApplicationError(fmt.Sprintf("cannot delete a namespace that is a target of a Nexus endpoint %s", entry.GetEndpoint().GetSpec().GetName()), errors.ValidationErrorErrType, nil, nil)
```
bergundy, line 129: This would need to be translated to a FailedPrecondition not an invalid argument error, can you change and test the error conversion works as expected?

bergundy (commented):
Overall LGTM, had a couple of small comments.

service/worker/deletenamespace/activities.go
alexshtin: This is a very good point. There are few other errors which are also should be `FailedPrecondition`. I have another PR focused on errors clean up, coming right after this one. Let me address this there.

service/worker/deletenamespace/workflow.go
alexshtin: This is very short lived WF (actual deletion is in separate WF which is started in abandon child mode). I think it is ok not to versioned it (I've already made few changes which breaks compatibility).

service/frontend/operator_handler.go
```diff
@@ -613,43 +614,9 @@ func (h *OperatorHandlerImpl) DeleteNamespace(
...
-				break
-			}
-		}
-	}
-
-	namespaceDeleteDelay := h.config.DeleteNamespaceNamespaceDeleteDelay()
-	if request.NamespaceDeleteDelay != nil {
-		namespaceDeleteDelay = request.NamespaceDeleteDelay.AsDuration()
+	// If NamespaceDeleteDelay is not provided, the default delay configured in the cluster should be used.
+	if request.NamespaceDeleteDelay == nil {
```
ychebotarev, line 618: minor:
in previous version you use new variable, here you are modifying incoming request. 
Which is not recommended usualy.

service/worker/deletenamespace/activities.go
```diff
@@ -99,6 +109,35 @@ func (a *localActivities) ValidateProtectedNamespacesActivity(_ context.Context,
 	return nil
 }
 
+func (a *localActivities) ValidateNexusEndpointsActivity(ctx context.Context, nsID namespace.ID, nsName namespace.Name) error {
+	if !a.allowDeleteNamespaceIfNexusEndpointTarget() {
```
ychebotarev, line 113: readability:

if a.allowDeleteNamespaceIfNexusEndpointTarget() { return nil}
...

less nesting, easier to read.

ychebotarev (approved):
assuming Roey comment will be addressed in the follow up PR

alexshtin:
#7029 has all follow ups.

PR #7023, opened 2024-12-20
populate workflow extended info in DescribeWorkflow

ychebotarev:
## What changed?

Populate WorkflowExtendedInfo as a part of DescribeWorfklow response.

## Why?

Feature request. Users want to see workflow expiration time as a part of DescribeWorfklow response.

## How did you test it?

Update unit tests.

## Is hotfix candidate?

No


service/history/api/describeworkflow/api.go
```diff
@@ -155,6 +155,15 @@ func Invoke(
...
 			VersioningInfo:               executionInfo.VersioningInfo,
 		},
+		WorkflowExtendedInfo: &workflowpb.WorkflowExecutionExtendedInfo{
+			ExecutionExpirationTime: executionInfo.WorkflowExecutionExpirationTime,
+			RunExpirationTime:       executionInfo.WorkflowRunExpirationTime,
+			CancelRequested:         executionInfo.CancelRequested,
+		},
+	}
+
+	if executionInfo.OriginalExecutionRunId != executionState.RunId {
```
yycptt, line 165: OriginalExecutionRunId is a new field and old workflows don't have it, so also need to check if the field is non-empty.

service/history/api/describeworkflow/api.go
```diff
@@ -155,6 +155,15 @@ func Invoke(
...
+		result.WorkflowExtendedInfo.LastResetTime = executionState.StartTime
```
yycptt, line 166: What about the `original_start_time` field?

service/history/api/describeworkflow/api.go
ychebotarev: This is the only way (I was able to find) to find out if workflow was reseted.  There is also `WasReset`, but it is set for "original" workflow, not for the new instance after reset.

Thus, if this is old workflow, this field will be empty. 
In this case this check will not pass (RunId is not empty), and LastResetTime will not be set.
Which is exactly what I would expect, extra check is not needed.

yycptt: for an old workflow OriginalExecutionRunId is empty, RunID is not empty. empty != non-empty, so the check will pass and the LastResetTime will be set right?

service/history/api/describeworkflow/api.go
ychebotarev: We didn't add it to WorkflowExtendedInfo. Which I now think was a mistake - I think StartTime ->is<- `original_start_time`, because it was coming from the event. But apparently it is coming from the new event.
So if we want `original_start_time`, we should add it to API proto. Chad will be happy I guess.

service/history/api/describeworkflow/api.go
yycptt: @gow addedthe IsResetRun() method on mutable state and can be used directly.

service/history/api/describeworkflow/api.go
yycptt: hmm... I can see this field in api main: https://github.com/temporalio/api/blob/master/temporal/api/workflow/v1/message.proto#L131

ychebotarev: oh. lol I forgot that I add it after all.
Ok, added.

service/history/api/describeworkflow/api.go
ychebotarev: will use that funciton

tests/describe_test.go
```diff
@@ -102,6 +102,10 @@ func (s *DescribeTestSuite) TestDescribeWorkflowExecution() {
 	s.Equal(id, wfInfo.RootExecution.GetWorkflowId())
 	s.Equal(we.RunId, wfInfo.RootExecution.GetRunId())
 	s.Equal(we.RunId, wfInfo.GetFirstRunId())
+	s.NotNil(dweResponse.WorkflowExtendedInfo)
+	s.Nil(dweResponse.WorkflowExtendedInfo.LastResetTime) // workflow was not reset
```
yycptt, line 106: nit: also assert OriginalStartTime.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -279,6 +279,11 @@ message WorkflowExecutionInfo {
     // Should be deprecated once the transition history is fully launched
     repeated VersionedTransition previous_transition_history = 100;
     VersionedTransition last_transition_history_break_point = 101;
+
+    // This is the execution time when the WorkflowExecutionStarted event was written.
+    // A workflow reset changes the workflow start time, but preserves this field .
+    google.protobuf.Timestamp original_execution_start_time = 102;
```
yycptt, line 285: I think we don't need this either. The describeWorkflowExecution api already loads the workflowStartedEvent, we can just get the event time as the OriginalStartTime

tests/describe_test.go
ychebotarev: done

proto/internal/temporal/server/api/persistence/v1/executions.proto
ychebotarev: yes, in this can there is no need to add it. I will reuse eventTime

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -279,6 +279,7 @@ message WorkflowExecutionInfo {
     // Should be deprecated once the transition history is fully launched
     repeated VersionedTransition previous_transition_history = 100;
     VersionedTransition last_transition_history_break_point = 101;
+
```
yycptt, line 282: nit: empty line

service/history/api/describeworkflow/api.go
```diff
@@ -155,6 +155,16 @@ func Invoke(
...
+		WorkflowExtendedInfo: &workflowpb.WorkflowExecutionExtendedInfo{
+			ExecutionExpirationTime: executionInfo.WorkflowExecutionExpirationTime,
+			RunExpirationTime:       executionInfo.WorkflowRunExpirationTime,
+			OriginalStartTime:       startEvent.EventTime,
+			CancelRequested:         executionInfo.CancelRequested,
+		},
+	}
+
+	if mutableState.IsResetRun() {
+		result.WorkflowExtendedInfo.LastResetTime = executionState.StartTime
```
yycptt, line 167: I wonder we should expose OriginalRunID instead instead of this. 
Both can be used to tell if the run is a reset run, but LastResetTime will always be the same as result.WorkflowExecutionInfo.StartTime.

yycptt (approved):

service/history/api/describeworkflow/api.go
ychebotarev: Yes, probably you right - by itself LastResetTime doesn't add much value, only the fact that reset happens.
OriginalRunID will add more value - information that reset happen and actual original run id.

good thing - now we have a means to expose whatever we want.

PR #7028, opened 2024-12-20
Wait for handover state update before serving the requests

yux0:
## What changed?
Wait for handover state update before serving the requests.

## Why?
We need to wait on handover state update before the redirect interceptor. It will serve the request once the replication state is updated and routes to the correct endpoint.

## How did you test it?
TODO


common/rpc/interceptor/namespace_validator.go
```diff
@@ -237,10 +242,35 @@ func (ni *NamespaceValidatorInterceptor) StateValidationIntercept(
...
+	handler grpc.UnaryHandler,
+	nsEntry *namespace.Namespace,
+) (interface{}, error) {
+
+	var response interface{}
+	op := func(ctx context.Context) error {
+		err := ni.checkReplicationState(nsEntry, info.FullMethod)
+		if err != nil {
+			var nsErr error
+			nsEntry, nsErr = ni.namespaceRegistry.RefreshNamespaceById(nsEntry.ID())
```
yycptt, line 261: hmmm this will keep refreshing the namespace? I mean all requests will perform a refresh upon retry.

also if the handover error is from history service then the refresh here is unnecessary

common/rpc/interceptor/namespace_validator.go
```diff
@@ -237,10 +242,35 @@ func (ni *NamespaceValidatorInterceptor) StateValidationIntercept(
...
+			}
+			return err
+		}
+
+		response, err = handler(ctx, req)
 		return err
 	}
-	return ni.checkReplicationState(namespaceEntry, fullMethod)
+	// Only retry on ErrNamespaceHandover, other errors will be handled by the retry interceptor
+	err := backoff.ThrottleRetryContext(ctx, op, handoverRetryPolicy, common.IsNamespaceHandoverError)
```
yycptt, line 272: I also worry this will cause server api latency alert to fire during the handover, we should probably exclude this backoff from nouserlatency metric.

common/rpc/interceptor/namespace_validator.go
```diff
@@ -237,10 +242,35 @@ func (ni *NamespaceValidatorInterceptor) StateValidationIntercept(
...
+				return nsErr
+			}
+			return err
+		}
+
+		response, err = handler(ctx, req)
 		return err
 	}
-	return ni.checkReplicationState(namespaceEntry, fullMethod)
+	// Only retry on ErrNamespaceHandover, other errors will be handled by the retry interceptor
```
yycptt, line 271: shall we just let retry interceptor handle the error? and this interceptor just block until ns is no longer in handover state?

yux0: I did not go with this because retry interceptor should be the most inner interceptor. If that is the case, can I add a handover interceptor after the retry interceptor?

common/rpc/interceptor/namespace_validator.go
yux0: Make sense. We can also just wait for ns refresh not do this proactive refresh.

common/rpc/interceptor/namespace_validator.go
yux0: ack

common/util.go
```diff
@@ -357,6 +357,10 @@ func IsServiceHandlerRetryableError(err error) bool {
 	return false
 }
 
+func IsNamespaceHandoverError(err error) bool {
+	return err.Error() == ErrNamespaceHandover.Error()
```
yycptt, line 361: nit: can you replace other places that perform the same change to use this func? e.g. L341 above.

common/rpc/interceptor/telemetry.go
```diff
@@ -215,10 +215,14 @@ func (ti *TelemetryInterceptor) RecordLatencyMetrics(ctx context.Context, startT
 		userLatencyDuration = time.Duration(val)
 		metrics.ServiceLatencyUserLatency.With(metricsHandler).Record(userLatencyDuration)
 	}
+	handoverRetryLatency := time.Duration(0)
+	if val, ok := metrics.ContextCounterGet(ctx, metrics.NamespaceHandoverRetryLatency.Name()); ok {
+		handoverRetryLatency = time.Duration(val)
```
yycptt, line 220: so we don't actually emit metric for this latency?

common/rpc/interceptor/namespace_validator.go
```diff
@@ -237,10 +243,41 @@ func (ni *NamespaceValidatorInterceptor) StateValidationIntercept(
...
+func (ni *NamespaceValidatorInterceptor) handleRequestWithHandoverRetry(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+	nsEntry *namespace.Namespace,
+) (interface{}, error) {
+
+	var response interface{}
+	var retryCount int
```
yycptt, line 258:
```suggestion
	var attempt int
```

common/rpc/interceptor/namespace_validator.go
```diff
@@ -237,10 +243,41 @@ func (ni *NamespaceValidatorInterceptor) StateValidationIntercept(
...
+		err := ni.checkReplicationState(nsEntry, info.FullMethod)
+		if err != nil {
+			return err
+		}
+
+		response, err = handler(ctx, req)
+		if retryCount > 1 {
+			retryLatency = retryLatency + handoverRetryPolicy.ComputeNextDelay(0, retryCount, nil)
+		}
+		retryCount++
```
yycptt, line 270: it's a bit hard to follow as it's calculating the retryLatency before the current attempt, but the code lives after the handler call...

common/rpc/interceptor/namespace_validator.go
yycptt: I couldn't remember why we moved retry interceptor to be the most inner interceptor, but I think there must be a reason. I think we should fully understand that to make sure the retry logic here won't cause any issue.

> can I add a handover interceptor after the retry interceptor?

hmmm then why not just retry handover error in retry interceptor?

I guess I was proposing something different in my comment, i.e. block on ns state change instead of retry (by registering a callback to namespace registry).

common/rpc/interceptor/telemetry.go
yux0: Follow the command to exclude the backoff retry delay from the no user latency.

yux0: We need to chat about this.

common/rpc/interceptor/namespace_validator.go
yux0: I cannot wait for ns callback as the error might return from history nodes and there is no callback from frontend.

common/rpc/interceptor/namespace_handover.go
```diff
@@ -0,0 +1,144 @@
...
+
+	startTime := time.Now()
+	namespaceData, err := i.namespaceRegistry.GetNamespace(namespaceName)
+	if err != nil {
+		return 0, err
+	}
+	if namespaceData.ReplicationState() == enumspb.REPLICATION_STATE_HANDOVER {
+		cbID := uuid.New()
+		waitReplicationStateUpdate := make(chan struct{})
+		i.namespaceRegistry.RegisterStateChangeCallback(cbID, func(ns *namespace.Namespace, deletedFromDb bool) {
```
yycptt, line 104: we probably should be change the impl to listen on changes for a specific namespace...
Can be done in a later PR.

common/rpc/interceptor/namespace_handover.go
```diff
@@ -0,0 +1,144 @@
...
+			if ns.State() != enumspb.NAMESPACE_STATE_REGISTERED ||
+				deletedFromDb ||
+				ns.ReplicationState() != enumspb.REPLICATION_STATE_HANDOVER ||
+				!ns.IsGlobalNamespace() {
+				// Stop wait on state change if:
+				// 1. namespace is deleting/deleted
+				// 2. namespace is not in handover
+				// 3. namespace is not global
+				select {
+				case <-waitReplicationStateUpdate:
```
yycptt, line 117: what's the purpose of this line? I don't see other senders on the channel.

common/rpc/interceptor/namespace_handover.go
```diff
@@ -0,0 +1,144 @@
...
+	methodName := api.MethodName(info.FullMethod)
+	namespaceName, err := GetNamespaceName(i.namespaceRegistry, req)
+	if err != nil {
+		return nil, err
+	}
+
+	if i.enabledForNS(namespaceName.String()) {
+		startTime := i.timeSource.Now()
+		defer func() {
+			metrics.HandoverWaitLatency.With(i.metricsHandler).Record(time.Since(startTime))
```
yycptt, line 76: I guess we only want to emit the metric when namespace actually waited on the handover state.

service/frontend/fx.go
```diff
@@ -267,6 +269,7 @@ func GrpcServerOptionsProvider(
 		namespaceLogInterceptor.Intercept, // TODO: Deprecate this with a outer custom interceptor
 		metrics.NewServerMetricsContextInjectorInterceptor(),
 		authInterceptor.Intercept,
+		namespaceHandoverInterceptor.Intercept,
```
yycptt, line 272: Can you add some comments explaining why it must be in this place?

yycptt: Is the handover state checking logic in StateValidationIntercept removed?

service/frontend/fx.go
```diff
@@ -267,6 +269,7 @@ func GrpcServerOptionsProvider(
...
 		redirectionInterceptor.Intercept,
```
yycptt, line 273: I guess theoretically we can retry in redirection Interceptor? since it has the knowledge of whether redirect will actually happen or not?

common/rpc/interceptor/namespace_handover.go
```diff
@@ -0,0 +1,144 @@
...
+				childCancelFn()
+			}()
+		}
+		select {
+		case <-childCtx.Done():
+			err = common.ErrNamespaceHandover
+		case <-waitReplicationStateUpdate:
+		}
+		i.namespaceRegistry.UnregisterStateChangeCallback(cbID)
+		if err != nil {
```
yycptt, line 138: nit: we don't need this check?

common/rpc/interceptor/namespace_handover.go
yux0: in case the channel is already closed.

service/frontend/fx.go
yux0: I guess so but the logic could be weird. What should we do if the redirect will not happen?

service/frontend/fx.go
yux0: I did not remove in this PR as I added the feature flag. Once we can confident we can remove it.

yycptt: do you want to disabled that the check in StateValidationIntercept if the feature flag is enabled?

common/rpc/interceptor/namespace_handover.go
```diff
@@ -0,0 +1,142 @@
...
+	return handler(ctx, req)
+}
+
+func (i *NamespaceHandoverInterceptor) waitNamespaceHandoverUpdate(
+	ctx context.Context,
+	namespaceName namespace.Name,
+	methodName string,
+) (waitTime *time.Duration, retErr error) {
+	if _, ok := allowedMethodsDuringHandover[methodName]; ok {
+		return
```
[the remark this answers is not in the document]
yux0: done

service/frontend/fx.go
yux0: Yes. Once I validate the function and set default to enable I will remove it

PR #7030, opened 2024-12-21
Handle system time move backward case in timer task processing

yycptt:
## What changed?

- Handle system time move backward case in timer task processing by using max(now(), task visibility timestamp)

## Why?

- Monotonic time is used in time.Time comparison only when both operands have monotonic time value. In our case, the timestamp stored/derived in/from mutable state doesn't have monotonic, thus the comparison logic will use wall clock time to decide if a timer task should be processed. If wall clock move backwards (after we verifies that now() > task visibility timestamp when submitting the task to the task scheduler), the timer task will be dropped and cause workflow to stuck.

## How did you test it?

- Unit tests

## Potential risks

- In worst case, we will execute a timer task logic earlier than expected.

## Is hotfix candidate?

- Could be. But the bug should be very rare as the use millisecond precision when doing time comparison in timer task processing.

### DIFF
--- a/service/history/queues/queue_scheduled.go
+++ b/service/history/queues/queue_scheduled.go
@@ -305,13 +305,17 @@ func (p *scheduledQueue) lookAheadTask() {
 // the reference time. The precision of the comparison is millisecond.
 // This function takes task as input and uses task's fire time (scheduled time)
 // as the minimal reference time to handle clock skew issue.
-// This check is only meaning for tasks with CategoryTypeScheduled as Immediate tasks
-// can be executed at any time.
+// This check is only meaning for tasks with CategoryTypeScheduled and always
+// return false for immediate tasks as they can be executed at any time.
 func IsTimeExpired(
 	task tasks.Task,
 	referenceTime time.Time,
 	testingTime time.Time,
 ) bool {
+	if task.GetCategory().Type() == tasks.CategoryTypeImmediate {
+		return false
+	}
+
 	// NOTE: Persistence layer may lose precision when persisting the task, which essentially moves
 	// task fire time backward. But we are already performing truncation here, so doesn't need to
 	// account for that.
--- a/service/history/queues/queue_scheduled.go
+++ b/service/history/queues/queue_scheduled.go
@@ -303,6 +303,10 @@ func (p *scheduledQueue) lookAheadTask() {
 
 // IsTimeExpired checks if the testing time is equal or before
 // the reference time. The precision of the comparison is millisecond.
+// This function takes task as input and uses task's fire time (scheduled time)
+// as the minimal reference time to handle clock skew issue.
+// This check is only meaning for tasks with CategoryTypeScheduled as Immediate tasks
+// can be executed at any time.
 func IsTimeExpired(
 	task tasks.Task,
 	referenceTime time.Time,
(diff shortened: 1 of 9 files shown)


service/history/queues/queue_scheduled.go
```diff
@@ -303,10 +304,14 @@ func (p *scheduledQueue) lookAheadTask() {
 // IsTimeExpired checks if the testing time is equal or before
 // the reference time. The precision of the comparison is millisecond.
 func IsTimeExpired(
```
bergundy, line 306: I would add a note explaining why we take the task here (to prevent clock skew issues).

service/history/queues/queue_scheduled.go
```diff
@@ -303,10 +304,14 @@ func (p *scheduledQueue) lookAheadTask() {
...
+	task tasks.Task,
 	referenceTime time.Time,
 	testingTime time.Time,
 ) bool {
-	referenceTime = referenceTime.Truncate(persistence.ScheduledTaskMinPrecision)
+	// NOTE: Persistence layer may lose precision when persisting the task, which essentially moves
+	// task fire time backward. But we are already performing truncation here, so doesn't need to
+	// account for that.
+	referenceTime = util.MaxTime(referenceTime, task.GetKey().FireTime).Truncate(persistence.ScheduledTaskMinPrecision)
```
bergundy, line 314: What's the difference between visibility timestamp and fire time and why take that here?

service/history/timer_queue_standby_task_executor.go
```diff
@@ -236,7 +240,7 @@ func (t *timerQueueStandbyTaskExecutor) executeActivityTimeoutTask(
 		// created.
 		isHeartBeatTask := timerTask.TimeoutType == enumspb.TIMEOUT_TYPE_HEARTBEAT
 		ai, heartbeatTimeoutVis, ok := mutableState.GetActivityInfoWithTimerHeartbeat(timerTask.EventID)
-		if isHeartBeatTask && ok && queues.IsTimeExpired(timerTask.GetVisibilityTime(), heartbeatTimeoutVis) {
+		if isHeartBeatTask && ok && queues.IsTimeExpired(timerTask, timerTask.GetVisibilityTime(), heartbeatTimeoutVis) {
```
bergundy, line 243: Shouldn't we use `referenceTime` to allow progressing beyond the visibility timestamp in case task processing is delayed for a while?

service/history/queues/queue_scheduled.go
yycptt: They are the same for Scheduled tasks. For immediate tasks, fire time is always 0. Since this function is only meant to be invoked by scheduled tasks, I think FireTime is better as it won't have any impact on immediate tasks.

service/history/timer_queue_standby_task_executor.go
yycptt: The goal here I believe is to clear the heartbeat timer task status only when we have to. Using referenceTime means we will clear the flag more often and causing more heartbeat timer task being created.

service/history/queues/queue_scheduled.go
bergundy: So for immediate tasks this function is expected to always return false? If so, LGTM.

bergundy (approved):

service/history/queues/queue_scheduled.go
yycptt: It was not, but I think that makes sense and changed it to be the case.

self-transition:
hi @yycptt @bergundy 

If I encounter this issue while using an older version, what is the expected fix? Is there a workaround available?

PR #7032, opened 2024-12-23
Fix ErrStateMachineNotFound handling in HSM state replication

justinp-tt:
## What changed?
- Modified `HSMStateReplicatorImpl.syncHSMNode()` to handle `ErrStateMachineNotFound` gracefully
- Added debug logging with correct field reference to `OriginalExecutionRunId` 
- Added unit test `TestSyncHSM_StateMachineNotFound` to verify behavior

## Why?
After adding support for state deletion in terminal states in Nexus, nightly tests started failing when sync HSM tasks tried to replicate state machines that had been legitimately deleted. Since the deletion is intentional for terminal states, we should gracefully handle these cases by logging and continuing replication of other state machines.

## How did you test it?
- Added unit test verifying graceful handling of `ErrStateMachineNotFound`
- Existing nightly test failures should be resolved by this change

## Potential risks
- If there are cases where a state machine is temporarily unavailable (rather than legitimately deleted), we might incorrectly continue processing
- However, based on the HSM implementation, state machines are either present in persistence or not - there is no transient state
- Suppressing `ErrStateMachineNotFound` could potentially mask other issues if the error occurs for unexpected reasons

## Documentation
No documentation changes required as this is an internal implementation detail handling error cases in the replication path.

## Is hotfix candidate?
No - while this fixes test failures, it's not causing production issues that would warrant a hotfix.


bergundy (commented):
Not super familiar with the replication code path, there may be a better place to put this error handling logic. Were the errors you observed on the source or target cluster?

service/history/ndc/hsm_state_replicator.go
```diff
@@ -183,10 +183,19 @@ func (r *HSMStateReplicatorImpl) syncHSMNode(
 		incomingNodePath := incomingNode.Path()
 		currentNode, err := currentHSM.Child(incomingNodePath)
 		if err != nil {
-			// 1. Already done history resend if needed before,
-			// and node creation today always associated with an event
-			// 2. Node deletion is not supported right now.
-			// Based on 1 and 2, node should always be found here.
+			// The node may not be found if:
+			// 1. The state machine was deleted (e.g. terminal state cleanup)
+			// 2. We're missing events that created this node
```
bergundy, line 188: That's not true based on the comment that you deleted.

> Already done history resend if needed before,
			// and node creation today always associated with an event

I would also clarify that creation and deletion are always associated with an event.

bergundy (approved):
Please clarify the code comments, otherwise LGTM.

service/history/ndc/hsm_state_replicator.go
```diff
@@ -183,10 +183,19 @@ func (r *HSMStateReplicatorImpl) syncHSMNode(
...
+			if errors.Is(err, hsm.ErrStateMachineNotFound) {
+				// In terminal state, nodes can be deleted
+				// Ignore the error and continue processing other nodes
+				r.logger.Debug("State machine not found - likely deleted in terminal state",
+					tag.WorkflowNamespaceID(mutableState.GetExecutionInfo().NamespaceId),
+					tag.WorkflowID(mutableState.GetExecutionInfo().WorkflowId),
+					tag.WorkflowRunID(mutableState.GetExecutionInfo().OriginalExecutionRunId),
+				)
+				return nil
```
yycptt, line 197: nit: I'd add a sanity check that version history in the mutable state is > the one in the request (same as the one on L265. or just return that info from compareVersionHistory), and return an error otherwise.

justinp-tt: Don't we already check this in compareVersionHistory?

yycptt: hmm not sure I follow. Error is returned from compareVersionHistory if last version history item of the (local) mutable state is < that in the request. The check I mentioned is for > (also not the same as the >= checked in compareVersionHistory)

yycptt (approved):

PR #7038, opened 2024-12-27
Add context deadline exceeded error to deep health check aggregator

laniehei:
## Why?
Context deadline exceeded should be treated as unhealthy any time it is emitted.


yux0 (approved):

yux0 (commented):
Can you add a unit test?

PR #7040, opened 2024-12-31
Refactor: remove key parameter from testvar methods

alexshtin:
## What changed?

Remove `key` parameter from `testvar` methods.

## Why?

It turn out to be confusing for people on how to use optional `key` parameter. @bergundy got a great idea to remove this parameter and use different instance of `tv` when more than one entity is needed.

## How did you test it?

Run all tests. Refactoring affects test code only.

## Potential risks

No risks.

## Documentation

Yes, updated testing docs.

## Is hotfix candidate?

No.


common/testing/testvars/any.go
```diff
@@ -80,6 +82,6 @@ func (a Any) ApplicationFailure() *failurepb.Failure {
 	}
 }
 
-func (a Any) RunID() string {
-	return uuid.New()
-}
+func (a Any) InfiniteTimeout() *durationpb.Duration {
```
ychebotarev, line 85: I have the same comment somewhere below.
Why adding this to Any? It is not random, it is very specific value.

common/testing/testvars/test_vars.go
```diff
@@ -104,38 +90,74 @@ func (tv *TestVars) clone() *TestVars {
...
+		vString = vStringer.String()
+	}
+	tv2.kv.Store(key, vString+"_"+suffix)
+
+	return tv2
 }
 
-// ----------- Methods for every "type" ------------
-// TODO: add more as you need them.
+// ----------- Methods for every entity ------------
```
ychebotarev, line 120: I think this all becoming more complicated.

I recommend to restrict this to "With" approach. And remove those "AppendTo" functions.
then users of this code will use just
`tv = tv.WithWorkflowID("specific_id")`
this will create custom "workflow_id" entry with "tv.uniqueString+"workflow-id_"+specific_id.

ychebotarev: I also will recommend to add "At" function:
`func (tv *TestVars) At(key string, initialVal...any)`
which will be basically return getOrCreate(tv.uniqueString, initialVal)

service/history/workflow/update/registry_test.go
```diff
@@ -458,9 +461,9 @@ func TestSendMessages(t *testing.T) {
 
 	t.Run("registry with 2 created updates has no messages to send", func(t *testing.T) {
 		var err error
-		upd1, _, err = reg.FindOrCreate(context.Background(), tv.UpdateID("1"))
+		upd1, _, err = reg.FindOrCreate(context.Background(), tv.UpdateID()+"_1")
```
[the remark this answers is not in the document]
ychebotarev: and same thoughts - tv.At("update_id_1") reads better.

tests/update_workflow_sdk_test.go
```diff
@@ -59,7 +59,10 @@ func TestUpdateWorkflowSdkSuite(t *testing.T) {
 func (s *UpdateWorkflowSdkSuite) TestUpdateWorkflow_TerminateWorkflowAfterUpdateAdmitted() {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
-	tv := testvars.New(s.T()).WithTaskQueue(s.TaskQueue()).WithNamespaceName(namespace.Name(s.Namespace()))
+	tv := testvars.New(s.T()).
+		WithTaskQueue(s.TaskQueue()).
+		WithNamespaceName(namespace.Name(s.Namespace())).
+		WithRunID("")
```
ychebotarev, line 65: this basically "reset run ID".
Why?

tests/update_workflow_sdk_test.go
```diff
@@ -97,7 +100,10 @@ WorkflowExecutionTerminated // This can be EventID=3 if WF is terminated before
 func (s *UpdateWorkflowSdkSuite) TestUpdateWorkflow_TimeoutWorkflowAfterUpdateAccepted() {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
-	tv := testvars.New(s.T()).WithTaskQueue(s.TaskQueue()).WithNamespaceName(namespace.Name(s.Namespace()))
+	tv := testvars.New(s.T()).
```
ychebotarev, line 103: same question

tests/update_workflow_test.go
```diff
@@ -166,7 +166,7 @@ func (s *UpdateWorkflowSuite) TestUpdateWorkflow_EmptySpeculativeWorkflowTask_Ac
   5 WorkflowTaskScheduled // Speculative WT events are not written to the history yet.
   6 WorkflowTaskStarted
 `, task.History)
-					return s.UpdateAcceptCompleteCommands(tv, "1"), nil
+					return s.UpdateAcceptCompleteCommands(tv), nil
```
ychebotarev, line 169: this "1" means that all messages has "1". You removed it, but I don't see any WithUpdateId above. So UpdateID() will be without "1". Is that ok? If so - what was the purpose of "1" suffix before?

tests/update_workflow_test.go
```diff
@@ -1131,7 +1131,7 @@ func (s *UpdateWorkflowSuite) TestUpdateWorkflow_ValidateWorkerMessages() {
 				updRequest := protoutils.UnmarshalAny[*updatepb.Request](s.T(), reqMsg.GetBody())
 				return []*protocolpb.Message{
 					{
-						Id:                 tv.MessageID("update-accepted", "1"),
+						Id:                 tv.MessageID() + "_1",
```
ychebotarev, line 1134: shouldn't it be `tv.MessageID()+"_update-accepted_1"`?

common/testing/testvars/test_vars.go
```diff
@@ -104,38 +90,74 @@ func (tv *TestVars) clone() *TestVars {
...
+// Add more as you need them following the pattern below.
+// Replace "Entity" with the name of the entity, i.e., UpdateID, ActivityType, etc.
+// Add only the necessary methods (in most cases only getter).
+/*
+func (tv *TestVars) Entity() string {
+	return getOrCreate(tv, "entity", tv.uniqueString)
+}
+func (tv *TestVars) WithEntity(entity string) *TestVars {
```
ychebotarev, line 128: usually "With" semantic means you are operating within the same object, rather then creating new object. It is kind of part of the constructor ("chaining" pattern). 
Code should look like
tv:=testvars.New(testvars.WithWorkflowId(), testvars.WithWorkflowId(), ...)

ychebotarev: those are main comments, rest are oservations.

bergundy (commented):
I'll defer to Yuri here, the general direction LGTM.

service/history/workflow/update/registry_test.go
```diff
@@ -262,13 +265,13 @@ func TestFindOrCreate(t *testing.T) {
...
-		upd1, existed, err := reg.FindOrCreate(context.Background(), tv.UpdateID("1"))
-		require.NoError(t, err, "creating update #1 should have beeen allowed")
+		upd1, existed, err := reg.FindOrCreate(context.Background(), tv1.UpdateID())
+		require.NoError(t, err, "creating update #1 should have been allowed")
 		require.False(t, existed)
 		require.Equal(t, 1, reg.Len())
 
 		t.Run("deny new update since it is exceeding the limit", func(t *testing.T) {
-			_, _, err = reg.FindOrCreate(context.Background(), tv.UpdateID("2"))
+			_, _, err = reg.FindOrCreate(context.Background(), tv2.UpdateID())
```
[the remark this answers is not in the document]
ychebotarev: with the new approach, it will be more readable as:
_, _, err = reg.FindOrCreate(context.Background(), tv.WithUpdateID("second_update"))

common/testing/testvars/test_vars.go
```diff
@@ -149,204 +184,170 @@ func (tv *TestVars) Namespace(key ...string) *namespace.Namespace {
...
-func (tv *TestVars) WorkflowID(key ...string) string {
-	return tv.getOrCreate("workflow_id", key).(string)
+func (tv *TestVars) WorkflowID() string {
+	return getOrCreate(tv, "workflow_id", tv.uniqueString, tv.stringNSetter)
 }
 
-func (tv *TestVars) BuildId(key ...string) string {
-	//revive:disable-next-line:unchecked-type-assertion
-	return tv.getOrCreate("build_id", key).(string)
+func (tv *TestVars) WithWorkflowIDN(n int) *TestVars {
```
ychebotarev, line 191: What this N means at the end? 

Why input parameter is int, and WorkflowID() return parameter is string? Why I can't add arbitrary string?

Why behavior of WithRunID is different from WithWorkflowIDN? (in WithRunID you call cloneSetVal, int WithWorkflowIDN you call cloneSetN)?

Why WithRunID accepting string, but WithWorkflowIDN accepting int?

This is very inconsistent and hard to understand.

I would (pretty strongly) suggest to
* remove N
* make all input parameters for WithXXXX strings instead of int if XXXX return string
* use the same approach for every WithXXX functions (either cloneSetN, or cloneSetVal)

common/testing/testvars/test_vars.go
```diff
@@ -149,204 +184,170 @@ func (tv *TestVars) Namespace(key ...string) *namespace.Namespace {
...
 
-func (tv *TestVars) WithWorkflowID(workflowID string, key ...string) *TestVars {
-	return tv.cloneSet("workflow_id", key, workflowID)
+func (tv *TestVars) BuildID() string {
+	return getOrCreate(tv, "build_id", tv.uniqueString, tv.stringNSetter)
 }
 
-func (tv *TestVars) RunID(key ...string) string {
-	return tv.getOrCreate("run_id", key, "").(string)
+func (tv *TestVars) WithBuildID(buildId string) *TestVars {
```
ychebotarev, line 218: code needs to be predictable.
tv.WithBuildID("1").BuildID() != tv.WithBuildIDN(1).BuildID()

this is unpredictable and not intuitive. What is the reason for that? Why we have 2 of those?

service/history/history_engine2_test.go
```diff
@@ -1279,7 +1271,7 @@ func makeCurrentWorkflowConditionFailedError(
 	lastWriteVersion := common.EmptyVersion
 	return &persistence.CurrentWorkflowConditionFailedError{
 		Msg:              "random message",
-		RequestID:        tv.String("PrevRequestID"),
+		RequestID:        tv.RequestID(),
```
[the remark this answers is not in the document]
alexshtin: Well, I think it does. So I removed that `String()` method at all. Let's create methods for every entity type.

service/history/workflow/update/registry_test.go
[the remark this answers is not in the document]
alexshtin: Sorry, I don't understand how new version is harder to understand than previous one.

service/history/workflow/update/registry_test.go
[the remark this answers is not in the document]
alexshtin: I changed all cases like this to use `tv.WithUpdateIDN(1).UpdateID()`. I agree it is a better, uniform way.

tests/namespace_delete_test.go
```diff
@@ -556,7 +556,7 @@ func (s *namespaceTestSuite) Test_NamespaceDelete_Protected() {
 	_, err := s.frontendClient.RegisterNamespace(ctx, &workflowservice.RegisterNamespaceRequest{
 		Namespace:                        tv.NamespaceName().String(),
 		Description:                      tv.Any().String(),
-		WorkflowExecutionRetentionPeriod: tv.InfiniteTimeout(),
+		WorkflowExecutionRetentionPeriod: tv.Any().InfiniteTimeout(),
```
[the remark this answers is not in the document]
alexshtin: `Any()` means, "I don't care, it doesn't matter for this test, I won't assert on it, I wouldn't set it at all, but server requires it".

common/testing/testvars/test_vars.go
alexshtin: I also don't like `AppendTo()`. I changed it to `WithEntityN()` (which means "with entity number N"). I still need `WithEntity()` because sometimes I need to set exact passed value, not generated one with some salt.

As for `At()`, I am against generic purpose methods in `tv`. I want them to be explicit, verbose, and simple.

tests/update_workflow_sdk_test.go
alexshtin: Because default behavior is to generate random one, but in these tests I need `WorkflowExecution()` with empty `RunID()`.

tests/update_workflow_test.go
alexshtin: Well... this is a primary goal for entire PR. "1" is now encapsulated in `tv`.

common/testing/testvars/test_vars.go
alexshtin: Using options pattern seems like an overkill here for me. This is a simple test helper which needs to be simple to use. Micro performance and memory allocations optimization is not important here.
As for naming convention, I believe, it is exactly the opposite. `With` indicates that you are getting new object "with specific property set to value". Otherwise it would be in go setter format: `tv.UpdateID("new_update_id")`.

ychebotarev (approved):
I assume you decide that those (like me) who want "WithWorkflowID(string)" will add that function themself?

alexshtin:
> I assume you decide that those (like me) who want "WithWorkflowID(string)" will add that function themself?

Yes, and I added doc on how to do it to `test_vars.go`. But I really hope that setting explicit value would be necessary in very rare cases.

PR #7043, opened 2025-01-02
Move shard.Engine interface into history/interfaces package

ychebotarev:
## What changed?

"Engine" interface was moved to "history/common" module

## Why?

Part of bigger refactoring efforts.
Goal is to reduce code entanglement, by separating interfaces and implementation, for some major components.
Currently, in lots of cases, interface and implementation are in a single file. This bloats the dependency graph, and introduce bad practices when dealing with "module cyclic dependency" error.

Once moving interfaces is complete:
* add "GetOrCreateContext" to shard context. It will receive cache as a parameter
* change s.workflowCache.GetOrCreateWorkflowExecution(s.hardContext, ...) to s.hardContext.GetOrCreateContext(...)
* add "cache" to shard context ->IMPLEMENTATION<- (this is important - there is (I hope) no need to actually expose workflow cache from shard context)
* remove cache as a parameter from GetOrCreateContext
* remove workflwCacha wcache from all those structures/calles/etc, where it is "bundled" with

This will open the way to other improvements.

## How did you test it?

unit tests


service/history/common/engine.go
```diff
@@ -24,7 +24,7 @@
 
 //go:generate mockgen -copyright_file ../../../LICENSE -package $GOPACKAGE -source $GOFILE -destination engine_mock.go
 
-package shard
+package common
```
ychebotarev, line 27: Note: initially I want to move it to something like "internal/common" or "internal/interfaces". But we can't do this - this is used in functional tests, and we technically put functional tests into separate "package"

bergundy (commented):
I'm still unclear on the problem you're trying to solve with this refactoring. Let's discuss this offline.

yycptt:
> I'm still unclear on the problem you're trying to solve with this refactoring. Let's discuss this offline.

I'd like to be part of the discussion as well. 


> Part of bigger refactoring efforts.

Also want to understand what the structure will eventually look like, plans for other interfaces, MutableState, Context, ShardContext etc. To me, it feels like this is a big refactoring project and need some doc describing the overall plan.

service/history/common/engine.go
bergundy, line 27: You can call this `historyinterfaces` to avoid the need to disambiguate the imports and have a name that better conveys the purpose of the package.

bergundy (commented):
No strong opinion on the new structure, mostly the naming.

Agree with @yycptt that it'd be better to understand the end goal before going to the implementation.

I think our problem is that we're going against Go best practices and defining interfaces on the producer side instead of the consumer.
If we stop using mocks and the pattern of interface + `Impl`, we wouldn't need to change the structure.

https://go.dev/wiki/CodeReviewComments#interfaces

service/history/common/engine.go
ychebotarev: I think about that, but ultimately decide against it.
'interfaces` is better.
* it is/will be consistent. 
* purpose is clear
* no redundant info (`history` is already part of `import ".../history/..."`)
* name is at least shorter

ychebotarev:
>If we stop using mocks and the pattern of interface + `Impl`, we wouldn't need to change the structure.
> 
> https://go.dev/wiki/CodeReviewComments#interfaces

"one step at a time". I don't think we can avoid interfaces altogether (I also think we shoulnd't, but that is another story). 
At least right now we can't avoid them, it will be too big of a change.
Thus separating interface and implementation is a good first step.

ychebotarev:
> Agree with @yycptt that it'd be better to understand the end goal before going to the implementation.
In short.
Short term goal - better code structure, better ownership hierarchy, remove access to persistence layer, remove not needed dependencies.
Long term - better separation of concerts, better testing, make codebase cleaner and more maintainable, etc. 
This sounds a bit abstract.

alexshtin:
I am ready to approve this.

service/history/api/respondworkflowtaskcompleted/api_test.go
```diff
@@ -63,6 +63,7 @@ import (
 	"go.temporal.io/server/service/history/api"
 	"go.temporal.io/server/service/history/events"
 	"go.temporal.io/server/service/history/hsm"
+	historyi "go.temporal.io/server/service/history/interfaces"
```
bergundy, line 66: I would still prefer to call the package `"go.temporal.io/server/service/history/historyinterfaces"` to avoid the need to alias the import everywhere.

bergundy (approved):
Talked offline, nothing blocking on my end.

Still think it'd be better to get another approver on this before merging such a fundamental restructure.

yycptt (approved):

yycptt:
cc @temporalio/cgs Please be aware of this interface change.

service/history/api/respondworkflowtaskcompleted/api_test.go
ychebotarev: We discuss it at length. historyinterfaces is too wordy. 
`historyi` seems like a good comproise.

PR #7049, opened 2025-01-06
Adding a Deployment scavenger 

Shivs11:
## What changed?

- [ WIP ]
- An initial draft of what the deployment scavenger could potentially look like

## Why?

- There is a hard limit on the number of deployments that can be present in a given namespace. Thus, deleting/scavenging old deployments is crucial.

## How did you test it?

- [ WIP ]


proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -56,6 +56,13 @@ message DeploymentLocalState {
         // truth for that is the value at the root of DeploymentLocalState.
         map<int32, TaskQueueData> task_queues = 1;
     }
+
+    StateInfo drained_state_info = 8;
```
Shivs11, line 60: This is placed as a temporary "solution" to deployment states. I only have this here since states and the scavenger are intertwined. Bound to change.

common/dynamicconfig/constants.go
```diff
@@ -1260,6 +1260,12 @@ these log lines can be noisy, we want to be able to turn on and sample selective
 		1000,
 		`MatchingMaxTaskQueuesInDeployment represents the maximum number of task-queues that can be registed in a single deployment`,
 	)
+	DeploymentScavengerDuration = NewNamespaceDurationSetting(
+		"deployment.scavenger.duration",
```
ShahabT, line 1264: something like "matching.DeploymentScavengerDuration" to make it consistent with other configs?

ShahabT: also probably should be called something like DeploymentScavengerRetryInterval to match its purpose.

service/worker/deployment/deployment_activities.go
```diff
@@ -116,3 +117,33 @@ func (a *DeploymentActivities) CheckUserDataPropagation(ctx context.Context, inp
...
+
+	response, err := a.deploymentClient.IsDeploymentDrainedOfAllWorkflows(ctx, a.namespace, input.Deployment)
+	if err != nil {
+		logger.Error("error when checking if deployment is scavenged", "error", err)
+		return nil, err
+	}
+	return &deploymentspb.IsDeploymentDrainedOfAllWorkflowsResponse{IsDrained: response}, nil
+}
+
+func (a *DeploymentActivities) GetNamespaceRetention() (time.Duration, error) {
```
ShahabT, line 143: better to use a [SideEffect](https://docs.temporal.io/develop/java/side-effects#side-effects) for this.

service/worker/deployment/deployment_reachability.go
```diff
@@ -87,29 +87,55 @@ func getDeploymentReachability(
...
 	return manager.CountWorkflowExecutionsRequest{
 		NamespaceID: namespace.ID(namespaceId),
 		Namespace:   namespace.Name(namespaceName),
-		Query:       makeDeploymentQuery(seriesName, buildId, open),
+		Query:       makeDeploymentQuery(seriesName, buildId, open, filterOnStatus),
 	}
 }
 
-func makeDeploymentQuery(seriesName, buildID string, open bool) string {
+func makeDeploymentQuery(seriesName, buildID string, open bool, filterOnStatus bool) string {
```
ShahabT, line 120: `open bool` and `filterOnStatus bool` together is awkward. better to make it an enum such as:
```
ExecutionStatusFilter {
OpenOnly
ClosedOnly
Any
}
```

service/worker/deployment/deployment_workflow.go
```diff
@@ -143,6 +201,14 @@ func (d *DeploymentWorkflowRunner) run(ctx workflow.Context) error {
 	// Listen to signals in a different goroutine to make business logic clearer
 	workflow.Go(ctx, d.listenToSignals)
 
+	// TODO (shivam): - Until deployment states are finalized, this go-routine runs regardless of the current state of the deployment.
+	// After deployment states are implemented, the go-routine will only run when the deployment is in "Draining State".
+	workflow.Go(ctx, func(ctx workflow.Context) {
+		d.retryWithBackoff(ctx, func(ctx workflow.Context) bool {
+			return d.IsDeploymentDrainedOfOpenWorkflows(ctx)
+		})
+	})
```
ShahabT, line 210: We can't do this on all deployments because of Visibility load, should only be done on the deployments that are not current.

ShahabT (commented):
Since any change right now will be hard to revert and we'd need to do patching. I'm thinking to park this and work it on the final design. I'm close to finalize the states design. Some things may change. This also feels overly complex. I want to bring in Delete API as well and rethink about the logic.

service/worker/deployment/deployment_workflow.go
Shivs11: Yep, I agree - have addressed this in the above TODO comment - I placed this here because we don't have the states designed finalized.

Shivs11:
> Since any change right now will be hard to revert and we'd need to do patching. I'm thinking to park this and work it on the final design. I'm close to finalize the states design. Some things may change. This also feels overly complex. I want to bring in Delete API as well and rethink about the logic.

I agree - we should not be shipping something right now if we are not sure if the design is finalized. Would love to hear why you thought this was complex as I can keep that in mind when developing this again after states is finalized.

PR #7050, opened 2025-01-07
OTEL for persistence

stephanos:
## What changed?

Adds (basic) OTEL tracing for all persistence requests:

Screenshot 2025-01-06 at 5 32 26 PM

## Why?

For improved observability during development.

## How did you test it?

By running `make OTEL=true start` and seeing the results in Grafana.

## Potential risks

OTEL is off by default, no behavior change is expected.


Makefile
```diff
@@ -182,6 +182,11 @@ GOTESTSUM := $(LOCALBIN)/gotestsum-$(GOTESTSUM_VER)
 $(GOTESTSUM): | $(LOCALBIN)
 	$(call go-install-tool,$(GOTESTSUM),gotest.tools/gotestsum,$(GOTESTSUM_VER))
 
+GOWRAP_VER := v1.4.1
+GOWRAP := $(LOCALBIN)/gowrap-$(GOWRAP_VER)
+$(GOWRAP): | $(LOCALBIN)
+	$(call go-install-tool,$(GOWRAP),github.com/hexdigest/gowrap/cmd/gowrap,$(GOWRAP_VER))
```
stephanos, line 188: Using `gowrap` removes the need to write yet another wrapper tool ourselves.

common/persistence/persistence-tests/persistence_test_base.go
```diff
@@ -191,6 +194,7 @@ func NewTestBaseForCluster(testCluster PersistenceTestCluster, logger log.Logger
 	return &TestBase{
 		DefaultTestCluster: testCluster,
 		Logger:             logger,
+		TracerProvider:     noop.NewTracerProvider(),
```
stephanos, line 197: By default, don't emit any traces in any test.

common/persistence/client/fx.go
```diff
@@ -201,6 +205,10 @@ func DataStoreFactoryProvider(
 		dataStoreFactory = faultinjection.NewFaultInjectionDatastoreFactory(defaultStoreCfg.FaultInjection, dataStoreFactory)
 	}
 
+	if otel.IsEnabled(tracerProvider) {
+		dataStoreFactory = telemetry.NewTelemetryDataStoreFactory(dataStoreFactory, tracerProvider.Tracer("persistence"))
+	}
```
stephanos, line 210: No OTEL, no interceptor. That's why the generated tracing code doesn't check again if it's recording or not.

common/persistence/telemetry/data_store_factory.go
```diff
@@ -0,0 +1,156 @@
...
+func (d *TelemetryDataStoreFactory) NewNexusEndpointStore() (persistence.NexusEndpointStore, error) {
+	if d.nexusEndpointStore == nil {
+		baseStore, err := d.baseFactory.NewNexusEndpointStore()
+		if err != nil {
+			return nil, err
+		}
+		d.nexusEndpointStore = newTelemetryNexusEndpointStore(baseStore, d.tracer)
+	}
+	return d.nexusEndpointStore, nil
+}
```
stephanos, line 156: ^ All that could also be generated, but it didn't seem worth the effort.

common/persistence/telemetry/gowrap_template
```diff
@@ -0,0 +1,38 @@
...
+            {{$method.ResultsNames}} = _d.{{$.Interface.Name}}.{{$method.Call}}
+            {{- if $method.ReturnsError}}
+            if err != nil {
+              span.RecordError(err)
+            }
+            {{end}}
+            return
+        }
+    {{end}}
+{{end}}
```
stephanos, line 38: The [gowrap README](https://github.com/hexdigest/gowrap?tab=readme-ov-file#hosted-templates) has a lot of examples.

common/persistence/telemetry/gowrap_template
```diff
@@ -0,0 +1,38 @@
...
+            ctx, span := _d.tracer.Start(ctx, "{{ printf "persistence.%s/%s" $.Interface.Name $method.Name }}")
+            defer span.End()
+
+            {{$method.ResultsNames}} = _d.{{$.Interface.Name}}.{{$method.Call}}
+            {{- if $method.ReturnsError}}
+            if err != nil {
+              span.RecordError(err)
+            }
+            {{end}}
+            return
```
stephanos, line 35: I'm keeping it simple for now. Just tracking the span, no extra data yet. That comes next.

This approach will work for many use cases, but not all of them. Specifically, if we got here from a RPC, this span will automatically be linked to that span. Future work will need to ensure everything has the correct Workflow ID tag, if there is no RPC.

Makefile
```diff
@@ -601,9 +606,10 @@ update-dependencies:
 	@go get -u -t $(PINNED_DEPENDENCIES) ./...
 	@go mod tidy
 
-go-generate: $(MOCKGEN) $(GOIMPORTS) $(STRINGER)
+go-generate: $(GOWRAP) $(MOCKGEN) $(GOIMPORTS) $(STRINGER)
 	@printf $(COLOR) "Process go:generate directives..."
 	@go generate ./...
+	$(MAKE) copyright
```
stephanos, line 612: Without this addition, the generated code has no copyright header. Fortunately, this step is idempotent (and fast).

... actually, I think none of the code generators should bother with the copyright header themselves and just rely on this being invoked tbh.

Makefile
```diff
@@ -211,6 +211,13 @@ $(STAMPDIR)/goimports-$(GOIMPORTS_VER): | $(STAMPDIR) $(LOCALBIN)
 	@touch $@
 $(GOIMPORTS): $(STAMPDIR)/goimports-$(GOIMPORTS_VER)
 
+GOWRAP_VER := v1.4.1
+GOWRAP := $(LOCALBIN)/gowrap
+$(STAMPDIR)/gowrap-$(GOWRAP_VER): | $(STAMPDIR) $(LOCALBIN)
+	$(call go-install-tool,$(GOWRAP),github.com/hexdigest/gowrap/cmd/gowrap,$(GOWRAP_VER))
+	@touch $@
+$(GOWRAP): $(STAMPDIR)/gowrap-$(GOWRAP_VER)
```
stephanos, line 219: Using `gowrap` removes the need to write yet another wrapper tool ourselves.

The [gowrap README](https://github.com/hexdigest/gowrap?tab=readme-ov-file#hosted-templates) has a lot of examples, if you're curious.

We might be able to delete a lot of code if we used this for other code generators we have in Go right now ...

common/persistence/telemetry/data_store_factory.go
```diff
@@ -0,0 +1,147 @@
...
+	if d.shardStore == nil {
+		baseStore, err := d.baseFactory.NewShardStore()
+		if err != nil {
+			return nil, err
+		}
+		d.shardStore = newTelemetryShardStore(baseStore, d.tracer)
+	}
+	return d.shardStore, nil
+}
+func (d *TelemetryDataStoreFactory) NewMetadataStore() (persistence.MetadataStore, error) {
```
dandavison, line 83:
```suggestion
}

func (d *TelemetryDataStoreFactory) NewMetadataStore() (persistence.MetadataStore, error) {
```

common/persistence/telemetry/cluster_metadata_store_gen.go
```diff
@@ -0,0 +1,143 @@
...
+// newTelemetryClusterMetadataStore returns telemetryClusterMetadataStore.
+func newTelemetryClusterMetadataStore(base _sourcePersistence.ClusterMetadataStore, tracer trace.Tracer) telemetryClusterMetadataStore {
+	return telemetryClusterMetadataStore{
+		ClusterMetadataStore: base,
+		tracer:               tracer,
+	}
+}
+
+// DeleteClusterMetadata wraps ClusterMetadataStore.DeleteClusterMetadata.
+func (_d telemetryClusterMetadataStore) DeleteClusterMetadata(ctx context.Context, request *_sourcePersistence.InternalDeleteClusterMetadataRequest) (err error) {
```
alexshtin, line 55: `_` is against go convention, but ok, I will leave it for code generators.

service/history/tasks/category.go
```diff
@@ -130,6 +131,10 @@ func (c Category) Type() CategoryType {
 	return c.cType
 }
 
+func (c Category) MarshalText() (text []byte, err error) {
+	return []byte(fmt.Sprintf("%d-%v-%v", c.id, c.cType, c.name)), nil
```
alexshtin, line 135: I think `c.name` would be good enough here.

alexshtin: Why do you need this at all?

alexshtin (approved):
I really want to see this in action!

common/persistence/telemetry/cluster_metadata_store_gen.go
stephanos: Good point. This comes from their templates. I'm happy to change it to `d`.

PR #7052, opened 2025-01-07
Dynamic config conversion improvements

### LINKED ISSUE #6756: Optimize dynamic config value conversion
**Is your feature request related to a problem? Please describe.**

Today dynamic config supports converting config values from primitive types and collections to other Go types.
The conversion code is run every time a value is retrieved and could end up taking up excessive CPU resources.

**Describe the solution you'd like**

Perform conversion only when values change. Based on discussions with @dnr, the best way to achieve this is to optimize for `NotifyingClient` implementations (an optional interface for `Client`).
The concern with this approach is that we would end up with two code paths to maintain and that users that provide an alternative `Client` implementation would not get the perf benefits. We should be careful if we decide to change the `Client` interface since it's public and we try not to break API compatibility.

dnr:
## What changed?
- Split implementation of "constrained default" settings from "plain default" settings. This is more code and the diff looks complex, but the individual paths are both simpler than the mixed version.
- Add conversion cache using a weak map.
- Remove GlobalCachedTypedValue.
- Use "raw" values for subscription dispatch deduping to avoid unnecessary conversions.
- Deep copy default values when using mapstructure, to avoid problems with merging over shared default values.

## Why?
- Fixes #6756
- Performance improvement for "plain default" settings (almost all of them)
- Performance improvement for settings with complex converters
- Remove footgun in defaults that aren't scalar values

## How did you test it?
existing+new unit tests


common/dynamicconfig/deepcopy.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		return nv
+	case reflect.Struct:
+		nv := reflect.New(v.Type()).Elem()
+		for i := range v.Type().NumField() {
+			nv.Field(i).Set(deepCopyValue(v.Field(i)))
+		}
+		return nv
+	default:
+		panic(fmt.Sprintf("Can't deep copy value of type %T: %v", v, v))
```
bergundy, line 80: How likely do you think it is that this will happen in practice and when? It seems like it should only happen during development/CI and adding new configs/defaults.

common/dynamicconfig/deepcopy.go
```diff
@@ -0,0 +1,82 @@
...
+		return nv
+	case reflect.Map:
+		if v.IsNil() {
+			return v
+		}
+		nv := reflect.MakeMapWithSize(v.Type(), v.Len())
+		for i := v.MapRange(); i.Next(); {
+			nv.SetMapIndex(i.Key(), deepCopyValue(i.Value()))
+		}
+		return nv
```
bergundy, line 58: nit: This case isn't tested, not blocking since the code looks good but wouldn't hurt to add a test.

Also, is there ever a need to deepcopy the key?

common/dynamicconfig/util.go
bergundy, line 1: Thanks for adding this.

bergundy (approved):
Thanks for doing this! I didn't have any blocking comments.

common/dynamicconfig/deepcopy.go
dnr: I don't think we need to deepcopy the key, you can't use reference types as keys, and for things like structs or arrays, you can't mutate a field in the key.

I'll add a test.

common/dynamicconfig/deepcopy.go
dnr: Yeah, this would happen on the first use of a new config with a non-copyable default value. It should definitely come up in initial testing/CI. Also this covers the basic types, it would have to be an interface or channel or something like that.

PR #7053, opened 2025-01-07
fix inconsistent label cardinality error

hai719:
## What changed?
fix inconsistent label cardinality error

`inconsistent label cardinality: expected 3 label values but got 4 in prometheus.Labels{"multi_region_namespace":"true", "operation":"SyncActivityTask", "service_name":"history", "temporal_service_type":"history"}`

`inconsistent label cardinality: expected 4 label values but got 5 in prometheus.Labels{"multi_region_namespace":"true", "namespace":"app-1996-frontline-uat.pfayb", "operation":"SyncActivityTask", "service_name":"history", "temporal_service_type":"history"}`

### DIFF
--- a/service/history/replication/task_executor.go
+++ b/service/history/replication/task_executor.go
@@ -141,6 +141,7 @@ func (e *taskExecutorImpl) handleActivityTask(
 		metrics.ServiceLatency.With(e.metricsHandler).Record(
 			time.Since(startTime),
 			metrics.OperationTag(metrics.SyncActivityTaskScope),
+			metrics.NamespaceTag(attr.GetNamespaceId()),
 		)
 	}()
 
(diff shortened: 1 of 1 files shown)


service/history/replication/task_executor.go
```diff
@@ -141,6 +141,7 @@ func (e *taskExecutorImpl) handleActivityTask(
 		metrics.ServiceLatency.With(e.metricsHandler).Record(
 			time.Since(startTime),
 			metrics.OperationTag(metrics.SyncActivityTaskScope),
+			metrics.NamespaceTag(attr.GetNamespaceId()),
```
hehaifengcn, line 144: is `NamespaceId` always available?

hai719: We already added ns here: [link](https://github.com/temporalio/temporal/blob/4c2297244636cc7c85da7969936d1c5e4b3e26de/service/history/replication/executable_task.go#L312)

meiliang86 (approved):

PR #7055, opened 2025-01-07
Add activity retry policy properties to replication

ychebotarev:
## What changed?

Add the following activity retry policy properties to the activity replication logic:
* InitialInterval
* MaximumInterval
* MaximumAttempts
* BackoffCoefficient

## Why?

Before there was assumption that activity retry policy can't change. Because of that it was not replicated.
Now we can change activity retry policy via UpdateActivity API. Those changes should be replicated.

## How did you test it?

Add func tests that check few updated properties.

## Potential risks

Increasing replication size.

## Is hotfix candidate?

No


service/history/replication/executable_activity_state_task.go
```diff
@@ -107,6 +107,10 @@ func NewExecutableActivityStateTask(
 			LastAttemptCompleteTime:    task.LastAttemptCompleteTime,
 			Stamp:                      task.Stamp,
 			Paused:                     task.Paused,
+			RetryInitialInterval:       task.RetryInitialInterval,
+			RetryMaximumInterval:       task.RetryMaximumInterval,
+			RetryMaximumAttempts:       task.RetryMaximumAttempts,
+			RetryBackoffCoefficient:    task.RetryBackoffCoefficient,
```
yux0, line 113: do you also need to update ActivitySyncInfo?

service/history/ndc/activity_state_replicator.go
```diff
@@ -155,6 +155,10 @@ func (r *ActivityStateReplicatorImpl) SyncActivityState(
 			LastAttemptCompleteTime:    request.LastAttemptCompleteTime,
 			Stamp:                      request.Stamp,
 			Paused:                     request.Paused,
+			RetryInitialInterval:       request.RetryInitialInterval,
```
yux0, line 158: also need it for SyncActivitiesState?

service/history/replication/executable_activity_state_task.go
ychebotarev: Updated, but not sure if it really needed - tests are passing without it. I wonder...

service/history/ndc/activity_state_replicator.go
ychebotarev: Not sure I understand this comment. Any action required?

service/history/ndc/activity_state_replicator.go
```diff
@@ -155,6 +155,10 @@ func (r *ActivityStateReplicatorImpl) SyncActivityState(
...
+			RetryMaximumInterval:       request.RetryMaximumInterval,
+			RetryMaximumAttempts:       request.RetryMaximumAttempts,
+			RetryBackoffCoefficient:    request.RetryBackoffCoefficient,
 		},
```
yycptt, line 162: Not related to this PR but related to the previous one.

Why the stamp check in `compareActivity()` method only checks for equality? I think we only want to sync the activity when stamp increases? If the incoming stamp is smaller (e.g. we are merging a dlqed sync activity task), we don't want the sync to happen I think.

ychebotarev: will change to <.

hai719:
Do we need to change the SyncActivityRequest [here](https://github.com/temporalio/temporal/blob/66aa23bd596c8b04aab8139404099ec572d38b88/service/history/replication/task_executor.go#L148)?

yycptt:
> Do we need to change the SyncActivityRequest [here](https://github.com/temporalio/temporal/blob/66aa23bd596c8b04aab8139404099ec572d38b88/service/history/replication/task_executor.go#L148)?

Do we want to support new features in the old rpc based replication stack?

service/history/ndc/activity_state_replicator.go
```diff
@@ -354,11 +358,16 @@ func (r *ActivityStateReplicatorImpl) compareActivity(
...
 	}
 
-	if activityInfo.Stamp != stamp {
+	if activityInfo.Stamp < stamp {
 		// stamp changed, should update activity
 		return true
 	}
 
+	if activityInfo.Stamp > stamp {
+		// stamp changed, should update activity
```
yycptt, line 367: nit: fix the comment.

yycptt (approved):

PR #7056, opened 2025-01-07
gowrap fault injection

stephanos:
## What changed?

Use [gowrap](https://github.com/hexdigest/gowrap) for generating fault injection code. Similar to https://github.com/temporalio/temporal/pull/7050.

## Why?

Less code to maintain.

Before: 238 LOC.

After: 34 line Go Template.

## How did you test it?

Existing tests + eyeballing the output.


Makefile
```diff
@@ -601,9 +608,10 @@ update-dependencies:
 	@go get -u -t $(PINNED_DEPENDENCIES) ./...
 	@go mod tidy
 
-go-generate: $(MOCKGEN) $(GOIMPORTS) $(STRINGER)
+go-generate: $(MOCKGEN) $(GOIMPORTS) $(STRINGER) $(GOWRAP)
 	@printf $(COLOR) "Process go:generate directives..."
 	@go generate ./...
+	$(MAKE) copyright
```
stephanos, line 614: Without this addition, the generated code has no copyright header. Fortunately, this step is idempotent (and fast).

... actually, I think none of the code generators should bother with the copyright header themselves and just rely on this being invoked tbh.

common/persistence/faultinjection/fault.go
```diff
@@ -95,31 +94,3 @@ func inject0(f *fault, op func() error) error {
 	}
 	return f.err
 }
```
stephanos, line 96: Simpler!

Makefile
ychebotarev: thoughts:  
1. should we copyright generated code? 
2. If we do - should we add it to the template? Otherwise it will still be "Uber 2020"
3. I remember we have this discussion about "do we even need copyright at every file"? Where does it go?

ychebotarev (approved):

Makefile
stephanos: All good questions.

(1) The other generated code does, so I followed here.
(2) Same as for (1)
(3) I'm told @alexshtin wants to follow up internally on that question.

alexshtin: I am on it. Let this line be there for now.

alexshtin (approved):
I am sorry to see my baby go but this is a good change!

PR #7058, opened 2025-01-08
Fix `TestDeploymentSuite/TestBatchUpdateWorkflowExecutionOptions_SetPinnedThenUnset` flake

carlydf:
## What changed?
- Increase max concurrent batch jobs from 1 to 3 in DeploymentTestSuite
- Sleep and retry if max concurrent batch operations is hit

## Why?
`TestDeploymentSuite/TestBatchUpdateWorkflowExecutionOptions_SetPinnedThenUnset` was flaking due to "Max concurrent batch operations is reached"

### DIFF
--- a/tests/deployment_test.go
+++ b/tests/deployment_test.go
@@ -28,6 +28,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -65,6 +66,10 @@ tests to write:
 2. Tests to register worker in a deployment and using DescribeDeployment for verification
 */
 
+const (
+	maxConcurrentBatchOps = 3
+)
+
 type (
 	DeploymentSuite struct {
 		testcore.FunctionalTestBase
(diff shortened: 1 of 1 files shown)


tests/deployment_test.go
```diff
@@ -1032,6 +1040,20 @@ func (s *DeploymentSuite) TestBatchUpdateWorkflowExecutionOptions_SetPinnedThenU
...
+func (s *DeploymentSuite) startBatchJobWithinConcurrentJobLimit(ctx context.Context, req *workflowservice.StartBatchOperationRequest) error {
+	var err error
+	s.Eventually(func() bool {
+		_, err = s.FrontendClient().StartBatchOperation(ctx, req)
+		if err == nil {
+			return true
+		} else if strings.Contains(err.Error(), "Max concurrent batch operations is reached") {
+			return false // retry
+		}
+		return true
```
Shivs11, line 1052: what happens if `startBatchOperation` were to return a different error than `Max concurrent batch operations is reached`? Right now, it seems that Eventually would return true but we probably want to retry it in this case as well right?

Shivs11 (approved):
good to go!

PR #7061, opened 2025-01-08
Fix stream sender update ack level on empty task

yux0:
## What changed?
Fix stream sender update ack level on empty task

## Why?
If there is no replication task between two cells, we should still update the ack level.

### DIFF
--- a/service/history/replication/stream_sender.go
+++ b/service/history/replication/stream_sender.go
@@ -489,10 +489,6 @@ Loop:
 			return fmt.Errorf("ReplicationServiceError StreamSender unable to get next replication task: %w", err)
 		}
 
-		if !s.shouldProcessTask(item) {
-			continue
-		}
-
 		skipCount++
 		// To avoid a situation: we are skipping a lot of tasks and never send any task, receiver side will not have updated high watermark,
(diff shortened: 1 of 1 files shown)


service/history/replication/stream_sender.go
```diff
@@ -510,10 +506,15 @@ Loop:
 				return err
 			}
 		}
+
```
hai719, line 509: should we reset skipCount after we send the watermark?

service/history/replication/stream_sender.go
```diff
@@ -489,10 +489,6 @@ Loop:
 			return fmt.Errorf("ReplicationServiceError StreamSender unable to get next replication task: %w", err)
 		}
 
-		if !s.shouldProcessTask(item) {
```
hehaifengcn, line 492: line 562 below https://github.com/temporalio/temporal/pull/7061/files#diff-3a5383e58f4357884dc90cd45cdc6b2650d1bf8e59c6dbb2d78ca561af96da57R562 , the code will send `WorkflowReplicationMessages` with `endExclusiveWatermark`. That is not enough to update Ack level?

PR #7062, opened 2025-01-08
Modify incoming batch query to select only running workflows for certain batch operations

ychebotarev:
## What changed?

Add "Status='Running'" to the query for certain batch operation types.
Note - if query is empty or already contains 'Status' cause - it will not be added.

## Why?

There is no need to run those operations on not-running workflows, but omitting this cause can significantly increase run time.

## How did you test it?

add unit test

## Potential risks
N/A

## Documentation

May be...

## Is hotfix candidate?

No

### DIFF
--- a/service/worker/batcher/activities.go
+++ b/service/worker/batcher/activities.go
@@ -29,6 +29,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"regexp"
 	"time"
 
 	"github.com/pborman/uuid"
@@ -50,7 +51,8 @@ import (
 )
 
 const (
-	pageSize = 1000
+	pageSize                 = 1000
+	statusRunningQueryFilter = "Status='Running'"
 )
 
 var (
@@ -108,11 +110,13 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
 		}
 	}
 
+	adjustedQuery := a.adjustQuery(batchParams)
+
 	if startOver {
 		estimateCount := int64(len(batchParams.Executions))
-		if len(batchParams.Query) > 0 {
+		if len(adjustedQuery) > 0 {
 			resp, err := sdkClient.CountWorkflow(ctx, &workflowservice.CountWorkflowExecutionsRequest{
-				Query: batchParams.Query,
+				Query: adjustedQuery,
 			})
 			if err != nil {
 				metrics.BatcherOperationFailures.With(metricsHandler).Record(1)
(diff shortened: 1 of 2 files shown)


service/worker/batcher/activities.go
```diff
@@ -212,6 +216,35 @@ func (a *activities) getActivityLogger(ctx context.Context) log.Logger {
...
+	}
+
+	// if query is not empty and contains status check - return it as is
+	if containsStatusCondition(batchParams.Query) {
+		return batchParams.Query
+	}
+
+	switch batchParams.BatchType {
+	// anything else?
+	case BatchTypeTerminate, BatchTypeDelete, BatchTypeCancel:
```
yiminc, line 241: No need to add this filter for BatchTypeDelete. 

Need to add BatchTypeSignal.

I'm not sure if BatchTypeUpdateOptions make sense to closed workflow or not.

service/worker/batcher/activities.go
```diff
@@ -212,6 +216,35 @@ func (a *activities) getActivityLogger(ctx context.Context) log.Logger {
...
+		return false
+	}
+	return matched
+}
+
+func (a *activities) adjustQuery(batchParams BatchParams) string {
+	if len(batchParams.Query) == 0 {
+		// don't add anything if query is empty
+		return batchParams.Query
+	}
```
yiminc, line 232: Should not special handle this case.

yiminc (commented):
The purpose of this additional filter is to filter out workflows that does not make sense for the batch operation type. For the action of terminate/cancel/signal, they don't apply to closed workflow, so we need to add this additional filter of "ExecutionStatus=Running" for them.

service/worker/batcher/activities.go
```diff
@@ -212,6 +216,35 @@ func (a *activities) getActivityLogger(ctx context.Context) log.Logger {
 	)
 }
 
+func containsStatusCondition(whereClause string) bool {
```
yiminc, line 219: We shouldn't care about this as well. We don't want user to start batch terminate job with query like "ExecutionStatus = Completed"

service/worker/batcher/activities.go
yiminc: OK, BatchTypeUpdateOptions also does not apply to closed workflow, so we should add it here as well.

ychebotarev: remove BatchTypeUpdateOptions, add BatchTypeSignal and BatchTypeUpdateOptions

service/worker/batcher/activities.go
ychebotarev: We don't process empty query. 
See
`if len(batchParams.Query) > 0 {` and such in BatchActivity function.

If I remove this - adjusted query will not be empty. In this case I will need to check the original query, and it all becomes ugly.

yiminc: make sense, I didn't know we special handled empty query.

service/worker/batcher/activities.go
yiminc: The reason I wanted that filter is to prevent user's mistake causing server to pay unnecessary cost.

yiminc (approved):

PR #7063, opened 2025-01-09
OTEL persistence attributes

stephanos:
## What changed?

Added persistence span attributes.

## Why?

Provide more details for debugging.

## How did you test it?

Ran it locally.

NOTE: Some persistence requests/responses contain encoded protobuf data. That will be addressed in a future PR.

### DIFF
--- a/common/persistence/persistence_interface.go
+++ b/common/persistence/persistence_interface.go
@@ -232,8 +232,8 @@ type (
 	// create the shard with the returned value.
 	InternalGetOrCreateShardRequest struct {
 		ShardID          int32
-		CreateShardInfo  func() (rangeID int64, shardInfo *commonpb.DataBlob, err error)
-		LifecycleContext context.Context // cancelled when shard is unloaded
+		CreateShardInfo  func() (rangeID int64, shardInfo *commonpb.DataBlob, err error) `json:"-"`
+		LifecycleContext context.Context                                                 // cancelled when shard is unloaded
 	}
 
 	// InternalGetOrCreateShardResponse is the response to GetShard
--- a/common/persistence/persistence_interface.go
+++ b/common/persistence/persistence_interface.go
@@ -232,7 +232,7 @@ type (
 	// create the shard with the returned value.
 	InternalGetOrCreateShardRequest struct {
 		ShardID          int32
-		CreateShardInfo  func() (rangeID int64, shardInfo *commonpb.DataBlob, err error) `json:"-"`
+		CreateShardInfo  func() (rangeID int64, shardInfo *commonpb.DataBlob, err error) `json:"-"` // cannot be serialized otherwise
 		LifecycleContext context.Context                                                 // cancelled when shard is unloaded
 	}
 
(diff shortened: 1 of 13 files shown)


common/persistence/persistence_interface.go
```diff
@@ -232,8 +232,8 @@ type (
 	// create the shard with the returned value.
 	InternalGetOrCreateShardRequest struct {
 		ShardID          int32
-		CreateShardInfo  func() (rangeID int64, shardInfo *commonpb.DataBlob, err error)
-		LifecycleContext context.Context // cancelled when shard is unloaded
+		CreateShardInfo  func() (rangeID int64, shardInfo *commonpb.DataBlob, err error) `json:"-"`
```
stephanos, line 235: Cannot be serialized as JSON.

service/history/tasks/category.go
```diff
@@ -130,6 +131,10 @@ func (c Category) Type() CategoryType {
 	return c.cType
 }
 
+func (c Category) MarshalText() (text []byte, err error) {
+	return []byte(fmt.Sprintf("[id:%d][type:%s][name:%s]", c.id, c.cType, c.name)), nil
+}
```
stephanos, line 136: This is needed to be able to serialize `Tasks map[tasks.Category][]tasks.Task`

common/persistence/telemetry/gowrap_template
```diff
@@ -22,12 +24,37 @@ func new{{upFirst $decorator}} (base {{.Interface.Type}}, tracer trace.Tracer) {
...
               span.RecordError(err)
             }
             {{end}}
+
+            if d.debugMode {
+                {{- if (gt (len $method.Params) 1) }}
+                {{ $request := (index $method.Params 1) }}
+                requestPayload, err := json.MarshalIndent({{ $request.Name }}, "", "    ")
+                if err != nil {
+                    fmt.Println("failed to serialize {{$request.Type}} for OTEL span: " + err.Error())
```
stephanos, line 42: This seems good enough as it's only for debugging, ie not for regular use.

common/persistence/persistence_interface.go
alexshtin: I think you should add a comment to this file explaining that this struct needs to be JSON serializable for tracing to work. And that those JSON settings is just for that.

service/history/tasks/category.go
alexshtin: Same comment as I left in previous PR: `c.name` should be good enough because it is in fact task key. Other fields adds more confusion than clarity.

common/persistence/telemetry/gowrap_template
alexshtin: Both Cass and SQL data store factories has logger and it shouldn't be hard to pass it to constructor of data stores themselves.

alexshtin:
```suggestion
                    fmt.Println("failed to serialize {{$request.Type}} for OTEL span:", err.Error())
```
If you decide to leave `fmt`.

alexshtin (approved):

PR #7064, opened 2025-01-09
Do not print error when reading tasks from an empty dlq using tdbg

prathyushpv:
## What changed?
tdbg should not print the error `GetDLQTasks failed. Error: queue not found` when trying to read tasks from an empty DLQ.

## Why?
If DLQ is not found, that means the queue is not yet created and effectively it is empty. HistoryTaskQUeueManager will create the queue when first message is enqueued.
Printing this error message will confuse users.

## How did you test it?
Running tdbg.

### DIFF
--- a/tools/tdbg/dlq_v2_service.go
+++ b/tools/tdbg/dlq_v2_service.go
@@ -193,6 +193,10 @@ func (ac *DLQV2Service) ReadMessages(c *cli.Context) (err error) {
 			}
 			res, err := adminClient.GetDLQTasks(ctx, request)
 			if err != nil {
+				// If the DLQ does not exist yet, it's effectively empty, so we can safely return without an error.
+				if strings.Contains(err.Error(), "queue not found:") {
+					return nil, nil, nil
+				}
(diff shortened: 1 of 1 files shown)


tools/tdbg/dlq_v2_service.go
```diff
@@ -193,6 +193,10 @@ func (ac *DLQV2Service) ReadMessages(c *cli.Context) (err error) {
 			}
 			res, err := adminClient.GetDLQTasks(ctx, request)
 			if err != nil {
+				// If the DLQ does not exist yet, it's effectively empty, so we can safely return without an error.
+				if strings.Contains(err.Error(), "queue not found:") {
```
yycptt, line 197: shall we return a notFound error from that api?

yycptt (approved):

tools/tdbg/dlq_v2_service.go
prathyushpv: It is returning serviceerror.NotFound. But it could be other things(eg: table not found). I couldn't think about a better way to just check "queue not found".

PR #7071, opened 2025-01-13
Refactor: return runID from startWorkflow instead of new testvars instance

alexshtin:
## What changed?

Refactor: return `runID` from `startWorkflow` instead of new `testvars` instance.

## Why?

While reviewing versioning3 functional tests, I found out that pattern when `tv` is getting reassigned from `startWorkflow` helper doesn't work there because of concurrency issues. And although this is not a case for Workflow Update functional tests, I want to follow same pattern in all functional tests.

## How did you test it?

Run them.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/tests/update_workflow_test.go
+++ b/tests/update_workflow_test.go
@@ -122,21 +122,21 @@ func (s *UpdateWorkflowSuite) speculativeWorkflowTaskOutcomes(
 
 func (s *UpdateWorkflowSuite) TestUpdateWorkflow_EmptySpeculativeWorkflowTask_AcceptComplete() {
 	testCases := []struct {
-		Name       string
-		ApplyRunID func(tv *testvars.TestVars, runID string) *testvars.TestVars
+		name     string
+		useRunID bool
 	}{
 		{
-			Name:       "with RunID",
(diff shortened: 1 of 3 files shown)


tests/update_workflow_test.go
```diff
@@ -253,27 +248,24 @@ func (s *UpdateWorkflowSuite) TestUpdateWorkflow_EmptySpeculativeWorkflowTask_Ac
...
-			Name:     "with RunID",
-			UseRunID: true,
+			Name:       "with RunID",
+			ApplyRunID: func(tv *testvars.TestVars, runID string) *testvars.TestVars { return tv.WithRunID(runID) },
 		},
 		{
-			Name:     "without RunID",
-			UseRunID: false,
+			Name:       "without RunID",
+			ApplyRunID: func(tv *testvars.TestVars, _ string) *testvars.TestVars { return tv },
```
stephanos, line 260: I like the `true/false` approach much better. It means I can read the test case inputs very quickly/easily. I don't think the extra `if` is an issue.

alexshtin: ok, let it be `bool` again.

stephanos (approved):

PR #7072, opened 2025-01-13
Initial support for filtering workflows

ychebotarev:
## What changed?

Add a where to parse "SQL_like" queries and evaluate them agains a mutable state

## Why?

There is a need for arbitrary mutable state filtering.
The immidiate need - to support filtering when applyting namespace-level rules (in development)

## How did you test it?

add unit tests

## Potential risks

n/a

## Documentation

not yet

## Is hotfix candidate?

no


service/history/workflow/matcher/evaluators_test.go
```diff
@@ -0,0 +1,204 @@
...
+		},
+		{
+			name:          "absent where clause",
+			query:         "",
+			expectedMatch: false,
+			expectedError: true,
+		},
+		{
+			name:          "workflow id where clause",
+			query:         fmt.Sprintf("where %s = 'workflow_id'", WorkflowID),
```
bergundy, line 63: our query syntax doesn't include `where`, there's no `select` statement.

service/history/workflow/matcher/evaluators_test.go
```diff
@@ -0,0 +1,204 @@
...
+			query:         fmt.Sprintf("where %s < '%s'", WorkflowStartTime, afterTimeStr),
+			expectedMatch: true,
+			expectedError: false,
+		},
+		{
+			name:          "workflow start time where < clause, before",
+			query:         fmt.Sprintf("where %s < '%s'", WorkflowStartTime, beforeTimeStr),
+			expectedMatch: false,
+			expectedError: false,
+		},
```
bergundy, line 170: Wanna do BETWEEN too?

service/history/workflow/matcher/evaluators_test.go
```diff
@@ -0,0 +1,204 @@
...
+			query:         "",
+			expectedMatch: false,
+			expectedError: true,
+		},
+		{
+			name:          "absent where clause",
+			query:         "",
+			expectedMatch: false,
+			expectedError: true,
+		},
```
bergundy, line 60: Duplicate test.

service/history/workflow/matcher/errors.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+
+	"go.temporal.io/api/serviceerror"
+)
+
+type (
+	Error struct {
+		message string
+	}
+)
```
bergundy, line 36: nit: We should really stop using this grouping for types. It's very non-standard and messes with generated documentation.

service/history/workflow/matcher/errors.go
```diff
@@ -0,0 +1,63 @@
...
+
+var (
+	malformedSqlQueryErrMessage = "malformed SQL query"
+	notSupportedErrMessage      = "operation is not supported"
+	invalidExpressionErrMessage = "invalid expression"
+)
+
+func NewMatcherError(format string, a ...interface{}) error {
```
bergundy, line 44:
```suggestion
func NewMatcherError(format string, a ...any) error {
```

service/history/workflow/matcher/errors.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"go.temporal.io/api/serviceerror"
+)
+
+type (
+	Error struct {
```
bergundy, line 33: Why not just use invalid argument instead of creating a new error type?

service/history/workflow/matcher/errors.go
```diff
@@ -0,0 +1,63 @@
...
+func NewMatcherError(format string, a ...interface{}) error {
+	message := fmt.Sprintf(format, a...)
+	return &Error{message: message}
+}
+
+func (c *Error) Error() string {
+	return c.message
+}
+
+func (c *Error) ToInvalidArgument() error {
```
bergundy, line 53: Return structs accept interfaces as a rule of thumb.

service/history/workflow/matcher/errors_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+// TestNewMatcherError ensures that NewMatcherError formats its output correctly
+// and returns an object of type *Error.
+func TestNewMatcherError(t *testing.T) {
+	tests := []struct {
+		name        string
+		format      string
+		args        []interface{}
```
bergundy, line 38:
```suggestion
		args        []any
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,281 @@
...
+		return false, NewMatcherError("%v is not an 'and' expression", sqlparser.String(expr))
+	}
+
+	leftResult, err := m.evaluate(andExpr.Left)
+	if err != nil {
+		return false, err
+	}
+	if leftResult == false {
+		return leftResult, nil
+	}
```
bergundy, line 101:
```suggestion
	if leftResult, err := m.evaluate(andExpr.Left); err != nil || !leftResult {
		return leftResult, err
	}
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,281 @@
...
+			return m.compareStartTimeBetween(fromValue, toValue)
+		case "not between":
+			result, err := m.compareStartTimeBetween(fromValue, toValue)
+			return !result, err
+		default:
+			return false, NewMatcherError("%s: range condition operator must be 'between' or 'not between'", invalidExpressionErrMessage)
+		}
+
+	default:
+		return false, fmt.Errorf("unknown filter name: %s", colNameStr)
```
bergundy, line 201: Shouldn't this also be a matcher error or invalid argument?

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil {
+			return false, err
+		}
+		toValue, err := convertToTime(sqlparser.String(rangeCond.To))
+		if err != nil {
+			return false, err
+		}
+
+		switch rangeCond.Operator {
+		case "between":
```
bergundy, line 191: Didn't see a test case for this.

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (m *MutableStateMatchEvaluator) compareWorkflowStatus(status string, operation string) (bool, error) {
+	if len(status) == 0 {
+		return false, NewMatcherError("workflow status cannot be empty")
+	}
+	msStatus := m.ms.GetExecutionState().Status.String()
+	switch operation {
+	case sqlparser.EqualStr:
+		return strings.EqualFold(msStatus, status), nil
```
bergundy, line 240: Would be nice to get test coverage here.

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,281 @@
...
+	case sqlparser.NotEqualStr:
+		return startTime != expectedTime, nil
+	default:
+		return false, NewMatcherError("%s: operation %s is not supported for StartTime", invalidExpressionErrMessage, operation)
+	}
+}
+
+func (m *MutableStateMatchEvaluator) compareStartTimeBetween(fromTime time.Time, toTime time.Time) (bool, error) {
+	startTime := m.ms.GetExecutionState().StartTime.AsTime()
+	return startTime.After(fromTime) && startTime.Before(toTime), nil
```
bergundy, line 280: Hmm.. is the range supposed to be inclusive here?

service/history/workflow/matcher/util.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	parsedTime, err := time.Parse(defaultDateTimeFormat, timestampStr)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return parsedTime, nil
+}
+
+// convertSqlValue returns a string, int64 or float64 if the parsing succeeds.
+func convertSqlValue(sqlValue string) (interface{}, error) {
```
bergundy, line 56: Not sure I'd use this approach, you know what you expect a value to be, don't try parsing without a knowing that the output type is supposed to be.

service/history/workflow/matcher/util.go
```diff
@@ -0,0 +1,77 @@
...
+	"strings"
+	"time"
+)
+
+const (
+	defaultDateTimeFormat = time.RFC3339
+)
+
+func extractStringValue(s string) (string, error) {
+	if len(s) >= 2 && s[0] == '\'' && s[len(s)-1] == '\'' {
```
bergundy, line 37: I'd want to dive into this more, want to see that we're accepting any form of quoting and allowing proper escaping.

bergundy (commented):
Left some comments. Didn't do a full review but this should get you started for now.

service/history/workflow/matcher/evaluators_test.go
ychebotarev: I add this because I found where we add "select * ..." to the query, but without "where" cause.

ychebotarev: found it...

```go
where = "where " + where
```

service/history/workflow/matcher/evaluators_test.go
ychebotarev: removed

service/history/workflow/matcher/errors.go
ychebotarev: I add this because I start from the existing way we work with queries.
Yeah, I don't see much value in this wrapper. All errors falls into "invalid argument" category, so I will remove it.

service/history/workflow/matcher/errors.go
ychebotarev: I add this because I start from the existing way we work with queries.
Yeah, I don't see much value in this wrapper. All errors falls into "invalid argument" category, so I will remove it.

service/history/workflow/matcher/errors.go
ychebotarev: for one single struct - yes. But in general I like this - it clearly separates declaration from the rest.

ychebotarev: anyway absolute since I will remove this declaration

service/history/workflow/matcher/evaluators.go
ychebotarev: removed

service/history/workflow/matcher/evaluators.go
ychebotarev: added, and for "not between" as well

service/history/workflow/matcher/evaluators.go
ychebotarev: there is "workflow status" test for that. Any specific test you have in mind?

service/history/workflow/matcher/evaluators.go
ychebotarev: good catch - The BETWEEN operator MUST be inclusive

ychebotarev: I will also add tests

service/history/workflow/matcher/util.go
ychebotarev: this assumes that "left" and "right" parts of the expression are parsed one after another.
another approach - parse "left" and "right", check if they match.
I'm not sure there is a strong benefit of the first approach over second. Let me know if you think there is any.

service/history/workflow/matcher/evaluators_test.go
ychebotarev: added between tests

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/temporalio/sqlparser"
+	"go.temporal.io/server/service/history/workflow"
+)
+
+// Supported Fields
+const (
+	WorkflowID              = "WorkflowId"
+	WorkflowTypeName        = "WorkflowType"
+	WorkflowStartTime       = "StartTime"
+	WorkflowExecutionStatus = "ExecutionStatus"
```
rodrigozhou, line 39:
```suggestion
	WorkflowID              = searchattribute.WorkflowId
	WorkflowTypeName        = searchattribute.WorkflowType
	WorkflowStartTime       = searchattribute.StartTime
	WorkflowExecutionStatus = searchattribute.ExecutionStatus
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+
+	leftResult, err := m.evaluate(andExpr.Left)
+	if err != nil {
+		return false, err
+	}
+	if leftResult == false {
+		return leftResult, nil
+	}
+	// if left is true, then right must be evaluated
+	return m.evaluate(andExpr.Right)
```
rodrigozhou, line 103:
```suggestion
	if leftResult {
		// if left is true, then right must be evaluated
		return m.evaluate(andExpr.Right)
	}
	return false, nil
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+}
+func (m *MutableStateMatchEvaluator) evaluateOr(expr sqlparser.Expr) (bool, error) {
+	andExpr, ok := expr.(*sqlparser.AndExpr)
```
rodrigozhou, line 106:
```suggestion
	orExpr, ok := expr.(*sqlparser.OrExpr)
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+	if !ok {
+		return false, NewMatcherError("%v is not an 'and' expression", sqlparser.String(expr))
```
rodrigozhou, line 108:
```suggestion
		return false, NewMatcherError("%v is not an 'or' expression", sqlparser.String(expr))
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	if leftResult, err := m.evaluate(andExpr.Left); err != nil || !leftResult {
+		return leftResult, err
+	}
```
rodrigozhou, line 113:
```suggestion
	if leftResult, err := m.evaluate(andExpr.Left); err != nil || leftResult {
		return leftResult, err
	}
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+		val, err := extractStringValue(valStr)
+		if err != nil {
+			return false, err
+		}
+		return m.compareWorkflowStatus(val, op)
+	case WorkflowStartTime:
+		return m.compareStartTime(valStr, op)
+
+	default:
+		return false, fmt.Errorf("unknown filter name: %s", colNameStr)
```
rodrigozhou, line 159:
```suggestion
		return false, fmt.Errorf("unknown or unsupported search attribute name: %s", colNameStr)
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+
+func (m *MutableStateMatchEvaluator) evaluateRange(expr sqlparser.Expr) (bool, error) {
+	rangeCond, ok := expr.(*sqlparser.RangeCond)
+	if !ok {
+		return false, NewMatcherError("%v is not an 'comparison' expression", sqlparser.String(expr))
+	}
+
+	colName, ok := rangeCond.Left.(*sqlparser.ColName)
+	if !ok {
+		return false, fmt.Errorf("invalid filter name: %s", sqlparser.String(rangeCond.Left))
```
rodrigozhou, line 171: ditto

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+	existingWorkflowType := m.ms.GetExecutionInfo().WorkflowTypeName
+	return m.compareWorkflowString(existingWorkflowType, workflowType, operation, WorkflowTypeName)
+}
+
+func (m *MutableStateMatchEvaluator) compareWorkflowID(workflowID string, operation string) (bool, error) {
+	existingWorkflowId := m.ms.GetExecutionInfo().WorkflowId
+	return m.compareWorkflowString(workflowID, existingWorkflowId, operation, WorkflowID)
+}
+
+func (m *MutableStateMatchEvaluator) compareWorkflowString(inStr string, expectedStr string, operation string, fieldName string) (bool, error) {
```
rodrigozhou, line 211:
```suggestion
func (m *MutableStateMatchEvaluator) compareString(inStr string, expectedStr string, operation string, fieldName string) (bool, error) {
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return false, err
+	}
+	startTime := m.ms.GetExecutionState().StartTime.AsTime()
+	switch operation {
+	case sqlparser.GreaterEqualStr:
+		if startTime == expectedTime {
+			return true, nil
+		}
+		return startTime.After(expectedTime), nil
```
rodrigozhou, line 255: Why not this?
```suggestion
		return startTime.Compare(expectedTime) >= 0, nil
```

rodrigozhou (changes requested):

service/history/workflow/matcher/evaluators.go
ychebotarev: means test is missing...will add some test

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+	"go.temporal.io/server/common/searchattribute"
+	"go.temporal.io/server/service/history/workflow"
+)
+
+// Supported Fields
+const (
+	WorkflowID              = searchattribute.WorkflowID
+	WorkflowTypeName        = searchattribute.WorkflowType
+	WorkflowStartTime       = searchattribute.StartTime
+	WorkflowExecutionStatus = searchattribute.ExecutionStatus
```
bergundy, line 40: Any reason to export these consts?

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+
+type (
+	ExprEvaluator interface {
+		Evaluate(expr sqlparser.Expr) (bool, error)
+	}
+
+	MutableStateMatchEvaluator struct {
+		ms workflow.MutableState
+	}
+)
```
bergundy, line 51:
```suggestion
type ExprEvaluator interface {
	Evaluate(expr sqlparser.Expr) (bool, error)
}

type MutableStateMatchEvaluator struct {
	ms workflow.MutableState
}
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+
+func newMutableStateMatchEvaluator(ms workflow.MutableState) *MutableStateMatchEvaluator {
+	return &MutableStateMatchEvaluator{ms: ms}
+}
+
+func (m *MutableStateMatchEvaluator) Evaluate(expr sqlparser.Expr) (bool, error) {
+	return m.evaluate(expr)
+}
+
+func (m *MutableStateMatchEvaluator) evaluate(expr sqlparser.Expr) (bool, error) {
```
bergundy, line 61: Why do you need this private method with the exact same signature as the public one?

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+		if err != nil {
+			return false, err
+		}
+		return m.compareWorkflowType(val, compExpr.Operator)
+	case WorkflowID:
+		val, err := extractStringValue(valStr)
+		if err != nil {
+			return false, err
+		}
+		return m.compareWorkflowID(val, compExpr.Operator)
```
bergundy, line 147: nit:

```suggestion
		return m.compareWorkflowType(val, op)
	case WorkflowID:
		val, err := extractStringValue(valStr)
		if err != nil {
			return false, err
		}
		return m.compareWorkflowID(val, op)
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+
+	default:
+		return false, fmt.Errorf("unknown or unsupported search attribute name: %s", colNameStr)
+	}
+}
+
+func (m *MutableStateMatchEvaluator) evaluateRange(expr sqlparser.Expr) (bool, error) {
+	rangeCond, ok := expr.(*sqlparser.RangeCond)
+	if !ok {
+		return false, NewMatcherError("%v is not an 'comparison' expression", sqlparser.String(expr))
```
bergundy, line 165:
```suggestion
		return false, NewMatcherError("%v is not a 'comparison' expression", sqlparser.String(expr))
```

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+	case WorkflowStartTime:
+		return m.compareStartTime(valStr, op)
+
+	default:
+		return false, fmt.Errorf("unknown or unsupported search attribute name: %s", colNameStr)
+	}
+}
+
+func (m *MutableStateMatchEvaluator) evaluateRange(expr sqlparser.Expr) (bool, error) {
+	rangeCond, ok := expr.(*sqlparser.RangeCond)
```
bergundy, line 163: Accept a *sqlparser.RangeCond and remove the need to cast again.

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	switch operation {
+	case sqlparser.EqualStr:
+		return strings.EqualFold(expectedStr, inStr), nil
+	case sqlparser.NotEqualStr:
+		return !strings.EqualFold(expectedStr, inStr), nil
+	case sqlparser.StartsWithStr:
+		return strings.HasPrefix(expectedStr, inStr), nil
+	case sqlparser.NotStartsWithStr:
+		return !strings.HasPrefix(expectedStr, inStr), nil
```
bergundy, line 222: `EqualFold` is case insensitive and `HasPrefix` isn't. I don't think we need to do case insensitive matching here at all. There are special cases like execution status that I see is handled separately.

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	// if left is false, then right must be evaluated
+	return m.evaluate(orExpr.Right)
+}
+
+func (m *MutableStateMatchEvaluator) evaluateComparison(expr sqlparser.Expr) (bool, error) {
+	compExpr, ok := expr.(*sqlparser.ComparisonExpr)
+	if !ok {
+		return false, NewMatcherError("%v is not an 'comparison' expression", sqlparser.String(expr))
+	}
```
bergundy, line 121: Better accept the concrete type, you should have casted before entering this function.

```suggestion
func (m *MutableStateMatchEvaluator) evaluateComparison(compExpr *sqlparser.ComparisonExpr) (bool, error) {
```

service/history/workflow/matcher/evaluators.go
bergundy: I must have missed it.

service/history/workflow/matcher/evaluators.go
bergundy: Also the interface isn't used anywhere, I'd just delete it since it's dead code.

service/history/workflow/matcher/util.go
bergundy: SGTM, we can defer this for later.

service/history/workflow/matcher/util.go
bergundy: Where is this function even used? I only see it in tests.

bergundy (approved):
I'm approving although there's still a few comments. Please use your judgement to determine if we need another round or not.

service/history/workflow/matcher/errors_test.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.temporal.io/api/serviceerror"
+)
+
+// TestNewMatcherError ensures that NewMatcherError formats its output correctly
+// and returns an object of type *Error.
+func TestNewMatcherError(t *testing.T) {
```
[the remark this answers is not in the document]
ychebotarev: at least there is a contract that the return error will be InvalidArgument, contract must be tested.

service/history/workflow/matcher/errors.go
```diff
@@ -0,0 +1,40 @@
...
+var (
+	malformedSqlQueryErrMessage = "malformed SQL query"
+	notSupportedErrMessage      = "operation is not supported"
+	invalidExpressionErrMessage = "invalid expression"
+)
+
+func NewMatcherError(format string, a ...any) error {
+	message := fmt.Sprintf(format, a...)
+	return serviceerror.NewInvalidArgument(message)
+}
```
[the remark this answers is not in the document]
ychebotarev: I think there is a value here - uniformity. Whatever is inside NewMatcherError - it is the same.

service/history/workflow/matcher/evaluators.go
ychebotarev: nope. removed

service/history/workflow/matcher/evaluators.go
ychebotarev: it there are multiple types I actually prefer them to be combined in type ( ... ).
But you right, I'm at the end I'm not using that interface right now, and not planning to. I will remove it

service/history/workflow/matcher/evaluators.go
ychebotarev: no need, just the artifact of few refactoring.

ychebotarev: actually, it is cleaner that way. "evaluate" is called from other internal functions like evaluate. I will make a different (better I hope) name for internal function

service/history/workflow/matcher/evaluators.go
ychebotarev: I will do the opposite - remove "op"

service/history/workflow/matcher/evaluators.go
ychebotarev: done

service/history/workflow/matcher/evaluators.go
ychebotarev: discussed. Our comparisons are case-sensitive, everywhere. I will change EqualFold to just '=='

service/history/workflow/matcher/evaluators.go
ychebotarev: removed

service/history/workflow/matcher/evaluators.go
ychebotarev: done for every function

service/history/workflow/matcher/util.go
ychebotarev: I'm planning to use it later, once we support columns like "Attempts"

service/history/workflow/matcher/mutable_state_matcher.go
```diff
@@ -0,0 +1,99 @@
...
+		strings.HasPrefix(strings.ToLower(query), "select ") {
+		return "", fmt.Errorf("invalid filter: %s", query)
+	}
+
+	// sqlparser can't parse just WHERE clause but instead accepts only valid SQL statement.
+	query = fmt.Sprintf("select * from table1 where %s", query)
+
+	return query, nil
+}
+*/
```
rodrigozhou, line 99: Why is this here?

rodrigozhou (changes requested):

service/history/workflow/matcher/mutable_state_matcher.go
ychebotarev: removed

service/history/workflow/matcher/util.go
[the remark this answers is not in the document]
ychebotarev: this is the exact same function that is used in 3 places in visibility.
Same answer as for Roey - yes, it should be re-used, but it is the subject for a separate PR

service/history/workflow/matcher/util.go
```diff
@@ -0,0 +1,77 @@
...
+		return s[1 : len(s)-1], nil
+	}
+	return "", fmt.Errorf("value %s is not a string value", s)
+}
+
+func convertToTime(timeStr string) (time.Time, error) {
```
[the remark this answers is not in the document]
ychebotarev: I agree, but see my comment above.

service/history/workflow/matcher/evaluators.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		switch rangeCond.Operator {
+		case "between":
+			return m.compareStartTimeBetween(fromValue, toValue)
+		case "not between":
+			result, err := m.compareStartTimeBetween(fromValue, toValue)
+			return !result, err
+		default:
+			return false, NewMatcherError("%s: range condition operator must be 'between' or 'not between'", invalidExpressionErrMessage)
```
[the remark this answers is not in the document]
ychebotarev: done, also add operator to the error message

rodrigozhou (approved):

PR #7073, opened 2025-01-13
clone and rename deployment workflows and activities

carlydf:

### DIFF
--- a/docs/architecture/workflow-update.md
+++ b/docs/architecture/workflow-update.md
@@ -169,19 +169,19 @@ An Update is aborted when:
 
 Full "Update state" and "Abort reason" matrix is the following:
 
-| Update State / Abort Reason             | (1) RegistryCleared        | (2) WorkflowCompleted                    | (3) WorkflowContinuing                   |
-|-----------------------------------------|----------------------------|------------------------------------------|------------------------------------------|
-| **Created**                             | `WorkflowUpdateAbortedErr` | `ErrWorkflowCompleted`                   | `ErrWorkflowClosing`                     |
-| **ProvisionallyAdmitted**               | `WorkflowUpdateAbortedErr` | `ErrWorkflowCompleted`                   | `ErrWorkflowClosing`                     |
-| **Admitted**                            | `WorkflowUpdateAbortedErr` | `ErrWorkflowCompleted`                   | `ErrWorkflowClosing`                     |
-| **Sent**                                | `WorkflowUpdateAbortedErr` | `ErrWorkflowCompleted`                   | `ErrWorkflowClosing`                     |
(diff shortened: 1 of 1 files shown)


service/worker/deployment/legacy_deployment_client.go
Shivs11, line 1: I think you will need to change the naming of the functions present here since the "new" deployment client would have the function names repeated under the same package name (deployment) + changing the function names should not be an issue wrt to existing users

Shivs11 (approved):
the renaming is nice + thanks for doing this - also, the tests seem to fail (my assumption is due to lint issues) and once that happens, we should keep an eye out if any of those failed or not

service/worker/deployment/legacy_deployment_client.go
carlydf: since the functions are hanging off of the DeploymentClient, I think renaming the new DeploymentClient --> WorkerDeploymentClient should be enough for name conflicts? The code would not have been able to build if there were name conflicts within the package, so I think the unit test is failing for some other reason, which I'll check

carlydf: for the activities I had to change the names actually

Shivs11 (approved):
reviewed the latest changes too and it's good to go - thanks for making them!
shall make the corresponding API changes now!

service/worker/deployment/util.go
```diff
@@ -40,6 +40,10 @@ import (
 
 const (
 	// Workflow types
+	WorkerDeploymentWorkflowType = "temporal-sys-worker-deployment-workflow"
+	WorkerBuildWorkflowType      = "temporal-sys-worker-build-workflow"
```
ShahabT, line 44: Let's change this to WorkerDeploymentVersion before merging.

service/worker/deployment/activities.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package deployment
```
ShahabT, line 25: thinking maybe we should clone the new stuff in a new package called worker-deployment?

PR #7076, opened 2025-01-14
Delete Namespace: make RPS dynamic

alexshtin:
## What changed?

Delete Namespace: make RPS dynamic.

## Why?

Previously `frontend.deleteNamespaceDeleteActivityRPS` was read once, before deletion of namespace is started and changes of that dynamic config didn't affect already started deletion. It was very inconvenient when huge namespaces were deleted because default RPS is just 100. Now it can be increased and decreased on the fly. 

NOTE: because delete workflow execution is async process, this RPS essentially controls rate at which delete execution tasks are created. This means that decrease of this value (let's say from `1000` to `10`) doesn't immediately slow down the process because already created tasks in transfer queue need to be drained first. 

## How did you test it?

Existing tests and few manual runs.

## Potential risks

No risks. Change is backward compatible. Old parameter is marked as deprecated and needs to be removed after next release (1.27). This is to cover rare case when new `DeleteExecutionsWorkflow` starts old `DeleteExecutionsActivity` during server deployment.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -987,26 +987,26 @@ to allow waiting on the "Accepted" lifecycle stage.`,
 		100,
 		`DeleteNamespaceDeleteActivityRPS is an RPS per every parallel delete executions activity.
 Total RPS is equal to DeleteNamespaceDeleteActivityRPS * DeleteNamespaceConcurrentDeleteExecutionsActivities.
-Default value is 100.`,
+Default value is 100. Despite starting with 'frontend.' this setting is used by a worker and can be changed while namespace is deleted.`,
 	)
 	DeleteNamespacePageSize = NewGlobalIntSetting(
 		"frontend.deleteNamespaceDeletePageSize",
 		1000,
 		`DeleteNamespacePageSize is a page size to read executions from visibility for delete executions activity.
-Default value is 1000.`,
+Default value is 1000. Read once before delete of specified namespace is started.`,
 	)
 	DeleteNamespacePagesPerExecution = NewGlobalIntSetting(
 		"frontend.deleteNamespacePagesPerExecution",
 		256,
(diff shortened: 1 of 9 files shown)


common/dynamicconfig/constants.go
```diff
@@ -987,26 +987,26 @@ to allow waiting on the "Accepted" lifecycle stage.`,
 		100,
 		`DeleteNamespaceDeleteActivityRPS is an RPS per every parallel delete executions activity.
 Total RPS is equal to DeleteNamespaceDeleteActivityRPS * DeleteNamespaceConcurrentDeleteExecutionsActivities.
-Default value is 100.`,
+Default value is 100. Despite starting with 'frontend.' this setting is used by a worker and can be changed while namespace is deleted.`,
```
alexshtin, line 990: I wish I can change DC setting name.

service/worker/deletenamespace/deleteexecutions/activities.go
```diff
@@ -180,7 +188,15 @@ func (a *Activities) DeleteExecutionsActivity(ctx context.Context, params Delete
 		logger.Error("Unable to list all workflow executions.", tag.Error(err))
 		return result, err
 	}
-	rateLimiter := quotas.NewRateLimiter(float64(params.RPS), params.RPS)
+	var rateLimiter *quotas.RateLimiterImpl
+	deleteRPS, cancelDeleteRPSWatch := a.deleteActivityRPS(func(newRPS int) {
+		if rateLimiter != nil {
+			rateLimiter.SetRPS(float64(newRPS))
```
alexshtin, line 194: Thanks to ability to subscribe to DC changes (@dnr) and change RPS on the fly (@wxing1292) this PR is straight forward.

yiminc (approved):

PR #7079, opened 2025-01-14
Refactor workflow id dedup, add callbacks to ms and update versioning override

rodrigozhou:
## What changed?

Refactoring code.

## How did you test it?

Existing unit tests.

### DIFF
--- a/service/history/api/workflow_id_dedup.go
+++ b/service/history/api/workflow_id_dedup.go
@@ -67,42 +67,77 @@ func ResolveDuplicateWorkflowID(
 	switch currentState {
 	// *running* workflow: apply WorkflowIdConflictPolicy
 	case enumsspb.WORKFLOW_EXECUTION_STATE_CREATED, enumsspb.WORKFLOW_EXECUTION_STATE_RUNNING:
-		switch wfIDConflictPolicy {
-		case enumspb.WORKFLOW_ID_CONFLICT_POLICY_FAIL:
-			msg := "Workflow execution is already running. WorkflowId: %v, RunId: %v."
-			return nil, generateWorkflowAlreadyStartedError(msg, currentStartRequestID, workflowKey)
-		case enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING:
-			return nil, ErrUseCurrentExecution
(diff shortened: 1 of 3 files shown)


service/history/api/workflow_id_dedup.go
```diff
@@ -67,42 +67,77 @@ func ResolveDuplicateWorkflowID(
...
+	case enumspb.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE:
+		msg := "Workflow execution already finished. WorkflowId: %v, RunId: %v. Workflow Id reuse policy: reject duplicate workflow Id."
+		return generateWorkflowAlreadyStartedError(msg, currentStartRequestID, workflowKey)
+	default:
+		return serviceerror.NewInternal(
+			fmt.Sprintf("Failed to process start workflow id reuse policy: %v.", wfIDReusePolicy),
+		)
+	}
+	// ie "allow" starting a new workflow
+	return nil
```
ychebotarev, line 140: this should be "unreachable code"

ychebotarev (approved):

service/history/api/workflow_id_dedup.go
rodrigozhou: It's not unreachable. The first two cases in the switch-case can reach here.

bergundy (approved):

service/history/workflow/mutable_state_impl.go
```diff
@@ -2469,6 +2443,40 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+			err := proto.Unmarshal(cb.GetInternal().GetData(), persistenceCB)
+			if err != nil {
+				return err
+			}
+		}
+		machine := callbacks.NewCallback(event.EventTime, callbacks.NewWorkflowClosedTrigger(), persistenceCB)
+		// Use the start event version and ID as part of the callback ID to ensure that callbacks have unique
+		// IDs that are deterministically created across clusters.
+		// TODO: Replicate the state machine state and allocate a uuid there instead of relying on history event
+		// replication.
```
bergundy, line 2471: This isn't relevant anymore, we could probably get rid of the namespace failover version in the state machine ID but it'd be tricky to change now.

PR #7080, opened 2025-01-14
Add OnConflictOptions to StartWorkflowExecution

rodrigozhou:
## What changed?

Add `OnConflictOptions` to `StartWorkflowExecution`.
- if the workflow id conflict policy is not `USE_EXISTING`, then no-op;
- if the field is nil, then no-op;
- otherwise, then it will add a `WorkflowExecutionOptionsUpdatedEvent` to the existing running workflow.

## Why?

Ability to attach completion callbacks to existing running workflows.

## How did you test it?

WIP: writing tests


service/frontend/workflow_handler.go
```diff
@@ -475,6 +475,10 @@ func (wh *WorkflowHandler) prepareStartWorkflowRequest(
 	enums.SetDefaultWorkflowIdReusePolicy(&request.WorkflowIdReusePolicy)
 	enums.SetDefaultWorkflowIdConflictPolicy(&request.WorkflowIdConflictPolicy, enumspb.WORKFLOW_ID_CONFLICT_POLICY_FAIL)
 
+	if err := wh.validateOnConflictOptions(request.OnConflictOptions); err != nil {
+		return nil, err
+	}
```
bergundy, line 480: Redundant for now but 🤷

service/history/history_engine2_test.go
```diff
@@ -1361,6 +1362,28 @@ func (s *engine2Suite) TestStartWorkflowExecution_Dedup_Running_SameRequestID()
...
+	startRequest.StartRequest.RequestId = s.tv.String("AttachedRequestID1")
+	resp, err := s.historyEngine.StartWorkflowExecution(metrics.AddMetricsContext(context.Background()), startRequest)
+	s.NoError(err)
+	s.True(resp.Started)
+	s.Equal(s.tv.RunID(), resp.GetRunId())
+
+	startRequest.StartRequest.RequestId = s.tv.String("AttachedRequestID2")
+	resp, err = s.historyEngine.StartWorkflowExecution(metrics.AddMetricsContext(context.Background()), startRequest)
+	s.NoError(err)
+	s.True(resp.Started)
```
bergundy, line 1383: Maybe I'm not following this but shouldn't `Started` be false here since we're attaching to an existing workflow?

service/history/historybuilder/event_factory.go
```diff
@@ -395,13 +395,19 @@ func (b *EventFactory) CreateWorkflowExecutionTerminatedEvent(
 
 func (b *EventFactory) CreateWorkflowExecutionOptionsUpdatedEvent(
 	versioningOverride *workflowpb.VersioningOverride,
+	attachRequestID string,
+	attachCompletionCallbacks []*commonpb.Callback,
+	links []*commonpb.Link,
```
bergundy, line 400: Maybe instead accept a historypb.WorkflowExecutionOptionsUpdatedEventAttributes and links here to avoid this getting out of control?

service/history/historybuilder/history_builder.go
```diff
@@ -413,8 +413,16 @@ func (b *HistoryBuilder) AddWorkflowExecutionTerminatedEvent(
 
 func (b *HistoryBuilder) AddWorkflowExecutionOptionsUpdatedEvent(
 	versioningOverride *workflowpb.VersioningOverride,
+	attachRequestID string,
+	attachCompletionCallbacks []*commonpb.Callback,
```
bergundy, line 417: Same as above, probably better to accept the attributes here.

service/history/workflow/mutable_state.go
```diff
@@ -224,7 +224,12 @@ type (
 		AddWorkflowExecutionStartedEvent(*commonpb.WorkflowExecution, *historyservice.StartWorkflowExecutionRequest) (*historypb.HistoryEvent, error)
 		AddWorkflowExecutionStartedEventWithOptions(*commonpb.WorkflowExecution, *historyservice.StartWorkflowExecutionRequest, *workflowpb.ResetPoints, string, string) (*historypb.HistoryEvent, error)
 		AddWorkflowExecutionTerminatedEvent(firstEventID int64, reason string, details *commonpb.Payloads, identity string, deleteAfterTerminate bool, links []*commonpb.Link) (*historypb.HistoryEvent, error)
-		AddWorkflowExecutionOptionsUpdatedEvent(versioningOverride *workflowpb.VersioningOverride) (*historypb.HistoryEvent, error)
+		AddWorkflowExecutionOptionsUpdatedEvent(
+			versioningOverride *workflowpb.VersioningOverride,
+			attachRequestID string,
+			attachCompletionCallbacks []*commonpb.Callback,
```
bergundy, line 230: Same, accept the event attributes here instead.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4410,6 +4425,14 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionOptionsUpdatedEvent(event *his
 	if err := ms.updateVersioningOverride(attributes.GetVersioningOverride()); err != nil {
 		return err
 	}
+	if attributes.GetAttachedRequestId() != "" {
+		ms.attachRequestID(attributes.GetAttachedRequestId())
+	}
+	if len(attributes.GetAttachedCompletionCallbacks()) > 0 {
```
bergundy, line 4431: We need to verify that the total number of attached callbacks on this workflow doesn't exceed the configured limit.

bergundy (commented):
I want to see functional test coverage for as many of the possible cases. I don't feel comfortable merging this otherwise.
Some tests off the top of my head:
- Test that the event is created as expected with links, callbacks, and request ID attached.
- Test that callbacks appear in the describe response.
- Test request ID based deduping works for the attached request ID

service/frontend/workflow_handler.go
rodrigozhou: Yes, I know is unnecessary for now. Just being preemptive adding the function to validate.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -294,6 +293,11 @@ message WorkflowExecutionState {
     temporal.api.enums.v1.WorkflowExecutionStatus status = 4;
     VersionedTransition last_update_versioned_transition = 5;
     google.protobuf.Timestamp start_time = 6;
+
+    // Request IDs that are attached to the workflow execution.
+    // They can be from StartWorkflowExecution requests with conflict policy USE_EXISTING and
+    // conflict options to attached completion callbacks.
+    repeated string attached_request_ids = 7;
```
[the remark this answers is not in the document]
rodrigozhou: Hm, if we decide to add some metadata later, we would still need to change the type, so I don't think setting it to `map[string]Empty` at this moment would make any difference.

Also, I'm not sure it makes sense to evict request ID/callbacks. If we reach our limit to attach request ids, shouldn't we reject the request all together? Say we evict callbacks, then some caller won't ever receive the callback without even knowing.

service/history/history_engine2_test.go
rodrigozhou: Oh, yes! Nice catch!

proto/internal/temporal/server/api/persistence/v1/executions.proto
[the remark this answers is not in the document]
rodrigozhou: I think we have to keep separate.
As you caught the bug below related to `resp.Started` needs to be `false` for attaching request id, but it should be `true` if deduping the `create_request_id`.

service/history/historybuilder/event_factory.go
rodrigozhou: Hm, I think it would make sense to use `historyservice.UpdateWorkflowExecutionOptionsRequest`, but it would add some logic in the event factory (do the mask processing here), which I'm not sure it makes sense.
I'd leave as it is for now, and refactor later. I'm already planning to refactor the `startworkflow` api to call the `updateworkflowwithoptions` api in which I can address this issue together.

common/persistence/serialization/blob.go
```diff
@@ -125,6 +125,18 @@ func QueueStateFromBlob(blob []byte, encoding string) (*persistencespb.QueueStat
...
+	return proto3Encode(requestIDs)
+}
+
+func WorkflowExecutionRequestIDsFromBlob(blob []byte, encoding string) (*persistencespb.WorkflowExecutionRequestIDs, error) {
+	if blob == nil {
+		return nil, nil
+	}
+	result := &persistencespb.WorkflowExecutionRequestIDs{}
+	return result, proto3Decode(blob, encoding, result)
+}
```
rodrigozhou, line 138: This is code duplication for now.
We need to refactor the code to pass the serializer object to the DB plugins to be able to serialize/deserialize fields from DB.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -294,6 +293,22 @@ message WorkflowExecutionState {
     temporal.api.enums.v1.WorkflowExecutionStatus status = 4;
     VersionedTransition last_update_versioned_transition = 5;
     google.protobuf.Timestamp start_time = 6;
+
+    // Request IDs that are attached to the workflow execution.
+    // They can be from StartWorkflowExecution requests with conflict policy USE_EXISTING and
+    // conflict options to attached completion callbacks.
+    WorkflowExecutionRequestIDs attached_request_ids = 7;
```
rodrigozhou, line 300: @bergundy With this, we can slowly deprecate the `create_request_id`, but I'm gonna do in a separate PR

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -294,6 +293,22 @@ message WorkflowExecutionState {
...
+}
+
+message WorkflowExecutionRequestIDs {
```
rodrigozhou, line 303: Should this be `IDs` or `Ids` (follow proto pattern)?

service/history/historybuilder/event_factory.go
```diff
@@ -396,14 +396,20 @@ func (b *EventFactory) CreateWorkflowExecutionTerminatedEvent(
 func (b *EventFactory) CreateWorkflowExecutionOptionsUpdatedEvent(
 	versioningOverride *workflowpb.VersioningOverride,
 	unsetVersioningOverride bool,
+	attachRequestID string,
+	attachCompletionCallbacks []*commonpb.Callback,
```
bergundy, line 400: Still think you should just accept `*historypb.WorkflowExecutionOptionsUpdatedEventAttributes` here and avoid this function signature growing every time we add another attribute.

service/history/workflow/mutable_state_impl.go
bergundy: I saw that you've added the check above, please make sure to have test coverage for this.

service/history/api/startworkflow/api.go
```diff
@@ -617,6 +625,79 @@ func (s *Starter) getWorkflowHistory(ctx context.Context, mutableState *mutableS
...
+	case consts.ErrWorkflowCompleted:
+		// Need to re-evaluate the reuse policy because the previous check didn't lock the current
+		// execution. So it's possible the workflow completed after the first call of
+		// api.ResolveDuplicateWorkflowID and before being able to update the existing workflow.
+		err := api.ResolveWorkflowIDReusePolicy(
+			workflowKey,
+			currentWorkflowConditionFailed.Status,
+			currentWorkflowConditionFailed.RequestID,
+			s.request.StartRequest.GetWorkflowIdReusePolicy(),
+		)
```
bergundy, line 689: Is `currentWorkflowConditionFailed.RequestID` the current request's ID? Should we be checking that the request IDs in the DB and request matches? (asking for lack of context on this part of the code).

bergundy (commented):
Note that we also need to ensure that the WorkflowExecutionOptionsUpdated event is cherry picked (reapplied) on conflict resolution.

service/history/api/startworkflow/api.go
rodrigozhou: `currentWorkflowConditionFailed.RequestID` is the existing current workflow, which can be running or completed already.

common/persistence/cassandra/mutable_state_store.go
```diff
@@ -748,12 +749,21 @@ func (d *MutableStateStore) ConflictResolveWorkflowExecution(
...
-		executionStateDatablob, err := serialization.WorkflowExecutionStateToBlob(&persistencespb.WorkflowExecutionState{
-			RunId:           runID,
-			CreateRequestId: createRequestID,
-			State:           state,
-			Status:          status,
-		})
+		executionStateDatablob, err := serialization.WorkflowExecutionStateToBlob(
+			&persistencespb.WorkflowExecutionState{
+				RunId:           runID,
+				CreateRequestId: createRequestID,
```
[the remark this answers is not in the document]
rodrigozhou: I'd leave for a while first to transition to use the new field.

common/persistence/sql/sqlplugin/history_execution.go
```diff
@@ -61,15 +61,17 @@ type (
...
+		NamespaceID                primitives.UUID
+		WorkflowID                 string
+		RunID                      primitives.UUID
+		CreateRequestID            string
+		StartTime                  *time.Time
+		LastWriteVersion           int64
+		State                      enumsspb.WorkflowExecutionState
+		Status                     enumspb.WorkflowExecutionStatus
+		AttachedRequestIDs         []byte `db:"attached_request_ids"`
+		AttachedRequestIDsEncoding string `db:"attached_request_ids_encoding"`
```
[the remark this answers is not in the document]
rodrigozhou: Changed to a generate data blob.

service/history/historybuilder/event_factory.go
rodrigozhou: It's a bit complicated. This is the event factory, ie., that's the function that builds `*historypb.WorkflowExecutionOptionsUpdatedEventAttributes`. I want to change to get the history request, but then it would add some logic to translate the request input to event attributes, which is not something we did so far.

common/persistence/data_interfaces.go
```diff
@@ -122,13 +122,18 @@ type (
 
 	// CurrentWorkflowConditionFailedError represents a failed conditional update for current workflow record
 	CurrentWorkflowConditionFailedError struct {
```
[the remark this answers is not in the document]
rodrigozhou: Hm, yes, we can do that, but it's kind of a big refactor. I can do this in another PR.

common/persistence/cassandra/mutable_state_store.go
```diff
@@ -747,13 +748,20 @@ func (d *MutableStateStore) ConflictResolveWorkflowExecution(
...
-		executionStateDatablob, err := serialization.WorkflowExecutionStateToBlob(&persistencespb.WorkflowExecutionState{
-			RunId:           runID,
-			CreateRequestId: createRequestID,
-			State:           state,
-			Status:          status,
-		})
+		startTime := executionState.StartTime
+		details := executionState.Details
+
+		// TODO: should this just serialize the executionState variable?
```
rodrigozhou, line 754: Wondering if we should just serialize `executionState` instead of building another object.

I already noticed that it was missing setting the `StartTime` below as well as the `LastUpdateVersioningTransition`.

For the `StartTime`, it's working because there's a handling for missing value since it's in transition phase from removing the field from `ExecutionInfo`.

For `LastUpdateVersioningTransition`, I don't know if it's supposed to copy from `executionState` set above in L741-746.

cc: @yycptt

common/testing/historyrequire/history_require.go
```diff
@@ -494,6 +494,13 @@ func (h HistoryRequire) equalExpectedMapToActualAttributes(expectedMap map[strin
 			require.Failf(h.t, "", "Expected property %s.%s wasn't found for EventID=%v", attrPrefix, attrName, eventID)
 		}
 
+		if expectedValue == nil {
+			if !actualV.IsNil() {
+				require.Failf(h.t, "", "Value of property %s.%s for EventID=%v expected to be nil", attrPrefix, attrName, eventID)
+			}
+			continue
+		}
+
```
rodrigozhou, line 503: This is a fix in the history event attributes assertion.
If the field is expected to be `nil`, the assertion below at L527 will fail because `actualV` is something like `var actualV any = (*T)(nil)`, and this is different than `var expectedValue any = nil`.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2317,8 +2334,13 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 	}
 
 	ms.approximateSize -= ms.executionInfo.Size()
+	ms.approximateSize -= ms.executionState.Size()
```
rodrigozhou, line 2337: I think this is a bug in the mutable approximate size. In L2339, it sets the `CreateRequestId`, but it's not updating the size.
So, I added this line and L2474 to correct the approximate size.

cc: @yycptt

common/persistence/serialization/blob.go
bergundy: Will you do the refactor? Not blocking this PR, but ideally we'd at least track this.

common/persistence/data_interfaces.go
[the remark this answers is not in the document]
bergundy: Sounds good.

common/persistence/cassandra/mutable_state_store.go
yycptt: 👍 Thanks! Yeah those two fields are missing. I didn't notice we were constructing a new state object here  when adding `LastUpdateVersioningTransition`. Guess the same thing happened for `StartTime`. 

Though at least for `LastUpdateVersioningTransition` I don't think there's any real impact. That field is only used via mutable state which is initialized by the concrete execution row that have the all the fields there. The blob constructed here (and I noticed the same thing for sql) is only for current execution row. 

cc @xwduan @hai719

service/history/workflow/mutable_state_impl.go
yycptt: Thanks!

common/persistence/cassandra/mutable_state_store.go
yycptt: so we can just use the serialized blob in the `request`?

rodrigozhou: Fixed! Now it's serializing the `executionState` var instead of copying.

common/persistence/cassandra/mutable_state_store.go
```diff
@@ -744,59 +744,34 @@ func (d *MutableStateStore) ConflictResolveWorkflowExecution(
...
-		status := executionState.Status
-
-		executionStateDatablob, err := serialization.WorkflowExecutionStateToBlob(&persistencespb.WorkflowExecutionState{
-			RunId:           runID,
-			CreateRequestId: createRequestID,
-			State:           state,
-			Status:          status,
-		})
+
+		executionStateDatablob, err := serialization.WorkflowExecutionStateToBlob(executionState)
```
yycptt, line 749: use the ExecutionStateBlob in the request?

common/persistence/serialization/serializer.go
```diff
@@ -427,7 +430,23 @@ func (t *serializerImpl) WorkflowExecutionStateToBlob(info *persistencespb.Workf
 
 func (t *serializerImpl) WorkflowExecutionStateFromBlob(data *commonpb.DataBlob) (*persistencespb.WorkflowExecutionState, error) {
 	result := &persistencespb.WorkflowExecutionState{}
-	return result, ProtoDecodeBlob(data, result)
+	if err := ProtoDecodeBlob(data, result); err != nil {
+		return nil, err
+	}
+	// Initialize the WorkflowExecutionStateDetails for old records.
```
yycptt, line 436: reuse WorkflowExecutionStateFromBlob?

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -294,6 +294,20 @@ message WorkflowExecutionState {
     temporal.api.enums.v1.WorkflowExecutionStatus status = 4;
     VersionedTransition last_update_versioned_transition = 5;
     google.protobuf.Timestamp start_time = 6;
+    // New fields added to WorkflowExecutionState should be added to WorkflowExecutionStateDetails
+    // to avoid modifying the SQL table schema.
+    WorkflowExecutionStateDetails details = 7;
```
yycptt, line 299: not a huge fan of of this... I feel this should be handled by the sql execution store impl (create this separate details message with certain fields in ExecutionState definition and do an extra encoding step.) instead of moving the concern up the layer.

I guess I still hope eventually, the current execution table will just have one blob column which has everything here. With the current approach, I don't feel that will happen as we are pushing all new fields to the Details message and end up with a weird definition for WorkflowExecutionState (only one Details field with everything in it.)

service/history/workflow/mutable_state_impl.go
```diff
@@ -2090,6 +2093,20 @@ func (ms *MutableStateImpl) DeleteSignalRequested(
 	ms.approximateSize -= len(requestID)
 }
 
+func (ms *MutableStateImpl) attachRequestID(requestID string, eventType enumspb.EventType) {
```
yycptt, line 2096: Set `executionStateUpdated` to true in this func so that the closeTransaction logic can update lastUpdateVersionedTransition for execution state.

service/history/api/startworkflow/api.go
```diff
@@ -317,9 +317,18 @@ func (s *Starter) handleConflict(
 	currentWorkflowConditionFailed *persistence.CurrentWorkflowConditionFailedError,
 ) (*historyservice.StartWorkflowExecutionResponse, StartOutcome, error) {
 	request := s.request.StartRequest
-	if currentWorkflowConditionFailed.RequestID == request.GetRequestId() {
-		resp, err := s.respondToRetriedRequest(ctx, currentWorkflowConditionFailed.RunID)
-		return resp, StartDeduped, err
+	currentWorkflowRequestIDs := currentWorkflowConditionFailed.RequestIDs
+	if requestIDInfo := currentWorkflowRequestIDs[request.GetRequestId()]; requestIDInfo != nil {
```
yycptt, line 321: nit
```suggestion
	if _, ok := currentWorkflowRequestIDs[request.GetRequestId()]; ok {
```

tests/workflow_test.go
```diff
@@ -127,6 +129,288 @@ func (s *WorkflowTestSuite) TestStartWorkflowExecution() {
 	})
 }
 
+func (s *WorkflowTestSuite) TestStartWorkflowExecution_UseExisting() {
+	id := "functional-start-workflow-use-existing-test"
```
bergundy, line 133: nit: `testcore.RandomizeString(s.T().Name())` and don't bother with the naming, it's nice for duplicating tests later.

Same for other variables here.

bergundy: Or just use testvars.

tests/workflow_test.go
```diff
@@ -127,6 +129,288 @@ func (s *WorkflowTestSuite) TestStartWorkflowExecution() {
...
+		},
+		{
+			name:                     "OnConflictOptions failed max callbacks per workflow",
+			WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
+			OnConflictOptions: &workflowpb.OnConflictOptions{
+				AttachRequestId:           true,
+				AttachCompletionCallbacks: true,
+				AttachLinks:               true,
+			},
+			ErrMessage:              "cannot attach more than 1 callbacks to a workflow (1 callbacks already attached)",
```
bergundy, line 212: Please also test that the request ID isn't attached on workflow execution state in the current executions table here.

tests/workflow_test.go
```diff
@@ -127,6 +129,288 @@ func (s *WorkflowTestSuite) TestStartWorkflowExecution() {
...
+
+	testCases := []struct {
+		name                     string
+		WorkflowIdConflictPolicy enumspb.WorkflowIdConflictPolicy
+		OnConflictOptions        *workflowpb.OnConflictOptions
+		ErrMessage               string
+		MaxCallbacksPerWorkflow  int
+	}{
+		{
+			name:                     "OnConflictOptions nil",
```
bergundy, line 184: Isn't this a duplicate of the test above?

tests/workflow_test.go
```diff
@@ -127,6 +129,288 @@ func (s *WorkflowTestSuite) TestStartWorkflowExecution() {
...
+	request.WorkflowIdConflictPolicy = enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
+	request.OnConflictOptions = &workflowpb.OnConflictOptions{
+		AttachRequestId: true,
+	}
+	we1, err1 := s.FrontendClient().StartWorkflowExecution(testcore.NewContext(), request)
+	s.NoError(err1)
+	s.Equal(we0.RunId, we1.RunId)
+	s.False(we1.Started)
+
+	we2, err2 := s.FrontendClient().StartWorkflowExecution(testcore.NewContext(), request)
```
bergundy, line 392: Worth doing another request without the OnConflictOptions to see that it gets deduped to for completeness.

bergundy (commented):
Overall LGTM but I'm missing a few test cases:

- Start without OnConflictOptions and same different request ID fails (for good measure)
- Start fails due to validation (e.g. max callback limit reached), does not attach a request ID to the current executions table (e.g. a following request doesn't get duduped).
- XDC test and/or reset test that verifies the update event is applied and start requests get properly deduped

proto/internal/temporal/server/api/persistence/v1/executions.proto
rodrigozhou: Changed so it's just a SQL implementation detail.

service/history/api/startworkflow/api.go
```diff
@@ -415,25 +424,22 @@ func (s *Starter) resolveDuplicateWorkflowID(
 		newRunID,
 		currentWorkflowConditionFailed.State,
 		currentWorkflowConditionFailed.Status,
-		currentWorkflowConditionFailed.RequestID,
+		currentWorkflowConditionFailed.CreateRequestID,
```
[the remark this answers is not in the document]
rodrigozhou: The error handling only needs/uses the start request ID.

common/persistence/data_interfaces.go
```diff
@@ -122,8 +122,12 @@ type (
...
-		Msg              string
-		RequestID        string
+		Msg string
+		// RequestIDs contains all request IDs associated with the workflow execution, ie., contain the
+		// request ID that started the workflow execution as well as the request IDs that were attached
+		// to the workflow execution when it was running.
+		RequestIDs       map[string]*persistencespb.RequestIDInfo
+		CreateRequestID  string
```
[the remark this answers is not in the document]
rodrigozhou: I only added this because some places needs the `CreateRequestID`. In the start workflow path, it needs to building the `WorkflowAlreadyStartedError`. In the signal with start path, it needs for deduping, but I can change to check on `RequestIDs`.

tests/workflow_test.go
rodrigozhou: Oh, yes. I'll dedup it.

common/persistence/sql/execution_util.go
```diff
@@ -1152,3 +1161,28 @@ func updateExecution(
 
 	return nil
 }
+
+func workflowExecutionStateFromCurrentExecutionsRow(
```
yycptt, line 1165: shall we change other places that uses current execution row to use the blob instead? e.g. GetCurrentExecution in sql/execution.go.

we still do double write so doesn't really matter right now.

yycptt (approved):

common/persistence/sql/execution_util.go
rodrigozhou: I thought about the same, but I'll check and see if it's gonna be an improvement or not in another PR.

tests/workflow_test.go
rodrigozhou: Created a new test to verify this.

rodrigozhou:
> XDC test and/or reset test that verifies the update event is applied and start requests get properly deduped

@bergundy I created a separate PR that reapplies the WorkflowExecutionOptionsUpdated history event: https://github.com/temporalio/temporal/pull/7221

bergundy (approved):

PR #7081, opened 2025-01-14
Use task context to save the callback result

bergundy:
Previously we were using the call context which can timeout and prevent saving the callback result if the call times out.

### DIFF
--- a/components/callbacks/executors.go
+++ b/components/callbacks/executors.go
@@ -143,7 +143,7 @@ func (e taskExecutor) executeInvocationTask(
 	defer cancel()
 
 	result := invokable.Invoke(callCtx, ns, e, task)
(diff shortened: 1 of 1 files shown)


pdoerner (approved):

components/callbacks/executors.go
```diff
@@ -143,7 +143,7 @@ func (e taskExecutor) executeInvocationTask(
 	defer cancel()
 
 	result := invokable.Invoke(callCtx, ns, e, task)
-	saveErr := e.saveResult(callCtx, env, ref, result)
+	saveErr := e.saveResult(ctx, env, ref, result)
```
yycptt, line 146: does this ctx have any timeout/deadline?

bergundy: That's the task context, it shouldn't have a deadline AFAIK.

PR #7082, opened 2025-01-14
Add handling for links in Nexus sync success responses

pdoerner:
## What changed?

Added handling for links in Nexus sync success responses

### DIFF
--- a/components/nexusoperations/completion.go
+++ b/components/nexusoperations/completion.go
@@ -204,7 +204,7 @@ func CompletionHandler(
 			if opFailedError != nil {
 				return handleUnsuccessfulOperationError(node, operation, opFailedError, CompletionSourceCallback)
 			}
-			return handleSuccessfulOperationResult(node, operation, result, links, CompletionSourceCallback)
+			return handleSuccessfulOperationResult(node, operation, result, nil, CompletionSourceCallback)
 		})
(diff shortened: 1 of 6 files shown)


components/nexusoperations/completion.go
```diff
@@ -202,7 +204,7 @@ func CompletionHandler(
 			if opFailedError != nil {
 				return handleUnsuccessfulOperationError(node, operation, opFailedError, CompletionSourceCallback)
 			}
-			return handleSuccessfulOperationResult(node, operation, result, CompletionSourceCallback)
+			return handleSuccessfulOperationResult(node, operation, result, links, CompletionSourceCallback)
```
pdoerner, line 207: The original intent for adding links to `CompletionHandler` was for fabricating them on started events when receiving a completion before a start. Not sure whether it makes sense to have separate fields for those links and the ones that should be attached to the completion event, but for now they are just being duplicated.

bergundy: The links on the completion handler go on the started event. The links on the sync completion go on the completed event.

bergundy: You should set this to nil here.

bergundy (approved):
Had a single comment, please address before merging.

PR #7083, opened 2025-01-14
Fix functional subtest execution

alexshtin:
## What changed?

Fix functional subtest execution.
Credits to @stephanos for the brilliant idea!

## Why?

Previously, if there is a assert failure (like `s.Equal`) in subtest, it failed entire test and stoped executing other subtests. Error message went to parent test as well, while subtest failed with wired message "subtest paniced". This is because we override `s.Assertions` in every test with `require.New` and it calls `FailNow` when condition is not meet. Same thing must to be done for every subtest. And every test helper which depends on `s.T()`.

## How did you test it?

Simulate failure of some sub tests and checked results.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/tests/testcore/client_suite.go
+++ b/tests/testcore/client_suite.go
@@ -136,6 +136,13 @@ func (s *ClientFunctionalSuite) TearDownTest() {
 	}
 }
 
+func (s *ClientFunctionalSuite) SetupSubTest() {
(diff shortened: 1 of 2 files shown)


stephanos (approved):
Perfect!

PR #7084, opened 2025-01-14
Use callerType to determine history requests priority

yycptt:
## What changed?

- Use caller type to determine history requests priority

## Why?

- Right now the priority feature is only used for prioritizing operator requests, but doesn't help with background requests increase case which can cause user requests to be throttled as well.

### DIFF
--- a/service/history/configs/quotas.go
+++ b/service/history/configs/quotas.go
@@ -36,81 +36,14 @@ const (
 )
 
 var (
-	APIToPriority = map[string]int{
-		"/temporal.server.api.historyservice.v1.HistoryService/CloseShard":                             1,
-		"/temporal.server.api.historyservice.v1.HistoryService/GetShard":                               1,
-		"/temporal.server.api.historyservice.v1.HistoryService/DeleteWorkflowExecution":                1,
-		"/temporal.server.api.historyservice.v1.HistoryService/DescribeHistoryHost":                    1,
-		"/temporal.server.api.historyservice.v1.HistoryService/DescribeMutableState":                   1,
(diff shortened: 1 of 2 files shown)


service/history/configs/quotas.go
```diff
@@ -36,81 +36,14 @@ const (
...
-		"/temporal.server.api.historyservice.v1.HistoryService/GetWorkflowExecutionHistoryReverse":     1,
-		"/temporal.server.api.historyservice.v1.HistoryService/GetWorkflowExecutionRawHistory":         1,
-		"/temporal.server.api.historyservice.v1.HistoryService/GetWorkflowExecutionRawHistoryV2":       1,
-		"/temporal.server.api.historyservice.v1.HistoryService/ForceDeleteWorkflowExecution":           1,
-		"/temporal.server.api.historyservice.v1.HistoryService/GetDLQTasks":                            1,
-		"/temporal.server.api.historyservice.v1.HistoryService/DeleteDLQTasks":                         1,
-		"/temporal.server.api.historyservice.v1.HistoryService/AddTasks":                               1,
-		"/temporal.server.api.historyservice.v1.HistoryService/ListQueues":                             1,
-		"/temporal.server.api.historyservice.v1.HistoryService/ListTasks":                              1,
-		"/temporal.server.api.historyservice.v1.HistoryService/CompleteNexusOperation":                 1,
```
yycptt, line 103: hmmm, looks like caller type is not specified for CompleteNexusOperation

yycptt: The logic will default the priority to 1, so actually increases the priority for this api, as other background task processing requests will be using 2. But this should be fixed.

prathyushpv (approved):

PR #7090, opened 2025-01-15
Allow disabling the Updates in-flight limit

stephanos:
## What changed?

If the maximum in-flight limit for updates is zero, disable the check.

## Why?

Preparation to move towards a payload-based check. Having the option to turn this limit off is useful.

## How did you test it?

Added test.

### DIFF
--- a/service/history/workflow/update/registry_test.go
+++ b/service/history/workflow/update/registry_test.go
@@ -213,6 +213,7 @@ func TestFindOrCreate(t *testing.T) {
 	tv1 := tv.WithUpdateIDNumber(1)
 	tv2 := tv.WithUpdateIDNumber(2)
 	tv3 := tv.WithUpdateIDNumber(3)
+	tv4 := tv.WithUpdateIDNumber(4)
 
 	t.Run("find stored update", func(t *testing.T) {
 		reg := update.NewRegistry(&mockUpdateStore{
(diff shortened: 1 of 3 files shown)


service/history/workflow/update/registry_test.go
```diff
@@ -304,7 +305,7 @@ func TestFindOrCreate(t *testing.T) {
 			limit += 1
 
 			_, existed, err = reg.FindOrCreate(context.Background(), tv2.UpdateID())
-			require.NoError(t, err, "creating update #2 should have beeen created after limit increase")
+			require.NoError(t, err, "update #2 should have beeen created after limit increase")
```
stephanos, line 308: wording tweak

service/history/workflow/update/registry_test.go
```diff
@@ -313,10 +314,19 @@ func TestFindOrCreate(t *testing.T) {
 			assertRejectUpdateInRegistry(t, reg, evStore, upd1)
 
 			_, existed, err = reg.FindOrCreate(context.Background(), tv3.UpdateID())
-			require.NoError(t, err, "update #3 should be created after #1 completed")
+			require.NoError(t, err, "update #3 should have been created after #1 completed")
```
stephanos, line 317: wording tweak

alexshtin (approved):

PR #7091, opened 2025-01-16
use UnsetVersioningOverride field in ApplyWorkflowExecutionOptionsUpdatedEvent

carlydf:
## What changed?
use UnsetVersioningOverride field in ApplyWorkflowExecutionOptionsUpdatedEvent

## Why?
So that users of this event don't need to load VersioningOverride from mutable state every time they create this event.
Now, a nil Versioning Override in this event means "no change" instead of "remove".
This reduces the chance that someone accidentally unsets an override in the future, and also is more efficient.
We've discussed this change internally in the server team and are ok with changing the meaning of this history event, because it is such a small change and the scope of impact is small (pre-release versioning users who have unset a versioning override and are building mutable state from that history).

## How did you test it?
Made sure that versioning override functional tests pass.

## Potential risks
Now, a nil Versioning Override in this event means "no change" instead of "remove".
If an event exists with the previous meaning and the mutable state is rebuilt, the Versioning Override would not be removed.
But the chance of that happening is very low.


service/history/workflow/mutable_state_impl.go
```diff
@@ -4409,6 +4410,9 @@ func (ms *MutableStateImpl) AddWorkflowExecutionOptionsUpdatedEvent(
 
 func (ms *MutableStateImpl) ApplyWorkflowExecutionOptionsUpdatedEvent(event *historypb.HistoryEvent) error {
 	override := event.GetWorkflowExecutionOptionsUpdatedEventAttributes().GetVersioningOverride()
+	if event.GetWorkflowExecutionOptionsUpdatedEventAttributes().GetUnsetVersioningOverride() {
+		override = nil
+	}
```
rodrigozhou, line 4415: You need to check if `override == nil && !unsetVersioningOverride`. In this case, you don't want to do anything, right?

```suggestion
	attributes := event.GetWorkflowExecutionOptionsUpdatedEventAttributes()
	override := attributes.GetVersioningOverride()
	if attributes.GetUnsetVersioningOverride() {
		override = nil
	} else if override == nil {
		return
	}
```

carlydf: omg thank you so much for catching that!

service/history/api/updateworkflowoptions/api.go
```diff
@@ -122,7 +126,12 @@ func Invoke(
 func getOptionsFromMutableState(ms workflow.MutableState) *workflowpb.WorkflowExecutionOptions {
 	opts := &workflowpb.WorkflowExecutionOptions{}
 	if versioningInfo := ms.GetExecutionInfo().GetVersioningInfo(); versioningInfo != nil {
-		opts.VersioningOverride = versioningInfo.GetVersioningOverride()
+		if override := versioningInfo.GetVersioningOverride(); override != nil {
+			opts.VersioningOverride = &workflowpb.VersioningOverride{
+				Behavior:   override.Behavior,
+				Deployment: override.Deployment,
+			}
+		}
```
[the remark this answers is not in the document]
carlydf: yes probably, I just didn't think of it since there were only two fields. but that's cleaner, thanks

rodrigozhou (approved):

PR #7096, opened 2025-01-16
Refactor: arrange and document dynamic config usages in functional tests

alexshtin:
## What changed?

- Consolidated all global DCs in one file with clear name for easy discovery.
- Documented how to override DCs properly. 
- Cleaned up duplicated configs.

## Why?

Quality of life for function test writers.

## How did you test it?

Run.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.


tests/testcore/onebox.go
```diff
@@ -140,16 +140,6 @@ type (
 	HistoryConfig struct {
 		NumHistoryShards int32
 		NumHistoryHosts  int
-
-		// TODO (alex): Remove all this limits and replace them with consts in sizelimit_tests.go
```
alexshtin, line 144: Addressed.

tests/testcore/dynamic_config_overrides.go
```diff
@@ -0,0 +1,83 @@
...
+var (
+	// Functional tests don't use any dynamic config files. All settings get their default values
+	// (defined where setting is declared), besides those which are overridden.
+	// There are 4 ways to override a setting:
+	// 1. Globally using this file. Every test suite creates a new test cluster using this overrides.
+	// 2. Per test suite using SetupSuiteWithCluster() and WithDynamicConfigOverrides() option.
+	// 3. Per test using s.OverrideDynamicConfig() method.
+	// 4. Per specific cluster per test (if test has more than one default cluster) using cluster.OverrideDynamicConfig() method.
+	//
+	// NOTE: settings which are not really dynamic (requires server restart to take effect) can't be overridden on test level.
```
alexshtin, line 43: Main idea of this PR: describe how to override DC and consolidate all global overrides in one file with clear name (for easy discovery).

tests/testcore/dynamic_config_overrides.go
```diff
@@ -0,0 +1,84 @@
...
+	// 3. Per test using FunctionalTestBase.OverrideDynamicConfig() method.
+	// 4. Per specific cluster per test (if test has more than one cluster) using TestCluster.OverrideDynamicConfig() method.
+	//
+	// NOTE1: settings which are not really dynamic (requires server restart to take effect) can't be overridden on test level,
+	//        i.e., must be overridden globally (1) or per test suite (2).
+	// NOTE2: per test overrides change value for the cluster, therefore, it affects not only specific test, but
+	//        all tests for this suite. Automatic cleanup reverts previous value and tests don't affect each other.
+	//        But it means that tests in the same suite can be run in parallel.This is not a problem because testify
+	//        doesn't allow parallel execution of tests in the same suite anyway. If one day, it will be allowed,
+	//        unique namespaces with overrides per namespace should be used for tests which require different settings.
```
alexshtin, line 49: Main idea of this PR: describe how to override DC and consolidate all global overrides in one file with clear name (for easy discovery).

stephanos, line 49:
```suggestion
	// NOTE2: per test overrides change the value for the cluster, therefore, it affects not only a specific test, but
	//        all tests for that suite. The automatic cleanup reverts to the  previous value and tests don't affect each other.
	//        But that means tests in the same suite can't be run in parallel. This is not a problem because testify
	//        doesn't allow parallel execution of tests in the same suite anyway. If one day, it will be allowed,
	//        unique namespaces with overrides per namespace should be used for tests that require overrides.
```

tests/testcore/dynamic_config_overrides.go
```diff
@@ -0,0 +1,84 @@
...
+
+var (
+	// Functional tests don't use dynamic config files. All settings get their default values
+	// defined in common/dynamicconfig/constants.go.
+	// There are 4 ways to override a setting:
+	// 1. Globally using this file. Every test suite creates a new test cluster using this overrides.
+	// 2. Per test suite using FunctionalTestBase.SetupSuiteWithCluster() and WithDynamicConfigOverrides() option.
+	// 3. Per test using FunctionalTestBase.OverrideDynamicConfig() method.
+	// 4. Per specific cluster per test (if test has more than one cluster) using TestCluster.OverrideDynamicConfig() method.
+	//
```
stephanos, line 42:
```suggestion
	// Functional tests don't use dynamic config files. All settings get their default values
	// defined in common/dynamicconfig/constants.go.
	//
	// There are 4 ways to override a setting:
	// 1. Globally using this file. Every test suite creates a new test cluster using this overrides.
	// 2. Per test suite using FunctionalTestBase.SetupSuiteWithCluster() and WithDynamicConfigOverrides() option.
	// 3. Per test using FunctionalTestBase.OverrideDynamicConfig() method.
	// 4. Per specific cluster per test (if test has more than one cluster) using TestCluster.OverrideDynamicConfig() method.
	//
```

Newline for emphasis.

stephanos (approved):
❤️

tests/testcore/dynamic_config_overrides.go
alexshtin: Especially `can` -> `can't` :man_facepalming:.

tests/activity_api_pause_test.go
```diff
@@ -52,15 +51,16 @@ type ActivityApiPauseClientTestSuite struct {
 	activityRetryPolicy *temporal.RetryPolicy
 }
 
-func (s *ActivityApiPauseClientTestSuite) SetupSuite() {
-	s.FunctionalTestSdkSuite.SetupSuite()
-	s.OverrideDynamicConfig(dynamicconfig.ActivityAPIsEnabled, true)
-	s.tv = testvars.New(s.T()).WithTaskQueue(s.TaskQueue()).WithNamespaceName(s.Namespace())
+func TestActivityApiPauseClientTestSuite(t *testing.T) {
+	s := new(ActivityApiPauseClientTestSuite)
```
dnr, line 55: can you also add `t.Parallel()` to all suites where it doesn't break things?

tests/testcore/onebox.go
```diff
@@ -262,13 +226,12 @@ func newTemporal(t *testing.T, params *TemporalParams) *TemporalImpl {
 		hostsByProtocolByService:         params.HostsByProtocolByService,
 	}
 
-	for k, v := range staticOverrides {
+	for k, v := range dynamicConfigOverrides {
```
dnr, line 229: It feels like it should be possible to merge the "global test overrides" and the ones from "params" (the per-suite ones?) earlier so that this code here is just one loop. Did you consider that?

Like maybe change https://github.com/temporalio/temporal/pull/7096/files#diff-b77e5f4402c434c0f9f4a58cbe72c7fa38e280e235f5d6c393b370301355cd3eR211 to `s.testClusterConfig.DynamicConfigOverrides = maps.Clone(dynamicConfigOverrides)`?

dnr (commented):
Thank you! This is so much better already. There's more to clean up but this is a good step.

PR #6938, opened 2024-12-05
Generic hooks for testing

dnr:
## What changed?
- Add generic hook interface for fine-grained control of behavior in tests
- Use the hooks for matching varying behavior tests (force load balancer to target partitions and disable sync match)
- Use the hooks to force a race condition in an update-with-start test (by @stephanos)

## Why?
To write integration/functional tests that require tweaking behavior of code under test, without affecting non-test builds.

## Potential risks
Hooks are disabled by default, so there should be zero risk to production code, and zero overhead (assuming the Go compiler can do very basic inlining and dead code elimination).

The downside is that functional tests now have to be run with `-tags test_dep` everywhere.

PR #7097, opened 2025-01-16
Follow ReusePolicy AllowDuplicateFailedOnly with TerminateExisting

stephanos:
## What changed?

Make WorkflowIdConflictPolicy TerminateExisting follow the WorkflowIdReusePolicy after an _unsuccessful_ termination.

NOTE: The frontend change was made in https://github.com/temporalio/temporal/pull/7099

## Why?

When the termination from TerminateExisting fails, the user would expect the WorkflowIdReusePolicy to be applied as the Workflow is not running.

## How did you test it?

Well ... there's no way to test this well right now. There are no unit tests; and the functional test cannot be written without the use of [testhooks](https://github.com/temporalio/temporal/pull/6938) since there is no other way to simulate the race condition here.

I manually added a `sync.Once` into the code that terminates the workflow and can confirm the expected behavior.

### DIFF
--- a/service/history/api/startworkflow/api.go
+++ b/service/history/api/startworkflow/api.go
@@ -27,6 +27,8 @@ package startworkflow
 import (
 	"context"
 	"errors"
+	"fmt"
+	"sync"
 	"time"
 
 	commonpb "go.temporal.io/api/common/v1"
@@ -37,6 +39,7 @@ import (
 	"go.temporal.io/server/api/historyservice/v1"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/definition"
+	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/enums"
 	"go.temporal.io/server/common/locks"
 	"go.temporal.io/server/common/metrics"
@@ -52,6 +55,8 @@ import (
 	"go.temporal.io/server/service/history/workflow/cache"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
+
+var killOnce sync.Once
 
 type (
 	eagerStartDeniedReason metrics.ReasonString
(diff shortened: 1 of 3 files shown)


service/history/api/startworkflow/api.go
```diff
@@ -431,6 +438,35 @@ func (s *Starter) resolveDuplicateWorkflowID(
...
+			},
+			nil,
+			s.shardContext,
+			s.workflowConsistencyChecker,
+		)
+		if err != nil {
+			panic(err)
+		}
+	})
+
```
stephanos, line 469: Obviously, I had to remove this again, but it demonstrates that the change works.

service/history/api/startworkflow/api.go
```diff
@@ -431,6 +438,35 @@ func (s *Starter) resolveDuplicateWorkflowID(
...
+				)
+			},
+			nil,
+			s.shardContext,
+			s.workflowConsistencyChecker,
+		)
+		if err != nil {
+			panic(err)
+		}
+	})
```
stephanos, line 468: Obviously, I had to remove this again, but it demonstrates that the change works.

service/history/api/startworkflow/api.go
```diff
@@ -493,8 +529,12 @@ func (s *Starter) resolveDuplicateWorkflowID(
 		resp, err := s.generateResponse(newRunID, mutableStateInfo.workflowTask, events)
 		return resp, StartNew, err
 	case consts.ErrWorkflowCompleted:
-		// current workflow already closed
-		// fallthough to the logic for only creating the new workflow below
+		if s.followReusePolicyAfterConflictPolicyTerminate(s.namespace.ID().String()) {
+			// Exit and retry again from the top.
+			// By returning an Unavailable service error, the entire Start request will be retried.
+			return nil, StartErr, serviceerror.NewUnavailable(fmt.Sprintf("Termination failed: %v", err))
```
stephanos, line 535: I think this is the cleanest way out of here. We're using the same pattern in Signal-with-Start and MultiOperation.

service/history/api/startworkflow/api.go
stephanos: Screenshot 2025-01-17 at 10 06 02 AM
Screenshot 2025-01-17 at 10 05 55 AM

alexshtin (approved):

rodrigozhou (approved):

PR #7099, opened 2025-01-16
Deny ReusePolicy RejectDuplicate for ConflictPolicy TerminateExisting

stephanos:
## What changed?

Making the combination of WorkflowIdReusePolicy RejectDuplicate and WorkflowIdConflictPolicy TerminateExisting invalid.

## Why?

The combination is not a desirable configuration for users. It's essentially "Terminate Workflow", ie no start would ever be performed.

## How did you test it?

Added test.

## Potential risks

This _could_ break a user's behavior. Although it's unlikely. But a dynamic config was added just in case.


common/dynamicconfig/constants.go
```diff
@@ -2077,6 +2077,12 @@ the user has not specified an explicit RetryPolicy`,
 		retrypolicy.DefaultDefaultRetrySettings,
 		`DefaultWorkflowRetryPolicy represents the out-of-box retry policy for unset fields
 where the user has set an explicit RetryPolicy, but not specified all the fields`,
+	)
+	AllowReusePolicyRejectWithConflictPolicyTerminate = NewNamespaceTypedSetting(
+		"history.allowReusePolicyRejectWithConflictPolicyTerminate",
+		false,
+		`Allows to use WorkflowIdReusePolicy RejectDuplicate with WorkflowIdReusePolicy TerminateExisting.
+If false, an invalid argument error is returned.`,
```
bergundy, line 2085: nit: Might be worth having a single flag for this and changing the behavior of RejectDuplicateFailedOnly.

I'd also consider adding a warning that this flag may go away in upcoming releases, I don't think we want to keep this forever.

```suggestion
	AllowReusePolicyRejectWithConflictPolicyTerminate = NewNamespaceTypedSetting(
		"history.allowReusePolicyRejectWithConflictPolicyTerminate",
		false,
		`Allows to use WorkflowIdReusePolicy RejectDuplicate with WorkflowIdReusePolicy TerminateExisting.
If false (the default), an invalid argument error is returned.`,
```

service/frontend/errors.go
```diff
@@ -85,7 +85,8 @@ var (
...
 	errUseVersioningWithoutNormalName                     = serviceerror.NewInvalidArgument("NormalName must be set on sticky queue if UseVersioning is true.")
 	errBuildIdTooLong                                     = serviceerror.NewInvalidArgument("Build ID exceeds configured limit.workerBuildIdSize, use a shorter build ID.")
-	errIncompatibleIDReusePolicy                          = serviceerror.NewInvalidArgument("Invalid WorkflowIDReusePolicy: WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING cannot be used together with a WorkflowIDConflictPolicy.")
+	errIncompatibleIDReusePolicyTerminateIfRunning        = serviceerror.NewInvalidArgument("Invalid WorkflowIDReusePolicy: WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING cannot be used together with a WorkflowIDConflictPolicy.")
+	errIncompatibleIDReusePolicyRejectDuplicate           = serviceerror.NewInvalidArgument("Invalid WorkflowIDReusePolicy: WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE cannot be used together with WorkflowIdConflictPolicy WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING.")
```
bergundy, line 89: nit: Typically it's recommended not to use punctuation in single sentence error strings. I know we've violated this fairly consistently but we have both forms in the codebase and I'd rather go with no punctuation, which is more standard (especially for Go codebases).

bergundy (approved):

stephanos:
I made these changes:

- removed period at end of error string
- expanded dynamic config name and description to cover the `RejectDuplicateFailedOnly` case, too (implementation follows in next PR)

common/dynamicconfig/constants.go
```diff
@@ -2077,6 +2077,13 @@ the user has not specified an explicit RetryPolicy`,
 		retrypolicy.DefaultDefaultRetrySettings,
 		`DefaultWorkflowRetryPolicy represents the out-of-box retry policy for unset fields
 where the user has set an explicit RetryPolicy, but not specified all the fields`,
+	)
+	EnableReusePolicyRejectOnConflictPolicyTerminate = NewNamespaceTypedSetting(
```
rodrigozhou, line 2081: Rename to `AllowReusePolicyRejectWithConflictPolicyTerminate`.

service/frontend/workflow_handler.go
```diff
@@ -4777,12 +4782,18 @@ func (wh *WorkflowHandler) validateVersionRuleBuildId(request *workflowservice.U
...
 	conflictPolicy enumspb.WorkflowIdConflictPolicy,
 ) error {
 	if conflictPolicy != enumspb.WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED &&
 		reusePolicy == enumspb.WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING {
-		return errIncompatibleIDReusePolicy
+		return errIncompatibleIDReusePolicyTerminateIfRunning
+	}
+	if conflictPolicy == enumspb.WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING &&
+		reusePolicy == enumspb.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE &&
+		wh.enableReusePolicyRejectOnConflictPolicyTerminate(namespaceName.String()) {
```
rodrigozhou, line 4795: This should be `!wh.allow...`, right?

rodrigozhou (changes requested):

common/dynamicconfig/constants.go
stephanos: Well, by using a single flag, this is becoming awkward in terms of wording it.

We want to _deny_ the use of the `RejectDuplicate` together with `TerminateExisting`, but _honor_ the `RejectDuplicateIfFailed` policy.

"Allow" doesn't quite work here because while will "allow" using `RejectDuplicateIfFailed` with `TerminateExisting` - we won't "allow"`RejectDuplicate`.

Open to suggestions.

stephanos: Maybe "Follow..."

rodrigozhou: My comment was just to be the same as your dynamic config name.

stephanos: Ah, that wasn't clear.

PR #7101, opened 2025-01-17
Handle shard ownership lost when reading history tasks

yycptt:
## What changed?

- Wrap ExecutionManager.GetHistoryTasks in shard context to handler shardOwnershipLost errors.
- Will need to follow up and audit all usage of this method and other methods on execution manager. This PR is only for fixing the issue we are seeing.

## Why?

- If shard has no (api) traffic, and ownership already got lost in the background (from some persistence implementations). Task processing can get stuck forever trying to load tasks and see this shard ownership lost error.

## How did you test it?

- Unit test

### DIFF
--- a/service/history/shard/context_impl.go
+++ b/service/history/shard/context_impl.go
@@ -574,6 +574,18 @@ func (s *ContextImpl) AddSpeculativeWorkflowTaskTimeoutTask(
 	return nil
 }
 
+func (s *ContextImpl) GetHistoryTasks(
+	ctx context.Context,
+	request *persistence.GetHistoryTasksRequest,
+) (*persistence.GetHistoryTasksResponse, error) {
+	if err := s.errorByState(); err != nil {
+		return nil, err
+	}
+
+	resp, err := s.executionManager.GetHistoryTasks(ctx, request)
+	return resp, s.handleReadError(err)
+}
+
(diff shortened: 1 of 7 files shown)


alfred-landrum (approved):

service/history/shard/context_impl.go
```diff
@@ -574,6 +574,18 @@ func (s *ContextImpl) AddSpeculativeWorkflowTaskTimeoutTask(
 	return nil
 }
 
+func (s *ContextImpl) GetHistoryTasks(
+	ctx context.Context,
+	request *persistence.GetHistoryTasksRequest,
+) (*persistence.GetHistoryTasksResponse, error) {
+	if err := s.errorByState(); err != nil {
```
ychebotarev, line 581: in I understand correctly main change - you "wrap" s.executionManager.GetHistoryTask with

s.errorByState()
and
s.handleReadError()
?

pdoerner (approved):

ychebotarev (approved):

service/history/shard/context_impl.go
yycptt: Yes. s.handleReadError is the key part which will check for shardOwnership lost and will put shard context in an invalid state. Then a new shard will be created once shard controller realizes the shard context is invalid.

yiminc (approved):

PR #7102, opened 2025-01-17
Refactor: remove cluster.yaml config

alexshtin:
## What changed?

Refactor: remove `cluster.yaml` config.

## Why?

It is the same as `es_cluster.yaml` but doesn't have `esConfig` section. The intent was to use this config file when Elasticsearch is not used. But this section is not used if SQL persistence is used, so there is no harm to always use `es_cluster.yaml`.

I am going to significantly decrease number of config files for functional tests (if not remove them all). This is small step in this direction.

## How did you test it?

Run tests.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/tests/advanced_visibility_test.go
+++ b/tests/advanced_visibility_test.go
@@ -71,16 +71,15 @@ import (
 const (
 	numOfRetry   = 50
 	waitTimeInMs = 400
+
+	testSearchAttributeKey = "CustomTextField"
+	testSearchAttributeVal = "test value"
 )
 
 type AdvancedVisibilitySuite struct {
 	testcore.FunctionalTestSuite
 
-	isElasticsearchEnabled bool
-
-	testSearchAttributeKey string
-	testSearchAttributeVal string
-	sdkClient              sdkclient.Client
+	sdkClient sdkclient.Client
 	// client for the system namespace
 	sysSDKClient sdkclient.Client
 }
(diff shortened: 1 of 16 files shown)


dnr:
I think we should remove all these yaml files and just write Go structs in code. Would it be easier if we just jumped to that directly?

alexshtin:
I would prefer to do it incrementally.

alexshtin:
Mainly because many of those config setting are no relevant at all. I will remove all unneeded ones first, and then convert all the rest to the code.

tests/advanced_visibility_test.go
```diff
@@ -103,15 +102,9 @@ func (s *AdvancedVisibilitySuite) SetupSuite() {
 		// Allow the scavenger to remove any build ID regardless of when it was last default for a set.
 		dynamicconfig.RemovableBuildIdDurationSinceDefault.Key(): time.Microsecond,
 	}
+	s.FunctionalTestSuite.SetupSuiteWithDefaultCluster(testcore.WithDynamicConfigOverrides(dynamicConfigOverrides))
```
dnr, line 105: can't you write this as
```suggestion
	s.SetupSuiteWithDefaultCluster(testcore.WithDynamicConfigOverrides(dynamicConfigOverrides))
```
?

tests/advanced_visibility_test.go
```diff
@@ -103,15 +102,9 @@ func (s *AdvancedVisibilitySuite) SetupSuite() {
...
 
-	if testcore.UsingSQLAdvancedVisibility() {
-		s.FunctionalTestSuite.SetupSuiteWithCluster("testdata/cluster.yaml", testcore.WithDynamicConfigOverrides(dynamicConfigOverrides))
-		s.Logger.Info(fmt.Sprintf("Running advanced visibility test with %s/%s persistence", testcore.TestFlags.PersistenceType, testcore.TestFlags.PersistenceDriver))
-		s.isElasticsearchEnabled = false
-	} else {
-		s.FunctionalTestSuite.SetupSuiteWithCluster("testdata/es_cluster.yaml", testcore.WithDynamicConfigOverrides(dynamicConfigOverrides))
-		s.Logger.Info("Running advanced visibility test with Elasticsearch persistence")
-		s.isElasticsearchEnabled = true
+	if !testcore.UseSQLVisibility() {
```
dnr, line 107: It feels a little weird that suites are calling the global function `testcore.UseSQLVisibility`... I would expect/prefer something more like:

There's one function in the code that examines flags/env vars to get persistence + vis config. TestCluster or maybe whatever calls TestCluster calls that function, then saves the results in TestCluster. The base suite has methods that read the persistence + vis type from there.

(Doesn't have to be this PR)

dnr (approved):

tests/advanced_visibility_test.go
alexshtin: This is a great idea and I was thinking about something similar. For now I am removing unnecessary indirections.

tests/advanced_visibility_test.go
alexshtin: It meant to be `FunctionalTestSuiteBase`. I am explicitly indicate struct which this setup method belongs too.

PR #7103, opened 2025-01-17
OTEL in task processing

stephanos:
## What changed?

Added support for OpenTelemetry (OTEL) to task processing.

## Why?

To extend the OTEL coverage and gain insight into task processing.

## How did you test it?

Ran Grafana Tempo locally:

Screenshot 2025-01-16 at 6 07 51 PM

## Potential risks

I don't expect any negative performance impact on task processing because most operations are noops.

### DIFF
--- a/service/history/queues/executable.go
+++ b/service/history/queues/executable.go
@@ -36,6 +36,8 @@ import (
 	"sync"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/common"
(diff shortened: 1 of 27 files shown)


service/history/queues/executable.go
```diff
@@ -277,6 +283,22 @@ func (e *executableImpl) Execute() (retErr error) {
...
+			attribute.Key(telemetry.WorkflowIDKey).String(e.GetWorkflowID()),
+			attribute.Key(telemetry.WorkflowRunIDKey).String(e.GetRunID()),
+			attribute.Key("task-type").String(e.GetType().String()),
+			attribute.Key("task-id").Int64(e.GetTaskID())))
+	defer func() {
+		if retErr != nil {
+			span.RecordError(retErr)
+		}
+		span.End()
+	}()
```
stephanos, line 300: This is the main change.

yycptt: I wonder how many extra memory allocation we will have with this even when it's a noopTracer. The allocation for those attributes will happen anyway? e.g. for loggers we have a lazy logger to avoid creating logger tags when it's not needed.
May need to monitor the cpu usage after the change.

yycptt (approved):

service/history/queues/executable.go
stephanos: I was wondering if you would pick up on that, too :) It might be best to wrap the entire thing in an `if` check given how critical this code path is. I'll do that 👍

alexshtin: I would add a comment why `if` check is needed.

alexshtin (commented):
:+1:

PR #7107, opened 2025-01-17
RegisterTaskQueue + DescribeVersion: versioning-3.1

Shivs11:
## What changed?

- Updated deployment entity wf's API's (naming) for versioning-3.1. Removed those which will not be required.
- Added functionality for RegisterTaskQueue +  DescribeVersion
- Added functional tests for verifying the above two work. 
- UserData stuff has been removed right now since there are pending discussions

## Why?

- Versioning-3.1

## How did you test it?

- Functional tests have been added. Note, the tests added are exact replicas (in terms of their core functionality) with the existing tests under `deployment_test.go`

## Potential risks

- None, feature branch.

## Is hotfix candidate?

- No


service/frontend/workflow_handler.go
```diff
@@ -3306,6 +3314,42 @@ func (wh *WorkflowHandler) SetCurrentDeployment(ctx context.Context, request *wo
 	}, nil
 }
 
+// Versioning-3 Public-Preview API's
+
+func (wh *WorkflowHandler) DescribeWorkerDeploymentVersion(ctx context.Context, request *workflowservice.DescribeWorkerDeploymentVersionRequest) (_ *workflowservice.DescribeWorkerDeploymentVersionResponse, retError error) {
```
Shivs11, line 3319: functionality copied from existing working API for deployments: `DescribeDeployment`

service/matching/physical_task_queue_manager.go
```diff
@@ -342,8 +343,17 @@ func (c *physicalTaskQueueManagerImpl) PollTask(
...
+	if c.partitionMgr.engine.config.EnableDeployments(namespaceEntry.Name().String()) {
+		if err = c.ensureRegisteredInDeployment(ctx, namespaceEntry, pollMetadata); err != nil {
+			return nil, err
+		}
+	}
+
+	if c.partitionMgr.engine.config.EnableDeploymentVersions(namespaceEntry.Name().String()) {
+		if err = c.ensureRegisteredInDeploymentVersion(ctx, namespaceEntry, pollMetadata); err != nil {
+			return nil, err
+		}
```
Shivs11, line 356: note - the additional `if` cases added here (`if c.partitionMgr.engine.config.EnableDeploymentVersions...`) serves as a way to let pre-release customers still be able to poll without having to face errors. Point being, `ensureRegisteredInDeploymentXXX` will throw an error when the corresponding dynamic config is not set (which is the right behaviour). Since we don't want our public-preview customers to face with a dynamic config error of not having `EnableDeployments` set, and also don't want our pre-release customers to face with a dynamic config error of not having `EnableDeploymentVersions` set, I added these extra if statements.

They will be removed when we remove pre-release versioning-related code.

service/matching/physical_task_queue_manager.go
```diff
@@ -619,6 +630,87 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeployment(
...
+		// deployment already registered
+		return nil
+	}
+
+	if c.deploymentRegisterError != nil {
+		// deployment not possible due to registration limits
+		return c.deploymentRegisterError
+	}
+
+	// userData, _, err := c.partitionMgr.GetUserDataManager().GetUserData()
```
Shivs11, line 661: UserData stuff will come in later since there are pending discussions on how to finalize this structure.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -328,7 +328,7 @@ message SyncDeploymentUserDataRequest {
     temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
 
     // This is the deployment being modified.
-    temporal.api.deployment.v1.Deployment deployment = 4;
+    temporal.api.deployment.v1.Deployment deployment = 4;
```
ShahabT, line 331: Nit: extra space

service/matching/physical_task_queue_manager.go
ShahabT: The logic before was to reject pollers with deployment info if deployments are not enabled in the system. And that was intentional. But I think this is also fine.

service/worker/workerdeployment/fx.go
```diff
@@ -93,7 +93,7 @@ func NewResult(
 	return fxResult{
 		Component: &workerComponent{
 			activityDeps: params,
-			enabledForNs: dynamicconfig.EnableDeployments.Get(dc),
+			enabledForNs: dynamicconfig.EnableDeploymentVersions.Get(dc),
```
ShahabT, line 96: as discussed before, this should be enabled for all NS, otherwise we'd need to take extra steps to make the per-ns workers refresh themselves after enabling the feature for a pre-release customer (same for old component)

ShahabT (commented):
overall looks good.

service/worker/workerdeployment/fx.go
Shivs11: hmm, good point + sorry this slipped my mind - I do wonder now if there is a strict need for running this by a per-namespace worker? Can we just not run it using a normal worker, which shall run for all namespaces, on the default worker task-queue instead of the per-ns-worker-task-queue?

ShahabT: It should still be in per-ns worker like schedules and batches. Just we don't need to disable/enable per ns.

service/matching/physical_task_queue_manager.go
```diff
@@ -619,6 +630,87 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeployment(
...
+	if workerDeployment == nil {
+		return nil
+	}
+	if !c.partitionMgr.engine.config.EnableDeploymentVersions(namespaceEntry.Name().String()) {
+		return errMissingDeploymentVersion
+	}
+
+	// lock so that only one poll does the update and the rest wait for it
+	c.deploymentLock.Lock()
+	defer c.deploymentLock.Unlock()
```
carlydf, line 649: is this code just copied for now? are you thinking to use the same lock to cover updates to both the old and the new deployment entity WFs, or two separate locks? (same lock might be the right choice, I'm not sure, just want to discuss)

service/worker/pernamespaceworker.go
```diff
@@ -398,7 +398,6 @@ func (w *perNamespaceWorker) refresh(args refreshArgs) (retErr error) {
 			componentSet += fmt.Sprintf("%p,", cmp)
 		}
 	}
-
```
carlydf, line 401: nit: add this line back / restore the file so this file is not in the diff

service/worker/workerdeployment/fx.go
Shivs11: Resolved offline - I learnt that we should still be running these on the per-ns workers they run stuff in the user's namespace and we want to keep this functionality for these internal wfs

service/worker/workerdeployment/client.go
```diff
@@ -187,23 +169,18 @@ func (d *ClientImpl) RegisterTaskQueueWorker(
 func (d *ClientImpl) DescribeVersion(
 	ctx context.Context,
 	namespaceEntry *namespace.Namespace,
-	seriesName string,
-	buildID string,
-) (_ *deploymentpb.DeploymentInfo, retErr error) {
+	version string,
```
carlydf, line 172: For now lets assume that version is globally unique, and we can always add the deploymentName parameter later if that changes

service/worker/workerdeployment/client.go
```diff
@@ -790,22 +683,28 @@ func (d *ClientImpl) record(operation string, retErr *error, args ...any) func()
...
 
 //nolint:staticcheck
-func stateToInfo(state *deploymentspb.VersionLocalState) *deploymentpb.DeploymentInfo {
+func stateToInfo(state *deploymentspb.VersionLocalState) *deploymentpb.WorkerDeploymentVersionInfo {
 	if state == nil {
 		return nil
 	}
 
-	taskQueues := make([]*deploymentpb.DeploymentInfo_TaskQueueInfo, 0, len(state.TaskQueueFamilies)*2)
+	taskQueues := make([]*deploymentpb.WorkerDeploymentVersionInfo_WorkerBuildTaskQueueInfo, 0, len(state.TaskQueueFamilies)*2)
```
carlydf, line 691: `WorkerBuildTaskQueueInfo` seems out of date now that we've settled on Worker Version -- does the api repo need to be fixed?

service/worker/workerdeployment/util.go
```diff
@@ -67,9 +67,9 @@ const (
 	WorkerDeploymentWorkflowIDPrefix             = "temporal-sys-worker-deployment"
 	WorkerDeploymentVersionWorkflowIDDelimeter   = ":"
 	WorkerDeploymentVersionWorkflowIDEscape      = "|"
-	WorkerDeploymentVersionWorkflowIDInitialSize = (2 * len(WorkerDeploymentVersionWorkflowIDDelimeter)) + len(WorkerDeploymentVersionWorkflowIDPrefix)
+	WorkerDeploymentVersionWorkflowIDInitialSize = len(WorkerDeploymentVersionWorkflowIDDelimeter) + len(WorkerDeploymentVersionWorkflowIDPrefix) // todo (Shivam): Do we need 2 * len(WorkerDeploymentVersionWorkflowIDDelimeter)?
```
carlydf, line 70: right now we have only one delimiter in the generated id, so I think just leave it at 1 for now. I think if we add another field / delimiter in the future then we can add the 2 * then

carlydf (approved):
approved once comments are addressed

service/worker/workerdeployment/util.go
Shivs11: sorry, I think that's an outdated comment which you addressed - yes, you are right btw

service/matching/physical_task_queue_manager.go
Shivs11: good question - I had thought about that and I ended up choosing to keep the locks the same since these locks just make sure that only one update happens at the same time - moreover, both the entity wf updates happen sequentially so I don't see the use-case of using two separate locks instead of just one (naming of the locks will change later though)

service/worker/workerdeployment/client.go
Shivs11: oh, good catch - yes, I think we do need to change this in the api repo. I am holding this merge until I do that. Watch out for that PR coming up shortly

common/dynamicconfig/constants.go
```diff
@@ -861,11 +861,11 @@ of Timeout and if no activity is seen even after that the connection is closed.`
 	EnableDeployments = NewNamespaceBoolSetting(
 		"system.enableDeployments",
 		false,
-		`EnableDeployments enables deployments (versioning v3) in all services,
+		`EnableDeployments enables deployments (deprecated versioning v3 pre-release) in all services,
 including deployment-related RPCs in the frontend, deployment entity workflows in the worker,
 and deployment interaction in matching and history.`,
 	)
-	EnableDeploymentVersions = NewNamespaceBoolSetting(
+	EnableDeploymentVersions = NewGlobalBoolSetting(
```
ShahabT, line 868: We should still use a per-ns dynamic config. In all other places (frontend/matching) the enabling is still per NS because we want to only allow private preview customers. The worker component does not need to use the per-ns config and should always return enabled though.

service/worker/workerdeployment/fx.go
```diff
@@ -93,14 +93,14 @@ func NewResult(
...
-			enabledForNs: dynamicconfig.EnableDeploymentVersions.Get(dc),
+			enabled:      dynamicconfig.EnableDeploymentVersions.Get(dc),
 		},
 	}
 }
 
 func (s *workerComponent) DedicatedWorkerOptions(ns *namespace.Namespace) *workercommon.PerNSDedicatedWorkerOptions {
 	return &workercommon.PerNSDedicatedWorkerOptions{
-		Enabled: s.enabledForNs(ns.Name().String()),
+		Enabled: s.enabled(),
```
ShahabT, line 103:
```suggestion
		Enabled: true,
```

service/worker/workerdeployment/fx.go
```diff
@@ -41,7 +41,7 @@ import (
 type (
 	workerComponent struct {
 		activityDeps activityDeps
-		enabledForNs dynamicconfig.BoolPropertyFnWithNamespaceFilter
+		enabled      dynamicconfig.BoolPropertyFn
```
ShahabT, line 44: no need for this field.

common/dynamicconfig/constants.go
carlydf: ohh I misunderstood this my bad

PR #7109, opened 2025-01-17
Move tasktoken serializer to package

stephanos:
## What changed?

Moved `TaskTokenSerializer` to the already existing package `common/tasktoken`.

## Why?

- the serializer's dependency on `utf8validator` prevents a refactoring I plan to do; moving it out of `common` helps
- serializer belongs in that package
- allows to drop the "TaskToken" part

## How did you test it?

Pure refactoring; no behavior change.

## Potential risks

This will break (at least) two references in other repos I've identified. I'll update those once this is merged.


common/task_token_serializer_interfaces.go
```diff
@@ -1,41 +0,0 @@
...
-type (
-	// TaskTokenSerializer serializes task tokens
-	TaskTokenSerializer interface {
-		Serialize(token *tokenspb.Task) ([]byte, error)
-		Deserialize(data []byte) (*tokenspb.Task, error)
-		SerializeQueryTaskToken(token *tokenspb.QueryTask) ([]byte, error)
-		DeserializeQueryTaskToken(data []byte) (*tokenspb.QueryTask, error)
-		SerializeNexusTaskToken(token *tokenspb.NexusTask) ([]byte, error)
-		DeserializeNexusTaskToken(data []byte) (*tokenspb.NexusTask, error)
-	}
```
stephanos, line 40: This interface is superfluous. No mocking is done anywhere. And prob won't ever be needed.

common/tasktoken/serializer.go
```diff
@@ -22,23 +22,21 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package common
+package tasktoken
```
ychebotarev, line 25: I though you just add it...
but looks like we already have tasktoken package.
In this case yes, serializer should be in this package.

ychebotarev (approved):

stephanos:
Note that I'll wait with merging until today's release is out to avoid compile error in other repos (which I'll address).

alexshtin (approved):

PR #7112, opened 2025-01-17
Move namespace replication to own package

stephanos:
## What changed?

Moved code from `common/namespace` to `common/namespace/nsreplication`. Next to the existing `common/namespace/nsregistry`.

## Why?

The goal is to remove all references from `common/namespace` to `common/persistence`. One step towards using `common/dynamicconfig` from `common/namespace`. 

## How did you test it?

No behavior changes here.


common/namespace/nsdlq/dlq_message_handler.go
alexshtin, line 1: This DLQ is just for namespace replication queue. I believe it can go under `nsreplication` package.

alexshtin (approved):

PR #7113, opened 2025-01-17
Fix flakey TestWorkflowNexusCallbacks_CarriedOver/WorkflowFailureRetry

pdoerner:
## What changed?

Added an Eventually condition to getting callback info to account for delays in the workflow retry in `TestWorkflowNexusCallbacks_CarriedOver/WorkflowFailureRetry`

## Why?

Fix flake

## How did you test it?

Ran a couple hundred times locally. Failed within 50 runs before change.

### DIFF
--- a/tests/callbacks_test.go
+++ b/tests/callbacks_test.go
@@ -46,7 +46,7 @@ import (
 	"go.temporal.io/sdk/worker"
 	"go.temporal.io/sdk/workflow"
 	"go.temporal.io/server/common/dynamicconfig"
-	"go.temporal.io/server/common/testing/protorequire"
+	"go.temporal.io/server/common/testing/protoassert"
 	"go.temporal.io/server/components/callbacks"
 	"go.temporal.io/server/internal/freeport"
 	"go.temporal.io/server/tests/testcore"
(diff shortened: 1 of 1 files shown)


stephanos (approved):

tests/callbacks_test.go
```diff
@@ -312,22 +314,24 @@ func (s *CallbacksSuite) TestWorkflowNexusCallbacks_CarriedOver() {
...
-					s.Nil(callbackInfo.LastAttemptFailure)
-				}
+				s.EventuallyWithT(func(col *assert.CollectT) {
+					description, err := sdkClient.DescribeWorkflowExecution(ctx, request.WorkflowId, "")
+					require.NoError(col, err)
+					require.Equal(col, 1, len(description.Callbacks))
+					callbackInfo := description.Callbacks[0]
+					protorequire.ProtoEqual(col, request.CompletionCallbacks[0], callbackInfo.Callback)
+					protorequire.ProtoEqual(col, &workflowpb.CallbackInfo_Trigger{Variant: &workflowpb.CallbackInfo_Trigger_WorkflowClosed{WorkflowClosed: &workflowpb.CallbackInfo_WorkflowClosed{}}}, callbackInfo.Trigger)
+					require.Equal(col, int32(attempt), callbackInfo.Attempt)
```
bergundy, line 324: This will still fail the test, you should change to assert and eager return.

```suggestion
					assert.Equal(col, int32(attempt), callbackInfo.Attempt)
					// Return to avoid evaluate the require statements below.
					if callbackInfo.Attempt != int32(attempt) {
						return
					}
```

bergundy (commented):
Generally the fix makes sense to me, please see the inline comment.

PR #7118, opened 2025-01-18
Exit test after timeout

stephanos:
## What changed?

Added per-test timeouts.

Unfortunately there's no way to stop a particular Goroutine from the outside. That's why `os.Exit` - the nuclear option - is really the only choice IMO.

## Why?

Go only allows to set a timeout on the entire test run. But sometimes a single test is stuck and never going to complete. But you still have to wait for the very long timeout.

With this, at least it will _fail fast_.

## Potential risks

It could prematurely abort slower (non-stuck) tests - and require more tweaking of the timeouts for these. That could be annoying.


tests/testcore/timeout.go
```diff
@@ -0,0 +1,56 @@
...
+				if time.Now().After(detonateAt) {
+					// Cannot use t.Fatalf since it will wait until the test completes.
+					t.Logf("test took more than %v to complete, exiting now", after)
+
+					// Print stack trace to help debugging.
+					rawStack := append(debug.Stack(), '\n', '\n')
+					_, _, err := stack.ScanSnapshot(bytes.NewReader(rawStack), os.Stdout, stack.DefaultOpts())
+					if err != nil {
+						t.Logf("failed to parse stack trace: %v", err)
+					}
```
stephanos, line 38: This output is a bit more readable than Go's default output.

.github/workflows/run-tests.yml
```diff
@@ -258,13 +258,21 @@ jobs:
...
+          make unit-test-coverage
         env:
           UNIT_TEST_DIRS: ${{ inputs.unit_test_directory }}
           TEST_ARGS: ${{ needs.set-up-single-test.outputs.single_test_args }}
           TEST_TIMEOUT: ${{ needs.set-up-single-test.outputs.test_timeout }}
 
+      - name: Highlight test timeout
+        if: ${{ steps.test_run.outcome == 'failure' && steps.test_run.outputs.exit_code == 124 }}
+        run: |
+          echo "::error::Test Test timed out"
```
stephanos, line 274: Example:
TODO

tests/testcore/functional_test_base.go
```diff
@@ -244,12 +245,21 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(clusterConfigFile string, opt
...
 	s.initAssertions()
 }
 
+func (s *FunctionalTestBase) TeardownTest() {
+	s.testTimeout.cancel()
+}
+
+func (s *FunctionalTestBase) SetTestTimeout(d time.Duration) {
+	s.testTimeout.set(d)
+}
```
stephanos, line 261: A test that needs more time can extend the timeout by calling this function.

(side effect: we can easily see which tests are very slow)

tests/testcore/functional_test_base.go
```diff
@@ -244,10 +247,21 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(clusterConfigFile string, opt
 func (s *FunctionalTestBase) SetupTest() {
 	s.checkTestShard()
 	s.initAssertions()
+	if s.testTimeout == nil {
+		s.testTimeout = newTestTimeout(s.T(), 30*time.Second*debug.TimeoutMultiplier)
```
stephanos, line 251: Can't be in `SetupSuite` as the SDK functional test suite overrides that.

tests/testcore/functional_test_base.go
```diff
@@ -244,10 +246,21 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(clusterConfigFile string, opt
 func (s *FunctionalTestBase) SetupTest() {
 	s.checkTestShard()
 	s.initAssertions()
+	if s.testTimeout == nil {
+		s.testTimeout = newTestTimeout(s.T(), 30*time.Second)
+	} else {
+		s.testTimeout.reset(s.T())
+	}
```
stephanos, line 253: Can't be in SetupSuite as the SDK functional test suite overrides that.

PR #7123, opened 2025-01-20
Ss/update old deployment component

Shivs11:
## What changed?

- Was decided to now enable the worker component to register itself always when existing dynamic config related to deployments, soon to be deprecated, was enabled on a namespace

## Why?

- Pre-release versioning-3

## How did you test it?

- Existing suite of tests


go.mod
```diff
@@ -154,3 +154,5 @@ require (
 	modernc.org/strutil v1.2.0 // indirect
 	modernc.org/token v1.1.0 // indirect
 )
+
+replace go.temporal.io/api => github.com/temporalio/api-go v1.43.2-0.20250117224308-f070b00264f6
```
carlydf, line 158: I think you need to take out this replace since this PR is going into main. I'll review once that is pushed and the diff is more readable!

Shivs11:
I think I messed up with my branch sync since the diff shouldn't be this long - moreover, the change is super tiny for me to further resolve all conflicts and still have the branch history look this bad - a better idea imo is branching off the current main, making the changes and opening a fresh new PR - shall be doing that! Sorry :(

PR #7127, opened 2025-01-21
Describe worker deployment in versioning 3.1

Shivs11:
## What changed?

- DescribeWorkerDeployment API for versioning-3.1

For this API to work, I realized that the worker deployment workflow has to have information about the various deployment versions it encapsulates. Now that version deployment workflows won't have a `deploymentName` present in their workflowID (for now), partial matching workflow ID's with the received deployment name to target version workflows was not feasible.

Another option was to store the mapping from "versions" to "the local state of a version workflow". However, this would have  made the deployment workflow arguments large in size since in theory, we could have many deployment versions each with large sized local states.

The option I ended up going with was storing the version in a deployment workflow during task-queue registration. During a `DescribeWorkerDeployment` call, we would call `Describe` on each stored version and return the aggregated info to the user. A version will be removed from a worker deployment when it becomes "scavenged" (to be implemented)

Potential future improvements include adding a caching mechanism.

## Why?

- Versioning-3.1

## How did you test it?

- Added a functional test + more tests to follow

## Potential risks

- None, going inside a feature.

## Is hotfix candidate?

- No


go.mod
```diff
@@ -155,4 +155,5 @@ require (
 	modernc.org/token v1.1.0 // indirect
 )
 
-replace go.temporal.io/api => github.com/temporalio/api-go v1.43.2-0.20250117224308-f070b00264f6
+// replace go.temporal.io/api => github.com/temporalio/api-go v1.43.2-0.20250117224308-f070b00264f6
+replace go.temporal.io/api => ../api-go
```
Shivs11, line 159: will be altering this to point to the latest commit in the versioning-3.1 branch before merging

tests/worker_deployment_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+		a.NotNil(resp.GetWorkerDeploymentInfo().GetVersionSummaries()[0].GetCreateTime())
+		a.NotNil(resp.GetWorkerDeploymentInfo().GetVersionSummaries()[1].GetCreateTime())
+
+		a.NotNil(resp.GetWorkerDeploymentInfo().GetCreateTime())
+	}, time.Second*10, time.Millisecond*1000)
+}
+
+// Name is used by testvars. We use a shorten test name in variables so that physical task queue IDs
+// do not grow larger that DB column limit (currently as low as 272 chars).
```
carlydf, line 138:
```suggestion
// Name is used by testvars. We use a shortened test name in variables so that physical task queue IDs
// do not grow larger than DB column limit (currently as low as 272 chars).
```

service/worker/workerdeployment/version_workflow.go
```diff
@@ -218,6 +218,17 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
 	// 	}
```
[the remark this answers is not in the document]
Shivs11: Will be coming in the very next PR, which depends on this PR for getting in

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -79,6 +79,9 @@ message WorkerDeploymentWorkflowArgs {
 message WorkerDeploymentLocalState {
     string current_version = 1;
     google.protobuf.Timestamp current_changed_time = 2;
+    // Versions that have been registered in this deployment.
```
carlydf, line 82: Sorry to be so nit-picky about this comment, but I was curious about whether items will ever be removed from this list, so I consulted the notion doc which says this:

```
  // Versions that are currently tracked in this Worker Deployment. A Version will be
  // cleaned up automatically if all the following conditions meet:
	//  - It does not receive new executions (see WorkerDeploymentVersionInfo.accepts_new_executions)
	//  - It has no active pollers (see WorkerDeploymentVersionInfo.pollers_status)
	//  - It is drained (see WorkerDeploymentVersionInfo.drainage_status)
```

If the goal is to eventually add this cleaning up logic, but it's not implemented yet, perhaps a TODO comment saying that we eventually intend to clean up versions from this list on the above conditions

Shivs11: yes, good point - sorry, you shouldn't have had to consulting the notion document since thats what comments are meant to be for - moreover, I like the comment you have pasted and will use it in this message for better clarity.

carlydf (approved):

PR #7131, opened 2025-01-21
Unify OTEL resource names

stephanos:
## What changed?

Moved all telemetry resource names to one place.

## Why?

Having them in one place.

### DIFF
--- a/service/history/queues/executable.go
+++ b/service/history/queues/executable.go
@@ -283,6 +283,7 @@ func (e *executableImpl) Execute() (retErr error) {
 	)
 	e.Unlock()
 
+	// Wrapped in if block to avoid unnecessary allocations when OTEL is disables.
 	if telemetry.IsEnabled(e.tracer) {
 		var span trace.Span
 		ctx, span = e.tracer.Start(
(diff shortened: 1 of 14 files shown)


service/history/queues/executable.go
```diff
@@ -292,8 +293,8 @@ func (e *executableImpl) Execute() (retErr error) {
 			trace.WithAttributes(
 				attribute.Key(telemetry.WorkflowIDKey).String(e.GetWorkflowID()),
 				attribute.Key(telemetry.WorkflowRunIDKey).String(e.GetRunID()),
-				attribute.Key("task-type").String(e.GetType().String()),
-				attribute.Key("task-id").Int64(e.GetTaskID())))
+				attribute.Key("task.type").String(e.GetType().String()),
+				attribute.Key("task.id").Int64(e.GetTaskID())))
```
stephanos, line 297: On second glance, most attributes are `.` separated, so might just do that here, too.

service/history/workflow/update/util.go
```diff
@@ -32,11 +32,10 @@ import (
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/telemetry"
 	"google.golang.org/protobuf/proto"
 )
 
-const libraryName = "go.temporal.io/service/history/workflow/update"
```
stephanos, line 38: Safe to remove as no spans were ever emitted here.

alexshtin (approved):

PR #7134, opened 2025-01-22
Forward priority to matching

stephanos:
## What changed?

Stores priorities from commands in MS, extracts it in task processing when sending to matching.

- [x] StartWorkflowExecutionRequest
- [x] StartChildWorkflowExecutionInitiatedEventAttributes
- [x] ActivityTaskScheduledEventAttributes
- [x] WorkflowExecutionStartedEventAttributes
- [x] PollActivityTaskQueueResponse
- [x] SignalWithStartWorkflowExecutionRequest
- [x] RecordActivityTaskStartedResponse
- [x] ScheduleActivityTaskCommandAttributes
- [x] DescribeWorkflowExecution
- [x] NewWorkflowExecutionInfo
- [x] GetPendingActivityInfo

NOTE: haven't added any frontend validation yet.

## How did you test it?

Wrote unit test for merging priorities.

Tested a few scenarios like Start Workflow and Activity by hand and checked the priority makes it to Matching.

Didn't write any functional tests yet, I suspect this will be covered by future priority functional tests.


service/history/priority_util.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return &commonpb.Priority{
+		PriorityKey:       priorityKey,
+		FairnessKey:       fairnessKey,
+		FairnessWeight:    fairnessWeight,
+		FairnessRateLimit: fairnessRateLimit,
+		OrderingKey:       orderingKey,
+	}
+}
```
stephanos, line 70: Didn't quite know where else to put this.

service/history/transfer_queue_active_task_executor.go
```diff
@@ -896,6 +900,7 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
 		inheritedBuildId,
 		initiatedEvent.GetUserMetadata(),
 		mutableState.GetExecutionInfo().GetVersioningInfo().GetVersioningOverride(),
+		mergePriorities(mutableState.GetExecutionInfo().Priority, attributes.Priority),
```
stephanos, line 903: I'm not sure about this one, I didn't see any mention of this case in the new API docs.

It seems reasonable to me for the child to inherit the parent's priority, but still be able to override it.

service/frontend/workflow_handler.go
```diff
@@ -3066,6 +3066,7 @@ func (wh *WorkflowHandler) CreateSchedule(ctx context.Context, request *workflow
 		WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_FAIL,
 		Memo:                     request.Memo,
 		SearchAttributes:         sa,
+		Priority:                 &commonpb.Priority{}, // ie default priority
```
stephanos, line 3069: Added a default priority where there's no custom one to make it easy to find them all.

common/priorities/priority_util.go
```diff
@@ -0,0 +1,73 @@
...
+	commonpb "go.temporal.io/api/common/v1"
+)
+
+func Merge(
+	base *commonpb.Priority,
+	override *commonpb.Priority,
+) *commonpb.Priority {
+	if base == nil || override == nil {
+		return cmp.Or(base, override)
+	}
```
stephanos, line 39: Could also return a non-nil, default priority object if both are nil.

service/history/transfer_queue_active_task_executor.go
```diff
@@ -896,6 +901,7 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
 		inheritedBuildId,
 		initiatedEvent.GetUserMetadata(),
 		mutableState.GetExecutionInfo().GetVersioningInfo().GetVersioningOverride(),
+		priorities.Merge(mutableState.GetExecutionInfo().Priority, attributes.Priority),
```
stephanos, line 904: I'm not sure about this one, I didn't see any mention of this case in the new API docs.

It seems reasonable to me for the child to inherit the parent's priority, but still be able to override it.

common/priorities/priority_util.go
```diff
@@ -0,0 +1,73 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package priorities
```
dnr, line 25: I feel like the singular `priority` would be a more natural package name? But then it would conflict with local variables named `priority`. Could be `priority_util` though that's not great either.

common/priorities/priority_util.go
```diff
@@ -0,0 +1,73 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
dnr, line 5: this is brand new, I think we can go with 2025 Temporal only

common/priorities/priority_util.go
dnr: I'd rather just use nils everywhere where possible, many or most wfs/activities/tasks will have no priority metadata and there's no need to allocate useless empty values everywhere

common/priorities/priority_util.go
```diff
@@ -0,0 +1,73 @@
...
+	if override.OrderingKey != 0 {
+		orderingKey = override.OrderingKey
+	}
+
+	return &commonpb.Priority{
+		PriorityKey:       priorityKey,
+		FairnessKey:       fairnessKey,
+		FairnessWeight:    fairnessWeight,
+		FairnessRateLimit: fairnessRateLimit,
+		OrderingKey:       orderingKey,
```
dnr, line 71:
```suggestion
	return &commonpb.Priority{
		PriorityKey:       cmp.Or(override.PriorityKey, base.PriorityKey),
		FairnessKey:       cmp.Or(override.FairnessKey, base.FairnessKey),
		FairnessWeight:    cmp.Or(override.FairnessWeight, base.FairnessWeight),
		FairnessRateLimit: cmp.Or(override.FairnessRateLimit, base.FairnessRateLimit),
		OrderingKey:       cmp.Or(override.OrderingKey, base.OrderingKey),
```

common/priorities/priority_util.go
```diff
@@ -0,0 +1,73 @@
...
+	orderingKey := base.OrderingKey
+	if override.OrderingKey != 0 {
+		orderingKey = override.OrderingKey
+	}
+
+	return &commonpb.Priority{
+		PriorityKey:       priorityKey,
+		FairnessKey:       fairnessKey,
+		FairnessWeight:    fairnessWeight,
+		FairnessRateLimit: fairnessRateLimit,
```
dnr, line 70: actually, I realized that the code that I have so far doesn't support a rate limit passed through the tasks, so we should remove this until we can. I'll remove it from api first though (and maybe the other fields if we're not doing them yet)

service/frontend/workflow_handler.go
dnr: I guess I'd rather leave it nil? but either way is fine here

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -145,6 +146,7 @@ func recordActivityTaskStarted(
 	response := &historyservice.RecordActivityTaskStartedResponse{
 		ScheduledEvent:              scheduledEvent,
 		CurrentAttemptScheduledTime: ai.ScheduledTime,
+		Priority:                    priorities.Merge(mutableState.GetExecutionInfo().Priority, ai.Priority),
```
dnr, line 149: This is interesting. So, there's technically a possibility that this may not be the same as the priority used to dispatch the task, since that one is calculated at schedule time and this is at start time. (Suppose the workflow was mutated between schedule and start.)

This priority is just for downstream uses, so I think actually this is better, the mutation takes effect immediately. The alternative would be storing the "scheduled priority" in activity info (recalculated on each retry), but I don't really see the point in doing that.

service/history/transfer_queue_active_task_executor.go
dnr: I thought I mentioned it somewhere.. yes, definitely makes sense for child to inherit parent's priority

service/matching/forwarder.go
```diff
@@ -154,6 +154,7 @@ func (fwdr *Forwarder) ForwardTask(ctx context.Context, task *internalTask) erro
 				ScheduleToStartTimeout: expirationDuration,
 				ForwardInfo:            fwdr.getForwardInfo(task),
 				VersionDirective:       task.event.Data.GetVersionDirective(),
+				Priority:               task.event.Data.Priority,
```
dnr, line 157: I'd use GetPriority() here just to be sure. I forgot if there were any circumstances where `task.event.Data` can be nil but the above lines do it

service/worker/deployment/deployment_client.go
```diff
@@ -468,6 +468,7 @@ func (d *DeploymentClientImpl) StartDeploymentSeries(
 		WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
 		SearchAttributes:         d.buildSearchAttributes(),
 		Memo:                     memo,
+		Priority:                 &commonpb.Priority{},
```
dnr, line 471: I'd leave these out, I don't think there's any value

service/worker/workerdeployment/client.go
```diff
@@ -467,6 +467,7 @@ func (d *ClientImpl) Start(
 		WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
 		SearchAttributes:         d.buildSearchAttributes(),
 		Memo:                     memo,
+		Priority:                 &commonpb.Priority{}, // ie default priority
```
dnr, line 470: same here

common/priorities/priority_util.go
stephanos: 👍  I just ran `make copyright`

common/priorities/priority_util.go
stephanos: I had that at first and was getting linter errors for my `priority` variables.

common/priorities/priority_util.go
stephanos: 👍 Ah, clever; I forgot it works on zero values, too.

service/frontend/workflow_handler.go
stephanos: I like specifying it because it avoids the ambiguity of "is priority been considered here - or is it missing?"

common/priorities/priority_util.go
stephanos: I took the hint from other common packages which are plural, too. (e.g. `locks`, `enums`, `quotas`)

service/history/api/recordactivitytaskstarted/api.go
stephanos: Right; I took your suggestion from our convo as "bind the priority as late as possible".

common/priorities/priority_util.go
```diff
@@ -0,0 +1,48 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2025 Uber Technologies, Inc.
```
dnr, line 5: 2025 Temporal and remove this line

common/priorities/priority_util.go
dnr: That's fair, I'm fine with the name

dnr (approved):

PR #7135, opened 2025-01-22
OTEL noop vars

stephanos:
## What changed?

Introduces vars for the noop tracer and trace provider.

## Why?

Communicates clearer what/where they are.


alexshtin (approved):
Readability is much better now!

PR #7138, opened 2025-01-22
Embed Unimplemented*ServiceServer instead of Unsafe*ServiceServer

alexshtin:
## What changed?

Embed `Unimplemented*ServiceServer` instead of `Unsafe*ServiceServer` in all gRPC handlers.

## Why?

Embedding `Unimplemented*ServiceServer` is the right way to keep backward compatibility with future interface changes.

## How did you test it?

Build and run.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/service/matching/handler.go
+++ b/service/matching/handler.go
@@ -52,7 +52,7 @@ import (
 type (
 	// Handler - gRPC handler interface for matchingservice
 	Handler struct {
-		matchingservice.UnsafeMatchingServiceServer
+		matchingservice.UnimplementedMatchingServiceServer
 
 		engine            Engine
 		config            *Config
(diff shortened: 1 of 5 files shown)


tdeebswihart (approved):

service/matching/handler.go
```diff
@@ -52,7 +52,7 @@ import (
 type (
 	// Handler - gRPC handler interface for matchingservice
 	Handler struct {
-		matchingservice.UnsafeMatchingServiceServer
+		matchingservice.UnimplementedMatchingServiceServer
```
ychebotarev, line 55: can we have some sort of test to be sure this is not deleted in the future?

ychebotarev (approved):

service/matching/handler.go
alexshtin: This can be done using reflection. I don't see much value in it though.

PR #7139, opened 2025-01-22
fix flaky xdc test TestCronWorkflowCompleteAndFailover

hai719:
## What changed?
Wait for clusters to be synced instead of using Sleep. Make sure second run is started before failover.

## Why?
Current implementation depends on the timing which is not reliable.

## How did you test it?
Repeatedly run the test locally and no failure found.

### DIFF
--- a/tests/xdc/failover_test.go
+++ b/tests/xdc/failover_test.go
@@ -844,7 +844,43 @@ func (s *FunctionalClustersTestSuite) TestTerminateFailover() {
 	s.logger.Info("PollAndProcessWorkflowTask", tag.Error(err))
 	s.NoError(err)
 
+	// check terminate done
+	getHistoryReq := &workflowservice.GetWorkflowExecutionHistoryRequest{
+		Namespace: namespace,
+		Execution: &commonpb.WorkflowExecution{
+			WorkflowId: id,
+		},
+	}
+
+	s.WaitForHistory(`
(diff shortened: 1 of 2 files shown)


ychebotarev (approved):

tests/xdc/failover_test.go
```diff
@@ -1896,6 +1925,54 @@ func (s *FunctionalClustersTestSuite) TestCronWorkflowStartAndFailover() {
...
+	namespace string,
+	execution *commonpb.WorkflowExecution,
+) string {
+	var newRunID string
+	s.Eventually(func() bool {
+		newRunID = s.getNewExecutionRunIdFromLastEvent(client, namespace, execution)
+		return newRunID != ""
+	}, 10*time.Second, 100*time.Millisecond)
+
+	s.NotEmpty(newRunID, "New run should have started")
```
alexshtin, line 1972: This will never fail.

tests/xdc/failover_test.go
```diff
@@ -1896,6 +1925,54 @@ func (s *FunctionalClustersTestSuite) TestCronWorkflowStartAndFailover() {
...
+	return newRunID
+}
+
```
alexshtin, line 1975: There is `s.WaitForHistoryEventsSuffix` instead all of these 3 funcs.

PR #7144, opened 2025-01-23
Report timed out tests

stephanos:
## What changed?

Extended the testrunner to scan the test output for timed out tests and report these (1) in the logs and (2) in the JUnit report.

## Why?

To track timed out tests better.

## How did you test it?

I set the test timeout to 5s. This is what you see, for example:
https://github.com/temporalio/temporal/actions/runs/12922374527/job/36038028569?pr=7144#step:8:52993

image

The developer will still have to connect the the test/suite name to the code location, but that shouldn't be too hard.

Also verified that the JUnit report is being written correctly.


tools/testrunner/testrunner.go
```diff
@@ -69,22 +71,109 @@ type attempt struct {
...
+
+		if len(timedoutTests) == 0 {
+			return exitErr
+		}
 
+		// write JUnit XML outpout for timed out tests since none will have been generated
+		var suites junit.Testsuites
+		for _, name := range timedoutTests {
+			suites.Suites = append(suites.Suites, junit.Testsuite{
+				Name: name,
```
stephanos, line 125: This isn't quite correct, but I don't think the suite name matters?

tools/testrunner/testdata/timeout-output.log
```diff
@@ -0,0 +1,30 @@
...
+		/temporal/common/persistence/cassandra/test.go:109
+		/temporal/common/persistence/persistence-tests/persistence_test_base.go:214
+		/temporal/tests/testcore/test_cluster.go:251
+		/temporal/tests/testcore/test_cluster.go:148
+		/temporal/tests/testcore/functional_test_base.go:224
+		/temporal/tests/testcore/functional_test_sdk_suite.go:84
+		/golang/1.23.2/packages/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:157
+		/temporal/tests/activity_api_reset_test.go:58
+		/golang/1.23.2/packages/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.4.darwin-arm64/src/testing/testing.go:1690
+
```
stephanos, line 30: ^ this is what's printed in the CI

tools/testrunner/testdata/junit-timeout-output.xml
```diff
@@ -0,0 +1,17 @@
...
+        <testcase name="TestNDCFuncTestSuite" classname="">
+            <failure message="timeout"></failure>
+        </testcase>
+    </testsuite>
+    <testsuite name="TestActivityApiStateReplicationSuite" tests="0" failures="0" errors="0" id="0" time="">
+        <testcase name="TestActivityApiStateReplicationSuite" classname="">
+            <failure message="timeout"></failure>
+        </testcase>
+    </testsuite>
+</testsuites>
```
stephanos, line 17: ^ this is the generated JUnit report

tools/testrunner/testdata/junit-timeout-output.xml
```diff
@@ -0,0 +1,13 @@
+<testsuites>
+    <testsuite name="suite" tests="0" failures="0" errors="0" id="0" time="">
```
stephanos, line 2: I don't think the testsuite name matters?

tools/testrunner/testrunner.go
```diff
@@ -69,30 +70,45 @@ type attempt struct {
...
-func (a *attempt) run(ctx context.Context, args []string) error {
+func (a *attempt) run(ctx context.Context, args []string) (error, string) {
 	args = append([]string{"--junitfile", a.junitXmlPath}, args...)
 	log.Printf("starting test attempt %d with args: %v", a.number, args)
 	cmd := exec.CommandContext(ctx, a.runner.gotestsumExecutable, args...)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
+	var output strings.Builder
+	cmd.Stdout = io.MultiWriter(os.Stdout, &output)
+	cmd.Stderr = io.MultiWriter(os.Stderr, &output) // we only need one output
```
stephanos, line 79: Needs to buffer everything in memory, which seems fine. A stream-based approach would be more efficient but overkill.

tools/testrunner/testrunner.go
```diff
@@ -296,22 +160,34 @@ func Main() {
...
+				// update JUnit XML outpout for timed out tests since none will have been generated
+				a.junitReport.updateWithTimedoutTests(timedoutTests)
+				log.Print(stacktrace)
+			}
+			// Exit the retry loop.
 			break
 		}
-		failures := r.attempts[retry].failures()
+
+		failures := a.junitReport.failures()
```
stephanos, line 186: Would this ever not be just `a`?

tools/testrunner/testrunner.go
```diff
@@ -24,10 +24,9 @@ package testrunner
```
stephanos, line 24: All the previous tests are still there, they were just moved around as needed.

tools/testrunner/testdata/junit-timeout-output.xml
ychebotarev: I actually not sure. doesn't matter for code that extracts job_id. But ->may<- matter for code that collects those reports and calculate statistics? You may ask Den (or just leave it as is and see if something is broken).

ychebotarev (approved):
LGFM, some minor refactoring

tools/testrunner/testdata/junit-timeout-output.xml
stephanos: @dandavison do you know if the testsuite name is relevant for your tool?

tools/testrunner/log.go
```diff
@@ -0,0 +1,90 @@
...
+				if strings.HasPrefix(line, "Test") {
+					timedoutTests = append(timedoutTests, strings.Split(line, " ")[0])
+				}
+				if line == "" {
+					break
+				}
+			}
+		} else if len(timedoutTests) > 0 {
+			// collect stracktrace
+			stacktrace += line + "\n"
```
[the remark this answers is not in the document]
stephanos: yup!

tools/testrunner/log.go
```diff
@@ -0,0 +1,90 @@
...
+		len(timedoutTests), strings.Join(timedoutTests, "\n\t"), testOnlyStacktrace(stacktrace))
+	return
+}
+
+// testOnlyStacktrace removes all but the test stacktraces from the full stacktrace.
+func testOnlyStacktrace(stacktrace string) string {
+	var res string
+	snap, _, err := stack.ScanSnapshot(strings.NewReader(stacktrace), io.Discard, stack.DefaultOpts())
+	if err != nil && err != io.EOF {
+		res = fmt.Sprintf("failed to parse stacktrace: %v", err)
```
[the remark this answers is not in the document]
stephanos: good point!

tools/testrunner/junit.go
```diff
@@ -0,0 +1,204 @@
...
+		combined.Failures += report.Testsuites.Failures
+		combined.Skipped += report.Testsuites.Skipped
+		combined.Disabled += report.Testsuites.Disabled
+		combined.Time += report.Testsuites.Time
+
+		if i == 0 {
+			combined.XMLName = report.Testsuites.XMLName
+			combined.Name = report.Testsuites.Name
+			combined.Suites = report.Testsuites.Suites
+			continue
```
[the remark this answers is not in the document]
stephanos: I only moved this code around, I didn't touch it. But I see your point. I've simplified it a little.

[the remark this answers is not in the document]
stephanos: Actually, I'm not sure I changed much there. IMO it's good enough for an internal testing tool.

tools/testrunner/testrunner.go
```diff
@@ -38,116 +37,27 @@ import (
...
-	suites       junit.Testsuites
+	runner      *runner
+	number      int
+	exitErr     *exec.ExitError
+	junitReport *junitReport
 }
 
-func (a *attempt) run(ctx context.Context, args []string) error {
-	args = append([]string{"--junitfile", a.junitXmlPath}, args...)
+func (a *attempt) run(ctx context.Context, args []string) (error, string) {
```
bergundy, line 49: This is very non-standard to return the error first.

tools/testrunner/testrunner.go
```diff
@@ -295,23 +167,41 @@ func Main() {
 		log.Fatalf("failed to parse command line options: %v", err)
 	}
 
+	var currentAttempt *attempt
```
bergundy, line 170: nit: could you instead extract the last attempt from the runner?

tools/testrunner/junit.go
```diff
@@ -0,0 +1,204 @@
...
+
+type junitReport struct {
+	path string
+	junit.Testsuites
+}
+
+func (j *junitReport) read() {
+	f, err := os.Open(j.path)
+	if err != nil {
+		log.Fatalf("failed to open junit report file: %v", err)
```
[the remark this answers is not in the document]
bergundy: I think it's fine to use in scripts, but I'd return an error from the helper functions and panic only at the top level. That makes more reusable code.

tools/testrunner/junit.go
```diff
@@ -0,0 +1,207 @@
...
+
+	"github.com/jstemmer/go-junit-report/v2/junit"
+)
+
+type junitReport struct {
+	path string
+	junit.Testsuites
+}
+
+func (j *junitReport) read() {
```
bergundy, line 41: Wondering if there's a better option to initialize the junitReport struct with an already open file instead of having the struct be mutable.

tools/testrunner/junit.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	defer f.Close()
+
+	decoder := xml.NewDecoder(f)
+	if err = decoder.Decode(&j.Testsuites); err != nil {
+		log.Fatalf("failed to read junit report file: %v", err)
+	}
+}
+
+func (j *junitReport) generateForTimedoutTests(timedoutTests []string) {
```
bergundy, line 54: I would make this a constructor and keep the report immutable.

tools/testrunner/testrunner.go
stephanos: Good point

tools/testrunner/log.go
```diff
@@ -0,0 +1,90 @@
...
+	} else if snap != nil {
```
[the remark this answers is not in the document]
stephanos: It's possible to _not_ receive an error from the library, but still get a `nil` snap back when the input is "a little off". I've had to add a newline somewhere to prevent that. Very finicky.

tools/testrunner/junit.go
stephanos: We could probably improve it slightly, but I think we're hitting diminishing returns here now (given its just an internal tool).

tools/testrunner/testrunner.go
```diff
@@ -38,116 +37,27 @@ import (
...
-	args = append([]string{"--junitfile", a.junitXmlPath}, args...)
+func (a *attempt) run(ctx context.Context, args []string) (error, string) {
+	args = append([]string{"--junitfile", a.junitReport.path}, args...)
 	log.Printf("starting test attempt %d with args: %v", a.number, args)
 	cmd := exec.CommandContext(ctx, a.runner.gotestsumExecutable, args...)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
+	var output strings.Builder
+	cmd.Stdout = io.MultiWriter(os.Stdout, &output)
+	cmd.Stderr = io.MultiWriter(os.Stderr, &output) // we only need one output
```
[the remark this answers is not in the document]
stephanos: Testing locally, it seems to just use stdout. So I removed the stderr part.

tools/testrunner/testrunner.go
stephanos: I don't quite follow. The `currentAttempt` is assigned to the result of `r.newAttempt()`

bergundy: My point is that if you have a method to get the last attempt from the runner, you can get rid of this mutable variable which takes away from the code's readability.
If `currentAttempt` is always `r.attempts[len(r.attempts)-1]`, there's not much value in doing this booking.

bergundy (approved):

PR #7149, opened 2025-01-23
flake + unit tests fix

Shivs11:
## What changed?

- Fix flaky `DescribeWorkerDeployment`. The flake was caused because I forgot to pass in the right parameters when creating version entity workflows. Moreover, recent proto changes meant that my assertions also had to be updated and have done so :) 
- Other breaking unit tests in the versioning-3.1 repo

## Why?

- don't like seeing red markers after CI is done running 

## How did you test it?

- Ran unit and functional tests

## Potential risks

- None, going to a feature.


workflowservicemock/v1/service_grpc.pb.mock.go
Shivs11, line 1: This file was by mistake created by me (still not sure how) when I merged this [PR](https://github.com/temporalio/temporal/pull/7107/files#diff-aca5cdd96011e9e1a409782cb07b5bc91fa3153f895633bab276aa1d24ace657) in. Double checked `main` to realize this file is not present and it's presence is not required.

Sorry about that!

carlydf (approved):

PR #7152, opened 2025-01-23
[Scheduled Actions V2] Invoker logic

lina-temporal:
## What changed?
The Invoker (start workflow) logic is added for HSM scheduler.

A few differences between workflow scheduler and HSM scheduler execution logic:
- A `ProcessBuffer` task is 1:1 with a call to `ProcessBuffer`; draining is not attempted more than once on the same task. 
- Any time processing indicates that a delay must be taken (rate limiter, exponential backoff..), another `ProcessBuffer` task is scheduled to wake after the delay. `ProcessBuffer` is used to drive retries.
  - In the event that some external event must occur for execution to continue (such as a running workflow closing), the Executor is transitioned to waiting/idle, and no further action will be taken until another state machine wakes it back up.
- Each `BufferedStart` has its own `Attempt` and `BackoffTime` to manage retries. The `Execute` task is therefore idempotent. 
  -  HSM Scheduler can therefore continue to accept API calls while a BufferedStart is retrying, and each individual BufferedStart can make retry progress without having to hold a workflow lock.
- `startWorkflow`, `terminateWorkflow`, and `cancelWorkflow` are no longer called within local activities. Instead, they are called with a short (5 sec) deadline within the `Execute` task.
- A single `Execute` task is generated whenever there are BufferedStarts, CancelWorkflows, or TerminateWorkflows available for execution. An `Execute` task will only ever make a single attempt on each work item before rescheduling a `ProcessBuffer` task.
  - `Execute` task completion will always schedule a `ProcessBuffer` task after completing work (unless the Execute task was a no-op). This is to avoid a race condition between reading the Invoker's current state, and choosing a transition (because applying a transition necessitates mutating the state field, which I'd rather just leave as-is). 

Still missing in HSM Invoker:
- Workflow Watcher support. This is something we don't plan to touch until the CHASM port over.
  - This implies `LastCompletionResult` also isn't wired up
- Rate Limiter support 

## How did you test it?
* New tests
* go test -v && make lint

## Potential risks
* HSM scheduler isn't presently deployed


components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+		fx.In
+
+		Config         *Config
+		MetricsHandler metrics.Handler
+		BaseLogger     log.Logger
+		HistoryClient  resource.HistoryClient
+		FrontendClient workflowservice.WorkflowServiceClient
+	}
+
+	executorTaskExecutor struct {
```
lina-temporal, line 64: I'm not sure about how everyone else feels, but this name makes my eyes bleed. At the same time, I didn't want to be inconsistent with other HSM naming. Maybe the path of least resistance is to just use a few synonyms for this? Like, `starterTaskExecutor` (and rename the state machine to `Starter`.. or something like that). Really open to suggestions. Ditto for renaming the tasks. In general I'm not satisfied with the naming.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+			return nil, temporal.NewApplicationError(
+				fmt.Sprintf("Rate limited while attempting to start %s", workflowID),
+				errTypeRateLimited,
+				rateLimitedDetails{Delay: delay},
+			)
+		}
+	}
+
+	// TODO - set last completion result/continued failure
+	// TODO - set search attributes
```
lina-temporal, line 355: I don't think there's anything blocking me from passing the search attributes through to `StartWorkflowExecution`, I just hadn't closed off the TODO. Will follow up.

bergundy:
> Rate Limiter support. The existing rate limiter acts per-namespace (given it is part of per-NS worker), and I'm not sure of the right way to wire that into an HSM executor (which is scoped wider than a single namespace).

You can keep a rate limiter in the executor. If the rate limiter is out of tokens, you'll have to back off either by failing the task, or committing any pending progress and adding a persistent timer task. It's dependent on how long you need to wait for the next token. Do not wait on the rate limiter in the shared executor.

components/scheduler/executor.go
```diff
@@ -114,9 +114,7 @@ var TransitionExecute = hsm.NewTransition(
 	},
 	enumsspb.SCHEDULER_EXECUTOR_STATE_EXECUTING,
 	func(e Executor, event EventExecute) (hsm.TransitionOutput, error) {
-		// We want Executor to immediately wake and attempt to buffer when new starts
-		// are added.
-		e.NextInvocationTime = nil
+		e.NextInvocationTime = timestamppb.New(event.Deadline)
```
bergundy, line 117: Shouldn't this always be the time of the first buffered start (chronologically)? I'm a bit rusty on the details so please correct me if I'm wrong. I'm just worried that this will invalidate a pending task if the event's buffered starts are further in the future.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
bergundy, line 5: nit:

```suggestion
// Copyright (c) 2025 Temporal Technologies Inc.  All rights reserved.
```

components/scheduler/executor_executors.go
bergundy: It's not just a starter though, right? Will this also e.g. terminate workflows? You can call it an invoker maybe?

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+		FrontendClient workflowservice.WorkflowServiceClient
+	}
+
+	executorTaskExecutor struct {
+		ExecutorTaskExecutorOptions
+	}
+
+	rateLimitedDetails struct {
+		// The requested interval to delay processing by rescheduilng.
+		Delay time.Duration
```
bergundy, line 70: Any reason to export this field?

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/resource"
+	"go.temporal.io/server/common/util"
+	"go.temporal.io/server/service/history/hsm"
+	scheduler1 "go.temporal.io/server/service/worker/scheduler"
+	"go.uber.org/fx"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type (
```
bergundy, line 53: Please stop grouping types, it's not a good practice. It generally messes with the generated docs.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+const (
+	startWorkflowMinDeadline = 5 * time.Second // Lower bound for the deadline in which buffered actions are dropped.
+	startWorkflowMaxDeadline = 1 * time.Hour   // Upper bound for the deadline in which buffered actions are dropped.
+
+	// Because the catchup window doesn't apply to a manual start, pick a custom
+	// execution deadline before timing out a start.
+	manualStartExecutionDeadline = startWorkflowMaxDeadline
+
+	// Upper bound on how many times starting an individual buffered action should be retried.
+	ExecutorMaxStartAttempts = 10
```
bergundy, line 83: Is there a good reason to export this?

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Make sure we have something to start. If not, we can clear the buffer and
+	// transition to the waiting state.
+	executionInfo := scheduler.Schedule.Action.GetStartWorkflow()
+	if executionInfo == nil || len(executor.BufferedStarts) == 0 {
+		return hsm.MachineTransition(node, func(e Executor) (hsm.TransitionOutput, error) {
+			e.ExecutorInternal.BufferedStarts = nil
```
bergundy, line 115: nit: I typically like to keep all of the mutations as part of the well defined transition methods to have all of the logic in a central location.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+		})
+	}
+
+	// Drain the Executor's BufferedStarts. drainBuffer will update Executor's
+	// BufferedStarts, as well as Scheduler metadata.
+	e.drainBuffer(logger, env, executor, scheduler, executionInfo)
+
+	// Update SchedulerInfo metadata.
+	err = hsm.MachineTransition(schedulerNode, func(s Scheduler) (hsm.TransitionOutput, error) {
+		s.Info = scheduler.Info
```
bergundy, line 128: Why is it important to cache this field?

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+			newBuffer = append(newBuffer, start)
+			continue
+		}
+
+		if env.Now().After(e.startWorkflowDeadline(scheduler, start)) {
+			// Drop expired starts.
+			continue
+		}
+
+		result, err := e.startWorkflow(env, scheduler, start, request)
```
bergundy, line 215: You shouldn't do any IO while holding the lock, you should do this in an immediate task, not a timer task (since the HSM framework can only execute timer tasks under a lock).

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	errTypeRetryLimitExceeded = "RetryLimitExceeded"
+	errTypeRateLimited        = "RateLimited"
+	errTypeAlreadyStarted     = "serviceerror.WorkflowExecutionAlreadyStarted"
+)
+
+func RegisterExecutorExecutors(registry *hsm.Registry, options ExecutorTaskExecutorOptions) error {
+	e := executorTaskExecutor{
+		ExecutorTaskExecutorOptions: options,
+	}
+	return hsm.RegisterTimerExecutor(registry, e.executeExecuteTask)
```
bergundy, line 94: Why is this a timer task?

components/scheduler/executor_executors.go
bergundy: Note that since you're (wrongly) executing all of this under a lock it's safe but this should really not be done that way. You are potentially overwriting this field with a stale value, this can happen if other transitions mutate this field while the lock is released.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+				Node:     node,
+				Deadline: retryAt,
+			})
+		})
+	}
+
+	// No more buffered starts, or remaining starts are waiting for a workflow to be
+	// closed. We can transition to waiting.
+	return hsm.MachineTransition(node, func(e Executor) (hsm.TransitionOutput, error) {
+		e.ExecutorInternal.BufferedStarts = executor.BufferedStarts
```
bergundy, line 166: Same note as above, you may lose buffered starts if more starts are buffered while you release the lock.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	executor Executor,
+	scheduler Scheduler,
+	request *workflowpb.NewWorkflowExecutionInfo,
+) {
+	metricsWithTag := e.MetricsHandler.WithTags(
+		metrics.StringTag(metrics.ScheduleActionTypeTag, metrics.ScheduleActionStartWorkflow))
+	isRunning := len(scheduler.Info.RunningWorkflows) > 0
+
+	// Resolve overlap policies and prepare next workflows to start.
+	action := scheduler1.ProcessBuffer(executor.BufferedStarts, isRunning, scheduler.resolveOverlapPolicy)
```
bergundy, line 187: Does this actually drain the entire buffer? I only see one action being returned here. (I don't have deep familiarity with the code so will just defer to you to confirm).

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+			Namespace:           scheduler.Namespace,
+			WorkflowExecution:   &commonpb.WorkflowExecution{WorkflowId: target.WorkflowId},
+			Reason:              "terminated by schedule overlap policy",
+			Identity:            scheduler.identity(),
+			FirstExecutionRunId: target.RunId,
+		},
+	}
+	ctx, cancelFunc := context.WithTimeout(context.Background(), e.Config.ServiceCallTimeout())
+	defer cancelFunc()
+	_, err := e.HistoryClient.TerminateWorkflowExecution(ctx, request)
```
bergundy, line 407: Wondering why you start the workflow with the frontend client and terminate it with the history client. Why not be consistent here?

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+}
+
+func (e executorTaskExecutor) loadExecutor(node *hsm.Node) (Executor, error) {
+	prevExecutor, err := hsm.MachineData[Executor](node)
+	if err != nil {
+		return Executor{}, err
+	}
+
+	return Executor{
+		ExecutorInternal: prevExecutor.ExecutorInternal,
```
bergundy, line 453: Note, you're not cloning here, when you change the code to release the lock, you'll want to clone.
I probably wouldn't return the whole structs in these `load` functions, instead take only what is relevant for the executor functionality.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	// Because the catchup window doesn't apply to a manual start, pick a custom
+	// execution deadline before timing out a start.
+	manualStartExecutionDeadline = startWorkflowMaxDeadline
+
+	// Upper bound on how many times starting an individual buffered action should be retried.
+	ExecutorMaxStartAttempts = 10
+
+	errTypeRetryLimitExceeded = "RetryLimitExceeded"
+	errTypeRateLimited        = "RateLimited"
+	errTypeAlreadyStarted     = "serviceerror.WorkflowExecutionAlreadyStarted"
```
bergundy, line 87: Ideally we'd use the standard errors package instead of performing string comparisons.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	return Executor{
+		ExecutorInternal: prevExecutor.ExecutorInternal,
+	}, nil
+}
+
+// translateError converts a dependent service error into an application error.
+// Errors are classified between retryable and non-retryable.
+func translateError(err error, msgPrefix string) error {
```
bergundy, line 459: Not sure I understand why you need this translation here. Why can't you use the original error? What's the value in wrapping with ApplicationError? That's an SDK concept and isn't relevant here.

components/scheduler/executor_executors.go
lina-temporal: True, it also terminates/cancels as needed. `Invoker` works for me!

components/scheduler/executor.go
lina-temporal: > Shouldn't this always be the time of the first buffered start (chronologically)?

Yes, it should; that's what `Deadline` gets set to. I suppose we could get rid of `Deadline` and just compute it within the transition, though, I'll make that change. 

> I'm just worried that this will invalidate a pending task if the event's buffered starts are further in the future.

Is only one pending timer task allowed? If so, yes, that could happen presently.

components/scheduler/executor_executors.go
lina-temporal: I'll follow up in a separate PR with this since it's from `make generate`.

components/scheduler/executor_executors.go
lina-temporal: Nope, will fix.

components/scheduler/executor_executors.go
lina-temporal: I reference it within the tests, which are in a separate `scheduler_test` package.

components/scheduler/executor_executors.go
lina-temporal: Sure, I'll update these.

components/scheduler/executor_executors.go
lina-temporal: > Does this actually drain the entire buffer? I only see one action being returned here.

It depends on the overlap policy; if something like `OVERLAP_ALL` is specified, it'll drain the buffer. It drains as much of the buffer as possible; I guess I could rename the calling function from `drainBuffer` to `processBuffer` as well.

components/scheduler/executor_executors.go
lina-temporal: This is what the existing workflow scheduler does; I believe that `StartWorkflowExecution` goes through frontend to make sure metering is applied, which wouldn't matter for `terminateWorkflow`/`cancelWorkflow`. Though, I'm not sure the reason to have the other two not go through frontend - @dnr ?

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+		NamespaceId: scheduler.NamespaceId,
+		TerminateRequest: &workflowservice.TerminateWorkflowExecutionRequest{
+			Namespace:           scheduler.Namespace,
+			WorkflowExecution:   &commonpb.WorkflowExecution{WorkflowId: target.WorkflowId},
+			Reason:              "terminated by schedule overlap policy",
+			Identity:            scheduler.identity(),
+			FirstExecutionRunId: target.RunId,
+		},
+	}
+	ctx, cancelFunc := context.WithTimeout(context.Background(), e.Config.ServiceCallTimeout())
```
dnr, line 405: You should set callerinfo on this context (and all other client contexts) to get proper prioritization and stuff. Search for `headers.SetCallerInfo` for examples. The workflow impl gets this from pernsworkermanager

components/scheduler/executor_executors.go
dnr: Yes, it goes through frontend for metering, and also namespace rps limit for safety.

Terminate/cancel could also, but I had already written them to go to history and it didn't seem worth changing at the time. I don't know if terminate/cancel are metered.. if they are, and if there's not a special case to exclude calls from schedules, that would be a visible metering change, so please check with someone who knows that stuff if you want to do that.

components/scheduler/executor_executors.go
dnr: :+1: temporal.ApplicationError is an sdk thing, don't use it (or any error wrapping) here

components/scheduler/executor_executors_test.go
```diff
@@ -0,0 +1,488 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package scheduler_test
```
dnr, line 25: This looks great.

Have you looked at the old workflow_test.go that much? Most of the detailed behaviors of the old impl are actually tested there; the functional tests are much more limited. Obviously it would be a bunch of work to port over, and a bunch is not applicable, but I think it would be pretty valuable to preserve most of those cases.

I'm thinking ones like `TestUpdateBetweenNominalAndJitter`, `TestUpdateNotRetroactive`, `TestBackfillInclusiveStartEnd`, and maybe even `TestHugeBackfillBuffer` though that would probably need more modification.

At least, maybe use it as inspiration for future tests :)

components/scheduler/executor_executors_test.go
```diff
@@ -0,0 +1,488 @@
...
+		}, nil)
+
+	// Run an Execute task.
+	err = e.registry.ExecuteTimerTask(e.env, e.executorNode, scheduler.ExecuteTask{})
+	require.NoError(e.T(), err)
+
+	// Buffer should be empty, scheduler metadata should be updated.
+	schedulerSm, err := hsm.MachineData[scheduler.Scheduler](e.schedulerNode)
+	require.NoError(e.T(), err)
+	require.Equal(e.T(), 0, len(executor.BufferedStarts))
```
dnr, line 142:
```suggestion
	require.Empty(e.T(), executor.BufferedStarts)
```
reads nicer

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+	for _, start := range allStarts {
+		// Ensure we can take more actions. Manual actions are always allowed.
+		if !start.Manual && !scheduler.useScheduledAction(true) {
+			// Drop buffered automated actions while paused.
+			continue
+		}
+
+		if env.Now().Before(start.BackoffTime.AsTime()) {
+			// BufferedStart is still backing off, push it back into the buffer.
+			newBuffer = append(newBuffer, start)
```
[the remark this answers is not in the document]
lina-temporal: Yes, that's right, multiple stops can be backing off concurrently. Right now, there's no cap on that. Following your other comment re: allowing retry until catchup window (which I am on-board for), we might want at least a metric on tasks concurrently backing off, since that's a metric I'd want to look at if we had schedules reaching their max buffer limits (still left at 1000 for now).

In terms of behavior, every time the Invoker/Executor wakes, it'll retry any failed tasks that have passed their backoff time that are eligible, and Invoker will create another delayed task for the next-earliest backoff if it wasn't able to retry all backing-off starts.

components/scheduler/config.go
```diff
@@ -35,44 +36,66 @@ type (
...
+		`The maximum backoff interval when retrying a failed workflow start.`,
+	)
 
-	ExecutionTimeout = dynamicconfig.NewNamespaceDurationSetting(
-		"component.scheduler.executionTimeout",
-		time.Second*10,
-		`ExecutionTimeout is the timeout for executing a single scheduler task.`,
+	ServiceCallTimeout = dynamicconfig.NewGlobalDurationSetting(
+		"component.scheduler.serviceCallTimeout",
+		5*time.Second,
```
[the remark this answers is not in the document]
lina-temporal: Sure, we can keep it at 10s. If anything, HSM scheduler should be less sensitive to long-tail latencies. Will fix.

components/scheduler/executor_executors_test.go
lina-temporal: Yes, good idea! I'll port over the tests you called out and give workflow_test.go another pass to see what else might make sense.

components/scheduler/executor_executors.go
```diff
@@ -0,0 +1,493 @@
...
+		timeout = manualStartExecutionDeadline
+	} else {
+		// Set request deadline based on the schedule's catchup window, which is the
+		// latest time that it's acceptable to start this workflow.
+		tweakables := e.Config.Tweakables(scheduler.Namespace)
+		timeout = catchupWindow(scheduler, tweakables)
+	}
+
+	timeout = max(timeout, startWorkflowMinDeadline)
+	timeout = min(timeout, startWorkflowMaxDeadline)
```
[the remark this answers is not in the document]
lina-temporal: That's fine with me. IMO, even in a worst case scenario operationally, a buffer could fill up entirely with backing off starts. But at that point, it's unlikely start 1000 would succeed where 999 didn't, so we'd probably want to get paged anyways.

components/scheduler/executor_executors.go
lina-temporal: Thanks, will fix.

components/scheduler/executor_executors_test.go
lina-temporal: I gave the existing tests in `workflow_test` a more thorough look. At a high level, I think that most of these cases actually are covered in the new tests, just broken up closer to each component. Some of the Update specs that deal with behavior we'll implement synchronously (like clearing the buffer when a schedule's action changes), so they aren't represented yet. Others, such as `TestLotsOfIterations`, the CAN tests, and the Signal tests, are obsolete given the state machine architecture. The `ExitScheduleWorkflow` tests I'd consider analogously covered by the "empty buffer"/"no work to do" tests.

- I've brought a few more test cases over relevant to the Invoker
- `TestUpdateBetweenNominalAndJitter`: this tests [this branch](https://github.com/temporalio/temporal/blob/main/service/worker/scheduler/workflow.go) in existing scheduler, which maps to [this branch](https://github.com/temporalio/temporal/blob/main/components/scheduler/spec_processor.go#L115-L119) in SpecProcessor. I've added the test case to SpecProcessor.
- `TestUpdateNotRetroactive`: I think this test case will come later, with the update handler, since we'll be processing updates synchronously where possible. 
- `TestBackfillInclusiveStartEnd`: I'll make sure this is in the spec for the Backfiller. 
- `TestHugeBackfillBuffer`: ditto

components/scheduler/config.go
```diff
@@ -35,44 +36,66 @@ type (
 		DefaultCatchupWindow              time.Duration // Default for catchup window
 		MinCatchupWindow                  time.Duration // Minimum for catchup window
 		MaxBufferSize                     int           // MaxBufferSize limits the number of buffered actions pending execution in total
-		BackfillsPerIteration             int           // How many backfilled actions to buffer per iteration (implies rate limit since min sleep is 1s)
 		CanceledTerminatedCountAsFailures bool          // Whether cancelled+terminated count for pause-on-failure
+		RecentActionCount                 int           // How many recent actions are recorded in SchedulerInfo.
```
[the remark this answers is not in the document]
lina-temporal: Reworded to, "Number of recent actions taken (workflow execution results) recorded in the ScheduleInfo metadata."

components/scheduler/invoker.go
```diff
@@ -0,0 +1,293 @@
...
+
+// TransitionEnqueue adds buffered starts to the Invoker's queue.
+var TransitionEnqueue = hsm.NewTransition(
+	[]enumsspb.SchedulerInvokerState{
+		enumsspb.SCHEDULER_INVOKER_STATE_UNSPECIFIED,
+		enumsspb.SCHEDULER_INVOKER_STATE_WAITING,
+		enumsspb.SCHEDULER_INVOKER_STATE_PROCESSING,
+		enumsspb.SCHEDULER_INVOKER_STATE_BACKING_OFF,
+	},
+	enumsspb.SCHEDULER_INVOKER_STATE_PROCESSING,
```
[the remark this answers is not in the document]
lina-temporal: That's what `WAITING` was, but I do agree with your other comment about combining `WAITING` and `BACKING_OFF`. The only difference between those states is whether or not another `ProcessBuffer` task is generated, and that's deterministically based off of `len(invoker.getEligibleBufferStarts() + cancels/terminates)`.

components/scheduler/invoker.go
```diff
@@ -0,0 +1,293 @@
...
+		i.LastProcessedTime = timestamppb.New(event.LastProcessedTime)
+		return i.output()
+	},
+)
+
+// update updates the Invoker's internal state based on result. Should be
+// called only during transitions.
+func (i Invoker) update(result *processBufferResult) {
+	discards := make(map[string]bool) // request ID -> is present
+	ready := make(map[string]bool)
```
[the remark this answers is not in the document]
lina-temporal: Yeah, will update, I see the `map[string]struct{}` more prevalent in the codebase.

components/scheduler/invoker.go
```diff
@@ -0,0 +1,293 @@
...
+	// Update attempt counts and backoffs for failed/retrying starts.
+	for _, start := range i.GetBufferedStarts() {
+		if retry, ok := retryable[start.RequestId]; ok {
+			start.Attempt++
+			start.BackoffTime = retry.GetBackoffTime()
+		}
+	}
+}
+
+// EarliestRetry returns the earliest possible time an attempted BufferedStart
```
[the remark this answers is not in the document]
lina-temporal: Yep, cancels and terminates are always immediate.

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,582 @@
...
+	result = result.Append(e.terminateWorkflows(logger, *scheduler, invoker.GetTerminateWorkflows()))
+	result = result.Append(e.cancelWorkflows(logger, *scheduler, invoker.GetCancelWorkflows()))
+	sres, startResults := e.startWorkflows(logger, env, *scheduler, eligibleStarts)
+	result = result.Append(sres)
+
+	// Write results.
+	return env.Access(ctx, ref, hsm.AccessWrite, func(node *hsm.Node) error {
+		// Record completed executions on the Invoker.
+		err := hsm.MachineTransition(node, func(i Invoker) (hsm.TransitionOutput, error) {
+			return TransitionCompleteExecution.Apply(i, EventCompleteExecution{
```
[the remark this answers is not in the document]
lina-temporal: Yes, true, I meant "complete" as in "complete execute task". Maybe `TransitionRecordExecution` would be a better name?

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,582 @@
...
+	metricsWithTag := e.MetricsHandler.WithTags(
+		metrics.StringTag(metrics.ScheduleActionTypeTag, metrics.ScheduleActionStartWorkflow))
+	for _, start := range starts {
+		startResult, err := e.startWorkflow(env, scheduler, start)
+		if err != nil {
+			logger.Error("Failed to start workflow", tag.Error(err))
+
+			// Don't count "already started" for the error metric or retry, as it is most likely
+			// due to misconfiguration.
+			if !isAlreadyStartedError(err) {
```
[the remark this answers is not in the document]
lina-temporal: It wouldn't be as as result of a retry from the scheduler, though (or we wouldn't get `already started` back). This is supposed to account more for the case that something external to the scheduler kicked off our target action.

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,582 @@
...
+	readyStarts := action.OverlappingStarts
+	if action.NonOverlappingStart != nil {
+		readyStarts = append(readyStarts, action.NonOverlappingStart)
+	}
+
+	// Update result metrics.
+	result.OverlapSkipped = action.OverlapSkipped
+
+	// Add starting workflows to result, trim others.
+	for _, start := range readyStarts {
```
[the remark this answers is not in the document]
lina-temporal: I agree that more actions could become invalid (even if just by virtue of them falling past the `startWorkflowDeadline`, after this loop completes), but I don't see why that's worth optimizing against, considering even if we checked discards right before we write the task's completion transition, we'd still be racing against the `startWorkflowDeadline` wall clock. Maybe I'm not following the condition you're calling out?

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,582 @@
...
+		// Ensure we can take more actions. Manual actions are always allowed.
+		if !start.Manual && !scheduler.useScheduledAction(true) {
+			// Drop buffered automated actions while paused.
+			result.DiscardStarts = append(result.DiscardStarts, start)
+			continue
+		}
+
+		if env.Now().After(e.startWorkflowDeadline(scheduler, start)) {
```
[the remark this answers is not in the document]
lina-temporal: > Note that once a request is put in the buffer you should execute it to completion it since you're risking creating orphans with task retries.

Yep, that is how it's set up; we invalidate within the `ProcessBufferTask`, but once an action is in scope for `ExecuteTask`, we'll drive it to completion. `processBuffer` is the last chance we have to back out of a stale workflow start.

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,582 @@
...
+) (*schedulepb.ScheduleActionResult, error) {
+	requestSpec := scheduler.GetSchedule().GetAction().GetStartWorkflow()
+	nominalTimeSec := start.NominalTime.AsTime().Truncate(time.Second)
+	workflowID := fmt.Sprintf("%s-%s", requestSpec.WorkflowId, nominalTimeSec.Format(time.RFC3339))
+
+	if start.Attempt >= ExecutorMaxStartAttempts {
+		return nil, errRetryLimitExceeded
+	}
+
+	// Get rate limiter permission once per buffered start, on the first attempt only.
```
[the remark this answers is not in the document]
lina-temporal: The rate that we're trying to limit is "schedules kicked off" moreso than "RPS we're calling `StartWorkflowExecution` at", it's intended to reflect the customer-facing "max actions taken per second" limit on a schedule.

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,582 @@
...
+	}
+	ctx, cancelFunc := e.newContext(scheduler.Namespace)
+	defer cancelFunc()
+	_, err := e.HistoryClient.RequestCancelWorkflowExecution(ctx, request)
+	return err
+}
+
+func (e invokerTaskExecutor) newContext(namespace string) (context.Context, context.CancelFunc) {
+	ctx, cancelFunc := context.WithTimeout(context.Background(), e.Config.ServiceCallTimeout())
+	ctx = headers.SetCallerInfo(ctx, headers.NewBackgroundCallerInfo(namespace))
```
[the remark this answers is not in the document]
lina-temporal: Looked into this. The `outboundQueueActiveTaskExecutor` [provides the context](https://github.com/temporalio/temporal/blob/main/service/history/outbound_queue_active_task_executor.go#L105) to immediate tasks. `Executable` [generates the context](https://github.com/temporalio/temporal/blob/main/service/history/queues/executable.go#L281-L284) that goes to task executor queues, which does indeed call `SetCallerInfo` with the proper namespace. So, yes, it does! 👍  will update accordingly.

components/scheduler/invoker.go
```diff
@@ -0,0 +1,293 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
lina-temporal: [Done as a separate PR](https://github.com/temporalio/temporal/pull/7418).

components/scheduler/invoker.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	return deadline
+}
+
+// getEligibleBufferedStarts returns all BufferedStarts that are marked for
+// execution (Attempt > 0), and aren't presently backing off, based on last
+// processed time.
+func (i Invoker) getEligibleBufferedStarts() []*schedulespb.BufferedStart {
+	return util.FilterSlice(i.GetBufferedStarts(), func(start *schedulespb.BufferedStart) bool {
+		return start.Attempt > 0 && start.BackoffTime.AsTime().Before(i.GetLastProcessedTime().AsTime())
```
yycptt, line 267: Please help add some comments about the purpose of the LastProcessedTime field :)

components/scheduler/invoker_executors.go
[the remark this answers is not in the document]
lina-temporal: Another thing I'd like to raise is if we still *want* to rate limit the scheduler itself, versus relying on backpressure from APS limits on `StartWorkflowExecution`/cancel, terminate. This would be a good knob to simplify for our customers, not to mention operations.

components/scheduler/invoker_executors.go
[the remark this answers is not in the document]
lina-temporal: Yep, starts have a unique/deterministic request ID.

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,613 @@
...
+
+func (e invokerTaskExecutor) startWorkflow(
+	ctx context.Context,
+	env hsm.Environment,
+	scheduler Scheduler,
+	start *schedulespb.BufferedStart,
+) (*schedulepb.ScheduleActionResult, error) {
+	requestSpec := scheduler.GetSchedule().GetAction().GetStartWorkflow()
+	nominalTimeSec := start.NominalTime.AsTime().Truncate(time.Second)
+	workflowID := fmt.Sprintf("%s-%s", requestSpec.WorkflowId, nominalTimeSec.Format(time.RFC3339))
```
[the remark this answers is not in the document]
lina-temporal: Discussed this offline - no, we don't (it also isn't supported in workflow scheduler), so we might want to deprecate the field.

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,613 @@
...
+		UserMetadata:             requestSpec.UserMetadata,
+	}
+	result, err := e.FrontendClient.StartWorkflowExecution(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	return &schedulepb.ScheduleActionResult{
+		ScheduleTime: start.ActualTime,
+		ActualTime:   timestamppb.New(env.Now()),
```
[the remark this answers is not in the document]
lina-temporal: Yeah; `start.ActualTime` is the scheduled time after jitter, whereas `ActualTime` is meant to be the time when it was actually kicked off. I suppose we could hide the detail by using `start.ActualTime` for both, but that might be a little confusing if a start was scheduled but failed to start for a period before success.

components/scheduler/invoker.go
lina-temporal: I recalled the reason - it's subtle, I'll update the comments. It's because of how the Invoker drives its backoffs. Invoker generates a timer task to wake on the earliest possible backoff, with a `ProcessBufferTask`. When the `ProcessBufferTask` runs, to handle pending backoffs, it simply moves the `LastProcessedTime` high water mark forward. With the MachineTransition that updates `LastProcessedTime`, tasks are generated, and since `LastProcessedTime` > `BackoffTime` for some start, an immediate `ExecuteTask` is generated.

The reason that we account for it separately in `LastProcessedTime` is to support replication and being able to re-generate tasks based on incoming state. This way, when we replicate the machine at a specific point in time, it'll generate an identical set of tasks regardless of when tasks were regenerated (as opposed to if we used say, `env.Now()`).

components/scheduler/invoker_executors.go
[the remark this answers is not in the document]
dnr: It's not "still", the support was never finished. We should support it as soon as we can.

(The tricky part is to handle all the interactions with KeepOriginal and other features, specifically ALLOW_ALL overlap policy: we should block create/update with both of those set, and decide what to do for trigger with ALLOW_ALL if KeepOriginal is set.)

components/scheduler/invoker_executors.go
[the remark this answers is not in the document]
dnr: Yeah I thought the consensus was to let the operations themselves get limited (either at frontend or history or both) and drop any additional limiting. (Also the transactions on the scheduler entity itself count against APS and OPS, right?)

components/scheduler/invoker_executors.go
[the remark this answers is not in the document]
dnr: Sorry, the confusing names are my fault :disappointed: 

They're two separate contexts:

When talking about scheduled times, we have "Nominal" vs "Actual", "nominal" is what matches the schedule spec, "actual" is that plus jitter. So "actual" is the time when we want the thing to run.

When talking about what actually happened, the "actual" from above is when we _want_ it to run, and then there's when it actually did run. So "actual" turns into ScheduleTime and then the "real" time turns into ActualTime. Of course I should have picked a third word. Or maybe we should call it ActualActualTime or ReallyRealTime :laughing:

components/scheduler/invoker_executors.go
```diff
@@ -0,0 +1,613 @@
...
+				ActionCount: int64(len(startResults)),
+				Results:     startResults,
+			})
+		})
+	})
+}
+
+// shouldYield returns true when the immediate task should complete/spawn a new task.
+func (e invokerTaskExecutor) shouldYield(scheduler Scheduler, actionsTaken int) bool {
+	tweakables := e.Config.Tweakables(scheduler.Namespace)
```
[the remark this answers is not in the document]
dnr: It'll be less expensive after https://github.com/temporalio/temporal/pull/7052 that caches conversions

yycptt (approved):

components/scheduler/invoker_executors.go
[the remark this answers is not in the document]
yycptt: ideally we want to do rate limiting as early as possible and avoid unnecessary task retries, so today for task processing requests we are rate limiting the task itself, not individual operations within it. 

But looks like here rate limited operation won't cause task to fail so yeah we can either introduce an ops limited history client or do it on the callee side.

PR #7153, opened 2025-01-23
Attach priority to query

stephanos:
WISOTT


service/history/api/queryworkflow/api.go
```diff
@@ -181,6 +182,7 @@ func Invoke(
 				rawMatchingClient,
 				matchingClient,
 				scope,
+				mutableState.GetExecutionInfo().Priority,
```
dnr, line 185: This is after `workflowLease.GetReleaseFn()(nil)`.. I don't think we can access `mutableState.GetExecutionInfo()` here?

service/history/api/queryworkflow/api.go
```diff
@@ -237,6 +239,7 @@ func Invoke(
...
```
dnr, line 242: This one is too

service/history/api/queryworkflow/api.go
stephanos: I wish you could use emojis in Go function names; I didn't notice the `GetReleaseFn` line there at all.

service/history/api/queryworkflow/api.go
```diff
@@ -149,6 +150,8 @@ func Invoke(
 		return nil, serviceerror.NewWorkflowNotReady("Unable to query workflow due to Workflow Task in failed state.")
 	}
 
+	priority := mutableState.GetExecutionInfo().Priority
```
dnr, line 153: In some cases we've had problems with stuff like this, which just keeps a pointer to the proto (shared with the mutable state cache) instead of cloning it. As long as we treat Priority messages as immutable and always construct new ones, this is safe.

dnr (approved):
We could keep this open until we decide for sure if we're going with the implementation that supports queries or not. Or just merge it and either revert or ignore the query priorities for now.

stephanos:
I'll merge for now and we can revert later as long-standing open PRs have been frowned upon in the past.

PR #7154, opened 2025-01-23
SetWorkerDeploymentCurrentVersion:  Versioning-3.1

Shivs11:
## What changed?

- title

## Why?

- versioning-3.1

## How did you test it?

- existing suite of tests + added functional tests

## Potential risks

- None, feature branch

## Is hotfix candidate?

- No


go.mod
```diff
@@ -7,7 +7,7 @@ retract (
 	v1.26.0 // Published accidentally.
 )
 
-replace go.temporal.io/api => go.temporal.io/api v1.43.2-0.20250121234928-f4e3b077daa6
+replace go.temporal.io/api => go.temporal.io/api v1.43.2-0.20250123223624-792f7c873972
```
Shivs11, line 10: pointing to a feature on api-go: ``ss-routing-info-in-worker-deployment``
Corresponding API PR: https://github.com/temporalio/api/pull/527

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -31,11 +31,24 @@ import "google/protobuf/timestamp.proto";
...
 
-// Data for each worker deployment version + task queue pair.
-// This is stored in each worker deployment version (for each task queue),
-// and synced to task queue user data (for each worker deployment version).
-message DeploymentVersionTaskQueueData {
-    // TODO (Shivam): Add PollerStatus after it gets added in the API repo.
+
+// The source of truth for this data is in the WorkerDeployment entity workflows, which is
+// synced to all TQs whenever the source changes.
+message DeploymentVersionData {
```
Shivs11, line 37: This directly concerns @ShahabT's work so tagging you - I moved this `message` out of persistence inside `deployment` to stay consistent with the way we were doing things for versioning-3.0. A stronger reason for this was the presence of cyclic dependencies whereby if this message was present in `persistence`, then using it inside deployment (which we have to) would not be possible - even when David had worked on some TaskQueueUserData stuff, the data to be stored was defined inside the deployment package

service/frontend/workflow_handler.go
```diff
@@ -3361,6 +3345,50 @@ func (wh *WorkflowHandler) DescribeWorkerDeploymentVersion(ctx context.Context,
...
+	}
+
+	// TODO (Shivam) : Verify if sending in uuid.New() is correct
+	resp, err := wh.workerDeploymentClient.SetCurrentVersion(ctx, namespaceEntry, request.DeploymentName, request.Version.Value, request.Identity, uuid.New())
+	if err != nil {
+		switch err.(type) {
+		case *serviceerror.AlreadyExists:
+			return &workflowservice.SetWorkerDeploymentCurrentVersionResponse{
+				PreviousVersion: request.Version.Value, // previousVersion is the version which was requested to be set as current
+			}, nil
```
Shivs11, line 3375: idempotency

service/worker/workerdeployment/client.go
```diff
@@ -268,72 +267,14 @@ func (d *ClientImpl) DescribeWorkerDeployment(
 	return d.deploymentStateToDeploymentInfo(ctx, namespaceEntry, deploymentName, queryResponse.State)
 }
 
-func (d *ClientImpl) GetCurrentVersion(
```
Shivs11, line 271: I was in two minds about this - should we keep this method purely for testing purposes and to verify if local data changes all the way to the version workflow later were indeed happening? Removed it for now since it's not an API exposed on the frontend

service/worker/workerdeployment/version_activities.go
```diff
@@ -68,13 +68,19 @@ func (a *VersionActivities) SyncDeploymentVersionUserData(
...
+
+			res, err = a.matchingClient.SyncDeploymentUserData(ctx, &matchingservice.SyncDeploymentUserDataRequest{
 				NamespaceId:   a.namespace.ID().String(),
 				TaskQueue:     syncData.Name,
 				TaskQueueType: syncData.Type,
-				Deployment:    nil, // TODO (Shivam): Pass nil until matchingservice.SyncDeploymentUserDataRequest gets updated.
-				Data:          nil, // TODO (Shivam): Pass nil until matchingservice.SyncDeploymentUserDataRequest gets updated.
+				Operation: &matchingservice.SyncDeploymentUserDataRequest_UpdateVersionData{
+					UpdateVersionData: syncData.Data,
+				},
```
Shivs11, line 81: Note: I have not added code when we shall remove a version from the task-queue - will be coming in later

service/worker/workerdeployment/client.go
```diff
@@ -827,6 +768,11 @@ func (d *ClientImpl) deploymentStateToDeploymentInfo(ctx context.Context, namesp
 	workerDeploymentInfo.Name = deploymentName
 	workerDeploymentInfo.CreateTime = state.CreateTime
 
+	// RoutingInfo
+	workerDeploymentInfo.RoutingInfo = &deploymentpb.RoutingInfo{
+		CurrentVersion: state.CurrentVersion,
```
carlydf, line 773:
```suggestion
		CurrentVersion: state.CurrentVersion,
		CurrentVersionUpdateTime: state.CurrentChangedTime,
```

service/worker/workerdeployment/client.go
```diff
@@ -357,27 +298,27 @@ func (d *ClientImpl) SetCurrentVersion(
 		requestID,
 	)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errNoChangeType {
-		return nil, nil, serviceerror.NewAlreadyExist(fmt.Sprintf("Build ID %q is already current for %q",
+		return nil, serviceerror.NewAlreadyExist(fmt.Sprintf("Version %q is already current for %q",
```
carlydf, line 305: I think you should just return nil error here, and response with `PreviousVersion` = whatever the customer just tried to make current.

service/worker/workerdeployment/workflow.go
```diff
@@ -136,9 +135,34 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+	if _, err := d.syncVersion(ctx, args.Version, versionUpdateTime); err != nil {
+		return nil, err
+	}
+
+	if prevCurrentVersion != "" {
+		// tell previous current that it's no longer current
+		if _, err := d.syncVersion(ctx, prevCurrentVersion, versionUpdateTime); err != nil {
+			return nil, err
+		}
+	}
```
carlydf, line 151: I thought the plan was to sync to the task queues, then commit the new current to the deployment WF, then sync the new current to the version WFs.

It seems here that you're syncing the new current to the previous current and new current version WFs, and then having those workflows sync their new state to the task queues in `VersionWorkflowRunner.handleSyncState`. 

I think the whole `// sync to task queues` section of `handleSyncState` should be moved to `handleSetCurrent`, let me know if you disagree and we should discuss! My thinking is that our focus should be on how to make `SetCurrent` return quickly and reliably while guaranteeing:
- DescribeDeployment will not return the _new_ Current until SetCurrent has completed and has consistency with all the TQs
- SetCurrent will not return success until it has consistency with all the TQs (it might time out or return error but not success)

After the task queues and the Deployment WF have the new current, then we send updates to the version workflows to tell them too. I'm not super opinionated about whether those should complete before we return success to the SetCurrent caller, but I think it's fine to return success as soon as the new current is persisted in the TQs and in the Deployment WF, since those are the sources of truth for workflow task dispatch and `DescribeDeployment.CurrentVersion`

service/worker/workerdeployment/workflow.go
```diff
@@ -136,9 +135,34 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+
+	// update local state
+	d.State.CurrentVersion = args.Version
+	d.State.CurrentChangedTime = versionUpdateTime
+
+	// update memo
+	if err = d.updateMemo(ctx); err != nil {
+		return nil, err
+	}
+
```
carlydf, line 161: only after the Deployment local state / memo are updated to reflect the new current should we notify the versions of the new current.

service/worker/workerdeployment/workflow.go
carlydf: cc @ShahabT in case I have the wrong understanding here

Shivs11: thanks for starting this discussion!
a couple of things:

a) I think syncing to the task-queues should happen inside each version workflow themselves. We need these operations to be atomic in nature and since these version workflows own the task-queues, syncing to the user-data should happen inside of `handleSyncState` rather than inside `handleSetCurrent`. 

The chain of flow right now, as implemented, is as follows:

1. Sync task-queues of the "to-be-current" version.
2. Update local state of the "to-be-current" version after syncing is complete.
3. Sync task-queues of the "previously-current" version.
4. Update local state of the "previously-current" version after syncing is successfully done.
5. Update local state of the `worker deployment` workflow.

b) Another point, and perhaps the more important one, is the ordering of this chain of flow. I am fully for the motion of returning from setCurrent after the TQ's are consistent. If I have understood your reasoning correctly, you reason for (2), (3), (4) to happen after (5) (in the above listed order). I thought about that and here are my concerns:

After (1) and (5), if we send async write operations to the version-workflow (in the form of signals) and return to the client, it *could* lead to a state inconsistency scenario - consider an example if another client were to send an update request which gets processed before the initial scheduled signals - this would be pretty bad since our version local state then would go out of sync and things would break

Thus, after thinking about this, I think we should keep all our updates in a "serial" fashion - Yes, it will be at more latency but we should prefer keeping things consistent as much as we can

Let me know if I have gotten this wrong and I am happy to be corrected!

service/frontend/workflow_handler.go
```diff
@@ -3361,6 +3345,50 @@ func (wh *WorkflowHandler) DescribeWorkerDeploymentVersion(ctx context.Context,
...
+	if !wh.config.EnableDeploymentVersions(request.Namespace) {
+		return nil, errDeploymentsNotAllowed
+	}
+
+	namespaceEntry, err := wh.namespaceRegistry.GetNamespace(namespace.Name(request.GetNamespace()))
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO (Shivam) : Verify if sending in uuid.New() is correct
```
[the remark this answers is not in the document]
Shivs11: Thanks for this thought provoking comment - I was struggling to come up with this thinking because the code was initially written by David in versioning-3.0 (I think) , hence the TODO comment - Since `setCurrent` as a whole is going to be idempotent, I don't think syncing it again makes a lot of sense since we would have already synced task-queues if A was already current earlier and any new task-queues added are synced when registered.

Hence, I think it's okay to get rid of this altogether :)

service/worker/workerdeployment/workflow.go
Shivs11: addressing this is directly tied to my previous comment - we should discuss this more in sync!

service/matching/matching_engine.go
```diff
@@ -1675,13 +1676,11 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
...
-				if old.GetRoutingUpdateTime().AsTime().Before(vd.GetRoutingUpdateTime().AsTime()) {
-					// only update if the timestamp is more recent
-					deploymentData.Versions[idx] = vd
-				} else {
-					// No-op
+				if old.GetRoutingUpdateTime().AsTime().After(vd.GetRoutingUpdateTime().AsTime()) {
 					return nil, false, errUserDataUnmodified
 				}
+				// only update if the timestamp is more recent
+				deploymentData.Versions[idx] = vd
```
Shivs11, line 1683: changed this to stop lint complaints - also easier to read imo

service/worker/workerdeployment/client.go
```diff
@@ -268,82 +267,24 @@ func (d *ClientImpl) DescribeWorkerDeployment(
...
 	deploymentName string,
 	version string,
-	updateMetadata *deploymentpb.UpdateDeploymentMetadata,
 	identity string,
-	requestID string,
-) (_ *deploymentpb.WorkerDeploymentVersionInfo, _ *deploymentpb.WorkerDeploymentVersionInfo, retErr error) {
+) (_ *deploymentspb.SetCurrentVersionResponse, retErr error) {
 	//revive:disable-next-line:defer
 	defer d.record("SetCurrentVersion", &retErr, namespaceEntry.Name(), deploymentName, version, identity)()
+	requestID := uuid.New()
```
carlydf, line 279: Since we're not using this, could you delete it from `SetCurrentVersionArgs` and not generate a random ID here?

carlydf (approved):

service/worker/workerdeployment/client.go
```diff
@@ -357,27 +298,27 @@ func (d *ClientImpl) SetCurrentVersion(
...
 	}
 
+	var res deploymentspb.SetCurrentVersionResponse
 	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errNoChangeType {
-		return nil, nil, serviceerror.NewAlreadyExist(fmt.Sprintf("Build ID %q is already current for %q",
-			version, deploymentName))
+		res.PreviousVersion = version
+		return &res, nil
 	} else if failure != nil {
 		// TODO: is there an easy way to recover the original type here?
```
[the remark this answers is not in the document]
carlydf: but I don't think you need to!

service/worker/workerdeployment/client.go
Shivs11: I think we still need a requestID being generated - we are not using it for de-dup, but we end up using it for updateID's (which are required for an update) as well as in our start workflow execution requests - both these cases require a unique identifier which is typically a uuid - having said that, I can remove it from `setCurrentVersionArgs` but I think the generation of the ID should still be here

carlydf: oh ok makes sense if we're passing it to update

service/worker/workerdeployment/workflow.go
carlydf: I'm convinced that we should sync to the version workflows sequentially because it would be hard to get out of a bad state if someone sends two different SetCurrent requests back-to-back while we are still eventually making the version workflows consistent with the already-returned first request.

PR #7158, opened 2025-01-24
Worker Deployment Version Drainage Status

carlydf:
## What changed?
Add DrainageStatus child workflow to worker deployment system

## Why?
To periodically update the version workflow with results from visibility.

## How did you test it?
Functional tests. Currently broken (see todo comment in code)


service/worker/workerdeployment/version_workflow.go
```diff
@@ -286,68 +296,26 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
+		state.CurrentSinceTime = args.GetCurrentSinceTime()
+	}
+	if args.GetRampingSinceTime() != nil {
+		state.RampingSinceTime = args.GetRampingSinceTime()
 	}
+	isAcceptingNewWorkflows := state.GetCurrentSinceTime() != nil || state.GetRampingSinceTime() != nil
 
-	// TODO (Shivam) - Move this when UpdateMetadata is implemented.
+	if wasAcceptingNewWorkflows && !isAcceptingNewWorkflows {
+		workflow.ExecuteChildWorkflow(ctx, WorkerDeploymentCheckDrainageWorkflowType)
```
carlydf, line 310: todo: if an update causes the version to start accepting workflows again, send a cancel request to the child workflow

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -111,15 +120,19 @@ message RegisterWorkerInVersionArgs {
...
-    }
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not current.
+    google.protobuf.Timestamp current_since_time = 2;
+
+    // (-- api-linter: core::0140::prepositions=disabled
+    //     aip.dev/not-precedent: 'Since' captures the field semantics despite being a preposition. --)
+    // Nil if not ramping.
+    google.protobuf.Timestamp ramping_since_time = 3;
```
Shivs11, line 135: thinking out loud: if we are planning on using `version_workflow.handleSyncState` to also set a version to ramping, this should have a ramp percentage set right?

service/worker/workerdeployment/util.go
```diff
@@ -40,8 +40,9 @@ import (
 
 const (
 	// Workflow types
-	WorkerDeploymentVersionWorkflowType = "temporal-sys-worker-deployment-version-workflow"
-	WorkerDeploymentWorkflowType        = "temporal-sys-worker-deployment-workflow"
+	WorkerDeploymentVersionWorkflowType       = "temporal-sys-worker-deployment-version-workflow"
+	WorkerDeploymentWorkflowType              = "temporal-sys-worker-deployment-workflow"
+	WorkerDeploymentCheckDrainageWorkflowType = "temporal-sys-worker-deployment-check-drainage-workflow"
```
Shivs11, line 45: `temporal-sys-worker-deployment-check-drainage-workflow` -> `temporal-sys-worker-deployment-drainage-workflow` 

no need to keep it super verbose; wdyt?

service/worker/workerdeployment/client.go
```diff
@@ -117,6 +118,11 @@ type Client interface {
 		identity string,
 		requestID string,
 	) (*deploymentspb.AddVersionToWorkerDeploymentResponse, error)
+
+	GetVersionDrainageStatus(
```
Shivs11, line 122: nit: comment explaining it's being used only internally could be nice here

service/worker/workerdeployment/version_workflow.go
```diff
@@ -303,70 +308,78 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
-	// for key, payload := range args.UpdateMetadata.GetUpsertEntries() {
-	//	d.VersionState.Metadata[key] = payload
-	// }
-	// for _, key := range args.UpdateMetadata.GetRemoveEntries() {
-	//	delete(d.VersionState.Metadata, key)
-	// }
+	if wasAcceptingNewWorkflows && !isAcceptingNewWorkflows {
+		state.DrainageInfo = &deploymentpb.VersionDrainageInfo{}
+		workflow.ExecuteChildWorkflow(ctx, WorkerDeploymentCheckDrainageWorkflowType, d.a.namespace)
+	}
```
Shivs11, line 375: say we have a version workflow that has become draining and kickstarts this workflow - now, say this version workflow CAN's - I might have missed this but I don't see a `PARENT_CLOSE_POLICY` for this child workflow

IMHO, we should: 

1. set the policy to terminate so that the workflow terminates when we CAN 
2. In the `version_workflow.run` function, if the drainage state is set to `draining`, we simply kickstart this child workflow again which shall continue execution 

This will also call for the todo statement you have inside of `drainage_workflow.go` to be implemented since if this version workflow were to accept executions, we would have to send a signal to this workflow to stop carrying out execution

wdyt?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -264,26 +279,16 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
...
-			// version will become current
-			if !d.VersionState.LastBecameCurrentTime.AsTime().Equal(set.LastBecameCurrentTime.AsTime()) {
-				return nil
-			}
-		}
+func (d *VersionWorkflowRunner) validateSyncState(args *deploymentspb.SyncVersionStateUpdateArgs) error {
+	if args.GetCurrentSinceTime().AsTime().Equal(d.GetVersionState().GetCurrentSinceTime().AsTime()) &&
+		args.GetRampingSinceTime().AsTime().Equal(d.GetVersionState().GetRampingSinceTime().AsTime()) {
+		res := &deploymentspb.SyncVersionStateResponse{VersionState: d.VersionState}
+		return temporal.NewApplicationError("no change", errNoChangeType, res)
```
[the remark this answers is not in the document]
carlydf: I think when `SetCurrent` is called, `ramping_since_time` will be set to nil, and we want to sync both values. Similarly, if `SetRamp` is called, `current_since_time` will be set to nil, which we want to sync across the workflows/task queues that need to know that.

proto/internal/temporal/server/api/deployment/v1/message.proto
carlydf: yes, I think so! I was thinking to wait to add that until the ramp PR, but we could pre-emptively add the field now. I'll at least add a todo

service/worker/workerdeployment/util.go
carlydf: nice I like that

service/worker/workerdeployment/version_workflow.go
carlydf: yes, I need to add the code handling CaN of the parent and also handling termination of the child workflow if the parent version starts accepting traffic again

service/worker/workerdeployment/workflow.go
```diff
@@ -135,17 +135,28 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+	if _, err := d.syncVersion(ctx, newCurrent, currUpdateArgs); err != nil {
 		return nil, err
 	}
 
-	if prevCurrentVersion != "" {
+	if prevCurrent != "" {
 		// tell previous current that it's no longer current
-		if _, err := d.syncVersion(ctx, prevCurrentVersion, versionUpdateTime); err != nil {
+		prevUpdateArgs := &deploymentspb.SyncVersionStateUpdateArgs{
+			RoutingUpdateTime: versionUpdateTime,
```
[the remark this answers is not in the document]
carlydf: I need to check how Shahab uses it in his routing code, but conceptually making a version not-current is definitely a change to its routing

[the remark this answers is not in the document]
carlydf: The versioning Info that is used to route a workflow is the one in the list that IsCurrent _and_ has the most recent routing update time. So if it is not current, then it is ignored.
```
	// Find new current and ramping
	for _, v := range deployments.GetVersions() {
		if v.RoutingUpdateTime != nil && v.GetIsCurrent() {
			if t := v.RoutingUpdateTime.AsTime(); t.After(current.GetRoutingUpdateTime().AsTime()) {
				current = v
			}
		}
		if v.RoutingUpdateTime != nil && v.GetRampPercentage() > 0 {
			if t := v.RoutingUpdateTime.AsTime(); t.After(ramping.GetRoutingUpdateTime().AsTime()) {
				ramping = v
			}
		}
	}
```
So I think we can update the routing time here.

service/worker/workerdeployment/version_workflow.go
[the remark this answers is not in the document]
carlydf: I changed this to make more sense

service/worker/workerdeployment/version_workflow.go
```diff
@@ -388,9 +401,12 @@ func (d *VersionWorkflowRunner) updateMemo(ctx workflow.Context) error {
...
 		WorkerDeploymentVersionMemoField: &deploymentspb.VersionWorkflowMemo{
-			DeploymentName: d.VersionState.DeploymentName,
-			Version:        d.VersionState.Version,
-			CreateTime:     d.VersionState.CreateTime,
+			DeploymentName:   d.VersionState.DeploymentName,
+			Version:          d.VersionState.Version,
+			CurrentSinceTime: d.VersionState.CurrentSinceTime,
+			RampingSinceTime: d.VersionState.RampingSinceTime,
+			CreateTime:       d.VersionState.CreateTime,
+			DrainageInfo:     d.VersionState.DrainageInfo,
```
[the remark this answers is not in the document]
carlydf: After our discussion on what the memo is used for (to give info via DescribeWorkflowExecution and ListWorkflowExecutions) I removed everything here, because DescribeVersion uses query, not memo. I have a hunch that we might find it useful to put drainage info in here at some point, but we should only add that when we have a use-case for it

service/worker/workerdeployment/workflow.go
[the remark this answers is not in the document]
Shivs11: > The versioning Info that is used to route a workflow is the one in the list that IsCurrent and has the most recent routing update time. So if it is not current, then it is ignored.
yeah, that is right my concern was not about an issue on the routing side since we have strong checks implemented by Shahab there - my concern was more or so about if a version is becoming non-current, should we update the RoutingTime

But reading this comment solved it - 
```
// Last time `is_current` or `ramp_percentage` of this version changed.
```

service/worker/workerdeployment/version_workflow.go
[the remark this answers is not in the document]
Shivs11: I like the change more - no need to evaluate the timing since current was set just check if it was current/ramping

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -58,19 +58,26 @@ message VersionLocalState {
 
     map<string, TaskQueueFamilyData> task_queue_families = 3;
 
-    bool is_current = 4;
-    google.protobuf.Timestamp last_became_current_time = 5;
+    // Last time `is_current` or `ramp_percentage` of this version changed.
+    google.protobuf.Timestamp routing_update_time = 4;
+    // If this version is the current version of its deployment.
+    bool is_current = 5;
+    // Range: [0, 100]. Must be zero if is_current is true. Must be non-zero if `version.version`
```
Shivs11, line 65: may have to change this comment since we don't have a `version.version` (or we may have in the near future based on Shahab's latest API refinements)

tests/deployment_version_test.go
```diff
@@ -192,3 +198,215 @@ func (s *DeploymentVersionSuite) Name() string {
...
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		a := assert.New(t)
+		resp, err := s.FrontendClient().DescribeWorkerDeploymentVersion(ctx, &workflowservice.DescribeWorkerDeploymentVersionRequest{
+			Namespace: s.Namespace().String(),
+			Version:   tv2.BuildID(),
+		})
+		a.NoError(err)
+		a.Equal(tv2.DeploymentSeries(), resp.GetWorkerDeploymentVersionInfo().GetDeploymentName())
+		a.Equal(tv2.BuildID(), resp.GetWorkerDeploymentVersionInfo().GetVersion())
+	}, time.Second*5, time.Millisecond*200)
```
Shivs11, line 291: this may not be what you are intending to do - I assume you wanted to start a new version workflow, under the same deployment name, right? However, both tv1 and tv2 are using tv.BuildID() which generates the same BuildID making only one version workflow registered

IMHO, maybe try using .WithBuildIDNumber(`number`) for both these separate tvs?

tests/deployment_version_test.go
```diff
@@ -192,3 +207,173 @@ func (s *DeploymentVersionSuite) Name() string {
 		farm.Fingerprint32([]byte(fullName)),
 	)
 }
+
+func (s *DeploymentVersionSuite) TestDrainageStatus_SetCurrentVersion_NoOpenWFs() {
```
carlydf, line 211: This tests the new parts of the feature (version stops accepting new executions --> child workflow starts --> child sets the status to DRAINING and sleeps for `visibilityGracePeriod` --> child queries visibility, counts 0 open workflows, and sets the status to DRAINED).

It does not test the visibility query, since there are no open workflows, but the visibility query code has not changed from the pre-release, so that should all still work. There are TODOs to test it in the comments below.

service/worker/workerdeployment/client.go
```diff
@@ -492,13 +499,23 @@ func (d *ClientImpl) updateWithStartWorkerDeploymentVersion(
...
-			DeploymentName: deploymentName,
-			Version:        version,
-			CreateTime:     timestamppb.Now(),
+			Version: &deploymentpb.WorkerDeploymentVersion{
+				DeploymentName: deploymentName,
+				BuildId:        version,
+			},
+			WorkflowVersioningMode: 0, // todo
+			CreateTime:             now,
+			RoutingUpdateTime:      now,
```
Shivs11, line 513: hmm, I think I might have realized there might be a lag in our implementation: Assume the following flow:

`RegisterTaskQueueWorker` -> `updateWithStartWorkerDeploymentVersion` -> `handleRegisterWorker`

In the piece of code highlighted, we update the `routingUpdateTime` to be `now` which is right. This shall make the local state of the version wf have the `routingUpdateTime` to also be `now`. 

Now, when the update is going to take place, in `handleRegisterWorker`, we update the `routingUpdateTime` for the task-queue to be registered by again doing `time.Now()`. This may cause a lag between the local state's update time and the time being sent to the task-queue.

```
In handleRegisterWorker:236

	// initial data
	data := &deploymentspb.DeploymentVersionData{
		Version:           d.VersionState.Version,
		RoutingUpdateTime: timestamppb.Now(),
		CurrentSinceTime:  nil, // not current
		RampingSinceTime:  nil, // not ramping
		RampPercentage:    0,   // not ramping
		FirstPollerTime:   args.FirstPollerTime,
	}

```

again, not a big deal given that routing will work - but just some food for thought if we should take the extra step and achieve no lag

Shivs11 (approved):
haven't *yet* peeked at the functional tests but the code looks nice - thanks for the annoying refactorings required with the latest API refinements too - I shall take a peek at the tests sometime later today but merging to iterate faster

Shivs11:
merging this to speed up development - more tests to follow
cc - @carlydf

service/worker/workerdeployment/client.go
carlydf: hm yeah I think it would be better to pass the original routing time along

tests/deployment_version_test.go
carlydf: I think since I initiated them with this, now the build ids are different whenever I do tv.BuildId(). Thanks for catching this!
```
	tv1 := testvars.New(s).WithBuildIDNumber(1)
	tv2 := testvars.New(s).WithBuildIDNumber(2)
```

PR #7160, opened 2025-01-24
Use namespace.ID in dynamicconfig

stephanos:


service/history/api/create_workflow_util_test.go
```diff
@@ -0,0 +1,113 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
stephanos, line 5: Using same dates as the file I moved this from.

common/constants.go
```diff
@@ -63,11 +63,6 @@ const (
 	CriticalLongPollTimeout = time.Second * 20
 )
 
-const (
-	// MaxWorkflowTaskStartToCloseTimeout sets the Max Workflow Task start to close timeout for a Workflow
-	MaxWorkflowTaskStartToCloseTimeout = 120 * time.Second
-)
-
```
stephanos, line 70: Moved to history service.

common/util_test.go
```diff
@@ -68,86 +67,6 @@ func TestIsContextCanceledErr(t *testing.T) {
 	require.True(t, IsContextCanceledErr(ctx.Err()))
 }
```
stephanos, line 69: All moved to history service (where it's actually used):

common/util_test.go
```diff
@@ -40,7 +40,6 @@ import (
 	enumsspb "go.temporal.io/server/api/enums/v1"
 	historyspb "go.temporal.io/server/api/history/v1"
 	persistencespb "go.temporal.io/server/api/persistence/v1"
-	"go.temporal.io/server/common/dynamicconfig"
```
stephanos, line 43: This was causing a circular dependency compile error.

cmd/tools/gendynamicconfig/main.go
```diff
@@ -95,8 +95,8 @@ var (
 		},
 		{
 			Name:   "NamespaceID",
-			GoArgs: "namespaceID string",
-			Expr:   "[]Constraints{{NamespaceID: namespaceID}, {}}",
+			GoArgs: "namespaceID namespace.ID",
+			Expr:   "[]Constraints{{NamespaceID: namespaceID.String()}, {}}",
```
dnr, line 99: Did you consider changing the type of the field in the Constraints struct also?

_Technically_ this is public (i.e. through ServerOptions) api, so we shouldn't, but we could argue it's trivial and very likely no one would notice.

dnr (approved):
Awesome, thanks!

cmd/tools/gendynamicconfig/main.go
stephanos: I haven't considered that yet, I felt most of the value was captured by ensuring typesafety on the API.

I've got no strong opinion either way there.

But first, I was going to do the same for namespace name to wrap that up.

PR #7165, opened 2025-01-27
OTEL annotates gRPC responses

stephanos:
## What changed?

Added support for our OTEL interceptor to parse tags from the gRPC response (not just the request).

## Why?

Without annotating the responses, traces from calls like `PollWorkflowTaskQueue` cannot be queried for since the ID is only on the response but not the request.

## How did you test it?

Verified in Tempo:

image

## Potential risks

There is one other caller of `Extract`, but they are only feeding in the request payload. So no behavior change is expected.

### DIFF
--- a/cmd/tools/genrpcserverinterceptors/main.go
+++ b/cmd/tools/genrpcserverinterceptors/main.go
@@ -43,10 +43,10 @@ import (
 	"go.temporal.io/server/api/matchingservice/v1"
 )
 
-const maxPayloadDepth = 5
+const maxMessageDepth = 5
 
 type (
-	payloadData struct {
+	messageData struct {
 		Type string
 
 		WorkflowIdGetter string
@@ -57,7 +57,7 @@ type (
 	grpcServerData struct {
 		Server   string
 		Imports  []string
-		Payloads []payloadData
+		Messages []messageData
 	}
 )
 
@@ -125,9 +125,9 @@ import (
 	"go.temporal.io/server/common/log/tag"
 )
 
-func (wt *WorkflowTags) extractFrom{{.Server}}Payload(payload any) []tag.Tag {
-	switch r := payload.(type) {
-	{{- range .Payloads}}
+func (wt *WorkflowTags) extractFrom{{.Server}}Message(message any) []tag.Tag {
+	switch r := message.(type) {
+	{{- range .Messages}}
 	case {{.Type}}:
 	{{- if or .TaskTokenGetter .WorkflowIdGetter .RunIdGetter}}
 		{{- if .TaskTokenGetter}}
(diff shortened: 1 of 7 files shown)


common/telemetry/grpc.go
```diff
@@ -166,10 +148,11 @@ func (c *customServerStatsHandler) HandleRPC(ctx context.Context, stat stats.RPC
 			span.SetAttributes(attribute.Key("rpc.request.type").String(msgType))
 		}
 	case *stats.OutPayload:
+		span := trace.SpanFromContext(ctx)
+		c.annotateTags(ctx, span, s.Payload)
```
stephanos, line 152: This is the main change: extracting the tags from the gRPC _response_.

common/rpc/interceptor/logtags/matching_service_server_gen.go
```diff
@@ -31,81 +31,151 @@ import (
...
 		return nil
+	case *matchingservice.PollNexusTaskQueueResponse:
+		return nil
 	case *matchingservice.PollWorkflowTaskQueueRequest:
 		return nil
+	case *matchingservice.PollWorkflowTaskQueueResponse:
+		return []tag.Tag{
+			tag.WorkflowID(r.GetWorkflowExecution().GetWorkflowId()),
+			tag.WorkflowRunID(r.GetWorkflowExecution().GetRunId()),
+		}
```
stephanos, line 135: This is one of the new cases.

common/telemetry/grpc.go
stephanos: And _could_ happen that it overrides a tag set by the request, I think that's alright for now.

common/rpc/interceptor/logtags/workflow_tags.go
```diff
@@ -53,22 +53,22 @@ func NewWorkflowTags(
 	}
 }
 
-func (wt *WorkflowTags) Extract(req any, fullMethod string) []tag.Tag {
-	if req == nil {
+func (wt *WorkflowTags) Extract(payload any, fullMethod string) []tag.Tag {
```
alexshtin, line 56: I would leave this as `req`. It is request, not generic payload.

cmd/tools/genrpcserverinterceptors/main.go
```diff
@@ -43,10 +43,10 @@ import (
 	"go.temporal.io/server/api/matchingservice/v1"
 )
 
-const maxRequestDepth = 5
+const maxPayloadDepth = 5
 
 type (
-	requestData struct {
+	payloadData struct {
```
alexshtin, line 49: Word "payload" have specific meaning in this project (and proto messages). May be `reqResp` instead of `request`?

alexshtin (approved):

cmd/tools/genrpcserverinterceptors/main.go
stephanos: How about "messageData"?

alexshtin: Also confusing, but I think it is better.

PR #7168, opened 2025-01-28
Refactoring: remove code duplication in query parsers

ychebotarev:
## What changed?
There was 3 function related to query parsing, that where copy/pasted in 4 different files.
Combine them into a single function shared across different modules.

## Why?

Violate "rule of three".

## How did you test it?

existing tests

## Potential risks

can't think of any if there was enough unit tests.

## Is hotfix candidate?

N/A

### DIFF
--- a/common/archiver/s3store/query_parser.go
+++ b/common/archiver/s3store/query_parser.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/temporalio/sqlparser"
+	"go.temporal.io/server/common/sqlquery"
 	"go.temporal.io/server/common/util"
 )
 
@@ -75,7 +76,7 @@ func NewQueryParser() QueryParser {
 }
 
 func (p *queryParser) Parse(query string) (*parsedQuery, error) {
-	stmt, err := sqlparser.Parse(fmt.Sprintf(util.QueryTemplate, query))
+	stmt, err := sqlparser.Parse(fmt.Sprintf(sqlquery.QueryTemplate, query))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


common/util/query.go
alexshtin, line 1: Please don't add any new code to `util` package. We need to remove it. `sqlquery` sounds like a good new one for your needs.

common/archiver/s3store/query_parser.go
```diff
@@ -70,18 +68,14 @@ const (
...
-)
 
 // NewQueryParser creates a new query parser for filestore
 func NewQueryParser() QueryParser {
 	return &queryParser{}
 }
 
 func (p *queryParser) Parse(query string) (*parsedQuery, error) {
-	stmt, err := sqlparser.Parse(fmt.Sprintf(queryTemplate, query))
+	stmt, err := sqlparser.Parse(fmt.Sprintf(util.QueryTemplate, query))
```
alexshtin: I would make this a function and unexport `QueryTemplate`:
```suggestion
	stmt, err := sqlquery.Parse(query)
```

alexshtin (approved):

common/util/query.go
ychebotarev: move all to the sqlquery folder in common.

common/archiver/s3store/query_parser.go
ychebotarev: sqlparser is a 3rd party package. I can't make this change

alexshtin: This is not what I ask for :-)

rodrigozhou (approved):

common/util/query.go
dnr: Why would we remove the `common/util` package? It's a wonderful little package. Are you thinking of `common` itself? That one should definitely go away.

However, I agree this code can't go there because the rule is `common/util` can't depend on any other server packages, so that it can't create dependency cycles.

alexshtin: "wonderful little". Let's keep it this way.

PR #7171, opened 2025-01-28
Actually delete HSMs on Nexus completion

bergundy:
## Why?

We didn't go through the same code path in active and passive clusters and never deleted the state machines as intended.

## How did you test it?

Added a functional test.

### DIFF
--- a/components/nexusoperations/completion.go
+++ b/components/nexusoperations/completion.go
@@ -42,11 +42,10 @@ func handleSuccessfulOperationResult(
 	operation Operation,
 	result *commonpb.Payload,
 	links []*commonpb.Link,
-	completionSource CompletionSource,
-) (hsm.TransitionOutput, error) {
+) error {
 	eventID, err := hsm.EventIDFromToken(operation.ScheduledEventToken)
 	if err != nil {
-		return hsm.TransitionOutput{}, err
+		return err
 	}
 	event := node.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_COMPLETED, func(e *historypb.HistoryEvent) {
 		// We must assign to this property, linter doesn't like this.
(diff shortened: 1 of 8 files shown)


components/nexusoperations/completion.go
```diff
@@ -196,17 +181,21 @@ func CompletionHandler(
...
+			return nil
+		}
+		if requestID != "" && operation.RequestId != requestID {
+			isRetryableNotFoundErr = false
+			return serviceerror.NewNotFound("operation not found")
+		}
+		if opFailedError != nil {
+			err = handleUnsuccessfulOperationError(node, operation, opFailedError)
+		} else {
+			err = handleSuccessfulOperationResult(node, operation, result, nil)
```
yux0, line 195: do you need to pass in links?

yux0:
Two questions:
1. how does it delete the hsm from passive cluster?
2. do you need to revert #7032 ?

components/nexusoperations/completion.go
```diff
@@ -196,17 +181,21 @@ func CompletionHandler(
...
-			}
-			return handleSuccessfulOperationResult(node, operation, result, nil, CompletionSourceCallback)
-		})
-		// TODO(bergundy): Remove this once the operation auto-deletes itself from the tree on completion.
+		operation, err := hsm.MachineData[Operation](node)
+		if err != nil {
+			return nil
+		}
+		if requestID != "" && operation.RequestId != requestID {
+			isRetryableNotFoundErr = false
```
justinp-tt, line 189: Not clear on why we bother setting this if we return on the next line.

justinp-tt (approved):

components/nexusoperations/completion.go
bergundy: Hmm... let me look into this, I think something may have been lost in the translation.

bergundy: Ah this is because we check the condition after the Access function call.

components/nexusoperations/completion.go
bergundy: No that happens only in the start event.

bergundy:
> Two questions:
> 
> 1. how does it delete the hsm from passive cluster?

The logic for deletion is in the event definitions, those Apply functions that we call now instead of transitioning the state machines directly. The Apply functions are called when history events are replicated.

> 2. do you need to revert [Fix ErrStateMachineNotFound handling in HSM state replication #7032](https://github.com/temporalio/temporal/pull/7032) ?

No that's still needed in case we get the sync HSM replication task after replicating history (e.g. out of order tasks).

PR #7173, opened 2025-01-28
ListWorkerDeployments : Versioning-3.1

Shivs11:
## What changed?

- ListWorkerDeployments API
- Also removed presence of locks when adding a version to a worker-deployment workflow. Causes deadlocks and we don't need them.

## Why?

- Versioning-3.1

## How did you test it?

- Added functional tests (happy-path)
- Existing suite

## Potential risks

- None, feature

## Is hotfix candidate?

- No


service/matching/physical_task_queue_manager.go
```diff
@@ -658,26 +658,26 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeploymentVersion(
...
-	// 	return err
-	// }
-
-	// deploymentVersionData := userData.GetData().GetPerType()[int32(c.queue.TaskType())].GetDeploymentVersionData()
-	// if findDeploymentVersion(deploymentVersionData, workerDeployment.SeriesName, workerDeployment.BuildId) >= 0 {
-	// 	// already registered in user data, we can assume the workflow is running.
-	// 	// TODO: consider replication scenarios where user data is replicated before
-	// 	// the deployment workflow.
-	// 	return nil
-	// }
```
Shivs11, line 672: had forgotten to uncomment this and the next block of code - important since it basically serves as a de-dup mechanism for multiple polls coming in with the same worker-deployment information!

service/worker/workerdeployment/util.go
```diff
@@ -44,7 +43,7 @@ const (
 	WorkerDeploymentWorkflowType        = "temporal-sys-worker-deployment-workflow"
 
 	// Namespace division
-	WorkerDeploymentVersionNamespaceDivision = "TemporalWorkerDeploymentVersion"
+	WorkerDeploymentVersionNamespaceDivision = "TemporalWorkerDeployment"
```
carlydf, line 46: Should this variable be called `WorkerDeploymentNamespaceDivision` ?

carlydf: I think the naming with "version" is left over from when "Deployment" (instance) was the "first-class-citizen" of this whole API

service/worker/workerdeployment/version_workflow.go
```diff
@@ -238,17 +238,6 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
...
-	activityCtx = workflow.WithActivityOptions(ctx, defaultActivityOptions)
-	err = workflow.ExecuteActivity(activityCtx, d.a.AddVersionToWorkerDeployment, &deploymentspb.AddVersionToWorkerDeploymentRequest{
-		DeploymentName: d.VersionState.DeploymentName,
-		Version:        d.VersionState.Version,
-		RequestId:      d.newUUID(ctx),
-	}).Get(ctx, nil)
-	if err != nil {
-		return err
-	}
-
```
carlydf, line 251: why is this deleted?

service/worker/workerdeployment/workflow.go
```diff
@@ -115,7 +117,7 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 }
 
 func (d *WorkflowRunner) validateSetCurrent(args *deploymentspb.SetCurrentVersionArgs) error {
-	if d.State.CurrentVersion != args.Version {
+	if d.State.RoutingInfo.CurrentVersion != args.Version {
```
carlydf, line 120: Nit: I think it's more idiomatic in Go, and more readable, to structure the function like
```
func foo(...) error {
   if errorCondition {
      return err
   }
   return nil
}
```

Returning success at the end and errors in the middle makes it easier to read. The section "A longer example" in https://8thlight.com/insights/exploring-error-handling-patterns-in-go this article also calls this out with another example.

service/worker/workerdeployment/client.go
```diff
@@ -539,7 +588,7 @@ func (d *ClientImpl) AddVersionToWorkerDeployment(
 	outcome, err := d.updateWithStart(
 		ctx,
 		namespaceEntry,
-		WorkerDeploymentVersionWorkflowType,
+		WorkerDeploymentWorkflowType,
```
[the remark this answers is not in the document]
Shivs11: I know - I wondered on this too....

service/worker/workerdeployment/version_workflow.go
Shivs11: ah sorry - we were adding the same version twice to the worker-deployment workflow (we were doing this twice)!

service/frontend/workflow_handler.go
```diff
@@ -3380,6 +3380,55 @@ func (wh *WorkflowHandler) SetWorkerDeploymentRampingVersion(ctx context.Context
...
+	if wh.config.DisableListVisibilityByFilter(namespaceEntry.Name().String()) {
+		return nil, errListNotAllowed
+	}
+
+	maxPageSize := int32(wh.config.VisibilityMaxPageSize(request.GetNamespace()))
+	if request.GetPageSize() <= 0 || request.GetPageSize() > maxPageSize {
+		request.PageSize = maxPageSize
+	}
+
+	resp, nextPageToken, err := wh.workerDeploymentClient.ListWorkerDeployments(ctx, namespaceEntry, int(request.PageSize), request.NextPageToken)
```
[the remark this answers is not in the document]
Shivs11: I think this was a design decision which was followed since the pre-release days and I didn't wanna alter it - fwiw, I agree with you - I shall keep things as is for now but not repeat the same thing in future

service/worker/workerdeployment/util.go
Shivs11: oops, yeah you are right - changing this to be called `WorkerDeploymentNamespaceDivision`

service/worker/workerdeployment/version_workflow.go
carlydf: oh I see the duplicate code above -- got it

carlydf (approved):

PR #7175, opened 2025-01-29
Refactor: use Eventually in xdc tests

alexshtin:
## What changed?

WIP


alexshtin:
#7243 is a better version of this PR.

PR #7176, opened 2025-01-29
add gauge metrics for gc count and pause

alfred-landrum:
## What changed?
Add 2 new metrics reporting runtime memory information for the number of GC's since program start, and the total time spent in GC pauses.

## Why?
While there were some existing metrics for these values, they were reported as a histogram and timer, and interpretation of them is complex. Reporting these values without any extra processing will make using them easier.

## How did you test it?
Ran locally & verified that these metrics were present via a local prometheus setup.

## Is hotfix candidate?
No.


stephanos (approved):
LGTM

PR #7177, opened 2025-01-29
handle state machine deletion for state-based replication

hai719:
## What changed?
Handle state machine deletion for state-based replication.

## Why?
Deletion is needed as we want to provide support for more nexus operations.

## How did you test it?
unit test.


service/history/workflow/mutable_state_impl.go
```diff
@@ -4264,6 +4264,29 @@ func (ms *MutableStateImpl) AddWorkflowExecutionUpdateAdmittedEvent(request *upd
 	return event, nil
 }
 
+func (ms *MutableStateImpl) DeleteSubStateMachine(path *persistencespb.StateMachinePath) error {
+	incomingPath := []hsm.Key{}
```
bergundy, line 4268:
```suggestion
	incomingPath := make([]hsm.Key, len(path))
```

And then you can just assign with the index.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4264,6 +4264,29 @@ func (ms *MutableStateImpl) AddWorkflowExecutionUpdateAdmittedEvent(request *upd
...
+		}
+		ms.logError(
+			fmt.Sprintf("unable to find path: %v in subStateMachine", incomingPath),
+			tag.ErrorTypeInvalidMutableStateAction,
+		)
+		// log data inconsistency instead of returning an error
+		ms.logDataInconsistency()
+		return nil
+	}
+	return root.DeleteChild(node.Key)
```
bergundy, line 4287: You'd need to find the parent of the node to delete the immediate child.

service/history/workflow/mutable_state_impl.go
```diff
@@ -7265,41 +7288,8 @@ func (ms *MutableStateImpl) syncExecutionInfo(current *persistencespb.WorkflowEx
 }
 
 func (ms *MutableStateImpl) syncSubStateMachinesByType(incoming map[string]*persistencespb.StateMachineMap) error {
```
bergundy, line 7290: @yycptt do you see an issue with this? I assume that we always just want to do a full sync here.

yycptt: hmm are we still using the `TransitionCount` field in StateMachineNode? If no, then I think this should work.

We can also remove the `PopTask()` logic in workflowStateReplicator.applySnapshot I think as we are no longer calling node.Sync which regenerates the tasks.

service/history/workflow/mutable_state_impl.go
```diff
@@ -7331,6 +7321,8 @@ func (ms *MutableStateImpl) applyTombstones(tombstoneBatches []*persistencespb.S
 				if _, ok := ms.pendingSignalInfoIDs[tombstone.GetSignalExternalInitiatedEventId()]; ok {
 					err = ms.DeletePendingSignal(tombstone.GetSignalExternalInitiatedEventId())
 				}
+			case *persistencespb.StateMachineTombstone_StateMachinePath:
+				err = ms.DeleteSubStateMachine(tombstone.GetStateMachinePath())
```
yycptt, line 7325: also need to `TrimStateMachineTimers` when nodes are being deleted.

I have a TODO in task_refresher.go about this (~L660). Probably need to find a way to trigger the trim only once.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4264,6 +4264,29 @@ func (ms *MutableStateImpl) AddWorkflowExecutionUpdateAdmittedEvent(request *upd
...
+	}
+
+	root := ms.HSM()
+	node, err := root.Child(incomingPath)
+	if err != nil {
+		if !errors.Is(err, hsm.ErrStateMachineNotFound) {
+			return err
+		}
+		ms.logError(
+			fmt.Sprintf("unable to find path: %v in subStateMachine", incomingPath),
```
yycptt, line 4280: hmm I think this could happen? If the mutation being synced may have a start point older than the workflow's current versioned transition. So some tombstones have already been synced before. 

well I guess I just can't prove that situation won't happen... 🤷‍♂️ 

It's just printing logs so no strong opinion.

service/history/workflow/mutable_state_impl.go
hai719: when source cluster is generating snapshot, the TransitionCount is reset to 1.

hai719: From the code, looks like `TransitionCount` is checked in `ValidateNotTransitioned`, and `ValidateNotTransitioned` is used only in components/dummy and hsmtest.

@bergundy can you confirm?

bergundy: We don't use `TransitionCount` anymore. It's only used in unit tests.

service/history/workflow/mutable_state_impl.go
bergundy: We need to trim every time when we close a transaction, @justinp-tt is working on it.

bergundy:
@hai719 note that you'll need to revert https://github.com/temporalio/temporal/pull/7163 and uncomment this TODO: https://github.com/temporalio/temporal/blob/e4cfc5af8a952668b6f157688bedcdf4b7bfee29/tests/xdc/nexus_state_replication_test.go#L75 for state based replication and deletion to work.

service/history/workflow/mutable_state_impl.go
hai719: add a flag to track if subStateMachine is deleted.

service/history/workflow/mutable_state_impl.go
hai719: updated

service/history/workflow/mutable_state_impl.go
hai719: updated

service/history/workflow/mutable_state_impl.go
hai719: updated.

service/history/workflow/mutable_state_impl.go
hai719: PopTask() removed.

hai719:
> @hai719 note that you'll need to revert #7163 and uncomment this TODO:
> 
> https://github.com/temporalio/temporal/blob/e4cfc5af8a952668b6f157688bedcdf4b7bfee29/tests/xdc/nexus_state_replication_test.go#L75
> 
> for state based replication and deletion to work.

enabled node deletion for state based replication and enabled tests.

components/nexusoperations/events.go
```diff
@@ -306,12 +306,5 @@ func findOperationNode(root *hsm.Node, event *historypb.HistoryEvent) (*hsm.Node
 }
 
 func maybeDeleteNode(node *hsm.Node) error {
```
bergundy, line 308: You can get rid of this function entirely, it's just confusing now.

service/history/workflow/mutable_state_impl.go
```diff
@@ -7351,41 +7375,8 @@ func (ms *MutableStateImpl) syncExecutionInfo(current *persistencespb.WorkflowEx
...
-			return err
-		}
-
-		return currentNode.Sync(incomingNode)
-	}); err != nil {
-		return err
-	}
-
+	ms.executionInfo.SubStateMachinesByType = incoming
+	ms.mustInitHSM()
```
bergundy, line 7379: Just please make sure that we regenerate the relevant timers from mutable state and delete any timers for state machines that were deleted during the full sync.
The important thing to note here is that the state machine timer group has a "scheduled" flag that indicates that there's a corresponding task in the queue, we want to make sure we don't regenerate the already scheduled tasks.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4335,6 +4336,29 @@ func (ms *MutableStateImpl) AddWorkflowExecutionUpdateAdmittedEvent(request *upd
...
+			return err
+		}
+		// node is already deleted.
+		return nil
+	}
+	err = node.Parent.DeleteChild(node.Key)
+	if err != nil {
+		return err
+	}
+	ms.subStateMachineDeleted = true
```
bergundy, line 4358: I don't see this being reset anywhere.

hai719: reset in cleanupTransaction()

service/history/workflow/mutable_state_impl.go
hai719: updated as discussed offline.

components/nexusoperations/events.go
hai719: removed.

service/history/workflow/state_machine_timers.go
```diff
@@ -36,10 +37,16 @@ import (
 // AddNextStateMachineTimerTask generates a state machine timer task if the first deadline doesn't have a task scheduled
 // yet.
 func AddNextStateMachineTimerTask(ms MutableState) {
-	timers := ms.GetExecutionInfo().StateMachineTimers
+	// filter out empty timer groups
+	timers := util.FilterSlice(ms.GetExecutionInfo().StateMachineTimers, func(timerGroup *persistencespb.StateMachineTimerGroup) bool {
```
bergundy, line 41: nit: I think you can use `slices.DeleteFunc` here but not critical.

bergundy (approved):

service/history/workflow/state_machine_timers.go
hai719: updated

PR #7178, opened 2025-01-29
Address Drainage PR comments: Pass RegisterVersion info to task queue, fix comments

carlydf:


proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -36,9 +36,6 @@ import "temporal/api/common/v1/message.proto";
 // The source of truth for this data is in the WorkerDeployment entity workflows, which is
 // synced to all TQs whenever the source changes.
 message DeploymentVersionData {
-    // Special case: `version.version` may be empty (but not `version.deployment_name`). That is
-    // only valid when `ramp_percentage` is non-zero, representing ramping to unversioned in a
-    // particular worker deployment.
```
carlydf, line 41: empty version is no longer a special case because now we have the `__unversioned__` special build id string to represent unversioned

Shivs11 (approved):

PR #7183, opened 2025-01-29
ramp - Versioning:3.1

Shivs11:
## What changed?

- Ramping PR for versioning-3.1
- now, if a ramping version becomes current, the ramp will be unset.

## Why?

- Versioning-3.1

## How did you test it?

- Added more functional tests
- Existing suite

## Potential risks

- None, going to feature.

## Is hotfix candidate?

No


service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,96 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+		}
+
+		if _, err := d.syncVersion(ctx, prevRampingVersion, rampUpdateArgs); err != nil {
+			return nil, err
+		}
+	} else {
+		// tell new ramping version it's ramping
+		rampUpdateArgs := &deploymentspb.SyncVersionStateUpdateArgs{
+			RoutingUpdateTime: updateTime,
+			RampingSinceTime:  updateTime,
```
carlydf, line 183: RampingSinceTime is a bit weird, because if the ramp % is changing but the ramping version is staying the same, RampingSinceTime will not change, because the time when the version started ramping is still the same.

I would make a variable called `newRampingSinceTime` and do:
```
newRampingSinceTime = updateTime // if this version is newly ramping
if prevRampingVersion == newRampingVersion { // this version was already ramping
  newRampingSinceTime = d.State.RoutingInfo.RampingVersionUpdateTime
}
```

service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,96 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+			if _, err := d.syncVersion(ctx, prevRampingVersion, prevRampUpdateArgs); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	// update local state
+	d.State.RoutingInfo.RampingVersion = newRampingVersion
+	d.State.RoutingInfo.RampingVersionPercentage = args.Percentage
+	d.State.RoutingInfo.RampingVersionUpdateTime = updateTime
```
carlydf, line 206: If the RampingVersion stayed the same, `d.State.RoutingInfo.RampingVersionUpdateTime` should not change

carlydf: possibly update `d.State.RoutingInfo.RampingVersionPercentageUpdateTime`

carlydf (commented):
really close! just left a few comments

service/frontend/workflow_handler.go
```diff
@@ -3377,7 +3380,48 @@ func (wh *WorkflowHandler) SetWorkerDeploymentCurrentVersion(ctx context.Context
...
+	}
+
+	if request.GetBuildId() == "" {
+		if request.GetPercentage() != 0 {
+			return nil, serviceerror.NewInvalidArgument("Empty value for build_id must be paired with percentage=0")
+		}
+	}
+
+	if request.GetPercentage() < 0 || request.GetPercentage() > 100 {
+		return nil, serviceerror.NewInvalidArgument("Percentage must be between 0 and 100")
```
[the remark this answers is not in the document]
Shivs11: yes, good point

service/worker/workerdeployment/util.go
```diff
@@ -47,10 +47,11 @@ const (
...
 
 	// Updates
-	RegisterWorkerInDeployment   = "register-task-queue-worker"       // for Worker Deployment Version wf
-	SyncVersionState             = "sync-version-state"               // for Worker Deployment Version wfs
-	SetCurrentVersion            = "set-current-version"              // for Worker Deployment wfs
-	AddVersionToWorkerDeployment = "add-version-to-worker-deployment" // for Worker Deployment wfs
+	RegisterWorkerInDeployment        = "register-task-queue-worker"            // for Worker Deployment Version wf
+	SyncVersionState                  = "sync-version-state"                    // for Worker Deployment Version wfs
+	SetCurrentVersion                 = "set-current-version"                   // for Worker Deployment wfs
+	SetWorkerDeploymentRampingVersion = "set-worker-deployment-ramping-version" // for Worker Deployment wfs
```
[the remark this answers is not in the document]
Shivs11: Agreed - don't mind shortening this to remain consistent with our other variables

service/worker/workerdeployment/client.go
```diff
@@ -376,6 +384,65 @@ func (d *ClientImpl) SetCurrentVersion(
...
+
+	var res deploymentspb.SetWorkerDeploymentRampingVersionResponse
+	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errNoChangeType {
+		res.PreviousVersion = version.GetBuildId()
+		res.PreviousPercentage = percentage
+		return &res, nil
+	} else if failure.GetApplicationFailureInfo().GetType() == errVersionAlreadyCurrentType {
+		res.PreviousVersion = ""
+		res.PreviousPercentage = 0
+		return &res, nil
```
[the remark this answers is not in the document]
Shivs11: hmm, do we? I ask because in the case of setCurrent, setting the same version current doesn't error out and is idempotent - moreover, we also don't return an error to the user in that case and let them think as if it's successful

I think we should stay consistent in both cases - wdyt?

service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,96 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+		d.lock.Unlock()
+	}()
+
+	prevRampingVersion := d.State.RoutingInfo.RampingVersion
+	prevRampingVersionPercentage := d.State.RoutingInfo.RampingVersionPercentage
+
+	newRampingVersion := args.Version
+	updateTime := timestamppb.New(workflow.Now(ctx))
+
+	// unset currently set ramping version
```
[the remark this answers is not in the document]
Shivs11: I don't want that to come in this PR only because I like PR's to only revolve around one change - but once again thank you for pointing this out! I'll make a TODO here, make a ticket for myself, and also quickly follow-up!

service/worker/workerdeployment/workflow.go
Shivs11: In my latest commit, `d.State.RoutingInfo.RampingVersionUpdateTime` now works correctly - I have not made the addition to percentage *yet* due to a opinion I had which I pasted on our slack thread

service/worker/workerdeployment/workflow.go
```diff
@@ -281,6 +279,13 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
 	d.State.RoutingInfo.CurrentVersion = args.Version
 	d.State.RoutingInfo.CurrentVersionUpdateTime = updateTime
 
+	// unset ramping version if it was set to current version
+	if d.State.RoutingInfo.CurrentVersion == d.State.RoutingInfo.RampingVersion {
+		d.State.RoutingInfo.RampingVersion = ""
+		d.State.RoutingInfo.RampingVersionPercentage = 0
+		d.State.RoutingInfo.RampingVersionUpdateTime = nil
```
carlydf, line 286: this should be `updateTime` not nil

tests/worker_deployment_test.go
```diff
@@ -572,6 +572,36 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_WithCurren
...
+		DeploymentName: tv.DeploymentSeries(),
+	})
+	s.NoError(err)
+	s.verifyDescribeWorkerDeployment(resp, &workflowservice.DescribeWorkerDeploymentResponse{
+		WorkerDeploymentInfo: &deploymentpb.WorkerDeploymentInfo{
+			Name: tv.DeploymentSeries(),
+			RoutingInfo: &deploymentpb.RoutingInfo{
+				RampingVersion:           "",  // no ramping info should be set
+				RampingVersionPercentage: 0,   // no ramping info should be set
+				RampingVersionUpdateTime: nil, // no ramping info should be set
```
carlydf, line 597: this should be the same as `CurrentVersionUpdateTime`, not nil

service/worker/workerdeployment/util.go
```diff
@@ -47,10 +47,11 @@ const (
...
 
 	// Updates
-	RegisterWorkerInDeployment   = "register-task-queue-worker"       // for Worker Deployment Version wf
-	SyncVersionState             = "sync-version-state"               // for Worker Deployment Version wfs
-	SetCurrentVersion            = "set-current-version"              // for Worker Deployment wfs
-	AddVersionToWorkerDeployment = "add-version-to-worker-deployment" // for Worker Deployment wfs
+	RegisterWorkerInDeployment   = "register-task-queue-worker"            // for Worker Deployment Version wf
+	SyncVersionState             = "sync-version-state"                    // for Worker Deployment Version wfs
+	SetCurrentVersion            = "set-current-version"                   // for Worker Deployment wfs
+	SetRampingVersion            = "set-worker-deployment-ramping-version" // for Worker Deployment wfs
```
carlydf, line 53: Nit: maybe make the string be "set-ramping-version" to match "set-current-version" above

service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,105 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+
+		rampUpdateArgs := &deploymentspb.SyncVersionStateUpdateArgs{
+			RoutingUpdateTime: routingUpdateTime,
+			RampingSinceTime:  nil, // remove ramp
+			RampPercentage:    0,   // remove ramp
+		}
+
+		if _, err := d.syncVersion(ctx, prevRampingVersion, rampUpdateArgs); err != nil {
+			return nil, err
+		}
```
carlydf, line 180: Nit: You could put all of these `d.syncVersion(ctx, prevRampingVersion, rampUpdateArgs)` calls outside of the if-else block, and just use the if-else block to construct the appropriate `rampUpdateArgs`

service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,105 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+	} else {
+		// setting ramp
+
+		if prevRampingVersion == newRampingVersion { // the version was alread ramping, user changing ramp %
+			rampingSinceTime = d.State.RoutingInfo.RampingVersionUpdateTime
+		} else {
+			rampingSinceTime = timestamppb.New(workflow.Now(ctx)) // version ramping for the first time
```
carlydf, line 187:
```suggestion
			rampingSinceTime = routingUpdateTime // version ramping for the first time
```

service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,105 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+		if prevRampingVersion != "" && prevRampingVersion != newRampingVersion {
+			prevRampUpdateArgs := &deploymentspb.SyncVersionStateUpdateArgs{
+				RoutingUpdateTime: routingUpdateTime,
+				RampingSinceTime:  nil, // remove ramp
+				RampPercentage:    0,   // remove ramp
+			}
+			if _, err := d.syncVersion(ctx, prevRampingVersion, prevRampUpdateArgs); err != nil {
+				return nil, err
+			}
+		}
```
carlydf, line 209: if someone is removing the ramp (`if newRampingVersion == "" {`), you still need to tell the previous ramping version that it's no longer ramping, so I think this needs to go outside of the `else` block

service/worker/workerdeployment/workflow.go
```diff
@@ -116,6 +127,105 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+	}
+
+	// update local state
+	d.State.RoutingInfo.RampingVersion = newRampingVersion
+	d.State.RoutingInfo.RampingVersionPercentage = args.Percentage
+	d.State.RoutingInfo.RampingVersionUpdateTime = rampingSinceTime
```
carlydf, line 215:
```suggestion
	d.State.RoutingInfo.RampingVersionUpdateTime = rampingSinceTime // this is still nil if ramp was unset
```
Nit: Suggesting to add this comment only because I had to re-read the code to convince myself of this. A comment could save future readers the trouble

PR #7185, opened 2025-01-30
use transition history if enabled when long poll history

hai719:
## What changed?
use transition history if enabled when long poll history

## Why?
For state-based replication, we do allow event version version to go backwards as long as the overall versioned transition is keep increasing. So we need to change token to use versioned transition.

## How did you test it?
unit test.


service/history/api/get_workflow_util.go
```diff
@@ -147,8 +166,19 @@ func GetOrPollMutableState(
...
+			if request.VersionedTransition != nil && workflow.TransitionHistoryStalenessCheck(transitionHistory, request.VersionedTransition) != nil {
+				logger.Warn(fmt.Sprintf("Request versioned transition and transition history prior to polling the mutable state. Request: %v, current: %v",
+					request.VersionedTransition,
+					transitionHistory[len(transitionHistory)-1]),
+					tag.WorkflowNamespaceID(workflowKey.GetNamespaceID()),
+					tag.WorkflowID(workflowKey.GetWorkflowID()),
+					tag.WorkflowRunID(workflowKey.GetRunID()))
+				return nil, serviceerrors.NewCurrentBranchChanged(response.CurrentBranchToken, request.CurrentBranchToken)
+			}
+		} else if !versionhistory.ContainsVersionHistoryItem(currentVersionHistory, request.VersionHistoryItem) {
```
yycptt, line 181: hmm shall we fallback to version history check when either `transitionHistory` or `request.VersionedTransition` is empty? 
Looks right now, if `transitionHistory` is no empty but `request.VersionedTransition` is empty, we will just skip the validation priori to polling mutable state.

service/history/api/get_workflow_util.go
```diff
@@ -199,8 +230,18 @@ func GetOrPollMutableState(
...
+					if request.VersionedTransition != nil && workflow.TransitionHistoryStalenessCheck(transitionHistory, request.VersionedTransition) != nil {
+						logger.Warn(fmt.Sprintf("Request versioned transition and transition history don't match after polling the mutable state. Request: %v, current: %v",
+							request.VersionedTransition,
+							transitionHistory[len(transitionHistory)-1]),
+							tag.WorkflowNamespaceID(workflowKey.GetNamespaceID()),
+							tag.WorkflowID(workflowKey.GetWorkflowID()),
+							tag.WorkflowRunID(workflowKey.GetRunID()))
+						return nil, serviceerrors.NewCurrentBranchChanged(response.CurrentBranchToken, request.CurrentBranchToken)
+					}
+				} else if !versionhistory.ContainsVersionHistoryItem(eventVersionHistory, request.VersionHistoryItem) {
```
yycptt, line 244: Same Q here.

service/history/events/notifier.go
```diff
@@ -115,6 +119,7 @@ func NewNotification(
 		WorkflowState:          workflowState,
 		WorkflowStatus:         workflowStatus,
 		VersionHistories:       versionhistory.CopyVersionHistories(versionHistories),
+		TransitionHistory:      transitionhistory.CopyVersionedTransitions(transitionHistory),
```
yycptt, line 122: we are copy versionHistories and transitionHistory for each update even if there's no subscriber on those events :(

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -86,29 +88,38 @@ func Invoke(
...
 		if err != nil {
-			return nil, "", 0, 0, false, nil, err
+			return nil, "", 0, 0, false, nil, nil, err
 		}
+
+		var lastVersionedTransition *persistencespb.VersionedTransition
+		transitionHistory := response.GetTransitionHistory()
+		if transitionHistory != nil {
+			lastVersionedTransition = transitionHistory[len(transitionHistory)-1]
+		}
```
yycptt, line 114: nit: make this a util function as well for getting the last versioned transition? I think I have seen this logic in couple of places.

service/history/events/notifier.go
hai719: > we are copy versionHistories and transitionHistory for each update even if there's no subscriber on those events :(

We have to make a copy to avoid data race.

service/history/api/get_workflow_util.go
yux0: I have the same question on if this is a valid case on `transitionHistory is no empty but request.VersionedTransition is empty`

yycptt: it will happen when we start to rollout state based replication so we have to handle it. I think we can just validate version history in that case.

service/history/api/get_workflow_util.go
```diff
@@ -147,8 +166,19 @@ func GetOrPollMutableState(
 		if err != nil {
 			return nil, err
 		}
-		// TODO: update to use transition version history
-		if !versionhistory.ContainsVersionHistoryItem(currentVersionHistory, request.VersionHistoryItem) {
+
+		transitionHistory := response.GetTransitionHistory()
+		if len(transitionHistory) != 0 {
+			if request.VersionedTransition != nil && workflow.TransitionHistoryStalenessCheck(transitionHistory, request.VersionedTransition) != nil {
+				logger.Warn(fmt.Sprintf("Request versioned transition and transition history prior to polling the mutable state. Request: %v, current: %v",
```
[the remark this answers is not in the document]
hai719: updated

service/history/api/get_workflow_util.go
hai719: updated

service/history/api/get_workflow_util.go
hai719: updated

service/history/api/getworkflowexecutionhistory/api.go
hai719: created a function to get the last versioned transition.

service/history/api/get_workflow_util.go
```diff
@@ -274,7 +316,14 @@ func GetMutableStateWithConsistencyCheck(
 		ctx,
 		nil,
 		func(mutableState workflow.MutableState) bool {
-			mutableState.GetExecutionInfo().GetVersionHistories()
+			transitionHistory := mutableState.GetExecutionInfo().GetTransitionHistory()
+			if len(transitionHistory) != 0 {
+				if versionedTransition != nil && workflow.TransitionHistoryStalenessCheck(transitionHistory, versionedTransition) != nil {
+					return false
+				}
+				return true
```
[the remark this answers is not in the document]
hai719: updated

service/history/api/get_workflow_util.go
```diff
@@ -99,6 +102,20 @@ func GetOrPollMutableState(
...
+	transitionHistory := response.GetTransitionHistory()
+	if len(transitionHistory) != 0 && request.VersionedTransition != nil {
+		if workflow.TransitionHistoryStalenessCheck(transitionHistory, request.VersionedTransition) != nil {
+			logger.Warn(fmt.Sprintf("Request versioned transition and transition history don't match. Request: %v, current: %v",
+				request.VersionedTransition,
+				transitionhistory.LastVersionedTransition(transitionHistory)),
+				tag.WorkflowNamespaceID(workflowKey.GetNamespaceID()),
+				tag.WorkflowID(workflowKey.GetWorkflowID()),
+				tag.WorkflowRunID(workflowKey.GetRunID()))
+			return nil, serviceerrors.NewCurrentBranchChanged(response.CurrentBranchToken, request.CurrentBranchToken)
```
yycptt, line 115: nit: we probably should add extra into to the CurrentBatchChanged error as a follow up. Branch token might be the same when transition history diverges and that could cause confusion.

yycptt (approved):

service/history/api/getworkflowexecutionhistory/api.go
yycptt: Thank you!

service/history/api/get_workflow_util.go
hai719: added transition history info to the error.

PR #7186, opened 2025-01-30
Add tracing logs for Nexus HTTP request retries

pdoerner:
## What changed?

Added additional HTTP tracing logs for the first 3 Nexus request retries.
Both minimum and maximum attempts to add additional tracing info to are configurable.

## Why?

To aid in debugging.


common/dynamicconfig/constants.go
```diff
@@ -865,6 +865,18 @@ used when the first cache layer has a miss. Requires server restart for change t
...
+		"system.nexusHTTPTraceMinAttempt",
+		2,
+		`The minimum attempt of a Nexus request which will log additional HTTP request tracing information.
+WARNING: setting to 0 or 1 will log additional tracing information for ALL Nexus HTTP requests and may be expensive.`,
+	)
+	NexusHTTPTraceMaxAttempt = NewNamespaceIntSetting(
+		"system.nexusHTTPTraceMaxAttempt",
+		5,
+		`The maximum attempt of a Nexus request which will log additional HTTP request tracing information.
+Set to 0 to disable tracing.`,
```
bergundy, line 879: This should be disabled by default. Generally, I don't think the attempt number matters here, it's more about what to trace IMHO. E.g. whether `TLSHandshakeStart` should be traced or not. Ideally the config would be for a set of methods, but It'd be hard to express in dynamic config as a set without adding a lot of overhead. We could probably provide a config per method we support intercepting though.

pdoerner: I think attempt count is useful for limiting the overhead and the overall log spam. Without a notion of attempt count do you imagine we would log trace info for every retry?

Instead of adding a dynamic config per-hook, we could just use fx to provide the `httptrace.ClientTrace`. Then anyone who wants to change the behavior could do so with fx.replace

bergundy: Discussed offline, agree that adding the counts is nice, also requested that we add the dynamic config and make it global so it can be efficiently cached (a limitation of the current DC implementation).

common/nexus/trace.go
```diff
@@ -0,0 +1,161 @@
...
+
+type HTTPClientTraceProvider interface {
+	// NewTrace returns a *httptrace.ClientTrace which provides hooks to invoke at each point in the HTTP request
+	// lifecycle. This trace must be added to the HTTP request context using httptrace.WithClientTrace for the hooks to
+	// be invoked. The provided logger should already be tagged with relevant request information
+	// e.g. using log.With(logger, tag.RequestID(id), tag.Operation(op), ...).
+	NewTrace(attempt int32, logger log.Logger) *httptrace.ClientTrace
+}
+
+var HTTPTraceConfig = dynamicconfig.NewGlobalTypedSetting(
```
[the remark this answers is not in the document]
pdoerner: Synced with @dnr offline and changed the conversion functions to the best we can do at the moment. We will want to revisit this setting after merging https://github.com/temporalio/temporal/pull/7052

PR #7187, opened 2025-01-30
Delete APIs implementation

carlydf:


service/worker/workerdeployment/version_workflow.go
```diff
@@ -161,17 +173,21 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
 	workflow.Go(ctx, d.listenToSignals)
 
 	// Wait on any pending signals and updates.
-	err := workflow.Await(ctx, func() bool { return d.pendingUpdates == 0 && d.signalsCompleted })
+	err := workflow.Await(ctx, func() bool { return (d.pendingUpdates == 0 && d.signalsCompleted) || d.done })
```
carlydf, line 176: this might be wrong

service/worker/workerdeployment/client.go
```diff
@@ -97,6 +94,21 @@ type Client interface {
...
+		namespaceEntry *namespace.Namespace,
+		deploymentName string,
+		identity string,
+	) error
+
+	DeleteWorkerDeploymentVersion(
+		ctx context.Context,
+		namespaceEntry *namespace.Namespace,
+		deploymentName string,
+		buildId string,
```
Shivs11, line 108: lets pass in `&deploymentpb.WorkerDeploymentVersion` - it was in our one of our TODO's earlier too and I passed this while working on the Ramp API's

service/worker/workerdeployment/client.go
```diff
@@ -376,6 +388,94 @@ func (d *ClientImpl) SetCurrentVersion(
 	return &res, nil
 }
 
+func (d *ClientImpl) DeleteWorkerDeployment(
+	ctx context.Context,
+	namespaceEntry *namespace.Namespace,
+	deploymentName string,
+	identity string,
+) (retErr error) {
+	// if len(deployment.versions) == nil, delete
```
Shivs11, line 397: stale comment?

service/worker/workerdeployment/version_workflow.go
Shivs11: curious why you think so - the logic right now seems to be that unless `d.done` is true, we wait for any pending updates and signals - if `d.done` is true, we don't care about the pending signals/updates and don't await

service/worker/workerdeployment/version_workflow.go
```diff
@@ -52,6 +52,7 @@ type (
 		pendingUpdates         int
 		signalsCompleted       bool
 		drainageWorkflowFuture *workflow.ChildWorkflowFuture
+		done                   bool
```
Shivs11, line 55: nit: `done` might seem hard to understand at first - wdyt about renaming this to something like `terminate/completed`? (may not be the best suggestion but just putting it out there)

service/worker/workerdeployment/workflow.go
```diff
@@ -47,6 +48,7 @@ type (
 		metrics        sdkclient.MetricsHandler
 		lock           workflow.Mutex
 		pendingUpdates int
+		done           bool
```
Shivs11, line 51: same renaming suggestion as the above comment

service/worker/workerdeployment/workflow.go
```diff
@@ -103,11 +105,38 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+			Validator: d.validateDeleteVersion,
+		},
+	); err != nil {
+		return err
+	}
+
 	// Wait until we can continue as new or are cancelled.
-	err = workflow.Await(ctx, func() bool { return workflow.GetInfo(ctx).GetContinueAsNewSuggested() && pendingUpdates == 0 })
+	err = workflow.Await(ctx, func() bool {
+		return (workflow.GetInfo(ctx).GetContinueAsNewSuggested() || d.done) && pendingUpdates == 0
```
Shivs11, line 132: I wonder if we should wait for any more pending updates if `d.done` is true - in other words, if we are ready to be deleted, should we be worrying about pending ongoing updates?????

Moreover, this logic is different from the await logic found in version_workflow - over there, we seem to be waiting for any pending ongoing updates even if `d.done` is true

service/worker/workerdeployment/workflow.go
```diff
@@ -103,11 +105,38 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+		ctx,
+		DeleteVersion,
+		d.handleDeleteVersion,
+		workflow.UpdateHandlerOptions{
+			Validator: d.validateDeleteVersion,
+		},
+	); err != nil {
+		return err
+	}
+
```
Shivs11, line 129: nice!

if you now notice, we are basically doing *almost everything* through the deployment workflow which goes through to the version workflow - the only thing we are doing the other way around is registering a task-queue worker

I know things work right now, but if we do have some time before the cut-off, lets change that too to make things consistent? I might as well just make a ticket

service/worker/workerdeployment/client.go
carlydf: removed :)

service/worker/workerdeployment/workflow.go
carlydf: I've made it now so that in the Deployment workflow the Await is like this:
```
err = workflow.Await(ctx, func() bool {
	return (workflow.GetInfo(ctx).GetContinueAsNewSuggested() || d.done) && pendingUpdates == 0
})
```

and in the Version workflow the Await is like this:
```
err := workflow.Await(ctx, func() bool { return (d.signalsCompleted || d.done) && d.pendingUpdates == 0 })
```
I think those match / make sense, but also it's late, so please help me sanity check that this is what we want

service/worker/workerdeployment/client.go
carlydf: Resolving this because of that other thread about when to pass Version object vs the two strings

service/worker/workerdeployment/version_activities.go
```diff
@@ -141,6 +144,33 @@ func (a *VersionActivities) CheckWorkerDeploymentUserDataPropagation(ctx context
...
+		if len(typesInfo[int32(enumspb.TASK_QUEUE_TYPE_WORKFLOW)].GetPollers()) > 0 {
+			return true, nil
+		}
+		if len(typesInfo[int32(enumspb.TASK_QUEUE_TYPE_ACTIVITY)].GetPollers()) > 0 {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
```
Shivs11, line 173: this is okay in here too but an idea that I had was placing this inside the deployment client, since the client serves as a nice bridge between the matchingEngine and the entity workflows

I shall be implementing the describeCalls for poller presence over there so shall take care of the fx stuff - I think this should go there but lmk what you think!

service/worker/workerdeployment/workflow.go
Shivs11: I agree that they should match but what I was maybe hinting/asking was if we should wait for signals/updates to be processed when we know we are `done`.

Now that I think about it, the answer lies in this question:

Will there ever be a scenario where the version workflow changes from being ready to be deleted (d.done is true) to suddenly ready to accept new executions - I think the answer is no, so IMO we should not wait for any pending updates/signals when d.done is true and just blatantly cancel execution

wdyt?

something like:

```
 err := workflow.Await(ctx, func() bool { return (d.signalsCompleted && d.pendingUpdates == 0) || d.done })

and

err = workflow.Await(ctx, func() bool {
	return (workflow.GetInfo(ctx).GetContinueAsNewSuggested() && pendingUpdates == 0) || d.done)
})
 
 ```

File: service/worker/workerdeployment/version_workflow.go (hunk 4 above)
Comment (carlydf) [reply service/worker/workerdeployment/version_workflow.go:56]: I chose `done` inspired by `ctx.Done()`, does that help it make sense to you at all?

File: service/worker/workerdeployment/workflow.go (hunk 6 above)
Comment (carlydf) [reply service/worker/workerdeployment/workflow.go:143]: Hm I think you're right, because the pending updates that I wanted to wait for would basically just be a SetCurrent/SetRamping. However, if the version is ready to be deleted, we know it is drained and has no pollers. If it's drained, we know it's not receiving any more signals from the drainage child workflow. And if it has no pollers, it can't be SetCurrent or SetRamping.

Comment (carlydf) [reply service/worker/workerdeployment/workflow.go:143]: changing both to be `(d.signalsCompleted && d.pendingUpdates == 0) || d.done` :)

File: service/worker/workerdeployment/workflow.go (hunk 5 above)
Comment (carlydf) [reply service/worker/workerdeployment/workflow.go:51]: see above

File: service/worker/workerdeployment/version_activities.go (hunk 8 above)
Comment (Shivs11) [reply service/worker/workerdeployment/version_activities.go:173]: hmm, good point I seemed to have missed that - sure, no worries we can skip this then 

btw, for now and in the interest of time, I implemented my describe call in the client but may change that and get that inside activities to just be consistent

File: service/worker/workerdeployment/version_workflow.go (hunk 4 above)
Comment (Shivs11) [reply service/worker/workerdeployment/version_workflow.go:56]: sure, was more of a nit sorry - no need to ponder over this

File: service/worker/workerdeployment/version_activities.go (hunk 8 above)
Comment (carlydf) [reply service/worker/workerdeployment/version_activities.go:173]: it's ok if it's not consistent as long as it works!

File: proto/internal/temporal/server/api/deployment/v1/message.proto [hunk 9]
```diff
@@ -229,6 +233,31 @@ message SetCurrentVersionResponse {
...
+    string identity = 1;
+    string deployment_name = 2;
+    string version = 3;
+    string request_id = 4;
+}
+
+// used as Worker Deployment Activity input:
+message CheckTaskQueuesHaveNoPollersActivityArgs {
+    repeated temporal.api.taskqueue.v1.TaskQueue task_queues = 1;
+    string build_id = 3;
```
Shivs11, line 253: nit: string build_id = 2;

service/worker/workerdeployment/client.go
```diff
@@ -476,6 +489,54 @@ func (d *ClientImpl) SetRampingVersion(
 	return &res, nil
 }
 
+func (d *ClientImpl) DeleteWorkerDeploymentVersion(
+	ctx context.Context,
+	namespaceEntry *namespace.Namespace,
+	deploymentName string,
+	buildId string,
+) (retErr error) {
+	// if version.drained and !version.has_pollers, delete
```
Shivs11, line 498: nit: comment

service/worker/workerdeployment/version_workflow.go
```diff
@@ -201,6 +218,96 @@ func (d *VersionWorkflowRunner) stopDrainage(ctx workflow.Context) error {
...
+	state := d.GetVersionState()
+
+	// describe all task queues in the deployment, if any have pollers, then cannot delete
+	var tqs []*taskqueuepb.TaskQueue
+	for tqName, _ := range state.TaskQueueFamilies {
+		tqs = append(tqs, &taskqueuepb.TaskQueue{
+			Name: tqName,
+			Kind: enumspb.TASK_QUEUE_KIND_NORMAL, // TODO (Carly): could this be sticky?
+		})
+	}
```
Shivs11, line 256: thinking out loud: 

`TaskQueueFamilies` is a map going from the task-queue -> DeploymentVersionData for *each* task-queue type of this task-queue

Say we were deleting a version `X` which only had the `TASK_QUEUE_TYPE_NEXUS` of a task-queue `foo`. Maybe version `Y` contains the `workflow type` of this `foo`.

This snippet of code seems to be ignoring the task-queue types in the Describe call since the Describe enhanced call, under the hood, will now return if all task-queue types for this task-queue has pollers or not

I am not sure if this is what we want - open to hearing your thoughts if I have been mistaken

Shivs11 (commented):
one important comment

service/worker/workerdeployment/version_workflow.go
carlydf: The DescribeTaskQueueEnhanced API only works on `TASK_QUEUE_TYPE_WORKFLOW`... but it also has the TaskQueueTypes field, which is maybe where we would want to pass the task queue type in that we have from this families map. Let me look at the code for DescribeTaskQueue to see what it does there

carlydf: ok so, for each type in `req.TaskQueueTypes`, for each partition within that type, we call `DescribeTaskQueuePartition`.

carlydf: Thanks for this comment, I forgot that technically an Activity Type task queue can be in a different deployment from a Workflow type task queue

PR #7194, opened 2025-01-31
[Reset] Add children_initialized_post_reset_point to WorkflowExecutionInfo for tracking child workflows post-reset

gow:
## What changed?
Added a new field `children_initialized_post_reset_point` to collect a set of children that need to be restarted after the parent is reset.

## Why?
We need to remember the set of children that need to be restarted after resetting a parent. So adding this field which will be populated in `workflowResetterImpl` (in the followup PR)

## How did you test it?
Existing tests. This is just the protochange.

## Potential risks
N/A

## Documentation
N/A

## Is hotfix candidate?
No


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -280,6 +280,11 @@ message WorkflowExecutionInfo {
     // Should be deprecated once the transition history is fully launched
     repeated VersionedTransition previous_transition_history = 100;
     VersionedTransition last_transition_history_break_point = 101;
+
+    // This is a set of child workflows that were initialized after the reset point in the parent workflow.
+    // The children are identified by the key "workflow_type:workflow_id". When the parent starts to make progress after reset, it uses this data to
```
yycptt, line 285: we don't have any restrictions on the char used in the workflowID & workflowType I think, so they could conflict. But i guess it doesn't matter too much in reality.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -280,6 +280,11 @@ message WorkflowExecutionInfo {
     // Should be deprecated once the transition history is fully launched
     repeated VersionedTransition previous_transition_history = 100;
     VersionedTransition last_transition_history_break_point = 101;
+
+    // This is a set of child workflows that were initialized after the reset point in the parent workflow.
```
yycptt, line 284: nit: add comment that this includes children in continue as new runs.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -280,6 +280,11 @@ message WorkflowExecutionInfo {
...
+    // The children are identified by the key "workflow_type:workflow_id". When the parent starts to make progress after reset, it uses this data to
+    // determine the right start policy to apply to the child.
+    map<string, bool> children_initialized_post_reset_point = 102;
```
yycptt, line 287: Is the `bool` here just a placeholder? 
In either cases, I'd create a separate proto message for this in case any new info needs to be stored later.

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: Yes. It should not be an issue since we are not parsing the value back. it's kind of like a hash. So even if they include special chars or `:` char, it should be ok.

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: I'm using the boolean indicator to track children that need to be restarted. Once restarted, I'm setting it false (but not deleting it). I plan to use the existence of false value to detect and ignore results from old child.
I agree that having a separate proto object for this is better (even if it contains a single bool field). Will update it.

proto/internal/temporal/server/api/persistence/v1/executions.proto
yycptt: oh I mean the hash could conflict for different workflowType and workflowID.

proto/internal/temporal/server/api/persistence/v1/executions.proto
yycptt: Ah I see. because we only support Restart option in the beginning, so it will always be set to `true` initially.

yycptt, line 287: Just calling out that this new field won't be replicated with today's replication stack. So the new feature will only work properly with in single dc setup

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: Added

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: ah! yes. There is a possibility of that. Maybe in future iterations we can build a good and signature.

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: Added `ResetChildInfo` message. It just contains a bool field (`should_terminate_and_start`) for now.

gow: Yes. I will be adding this and a few other fields (`reset_run`) to replication stack.

yycptt (approved):

PR #7198, opened 2025-01-31
Update Deployment Workflow ID's and accept string version values

Shivs11:
## What changed?

- Deployment BuildId's are no longer unique across a namespace. The combination of <DeploymentName, BuildID> will be.
- Constraints for not allowing "/" and "__" in deploymentName and buildID respectively
- We want the APIs to leave open the possibility of using a custom version id instead of deployment_name/build_id to set current or ramping. Also we want to accept the "__unversioned__" string without having to accep an "unversioned version" . To support this, refactored APIs to accept version strings. Refactored internal code to use build id string for build id only, and version string for fully-qualified version string only.

## Why?

- Versioning-3.1

## How did you test it?

- Existing suite of tests


carlydf (approved):

service/worker/workerdeployment/client.go
```diff
@@ -339,8 +349,15 @@ func (d *ClientImpl) SetCurrentVersion(
...
-	defer d.record("SetCurrentVersion", &retErr, namespaceEntry.Name(), deploymentName, version, identity)()
+	defer d.record("SetCurrentVersion", &retErr, namespaceEntry.Name(), version, identity)()
 	requestID := uuid.New()
+	versionObj, err := worker_versioning.WorkerDeploymentVersionFromString(version)
+	if err != nil {
+		return nil, serviceerror.NewInvalidArgument("invalid version string: " + err.Error())
+	}
+	if versionObj.GetDeploymentName() != deploymentName {
+		return nil, serviceerror.NewInvalidArgument(fmt.Sprintf("invalid version string '%s' does not match deployment name '%s'", version, deploymentName))
+	}
```
Shivs11, line 360: there's a chance, right now atleast, that `versionObj` could be `nil` (in the event that we deal with unversioned workers)

Either we change the code in physical_task_queue_manager to only deal with `VERSIONING_BEHAVIORS` workers and then this won't be a problem or safeguard against this

wdyt?

service/worker/workerdeployment/client.go
```diff
@@ -206,25 +207,34 @@ func (d *ClientImpl) RegisterTaskQueueWorker(
 func (d *ClientImpl) DescribeVersion(
 	ctx context.Context,
 	namespaceEntry *namespace.Namespace,
-	version string,
+	workerDeploymentVersion *deploymentpb.WorkerDeploymentVersion,
 ) (_ *deploymentpb.WorkerDeploymentVersionInfo, retErr error) {
+	deploymentName := workerDeploymentVersion.GetDeploymentName()
+	buildID := workerDeploymentVersion.GetBuildId()
```
Shivs11, line 213: we may need to have `nil` checks here (right now) just to safeguard but I don't think we do in the future since "unversioned" would just be present in a new field (`id`) inside `workerDeploymentVersion` - might be conservative programming rn but better to be safe than sorry

service/worker/workerdeployment/client.go
```diff
@@ -384,20 +402,37 @@ func (d *ClientImpl) SetCurrentVersion(
...
+) (_ *deploymentspb.SetRampingVersionResponse, retErr error) {
 	//revive:disable-next-line:defer
 	defer d.record("SetWorkerDeploymentRampingVersion", &retErr, namespaceEntry.Name(), version, percentage, identity)()
 	requestID := uuid.New()
+	var versionObj *deploymentpb.WorkerDeploymentVersion
+	var err error
+	if version == "" {
+		versionObj = &deploymentpb.WorkerDeploymentVersion{
+			DeploymentName: deploymentName,
+			BuildId:        "",
```
Shivs11, line 421: nit: took me a min to realize that this was unsetting the set ramp - maybe a comment would be nice

service/worker/workerdeployment/client.go
```diff
@@ -59,7 +59,7 @@ type Client interface {
 	RegisterTaskQueueWorker(
 		ctx context.Context,
 		namespaceEntry *namespace.Namespace,
-		deploymentName, version string,
+		deploymentName, buildId string,
```
Shivs11, line 62: if you notice, this is the only method which accepts a `build_id` rather than other methods now accepting `version` which will be the concatenation of deploymentName/build_id - is there any way we can make a change here for consistency purposes? 

More than consistency, it's so hard to follow now what to pass in where and almost always think if you are passing in the wrong version/build_id value - Maybe keeping things consistent could help us here?

Shivs11 (commented):
some comments

PR #7203, opened 2025-01-31
Add Conflict Tokens to Deployment APIs

carlydf:


service/frontend/workflow_handler.go
```diff
@@ -3514,14 +3514,7 @@ func (wh *WorkflowHandler) DescribeWorkerDeployment(ctx context.Context, request
...
 	}
 
-	workerDeploymentInfo, err := wh.workerDeploymentClient.DescribeWorkerDeployment(ctx, namespaceEntry, request.DeploymentName)
-	if err != nil {
-		return nil, err
-	}
-
-	return &workflowservice.DescribeWorkerDeploymentResponse{
-		WorkerDeploymentInfo: workerDeploymentInfo,
-	}, nil
```
Shivs11, line 3524: hmmm, can we keep this? The reason why I had introduced this is because I wanted the client to only accept/return internal deployment related protos and this breaks it

this isn't a blocker - just might be better design imo

service/worker/workerdeployment/workflow.go
```diff
@@ -66,7 +69,8 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 	if d.State == nil {
 		d.State = &deploymentspb.WorkerDeploymentLocalState{}
 		d.State.CreateTime = timestamppb.New(time.Now())
-		d.State.RoutingInfo = &deploymentpb.RoutingInfo{}
+		d.State.RoutingInfo = &deploymentpb.RoutingInfo{CurrentVersion: worker_versioning.UnversionedBuildId}
```
Shivs11, line 72: a bit confused about this actually and just want clarity:

- This would mean that the default current version for this worker-deployment, on-start, will be "unversioned"
- This would also mean that when users query make a DescribeWorkerDeployment call, they can see this value in the RoutingConfig


If the above two points are right, then we:

1. Have to make sure that the we have "__unversioned__" as the routing for these task-queues (Shahab's changes + discussion this morning now would mean we don't need to handle this specially)
2. `prevCurrentVersion` will never be "" now so we need to have changes there (not done)


I think it's best we get these changes in this PR - wdyt?

service/worker/workerdeployment/workflow.go
```diff
@@ -127,12 +131,15 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
...
+func (d *WorkflowRunner) validateSetRampingVersion(args *deploymentspb.SetRampingVersionArgs) error {
+	if args.ConflictToken != nil && !bytes.Equal(args.ConflictToken, d.State.ConflictToken) {
+		return temporal.NewApplicationError("conflict token mismatch", errConflictTokenMismatchType)
+	}
 	if args.Version == d.State.RoutingInfo.RampingVersion && args.Percentage == d.State.RoutingInfo.RampingVersionPercentage {
 		d.logger.Info("version already ramping, no change")
 		return temporal.NewApplicationError("version already ramping, no change", errNoChangeType)
 	}
-	// todo: this will only work when "__unversioned__" is the default current-version of a deployment.
+	// TODO (Shivam): this will only work when "__unversioned__" is the default current-version of a deployment.
```
Shivs11, line 142: thanks!

service/history/workflow/mutable_state_impl.go
```diff
@@ -4495,6 +4495,7 @@ func (ms *MutableStateImpl) updateVersioningOverride(
 				worker_versioning.DeploymentVersionFromDeployment(d))
 		}
 	} else if ms.GetExecutionInfo().GetVersioningInfo() != nil {
+		// TODO (shahab): this behavior has changed in main branch. Update it when merging to main.
```
carlydf, line 4498: not sure if this comment is still relevant

service/worker/workerdeployment/workflow.go
carlydf: I'm removing the unversioned thing from this PR

service/history/workflow/mutable_state_impl.go
ShahabT: it didn't seem it is so I removed it.

tests/worker_deployment_test.go
```diff
@@ -200,6 +200,72 @@ func (s *WorkerDeploymentSuite) TestDescribeWorkerDeployment_SetCurrentVersion()
...
 
+func (s *WorkerDeploymentSuite) TestConflictToken_Describe_SetCurrent_SetRamping() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	defer cancel()
+	tv := testvars.New(s)
+
+	firstVersion := tv.WithBuildIDNumber(1)
+	secondVersion := tv.WithBuildIDNumber(2)
+
+	// Start deployment version workflow + worker-deployment workflow. Only one version is stared manually
```
Shivs11, line 211: nit: started

Also, struggling to understand the comment actually

tests/worker_deployment_test.go
Shivs11, line 1: nice test!

maybe we can also have a test that errors out because an invalid conflict token has been passed? Not sure if this is tedious but might confirm if things are fully working or not

not a blocker if you think things are working though

service/history/workflow/mutable_state_impl.go
ShahabT:
```suggestion
```

Shivs11 (approved):

tests/activity_api_pause_test.go
```diff
@@ -1,6 +1,6 @@
 // The MIT License
 //
-// Copyright (c) 2024 Temporal Technologies Inc.  All rights reserved.
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
```
ShahabT, line 3:
```suggestion
// Copyright (c) 2024 Temporal Technologies Inc.  All rights reserved.
```

ShahabT (approved):

PR #7204, opened 2025-01-31
Limit max deployments and other dynamic config changes

carlydf:
## What changed?
- Add max deployments limit
- Add max versions per deployment limit
- Change how drainage durations were being passed to the drainage workflow; instead of binding the times to the workflow function at registration (which I think would have caused NDE/versioning issues if the values changed), pass the times to the version workflow when it starts, and store it in the VersionLocalState so that they are remembered if the version workflow continues as new before draining.


common/dynamicconfig/constants.go
```diff
@@ -1275,10 +1275,20 @@ these log lines can be noisy, we want to be able to turn on and sample selective
...
+	MatchingMaxVersionsInDeployment = NewNamespaceIntSetting(
+		"matching.maxVersionsInDeployment",
+		1000,
+		`MatchingMaxVersionsInDeployment represents the maximum number of versions that can be registered in a single worker deployment`,
+	)
 	MatchingMaxTaskQueuesInDeployment = NewNamespaceIntSetting(
 		"matching.maxTaskQueuesInDeployment",
 		1000,
-		`MatchingMaxTaskQueuesInDeployment represents the maximum number of task-queues that can be registed in a single deployment`,
+		`MatchingMaxTaskQueuesInDeployment represents the maximum number of task-queues that can be registered in a single worker deployment`,
```
Shivs11, line 1291: this should be changed to `MatchingMaxTaskQueuesInDeploymentVersions` since this would limit the total number of task-queues in a single deployment version, right?

service/worker/workerdeployment/client.go
```diff
@@ -451,6 +456,15 @@ func (d *ClientImpl) StartWorkerDeployment(
 	//revive:disable-next-line:defer
 	defer d.record("StartWorkerDeployment", &retErr, namespaceEntry.Name(), deploymentName, identity)()
 
+	// TODO (Carly): either max page size or default page size is 1000, so if there are > 1000 this would not catch it
```
Shivs11, line 459: this is a pretty important catch imo - maybe it's not of the utmost priority per-say for this deadline though

service/worker/workerdeployment/util.go
```diff
@@ -80,6 +81,10 @@ const (
 	errVersionAlreadyExistsType   = "errVersionAlreadyExists"
 	errMaxTaskQueuesInVersionType = "errMaxTaskQueuesInVersion"
 	errVersionAlreadyCurrentType  = "errVersionAlreadyCurrent"
+
+	// Defaults used in unexpected case where VersionLocalState is nil
```
Shivs11, line 85: this is a good safe-guard but I do wonder if this will *ever* happen ; nevertheless, good to always be safe

service/worker/workerdeployment/workflow.go
```diff
@@ -298,21 +298,21 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
 
 }
 
-func (d *WorkflowRunner) validateAddVersionToWorkerDeployment(version string) error {
-	if d.State.Versions == nil {
-		return nil
+func (d *WorkflowRunner) validateAddVersionToWorkerDeployment(args *deploymentspb.AddVersionToWorkerDeploymentUpdateArgs) error {
+	if int32(len(d.State.Versions)) >= args.MaxVersions {
+		return temporal.NewApplicationError("maximum versions in deployment, adjust scavenger speed or delete manually to continue deploying", "todo (carly)")
```
Shivs11, line 303: we can remove the "todo" here since we have the manual deletion process now

service/worker/workerdeployment/workflow.go
```diff
@@ -298,21 +298,21 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+		if v == args.Version {
 			return temporal.NewApplicationError("deployment version already registered", errVersionAlreadyExistsType)
 		}
 	}
 
 	return nil
 }
 
-func (d *WorkflowRunner) handleAddVersionToWorkerDeployment(ctx workflow.Context, version string) error {
+func (d *WorkflowRunner) handleAddVersionToWorkerDeployment(ctx workflow.Context, args *deploymentspb.AddVersionToWorkerDeploymentUpdateArgs) error {
```
Shivs11, line 315: thank you for making this a structure

Shivs11 (commented):
just one comment

common/dynamicconfig/constants.go
```diff
@@ -1275,10 +1275,20 @@ these log lines can be noisy, we want to be able to turn on and sample selective
...
 		`MatchingDropNonRetryableTasks states if we should drop matching tasks with Internal/Dataloss errors`,
 	)
+	MatchingMaxDeployments = NewNamespaceIntSetting(
+		"matching.maxDeployments",
+		1000,
+		`MatchingMaxDeployments represents the maximum number of worker deployments that can be registered in a single namespace`,
+	)
+	MatchingMaxVersionsInDeployment = NewNamespaceIntSetting(
+		"matching.maxVersionsInDeployment",
+		1000,
```
ShahabT, line 1285: I'm inclined to reduce this limit to say 100 for now until we get more confidence.

common/dynamicconfig/constants.go
```diff
@@ -1275,10 +1275,20 @@ these log lines can be noisy, we want to be able to turn on and sample selective
...
+		`MatchingMaxVersionsInDeployment represents the maximum number of versions that can be registered in a single worker deployment`,
+	)
 	MatchingMaxTaskQueuesInDeployment = NewNamespaceIntSetting(
 		"matching.maxTaskQueuesInDeployment",
 		1000,
```
ShahabT, line 1290: same for this one, maybe reduce it to 100 for now.

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -110,6 +111,9 @@ message VersionLocalState {
         // truth for that is the value at the root of WorkerDeploymentLocalState.
         map<int32, DeploymentVersionData> task_queues = 1;
     }
+
+    google.protobuf.Duration drainage_refresh_interval = 12;
+    google.protobuf.Duration drainage_visibility_grace_period = 13;
```
ShahabT, line 116: Please educate me, why we need to store these values in the state? can't we read the latest config in a `SideEffect` or something and then schedule the timer based on that, as opposed to sticking to potentially old config.

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -196,6 +200,13 @@ message StartWorkerDeploymentRequest {
     string request_id = 2;
 }
 
+message DrainageWorkflowArgs {
+    temporal.api.deployment.v1.WorkerDeploymentVersion version = 1;
+    bool first = 2;
+    google.protobuf.Duration refresh_interval = 3;
+    google.protobuf.Duration visibility_grace_period = 4;
```
ShahabT, line 207: same for these ones

proto/internal/temporal/server/api/deployment/v1/message.proto
carlydf: I didn't know any other way to read the config besides doing it in the fx

service/worker/workerdeployment/workflow.go
carlydf: we don't have the scavenger yet, so im going to keep it until then

PR #7205, opened 2025-01-31
Poller presence + DeleteVersion

Shivs11:
## What changed?

- added poller presence (called MissingTaskQueue in code) checks when a version starts to ramp or wants to be set as current
- happy path tests for delete version have been added
- happy path tests for poller presence checks have been added

## Why?

- versioning-3.1

## How did you test it?

- NO TESTS have been added - just an initial prototype for review.


service/worker/workerdeployment/workflow.go
```diff
@@ -184,7 +184,13 @@ func (d *WorkflowRunner) handleSetWorkerDeploymentRampingVersion(ctx workflow.Co
 			rampingSinceTime = d.State.RoutingInfo.RampingVersionChangedTime
 			rampingVersionUpdateTime = d.State.RoutingInfo.RampingVersionChangedTime
 		} else {
-			rampingSinceTime = routingUpdateTime // version ramping for the first time
+			// version ramping for the first time
+
+			if _, err := d.verifyPollerPresenceInVersion(ctx, prevRampingVersion, newRampingVersion); err != nil {
```
Shivs11, line 189: request flag to be added

cc - @Shivs11

go.mod
```diff
@@ -7,7 +7,7 @@ retract (
 	v1.26.0 // Published accidentally.
 )
 
-replace go.temporal.io/api => go.temporal.io/api v1.43.2-0.20250130194036-6d83c927eaed
+replace go.temporal.io/api => go.temporal.io/api v1.43.2-0.20250131161853-e69f12c6e1dc
```
Shivs11, line 10: latest @ versioning-3.1

service/worker/workerdeployment/version_workflow.go
```diff
@@ -503,7 +539,7 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
 
 	// stopped accepting new workflows --> start drainage child wf
 	if wasAcceptingNewWorkflows && !isAcceptingNewWorkflows {
-		state.DrainageInfo = &deploymentpb.VersionDrainageInfo{}
```
Shivs11, line 506: we now pass this on task-queue registration

service/worker/workerdeployment/version_activities.go
```diff
@@ -149,33 +149,32 @@ func (a *VersionActivities) CheckWorkerDeploymentUserDataPropagation(ctx context
...
-				Versions:      &taskqueuepb.TaskQueueVersionSelection{BuildIds: []string{args.BuildId}},
-				ReportPollers: true,
-				TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+				Namespace:      a.namespace.Name().String(),
+				TaskQueue:      tq,
+				ApiMode:        enumspb.DESCRIBE_TASK_QUEUE_MODE_ENHANCED,
+				Versions:       &taskqueuepb.TaskQueueVersionSelection{BuildIds: []string{worker_versioning.WorkerDeploymentVersionToString(args.WorkerDeploymentVersion)}}, // todo (Shivam) - pass a full version here
+				ReportPollers:  true,
+				TaskQueueType:  enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+				TaskQueueTypes: []enumspb.TaskQueueType{enumspb.TASK_QUEUE_TYPE_WORKFLOW},
```
Shivs11, line 162: I think your latest PR shall override these changes so we can ignore this

service/worker/workerdeployment/activities.go
```diff
@@ -58,6 +58,22 @@ func (a *Activities) SyncWorkerDeploymentVersion(ctx context.Context, args *depl
...
 
+func (a *Activities) IsVersionMissingTaskQueues(ctx context.Context, args *deploymentspb.IsVersionMissingTaskQueuesArgs) (*deploymentspb.IsVersionMissingTaskQueuesResult, error) {
+	res, err := a.deploymentClient.IsVersionMissingTaskQueues(
+		ctx,
+		a.namespace,
+		args.PrevCurrentVersion,
+		args.NewCurrentVersion,
+	)
+	if err != nil {
+		// todo (Shivam): do we return a non-retryable error here since we want to fail the operation if this check has failed.
```
carlydf, line 69: If the call to deploymentClient / matchingClient failed, it could be due to a network partition, which is retryable, we don't really know if it's retryable or not, so I think we want the Activity to retry a few times.

Ultimately, if the task queues are missing, the grpc error type we should return to the user is `serviceerror.NewFailedPrecondition` because the state of the system needs to change for this request to work.

service/worker/workerdeployment/client.go
```diff
@@ -520,6 +540,8 @@ func (d *ClientImpl) DeleteWorkerDeploymentVersion(
 	if err != nil {
 		return err
 	}
+
+	fmt.Printf("Inside delete, deployment name is %s\n", deploymentName)
```
carlydf, line 544: remove?

service/worker/workerdeployment/client.go
```diff
@@ -1100,3 +1127,128 @@ func makeDeploymentQuery(deploymentName, buildID string) string {
 	statusFilter = "= 'Running'"
 	return fmt.Sprintf("%s %s AND %s %s", searchattribute.BuildIds, deploymentFilter, searchattribute.ExecutionStatus, statusFilter)
 }
+
+func (d *ClientImpl) IsVersionMissingTaskQueues(ctx context.Context, namespaceEntry *namespace.Namespace, prevCurrentVersion, newVersion string) (bool, error) {
+	// Check if all the task-queues in the prevCurrentVersion are present in the newCurrentVersion
```
carlydf, line 1132:
```suggestion
	// Check if all the task-queues in the prevCurrentVersion are present in the newVersion (newVersion is either the new ramping version or the new current version)
```

service/worker/workerdeployment/client.go
```diff
@@ -1100,3 +1127,128 @@ func makeDeploymentQuery(deploymentName, buildID string) string {
...
+		return false, err
+	}
+
+	missingTaskQueues, err := d.checkForMissingTaskQueues(prevCurrentVersionInfo, newVersionInfo)
+	if err != nil {
+		return false, err
+	}
+
+	if len(missingTaskQueues) == 0 {
+		d.logger.Info("No missing task-queues found")
```
carlydf, line 1149: I think this might create log spam, since this will be true in most cases

service/worker/workerdeployment/client.go
```diff
@@ -1100,3 +1127,128 @@ func makeDeploymentQuery(deploymentName, buildID string) string {
...
+			return false, err
+		}
+		if isExpectedInNewVersion {
+			// one of the missing task queues is expected in the new version
+			return true, nil
+		}
+	}
+
+	// all expected task queues are present in the new version
+	d.logger.Info("All expected task queues are present in the new version")
```
carlydf, line 1166: I also worry that this could produce log-spam

carlydf: maybe d.logger.Debug ?

service/worker/workerdeployment/client.go
```diff
@@ -1100,3 +1127,128 @@ func makeDeploymentQuery(deploymentName, buildID string) string {
...
+		NamespaceId: namespaceEntry.ID().String(),
+		DescRequest: &workflowservice.DescribeTaskQueueRequest{
+			ApiMode: enumspb.DESCRIBE_TASK_QUEUE_MODE_ENHANCED,
+			TaskQueue: &taskqueuepb.TaskQueue{
+				Name: taskQueue.Name,
+				Kind: enumspb.TaskQueueKind(enumspb.TASK_QUEUE_KIND_NORMAL),
+			},
+			TaskQueueTypes: []enumspb.TaskQueueType{enumspb.TaskQueueType(taskQueue.Type)},
+			Versions: &taskqueuepb.TaskQueueVersionSelection{
+				BuildIds: []string{prevCurrentVersionInfo.GetVersion()},
```
carlydf, line 1211:
```suggestion
				BuildIds: []string{prevCurrentVersionInfo.GetVersion()}, // pretending the version string is a build id
```

service/matching/task_queue_partition_manager.go
```diff
@@ -526,25 +526,51 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 	buildIds map[string]bool,
 	includeAllActive, reportStats, reportPollers, internalTaskQueueStatus bool) (*matchingservice.DescribeTaskQueuePartitionResponse, error) {
 	pm.versionedQueuesLock.RLock()
+
```
carlydf, line 529: would love to hop on a call to discuss the changes to this file -- I think that will be the fastest way for me to review this part

service/worker/workerdeployment/version_activities.go
```diff
@@ -149,33 +149,32 @@ func (a *VersionActivities) CheckWorkerDeploymentUserDataPropagation(ctx context
 // CheckIfTaskQueuesHavePollers returns true if any of the given task queues has any pollers
 func (a *VersionActivities) CheckIfTaskQueuesHavePollers(ctx context.Context, args *deploymentspb.CheckTaskQueuesHaveNoPollersActivityArgs) (bool, error) {
```
carlydf, line 150: I'm realizing this args needs to be called `CheckTaskQueuesHavePollersActivityArgs` instead of `CheckTaskQueuesHaveNoPollersActivityArgs` cuz there's a big difference between the two

carlydf: I can fix that in my PR though

service/worker/workerdeployment/version_workflow.go
```diff
@@ -245,28 +250,37 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context) error
...
-		tqs = append(tqs, &taskqueuepb.TaskQueue{
-			Name: tqName,
-			Kind: enumspb.TASK_QUEUE_KIND_NORMAL, // TODO (Carly): could this be sticky?
-		})
+	// Manual deletion of versions is only possible when:
+	// 1. The version is not current or ramping
+	// 2. The version is drained. (check skipped when `skip-drainage=true` )
+	// 3. The version has no active pollers.
+
+	// 1. Check if the version is not current or ramping.
```
carlydf, line 260: technically if a version is drained it should never be current or ramping, but doesn't hurt to check

service/worker/workerdeployment/version_workflow.go
```diff
@@ -245,28 +250,37 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context) error
...
-	err = workflow.ExecuteActivity(activityCtx, d.a.CheckIfTaskQueuesHavePollers, checkPollersReq).Get(ctx, &hasPollers)
 	if err != nil {
+		// some other error allowing activity retries
 		return err
 	}
-	if hasPollers {
-		return serviceerror.NewFailedPrecondition("cannot delete, task queues in this version still have pollers")
-	}
+
+	d.logger.Info("Version is eligible for deletion")
```
carlydf, line 283: d.logger.Debug I think

service/worker/workerdeployment/version_activities.go
```diff
@@ -149,33 +149,32 @@ func (a *VersionActivities) CheckWorkerDeploymentUserDataPropagation(ctx context
...
 		}
-		typesInfo := res.GetDescResponse().GetVersionsInfo()[args.BuildId].GetTypesInfo()
+		typesInfo := res.GetDescResponse().GetVersionsInfo()[worker_versioning.WorkerDeploymentVersionToString(args.WorkerDeploymentVersion)].GetTypesInfo()
 		if len(typesInfo[int32(enumspb.TASK_QUEUE_TYPE_WORKFLOW)].GetPollers()) > 0 {
 			return true, nil
 		}
 		if len(typesInfo[int32(enumspb.TASK_QUEUE_TYPE_ACTIVITY)].GetPollers()) > 0 {
 			return true, nil
 		}
+		if len(typesInfo[int32(enumspb.TASK_QUEUE_TYPE_NEXUS)].GetPollers()) > 0 {
```
carlydf, line 175: are we versioning nexus task queues as well? I didn't know that

service/worker/workerdeployment/version_workflow.go
```diff
@@ -308,6 +322,28 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context) error
 	return nil
 }
 
+// doesVersionHaveActivePollers returns true if the version has active pollers.
+func (d *VersionWorkflowRunner) doesVersionHaveActivePollers(ctx workflow.Context) (bool, error) {
+	var tqs []*taskqueuepb.TaskQueue
+	for tqName := range d.VersionState.TaskQueueFamilies {
+		tqs = append(tqs, &taskqueuepb.TaskQueue{
```
carlydf, line 329: this will be changed in my PR

service/worker/workerdeployment/workflow.go
```diff
@@ -196,7 +197,20 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
 			rampingSinceTime = d.State.RoutingConfig.RampingVersionChangedTime
 			rampingVersionUpdateTime = d.State.RoutingConfig.RampingVersionChangedTime
 		} else {
-			rampingSinceTime = routingUpdateTime // version ramping for the first time
+			// version ramping for the first time
+
+			currentVersion := d.State.RoutingConfig.CurrentVersion
+			if !args.IgnoreMissingTaskQueues && currentVersion != "" {
```
carlydf, line 203: currentVersion will never be "" since it should start as "__unversioned__"

carlydf: maybe that's not in yet though

service/worker/workerdeployment/workflow.go
```diff
@@ -196,7 +197,20 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
...
+				isMissingTaskQueues, err := d.isVersionMissingTaskQueues(ctx, currentVersion, newRampingVersion)
+				if err != nil {
+					d.logger.Info("Error verifying poller presence in version", "error", err)
+					return nil, err
+				}
+				if isMissingTaskQueues {
+					return nil, serviceerror.NewFailedPrecondition("New ramping version does not have all the task queues from the previous current version and some missing task queues are unversioned and active")
```
carlydf, line 210:
```suggestion
					return nil, serviceerror.NewFailedPrecondition("New ramping version does not have all the task queues from the previous current version and some missing task queues are active and would become unversioned after this operation")
```

service/worker/workerdeployment/workflow.go
```diff
@@ -307,6 +321,17 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
 	updateTime := timestamppb.New(workflow.Now(ctx))
 
+	if !args.IgnoreMissingTaskQueues && prevCurrentVersion != "" {
+		isMissingTaskQueues, err := d.isVersionMissingTaskQueues(ctx, prevCurrentVersion, newCurrentVersion)
+		if err != nil {
+			d.logger.Info("Error verifying poller presence in version", "error", err)
+			return nil, err
+		}
+		if isMissingTaskQueues {
+			return nil, serviceerror.NewFailedPrecondition("New current version does not have all the task queues from the previous current version and some missing task queues are unversioned and active")
```
carlydf, line 331:
```suggestion
			return nil, serviceerror.NewFailedPrecondition("New current version does not have all the task queues from the previous current version and some missing task queues are active and would become unversioned after this operation")
```

service/worker/workerdeployment/workflow.go
```diff
@@ -377,6 +402,8 @@ func (d *WorkflowRunner) handleAddVersionToWorkerDeployment(ctx workflow.Context
 	if d.State.Versions == nil {
 		d.State.Versions = make([]string, 0)
 	}
+
+	fmt.Printf("Adding version %s to deployment %s\n", version, d.DeploymentName)
```
carlydf, line 406: remove?

carlydf (commented):
left some comments

service/matching/task_queue_partition_manager.go
carlydf: change request: comment at top of fn and in the two build-id cases about how you're accepting version id strings as build ids and returning them as build ids

service/worker/workerdeployment/client.go
Shivs11: removed them altogether

carlydf (approved):

service/worker/workerdeployment/client.go
```diff
@@ -726,11 +752,11 @@ func (d *ClientImpl) updateWithStartWorkerDeploymentVersion(
 			},
 			CreateTime:        now,
 			RoutingUpdateTime: now,
```
ShahabT, line 754: I don't think `RoutingUpdateTime` this should have any value at the beginning because no setCurrent or getCurrent was called for this version yet.

service/worker/workerdeployment/client.go
```diff
@@ -726,11 +752,11 @@ func (d *ClientImpl) updateWithStartWorkerDeploymentVersion(
...
-			CurrentSinceTime:  nil, // not current
-			RampingSinceTime:  nil, // not ramping
-			RampPercentage:    0,   // not ramping
-			DrainageInfo:      nil, // not draining or drained
-			Metadata:          nil, // todo
+			CurrentSinceTime:  nil,                                 // not current
+			RampingSinceTime:  nil,                                 // not ramping
+			RampPercentage:    0,                                   // not ramping
+			DrainageInfo:      &deploymentpb.VersionDrainageInfo{}, // not draining or drained
+			Metadata:          nil,                                 // todo
```
ShahabT, line 759: I think the client should be able to not mention these values at all and just zero values work fine. otherwise, we are violating the boundaries because client should not have responsibility of keeping workflow local state correct.

carlydf (approved):

PR #7207, opened 2025-01-31
Update DescribeHistoryHost to check shard ownership

yux0:
## What changed?
Update DescribeHistoryHost to check shard ownership

## Why?
When providing shard id to describe history host, it needs to check shard ownership to make sure it connect to the correct history host.

## How did you test it?
New unit test


hai719 (approved):

yycptt:
How will this API be used in the replication stack?

yux0:
> How will this API be used in the replication stack?

We will call this API to and it will trigger a refresh to the history client connection cache. https://github.com/temporalio/temporal/blob/0682c477181f1ffda5b618f22481c707e2494980/service/frontend/admin_handler.go#L1893

yycptt (approved):
Synced offline. Let's include change for using this API the into this PR as well.

PR #7209, opened 2025-01-31
Reduce default namespace cache refresh interval to 2s

pdoerner:
## What changed?

Reduced the default namespace cache refresh interval (dynamic config `system.namespaceCacheRefreshInterval`) from 10s -> 2s

## Why?

1. To increase responsiveness of the system
2. Migration handover wait was reduced to 10s and the cache refresh interval should be shorter than this value.

## How did you test it?

Existing tests

## Potential risks

Additional load on persistence from increased calls to refresh namespaces in the cache.

### DIFF
--- a/service/worker/deletenamespace/reclaimresources/workflow.go
+++ b/service/worker/deletenamespace/reclaimresources/workflow.go
@@ -42,7 +42,7 @@ import (
 const (
 	WorkflowName = "temporal-sys-reclaim-namespace-resources-workflow"
 
-	namespaceCacheRefreshDelay = 11 * time.Second
+	namespaceCacheRefreshDelay = 3 * time.Second
 )
 
 type (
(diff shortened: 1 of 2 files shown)


meiliang86 (approved):

yux0 (approved):

service/worker/deletenamespace/reclaimresources/workflow.go
```diff
@@ -42,7 +42,7 @@ import (
 const (
 	WorkflowName = "temporal-sys-reclaim-namespace-resources-workflow"
 
-	namespaceCacheRefreshDelay = 11 * time.Second
+	namespaceCacheRefreshDelay = 3 * time.Second
```
yycptt, line 45: 🤦

Let's may not change it? In production we are using 2s and 11s combination and thing are fine. 2s and 3s I am not sure. Relying on this delay to make sure no new workflows are created is kinda dangerous.

cc @alexshtin WDYT?

pdoerner: Sure, no strong opinion for me. Just did a quick search to try to find anywhere else we were redefining it as a constant.

alexshtin: I didn't realized that we have DC for this interval. This WF should read DC, not have its own const. I will change it. Revert it to `11` for now.

alexshtin: #7215

PR #7210, opened 2025-01-31
[Reset] Track child workflows initiated after reset

gow:
## What changed?
In this PR I'm recording a set of children that were initialized after the reset point. The children are identified by "workflow_type:workflow_id" key.

## Why?
When the parent starts to make progress after reset, it needs to know all the children that were already started after the reset point. In the current iteration of the feature we will use this information to terminate any running child before starting another instance.
We need this information mainly for 2 reasons.
1. If we don't attempt to terminate the parent may not make progress due to ID conflict policy 
2. On the other hand, always terminating all children before starting is also wrong since we may,
    -  Terminate any children that may be started in the future that were not present at the time of reset.
    - Terminate children belonging to a different parent.
So, we need to record an exact set of children that were observed (at the time of reset) to have been initialized after the reset point.


service/history/ndc/workflow_resetter.go
```diff
@@ -59,6 +59,14 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+const (
+	maxChildrenInResetMutableState = 1000 // max number of children tracked in reset mutable state
+)
+
+var (
+	errWorkflowResetterMaxChildren = serviceerror.NewInternal(fmt.Sprintf("WorkflowResetter encountered max allowed children [%d] while resetting.", maxChildrenInResetMutableState))
```
yycptt, line 67: Invalid argument?

service/history/ndc/workflow_resetter.go
```diff
@@ -648,6 +656,11 @@ func (r *workflowResetterImpl) reapplyContinueAsNewWorkflowEvents(
 		// All subsequent events should be excluded from being re-applied. So, do nothing and return.
 		return lastVisitedRunID, nil
 	}
+
+	// track the child workflow initiated (but not yet completed) events after reset.
```
yycptt, line 660: What's the story around completed child workflows? 
I mean if workflow being reset is using workflowIDReusePolicyRejectDuplicate then it will fail?

service/history/ndc/workflow_resetter.go
```diff
@@ -779,6 +811,38 @@ func (r *workflowResetterImpl) reapplyEventsFromBranch(
...
+	// nolint:exhaustive
+	switch event.GetEventType() {
+	// track the child if it has been initiated after reset
+	case enumspb.EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED:
+		childInitEventsAfterReset[event.GetEventId()] = event
+
+	// remove the child if it has reached a terminal state
+	case enumspb.EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_FAILED:
+		attr := event.GetStartChildWorkflowExecutionFailedEventAttributes()
+		delete(childInitEventsAfterReset, attr.InitiatedEventId)
```
yycptt, line 825: hmm InitiatedEventId could duplicate across runs.

service/history/ndc/workflow_resetter.go
gow: ah! yes. I should just record all initialized children after reset-point (whether they have reached a terminal state or not). Will make that change.

service/history/ndc/workflow_resetter.go
gow: Good catch.
Now that I won't be tracking completed children, I'll directly construct the map using child ID strings. No need to track init events.

service/history/ndc/workflow_resetter.go
```diff
@@ -733,6 +748,15 @@ func (r *workflowResetterImpl) reapplyContinueAsNewWorkflowEvents(
...
 		}
 	}
+	childrenInitializedAfterReset := make(map[string]bool)
+	for _, event := range childInitEventsAfterReset {
+		attr := event.GetStartChildWorkflowExecutionInitiatedEventAttributes()
+		childID := fmt.Sprintf("%s:%s", attr.GetWorkflowType().Name, attr.GetWorkflowId())
+		childrenInitializedAfterReset[childID] = true
+	}
+	if len(childrenInitializedAfterReset) > 0 {
+		resetMutableState.SetChildrenInitializedPostResetPoint(childrenInitializedAfterReset)
```
yycptt, line 758: are we carrying over this field upon continueAsNew? 
Same Q for other fields like ResetRunID and original executionRunID. (maybe we don't need to carry over them, just checking.)

service/history/ndc/workflow_resetter.go
gow: What do you mean by invalid arg? I want to include the `maxChildrenInResetMutableState` value in the error message (in case we change these limits).
Or did you mean something else?

service/history/ndc/workflow_resetter.go
gow: We need to carry them over since this data is used at the time of starting the child. And it's also used while accepting the results from the child.

service/history/ndc/workflow_resetter.go
gow: Fixed it. Now tracking all children that were initialized.

service/history/ndc/workflow_resetter.go
gow: This should not be an issue now.

service/history/ndc/workflow_resetter.go
yycptt: I mean the error type should be invalid argument, otherwise it counts against SLA.

service/history/ndc/workflow_resetter.go
yycptt: Synced offline. There will be a separate PR for carrying over the information.

yycptt (approved):

service/history/ndc/workflow_resetter.go
gow: Got it. Changed it to `InvalidArgument`

PR #7212, opened 2025-01-31
OTEL omit redundant persistence blobs

stephanos:
## What changed?

Omitting redundant persistence data blobs.

## Why?

Since the original data is already there, serializing the blobs is not useful. Removing it reduces noise.


alexshtin (approved):
Unless there is a bug in serialization logic and you want manually compare blob with content :-). I believe tracing shouldn't be used for that anyway.

PR #7221, opened 2025-02-03
Reapply WorkflowExecutionOptionsUpdated history event

rodrigozhou:
## What changed?

Reapply WorkflowExecutionOptionsUpdated history event in conflict resolution

### DIFF
--- a/service/history/ndc/workflow_resetter.go
+++ b/service/history/ndc/workflow_resetter.go
@@ -850,6 +850,18 @@ func reapplyEvents(
 	var reappliedEvents []*historypb.HistoryEvent
 	for _, event := range events {
 		switch event.GetEventType() {
+		case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED:
+			attr := event.GetWorkflowExecutionOptionsUpdatedEventAttributes()
+			if _, err := mutableState.AddWorkflowExecutionOptionsUpdatedEvent(
+				attr.GetVersioningOverride(),
+				attr.GetUnsetVersioningOverride(),
+				attr.GetAttachedRequestId(),
+				attr.GetAttachedCompletionCallbacks(),
+				event.Links,
+			); err != nil {
+				return reappliedEvents, err
+			}
+			reappliedEvents = append(reappliedEvents, event)
 		case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED:
 			if excludeSignal || isDuplicate(event) {
 				continue
--- a/service/history/ndc/workflow_resetter.go
+++ b/service/history/ndc/workflow_resetter.go
@@ -850,18 +850,6 @@ func reapplyEvents(
 	var reappliedEvents []*historypb.HistoryEvent
 	for _, event := range events {
 		switch event.GetEventType() {
-		case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED:
-			attr := event.GetWorkflowExecutionOptionsUpdatedEventAttributes()
-			if _, err := mutableState.AddWorkflowExecutionOptionsUpdatedEvent(
-				attr.GetVersioningOverride(),
-				attr.GetUnsetVersioningOverride(),
-				attr.GetAttachedRequestId(),
-				attr.GetAttachedCompletionCallbacks(),
-				event.Links,
-			); err != nil {
-				return reappliedEvents, err
-			}
-			reappliedEvents = append(reappliedEvents, event)
 		case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED:
 			if excludeSignal || isDuplicate(event) {
 				continue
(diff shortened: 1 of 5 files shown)


bergundy (approved):
LGTM, I would appreciate at least a manual test confirming that a second start request would be deduped post replication.

service/history/ndc/workflow_resetter.go
```diff
@@ -898,6 +898,21 @@ func reapplyEvents(
 				return reappliedEvents, err
 			}
 			reappliedEvents = append(reappliedEvents, event)
+		case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED:
```
yycptt, line 901: Two questions:
- So we think it makes sense to reapply this event even across runs? 
- Will the callback logic get triggered on a run that get terminated due to version conflict?

yycptt (commented):
Code LGTM, just want to confirm desired behavior.

service/history/ndc/workflow_resetter.go
rodrigozhou: > So we think it makes sense to reapply this event even across runs?

What do you mean by "across runs"?

> Will the callback logic get triggered on a run that get terminated due to version conflict?

Currently, callbacks are called in the workflow transitions to the following statuses:
- `WORKFLOW_EXECUTION_STATUS_COMPLETED`
- `WORKFLOW_EXECUTION_STATUS_FAILED`
- `WORKFLOW_EXECUTION_STATUS_TIMED_OUT`
- `WORKFLOW_EXECUTION_STATUS_CANCELED`
- `WORKFLOW_EXECUTION_STATUS_TERMINATED`

So, basically, only when it's continued as new it won't call the callbacks.
I suppose we want to change to not trigger in case of resets and version conflicts.
cc: @bergundy

bergundy: > Will the callback logic get triggered on a run that get terminated due to version conflict?

We didn't explicitly test that, so I can't say for sure. I'm not 100% what this scenario is, would another run take over as current and the losing run made a zombie? If that's the case, can we copy over the callbacks to the conflict winner? I'm not sure it completely makes sense to do that, it's kind of undefined behavior, probably best to trigger the callback and report termination of the loser and not apply to the winner.

yycptt: > you mean a workflow reset?

No I mean reapply from zombie run to current run. I think it makes sense, so just confirm here.

> would another run take over as current and the losing run made a zombie?

So all zombie runs will eventually be terminated (when the cluster creates the zombie run discovers there's a newer new run, it just terminate the run it has instead of putting it into zombie.) 
The situation I am thinking is one workflow W starts a workflow W1 in cluster A, after a (forced) failover, it creates re-runs the operation and creates W2. W1 will eventually be terminated and W2 may still be running. Once terminated, callback in W1 will (or maybe not) run. So now W will see a failed nexus operation despite that W2 is still running. 

> can we copy over the callbacks to the conflict winner?

I think it's already the case since we are reapplying WorkflowExecutionOptionsUpdatedEvent from zombie to current run? If the zombie (eventually terminated) run will also try to report back then it's a race condition. and may not be the behavior we want.

rodrigozhou: How/Where in the code the zombie workflow is terminated? Is it [here](https://github.com/temporalio/temporal/blob/de8f006301cfa527aeb05379b2982863db032bfe/service/history/workflow/mutable_state_impl.go#L4569)?
How do we know it's a zombie there?

rodrigozhou: Updated the PR to not process the callbacks when it's a zombie workflow.

> The situation I am thinking is one workflow W starts a workflow W1 in cluster A, after a (forced) failover, it creates re-runs the operation and creates W2. W1 will eventually be terminated and W2 may still be running. Once terminated, callback in W1 will (or maybe not) run. So now W will see a failed nexus operation despite that W2 is still running.

Discussed offline: it should be a rare case, and we'll handle this case another time.

service/history/ndc/workflow.go
```diff
@@ -195,6 +195,14 @@ func (r *WorkflowImpl) SuppressBy(
 	currentCluster := r.clusterMetadata.GetCurrentClusterName()
 
 	if currentCluster == lastWriteCluster {
+		// Update the workflow state to zombie so that ApplyWorkflowExecutionTerminatedEvent knows it
+		// was a zombie and it won't process the completion callbacks.
+		if err := r.mutableState.UpdateWorkflowStateStatus(
```
yycptt, line 200: Should be safe from what I can see. but cc @xwduan @hai719 @yux0 FYI.

service/history/ndc/workflow_resetter.go
yycptt: Not processing callback for zombie workflows already fixed ^ issue since W1 will no longer report back to W. What we discussed offline is a different case where there are two levels of zombie workflows.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4558,7 +4559,10 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionTerminatedEvent(
 	ms.ClearStickyTaskQueue()
 	ms.writeEventToCache(event)
 
-	return ms.processCloseCallbacks()
+	if currentState != enumsspb.WORKFLOW_EXECUTION_STATE_ZOMBIE {
```
yycptt, line 4562: Is this good enough? 
I mean this will prevent the current cluster from generating callback tasks. But the termination event will eventually be replicated (in event-based replication) to another cluster, then in that cluster callback tasks will still be generated (and processed if failover happens)?

yycptt (commented):
I don't want to block this PR and at least the reapply logic within one run.

The zombie workflow cases I described are very rare and will only happen during forced failover, so we can address them later. 

Feel free to revert to the previous version of the PR and land it.

service/history/ndc/workflow.go
xwduan: Yeah. I think it is safe. This also may helped on a case where the terminated events comes from suppress should not be reapplied.

rodrigozhou:
@yycptt I reverted the last commit as it's actually causing tests to fail, and there's more things to consider. I'll follow up on that in another PR.

PR #7223, opened 2025-02-03
Proto definition for CHASM nodes

yycptt:
## What changed?

- Proto definition for CHASM nodes

## Why?

- CHASM project


proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -0,0 +1,72 @@
...
+package temporal.server.api.persistence.v1;
+option go_package = "go.temporal.io/server/api/persistence/v1;persistence";
+
+import "google/protobuf/timestamp.proto";
+
+import "temporal/server/api/persistence/v1/hsm.proto";
+import "temporal/api/common/v1/message.proto";
+
+message ChasmNode {
+    // Versioned transition when the node was last updated.
```
lina-temporal, line 34: (minor) I think these comments were meant to swap lines :)

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -0,0 +1,72 @@
...
+        ChasmDataAttributes data_attributes = 12;
+        ChasmCollectionAttributes collection_attributes = 13;
+        ChasmPointerAttributes pointer_attributes = 14;
+    }
+}
+
+message ChasmComponentAttributes {
+    message Task {
+        // Fully qualified name of a registered task.
+        string name = 1;
```
lina-temporal, line 50: minor/naming - `name` would usually be a constant defined in a CHASM component, for the task registry, right? IMO, `type` would be clearer to me as an indication that a class of `Task`s is registered/referred to in CHASM by a specific identifier, versus `name` which I tend to think of as more synonymous with 'key'.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -0,0 +1,72 @@
...
+        string destination = 2;
+        google.protobuf.Timestamp scheduled_time = 3;
+        temporal.api.common.v1.DataBlob data = 4;
+        // Versioned transition of the Entity when the task was created.
+        VersionedTransition versioned_transition = 5;
+    }
+
+    // Fully qualified name of a registered component.
+    string name = 1;
```
lina-temporal, line 59: What does fully qualified mean w.r.t the component name? When I look at [ref.go](https://github.com/temporalio/temporal/pull/6987/files#diff-76564ca39466189d27d8372ae7f27e8837caa9e10d5d5e4e79103621d0082493) for the interface draft, I'm assuming "fully qualified" refers to the path you'd get if you did `rootComponentPath+componentPath+componentName` (or however that's supposed to be joined).

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -0,0 +1,72 @@
...
+}
+
+message ChasmDataAttributes {
+    temporal.api.common.v1.DataBlob data = 1;
+}
+
+message ChasmCollectionAttributes {}
+
+message ChasmPointerAttributes {
+    repeated string node_path = 1;
```
lina-temporal, line 71: Can a node's path ever change? I ask since I believe we'd been discussing something relating to the ability of a CHASM component being able to essentially update its own type, maybe by replacing itself in the node graph with a different component. Lmk if I've misremembered.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt: 🤦 Fixed. Thanks!

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt: Check [this](https://github.com/temporalio/temporal/blob/main/chasm/registry.go#L95). 
It's basically the library name + component name.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt: A node can't be moved. At least I am not aware of such discussion so far.

But yes, you can delete the old node and create a new node with the same path.

Pointers needs to be validated in that case via the initialVersionedTransition field (pointer is not valid if the initialVT of the pointer is older than the initialVT of the node it points to). and we can decide what to do for invalid pointers, either failed the transition directly or just return NotFound when the pointer is dereferenced.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt: hmm yeah it's a bit confusing. Agree `type` is more clear here. 

I also need to change other files under `chasm` to use `type`. Though `type` alone won't work for variable names, so will be `componentType`, or `taskType` I think.

yycptt: https://github.com/temporalio/temporal/pull/7225

alexshtin (approved):
I like that you put comments on some fields. I found proto documentation very useful. I believe we should document every field in this file. I am ok to postpone it as we might change these messages a lot.

bergundy (commented):
The protos look good to me but I would really appreciate if we do better with documentation. These are protos that are at the core of the platform. We have an opportunity to raise the documentation bar here. Same goes for the chasm framework interfaces, we're building a framework that should be very easy for devs to use, treat it as you would a user facing SDK please.

PR #7224, opened 2025-02-03
Dynamic config subscriber in Update Registry

stephanos:
## What changed?

Using the new dynamic config subscribers in the Update registry.

## Why?

To keep the Update registry's settings up-to-date.

## How did you test it?

Updated existing tests.

### DIFF
--- a/service/history/workflow/context.go
+++ b/service/history/workflow/context.go
@@ -167,6 +167,8 @@ type (
 		lock           locks.PrioritySemaphore
 		MutableState   MutableState
 		updateRegistry update.Registry
+
+		onClose []func()
 	}
 )
 
@@ -985,28 +987,15 @@ func (c *ContextImpl) UpdateRegistry(ctx context.Context) update.Registry {
 	}
 
 	if c.updateRegistry == nil {
-		nsIDStr := c.MutableState.GetNamespaceEntry().ID().String()
-
+		ns := c.MutableState.GetNamespaceEntry().Name()
 		c.updateRegistry = update.NewRegistry(
 			c.MutableState,
 			update.WithLogger(c.logger),
 			update.WithMetrics(c.metricsHandler),
 			update.WithTracerProvider(trace.SpanFromContext(ctx).TracerProvider()),
-			update.WithInFlightLimit(
-				func() int {
-					return c.config.WorkflowExecutionMaxInFlightUpdates(nsIDStr)
-				},
-			),
-			update.WithRegistrySizeLimit(
-				func() int {
-					return c.config.WorkflowExecutionMaxInFlightUpdatePayloads(nsIDStr)
-				},
-			),
-			update.WithTotalLimit(
-				func() int {
-					return c.config.WorkflowExecutionMaxTotalUpdates(nsIDStr)
-				},
-			),
+			update.WithInFlightLimit(ns, c.config.WorkflowExecutionMaxInFlightUpdates),
+			update.WithRegistrySizeLimit(ns, c.config.WorkflowExecutionMaxInFlightUpdatePayloads),
+			update.WithTotalLimit(ns, c.config.WorkflowExecutionMaxTotalUpdates),
 		)
 	}
 	return c.updateRegistry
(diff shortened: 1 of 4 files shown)


service/history/workflow/update/registry.go
```diff
@@ -172,13 +197,10 @@ func NewRegistry(
 	opts ...Option,
 ) Registry {
 	r := &registry{
-		updates:              make(map[string]*Update),
-		store:                store,
-		instrumentation:      noopInstrumentation,
-		maxRegistrySizeLimit: func() int { return 0 }, // ie disabled
-		maxInFlight:          func() int { return 0 }, // ie disabled
-		maxTotal:             func() int { return 0 }, // ie disabled
```
stephanos, line 180: Not needed anymore.

service/history/workflow/context.go
```diff
@@ -985,28 +987,15 @@ func (c *ContextImpl) UpdateRegistry(ctx context.Context) update.Registry {
 	}
 
 	if c.updateRegistry == nil {
-		nsIDStr := c.MutableState.GetNamespaceEntry().ID().String()
-
+		ns := c.MutableState.GetNamespaceEntry().Name()
```
stephanos, line 990: Drive-by fix: this was "namespace ID" but should be "namespace name"

service/history/workflow/update/registry.go
```diff
@@ -358,6 +380,10 @@ func (r *registry) Send(
 func (r *registry) Clear() {
 	r.Abort(AbortReasonRegistryCleared)
 
+	for _, f := range r.onClear {
+		f()
+	}
```
alexshtin, line 385: I am not sure about it. `cancel()` cancels subscription created by `Subscribe()` which is called once per process? `r.Clear()` is called many times. Wouldn't it stop getting notification after first `Clear()` call?

alexshtin (changes requested):

service/history/workflow/update/registry.go
stephanos: Maybe it should be called `Destroy` instead of `Clear`. Clear implies it can be used after.

dnr (commented):
It's not clear to me this is a good use of subscriptions. The uses of these settings appear to be "reactive", that is, it reads and uses the setting when it needs it, and doesn't require any proactive work to be done at change time. So changes take effect immediately anyway, there's no additional delay that can be removed with subscriptions.

Also, the lifecycle of this registry looks somewhat complex and is definitely not "for the lifetime of the process" which is the common case for subscriptions. It works, but if the ratio between "use value" and "create/delete subscription" isn't extremely high, there may not be any performance benefit, and it might be worse overall: subscriptions require more allocation than just getting the value, so the bookkeeping may outweigh any gains.

What is the motivation here?

[edit: A good rule of thumb is: if your subscription handler just stores a value somewhere, then probably don't use subscriptions, just use Get. If it _pushes_ a value into some other thing that can't be modified to do a pull, that's appropriate.]

stephanos:
@dnr Oh, that's good feedback. The intention was just to keep the settings in-sync and not just read them once at the start of the registry creation. It wasn't clear to me that the overhead of just reading the dynamic config every time would be smaller. In that case, I can just evaluate of the dynamic config every time it's accessed to achieve the same goal 👍

dnr:
> just evaluate of the dynamic config every time it's accessed to achieve the same goal

The old code already evaluated them on every access

stephanos:
> The old code already evaluated them on every access

But the registry was caching the evaluation across it's lifetime: https://github.com/temporalio/temporal/pull/7224/files#diff-c365166a3b00b43950b46bc9be819b9c407f41947377a563c8b18369bd5d897fL997

dnr:
> But the registry was caching the evaluation across it's lifetime

That call is inside a closure, just to bind the namespace. `maxInFlight`, `maxTotal`, etc. were being called each time

stephanos:
🤦 You're right.

stephanos:
Revert: https://github.com/temporalio/temporal/pull/7251

stephanos:
Let's all forget this ever happened.

image

PR #7225, opened 2025-02-03
CHASM: Rename name to type

yycptt:
## What changed?

- Rename component/taskName to component/TaskType in chasm framework to avoid confusion.

### DIFF
--- a/chasm/ref.go
+++ b/chasm/ref.go
@@ -41,18 +41,22 @@ type EntityKey struct {
 type ComponentRef struct {
 	EntityKey
 
-	// This is needed for routing
-	rootComponentName string
-	// or maybe make it specific
+	// Fully qualified component type name for routing.
+	rootComponentType string
+	// We could also make the routing infomation clear with
 	// routingKey string
 
+	// We can also simply put shardID here.
+	// TODO: Remove one of shardID and rootComponentType.
 	shardID int32
 
-	// From the component name, we can find the component struct definition
-	// use reflection to find sub-components and understand if those sub-components
-	// needs to be loaded or not. we only need to do this for sub-components of the component.
-	// path for parent/ancenstor component can be inferred from the path.
-	componentName      string
+	// Fully qualified component type name.
+	// From the component type, we can find the component struct definition,
+	// then use reflection to find sub-components and understand if those sub-components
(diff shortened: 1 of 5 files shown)


chasm/registry.go
```diff
@@ -33,30 +33,30 @@ import (
...
+		componentByType:   make(map[string]*RegistrableComponent),
+		componentByGoType: make(map[reflect.Type]*RegistrableComponent),
+		taskByType:        make(map[string]*RegistrableTask),
+		taskByGoType:      make(map[reflect.Type]*RegistrableTask),
 	}
 }
 
 func (r *Registry) Register(lib Library) error {
-	if err := r.validateName(lib.Name()); err != nil {
+	if err := r.validateTypeName(lib.Name()); err != nil {
```
yycptt, line 59: :( It's still library.Name().library type will be weird I think.

Maybe I should use `typeName` everywhere. That's also what  golang does. There's a Name() method  in reflect.Type.

alexshtin: I think it is fine. Go package has name, types inside it also have names. FQN on usage size looks like `<package_name>.<type_name>`. I would revert `validateTypeName` back to `validateName` because it validates all different names, not only types. Same for `typeNameValidator` var.

chasm/ref.go
```diff
@@ -41,18 +41,22 @@ type EntityKey struct {
...
-	// needs to be loaded or not. we only need to do this for sub-components of the component.
-	// path for parent/ancenstor component can be inferred from the path.
-	componentName      string
+	// Fully qualified component type name.
+	// From the component type, we can find the component struct definition,
+	// then use reflection to find sub-components and understand if those sub-components
+	// need to be loaded or not.
+	// We only need to do this for sub-components, path for parent/ancenstor components
+	// can be inferred from the current component path.
+	componentType      string
```
alexshtin, line 59: I am a little in doubt between `componentType` and `componentTypeName`. I am ok to use the former for now, and see if it is confusing. We can rename it once again later.

alexshtin (approved):

chasm/registry.go
yycptt: Done. Reverted back to `nameValidator` and `validateName`.

PR #7227, opened 2025-02-03
skip processing old tasks if stream restarts

hai719:

### DIFF
--- a/service/history/replication/executable_task.go
+++ b/service/history/replication/executable_task.go
@@ -256,8 +256,7 @@ func (e *ExecutableTaskImpl) Cancel() {
 		"replication task: %v encountered cancellation event",
 		e.taskID,
 	))
-	now := time.Now().UTC()
-	e.emitFinishMetrics(now)
+	// should not emit metrics for cancelled task
 }
 
 func (e *ExecutableTaskImpl) Reschedule() {
--- a/service/history/replication/executable_task_tracker.go
+++ b/service/history/replication/executable_task_tracker.go
@@ -58,6 +58,7 @@ type (
 		LowWatermark() *WatermarkInfo
 		Size() int
 		Cancel()
+		Resume()
 	}
 	ExecutableTaskTrackerImpl struct {
 		logger         log.Logger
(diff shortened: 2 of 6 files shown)


service/history/replication/executable_task_tracker.go
```diff
@@ -205,6 +209,9 @@ func (t *ExecutableTaskTrackerImpl) Cancel() {
 	t.cancelled = true
 	t.cancelLocked()
 }
+func (t *ExecutableTaskTrackerImpl) Resume() {
```
xwduan, line 212: When `Resume()`, we also needs to pop all the tasks that is canceled. Otherwise, the ACK cannot be move forward.

xwduan: And when calling `Cancel()`, we need to move the HighReadWatermark back to the last executed task. I think we may need to redesign the task tracker to not allow cancel, otherwise it is hard to make HighReadWatermark going back.

hai719: For tasks that are canceled, task.Execute() will return directly and Ack() in the scheduler.

hai719: https://github.com/temporalio/temporal/blob/4e7c6628b698f19795515c2d594057fcbeb0bff3/service/history/queues/executable.go#L267

hai719: Let me move tracker.Cancel() out of StreamReceiver so it can still work and won't mess up the HighReadWatermark

service/history/replication/executable_task.go
```diff
@@ -256,8 +256,7 @@ func (e *ExecutableTaskImpl) Cancel() {
 		"replication task: %v encountered cancellation event",
 		e.taskID,
 	))
-	now := time.Now().UTC()
-	e.emitFinishMetrics(now)
+	// should not emit metrics for cancelled task
```
yux0, line 259: do we ever use cancel? I thought we only use abort.

service/history/replication/executable_task_tracker.go
```diff
@@ -205,6 +209,9 @@ func (t *ExecutableTaskTrackerImpl) Cancel() {
...
+	t.cancelled = false
```
yux0, line 213: get lock?

service/history/replication/stream_receiver.go
```diff
@@ -87,20 +87,17 @@ func NewClusterShardKey(
 }
 
 func NewStreamReceiver(
+	logger log.Logger,
 	processToolBox ProcessToolBox,
 	taskConverter ExecutableTaskConverter,
 	clientShardKey ClusterShardKey,
 	serverShardKey ClusterShardKey,
+	highPriorityTaskTracker ExecutableTaskTracker,
+	lowPriorityTaskTracker ExecutableTaskTracker,
```
yux0, line 96: if we move this up. Then the stream receiver should not call cancel/stop on task tracker?

PR #7228, opened 2025-02-04
Fix on state based replication

xwduan:
## What changed?

1. Improve logging on replication task execution
2. Check versioned transition existence before reading it. (avoid panic)
3. When SyncState returning `nil` error and 'false` doContinue, ignore the task instead of keep retrying. Otherwise this will cause DLQ.
4. release workflow lock before `convertTaskEquivalents`, otherwise deadlock.
## Why?

Fix on state based replication
## How did you test it?

Ran tasks on test clusters
## Potential risks

n/a
## Documentation

n/a
## Is hotfix candidate?

no

### DIFF
--- a/service/history/replication/executable_backfill_history_events_task.go
+++ b/service/history/replication/executable_backfill_history_events_task.go
@@ -187,7 +187,6 @@ func (e *ExecutableBackfillHistoryEventsTask) HandleErr(err error) error {
 				)
 				return err
 			}
-			// return original task processing error
 			return nil
 		}
 		return e.Execute()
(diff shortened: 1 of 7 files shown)


service/history/replication/executable_backfill_history_events_task.go
```diff
@@ -171,16 +178,17 @@ func (e *ExecutableBackfillHistoryEventsTask) HandleErr(err error) error {
...
+				e.Logger.Error("BackFillHistoryEvent replication task encountered error during sync state",
 					tag.WorkflowNamespaceID(e.NamespaceID),
 					tag.WorkflowID(e.WorkflowID),
 					tag.WorkflowRunID(e.RunID),
 					tag.TaskID(e.ExecutableTask.TaskID()),
 					tag.Error(syncStateErr),
 				)
+				return err
 			}
 			// return original task processing error
```
hai719, line 190: nit: remove the comment

hai719 (approved):

PR #7234, opened 2025-02-04
Pass task queue types to Describe and take unecessary info out of TaskQueueFamilyData

carlydf:
## What changed?
1. Pass Task Queue Types to `CheckIfTaskQueuesHavePollers` activity so that we check pollers for only the task queue types that are registered in that version.
2. Remove `DeploymentVersionData` from the per-type map in `TaskQueueFamilyData`. `DeploymentVersionData` stores the same info as `VersionLocalState`, so repeating it num_task_queue times inside the VersionLocalState was a waste of space, and kind of confusing. I spoke with Shahab yesterday about how `TaskQueueFamilyData` should contain information that is specific to each task queue type within that version. So far, the only data that is specific to the version + task queue tuple is `first_poller_time`, so I put that in a new `TaskQueueVersionData` struct and put it in `TaskQueueFamilyData`.

## Why?
See explanation above.

## How did you test it?
Tested that the existing Versioning 3.1 functional tests still work.


proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -125,6 +125,9 @@ message DeploymentData {
     // (or unspecified) WorkflowVersioningMode will receive the tasks.
     repeated temporal.server.api.deployment.v1.DeploymentVersionData versions = 2;
 
+    // Present if the task queue's ramping version or current version is unversioned.
```
carlydf, line 128: this is what we will do when "set unversioned ramp" or "set unversioned current" is synced to the task queue. `unversioned_data.version=nil` always. Theoretically it could go in the `versions` with VersionData.version=nil, but Shahab and I had the idea to make it extra obvious with this separate one.

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -112,13 +109,17 @@ message VersionLocalState {
...
-    }
+message TaskQueueFamilyData {
+    // Key: Task Queue Type
+    map<int32, TaskQueueVersionData> task_queues = 1;
+}
+
+// Data specific to a task queue, from the perspective of a worker deployment version.
+message TaskQueueVersionData {
+    // Experimental. First time this task queue received a poller with this worker version.
+    google.protobuf.Timestamp first_poller_time = 1;
```
Shivs11, line 122: good catch! completely missed that we were actually wasting space and just using our local state whenever syncing to task-queue - sorry!

I quite like these structs - maybe we can keep `TaskQueueFamilyData` nested inside `VersionLocalState` since it's specific to it though - wdyt

service/worker/workerdeployment/version_workflow.go
```diff
@@ -247,17 +246,18 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context) error
...
+		for tqType, _ := range tqFamilyData.TaskQueues {
+			tqTypes = append(tqTypes, enumspb.TaskQueueType(tqType))
+		}
+		tqNameToTypes[tqName] = &deploymentspb.CheckTaskQueuesHaveNoPollersActivityArgs_TaskQueueTypes{Types: tqTypes}
+	}
 	checkPollersReq := &deploymentspb.CheckTaskQueuesHaveNoPollersActivityArgs{
-		TaskQueues: tqs,
-		BuildId:    d.VersionState.Version.BuildId,
+		TaskQueuesAndTypes: tqNameToTypes,
+		BuildId:            d.VersionState.Version.BuildId,
```
Shivs11, line 260: based on our discussion yesterday, we will be passing in a full-fledged version rather than a buildID here - but not a blocker since I can get the change in

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
Shivs11: hmm, I do wonder if we *really* do need this - When we do wanna make unversioned as ramping or current, we would have a `DeploymentVersionData.version` as nil. We can extract all the necessary information from that instance, even if the task-queue is unversioned.

Right now, the code in `worker-versioning` doesn't need to deal with this additional structure and can infer whether version is current/ramping from the deployment versions present + routing time too 

I see how making it obvious would make things easier to understand when developing, but not sure if im fully convinced on if we need this

Shivs11 (approved):
approved, but with a comment or two

service/worker/workerdeployment/version_workflow.go
```diff
@@ -273,16 +273,6 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context) error
...
-	for tqName, byType := range state.TaskQueueFamilies {
-		for tqType, oldData := range byType.TaskQueues {
-
-			syncReq.Sync = append(syncReq.Sync, &deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData{
-				Name: tqName,
-				Type: enumspb.TaskQueueType(tqType),
-				Data: oldData,
-			})
-		}
-	}
```
Shivs11, line 285: I seem to be missing something here - why are we removing this piece of code? According to me, when forgetting a version, we still have to specify the task-queues and the types from where the version would be forgotten from - this piece of code seems to be doing that

proto/internal/temporal/server/api/deployment/v1/message.proto
carlydf: oh yes! I took it out because I was using it in `CheckTaskQueuesHaveNoPollersActivityArgs`, but then I changed that, so I'll put it back in the nested.

service/worker/workerdeployment/version_workflow.go
carlydf: hm ok, I'll leave that to your PR since I feel like it requires other changes too?

service/worker/workerdeployment/version_workflow.go
carlydf: I got confused, just meant to remove the data part. Fixing

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
carlydf: changed to this per our convo in standup:
```
    temporal.server.api.deployment.v1.DeploymentVersionData unversioned_ramp_data = 3;
```

PR #7211, opened 2025-01-31
Use otel metrics by default

yiminc:
## What changed?
Use otel metrics by default if not specified. 

## Why?
Otel metrics is stable and tally is abandoned 

## How did you test it?
Unit test

## Potential risks
The metrics may not be 100% exactly the same between otel and tally.

## Is hotfix candidate?
No

PR #7235, opened 2025-02-04
Add backward compatibility configs for tally to otel migration

yycptt:
## What changed?

- Add backward compatibility configs for tally to otel migration

## Why?

- There are some breaking changes when switching from tally to otel metric framework.
1. Unit name will be appended to the metric name
2. `_total` will be appended to counter metrics
3. Timer metric will be emitted as milliseconds instead of seconds
This PR adds configs for controlling both behaviors when using otel framework so that the behavior will be the same as tally.

## How did you test it?

- Unit tests
- Tested locally with various combination of configurations.

## Potential risks

- No risk by default. If new options are enabled, timer metric value may be wrong or not bucketed correctly.

## Is hotfix candidate?

- No.

### DIFF
--- a/common/metrics/config.go
+++ b/common/metrics/config.go
@@ -74,11 +74,14 @@ type (
 		PerUnitHistogramBoundaries map[string][]float64 `yaml:"perUnitHistogramBoundaries"`
 
 		// Following configs are added for backwards compatibility when switching from tally to opentelemetry
-		// Both configs should be set to true when using opentelemetry framework to have the same behavior as tally.
+		// All configs should be set to true when using opentelemetry framework to have the same behavior as tally.
 
-		// WithoutUnits controls if metric unit should be added to the metric name as a suffix.
+		// WithoutUnitSuffix controls the additional of unit suffixes to metric names.
 		// This config only takes effect when using opentelemetry framework.
-		WithoutUnits bool `yaml:"withoutUnits"`
+		WithoutUnitSuffix bool `yaml:"withoutUnitSuffix"`
+		// WithoutCounterSuffix controls the additional of _total suffixes to counter metric names.
+		// This config only takes effect when using opentelemetry framework.
+		WithoutCounterSuffix bool `yaml:"withoutCounterSuffix"`
 		// RecordTimerInSeconds controls if Timer metric should be emitted as number of seconds
 		// (instead of milliseconds).
 		// This config only takes effect when using opentelemetry framework.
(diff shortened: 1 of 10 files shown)


yycptt:
Related to https://github.com/temporalio/temporal/pull/7211

bergundy (approved):
LGTM, looks like there's good test coverage.

yiminc (approved):

common/metrics/config.go
```diff
@@ -72,6 +72,20 @@ type (
...
+
+		// Following configs are added for backwards compatibility when switching from tally to opentelemetry
+		// All configs should be set to true when using opentelemetry framework to have the same behavior as tally.
+
+		// WithoutUnitSuffix controls the additional of unit suffixes to metric names.
+		// This config only takes effect when using opentelemetry framework.
+		WithoutUnitSuffix bool `yaml:"withoutUnitSuffix"`
+		// WithoutCounterSuffix controls the additional of _total suffixes to counter metric names.
+		// This config only takes effect when using opentelemetry framework.
+		WithoutCounterSuffix bool `yaml:"withoutCounterSuffix"`
```
yycptt, line 84: Also added config for disabling the addition of `_total` to Counters

1nu:
Hi! I might be wrong, but to me it seems like OTEL is missing backwards compatibility for tally users that configured `prefix`.

yycptt:
> Hi! I might be wrong, but to me it seems like OTEL is missing backwards compatibility for tally users that configured `prefix`.

Hi @1nu, thanks for reporting! 

Yeah, looks like it's the case and `Prefix` was never supported when framework is set to `opentelementry`. I will look into how to keep the behavior the same for tally users.

yycptt:
https://github.com/temporalio/temporal/pull/7297

1nu:
I just discovered another potential backward compatibility issue in handling `tags` keys and values with dashes, reporting here just in case. Tally replaces dashes with underscores in both key and value, opentelemetry only replaces dashes in the key.
For example:
```
...
    tags:
      temporal-cluster: "order-management"
...
```
becomes `temporal_cluster="order_management"` using tally and `temporal_cluster="order-management"` with opentelemetry.

yycptt:
I have [reverted](https://github.com/temporalio/temporal/pull/7303) the change to always use OTEL as the framework, so there will be no impact for Tally users in the next release.

We will look into those issues and do some more testing. Also needs to continue to support statsD users somehow with otel before fully deprecate tally.

Thanks again @1nu for flagging those issues!

PR #7236, opened 2025-02-04
[Reset] Override ID reuse and conflict policy when starting child

gow:
## What changed?
- Added `is_reset_run` in ParentExecutionInfo
- Overriding `startRequest.WorkflowIdReusePolicy` & `startRequest.WorkflowIdConflictPolicy` for children that were recorded in the `executionInfo.ChildrenInitializedPostResetPoint`
- Once a child is started with the overridden policies, we set `executionInfo.ChildrenInitializedPostResetPoint[childID] = false` so that we don't do this again if the parent starts another instance of the same child.

## Why?
This is needed so that the parent can terminate any running children that were started in the previous run. Without this after reset, the parent may not make progress.

## How did you test it?
Manual testing. Will add functional tests in followup PRs once the feature is ready.

## Potential risks
N/A

## Documentation
No. Documentation change is pending.

## Is hotfix candidate?
No


proto/internal/temporal/server/api/workflow/v1/message.proto
```diff
@@ -37,6 +37,7 @@ message ParentExecutionInfo {
     int64 initiated_id = 4;
     temporal.server.api.clock.v1.VectorClock clock = 5;
     int64 initiated_version = 6;
+    bool is_reset_run = 7;
```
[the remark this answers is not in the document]
gow: Yeah. Looks like it's an existing pattern. We can use it here. Made the change.

service/history/transfer_queue_active_task_executor.go
```diff
@@ -928,6 +938,16 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
 	t.logger.Debug("Child Execution started successfully",
 		tag.WorkflowID(attributes.WorkflowId), tag.WorkflowRunID(childRunID))
 
+	// if shouldTerminateAndStartChild is true, it means the child was started after the reset point and we attempted to terminate it before starting a new one.
+	// We should update the parent execution info to reflect that the child was potentially terminated and started.
+	if shouldTerminateAndStartChild {
+		childrenInitializedPostResetPoint := executionInfo.ChildrenInitializedPostResetPoint
+		childrenInitializedPostResetPoint[resetChildID] = &persistencespb.ResetChildInfo{
+			ShouldTerminateAndStart: false,
```
yycptt, line 946: do we ever clean up this field?

service/history/transfer_queue_active_task_executor.go
```diff
@@ -843,33 +843,42 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
...
-				return err
+				parentClock, err := t.shardContext.NewVectorClock()
+				if err != nil {
+					return err
+				}
+				return t.createFirstWorkflowTask(ctx, task.TargetNamespaceID, childExecution, parentClock, childClock)
+			}
+		} else {
+			// child was started after reset-point. But we need to first check if this child was recorded at the time of reset.
+			if resetChildInfo, ok := mutableState.GetExecutionInfo().GetChildrenInitializedPostResetPoint()[resetChildID]; ok {
```
yycptt, line 879: per our discussion: this check needs to be done even if `mutableState.IsResetRun()` is false to support continue as new case.

service/history/transfer_queue_active_task_executor.go
gow: I don't have any plans of removing this data yet. We need it to prevent forwarded results from old children. So, after the child completes we can delete this data.

service/history/transfer_queue_active_task_executor.go
gow: ah! yes. Fixed it.

service/history/transfer_queue_active_task_executor.go
yycptt: Please track the deletion work.

yycptt (approved):

service/history/transfer_queue_active_task_executor.go
gow: PR coming up.

gow: https://github.com/temporalio/temporal/pull/7295

PR #7240, opened 2025-02-04
pass only version from worker deployment client when starting version wf

carlydf:

### DIFF
--- a/service/worker/workerdeployment/version_workflow.go
+++ b/service/worker/workerdeployment/version_workflow.go
@@ -119,7 +119,9 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
 	if d.GetVersionState().Version == nil {
 		return fmt.Errorf("version cannot be nil on start")
 	}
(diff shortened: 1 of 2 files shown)


service/worker/workerdeployment/version_workflow.go
```diff
@@ -113,9 +116,10 @@ func (d *VersionWorkflowRunner) listenToSignals(ctx workflow.Context) {
 }
 
 func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
-	if d.VersionState == nil {
-		d.VersionState = &deploymentspb.VersionLocalState{}
+	if d.GetVersionState().Version == nil {
+		return fmt.Errorf("version cannot be nil on start")
 	}
+	d.VersionState.CreateTime = timestamppb.New(workflow.Now(ctx))
```
ShahabT, line 122: Be careful with CaNs. This should be set only if it's nil.

ShahabT (approved):

PR #7242, opened 2025-02-05
Cleanup state machine timers on machine deletion

justinp-tt:
## What changed?

Delete state machine timers every time we close a transaction on mutable state.

## Why?

Part of the work on state machine deletion

## How did you test it?

Unit tests

### DIFF
--- a/service/history/workflow/task_generator.go
+++ b/service/history/workflow/task_generator.go
@@ -27,7 +27,6 @@
 package workflow
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


bergundy (commented):
We want to trim state machine timers not just when transition history is enabled, any timer that we are tracking where the relevant state machine has been deleted, can also be deleted.

To achieve this, I probably wouldn't use the current implementation of `TrimStateMachineTimers`, it will recreated all of the timers when transition history is disabled, generating extra "physical" timers in the queue.

You can instead react to nodes being deleted in the opLog (line 323 in the current code version) and delete the timers there.

Another thing you need to do is fix the TODO in `task_refresher.go` line 659, here's what I have in mind there (untested):

```go
	haveNewDeletedStateMachines := slices.ContainsFunc(mutableState.GetExecutionInfo().SubStateMachineTombstoneBatches, func(batch *persistencespb.StateMachineTombstoneBatch) bool {
		isFreshBatch := CompareVersionedTransition(
			batch.VersionedTransition,
			minVersionedTransition,
		) >= 0
		return isFreshBatch || slices.ContainsFunc(batch.StateMachineTombstones, func(ts *persistencespb.StateMachineTombstone) bool {
			// At least one state machine node was deleted.
			return ts.GetStateMachinePath() != nil
		})
	})

	if len(nodesToRefresh) != 0 || haveNewDeletedStateMachines {
		if err := TrimStateMachineTimers(mutableState, minVersionedTransition); err != nil {
			return err
		}
	}
```

service/history/workflow/task_generator.go
```diff
@@ -328,7 +347,10 @@ func (r *TaskGeneratorImpl) GenerateDirtySubStateMachineTasks(
 		}
 	}
 
-	AddNextStateMachineTimerTask(r.mutableState)
+	if r.mutableState.IsTransitionHistoryEnabled() {
+		// Ensure there's a timer task scheduled if needed
+		AddNextStateMachineTimerTask(r.mutableState)
+	}
```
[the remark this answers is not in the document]
justinp-tt: Done

service/history/workflow/task_refresher.go
```diff
@@ -656,10 +657,21 @@ func (r *TaskRefresherImpl) refreshTasksForSubStateMachines(
...
+				batch.VersionedTransition,
+				minVersionedTransition,
+			) >= 0
+			return isFreshBatch || slices.ContainsFunc(batch.StateMachineTombstones, func(ts *persistencespb.StateMachineTombstone) bool {
+				// At least one state machine node was deleted.
+				return ts.GetStateMachinePath() != nil
+			})
+		})
+
+		if len(nodesToRefresh) != 0 || haveNewDeletedStateMachines {
```
yycptt, line 671: `len(nodesToRefresh) != 0` is always true here right, we already checked it on L659.

I think you want to move the haveNewDeletedStateMachines logic before L659?

service/history/workflow/task_refresher.go
```diff
@@ -656,10 +657,21 @@ func (r *TaskRefresherImpl) refreshTasksForSubStateMachines(
...
-		// TODO: after hsm node tombstone is tracked in mutable state,
-		// also trigger trim when there are new tombstones after minVersionedTransition
-		if err := TrimStateMachineTimers(mutableState, minVersionedTransition); err != nil {
-			return err
+		haveNewDeletedStateMachines := slices.ContainsFunc(mutableState.GetExecutionInfo().SubStateMachineTombstoneBatches, func(batch *persistencespb.StateMachineTombstoneBatch) bool {
+			isFreshBatch := CompareVersionedTransition(
+				batch.VersionedTransition,
+				minVersionedTransition,
+			) >= 0
+			return isFreshBatch || slices.ContainsFunc(batch.StateMachineTombstones, func(ts *persistencespb.StateMachineTombstone) bool {
```
yycptt, line 665: `&&` instead of `||` ?

service/history/workflow/task_generator.go
```diff
@@ -300,31 +300,50 @@ func (r *TaskGeneratorImpl) GenerateDirtySubStateMachineTasks(
...
-		if err != nil {
-			return err
-		}
+	for _, op := range opLog {
+		switch transitionOp := op.(type) {
+		case hsm.DeleteOperation:
+			execInfo := r.mutableState.GetExecutionInfo()
+			trimmedTimers := make([]*persistencespb.StateMachineTimerGroup, 0, len(execInfo.StateMachineTimers))
+
+			for _, group := range execInfo.StateMachineTimers {
```
bergundy, line 310: nit: this would be neater if it was a separate function.

bergundy (approved):
Overall LGTM, wondering if @yycptt sees something I don't.

service/history/workflow/task_refresher.go
```diff
@@ -656,10 +657,21 @@ func (r *TaskRefresherImpl) refreshTasksForSubStateMachines(
 	}
 
 	if len(nodesToRefresh) != 0 {
-		// TODO: after hsm node tombstone is tracked in mutable state,
-		// also trigger trim when there are new tombstones after minVersionedTransition
-		if err := TrimStateMachineTimers(mutableState, minVersionedTransition); err != nil {
-			return err
+		haveNewDeletedStateMachines := slices.ContainsFunc(mutableState.GetExecutionInfo().SubStateMachineTombstoneBatches, func(batch *persistencespb.StateMachineTombstoneBatch) bool {
```
yycptt, line 660: hmm I am afraid it's not that easy. The tombstone are capped to a certain length, so there could be nodes that got deleted but are not tracked in the batches. i.e. minVersionedTransition < the versionedTransition of the first batch. I forgot the exact contract here. @xwduan Can you help comment here?

And we can always trim in that case I think. 

Also this is only relevant to the state-based replication I think. so if we are under tight deadline, we can do this in a separate PR and only make sure trimming is working in active cluster and standby cluster via existing event based replication.

xwduan: Yeah that correct. We still have the trim. The first element of tombstone batch indicating the first state transition it tracks. That means all sub statemachines deleted before this state transition are not tracked.

yycptt: @xwduan Will the tombstoneBatches be empty? or we always but a placeholder batch there? I remember we discussed that but could recall the decision.

service/history/workflow/task_generator.go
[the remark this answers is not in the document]
bergundy: I don't mind too much since we don't plan on adding more use cases for the HSM framework.

service/history/workflow/task_refresher.go
bergundy: @justinp-tt do you need help figuring out how to translate this to code?

service/history/workflow/task_refresher.go
bergundy: Ah yes, thanks!

service/history/workflow/task_refresher.go
bergundy: Let's do the state replication part later since we're on a deadline.

PR #7244, opened 2025-02-05
Handle Unversioned Current and Ramping Version

carlydf:
## What changed?
- The initial current version of a worker deployment is now `"__unversioned__"` instead of `""`
- When the ramping version of a worker deployment becomes unversioned, the deployment tells all the task queues in the current version that they now have an unversioned ramp.
- If the unversioned ramp is promoted to current version, or if the unversioned ramping version is unset, the deployment tells all the task queues of the previous current version that they no longer have an unversioned ramp.

## How did you test it?
- Made sure all existing WorkerDeploymentSuite and DeploymentVersionSuite tests pass
- Wrote new tests in WorkerDeploymentSuite for the specific edge cases that this code handles:
   - `TestSetCurrentVersion_Unversioned_NoRamp`: Test that when current version changes from versioned -> unversioned, the task queues become unversioned
   - `TestSetCurrentVersion_Unversioned_PromoteUnversionedRamp`: Test that when the current version changes from versioned -> unversioned and unversioned was previously ramping, the task queues become unversioned with no more unversioned ramp
   - `TestSetRampingVersion_Unversioned_UnversionedCurrent`: Test that this fails with "Ramping version __unversioned__ is already current" error
   - `TestSetRampingVersion_Unversioned_VersionedCurrent`: Test that the ramping version of the current version's task queues becomes unversioned


common/worker_versioning/worker_versioning.go
```diff
@@ -392,10 +392,6 @@ func FindDeploymentVersionForWorkflowID(
 ) *deploymentspb.WorkerDeploymentVersion {
 	ramp := ramping.GetRampPercentage()
 	rampingVersion := ramping.GetVersion()
-	if rampingVersion.GetBuildId() == "" {
-		// Ramping to unversioned
-		rampingVersion = nil
-	}
```
carlydf, line 398: `ramping.GetVersion()` will already be nil if there is an unversioned ramp

service/worker/workerdeployment/workflow.go
```diff
@@ -338,19 +358,37 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+			RampPercentage:    0,   // remove ramp for that version if it was ramping
+		}
+		if _, err := d.syncVersion(ctx, newCurrentVersion, currUpdateArgs); err != nil {
+			return nil, err
+		}
+	} else if d.State.RoutingConfig.RampingVersion == worker_versioning.UnversionedVersionId {
+		// If the new current is unversioned, and it was previously ramping, we need to tell
+		// all the task queues with unversioned ramp that they no longer have unversioned ramp.
+		// The task queues with unversioned ramp are the task queues of the previous current version.
+		// TODO (Carly): Should we ban people from changing the task queues in the current version while they have an unversioned ramp?
```
carlydf, line 376: discuss

carlydf: reasoning for banning this is that we use the current version's task queues to know what task queues the unversioned ramp is set on. We need that list to be able to unset the unversioned ramp.

ShahabT: task queues can only be added to the current version, right? so the list will only get larger so you'd still unset all the unversioned ramps from the TQs who got the unversioned ramp.

ShahabT (approved):

service/worker/workerdeployment/workflow.go
carlydf: If someone does ignoreMissingTaskQueues = true then they can do whatever they want though right? including removing task queues from the current version?

carlydf: We would be unable to tell those task queues "you are done with unversioned ramp"

carlydf: Other option is to maintain a separate list inside the DeploymentLocalState of the task queues that we set unversioned ramp on ...

PR #7245, opened 2025-02-05
Inherit pinned Version from parent

ShahabT:
## What changed?

Child workflows start in parent's pinned Worker Deployment Version.

## Why?

So user do not have to worry about interface compatibility between pinned parents and children.

## How did you test it?

Added new tests.


tests/versioning_3_test.go
```diff
@@ -878,6 +877,403 @@ func (s *Versioning3Suite) testIndependentActivity(behavior enumspb.VersioningBe
 	s.verifyWorkflowVersioning(tvWf, behavior, tvWf.Deployment(), nil, nil)
 }
 
+func (s *Versioning3Suite) TestChildWorkflowInheritance_PinnedParent() {
+	s.testChildWorkflowInheritance_ExpectInherit(false, false, vbPinned)
```
carlydf, line 881:
```suggestion
        // same task queue, no override, parent is pinned -> expect child to start on parent's pinned version
	s.testChildWorkflowInheritance_ExpectInherit(false, false, vbPinned)
```
Writing this partially to confirm we're on the same page, partially because I do think little comments here like this would make it easier for future readers (maybe us in a year) to quickly understand this

tests/versioning_3_test.go
```diff
@@ -878,6 +877,403 @@ func (s *Versioning3Suite) testIndependentActivity(behavior enumspb.VersioningBe
...
+}
+
```
carlydf, line 883: would it make sense to add one of these tests doing
`s.testChildWorkflowInheritance_ExpectInherit(false, true, vbPinned)` to check that the override is inherited?

proto/internal/temporal/server/api/workflow/v1/message.proto
```diff
@@ -37,6 +37,11 @@ message ParentExecutionInfo {
     int64 initiated_id = 4;
     temporal.server.api.clock.v1.VectorClock clock = 5;
     int64 initiated_version = 6;
+    // When present, child workflow starts as Pinned to this Worker Deployment Version.
+    // Set only if the parent execution is Pinned to a Worker Deployment Version, and the child
```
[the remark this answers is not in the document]
carlydf:
```suggestion
    // Set only if the parent execution is effectively Pinned to a Worker Deployment Version when it first starts the child workflow, and the child
```

service/history/transfer_queue_active_task_executor.go
```diff
@@ -886,6 +887,19 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
 		},
 	}
 
+	parentPinnedVersion := ""
+	var parentPinnedOverride *workflowpb.VersioningOverride
+	if attributes.TaskQueue.GetName() == mutableState.GetExecutionInfo().GetTaskQueue() {
+		// TODO (shahab): also inherit when the child TQ is different, but in the same Version
+		if mutableState.GetEffectiveVersioningBehavior() == enumspb.VERSIONING_BEHAVIOR_PINNED {
```
carlydf, line 894: I'm not convinced that this should be the `EffectiveVersioningBehavior`. If the parent is unpinned base with a pinned override, I think the child should be unpinned base with a pinned override. Right now, the child would be pinned base with a pinned override.

tests/versioning_3_test.go
ShahabT: yes

service/history/transfer_queue_active_task_executor.go
ShahabT: If the parent has a pinned override, it doesn't matter because the child will get the pinned override anyways.

Now, the parent is pinned but has an unpinned, then I don't think there is any point to start the child in the same build because the parent is effectively unpinned and hence cannot expect interface compatibility protection.

tests/versioning_3_test.go
ShahabT: done

service/history/transfer_queue_active_task_executor.go
carlydf: If the parent has (base: unpinned, override: pinned) I think it does matter for the child to have (base: unpinned, override: pinned) instead of (base: pinned, override: pinned), because if the user does a batch job to unset the override on a batch that covers all the parents and children of one workflow type, then the parent would become unpinned but the child would remain pinned, which is probably not what the user wants / expects.

carlydf: actually, thinking about what Antonio said is making me reconsider this.

The true base pinned-ness or unpinned-ness of the child ultimately comes from how the child wftype is registered in the code.

If the child starts on the pinned override of their unpinned parent (let's say Version A), hopefully, if the user expects the child to be unpinned, the Version A code of the child will say that the child is unpinned. Then, on WFT completion, the child's base behavior will immediately become unpinned, as the user expects.

carlydf: I think the way you have it is correct. And now that the comment says "effectively Pinned", the proto is technically accurate as well.

carlydf (approved):

PR #7251, opened 2025-02-05
Remove Subscription from Update Registry

stephanos:
## What changed?

Reverts https://github.com/temporalio/temporal/pull/7224 but keeps the bug fix.

## Why?

Change was unnecessary.

## How did you test it?

Modified existing tests.

### DIFF
--- a/service/history/workflow/context.go
+++ b/service/history/workflow/context.go
@@ -985,7 +985,7 @@ func (c *ContextImpl) UpdateRegistry(ctx context.Context) update.Registry {
 	}
 
 	if c.updateRegistry == nil {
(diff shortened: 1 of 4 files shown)


service/history/workflow/context.go
```diff
@@ -987,15 +985,28 @@ func (c *ContextImpl) UpdateRegistry(ctx context.Context) update.Registry {
 	}
 
 	if c.updateRegistry == nil {
-		ns := c.MutableState.GetNamespaceEntry().Name()
+		nsName := c.MutableState.GetNamespaceEntry().Name().String()
```
stephanos, line 988: Keeping the bug fix.

dnr (approved):

PR #7179, opened 2025-01-29
Send raw history blobs from history service to frontend

prathyushpv:
## What changed?
Change to send raw history blobs from history service to frontend service. History service returns a new proto message that has a repeated bytes history field.
This response is wire compatible with the original response which has temporal.api.history.v1.History type for this field. This allows history service to not deserialize events from this data blob. This considerably reduces CPU usage.

History service still needs event_id and version decoded from history events. For this we use a new proto message StrippedHistoryEvent which has these two fields only. It takes considerably less CPU to decode events to this struct.

## Why?
We have seen incidents of high history CPU usage when large number of GetWorkflowExecutionHistory calls are made to workflows which has large history. With this change we can reduce the CPU burden on history service during this API call.

## How did you test it?
Existing unit tests and manual test to run workflows.

## Is hotfix candidate?
No

PR #7252, opened 2025-02-05
Send raw history events from history to matching service

prathyushpv:
## What changed?
Send raw history events from history to matching service similar to the change in https://github.com/temporalio/temporal/pull/7179.

Instead of deserializing history blobs to history events, history service will send the raw bytes to matching service. History service uses a different proto message to do that. Matching client response proto has History message in the same field. GRPC's proto deserialization will automatically deserialize raw history batch bytes to History message. Matching to frontend response still have deserialized History message in the response. And there is not much CPU usage to save by changing this.

## Why?
We have seen incidents of high history CPU usage when large amount of history events are pulled. With this change we can reduce the CPU burden on history service during this time.

## How did you test it?
Unit tests

## Is hotfix candidate?
No


service/frontend/workflow_handler.go
```diff
@@ -917,6 +917,12 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
 			tag.Error(err))
 		return nil, err
 	}
+	if matchingResp.History != nil {
+		err = api.ProcessOutgoingSearchAttributes(wh.namespaceRegistry, wh.saProvider, wh.saMapperProvider, matchingResp.History.Events, namespaceID, wh.visibilityMgr)
```
yycptt, line 921: hmmm during rollout this can potentially process search attributes 0 time or twice I think.

yycptt: Actually the same thing could happen for the previous GetWorkflowExecutionHistory change?

service/history/api/recordworkflowtaskstarted/api.go
```diff
@@ -408,3 +417,69 @@ func CreateRecordWorkflowTaskStartedResponse(
 
 	return response, nil
 }
+
+func CreateRecordWorkflowTaskStartedResponseWithRawHistory(
```
yycptt, line 421: so the logic here is the same as `CreateRecordWorkflowTaskStartedResponse`, only the value type is different? 

I wonder if we should have only one copy of the logic for determining the values and then a conversion func between those two types.

service/frontend/workflow_handler.go
prathyushpv: Right.. let me see how to avoid this issue.

common/dynamicconfig/constants.go
```diff
@@ -2384,7 +2384,7 @@ that task will be sent to DLQ.`,
 	)
 	SendRawHistoryBetweenInternalServices = NewGlobalBoolSetting(
 		"history.sendRawHistoryBetweenInternalServices",
-		false,
+		true,
```
yycptt, line 2387: this is for testing and will be reverted?

yycptt (approved):

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -275,6 +275,31 @@ message RecordWorkflowTaskStartedResponse {
     bytes next_page_token = 19;
 }
 
+// RecordWorkflowTaskStartedResponseWithRawHistory should be wire compatible with RecordWorkflowTaskStartedResponse.
+// The only difference is that RecordWorkflowTaskStartedResponseWithRawHistory has a `history` field that contains the
+// raw history batches.
+message RecordWorkflowTaskStartedResponseWithRawHistory {
```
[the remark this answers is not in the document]
prathyushpv: Done

PR #7253, opened 2025-02-05
Nits about Deployment Name and Build ID validation

carlydf:


ShahabT:
added changes to https://github.com/temporalio/temporal/pull/7268/files

PR #7254, opened 2025-02-05
Enforce Update Registry size limit

stephanos:
## What changed?

Enforce the configurable inflight size limit for Updates.

## Why?

Instead of only using a fixed number of inflight Updates, a size-based approach is more helpful.

The limit is at 10MB right now. But will likely be lowered very soon.

## How did you test it?

Added tests.

### DIFF
--- a/service/history/workflow/update/registry.go
+++ b/service/history/workflow/update/registry.go
@@ -112,15 +112,14 @@ type (
 		updates map[string]*Update
 		// A store from which Registry is constructed with NewRegistry function,
 		// and completed Updates are loaded. Practically it is a Mutable State.
-		store UpdateStore
-
-		instrumentation instrumentation
+		store           UpdateStore
 		completedCount  int
 		failoverVersion int64
+		instrumentation instrumentation
 
-		maxInFlight                           func() int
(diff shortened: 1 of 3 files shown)


service/history/workflow/update/registry.go
```diff
@@ -176,13 +176,13 @@ func NewRegistry(
...
-		instrumentation:      noopInstrumentation,
-		maxRegistrySizeLimit: func() int { return 0 }, // ie disabled
-		maxInFlight:          func() int { return 0 }, // ie disabled
-		maxTotal:             func() int { return 0 }, // ie disabled
-		failoverVersion:      store.GetCurrentVersion(),
+		updates:               make(map[string]*Update),
+		store:                 store,
+		instrumentation:       noopInstrumentation,
+		maxInflightUpdateSize: func() int { return 0 }, // ie disabled
+		maxInFlight:           func() int { return 0 }, // ie disabled
```
alexshtin, line 183: Shouldn't this be `maxInFlightUpdateCount` now?

alexshtin (approved):

alexshtin:
Internal doc needs to be updated for this too.

PR #7257, opened 2025-02-05
Fix passing callbacks to continued as new workflow and retries

rodrigozhou:
## What changed?

Fix passing callbacks to continued as new workflow and retries.

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -555,21 +555,6 @@ func NewMutableStateInChain(
 	// carry over necessary fields from current mutable state
 	newMutableState.executionInfo.WorkflowExecutionTimerTaskStatus = currentMutableState.GetExecutionInfo().WorkflowExecutionTimerTaskStatus
 
-	// Copy completion callbacks to new run.
-	oldCallbacks := callbacks.MachineCollection(currentMutableState.HSM())
-	newCallbacks := callbacks.MachineCollection(newMutableState.HSM())
-	for _, node := range oldCallbacks.List() {
-		cb, err := oldCallbacks.Data(node.Key.ID)
-		if err != nil {
-			return nil, err
-		}
-		if _, ok := cb.GetTrigger().GetVariant().(*persistencespb.CallbackInfo_Trigger_WorkflowClosed); ok {
-			if _, err := newCallbacks.Add(node.Key.ID, cb); err != nil {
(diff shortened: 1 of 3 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -555,21 +555,6 @@ func NewMutableStateInChain(
 	// carry over necessary fields from current mutable state
 	newMutableState.executionInfo.WorkflowExecutionTimerTaskStatus = currentMutableState.GetExecutionInfo().WorkflowExecutionTimerTaskStatus
```
rodrigozhou, line 556: @yycptt I wonder if this is also wrong as it's not reapplied from a history event.

bergundy (approved):
Thanks!

service/history/workflow/mutable_state_impl.go
yycptt: hmm not sure I follow. What's the issue here? 
The taskStatus only exists in mutable state so can't be applied from a event.

yycptt: and I think NewMutableStateInChain is used in continue as new and retry cases as well.
I guess I am just lacking context here.

pdoerner: I'm also confused what was the issue here? I thought we had functional tests to cover this case https://github.com/temporalio/temporal/blob/main/tests/callbacks_test.go#L211

rodrigozhou: The test checks that the callbacks are in the mutable state. The issue is in the mutable state rebuilder. Since it's currently copying the callbacks in the `NewMutableStateInChain`, and not storing them in the start event, if the mutable state needs to be rebuilt, it won't have the callbacks.

rodrigozhou: In this case, for task status, after you write it here, it will be overwritten here again: https://github.com/temporalio/temporal/blob/107918aa5d3d2fe38c0af7704d6aacd28009b504/service/history/workflow/mutable_state_impl.go#L2321

yycptt: >  it will be overwritten again in AddWorkflowExecutionStartedEventWithOptions
yes that's expected. The `GenerateWorkflowStartTasks` logic will first read the taskStatus and skip generating new tasks if the Status is already Create.

PR #7258, opened 2025-02-05
Sync multiple TQ types at once

ShahabT:
## What changed?

Syncing TQ User Data for all types of a single TQ name at once to reduce the SetCurrent and SetRamping latency and flakiness of tests.


service/worker/workerdeployment/activities.go
```diff
@@ -81,10 +82,11 @@ func (a *Activities) SyncUnversionedRamp(
 	}
 	var taskQueueSyncs []*deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData
 	for _, tqInfo := range currVersionInfo.GetTaskQueueInfos() {
+		// TODO: group by TQs by name and make only one sync req per name
```
carlydf, line 85:
```suggestion
		// TODO (Carly): group by TQs by name and make only one sync req per name
```
unless you want to do it in this PR

carlydf (approved):
approved with one comment

carlydf (approved):

PR #7259, opened 2025-02-05
Propagate Nexus failure source header to outbound executor metrics

pdoerner:
## What changed?

Propagating the Nexus response header `Temporal-Nexus-Failure-Source` to the outbound queue executor.

## Why?

This header is included as a label on the outbound queue metrics so that we can exclude expected user errors from our alerts.

### DIFF
--- a/common/nexus/failure.go
+++ b/common/nexus/failure.go
@@ -48,6 +48,10 @@ type failureSourceContextKeyType struct{}
 
 var FailureSourceContextKey = failureSourceContextKeyType{}
 
+type FailureSourceContextValue struct {
+	Source string
+}
+
 var failureTypeString = string((&failurepb.Failure{}).ProtoReflect().Descriptor().FullName())
 
 // ProtoFailureToNexusFailure converts a proto Nexus Failure to a Nexus SDK Failure.
--- a/common/nexus/failure.go
+++ b/common/nexus/failure.go
@@ -44,7 +44,9 @@ const (
 	FailureSourceWorker = "worker"
 )
 
-type FailureSourceContextKey struct{}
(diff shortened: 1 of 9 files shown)


common/nexus/failure.go
```diff
@@ -37,6 +37,15 @@ import (
...
 )
 
+const (
+	// FailureSourceHeaderName is the header used to indicate from where the Nexus failure originated.
+	FailureSourceHeaderName = "Temporal-Nexus-Failure-Source"
+	// FailureSourceWorker indicates the failure originated from outside the server (e.g. bad request or on the Nexus worker).
+	FailureSourceWorker = "worker"
+)
+
+type FailureSourceContextKey struct{}
```
bergundy, line 47: You'll want to create an instance of this struct to use as a unique key.

```suggestion
type FailureSourceContextKeyType struct{}
var FailureSourceContextKey FailureSourceContextKeyType{}
```

bergundy (commented):
Looks great, can you please also add a test that shows we actually extract the tag?

common/nexus/failure.go
```diff
@@ -37,6 +37,21 @@ import (
...
+	FailureSourceWorker = "worker"
+)
+
+type failureSourceContextKeyType struct{}
+
+var FailureSourceContextKey = failureSourceContextKeyType{}
+
+type FailureSourceContextValue struct {
+	Source string
+}
```
bergundy, line 53: Use an atomic value instead, what you have isn't concurrency safe: https://pkg.go.dev/sync/atomic#Value

bergundy (approved):
LGTM, please address the comment before merging.

PR #7261, opened 2025-02-06
[Reset] Verify parent before terminating and starting child

gow:
## What changed?
Modified `service/history/api/startworkflow/api.go` to first verify the parent-child relationship before terminating running child.
 
## Why?
We need this verification since the parent overrides the ID reuse/conflict policies in reset runs. But by the time the parent gets to "restart" the child, that child could have been claimed by someone else. So we need to ensure that the currently running child was indeed started by the parent issuing this request.

## How did you test it?
Manual tests. Will add functional tests in followup PRs.

## Potential risks
N/A

## Documentation
Documentation of the feature is pending. But the feature itself is in pre-release (gated behind a dynamic config)

## Is hotfix candidate?
No


service/history/api/signalwithstartworkflow/signal_with_start_workflow.go
```diff
@@ -186,6 +186,8 @@ func createWorkflowMutationFunction(
 		workflowIDReusePolicy,
 		workflowIDConflictPolicy,
 		currentWorkflowStartTime,
+		nil,
+		false,
```
gow, line 190: Maybe keeping the flag in `ParentExecutionInfo` would have been better? Now we have to pass 2 params all the way down.

yycptt: if ChildWorkflowOnly is false then we don't need to pass parentExecutionInfo down...  so for ResolveDuplicateWorkflowID we can say as long as parentExecutionInfo is specified, we will verify the relationship. That could be confusing for the caller though... 

I am ok with passing both down. I didn't know we have that many layers though 🤦 

I still feel it's a bit weird for it to be in `ParentExecutionInfo` because the info is stored in the child, but the verification and termination is done on a different workflow.

gow: Agree. Passing both is the best possible option.

yycptt (approved):

PR #7265, opened 2025-02-06
Always use OTEL as metric handler

yycptt:
## What changed?

- Always use OTEL as metric handler

## Why?

- Deprecate Tally metrics framework

## How did you test it?

- Will test locally and also monitor CI

## Potential risks

- Metric could be different and break existing dashboard & alerting systems for existing tally users.

## Documentation

- No. We keep the behavior unchanged for tally users.

## Is hotfix candidate?

- No

### DIFF
--- a/common/metrics/config.go
+++ b/common/metrics/config.go
@@ -77,14 +77,14 @@ type (
 		// All configs should be set to true when using opentelemetry framework to have the same behavior as tally.
 
 		// WithoutUnitSuffix controls the additional of unit suffixes to metric names.
-		// This config only takes effect when using opentelemetry framework.
+		// This config only takes effect when explicitly using "opentelemetry" as the framework.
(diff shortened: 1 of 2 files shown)


yiminc (approved):

common/metrics/config.go
```diff
@@ -475,25 +480,24 @@ func newPrometheusScope(
 
 // MetricsHandlerFromConfig is used at startup to construct a MetricsHandler
 func MetricsHandlerFromConfig(logger log.Logger, c *Config) (Handler, error) {
-	if c == nil {
+	if c == nil || c.Prometheus == nil {
```
yycptt, line 483: This is existing behavior, when creating tally handler, `NewScope` will return a noop scope if c.Prometheus is nil.

PR #7266, opened 2025-02-06
Deletes are now idempotent

Shivs11:
## What changed?

- make deletes idempotent

## How did you test it?

- added functional tests 
- also changes to functional test to make them have a stricter consistency check

### DIFF
--- a/tests/deployment_version_test.go
+++ b/tests/deployment_version_test.go
@@ -138,6 +138,19 @@ func (s *DeploymentVersionSuite) startVersionWorkflow(ctx context.Context, tv *t
 		})
 		a.NoError(err)
 		a.Equal(tv.DeploymentVersionString(), resp.GetWorkerDeploymentVersionInfo().GetVersion())
+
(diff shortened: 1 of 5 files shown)


tests/deployment_version_test.go
```diff
@@ -138,6 +138,19 @@ func (s *DeploymentVersionSuite) startVersionWorkflow(ctx context.Context, tv *t
...
+			DeploymentName: tv.DeploymentSeries(),
+		})
+		a.NoError(err)
+
+		var versionSummaryNames []string
+		for _, versionSummary := range newResp.GetWorkerDeploymentInfo().GetVersionSummaries() {
+			versionSummaryNames = append(versionSummaryNames, versionSummary.GetVersion())
+		}
+		a.Contains(versionSummaryNames, tv.DeploymentVersionString())
+
```
Shivs11, line 153: this was done because just verifying a version workflow was not enough - we want a strong guarantee in that the deployment workflow should be created + the version should be registered in the deployment workflow's local list

a test was flaking and this fixed it

special thanks to @carlydf for the pair programming session on this one!

carlydf (approved):

PR #7267, opened 2025-02-06
Describe closed Deployment Workflows returns NotFound + test ListDeployments excludes closed

carlydf:
## What changed?
Make describe closed wf return not found and add tests


ShahabT (approved):
This works, but I wonder if it was better for the query handler to just return an error when the wf is closed rather than pushing logic of checking `closed` to the client. That way, keeping just a in-memory bool in wf memory would be enough, without need for augmenting protos.

service/worker/workerdeployment/client.go
```diff
@@ -281,19 +281,21 @@ func (d *ClientImpl) DescribeVersion(
...
 			},
-			Query: &querypb.WorkflowQuery{QueryType: QueryDescribeVersion},
+			Query:                &querypb.WorkflowQuery{QueryType: QueryDescribeVersion},
+			QueryRejectCondition: enumspb.QUERY_REJECT_CONDITION_NOT_OPEN,
 		},
 	}
 
 	res, err := d.historyClient.QueryWorkflow(ctx, req)
 	if err != nil {
-		var notFound *serviceerror.NotFound
```
ShahabT, line 290: why are we deleting this part, isn't possingle for this err to be notFound?

Shivs11: ah sorry, this is my bad - restored

PR #7268, opened 2025-02-06
Use MutableSideEffect to access dynamic config from inside workflows

carlydf:
## What changed?
Use MutableSideEffect to access dynamic config from inside workflows.

## Why?
To prevent non-determinism in the workflow code.
To allow the workflows to use the latest dynamic config values at all times.

## How did you test it?
Tests are failing for other reasons, so I'm unable to test until I pull in those changes.


service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -82,11 +91,48 @@ func DrainageWorkflowWithDurations(visibilityGracePeriod, refreshInterval time.D
...
+			refresh, err := getRefreshInterval(ctx, dc, ns)
+			if err != nil {
+				return err
+			}
+			_ = workflow.Sleep(ctx, refresh)
 		}
 	}
 }
+
+func getRefreshInterval(ctx workflow.Context, dc *dynamicconfig.Collection, ns string) (time.Duration, error) {
```
dnr, line 106: Can we consolidate all these so there's one mutablesideeffect that returns a struct of all the needed values?

service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -82,11 +91,48 @@ func DrainageWorkflowWithDurations(visibilityGracePeriod, refreshInterval time.D
...
+func durationEq(a, b interface{}) bool {
+	aDur, ok := a.(time.Duration)
+	if !ok {
+		return false
+	}
+	bDur, ok := b.(time.Duration)
+	if !ok {
+		return false
+	}
+	return aDur == bDur
```
dnr, line 137: does this not work?
```suggestion
	return a == b
```

service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -82,11 +91,48 @@ func DrainageWorkflowWithDurations(visibilityGracePeriod, refreshInterval time.D
...
+			if err != nil {
+				return err
+			}
+			_ = workflow.Sleep(ctx, refresh)
 		}
 	}
 }
+
+func getRefreshInterval(ctx workflow.Context, dc *dynamicconfig.Collection, ns string) (time.Duration, error) {
+	get := func(ctx workflow.Context) interface{} {
```
dnr, line 107: use `any` everywhere, it's more readable

service/worker/workerdeployment/workflow.go
```diff
@@ -154,7 +161,7 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 	// Note, if update requests come in faster than they
 	// are handled, there will not be a moment where the workflow has
 	// nothing pending which means this will run forever.
-	return workflow.NewContinueAsNewError(ctx, Workflow, d.WorkerDeploymentWorkflowArgs)
+	return workflow.NewContinueAsNewError(ctx, WorkflowWithDC(d.dc), d.WorkerDeploymentWorkflowArgs)
```
dnr, line 164: Use the name here, I'm worried it'll get confused with all the closures.
```suggestion
	return workflow.NewContinueAsNewError(ctx, WorkerDeploymentWorkflowType, d.WorkerDeploymentWorkflowArgs)
```

service/worker/workerdeployment/workflow.go
```diff
@@ -477,7 +484,26 @@ func (d *WorkflowRunner) validateAddVersionToWorkerDeployment(version string) er
...
 		}
 	}
+	return nil
+}
 
+func (d *WorkflowRunner) checkMaxVersions(ctx workflow.Context) error {
+	getMaxVersionsInDeployment := func(ctx workflow.Context) interface{} {
+		return dynamicconfig.MatchingMaxVersionsInDeployment.Get(d.dc)(d.a.namespace.Name().String())
+	}
+	intEq := func(a, b interface{}) bool {
```
dnr, line 494: same == here?

service/worker/workerdeployment/workflow.go
```diff
@@ -477,7 +484,26 @@ func (d *WorkflowRunner) validateAddVersionToWorkerDeployment(version string) er
...
+		aInt, _ := a.(int)
+		bInt, _ := b.(int)
+		return aInt == bInt
+	}
+	var maxVersions int
+	if err := workflow.MutableSideEffect(ctx, "getMaxVersions", getMaxVersionsInDeployment, intEq).Get(&maxVersions); err != nil {
+		d.logger.Error("error decoding max versions: ", err)
+		return err
+	}
+	if len(d.State.Versions) >= maxVersions {
```
dnr, line 504: mixing the sideeffect and logic in one function feels a little weird, can you have one function to handle all the side effect business, returning a struct of concrete values (or a single one) and another function for the workflow logic (comparison)?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -411,7 +411,9 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
 	err = workflow.Await(ctx, func() bool { return d.VersionState.StartedDeploymentWorkflow })
 	if err != nil {
 		d.logger.Error("Update canceled before deployment workflow started")
-		return err
+		// TODO: This is likely due to too many deployments, but make sure we exculed other possible errors here and send a proper error message all the time.
```
carlydf, line 414:
```suggestion
		// TODO (Carly): This is likely due to too many deployments, but make sure we exculed other possible errors here and send a proper error message all the time.
```

service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -34,59 +34,75 @@ import (
...
+			return err
+		}
+		if info.Status == enumspb.VERSION_DRAINAGE_STATUS_DRAINED {
+			return nil
+		}
+		refresh, err := getSafeDurationConfig(ctx, "getDrainageRefreshInterval", unsafeRefreshIntervalGetter, defaultVisibilityRefresh)
+		if err != nil {
+			return err
 		}
+		_ = workflow.Sleep(ctx, refresh)
```
dnr, line 106: the delay of a workflow.Sleep is allowed to change without determinism error, so actually you can use the config directly here. also ok to use sideeffect to consistent if you want.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -438,8 +440,22 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
...
+	err = workflow.ExecuteActivity(activityCtx, d.a.AddVersionToWorkerDeployment, &deploymentspb.AddVersionToWorkerDeploymentRequest{
+		DeploymentName: d.VersionState.Version.GetDeploymentName(),
+		UpdateArgs: &deploymentspb.AddVersionUpdateArgs{
+			Version:    worker_versioning.WorkerDeploymentVersionToString(d.VersionState.Version),
+			CreateTime: d.VersionState.CreateTime,
+		},
+		RequestId: d.newUUID(ctx),
+	}).Get(ctx, nil)
+	if err != nil {
+		// TODO: make sure the error message that goes to the user is informative and has the limit mentioned
```
carlydf, line 454:
```suggestion
		// TODO (Carly): make sure the error message that goes to the user is informative and has the limit mentioned
```

service/worker/workerdeployment/version_workflow.go
```diff
@@ -411,7 +411,9 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
 	err = workflow.Await(ctx, func() bool { return d.VersionState.StartedDeploymentWorkflow })
 	if err != nil {
 		d.logger.Error("Update canceled before deployment workflow started")
-		return err
+		// TODO: This is likely due to too many deployments, but make sure we exculed other possible errors here and send a proper error message all the time.
+		// TODO: mention the limit in here or make sure matching does in the error returned to the poller
```
carlydf, line 415:
```suggestion
		// TODO (Carly): mention the limit in here or make sure matching does in the error returned to the poller
```

service/worker/workerdeployment/util.go
```diff
@@ -139,3 +144,18 @@ func DecodeWorkerDeploymentMemo(memo *commonpb.Memo) *deploymentspb.WorkerDeploy
 	}
 	return &workerDeploymentWorkflowMemo
 }
+
+func getSafeDurationConfig(ctx workflow.Context, id string, unsafeGetter func() any, defaultValue time.Duration) (time.Duration, error) {
```
dnr, line 148: this can easily be generic if you want to do that now

service/worker/workerdeployment/version_workflow.go
dnr, line 414:
```suggestion
		// TODO: This is likely due to too many deployments, but make sure we excluded other possible errors here and send a proper error message all the time.
```

service/worker/workerdeployment/workflow.go
```diff
@@ -483,21 +488,36 @@ func (d *WorkflowRunner) validateAddVersionToWorkerDeployment(args *deploymentsp
...
-	if len(d.State.Versions) >= defaultMaxVersions {
-		err := d.tryDeleteVersion(ctx)
-		if err != nil {
-			return err
-		}
+
+	maxVersions := d.getMaxVersions(ctx)
+
+	if len(d.State.Versions) >= maxVersions {
+		return temporal.NewApplicationError(fmt.Sprintf("cannot add version, already at max versions %d", maxVersions), errTooManyVersions)
```
carlydf, line 518:
```suggestion
                err := d.tryDeleteVersion(ctx)
                if err != nil {
                		return temporal.NewApplicationError(fmt.Sprintf("cannot add version, already at max versions %d", maxVersions), errTooManyVersions)
                }
```

carlydf: important

carlydf: the test that tests this was skipped.. which is why it didn't catch this regfression

service/worker/workerdeployment/workflow.go
```diff
@@ -483,21 +488,36 @@ func (d *WorkflowRunner) validateAddVersionToWorkerDeployment(args *deploymentsp
 			return temporal.NewApplicationError("deployment version already registered", errVersionAlreadyExistsType)
 		}
 	}
-
 	return nil
 }
 
+func (d *WorkflowRunner) getMaxVersions(ctx workflow.Context) int {
```
dnr, line 494: make `getSafeDurationConfig` generic and use it here?

dnr (approved):
some suggestions and nitpicks but nothing blocking

service/worker/workerdeployment/workflow.go
ShahabT: Oh, good catch!

service/worker/workerdeployment/util.go
ShahabT: have to do it later

PR #7270, opened 2025-02-06
[Reset] Carry over ChildrenInitializedPostResetPoint when continue-as-new

gow:
## What changed?
Copying `executionInfo.ChildrenInitializedPostResetPoint` over into new run in continue-as-new operation.

## Why?
This is needed so that we can apply appropriate reset policy (if needed) to children started in new runs.

## How did you test it?
manual test.

## Potential risks
N/A

## Documentation
Documentation pending.

## Is hotfix candidate?
No

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -555,6 +555,7 @@ func NewMutableStateInChain(
 
 	// carry over necessary fields from current mutable state
 	newMutableState.executionInfo.WorkflowExecutionTimerTaskStatus = currentMutableState.GetExecutionInfo().WorkflowExecutionTimerTaskStatus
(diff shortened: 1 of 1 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -555,6 +555,7 @@ func NewMutableStateInChain(
 
 	// carry over necessary fields from current mutable state
 	newMutableState.executionInfo.WorkflowExecutionTimerTaskStatus = currentMutableState.GetExecutionInfo().WorkflowExecutionTimerTaskStatus
+	newMutableState.executionInfo.ChildrenInitializedPostResetPoint = currentMutableState.GetExecutionInfo().ChildrenInitializedPostResetPoint
```
gow, line 558: This should be ok since entries in `currentMutableState.GetExecutionInfo().ChildrenInitializedPostResetPoint` will not be updated beyond this point. If there are any possibility of that happening, we should deep copy the map contents here. Please let me know.

yycptt (approved):

PR #7271, opened 2025-02-06
try to delete version if we are at limit

carlydf:
## What changed?
Make updates and signals use args
Try to delete oldest to newest version if we hit limit

## How did you test it?
New functional test


proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -182,6 +189,24 @@ message SyncVersionStateResponse {
...
+// Sent from Drainage child workflow to Version parent
+message SyncDrainageInfoSignalArgs {
+    temporal.api.deployment.v1.VersionDrainageInfo drainage_info = 1;
+}
+
+// Sent from Version workflow to Worker Deployment workflow
+message SyncDrainageStatusSignalArgs {
+    string version = 1;
+    temporal.api.enums.v1.VersionDrainageStatus drainage_status = 2;
+}
```
carlydf, line 207: these are not currently used, I wanted to keep this PR short, but I will have another PR using them so that we can keep the VersionSummaries up to date in the deployment WF

tests/worker_deployment_version_test.go
```diff
@@ -475,20 +483,94 @@ func (s *DeploymentVersionSuite) TestDeleteVersion_ValidDelete() {
...
+	s.signalAndWaitForDrained(ctx, tvs[1])
+	s.signalAndWaitForDrained(ctx, tvs[2])
+	// Wait for pollers going away
+	s.waitForNoPollers(ctx, tvs[1])
+	s.waitForNoPollers(ctx, tvs[2])
+
+	fmt.Println("CARLY: AFTER DRAINAGE")
+
+	// try to add a version again, and it succeeds, after deleting the second version but not the third (both are eligible)
+	// TODO: This fails if I try to add tvMax again...
```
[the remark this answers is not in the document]
carlydf: or, I think it is a problem but I'm not sure how to fix it. Hopefully in most cases when this is happening, at least one out of 100 versions will be eligible deletion so the first add would not fail.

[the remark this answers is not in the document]
carlydf: I think it's not worth fixing this until everything else we need to do is done

service/worker/workerdeployment/workflow.go
```diff
@@ -467,35 +472,68 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+		// cmp(a, b) should return a negative number when a < b, a positive number when a > b,
+		// and zero when a == b or a and b are incomparable in the sense of a strict weak ordering.
+		if a.GetCreateTime().AsTime().After(b.GetCreateTime().AsTime()) {
+			return 1
+		} else if a.GetCreateTime().AsTime().Before(b.GetCreateTime().AsTime()) {
+			return -1
+		}
+		return 0
+	})
+	for _, v := range d.State.Versions {
```
ShahabT, line 524: Do you want to check for version not being drained here, or that info is not sent to the deployment wf yet?

Shivs11 (approved):

service/worker/workerdeployment/workflow.go
carlydf: it's not sent to the deployment wf yet, but when I do that I will

PR #7275, opened 2025-02-06
Refactor: remove enablerepicator setting from functional test config

alexshtin:
## What changed?

Refactor: remove `enablerepicator` setting from functional test config.

## Why?

This is legacy setting which is not really used any more, but there were configs which are different just in this setting value. Unified these configs too.

I hope all these test configs will go away soon, but I want to do it slowly.

## How did you test it?

Run tests.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.


tests/testcore/onebox.go
```diff
@@ -541,9 +540,6 @@ func (c *TemporalImpl) startWorker() {
 		CurrentClusterName:       c.clusterMetadataConfig.CurrentClusterName,
 		ClusterInformation:       maps.Clone(c.clusterMetadataConfig.ClusterInformation),
 	}
-	if c.workerConfig.EnableReplicator {
-		clusterConfigCopy.EnableGlobalNamespace = true
```
stephanos, line 545: So this has no actual effect then?

stephanos (approved):

tests/testcore/onebox.go
alexshtin: For non-xdc functional tests we don't need global namespace support: they all are for single cluster. I am setting this explicitly in `xdcBaseSuite` and actually, I see no harm enabling it always. I will probably do it later.

PR #7279, opened 2025-02-06
Allow batched user data updates to succeed individually

dnr:
## What changed?
Following from #7039, the batch update isn't required to be transactional. The persistence interface returns an error if any failed, but update success is reported individually.

## Why?
ease adaptation to less-transactional stores

## How did you test it?
updated integration test


stephanos (approved):
LGTM

PR #7280, opened 2025-02-06
[Reset] Gate auto restarting of children behind the feature flag

gow:
## What changed?
This PR gates this feature behind the feature flag before the full release.

## Why?
The feature is brand new and we may not have identified all edge cases. So it's better to roll it out slowly.

## How did you test it?
Manually flipped the flag off and verified that the old child was not terminated.

## Potential risks
N/A

## Documentation
N/A

## Is hotfix candidate?
No.


service/history/ndc/workflow_resetter.go
```diff
@@ -762,6 +767,7 @@ func (r *workflowResetterImpl) reapplyEventsFromBranch(
 	nextEventID int64,
 	branchToken []byte,
 	resetReapplyExcludeTypes map[enumspb.ResetReapplyExcludeType]struct{},
+	allowResetWithPendingChildren bool,
 	childrenInitializedAfterReset map[string]*persistencespb.ResetChildInfo,
```
gow, line 771: Thought of sending a nil value of `childrenInitializedAfterReset` (when allowResetWithPendingChildren is false) instead of passing the flag all the way here. But check for nil map and then populating it is an antipattern. So just decided to pass the flag explicitly.

yycptt (approved):

PR #7281, opened 2025-02-06
Fix history batch validations

prathyushpv:
## What changed?
Move history batch validations from ExecutionManager to history/api package.

## Why?
ReadRawHistoryBranch in ExecutionManager may not return full list of history events in some situations. So we cannot check if the history is complete in that layer.
Moving this check to a layer above that which will have a full list of history events

## How did you test it?
Unit tests


common/persistence/history_manager.go
```diff
@@ -955,17 +930,51 @@ func (m *executionManagerImpl) readHistoryBranch(
...
+			tag.TokenLastEventID(token.LastEventID),
+		}
+	}
+
 	for _, batch := range dataBlobs {
 		events, err := m.serializer.DeserializeEvents(batch)
 		if err != nil {
 			return nil, nil, nil, nil, dataSize, err
 		}
+		if len(events) == 0 {
```
prathyushpv, line 954: This check was moved from this function to readRawHistoryBranchAndFilter() in the previous PR. readRawHistoryBranchAndFilter() is called from this function. So this check was not needed here. Now I moved that check from readRawHistoryBranchAndFilter() to a layer above. Putting back this check here.

service/history/api/get_history_util.go
```diff
@@ -91,15 +123,15 @@ func GetRawHistory(
 			return nil, nil, err
 		}
 
-		for _, event := range transientWorkflowTaskInfo.HistorySuffix {
-			blob, err := shard.GetPayloadSerializer().SerializeEvent(event, enumspb.ENCODING_TYPE_PROTO3)
+		if len(transientWorkflowTaskInfo.HistorySuffix) > 0 {
+			blob, err := shard.GetPayloadSerializer().SerializeEvents(transientWorkflowTaskInfo.HistorySuffix, enumspb.ENCODING_TYPE_PROTO3)
```
prathyushpv, line 127: If we create a history blob using SerializeEvent(), it can not be deserialized into History object in frontend. We have to use SerializeEvents()

common/persistence/history_manager.go
```diff
@@ -813,20 +812,9 @@ func (m *executionManagerImpl) readRawHistoryBranchAndFilter(
...
-	allEvents := make([]*historyspb.StrippedHistoryEvent, 0, request.PageSize)
 	if len(nodes) > 0 {
 		dataBlobs = make([]*commonpb.DataBlob, len(nodes))
 		for index, node := range nodes {
-			events, err := m.serializer.DeserializeStrippedEvents(node.Events)
-			if err != nil {
-				return nil, nil, nil, nil, 0, err
-			}
-			allEvents = append(allEvents, events...)
-			err = ValidateBatch(events, branchToken, token, m.logger)
```
yycptt, line 825: we can still validate batch here?

prathyushpv: We can. But we will have to deserialize to stripped events twice. Once here and once in [get_history_util.go](https://github.com/temporalio/temporal/pull/7281/files#diff-1fa11c967788bd20a3f238ddc043d8d99f20e92d8d295ca9416024ee8d5b6920) to verify history is complete.

yycptt (approved):

PR #7282, opened 2025-02-06
ForceCAN bug fix

Shivs11:
## What changed?

- ForceCAN bug fix


carlydf (approved):
nice! Does the same have to be done in the Worker Deployment workflow?

PR #7284, opened 2025-02-06
Log prometheus listener error by default

yycptt:
## What changed?

- Log error instead of panic for existing tally users when prometheus listener can't be created.
- No behavior change for existing OTEL users

## Why?

- For backward compatibility

### DIFF
--- a/common/metrics/opentelemetry_provider.go
+++ b/common/metrics/opentelemetry_provider.go
@@ -107,7 +107,7 @@ func initPrometheusListener(
 	config *PrometheusConfig,
 	reg *prometheus.Registry,
 	logger log.Logger,
-	onError func(e error),
+	onListenerError func(error),
 ) *http.Server {
 	handlerPath := config.HandlerPath
 	if handlerPath == "" {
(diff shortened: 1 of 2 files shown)


common/metrics/opentelemetry_provider.go
```diff
@@ -102,7 +103,12 @@ func NewOpenTelemetryProvider(
 	return reporter, nil
 }
 
-func initPrometheusListener(config *PrometheusConfig, reg *prometheus.Registry, logger log.Logger) *http.Server {
+func initPrometheusListener(
+	config *PrometheusConfig,
+	reg *prometheus.Registry,
+	logger log.Logger,
+	onError func(e error),
```
dnr, line 110: maybe call it onListenError to be more specific
```suggestion
	onListenError func(error),
```

dnr: I'd also be fine with just a `listenErrorIsFatal bool` :shrug:

dnr (approved):

common/metrics/opentelemetry_provider.go
yycptt: makes sense. switched to bool flag.

PR #7287, opened 2025-02-07
Use Worker Deployment Search Attributes

carlydf:
## What changed?
Write to WorkerDeployment, WorkerDeploymentVersion, and WorkflowVersioningBehavior Search Attributes in all the places where we used to only write to BuildIds SA.

If using worker deployments and pinned, write `pinned:version` to BuildIds list instead of `pinned:series_name:build_id`.

Use `pinned:version` instead of `pinned:series_name:build_id` for deployment version drainage status.

## How did you test it?
- Confirmed that all Deployment Reachability tests pass
- New tests (in progress)


service/history/workflow/mutable_state_impl.go
```diff
@@ -2846,17 +2846,26 @@ func (ms *MutableStateImpl) UpdateBuildIdAssignment(buildId string) error {
...
-// pinned:<deployment_series_name>:<deployment_build_id> to the BuildIds search attribute,
-// if it does not already exist there. The deployment will be execution_info.deployment, or the override deployment if
-// a pinned override is set.
-// For all other workflows (ms.GetEffectiveVersioningBehavior() != PINNED), this will append a tag based on the
-// workflow's versioning status.
-func (ms *MutableStateImpl) updateBuildIdsSearchAttribute(stamp *commonpb.WorkerVersionStamp, maxSearchAttributeValueSize int) error {
-	changed, err := ms.addBuildIdToSearchAttributesWithNoVisibilityTask(stamp, maxSearchAttributeValueSize)
+// Sets TemporalWorkerDeployment to the override DeploymentName or`ms.executionInfo.WorkerDeploymentName` if present.
+// Sets TemporalWorkerDeploymentVersion to the override PinnedVersion or `ms.executionInfo.VersioningInfo.Version` if present.
+// Sets TemporalWorkflowVersioningBehavior to the override Behavior or `ms.executionInfo.VersioningInfo.Behavior` if specified.
```
ShahabT, line 2854: this should be Pinned if override PinnedVersion is present

service/history/workflow/mutable_state_impl.go
```diff
@@ -2846,17 +2846,26 @@ func (ms *MutableStateImpl) UpdateBuildIdAssignment(buildId string) error {
...
+// For pinned workflows using WorkerDeployment APIs (ms.GetEffectiveVersioningBehavior() == PINNED &&
+// ms.executionInfo.VersioningInfo.Version != ""), this will append a tag formed as `pinned:<version>`
+// to the BuildIds search attribute, if it does not already exist there. The version used will be the
+// effective version of the workflow (aka, the override version if override is set).
+//
+// If deprecated Deployment-based APIs are in use and the workflow is pinned, `pinned:<deployment_series_name>:<deployment_build_id>`
+// will be appended to the BuilIds list if it is not already present. The deployment will be
+// the effective deployment of the workflow (aka the override deployment_series and build_id if set).
+//
+// For all other workflows (ms.GetEffectiveVersioningBehavior() != PINNED), this will append a tag  to BuildIds
```
ShahabT, line 2865: the build Id (and not version) will still be appended even for pinned wfs, right?

service/history/workflow/mutable_state_impl.go
```diff
@@ -7507,6 +7621,36 @@ func (ms *MutableStateImpl) GetEffectiveDeployment() *deploymentpb.Deployment {
...
+	}
+	return ""
+}
+
+func (ms *MutableStateImpl) GetWorkflowVersioningBehaviorSA() string {
+	s := ms.GetEffectiveVersioningBehavior().String()
+	if s == enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED.String() {
+		s = ""
+	}
+	return s
```
ShahabT, line 7651: alse return pinned, if override is pinned.

service/history/workflow/mutable_state_impl.go
carlydf: oh yes. `GetEffectiveBehavior` is what is implemented, the comment is just wrong. will fix

carlydf: the comment says `the override Behavior or `ms.executionInfo.VersioningInfo.Behavior` if specified`, do you think it's not clear that override takes priority?

service/history/workflow/mutable_state_impl.go
carlydf: ah I should have realized that was something we wanted to do. I didn't do that for Deployment APIs in november either. Will do that for both here

carlydf: Just to be clear, do you mean `versioned:<build_id>` ?

service/history/workflow/mutable_state_impl.go
carlydf: `GetEffectiveVersioningBehavior()` takes into account the override, and `.String()` will return `Pinned`, `AutoUpgrade`, or `Unspecified` for each behavior. So this already returns pinned if the override is pinned.

All I'm doing here is converting `Unspecified`->`""`.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2914,7 +2946,18 @@ func (ms *MutableStateImpl) addBuildIdToLoadedSearchAttribute(
 	if !stamp.GetUseVersioning() && effectiveBehavior == enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED { // unversioned workflows may still have non-nil deployment, so we don't check deployment
 		newValues = append(newValues, worker_versioning.UnversionedSearchAttribute)
 	} else if effectiveBehavior == enumspb.VERSIONING_BEHAVIOR_PINNED {
-		newValues = append(newValues, worker_versioning.PinnedBuildIdSearchAttribute(ms.getPinnedDeployment()))
+		vInfo := ms.GetExecutionInfo().GetVersioningInfo()
+		if vInfo.GetVersioningOverride().GetPinnedVersion() != "" {
+			if vInfo.GetVersion() != "" { // worker deployments
+				newValues = append(newValues, worker_versioning.PinnedBuildIdSearchAttribute(nil, vInfo.GetVersioningOverride().GetPinnedVersion()))
```
ShahabT, line 2952: similar to `assigned:<bld>` and `unversioned`, we only want one `pinned:...` value in the list. old ones should be added as ordinary values (I guess they still have a `versioned:` or `unversioned:` prefixes).

carlydf: ah ok, I think I didn't handle that correctly in the last release. Let me refactor this ..

service/history/workflow/mutable_state_impl.go
ShahabT: yes `versioned:<build_id>` same as what would appear in v2.

service/history/workflow/mutable_state_impl.go
ShahabT: it's good

service/history/workflow/mutable_state_impl.go
ShahabT: sorry I didn't pay attention to the `Effective` part. But we don't want to use effective, because of transitions. If the wf is pinned but there is an ongoing transition GetEffectiveVersioningBehavior will returned autoUpgrade but we want it to be pinned and consistent with the DeploymentVersion SA.

service/history/workflow/mutable_state_impl.go
carlydf: I fixed the sentence

service/history/workflow/mutable_state_impl.go
carlydf: is that right? GetEffectiveVersioningBehavior currently doesn't consider transition, and transition does not have a behavior field. Lmk if that should change.

```
// GetEffectiveVersioningBehavior returns the effective versioning behavior in the following
// order:
//  1. VersioningOverride.Behavior: this is returned when user has set a behavior override
//     at wf start time, or later via UpdateWorkflowExecutionOptions.
//  2. Behavior: this is returned when there is no override (most common case). Behavior is
//     set based on the worker-sent deployment in the latest WFT completion.
func GetEffectiveVersioningBehavior(versioningInfo *workflowpb.WorkflowExecutionVersioningInfo) enumspb.VersioningBehavior {
	if versioningInfo == nil {
		return enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED
	} else if override := versioningInfo.GetVersioningOverride(); override != nil {
		return override.GetBehavior()
	}
	return versioningInfo.GetBehavior()
}
```

ShahabT: oh ok. I think it's correct now because transitions never start on pinned wfs. but once we have safe moves I think we need to update this.

ShahabT: Still better for SAs to have their own function and not rely on Effective.

carlydf: ok I made it independent 👍

service/history/workflow/mutable_state_impl.go
```diff
@@ -2932,6 +2932,49 @@ func (ms *MutableStateImpl) getPinnedDeployment() *deploymentpb.Deployment {
...
+func (ms *MutableStateImpl) getVersionedBuildId() string {
+	versioningInfo := ms.executionInfo.GetVersioningInfo()
+	if override := versioningInfo.GetVersioningOverride(); override != nil {
+		if versioningInfo.GetVersion() != "" { // worker deployments
+			if override.GetBehavior() == enumspb.VERSIONING_BEHAVIOR_PINNED {
+				v, _ := worker_versioning.WorkerDeploymentVersionFromString(override.GetPinnedVersion())
+				return v.GetBuildId()
+			}
+			if override.GetBehavior() == enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE {
+				return versioningInfo.GetVersion()
```
ShahabT, line 2959: should return build id only

service/history/workflow/mutable_state_impl.go
```diff
@@ -2910,36 +2985,47 @@ func (ms *MutableStateImpl) addBuildIdToLoadedSearchAttribute(
...
-			if existingValue == buildId {
-				found = true
-			}
-			if !worker_versioning.IsUnversionedOrAssignedBuildIdSearchAttribute(existingValue) {
-				newValues = append(newValues, existingValue)
-			}
+		buildId = worker_versioning.VersionStampToBuildIdSearchAttribute(stamp)
+	} else { // versioning behaviors
+		buildId = worker_versioning.VersionedBuildIdSearchAttribute(ms.getVersionedBuildId())
+	}
```
ShahabT, line 3012: I think you should only add build ID if the stamp is given. no need to extract build ID using VersionedBuildIdSearchAttribute.
Note that `versioned:..` `unversioned:..` are supposed to be populated only for builds which completed wfts.

Instead of this you should ensure that afterAddWorkflowTaskCompletedEvent calls updateBuildIdsAndDeploymentSearchAttributes with a stamp (which it does). But what we need to additionally cover is to make the updateBuildIdsAndDeploymentSearchAttributes call with a stamp made out of attrs.VersioningBehavior and attrs.WorkerDeploymentVersion even if attrs.GetWorkerVersion() is nill. (Note that we've deprecated WorkerVersionStamp since v3.1)
Once you do ^ then you can only worry about adding `versioned:..` `unversioned:..` items IF stamp is not nil.

carlydf: ah ok. makes sense, let me check

ShahabT (approved):

tests/worker_deployment_version_test.go
```diff
@@ -370,30 +353,119 @@ func (s *DeploymentVersionSuite) TestDrainageStatus_SetCurrentVersion_NoOpenWFs(
...
-	// todo carly: test with open workflows on the draining version that then complete
+	// tv1 should now be "draining" for visibilityGracePeriod duration
+	s.checkVersionDrainage(ctx, tv1, &deploymentpb.VersionDrainageInfo{
+		Status:          enumspb.VERSION_DRAINAGE_STATUS_DRAINING,
+		LastChangedTime: nil, // don't test this now
+		LastCheckedTime: nil, // don't test this now
+	}, false, false)
+
+	// tv1 should still be "draining" for visibilityGracePeriod duration
+	s.checkVersionDrainage(ctx, tv1, &deploymentpb.VersionDrainageInfo{
```
ShahabT, line 417: this doesn't wait for refresh period because Eventually will see the desired result in the first try.

To check that drainage happened after grace+drainage period, you can instead use the drainageChangedTime after you get Drained and make sure it was at least grace+drainage time after the time you v1 became non-current.

carlydf: The CI is almost passed, so I'm going to let this merge first, and then make this test change in a future PR

PR #7292, opened 2025-02-07
Eliminate DescribeWorkerDeployment fan-out

ShahabT:
## What changed?

- Eliminate the DescribeWorkerDeployment fan-out to Versions for getting the summaries.
- Sync drainage status with the deployment workflow when it changes.
- Drainage WF now can CaN!
- Move drainage args to it's own message to be future proof.
- Made WorkerDeploymentLocalState.versions a map for easier access to versions.


service/worker/workerdeployment/util.go
```diff
@@ -60,24 +60,22 @@ const (
 	DeleteDeployment             = "delete-deployment"                // for WorkerDeployment wfs
 
 	// Signals
-	ForceCANSignalName      = "force-continue-as-new" // for Worker Deployment Version _and_ Worker Deployment wfs
-	SyncDrainageSignalName  = "sync-drainage-status"
-	TerminateDrainageSignal = "terminate-drainage"
+	ForceCANSignalName       = "force-continue-as-new" // for Worker Deployment Version _and_ Worker Deployment wfs
+	SyncDrainageSignalName   = "sync-drainage-status"
```
carlydf, line 64:
```suggestion
	SyncDrainageSignalName   = "sync-drainage-status" // for Worker Deployment Version _and_ Worker Deployment wfs
```

service/worker/workerdeployment/version_workflow.go
```diff
@@ -209,7 +210,7 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
 	d.logger.Debug("Version doing continue-as-new")
 	nextArgs := d.WorkerDeploymentVersionWorkflowArgs
 	nextArgs.VersionState = d.VersionState
-	return workflow.NewContinueAsNewError(ctx, VersionWorkflow, nextArgs)
```
carlydf, line 212: did `VersionWorkflow` not work? like, is passing `WorkerDeploymentVersionWorkflowType` an optimization or a bug fix (just curious)

carlydf (approved):

service/worker/workerdeployment/util.go
ShahabT: I'm not using SyncDrainageSignalName in the deployment wf

service/worker/workerdeployment/version_workflow.go
ShahabT: I think in simple situations it works, but once you start overriding the name in register call (which we do) and especially if you start wrapping the wf func in a closure etc it becomes dangerous IMO.

service/worker/workerdeployment/client.go
```diff
@@ -1272,14 +1272,10 @@ func (d *ClientImpl) deploymentStateToDeploymentInfo(ctx context.Context, namesp
...
-		if err != nil {
-			return nil, err
-		}
 		workerDeploymentInfo.VersionSummaries = append(workerDeploymentInfo.VersionSummaries, &deploymentpb.WorkerDeploymentInfo_WorkerDeploymentVersionSummary{
-			Version:        versionInfo.Version,
-			CreateTime:     versionInfo.CreateTime,
-			DrainageStatus: versionInfo.GetDrainageInfo().GetStatus(),
+			Version:        v.GetVersion(),
+			CreateTime:     v.GetCreateTime(),
+			DrainageStatus: v.GetDrainageStatus(),
```
Shivs11, line 1278: nice!

service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -58,7 +61,7 @@ func DrainageWorkflow(
 	})
 
 	// Set status = DRAINING and then sleep for visibilityGracePeriod (to let recently-started workflows arrive in visibility)
-	if first { // skip if resuming after the parent continued-as-new
+	if !args.IsCan { // skip if resuming after the parent continued-as-new
 		parentWf := workflow.GetInfo(ctx).ParentWorkflowExecution
 		now := timestamppb.Now()
 		drainingInfo := &deploymentpb.VersionDrainageInfo{
```
Shivs11, line 67: sanity check - earlier on, the point of using `first` was that if the parent version workflow, already having been set to `draining`, CAN'ned, the block ranging from lines 64 to 81 would not execute since the drainage state is already draining.

I know there is no harm in sending the signal again and sleeping for the grace period doesn't really make a difference but my point was that maybe this was forgotten when being implemented

cc - @ShahabT

ShahabT: fixing in https://github.com/temporalio/temporal/pull/7293/commits/cdd5d9afa9b0de717790a8079b98b912f8e54dd8

PR #7293, opened 2025-02-07
Test and fix Deployment SAs

ShahabT:
## What changed?

Making sure the following SAs are populated correctly for Versioning v3.1
- TemporalWorkerDeployment 
- TemporalWorkerDeploymentVersion 
- TemporalWorkflowVersioningBehavior
- BuildIds

### DIFF
--- a/service/history/api/respondworkflowtaskcompleted/api.go
+++ b/service/history/api/respondworkflowtaskcompleted/api.go
@@ -215,10 +215,12 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 	}
 
 	behavior := request.GetVersioningBehavior()
-	if behavior != enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED && request.GetDeployment() == nil {
+	//nolint:staticcheck // SA1019 deprecated Deployment will clean up later
+	if behavior != enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED && request.GetDeployment() == nil &&
+		(request.GetDeploymentOptions() == nil || request.GetDeploymentOptions().GetWorkerVersioningMode() != enumspb.WORKER_VERSIONING_MODE_VERSIONED) {
 		// Mutable state wasn't changed yet and doesn't have to be cleared.
 		releaseLeaseWithError = false
-		return nil, serviceerror.NewInvalidArgument("deployment must be set when versioning behavior specified")
+		return nil, serviceerror.NewInvalidArgument("versioning behavior cannot be specified without deployment options being set with versioned mode")
 	}
 
 	assignedBuildId := ms.GetAssignedBuildId()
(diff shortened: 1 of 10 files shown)


service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -252,7 +254,10 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
...
 
 	// It's an error if the workflow has used versioning in the past but this task has no versioning info.
-	if ms.GetMostRecentWorkerVersionStamp().GetUseVersioning() && !request.GetWorkerVersionStamp().GetUseVersioning() {
+	if ms.GetMostRecentWorkerVersionStamp().GetUseVersioning() &&
+		//nolint:staticcheck // SA1019 deprecated stamp will clean up later
+		!request.GetWorkerVersionStamp().GetUseVersioning() &&
+		request.GetDeploymentOptions().GetWorkerVersioningMode() != enumspb.WORKER_VERSIONING_MODE_VERSIONED {
 		// Mutable state wasn't changed yet and doesn't have to be cleared.
 		releaseLeaseWithError = false
 		return nil, serviceerror.NewInvalidArgument("Workflow using versioning must continue to use versioning.")
```
carlydf, line 263: how does "ramp to unversioned" work with this? If someone uses versioning in a version stamp, they can't ramp to unversioned? (It makes sense to me that v1/v2 can't ramp to unversioned, but v3.1 should be able to, right?)

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -478,7 +478,9 @@ func (handler *workflowTaskCompletedHandler) handleCommandScheduleActivity(
 
 	// TODO: versioning 3 allows eager activity dispatch for both pinned and unpinned workflows, no
 	// special consideration is need. Remove the versioning logic from here. [cleanup-old-wv]
-	oldVersioningUsed := handler.mutableState.GetMostRecentWorkerVersionStamp().GetUseVersioning()
+	oldVersioningUsed := handler.mutableState.GetMostRecentWorkerVersionStamp().GetUseVersioning() &&
+		// for V3 versioning it's ok to dispatch eager activities
+		handler.mutableState.GetEffectiveVersioningBehavior() == enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED
 	newVersioningUsed := handler.mutableState.GetExecutionInfo().GetAssignedBuildId() != ""
```
carlydf, line 484: is `newVersioningUsed` correct? it seems like this only checks that the WF never used versioning v2.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -580,7 +581,7 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
 
 	// stopped accepting new workflows --> start drainage child wf
 	if wasAcceptingNewWorkflows && !isAcceptingNewWorkflows {
-		d.startDrainage(ctx, true)
+		d.startDrainage(ctx, false)
```
carlydf, line 584: :o nice catch!

tests/versioning_3_test.go
```diff
@@ -1695,33 +1708,12 @@ func respondCompleteWorkflow(
...
-						Service:   tv.Service(),
-						Operation: tv.Operation(),
-						Input:     tv.Any().Payload(),
-					},
-				},
-			},
+		DeploymentOptions: &deploymentpb.WorkerDeploymentOptions{
+			BuildId:              tv.BuildID(),
+			DeploymentName:       tv.DeploymentSeries(),
+			WorkerVersioningMode: enumspb.WORKER_VERSIONING_MODE_VERSIONED,
```
carlydf, line 1715: do you need a version stamp here too?

carlydf (approved):

service/history/api/respondworkflowtaskcompleted/api.go
ShahabT: yeah they could not in v1-2. now that I exclude v3.1 from the condition it should be possible for v3.1 even if stamp is not sent.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
ShahabT: I think it is correct. once a wf uses v2 it cannot go back! in here I just want to exclude v3 from the check. not touching v1-2 suff.

tests/versioning_3_test.go
ShahabT: well, I someone whats to check the SAs after wf completion yes but I didn't have that use and left it out for now because wanted to make sure the request goes through without stamp as well.

PR #7295, opened 2025-02-07
[Reset] Handle completion events from restarted children

gow:
## What changed?
In this PR we are doing 2 things,
1. Cleaning up the entries in `ChildrenInitializedPostResetPoint`
2. Verifying that the results are indeed from the correct restarted child before accepting the completion event.

## Why?
- Cleans up the mutable state and reduces bloat.
- Avoids a race condition where a old child completed in between the time when new child Init is recorded and when it is actually started.

## How did you test it?
Added unit tests + manual testing.

## Potential risks
N/A

## Documentation
Pending

## Is hotfix candidate?
No

### DIFF
--- a/service/history/api/recordchildworkflowcompleted/api.go
+++ b/service/history/api/recordchildworkflowcompleted/api.go
@@ -157,7 +157,7 @@ func recordChildWorkflowCompleted(
 				return nil, consts.ErrChildExecutionNotFound
 			}
 
-			childrenInitializedAfterResetPoint := mutableState.GetExecutionInfo().GetChildrenInitializedPostResetPoint()
+			childrenInitializedAfterResetPoint := mutableState.GetChildrenInitializedPostResetPoint()
 			if len(childrenInitializedAfterResetPoint) > 0 {
 				// This parent was reset and it also has some children that potentially were restarted.
 				initiatedEvent, err := mutableState.GetChildExecutionInitiatedEvent(ctx, parentInitiatedID)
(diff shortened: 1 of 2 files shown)


service/history/api/recordchildworkflowcompleted/api.go
```diff
@@ -153,6 +157,28 @@ func recordChildWorkflowCompleted(
...
 
+			childrenInitializedAfterResetPoint := mutableState.GetChildrenInitializedPostResetPoint()
+			if len(childrenInitializedAfterResetPoint) > 0 {
+				// This parent was reset and it also has some children that potentially were restarted.
+				initiatedEvent, err := mutableState.GetChildExecutionInitiatedEvent(ctx, parentInitiatedID)
+				if err != nil {
+					return nil, consts.ErrChildExecutionNotFound
+				}
+				initiatedAttr := initiatedEvent.GetStartChildWorkflowExecutionInitiatedEventAttributes()
+				childID := fmt.Sprintf("%s:%s", initiatedAttr.GetWorkflowType().Name, initiatedAttr.GetWorkflowId())
```
yycptt, line 168: nit: we really should create a util function for this concatenation logic. I think I've seen it in 3 different places.

service/history/api/recordchildworkflowcompleted/api.go
```diff
@@ -153,6 +157,28 @@ func recordChildWorkflowCompleted(
 				return nil, consts.ErrChildExecutionNotFound
 			}
 
+			childrenInitializedAfterResetPoint := mutableState.GetChildrenInitializedPostResetPoint()
+			if len(childrenInitializedAfterResetPoint) > 0 {
+				// This parent was reset and it also has some children that potentially were restarted.
+				initiatedEvent, err := mutableState.GetChildExecutionInitiatedEvent(ctx, parentInitiatedID)
```
yycptt, line 163: hmm I don't think we need to load the initiatedEvent. Both child workflow type and child workflowID are available in ChildExecutionInfo in mutable state I think.

service/history/api/recordchildworkflowcompleted/api.go
```diff
@@ -153,6 +157,28 @@ func recordChildWorkflowCompleted(
...
+				_, ok := childrenInitializedAfterResetPoint[childID]
+				if ok {
+					// The child sending this request was restarted. Do not accept any forwarded completion events because the restarted child will directly send its completion event to this parent.
+					// Sometimes the old child will race to complete before being restarted and it also happens to have the same initialized event ID. In such cases the result will come as forwarded request which we ignore here.
+					if isForwarded {
+						return nil, consts.ErrChildExecutionNotFound
+					}
+					// The results are from the child that this parent initiated. We should stop tracking the child and process the result now.
+					delete(childrenInitializedAfterResetPoint, childID)
+					mutableState.SetChildrenInitializedPostResetPoint(childrenInitializedAfterResetPoint)
```
yycptt, line 178: plz update `ms.approximateSize` when setting the reset point.

PR #7297, opened 2025-02-07
Support metric prefix in OTEL

yycptt:
## What changed?

- Support metric prefix in OTEL
- NOTE: this is a breaking change for OTEL users who current have the `Prefix` field set but also expect NOT to see the prefix in the emitted metric name.

## Why?

- It's needed for backward compatibility for existing tally users as now we always otel implementation under the hood.
- It was missed when doing OTEL implementation initially. 

## How did you test it?

- Unit test
- Tested locally

### DIFF
--- a/common/metrics/config.go
+++ b/common/metrics/config.go
@@ -61,6 +61,11 @@ type (
 		// Each value in values list will white-list tag values to be reported as usual.
 		ExcludeTags map[string][]string `yaml:"excludeTags"`
 		// Prefix sets the prefix to all outgoing metrics
+		// When migrating from tally to opentelemetry and to be backward compatible with the existing metric names,
+		// if the prefix has a "_" suffix, add an additional "_" at the end.
+		// i.e. "temporal" -> "temporal", but "temporal_" -> "temporal__", "temporal__" -> "temporal___".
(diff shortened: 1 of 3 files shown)


bergundy (commented):
Would be great if the configured prefix would have the same behavior independent on the framework.

bergundy:
AFAIU it shouldn't be a breaking change to have the prefix behave similarly.

yycptt:
> Would be great if the configured prefix would have the same behavior independent on the framework.

> AFAIU it shouldn't be a breaking change to have the prefix behave similarly.

We can, but Imao the otel behavior makes more sense and we should standardize to that.

The different behavior is only an issue if someone is updating the config file from "tally" to "otel". So they are either:
1. Wish to keep backward compatibility, then they need to set other fields like "WithoutUnit" "RecordTimeInSeconds", `Prefix` is just one more field they need to change.  However, I don't think the case makes sense from the beginning as they can just continue to use "tally" in the config.
2. Willing to use the default behavior of otel and change their dashboard,  alert rules etc. Then they anyway need to change the metric name in their dashboards/alerts.

bergundy (approved):
What you have is fine, just wanted to avoid confusing users since they probably won't care or know what the default otel behavior is. Nothing blocking on my end.

PR #7299, opened 2025-02-08
Identify duplicated replication task and skip emitting lag metrics for duplicated replication task

xwduan:
## What changed?

Introduce ErrDuplicate at replication code path to distinguish duplicated replication task
## Why?

To better emit replication lag metrics
## How did you test it?

test will be added later.

### DIFF
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -26,6 +26,7 @@ package history
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"sync"
@@ -1542,10 +1543,10 @@ func (h *Handler) ReplicateWorkflowState(
 	}
 
 	err = engine.ReplicateWorkflowState(ctx, request)
-	if err != nil {
-		return nil, err
+	if err == nil || errors.Is(err, replication.ErrDuplicatedReplicationRequest) {
+		return &historyservice.ReplicateWorkflowStateResponse{}, nil
 	}
-	return &historyservice.ReplicateWorkflowStateResponse{}, nil
+	return nil, err
 }
 
 // SyncShardStatus is called by processor to sync history shard information from another cluster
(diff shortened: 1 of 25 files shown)


service/history/replication/executable_activity_state_task.go
```diff
@@ -196,6 +198,10 @@ func (e *ExecutableActivityStateTask) Execute() error {
 }
 
 func (e *ExecutableActivityStateTask) HandleErr(err error) error {
+	if errors.Is(err, consts.ErrDuplicate) {
```
hai719, line 201: Handle ErrDuplicate in task scheduler? same as followings.

service/history/handler.go
```diff
@@ -1509,7 +1510,7 @@ func (h *Handler) ReplicateEventsV2(ctx context.Context, request *historyservice
 	}
 
 	err2 := engine.ReplicateEventsV2(ctx, request)
-	if err2 != nil {
+	if err2 != nil && errors.Is(err2, consts.ErrDuplicate) {
```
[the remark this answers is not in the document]
xwduan: Thanks. Will fix.

service/history/replication/executable_activity_state_task.go
xwduan: Handling ErrDuplicate in task scheduler may reduce the duplicate code, but i think have err handling logic for each task in one place makes more sense. Let me know if you have strong opinion on this.

service/history/replication/executable_task.go
```diff
@@ -298,9 +301,17 @@ func (e *ExecutableTaskImpl) Attempt() int {
...
 
+func (e *ExecutableTaskImpl) MarkTaskDuplicated() {
+	e.isDuplicated = true
+}
+
 func (e *ExecutableTaskImpl) emitFinishMetrics(
 	now time.Time,
 ) {
+	if e.isDuplicated {
+		metrics.ReplicationDuplicatedTaskCount.With(e.MetricsHandler).Record(1)
```
yux0, line 312: is it possible to have the operation tag and ns tag?

hai719 (approved):

PR #7300, opened 2025-02-08
Poller Scaling Decisions

Sushisource:
> [!IMPORTANT]
> Needs  https://github.com/temporalio/api/pull/553 to be merged for CI to pass

See also the SDK side here: https://github.com/temporalio/sdk-core/pull/874

## What changed?
Optionally attach scaling decisions about whether SDK should increase or decrease pollers to task responses

## Why?
Part of the worker management effort to simplify configuration of workers for users.

## How did you test it?
Unit and basic functional tests are added here. More comprehensive benchmarking was done in conjunction with the SDK changes. Overall, there can be as much as 100% performance increases in task throughput compared to a static number of pollers in very bursty scenarios. At worst, there is no difference. 

A sample of some of the benchmarking results:
- 250 workflows running 5 activities each
    - Autoscale ~5s / No Autoscale ~10s
- 1000 workflows 5 activities each
    - Autoscale ~28s / No Autoscale ~44 seconds
- 1000 workflows 5 activities each 4 workers
    - No autoscale - ~20 seconds first load ~25 seconds second load
    - Autoscale - ~14 seconds first load ~25 seconds second load

## Potential risks
This will not be flipped on by default in SDKs until we want to. That said, the obvious risk here is overall significantly increased polling RPCs. That said, the SDK logic respects rate-limiting errors too.

## Documentation
Docs will be coming as we roll out the SDK flag to enable this.

## Is hotfix candidate?
No


common/dynamicconfig/constants.go
```diff
@@ -1254,6 +1254,36 @@ these log lines can be noisy, we want to be able to turn on and sample selective
 		100,
 		`MatchingMaxTaskQueuesInDeployment represents the maximum number of task-queues that can be registered in a single worker deployment version`,
 	)
+	MatchingPollerScalingMinimumBacklog = NewTaskQueueIntSetting(
+		"matching.pollerScalingMinimumBacklog",
+		1,
+		`MatchingPollerScalingMinimumBacklog is the minimum backlog size that must be accumulated before
```
ShahabT, line 1260: I wonder if this should be based on backlog age and not count. count is more relative because a count of 100 tasks can be anything between negligible or very large depending on the scale in which user is operating. Also given that we start with root partition stats only, this becomes a single partition count which is also dependent on the number of partitions. Number of partitions can change any time (manually now, and automatically in the future).

ShahabT: Looking at `pollerScalingSyncMatchWaitTime` it gives another reason to make this threshold also a time threshold, on poller wait time (for scale down), another task wait time (for scale up).

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1086,41 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
+		pd = nil
+	} else if task.source == serverenumspb.TASK_SOURCE_HISTORY &&
+		pollWaitTime >= pm.config.PollerScalingSyncMatchWaitTime() {
+		// Decrease if any poll matched after sitting idle for some configured period
+		pd.PollerDelta = -1
+	} else if stats.TasksAddRate/stats.TasksDispatchRate > pm.config.PollerScalingDispatchUpFraction() {
+		// Increase if we're adding tasks faster than we're dispatching them. This case is particularly useful when
+		// a new burst of traffic arrives. The backlog may stay at or bounce off of zero as tasks are delivered, and
+		// this allows the pollers to start scaling without accumulating a backlog.
```
ShahabT, line 1111: I tried for 5 minutes and could not come with an example that this is possible. i.e. have a zero backlog and stats.TasksAddRate/stats.TasksDispatchRate being >1. (Note that both rates are based on the last 30s)
Am I missing something?

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1086,41 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+		pd.PollerDelta = 1
+	} else if stats.TasksAddRate/stats.TasksDispatchRate < pm.config.PollerScalingDispatchDownFraction() {
+		// Decrease if we're dispatching tasks faster than we're adding them. This case can come up as the converse of
+		// the above, where we have cleared the backlog but are still not hitting the wait-time sync matching case. We
+		// still can will begin to scale down before hitting that case.
```
ShahabT, line 1116: For this I could imagine a scenario but still not convinced it's good to have it. The scenario it is covering is an edge case in which we want to scale pollers down because the backlog just got drained. 
- But why don't we want to wait until the regular condition of `pollWaitTime >= pm.config.PollerScalingSyncMatchWaitTime()` is met? What do we get for reducing the poller count just 1s ealier?
- It seems the equilibrium of poller count is where pollers wait just bellow PollerScalingSyncMatchWaitTime. So we do not want to ask SDK to reduce pollers below this equilibrium because it'll have to increase the poller count again to go back to the equilibrium. In other words, I want to say maybe this behavior will add unnecessary "swinging".
- Without it code will be simpler, easier to reason about, and we'd have one less knob to tune.

common/dynamicconfig/constants.go
Sushisource: Well the downside with this is that you (very likely) start scaling up slower. If we make the value low I can see how that's a potentially good option though. 

I can certainly test it out and see how it behaves. It might help with avoiding overshooting too. Curious what others think.

service/matching/task_queue_partition_manager.go
Sushisource: It's definitely possible because I've seen it happen (and not infrequently). I agree it's not intuitive why it happens, though. Perhaps something is not exactly accurate in the way the rates are calculated, but it seems like it happens when a bunch of tasks are added after a period of inactivity.

service/matching/task_queue_partition_manager.go
Sushisource: I am totally fine with removing both of these conditions given that they likely don't contribute a whole lot, but I haven't tested without them. If they aren't doing much (which is probably likely) I'm happy to get rid of them.

ShahabT: then I'd vote for deleting them in the interest of simplicity.

common/dynamicconfig/constants.go
dnr: I like the symmetry of using time for scaling in both directions. The backlog age is basically "how long a task has waited", similar to "how long a poll has waited". Though measured slightly differently.

service/matching/task_queue_partition_manager.go
dnr: The add/dispatch rates include sync match, so it's possible that a bunch of new tasks all get sync matched to waiting pollers with no backlog. That would exactly match Spencer's description.

service/matching/task_queue_partition_manager.go
dnr: I guess I'm neutral on this.. I'm always for simplicity but ultimately these are all heuristics, if it performs better in practice and isn't expensive/very complicated to do, then there's not much harm in keeping a few extra conditions.

common/dynamicconfig/constants.go
```diff
@@ -1254,6 +1254,24 @@ these log lines can be noisy, we want to be able to turn on and sample selective
...
+		"matching.pollerScalingSyncMatchWaitTime",
+		1*time.Second,
+		`MatchingPollerScalingSyncMatchWaitTime is the duration a sync-matched poller must exceed before
+a decision to scale down the number of pollers will be issued`,
+	)
+	MatchingPollerScalingDecisionsPerSecond = NewTaskQueueFloatSetting(
+		"matching.pollerScalingDecisionsPerSecond",
+		10,
+		`MatchingPollerScalingDecisionsPerSecond is the maximum number of scaling decisions that will be issued per
+second by one partition manager`,
```
dnr, line 1273: One more note: ideally this would be proportional to worker count, or even a rate limit per worker count. For my projects I'm going to have a "approximate per-key rate limiter" that accepts arbitrary keys but uses only fixed memory. We could use that with poller identity to enforce a rate limit per worker.

Or we can actually just count recent pollers from poller history and multiply the rate limit by that (and then make it like 1 or 2 per worker)

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1085,31 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+	pd := &taskqueuepb.PollerScalingDecision{}
+	pollWaitTime := pm.engine.timeSource.Since(pollStartTime)
+	if stats.ApproximateBacklogCount > 0 && stats.ApproximateBacklogAge.AsDuration() > pm.config.PollerScalingBacklogAgeScaleUp() {
+		// Always increase when there is a backlog, even if we're a partition. It's also important to increase for
+		// sticky queues.
+		pd.PollRequestDeltaSuggestion = 1
+	} else if !pm.partition.IsRoot() {
+		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
+		pd = nil
+	} else if task.source == serverenumspb.TASK_SOURCE_HISTORY &&
```
dnr, line 1103: Why do we check source here? Maybe a comment? The existing comment just says "if any poll matched after sitting idle", which doesn't have anything to do with the source of the task

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1085,31 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+	stats *taskqueuepb.TaskQueueStats, task *internalTask, pollStartTime time.Time,
+) *taskqueuepb.PollerScalingDecision {
+	pd := &taskqueuepb.PollerScalingDecision{}
+	pollWaitTime := pm.engine.timeSource.Since(pollStartTime)
+	if stats.ApproximateBacklogCount > 0 && stats.ApproximateBacklogAge.AsDuration() > pm.config.PollerScalingBacklogAgeScaleUp() {
+		// Always increase when there is a backlog, even if we're a partition. It's also important to increase for
+		// sticky queues.
+		pd.PollRequestDeltaSuggestion = 1
+	} else if !pm.partition.IsRoot() {
+		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
```
dnr, line 1101: Is this still true if we're not using add/dispatch rates?

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1085,31 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
 	perType := userData.GetData().GetPerType()[int32(pm.Partition().TaskType())]
 	return perType, userDataChanged, nil
 }
+
+// makePollerScalingDecision makes a decision on whether to scale pollers up or down based on the current state of the
+// task queue and the task about to be returned. Does not modify inputs.
+func (pm *taskQueuePartitionManagerImpl) makePollerScalingDecision(
```
dnr, line 1091: Should we make scaling decisions for forwarded polls? I don't see a clear reason to say yes or no yet...

service/matching/task_queue_partition_manager.go
Sushisource: Ahh, that makes sense then that's definitely what's gone on.

service/matching/task_queue_partition_manager.go
Sushisource: Based on the tests I've done they made next-to-no difference. Of course it's possible they maybe would make a difference for some scenario I didn't try out, but, :man_shrugging: . They are super simple. The config knobs are maybe the most annoying thing about keeping them in. We could keep them in with hardcoded values?

common/dynamicconfig/constants.go
ShahabT: +1 on proportionality to worker count. Especially because root is exclusively making the scale down decisions, even if number of partitions change when workers scale out, the scale down decisions will be rate limited by this value globally. 
We don't want the rate of poller scale down be the same (in terms of pollers/sec) when there is one worker vs when there are 1000 workers, right?

Sushisource: mmmmm... I'm not sure about this. A lot of my testing I did with just one worker, and fast (Rust, Go, Java) workers can perform easily a huge amount of tasks per second when they're simple. (Which isn't uncommon). So, being able to burst up pollers pretty fast for a _single_ worker is fairly desirable.

Now, I'd agree maybe it should be _even faster_ for more workers, but, I wouldn't want to only do 1 per sec for 1 worker.

service/matching/task_queue_partition_manager.go
Sushisource: Well, I wanted to see that it was a sync match and when we discussed it you mentioned this would be the preferred way to do that. Maybe that doesn't really matter though, I suppose any reason we waited is just as valid.

service/matching/task_queue_partition_manager.go
Sushisource: I'm not sure. This is why I had that open question about keeping track of the overall wait time including forwarding, since that would affect that I think.

service/matching/task_queue_partition_manager.go
Sushisource: Maybe not. I will run some tests without it.

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1085,31 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+	stats *taskqueuepb.TaskQueueStats, task *internalTask, pollStartTime time.Time,
+) *taskqueuepb.PollerScalingDecision {
+	pd := &taskqueuepb.PollerScalingDecision{}
```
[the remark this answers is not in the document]
Sushisource: Good point - an artifact of when I had more in there previously

service/matching/task_queue_partition_manager.go
```diff
@@ -326,6 +330,12 @@ func (pm *taskQueuePartitionManagerImpl) PollTask(
 	}
 
 	task, err := dbq.PollTask(ctx, pollMetadata)
+
+	if task != nil {
+		taskQueueInfo := dbq.GetStats()
```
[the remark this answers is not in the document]
Sushisource: Done

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1085,31 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+func (pm *taskQueuePartitionManagerImpl) makePollerScalingDecision(
+	stats *taskqueuepb.TaskQueueStats, task *internalTask, pollStartTime time.Time,
+) *taskqueuepb.PollerScalingDecision {
+	pd := &taskqueuepb.PollerScalingDecision{}
+	pollWaitTime := pm.engine.timeSource.Since(pollStartTime)
+	if stats.ApproximateBacklogCount > 0 && stats.ApproximateBacklogAge.AsDuration() > pm.config.PollerScalingBacklogAgeScaleUp() {
+		// Always increase when there is a backlog, even if we're a partition. It's also important to increase for
+		// sticky queues.
+		pd.PollRequestDeltaSuggestion = 1
+	} else if !pm.partition.IsRoot() {
```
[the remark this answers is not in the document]
Sushisource: Yeah that's strange. I would expect it to be root for a sticky queue.

tests/poller_scaling_test.go
```diff
@@ -0,0 +1,207 @@
...
+	feClient := s.FrontendClient()
+	// This needs to be done in an eventually loop because nexus endpoints don't become available immediately...
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		resp, err := feClient.PollWorkflowTaskQueue(ctx, &workflowservice.PollWorkflowTaskQueueRequest{
+			Namespace: s.Namespace().String(),
+			TaskQueue: &taskqueuepb.TaskQueue{Name: tq, Kind: enumspb.TASK_QUEUE_KIND_NORMAL},
+		})
+		assert.NoError(t, err)
+		assert.NotNil(t, resp.PollerScalingDecision)
+		assert.GreaterOrEqual(t, int32(1), resp.PollerScalingDecision.PollRequestDeltaSuggestion)
```
bergundy, line 115: `assert` calls don't stop execution and you can't use `require` in `EventuallyWithT` blocks. You'll want to ensure your code won't panic if it gets passed an assertion.

tests/poller_scaling_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+	_, err := s.OperatorClient().CreateNexusEndpoint(ctx, &operatorservice.CreateNexusEndpointRequest{
+		Spec: &nexuspb.EndpointSpec{
+			Name: endpointName,
+			Target: &nexuspb.EndpointTarget{
+				Variant: &nexuspb.EndpointTarget_External_{
+					External: &nexuspb.EndpointTarget_External{
+						Url: "http://" + listenAddr,
+					},
+				},
```
bergundy, line 87: Change the target to a worker instead of an external server.

tests/poller_scaling_test.go
Sushisource: Yeah, I wasn't expecting them to, if the eventually passes that's all I need here

service/matching/task_queue_partition_manager.go
Sushisource: I think we might need to keep these for Nexus (or, at least keep the scale-up condition - the existing wait-time can be used for scale-down).

Since Nexus has no backlog, we need some way to decide when to scale up and this is really the only appropriate thing I can think of. Thoughts @bergundy / @dnr / @ShahabT  ?

bergundy: We technically do have a backlog for queries and nexus tasks, just not a persistent one. But I think using the incoming / outgoing rate is probably a pretty good signal.

common/dynamicconfig/constants.go
dnr: See `GetAllPollerInfo` and similar functions. You could add new ones to get just a count

service/matching/task_queue_partition_manager.go
```diff
@@ -1075,3 +1084,34 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
+		return nil
+	} else if task.source == enumsspb.TASK_SOURCE_HISTORY &&
+		pollWaitTime >= pm.config.PollerScalingSyncMatchWaitTime() {
+		// Decrease if any poll matched after sitting idle for some configured period
+		delta = -1
+	}
+
+	return &taskqueuepb.PollerScalingDecision{
```
dnr, line 1114:
```suggestion
	if delta != 0 {
	 ...
```
?

service/matching/task_queue_partition_manager.go
dnr: It 's not a "backlog" that comes out in `stats.ApproximateBacklogCount`, so it needs extra logic somehow (or we can change backlog stats to include in-mem queries + nexus, that would be interesting...)

service/matching/task_queue_partition_manager.go
dnr: Yeah, my general feeling is the more we ignore stuff like the source of a task, the easier it is to understand overall

service/matching/task_queue_partition_manager.go
dnr: Good point. Though I'm not sure we actually need to pass the start time through forwarding: either a poll is forwarded immediately or not at all, so the wait time on the parent should be only slightly less than the wait time on the child

service/matching/task_queue_partition_manager.go
Sushisource: Right, @dnr what I mean is that branch wouldn't be used for nexus, we'd instead rely on the dispatch rate info - unless I'm misunderstanding?

dnr: I'm just pointing out an alternative solution where we do include nexus in the "backlog count" (i.e. make it durable backlog plus in-mem backlog). Right now it might be a little awkward to get that count, though. The new matcher will make that easier. I'm also totally fine with the rate.

Thinking through it a little more: the count is the integral of the rate. The backlog condition is basically: has the count been positive for x time (age > x), or has the count been 0 for y time (poller waited > y)? If the rate is > z then we can "guess" that the count will be positive for a while (this seems safe), or if it's < z then we can guess that the count will soon be 0 (this one seems less safe... but the number of query/nexus in memory is implicitly limited by a timeout, so maybe it's okay?)

bergundy: I think for now the rate is okay but note that we want to also have a way to instruct users to scale out their workers to handle more Nexus traffic and I think the transient backlog size should fit in the current worker scaling model.
(We could use transient backlog size to scale for high query load as well).

common/dynamicconfig/constants.go
Sushisource: Thought a bit more about this - I ran some tests where I simply moved the scale down on wait-time to be before any ratelimiting, which seemingly makes good sense anyway. If all those polls actually all did wait that long, they should probably scale down. It works perfectly fine

Then, for scale up, the current ratelimiting seems to be plenty fast enough. 

I would have to add a new ratelimiter implementation to do the multiplicative scaling and maybe it makes more sense to just wait for what @dnr is likely to add.

service/matching/task_queue_partition_manager.go
Sushisource: We are now, and, additionally I moved the time-based scale down check to be in front of this which I think makes sense.

common/constants.go
```diff
@@ -60,7 +60,7 @@ const (
 	MinLongPollTimeout = time.Second * 2
 	// CriticalLongPollTimeout is a threshold for the context timeout passed into long poll API,
 	// below which a warning will be logged
-	CriticalLongPollTimeout = time.Second * 20
+	CriticalLongPollTimeout = time.Second * 10
```
Sushisource, line 63: Changed this since, it's useful SDK side, to reduce the timeout while under load (so that when load runs out, we scale down faster).

Talked w/ David and it seems like reducing this ought to be fine, but we can undo if any objections.

tests/poller_scaling_test.go
bergundy: If the test panics it will fail without rerunning the eventually block. I don't think that's what you want.

common/dynamicconfig/constants.go
dnr: My trick for multiplicative scaling that I was going to suggest: let the rate limit be 10*1e6 with proportional burst, then each poll, call AllowN(1e6 / numPollers)

tests/poller_scaling_test.go
Sushisource: I'm not following you. This is doing exactly what I expect, it re-runs the block until the asserts pass. Nothing is panicking.

common/dynamicconfig/constants.go
Sushisource: Ah, ok yeah good trick

tests/poller_scaling_test.go
ShahabT: I think what @bergundy is saying is that line 138 will run and panic even when `assert.NotNil(t, resp.PollerScalingDecision)` is false. You may want to say something like:
```suggestion
		assert.NoError(t, err)
		if assert.NotNil(t, resp.PollerScalingDecision) {
		  assert.GreaterOrEqual(t, int32(1), resp.PollerScalingDecision.PollRequestDeltaSuggestion)
```

service/matching/physical_task_queue_manager.go
```diff
@@ -638,3 +646,53 @@ func (c *physicalTaskQueueManagerImpl) ShouldEmitGauges() bool {
...
+	stats := statsFn()
+	if stats.ApproximateBacklogCount > 0 &&
+		stats.ApproximateBacklogAge.AsDuration() > c.partitionMgr.config.PollerScalingBacklogAgeScaleUp() {
+		// Always increase when there is a backlog, even if we're a partition. It's also important to increase for
+		// sticky queues.
+		delta = 1
+	} else if !c.queue.Partition().IsRoot() {
+		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
+		return nil
+	} else if (stats.TasksAddRate / stats.TasksDispatchRate) > 1.2 {
```
ShahabT, line 686: Sounds good, but as discussed before with @bergundy, the plan is to include the in-memory "backlog" of Nexus and Queries in the stats.ApproximateBacklogCount.

ShahabT, line 686: As this is also an increase decision, Should we move it above the `else if !c.queue.Partition().IsRoot()` block? it seems non-root partitions should be able to make this decisions, right?

service/matching/physical_task_queue_manager.go
```diff
@@ -638,3 +646,53 @@ func (c *physicalTaskQueueManagerImpl) ShouldEmitGauges() bool {
...
+		}
+	}
+
+	// Avoid spiking pollers crazy fast by limiting how frequently change decisions are issued. Be more permissive when
+	// there are more recent pollers.
+	numPollers := c.pollerHistory.history.Size()
+	if numPollers == 0 {
+		numPollers = 1
+	}
+	if !c.pollerScalingRateLimiter.AllowN(pollStartTime, 1e6/numPollers) {
```
ShahabT, line 672: wondering is there is a significance to use pollStartTime instead of Now?

common/dynamicconfig/constants.go
ShahabT: "by one partition manager" -> "by one physical queue manager"?

Also, I think we should say something like: "The effective rate limit would be this value multiplied by the number of active pollers".

ShahabT (approved):
Put some comments, approved otherwise.

service/matching/physical_task_queue_manager.go
Sushisource: I think you can kinda go either way on this one. My thought is that some partition might see a spike that really isn't representative. I think sort of either way you pick, things end up averaging out anyway.

service/matching/physical_task_queue_manager.go
Sushisource: There's really not, changed

common/dynamicconfig/constants.go
```diff
@@ -1260,6 +1260,24 @@ these log lines can be noisy, we want to be able to turn on and sample selective
 		100,
 		`MatchingMaxTaskQueuesInDeployment represents the maximum number of task-queues that can be registered in a single worker deployment version`,
 	)
+	MatchingPollerScalingBacklogAgeScaleUp = NewTaskQueueDurationSetting(
+		"matching.pollerScalingMinimumBacklog",
+		200*time.Millisecond,
+		`MatchingPollerScalingBacklogAgeScaleUp is the minimum backlog age that must be accumulated before 
+a decision to scale up the number of pollers will be issued`,
+	)
+	MatchingPollerScalingSyncMatchWaitTime = NewTaskQueueDurationSetting(
```
dnr, line 1269: is it still specific to sync match, or can we shorten/simplify the name?
```suggestion
	MatchingPollerScalingPollWaitTime = NewTaskQueueDurationSetting(
```

service/matching/physical_task_queue_manager_interface.go
```diff
@@ -73,5 +73,8 @@ type (
 		// ShouldEmitGauges determines whether the gauge metrics should be emitted or not for this particular physical
 		// queue based on dynamic configs.
 		ShouldEmitGauges() bool
+		// MakePollerScalingDecision makes a decision on whether to scale pollers up or down based on the current state
+		// of the task queue and the task about to be returned. Does not modify inputs.
```
dnr, line 77: `time.Time` is immutable, no need to say this
```suggestion
		// of the task queue and the task about to be returned.
```

tests/poller_scaling_test.go
```diff
@@ -0,0 +1,237 @@
...
+		_, err = feClient.RespondWorkflowTaskCompleted(ctx, &workflowservice.RespondWorkflowTaskCompletedRequest{
+			Identity:  "test",
+			TaskToken: resp.TaskToken,
+			Commands:  commands,
+		})
+		assert.NoError(t, err)
+	}, 20*time.Second, 200*time.Millisecond)
+
+	// Wait a little time to ensure add rate exceeds dispatch rate & backlog age grows
+	time.Sleep(500 * time.Millisecond) //nolint:forbidigo // can't poll in an eventually, then we'd drain the tasks
```
dnr, line 176: you can use Eventually with describetaskqueue to check backlog age:
https://github.com/dnr/temporal/blob/pri-light/tests/priority_fairness_test.go#L131-L144
does that work here?

dnr (approved):

PR #7304, opened 2025-02-10
Run integration tests again

stephanos:
## What changed?

Run persistence integration test again.

## Why?

They should be running 🙃 Somehow they were omitted when either `gotestsum` or `go test` parse the space-separated list of packages to test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -413,13 +413,13 @@ prepare-coverage-test: $(GOTESTSUM) $(TEST_OUTPUT_ROOT)
 
 unit-test-coverage: prepare-coverage-test
 	@printf $(COLOR) "Run unit tests with coverage..."
-	go run ./cmd/tools/test-runner $(GOTESTSUM) -retries $(FAILED_TEST_RETRIES) --junitfile $(NEW_REPORT) --packages $(UNIT_TEST_DIRS) -- \
+	go run ./cmd/tools/test-runner $(GOTESTSUM) -retries=$(FAILED_TEST_RETRIES) --junitfile=$(NEW_REPORT) --packages="$(UNIT_TEST_DIRS)" -- \
 		$(COMPILED_TEST_ARGS) \
 		-coverprofile=$(NEW_COVER_PROFILE)
 
 integration-test-coverage: prepare-coverage-test
 	@printf $(COLOR) "Run integration tests with coverage..."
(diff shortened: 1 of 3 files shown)


common/persistence/tests/sqlite_test.go
```diff
@@ -1187,7 +1187,7 @@ func TestSQLiteTransactionContextCancellation(t *testing.T) {
 	cancel()
 
 	err = tx.Commit()
-	assert.ErrorAs(t, err, &context.Canceled)
```
stephanos, line 1190: This yields a data race when the assertion library is trying to wrap the error.

alexshtin: How this could possible work at all?

tools/tdbg/dlq_service_test.go
```diff
@@ -309,9 +297,8 @@ func TestDLQCommand_V2(t *testing.T) {
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
```
stephanos, line 312: No data race if you don't run it in parallel :smart:

common/persistence/tests/sqlite_test.go
stephanos: 🤷

Makefile
```diff
@@ -413,13 +413,13 @@ prepare-coverage-test: $(GOTESTSUM) $(TEST_OUTPUT_ROOT)
 
 unit-test-coverage: prepare-coverage-test
 	@printf $(COLOR) "Run unit tests with coverage..."
-	go run ./cmd/tools/test-runner $(GOTESTSUM) -retries $(FAILED_TEST_RETRIES) --junitfile $(NEW_REPORT) --packages $(UNIT_TEST_DIRS) -- \
+	go run ./cmd/tools/test-runner $(GOTESTSUM) -retries $(FAILED_TEST_RETRIES) --junitfile $(NEW_REPORT) --packages="$(UNIT_TEST_DIRS)" -- \
```
alexshtin, line 416: Are you sure you need `=` here?

stephanos: No, but it aligns with the change I'm making in the code coverage PR (where I'm changing all of them to use `=`)

alexshtin (approved):

Makefile
alexshtin: Add them for other flags on this line then.

stephanos: I can't yet, because testrunner expects them to be like this right now.

PR #7311, opened 2025-02-10
Add configuration to enrich Nexus operation metrics

nikki-dag:
## What changed?

Add configuration that controls which additional metric tags are included with Nexus operation metrics

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -342,13 +342,6 @@ func NexusOperationTag(value string) Tag {
 	return &tagImpl{key: nexusOperationTagName, value: value}
 }
 
-func NexusRequestHeaderTag(name, value string) Tag {
-	if len(value) == 0 {
-		value = unknownValue
-	}
-	return &tagImpl{key: name, value: value}
-}
-
 // HttpStatusTag returns a new httpStatusTag.
 func HttpStatusTag(value int) Tag {
 	return &tagImpl{key: httpStatusTagName, value: strconv.Itoa(value)}
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -334,6 +334,18 @@ func NexusEndpointTag(value string) Tag {
 	return &tagImpl{key: nexusEndpointTagName, value: value}
 }
 
+func NexusServiceTag(value string) Tag {
+	return &tagImpl{key: nexusServiceTagName, value: value}
+}
+
+func NexusOperationTag(value string) Tag {
+	return &tagImpl{key: nexusOperationTagName, value: value}
+}
+
+func NexusRequestHeaderTag(name, value string) Tag {
+	return &tagImpl{key: name, value: value}
+}
+
 // HttpStatusTag returns a new httpStatusTag.
 func HttpStatusTag(value int) Tag {
 	return &tagImpl{key: httpStatusTagName, value: strconv.Itoa(value)}
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -343,6 +343,9 @@ func NexusOperationTag(value string) Tag {
 }
 
 func NexusRequestHeaderTag(name, value string) Tag {
+	if len(value) == 0 {
+		value = unknownValue
+	}
 	return &tagImpl{key: name, value: value}
 }
 
--- a/components/nexusoperations/config.go
+++ b/components/nexusoperations/config.go
@@ -149,6 +149,23 @@ var RetryPolicyMaximumInterval = dynamicconfig.NewGlobalDurationSetting(
 	`The maximum backoff interval between every nexus StartOperation or CancelOperation request for a given operation.`,
 )
 
+var MetricTagConfiguration = dynamicconfig.NewNamespaceTypedSetting(
+	"component.nexusoperations.metrics.tags",
+	&NexusMetricTagConfig{
+		IncludeServiceTag:   false,
+		IncludeOperationTag: false,
+		HeaderTagMappings:   []NexusHeaderTagMapping{},
+	},
+	`Controls which metric tags are included with Nexus operation metrics. This configuration supports:
+1. Service name tag - adds the Nexus service name as a metric dimension (IncludeServiceTag)
+2. Operation name tag - adds the Nexus operation name as a metric dimension (IncludeOperationTag)
+3. Header-based tags - maps values from request headers to metric tags (HeaderTagMappings)
+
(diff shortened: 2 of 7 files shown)


common/metrics/tags.go
```diff
@@ -334,6 +334,18 @@ func NexusEndpointTag(value string) Tag {
...
+	return &tagImpl{key: nexusServiceTagName, value: value}
+}
+
+func NexusOperationTag(value string) Tag {
+	return &tagImpl{key: nexusOperationTagName, value: value}
+}
+
+func NexusRequestHeaderTag(name, value string) Tag {
+	return &tagImpl{key: name, value: value}
+}
```
nikki-dag, line 347: The tag key will be based on the dynamic config in this case.

service/frontend/nexus_handler.go
```diff
@@ -355,6 +390,7 @@ func (h *nexusHandler) StartOperation(
 		return nil, err
 	}
 	ctx = oc.augmentContext(ctx, options.Header)
+	oc.enrichNexusOperationMetrics(service, operation, options.Header)
```
nikki-dag, line 393: We call the `enrichNexusOperationMetrics` method before calling interceptors.
When we have external callers, we might want to include data set by the interceptors. (For example, the auth interceptor might set the caller type and identity.) Since we are still in the design phase and don't know how the interceptors will interact, I left it as is.

components/nexusoperations/config.go
```diff
@@ -149,6 +149,23 @@ var RetryPolicyMaximumInterval = dynamicconfig.NewGlobalDurationSetting(
 	`The maximum backoff interval between every nexus StartOperation or CancelOperation request for a given operation.`,
 )
 
+var MetricTagConfiguration = dynamicconfig.NewNamespaceTypedSetting(
```
bergundy, line 152: Make this a global typed setting please so it can be properly cached with `NewGlobalCachedTypedValue` later. This is a limitation of the dynamic config implementation today.

service/frontend/nexus_handler.go
```diff
@@ -517,6 +553,7 @@ func (h *nexusHandler) CancelOperation(ctx context.Context, service, operation,
 	if err != nil {
 		return err
 	}
+	oc.enrichNexusOperationMetrics(service, operation, options.Header)
```
bergundy, line 556: Looks like we're missing a call to `augmentContext` here, could you please add that while you're here?

service/frontend/nexus_handler.go
```diff
@@ -273,6 +275,37 @@ func (c *operationContext) shouldForwardRequest(ctx context.Context, header nexu
 		c.forwardingEnabledForNamespace(c.namespaceName)
 }
 
+// enrichNexusOperationMetrics enhances metrics with additional Nexus operation context based on configuration.
+func (c *operationContext) enrichNexusOperationMetrics(service, operation string, requestHeader map[string]string) {
```
bergundy, line 279:
```suggestion
func (c *operationContext) enrichNexusOperationMetrics(service, operation string, requestHeader nexus.Header) {
```

service/frontend/nexus_handler.go
```diff
@@ -273,6 +275,37 @@ func (c *operationContext) shouldForwardRequest(ctx context.Context, header nexu
...
+	if conf.IncludeServiceTag {
+		tags = append(tags, metrics.NexusServiceTag(service))
+	}
+
+	if conf.IncludeOperationTag {
+		tags = append(tags, metrics.NexusOperationTag(operation))
+	}
+
+	for _, mapping := range conf.HeaderTagMappings {
+		value, ok := requestHeader[mapping.SourceHeader]
```
bergundy, line 296: `Get` is case insensitive.

```suggestion
		value, ok := requestHeader.Get(mapping.SourceHeader)
```

bergundy (commented):
Thanks! Left a few small comments.

common/metrics/tags.go
```diff
@@ -334,6 +334,21 @@ func NexusEndpointTag(value string) Tag {
 	return &tagImpl{key: nexusEndpointTagName, value: value}
 }
 
+func NexusServiceTag(value string) Tag {
+	return &tagImpl{key: nexusServiceTagName, value: value}
+}
+
+func NexusOperationTag(value string) Tag {
```
pdoerner, line 341: I think a few other places we are just using the existing `metrics.OperationTag`, e.g. in `nexus_handler.getOperationContext`. Should we switch all such instances to this new tag?

nikki-dag: I might be wrong, but in the [nexus_handler.getOperationContext](https://github.com/temporalio/temporal/blob/e0ef18fb17cf750aba31ddd471fb4001784cd313/service/frontend/nexus_handler.go#L317) context, we are referring to the internal `method` that takes `StartOperation` or `CancelOperation` as values. Here, I am referring to the nexus service and operation names defined by the users. I am open to suggestions if you believe we should clarify these further.

pdoerner (approved):

components/nexusoperations/config.go
```diff
@@ -149,6 +149,23 @@ var RetryPolicyMaximumInterval = dynamicconfig.NewGlobalDurationSetting(
 	`The maximum backoff interval between every nexus StartOperation or CancelOperation request for a given operation.`,
 )
 
+var MetricTagConfiguration = dynamicconfig.NewGlobalTypedSetting(
+	"component.nexusoperations.metrics.tags",
+	&NexusMetricTagConfig{
+		IncludeServiceTag:   false,
+		IncludeOperationTag: false,
+		HeaderTagMappings:   []NexusHeaderTagMapping{},
+	},
```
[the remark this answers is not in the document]
nikki-dag: Thanks for catching this!

service/frontend/service.go
```diff
@@ -356,6 +357,16 @@ func NewConfig(
 				return util.WildCardStringsToRegexp(patterns)
 			},
 		),
+		NexusOperationsMetricTagConfig: dynamicconfig.NewGlobalCachedTypedValue(
+			dc,
+			nexusoperations.MetricTagConfiguration,
+			func(config *nexusoperations.NexusMetricTagConfig) (*nexusoperations.NexusMetricTagConfig, error) {
```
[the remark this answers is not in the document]
nikki-dag: I included it following this comment. https://github.com/temporalio/temporal/pull/7311#discussion_r1950982894
So the cache is only for the conversion?

PR #7312, opened 2025-02-10
Use metrics.ClientConfig.Prefix as otel namespace

cludden:
## What changed?
Propagates `go.temporal.io/sdk/common/metrics.ClientConfig.Prefix` to opentelemetry exporter options in `NewOpenTelemetryProvider`

## Why?
Previously, this was handled by `NewScope`, which is no longer invoked as of https://github.com/temporalio/temporal/pull/7265


yycptt:
Dup of https://github.com/temporalio/temporal/pull/7297

yycptt:
I've also [reverted](https://github.com/temporalio/temporal/pull/7303)  https://github.com/temporalio/temporal/pull/7265

PR #7317, opened 2025-02-11
fix more flakes

Shivs11:
## What changed?

- Attempt to fix the seen flakes for versioning-3.1

## How did you test it?

 - Tested by running 50 times against all databases - run: https://github.com/temporalio/temporal/actions/runs/13268736983


tests/worker_deployment_test.go
```diff
@@ -934,15 +934,13 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_NoRamp() {
...
-
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 	defer cancel()
 	tv := testvars.New(s)
 	currentVars := tv.WithBuildIDNumber(1)
-	s.pollFromDeployment(ctx, currentVars)
-	s.ensureCreateVersion(ctx, currentVars)
+
+	go s.pollFromDeployment(ctx, currentVars)
+	s.ensureCreateVersionInDeployment(currentVars)
```
Shivs11, line 943: making this a separate go-routine firstly decreased the total time for a single run for this test from 60 seconds -> 2 seconds

also changed the call from `ensureCreateVersion` to `ensureCreateVersionInDeployment` since `setCurrentVersion` requires the version to be in the deployment list - having a stricter guarantee before the setCurrent call will resolve the errors such as: `version not found in deployment`

carlydf (approved):

PR #7320, opened 2025-02-12
Keep syncing User Data as long as version changes

ShahabT:
## What changed?

UserDataManger in child partitions constantly fetches data from the parent partition. There is a safely mechanism to prevent busy loop in case the parent returns result immediately instead of only when the data is updated.
The safety mechanism did not work well with Versioning 3 because many updates are involved in a single tests. With this update we relax the safety check to only apply when the user data version remains the same and continue to fetch the data asap as long as each time it gets new data.

Here, also I'm relaxing the "requested task queue user data for version greater than known version" errors to be thrown only for the root partition, as they can normally happen for child partitions.

## Why?

So that back-to-back Versioning 3 APIs do not have to wait for a long time.

## How did you test it?

Existing tests.

## Is hotfix candidate?

Yes.


service/matching/user_data_manager.go
```diff
@@ -575,7 +584,28 @@ func (m *userDataManagerImpl) HandleGetUserDataRequest(
...
+			// due to an edge case in during ownership transfer.
+			// We rely on client retries in this case to let the system eventually self-heal.
+			m.logger.Error("requested task queue user data for version greater than known version",
+				tag.NewInt64("request-known-version", version),
+				tag.UserDataVersion(userData.Version),
+			)
+			return nil, errRequestedVersionTooLarge
+		}
+		if req.WaitNewData && userData.GetVersion() <= version {
 			// long-poll: wait for data to change/appear
```
Shivs11, line 609: was wondering - if you had decided to keep the contents of this function the same and only change the logic of the safety mechanism to now only check when the new and old versions were the same, would the same effect have been applied?

I was thinking if this change is a must here

carlydf: I think that this change reduces client retries, which I imagine is a big source of delays. If, for out-of-date non-root-partitions, the client retry can be replaced with an internal `wait on userDataChanged channel`, and then try the loop again and potentially return success, that is a faster loop than pushing the "retry" back onto the client. Especially because retries are expected for non-root partitions, since they need to wait for the user data change to be propagated to them.

carlydf: (If I understand the change correctly, see my other comment clarifying)

service/matching/user_data_manager.go
```diff
@@ -575,7 +584,28 @@ func (m *userDataManagerImpl) HandleGetUserDataRequest(
...
+			// concurrent user data updates in flight. We do not log an error in that case.
+
+			// This is highly unlikely to happen in the owner/root partition but may happen
+			// due to an edge case in during ownership transfer.
+			// We rely on client retries in this case to let the system eventually self-heal.
+			m.logger.Error("requested task queue user data for version greater than known version",
+				tag.NewInt64("request-known-version", version),
+				tag.UserDataVersion(userData.Version),
+			)
+			return nil, errRequestedVersionTooLarge
```
carlydf, line 606: Previously, when a non-root partition had out-of-date user data, we returned `errRequestedVersionTooLarge`. Now we only return that if the root partition is out of date, so the non-root partition won't be "self-healed" through client retries.

I just want to make sure that the non-root partition does not return success to the caller until it actually has the up-to-date user data.

It looks like on line 608, partitions with out-of-date versions trigger a `wait on userDataChanged channel`, and if they receive that signal, they'll continue through the for-loop and possibly succeed this time -- is that how non-root partitions are now expected to retry and eventually get the correct data and return success?

carlydf (approved):
approved if my understanding is correct that we still do not return _success_ unless the data is up-to-date. (difference is we wait and retry internally instead of pushing the retry to the caller)

service/matching/user_data_manager.go
ShahabT: yes, previously we'd error if the known version was higher than the loaded one, now we keep fetching until we pass the known version (for child partitions).

service/matching/user_data_manager.go
Shivs11: ah this makes sense - saw the other comment too and I see how this affects client retries
should have seen this earlier - sorry!

Shivs11 (approved):

service/matching/user_data_manager.go
```diff
@@ -337,14 +337,23 @@ func (m *userDataManagerImpl) fetchUserData(ctx context.Context) error {
...
 		start := time.Now()
+		m.lock.Lock()
+		previousVersion := m.userData.GetVersion()
+		m.lock.Unlock()
+
 		_ = backoff.ThrottleRetryContext(ctx, op, m.config.GetUserDataRetryPolicy, nil)
+
+		m.lock.Lock()
+		currentVersion := m.userData.GetVersion()
+		m.lock.Unlock()
```
dnr, line 348: Instead of these GetVersion calls, can we just have `op` set a local variable to false at the top of the function, and set it to true when it gets a new version? Then we can just use that value below

dnr (approved):

PR #7325, opened 2025-02-12
Fix Nexus disallow header dynamic config usage

bergundy:
## What changed?

Made `component.nexusoperations.disallowedHeaders` a global setting so it can be efficiently cached and fixed the usage of the default value.

## Why?

The previous behavior was to append to the default value which is wrong.

### DIFF
--- a/components/nexusoperations/config.go
+++ b/components/nexusoperations/config.go
@@ -96,6 +96,7 @@ Uses Go's len() function on header keys and values to determine the total size.`
 )
 
 // defaultDisallowedOperationHeaders - set in the convert function below due to a limitation in the dynamic config framework.
+// TODO: restore after an upgrade to Go 1.24 and merging #7052.
 var defaultDisallowedOperationHeaders = []string{
 	"request-timeout",
 	interceptor.DCRedirectionApiHeaderName,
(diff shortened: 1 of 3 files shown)


components/nexusoperations/config.go
```diff
@@ -96,26 +95,19 @@ ScheduleNexusOperation commands with a "nexus_header" field that exceeds this li
 Uses Go's len() function on header keys and values to determine the total size.`,
 )
 
-var DisallowedOperationHeaders = dynamicconfig.NewNamespaceTypedSettingWithConverter(
+// defaultDisallowedOperationHeaders - set in the convert function below due to a limitation in the dynamic config framework.
```
dnr, line 98: let's add a TODO somewhere to restore this after 7052?

dnr (approved):

components/nexusoperations/config.go
bergundy:
```suggestion
// defaultDisallowedOperationHeaders - set in the convert function below due to a limitation in the dynamic config framework.
// TODO restore after an upgrade to Go 1.24 and merging #7052.
```

PR #6935, opened 2024-12-04
[Reset] Ignore ParentClosePolicy when a parent is closed due to reset

gow:
## What changed?
This PR prevents the parentClosePolicy from triggering when a currently running execution was terminated due to reset.

## Why?
This is needed to implement the "Reconnect" reset policy.

## How did you test it?
Added unit test

## Potential risks
N/A

## Documentation
N/A

## Is hotfix candidate?
No

PR #7326, opened 2025-02-12
[Reset] Skip parent close policy only if the parent is reset and we are using the new reset flow.

gow:
## What changed?
The parent will skip applying parent close policy only when it is reset and if we are taking the new reset flow (i.e `AllowResetWithPendingChildren` config is enabled)

## Why?
This fixes a bug I introduced in https://github.com/temporalio/temporal/pull/6935. I am skipping the parent close policy whenever a parent is reset. This was done under the assumption that the server never allowed a parent to be reset which has a pending child. Unfortunately this assumption is not true and the server allowed resetting of parents with pending children in some cases. And in these cases the parent would apply the given parent close policy to the child. So, this change restores the original behavior.
Additionally we need to further fix this by skipping parent close policies only to children that were started *before* the reset point. Will work on that in a following PR since it needs mutable state change.

## How did you test it?
Manual test

## Potential risks
N/A

## Documentation
N/A

## Is hotfix candidate?
Yes

### DIFF
--- a/service/history/transfer_queue_active_task_executor.go
+++ b/service/history/transfer_queue_active_task_executor.go
@@ -344,9 +344,10 @@ func (t *transferQueueActiveTaskExecutor) processCloseExecution(
 		RunId:      task.GetRunID(),
 	}
 	executionInfo := mutableState.GetExecutionInfo()
+	children := copyChildWorkflowInfos(mutableState.GetPendingChildExecutionInfos())
 	var completionEvent *historypb.HistoryEvent // needed to report close event to parent workflow
 	replyToParentWorkflow := mutableState.HasParentExecution() && executionInfo.NewExecutionRunId == ""
-	if replyToParentWorkflow {
+	if replyToParentWorkflow || len(children) > 0 {
 		// only load close event if needed.
 		completionEvent, err = mutableState.GetCompletionEvent(ctx)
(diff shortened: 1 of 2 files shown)


service/history/transfer_queue_active_task_executor.go
```diff
@@ -403,7 +403,13 @@ func (t *transferQueueActiveTaskExecutor) processCloseExecution(
 	// process parentClosePolicy except when the execution was reset. In case of reset, we need to keep the children around so that we can reconnect to them.
 	// We know an execution was reset when ResetRunId was populated in it.
 	// TODO (Chetan): update this condition as new reset policies are added. For now we keep all children since "Reconnect" is the only policy available.
-	if executionInfo.GetResetRunId() == "" {
+	allowResetWithPendingChildren := t.shardContext.GetConfig().AllowResetWithPendingChildren(namespaceName.String())
+	shouldSkipParentClosePolicy := false
+	if executionInfo.GetResetRunId() != "" && allowResetWithPendingChildren {
```
yycptt, line 408: nit: maybe comment that we will ignore the race condition of dc value update and the reset as reset is rarely used.

yycptt (approved):

yycptt (approved):

PR #7333, opened 2025-02-13
Use modernc.org/sqlite version v1.34.1

prathyushpv:
## What changed?
Use version v1.34.1 of modernc.org/sqlite library.

## Why?
We had fixed an issue with contexts being cancelled during a transaction. But a later change in this library reintroduced this issue.
While we work with the maintainer to fix this, use the previous version.

## Is hotfix candidate?
No


yiminc (approved):

Makefile
```diff
@@ -117,6 +117,7 @@ UNIT_TEST_DIRS := $(filter-out $(FUNCTIONAL_TEST_ROOT)% $(FUNCTIONAL_TEST_XDC_RO
 endif
 
 PINNED_DEPENDENCIES := \
+	modernc.org/sqlite@v1.34.0
```
bergundy, line 120: Please document why this is pinned.

bergundy: Can we also pin to a more recent patch or does it have to be 1.34.0?

prathyushpv: We can use [v1.34.1](https://gitlab.com/cznic/sqlite/-/tags/v1.34.1). v1.34.2 has the problematic commit

prathyushpv: Added comment.

rodrigozhou:
Not sure how big of an issue is, but the package page has this:

> ### Fragile modernc.org/libc dependency
> When you import this package you should use in your go.mod file the exact same version of modernc.org/libc as seen in the go.mod file of this repository.
> 
> See the discussion at https://gitlab.com/cznic/sqlite/-/issues/177 for more details.

In v1.34.1, they use `modernc.org/libc@v1.55.3`: https://gitlab.com/cznic/sqlite/-/blob/v1.34.1/go.mod?ref_type=tags#L10

prathyushpv:
> Not sure how big of an issue is, but the [package docs](https://pkg.go.dev/modernc.org/sqlite@v1.34.1) has this:
> 
> > ### Fragile modernc.org/libc dependency
> > When you import this package you should use in your go.mod file the exact same version of modernc.org/libc as seen in the go.mod file of this repository.
> > See the discussion at https://gitlab.com/cznic/sqlite/-/issues/177 for more details.
> 
> In v1.34.1, they use `modernc.org/libc@v1.55.3`: https://gitlab.com/cznic/sqlite/-/blob/v1.34.1/go.mod?ref_type=tags#L10

Thanks @rodrigozhou for finding that! I have pinned libc to this version as well.

rodrigozhou (approved):

PR #7334, opened 2025-02-13
Retry to get workflow result when current branch changed

hai719:
## What changed?
Retry to get workflow result when current branch changed. 

## Why?
When current branch is changed, if caller is waiting for workflow result, we should not return error. Instead, we should get workflow result from the new branch.

## How did you test it?
unit test.

### DIFF
--- a/service/history/api/getworkflowexecutionhistory/api.go
+++ b/service/history/api/getworkflowexecutionhistory/api.go
@@ -40,6 +40,7 @@ import (
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/failure"
 	"go.temporal.io/server/common/headers"
+	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/persistence/transitionhistory"
 	"go.temporal.io/server/common/persistence/versionhistory"
@@ -101,7 +102,12 @@ func Invoke(
 
 		var branchErr *serviceerrors.CurrentBranchChanged
 		if errors.As(err, &branchErr) && isCloseEventOnly {
-			shardContext.GetLogger().Info("Got CurrentBranchChanged, retry with empty branch token")
+			shardContext.GetLogger().Info("Got CurrentBranchChanged, retry with empty branch token",
+				tag.WorkflowNamespaceID(namespaceUUID.String()),
+				tag.WorkflowID(execution.GetWorkflowId()),
+				tag.WorkflowRunID(execution.GetRunId()),
+				tag.Error(err),
+			)
 			// if we are only querying for close event, and encounter CurrentBranchChanged error, then we retry with empty branch token to get the close event
 			response, err = api.GetOrPollMutableState(
 				ctx,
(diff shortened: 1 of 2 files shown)


service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -94,6 +96,24 @@ func Invoke(
...
+					ExpectedNextEventId: expectedNextEventID,
+					CurrentBranchToken:  nil,
+					VersionHistoryItem:  nil,
+					VersionedTransition: nil,
+				},
+				workflowConsistencyChecker,
+				eventNotifier,
+			)
+		}
 		if err != nil {
```
yux0, line 117: nit: combine those two err != nil check?

tests/xdc/history_replication_signals_and_updates_test.go
```diff
@@ -128,7 +128,8 @@ func TestHistoryReplicationSignalsAndUpdatesTestSuite(t *testing.T) {
 
 func (s *hrsuTestSuite) SetupSuite() {
 	s.dynamicConfigOverrides = map[dynamicconfig.Key]any{
-		dynamicconfig.EnableReplicationStream.Key(): true,
+		dynamicconfig.EnableReplicationStream.Key():           true,
+		dynamicconfig.HistoryLongPollExpirationInterval.Key(): 1 * time.Second,
```
yux0, line 132: do you want to let the poll timeout? 1 second may be too long for test

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -94,6 +96,24 @@ func Invoke(
 			workflowConsistencyChecker,
 			eventNotifier,
 		)
+		if err != nil && isCloseEventOnly {
```
yux0, line 99: can we also check the error type is CurrentBranchChanged

service/history/api/getworkflowexecutionhistory/api.go
hai719: They cannot combine together because if retry, err will be the return error of second call.

tests/xdc/history_replication_signals_and_updates_test.go
hai719: set to 100ms, also add check to ensure there are multiple poll calls.

service/history/api/getworkflowexecutionhistory/api.go
hai719: updated. the check was lost.

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -94,6 +98,26 @@ func Invoke(
 			workflowConsistencyChecker,
 			eventNotifier,
 		)
+
+		var branchErr *serviceerrors.CurrentBranchChanged
+		if errors.As(err, &branchErr) && isCloseEventOnly {
+			shardContext.GetLogger().Info("Got CurrentBranchChanged, retry with empty branch token")
```
yycptt, line 104: nit: include workflow information as well?

yycptt (approved):

service/history/api/getworkflowexecutionhistory/api.go
hai719: added.

PR #7336, opened 2025-02-13
[Scheduled Actions V2] Backfiller component

lina-temporal:
## What changed?
Added the HSM Scheduler's Backfiller component, responsible for buffering manual actions. 

Key differences between workflow scheduler and HSM backfiller logic:
* A Backfiller component is 1:1 for each Backfill request, spawned on request and deleted on completion.
* Backfillers each generate their own unique IDs at spawn, used as part of the generated request IDs for deduplication.
* When the Invoker's buffer is full, the Backfiller will exponentially back off and retry filling, giving the Invoker a chance to catch up.

## How did you test it?
* New tests, including those called out in #7152
* `go test -v`

## Potential risks
* Not in production yet
* Missing subtle behavior from the previous backfill mechanism - although I believe the test cases are well represented.


components/scheduler/backfiller_tasks.go
```diff
@@ -0,0 +1,75 @@
...
+func (b BackfillTask) Deadline() time.Time {
+	return b.deadline
+}
+
+func (BackfillTask) Destination() string {
+	return ""
+}
+
+func (BackfillTask) Validate(_ *persistencespb.StateMachineRef, node *hsm.Node) error {
+	// Backfiller only has a single task/state, so no validation is done here.
```
bergundy, line 61: You'll need a way to validate the task (probably by checking if there's an expected action for the given deadline). Otherwise, a standby cluster won't be able to discard these tasks when they've been executed in the active cluster.

bergundy: You can reach out to @yycptt for guidance in case I can't do a followup for this review.

components/scheduler/backfiller_tasks.go
```diff
@@ -0,0 +1,75 @@
...
+	return []hsm.Task{BackfillTask{deadline: b.NextInvocationTime.AsTime()}}, nil
+}
+
+func (b Backfiller) output() (hsm.TransitionOutput, error) {
+	tasks, err := b.tasks()
+	if err != nil {
+		return hsm.TransitionOutput{}, err
+	}
+	return hsm.TransitionOutput{Tasks: tasks}, nil
+}
```
bergundy, line 75: I was looking for this in `backfiller.go`, got used to seeing task generation next to the state machine transition definitions.

proto/internal/buf.yaml
```diff
@@ -12,6 +12,7 @@ breaking:
   # Uncomment this to temporarily ignore specific files or directories:
   ignore:
     - temporal/server/api/historyservice/v1/
+    - temporal/server/api/schedule/v1/message.proto
```
bergundy, line 15: Put a comment to remove this exception once this PR is merged?
Also we should remove the exception for historyservice.

bergundy (commented):
Overall this looks great. Didn't have a lot of comments here.

components/scheduler/backfiller_executors.go
```diff
@@ -0,0 +1,274 @@
...
+	// Otherwise, update progress and reschedule.
+	return hsm.MachineTransition(node, func(b Backfiller) (hsm.TransitionOutput, error) {
+		return TransitionBackfillProgress.Apply(b, EventBackfillProgress{
+			Node:                   node,
+			backfillProgressResult: result,
+		})
+	})
+}
+
+func (backfillerTaskExecutor) enqueue(schedulerNode *hsm.Node, starts []*schedulespb.BufferedStart) error {
```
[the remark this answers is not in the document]
yycptt: This can be removed now?

components/scheduler/scheduler.go
```diff
@@ -256,6 +256,73 @@ func (s *Scheduler) updateConflictToken() {
...
+	err = hsm.MachineTransition(invokerNode, func(e Invoker) (hsm.TransitionOutput, error) {
+		return TransitionEnqueue.Apply(e, EventEnqueue{
+			Node:           invokerNode,
+			BufferedStarts: starts,
+		})
+	})
+	return err
+}
+
+// RequestImmediate spawns a new Backfiller node to the scheduler tree for a
```
yycptt, line 278:
```suggestion
// RequestBackfill spawns a new Backfiller node to the scheduler tree for a
```

components/scheduler/backfiller_tasks.go
yycptt: Yes. All tasks needs a Validate() function implemented. In CHASM world, we will rely on this to manage tasks even in active cluster.

I think you can put attempt count in the task and validate if the one in the task matches with that in the backfiller state. If the Backfiller completes, the node will be removed, and we don't need to worry about task validation.

components/scheduler/backfiller.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil, fmt.Errorf("invalid backfiller state provided: %v", state)
+}
+
+func (backfillerMachineDefinition) Deserialize(body []byte) (any, error) {
+	state := &schedulespb.BackfillerInternal{}
+	return Backfiller{state}, proto.Unmarshal(body, state)
+}
+
+func (backfillerMachineDefinition) CompareState(a any, b any) (int, error) {
+	panic("TODO: CompareState not yet implemented for Executor")
```
yycptt, line 110: yeah we no longer need this in CHASM. State-based replication will take care of it :)

components/scheduler/backfiller_tasks.go
lina-temporal: I put a TODO to validate task attempt count against state in `Validate` for now, as I'm just about to start moving these over to CHASM :).

components/scheduler/backfiller.go
```diff
@@ -0,0 +1,153 @@
...
+	BufferedStarts []*schedulespb.BufferedStart
+
+	// When the next batch of backfills should attempt buffering.
+	NextInvocationTime time.Time
+
+	// High water mark for when state was last updated.
+	LastProcessedTime time.Time
+
+	// When true, the backfill has completed and the node can be deleted.
+	Complete bool
```
yycptt, line 132: do we need this if Scheduler will delete Backfiller when it's completed?

components/scheduler/backfiller.go
```diff
@@ -0,0 +1,153 @@
...
+
+// Applied when the Backfiller has made partial progress on its request and needs
+// to retry.
+var TransitionBackfillProgress = hsm.NewTransition(
+	[]BackfillerMachineState{BackfillerMachineStateRunning},
+	BackfillerMachineStateRunning,
+	func(b Backfiller, event EventBackfillProgress) (hsm.TransitionOutput, error) {
+		b.LastProcessedTime = timestamppb.New(event.LastProcessedTime)
+		b.NextInvocationTime = timestamppb.New(event.NextInvocationTime)
+		b.Attempt++
```
yycptt, line 150: so Attempt will be high either when backfilling itself is keep failing or we just have a lot of things to backfill?

yycptt (approved):

components/scheduler/backfiller.go
lina-temporal: Probably not, when it's wired up in CHASM I'll clean up state we don't need.

components/scheduler/backfiller.go
lina-temporal: Yes to both cases, yeah. You'd need to look at service metrics to be able to distinguish the two cases.

PR #6631, opened 2024-10-09
Refactor protobuf duration validation

pdoerner:
## What changed?

Refactored and normalized protobuf duration validation.
Added additional validation for durations with mismatched signs.
Removed `timer.validation.ValidateAndCapTimer` and will instead rely on task dropping logic in mutable state for long duration timers.

## Why?

Durations with mismatched signs can be converted to Golang durations but they cannot be serialized to a single string which causes an error when trying to view histories with these durations.

## How did you test it?

Existing tests

## Potential risks

Low risk that this additional validation rejects durations which were previously accepted and processed without issue. It is our assumption that temporal created SDKs are not sending these invalid durations. The error when viewing history has only been reported by one customer using a custom Haskell SDK.

PR #7337, opened 2025-02-13
Add back 250 year max duration cap

pdoerner:
## What changed?

Adding back a cap for durations at 250 years.

## Why?

To prevent serialization errors for very large proto durations.

## How did you test it?

New unit test

### DIFF
--- a/common/primitives/timestamp/duration.go
+++ b/common/primitives/timestamp/duration.go
@@ -31,15 +31,17 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
-// 250 years. Maximum representable golang time.Duration is approximately 290 years (INT64_MAX * time.Nanosecond)
-const maxAllowedDuration = 250 * 365 * 24 * time.Hour
+const (
+	// 250 years. Maximum representable golang time.Duration is approximately 290 years (INT64_MAX * time.Nanosecond)
+	maxAllowedDuration = 250 * 365 * 24 * time.Hour
+	maxNanos           = int32(0) // A year is always a round number of seconds.
+)
 
 var (
 	errNegativeDuration = fmt.Errorf("negative duration")
 	errMismatchedSigns  = fmt.Errorf("duration has seconds and nanos with different signs")
 
 	maxSeconds = maxAllowedDuration.Nanoseconds() / 1e9
-	maxNanos   = int32(maxAllowedDuration.Nanoseconds() % 1e9)
 )
 
 func DurationValue(d *durationpb.Duration) time.Duration {
--- a/common/primitives/timestamp/duration.go
+++ b/common/primitives/timestamp/duration.go
@@ -39,7 +39,7 @@ var (
 	errMismatchedSigns  = fmt.Errorf("duration has seconds and nanos with different signs")
 
 	maxSeconds = maxAllowedDuration.Nanoseconds() / 1e9
-	maxNanos   = int32(maxAllowedDuration.Nanoseconds() - maxSeconds*1e9)
+	maxNanos   = int32(maxAllowedDuration.Nanoseconds() % 1e9)
 )
 
 func DurationValue(d *durationpb.Duration) time.Duration {
--- a/common/primitives/timestamp/duration.go
+++ b/common/primitives/timestamp/duration.go
@@ -82,6 +82,7 @@ func durationMultipleOf(amt int64, mult time.Duration) *durationpb.Duration {
 //  2. the golang representation of the duration is not negative
 //
 // Durations are capped to 250 years to prevent overflow and serialization errors.
+// NB: to cap durations, the proto Seconds and Nanos fields are modified!
 //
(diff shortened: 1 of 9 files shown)


common/primitives/timestamp/duration.go
```diff
@@ -31,9 +31,15 @@ import (
...
 
+// 250 years. Maximum representable golang time.Duration is approximately 290 years (INT64_MAX * time.Nanosecond)
+const maxAllowedDuration = 250 * 365 * 24 * time.Hour
+
 var (
 	errNegativeDuration = fmt.Errorf("negative duration")
 	errMismatchedSigns  = fmt.Errorf("duration has seconds and nanos with different signs")
+
+	maxSeconds = maxAllowedDuration.Nanoseconds() / 1e9
+	maxNanos   = int32(maxAllowedDuration.Nanoseconds() - maxSeconds*1e9)
```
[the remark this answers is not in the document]
pdoerner: Proto durations use `int64` for the Seconds field, and `int32` for the Nanos field, so that's why the types are like that. Updated to use modulo instead of subtraction.

common/primitives/timestamp/duration.go
```diff
@@ -95,5 +102,10 @@ func ValidateProtoDuration(d *durationpb.Duration) error {
 		return errNegativeDuration
 	}
 
+	if d.AsDuration() > maxAllowedDuration {
+		d.Seconds = maxSeconds
+		d.Nanos = maxNanos
```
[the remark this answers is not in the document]
pdoerner: That would work but the linter will complain about copying the mutex in the proto duration. I think setting the fields is fine, that's how we implemented it before https://github.com/temporalio/temporal/pull/6631

common/primitives/timestamp/duration.go
```diff
@@ -72,15 +78,16 @@ func durationMultipleOf(amt int64, mult time.Duration) *durationpb.Duration {
...
 //
+// Durations are capped to 250 years to prevent overflow and serialization errors.
+//
 // nil durations are considered valid because they will be treated as the zero value.
 // durationpb.CheckValid cannot be used directly because it will return an error for
-// very large durations but we are okay with truncating these. durationpb.AsDuration()
-// caps the upper bound for timers at 10,000 years to prevent overflow.
-func ValidateProtoDuration(d *durationpb.Duration) error {
+// very large durations, but we are okay with truncating these.
+func ValidateAndCapProtoDuration(d *durationpb.Duration) error {
```
dnr, line 90: Mutating protos is suspicious, and doing it to messages deeply nested in other ones is even more... this would feel a lot better if it returned a new durationpb.Duration. Is it plausible to do that without tearing up huge amounts of existing code?

common/primitives/timestamp/duration.go
```diff
@@ -31,9 +31,15 @@ import (
...
 
+// 250 years. Maximum representable golang time.Duration is approximately 290 years (INT64_MAX * time.Nanosecond)
+const maxAllowedDuration = 250 * 365 * 24 * time.Hour
+
 var (
 	errNegativeDuration = fmt.Errorf("negative duration")
 	errMismatchedSigns  = fmt.Errorf("duration has seconds and nanos with different signs")
+
+	maxSeconds = maxAllowedDuration.Nanoseconds() / 1e9
+	maxNanos   = int32(maxAllowedDuration.Nanoseconds() % 1e9)
```
dnr, line 42: this is obviously always zero since an hour is a round number of seconds, can we just not make a constant?

common/primitives/timestamp/duration.go
pdoerner: It would definitely be a more invasive change. We would still be mutating other protos with the new duration, so I'm not convinced there would be much benefit.

dnr: Yeah, I looked at 6631 again and I see that it used to do that, so I guess it's okay. Just make sure there's a loud comment saying that this can mutate its argument

common/primitives/timestamp/duration.go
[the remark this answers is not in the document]
rodrigozhou: By the way, can't you just set the nanos field to 0? That's what it is anyway.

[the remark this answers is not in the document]
rodrigozhou: Nvm, I see you just changed.

common/primitives/timestamp/duration.go
```diff
@@ -95,5 +102,10 @@ func ValidateProtoDuration(d *durationpb.Duration) error {
 		return errNegativeDuration
 	}
 
+	if d.AsDuration() > maxAllowedDuration {
+		d.Seconds = maxSeconds
+		d.Nanos = int32(0) // A year is always a round number of seconds.
```
dnr, line 107:
```suggestion
		d.Nanos = 0 // A year is always a round number of seconds.
```

dnr (approved):

PR #7339, opened 2025-02-13
De-dup concurrent requests in ramping, current and delete updates

Shivs11:
## What changed?

- Added a layer of validation for SetCurrent, SetRamping and DeleteVersion update requests *inside* each update handler so that concurrent stale requests don't go through and break state.
- Tests testing the above behaviour.

## Why?

- Right now, it's possible for concurrent updates to bypass the validator and get executed. This is bad since it might leave workflow in a bad state.

## How did you test it?

- Added functional tests testing this behaviour for the three operations in consideration.

## Potential risks

- None, in fact this might make the feature better.

## Is hotfix candidate?

- No


service/worker/workerdeployment/client.go
```diff
@@ -503,13 +504,26 @@ func (d *ClientImpl) SetCurrentVersion(
 		return nil, err
 	}
 
+	updateID := uuid.New()
+	if conflictToken != nil {
+		// When a conflict token is provided, we use the hash of the conflict token, deploymentName and the version. This is done
+		// to ensure the automatic de-duplication of updates with the same UpdateID.
+
+		combinedBytes := make([]byte, 0, len(conflictToken)+len(deploymentName)+len(version))
```
ShahabT, line 512: I'm thinking it's better to just serialize the whole input and take its hash. It'll be safer for later when more inputs are added. and also it covers the whole request including client identity etc. (I think it client identity is different, we should not de-dup it)

service/worker/workerdeployment/workflow.go
```diff
@@ -250,6 +254,16 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
...
 
+	// Validating the state before starting the SetRampingVersion operation. This is required due to the following reason:
+	// The validator accepts/rejects updates based on the state of the deployment workflow. Theoretically, two concurrent update requests (with no conflict tokens)
+	// might be accepted by the validator since the state of the workflow, at that point in time, is valid for the updates to take place. Since this update handler
+	// enforces sequential updates, after the first update completes, the local state of the deployment workflow will change. The second update,
+	// now already accepted by the validator, should now not be allowed to run since the state of the workflow is different.
+	err = d.validateStateBeforeAcceptingRampingUpdate(args)
+	if err != nil {
+		return nil, err
+	}
```
ShahabT, line 265: so if the request is not sending conflict token, none of these two changes will help them. Do you plan to also add a noop check here to return success for repeated calls without conflict token? (doesn't have to be in this PR)

Shivs11: so, if the request is not sending a conflict token, the `validateStateBeforeAcceptingRampingUpdate` would still catch duplicate requests (have a peek at my tests where I show how de-duping is done between requests that come in with conflict token and those without)

In short, duplicate concurrent conflict token-less requests, that pass the validator step, will get caught for checks like (in the case of ramping):

```
	if args.Version == d.State.RoutingConfig.RampingVersion && args.Percentage == d.State.RoutingConfig.RampingVersionPercentage {
		d.logger.Info("version already ramping, no change")
		return temporal.NewApplicationError("version already ramping, no change", errNoChangeType)
	}
	if args.Version == d.State.RoutingConfig.CurrentVersion {
		return temporal.NewApplicationError("version can't be set to ramping since it is already current", errVersionAlreadyCurrentType)
	}
	if _, ok := d.State.Versions[args.Version]; !ok && args.Version != "" && args.Version != worker_versioning.UnversionedVersionId {
		d.logger.Info("version not found in deployment")
		return temporal.NewApplicationError("version not found in deployment", errVersionNotFound)
	}

````

all of these checks in the update handler will ensure no stale update request ever runs.

File: service/worker/workerdeployment/client.go (hunk 1 above)
Comment (Shivs11) [reply service/worker/workerdeployment/client.go:512]: I'm not sure about including client identity as part of the hash - If two exact requests are coming in from different call sites (hence different identities), we want them to be considered as duped right?

De-dup imo should be based solely on the inputs sent by clients and if the system has already accepted the same set of inputs, why let it do work again to achieve the same state? For requests that do get duped, we are also not returning errors and the right response objects so that calls from a different call site not face any odd errors

File: service/worker/workerdeployment/workflow.go (hunk 2 above)
Comment (ShahabT) [reply service/worker/workerdeployment/workflow.go:265]: But what about conflict token? we should also check conflict token match here again (if it's given).

Comment (ShahabT) [reply service/worker/workerdeployment/workflow.go:265]: oh ok it is also checking the conflict token 👍

File: service/worker/workerdeployment/workflow.go [hunk 3]
```diff
@@ -220,14 +220,15 @@ func (d *WorkflowRunner) handleDeleteDeployment(ctx workflow.Context) error {
 	return nil
 }
 
-func (d *WorkflowRunner) validateSetRampingVersion(args *deploymentspb.SetRampingVersionArgs) error {
-	if args.ConflictToken != nil && !bytes.Equal(args.ConflictToken, d.State.ConflictToken) {
-		return temporal.NewApplicationError("conflict token mismatch", errConflictTokenMismatchType)
-	}
-	if args.Version == d.State.RoutingConfig.RampingVersion && args.Percentage == d.State.RoutingConfig.RampingVersionPercentage {
+func (d *WorkflowRunner) validateStateBeforeAcceptingRampingUpdate(args *deploymentspb.SetRampingVersionArgs) error {
+	if args.Version == d.State.RoutingConfig.RampingVersion && args.Percentage == d.State.RoutingConfig.RampingVersionPercentage && args.Identity == d.State.LastModifierIdentity {
```
Shivs11, line 224: based on the discussion with @antlai-temporal, a request can be classified as a no-op only if the identity and the args are the same.

ShahabT: what conflict token client would send to the caller? it seems we need to embed the current conflict token in the noChange error?

Shivs11: oh yes, sorry - I haven't included the change to return the latest conflict token when there is `noChange`  - on it

ShahabT (approved):

PR #7342, opened 2025-02-13
Send raw history events from history to frontend service

prathyushpv:
## What changed?
Change to send raw history blobs from history service to frontend service. History service returns a new proto message that has a repeated bytes history field.
This response is wire compatible with the original response which has temporal.api.history.v1.History type for this field. This allows history service to not deserialize events from this data blob. This considerably reduces CPU usage.

History service still needs event_id and version decoded from history events. For this we use a new proto message StrippedHistoryEvent which has these two fields only. It takes considerably less CPU to decode events to this struct.

## Why?
We have seen incidents of high history CPU usage when large number of GetWorkflowExecutionHistory calls are made to workflows which has large history. With this change we can reduce the CPU burden on history service during this API call.

## How did you test it?
Existing unit and functional tests.


proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -982,6 +982,13 @@ message GetWorkflowExecutionHistoryRequest {
 
 message GetWorkflowExecutionHistoryResponse {
     temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse response = 1;
+    temporal.api.history.v1.History history = 2;
+}
+
+// This message must be wire compatible with GetWorkflowExecutionHistoryResponse.
+message GetWorkflowExecutionHistoryResponseWithRaw {
+    temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse response = 1;
+    repeated bytes history = 2;
```
prathyushpv, line 991: This is the change from the previous PR. History service will send raw history bytes to this field if config `history.sendRawHistoryBetweenInternalService` is set to true. Frontend service can check this field and do the modifications like process search attributes, fix the last failed events, filter close events etc and move this history to the correct temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse.History field.

This way it is safe during upgrades and downgrades. We can enable `temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse` after upgrade has finished. And disable it before downgrading.

common/dynamicconfig/constants.go
```diff
@@ -2376,6 +2376,11 @@ that task will be sent to DLQ.`,
 		0.90,
 		"History service health check on persistence error ratio",
 	)
+	SendRawHistoryBetweenInternalServices = NewGlobalBoolSetting(
+		"history.sendRawHistoryBetweenInternalServices",
+		true,
```
prathyushpv, line 2381: Setting this to true to verify all functional tests and integration tests passes for this PR. I will set it back to false before merging.

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -300,6 +301,53 @@ func Invoke(
...
+	// workflowservice.GetWorkflowExecutionHistoryResponse.
+	if sendRawHistoryBetweenInternalServices && !config.SendRawWorkflowHistory(request.Request.GetNamespace()) {
+		rawHistory = make([][]byte, 0, len(historyBlob))
+		for _, blob := range historyBlob {
+			rawHistory = append(rawHistory, blob.Data)
+		}
+		// If there are no events in the history, frontend will not be able to deserialize the response to History object.
+		// In that case, create an empty history object and set it in the response.
+		if len(rawHistory) == 0 {
+			blob, err := history.Marshal()
```
yycptt, line 325: so the `history` must be empty here? shall we make that explicit?

yycptt (approved):

service/history/api/getworkflowexecutionhistory/api.go
prathyushpv: If we return nil for history, frontend will just have a nil pointer in History field. This is causing nil-pointer exception in SDK. That is why we are putting an empty History object here. History is anyways empty here.

service/matching/matching_engine.go
```diff
@@ -775,6 +783,26 @@ func (e *matchingEngineImpl) getHistoryForQueryTask(
 		return nil, nil, err
 	}
 
+	// History service can send history events in response.History.Events. In that case use that directly.
+	// This happens when history.sendRawHistoryBetweenInternalServices is enabled.
+	if resp.History != nil {
```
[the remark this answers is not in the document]
prathyushpv: Done

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -300,6 +301,53 @@ func Invoke(
...
+			return nil, err
+		}
+	}
+
+	var rawHistory [][]byte
+	// if sendRawHistoryBetweenInternalServices is true and SendRawWorkflowHistory is not enabled for this namespace,
+	// send history in raw format in History field of historyservice.GetWorkflowExecutionHistoryResponseWithRaw.
+	// If SendRawWorkflowHistory is enabled for this namespace, raw history will be appended to RawHistory field in
+	// workflowservice.GetWorkflowExecutionHistoryResponse.
+	if sendRawHistoryBetweenInternalServices && !config.SendRawWorkflowHistory(request.Request.GetNamespace()) {
```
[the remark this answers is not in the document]
prathyushpv: Done

service/history/api/getworkflowexecutionhistory/api.go
yycptt: > shall we make that explicit?

I meant Marshal an empty history. As during review it took me some time to confirm `history` will always be empty at that point.

>This is not relevant

Ah, cool. That's even better!

PR #7343, opened 2025-02-14
set status to DRAINING immediately after stop accepting new workflows

carlydf:
## What changed?
Set version to `DRAINING` before starting child instead of relying on child's signal


service/worker/workerdeployment/version_workflow.go
```diff
@@ -96,14 +96,17 @@ func (d *VersionWorkflowRunner) listenToSignals(ctx workflow.Context) {
...
 		if d.VersionState.GetDrainageInfo().GetStatus() != newInfo.Status {
-			d.VersionState.DrainageInfo.Status = newInfo.Status
-			d.VersionState.DrainageInfo.LastChangedTime = newInfo.LastCheckedTime
+			mergedInfo.Status = newInfo.Status
+			mergedInfo.LastChangedTime = newInfo.LastCheckedTime
+			d.VersionState.DrainageInfo = mergedInfo
 			d.syncSummary(ctx)
+		} else {
+			mergedInfo.Status = d.VersionState.DrainageInfo.Status
+			mergedInfo.LastChangedTime = d.VersionState.DrainageInfo.LastChangedTime
```
ShahabT, line 108: Should change the child wf to not send a singal at the beginning now? it's awkward for LastCheckedTime to change a few milliseconds.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -233,6 +236,14 @@ func (d *VersionWorkflowRunner) handleUpdateVersionMetadata(ctx workflow.Context
 }
 
 func (d *VersionWorkflowRunner) startDrainage(ctx workflow.Context, isCan bool) {
+	if d.VersionState.GetDrainageInfo().GetStatus() == enumspb.VERSION_DRAINAGE_STATUS_UNSPECIFIED {
+		now := timestamppb.New(workflow.Now(ctx))
+		d.VersionState.DrainageInfo = &deploymentpb.VersionDrainageInfo{
```
ShahabT, line 241: hopefully no logic depends on `VersionState.DrainageInfo` otherwise we have to patch.

carlydf: I remember making sure that everything did `VersionState.GetDrainageInfo().GetStatus()` so it /should/ just say unspecified if it's nil due to the gap between child starting and signal arriving

service/worker/workerdeployment/version_workflow.go
carlydf: done

service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -62,17 +61,6 @@ func DrainageWorkflow(
 
 	// Set status = DRAINING and then sleep for visibilityGracePeriod (to let recently-started workflows arrive in visibility)
 	if !args.IsCan { // skip if resuming after the parent continued-as-new
-		parentWf := workflow.GetInfo(ctx).ParentWorkflowExecution
```
ShahabT, line 65: you need a wf patch here

service/worker/workerdeployment/version_workflow.go
ShahabT: I mean workflow patch for NDE

service/worker/workerdeployment/drainage_workflow.go
carlydf: done

service/worker/workerdeployment/version_workflow.go
carlydf: I added a patch because I'm adding a signal via syncSummary

service/worker/workerdeployment/drainage_workflow.go
```diff
@@ -62,16 +62,22 @@ func DrainageWorkflow(
...
-		err := workflow.SignalExternalWorkflow(ctx, parentWf.ID, parentWf.RunID, SyncDrainageSignalName, drainingInfo).Get(ctx, nil)
-		if err != nil {
-			return err
+		// When workflow.GetVersion() is run for the new Workflow Execution, it records a marker in the
+		// Workflow history so that all future calls to GetVersion for this change Id—Step 1 in the example—
+		// on this Workflow Execution will always return the given version number, which is 1 in the example.
+		v := workflow.GetVersion(ctx, "Step1", workflow.DefaultVersion, 1)
+		if v == workflow.DefaultVersion { // needs patching because we removed a Signal call
+			parentWf := workflow.GetInfo(ctx).ParentWorkflowExecution
+			now := timestamppb.Now()
```
dnr, line 71: is this indirectly using time.Now()? this needs to use workflow.Now instead (I think even the old version here should be changed)

service/worker/workerdeployment/version_workflow.go
```diff
@@ -96,14 +96,17 @@ func (d *VersionWorkflowRunner) listenToSignals(ctx workflow.Context) {
 	selector.AddReceive(drainageStatusSignalChannel, func(c workflow.ReceiveChannel, more bool) {
 		var newInfo *deploymentpb.VersionDrainageInfo
 		c.Receive(ctx, &newInfo)
-		if d.VersionState.GetDrainageInfo() == nil {
-			d.VersionState.DrainageInfo = &deploymentpb.VersionDrainageInfo{}
-		}
-		d.VersionState.DrainageInfo.LastCheckedTime = newInfo.LastCheckedTime
+		mergedInfo := &deploymentpb.VersionDrainageInfo{}
```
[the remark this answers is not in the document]
carlydf: I wanted to remove any possibility of the drainage info being empty, which I think this change does. I've verified that this replays on the old history, so I feel confident about it

service/worker/workerdeployment/drainage_workflow.go
carlydf: great call out! fixed in previous PR

service/worker/workerdeployment/version_workflow.go
```diff
@@ -235,6 +238,18 @@ func (d *VersionWorkflowRunner) handleUpdateVersionMetadata(ctx workflow.Context
...
+	v := workflow.GetVersion(ctx, "Step1", workflow.DefaultVersion, 1)
+	if v != workflow.DefaultVersion { // needs patching because we added a Signal call via d.syncSummary
+		if d.VersionState.GetDrainageInfo().GetStatus() == enumspb.VERSION_DRAINAGE_STATUS_UNSPECIFIED {
+			now := timestamppb.New(workflow.Now(ctx))
+			d.VersionState.DrainageInfo = &deploymentpb.VersionDrainageInfo{
+				Status:          enumspb.VERSION_DRAINAGE_STATUS_DRAINING,
+				LastChangedTime: now,
+				LastCheckedTime: now,
+			}
+			d.syncSummary(ctx)
```
ShahabT, line 250: should we set the drainage status in the Deployment wf too (in VersionSummary) when a version stops being ramping/current? In case some one is using that as checking drainage status. Does not have to go to this PR.

ShahabT (approved):

service/worker/workerdeployment/version_workflow.go
carlydf: done

PR #7344, opened 2025-02-14
Making poller history a dynamic config

Shivs11:
## What changed?

- PollerHistoryTTL is now a dynamic config

## Why?

- Versioning-3.1 tests required setting a low value for this config in order for some tests to work - right now, those tests were excluded from CI runs and that's not good.

## How did you test it?

- Unskipped the tests that were being skipped previously.

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -452,6 +452,12 @@ to exceed this number will fail with a FailedPrecondition error.`,
 		`MatchingDeletedRuleRetentionTime is the length of time that deleted Version Assignment Rules and
 Deleted Redirect Rules will be kept in the DB (with DeleteTimestamp). After this time, the tombstones are deleted at the next time update of versioning data for the task queue.`,
 	)
+	PollerHistoryTTL = NewNamespaceDurationSetting(
+		"matching.wv.PollerHistoryTTL",
+		5*time.Minute,
+		`PollerHistoryTTL is the time to live for poller histories in the pollerHistory cache of a physical task queue. Poller histories are fetched when
+		requiring a list of pollers that polled a given task queue.`,
+	)
 	ReachabilityBuildIdVisibilityGracePeriod = NewNamespaceDurationSetting(
 		"matching.wv.ReachabilityBuildIdVisibilityGracePeriod",
 		3*time.Minute,
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -453,7 +453,7 @@ to exceed this number will fail with a FailedPrecondition error.`,
 Deleted Redirect Rules will be kept in the DB (with DeleteTimestamp). After this time, the tombstones are deleted at the next time update of versioning data for the task queue.`,
 	)
 	PollerHistoryTTL = NewNamespaceDurationSetting(
-		"matching.wv.PollerHistoryTTL",
+		"matching.PollerHistoryTTL",
 		5*time.Minute,
 		`PollerHistoryTTL is the time to live for poller histories in the pollerHistory cache of a physical task queue. Poller histories are fetched when
 		requiring a list of pollers that polled a given task queue.`,
--- a/service/matching/physical_task_queue_manager.go
+++ b/service/matching/physical_task_queue_manager.go
@@ -144,7 +144,8 @@ func newPhysicalTaskQueueManager(
 		tasksAddedInIntervals:      newTaskTracker(clock.NewRealTimeSource()),
 		tasksDispatchedInIntervals: newTaskTracker(clock.NewRealTimeSource()),
 	}
-	pqMgr.pollerHistory = newPollerHistory(e.config.PollerHistoryTTL(partitionMgr.Namespace().Name().String()))
+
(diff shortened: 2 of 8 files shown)


tests/worker_deployment_test.go
```diff
@@ -1116,13 +1116,16 @@ func (s *WorkerDeploymentSuite) TestDeleteWorkerDeployment_ValidDelete() {
...
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		a := assert.New(t)
+		listResp, err := s.FrontendClient().ListWorkerDeployments(ctx, &workflowservice.ListWorkerDeploymentsRequest{
+			Namespace: s.Namespace().String(),
+		})
+		a.Nil(err)
+		for _, dInfo := range listResp.GetWorkerDeployments() {
+			a.NotEqual(tv1.DeploymentSeries(), dInfo.GetName())
+		}
+	}, time.Second*5, time.Millisecond*200)
```
Shivs11, line 1128: this has to go in an eventually call since there might be some delays in visibility - running this locally on a cassandra DB was failing too (without the eventually block)

tests/worker_deployment_test.go
```diff
@@ -1031,7 +1031,7 @@ func (s *WorkerDeploymentSuite) verifyTaskQueueVersioningInfo(ctx context.Contex
 }
 
 func (s *WorkerDeploymentSuite) TestDeleteWorkerDeployment_ValidDelete() {
-	s.T().Skip("skipping this test for now until I make TTL of pollerHistoryTTL configurable by dynamic config.")
+	s.OverrideDynamicConfig(dynamicconfig.PollerHistoryTTL, 5*time.Second)
```
ShahabT, line 1034: 5 sec is still quite high for a test. maybe 500ms or so?

ShahabT (approved):

common/dynamicconfig/constants.go
```diff
@@ -452,6 +452,12 @@ to exceed this number will fail with a FailedPrecondition error.`,
 		`MatchingDeletedRuleRetentionTime is the length of time that deleted Version Assignment Rules and
 Deleted Redirect Rules will be kept in the DB (with DeleteTimestamp). After this time, the tombstones are deleted at the next time update of versioning data for the task queue.`,
 	)
+	PollerHistoryTTL = NewNamespaceDurationSetting(
+		"matching.wv.PollerHistoryTTL",
```
dnr, line 456: this doesn't have anything to do with versioning
```suggestion
		"matching.PollerHistoryTTL",
```

service/matching/physical_task_queue_manager.go
```diff
@@ -144,7 +144,7 @@ func newPhysicalTaskQueueManager(
 		tasksAddedInIntervals:      newTaskTracker(clock.NewRealTimeSource()),
 		tasksDispatchedInIntervals: newTaskTracker(clock.NewRealTimeSource()),
 	}
-	pqMgr.pollerHistory = newPollerHistory()
+	pqMgr.pollerHistory = newPollerHistory(e.config.PollerHistoryTTL(queue.NamespaceId()))
```
dnr, line 147: This needs a namespace name, not an id. Unfortunately we don't use the specific types pervasively so the type system doesn't catch this.

Shivs11: thanks for this catch, missed this - fixed in my latest push

service/matching/physical_task_queue_manager.go
```diff
@@ -144,7 +144,7 @@ func newPhysicalTaskQueueManager(
 		tasksAddedInIntervals:      newTaskTracker(clock.NewRealTimeSource()),
 		tasksDispatchedInIntervals: newTaskTracker(clock.NewRealTimeSource()),
 	}
-	pqMgr.pollerHistory = newPollerHistory()
+	pqMgr.pollerHistory = newPollerHistory(e.config.PollerHistoryTTL(partitionMgr.Namespace().Name().String()))
```
Shivs11, line 147: fetching the namespaceName from the partitionMgr directly - was thinking of doing it using the ns registry but think the value stored here would eventually evaluate to be the same

cc - @dnr

dnr: I think really the cleanest way to do this is to put it in `taskQueueConfig` as well (which is `partitionMgr.config` here). Reaching up to the service-level `Config` shouldn't be necessary.

Shivs11: hmm, the only reason why I was hesitating to put that in there was that it maybe doesn't belong there - I noticed most taskQueueConfigs are those that have a task-queue name and type filter - I can simply place the config there and keep it namespace scoped, sure, but is it okay to place it there if it's *really* not a task-queue level config?

dnr: It's fine. There are some things in there that are constants :)  If it's a setting used by an individual task queue partition, it should probably be there.

Shivs11: cool, making the change - thanks!

dnr: great!

dnr (approved):

PR #7345, opened 2025-02-14
format version string correctly in deployment API error messages

carlydf:


ShahabT (approved):
seems like an OSS release patch candidate.

PR #7347, opened 2025-02-14
deletes not possible on "draining" workflows

Shivs11:
## What changed?

- Deleting a version that is draining should never be possible.

## How did you test it?

- modified a test to now test for this specific behaviour

### DIFF
--- a/service/worker/workerdeployment/version_workflow.go
+++ b/service/worker/workerdeployment/version_workflow.go
@@ -316,8 +316,6 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context, args *
 		return err
 	}
 
-	d.logger.Info("Version is eligible for deletion")
-
 	// sync version removal to task queues
 	syncReq := &deploymentspb.SyncDeploymentVersionUserDataRequest{
 		Version:       state.GetVersion(),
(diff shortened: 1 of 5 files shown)


service/worker/workerdeployment/version_workflow.go
```diff
@@ -299,9 +299,9 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context, args *
 
 	// 2. Check if the version is drained.
```
carlydf, line 300:
```suggestion
	// 2. Check if the version is not draining.
```

tests/worker_deployment_version_test.go
```diff
@@ -513,16 +513,47 @@ func (s *DeploymentVersionSuite) TestDeleteVersion_DeleteRampedVersion() {
...
 }
 
-func (s *DeploymentVersionSuite) TestDeleteVersion_NotDrained() {
+func (s *DeploymentVersionSuite) TestDeleteVersion_DrainingVersion() {
 	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
 	defer cancel()
 	tv1 := testvars.New(s).WithBuildIDNumber(1)
 
 	// Start deployment workflow 1 and wait for the deployment version to exist
 	s.startVersionWorkflow(ctx, tv1)
```
carlydf, line 523: this looks great overall. could you also start another tv2 Deployment Version, try to delete that? It should succeed if it was just created and never current/ramping/draining

carlydf (commented):
request for one more test case, otherwise looks good

tests/worker_deployment_version_test.go
Shivs11: the test case I added creates a new version, tv2, and sets it as current so deleting it won't be possible - I can make a new test case and start a version, not make it current/ramping, and try deleting that

carlydf (approved):

PR #7350, opened 2025-02-15
WIP: Add version column to SQLlite, MySQL, and Postgress VisibilityStores

jmbarzee:
## What changed?
Add version column to SQLlite, MySQL, and Postgress VisibilityStores

## Why?
There is no built in support for versioning in SQLlite, MySQL, and Postgress, which leaves open the possibility for out-of-order updates. We need to explicitly have a column for version to prevent this. 

## How did you test it?
Manually tested that Server can run normally (still some bugs). There will be more tests with nightly pipelines and mid pipeline version changes to test backwards compatibility.

## Potential risks
High. A failure during table alteration or in compatibility guarantees would entirely halt all visibility updates.

## Documentation
Not that I am aware of. We will have to close several github issues though!

## Is hotfix candidate?
Not for cloud. Aiming for upcoming OSS release.


schema/mysql/v8/visibility/schema.sql
```diff
@@ -85,6 +86,7 @@ CREATE INDEX by_temporal_namespace_division   ON executions_visibility (namespac
 CREATE TABLE custom_search_attributes (
   namespace_id      CHAR(64)  NOT NULL,
   run_id            CHAR(64)  NOT NULL,
+  version           BIGINT    NOT NULL DEFAULT 0,
```
jmbarzee, line 89: I'm uncertain we need this. It would hinge on the way a write returns in MySQL. I'm fairly sure that currently if the proposed version is smaller than the written version we won't write, but also won't error. This needs to be addressed.

common/persistence/sql/sqlplugin/visibility.go
```diff
@@ -43,6 +43,7 @@ import (
 
 var (
 	ErrInvalidKeywordListDataType = errors.New("Unexpected data type in keyword list")
+	VersionColumnName = "version"
```
jmbarzee, line 46: Reminder for me to add a lightweight test to verify the field name matches/exists on the row.

rodrigozhou: I wonder if the column name `version` is too generic and cause conflicts later. Should we name something like `_version`?

common/persistence/sql/sqlplugin/mysql/visibility.go
```diff
@@ -82,7 +82,9 @@ var (
 func buildOnDuplicateKeyUpdate(fields ...string) string {
 	items := make([]string, len(fields))
 	for i, field := range fields {
-		items[i] = fmt.Sprintf("%s = VALUES(%s)", field, field)
+		// This line is to ensure that no update occurs (for any column) if the version is behind the saver version.
+		items[i] = fmt.Sprintf("%v = IF(%v < VALUES(%v), VALUES(%v), %v)",
+		 field, sqlplugin.VersionColumnName, sqlplugin.VersionColumnName, field, field)
```
rodrigozhou, line 87: We need to make sure the `version` column is the last to update. Otherwise, the condition will fail after the `version` column is updated.

common/persistence/sql/sqlplugin/visibility.go
```diff
@@ -223,6 +225,25 @@ func getDbFields() []string {
...
+	for i := 0; i < len(dbFields)-1; i++ {
+		if dbFields[i] == VersionColumnName {
+			foundVersion = true
+		}
+		if foundVersion {
+			dbFields[i] = dbFields[i+1]
+		}
+	}
+	end := len(dbFields) - 1
+	dbFields[end] = VersionColumnName
```
rodrigozhou, line 245: Oh, should we do this for MySQL only in the `buildOnDuplicateKeyUpdate` function?

common/persistence/sql/sqlplugin/visibility.go
```diff
@@ -54,6 +55,7 @@ type (
 	VisibilityRow struct {
 		NamespaceID          string
 		RunID                string
+		Version 		     int64
```
rodrigozhou, line 58: Can you maybe separate this from the rest, and add comments explaining this field?
```
VisibilityRow struct {
	// existing fields

	// explain version
	Version int64
}
```

schema/mysql/v8/visibility/schema.sql
```diff
@@ -1,6 +1,7 @@
 CREATE TABLE executions_visibility (
   namespace_id            CHAR(64)      NOT NULL,
   run_id                  CHAR(64)      NOT NULL,
+  version                 BIGINT        NOT NULL DEFAULT 0,
```
rodrigozhou, line 4: ditto: maybe put in a separate block and add comments explaining it

schema/postgresql/v12/visibility/schema.sql
```diff
@@ -10,6 +10,7 @@ $$ LANGUAGE plpgsql IMMUTABLE RETURNS NULL ON NULL INPUT;
 CREATE TABLE executions_visibility (
   namespace_id            CHAR(64)      NOT NULL,
   run_id                  CHAR(64)      NOT NULL,
+  version                 BIGINT        NOT NULL DEFAULT 0,
```
rodrigozhou, line 13: ditto

schema/sqlite/v3/visibility/schema.sql
```diff
@@ -1,6 +1,7 @@
 CREATE TABLE executions_visibility (
   namespace_id            CHAR(64)      NOT NULL,
   run_id                  CHAR(64)      NOT NULL,
+  version                 BIGINT        NOT NULL DEFAULT 0,
```
rodrigozhou, line 4: ditto

schema/mysql/v8/visibility/schema.sql
rodrigozhou: I think you need this.

rodrigozhou (changes requested):
You need to update the version in `schema/mysql/version.go` (same for `postgresql`).

common/persistence/sql/sqlplugin/visibility.go
jmbarzee: So, I'm not deeply familiar with the `Values(...)` and that might be throwing me off here. Do the orders of the columns need to match in every part of the query? Some part of my very old SQL Experience says yes.

common/persistence/sql/sqlplugin/visibility.go
jmbarzee: I choose the position to match the order of the columns in the table (if you just use create statements) 

but, the more I think about this the more I like it. Because Go reflects fields in the order of declaration, we would remove the need for the hacky reordering and have no SQL DB type specific code in the main plugin. I'm going to do that.

common/persistence/sql/sqlplugin/visibility.go
jmbarzee: Yea, as I think about this, something like `_orderingVersion` would be more descriptive and help avoid collisions. What do you think?

common/persistence/sql/sqlplugin/mysql/visibility.go
jmbarzee: Yes. I put that in a different section, but I think we can get away with just ordering the field in the struct. I'll make a test to ensure the order comes back correct, but golang also guarantees field order matches declaration (except for embedding) in reflection.

schema/mysql/v8/visibility/schema.sql
jmbarzee: So, its only an optimization, but generally fields which are read more will be more performant if earlier in column order.

jmbarzee: Considering that we will always read this on write, it seems worth doing. 

On the other hand, it seems like we read most columns, if not all.

schema/mysql/v8/visibility/schema.sql
jmbarzee: Yea, particularly when I consider that I think our query will return with 1 row written no matter what happens with version. 

And now that I think of it, we could have an awful situation were one table is ahead and the other isn't. Yuck.

schema/postgresql/v12/visibility/schema.sql
jmbarzee: re: ditto, ditto

schema/sqlite/v3/visibility/schema.sql
jmbarzee: re: ditto, ditto

common/persistence/sql/sqlplugin/visibility.go
rodrigozhou: We explicitly list the columns when inserting: `INSERT INTO executions_visibility (<columns>) VALUES (<values>)`. The order in `<values>` is the same for `<columns>`.

common/persistence/sql/sqlplugin/visibility.go
rodrigozhou: I suggested `_version` because it's the name Elasticsearch uses. I also like `_row_version`.

schema/mysql/v8/visibility/schema.sql
rodrigozhou: Hm, interesting. If that's the case, our order is not optimized at all. It should've been `namespace_id`, `workflow_id`, `run_id`, `version`, ...
Anyway, I don't think we're trying to optimize this much, but sure, not a big deal for me. But please, add some comments for the `version` column.

common/persistence/sql/sqlplugin/postgresql/visibility.go
```diff
@@ -45,10 +45,12 @@ var (
 	templateUpsertWorkflowExecution = fmt.Sprintf(
 		`INSERT INTO executions_visibility (%s)
 		VALUES (%s)
-		%s`,
+		%s
+		WHERE executions_visibility.%s < EXCLUDED.%s`, // This line is to ensure that no update occurs if the version is behind the saved version.
```
rodrigozhou, line 49: The `WHERE` part is part of the `ON CONFLICT` expression. Please put this together in the `buildOnDuplicateKeyUpdate` function.

common/persistence/sql/sqlplugin/sqlite/visibility.go
```diff
@@ -49,10 +49,12 @@ var (
 	templateUpsertWorkflowExecution = fmt.Sprintf(
 		`INSERT INTO executions_visibility (%s)
 		VALUES (%s)
-		%s`,
+		%s
+		WHERE executions_visibility.%s < EXCLUDED.%s`, // This line is to ensure that no update occurs if the version is behind the saved version.
```
rodrigozhou, line 53: ditto

schema/mysql/v8/visibility/versioned/v1.9/add_version_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE executions_visibility ADD COLUMN _version BIGINT NOT NULL DEFAULT 0;
+ALTER TABLE custom_search_attributes ADD COLUMN _version BIGINT NOT NULL DEFAULT 0;
```
rodrigozhou, line 2: The "⛔" is because it doesn't have a final line break ("blank line" after line 2. It's kind of a standard for the last line of a file to be a blank line.

common/persistence/sql/sqlplugin/visibility.go
```diff
@@ -72,6 +73,11 @@ type (
 		ParentRunID          *string
 		RootWorkflowID       string
 		RootRunID            string
+
+		// Version must be at the end because the version column has to be the last column in the insert statement.
+		// Otherwise we may do partial updates as the version changes halfway through.
+		// This is because MySQL doesn't support row versioning in a way that prevents out-of-order updates.
+		Version int64 `db:"_version"`
```
rodrigozhou, line 80: Please add a unit test to verify this is always the last field.

rodrigozhou (changes requested):

common/persistence/sql/sqlplugin/mysql/visibility.go
```diff
@@ -61,8 +61,9 @@ var (
 	templateUpsertCustomSearchAttributes = `
 		INSERT INTO custom_search_attributes (
 			namespace_id, run_id, search_attributes
-		) VALUES (:namespace_id, :run_id, :search_attributes)
-		ON DUPLICATE KEY UPDATE search_attributes = VALUES(search_attributes)`
+		) VALUES (:namespace_id, :run_id, :search_attributes)` +
+		fmt.Sprintf(`ON DUPLICATE KEY UPDATE search_attributes = IF(%v < VALUES(%v), VALUES(search_attributes), search_attributes)`,
+			sqlplugin.VersionColumnName, sqlplugin.VersionColumnName)
```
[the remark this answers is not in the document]
jmbarzee: Sorry, I'm having a hard time seeing what you are trying to point out here.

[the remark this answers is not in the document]
jmbarzee: Ahhh, The `INSERT ... _version"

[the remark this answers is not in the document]
rodrigozhou: Yes, it's missing the `_version` column.

common/persistence/sql/sqlplugin/postgresql/visibility.go
jmbarzee: Ok, I tossed it here only because it references the table, but if you feel like the function should represent the full expression, I'm fine with that.

common/persistence/sql/sqlplugin/visibility.go
jmbarzee: Yep, Haven't resolved my reminder yet. Just fixing all the broken tests first.

schema/mysql/v8/visibility/versioned/v1.9/add_version_column.sql
jmbarzee: Thanks, I've never noticed that symbol.

common/persistence/sql/sqlplugin/postgresql/visibility.go
rodrigozhou: Do you need to reference the table?

common/persistence/sql/sqlplugin/visibility.go
jmbarzee: Done

rodrigozhou (approved):

schema/mysql/v8/visibility/schema.sql
jmbarzee: Added comments

common/persistence/sql/sqlplugin/postgresql/visibility.go
jmbarzee: Yes. We got an error back from postgress saying  `column reference \"_version\" is ambiguous`

PR #7356, opened 2025-02-19
[Scheduled Actions] Limit the length of the ScheduleInfo Spec fields in the memo block (ListSchedules)

lina-temporal:
## What changed?
- The Spec fields on `ScheduleInfo` have an upper bound applied to their length with the LimitMemoSpecSize version applied.

## Why?
- Large ScheduleInfo blocks in the memo lead to unnecessarily large database transactions. Customers can still use `DescribeSchedule` to get the full spec.

## How did you test it?
- New functional test.
- `go test -tags test_dep -v ./tests/ -run ScheduleFunctionalSuite/TestLimitMemoSpecSize && make lint`

## Potential risks
- We break a customer who'd been relying on an exhaustive ScheduleInfo block in ListSchedules responses upon rollout.


service/worker/scheduler/workflow.go
```diff
@@ -116,6 +118,8 @@ const (
 	nextTimeCacheV1Size = 10
 
 	impossibleHistorySize = 1e6 // just for testing, no real history can be this long
+
+	SpecFieldLengthLimit = 10 // 10 items per spec field on the ScheduleInfo memo
```
dnr, line 122: could go in tweakables? but it doesn't matter, I doubt there's any need to change it

dnr (approved):

service/worker/scheduler/workflow.go
lina-temporal: Moved it to tweakables.

PR #7360, opened 2025-02-19
Consolidate task queue background contexts

dnr:
## What changed?
- Consolidate all task queue contexts into one, at physicalTaskQueueManager.
- Most objects under physicalTaskQueueManager have a reference to the context and use it for background operations.
- Task reader/writer also use the same context to know when to shut down.
- Currently taskQueueDB does not hold the context and requires contexts passed in, but that could happen also as a second step.
- Timeout context creation for taskGC is moved inside, to avoid creating a timer when we're not going to gc.

## Why?
Simplify code, especially startup/shutdown: everything shares one context for lifecycle, there's no more context construction or tagging functions being passed around, everything can just use that context and WithTimeout.

## How did you test it?
updated unit tests, existing functional tests


service/matching/task_reader.go
```diff
@@ -83,30 +79,8 @@ func newTaskReader(backlogMgr *backlogManagerImpl) *taskReader {
 
 // Start taskReader background goroutines.
 func (tr *taskReader) Start() {
-	if !atomic.CompareAndSwapInt32(
-		&tr.status,
-		common.DaemonStatusInitialized,
-		common.DaemonStatusStarted,
-	) {
-		return
-	}
```
stephanos, line 92: I came to the conclusion that this isn't needed since the only place this is started from is the `backlogManagerImpl` which in turn is started by `physicalTaskQueueManagerImpl` - which has this protection already.

service/matching/task_reader.go
```diff
@@ -83,30 +79,8 @@ func newTaskReader(backlogMgr *backlogManagerImpl) *taskReader {
...
-
-	tr.gorogrp.Go(tr.dispatchBufferedTasks)
-	tr.gorogrp.Go(tr.getTasksPump)
-}
-
-// Stop taskReader goroutines.
-// Note that this does not wait until
```
stephanos, line 99: Not even the author of this comment was willing to wait until they

service/matching/task_reader.go
```diff
@@ -51,11 +50,9 @@ const (
 
 type (
 	taskReader struct {
-		status     int32
 		taskBuffer chan *persistencespb.AllocatedTaskInfo // tasks loaded from persistence
 		notifyC    chan struct{}                          // Used as signal to notify pump of new tasks
 		backlogMgr *backlogManagerImpl
-		gorogrp    goro.Group
```
stephanos, line 58: ❤️

service/matching/task_gc.go
```diff
@@ -33,11 +33,13 @@ import (
 )
 
 type taskGC struct {
+	backlogMgr *backlogManagerImpl
```
stephanos, line 36: Hm, this (and the others) make the cyclical dependency situation worse. Are we going to declare design bankruptcy on these components and merge them all into one place in the near future?

stephanos (approved):

service/matching/task_gc.go
dnr: Oh, funny, on the tip of my working branch I switched it back to taking separate dependencies... I extracted this PR from a few in the middle and forgot about that. I'll change it to use separates now.

I have definitely considered merging them and still think it may make sense. It's partly a question of how much you want to review as "diff" and how much as new code.

Inherently there is always going to be some cycle somewhere since I think backlog manager <--> task reader is irreducible. We can hide it inside closures if we want, but I feel like that can obscure more than it helps.

service/matching/task_reader.go
dnr: We don't need this status cas stuff anywhere and I'm slowly ripping it out. It's protection against calling Start/Stop multiple times and we should just never do that, and treat it as a bug if it somehow happens. For components with the same lifecycle as the process, fx takes care of everything. For things like this, matching engine should manage the lifecycle accurately and never start/stop twice.

service/matching/task_reader.go
dnr: Oh, um, that was me :grimacing:  I must have gotten distracted mid-sentence

dnr (commented):
btw, I don't love the name `tqCtx`, other ideas welcome. I want to distinguish it from a request context, so something like `lifecycleCtx` or `backgroundCtx`?

service/matching/task_reader.go
dnr: I don't _hate_ goro.Group/goro.Handle, but I'm souring on them slightly: they package up a few common patterns, but encapsulate things so that they're not composable with themselves or other startup/shutdown patterns.

stephanos: I haven't used it myself, but using the Go primitives instead where possible seems like a great change.

service/matching/task_gc.go
stephanos: Sounds good!

I'd support merging them as an immediate step since they are too entangled right now anyway. Then, it might be possible to chip off parts again one-by-one to reduce the size where reasonable/possible.

I think the major downsides of the cycles are (1) testing (interfaces help, though) and (2) reasoning about it. A reset might help identify the parts that can actually be safely extracted. Only where needed/valuable ofc.

stephanos:
I don't mind `tqCtx`. `lifecycleCtx` would work for me, too. `tqLifecycleCtx` seems even clearer and isn't too long. `backgroundCtx` seems too easy to confuse with `context.Background()`, tough.

PR #7361, opened 2025-02-19
Refactoring: RegisterWorker via WorkerDeployment 

Shivs11:
## What changed?

- RegisterWorker will now happen from within WorkerDeployment workflow.

## Why?

- Without this change, a successful `DescribeVersion` call would not mean that the version was registered in the deployment workflow. Users, and current tests in the SDK, have to further do a `DescribeWorkerDeployment` and verify if the version was present in the deployment before doing operations like `SetCurrent/SetRamping`. This was not desirable and this PR fixes that.

## How did you test it?

- Passing existing suite of tests work. Tested SDK tests locally to verify too.


service/matching/physical_task_queue_manager.go
```diff
@@ -578,7 +580,7 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeploymentVersion(
 		ctx, namespaceEntry, workerDeployment.SeriesName, workerDeployment.BuildId, c.queue.TaskQueueFamily().Name(), c.queue.TaskType(),
 		"matching service", uuid.New())
 	if err != nil {
-		var errTooMany deployment.ErrMaxTaskQueuesInDeployment
+		var errTooMany workerdeployment.ErrMaxTaskQueuesInDeployment
```
Shivs11, line 583: this seems to be an error and is something I directly worked on :( 
Sorry! Caught it now and verified things are working by running the test `TestDeploymentVersionLimits` and having a peek at it's output.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -136,7 +136,7 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
 
 	if err := workflow.SetUpdateHandlerWithOptions(
 		ctx,
-		RegisterWorkerInDeployment,
+		RegisterWorkerInDeploymentVersion, // TODO (Shivam): do we need patching for this?
```
Shivs11, line 139: was not totally sure on this - since this change won't go in the cloud-release, and the change seems to be backwards-incompatible, my guess is that I have to use patching here - happy to be proven wrong

service/worker/workerdeployment/workflow.go
```diff
@@ -384,10 +425,14 @@ func (d *WorkflowRunner) handleDeleteVersion(ctx workflow.Context, args *deploym
 		d.lock.Unlock()
 	}()
 
+	return d.deleteVersion(ctx, args)
+}
+
+func (d *WorkflowRunner) deleteVersion(ctx workflow.Context, args *deploymentspb.DeleteVersionArgs) error {
```
Shivs11, line 431: made a helper function because when an individual delete version request were to come in, we need to acquire the lock vs `tryDelete` requests coming while registering a worker cannot acquire a lock since the workflow, at that point in time, is already holding on to it

Shivs11: PS - Since consolidating into a helper won't necessarily change the execution path of old currently-running worker-deployment workflows, this change won't require patching.

service/worker/workerdeployment/client.go
```diff
@@ -1381,7 +1442,7 @@ func (d *ClientImpl) isTaskQueueExpectedInNewVersion(
 			Versions: &taskqueuepb.TaskQueueVersionSelection{
 				BuildIds: []string{prevCurrentVersionInfo.GetVersion()}, // pretending the version string is a build id
 			},
-			TaskQueueType: taskQueue.Type, // since request doesn't pass through frontend, this field is not automatically populated
+			TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW, // since request doesn't pass through frontend, this field is not automatically populated
```
Shivs11, line 1445: this might be an error on my end - if for example, the missing task-queue was an activity type, sending an activity type in this field will result in the Describe call returning an error : `Describe cannot be called on an activity type ` - in this example, to fetch activity task-queue's stats, you pass the type in the types field (which I am doing)

PR #7363, opened 2025-02-19
Fix callbacks state machine id

rodrigozhou:
## What changed?

Make `OnConflictOptions.AttachRequestId` required if `OnConflictOptions.AttachCompletionCallbacks` is set.
Change the callback state machine ID builder to use the request ID instead of the history event ID.

## Why?

Attaching a callback to a running workflow creates the `WorkflowExecutionOptionsUpdated` event, which can be buffered. In this case, the history event ID does not "exist" yet, and cannot be used to generate the callback state machine ID.

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -2391,7 +2391,11 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 	ms.executionInfo.DefaultWorkflowTaskTimeout = event.GetWorkflowTaskTimeout()
 	ms.executionInfo.OriginalExecutionRunId = event.GetOriginalExecutionRunId()
 
-	if err := ms.addCompletionCallbacks(startEvent, event.GetCompletionCallbacks()); err != nil {
+	if err := ms.addCompletionCallbacks(
+		startEvent,
+		requestID,
+		event.GetCompletionCallbacks(),
+	); err != nil {
 		return err
 	}
 	if err := ms.UpdateWorkflowStateStatus(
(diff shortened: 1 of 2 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -2567,11 +2572,18 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+		// This is for backwards compatibility: callbacks was initially only attached when the workflow
+		// execution started, but now it can be attached while it's running.
+		if event.GetEventType() == enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED {
+			// Use the start event version and ID as part of the callback ID to ensure that callbacks have unique
+			// IDs that are deterministically created across clusters.
+			// TODO: Replicate the state machine state and allocate a uuid there instead of relying on history event
+			// replication.
+			id = fmt.Sprintf("%d-%d-%d", event.GetVersion(), event.GetEventId(), idx)
+		} else {
+			id = fmt.Sprintf("%s-%d", requestID, idx)
```
bergundy, line 2585: Put a prefix here so we can ensure we avoid collisions in case the request ID is an unexpected value like `0-1`.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2567,11 +2572,18 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
-		// replication.
-		id := fmt.Sprintf("%d-%d-%d", event.GetVersion(), event.GetEventId(), idx)
+		id := ""
+		// This is for backwards compatibility: callbacks was initially only attached when the workflow
+		// execution started, but now it can be attached while it's running.
+		if event.GetEventType() == enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED {
+			// Use the start event version and ID as part of the callback ID to ensure that callbacks have unique
+			// IDs that are deterministically created across clusters.
+			// TODO: Replicate the state machine state and allocate a uuid there instead of relying on history event
+			// replication.
```
bergundy, line 2582: You can remove the TODO here, this will be made obsolete with CHASM and state based replication.

bergundy (approved):

yycptt (approved):
Not sure about the requirement on testing in this case. I will leave that to the nexus crew to decide.
Logic LGTM!

PR #7368, opened 2025-02-20
make sure child completion reports to correct branch

hai719:
## What changed?
Check the child's FirstExecutionRunId when RecordChildExecutionCompleted.

## Why?
When there is conflict resolution and parent's branch changes, there may be a new child run2. Old child run1 and new run2 have same initiated event id, but different started events on different branches. If the old child run1 reports execution completion to parent, parent will incorrectly complete new run2.

## How did you test it?
unit test.


service/history/api/recordchildworkflowcompleted/api.go
```diff
@@ -153,6 +153,11 @@ func recordChildWorkflowCompleted(
 				return nil, consts.ErrChildExecutionNotFound
 			}
 
+			if ci.GetStartedRunId() != request.GetChildFirstExecutionRunId() {
```
yycptt, line 156: hmmm then we need to change the child runID recorded in `recordStartedEventIfMissing` and also [here](https://github.com/temporalio/temporal/blob/8398b630efe602bc131c227503d1a85622910c67/service/history/transfer_queue_active_task_executor.go#L871) when reattaching to a child workflow upon reset to be firstExecutionRunId.

hai719: thanks for pointing out! updated.

yycptt (approved):

PR #7369, opened 2025-02-20
put drainage workflow in TemporalWorkerDeployment namespace division to hide it from default wf list

carlydf:

### DIFF
--- a/service/worker/workerdeployment/version_workflow.go
+++ b/service/worker/workerdeployment/version_workflow.go
@@ -26,8 +26,6 @@ package workerdeployment
 
 import (
 	"fmt"
-	"time"
-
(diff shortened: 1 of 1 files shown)


service/worker/workerdeployment/version_workflow.go
```diff
@@ -243,6 +245,12 @@ func (d *VersionWorkflowRunner) startDrainage(ctx workflow.Context, isCan bool)
 	d.drainageWorkflowFuture = &fut
```
dnr, line 245: Not this PR but I just happened to see it: the future is an interface type, you should just use it directly, not use a pointer to it

dnr (approved):

PR #7375, opened 2025-02-21
Update stringer version to 0.30

ychebotarev:


dnr (approved):
no objections, though I'm surprised you're having trouble with the current version and this fixes it. I don't see any changes that would affect things here

PR #7378, opened 2025-02-24
Replace cache with map in namespace registry

ychebotarev:
## What changed?

Change "cache" with "map" in nsRegistry.

## Why?

There is no cache in registry per se. There is no adding one entry at a time, or removing one entry at a time.
Practically it is just a map. This makes code cleaner. 

## How did you test it?

Existing test.

## Potential risks

Can't think of anything realistic.

### DIFF
--- a/common/namespace/nsregistry/registry.go
+++ b/common/namespace/nsregistry/registry.go
@@ -236,8 +236,10 @@ func (r *registry) getAllNamespace() []*namespace.Namespace {
 
 func (r *registry) getAllNamespaceLocked() []*namespace.Namespace {
 	result := make([]*namespace.Namespace, len(r.idToNamespace))
+	i := 0
 	for _, value := range r.idToNamespace {
-		result = append(result, value)
+		result[i] = value
+		i++
 	}
 	return result
 }
--- a/common/namespace/nsregistry/registry.go
+++ b/common/namespace/nsregistry/registry.go
@@ -109,7 +109,7 @@ type (
 		logger                  log.Logger
 		refreshInterval         dynamicconfig.DurationPropertyFn
 
-		// registryLock protects nameToID, idToNamespace and stateChangeCallbacks.
+		// nsMapsLock protects nameToID, idToNamespace and stateChangeCallbacks.
 
 		nsMapsLock    sync.RWMutex
 		nameToID      map[namespace.Name]namespace.ID
--- a/common/namespace/nsregistry/registry.go
+++ b/common/namespace/nsregistry/registry.go
@@ -49,7 +49,6 @@ import (
 
 const (
 	cacheMaxSize = 64 * 1024
-	cacheTTL     = 0 // 0 means infinity
 	// CacheRefreshFailureRetryInterval is the wait time
 	// if refreshment encounters error
 	CacheRefreshFailureRetryInterval = 1 * time.Second
@@ -66,9 +65,6 @@ const (
 )
 
 var (
-	cacheOpts = cache.Options{
-		TTL: cacheTTL,
-	}
 	readthroughNotFoundCacheOpts = cache.Options{
 		TTL: readthroughCacheTTL,
 	}
@@ -114,9 +110,11 @@ type (
 		refreshInterval         dynamicconfig.DurationPropertyFn
 
 		// cacheLock protects cachNameToID, cacheByID and stateChangeCallbacks.
-		cacheLock                     sync.RWMutex
-		cacheNameToID                 cache.Cache
-		cacheByID                     cache.Cache
+
+		registryLock  sync.RWMutex
+		nameToID      map[namespace.Name]namespace.ID
+		idToNamespace map[namespace.ID]*namespace.Namespace
+
 		stateChangeCallbacks          map[any]namespace.StateChangeCallbackFn
 		stateChangedDuringReadthrough []*namespace.Namespace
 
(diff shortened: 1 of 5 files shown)


common/namespace/nsregistry/registry.go
```diff
@@ -49,7 +49,6 @@ import (
 
 const (
 	cacheMaxSize = 64 * 1024
```
dnr, line 51: rename this `readthroughCacheSize`?

common/namespace/nsregistry/registry.go
```diff
@@ -114,9 +110,11 @@ type (
 		refreshInterval         dynamicconfig.DurationPropertyFn
 
 		// cacheLock protects cachNameToID, cacheByID and stateChangeCallbacks.
```
dnr, line 112: update comment

common/namespace/nsregistry/registry.go
```diff
@@ -220,52 +218,37 @@ func (r *registry) GetPingChecks() []pingable.Check {
...
-		}
-		result[id] = aResult
+func (r *registry) getAllNamespaceLocked() []*namespace.Namespace {
+	result := make([]*namespace.Namespace, len(r.idToNamespace))
+	i := 0
+	for _, value := range r.idToNamespace {
+		result[i] = value
+		i++
 	}
 	return result
```
dnr, line 244:
```suggestion
	return expmaps.Values(r.idToNamespace)
```
where `expmaps` is `import expmaps "golang.org/x/exp/maps"`

common/namespace/nsregistry/registry.go
```diff
@@ -414,22 +397,23 @@ func (r *registry) refreshNamespaces(ctx context.Context) error {
 	}
 
 	// Make a copy of the existing namespace cache (excluding deleted), so we can calculate diff and do "compare and swap".
```
dnr, line 399:
```suggestion
	// Make a copy of the existing namespace maps (excluding deleted), so we can calculate diff and atomically swap.
```
it looks like we don't do compare-and-swap, we just assume only one thread is ever updating

common/namespace/nsregistry/registry.go
```diff
@@ -114,9 +110,11 @@ type (
 		refreshInterval         dynamicconfig.DurationPropertyFn
 
 		// cacheLock protects cachNameToID, cacheByID and stateChangeCallbacks.
-		cacheLock                     sync.RWMutex
-		cacheNameToID                 cache.Cache
-		cacheByID                     cache.Cache
+
+		registryLock  sync.RWMutex
+		nameToID      map[namespace.Name]namespace.ID
+		idToNamespace map[namespace.ID]*namespace.Namespace
```
alexshtin, line 116: It was clear before that these 3 fields play together. While I agree, `cache` was a bad word for grouping them, not having it at all is not good either. What's about:
```suggestion
		mapsLock  sync.RWMutex
		nameToIDMap      map[namespace.Name]namespace.ID
		idToNamespaceMap map[namespace.ID]*namespace.Namespace
```
?

common/namespace/nsregistry/registry.go
```diff
@@ -166,9 +164,9 @@ func NewRegistry(
 // GetCacheSize observes the size of the by-name and by-ID caches in number of
 // entries.
 func (r *registry) GetCacheSize() (sizeOfCacheByName int64, sizeOfCacheByID int64) {
```
alexshtin, line 166: I believe you need to search through this package and replace all instances of word `Cache` with something else.

common/namespace/nsregistry/registry.go
alexshtin: Ideally, it would be great to create another struct which combines these 3 fields.

common/namespace/nsregistry/registry.go
ychebotarev: done

common/namespace/nsregistry/registry.go
ychebotarev: done

common/namespace/nsregistry/registry.go
ychebotarev: done

common/namespace/nsregistry/registry.go
ychebotarev: you added "map" to those fields.
*  xxToYy already imply by convention its a map
* adding type to a field name is generally considered a bad practice in modern programming

That said - yeah, registryLock is a bit generic, I will rename it to nsMapsLock

common/namespace/nsregistry/registry.go
ychebotarev: yeah, now some names are weird. Renamed, a lot

dnr (approved):
Please don't force-push once a PR is already in review.. it breaks the "Show changes since your last review" functionality in github

service/history/handler.go
```diff
@@ -693,7 +693,7 @@ func (h *Handler) DescribeHistoryHost(_ context.Context, req *historyservice.Des
 		}
 	}
 
-	itemsInCacheByIDCount, itemsInCacheByNameCount := h.namespaceRegistry.GetCacheSize()
+	itemsInCacheByIDCount, itemsInCacheByNameCount := h.namespaceRegistry.GetRegistrySize()
```
alexshtin, line 696: Well, this needs to be renamed too. As well as `namespacespb.NamespaceCacheInfo` proto fields. I am ok to leave it as is though.

alexshtin (approved):

PR #7380, opened 2025-02-24
Register task-queue workers through worker deployment 

Shivs11:
## What changed?

- RegisterWorker will now happen from within WorkerDeployment workflow.
- Also added a script which saves new workflow histories for both the deployment entity workflows and shall be useful for detecting NDE's if someone makes a backwards incompatible change in the near future.

## Why?

- Without this change, a successful DescribeVersion call would not mean that the version was registered in the deployment workflow. Users, and current tests in the SDK, have to further do a DescribeWorkerDeployment and verify if the version was present in the deployment before doing operations like SetCurrent/SetRamping. This was not desirable and this PR fixes that

## How did you test it?

- Passing existing suite of tests work. Tested SDK tests locally to verify too.


service/matching/physical_task_queue_manager.go
```diff
@@ -580,7 +581,7 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeploymentVersion(
 		ctx, namespaceEntry, workerDeployment.SeriesName, workerDeployment.BuildId, c.queue.TaskQueueFamily().Name(), c.queue.TaskType(),
 		"matching service", uuid.New())
 	if err != nil {
-		var errTooMany deployment.ErrMaxTaskQueuesInDeployment
+		var errTooMany workerdeployment.ErrMaxTaskQueuesInDeployment
```
Shivs11, line 584: this seems to be an error and is something I directly worked on :(
Sorry! Caught it now and verified things are working by running the test TestDeploymentVersionLimits and having a peek at it's output.

service/worker/workerdeployment/client.go
```diff
@@ -1381,7 +1445,9 @@ func (d *ClientImpl) isTaskQueueExpectedInNewVersion(
 			Versions: &taskqueuepb.TaskQueueVersionSelection{
 				BuildIds: []string{prevCurrentVersionInfo.GetVersion()}, // pretending the version string is a build id
 			},
-			TaskQueueType: taskQueue.Type, // since request doesn't pass through frontend, this field is not automatically populated
+			// Since request doesn't pass through frontend, this field is not automatically populated.
+			// Moreover, DescribeTaskQueueEnhanced requires this field to be set to WORKFLOW type.
+			TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW,
```
Shivs11, line 1450: this might be an error on my end - if for example, the missing task-queue was an activity type, sending an activity type in this field will result in the Describe call returning an error : Describe cannot be called on an activity type  - in this example, to fetch activity task-queue's stats, you pass the type in the types field (which I am doing

service/worker/workerdeployment/workflow.go
```diff
@@ -407,6 +490,27 @@ func (d *WorkflowRunner) validateDeleteVersion(args *deploymentspb.DeleteVersion
 	return d.validateStateBeforeAcceptingDeleteVersion(args)
 }
 
+func (d *WorkflowRunner) deleteVersion(ctx workflow.Context, args *deploymentspb.DeleteVersionArgs) error {
```
Shivs11, line 493: consolidated the main logic of deleting a version into a helper because when an individual delete version request were to come in, we need to acquire the lock vs tryDelete requests spurring up when registering a worker cannot acquire a lock since the workflow, at that point in time, is already holding on to it

PS - Since consolidating into a helper won't necessarily change the execution path of old currently-running worker-deployment workflows, this change won't require patching.

service/matching/physical_task_queue_manager.go
```diff
@@ -53,6 +53,7 @@ import (
 	"go.temporal.io/server/common/testing/testhooks"
 	"go.temporal.io/server/common/worker_versioning"
 	"go.temporal.io/server/service/worker/deployment"
```
carlydf, line 55: The fact that this is still here makes me concerned that we're accidentally using the `deployment` package somewhere else? Could you double check that wherever this file uses this package is an appropriate use?

service/worker/workerdeployment/client.go
```diff
@@ -163,6 +163,7 @@ type Client interface {
 	) error
 
 	// Used internally by the Worker Deployment Version workflow in its AddVersionToWorkerDeployment Activity
```
carlydf, line 165: is this comment still true?

service/worker/workerdeployment/workflow.go
```diff
@@ -130,6 +131,14 @@ func (d *WorkflowRunner) run(ctx workflow.Context) error {
 		return err
 	}
 
+	if err := workflow.SetUpdateHandler(
```
carlydf, line 134: I'm not sure if this `SetUpdateHandler` call needs to be in a patch too. Like, does `SetUpdateHandler` create a wf event?

carlydf (commented):
Could you add a replay unit test to this PR? Here are two examples of the test:

https://github.com/temporalio/temporal/blob/beca2ea7664299c1c28cfc7d92f50e859f2e17ad/service/worker/workerdeployment/replay_test.go#L43

https://github.com/temporalio/temporal/blob/0836b0fff0bbd6637ea1723f7e5a675519029354/service/worker/scheduler/replay_test.go

The scheduler test includes a script for generating new history data, which is "best practice", but I'd be happy with a manually-created-and-downloaded history (you can download a workflow history from the UI) which is what I did for my test.

go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
 	go.temporal.io/api v1.45.0
-	go.temporal.io/sdk v1.32.1
+	go.temporal.io/sdk v1.33.0
```
Shivs11, line 61: I guess the sdk version had to be updated so that the `./generate_history` script could work as it requires a poller using the latest `WorkerDeploymentOptions` - Hoping that's okay?

carlydf: yeah the main thing is that the server depends on a real SDK version (not a random commit hash). updating it is good!

service/worker/workerdeployment/replaytester/replay_test.go
```diff
@@ -49,23 +51,25 @@ func TestReplays(t *testing.T) {
...
+	replayer.RegisterWorkflowWithOptions(workerdeployment.VersionWorkflow, workflow.RegisterOptions{Name: workerdeployment.WorkerDeploymentVersionWorkflowType})
+	replayer.RegisterWorkflowWithOptions(deploymentWorkflow, workflow.RegisterOptions{Name: workerdeployment.WorkerDeploymentWorkflowType})
+	replayer.RegisterWorkflowWithOptions(drainageWorkflow, workflow.RegisterOptions{Name: workerdeployment.WorkerDeploymentDrainageWorkflowType})
 
 	files, err := filepath.Glob("testdata/replay_*.json.gz")
 	require.NoError(t, err)
 
 	logger := log.NewSdkLogger(log.NewTestLogger())
 
+	fmt.Println("Length of files", len(files))
```
carlydf, line 71: do you want to leave this in?

carlydf (approved):
Thanks for adding the replay tests!

One important thing: Make sure you use an old server with a specific version (I think either the latest OSS tag used in cloud, or the latest OSS version would be the right one to use) to generate the histories. Then, include that version in the file name of the history, so we know what version of the server generated that history. The test with the current code should pass on histories generated by a server running the old code.

service/worker/workerdeployment/replaytester/generate_history.sh
```diff
@@ -0,0 +1,25 @@
...
+# Consider running this script to generate a new history for TestReplays
+# whenever there's some change to the worker-deployment or worker-deployment-version workflow.
+# To use it, run a local server (any backend) with the dynamic config for worker-versioning 3
+# (system.enableDeploymentVersions=true) enabled and run this script.
+#
+# Note: this requires temporal cli >= 0.12 and sdk >= v1.33.0
+
+deploymentName="foo"
+version="1.0"
+
```
carlydf, line 12: make sure that when you run this script, the server you used to generate the worker deployment wf was the old server version (without the changes in this branch)

carlydf: [v1.27.1](https://github.com/temporalio/temporal/releases/tag/v1.27.1) is the latest OSS version.
[v1.27.0-128.5](https://github.com/temporalio/temporal/releases/tag/v1.27.0-128.5) is the latest OSS tag used in cloud (used in [v3.128.5+oss1.27.0-128.5](https://github.com/temporalio/saas-temporal/releases/tag/v3.128.5%2Boss1.27.0-128.5) of saas-temporal).

service/worker/workerdeployment/replaytester/replay_test.go
Shivs11: removed!

PR #7383, opened 2025-02-24
Fix data_encoding column in current_executions table

rodrigozhou:
## What changed?

Fix `data_encoding` column in `current_executions` table in SQL database.
The column must be `NOT NULL`.

### DIFF
--- a/common/persistence/sql/execution_util.go
+++ b/common/persistence/sql/execution_util.go
@@ -1165,7 +1165,7 @@ func updateExecution(
 func workflowExecutionStateFromCurrentExecutionsRow(
 	row *sqlplugin.CurrentExecutionsRow,
 ) (*persistencespb.WorkflowExecutionState, error) {
-	if len(row.Data) > 0 {
+	if len(row.Data) > 0 && row.DataEncoding != "" {
 		return serialization.WorkflowExecutionStateFromBlob(row.Data, row.DataEncoding)
 	}
 
--- a/schema/mysql/v8/temporal/schema.sql
+++ b/schema/mysql/v8/temporal/schema.sql
@@ -57,7 +57,7 @@ CREATE TABLE current_executions(
   last_write_version BIGINT NOT NULL,
   -- `data` contains the WorkflowExecutionState (same as in `executions.state` above)
   data MEDIUMBLOB NULL,
-  data_encoding VARCHAR(16) NULL,
+  data_encoding VARCHAR(16) NOT NULL DEFAULT '',
   PRIMARY KEY (shard_id, namespace_id, workflow_id)
 );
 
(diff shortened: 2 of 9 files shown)


schema/mysql/v8/temporal/schema.sql
```diff
@@ -57,7 +57,7 @@ CREATE TABLE current_executions(
   last_write_version BIGINT NOT NULL,
   -- `data` contains the WorkflowExecutionState (same as in `executions.state` above)
   data MEDIUMBLOB NULL,
-  data_encoding VARCHAR(16) NULL,
+  data_encoding VARCHAR(16) NOT NULL DEFAULT '',
```
yiminc, line 60: Is this necessary?

common/persistence/sql/execution_util.go
```diff
@@ -1165,7 +1165,7 @@ func updateExecution(
 func workflowExecutionStateFromCurrentExecutionsRow(
 	row *sqlplugin.CurrentExecutionsRow,
 ) (*persistencespb.WorkflowExecutionState, error) {
-	if len(row.Data) > 0 {
+	if len(row.Data) > 0 && row.DataEncoding != "" {
```
yiminc, line 1168: maybe len(raw.DataEncoding) > 0 to handle nil case as well?

schema/mysql/v8/temporal/schema.sql
bergundy: Wondering if this will cause a backfill on the entire table, if that's the case we should avoid that for this table which could be potentially very large.

rodrigozhou: Both MySQL and PostgreSQL support adding new columns with default values instantly. It's just a table metadata change.

rodrigozhou: > Is this necessary?

Not strictly, but makes the code cleaner.

common/persistence/sql/execution_util.go
rodrigozhou: `DataEncoding` is a string, so comparing with empty string is the same as comparing the size `> 0`.

yiminc (commented):
Do you need to change for sqlite as well?

yiminc (approved):

kellyRLagnew:
Out of curiosity, when is 1.27.1 expected to be released? Just trying to figure out if it is worth it to rollback our staging server to 1.26.2 to make it stable again, but if 1.27.1 is rolling out in the next couple days I won't bother

rodrigozhou:
We just released v1.27.1: https://github.com/temporalio/temporal/releases/tag/v1.27.1
Be sure to upgrade schema before upgrading Temporal to v1.27.1.

kellyRLagnew:
> We just released v1.27.1: https://github.com/temporalio/temporal/releases/tag/v1.27.1 Be sure to upgrade schema before upgrading Temporal to v1.27.1.

Amazing, thanks

PR #7387, opened 2025-02-25
CHASM: define exported Tree methods

yycptt:
## What changed?

- CHASM: define exported Tree methods

## Why?

- CHASM tree is the core of the entire CHASM framework. Although its own implementation is complex, by defining its exported method, we can unblock the work needed in mutable state, chasm engine & replication

## How did you test it?

- N/A, not actual implementation in this PR.

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -37,13 +37,17 @@ type (
 		// persistence *persistencespb.ChasmNode
 	}
 
+	// NodesMutation is a set of mutations for all nodes rooted at a given node n,
+	// including the node n itself.
 	NodesMutation struct {
-		UpdatedNodes map[string]*persistencespb.ChasmNode
+		UpdatedNodes map[string]*persistencespb.ChasmNode // flattened node path -> chasm node
 		DeletedNodes map[string]struct{}
 	}
 
+	// NodesSnapshot is a snapshot for all nodes rooted at a given node n,
+	// including the node n itself.
 	NodesSnapshot struct {
-		Nodes map[string]*persistencespb.ChasmNode
+		Nodes map[string]*persistencespb.ChasmNode // flattened node path -> chasm node
 	}
 
 	NodeBackend interface {
@@ -86,7 +90,9 @@ func (n *Node) Ref(
 }
 
 // Now implements the CHASM Context interface
-func (n *Node) Now() time.Time {
+func (n *Node) Now(
+	component Component,
+) time.Time {
 	panic("not implemented")
 }
 
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -1,0 +1,135 @@
+// The MIT License
+//
+// Copyright (c) 2025 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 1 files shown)


chasm/tree.go
```diff
@@ -0,0 +1,135 @@
...
+		// TODO: add necessary fields here, e.g.
+		//
+		// parent   *Node
+		// children map[string]*Node
+		// persistence *persistencespb.ChasmNode
+	}
+
+	NodesMutation struct {
+		UpdatedNodes map[string]*persistencespb.ChasmNode
+		DeletedNodes map[string]struct{}
```
lina-temporal, line 42: What key would represent the root node itself, is it `nil` or is it the root's path? I ask because if I were to implement `ApplyMutation`, I can go from the `UpdatedNodes[string]*ChasmNode` to `children map[string]*Node` by comparing their paths, but I'm not sure what would apply to `Node.persistence` itself. 

Should we add `Path()` to `Node`?

chasm/tree.go
```diff
@@ -0,0 +1,135 @@
...
+// It also performs consistency, access rule, and task validation checks
+// (for task processing requests) before returning the component.
+func (n *Node) Component(
+	chasmContext Context,
+	ref ComponentRef,
+) (Component, error) {
+	panic("not implemented")
+}
+
+// Ref implements the CHASM Context interface
```
lina-temporal, line 81: (minor) I think adding a `var _ chasm.Context = &Node{}`  to assert which interfaces are implemented adds to the readability.

chasm/tree.go
```diff
@@ -0,0 +1,135 @@
...
+// mutations from the source cluster.
+func (n *Node) ApplyMutation(
+	mutation NodesMutation,
+) error {
+	panic("not implemented")
+}
+
+// ApplySnapshot is used by replication stack to apply node
+// snapshot from the source cluster.
+//
```
lina-temporal, line 124: How are deletes intended to work with snapshots? Because we might process replication tasks out-of-order, I assume we can't simply remove any nodes that aren't present in the snapshot, but at the same time I don't see any built-in tombstoning. Do we just ignore keys present locally but missing in remote/the incoming snapshot until a `NodesMutation` with `DeletedNodes` is processed?

chasm/tree.go
yycptt: The `string` in the mutation/snapshot is the flattened/encoded _**path**_ for each node, and the `string` in `children map[string]*Node` is child name under it's parent, and not a full path.

The Tree implementation will decide how a path (`[]string`) should be flattened into a single `string` for persistence purpose while also supporting certain range query use cases (I have a separate doc on this). 

But this path conversion happens inside the tree implementation and at least right now I am not seeing a need for exporting this conversion logic. 

> what would apply to Node.persistence itself

Not sure I follow this question. Basically first conversion the flattened `string` to path (`[]string`), then the node in the tree and then override the `persistence` field with the node in the mutation. 

Probably the name `NodesMutation` causes the confusion? It's the mutation for all the nodes rooted at a given node. (though in practice the Mutation is also for the entire tree and mutate state only invokes method on the root of the tree)

chasm/tree.go
yycptt: Replication task might be processed out of order, but replication task processing logic itself won't apply a old snapshot to a workflow with newer state. So the workflow state will only keep moving forward and the implementation of apply snapshot  can simply remove any nodes not present in the snapshot.

lina-temporal: Oh, cool, that's a nice guarantee; does that ordering guarantee apply to mutations as well, or just snapshots?

chasm/tree.go
yycptt: > I'd been thinking this would also update the root's state.

I mean yeah it will include update for the root itself, if the map contains a key that is, say, an empty string. A node is also rooted at itself :) 

Yup I will add comments for Mutation & Snapshot struct.

chasm/tree.go
yycptt: It applies to both mutation and snapshot. 

Basically a mutable state mutation will say I am the mutation from versioned transition X to Y and a snapshot will say I am the snapshot for versioned transition Y.  Then all later mutations & snapshots applied by replication stack must have a versioned transition > Y.

chasm/tree.go
yycptt: Done.

chasm/tree.go
yycptt: Actually I can't write the assertion because chasm.Context has one method `getContext() context.Context`(maybe more in the future) not implemented by `*Node`, since it's still a `Context` :). 

So chasm.Context will be a very thin wrapper on top of Node, to also track the `context.Context`

alexshtin:
This will have minor conflicts with my work (I also added `tree.go`) but I can easily resolve them.

PR #7389, opened 2025-02-26
Fix start time calculation in schedule spec

dnr:
## What changed?
- Fix handling of start time in schedule specs that could result in GetNextTime returning the time that it was given.
- Introduce a SpecVersion as part of tweakables but separate from workflow logic version to control this behavior.
- Non-workflow logic can always use the latest version.
- Workflow logic is initialized with the fixed version, which is technically not downgrade-compatible, but it should affect very few workflows.

## Why?
This would lead to incorrect results and infinite loops.

## How did you test it?
new unit test, manual replay tests

## Potential risks
This was done in a way to be upgrade-compatible, but is not downgrade-compatible for workflows that hit this edge case after the upgrade.

### DIFF
--- a/service/worker/scheduler/spec.go
+++ b/service/worker/scheduler/spec.go
@@ -36,6 +36,7 @@ import (
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/cache"
 	"go.temporal.io/server/common/primitives/timestamp"
+	"go.temporal.io/server/common/util"
 )
 
 type (
@@ -295,9 +296,7 @@ func (cs *CompiledSpec) CanonicalForm() *schedulepb.ScheduleSpec {
 func (cs *CompiledSpec) GetNextTime(jitterSeed string, after time.Time) GetNextTimeResult {
 	// If we're starting before the schedule's allowed time range, jump up to right before
 	// it (so that we can still return the first second of the range if it happens to match).
-	if cs.spec.StartTime != nil && after.Before(timestamp.TimeValue(cs.spec.StartTime)) {
-		after = cs.spec.StartTime.AsTime().Add(-time.Second)
-	}
+	after = util.MaxTime(after, cs.spec.StartTime.AsTime().Add(-time.Second))
 
 	pastEndTime := func(t time.Time) bool {
 		return cs.spec.EndTime != nil && t.After(cs.spec.EndTime.AsTime())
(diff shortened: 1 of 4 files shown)


service/worker/scheduler/spec.go
```diff
@@ -295,9 +296,7 @@ func (cs *CompiledSpec) CanonicalForm() *schedulepb.ScheduleSpec {
 func (cs *CompiledSpec) GetNextTime(jitterSeed string, after time.Time) GetNextTimeResult {
 	// If we're starting before the schedule's allowed time range, jump up to right before
 	// it (so that we can still return the first second of the range if it happens to match).
-	if cs.spec.StartTime != nil && after.Before(timestamp.TimeValue(cs.spec.StartTime)) {
-		after = cs.spec.StartTime.AsTime().Add(-time.Second)
-	}
+	after = util.MaxTime(after, cs.spec.StartTime.AsTime().Add(-time.Second))
```
yiminc, line 299: is there risk of NPE on cs.spec.StartTime?

dnr: relying on AsTime to return unix epoch, which is way in the past. does it need a comment?

service/worker/scheduler/workflow.go
```diff
@@ -175,6 +175,7 @@ type (
 		ReuseTimer                        bool                     // Whether to reuse timer. Used for workflow compatibility.
 		NextTimeCacheV2Size               int                      // Size of next time cache (v2)
 		SpecFieldLengthLimit              int                      // item limit per spec field on the ScheduleInfo memo
+		SpecVersion                       SpecVersion              // version for spec logic
```
yiminc, line 178: Just curious, why not just a simple bool?

yiminc (approved):

service/worker/scheduler/spec_test.go
```diff
@@ -407,6 +407,20 @@ func (s *specSuite) TestSpecStartTime() {
 	)
 }
 
+func (s *specSuite) TestSpecStartTimeMinusOneSecond() {
```
lina-temporal, line 410: (minor) Maybe a comment to call out that the test case validates the `FixStartTimeBug` version case?

lina-temporal (approved):

service/worker/scheduler/workflow.go
dnr: Basically the same reason for `Version`: if we added a bool for each fix, the struct would get bigger and bigger. This way we can just increment a number.

PR #7393, opened 2025-02-26
Fix start time calculation in schedule spec

stephanos:
cherry-picked https://github.com/temporalio/temporal/commit/8985dc38910128f7b569b2c74934f3eef601678b

---

## What changed?
Fix handling of start time in schedule specs that could result in GetNextTime returning the time that it was given.

## Why?
This would lead to incorrect results and infinite loops.

## How did you test it?
new unit test, manual replay tests


dnr (approved):
approved if you want to merge, though we usually don't do PRs for release branch cherry-picks

PR #7395, opened 2025-02-26
Fix flaky test TestStartWorkflowExecution_UseExisting_OnConflictOptions

rodrigozhou:
## What changed?

Fix flaky tests `TestStartWorkflowExecution_UseExisting_OnConflictOptions`.

Fix proto assert library bugs.

## Why?

Callbacks in the describe workflow response is not in the same order as they are added.


bergundy (approved):
Thanks!

PR #7396, opened 2025-02-26
Fix potential NPE in workflow context

yycptt:
## What changed?

- Set workflowContext.MutableState only when mutable state is not nil
- The case won't happen today as `NewMutableStateFromDB` never returns nil, but could happen in the future.

## Why?

- Prevent potential NPE when the interface value is not nil but points to a nil value. 

## How did you test it?

- Tested in cicd

## Potential risks

- N/A

## Documentation

- N/A

## Is hotfix candidate?

- No

### DIFF
--- a/service/history/workflow/context.go
+++ b/service/history/workflow/context.go
@@ -282,8 +282,9 @@ func (c *ContextImpl) LoadMutableState(ctx context.Context, shardContext shard.C
 
 		// NOTE: we can't not assigned the result directly to c.MutableState like below
 		// c.MutableState, err = NewMutableStateFromDB(...)
-		// Otherwise c.MutableState itself will not be nil, but can potentially point to a nil *MutableStateImpl,
-		// and causing NPE (e.g. when calling c.Clear()) or other unexpected behavior.
+		// Otherwise c.MutableState (an interface) will not be nil, but can point to a nil *MutableStateImpl pointer
+		// returned by NewMutableStateFromDB().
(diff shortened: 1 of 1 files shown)


prathyushpv (approved):

service/history/workflow/context.go
```diff
@@ -279,6 +279,12 @@ func (c *ContextImpl) LoadMutableState(ctx context.Context, shardContext shard.C
 		if err != nil {
 			return nil, err
 		}
+
+		// NOTE: we can't not assigned the result directly to c.MutableState like below
+		// c.MutableState, err = NewMutableStateFromDB(...)
+		// Otherwise c.MutableState itself will not be nil, but can potentially point to a nil *MutableStateImpl,
+		// and causing NPE (e.g. when calling c.Clear()) or other unexpected behavior.
```
yiminc, line 286: Could you add comment that c.MutableState is an interface, and NewMutableStateFromDB() returns a pointer, thus it would end up be an non nil interface pointing to a nil pointer.

yiminc: Also, I think if we change the func NewMutableStateFromDB() to return MutableState interface, we would avoid this confusing behavior. Shall we do that?

yiminc (approved):

service/history/workflow/context.go
yycptt: hmm it's not trivial to just switch to return interface, some other implementations update the mutable state struct returned by it.  
There's also the golang "accept interface and return struct" convention, we mostly follow in the codebase.

PR #7406, opened 2025-02-28
Fix rate limiting in priority matcher

dnr:
## What changed?
- Add interface for task-specific absolute ready time (currently not task-specific yet)
- Use ready time to delay match
- Add new simple limiter that works with ready times
- Hook up existing dynamic configs and user-controlled limit

## Why?
- Accurate rate limiting in priority matcher
- Hooks for future per-priority, per-fairness key, etc. limits

## How did you test it?
not tested yet, on feature branch


service/matching/matcher_data.go
```diff
@@ -540,6 +578,28 @@ func (rt *resettableTimer) unset() {
...
+}
+
+func (s *simpleLimiter) set(rate float64, burst int) {
+	s.interval = int64(float64(time.Second) / rate)
+	s.burst = s.interval * int64(burst)
+}
+
+func (s *simpleLimiter) consume(now int64, tokens int64) {
+	s.ready = max(now, s.ready+s.burst) - s.burst + tokens*s.interval
+}
```
stephanos, line 601: As discussed, additional comment would be helpful to explain the overall intuition.

service/matching/matcher_data.go
```diff
@@ -540,6 +578,28 @@ func (rt *resettableTimer) unset() {
...
+type simpleLimiter struct {
+	// parameters:
+	interval int64 // nanos/task
+	burst    int64 // burst duration in nanos
+
+	// state:
+	ready int64 // unix nanos
+}
+
+func (s *simpleLimiter) set(rate float64, burst int) {
```
stephanos, line 594: As discussed, passing in burst duration seems slightly easier to follow.

stephanos (approved):
Approved with comments, as discussed.

PR #7407, opened 2025-02-28
CHASM: Wire up noop chasm tree with mutable state

yycptt:
## What changed?

- Wire up noop chasm tree with mutable state

## Why?

- To make sure we have the right methods on chasm tree and can work with rest of the stack.

## How did you test it?

- Unit test WIP

## Potential risks

- N/A, noop tree impl is used.

## Documentation

- N/A

## Is hotfix candidate?

- No


yycptt (commented):
cc @xwduan 
Please help review replication related changes: state retriever & applyMutation/Snapshot in mutable state. Thank you!

service/history/replication/sync_state_retriever.go
```diff
@@ -350,21 +350,18 @@ func (s *SyncStateRetrieverImpl) getNewRunInfo(ctx context.Context, namespaceId
 	}, nil
 }
 
-func (s *SyncStateRetrieverImpl) getMutation(mutableState workflow.MutableState, versionedTransition *persistencespb.VersionedTransition) (*persistencespb.WorkflowMutableStateMutation, error) {
+func (s *SyncStateRetrieverImpl) getMutation(
```
yycptt, line 353: I am separating the rewrite here into a separate PR as it fixes a race condition issue and also improves the performance of getMutation.

yycptt: https://github.com/temporalio/temporal/pull/7408

chasm/tree.go
```diff
@@ -105,20 +108,27 @@ func (n *Node) AddTask(
...
+) NodesSnapshot {
 	panic("not implemented")
 }
 
 // ApplyMutation is used by replication stack to apply node
 // mutations from the source cluster.
+//
+// NOTE: The behavior is undefined if UpdatedNodes and DeletedNodes have overlapping keys,
+// as the CHASM tree does not have enough information to known if the deletion happen
+// before or after the update.
```
lina-temporal, line 131: Should we define this as either add-wins or remove-wins up-front, so component authors understand the behavior during a conflict? I get that not leaving it up to the engine may leave an optimization off the table, but I'd find it very difficult to reason about this sort of conflict unless the behavior is clearly defined. Especially since it's unlikely we'd want this behavior to vary by engine.

yycptt: It's not related to component authors actually. From their pov, there's no mutation concept, they only see a consistent chasm tree. 

This method is for mutable state to move chasm tree from one state to another, and the comment essentially means mutable state should NOT provide a mutation with conflicting keys. The implementation of the method thus doesn't need to care about the order updates and deletions should be applied. So this undefined behavior is only for mutable state, not for component author. And mutable state must decide for conflicting keys, updates or deletions should be applied first. (see the applyMutation() method in mutable state, we always apply deletions first.)

lina-temporal: > And mutable state must decide for conflicting keys, updates or deletions should be applied first. (see the applyMutation() method in mutable state, we always apply deletions first.)

Okay, so it is defined as far as our engine goes, just defined at a level lower than CHASM itself, got it.

> the comment essentially means mutable state should NOT provide a mutation with conflicting keys. 

If it's never a valid configuration to pass both at once, and our own implementation wouldn't do that, then how about making it an error here instead of leaving it undefined?

yycptt: oh yeah that makes sense. I can update the comment and say conflicting key will be an error. 
The implementation of this method will need to check for this case though.

lina-temporal (approved):

PR #7409, opened 2025-03-01
CHASM: tree node serialization methods

alexshtin:
## What changed?

Added methods to serialized CHASM components to its proto representation.
Added many TODOs to address later.

## Why?

Part of CHASM workstream.

## How did you test it?

Added unit tests.

## Potential risks

No risks. Feature is still at the very early stage.

## Documentation

No.

## Is hotfix candidate?

No.


chasm/tree.go
```diff
@@ -39,28 +46,37 @@ type (
...
 		*nodeBase
 
-		parent      *Node
-		children    map[string]*Node
-		persistence *persistencespb.ChasmNode
+		parent   *Node
+		children map[string]*Node
+		pNode    *persistencespb.ChasmNode // serialized component/data/collection
+		// TODO: what is the better name?
+		component any // deserialized component/data/collection
```
lina-temporal, line 53: I think I like plain old `component` as well. The [HSM nomenclature was "`cache`"](https://github.com/temporalio/temporal/blob/main/service/history/hsm/tree.go#L205-L206), which does help illustrate that you're working on a local in-memory copy.

chasm/tree.go
```diff
@@ -39,28 +46,37 @@ type (
 	Node struct {
 		*nodeBase
 
-		parent      *Node
-		children    map[string]*Node
-		persistence *persistencespb.ChasmNode
+		parent   *Node
+		children map[string]*Node
+		pNode    *persistencespb.ChasmNode // serialized component/data/collection
```
lina-temporal, line 51: (minor, just my opinion) I'm not huge on `pNode` myself, just because my mind goes to Microsoft/"Hungarian-style" notation where `pFoo` meant 'pointer to Foo'. I'd find `node`, `persistence`/`persisted`, `persistedNode` a little more indicative.

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+	case *persistencespb.ChasmNode_ComponentAttributes:
+		return n.serializeComponentNode(mutation, currentPath)
+	case *persistencespb.ChasmNode_DataAttributes:
+		return n.serializeDataNode()
+	case *persistencespb.ChasmNode_CollectionAttributes:
+		panic("not implemented")
+	case *persistencespb.ChasmNode_PointerAttributes:
+		panic("not implemented")
+	default:
+		return serviceerror.NewInternal("unknown pNode type")
```
lina-temporal, line 162: I think `unknown ChasmNode type` would be a little clearer in the logs.

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+	}
+}
+
+func (n *Node) serializeComponentNode(
+	mutation *NodesMutation,
+	currentPath []string,
+) error {
+	componentT := reflect.TypeOf(n.component)
+	componentV := reflect.ValueOf(n.component)
+	for componentT.Kind() == reflect.Ptr {
```
lina-temporal, line 172:
```suggestion
	if componentT.Kind() == reflect.Ptr {
```

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+	}
+
+	protoMessageFound := false
+	processedChildren := make(map[string]struct{})
+	for i := 0; i < componentT.NumField(); i++ {
+		fieldV := componentV.Field(i)
+		fieldT := fieldV.Type()
+		if fieldT.AssignableTo(reflect.TypeFor[proto.Message]()) {
+			if protoMessageFound {
+				return serviceerror.NewInternal("only one proto field allowed in component")
```
lina-temporal, line 187: Maybe, `only one proto.Message field can be assignable on a Component`; if I saw "only one proto field allowed" as a component author, I might think it refers to my business logic.

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+		// TODO: support chasm name tag on the field
+		fieldName := componentT.Field(i).Name
+		processedChildren[fieldName] = struct{}{}
+
+		for fieldT.Kind() == reflect.Ptr {
+			fieldT = fieldT.Elem()
+			fieldV = fieldV.Elem()
+		}
+
+		if strings.HasPrefix(fieldT.String(), chasmFieldTypeName) { // not string comparison because chasm.Field is generic
```
lina-temporal, line 216: Do we ever have a field whose type is directly `chasm.Field`, or will it always be `chasm.FieldFoo`? Maybe `chasmFieldTypePrefix`/`chasmFieldTypeRoot`

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+		}
+	}
+
+	// remove deleted children
+	// TODO: this needs to be done recursively
+	// i.e remove the entire subtree
+	for childName := range n.children {
+		if _, processedChild := processedChildren[childName]; !processedChild {
+			mutation.DeletedNodes[strings.Join(append(currentPath, childName), "/")] = struct{}{}
+			delete(n.children, childName)
```
lina-temporal, line 259: I don't think I quite follow how this fits together, could you clarify for me? The `mutation` that we're getting in `serialize` is meant to be the mutation generated specifically during `CloseTransaction`, right, so we can finish populating it with deletes? If so, you might want to add a comment for that on the method that clarifies that (and also indicates that we mutate `mutations`).

yycptt (commented):
Haven't review the serialize logic yet but I feel it's better to split it into a separate PR given how complex it is (we need to update tree structure as well). Let's focusing on getting the decoding part fully implemented?

chasm/tree.go
alexshtin: It does. Limitation (at least for now) is that you can't define two `persistancespb.ActivityInfo` fields in your component.

chasm/field.go
```diff
@@ -30,17 +30,31 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// Used by reflection.
+const chasmFieldTypeName = "chasm.Field"
```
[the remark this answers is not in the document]
alexshtin: I added TODO for now.

chasm/tree.go
alexshtin: renamed this to `serializedValue`.

chasm/tree.go
alexshtin: and this to just `value`.

chasm/tree.go
```diff
@@ -70,6 +86,8 @@ type (
 	// where MutableState is defined.
 	NodeBackend interface {
 		// TODO: Add methods needed from MutateState here.
+		GetCurrentVersion() int64
+		NextTransitionCount() int64
```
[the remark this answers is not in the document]
alexshtin: not in this PR. Anyway I don't so far understand how we gonna use it.

chasm/tree.go
alexshtin: this is consequences of renames in GoLand. It sometimes does "smart" renames.

chasm/tree.go
alexshtin: My idea was to unroll all pointers if there are more than one. Inspired by `json.Unmarshal` which supports fields like `counter *******int` but I agree, it is pretty much useless, so I removed it.

chasm/tree.go
alexshtin: I added `[` to const.

chasm/tree.go
alexshtin: I rewrote children synchronization.

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+
+func (n *Node) deserializeNode(
+	instanceT reflect.Type,
+) error {
+	switch n.pNode.Attributes.(type) {
+	case *persistencespb.ChasmNode_ComponentAttributes:
+		return n.deserializeComponentNode(instanceT)
+	case *persistencespb.ChasmNode_DataAttributes:
+		return n.deserializeDataNode(instanceT)
+	case *persistencespb.ChasmNode_CollectionAttributes:
```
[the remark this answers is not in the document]
alexshtin: No. In the following one.

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+		panic("not implemented")
+	case *persistencespb.ChasmNode_PointerAttributes:
```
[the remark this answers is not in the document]
alexshtin: Added TODO for now. Will address it when I get to "pointer" support.

chasm/tree.go
```diff
@@ -106,17 +115,294 @@ func NewTree(
...
+	}
+	return nil
+}
+
+func (n *Node) deserializeComponentNode(
+	instanceT reflect.Type,
+) error {
+	if instanceT.Kind() != reflect.Struct {
+		return serviceerror.NewInternal("only struct is supported for component")
```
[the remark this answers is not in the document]
alexshtin: After putting some thoughts on it, I came to conclusion that "only pointers to structs" should be supported. Interfaces also (if they are implemented by pointer receiver).

chasm/tree.go
```diff
@@ -187,30 +473,12 @@ func (n *Node) ApplySnapshot(
...
-	}
-	childNode.insert(nodePath[1:], pNode, nodeBase)
-}
-
 func newNode(
 	base *nodeBase,
 	parent *Node,
 ) *Node {
 	return &Node{
+		dirty:    true,
```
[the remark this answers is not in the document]
alexshtin: yeah, this is wrong. `dirty` flag is not used/set yet. I left TODO for that too.

alexshtin:
Sorry, I did rebase and most of the comments are "Outdated" now. Which is actually true because I changed a lot. I addressed all comments though.

chasm/tree.go
```diff
@@ -214,36 +601,18 @@ func (n *Node) ApplySnapshot(
...
-		n.children[childName] = childNode
-	}
-	childNode.insert(nodePath[1:], pNode, nodeBase)
-}
-
 // IsDirty returns true if any node rooted at Node n has been modified.
 // The result will be reset to false after a call to CloseTransaction().
 func (n *Node) IsDirty() bool {
-	panic("not implemented")
+	return n.dirty
```
yycptt, line 607: this needs to recursively check all child nodes. The `dirty` flag is only for the current node, otherwise we can't compute mutations at the end of a transaction.

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+
+			value, err := n.unmarshalProto(n.serializedValue.GetComponentAttributes().GetData(), fieldT)
+			if err != nil {
+				return err
+			}
+			fieldV.Set(value)
+			continue
+		}
+
+		// chasm.Field field must be a pointer, i.e. *chasm.Field[T].
```
yycptt, line 461: I don't think it has to be a pointer. As long as the registered component that's using chasm.Field is a pointer we should be good I think. 
Is there a case you have in mind that doesn't work when Field is used without a pointer?

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+			}
+
+		case chasmCollectionTypePrefix:
+			// TODO: support collection
+			// init the map and populate
+			panic("not implemented")
+		}
+
+		// TODO: error or just go around?
+		// return serviceerror.NewInternal("unsupported field type in component " + fieldT.String())
```
yycptt, line 487: let's be strict in the beginning I'd say. It's always easier to relax a check than enforcing one.

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+		}
+		protoMessageFound = true
+
+		blob, err := serialization.ProtoEncodeBlob(fieldV.Interface().(proto.Message), enumspb.ENCODING_TYPE_PROTO3)
+		if err != nil {
+			return err
+		}
+
+		n.serializedValue.GetComponentAttributes().Data = blob
+		n.serializedValue.GetComponentAttributes().Type = nodeValueT.Elem().String() // type name w/o *
```
yycptt, line 311: This should be the registered component name, no go type name.

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+		if n.serializedValue.GetInitialVersionedTransition() == nil {
```
yycptt, line 312: initial versioned transition should only be set when the creating a new chasm node. It looks like you can do this as part of `setValue()`, though I am not 100% sure when that method will be called.

Then LastUpdateVersionedTransition is updated every time the node is mutated. Guess you can do it here.

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+	return err
+}
+
+func (n *Node) deleteChildren(removedPaths *[]string, childrenToKeep map[string]struct{}, currentPath []string) error {
+	for childName, childNode := range n.children {
+		if _, childToKeep := childrenToKeep[childName]; !childToKeep {
+			if err := childNode.deleteChildren(removedPaths, nil, append(currentPath, childName)); err != nil {
+				return err
+			}
+			path, err := n.pathEncoder.Encode(n, append(currentPath, childName))
```
yycptt, line 388:
```suggestion
			path, err := n.pathEncoder.Encode(childNode, append(currentPath, childName))
```

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+			}
+
+			childrenToKeep[fieldN] = struct{}{}
+		case chasmCollectionTypePrefix:
+			childrenToKeep[fieldN] = struct{}{}
+			// TODO: need to go over every item in collection and update children for it.
+			panic("not implemented")
+		}
+	}
+
```
yycptt, line 377: What about newly added fields? when will we add them to the child map?

Is that logic `reflectSubcomponents` or setValues()? I guess I am a bit confused where/when those two methods will be invoked.

yycptt (commented):
Would you mind give an overview of what are the main helper methods introduced by this PR and how they should be used by the closeTransaction() implementation?

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+	return nil
+}
+
+// Sync tree starting from node n with underlining component value:
+//   - if a child is nil in a corresponding component, it is removed from the tree,
+//   - if a child is no longer in a corresponding component, it is removed from the tree,
+//   - when a child is removed, all its children are removed too.
+//
+// Returns slice of paths to removed nodes.
+func (n *Node) syncChildren() ([]string, error) {
```
lina-temporal, line 328: When is this called in a node's/component's lifecycle? Is it just within `CloseTransaction` to get a list of removed paths, after an application deletes some of its nodes?

chasm/tree.go
```diff
@@ -137,6 +152,378 @@ func NewEmptyTree(
...
+	return f.Name
+}
+
+func (n *Node) reflectSubcomponents() error {
+	nodeValueT := reflect.TypeOf(n.value)
+	nodeValueV := reflect.ValueOf(n.value)
+
+	for i := 0; i < nodeValueT.Elem().NumField(); i++ {
+		fieldV := nodeValueV.Elem().Field(i)
+		fieldT := fieldV.Type()
```
lina-temporal, line 207: Have you considered creating a `Walk` method for walking a node's tree, to share this logic between `reflectSubcomponents`/`{de}serializeComponentNode`/`syncChildrenInternal`?

chasm/tree.go
alexshtin: I just don't want to support both cases and existing `Field` constructors return `*Field`. I agree that `Field` look more natural in component definition than `*Field`. I am flipping it to use `Field` and forbid `*Field`.

chasm/tree.go
```diff
@@ -344,6 +354,183 @@ func (n *Node) setSerializedNode(
...
+		if !fieldT.AssignableTo(protoMessageT) {
+			continue
+		}
+
+		if fieldV.IsNil() {
+			continue
+		}
```
yycptt, line 388: just to confirm,  this means component state should never be nil right?

chasm/tree.go
```diff
@@ -344,6 +354,183 @@ func (n *Node) setSerializedNode(
...
+		if err != nil {
+			return err
+		}
+
+		rc, ok := n.registry.componentFor(n.value)
+		if !ok {
+			return serviceerror.NewInternal(fmt.Sprintf("component type %s is not registered", nodeValueT.String()))
+		}
+
+		n.serializedNode.Data = blob
```
yycptt, line 405: so we guarantee `n.serializedNode` is not nil here right? even for newly created Fields?
and ^ is guaranteed by `syncSubComponents` below? So when later implementing CloseTransaction, we will call `syncSubComponents` and then `serializeComponentNode`?

yycptt (approved):

chasm/tree.go
```diff
@@ -344,6 +354,183 @@ func (n *Node) setSerializedNode(
 	childNode.setSerializedNode(nodePath[1:], serializedNode)
 }
 
+// serialize sets or updates serializedValue field of the node n with serialized value.
+func (n *Node) serialize() error {
```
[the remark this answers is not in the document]
alexshtin: I would do it in separate PR because it needs to be updated in `deserialize` too. TODO is already there.

chasm/tree.go
alexshtin: It can be `nil` and it indicates that `Data` field needs to be cleared. I am changing this part here.

chasm/tree.go
alexshtin: yes. `serializedNode` guaranteed to be not nil for top level component and for all sub component `syncSubComponents` should be called first. I split "sync" and "serialize" into two functions, because "sync" is recursive and applicable only to top level node, and "serialize" should be called individually on every node.

chasm/tree.go
```diff
@@ -344,6 +354,183 @@ func (n *Node) setSerializedNode(
...
+			// TODO: need to go over every item in collection and update children for it.
+			panic("not implemented")
+		}
+	}
+
+	err := n.deleteChildren(removedPaths, childrenToKeep, nodePath)
+	return err
+}
+
+func (n *Node) deleteChildren(removedPaths *[]string, childrenToKeep map[string]struct{}, currentPath []string) error {
```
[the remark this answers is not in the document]
alexshtin: It wasn't there when I started this PR :-). I am changing it.

PR #7410, opened 2025-03-01
Refactoring. Clean multableState interface

ychebotarev:
## What changed?

move stateMachineDefinition into a separate file.
Move some static definitions to mutable_state_impl.go

## Why?

Housekeeping.

## How did you test it?

existing tests.

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -136,6 +136,8 @@ var (
 	timeZeroUTC = time.Unix(0, 0).UTC()
 )
 
+var emptyTasks = []tasks.Task{}
+
 type (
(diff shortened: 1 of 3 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -136,6 +136,8 @@ var (
 	timeZeroUTC = time.Unix(0, 0).UTC()
 )
 
+var emptyTasks = []tasks.Task{}
```
alexshtin, line 139: I would remove this thing completely and use `nil` everywhere. I understand that it is not part of this PR.

alexshtin (approved):
I am always for keeping structs in dedicated files.

PR #7411, opened 2025-03-03
Assertion function

stephanos:
## What changed?

- Added a new function `softassert.That` that logs an error if the provided condition is false
- Added `testlogger.Testlogger` (copied and slightly modified from our internal repository) to functional tests

## Why?

To allow developers to verify pre- and post-conditions in their code.

**Why not panic?** Maybe in the future. For now, we're happy with finding these failed assertions in functional tests.

**What about nightlies/long-haul tests?** I've filed a ticket to detect a failed assertion there and surface them as an alert.

## How did you test it?

Planted a failing assertion.

**The error log:**
```
logger.go:146: 2025-03-04T20:50:08.578Z	ERROR	log/zap_logger.go:154	Unexpected Error log encountered: failed assertion: wake called twice	{"host": "127.0.0.1:35427", "component": "matching-engine", "wf-task-queue-name": "/_sys/temporal-sys-processor-parent-close-policy/1", "wf-task-queue-type": "Workflow", "wf-namespace": "temporal-system", "worker-build-id": "_unversioned_", "failed-assertion": true, "logging-call-at": "/home/runner/work/temporal/temporal/common/log/with_logger.go:72"}
        go.temporal.io/server/common/log.(*zapLogger).Error
        	/home/runner/work/temporal/temporal/common/log/zap_logger.go:154
        go.temporal.io/server/common/testing/testlogger.(*TestLogger).Error
        	/home/runner/work/temporal/temporal/common/testing/testlogger/testlogger.go:359
        go.temporal.io/server/common/log.(*withLogger).Error
        	/home/runner/work/temporal/temporal/common/log/with_logger.go:72
        go.temporal.io/server/common/log.(*withLogger).Error
        	/home/runner/work/temporal/temporal/common/log/with_logger.go:72
        go.temporal.io/server/common/log.(*withLogger).Error
        	/home/runner/work/temporal/temporal/common/log/with_logger.go:72
        go.temporal.io/server/common/log.(*withLogger).Error
        	/home/runner/work/temporal/temporal/common/log/with_logger.go:72
        go.temporal.io/server/common/softassert.That
        	/home/runner/work/temporal/temporal/common/softassert/softassert.go:50
        go.temporal.io/server/service/matching.(*waitableMatchResult).wake
        	/home/runner/work/temporal/temporal/service/matching/matcher_data.go:560
```

https://github.com/temporalio/temporal/actions/runs/13662479940/job/38196953773?pr=7411#step:9:1638

**The test result:**
```
=== Failed
=== FAIL: tests TestPriorityFairnessSuite/TestPriority_Activity_Basic (0.00s)
    functional_test_base.go:301: Running TestPriorityFairnessSuite/TestPriority_Activity_Basic in test shard 2/3
    functional_test_base.go:373: 
        	Error Trace:	/home/runner/work/temporal/temporal/tests/testcore/functional_test_base.go:373
        	            				/home/runner/work/temporal/temporal/tests/testcore/functional_test_base.go:272
        	            				/home/runner/work/temporal/temporal/tests/testcore/functional_test_base.go:254
        	            				/home/runner/work/temporal/temporal/tests/testcore/functional_test_sdk_suite.go:97
        	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:192
        	Error:      	Failing test as unexpected error logs were found.
        	            	Look for 'Unexpected Error log encountered'.
        	Test:       	TestPriorityFairnessSuite/TestPriority_Activity_Basic
```
https://github.com/temporalio/temporal/actions/runs/13662479940/job/38196953773?pr=7411#step:9:1053

### DIFF
--- a/common/softassert/softassert.go
+++ b/common/softassert/softassert.go
@@ -44,8 +44,7 @@ import (
 func That(logger log.Logger, condition bool, format string, args ...any) bool {
 	if !condition {
 		// By using the same prefix for all assertions, they can be reliably found in logs.
-		msg := fmt.Sprintf("failed assertion: %v", fmt.Sprintf(format, args...))
-		logger.Error(msg, tag.FailedAssertion())
+		logger.Error("failed assertion: "+fmt.Sprintf(format, args...), tag.FailedAssertion())
 	}
 	return condition
 }
--- a/common/softassert/softassert.go
+++ b/common/softassert/softassert.go
@@ -1,0 +1,51 @@
+// The MIT License
+//
+// Copyright (c) 2025 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package softassert
+
+import (
+	"fmt"
+
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+)
+
+// That performs a soft assertion by logging an error if the given condition is false.
+// It is meant to indicate a condition that are always expected to be true.
+// Returns true if the condition is met, otherwise false.
+//
+// Example:
+// softassert.That(logger, object.state == "ready", "object is not ready: %v", object.state)
+//
+// Best practices:
+// - Use it to check for programming errors and invariants.
+// - Use it to communicate assumptions about the code.
+// - Use it to abort or recover from an unexpected state.
+// - Never use it as a substitute for regular error handling, validation, or control flow.
+func That(logger log.Logger, condition bool, format string, args ...any) bool {
+	if !condition {
+		// By using the same prefix for all assertions, they can be reliably found in logs.
+		msg := fmt.Sprintf("failed assertion: %v", fmt.Sprintf(format, args...))
+		logger.Error(msg, tag.FailedAssertion())
+	}
+	return condition
+}
--- a/common/softassert/softassert.go
+++ b/common/softassert/softassert.go
@@ -23,8 +23,6 @@
 package softassert
 
 import (
-	"fmt"
-
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
 )
@@ -34,17 +32,17 @@ import (
 // Returns true if the condition is met, otherwise false.
 //
 // Example:
-// softassert.That(logger, object.state == "ready", "object is not ready: %v", object.state)
+// softassert.That(logger, object.state == "ready", "object is not ready")
 //
 // Best practices:
 // - Use it to check for programming errors and invariants.
 // - Use it to communicate assumptions about the code.
 // - Use it to abort or recover from an unexpected state.
 // - Never use it as a substitute for regular error handling, validation, or control flow.
-func That(logger log.Logger, condition bool, format string, args ...any) bool {
+func That(logger log.Logger, condition bool, msg string) bool {
 	if !condition {
 		// By using the same prefix for all assertions, they can be reliably found in logs.
-		logger.Error("failed assertion: "+fmt.Sprintf(format, args...), tag.FailedAssertion)
+		logger.Error("failed assertion: "+msg, tag.FailedAssertion)
 	}
 	return condition
 }
--- a/common/softassert/softassert.go
+++ b/common/softassert/softassert.go
@@ -44,7 +44,7 @@ import (
 func That(logger log.Logger, condition bool, format string, args ...any) bool {
 	if !condition {
 		// By using the same prefix for all assertions, they can be reliably found in logs.
-		logger.Error("failed assertion: "+fmt.Sprintf(format, args...), tag.FailedAssertion())
+		logger.Error("failed assertion: "+fmt.Sprintf(format, args...), tag.FailedAssertion)
 	}
 	return condition
 }
--- a/common/testing/testlogger/testlogger.go
+++ b/common/testing/testlogger/testlogger.go
@@ -1,0 +1,504 @@
+// The MIT License
+//
+// Copyright (c) 2025 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package testlogger
+
+import (
+	"container/list"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zaptest"
+)
+
+type Level string
+
+const (
+	Debug Level = "debug"
+	Info  Level = "info"
+	Warn  Level = "warn"
+	// Panics, DPanics, and Fatal logs are considered errors
+	Error Level = "error"
+)
+
+// This is a subset of the testing.T interface that we use in this package that is
+// also shared with *rapid.T.
+type TestingT interface {
+	zaptest.TestingT
+	Helper()
+	Name() string
+	Logf(format string, args ...any)
+	Log(args ...any)
+	Fatalf(format string, args ...any)
+	Fatal(args ...any)
+}
+
+// CleanupCapable is an interface that allows a test to register cleanup functions.
+// This is /not/ implemented by *rapid.T but is by *testing.T
+type CleanupCapableT interface {
+	TestingT
+	Cleanup(func())
+}
+
+func (l Level) String() string {
+	return string(l)
+}
+
+// Expectations represent errors we expect to happen in tests.
+// Their only purpose is to allow un-expecting (hah!) an error we've
+// marked as expected
+type Expectation struct {
+	e          *list.Element
+	testLogger *TestLogger
+	lvl        Level
+	unexpected bool
+}
+
+// Forget removes a previously registered expectation.
+// It will no longer be used to determine if a log message is expected or unexpected.
+func (e *Expectation) Forget() {
+	e.testLogger.Forget(e)
+}
+
+type matcher struct {
+	expectation *Expectation
+	msg         *regexp.Regexp
+	tags        map[string]*regexp.Regexp
+}
+
+func newMatcher(msg string, tags []tag.Tag, e *Expectation) matcher {
+	var rgx *regexp.Regexp
+	if msg != "" {
+		rgx = regexp.MustCompile(msg)
+	}
+	m := matcher{
+		expectation: e,
+		msg:         rgx,
+		tags:        map[string]*regexp.Regexp{},
+	}
+	for _, t := range tags {
+		m.tags[t.Key()] = regexp.MustCompile(formatValue(t))
+	}
+	return m
+}
+
+func (m matcher) Matches(msg string, tags []tag.Tag) bool {
+	// A nil regexp (meaning original string "") means we're only matching based on tags.
+	if m.msg != nil && !m.msg.MatchString(msg) {
+		return false
+	}
+
+	if len(m.tags) == 0 {
+		return true
+	}
+
+	// We need to match all tags specified in the matcher,
+	// not all tags on the message
+	remainingMatches := len(m.tags)
+	for _, t := range tags {
+		pat, ok := m.tags[t.Key()]
+		if !ok {
+			continue
+		}
+		if pat.MatchString(formatValue(t)) {
+			remainingMatches--
+		}
+	}
+	if remainingMatches == 0 {
+		return true
+	}
+	return false
+}
+
+type SharedTestLoggerState struct {
+	failOnDPanic atomic.Bool
+	failOnError  atomic.Bool
+	t            TestingT
+	mu           struct {
+		sync.RWMutex
+		expectations   map[Level]*list.List // Map[Level]List[matcher]
+		closed         bool
+		unexpectedErrs bool
+	}
+	logExpectations bool
+	logCaller       bool
+	level           zapcore.Level
+}
+
+// TestLogger is a log.Logger implementation that logs to the test's logger
+// _but_ will fail the test if log levels _above_ Warn are present
+type TestLogger struct {
+	wrapped log.Logger
+	state   *SharedTestLoggerState
+	tags    []tag.Tag
+}
+
+type LoggerOption func(*TestLogger)
+
+func DontFailOnError(t *TestLogger) *TestLogger {
+	t.state.failOnError.Store(false)
+	return t
+}
+
+func DontFailOnDPanic(t *TestLogger) *TestLogger {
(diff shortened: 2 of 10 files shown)


common/testing/testlogger/testlogger.go
stephanos, line 1: Taken verbatim from https://github.com/temporalio/saas-temporal/blob/main/common/testx/testlogger.go

common/softassert/softassert.go
```diff
@@ -0,0 +1,53 @@
...
+// Best practices:
+// - Use it to check for programming errors and internal invariants.
+// - Use it to communicate assumptions about the code.
+// - Use it to abort or recover from an unexpected state.
+// - Never use it as a substitute for regular error handling, validation, or control flow.
+func That(logger log.Logger, condition bool, format string, args ...any) bool {
+	if !condition {
+		// By using the same prefix for all assertions, they can be reliably found in logs.
+		msg := fmt.Sprintf("failed assertion: %v", fmt.Sprintf(format, args...))
+		logger.Error(msg, tag.FailedAssertion())
```
stephanos, line 50: Not sure if it needs both a prefix and a log tag? I'm not quite sure yet how people will search for this.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+}
+
+// Expectations represent errors we expect to happen in tests.
+// Their only purpose is to allow un-expecting (hah!) an error we've
+// marked as expected
+type Expectation struct {
+	e          *list.Element
+	testLogger *TestLogger
+	lvl        Level
+	unexpected bool
```
stephanos, line 81: Change from original: added this to denote expected und "unexpected" expectations.

PS: I considered renaming it to `Rule` but stuck with the current name.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+		tl.state.t.Helper()
+		tl.wrapped.Error(msg, tags...)
+		panic(failureMessage("Error", msg, tags))
+	}
+
+	// Labeling the error as unexpected; so it can be found later.
+	tl.wrapped.Error(errorMessage("Error", msg), tags...)
+	tl.state.mu.Lock()
+	tl.state.mu.unexpectedErrs = true
+	tl.state.mu.Unlock()
```
stephanos, line 386: Still logs the error, but adds a prefix to make it possible to find the error in the logs.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+}
+
+func (tl *TestLogger) isUnexpected(level Level, msg string, tags []tag.Tag) bool {
+	expectations := tl.state.mu.expectations[level]
+	for e := expectations.Front(); e != nil; e = e.Next() {
+		pat, ok := e.Value.(matcher)
+		if !ok {
+			tl.state.t.Fatalf("Bug in TestLogger: invalid %T value in matcher list", e.Value)
+		}
+		if pat.Matches(msg, tags) {
```
stephanos, line 314: This had a `if len(pat.tags) > 0 {` check but AFAICT the behavior is the same if the tags are passed in, empty or ot.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+			return pat.expectation.unexpected
```
stephanos, line 315: This was `return true` before, not it depends on the expectation type.

service/matching/pri_matcher.go
```diff
@@ -178,7 +183,7 @@ func (tm *priTaskMatcher) forwardTasks(lim quotas.RateLimiter, retrier backoff.R
 		if res.ctxErr != nil {
 			return // task queue closing
 		}
-		bugIf(res.task == nil, "bug: bad match result in forwardTasks")
+		softassert.That(tm.logger, res.task != nil, "expected a task from match")
```
stephanos, line 186: I removed the function name from these as the logger will indicate the location of the failed assertion (including a stacktrace).

tests/testcore/functional_test_base.go
```diff
@@ -193,7 +201,17 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(clusterConfigFile string, opt
 
 	// Logger might be already set by the test suite.
 	if s.Logger == nil {
-		s.Logger = log.NewTestLogger()
```
stephanos, line 196: I filed a ticket to port all `log.NewTestLogger()` calls to the new logger. Including xdc tests, too.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+func (tl *TestLogger) isUnexpected(level Level, msg string, tags []tag.Tag) bool {
+	expectations := tl.state.mu.expectations[level]
+	for e := expectations.Front(); e != nil; e = e.Next() {
+		m, ok := e.Value.(matcher)
+		if !ok {
+			tl.state.t.Fatalf("Bug in TestLogger: invalid %T value in matcher list", e.Value)
+		}
+		if m.Matches(msg, tags) {
+			return m.expectation.unexpected
+		}
```
stephanos, line 316: This had an `if len(pat.tags) > 0 {` check but AFAICT the behavior is the same if the tags are passed in, empty or not.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+
+func (tl *TestLogger) isUnexpected(level Level, msg string, tags []tag.Tag) bool {
+	expectations := tl.state.mu.expectations[level]
+	for e := expectations.Front(); e != nil; e = e.Next() {
+		m, ok := e.Value.(matcher)
+		if !ok {
+			tl.state.t.Fatalf("Bug in TestLogger: invalid %T value in matcher list", e.Value)
+		}
+		if m.Matches(msg, tags) {
+			return m.expectation.unexpected
```
stephanos, line 315: This was return true before, not it depends on the expectation type.

service/matching/pri_matcher.go
```diff
@@ -247,11 +252,13 @@ func (tm *priTaskMatcher) validateTasksOnRoot(lim quotas.RateLimiter, retrier ba
 		if res.ctxErr != nil {
 			return // task queue closing
 		}
-		bugIf(res.task == nil, "bug: bad match result in validateTasksOnRoot")
+		softassert.That(tm.logger, res.task != nil, "expected a task from match")
 
 		task := res.task
-		bugIf(task.forwardCtx != nil || task.isSyncMatchTask() || task.source != enumsspb.TASK_SOURCE_DB_BACKLOG,
```
stephanos, line 253: I split these out so it would be easier to know which one actually failed the assertion.

tests/testcore/functional_test_base.go
```diff
@@ -252,6 +270,7 @@ func (s *FunctionalTestBase) initAssertions() {
 	s.ProtoAssertions = protorequire.New(s.T())
 	s.HistoryRequire = historyrequire.New(s.T())
 	s.UpdateUtils = updateutils.New(s.T())
+	s.checkNoUnexpectedErrorLogs() // should have already been called in TearDownTest, but just in case
```
stephanos, line 273: This is not ideal as it labels the wrong test as failed, but it's better than forgetting about it altogether.

tests/testcore/functional_test_base.go
```diff
@@ -193,7 +201,17 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(clusterConfigFile string, opt
...
-		s.Logger = log.NewTestLogger()
+		tl := testlogger.NewTestLogger(s.T())
+
+		// Instead of failing immediately, TearDownTest will check for unexpected
+		// errors after each test completed. This is better since otherwise is would fail inside
+		// the server and not the test, creating a lot of noise and possibly stuck tests.
+		testlogger.DontFailOnError(tl)
+
+		// Fail test when an assertion fails (see `softassert` package).
+		tl.DontExpect(testlogger.Error, ".*", tag.FailedAssertion())
```
stephanos, line 212: At first I tried to explicitly `Expect` all errors, but after number 20 I realized that was a foolish endeavour.

service/matching/physical_task_queue_manager.go
```diff
@@ -206,7 +206,15 @@ func newPhysicalTaskQueueManager(
...
 			}
 		}
-		pqMgr.priMatcher = newPriTaskMatcher(tqCtx, config, queue.partition, fwdr, pqMgr.taskValidator, pqMgr.metricsHandler)
+		pqMgr.priMatcher = newPriTaskMatcher(
+			tqCtx,
+			config,
+			queue.partition,
+			fwdr,
+			pqMgr.taskValidator,
+			logger,
```
stephanos, line 215: Added the logger.

common/softassert/softassert.go
alexshtin: I am ok with both but nested `Sprintf`s is overkill.

common/testing/testlogger/testlogger.go
alexshtin, line 1: I prefer file per struct. But I will leave it up to you.

tests/deployment_test.go
```diff
@@ -117,6 +117,7 @@ func (s *DeploymentSuite) SetupTest() {
 }
 
 func (s *DeploymentSuite) TearDownTest() {
+	s.FunctionalTestBase.TearDownTest()
```
alexshtin, line 120: Was this just missing?

tests/testcore/functional_test_base.go
```diff
@@ -179,6 +180,13 @@ func (s *FunctionalTestBase) SetupSuite() {
 }
 
 func (s *FunctionalTestBase) TearDownSuite() {
+	// NOTE: We can't make s.Logger a testlogger.TestLogger because of AcquireShardSuiteBase.
```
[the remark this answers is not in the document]
alexshtin: `AcquireShardSuiteBase` this probably can be refactored now. As far as I remember it has custom logger for similar reason.

tests/testcore/functional_test_base.go
alexshtin: assign it to me.

tests/testcore/functional_test_base.go
```diff
@@ -193,7 +201,17 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(clusterConfigFile string, opt
 
 	// Logger might be already set by the test suite.
```
alexshtin, line 202: This also needs to be changed. It should be set only here.

alexshtin (approved):
This is really cool. Can't wait to use it!

common/testing/testlogger/testlogger.go
stephanos: I didn't want to make too many changes at oncee so Tim can still recognize his original code for reviewing purposes.

common/softassert/softassert.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// Example:
+// softassert.That(logger, object.state == "ready", "object is not ready: %v", object.state)
+//
+// Best practices:
+// - Use it to check for programming errors and invariants.
+// - Use it to communicate assumptions about the code.
+// - Use it to abort or recover from an unexpected state.
+// - Never use it as a substitute for regular error handling, validation, or control flow.
+func That(logger log.Logger, condition bool, format string, args ...any) bool {
```
dnr, line 44: arguably, it should be called `Thatf` (like `Printf`) but I'm not sure I like that

dnr, line 44: Since this might get used in hot loops, I wonder if a simpler version that just takes a static string is enough to give most of the benefits? I'm worried about the varargs doing allocation even when the condition is false. Ideally the whole thing could be inlined

common/softassert/softassert.go
```diff
@@ -0,0 +1,50 @@
...
+	if !condition {
+		// By using the same prefix for all assertions, they can be reliably found in logs.
+		logger.Error("failed assertion: "+fmt.Sprintf(format, args...), tag.FailedAssertion())
```
dnr, line 47: similarly, in the interest of keeping the code as small as possible, I'd get rid of the string prefix

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,504 @@
...
+			continue
+		}
+		if pat.MatchString(formatValue(t)) {
+			remainingMatches--
+		}
+	}
+	if remainingMatches == 0 {
+		return true
+	}
+	return false
```
dnr, line 137: simplify?
```suggestion
	return remainingMatches == 0
```

service/matching/matcher_data.go
```diff
@@ -243,15 +245,17 @@ type matcherData struct {
 	pollers pollerPQ
 	tasks   taskPQ
 
+	logger     log.Logger
```
dnr, line 248: should go right under `config` (always constant dependencies first, then mutable stuff)

tests/testcore/functional_test_base.go
dnr: iirc there was another test logger somewhere that did assertions.. should we replace that one too?

ah yeah: https://github.com/temporalio/temporal/blob/main/temporal/server_test.go#L188

tests/worker_deployment_version_test.go
```diff
@@ -109,6 +109,7 @@ func (s *DeploymentVersionSuite) SetupTest() {
 }
 
 func (s *DeploymentVersionSuite) TearDownTest() {
+	s.FunctionalTestBase.TearDownTest()
```
dnr, line 112: here and all other similar lines: should we do this after stopping the worker+client?

dnr (commented):
Do you want to do the new stuff on main and then rebase the branch to use it?

common/softassert/softassert.go
stephanos: Alright. I think we lose a bit of discovery there but the testlogger prefix and the logger tag hopefully make up for that.

tests/worker_deployment_version_test.go
stephanos: Good point.

service/matching/matcher_data.go
stephanos: 👍 I didn't know about that best practice.

common/softassert/softassert.go
stephanos: Hm, on second thought, I like how explicit and clear it is with the prefix. Since it's only prefixing on a failed assertion, that shouldn't be too often.

tests/testcore/functional_test_base.go
stephanos: I'll add that to the ticket 👍

common/softassert/softassert.go
dnr: it's more about the code size (which determines whether this call can be inlined) than the run time when the condition is false (which should be never of course)

dnr: on that note, can we make `tag.FailedAssertion` a static value instead of a function, so it's just a load instead of a function call?

service/matching/matcher_data.go
dnr: I'm not sure it's that well-established, I try to follow it in new code and sometimes refactor old. it makes the scope of locks and things more explicit (constants above a lock, things protected by it below, etc.)

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,501 @@
...
+// marked as expected
+type Expectation struct {
+	e          *list.Element
+	testLogger *TestLogger
+	lvl        Level
+	unexpected bool
+}
+
+// Forget removes a previously registered expectation.
+// It will no longer be used to determine if a log message is expected or unexpected.
```
tdeebswihart, line 85:
```suggestion
// A forgotten expectation will no longer be evaluated when errors are encountered.
```

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,501 @@
...
+		)
+		tl.wrapped = log.NewZapLogger(ztl).Skip(1)
+	}
+
+	// Only possible with a *testing.T until *rapid.T supports `Cleanup`
+	if ct, ok := t.(CleanupCapableT); ok {
+		// NOTE(tim): We don't care about anything logged after the test completes. Sure, this is racy,
+		// but it reduces the likelihood that we see stupid errors due to testing.T.Logf race conditions...
+		ct.Cleanup(tl.Close)
+	}
```
tdeebswihart, line 249: Note to OSS as new users of this code: if you call `t.Logf` after the test completes it will panic. When using this test logger you _must_ make sure to clean up your dependencies/goroutines before the test completes. The zaptest.Logger we rely upon uses `t.Log*` internally, so any logs that occur after the `testing.T` is marked as complete will cause a failure.

Ask me how I know 😅

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,501 @@
...
+}
+
+// ResetUnexpectedErrors resets the unexpected error state, returning the previous value.
+func (tl *TestLogger) ResetUnexpectedErrors() bool {
+	tl.state.mu.Lock()
+	defer tl.state.mu.Unlock()
+	prevUnexpectedLogs := tl.state.mu.unexpectedErrs
+	tl.state.mu.unexpectedErrs = false
+	return prevUnexpectedLogs
+}
```
tdeebswihart, line 261: `unexpectedErrs` is a confusing name for this, and has confused the logic.

What about changing it from a bool to a `Mode`, where the default is `FailOnUnexpectedErrors` and the alternate is `FailOnExpectedErrors`? That's much clearer at least to me

common/testing/testlogger/testlogger.go
tdeebswihart: Why are you setting `unexpectedErrs` here if it wasn't already set?

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,501 @@
...
+
+// Forget removes a previously registered expectation.
+// It will no longer be used to determine if a log message is expected or unexpected.
+func (tl *TestLogger) Forget(e *Expectation) {
+	tl.state.mu.Lock()
+	defer tl.state.mu.Unlock()
+	tl.state.mu.expectations[e.lvl].Remove(e.e)
+}
+
+func (tl *TestLogger) isUnexpected(level Level, msg string, tags []tag.Tag) bool {
```
tdeebswihart, line 304: I'm not a fan of `isUnexpected` as the new name for this (the original is `matchesExpectation`), as that combined with your new bool is confusing.

How about `shouldFailTest(...)`? That's clear about whats expected regardless of the mode this testlogger is in

tdeebswihart (changes requested):
I'm a big fan of eliminating spurious errors but the changes to the core logic are confusing. This needs some renames so its clear what the two modes actually do

common/testing/testlogger/testlogger.go
stephanos: Good idea 👍

common/testing/testlogger/testlogger.go
stephanos: I'm sorry, I don't understand the question.

stephanos:
FYI I rebased it on `main` and removed the priority-specific code. For some reason `TestAcquireShard_EventualSuccess` fails now. Investigating.

common/testing/testlogger/testlogger.go
```diff
@@ -0,0 +1,514 @@
...
+func (tl *TestLogger) Info(msg string, tags ...tag.Tag) {
+	tl.state.mu.RLock()
+	defer tl.state.mu.RUnlock()
+	if tl.state.mu.closed {
+		return
+	}
+	if tl.tags != nil {
+		tags = append(tags, tl.tags...)
+	}
+	_ = tl.shouldFailTest(Info, msg, tags)
```
tdeebswihart, line 423: Calls to this in levels below ERROR are vestigial and can be removed. Their purpose has long since been excised from the original codebase

tdeebswihart (approved):

common/testing/testlogger/testlogger.go
stephanos: 👍 Alright, I can remove them. I thought the purpose for them was logging.

PR #7414, opened 2025-03-04
[CHASM] Cassandra Persistence for CHASM nodes

lina-temporal:
## What changed?
Cassandra persistence support for Cassandra. MutableStateImpl implementations, including `ApplySnapshot` and `ApplyMutation`, are still WIP.

- Cassandra schema version bumped for new fields
- New queries for read/update/delete CHASM nodes on the executions table
- ExecutionManager/Serializer updates for CHASM nodes
- Updates to `persistence_interface`
- Wired in `chasm.Registry` under history's config
- Commented MutableStateImpl callsites, some stuff relies on tree implementation being merged
- common/persistence tests updated for chasm nodes

## How did you test it?
- `make unit-test && make lint`

## Potential risks
- Not in prod


proto/internal/temporal/server/api/persistence/v1/workflow_mutable_state.proto
```diff
@@ -36,6 +37,7 @@ message WorkflowMutableState{
     map<int64, ChildExecutionInfo> child_execution_infos = 3;
     map<int64, RequestCancelInfo> request_cancel_infos = 4;
     map<int64, SignalInfo> signal_infos = 5;
+    map<string, ChasmNode> chasm_nodes = 12;
```
lina-temporal, line 40: @yycptt  I followed your diff precisely when I saw we both depended on the same changes, so, whoever's merges first, I suppose :)

common/persistence/cassandra/mutable_state_store.go
yycptt, line 1: Plz update `applyWorkflowSnapshotBatchAsReset` path as well.

The read path will be in a separate PR?

service/history/workflow/mutable_state_impl.go
yycptt, line 1: I think my PR will cover all the TODOs here.

service/history/fx.go
yycptt, line 1: Thanks for wiring update the Registry! I'd suggest split this to a separate PR though, to keep this PR focused on persistence layer changes.

yycptt (commented):
Shall we focus on persistence layer changes in this PR? Wiring up with mutable state and shard context can be done in separate PRs I think.

It will also be ideal to add some test for the persistence implementation. and only exercise the code path for cassandra tests. I am thinking change `RandomSnapshot` and `RandomMutation` to only populate chasm related fields if test name starts with `TestCassandra`, then once we have the implementation for all persistence layers we can get rid of the special check.

lina-temporal:
> Shall we focus on persistence layer changes in this PR? Wiring up with mutable state and shard context can be done in separate PRs I think.

Yeah, I can split them out. 

> I am thinking change `RandomSnapshot` and `RandomMutation` to only populate chasm related fields if test name starts with `TestCassandra`, then once we have the implementation for all persistence layers we can get rid of the special check.

Sure, will do.

common/persistence/data_interfaces.go
```diff
@@ -732,6 +736,7 @@ type (
 		BufferedEventsCount    int
 		TaskCountByCategory    map[string]int
 		UpdateInfoCount        int
+		ChasmSizeByPath        map[string]int // size of each individual CHASM node by path
```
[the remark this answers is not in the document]
lina-temporal: > This should be the # of (pending) chasm nodes in the map I think. 

We don't need to track each chasm node's size? Plus, you can get the current number of chasm nodes with `len(ChasmSizeByPath)`.

common/persistence/cassandra/mutable_state_store.go
lina-temporal: > Plz update applyWorkflowSnapshotBatchAsReset path as well.

Will do.

> The read path will be in a separate PR?

I was planning to add it to this PR, I'd wanted to get feedback on the write path early. Happy to put it in a separate one if that's preferable, though!

yycptt: ah the PR is not a draft so I thought you want to split the read logic out. One PR for write + read makes sense to me.

common/persistence/data_interfaces.go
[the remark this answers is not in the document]
yycptt: good question. I don't see a good use case if we just emit size metric for each chasm node naively. The percentile may be skewed by collection and pointer as they don't have any payload. To make it useful, we probably want to use separate histogram for each registered chasm component typed.

I guess right now I mostly care about the total size. pending & total count we can probably add them later when we have a clear idea how we want to use them.

service/history/fx.go
lina-temporal: Moved registry to [a separate PR](https://github.com/temporalio/temporal/pull/7419)

common/persistence/data_interfaces.go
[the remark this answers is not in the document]
lina-temporal: > good question. I don't see a good use case if we just emit size metric for each chasm node naively. The percentile may be skewed by collection and pointer as they don't have any payload. To make it useful, we probably want to use separate histogram for each registered chasm component typed.

Yeah, a histogram for specific types would be very useful from the "temporal operator" perspective. I'd been imagining this used as a way for CHASM component authors to measure and optimize the sizes of their different component types' state blobs (particularly useful since we'll support partial writes). 

> I guess right now I mostly care about the total size. pending & total count we can probably add them later when we have a clear idea how we want to use them.

Sure, I'll remove the other two till we better define them.

common/persistence/cassandra/util.go
```diff
@@ -1034,71 +1131,15 @@ func updateBufferedEvents(
...
-		rcMap[initiatedID] = blob.Data
-	}
-
-	return rcMap, encoding, nil
-}
-
-func resetSignalInfoMap(
-	signalInfos map[int64]*commonpb.DataBlob,
-) (map[int64][]byte, enumspb.EncodingType, error) {
+func convertBlobMapToByteMap[T comparable](
```
yycptt, line 1134: 👍

common/persistence/data_interfaces.go
```diff
@@ -720,6 +723,7 @@ type (
 		SignalInfoSize        int
 		SignalRequestIDSize   int
 		BufferedEventsSize    int
+		ChasmTotalSize        int // total size of all CHASM nodes within a record
```
yycptt, line 726: populate & emit metric for this field? similar to other fields defined in the struct.

lina-temporal: Thanks, will fix.

alexshtin (approved):

PR #7420, opened 2025-03-05
Support updating closed workflows by mutation

yycptt:
## What changed?

- Support update closed workflow by mutation

## Why?

- Today, any updates (e.g. update callbacks after workflows & update branch token. With CHASM, there will be more cases) for closed workflows require persisting the entire mutable snapshot using the `SetWorkflowExecution` persistence API in order to skip the current record validation. 
- Application logic is complex and have to call different APIs based on workflow state
- Performance is also bad as we are writing unnecessary data when persisting a snapshot, even if they haven't been mutated.

## How did you test it?

- Unit & persistence integration tests

## Is hotfix candidate?

- No.

### DIFF
--- a/common/persistence/data_interfaces.go
+++ b/common/persistence/data_interfaces.go
@@ -82,12 +82,12 @@ const (
 	// UpdateWorkflowModeBypassCurrent update workflow, without current record
 	// NOTE: current record CANNOT point to the workflow to be updated
 	UpdateWorkflowModeBypassCurrent
-	// UpdateWorkflowModeSkipCurrent update workflow, without checking or update current record
-	// This mode should only be used when we don't know if the workflow being updated is the current workflow or not.
+	// UpdateWorkflowModeIgnoreCurrent update workflow, without checking or update current record.
+	// This mode should only be used when we don't know if the workflow being updated is the current workflow or not in DB.
 	// For example, when updating a closed workflow, it may or may not be the current workflow.
 	// This is similar to SetWorkflowExecution, but UpdateWorkflowExecution with this mode persists the workflow as a mutation,
 	// instead of a snapshot.
-	UpdateWorkflowModeSkipCurrent
+	UpdateWorkflowModeIgnoreCurrent
 )
 
 // ConflictResolveWorkflowMode conflict resolve mode
--- a/common/persistence/data_interfaces.go
+++ b/common/persistence/data_interfaces.go
@@ -82,6 +82,12 @@ const (
 	// UpdateWorkflowModeBypassCurrent update workflow, without current record
 	// NOTE: current record CANNOT point to the workflow to be updated
 	UpdateWorkflowModeBypassCurrent
+	// UpdateWorkflowModeSkipCurrent update workflow, without checking or update current record
+	// This mode should only be used when we don't know if the workflow being updated is the current workflow or not.
+	// For example, when updating a closed workflow, it may or may not be the current workflow.
+	// This is similar to SetWorkflowExecution, but UpdateWorkflowExecution with this mode persists the workflow as a mutation,
+	// instead of a snapshot.
+	UpdateWorkflowModeSkipCurrent
 )
 
 // ConflictResolveWorkflowMode conflict resolve mode
--- a/service/history/ndc/activity_state_replicator.go
+++ b/service/history/ndc/activity_state_replicator.go
@@ -176,11 +176,11 @@ func (r *ActivityStateReplicatorImpl) SyncActivityState(
 		return err
 	}
 
-	if r.shardContext.GetConfig().EnableUpdateClosedWorkflowByMutation() {
+	if r.shardContext.GetConfig().EnableUpdateWorkflowModeIgnoreCurrent() {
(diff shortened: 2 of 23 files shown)


common/persistence/data_interfaces.go
```diff
@@ -82,6 +82,12 @@ const (
 	// UpdateWorkflowModeBypassCurrent update workflow, without current record
 	// NOTE: current record CANNOT point to the workflow to be updated
 	UpdateWorkflowModeBypassCurrent
+	// UpdateWorkflowModeSkipCurrent update workflow, without checking or update current record
+	// This mode should only be used when we don't know if the workflow being updated is the current workflow or not.
+	// For example, when updating a closed workflow, it may or may not be the current workflow.
+	// This is similar to SetWorkflowExecution, but UpdateWorkflowExecution with this mode persists the workflow as a mutation,
+	// instead of a snapshot.
+	UpdateWorkflowModeSkipCurrent
```
yycptt, line 90: cc @alfred-landrum 
This could be a topic for our Thursday meeting :)

yux0:
LGTM. No concern on this change.

service/history/ndc/activity_state_replicator.go
```diff
@@ -176,6 +176,11 @@ func (r *ActivityStateReplicatorImpl) SyncActivityState(
 		return err
 	}
 
+	if r.shardContext.GetConfig().EnableUpdateClosedWorkflowByMutation() {
+		return executionContext.UpdateWorkflowExecutionAsPassive(ctx, r.shardContext)
+	}
+
+	// TODO: remove following code once EnableUpdateClosedWorkflowByMutation config is deprecated.
 	updateMode := persistence.UpdateWorkflowModeUpdateCurrent
 	if state, _ := mutableState.GetWorkflowStateStatus(); state == enumsspb.WORKFLOW_EXECUTION_STATE_ZOMBIE {
```
yycptt, line 185: The previous logic is kinda wrong/misleading as it making decision based on the in memory workflow state instead of the workflow state in DB. The `updateMode` though, is for doing current record CAS in db, so we need to pick it based on the state in db (i.e. the workflow state before the current mutable state transaction)

The logic still works because activity state replicator (and same for the hsm state replicator) below don't really change the workflow state, so in memory state == state in DB.

service/history/statemachine_environment.go
```diff
@@ -395,6 +395,11 @@ func (e *stateMachineEnvironment) Access(ctx context.Context, ref hsm.Ref, acces
 		return nil
 	}
 
+	if e.shardContext.GetConfig().EnableUpdateClosedWorkflowByMutation() {
+		return wfCtx.UpdateWorkflowExecutionAsActive(ctx, e.shardContext)
+	}
+
+	// TODO: remove following code once EnableUpdateClosedWorkflowByMutation config is deprecated.
 	if ms.GetExecutionState().State == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED {
```
yycptt, line 403: Same here. The hsm framework doesn't really change workflow state right now.

common/persistence/data_interfaces.go
gow, line 90: This does conflict a lot with `UpdateWorkflowModeBypassCurrent`.
Can we consider something like `UpdateWorkflowModeIgnoreCurrent`? It'll be even better if we can get rid of one of these modes all together (if possible)

service/history/visibility_queue_task_executor.go
```diff
@@ -461,6 +461,12 @@ func (t *visibilityQueueTaskExecutor) cleanupExecutionInfo(
 	executionInfo.Memo = nil
 	executionInfo.SearchAttributes = nil
 	executionInfo.RelocatableAttributesRemoved = true
+
+	if t.shardContext.GetConfig().EnableUpdateClosedWorkflowByMutation() {
+		return weContext.UpdateWorkflowExecutionAsPassive(ctx, t.shardContext)
```
gow, line 466: Curious why `UpdateWorkflowExecutionAsPassive` here and HSM state replicator. Why not `UpdateWorkflowExecutionAsActive`?

gow (approved):
Looks good. Some nit suggestions and comments below.

common/persistence/data_interfaces.go
yycptt: yeah makes sense. `IgnoreCurrent` is a better name.

This updateMode is part of the persistence interface so not trivial to get rid of any existing ones. 
Long term if we ever get a chance to redo our persistence interface, I think the interface can explicit ask to the caller to state which columns/rows to update and to what, including changes for current rows. That will make things much more explicit. Right now the current row concept, which is an application concept is almost entirely in persistence layer...

service/history/visibility_queue_task_executor.go
yycptt: HSM/Activity state replicator are replication logic for applying changes in the standby cluster of a namespace. And when "AsPassive" is used, the mutable state logic will prevent another replication task from being generated (there are some other differences, but this is the most important one).

For visibility, you can think this the removal of Memo & SearchAttributes is a cluster local operation (move the data from mutable state to ES).  and this operation doesn't need to be replicated.

PR #7423, opened 2025-03-05
Double QPS queue size for estimator

zigeH:
## What changed?

Estimator should take more data point to stabilize the estimation.
More discussion here: 
https://temporaltechnologies.slack.com/archives/C04K4CSN8MC/p1741032584001869?thread_ts=1741029359.041549&cid=C04K4CSN8MC

### DIFF
--- a/service/worker/migration/force_replication_workflow.go
+++ b/service/worker/migration/force_replication_workflow.go
@@ -502,7 +502,7 @@ func enqueueReplicationTasks(ctx workflow.Context, workflowExecutionsCh workflow
 // running simultaneously and the initial starting point.
 func NewQPSQueue(concurrentActivityCount int) QPSQueue {
 	return QPSQueue{
-		Data:    make([]QPSData, 0, concurrentActivityCount+1),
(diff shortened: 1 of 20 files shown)


service/worker/migration/force_replication_workflow.go
```diff
@@ -502,7 +502,7 @@ func enqueueReplicationTasks(ctx workflow.Context, workflowExecutionsCh workflow
 // running simultaneously and the initial starting point.
 func NewQPSQueue(concurrentActivityCount int) QPSQueue {
 	return QPSQueue{
-		Data:    make([]QPSData, 0, concurrentActivityCount+1),
+		Data:    make([]QPSData, 0, max(0, 2*concurrentActivityCount+1)),
```
hehaifengcn, line 505: how about making multiplier factor (number 2) here configurable so we can try out different number at least?

hehaifengcn (approved):

yux0 (approved):

zigeH:
> Approval for Poc. I am not sure how good it can solve the problem.

agree. I think we should further discuss it.

PR #7424, opened 2025-03-05
Don’t reapply termination event if workflow is terminated by system

hai719:
## What changed?
Don’t reapply termination event if workflow is terminated by system

## Why?
Workflow excutions can be terminated by user, or system (reset, conflict resolution, size exceeds limit, etc). Only termination initiated by user should be reapplied.

## How did you test it?
unit test.

### DIFF
--- a/service/history/ndc/workflow_resetter.go
+++ b/service/history/ndc/workflow_resetter.go
@@ -962,6 +962,9 @@ func reapplyEvents(
 				continue
 			}
 			attr := event.GetWorkflowExecutionTerminatedEventAttributes()
+			if attr.GetIdentity() == consts.IdentityHistoryService || attr.GetIdentity() == consts.IdentityResetter {
+				continue
+			}
 			if err := workflow.TerminateWorkflow(
 				mutableState,
 				attr.GetReason(),
--- a/service/history/workflow/util.go
+++ b/service/history/workflow/util.go
@@ -253,16 +253,18 @@ func GetEffectiveVersioningBehavior(versioningInfo *workflowpb.WorkflowExecution
 	return versioningInfo.GetBehavior()
 }
 
-// shouldReapplyEvent returns true if the event should be reapplied to the workflow execution.
-func shouldReapplyEvent(stateMachineRegistry *hsm.Registry, event *historypb.HistoryEvent) bool {
+// ShouldReapplyEvent returns true if the event should be reapplied to the workflow execution.
+func ShouldReapplyEvent(stateMachineRegistry *hsm.Registry, event *historypb.HistoryEvent) bool {
 	switch event.GetEventType() { // nolint:exhaustive
 	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED,
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ADMITTED,
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED,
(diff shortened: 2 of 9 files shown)


service/history/workflow/util.go
```diff
@@ -253,16 +253,18 @@ func GetEffectiveVersioningBehavior(versioningInfo *workflowpb.WorkflowExecution
 	return versioningInfo.GetBehavior()
 }
 
-// shouldReapplyEvent returns true if the event should be reapplied to the workflow execution.
-func shouldReapplyEvent(stateMachineRegistry *hsm.Registry, event *historypb.HistoryEvent) bool {
+// ShouldReapplyEvent returns true if the event should be reapplied to the workflow execution.
+func ShouldReapplyEvent(stateMachineRegistry *hsm.Registry, event *historypb.HistoryEvent) bool {
```
yycptt, line 257: is this function being used outside workflow package?

service/history/workflow/util.go
```diff
@@ -253,16 +253,18 @@ func GetEffectiveVersioningBehavior(versioningInfo *workflowpb.WorkflowExecution
...
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ADMITTED,
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED,
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_CANCEL_REQUESTED,
-		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED,
-		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_TERMINATED:
+		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED:
 		return true
+	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_TERMINATED:
+		attr := event.GetWorkflowExecutionTerminatedEventAttributes()
+		return attr.GetIdentity() != consts.IdentityHistoryService && attr.GetIdentity() != consts.IdentityResetter
```
yycptt, line 267: Do we need the same check in workflow_resetter.go reapplyEvents()?

hai719: Moved the check into workflow_resetter.go reapplyEvents().

service/history/workflow/util.go
hai719: No. Moved the check out of this function.

service/history/workflow/util.go
yycptt: hmm I thought we need both?

hai719: All paths (reset, conflict resolution) call workflow_resetter.go reapplyEvents().

yycptt: oh I see. yeah... eventually it will be filtered out. We might be able to save a rpc call is the check is done in both places. no strong opinion either way. I guess conflict resolution is rare enough and we don't care about the extra call here.

service/history/ndc/workflow_resetter.go
```diff
@@ -962,6 +962,9 @@ func reapplyEvents(
 				continue
 			}
 			attr := event.GetWorkflowExecutionTerminatedEventAttributes()
+			if attr.GetIdentity() == consts.IdentityHistoryService || attr.GetIdentity() == consts.IdentityResetter {
```
yycptt, line 965: actually shall we check the terminate reason? I think history service can terminate workflows for bunch of other reasons

yycptt (approved):

PR #7425, opened 2025-03-05
Fix sync workflow state task

xwduan:

### DIFF
--- a/service/history/ndc/workflow_state_replicator.go
+++ b/service/history/ndc/workflow_state_replicator.go
@@ -79,6 +79,14 @@ type (
 		) error
 	}
 
+	WorkflowStateReplicatorProvider func(
+		shard.Context,
+		wcache.Cache,
+		EventsReapplier,
+		serialization.Serializer,
+		log.Logger,
+	) WorkflowStateReplicator
+
 	WorkflowStateReplicatorImpl struct {
 		shardContext      shard.Context
 		namespaceRegistry namespace.Registry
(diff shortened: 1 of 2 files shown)


service/history/ndc/workflow_state_replicator.go
```diff
@@ -79,6 +79,14 @@ type (
 		) error
 	}
 
+	WorkflowStateReplicatorProvider func(
```
yycptt, line 82: hmm I don't see this used anywhere? looks like we don't need it?

xwduan: yeah, no longer needed. Will delete.

yycptt (approved):

service/history/replication/executable_workflow_state_task.go
```diff
@@ -147,6 +147,35 @@ func (e *ExecutableWorkflowStateTask) HandleErr(err error) error {
...
+		ctx, cancel := newTaskContext(namespaceName, e.Config.ReplicationTaskApplyTimeout())
+		defer cancel()
+
+		if doContinue, syncStateErr := e.SyncState(
+			ctx,
+			retryErr,
+			ResendAttempt,
+		); syncStateErr != nil || !doContinue {
+			if syncStateErr != nil {
+				e.Logger.Error("VerifyVersionedTransition replication task encountered error during sync state",
```
hai719, line 167: nit: "workflow state replication task"

hai719 (approved):

PR #7427, opened 2025-03-05
Functional tests for reset workflow with child

prathyushpv:
## What changed?
Added functional tests for reset workflows with child workflows.

## Why?
More test coverage.


service/history/ndc/workflow_resetter.go
```diff
@@ -796,22 +796,22 @@ func (r *workflowResetterImpl) reapplyEventsFromBranch(
 		}
 		// TODO: uncomment this code to enable phase 2 of reset workflow feature.
 		// track the child workflows initiated after reset-point
-		// if allowResetWithPendingChildren {
```
yycptt, line 799: we need to keep this feature disabled right now. If we want to exercise the code path in functional test, we need a separate flag for this I think.

prathyushpv: ack

prathyushpv: I have made changes to comment these lines and skip these tests. We can enable these tests when we decide to enable this feature.

tests/workflow_reset_with_child_test.go
```diff
@@ -0,0 +1,899 @@
...
+//     c: If the child was terminated/failed at the time of reset: Expect the parent to make progress by retaining the
+//     termination/failure from the child.
+type WorkflowResetWithChildSuite struct {
+	testcore.FunctionalTestSdkSuite
+}
+
+func TestWorkflowResetWithChildTestSuite(t *testing.T) {
+	t.Parallel()
+	// TODO: Enable this test when reset phase 2 is enabled.
+	t.Skip("Skipping until reset phase 2 is enabled")
```
yycptt, line 72: I think we can still enabled tests for case 3? Reset point before child init is reset phase 1.

tests/workflow_reset_with_child_test.go
```diff
@@ -0,0 +1,899 @@
...
+		Namespace: s.Namespace().String(),
+		WorkflowExecution: &commonpb.WorkflowExecution{
+			WorkflowId: wfID,
+			RunId:      firstRun.GetRunID(),
+		},
+		Reason: "integration test",
+	}
+
+	// (reset #2) - resetting the new workflow execution after child-1 while child-1 is still running
+	resetRequest.RequestId = "reset-request-2"
```
yycptt, line 603: nit: reset-1 ?

tests/workflow_reset_with_child_test.go
```diff
@@ -0,0 +1,899 @@
...
+		},
+		Reason: "integration test",
+	}
+
+	// Let the second child finish now
+	err = s.SdkClient().SignalWorkflow(context.Background(), initialChildExecutions[0].WorkflowId, initialChildExecutions[0].RunId, "continue", "")
+	s.NoError(err)
+
+	// (reset #2) - resetting the new workflow execution after child-1 while child-1 is still running
+	resetRequest.RequestId = "reset-request-2"
```
yycptt, line 657: > // Let the second child finish now

There's only one child?

> (reset #2)

reset # 1?

And do we need to explicit check the child workflow is completed or child completion is recorded in parent before sending the reset request?

yycptt (approved):

tests/workflow_reset_with_child_test.go
prathyushpv: Done

tests/workflow_reset_with_child_test.go
prathyushpv: Sorry that is a mistake when I did copy-paste. 
Also, yes we have to check if child workflow has completed. Let me add that. Thanks!

PR #7429, opened 2025-03-05
Move two settings into dynamic config

dnr:
## What changed?
- Add dynamic config for `matching.getTasksReloadAt`
- Add dynamic config for `matching.backlogTaskForwardTimeout`

## Why?
tuning ability

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1287,12 +1287,17 @@ second per poller by one physical queue manager`,
 	MatchingUseNewMatcher = NewTaskQueueBoolSetting(
 		"matching.useNewMatcher",
 		false,
-		`Use priority-enabled TaskMatcher.`,
+		`Use priority-enabled TaskMatcher`,
 	)
 	MatchingPriorityLevels = NewTaskQueueIntSetting(
 		"matching.priorityLevels",
 		5,
-		`Number of simple priority levels`,
+		`Number of simple priority levels (requires new matcher)`,
+	)
+	MatchingBacklogTaskForwardTimeout = NewTaskQueueDurationSetting(
+		"matching.backlogTaskForwardTimeout",
+		60*time.Second,
+		`Timeout for forwarded backlog task (requires new matcher)`,
 	)
 
 	// keys for history
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1056,7 +1056,12 @@ See DynamicRateLimitingParams comments for more details.`,
 	MatchingGetTasksBatchSize = NewTaskQueueIntSetting(
 		"matching.getTasksBatchSize",
 		1000,
-		`MatchingGetTasksBatchSize is the maximum batch size to fetch from the task buffer`,
+		`How many backlog tasks to read from persistence at once`,
+	)
+	MatchingGetTasksReloadAt = NewTaskQueueIntSetting(
+		"matching.getTasksReloadAt",
+		100,
+		`Reload a batch of tasks when there are this many remaining. Must be less than MatchingGetTasksBatchSize.`,
 	)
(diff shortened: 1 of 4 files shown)


common/dynamicconfig/constants.go
```diff
@@ -1056,7 +1056,12 @@ See DynamicRateLimitingParams comments for more details.`,
 	MatchingGetTasksBatchSize = NewTaskQueueIntSetting(
 		"matching.getTasksBatchSize",
 		1000,
-		`MatchingGetTasksBatchSize is the maximum batch size to fetch from the task buffer`,
+		`How many backlog tasks to read from persistence at once`,
+	)
+	MatchingGetTasksReloadAt = NewTaskQueueIntSetting(
```
stephanos, line 1061: Should we also annotate this as "priority/new matcher"-only like the other one?

service/matching/pri_task_reader.go
```diff
@@ -170,7 +165,7 @@ func (tr *priTaskReader) completeTask(task *internalTask, res taskResponse) {
 
 	// use == so we just signal once when we cross this threshold
 	// TODO(pri): is this safe? maybe we need to improve this
-	if tr.loadedTasks == tr.backlogMgr.config.GetTasksBatchSize()/reloadFraction {
+	if tr.loadedTasks == tr.backlogMgr.config.GetTasksReloadAt() {
 		tr.Signal()
```
stephanos, line 169: On a completely unrelated note; do you think there's a better function name? While I now know what it means, `Signal` is quite vague. Since we're using `wake` for task matching, that would be an option (although it's still a bit vague). Or maybe more explicit such as `RequestLoadingTasks`.

common/dynamicconfig/constants.go
```diff
@@ -1056,7 +1056,12 @@ See DynamicRateLimitingParams comments for more details.`,
...
+		"matching.getTasksReloadAt",
+		100,
+		`Reload a batch of tasks when there are this many remaining. Must be less than MatchingGetTasksBatchSize.`,
```
stephanos, line 1064: Is my understanding correct that if `MatchingGetTasksBatchSize + MatchingGetTasksReloadAt >= MatchingGetTasksBatchSize` we would never load tasks from the backlog? I'm wondering if there's a trapdoor here for operators (like ourselves). Since this can be changed on the fly, a hard fail on startup isn't an option. But we could consider capping the value to `MatchingGetTasksBatchSize`?

common/dynamicconfig/constants.go
dnr: yeah

common/dynamicconfig/constants.go
dnr: Did you miss a 2× in that comparison?

Yeah, it's possible to set this so that it breaks things. But that's possible for a lot of dynamic configs. I'd like to create a general validation mechanism eventually, but that'll need some design work.

For this one, I'm thinking of it as almost temporary, ideally we'd have some code that tracks the dispatch rate and reloads based on time left (e.g. 1s before we're about to run out) rather than tasks left.

service/matching/pri_task_reader.go
dnr: Signal works for me, but I've read it a bunch so :shrug:  I wonder if that name is too specific, it's more like "wake up and apply some logic to figure out if you should load more tasks"

stephanos: `SignalTaskLoading`?

stephanos (approved):
The two suggestions are optional.

common/dynamicconfig/constants.go
dnr: Then we'd have to read two configs instead of one each time we check it?  I think the logic can be improved (e.g. if the config is changed it might miss the trigger), but I'd like to do it in another PR

PR #7430, opened 2025-03-05
Small improvements for new matcher

dnr:
## What changed?
- Don't call `allowForwarding` on a root partition
- Use absolute times for backlog head age so we only call `time.Since` once
- Return a slice of `taskqueuespb.InternalTaskQueueStatus` for subqueues

## Why?
Small optimizations, more detailed internal info

## How did you test it?
updated tests


proto/internal/temporal/server/api/taskqueue/v1/message.proto
```diff
@@ -77,7 +79,7 @@ message PhysicalTaskQueueInfo {
     // Unversioned workers (with `useVersioning=false`) are reported in unversioned result even if they set a Build ID.
     repeated temporal.api.taskqueue.v1.PollerInfo pollers = 1;
     temporal.api.taskqueue.v1.TaskQueueStats task_queue_stats = 2;
-    InternalTaskQueueStatus internal_task_queue_status = 3;
+    repeated InternalTaskQueueStatus internal_task_queue_status = 3;
```
stephanos, line 82: Similarly, doesn't this break for existing callers?

service/matching/backlog_manager.go
```diff
@@ -220,6 +218,21 @@ func (c *backlogManagerImpl) TotalApproximateBacklogCount() int64 {
...
 }
 
+func (c *backlogManagerImpl) InternalStatus() []*taskqueuespb.InternalTaskQueueStatus {
+	return []*taskqueuespb.InternalTaskQueueStatus{
+		&taskqueuespb.InternalTaskQueueStatus{
+			ReadLevel: c.taskAckManager.getReadLevel(),
+			AckLevel:  c.taskAckManager.getAckLevel(),
+			TaskIdBlock: &taskqueuepb.TaskIdBlock{
+				StartId: c.taskWriter.taskIDBlock.start,
+				EndId:   c.taskWriter.taskIDBlock.end,
```
stephanos, line 228: Would the same race condition exist here?

service/matching/pri_backlog_manager.go
```diff
@@ -311,6 +322,31 @@ func (c *priBacklogManagerImpl) TotalApproximateBacklogCount() int64 {
 	return c.db.getTotalApproximateBacklogCount()
 }
 
+func (c *priBacklogManagerImpl) InternalStatus() []*taskqueuespb.InternalTaskQueueStatus {
+	// TODO(pri): this is a data race, it should only be read by taskWriterLoop
+	idBlock := &taskqueuepb.TaskIdBlock{
+		StartId: c.taskWriter.taskIDBlock.start,
+		EndId:   c.taskWriter.taskIDBlock.end,
+	}
```
stephanos, line 330: Do you have a plan on how to address the data race?

service/matching/backlog_manager.go
```diff
@@ -220,6 +218,21 @@ func (c *backlogManagerImpl) TotalApproximateBacklogCount() int64 {
...
+			},
+			LoadedTasks:  c.taskAckManager.getBacklogCountHint(),
```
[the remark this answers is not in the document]
stephanos: Ohhhh, okay. Could we add your explanation somewhere in the code (abridged)? If it's already there, it's very well hidden. I now see there's a helpful note on the proto definition; but I never made it that far down.

proto/internal/temporal/server/api/taskqueue/v1/message.proto
stephanos: Oh, interesting. I didn't know that about protos.

stephanos (approved):

proto/internal/temporal/server/api/taskqueue/v1/message.proto
dnr: It looks like buf breaking complains about it, but it's too picky, we can ignore it

service/matching/backlog_manager.go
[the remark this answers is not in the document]
dnr: I added a comment to the backlogManager interface. I don't think there should be an interface there after cleanup, but we can move the comment at that point. Anywhere else for now?

PR #7431, opened 2025-03-05
Avoid unnecessary task queue metadata writes

dnr:
## What changed?
- Track last changed time and last written time in taskQueueDB.
- Skip write in SyncState when unchanged, unless it's been over 12 hours. Note this only applies to the new backlog manager.

## Why?
Avoid unnecessary persistence operations, improve scalability

## How did you test it?
will be tested in load tests for new matcher/backlog manager

### DIFF
--- a/service/matching/db.go
+++ b/service/matching/db.go
@@ -60,6 +60,10 @@ type (
 		store     persistence.TaskManager
 		logger    log.Logger
 		subqueues []*dbSubqueue
+
+		// used to avoid unnecessary metadata writes:
+		lastChange time.Time // updated when metadata is changed in memory
+		lastWrite  time.Time // udpated when metadata is successfully written to db
 	}
 
 	dbSubqueue struct {
@@ -146,7 +150,7 @@ func (db *taskQueueDB) RenewLease(
 			return taskQueueState{}, err
 		}
 	} else {
-		if err := db.renewTaskQueueLocked(ctx, db.rangeID+1); err != nil {
+		if err := db.updateTaskQueueLocked(ctx, true); err != nil {
 			return taskQueueState{}, err
 		}
 	}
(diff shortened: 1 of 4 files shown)


service/matching/db.go
```diff
@@ -60,6 +60,10 @@ type (
 		store     persistence.TaskManager
 		logger    log.Logger
 		subqueues []*dbSubqueue
+
+		// used to avoid unnecessary metadata writes:
+		lastChange time.Time // updated when metadata is changed in memory
+		lastWrite  time.Time // udpated when metadata is successfully written to db
```
stephanos, line 66:
```suggestion
		lastWrite  time.Time // updated when metadata is successfully written to db
```

service/matching/db.go
```diff
@@ -167,54 +171,61 @@ func (db *taskQueueDB) takeOverTaskQueueLocked(
...
-) error {
+func (db *taskQueueDB) updateTaskQueueLocked(ctx context.Context, incrementRangeId bool) error {
+	newRangeID := db.rangeID
+	if incrementRangeId {
+		newRangeID++
+	}
 	if _, err := db.store.UpdateTaskQueue(ctx, &persistence.UpdateTaskQueueRequest{
-		RangeID:       rangeID,
+		RangeID:       newRangeID,
 		TaskQueueInfo: db.cachedQueueInfo(),
```
stephanos, line 222: Q: What part of `cachedQueueInfo` is "cached"? Or does that mean "local" / "previously read"?

service/matching/db.go
```diff
@@ -167,54 +171,61 @@ func (db *taskQueueDB) takeOverTaskQueueLocked(
...
 	default:
 		return err
 	}
 }
 
-func (db *taskQueueDB) renewTaskQueueLocked(
-	ctx context.Context,
-	rangeID int64,
-) error {
+func (db *taskQueueDB) updateTaskQueueLocked(ctx context.Context, incrementRangeId bool) error {
```
stephanos, line 215: Could we find name for `incrementRangeId` that is less technically descriptive and more explanatory / domain-specific? Something that describes the intent.

dnr: I'm not sure what it would be... we need to increment the range id for two distinct reasons: to take over from another owner, and when we hit the end of the range and need to move to a new range. But the action for both is the same.

We could do something like
```
const updateSameRangeId = 0
const updateTakeoverRangeId = 1
const updateNextRangeId = 1
```
and pass one of those constants, then do `newRangeID := db.RangeID + updateIncrement`?

service/matching/db.go
dnr: Think write-back cache, not write-through/read-through. Updated in memory but maybe not persisted yet.

service/matching/db.go
stephanos: I was thinking about this for a bit now and the best I came up with is `updateInPlace bool` (where `true` means don't increment range ID). But honestly, I think I'm trying to lift this part to a higher abstraction than it allows. Since `rangeID` is ubiquitous and not an implementation detail; let's leave it this way.

stephanos (approved):

PR #7432, opened 2025-03-06
Port previous NewTestLogger flags over

stephanos:
## What changed?

Ports `TEMPORAL_TEST_LOG_FORMAT` and `TEMPORAL_TEST_LOG_LEVEL` to testlogger.TestLogger.

## Why?

To not break developer setups.

### DIFF
--- a/common/log/zap_logger.go
+++ b/common/log/zap_logger.go
@@ -38,9 +38,10 @@ import (
 const (
 	skipForZapLogger = 3
 	// we put a default message when it is empty so that the log can be searchable/filterable
-	defaultMsgForEmpty  = "none"
-	testLogFormatEnvVar = "TEMPORAL_TEST_LOG_FORMAT" // set to "json" for json logs in tests
-	testLogLevelEnvVar  = "TEMPORAL_TEST_LOG_LEVEL"  // set to "debug" for debug level logs in tests
+	defaultMsgForEmpty = "none"
+	// TODO: once `NewTestLogger` has been removed, move these vars into testlogger.TestLogger
+	TestLogFormatEnvVar = "TEMPORAL_TEST_LOG_FORMAT" // set to "json" for json logs in tests
+	TestLogLevelEnvVar  = "TEMPORAL_TEST_LOG_LEVEL"  // set to "debug" for debug level logs in tests
 )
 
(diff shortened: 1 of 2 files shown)


common/testing/testlogger/testlogger.go
```diff
@@ -204,18 +204,6 @@ func LogLevel(level zapcore.Level) LoggerOption {
...
-		t.state.logCaller = false
-	}
-}
-
-func LogCaller() LoggerOption {
-	return func(t *TestLogger) {
-		t.state.logCaller = false
-	}
-}
-
```
stephanos, line 218: This doesn't work together with  `log.BuildZapLogger` which defaults to disabling this. Removing it for now.

common/log/zap_logger.go
```diff
@@ -55,13 +56,13 @@ var _ Logger = (*zapLogger)(nil)
 
 // NewTestLogger returns a logger for tests
 func NewTestLogger() *zapLogger {
```
stephanos, line 58: There's a ticket to remove this function and replace it with testlogger.TestLogger.

common/log/zap_logger.go
```diff
@@ -54,14 +55,15 @@ type (
...
+	format := os.Getenv(TestLogFormatEnvVar)
 	if format == "" {
 		format = "console"
 	}
 
 	logger := BuildZapLogger(Config{
-		Level:       os.Getenv(testLogLevelEnvVar),
+		Level:       os.Getenv(TestLogLevelEnvVar),
 		Format:      format,
 		Development: true,
```
dnr, line 68: should we port this too? it looks like it controls whether it panics on dpanic

alexshtin: Why do we need envs to control test logger? Do we use it anywhere?

stephanos: David does to switch it to JSON format; and I noticed actually my IDE is set up to log on error level instead of debug level.

alexshtin (approved):

common/log/zap_logger.go
alexshtin: I didn't know we have it.

dnr (approved):

PR #7433, opened 2025-03-06
Use assertion function

stephanos:

### DIFF
--- a/service/matching/matcher_data.go
+++ b/service/matching/matcher_data.go
@@ -234,7 +234,7 @@ func (t *taskPQ) ForEachTask(pred func(*internalTask) bool, post func(*internalT
 
 type matcherData struct {
 	config     *taskQueueConfig
-	logger log.Logger
+	logger     log.Logger
 	canForward bool
 
 	lock sync.Mutex // covers everything below, and all fields in any waitableMatchResult
@@ -252,8 +252,8 @@ type matcherData struct {
 
 func newMatcherData(config *taskQueueConfig, logger log.Logger, canForward bool) matcherData {
 	return matcherData{
-		config: config,
-		logger: logger,
+		config:     config,
+		logger:     logger,
 		canForward: canForward,
 		tasks: taskPQ{
 			ages: newBacklogAgeTracker(),
--- a/service/matching/matcher_data.go
+++ b/service/matching/matcher_data.go
@@ -30,6 +30,8 @@ import (
 	"time"
 
 	enumsspb "go.temporal.io/server/api/enums/v1"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/softassert"
 	"go.temporal.io/server/common/util"
 )
 
@@ -232,6 +234,7 @@ func (t *taskPQ) ForEachTask(pred func(*internalTask) bool, post func(*internalT
 
 type matcherData struct {
 	config     *taskQueueConfig
+	logger log.Logger
 	canForward bool
 
 	lock sync.Mutex // covers everything below, and all fields in any waitableMatchResult
@@ -247,9 +250,10 @@ type matcherData struct {
 	lastPoller time.Time // most recent poll start time
 }
 
-func newMatcherData(config *taskQueueConfig, canForward bool) matcherData {
+func newMatcherData(config *taskQueueConfig, logger log.Logger, canForward bool) matcherData {
 	return matcherData{
 		config:     config,
+		logger: logger,
 		canForward: canForward,
 		tasks: taskPQ{
 			ages: newBacklogAgeTracker(),
@@ -295,7 +299,7 @@ func (d *matcherData) EnqueueTaskAndWait(ctxs []context.Context, task *internalT
 
 			if task.matchResult == nil {
 				d.tasks.Remove(task)
-				task.wake(&matchResult{ctxErr: ctx.Err(), ctxErrIdx: i})
+				task.wake(d.logger, &matchResult{ctxErr: ctx.Err(), ctxErrIdx: i})
 			}
 		})
 		defer stop() // nolint:revive // there's only ever a small number of contexts
@@ -343,7 +347,7 @@ func (d *matcherData) EnqueuePollerAndWait(ctxs []context.Context, poller *waiti
 				if poller.matchHeapIndex >= 0 {
 					d.pollers.Remove(poller)
 				}
-				poller.wake(&matchResult{ctxErr: ctx.Err(), ctxErrIdx: i})
+				poller.wake(d.logger, &matchResult{ctxErr: ctx.Err(), ctxErrIdx: i})
 			}
 		})
 		defer stop() // nolint:revive // there's only ever a small number of contexts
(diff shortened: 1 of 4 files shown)


service/matching/pri_matcher.go
```diff
@@ -178,7 +183,7 @@ func (tm *priTaskMatcher) forwardTasks(lim quotas.RateLimiter, retrier backoff.R
 		if res.ctxErr != nil {
 			return // task queue closing
 		}
-		bugIf(res.task == nil, "bug: bad match result in forwardTasks")
+		softassert.That(tm.logger, res.task != nil, "expected a task from match")
```
dnr, line 186: this should continue or return on failure, otherwise it'll just panic immediately

service/matching/pri_matcher.go
```diff
@@ -247,11 +252,13 @@ func (tm *priTaskMatcher) validateTasksOnRoot(lim quotas.RateLimiter, retrier ba
 		if res.ctxErr != nil {
 			return // task queue closing
 		}
-		bugIf(res.task == nil, "bug: bad match result in validateTasksOnRoot")
+		softassert.That(tm.logger, res.task != nil, "expected a task from match")
```
dnr, line 255: this too

service/matching/pri_matcher.go
```diff
@@ -276,7 +283,7 @@ func (tm *priTaskMatcher) forwardPolls() {
 		if res.ctxErr != nil {
 			return // task queue closing
 		}
-		bugIf(res.poller == nil, "bug: bad match result in forwardPolls")
+		softassert.That(tm.logger, res.poller != nil, "expected a poller from match")
```
dnr, line 286: needs to continue or return

service/matching/pri_matcher.go
```diff
@@ -324,7 +331,7 @@ func (tm *priTaskMatcher) forwardPolls() {
 func (tm *priTaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, error) {
 	finish := func() (bool, error) {
 		res, ok := task.getResponse()
-		bugIf(!ok, "Offer must be given a sync match task")
+		softassert.That(tm.logger, ok, "expected a sync match task")
```
dnr, line 334: needs to return error?

service/matching/pri_task_reader.go
```diff
@@ -410,7 +414,7 @@ func (tr *priTaskReader) signalNewTasks(resp subqueueCreateTasksResponse) {
 	// adding these tasks to outstandingTasks. So they should definitely not be there.
 	for _, t := range resp.tasks {
 		_, found := tr.outstandingTasks.Get(t.TaskId)
-		bugIf(found, "bug: newly-written task already present in outstanding tasks")
+		softassert.That(tr.logger, !found, "newly-written task already present in outstanding tasks")
```
dnr, line 417: this should probably do a slices.DeleteFunc and remove the ones that are found

service/matching/pri_task_reader.go
```diff
@@ -445,8 +449,8 @@ func (tr *priTaskReader) getLoadedTasks() int {
 
 func (tr *priTaskReader) ackTaskLocked(taskId int64) int64 {
 	wasAlreadyAcked, found := tr.outstandingTasks.Get(taskId)
-	bugIf(!found, "bug: completed task not found in outstandingTasks")
-	bugIf(wasAlreadyAcked.(bool), "bug: completed task was already acked")
+	softassert.That(tr.logger, found, "completed task not found in oustandingTasks")
+	softassert.That(tr.logger, !wasAlreadyAcked.(bool), "completed task was already acked")
```
dnr, line 453: probably just return if either of these are true?

service/matching/pri_matcher.go
stephanos: Seems like `return false, nil` has the same effect as an error.

service/matching/matcher_data.go
```diff
@@ -552,9 +556,9 @@ func (w *waitableMatchResult) initMatch(d *matcherData) {
 // call with matcherData.lock held.
 // w.matchResult must be nil (can't call wake twice).
 // w must not be in queues anymore.
-func (w *waitableMatchResult) wake(res *matchResult) {
-	bugIf(w.matchResult != nil, "bug: wake called twice")
-	bugIf(w.matchHeapIndex >= 0, "bug: wake called but still in heap")
+func (w *waitableMatchResult) wake(logger log.Logger, res *matchResult) {
+	softassert.That(logger, w.matchResult == nil, "wake called twice")
+	softassert.That(logger, w.matchHeapIndex < 0, "wake called but still in heap")
```
stephanos, line 561: Only ones left without a mitigation; not sure what that would be?

service/matching/matcher_data.go
```diff
@@ -246,12 +249,13 @@ type matcherData struct {
...
 }
 
-func newMatcherData(config *taskQueueConfig) matcherData {
+func newMatcherData(config *taskQueueConfig, logger log.Logger) matcherData {
 	return matcherData{
 		config: config,
 		tasks: taskPQ{
 			ages: newBacklogAgeTracker(),
 		},
+		logger: logger,
```
dnr, line 258: move to line 255 to match struct order

service/matching/matcher_data.go
dnr: I think it's fine to just do nothing and let it wake again

service/matching/pri_matcher.go
```diff
@@ -403,9 +423,13 @@ again:
 		}
 		return nil, res.ctxErr
 	}
-	bugIf(res.poller == nil, "bug: bad match result in syncOfferTask")
+	if !softassert.That(tm.logger, res.poller != nil, "expected poller from match") {
+		return nil, errors.New("internal error: no poller found on sync match")
```
dnr, line 427: should we just define a `errInternalMatchError = serviceerror.NewInternal("internal matcher error")` and use that?
just a thought, either way is fine

service/matching/pri_task_reader.go
```diff
@@ -408,10 +412,10 @@ func (tr *priTaskReader) signalNewTasks(resp subqueueCreateTasksResponse) {
 
 	// Because we checked readLevel above, we know that getTasksPump can't have beat us to
 	// adding these tasks to outstandingTasks. So they should definitely not be there.
-	for _, t := range resp.tasks {
+	resp.tasks = slices.DeleteFunc(resp.tasks, func(t *persistencespb.AllocatedTaskInfo) bool {
 		_, found := tr.outstandingTasks.Get(t.TaskId)
-		bugIf(found, "bug: newly-written task already present in outstanding tasks")
-	}
+		return !softassert.That(tr.logger, !found, "newly-written task already present in outstanding tasks")
+	})
```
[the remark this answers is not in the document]
dnr: I'm kinda thinking that if we do hit this, we should `tr.lock.Unlock(); tr.Signal(); return`, i.e. abort the optimized path and just signal the reader. It seems weird to trim down the slice (of course this should definitely never happen...)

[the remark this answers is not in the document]
stephanos: I can move the check between `canAddDirect :=` and `if !canAddDirect {`; and set `canAddDirect = false` if we fail the assert?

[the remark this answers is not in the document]
dnr: yeah, that's nice. it should be before setting readLevel

dnr (approved):

PR #7436, opened 2025-03-06
More Versioning flake fixes

Shivs11:
## What changed?

Fixing the following tests:

- TestDescribeTaskQueueVersioningInfo
- TestSetCurrentVersion_ConcurrentUpdates_NonIdempotentRequests (removed, see discussion for reason)
- TestSetRampingVersion_ConcurrentUpdates_NonIdempotentRequests (removed, see discussion for reason)
- TestSetRampingVersion_ConcurrentUpdates_IdempotentRequests (removed, see discussion for reason)
- TestSetCurrentVersion_ConcurrentUpdates_IdempotentRequests (removed, see discussion for reason)

## Why?

- To hopefully not see them in the flaky test reports.

## How did you test it?

- Ran all of them in CI 50 times with different DB's and all passed.
- Ran locally as well.


tests/worker_deployment_test.go
```diff
@@ -451,7 +451,7 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_ConcurrentUpdates_NonIdemp
 
 	// To allow the first go-routine to start before the second one.
 	//nolint:forbidigo
-	time.Sleep(1 * time.Millisecond)
+	time.Sleep(10 * time.Millisecond)
```
Shivs11, line 454: The errors seen in the WorkerDeploymentSuite were mostly because the second go-routine's updates happened before the first one. I was a little surprised that this was possible given there was a 1 millisecond time-gap between the two go-routines.

However, the time gap has been increased and I hope to see the errors resolved. Moreover, have also run them in the CI and they have passed.

stephanos: Is the 1st one `go s.pollFromDeployment(ctx, tv)` and the second one `go func() {` on line 441?

The Go scheduler doesn't necessarily pick up a goroutine immediately. It will park them; and once it decides to pause the current goroutine, it picks a random one that is unblocked to continue. So you can never count on any ordering there.

Furthermore, increasing the sleep is not a sure way to keep this non-flaky. You need some kind of signal/wait condition to know you're good to continue. Anything based on sleeps is either brittle or slow, unfortunately.

Do you think you can replace the Sleep?

tests/worker_deployment_test.go
```diff
@@ -621,7 +621,7 @@ func (s *WorkerDeploymentSuite) TestSetRampingVersion_ConcurrentUpdates_NonIdemp
 			Version:        tv.DeploymentVersionString(),
 			Percentage:     5,
 			ConflictToken:  cT,
-			Identity:       tv.Any().String(), // note: different identity
+			Identity:       tv.Any().String(), // note: different identity making this request different from the first one.
```
stephanos, line 624: Is it necessary for it to be a different identity?

If it is, I'd recommend `tv.ClientIdentity() + "-OtherClient"` so you preserve the original client ID (since it contains meta info that points to this test; which can be helpful when debugging).

Shivs11: > Is it necessary for it to be a different identity

yes! The idea for this specific test case was to have two concurrent requests come in which had different identities but the same conflict token - I wanted to check for the case of a request having a stale conflict token doesn't go ahead and change the state of the entity workflow - The different identity helps the test verify if the stale request changed state or not

tests/worker_deployment_test.go
Shivs11: Thinking about it a bit more, this doesn't seem to be as easy as I first thought it could be - The aim of this function is to send in two update requests such that both *pass* the update validator's checks but the second request gets caught in the update handler's validator check. Because of this constraint, I can't send in these requests sequentially.

What I desire is the following:

The first go-routine's request passes the validator, enters the handlers and gives up control. Then, the second go-routine passes the validator (it will since the first request is not fully complete and has not changed the state), and then gives up control. However, the second go-routine's request shall never complete successfully since the control goes back to the first go-routine and it changes state. On switching control back to the second go-routine, it conducts a state check again and realizes it can't proceed since the state has changed.

From what you just mentioned, I gauge that having a definitive ordering guarantee between the two go-routines may not be possible.

Shivs11: resolution: thought about this for quite a bit and realized there is no way around this. Made a decision to remove sleep and rely on a deterministic ordering for my test cases. Instead, I now check if either one of them completed which shall hopefully resolve the flakes.

Thank you @stephanos for helping me learn this!

Shivs11: agreed and thanks! 
removed the tests which were serving the purpose of testing that specific behaviour you just pointed out.

stephanos (approved):

PR #7437, opened 2025-03-06
Add workflow rules

ychebotarev:
## What changed?

Add implementation to the following API:
* CreateWorkflowRule
* DescribeWorkflowRule
* DeleteWorkflowRule
* ListWorkflowRules

Add initial rule implementation - activity will be paused on activity Retry.
Add functional test for that.

All functions are behind feature flag.

## Why?

Part of workflow rules work.

## How did you test it?

Add unit test.
Add func test.

## Potential risks

N/A


common/namespace/namespace.go
```diff
@@ -299,6 +301,13 @@ func (ns *Namespace) CustomSearchAttributesMapper() CustomSearchAttributesMapper
 	return ns.customSearchAttributesMapper
 }
 
+func (ns *Namespace) GetWorkflowRules() []*rulespb.WorkflowRule {
+	if ns.config.WorkflowRules == nil {
+		return []*rulespb.WorkflowRule{}
```
alexshtin, line 306: Why not `nil`? This is extra allocation, and empty slices and nil slices should be treated in same way.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -41,6 +41,7 @@ import "temporal/api/protocol/v1/message.proto";
 import "temporal/api/failure/v1/message.proto";
 import "temporal/api/nexus/v1/message.proto";
 
+
```
alexshtin, line 44:
```suggestion
```

service/frontend/namespace_handler.go
alexshtin, line 1: I always wanted to remove `namespace_handler` or at least change its name and interfaces. `handler` is suffix reserved for gRPC handlers, I wouldn't add new code to it but add it directly to `workflow_handler.go`.

service/frontend/namespace_handler.go
```diff
@@ -684,6 +685,126 @@ func (d *namespaceHandler) DeprecateNamespace(
...
+		return nil, err
+	}
+	getResponse, err := d.metadataMgr.GetNamespace(ctx, &persistence.GetNamespaceRequest{Name: nsName})
+	if err != nil {
+		return nil, err
+	}
+
+	existingNamespace := getResponse.Namespace
+
+	config := existingNamespace.Config
```
alexshtin, line 707: I personally don't like such variables. `getResponse` (btw, it is a very bad name) is still accessible and someone will use one day. If you just rename `getResponse` to `existing` you will get the same English but w/o extra vars.
I agree, it is matter of personal preferences though.

service/frontend/namespace_handler.go
alexshtin: Here you have part of validation in `workflow_handler.go` and another part here. It adds more confusion than help.

service/frontend/workflow_handler.go
```diff
@@ -5873,3 +5873,110 @@ func (wh *WorkflowHandler) ResetActivity(
...
 }
+
+func (wh *WorkflowHandler) CreateWorkflowRule(
+	ctx context.Context,
+	request *workflowservice.CreateWorkflowRuleRequest,
+) (_ *workflowservice.CreateWorkflowRuleResponse, retError error) {
+	defer log.CapturePanic(wh.logger, &retError)
+
+	if !wh.config.WorkflowRulesAPIsEnabled(request.GetNamespace()) {
+		return nil, status.Errorf(codes.Unimplemented, "method CreateWorkflowRule not supported")
```
alexshtin, line 5884: I understand you copied it from somewhere else but strong believe is that it should be:
```suggestion
		return nil, serviceerror.NewUnimplemented("method CreateWorkflowRule not supported")
```
And I have a reason for it.

service/history/workflow/matcher/activity_evaluator.go
```diff
@@ -241,7 +241,15 @@ func (m *activityMatchEvaluator) compareActivityStatus(status string, operator s
...
-	activityState := workflow.GetActivityState(m.ai)
+
+	activityState := enumspb.PENDING_ACTIVITY_STATE_SCHEDULED
+	if m.ai.CancelRequested {
+		activityState = enumspb.PENDING_ACTIVITY_STATE_CANCEL_REQUESTED
+	}
+	if m.ai.StartedEventId != common.EmptyEventID {
+		activityState = enumspb.PENDING_ACTIVITY_STATE_STARTED
+	}
+
```
alexshtin, line 252: Inline function?

service/history/workflow/matcher/activity_matcher.go
```diff
@@ -58,3 +58,31 @@ func MatchActivity(ai *persistencespb.ActivityInfo, query string) (bool, error)
 	evaluator := newActivityMatchEvaluator(ai)
 	return evaluator.Evaluate(query)
 }
+
+/*
+func MatchWorfklowRules(ai *persistencespb.ActivityInfo, namespaceEntry *namespace.Namespace) (bool, error) {
```
alexshtin, line 63: We generally don't check-in commented code. If you want to do it, you need to add good comment (comment for comment) on why it is commented out and when it should be uncommented. But this seems just some left over.

service/history/workflow/matcher/mutable_state_evaluator.go
```diff
@@ -41,11 +41,17 @@ const (
 )
 
 type mutableStateMatchEvaluator struct {
-	ms workflow.MutableState
+	executionInfo  *persistencespb.WorkflowExecutionInfo
+	executionState *persistencespb.WorkflowExecutionState
```
alexshtin, line 45: :+1:

alexshtin: This can be separate PR, btw.

tests/activity_api_rules_test.go
```diff
@@ -0,0 +1,260 @@
...
+	s.NotNil(createRuleResponse)
+
+	// verify that frontend has updated namespaces
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		nsResp, err := s.FrontendClient().ListWorkflowRules(ctx, &workflowservice.ListWorkflowRulesRequest{
+			Namespace: s.Namespace().String(),
+		})
+		s.NoError(err)
+		s.NotNil(nsResp)
+		s.NotNil(nsResp.Rules)
```
alexshtin, line 212: You can't use `s` here :-(. And remember to check for `assert.NoError` return value.

tests/activity_api_rules_test.go
```diff
@@ -0,0 +1,260 @@
...
+			s.Equal(ruleID, nsResp.Rules[0].Spec.Id)
+		}
+	}, 5*time.Second, 200*time.Millisecond)
+
+	// wait for activity to be paused by rule
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		assert.NoError(t, err)
+		if description.GetPendingActivities() != nil {
+			assert.Len(t, description.PendingActivities, 1)
```
alexshtin, line 224: Like here. I needs to be:
```suggestion
			if(assert.Len(t, description.PendingActivities, 1)) { ...
```

tests/activity_api_rules_test.go
```diff
@@ -0,0 +1,260 @@
...
+		assert.Less(t, int32(1), testWorkflow.startedActivityCount.Load())
+	}, 5*time.Second, 200*time.Millisecond)
+
+	// unblock the activity and let the workflow finish
+	testWorkflow.letActivitySucceed.Store(true)
+	testWorkflow.activityCompleteCn <- struct{}{}
+
+	// wait for workflow to finish
+	var out string
+	err = workflowRun.Get(ctx, &out)
```
alexshtin, line 259: I agree with linter :-)

tests/activity_api_rules_test.go
```diff
@@ -0,0 +1,260 @@
...
+}
```
alexshtin, line 260: I think you need more tests here. Concurrent ns updates (changing other properties), more advanced scenarios...

alexshtin (commented):
Looks good to me. Ready to approve after comments are addressed.

service/frontend/workflow_handler.go
ychebotarev: which is? Because in this case I will need to change it in other places as well.

alexshtin: What will be the error on the caller side if caller is using `CreateRule` function from SDK? It will be `serviceerror.NotImplemented`. The idea of `serviceserror` is that it hides gRPC statuses from our code and allows caller to handle exact same error as returned from handler. This is one of them. General consistency with other errors even in this func, is anther.

common/api/metadata.go
```diff
@@ -158,6 +158,10 @@ var (
 		"DeleteWorkerDeploymentVersion":         {Scope: ScopeNamespace, Access: AccessWrite},
 		"UpdateWorkerDeploymentVersionMetadata": {Scope: ScopeNamespace, Access: AccessWrite},
 		"ListWorkerDeployments":                 {Scope: ScopeNamespace, Access: AccessReadOnly},
+		"CreateWorkflowRule":                    {Scope: ScopeNamespace, Access: AccessReadOnly},
+		"DescribeWorkflowRule":                  {Scope: ScopeNamespace, Access: AccessReadOnly},
+		"DeleteWorkflowRule":                    {Scope: ScopeNamespace, Access: AccessReadOnly},
+		"ListWorkflowRules":                     {Scope: ScopeNamespace, Access: AccessReadOnly},
```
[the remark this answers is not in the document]
ychebotarev: add write access

service/frontend/namespace_handler.go
ychebotarev: ok. I think I copy/paster it from somewhere, getResponse was a weird name.

service/history/workflow/matcher/activity_evaluator.go
ychebotarev: moved to a helper function. I didn't remove it, why do you think I am?

service/history/workflow/matcher/activity_matcher.go
ychebotarev: deleted

tests/activity_api_rules_test.go
ychebotarev: replaced with assert.XXX
no need to check return value.

tests/activity_api_rules_test.go
ychebotarev: I add test for API. For implementation they will come later.

alexshtin (approved):
I still don't agree with `namespace_handler.go` separation, but I won't fight for it.

PR #7439, opened 2025-03-06
Suppress OTEL export errors during startup/shutdown

stephanos:
## What changed?

Disabling OTEL error logging before startup _and_ after shutdown was initiated. 

## Why?

Those logs are just noise.

## How did you test it?

Run before/after fix and saw these gone:

```
{"level":"warn","ts":"2025-03-06T11:44:31.484-0800","msg":"OTEL error","error":"traces export: the client is shutdown","service-error-type":"errors.errorString","logging-call-at":"/Users/stephan/Workspace/default/temporal/temporal/fx.go:877"}
```

### DIFF
--- a/temporal/fx.go
+++ b/temporal/fx.go
@@ -30,6 +30,7 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"sync/atomic"
 
 	"github.com/pborman/uuid"
 	"go.opentelemetry.io/otel"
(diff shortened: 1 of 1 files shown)


temporal/fx.go
```diff
@@ -865,18 +866,20 @@ type SpanExporterInputs struct {
 	Config     *config.Config `optional:"true"`
 }
 
+var tracingReady atomic.Bool
```
stephanos, line 869: I tried a few things and settled on this approach.

I tried registering it OnStart, but then the logs are still written using the default logger.
I tried de-registering it on OnStop, but then it was still happening at startup.

alexshtin: Global var in almost root module :man_facepalming:. At least it is not exported.

alexshtin (approved):

temporal/fx.go
stephanos: Oh, your comment made me realize that our functional tests actually run multiple fx instances!
Is there an fx context or sth I can attach it to?

stephanos:
Okay, I found a way to avoid the global var. I think.

alexshtin (approved):
this is better!

PR #7442, opened 2025-03-07
Reload task queues on change to new matcher setting

dnr:
## What changed?
When `MatchingUseNewMatcher` changes, unload task queue (and let it get reloaded).

## Why?
Easily enable/disable new matcher without manually reloading task queues.

## How did you test it?
New functional test to flip back and forth, manually verified reload was happening.

### DIFF
--- a/tests/priority_fairness_test.go
+++ b/tests/priority_fairness_test.go
@@ -25,6 +25,7 @@ package tests
 import (
 	"context"
 	"math/rand"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


tests/priority_fairness_test.go
```diff
@@ -128,6 +129,82 @@ func (s *PriorityFairnessSuite) TestPriority_Activity_Basic() {
...
+
+	s.T().Log("switching back to old matcher")
+	s.OverrideDynamicConfig(dynamicconfig.MatchingUseNewMatcher, false)
+
+	s.T().Log("unblocking activities")
+	unblockActivities <- struct{}{}
+
+	s.T().Log("waiting for last 100 activites")
+	s.Eventually(func() bool { return activitiesCompleted.Load() == 300 }, 10*time.Second, 10*time.Millisecond)
+}
```
stephanos, line 206: What do you think about adding a `matcher_version` tag to all metrics that are emitted by the old and new components? That way we could (a) monitor those metrics by version in prod and also (b) use them here to verify that both versions were live (the test would have prob worked before the change, too?).

stephanos (approved):

tests/priority_fairness_test.go
dnr: I'm a little wary of adding tags that multiply cardinality. Maybe instead we should add a few new metrics for some specific things we want to be aware of (e.g. reader bypass)?

Yes, this test would pass before and after this change, it wasn't intended to test this change specifically, it was something I wanted to test anyway, and this change made it easy so I did it now (and I could manually verify that it was reloading).

I was thinking about an extended version of this test that would use subqueues. That would give a clearer signal about what was happening.

stephanos: +1 to adding net new metrics for new behavior.

+1 to future version with subqueues

PR #7443, opened 2025-03-07
Refactoring. Split interface/implementation for Shard/Workflow Context

ychebotarev:
## What changed?

Main changes:
* move shard.Context to interfaces (and rename it to ShardContest). 
* move workflow.Context to interfaces (and rename it to WorkflowContext). 
* Many small improvements, fixing linter, etc.
This effectively splits interface/implementation for both of those contexts.
Also  again remove LOTS of dependencies. More to come.

## Why?

[Previous PR](https://github.com/temporalio/temporal/pull/7440)

Part of bigger refactoring efforts.
Goal is to reduce code entanglement, by separating interfaces and implementation, for some major components.
Currently, in lots of cases, interface and implementation are in a single file. This bloats the dependency graph, and introduce bad practices when dealing with "module cyclic dependency" error.

Nearest goal will be to split/move the following interfaces:
* Engine - done
* MutalbleState - done
* Shard Context - done
* Workflow Context

Once moving interfaces is complete:

add "GetOrCreateContext" to shard context. It will receive cache as a parameter
change s.workflowCache.GetOrCreateWorkflowExecution(s.hardContext, ...) to s.hardContext.GetOrCreateContext(...)
add "cache" to shard context ->IMPLEMENTATION<- (this is important - there is (I hope) no need to actually expose workflow cache from shard context)
remove cache as a parameter from GetOrCreateContext
remove workflwCacha wcache from all those structures/calles/etc, where it is "bundled" with

This will open the way to other improvements - introducing ServiceLocator and easier tests/code structure.

## How did you test it?

Existing tests.

## Potential risks

Can require changes in saas-temporal  and CDS.

## Is hotfix candidate?

No


service/history/ndc_task_util.go
```diff
@@ -337,10 +335,10 @@ func getNamespaceTagAndReplicationStateByID(
...
 	namespaceID string,
 ) (metrics.Tag, enumspb.ReplicationState) {
-	namespace, err := registry.GetNamespaceByID(namespace.ID(namespaceID))
+	namespaceName, err := registry.GetNamespaceByID(namespace.ID(namespaceID))
 	if err != nil {
 		return metrics.NamespaceUnknownTag(), enumspb.REPLICATION_STATE_UNSPECIFIED
 	}
 
-	return metrics.NamespaceTag(namespace.Name().String()), namespace.ReplicationState()
+	return metrics.NamespaceTag(namespaceName.Name().String()), namespaceName.ReplicationState()
```
yycptt, line 343: nit:
```suggestion
	return metrics.NamespaceTag(namespaceEntry.Name().String()), namespaceName.ReplicationState()
```

yycptt (approved):

yycptt:
Moving interface LGTM. 

I am not fully following the future steps in the description. That looks like a different kind of refactoring to me (using shard as property bag vs. grouping/moving interfaces) and probably need discussion within the team.

PR #7446, opened 2025-03-09
Jbarzee/remove migrated cluster handlers

jmbarzee:
## What changed?
Removed Handlers in admin service that have finished migrating.

## Why?
Dropping duplicated code.

## How did you test it?
Testing coverage, reviewed dependencies.

## Potential risks
Little to none, admin service is considered internal only.

## Documentation
N/A

## Is hotfix candidate?
No


service/frontend/admin_handler.go
```diff
@@ -1117,40 +1116,8 @@ func (adh *AdminHandler) DescribeCluster(
 	}, nil
 }
 
-// ListClusters return information about temporal clusters
-// TODO: Remove this API after migrate tctl to use operator handler
```
jmbarzee, line 1121: TCTL is deprecated, and CLI doesn't use admin service at all.

From Slack:
>The year or two ago they made the temporal CLI, they decided not to port admin things and only access frontend. I think their reasoning was that the server or tdbg or some other CLI could do it, unsure

service/frontend/admin_handler.go
```diff
@@ -1117,40 +1116,8 @@ func (adh *AdminHandler) DescribeCluster(
...
-		clusterMetadataList = append(clusterMetadataList, clusterResp.ClusterMetadata)
-	}
-	return &adminservice.ListClustersResponse{
-		Clusters:      clusterMetadataList,
-		NextPageToken: resp.NextPageToken,
-	}, nil
-}
 
 // ListClusterMembers
-// TODO: Remove this API after migrate tctl to use operator handler
```
jmbarzee, line 1153: This comment was made somewhat in error. This api is still critical, not duplicated anywhere and still used by tdbg.

alexshtin (approved):

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -288,32 +287,6 @@ message DescribeClusterResponse {
   string http_address = 14;
 }
 
-message ListClustersRequest {
-  int32 page_size = 1;
-  bytes next_page_token = 2;
-}
-
-message ListClustersResponse {
-  repeated temporal.server.api.persistence.v1.ClusterMetadata clusters = 1;
```
yycptt, line 297: operator service seem to return a different type here?

yycptt (commented):
cc @cgs

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
yux0: I think this is ok. But we should try to migrate use cases to operator service before merge this.

yux0 (approved):

PR #7447, opened 2025-03-09
remove search attribute RPC handlers from admin service (migrated to …

jmbarzee:
What changed?
Removed Handlers in admin service that have finished migrating.

Why?
Dropping duplicated code.

How did you test it?
Testing coverage, reviewed dependencies.

Potential risks
Little to none, admin service is considered internal only.

Documentation
N/A

Is hotfix candidate?
No


yycptt (approved):
cc @rodrigozhou

rodrigozhou (changes requested):
I'm not sure we can remove those yet, not until we completely deprecate `tctl`.

PR #7448, opened 2025-03-10
stop infinite activity retries in the entity workflows

Shivs11:
## What changed?

- Right now, entity workflows related to versioning have infinite number of retries happening. In the case of an unforeseen error, this might leave our internal entity workflows locked and might require manual intervention.
- The activity context, being used in all of the activities related to the worker-deployment workflows, now has a maximum number of attempts set to 5. This shall ensure that on the eve of an unforeseen error, the deployment workflow will unlock itself after a maximum time of ~5 mins. [ ~5 mins calculated by doing math involving backoff coefficient, initial interval and maximum attempts ]  

## Why?

- To not lock our own workflows.

## How did you test it?

- Current Suite of tests to verify nothing is broken.

### DIFF
--- a/service/worker/workerdeployment/util.go
+++ b/service/worker/workerdeployment/util.go
@@ -111,7 +111,6 @@ var (
 		StartToCloseTimeout: 1 * time.Minute,
 		RetryPolicy: &temporal.RetryPolicy{
 			InitialInterval: 100 * time.Millisecond,
-			MaximumInterval: 60 * time.Second,
 			MaximumAttempts: 5,
 		},
 	}
--- a/service/worker/workerdeployment/util.go
+++ b/service/worker/workerdeployment/util.go
@@ -32,6 +32,7 @@ import (
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/workflow"
 	deploymentspb "go.temporal.io/server/api/deployment/v1"
 	"go.temporal.io/server/common"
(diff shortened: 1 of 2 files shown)


service/worker/workerdeployment/util.go
```diff
@@ -105,10 +106,19 @@ var (
 	)
 )
 
+var (
+	defaultActivityOptions = workflow.ActivityOptions{
+		StartToCloseTimeout: 1 * time.Minute,
```
yiminc, line 111: Is this timeout too long?

service/worker/workerdeployment/util.go
```diff
@@ -105,10 +106,19 @@ var (
...
+		RetryPolicy: &temporal.RetryPolicy{
+			InitialInterval: 100 * time.Millisecond,
+			MaximumInterval: 60 * time.Second,
+			MaximumAttempts: 5,
```
yiminc, line 115: With max 5 attempt, max retry interval would be 800ms, the MaxInterval 60s set here is no use.

service/worker/workerdeployment/util.go
Shivs11: I thought about it too but I don't think so - In the worst case scenario, we could have userData syncs happening for 100 different task-queues and I think it's okay for us to be a little relaxed with the time constraint here

service/worker/workerdeployment/util.go
Shivs11: sure, removing this.

service/worker/workerdeployment/util.go
Shivs11: also discussed this with @stephanos and we came to a conclusion to be conservative and go ahead with this option.

stephanos: Based on my understanding, using a longer timeout for the time being addresses the bigger issue of never finishing and leaving things in a bad/partial state. And there will be follow-up discussions on finding a smoother user experience.

yiminc (approved):

PR #7450, opened 2025-03-10
Refactoring. Move workflow context release out of cache package

ychebotarev:
## What changed?

* Move `ReleaseCacheFunc` out of cache, to workflow context, and rename it to `ReleaseWorkflowContextFunc`
* Move OperationPossiblySucceeded to common/persistence

## Why?

Part of continuous refactoring work.
* `ReleaseCacheFunc` should't be in cache to start with.
* `ReleaseCacheFunc`  shouldn't have `cache` in the name, it is internal detail
* it removes multiple imports to `wcache` and `shard`

More details:

Here is a typical code that use shard and cache:
* Some structure is initialized with both shard and cache:
```
historyEngImpl.nDCActivityStateReplicator = ndc.NewActivityStateReplicator(
  shard,
  workflowCache,
...
```

then both cache and shard are used to get workflow context:

```
executionContext, release, err := r.workflowCache.GetOrCreateWorkflowExecution(
  ctx,
  r.shardContext,
...
```

Desired state - getting workflow context without explicitly exposing/using cache.

```
historyEngImpl.nDCActivityStateReplicator = ndc.NewActivityStateReplicator(
  shard,
...
```
Using to get context:
```
wfContext, release, err := r.shard.GetOrCreateConxtext(
   ctx,
   ...
```

This will remove all the dependencies on cache, and tons of extra struct members

## How did you test it?

Existing test

## Potential risks

saas-temporal may be affected


yycptt (commented):
Do you plan to:
- Move `Cache` to interface package as well?
- Split Cache into an actual cache and a locking component? I mean if no, then this releaseFunc _is_ part of the Cache interface since it's also doing locking.

ychebotarev:
> Do you plan to:
> 
> * Move `Cache` to interface package as well?
> * Split Cache into an actual cache and a locking component? I mean if no, then this releaseFunc _is_ part of the Cache interface since it's also doing locking.

For this specific aspect (locking in ReleaseFunc):
* I don't plan to move `Cache` to the interface package unless I absolutely must. It is already a bit controversial change, I don't want to make interfaces "free for all" directory.
* I do plan to move ->workflow lease<- to interfaces, and return WorkflowLease instead of (WorkflowContext, ReleaseFunc). We already have it in some places, and in some we don't. Not sure why.
* Because of this I'm not sure where locking will be at the end, so I don't know if I split cache or not.

service/history/interfaces/shard_context.go
```diff
@@ -120,6 +121,9 @@ type (
 		// If branchToken != nil, then delete history also, otherwise leave history.
 		DeleteWorkflowExecution(ctx context.Context, workflowKey definition.WorkflowKey, branchToken []byte, closeExecutionVisibilityTaskID int64, workflowCloseTime time.Time, stage *tasks.DeleteWorkflowExecutionStage) error
 
+		GetWorkflowContext(ctx context.Context, namespaceID namespace.ID, execution *commonpb.WorkflowExecution, lockPriority locks.Priority) (WorkflowContext, ReleaseWorkflowContextFunc, error)
+		GetCurrentWorkflowContext(ctx context.Context, namespaceID namespace.ID, workflowID string, lockPriority locks.Priority) (ReleaseWorkflowContextFunc, error)
+
```
yycptt, line 126: I think this needs team wide discussion. 
- It practically means we are going further on the road of using shard context as property bag and make workflow cache part of it for loading workflow context.
- It means we start to have both low level (like persistence facing) and high level method in shard context.

yycptt (commented):
> I do plan to move ->workflow lease<- to interfaces, and return WorkflowLease instead of (WorkflowContext, ReleaseFunc). We already have it in some places, and in some we don't. Not sure why.

WorkflowLease is only used in API packages (for those UpdateWorkflowAction functions I guess) and a very thin wrapper of WorkflowContext, mutableState & releasefunc. It's not used at all in task processing side and I am not sure why it needs to be an interface as it's only for providing access to those three things. 

However the way it's defined means we can not reload mutable state after getting a lease (the mutableState in it will always point to the stale one), which is required for doing consistency checks. So if WorkflowLease is returned from WorkflowCache, either it means we need to move all the consistency check logic into WorkflowCache, or add another method to the lease interface to fix the stale mutable state reference.

So all those are doable, task processing side can also be fixed after some nontrivial work. But can you provide some more insight into the benefit we get from this refactoring? So far it's not very clear to me and seems to be yet another layer on top of WorkflowContext and it needs to be more complicated to be adopted by lower levels. Maybe I am missing some context here and the goal you are trying to achieve.

service/history/interfaces/shard_context.go
ychebotarev: right now we expose persistence layer (cache) to tons of classes, alongside the shard context.
Moving cache into shard context hides this implementation detail, and remove dependency on cache for every case where WorkflowContext is locked.

That said - I will remove it from this PR, it was actually next step.

ychebotarev: discussed offline. I'm closing this since we reach the agreement.

service/history/interfaces/shard_context.go
```diff
@@ -120,6 +121,9 @@ type (
 		// If branchToken != nil, then delete history also, otherwise leave history.
 		DeleteWorkflowExecution(ctx context.Context, workflowKey definition.WorkflowKey, branchToken []byte, closeExecutionVisibilityTaskID int64, workflowCloseTime time.Time, stage *tasks.DeleteWorkflowExecutionStage) error
 
+		GetCachedWorkflowContext(ctx context.Context, namespaceID namespace.ID, execution *commonpb.WorkflowExecution, lockPriority locks.Priority) (WorkflowContext, ReleaseWorkflowContextFunc, error)
+		GetCurrentCachedWorkflowContext(ctx context.Context, namespaceID namespace.ID, workflowID string, lockPriority locks.Priority) (ReleaseWorkflowContextFunc, error)
```
yycptt, line 125: I still don't view Cache as an implementation detail of shard context which is persistence facing, and think Cache is one layer on top of that. 

I view those two methods as providing syntax sugar for doing `shardContext.GetWorkflowCache().GetWorkflowContext()` and using shard context as property bag for providing access to workflow cache (a host level component) thus saving us from always passing both shard context and workflow cache around to all components.

That also means if later we agreed on an approach for not using shard context as the property bag, those two methods should be removed as well. 

My approval is based on ^ interpretation of those two methods.

service/history/shard/context_impl.go
```diff
@@ -2248,6 +2248,22 @@ func (s *ContextImpl) ChasmRegistry() *chasm.Registry {
 	return s.chasmRegistry
 }
 
+func (s *ContextImpl) GetCachedWorkflowContext(
+	ctx context.Context,
+	namespaceID namespace.ID,
+	execution *commonpb.WorkflowExecution,
+	lockPriority locks.Priority,
+) (historyi.WorkflowContext, historyi.ReleaseWorkflowContextFunc, error) {
+	return nil, nil, nil
```
yycptt, line 2257: return NotImplemented error or panic here.

yycptt (approved):

PR #7455, opened 2025-03-12
[CHASM] SQL persistence for CHASM nodes

lina-temporal:
## What changed?
SQL persistence support has been added for CHASM nodes. 

- A new `chasm_node_maps` table has been added to SQL schemas
- SQL test suites updated
- SQL execution manager/util updated
- SQL plugins for `chasm_node_map` table

## Why?
- We've landed Cassandra persistence support for CHASM nodes, this brings SQL support up to parity.

## How did you test it?
- New tests ran on all SQL plugins
- `go test -v ./...`

## Potential risks
- No application paths write chasm_node_maps yet, but change does involve a schema change

### DIFF
--- a/common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
+++ b/common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
@@ -1,3 +1,27 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
 package tests
 
 import (
--- a/common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
+++ b/common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
@@ -87,8 +87,8 @@ func (s *historyExecutionChasmSuite) runTestCase(tc *testCase) {
 		s.NoError(err)
 
 		// We set clientFoundRows to true in our MySQL session, which makes the result count
-		// for updates not useful for comparison here, as rows matched by WHERE (not just
-		// those updated with IN) are counted:
+		// for updates not useful for comparison here, as rows that have been updated
+		// are double-counted in `INSERT ... ON DUPLICATE KEY UPDATE` statements:
 		//
 		// https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_row-count
 		//
--- a/common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
+++ b/common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
@@ -1,0 +1,251 @@
+package tests
+
+import (
+	"math/rand/v2"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"go.temporal.io/server/common/persistence/sql/sqlplugin"
+	"go.temporal.io/server/common/primitives"
+)
+
+type (
+	historyExecutionChasmSuite struct {
+		suite.Suite
+		*require.Assertions
+
+		store sqlplugin.HistoryExecutionChasm
+	}
+)
+
+func NewHistoryExecutionChasmSuite(
+	t *testing.T,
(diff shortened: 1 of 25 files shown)


common/persistence/tests/util.go
```diff
@@ -122,12 +115,6 @@ func RandomMutation(
 	dbRecordVersion int64,
 	branchToken []byte,
 ) (*p.WorkflowMutation, []*p.WorkflowEvents) {
-	// TODO - remove this branching when other persistence implementations land for CHASM
-	var chasmNodes map[string]*persistencespb.ChasmNode
-	if strings.HasPrefix(t.Name(), "TestCassandra") {
-		chasmNodes = RandomChasmNodeMap()
-	}
```
yycptt, line 129: Still need this to skip the chasm part for internal persistence impl.

schema/mysql/v8/temporal/schema.sql
```diff
@@ -251,6 +251,18 @@ CREATE TABLE signals_requested_sets (
   PRIMARY KEY (shard_id, namespace_id, workflow_id, run_id, signal_id)
 );
 
+CREATE TABLE chasm_node_maps (
+  shard_id INT NOT NULL,
+  namespace_id BINARY(16) NOT NULL,
+  workflow_id VARCHAR(255) NOT NULL,
+  run_id BINARY(16) NOT NULL,
+  chasm_path VARCHAR(384) NOT NULL,
```
yycptt, line 259: shall we specify the column to use bin collation here? I think that's not the default.

I understand collation can also be specified at query time, but not sure if there's a performance implication. I'd imaging there might be one if the order from the collation specified is different from the one of the primary index?

schema/postgresql/v12/temporal/schema.sql
```diff
@@ -251,6 +251,18 @@ CREATE TABLE signals_requested_sets (
   PRIMARY KEY (shard_id, namespace_id, workflow_id, run_id, signal_id)
 );
 
+CREATE TABLE chasm_node_maps (
+  shard_id INTEGER NOT NULL,
+  namespace_id BYTEA NOT NULL,
+  workflow_id VARCHAR(255) NOT NULL,
+  run_id BYTEA NOT NULL,
+  chasm_path VARCHAR(384) NOT NULL,
```
yycptt, line 259: same question if "C" collate should be specified here.

It also seems like the "C" collate compare string byte by byte even if the code point is multi-bytes. So it's more similar to mysql's varbinary I think. Considering cassandra string is also doing byte by byte comparison. I feel it would be better to make all implementations consistent? We also need to consider only one implementation when deciding the chasm path encoding alg.

schema/mysql/v8/temporal/schema.sql
lina-temporal: > shall we specify the column to use bin collation here? I think that's not the default.

Yeah, we can specify that here. 

> I understand collation can also be specified at query time, but not sure if there's a performance implication. I'd imaging there might be one if the order from the collation specified is different from the one of the primary index?

From what I've read, statement-time collation only has a performance hit if [it has to perform a character set conversion](https://dev.mysql.com/doc/refman/8.4/en/optimize-character.html) between strings (which it shouldn't, since all of our VARCHARs are left as the default `utf8mb4`). I'm on-board with specifying it right on the column, though, since that simplifies queries.

schema/postgresql/v12/temporal/schema.sql
lina-temporal: > It also seems like the "C" collate compare string byte by byte even if the code point is multi-bytes. So it's more similar to mysql's varbinary I think. 

Yeah, that's true. I think the biggest practical difference for our case would be the difference in handling padding/space characters, since we aren't otherwise doing any uppercase/lowercase conversions in our range compares. Still, it's probably nicer to reason about only having binary strings in all implementations, so I'm in favor of making this a `BYTEA` (and `VARBINARY` for MySQL).

common/persistence/tests/util.go
lina-temporal: Sorry, I'm not following, could you point me to this test group? I haven't seen any failures relating to a missing impl.

common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
```diff
@@ -0,0 +1,275 @@
...
+		affected, err := res.RowsAffected()
+		s.NoError(err)
+
+		// We set clientFoundRows to true in our MySQL session, which makes the result count
+		// for updates not useful for comparison here, as rows matched by WHERE (not just
+		// those updated with IN) are counted:
+		//
+		// https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_row-count
+		//
+		// See common/persistence/sql/sqlplugin/mysql/session/session.go
```
yycptt, line 95: is it because we don't know if it's a new row (counted as 1) or updating existing row (counted as 2)?

Do we have the same issue in other mysql tests?

common/persistence/tests/util.go
yycptt: Synced offline. One approach is to check execution store name or we need to rename some internal tests. The latter one seems to be easier if team is ok with that.

common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
lina-temporal: > is it because we don't know if it's a new row (counted as 1) or updating existing row (counted as 2)?

Because of the parameter we have set, `clientFoundRows` is set to the overall set selected by the `WHERE` clause, not the number of rows mutated. Older tests don't trigger the same conditions, correct.

common/persistence/tests/util.go
lina-temporal: [PR opened to SaaS](https://github.com/temporalio/saas-temporal/pull/3101)

common/persistence/sql/sqlplugin/tests/history_execution_chasm.go
yycptt: hmm I probably missed something here, the linked doc seem to suggest ^ is for `UPDATE` Statements. 

The query for ReplaceIntoChasmNodeMaps is using `INSERT ... ON DUPLICATE KEY UPDATE` though.

lina-temporal: Oh, yeah, you're right, for this assertion it's the 

> For [INSERT ... ON DUPLICATE KEY UPDATE](https://dev.mysql.com/doc/refman/8.4/en/insert-on-duplicate.html) statements, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the CLIENT_FOUND_ROWS flag, the affected-rows value is 1 (not 0) if an existing row is set to its current values.

behavior. Yeah, it's counted at 2x because we're updating existing rows.

lina-temporal: Updated the comment to reflect.

alexshtin (approved):

PR #7459, opened 2025-03-13
Adding replication task backlog count

yux0:
## What changed?

Adding replication task backlog count

## Why?

This metrics tells the task count in the backlog so we have task count and time based lagging info.

### DIFF
--- a/service/history/replication/ack_manager.go
+++ b/service/history/replication/ack_manager.go
@@ -60,7 +60,7 @@ type (
 		GetTasks(ctx context.Context, pollingCluster string, queryMessageID int64) (*replicationspb.ReplicationMessages, error)
 		GetTask(ctx context.Context, taskInfo *replicationspb.ReplicationTaskInfo) (*replicationspb.ReplicationTask, error)
 
-		SubscribeNotification() (<-chan struct{}, string)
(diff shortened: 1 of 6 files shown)


service/history/replication/ack_manager.go
```diff
@@ -509,14 +519,21 @@ func (p *ackMgrImpl) UnsubscribeNotification(subscriberID string) {
...
 
-	for _, channel := range p.subscribers {
+	for _, notification := range p.subscribers {
 		select {
-		case channel <- struct{}{}:
+		case notification.notifyCh <- struct{}{}:
+			// This tells the backlog between two notifications. This is the best effort to know lagging task count.
+			if notification.backlogCount > 0 {
+				metrics.ReplicationTaskSendBacklog.With(p.metricsHandler).
+					Record(notification.backlogCount, metrics.TargetClusterTag(notification.clusterName))
```
hai719, line 532: nit: can we add shardId as well?

hai719 (approved):

service/history/replication/ack_manager.go
yux0: We cannot. There  will be issue on cardinality.

PR #7466, opened 2025-03-13
versioning test: concurrent stale updates

Shivs11:
## What changed?

- Added tests for testing concurrent stale updates. 

## Why?

- more testing coverage.

## How did you test it?

- Added new unit tests + existing suite

## Potential risks

- None

## Is hotfix candidate?

- No

### DIFF
--- a/service/worker/workerdeployment/workflow_test.go
+++ b/service/worker/workerdeployment/workflow_test.go
@@ -122,17 +122,18 @@ func (s *WorkerDeploymentSuite) Test_SetCurrentVersion_RejectStaleConcurrentUpda
 		}, updateArgs)
 	}, 1*time.Millisecond)
 
-	// Update #1 clears the validator but is then halted because of a sleep in the activity. This is
+	// Update #1 clears the validator and is then halted during activity processing. This is
 	// to simulate the workflow giving up control which shall allow update #2 to clear the validator.
-	// However, update #2 cannot yet start being processed since update #1 holds the lock and will only
-	// release it after it has completed processing.
+	// However, update #2 cannot yet start being processed since update #1 holds the workflow lock and
+	// will only release it after it has completed processing.
+	haltUpdate := make(chan struct{})
 	var a *Activities
 	s.env.RegisterActivity(a.SyncWorkerDeploymentVersion)
 	s.env.OnActivity(a.SyncWorkerDeploymentVersion, mock.Anything, mock.Anything).Once().Return(
 		func(ctx context.Context, args *deploymentspb.SyncVersionStateActivityArgs) (*deploymentspb.SyncVersionStateActivityResult, error) {
 			if args.Version == tv.DeploymentVersionString() {
-				// Sleep in the activity to simulate workflow giving up control while processing the first update.
-				time.Sleep(1 * time.Second)
+				// Halt the activity to simulate workflow giving up control while processing the first update.
+				<-haltUpdate
 			}
 			return &deploymentspb.SyncVersionStateActivityResult{}, nil
 		},
@@ -151,12 +152,18 @@ func (s *WorkerDeploymentSuite) Test_SetCurrentVersion_RejectStaleConcurrentUpda
 			OnAccept: func() {
 			},
 			OnComplete: func(a interface{}, err error) {
-				// Update #2 clears the validator and waits for the first update to complete. However,
-				// during processing, it should be rejected since completion of the first update changed the state.
+				// Update #2 clears the validator and waits for the first update to complete. Once it starts
+				// being processed, it should be rejected since completion of the first update changed the state.
 				s.ErrorContains(err, errNoChangeType)
 			},
 		}, updateArgs)
 	}, 1*time.Millisecond)
+
+	// Close the channel to unblock the activity and allow update #1 to complete.
+	// Note: This is called after update #2 begins being processed since it's fired after 2 milliseconds.
+	s.env.RegisterDelayedCallback(func() {
+		close(haltUpdate)
+	}, 2*time.Millisecond)
 
 	deploymentWorkflow := func(ctx workflow.Context, args *deploymentspb.WorkerDeploymentWorkflowArgs) error {
 		maxVersionsGetter := func() int {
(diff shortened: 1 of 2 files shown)


service/worker/workerdeployment/workflow_test.go
```diff
@@ -47,9 +56,239 @@ func (s *WorkerDeploymentSuite) SetupTest() {
...
+func (s *WorkerDeploymentSuite) Name() string {
+	fullName := s.T().Name()
+	if len(fullName) <= 30 {
+		return fullName
+	}
+	short := fmt.Sprintf("%s-%08x",
+		fullName[len(fullName)-21:],
+		farm.Fingerprint32([]byte(fullName)),
+	)
+	return strings.Replace(short, ".", "|", -1)
```
stephanos, line 77: You can use `[ReplaceAll](https://cs.opensource.google/go/go/+/go1.24.1:src/strings/strings.go;l=1180) [¶](https://pkg.go.dev/strings#ReplaceAll) for this instead.

service/worker/workerdeployment/workflow_test.go
```diff
@@ -47,9 +56,239 @@ func (s *WorkerDeploymentSuite) SetupTest() {
...
 func (s *WorkerDeploymentSuite) TearDownTest() {
 	s.controller.Finish()
 	s.env.AssertExpectations(s.T())
 }
+
+// To incorporate testVars in our tests.
+func (s *WorkerDeploymentSuite) Name() string {
+	fullName := s.T().Name()
+	if len(fullName) <= 30 {
+		return fullName
```
stephanos, line 71: Why not always use the full name?

service/worker/workerdeployment/workflow_test.go
```diff
@@ -47,9 +56,239 @@ func (s *WorkerDeploymentSuite) SetupTest() {
...
+	s.env.RegisterDelayedCallback(func() {
+		updateArgs := &deploymentspb.SetCurrentVersionArgs{
+			Identity:                tv.ClientIdentity(),
+			Version:                 tv.DeploymentVersionString(),
+			IgnoreMissingTaskQueues: true,
+		}
+		// Setting the new version as current version
+		s.env.UpdateWorkflow(SetCurrentVersion, "", &testsuite.TestUpdateCallback{
+			OnReject: func(err error) {
+				s.Fail("update should not have failed with error %v", err)
```
stephanos, line 116:
```suggestion
				s.Fail("update #1 should not have failed with error %v", err)
```

service/worker/workerdeployment/workflow_test.go
```diff
@@ -47,9 +56,239 @@ func (s *WorkerDeploymentSuite) SetupTest() {
...
+	s.env.ExecuteWorkflow(deploymentWorkflow, &deploymentspb.WorkerDeploymentWorkflowArgs{
+		NamespaceName:  tv.NamespaceName().String(),
+		NamespaceId:    tv.NamespaceID().String(),
+		DeploymentName: tv.DeploymentSeries(),
+	})
+
+	s.True(s.env.IsWorkflowCompleted())
+	// Workflow times out so an error is expected.
+	s.Error(s.env.GetWorkflowError())
+	s.ErrorContains(s.env.GetWorkflowError(), "deadline exceeded (type: ScheduleToClose)")
```
stephanos, line 184: Why does it timeout?

service/worker/workerdeployment/workflow_test.go
```diff
@@ -47,9 +56,239 @@ func (s *WorkerDeploymentSuite) SetupTest() {
...
+		updateArgs := &deploymentspb.SetRampingVersionArgs{
+			Identity:                tv.ClientIdentity(),
+			Version:                 tv.DeploymentVersionString(),
+			Percentage:              50,
+			IgnoreMissingTaskQueues: true,
+		}
+		// Setting the new version as current version
+		s.env.UpdateWorkflow(SetRampingVersion, "", &testsuite.TestUpdateCallback{
+			OnReject: func(err error) {
+				s.Fail("update should not have failed with error %v", err)
```
stephanos, line 224:
```suggestion
				s.Fail("update #1 should not have failed with error %v", err)
```

service/worker/workerdeployment/workflow_test.go
Shivs11: I think I can do one better - I can remove the need for all of this and just use `s.T()` whose `Name()` function has what you desire.

service/worker/workerdeployment/workflow_test.go
Shivs11: The answer is because of the fact that the completion of the `ExecuteWorkflow` call results in the workflow execution timer firing. For these examples, the workflow execution time for each test is set to 30 seconds. If you try running the test, you shall see that after the `ExecuteWorkflow` command is run, you get a : `Auto fire timer TimerID 3 TimerDuration 30s TimeSkipped 29.998s`, signifying that the maximum time allowed to run this workflow has elapsed which results in an error.

So to understand that this was an expected error took me quite some time, but I eventually found similarity [here](https://github.com/temporalio/samples-go/blob/8334deef6d345fffedea657cb4952ee319f582b2/await-signals/await_signals_workflow_test.go#L25)

Docs: https://docs.temporal.io/encyclopedia/detecting-workflow-failures#workflow-execution-timeout

service/worker/workerdeployment/workflow_test.go
```diff
@@ -47,9 +56,239 @@ func (s *WorkerDeploymentSuite) SetupTest() {
 	s.controller = gomock.NewController(s.T())
 	s.env = s.WorkflowTestSuite.NewTestWorkflowEnvironment()
 	s.env.RegisterWorkflow(Workflow)
+	s.env.SetWorkflowRunTimeout(30 * time.Second)
```
Shivs11, line 59: run timeout being set here 
cc - @stephanos

service/worker/workerdeployment/workflow_test.go
stephanos: Could we just remove the two assertions and `s.env.SetWorkflowRunTimeout(30 * time.Second)`? `IsWorkflowCompleted` seems to be sufficient to me; since you've tested all you needed.

stephanos (approved):
Good work!

PR #7469, opened 2025-03-14
Re-use unit tests

stephanos:
## What changed?

Re-using existing unit test suites for new priority/backlog/forwarder and:

- fixes two data races
- uses testlogger.Testlogger

Incompatible tests are skipped: `s.T().Skip("not supported by new matcher")`. They can either be ported or left as-is in follow-up PRs.

## Why?

Increase test coverage.

## Potential risks

A handful of tests had to be slightly tweaked to make them compatible. It's possible there are regressions in flakiness.

### DIFF
--- a/service/matching/backlog_manager.go
+++ b/service/matching/backlog_manager.go
@@ -67,6 +67,9 @@ type (
 		TotalApproximateBacklogCount() int64
 		BacklogHeadAge() time.Duration
 		InternalStatus() []*taskqueuespb.InternalTaskQueueStatus
+
+		// TODO(pri): remove
+		getDB() *taskQueueDB
 	}
 
 	backlogManagerImpl struct {
@@ -232,8 +235,9 @@ func (c *backlogManagerImpl) InternalStatus() []*taskqueuespb.InternalTaskQueueS
 				StartId: c.taskWriter.taskIDBlock.start,
 				EndId:   c.taskWriter.taskIDBlock.end,
 			},
-			LoadedTasks:  c.taskAckManager.getBacklogCountHint(),
-			MaxReadLevel: c.db.GetMaxReadLevel(subqueueZero),
+			LoadedTasks:             c.taskAckManager.getBacklogCountHint(),
+			MaxReadLevel:            c.db.GetMaxReadLevel(subqueueZero),
+			ApproximateBacklogCount: c.db.getApproximateBacklogCount(subqueueZero),
 		},
 	}
 }
@@ -305,4 +309,8 @@ func executeWithRetry(
 
 func (c *backlogManagerImpl) queueKey() *PhysicalTaskQueueKey {
 	return c.pqMgr.QueueKey()
+}
+
+func (c *backlogManagerImpl) getDB() *taskQueueDB {
+	return c.db
 }
--- a/service/matching/backlog_manager_test.go
+++ b/service/matching/backlog_manager_test.go
@@ -272,7 +272,7 @@ func (s *BacklogManagerTestSuite) TestApproximateBacklogCount_IncrementedBySpool
 	defer s.blm.Stop()
 
 	taskCount := 10
-	s.ptqMgr.EXPECT().AddSpooledTask(gomock.Any()).Return(nil).MaxTimes(taskCount)
+	s.ptqMgr.EXPECT().AddSpooledTask(gomock.Any()).Return(nil).AnyTimes()
 	for i := 0; i < taskCount; i++ {
 		s.NoError(s.blm.SpoolTask(&persistencespb.TaskInfo{
 			ExpiryTime: timestamp.TimeNowPtrUtcAddSeconds(3000),
@@ -291,7 +291,7 @@ func (s *BacklogManagerTestSuite) TestApproximateBacklogCount_IncrementedBySpool
 	defer s.blm.Stop()
 
 	taskCount := 10
-	s.ptqMgr.EXPECT().AddSpooledTask(gomock.Any()).Return(nil).MaxTimes(taskCount)
+	s.ptqMgr.EXPECT().AddSpooledTask(gomock.Any()).Return(nil).AnyTimes()
 	for i := 0; i < taskCount; i++ {
 		s.Error(s.blm.SpoolTask(&persistencespb.TaskInfo{
 			ExpiryTime: timestamp.TimeNowPtrUtcAddSeconds(3000),
--- a/service/matching/backlog_manager_test.go
+++ b/service/matching/backlog_manager_test.go
@@ -29,28 +29,64 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
 	enumspb "go.temporal.io/api/enums/v1"
-	"go.temporal.io/server/api/matchingservicemock/v1"
 	persistencespb "go.temporal.io/server/api/persistence/v1"
 	"go.temporal.io/server/common/dynamicconfig"
-	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/primitives/timestamp"
+	"go.temporal.io/server/common/testing/testlogger"
 	"go.temporal.io/server/common/tqid"
 	"go.uber.org/mock/gomock"
 )
 
-func TestReadLevelForAllExpiredTasksInBatch(t *testing.T) {
-	controller := gomock.NewController(t)
+type BacklogManagerTestSuite struct {
+	suite.Suite
 
-	// TODO: do not create pq manager, directly create backlog manager
-	tlm := mustCreateTestPhysicalTaskQueueManager(t, controller)
-	blm := tlm.backlogMgr.(*backlogManagerImpl)
-	require.NoError(t, blm.taskWriter.initReadWriteState())
-	require.Equal(t, int64(1), blm.db.rangeID)
-	require.Equal(t, int64(0), blm.taskAckManager.getAckLevel())
-	require.Equal(t, int64(0), blm.taskAckManager.getReadLevel())
+	logger     *testlogger.TestLogger
+	blm        *backlogManagerImpl
+	controller *gomock.Controller
+	cancelCtx  context.CancelFunc
+	taskMgr    *testTaskManager
+	ptqMgr     *MockphysicalTaskQueueManager
+}
+
+func TestBacklogManagerTestSuite(t *testing.T) {
+	t.Parallel()
+	suite.Run(t, &BacklogManagerTestSuite{})
+}
+
+func (s *BacklogManagerTestSuite) SetupTest() {
+	s.controller = gomock.NewController(s.T())
+	s.logger = testlogger.NewTestLogger(s.T(), testlogger.FailOnAnyUnexpectedError)
+	s.taskMgr = newTestTaskManager(s.logger)
+
+	cfg := NewConfig(dynamicconfig.NewNoopCollection())
+	f, _ := tqid.NewTaskQueueFamily("", "test-queue")
+	prtn := f.TaskQueue(enumspb.TASK_QUEUE_TYPE_WORKFLOW).NormalPartition(0)
+	queue := UnversionedQueueKey(prtn)
+	tlCfg := newTaskQueueConfig(prtn.TaskQueue(), cfg, "test-namespace")
+
+	s.ptqMgr = NewMockphysicalTaskQueueManager(s.controller)
+	s.ptqMgr.EXPECT().QueueKey().Return(queue).AnyTimes()
+	s.ptqMgr.EXPECT().ProcessSpooledTask(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+
+	var ctx context.Context
+	ctx, s.cancelCtx = context.WithCancel(context.Background())
+	s.T().Cleanup(s.cancelCtx)
+
+	s.blm = newBacklogManager(ctx, s.ptqMgr, tlCfg, s.taskMgr, s.logger, s.logger, nil, metrics.NoopMetricsHandler)
+}
+
+func (s *BacklogManagerTestSuite) TearDownTest() {
+	s.controller.Finish()
+}
+
+func (s *BacklogManagerTestSuite) TestReadLevelForAllExpiredTasksInBatch() {
+	s.NoError(s.blm.taskWriter.initReadWriteState())
+	s.Equal(int64(1), s.blm.db.rangeID)
+	s.Equal(int64(0), s.blm.taskAckManager.getAckLevel())
+	s.Equal(int64(0), s.blm.taskAckManager.getReadLevel())
 
 	// Add all expired tasks
 	tasks := []*persistencespb.AllocatedTaskInfo{
@@ -70,12 +106,12 @@ func TestReadLevelForAllExpiredTasksInBatch(t *testing.T) {
 		},
 	}
 
-	require.NoError(t, blm.taskReader.addTasksToBuffer(context.TODO(), tasks))
-	require.Equal(t, int64(0), blm.taskAckManager.getAckLevel())
-	require.Equal(t, int64(12), blm.taskAckManager.getReadLevel())
+	s.NoError(s.blm.taskReader.addTasksToBuffer(context.TODO(), tasks))
+	s.Equal(int64(0), s.blm.taskAckManager.getAckLevel())
+	s.Equal(int64(12), s.blm.taskAckManager.getReadLevel())
 
 	// Now add a mix of valid and expired tasks
-	require.NoError(t, blm.taskReader.addTasksToBuffer(context.TODO(), []*persistencespb.AllocatedTaskInfo{
+	s.NoError(s.blm.taskReader.addTasksToBuffer(context.TODO(), []*persistencespb.AllocatedTaskInfo{
 		{
 			Data: &persistencespb.TaskInfo{
 				ExpiryTime: timestamp.TimeNowPtrUtcAddSeconds(-60),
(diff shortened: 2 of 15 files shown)


service/matching/physical_task_queue_manager.go
```diff
@@ -144,7 +144,6 @@ var (
 func newPhysicalTaskQueueManager(
 	partitionMgr *taskQueuePartitionManagerImpl,
 	queue *PhysicalTaskQueueKey,
-	opts ...taskQueueManagerOpt,
```
stephanos, line 147: Not needed anymore.

service/matching/pri_task_reader.go
```diff
@@ -518,11 +518,11 @@ func (tr *priTaskReader) doGC() {
 	ctx, cancel := context.WithTimeout(tr.backlogMgr.tqCtx, ioTimeout)
 	defer cancel()
 
-	n, err := tr.backlogMgr.db.CompleteTasksLessThan(ctx, tr.ackLevel+1, batchSize, tr.subqueue)
```
stephanos, line 521: Several unit tests produced a data race if `tr.ackLevel` is read outside of the lock.

service/matching/user_data_manager_test.go
```diff
@@ -971,3 +979,11 @@ func TestUserData_CheckPropagation(t *testing.T) {
...
 	}, 100*time.Millisecond, 5*time.Millisecond, "CheckTaskQueueUserDataPropagation did not return fast enough")
 }
+
+func defaultTqmTestOpts(controller *gomock.Controller) *tqmTestOpts {
+	return &tqmTestOpts{
+		config:             defaultTestConfig(),
+		dbq:                defaultTqId(),
+		matchingClientMock: matchingservicemock.NewMockMatchingServiceClient(controller),
+	}
+}
```
stephanos, line 989: ^ moved here since it's the only file left that uses it.

service/matching/backlog_manager.go
```diff
@@ -232,8 +235,9 @@ func (c *backlogManagerImpl) InternalStatus() []*taskqueuespb.InternalTaskQueueS
 				StartId: c.taskWriter.taskIDBlock.start,
 				EndId:   c.taskWriter.taskIDBlock.end,
 			},
-			LoadedTasks:  c.taskAckManager.getBacklogCountHint(),
-			MaxReadLevel: c.db.GetMaxReadLevel(subqueueZero),
+			LoadedTasks:             c.taskAckManager.getBacklogCountHint(),
+			MaxReadLevel:            c.db.GetMaxReadLevel(subqueueZero),
+			ApproximateBacklogCount: c.db.getApproximateBacklogCount(subqueueZero),
```
stephanos, line 240: Added this for parity with new backlog manager.

service/matching/backlog_manager_test.go
```diff
@@ -29,28 +29,64 @@ import (
...
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/primitives/timestamp"
+	"go.temporal.io/server/common/testing/testlogger"
 	"go.temporal.io/server/common/tqid"
 	"go.uber.org/mock/gomock"
 )
 
-func TestReadLevelForAllExpiredTasksInBatch(t *testing.T) {
-	controller := gomock.NewController(t)
+type BacklogManagerTestSuite struct {
```
stephanos, line 43: I revamped this test suite to understand it better and make it easier to port (what's needed). Didn't quite get there yet, though.

service/matching/pri_task_reader.go
```diff
@@ -173,9 +173,9 @@ func (tr *priTaskReader) completeTask(task *internalTask, res taskResponse) {
 		tr.SignalTaskLoading()
 	}
 
-	tr.lock.Unlock()
+	tr.backlogMgr.db.updateAckLevelAndCount(tr.subqueue, tr.ackLevel, -numAcked)
 
-	tr.backlogMgr.db.updateAckLevelAndCount(tr.subqueue, newAckLevel, -numAcked)
+	tr.lock.Unlock()
```
stephanos, line 178: There's a race causing the `should not move backwards ` assertion failure if the unlock happens before the call to `updateAckLevelAndCount`.

service/matching/pri_task_reader.go
```diff
@@ -159,11 +159,11 @@ func (tr *priTaskReader) completeTask(task *internalTask, res taskResponse) {
 	}
 
 	tr.lock.Lock()
+	defer tr.lock.Unlock()
```
stephanos, line 162: There's a race causing the `should not move backwards ` assertion failure if the unlock happens before the call to `updateAckLevelAndCount`.

service/matching/physical_task_queue_manager_test.go
```diff
@@ -287,284 +274,245 @@ func TestLegacyDescribeTaskQueue(t *testing.T) {
...
+
+	s.EventuallyWithT(func(collect *assert.CollectT) {
+		assert.Equal(collect, 1, tm.getUpdateCount(s.physicalTaskQueueKey))
+	}, 5*time.Second, 100*time.Millisecond)
 }
 
-func TestTQMInterruptsPollOnClose(t *testing.T) {
-	t.Parallel()
+func (s *PhysicalTaskQueueManagerTestSuite) TestTQMInterruptsPollOnClose() {
+	if s.newMatcher {
```
dnr, line 415: I think this should work too? I want to avoid a regression here

service/matching/pri_task_reader.go
dnr: I see, but we can't do io inside tr.lock, that's very deliberate. we should pass in the ack level in from maybeGCLocked

service/matching/task_queue_partition_manager_test.go
```diff
@@ -61,37 +63,57 @@ const (
...
 	pm.Start()
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
 	defer cancel()
 	err = pm.WaitUntilInitialized(ctx)
-	s.Assert().NoError(err)
+	s.NoError(err)
+}
+
+func (s *PartitionManagerTestSuite) TearDownTest() {
+	s.controller.Finish()
```
dnr, line 116: not needed

service/matching/user_data_manager_test.go
dnr: hmm, I feel like there's way too many of these test-related structs floating around, we should (eventually) consolidate them

service/matching/pri_task_reader.go
dnr: Oh, duh, actually it doesn't matter if we take db lock inside taskReader lock or not, since we do take it in this function. So maybe we can leave this here.

I do want to fix this later. Also the new pri- implementations do have one regression from the old ones in that the stats+status api calls block on db.lock, where before they just read atomics. I don't think that's a big deal but it should also be fixed eventually.

service/matching/physical_task_queue_manager.go
stephanos: Good call; the type was still there 👍

service/matching/pri_task_reader.go
stephanos: Let me try that 👍

service/matching/user_data_manager_test.go
stephanos: Yeah, I tried my best consolidating what I encountered. If this was a test suite, we could just merge the struct into that. But I considered this file out of scope for this PR.

service/matching/db.go
```diff
@@ -278,19 +280,22 @@ func (db *taskQueueDB) SyncState(ctx context.Context) error {
...
+func (db *taskQueueDB) updateAckLevelAndCount(subqueue int, newAckLevel int64, delta int64) {
 	db.Lock()
 	defer db.Unlock()
 
 	db.lastChange = time.Now()
-	if ackLevel < db.subqueues[subqueue].AckLevel {
-		db.logger.DPanic("bug: ack level should not move backwards!")
+	dbQueue := db.subqueues[subqueue]
+	if newAckLevel < dbQueue.AckLevel {
+		softassert.Fail(db.logger,
```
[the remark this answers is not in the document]
stephanos: I've considered that, but I thought you might object to the `Sprintf` part being (eagerly) evaluated even if the assertion does not fail.

service/matching/forwarder_test.go
```diff
@@ -305,12 +327,17 @@ func (t *ForwarderTestSuite) TestForwardPollForActivity() {
 	t.Equal(t.partition.TaskQueue().NamespaceId(), request.GetNamespaceId())
 	t.Equal("id1", request.GetPollRequest().GetIdentity())
 	t.Equal(mustParent(t.partition, 20).RpcName(), request.GetPollRequest().GetTaskQueue().GetName())
-	t.Equal(t.fwdr.partition.Kind(), request.GetPollRequest().GetTaskQueue().GetKind())
+	t.Equal(t.partition.Kind(), request.GetPollRequest().GetTaskQueue().GetKind())
 	t.Equal(resp, task.pollActivityTaskQueueResponse())
 	t.Nil(task.pollWorkflowTaskQueueResponse())
 }
 
 func (t *ForwarderTestSuite) TestMaxOutstandingConcurrency() {
```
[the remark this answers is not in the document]
stephanos: You mean only this one or all of them that contain `t.T().Skip("priority forwarder is not compatible with this test")`?

service/matching/physical_task_queue_manager_test.go
```diff
@@ -287,284 +274,245 @@ func TestLegacyDescribeTaskQueue(t *testing.T) {
...
-
-	controller := gomock.NewController(t)
-
-	cfg := NewConfig(dynamicconfig.NewNoopCollection())
-	cfg.MaxTaskQueueIdleTime = dynamicconfig.GetDurationPropertyFnFilteredByTaskQueue(1 * time.Second)
-	tqCfg := defaultTqmTestOpts(controller)
-	tqCfg.config = cfg
+func (s *PhysicalTaskQueueManagerTestSuite) TestTQMDoesFinalUpdateOnIdleUnload() {
+	if s.newMatcher {
+		s.T().Skip("not supported by new matcher")
```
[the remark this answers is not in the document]
stephanos: Technically yes, but it's very flaky. Same as `TestTQMDoesNotDoFinalUpdateOnOwnershipLost`. I didn't find out why yet. They also occasionally fail for the current matcher. That's why I tried using `EventuallyWithT`, but it might not have helped.

service/matching/matching_engine_test.go
```diff
@@ -1979,55 +1869,52 @@ func (s *matchingEngineSuite) TestMultipleEnginesWorkflowTasksRangeStealing() {
...
+	s.NotEqual(task1.event.GetTaskId(), task2.event.GetTaskId(), "IDs should not match")
 
 	task2.finish(nil, true)
 	s.EqualValues(0, s.taskManager.getTaskCount(dbq))
 }
 
 // TODO: should be moved to backlog_manager_test
 func (s *matchingEngineSuite) TestTaskQueueManagerGetTaskBatch() {
+	if s.newMatcher {
+		s.T().Skip("not supported by new matcher")
```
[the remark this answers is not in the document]
stephanos: Wasn't clear to me how to make it work for both impls.

Screenshot 2025-03-18 at 8 47 10 AM

service/matching/matching_engine_test.go
```diff
@@ -1979,55 +1869,52 @@ func (s *matchingEngineSuite) TestMultipleEnginesWorkflowTasksRangeStealing() {
 }
 
 func (s *matchingEngineSuite) TestAddTaskAfterStartFailure() {
+	if s.newMatcher {
+		s.T().Skip("not supported by new matcher")
```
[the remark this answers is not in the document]
stephanos: It's giving me:

```
 matching_engine_test.go:1901: 
        	Error Trace:	/Users/stephan/Workspace/default/temporal/service/matching/matching_engine_test.go:1901
        	Error:      	Should not be: 1
        	Test:       	TestMatchingEngineWithNewMatcherSuite/TestAddTaskAfterStartFailure
        	Messages:   	IDs should not match
```

I excluded any of the tests I couldn't immediately fix to timebox the PR.

service/matching/matching_engine_test.go
```diff
@@ -2124,67 +2011,34 @@ func (s *matchingEngineSuite) TestTaskQueueManagerGetTaskBatch() {
...
-	s.NoError(err)
-}
-
 func (s *matchingEngineSuite) TestTaskQueueManager_CyclingBehavior() {
-	namespaceId := uuid.New()
-	tl := "makeToast"
-	dbq := newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)
-	config := defaultTestConfig()
+	if s.newMatcher {
+		s.T().Skip("not supported by new matcher")
```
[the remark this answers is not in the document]
stephanos: This one is giving me

```
 matching_engine_test.go:2027: 
        	Error Trace:	/Users/stephan/Workspace/default/temporal/service/matching/matching_engine_test.go:2027
        	Error:      	"2" is not less than or equal to "1"
        	Test:       	TestMatchingEngineWithNewMatcherSuite/TestTaskQueueManager_CyclingBehavior
```

service/matching/matching_engine_test.go
```diff
@@ -3280,74 +3133,71 @@ func (s *matchingEngineSuite) resetBacklogCounter(numWorkers int, taskCount int,
...
 
 // TestResettingBacklogCounter tests the scenario where approximateBacklogCounter over-counts and resets it accordingly
 func (s *matchingEngineSuite) TestResetBacklogCounterNoDBErrors() {
-
 	s.resetBacklogCounter(2, 2, 2)
 }
 
 func (s *matchingEngineSuite) TestResetBacklogCounterDBErrors() {
+	if s.newMatcher {
+		s.T().Skip("not supported by new matcher")
```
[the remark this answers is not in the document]
stephanos: For this one I'm consistently seeing

```
Error Trace:	/Users/stephan/Workspace/default/temporal/service/matching/matching_engine_test.go:3143
        	            				/Users/stephan/.asdf/installs/golang/1.23.2/packages/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.6.darwin-arm64/src/runtime/asm_arm64.s:1223
        	Error:      	Not equal: 
        	            	expected: 0
        	            	actual  : 1
    matching_engine_test.go:3142: 
        	Error Trace:	/Users/stephan/Workspace/default/temporal/service/matching/matching_engine_test.go:3142
        	            				/Users/stephan/Workspace/default/temporal/service/matching/matching_engine_test.go:3155
        	Error:      	Condition never satisfied
        	Test:       	TestMatchingEngineWithNewMatcherSuite/TestResetBacklogCounterDBErrors
        	Messages:   	backlog counter should have been reset
```

service/matching/pri_task_reader.go
stephanos: I was completely with you on your first comment; but the second one I can't quite follow. Why doesn't it actually matter if we still hold the taskReader lock while also holding the db lock?

service/matching/physical_task_queue_manager_test.go
```diff
@@ -287,284 +272,237 @@ func TestLegacyDescribeTaskQueue(t *testing.T) {
...
+	}, 5*time.Second, 100*time.Millisecond)
 }
 
-func TestTQMInterruptsPollOnClose(t *testing.T) {
-	t.Parallel()
-
-	controller := gomock.NewController(t)
-	tqm := mustCreateTestPhysicalTaskQueueManager(t, controller)
-	tqm.Start()
+func (s *PhysicalTaskQueueManagerTestSuite) TestTQMInterruptsPollOnClose() {
```
stephanos, line 408: After re-enabling this test, I sometimes see this (odd!) data race

```
WARNING: DATA RACE
Write at 0x00c002187018 by goroutine 78959:
  go.temporal.io/server/service/matching.(*PhysicalTaskQueueManagerTestSuite).SetupTest()
      /Users/stephan/Workspace/default/temporal/service/matching/physical_task_queue_manager_test.go:100 +0x3a0
  github.com/stretchr/testify/suite.Run.func1()
      /Users/stephan/.asdf/installs/golang/1.23.2/packages/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:192 +0x47c
  testing.tRunner()
      /Users/stephan/.asdf/installs/golang/1.23.2/packages/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.6.darwin-arm64/src/testing/testing.go:1690 +0x274
  testing.(*T).Run.gowrap1()
      /Users/stephan/.asdf/installs/golang/1.23.2/packages/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.6.darwin-arm64/src/testing/testing.go:1743 +0x4c

Previous read at 0x00c002187018 by goroutine 78958:
  go.temporal.io/server/service/matching.(*PhysicalTaskQueueManagerTestSuite).TestTQMDoesNotDoFinalUpdateOnOwnershipLost.func1()
      /Users/stephan/Workspace/default/temporal/service/matching/physical_task_queue_manager_test.go:404 +0x50
  github.com/stretchr/testify/assert.EventuallyWithT.func1()
      /Users/stephan/.asdf/installs/golang/1.23.2/packages/pkg/mod/github.com/stretchr/testify@v1.10.0/assert/assertions.go:2044 +0x14c
```

stephanos: Especially odd that it references `TestTQMDoesNotDoFinalUpdateOnOwnershipLost`, not itself. So that implies that the `SetupTest` and `TestTQMDoesNotDoFinalUpdateOnOwnershipLost` are running concurrently :-?

dnr: hmm. I think I remember reading some issues in the testify/suite repo about goroutine/threading issues with Eventually. but that is weird, there are no explicit goroutines here, the Eventually must return before the test finishes, and we defer s.tqMgr.Stop() so everything in the background should be stopped.

I assume SetupTest creates a new taskManager? yeah, it's in createTestMatchingEngine. so I don't have an explanation yet

service/matching/pri_task_reader.go
```diff
@@ -510,7 +510,7 @@ func (tr *priTaskReader) shouldGCLocked() bool {
 }
 
 // called in new goroutine
-func (tr *priTaskReader) doGC() {
+func (tr *priTaskReader) doGC(ackLevel int64) {
```
dnr, line 513: have to actually use this in the code below and move the db call outside the lock?

service/matching/matching_engine_test.go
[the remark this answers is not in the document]
dnr: agreed, let's aim to merge it without significant changes and we can tackle these individually later

service/matching/backlog_manager_test.go
```diff
@@ -91,191 +127,162 @@ func TestReadLevelForAllExpiredTasksInBatch(t *testing.T) {
...
-func TestAddTasksValidateBacklogCounter_ServiceError(t *testing.T) {
-	controller := gomock.NewController(t)
-	backlogMgr := newBacklogMgr(t, controller, true)
-
-	// mock error signals
-	logger := backlogMgr.logger.(*log.MockLogger) // nolint:revive
-	logger.EXPECT().Error("Persistent store operation failure", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+func (s *BacklogManagerTestSuite) TestAddTasksValidateBacklogCounter_ServiceError() {
+	s.logger.Expect(testlogger.Error, "Persistent store operation failure")
+	s.taskMgr.dbServiceError = true
```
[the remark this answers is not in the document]
stephanos: I don't think we do; I'll add it.

service/matching/pri_task_reader.go
dnr: In general, if they can be independent they should be. But I agree in this case it doesn't really matter.

I think the db lock should be reworked, maybe including this one too. There's two things I would like to be true:

1. tr.completeTask (which is in the poll path) never has to wait for db ops.
2. The stats/status calls  (tr.getLevels, tr.getLoadedTasks, db.getApproximateBacklogCount) never have to wait for db ops.

I think 2 is mostly true on main but regressed on the branch. I think that's okay though, they're just informative calls. 1 isn't true now but it should be, and with all the refactoring, we're closer to making it true.

service/matching/db.go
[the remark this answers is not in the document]
dnr: Oh yeah, good point. Once we're confident in the fix we could remove the printf?

service/matching/backlog_manager_test.go
stephanos: I ported some of the tests now; and skipped the ones that cannot be fixed easily.

dnr (approved):
LG. I think we should merge this now and try to fix the discrepancies later.

PR #7470, opened 2025-03-14
Prevent taskIdBlock data race

stephanos:
## What changed?

Prevent data race when accessing task writer's `taskIdBlock`.

### DIFF
--- a/service/matching/pri_task_writer.go
+++ b/service/matching/pri_task_writer.go
@@ -26,6 +26,7 @@ package matching
 
 import (
 	"context"
+	"sync/atomic"
 	"time"
 
 	enumspb "go.temporal.io/api/enums/v1"
@@ -52,12 +53,13 @@ type (
 
 	// priTaskWriter writes tasks persistence split among subqueues
 	priTaskWriter struct {
-		backlogMgr  *priBacklogManagerImpl
-		config      *taskQueueConfig
-		db          *taskQueueDB
-		logger      log.Logger
-		appendCh    chan *writeTaskRequest
-		taskIDBlock taskIDBlock
+		backlogMgr      *priBacklogManagerImpl
+		config          *taskQueueConfig
+		db              *taskQueueDB
+		logger          log.Logger
+		appendCh        chan *writeTaskRequest
+		taskIDBlock     taskIDBlock
+		lastTaskIDBlock taskIDBlock // copy of the last taskIDBlock for safe concurrent access via currentTaskIDBlock()
 	}
 )
 
@@ -91,7 +93,6 @@ func (w *priTaskWriter) appendTask(
 	subqueue int,
 	taskInfo *persistencespb.TaskInfo,
 ) error {
-
 	select {
 	case <-w.backlogMgr.tqCtx.Done():
 		return errShutdown
@@ -174,6 +175,7 @@ func (w *priTaskWriter) initState() error {
 		return err
 	}
 	w.taskIDBlock = rangeIDToTaskIDBlock(state.rangeID, w.config.RangeSize)
+	w.lastTaskIDBlock = w.taskIDBlock
 	w.backlogMgr.initState(state, nil)
 	return nil
 }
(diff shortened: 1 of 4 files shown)


stephanos:
> What issue could this data race have caused?

Stale data and failed test runs.

> Could it have caused stuck workflow/pending activities?

No.

service/matching/pri_task_writer.go
```diff
@@ -26,6 +26,7 @@ package matching
 
 import (
 	"context"
+	"sync/atomic"
```
stephanos, line 29: I was using a mutex at first, but atomics were simpler in the end.

service/matching/pri_task_writer.go
```diff
@@ -249,3 +254,10 @@ func (w *priTaskWriter) allocTaskIDBlock(prevBlockEnd int64) (taskIDBlock, error
 	}
 	return rangeIDToTaskIDBlock(state.rangeID, w.config.RangeSize), nil
 }
+
+func (w *priTaskWriter) currentTaskIDBlock() taskIDBlock {
+	return taskIDBlock{
+		start: atomic.LoadInt64(&w.lastTaskIDBlock.start),
+		end:   atomic.LoadInt64(&w.lastTaskIDBlock.end),
+	}
```
stephanos, line 262: This is un-used (CodeCov tells me), but I think https://github.com/temporalio/temporal/pull/7469 might change that.

service/matching/task_writer.go
```diff
@@ -248,3 +254,11 @@ func (w *taskWriter) allocTaskIDBlock(prevBlockEnd int64) (taskIDBlock, error) {
 	}
 	return rangeIDToTaskIDBlock(state.rangeID, w.config.RangeSize), nil
 }
+
+// getTaskIDBlock returns the latest taskIDBlock. Safe to be called concurrently.
+func (w *taskWriter) getTaskIDBlock() taskIDBlock {
```
dnr, line 259: the names don't match between old/pri.. should we name this currentTaskIDBlock, or rename the priTaskWriter version to getTaskIDBlock?

service/matching/pri_task_writer.go
```diff
@@ -52,12 +53,13 @@ type (
...
-		logger      log.Logger
-		appendCh    chan *writeTaskRequest
-		taskIDBlock taskIDBlock
+		backlogMgr      *priBacklogManagerImpl
+		config          *taskQueueConfig
+		db              *taskQueueDB
+		logger          log.Logger
+		appendCh        chan *writeTaskRequest
+		taskIDBlock     taskIDBlock
+		lastTaskIDBlock taskIDBlock // copy of the last taskIDBlock for safe concurrent access via currentTaskIDBlock()
```
dnr, line 62: last/current naming is confusing, can we use the same name for the data and the accessor except with a "get"?

service/matching/pri_task_writer.go
```diff
@@ -185,6 +187,9 @@ func (w *priTaskWriter) taskWriterLoop() {
 
 	var reqs []*writeTaskRequest
 	for {
+		atomic.StoreInt64(&w.lastTaskIDBlock.start, w.taskIDBlock.start)
```
dnr, line 190: we could use an atomic.Pointer to a taskIDBlock to eliminate the possibility of skew on read, but it's already better than before, and this is just informative, so it's fine with me.

if you think about it, the `end` is already redundant info, it can always be computed from the start. so we could refactor everything to use one value. but we don't need to bother with that now.

service/matching/task_writer.go
stephanos: Good catch; I went back and forth on some of the naming. 👍

dnr (approved):

service/matching/pri_task_writer.go
stephanos: Oh, right; using `RangeSize`? I see it's a static value (the only one in the config, it seems). That would be simper (assuming we don't make that dynamically configurable in the future).

PR #7472, opened 2025-03-14
Log only unexpected task processing error

yycptt:
## What changed?

- Log only unexpected task processing error

## Why?

- Prevent log spamming when task is keep failing with expected errors like resource exhausted.
- For expected errors, we still have metrics emitted.

## How did you test it?

- Existing tests

### DIFF
--- a/service/history/queues/executable.go
+++ b/service/history/queues/executable.go
@@ -524,41 +524,17 @@ func (e *executableImpl) HandleErr(err error) (retErr error) {
 			// if err is due to workflow busy or APS limit, do not take any latency related to this attempt into account
 			e.inMemoryNoUserLatency += e.scheduleLatency + e.attemptNoUserLatency
 		}
-
-		if retErr != nil {
-			e.Lock()
-			defer e.Unlock()
-
-			e.attempt++
-			if e.attempt > taskCriticalLogMetricAttempts {
-				metrics.TaskAttempt.With(e.metricsHandler).Record(int64(e.attempt))
-				e.logger.Error("Critical error processing task, retrying.",
-					tag.Attempt(int32(e.attempt)),
-					tag.UnexpectedErrorAttempts(int32(e.unexpectedErrorAttempts)),
-					tag.Error(err),
-					tag.OperationCritical)
-			}
-		}
 	}()
 
-	if len(e.dlqErrorPattern()) > 0 {
-		match, mErr := regexp.MatchString(e.dlqErrorPattern(), err.Error())
-		if mErr != nil {
-			e.logger.Error(fmt.Sprintf("Failed to match task processing error with %s", dynamicconfig.HistoryTaskDLQErrorPattern.Key()))
-		} else if match {
-			e.logger.Error(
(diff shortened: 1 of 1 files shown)


service/history/queues/executable.go
```diff
@@ -524,51 +524,38 @@ func (e *executableImpl) HandleErr(err error) (retErr error) {
...
-			return fmt.Errorf("%w: %v", ErrTerminalTaskFailure, err)
-		}
-	}
+	err = e.convertDLQErrorPattern(err)
 
 	if e.isSafeToDropError(err) {
 		return nil
 	}
 
+	attempt := e.incAttempt()
```
pdoerner, line 535: This changes the logic so the attempt is not incremented when the error is safe to drop, but shouldn't those attempts still be counted?

service/history/queues/executable.go
```diff
@@ -524,51 +524,38 @@ func (e *executableImpl) HandleErr(err error) (retErr error) {
...
-				fmt.Sprintf("Error matches with %s. Marking task as terminally failed, will send to DLQ",
-					dynamicconfig.HistoryTaskDLQErrorPattern.Key()),
-				tag.Error(err),
-				tag.ErrorType(err))
-			e.terminalFailureCause = err
-			metrics.TaskTerminalFailures.With(e.metricsHandler).Record(1)
-			return fmt.Errorf("%w: %v", ErrTerminalTaskFailure, err)
-		}
-	}
+	err = e.convertDLQErrorPattern(err)
```
pdoerner, line 529: This removes the early return for DLQ errors. Will the below error handling deal with these errors correctly? Do we have test coverage for that?

service/history/queues/executable.go
yycptt: Attempt is only incremented when returned error is not nil. L528 in the old logic.

service/history/queues/executable.go
yycptt: hmm good catch, it will cause some log/metric to be double emitted. We have existing test coverage for making sure task will be properly dlqed when the pattern is match. 

Let me fix the log/metric issue.

yycptt: hmm I bring the early return back, it seems to be the easiest way to fix it.

service/history/queues/executable.go
```diff
@@ -524,51 +524,38 @@ func (e *executableImpl) HandleErr(err error) (retErr error) {
...
-			e.logger.Error(
-				fmt.Sprintf("Error matches with %s. Marking task as terminally failed, will send to DLQ",
-					dynamicconfig.HistoryTaskDLQErrorPattern.Key()),
-				tag.Error(err),
-				tag.ErrorType(err))
-			e.terminalFailureCause = err
-			metrics.TaskTerminalFailures.With(e.metricsHandler).Record(1)
-			return fmt.Errorf("%w: %v", ErrTerminalTaskFailure, err)
-		}
+	if matchedErr := e.matchDLQErrorPattern(err); matchedErr != nil {
```
yycptt, line 529: This internally emits a log, so the logging below isn't really necessary. 

We don't log "Fail to process task" before the change as well.

pdoerner (approved):

PR #7473, opened 2025-03-17
versioning flakes pt2

Shivs11:
## What changed?

fixed some code to reduce versioning flakes, mainly:
[TestVersioning3FunctionalSuite/TestUnpinnedWorkflowWithRamp_ToUnversioned](https://github.com/temporalio/temporal/actions/runs/13818885947/job/38659266639) (with_logger data-race ✅)

[TestWorkerDeploymentSuite/TestSetWorkerDeploymentRampingVersion_Unversioned_VersionedCurrent](https://github.com/temporalio/temporal/actions/runs/13818885947/job/38659261367) ( context timeout - maybe fixed by having a separate go-routine ✅ )

[TestDeploymentVersionSuite/TestDrainageStatus_SetCurrentVersion_YesOpenWFs](https://github.com/temporalio/temporal/actions/runs/13818885947/job/38659258964)  (with_logger data-race ✅)

[TestVersioningFunctionalSuite/TestDispatchActivityFailCrossTq (retry 2)](https://github.com/temporalio/temporal/actions/runs/13712084173/job/38350437309) (with_logger data-race ✅)

## Why?

- To stop these from appearing as flakes in the test reports.

## How did you test it?

- Ran this locally with `-race` and now see no errors (errored out previously)
- Ran this on CI a bunch of times without errors/flakes

### DIFF
--- a/common/log/with_logger.go
+++ b/common/log/with_logger.go
@@ -49,7 +49,11 @@ func newWithLogger(logger Logger, tags ...tag.Tag) *withLogger {
 }
 
 func (l *withLogger) prependTags(tags []tag.Tag) []tag.Tag {
-	return append(l.tags, tags...)
+	allTags := make([]tag.Tag, len(l.tags)+len(tags))
+	copy(allTags, l.tags)
+	copy(allTags[len(l.tags):], tags)
+
+	return allTags
 }
 
 // Debug writes message to the log (if enabled).
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -479,7 +479,6 @@ func (s *Versioning3Suite) TestUnpinnedWorkflowWithRamp_ToUnversioned() {
 	)
 }
 
-// testing.go:1399: race detected during execution of test
 func (s *Versioning3Suite) testUnpinnedWorkflowWithRamp(toUnversioned bool) {
 	// This test sets a 50% ramp and runs 50 wfs and ensures both versions got some wf and
 	// activity tasks.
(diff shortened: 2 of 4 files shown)


stephanos (approved):

common/log/with_logger.go
```diff
@@ -49,7 +49,11 @@ func newWithLogger(logger Logger, tags ...tag.Tag) *withLogger {
 }
 
 func (l *withLogger) prependTags(tags []tag.Tag) []tag.Tag {
-	return append(l.tags, tags...)
+	allTags := make([]tag.Tag, len(l.tags)+len(tags))
```
Shivs11, line 52: pasting this just for a reader in the future: this function was being accessed by multiple go-routines and was resulting in a race condition. `append` becomes a read+write operation when the underlying array for the slice has it's length reached the slice's max-capacity. This results in a new slice being made (hence a write operation) leading to potential data-races.

For a better, more detailed explanation:

common/log/with_logger.go
```diff
@@ -49,7 +49,11 @@ func newWithLogger(logger Logger, tags ...tag.Tag) *withLogger {
...
+	copy(allTags, l.tags)
+	copy(allTags[len(l.tags):], tags)
```
yiminc, line 54: Is it possible that the l.tags changed (like expanded) after allTags is created, will that cause trouble as allTags may not be big enough to hold l.tags?

Shivs11: Looking at the code now, I don't think so - the reason being, this change effectively removes appending to `l.tags` without making a new `withLogger` instance - in other words, `l.tags` will grow when someone adds new tags to it which happens in `WithLogger.With`. However, `WithLogger.With` returns a new instance of the logger which shall have a new instance of `l.tags`

PR #7476, opened 2025-03-17
Add internal logging to the mask interceptor

ychebotarev:

### DIFF
--- a/common/rpc/interceptor/mask_internal_error.go
+++ b/common/rpc/interceptor/mask_internal_error.go
@@ -30,8 +30,13 @@ import (
 
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/api"
 	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/rpc/interceptor/logtags"
+	"go.temporal.io/server/common/tasktoken"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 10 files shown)


common/rpc/interceptor/service_error_interceptor_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Message string
+	}
+)
+
+func (e *ErrorWithoutStatus) Error() string {
+	return e.Message
+}
+
+// Error returns string message.
+func TestServiceErrorInterceptorUnknown(t *testing.T) {
```
yycptt, line 48: nit: test serialization error as well

common/rpc/interceptor/telemetry.go
```diff
@@ -132,9 +132,9 @@ func NewTelemetryInterceptor(
 	}
 }
 
-// Use this method to override scope used for reporting a metric.
+// TelemetryUnaryOverrideOperationTag is used to override scope used for reporting a metric.
 // Ideally this method should never be used.
-func (ti *TelemetryInterceptor) unaryOverrideOperationTag(fullName, operation string, req any) string {
+func TelemetryUnaryOverrideOperationTag(fullName, operation string, req any) string {
```
yycptt, line 137: why those methods need to be exported?

common/rpc/interceptor/telemetry.go
yycptt, line 1: The hashing logic in the telemetry interceptor can be removed I think.

common/rpc/interceptor/mask_internal_error.go
```diff
@@ -42,51 +47,90 @@ var errorFrontendMasked = "something went wrong, please retry"
...
+	} else {
+		logTags = []tag.Tag{tag.Operation(overridedMethodName), tag.WorkflowNamespace(nsName.String())}
+	}
+
+	logTags = append(logTags, tag.NewStringTag("hash", errorHash))
+
+	logTags = append(logTags, tag.NewStringTag("grpc_code", statusCode.String()))
+	logTags = append(logTags, mi.workflowTags.Extract(req, fullMethod)...)
+
+	mi.logger.Error("service failures", append(logTags, tag.Error(err))...)
```
yycptt, line 135: Can we use a different error message here? so it would be easy to know if the log is from telemetry interceptor or mask error interceptor.

common/rpc/interceptor/telemetry.go
ychebotarev: ...because I keep forgetting those exporting rules by capitalization...
They don't. I need to reuse them in mask interceptor, but mask interceptor and telemetry are in the same package.

common/rpc/interceptor/telemetry.go
ychebotarev: removed

common/rpc/interceptor/mask_internal_error.go
ychebotarev: there will be a hash tag to differentiate. 
But yeah, just a filter will be nice. I will change to "masked service failures"

common/rpc/interceptor/service_error_interceptor_test.go
ychebotarev: added

yycptt (approved):

PR #7477, opened 2025-03-17
CHASM: Snapshot implementation

yycptt:
## What changed?

- Implement chasmTree.Snapshot method
- Move workflow.CompareVersionedTransition to transitionhistory package to avoid cycle dependency. 

## How did you test it?

- Unit test

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -186,7 +186,7 @@ func (n *Node) CloseTransaction() (NodesMutation, error) {
 }
 
 // Snapshot returns all nodes in the tree that have been modified after the given min versioned transition.
-// A nil minVT will be treated as the same as the zero versioned transition and returns all nodes in the tree.
+// A nil exclusiveMinVT will be treated as the same as the zero versioned transition and returns all nodes in the tree.
 // This method should only be invoked on root CHASM node when IsDirty() is false.
 func (n *Node) Snapshot(
 	exclusiveMinVT *persistencespb.VersionedTransition,
@@ -194,6 +194,8 @@ func (n *Node) Snapshot(
 	if !softassert.That(n.logger, n.parent == nil, "chasm.Snapshot() should only be called on the root node") {
 		panic(fmt.Sprintf("chasm.Snapshot() called on child node: %+v", n))
 	}
+
+	// TODO: add assertion on IsDirty() once implemented
 
 	nodes := make(map[string]*persistencespb.ChasmNode)
 	n.snapshotInternal(exclusiveMinVT, []string{}, nodes)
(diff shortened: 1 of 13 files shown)


chasm/tree.go
```diff
@@ -180,11 +187,52 @@ func (n *Node) CloseTransaction() (NodesMutation, error) {
...
+	currentPath []string,
+	nodes map[string]*persistencespb.ChasmNode,
+) error {
+	if n == nil {
+		return nil
+	}
+
+	if transitionhistory.Compare(n.persistence.LastUpdateVersionedTransition, exclusiveMinVT) > 0 {
+		encodedPath, err := n.pathEncoder.Encode(n, currentPath)
+		if !softassert.That(n.logger, err == nil, "chasm path encoding should always succeed on clean tree") {
```
yycptt, line 219: I could return an error here and update all the callsite, but feels unnecessary as it should never happen. Tree is clean means we have already encoded & persisted all the paths.

chasm/tree.go
```diff
@@ -180,11 +187,46 @@ func (n *Node) CloseTransaction() (NodesMutation, error) {
 
 // Snapshot returns all nodes in the tree that have been modified after the given min versioned transition.
 // A nil minVT will be treated as the same as the zero versioned transition and returns all nodes in the tree.
```
lina-temporal, line 189:
```suggestion
// A nil exclusiveMinVT will be treated as the same as the zero versioned transition and returns all nodes in the tree.
```

chasm/tree.go
```diff
@@ -180,11 +187,46 @@ func (n *Node) CloseTransaction() (NodesMutation, error) {
...
+// This method should only be invoked on root CHASM node when IsDirty() is false.
 func (n *Node) Snapshot(
-	minVT *persistencespb.VersionedTransition,
+	exclusiveMinVT *persistencespb.VersionedTransition,
 ) NodesSnapshot {
-	panic("not implemented")
+	if !softassert.That(n.logger, n.parent == nil, "chasm.Snapshot() should only be called on the root node") {
+		panic(fmt.Sprintf("chasm.Snapshot() called on child node: %+v", n))
+	}
+
```
lina-temporal, line 197: Should we also `softassert` on `IsDirty() == false`?

lina-temporal (approved):
Two small suggestions, but LGTM!

chasm/tree.go
yycptt: Make sense, however IsDirty() is not implemented right now. I will wait for Alex to land his PRs first before adding the assertion.

PR #7483, opened 2025-03-18
Add grpc keep alive server and client config for server and client initialization

xwduan:
## What changed?

1. Add grpc keepalive config in the yaml config.
## Why?

To support configuring grpc keep alive for service and client.


service/history/configs/config.go
```diff
@@ -384,6 +384,22 @@ type Config struct {
...
+	// Close the connection if a client is idle.
+	KeepAliveMaxConnectionIdle dynamicconfig.DurationPropertyFn
+	// Close the connection if it is too old.
+	KeepAliveMaxConnectionAge dynamicconfig.DurationPropertyFn
+	// Additive period after MaxConnectionAge after which the connection will be forcibly closed.
+	KeepAliveMaxConnectionAgeGrace dynamicconfig.DurationPropertyFn
+	// Ping the client if it is idle to ensure the connection is still active.
+	KeepAliveTime dynamicconfig.DurationPropertyFn
+	// Wait for the ping ack before assuming the connection is dead.
+	KeepAliveTimeout dynamicconfig.DurationPropertyFn
```
dnr, line 402: Wouldn't it make more sense to put these in static config instead of dynamic config? They can't be changed after server startup, right? And existing RPC-related settings (ports, bind addresses) are already in static config.

xwduan: Yeah, I think it make sense. Shall I create a Config struct in history/service.go similar as this one: https://github.com/temporalio/temporal/blob/9ea4876aae81e62ad8ef9c3249784d20df37c763/service/frontend/service.go#L60?

dnr: That's dynamic config too. I was thinking of putting it in here: https://github.com/temporalio/temporal/blob/main/common/config/config.go#L90

dnr: In general, sure. It's just a yaml file.

If you mean cloud: yes, it's part of the deployment though. I can point it out on slack.

common/resource/fx.go
```diff
@@ -405,6 +405,7 @@ func RPCFactoryProvider(
 	if tracingStatsHandler != nil {
 		options = append(options, grpc.WithStatsHandler(tracingStatsHandler))
 	}
+	options = append(options, grpc.WithKeepaliveParams(svcCfg.RPC.GetKeepAliveClientParameters()))
```
dnr, line 408: As far as I can tell, this affects remote connections also, right? Is that intended? Do we need to differentiate local vs remote?

dnr (commented):
Also, update the PR title since it's not just history server anymore, it's generic to all grpc connections

common/config/config.go
```diff
@@ -107,6 +108,10 @@ type (
 		// forwarded from HTTP to gRPC. Any value with a trailing * will match the prefix before
 		// the asterisk (eg. `x-internal-*`)
 		HTTPAdditionalForwardedHeaders []string `yaml:"httpAdditionalForwardedHeaders"`
+		// KeepAliveServerConfig parameters for the server
```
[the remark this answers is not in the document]
xwduan: Yeah, after reading code more carefully, you are right. And I think it does no make sense to configure this way. So I updated to code for the client config to be used by other service when initialize the client.

common/resource/fx.go
xwduan: I updated the code so we get some flexibility to configure it separately.

client/clientfactory.go
```diff
@@ -236,6 +241,11 @@ func (cf *rpcClientFactory) newFrontendClient(
 	return client
 }
 
+func (cf *rpcClientFactory) getClientDailOption(serviceName primitives.ServiceName) grpc.DialOption {
```
dnr, line 244:
```suggestion
func (cf *rpcClientFactory) getClientDialOption(serviceName primitives.ServiceName) grpc.DialOption {
```

client/clientfactory.go
```diff
@@ -236,6 +241,11 @@ func (cf *rpcClientFactory) newFrontendClient(
...
+	matchingConfig := cf.config.Services[string(serviceName)]
```
dnr, line 245:
```suggestion
	serviceConfig := cf.config.Services[string(serviceName)]
```

client/history/client_test.go
```diff
@@ -185,7 +185,7 @@ func (s *testHistoryService) DeleteDLQTasks(
 	return &historyservice.DeleteDLQTasksResponse{}, nil
 }
 
-func (t *testRPCFactory) CreateInternodeGRPCConnection(rpcAddress string) *grpc.ClientConn {
+func (t *testRPCFactory) CreateInternodeGRPCConnection(rpcAddress string, dialOptions ...grpc.DialOption) *grpc.ClientConn {
 	t.dialedAddresses = append(t.dialedAddresses, rpcAddress)
 	return t.base.CreateInternodeGRPCConnection(rpcAddress)
```
dnr, line 190: I know it doesn't really matter, but shouldn't this pass through `dialOptions...`?

common/config/config.go
```diff
@@ -652,3 +678,78 @@ func (p *JWTKeyProvider) HasSourceURIsConfigured() bool {
...
+	// 0 are treated as default values in grpc: https://github.com/grpc/grpc-go/blob/43a4a84abcd406230e7471062dd2ee5c7ba2485d/internal/transport/http2_server.go#L215-L225
+	// the default config is same as grpc default config, same for the below client config and enforcement policy
+	defaultConfig := keepalive.ServerParameters{
+		MaxConnectionIdle:     0 * time.Minute,
+		MaxConnectionAge:      0,
+		MaxConnectionAgeGrace: 0 * time.Second,
+		Time:                  2 * time.Hour,
+		Timeout:               20 * time.Second,
+	}
+	if rpc.KeepAliveServerConfig == nil {
```
dnr, line 692:
```suggestion
	if rpc == nil || rpc.KeepAliveServerConfig == nil {
```
rpc == nil shouldn't happen, but maybe it makes sense to be defensive and not panic?

common/config/config.go
```diff
@@ -652,3 +678,78 @@ func (p *JWTKeyProvider) HasSourceURIsConfigured() bool {
...
+}
+
+func (rpc *RPC) GetKeepAliveClientParameters() keepalive.ClientParameters {
+	defaultConfig := keepalive.ClientParameters{
+		Time:                0,
+		Timeout:             20 * time.Second,
+		PermitWithoutStream: true,
+	}
+
+	if rpc.KeepAliveClientConfig == nil {
```
dnr, line 720: same for this one

common/config/config.go
```diff
@@ -652,3 +678,78 @@ func (p *JWTKeyProvider) HasSourceURIsConfigured() bool {
...
+	return defaultConfig
+}
+
+func (rpc *RPC) GetKeepAliveEnforcementPolicy() keepalive.EnforcementPolicy {
+	defaultConfig := keepalive.EnforcementPolicy{
+		MinTime:             5 * time.Minute,
+		PermitWithoutStream: false,
+	}
+
+	if rpc.KeepAliveEnforcementPolicy == nil {
```
dnr, line 743: and this one

tests/testcore/onebox.go
```diff
@@ -656,6 +659,20 @@ func (c *TemporalImpl) frontendConfigProvider() *config.Config {
 	}
 }
 
+func (c *TemporalImpl) configProvider(serviceName primitives.ServiceName) func() *config.Config {
+	return func() *config.Config {
+		return &config.Config{
+			Services: map[string]config.Service{
+				string(serviceName): {
+					RPC: config.RPC{
+						HTTPPort: int(mustPortFromAddress(c.FrontendHTTPAddress())),
```
dnr, line 668: this is kind of misleading, only frontend has an http port. is this needed for some kind of validation or can it be removed?

tests/testcore/onebox.go
```diff
@@ -427,6 +426,7 @@ func (c *TemporalImpl) startHistory() {
 				serviceName,
 				c.mockAdminClient,
 			),
+			fx.Provide(c.configProvider(primitives.HistoryService)),
```
dnr, line 429: hah, I didn't realize history/matching/worker had no `*config.Config` in the test cluster fx graph

tests/testcore/onebox.go
```diff
@@ -656,6 +659,20 @@ func (c *TemporalImpl) frontendConfigProvider() *config.Config {
 	}
 }
 
+func (c *TemporalImpl) configProvider(serviceName primitives.ServiceName) func() *config.Config {
+	return func() *config.Config {
+		return &config.Config{
```
dnr, line 664: I'd write this like:
```suggestion
func (c *TemporalImpl) configProvider(serviceName primitives.ServiceName) *config.Config {
	return &config.Config{
```
and then just `fx.Provide(c.configProvider)` above. The fx graph already has a `primitives.ServiceName` corresponding to the service.

common/rpc.go
```diff
@@ -38,9 +38,9 @@ type RPCFactory interface {
 	GetFrontendGRPCServerOptions() ([]grpc.ServerOption, error)
 	GetInternodeGRPCServerOptions() ([]grpc.ServerOption, error)
 	GetGRPCListener() net.Listener
-	CreateRemoteFrontendGRPCConnection(rpcAddress string) *grpc.ClientConn
+	CreateRemoteFrontendGRPCConnection(rpcAddress string, dialOptions ...grpc.DialOption) *grpc.ClientConn
```
dnr, line 41: I have to admit it feels a little weird to have to pass in extra DialOptions here: the RPCFactory is supposed to encapsulate all the options and stuff, and make it easy to just create connections when required. Like, it is the source of the `grpc.ServerOption`s, so it seems like it should be the source of the `grpc.DialOption`s too. And the implementation of this interface has a reference to the static config already, so it should know everything it needs.

There's two problems that I see:

1. It doesn't have a reference to the _whole_ static config, it only has one `config.RPC` section corresponding to the server role. This is easy to fix: just have `common/rpc.NewFactory` take in the whole `*config.Config`, and then pull out the server section that it wants. It already accepts the service name too. Then it can pull out a different section for client options.
2. CreateInternodeGRPCConnection doesn't know if it's creating a connection for matching or history. This is also easy to fix: either have it accept a ServiceName also, or split it into two functions.

With those, I think it should be possible to do all the keepalive stuff in RPCFactory and not touch clientfactory at all. What do you think?

xwduan: I agree.

For 2, I split it to 2 functions.

tests/testcore/onebox.go
```diff
@@ -684,8 +697,16 @@ func (c *TemporalImpl) newRPCFactory(
...
 		options = append(options, grpc.WithStatsHandler(tracingStatsHandler))
 	}
+	rpcConfig := config.RPC{BindOnIP: host, GRPCPort: port, HTTPPort: int(httpPort)}
+	cfg := &config.Config{
+		Services: map[string]config.Service{
+			string(sn): {
+				RPC: rpcConfig,
+			},
+		},
+	}
```
dnr, line 707: I think this logic should just go in `configProvider` (instead of returning an empty struct) and then `newRPCFactory` can just accept a `*config.Config` and pass it through to `rpc.NewFactory`

I don't want to hold up this PR much more so I'll approve now and you or someone else can clean that up in a followup PR if you want

dnr (approved):
Thanks, I like this version a lot better!

tests/testcore/onebox.go
xwduan: Thanks. I will merge it for now and prepare a new pr to address this.

PR #7485, opened 2025-03-19
Ignore updating activity retry policy for backward compatible

xwduan:
## What changed?

Backward compatible support for this commit: https://github.com/temporalio/temporal/commit/c49995f27
## Why?

When replicating between different version of temporal, specifically replicating from version does not ^ above commit and to version has ^ commit, the activity retry policy will be override to 0 value and will cause activity fail after failover. 
## How did you test it?

tested locally with different version of temporal
## Potential risks

no risk.
## Documentation

n/a
## Is hotfix candidate?

yes

### DIFF
--- a/service/history/replication/raw_task_converter.go
+++ b/service/history/replication/raw_task_converter.go
@@ -50,6 +50,7 @@ import (
 	"go.temporal.io/server/service/history/tasks"
 	"go.temporal.io/server/service/history/workflow"
 	wcache "go.temporal.io/server/service/history/workflow/cache"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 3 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -1747,11 +1747,12 @@ func (ms *MutableStateImpl) UpdateActivityInfo(
 	ai.Stamp = incomingActivityInfo.GetStamp()
 
 	ai.Paused = incomingActivityInfo.GetPaused()
-
-	ai.RetryInitialInterval = incomingActivityInfo.GetRetryInitialInterval()
-	ai.RetryMaximumInterval = incomingActivityInfo.GetRetryMaximumInterval()
-	ai.RetryMaximumAttempts = incomingActivityInfo.GetRetryMaximumAttempts()
-	ai.RetryBackoffCoefficient = incomingActivityInfo.GetRetryBackoffCoefficient()
+	if incomingActivityInfo.RetryInitialInterval != nil {
```
yux0, line 1750: Can we assume:
1. If the retry policy is set, the initial interval will not be nil?
2. We always apply the activity info as snapshot?

xwduan: 1. I will add code change on the sender side, when it is nil, make it non-nil, so we can distinguish if they are not set or 0 value
2. This is the only place we update those info.

service/history/replication/raw_task_converter.go
```diff
@@ -170,6 +171,13 @@ func convertActivityStateReplicationTask(
 			if lastStartedBuildId == "" {
 				lastStartedBuildId = activityInfo.GetUseWorkflowBuildIdInfo().GetLastUsedBuildId()
 			}
+			// We will use field to distinguish between nil and zero value for backward compatibility
+			retryInitialInterval := activityInfo.RetryInitialInterval
+			if retryInitialInterval == nil {
+				retryInitialInterval = &durationpb.Duration{
```
ychebotarev, line 177: I'm not sure server treat nil **GetRetryInitialInterval** and 0 the same way.

ychebotarev (approved):

service/history/replication/raw_task_converter.go
xwduan: Duration proto handled nil as 0.

PR #7486, opened 2025-03-19
Add interface for NamespaceRateLimitInterceptor

prathyushpv:
## What changed?
Split NamespaceRateLimitInterceptor to interface and implementation. Also move nettest package out of internal, so that it can be used in other repositories.

## Why?
Splitting NamespaceRateLimitInterceptor to interface will allow us to change this interface type when building fx graph.

## How did you test it?
Existing tests.

## Potential risks
None


common/nettest/listener.go
dnr, line 1: Maybe consider `common/testing/nettest/...`?

prathyushpv: Right! Moved it to common/testing

client/history/client_test.go
```diff
@@ -36,7 +36,7 @@ import (
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/membership"
-	"go.temporal.io/server/internal/nettest"
+	nettest2 "go.temporal.io/server/common/testing/nettest"
```
yycptt, line 39: hmm why it needs to be `nettest2`?

yycptt (approved):

client/history/client_test.go
prathyushpv: Fixed it. That was unintentional.

PR #7489, opened 2025-03-19
fix data race using logger in goroutine

carlydf:
## What changed?
Fix data race using logger in goroutine

## Why?
It was causing tests to flake


dnr (commented):
Isn't this fixed by https://github.com/temporalio/temporal/pull/7473 ?

carlydf:
> Isn't this fixed by #7473 ?

yeah, I didn't realize that until this morning. Closing

PR #7491, opened 2025-03-19
Add event ID to RequestIDInfo

rodrigozhou:
## What changed?

Add event ID to `RequestIDInfo`.
Also added `RequestIdInfos` to `DescribeWorkflowExecutionResponse`.

## Why?

Be to able to find which event was added by the give request ID.

### DIFF
--- a/common/persistence/sql/sqlplugin/tests/history_current_execution.go
+++ b/common/persistence/sql/sqlplugin/tests/history_current_execution.go
@@ -34,6 +34,7 @@ import (
 	enumspb "go.temporal.io/api/enums/v1"
 	enumsspb "go.temporal.io/server/api/enums/v1"
 	persistencespb "go.temporal.io/server/api/persistence/v1"
+	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/persistence/serialization"
 	"go.temporal.io/server/common/persistence/sql/sqlplugin"
 	"go.temporal.io/server/common/primitives"
@@ -342,6 +343,7 @@ func (s *historyCurrentExecutionSuite) newRandomCurrentExecutionRow(
 		RequestIds: map[string]*persistencespb.RequestIDInfo{
 			uuid.NewString(): {
 				EventType: enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED,
+				EventId:   common.BufferedEventID,
 			},
 		},
 	}
--- a/service/history/api/describeworkflow/api.go
+++ b/service/history/api/describeworkflow/api.go
@@ -162,11 +162,23 @@ func Invoke(
 			OriginalStartTime:       startEvent.EventTime,
 			CancelRequested:         executionInfo.CancelRequested,
 			ResetRunId:              executionInfo.ResetRunId,
+			RequestIdInfos:          make(map[string]*workflowpb.RequestIdInfo),
 		},
 	}
 
 	if mutableState.IsResetRun() {
 		result.WorkflowExtendedInfo.LastResetTime = executionState.StartTime
+	}
+
+	for requestID, requestIDInfo := range mutableState.GetExecutionState().GetRequestIds() {
+		info := &workflowpb.RequestIdInfo{
+			EventType: requestIDInfo.EventType,
+			Buffered:  requestIDInfo.EventId == common.BufferedEventID,
+		}
+		if !info.Buffered {
+			info.EventId = requestIDInfo.EventId
+		}
+		result.WorkflowExtendedInfo.RequestIdInfos[requestID] = info
 	}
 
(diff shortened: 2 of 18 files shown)


common/persistence/sql/sqlplugin/tests/history_current_execution.go
```diff
@@ -342,6 +343,7 @@ func (s *historyCurrentExecutionSuite) newRandomCurrentExecutionRow(
 		RequestIds: map[string]*persistencespb.RequestIDInfo{
 			uuid.NewString(): {
 				EventType: enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED,
+				EventId:   common.FirstEventID,
```
pdoerner, line 346: should be `common.BufferedEventID`?

pdoerner (commented):
What happens in the event that the request ID is not populated? It is not strictly required by Nexus. Can we have fallback to link to the WorkflowExecutionStarted event at least?

rodrigozhou:
@pdoerner There's always a request ID, if it's not set by Nexus, it's auto-generated by the SDK.

common/testing/protoassert/assert.go
```diff
@@ -79,7 +81,7 @@ func ProtoSliceEqual[T proto.Message](t assert.TestingT, a []T, b []T) bool {
 	}
 	for i := 0; i < len(a); i++ {
 		if diff := cmp.Diff(a[i], b[i], protocmp.Transform()); diff != "" {
-			return assert.Fail(t, "Proto mismatch at index %d (-want +got):\n%v", i, diff)
+			return assert.Fail(t, fmt.Sprintf("Proto mismatch at index %d (-want +got):\n%v", i, diff))
```
rodrigozhou, line 84: This was a bug in the `protoassert` library.

go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
 	go.temporal.io/api v1.48.1-0.20250426005355-d6c89de822db
-	go.temporal.io/sdk v1.34.0
+	go.temporal.io/sdk v1.34.1-0.20250429051454-eb7575730f9a
```
bergundy, line 63: Are we going to release this SDK version before the next server release? We don't want an untagged SDK version in a server release.

service/history/api/describeworkflow/api.go
```diff
@@ -162,13 +162,25 @@ func Invoke(
...
 	}
 
+	for requestID, requestIDInfo := range mutableState.GetExecutionState().GetRequestIds() {
+		info := &workflowpb.RequestIdInfo{
+			EventType: requestIDInfo.EventType,
+			Buffered:  requestIDInfo.EventId == common.BufferedEventID,
+		}
+		if !info.Buffered {
+			info.EventId = requestIDInfo.EventId
+		}
```
bergundy, line 180: Can we backfill event ID to `1` for `WorkflowExecutionStarted` `EventType`?

bergundy (approved):

bergundy:
Can you also test that `RequestIdToEventId` is populated in MS after CaN?

service/history/api/describeworkflow/api.go
rodrigozhou: For the `WorkflowExecutionStarted`, the request ID info will already have the correct ID from the start.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6769,6 +6788,14 @@ func (ms *MutableStateImpl) updatePendingEventIDs(
 			continue
 		}
 	}
+	if len(requestIDToEventID) > 0 {
+		for requestID, eventID := range requestIDToEventID {
+			if requestIDInfo, ok := ms.executionState.RequestIds[requestID]; ok {
+				requestIDInfo.EventId = eventID
+			}
+		}
+		ms.executionStateUpdated = true
```
yycptt, line 6797: Synced offline. Let's not set this flag when updating RequestIds (here and in `attachRequestID`). 
It is meant to only capture changes of the State/Status field in executionState, otherwise standby side will refresh workflow start related tasks unnecessarily. 

I will send out a follow up PR to rename those fields.

go.mod
rodrigozhou: Removed this change from this PR. But it's needed in the next PR.

rodrigozhou:
Added tests and addressed comments.

service/history/workflow/mutable_state_impl.go
rodrigozhou: Removed

yycptt (approved):

PR #7504, opened 2025-03-20
Batch SyncDeploymentVersionUserData into 25 TQs

Shivs11:
## What changed?

WISOTT

## Why?

- optimization for versioning-3.1

## How did you test it?

- Added unit tests which confirm that batching is happening correctly.
- Added functional tests sending in requests with 50 task-queues to confirm nothing breaks with batching enabled. Since these are functional tests, I verified that the right number of batches were being made by looking at the logs. It wasn't needed since the unit tests covered this but we rather stay safe than be sorry later on.

## Potential risks

- None.

## Is hotfix candidate?

- No


tests/worker_deployment_test.go
```diff
@@ -967,12 +990,91 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_NoCurrent_
...
+
+// SetCurrent tests
+
+func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Batching() {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+	tv := testvars.New(s)
+
+	// register 100 task-queues in the version
+	taskQueues := 100
```
Shivs11, line 1039: 100 task queues were chosen and not a larger number (like 500) as a way to differentiate between a functional test and a load test - locally, I verified things work when run with 500 task-queues. However, a number like 500 would make each of our CI run slower and this is something we don't desire.

tests/worker_deployment_test.go
```diff
@@ -967,12 +990,91 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_NoCurrent_
 	})
 }
 
+func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_Batching() {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+	tv := testvars.New(s)
+
+	// register 50 task-queues in the version
+	taskQueues := 50
```
Shivs11, line 999: 50 task queues were chosen and not a larger number (like 500) as a way to differentiate between a functional test and a load test - locally, I verified things work when run with 500 task-queues. However, a number like 500 made CI much slower and throw errors like `sorry, too many clients already (SQLSTATE 53300)` suggesting that this number was above a current bottleneck for db connections.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -552,45 +552,110 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
+				}).Get(ctx, nil)
+			if err != nil {
+				// TODO (Shivam): Compensation functions required to roll back the local state + activity changes.
+				return nil, err
+			}
+		}
+	} else {
+
+		// send in the task-queue families in batches of syncBatchSize
+		batches := make([][]*deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData, 0)
```
carlydf, line 601: nit: you have enough info here to pre-calculate the # of batches instead of instantiating a zero-length list. Technically that helps with runtime and I think more optimal memory allocation, but unsure how big of a benefit you would see here.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -552,45 +552,110 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
+		for _, tqName := range workflow.DeterministicKeys(state.TaskQueueFamilies) {
+			byType := state.TaskQueueFamilies[tqName]
+			data := &deploymentspb.DeploymentVersionData{
+				Version:           d.VersionState.Version,
+				RoutingUpdateTime: args.RoutingUpdateTime,
+				CurrentSinceTime:  args.CurrentSinceTime,
+				RampingSinceTime:  args.RampingSinceTime,
+				RampPercentage:    args.RampPercentage,
+			}
+			var types []enumspb.TaskQueueType
```
carlydf, line 612: nit: same here (`types` could be initialized as a list with size = `len(byType.TaskQueues)`)

service/worker/workerdeployment/version_workflow.go
```diff
@@ -552,45 +552,110 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
+
+			syncReq.Sync = append(syncReq.Sync, &deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData{
+				Name:  tqName,
+				Types: types,
+				Data:  data,
+			})
+
+			if len(syncReq.Sync) == syncBatchSize {
+				batches = append(batches, syncReq.Sync)
+				syncReq.Sync = make([]*deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData, 0) // reset the syncReq.Sync slice for the next batch
```
carlydf, line 625: When weighing whether to use an empty initial slice + append vs a pre-set initial slice and then index into it, a big thing I consider is readability. Like, pre-allocating a slice is "nice to have" if you get it for free, but not if it adds risk of bugs and hard-to-maintain code.

In some cases when the total slice size is simple to calculate, and the for-loop where you have to keep track of the index is also straightforward, then it's easy to read the optimized pre-allocated way, so I go for it. IMO, the `types` slice above is a good example of this (you're just copying keys from a known-size-map into a same-size slice with a different type).

On the other hand, instantiating an empty `syncReq.Sync` and using append is nicely readable, and I think worrying about trying to calculate the correct size for the next batch could distract from readability of the code and possibly introduce bugs. So I would lean towards the nil slice + append method there.

TLDR, it's up to you how you want to do these list-building loops, just sharing how I think about it / pointing out that you have the option to pre-allocate.

carlydf (approved):

service/worker/workerdeployment/version_workflow.go
Shivs11: Thanks for chipping in on this! I did consider this while writing up the code and initially had a fixed length for the slice along with indexing to add elements to it. However, I chose to step away from it for the following reasons:

- I wanted things to be consistent between the way I was initializing and maintaining `syncReq.Sync`, `batches` as well as `types`. I agree that from an optimization perspective, going with a fixed size is always better but I thought in this case I wanted things to be streamlined between the three data-structures mainly because it would make reading easy.
- With fixed size slicing, we do have to care about an extra variable which maintains the position of the next element to be added. While this is simple, I just thought things would look a bit cluttered and difficult to follow if some slices were using fixed-indexing and others weren't. 
- For a potentially new reader, we already have a lot going on in the `handleSyncState` function which they have to be wary about. 

I think for all of the above reasons, I chose to replace the optimization for these data-structures with easy readability.

PR #7507, opened 2025-03-20
[CHASM] Avoid a double-encode by passing Metadata and Data as separate fields

lina-temporal:
## What changed?
- CHASM nodes are now internally represented as split distinctly between Metadata and Data parts. 
- Additional columns were added to SQL plugins, bumping the SQL schema
- No change was made to the Cassandra columns; instead, ExecutionManager ensures that when running on Cassandra, the entire node continues to be represented in a single blob, to avoid lowering Cass performance post-change.

## Why?
Before, a ChasmNode contained both its metadata fields, and the user data blob. Decoding a ChasmNode would cause an additional (often large, relative to the metadata) allocation to store the inner data blob, as well as additional memcpys while encoding/decoding. With this change, the data field can be passed along as-is, opaquely, while the metadata field can be decoded without additional penalty from a large data blob.

## How did you test it?
- Updated tests


common/persistence/persistence_interface.go
```diff
@@ -508,6 +508,19 @@ type (
 		Checksum *commonpb.DataBlob
 	}
 
+	InternalChasmNode struct {
+		Metadata *commonpb.DataBlob
+		Data     *commonpb.DataBlob
+
+		// Only set when Cassandra is used as the persistence layer. When set, Metadata
+		// and Data will be unset. *No* code outside of the ExecutionManager or Cassandra
+		// plugin should reference this field.
```
yycptt, line 517:
```suggestion
		// and Data will be unset. *No* code outside of the ExecutionManager or Cassandra
		// store should reference this field.
```

yycptt (commented):
LGTM! I am ready to approve after the schema changes are merged into the last version.

yycptt (approved):

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt, line 1: Let's [disable](https://github.com/yycptt/temporal/blob/e2ab6b450065270dec8a004fdeefbbee61cb011a/proto/internal/buf.yaml#L13) the buf breaking check for this file.

PR #7512, opened 2025-03-21
test concurrent setCurrent/setRamping requests 

Shivs11:
## What changed?

- Added some more tests
- [Description: ](https://temporalio.atlassian.net/jira/software/c/projects/OSS/boards/31?assignee=712020:e7840624-ab5e-4cba-b6b6-189c3d835eea&assignee=712020:479e6c4f-7c7d-4472-afdd-5d4ce87fb15a&selectedIssue=OSS-3901)

## Why?

- verifying if the versioning feature can handle concurrent requests
- desired behavior: back-to-back setCurrent/ramping requests with different inputs should not throw weird errors, only throttling or timeout, or successful result.

## How did you test it?

- Ran this suite locally and CI 

## Is hotfix candidate?

- No


tests/worker_deployment_test.go
```diff
@@ -967,6 +936,37 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_NoCurrent_
 	})
 }
 
+func (s *WorkerDeploymentSuite) TestSetCurrent_NoDeploymentVersion() {
```
Shivs11, line 939: just moved this test to an area which had all other `setCurrent` tests - no need to review this.

carlydf (commented):
Have you had a chance to test these new tests for flakiness?

tests/worker_deployment_test.go
```diff
@@ -1021,6 +1021,170 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
...
+				IgnoreMissingTaskQueues: true,
+				Identity:                tv.ClientIdentity(),
+			})
+			errChan <- err
+		}()
+	}
+
+	for i := 0; i < versions; i++ {
+		err := <-errChan
+		s.NoError(err)
```
carlydf, line 1054: I think some of these requests might be expected to get throttling error or context timeout error. If you decrease the # of WorkflowExecutionMaxInFlightUpdates to the default (which is 10), do you see anything weird, or just throttling + timeout errors?

tests/worker_deployment_test.go
```diff
@@ -1021,6 +1021,170 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
 	s.verifyTaskQueueVersioningInfo(ctx, currentVars.TaskQueue(), worker_versioning.UnversionedVersionId, "", 0)
 }
 
+func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Concurrent_DifferentVersions() {
+	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 100)
```
carlydf, line 1025: I think this test should use the default value for this (which is 10). What do you think?

tests/worker_deployment_test.go
carlydf: I think instead of `s.NoError` you could keep the rest of this test the same, and just assert that the error is either:
- nil
- timeout
- throttling

carlydf: just bringing this up since the ticket said "confirm that concurrent updates with different inputs don't result in any weird errors, only throttling or timeout or success"

tests/worker_deployment_test.go
```diff
@@ -1021,6 +1021,170 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
...
+	resp, err := s.FrontendClient().DescribeWorkerDeployment(ctx, &workflowservice.DescribeWorkerDeploymentRequest{
+		Namespace:      s.Namespace().String(),
+		DeploymentName: tv.DeploymentVersion().GetDeploymentName(),
+	})
+	s.NoError(err)
+	s.NotEqual(worker_versioning.UnversionedVersionId, resp.GetWorkerDeploymentInfo().GetRoutingConfig().GetCurrentVersion())
+}
+
+func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Concurrent_SameVersion() {
+	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 100)
```
carlydf, line 1067: same here, we should be testing this with default config

tests/worker_deployment_test.go
```diff
@@ -1021,6 +1021,170 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
...
+				IgnoreMissingTaskQueues: true,
+				Identity:                tv.ClientIdentity(),
+			})
+			errChan <- err
+		}()
+	}
+
+	for i := 0; i < 10; i++ {
+		err := <-errChan
+		s.NoError(err)
```
carlydf, line 1093: same here, throttling or timeout is acceptable

carlydf (commented):
Based on my understanding of the ticket, I think that these tests should not override the MaxInFlightUpdates -- we should test against the default because that's what customers will have. Ideally that will only cause throttling or timeout errors and nothing "weird." If any other error type comes up, that's a bug that should be addressed.

tests/worker_deployment_test.go
Shivs11: I think that there is one important point to consider - when a user sends in a setCurrent request (for example), there are going to be 2 in-flight updates (inflight updates are those updates that have been admitted but not completed). For each set request (1 in-flight update since it's not completed), we have a sync call request (another in-flight update). Thus, for 10 setCurrent requests, we have to have the limit set to 20. So, setting the limit to 10 shall give you `ResourceExhaustedErrors`.

Looking at the ticket again, I took a step back and tried understanding what it is asking us to test - It seems like the main motivation behind the ticket is that when we do send in 10 concurrent requests, we should not get any "weird" errors - however, I think we should slightly change what the ticket is asking:

We should be testing that given enough time for our system to process these 10 concurrent updates and are system is ready to handle these 10 concurrent updates, does anything weird at all happen? This makes it easy to catch for things we wanted to catch from the get-go: any weird errors. 

wdyt?

carlydf: I see it the same way as you put it in your second comment! Are you on board with writing the test that way?

tests/worker_deployment_test.go
```diff
@@ -1123,6 +1092,206 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
 	s.verifyTaskQueueVersioningInfo(ctx, currentVars.TaskQueue(), worker_versioning.UnversionedVersionId, "", 0)
 }
 
+func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Concurrent_DifferentVersions_NoUnexpectedErrors() {
+	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10)
```
carlydf, line 1096:
```suggestion
	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10) // this is the default
```

tests/worker_deployment_test.go
```diff
@@ -1123,6 +1092,206 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
...
+	resp, err := s.FrontendClient().DescribeWorkerDeployment(ctx, &workflowservice.DescribeWorkerDeploymentRequest{
+		Namespace:      s.Namespace().String(),
+		DeploymentName: tv.DeploymentVersion().GetDeploymentName(),
+	})
+	s.NoError(err)
+	s.NotEqual(worker_versioning.UnversionedVersionId, resp.GetWorkerDeploymentInfo().GetRoutingConfig().GetCurrentVersion())
+}
+
+func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Concurrent_SameVersion_NoUnexpectedErrors() {
+	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10)
```
carlydf, line 1147:
```suggestion
	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10) // this is the default
```

tests/worker_deployment_test.go
```diff
@@ -1123,6 +1092,206 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
...
+	resp, err := s.FrontendClient().DescribeWorkerDeployment(ctx, &workflowservice.DescribeWorkerDeploymentRequest{
+		Namespace:      s.Namespace().String(),
+		DeploymentName: tv.DeploymentVersion().GetDeploymentName(),
+	})
+	s.NoError(err)
+	s.Equal(tv.DeploymentVersionString(), resp.GetWorkerDeploymentInfo().GetRoutingConfig().GetCurrentVersion())
+}
+
+func (s *WorkerDeploymentSuite) TestSetRampingVersion_Concurrent_DifferentVersions_NoUnexpectedErrors() {
+	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10)
```
carlydf, line 1195:
```suggestion
	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10) // this is the default
```

tests/worker_deployment_test.go
```diff
@@ -1123,6 +1092,206 @@ func (s *WorkerDeploymentSuite) TestSetCurrentVersion_Unversioned_PromoteUnversi
...
+	resp, err := s.FrontendClient().DescribeWorkerDeployment(ctx, &workflowservice.DescribeWorkerDeploymentRequest{
+		Namespace:      s.Namespace().String(),
+		DeploymentName: tv.DeploymentVersion().GetDeploymentName(),
+	})
+	s.NoError(err)
+	s.NotNil(resp.GetWorkerDeploymentInfo().GetRoutingConfig().GetRampingVersion())
+}
+
+func (s *WorkerDeploymentSuite) TestSetRampingVersion_Concurrent_SameVersion_NoUnexpectedErrors() {
+	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10)
```
carlydf, line 1247:
```suggestion
	s.OverrideDynamicConfig(dynamicconfig.WorkflowExecutionMaxInFlightUpdates, 10) // this is the default
```

carlydf (approved):
I think it's worth stating that your override is simply ensuring that the MaxUpdates value is the default for these tests! Other than that looks great

PR #7513, opened 2025-03-21
Testing something. not for merging

prathyushpv:


common/dynamicconfig/memory_client.go
```diff
@@ -71,11 +71,14 @@ func (d *MemoryClient) getValueLocked(key Key) []ConstrainedValue {
 
 func (d *MemoryClient) Subscribe(f ClientUpdateFunc) (cancel func()) {
 	d.lock.Lock()
-	defer d.lock.Unlock()
-
 	d.subscriptionIdx++
 	id := d.subscriptionIdx
 	d.subscriptions[id] = f
+	d.lock.Unlock()
```
dnr, line 77: This change doesn't do anything? It's still unlocking at the same place. `f` is not called here

common/dynamicconfig/memory_client.go
```diff
@@ -90,43 +93,65 @@ func (d *MemoryClient) OverrideSetting(setting GenericSetting, value any) (clean
...
 	idx.Store(int64(len(d.overrides)))
 
 	d.overrides = append(d.overrides, kvpair{valid: true, key: key, value: value})
-
 	newValue := d.getValueLocked(key)
 	changed := map[Key][]ConstrainedValue{key: newValue}
-	for _, update := range d.subscriptions {
-		update(changed)
+
+	// Copy out subscriptions to call after we release the lock
```
dnr, line 104: yup, same fix, just manual code instead of library function

common/dynamicconfig/memory_client.go
```diff
@@ -90,43 +93,65 @@ func (d *MemoryClient) OverrideSetting(setting GenericSetting, value any) (clean
...
 			d.remove(removeIdx)
 		}
 	}
 }
 
 func (d *MemoryClient) remove(idx int) {
 	d.lock.Lock()
-	defer d.lock.Unlock()
+
+	if idx < 0 || idx >= len(d.overrides) || !d.overrides[idx].valid {
```
dnr, line 130: personally I would make this a panic or assert.. if the rest of the logic is correct this should never be the case

common/dynamicconfig/memory_client.go
```diff
@@ -90,43 +93,65 @@ func (d *MemoryClient) OverrideSetting(setting GenericSetting, value any) (clean
...
+		d.lock.Unlock()
+		return
+	}
 
 	key := d.overrides[idx].key
-	// mark this pair deleted
-	d.overrides[idx] = kvpair{}
+	d.overrides[idx].valid = false
```
dnr, line 136: I don't like this change, I deliberately reset the whole struct to drop the reference to the old value. This keeps the reference alive unnecessarily.

PR #7514, opened 2025-03-21
Fix locking bug in dynamicconfig MemoryClient

dnr:
## What changed?
MemoryClient should release its own lock before notifying subscribers.

## Why?
Lock ordering between Collection's subscriptionLock and MemoryClient's lock: they were taken in both orders which could result in a deadlock.

## How did you test it?
Existing tests. Note that MemoryClient is only used in test environments.

### DIFF
--- a/common/dynamicconfig/memory_client.go
+++ b/common/dynamicconfig/memory_client.go
@@ -27,6 +27,8 @@ package dynamicconfig
 import (
 	"sync"
 	"sync/atomic"
+
+	expmaps "golang.org/x/exp/maps"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


stephanos (approved):

yiminc (approved):

common/dynamicconfig/memory_client.go
```diff
@@ -99,7 +100,12 @@ func (d *MemoryClient) OverrideValue(key Key, value any) (cleanup func()) {
 
 	newValue := d.getValueLocked(key)
 	changed := map[Key][]ConstrainedValue{key: newValue}
-	for _, update := range d.subscriptions {
+	subscriptions := expmaps.Values(d.subscriptions)
+
+	d.lock.Unlock()
+
+	// do not hold lock while notifying subscriptions
+	for _, update := range subscriptions {
```
prathyushpv, line 108: Subscriptions not need to be protected from concurrent access?

dnr: this `subscriptions` is a local variable now

Issue #7454, opened 2025-03-12
SQL tool update-schema not working on windows


san127127:
## Expected Behavior
update-schema should work correctly on windows

## Actual Behavior
update-schema doesn't work correctly on windows

## Steps to Reproduce the Problem

  1. Download the temporal-sql-tool and the SQL scripts
  2. Start a postgres DB
  3. Create a DB and setup the schema version tables
      - `.\temporal-sql-tool.exe --ep "127.0.0.1" -p "5432" -u "postgres" -pw "123456" --plugin postgres12 --db temporal setup-schema -v 0.0`
  4. Run `.\temporal-sql-tool.exe --ep "127.0.0.1" -p "5432" -u "postgres" -pw "123456" --plugin postgres12 --db temporal update-schema -d "./postgresql/v12/temporal/versioned"`
  5. Will get the following error:
```
ERROR   Unable to update SQL schema.    {"error": "error processing manifest for version v1.0:readfile v1.0\\manifest.json: invalid argument", "logging-call-at": "/home/runner/work/temporal/temporal/tools/sql/handler.go:77"}
```
## Specifications

  - Version: 1.27.1
  - Platform: windows 11

## Remarks
The problem seems to be due to the use of `filepath.Join()`.
On windows, `filepath.Join()` returns a path with "\\" as the separator like `v1.0\manifest.json` and paths like this don't work with `fs.ReadFile()`.
Here is a code snippet to reproduce the problem:
```
var fsys = os.DirFS("./schema/postgresql/v12/temporal/versioned")
_, err := fs.ReadFile(fsys, "v1.0/manifest.json") // this works fine
_, err2 := fs.ReadFile(fsys, "v1.0\\manifest.json") // this returns error
```
I tried to change `filepath.Join()` to `path.Join()`, which returns a path using "/" as the separator, in `updatetask.go` and the update-schema will work.

I am not familiar with golang and this project, therefore I am not sure what is the correct way to fix this and I can't raise a PR for this. :disappointed_relieved:

yycptt:
Hi, thanks for reporting the issue! We will look into it.

yiminc:
@san127127 would you be interest to put up a PR to fix this issue?

san127127:
@yiminc 
Not sure if it is indeed the correct fix, but here it is

gow:
[...]

PR #7519, opened 2025-03-24
Fix sql tool error in windows

san127127:
## What changed?
Use `path.Join()` instead of `filepath.Join()`

## Why?
#7454

## How did you test it?
Tested locally

## Potential risks

None

## Documentation

None

## Is hotfix candidate?

No

### DIFF
--- a/tools/common/schema/updatetask.go
+++ b/tools/common/schema/updatetask.go
@@ -30,7 +30,6 @@ import (
 	// should not be used for anything more important (password hashes etc.). Marking it as #nosec because of how it's
 	// being used.
 	"crypto/md5" // #nosec
-	"embed"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -38,7 +37,6 @@ import (
 	"io/fs"
 	"os"
 	"path"
-	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


tools/common/schema/updatetask.go
```diff
@@ -30,15 +30,13 @@ import (
 	// should not be used for anything more important (password hashes etc.). Marking it as #nosec because of how it's
 	// being used.
 	"crypto/md5" // #nosec
-	"embed"
```
yiminc, line 33: i think the idea is to embed the schema into the tool's binary so the schema become optional at runtime, but you can still provide your schema if you want to otherwise it will use the embedded version. 
Would this PR break that logic?

san127127: I just gave it a try.
I found that actually embedded schema cannot not be used in windows because the embed filesystem uses "/" while the allowed values of the `--schema-name` cli option is "\" on windows

It gives this error on windows
```
Config Error:(-schema-name) must be one of: [mysql\\v8\\temporal mysql\\v8\\visibility postgresql\\v12\\temporal postgresql\\v12\\visibility]
```

Do you want to the include a fix for this as well in the PR?

yiminc: Yes, we should fix that, I think we want to keep the embed schema working.

san127127: @yiminc 
Fixed in `embed.go`
I updated the readme as well. TBH I didn't know the schema is embedded until you told me.

gow (approved):
This looks good.
Apparently the issue is in embed stdlib. It doesn't consider OS specific paths and the recommended solution is to use `path.Join()` - https://github.com/golang/go/issues/44305
@san127127 assuming you tested it on windows, this is good to go once the CI passes. I will merge it after that. Thank you for the fix and contributing to temporal.

PR #7522, opened 2025-03-25
versioning-3.1 optimization: batch unversionedRamp

Shivs11:
## What changed?

WISOTT

## Why?

- optimization for versioning-3.1

## How did you test it?

- Added unit tests which confirm that batching is happening correctly.
- Added functional tests sending in requests with 50 task-queues to confirm nothing breaks with batching enabled. Since these are functional tests, I verified that the right number of batches were being made by looking at the logs. It wasn't needed since the unit tests covered this but we rather stay safe than be sorry later on.


service/worker/workerdeployment/workflow.go
```diff
@@ -765,25 +765,109 @@ func (d *WorkflowRunner) syncVersion(ctx workflow.Context, targetVersion string,
...
+			}).Get(ctx, &res)
+		if err != nil {
+			return err
+		}
+
+		// send in the task-queue families in batches of syncBatchSize
+		batches := make([][]*deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData, 0)
+		syncReqs := make([]*deploymentspb.SyncDeploymentVersionUserDataRequest_SyncUserData, 0)
+
+		// Grouping by task-queue name
```
carlydf, line 812: Would you mind also deleting the line `// TODO (Carly): group by TQs by name and make only one sync req per name` from the activities.go file?

carlydf (approved):

PR #7524, opened 2025-03-25
Update Makefile tools

stephanos:
## What changed?

Updated most tools installed and run by Makefile.

Other tools will be upgraded in separate PRs.

## Why?

Dependency hygiene. 

## How did you test it?

Relying on CI here.


api/checksum/v1/message.pb.go
```diff
@@ -295,138 +293,52 @@ func (x *MutableStateChecksumPayload) GetVersionHistories() *v12.VersionHistorie
 
 var File_temporal_server_api_checksum_v1_message_proto protoreflect.FileDescriptor
 
-var file_temporal_server_api_checksum_v1_message_proto_rawDesc = []byte{
```
stephanos, line 298: Note: https://go-review.googlesource.com/c/protobuf/+/657895

Makefile
```diff
@@ -145,6 +143,10 @@ endef
 print-go-version:
 	@go version
 
+clean-tools:
+	rm -rf $(STAMPDIR)
+	rm -rf $(LOCALBIN)
```
stephanos, line 148: For convenience.

Makefile
```diff
@@ -155,19 +157,18 @@ $(LOCALBIN):
 .PHONY: golangci-lint
 GOLANGCI_LINT_BASE_REV ?= $(MAIN_BRANCH)
 GOLANGCI_LINT_FIX ?= true
-# TODO: change to release once newer version than 1.62.2 exists: https://github.com/golangci/golangci-lint/releases
-GOLANGCI_LINT_VERSION := dafd65537336fdce063c492a7ab2a68cc89f8d52
```
stephanos, line 159: We can use the release now.

Makefile
```diff
@@ -182,26 +183,26 @@ BUF := $(LOCALBIN)/buf-$(BUF_VER)
 $(BUF): | $(LOCALBIN)
 	$(call go-install-tool,$(BUF),github.com/bufbuild/buf/cmd/buf,$(BUF_VER))
 
-GO_API_VER := v1.33.0
+GO_API_VER := v1.46.0
```
stephanos, line 186: Matches latest release.

Makefile
```diff
@@ -9,13 +9,20 @@ bins: temporal-server temporal-cassandra-tool temporal-sql-tool tdbg
 all: clean proto bins check test
 
 # Used in CI
-ci-build-misc: print-go-version proto go-generate buf-breaking shell-check copyright-check goimports gomodtidy ensure-no-changes
+ci-build-misc: \
+	print-go-version \
+	clean-tools \
```
stephanos, line 14: Added `clean-tools`. Why? It will do nothing in CI (those folders will be empty); but when you try to repro it locally after it failed, using `clean-tools` ensures that you're getting the latest tool versions.

Makefile
```diff
@@ -366,32 +376,33 @@ shell-check:
 check: copyright-check lint shell-check
 
 ##### Tests #####
-clean-test-results:
-	@rm -f test.log $(TEST_OUTPUT_ROOT)/*
```
ychebotarev, line 370: I don't see this like in the new makefile.
Not needed anymore?
Or you want to preserve test.log?

ychebotarev (approved):

Makefile
stephanos: I replaced it with `clean-test-output`; which deletes the whole folder. It used to only contain `test.log`, but now it also contains JUnit XML reports and code coverage, too.

PR #7526, opened 2025-03-25
Move packages from internal to common

stephanos:
## What changed?

Moved all packages inside `/internal` to `/common`.

## Why?

While the `internal` package provides the benefit of preventing any external party from depending on them, we find that in practice this has little benefit.

Our compatibility guarantees do not coverage any APIs in `common` anyway.

Furthermore, splitting commonly-used packages into two places hinders discoverability. And adds (albeit minor) mental friction.

## How did you test it?

Compiler.

## Potential risks

External users could start relying on these packages. But that's their risk assessment to make. There are no guarantees (just like for any other package in this repo!) outside of the external APIs.


temporaltest/server.go
```diff
@@ -38,8 +38,8 @@ import (
 	"go.temporal.io/sdk/worker"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/log"
-	"go.temporal.io/server/internal/temporalite"
 	"go.temporal.io/server/temporal"
+	temporalite "go.temporal.io/temporaltest/internal"
```
stephanos, line 42: NOTE: I moved this package in here under `internal` as it's only used here and I was told once we might want to refactor this in a major way. Therefore, I'm not exposing that particular package.

Makefile
```diff
@@ -111,9 +111,9 @@ FUNCTIONAL_TEST_XDC_ROOT      := ./tests/xdc
 FUNCTIONAL_TEST_NDC_ROOT      := ./tests/ndc
 DB_INTEGRATION_TEST_ROOT      := ./common/persistence/tests
 DB_TOOL_INTEGRATION_TEST_ROOT := ./tools/tests
-INTEGRATION_TEST_DIRS := $(DB_INTEGRATION_TEST_ROOT) $(DB_TOOL_INTEGRATION_TEST_ROOT) ./temporaltest ./internal/temporalite
```
stephanos, line 114: `./internal/temporalite` was merged into `temporaltest`

common/freeport/freeport.go
dnr, line 1: Maybe should go in `common/testing/freeport`?

dnr (approved):

common/freeport/freeport.go
stephanos: Good call 👍

PR #7530, opened 2025-03-25
CHASM: Get Component & Context implementation

yycptt:
## What changed?

- Implement Component() method on chasm tree
- Implement chasm Context

## How did you test it?

- Added unit test


chasm/tree.go
```diff
@@ -155,12 +164,88 @@ func NewEmptyTree(
...
+	if ref.validationFn != nil {
+		if err := ref.validationFn(chasmContext, componentValue); err != nil {
+			return nil, err
+		}
+	}
+
+	return componentValue, nil
+}
+
+func (n *Node) prepareComponentValue(
```
yycptt, line 215: This method or at least some part of it can be reused by Field.Get() implementation I think. cc @alexshtin

lina-temporal (approved):

chasm/test_component_test.go
```diff
@@ -107,7 +107,7 @@ func testComponentSerializedNodes() map[string]*persistencespb.ChasmNode {
 				},
 				Attributes: &persistencespb.ChasmNodeMetadata_ComponentAttributes{
 					ComponentAttributes: &persistencespb.ChasmComponentAttributes{
-						Type:  "test_component",
+						Type:  "TestLibrary.test_component",
```
yycptt, line 110: @alexshtin Is this the right place to change? I don't see `generateMapInit` setting the Type field or probably I missed it.

alexshtin: because it is in #7409.

alexshtin (approved):

PR #7532, opened 2025-03-26
CHASM: use chasm.Field instead of *chasm.Field

alexshtin:
## What changed?

CHASM: use `chasm.Field` instead of `*chasm.Field`.

## Why?

Non-pointer type looks more natural in component definition and there is no reason to support both of them.

## How did you test it?

Added unit tests.

## Potential risks

No.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/chasm/field.go
+++ b/chasm/field.go
@@ -26,8 +26,18 @@ package chasm
 
 import (
 	"reflect"
+	"strings"
 
 	"google.golang.org/protobuf/proto"
+)
+
+const (
+	// Used by reflection.
+	chasmFieldTypePrefix      = "chasm.Field["
+	chasmCollectionTypePrefix = "chasm.Collection["
+	internalFieldName         = "Internal"
+
+	fieldNameTag = "name"
 )
 
 // This struct needs to be create via reflection
@@ -37,16 +47,25 @@ type Field[T any] struct {
 }
 
 type fieldInternal struct {
-	fieldType int // data, component, componentPointer
-	component reflect.Value
+	fieldType fieldType
+	value     any // Component | Data | Collection
 
-	// backingNode *nodeInfo
+	// Pointer to corresponding tree node. Can be nil for the jsut assigned fields.
+	node *Node
 }
 
-func (d *Field[T]) Get(Context) (T, error) {
+func (fi fieldInternal) IsEmpty() bool {
+	return fi.value == nil
+}
+
+func (f Field[T]) Get(Context) (T, error) {
 	// remember to handle d == nil case
 
 	panic("not implemented")
+}
+
+func NewEmptyField[T any]() Field[T] {
+	return Field[T]{}
 }
 
 // re. Data v.s. Component.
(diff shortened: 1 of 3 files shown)


chasm/field.go
```diff
@@ -26,29 +26,48 @@ package chasm
...
 // but reflection can't set prviate fields...
 type Field[T any] struct {
 	Internal fieldInternal
 }
 
 type fieldInternal struct {
-	fieldType int // data, component, componentPointer
-	component reflect.Value
+	fieldType fieldType
+	value     any // Component | Data | Collection
```
yycptt, line 51:
```suggestion
	value     any // Component | Data | Pointer
```

chasm/field.go
```diff
@@ -26,29 +26,48 @@ package chasm
...
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	// Used by reflection.
+	chasmFieldTypePrefix      = "chasm.Field["
+	chasmCollectionTypePrefix = "chasm.Collection["
+	internalFieldName         = "Internal"
+
+	fieldNameTag = "name"
```
yycptt, line 40: what's this for?

chasm/field.go
```diff
@@ -26,29 +26,48 @@ package chasm
...
+
+func (f Field[T]) Get(Context) (T, error) {
 	// remember to handle d == nil case
 
 	panic("not implemented")
 }
 
+func NewEmptyField[T any]() Field[T] {
+	return Field[T]{}
+}
```
yycptt, line 69: nit: not sure how useful this is. Component author can just do Field[T]{} themselves to clear/delete a field.

chasm/field.go
```diff
@@ -26,29 +26,48 @@ package chasm
...
+	chasmCollectionTypePrefix = "chasm.Collection["
+	internalFieldName         = "Internal"
+
+	fieldNameTag = "name"
+)
+
 // This struct needs to be create via reflection
 // but reflection can't set prviate fields...
 type Field[T any] struct {
 	Internal fieldInternal
```
yycptt, line 46: I wonder if we need to use `*fieldInternal`, in case user value is expensive to copy?
Probably not important right now. Component needs to be *struct and data is proto message, which is implement by pointer as well.

chasm/field.go
```diff
@@ -26,29 +26,48 @@ package chasm
...
 }
 
 type fieldInternal struct {
-	fieldType int // data, component, componentPointer
-	component reflect.Value
+	fieldType fieldType
+	value     any // Component | Data | Collection
 
-	// backingNode *nodeInfo
+	// Pointer to corresponding tree node. Can be nil for the jsut assigned fields.
```
yycptt, line 53:
```suggestion
	// Pointer to corresponding tree node. Can be nil for the just assigned fields.
```

yycptt (approved):
LGTM. Comments are not blocking.

chasm/field_test.go
```diff
@@ -0,0 +1,90 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
yycptt, line 5:
```suggestion
// Copyright (c) 2025 Temporal Technologies Inc.  All rights reserved.
```

chasm/field.go
alexshtin: to use `name` tag on the fields instead of actual go struct field name.

chasm/field.go
alexshtin: Yes, user `value` is pointer to the struct and it will be enforced. I was thinking about it and decided to leave as value for now.

chasm/field.go
alexshtin: I think it looks better with constructor. It's a matter of personal preferences though.

PR #7536, opened 2025-03-27
Use shard readiness for history health check

dnr:
## What changed?
History health check should only return serving after the first round of shards have been acquired

## Why?
Can be used to speed up history rollouts

## How did you test it?
unit tests, simple manual test


service/history/service.go
```diff
@@ -92,7 +94,19 @@ func (s *Service) Start() {
...
 	historyservice.RegisterHistoryServiceServer(s.server, s.handler)
 	healthpb.RegisterHealthServer(s.server, s.healthServer)
-	s.healthServer.SetServingStatus(serviceName, healthpb.HealthCheckResponse_SERVING)
+
+	// start as NOT_SERVING, update to SERVING after initial shards acquired
+	s.healthServer.SetServingStatus(serviceName, healthpb.HealthCheckResponse_NOT_SERVING)
+	readinessCtx, readinessCancel := context.WithCancel(context.Background())
+	s.readinessCancel = readinessCancel
+	go func() {
+		if s.handler.controller.InitialShardsAcquired(readinessCtx) == nil {
```
dynajoe, line 103: what should happen if InitialShardsAcquired returns an error or context is canceled? It won't try again. Is that expected?

dnr: It doesn't return any error besides a context error. And the context is only canceled by Service.Stop, in which case I want this to abort (otherwise it could race and set status back to SERVING after Stop sets it to NOT_SERVING)

service/history/shard/controller_impl.go
```diff
@@ -446,12 +464,63 @@ func (c *ControllerImpl) acquireShards(ctx context.Context) {
...
 
 	c.RLock()
+
 	numOfOwnedShards := len(c.historyShards)
+	// Readiness check: We should set initialShardsAcquired when:
+	// 1. It's not already set.
+	// 2. We should own at least one shard (i.e. not before we join membership).
+	// 3. We have ownership of all the shards we're supposed to own.
+	if !c.initialShardsAcquired.Ready() && numOfOwnedShards > 0 {
+		go c.checkShardReadiness(readinessCtx, readinessCancel, expmaps.Values(c.historyShards))
```
dynajoe, line 474: Perhaps we could avoid passing readinessCancel and rewrite to:

```
go func() {
    defer readinessCancel()
    c.checkShardReadiness(readinessCtx, readinessCancel, expmaps.Values(c.historyShards))
}()
```

That way, both branches of this predicate clearly handle canceling the readiness context.

service/history/shard/controller_impl.go
```diff
@@ -446,12 +464,63 @@ func (c *ControllerImpl) acquireShards(ctx context.Context) {
...
+	shards []historyi.ControllableContext,
+) {
+	defer cancel()
+
+	concurrency := int64(max(c.config.AcquireShardConcurrency(), 1))
+	sem := semaphore.NewWeighted(concurrency)
+	var ready atomic.Int32
+	for _, shard := range shards {
+		if sem.Acquire(ctx, 1) != nil {
+			return
```
dynajoe, line 496: It's possible to return from this function while a subset of go routines below are running. Though, perhaps since each of the potentially blocking functions in the goroutine share this same context it's probably okay to assume they'll return due to cancelation?

dynajoe (approved):

service/history/shard/controller_impl.go
dnr: sure, that's nice

service/history/shard/controller_impl.go
dnr: Yes and no... I originally tried to use contexts to get everything to clean up neatly, but it turns out AssertOwnership uses a detached context. So we can't actually interrupt it, but that's fine, the goroutine isn't doing anything interesting after it completes

service/history/shard/controller_impl.go
```diff
@@ -402,6 +414,12 @@ func (c *ControllerImpl) acquireShards(ctx context.Context) {
 
 	ctx = headers.SetCallerInfo(ctx, headers.SystemBackgroundCallerInfo)
 
+	// cancel previous readiness check goroutine
```
alfred-landrum, line 417: I suggest making explicit what I think is happening here: it's intended that the readiness check does not run concurrently with the tryAcquire loop.

service/history/shard/controller_impl.go
```diff
@@ -446,12 +464,63 @@ func (c *ControllerImpl) acquireShards(ctx context.Context) {
...
+	} else {
+		readinessCancel()
+	}
+
 	c.RUnlock()
 	metrics.NumShardsGauge.With(c.taggedMetricsHandler).Record(float64(numOfOwnedShards))
 	c.publishShardCountUpdate(numOfOwnedShards)
 }
 
+func (c *ControllerImpl) checkShardReadiness(
```
alfred-landrum, line 487: A suggestion: this could have return value(s) that indicate whether all the shards are ready,  then all of the concurrency/cancelation logic could live in `acquireShards` instead of being spread between the 2 functions.

service/history/shard/controller_impl.go
dnr: Sort of.. my goal here wasn't really about concurrency, it's that the readiness check should be invalidated as soon as membership changes. I.e. fully acquiring one set of shards shouldn't count as "ready" if the set of owned shards is now different.

service/history/shard/controller_impl.go
dnr: After Joe's suggestion I think this is the case? Are you suggesting just moving the last few lines (with `c.initialShardsAcquired.SetIfNotReady`) into acquireShards? I guess that's okay but I don't think it's an improvement

alfred-landrum: > Are you suggesting just moving the last few lines ...

yes, that's my suggestion. I think it's a little easier to understand the management of `c.initialShardsAcquired` if all of the modifications are in the same function. But up to you.

dnr: ok, that's a good point

service/history/shard/controller_impl.go
```diff
@@ -446,10 +481,67 @@ func (c *ControllerImpl) acquireShards(ctx context.Context) {
...
+	if readinessCtx != nil {
+		if len(ownedShards) > 0 {
+			go func() {
+				defer readinessCancel()
+				if c.checkShardReadiness(readinessCtx, ownedShards) {
+					c.initialShardsAcquired.SetIfNotReady(struct{}{}, nil)
+				}
+			}()
+		} else {
+			readinessCancel()
```
yiminc, line 503: put a comment explain when this would be true that ownedShards is empty, is this for shutdown?

service/history/shard/controller_impl.go
```diff
@@ -446,10 +481,67 @@ func (c *ControllerImpl) acquireShards(ctx context.Context) {
...
+			// Note that AssertOwnership uses a detached context for the actual persistence
+			// op so we can't cancel it. If context is canceled, the final Acquire will
+			// fail and we won't do anything.
+			if shard, err := c.GetShardByID(shardID); err != nil {
+				return
+			} else if _, err := shard.GetEngine(ctx); err != nil {
+				return
+			} else if shard.AssertOwnership(ctx) != nil {
+				return
+			}
```
yiminc, line 530: are any of those expected? if not, shall we wrote log with enough debug info, or emit metrics etc?

service/history/service.go
```diff
@@ -92,7 +94,19 @@ func (s *Service) Start() {
...
+			// add a few seconds for stabilization
+			if util.InterruptibleSleep(readinessCtx, 5*time.Second) == nil {
+				s.healthServer.SetServingStatus(serviceName, healthpb.HealthCheckResponse_SERVING)
+			}
```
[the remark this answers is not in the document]
dnr: I don't think anyone is doing grpc health checks against a dev server, are they?

service/history/shard/controller_impl.go
dnr: line 492:
> i.e. not before we join membership

service/history/shard/controller_impl.go
dnr: `acquireShards` already has lots of metrics and some logs. The main observable result of this readiness check is the grpc health check status. I don't think it needs any other logs or metrics itself (though there is one log line for the whole thing below)

service/history/service.go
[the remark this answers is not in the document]
yiminc: There could be folks doing that for testing, I could think of SDK team might be doing that.

yiminc (approved):

service/history/service.go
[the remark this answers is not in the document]
dnr: The CLI uses this Go SDK method, which uses WorkflowService as the service name in health check queries: https://github.com/temporalio/sdk-go/blob/4c897cd0de70ae042881f303b3e321a70990cfc5/internal/internal_workflow_client.go#L1257
This affects queries for HistoryService specifically. I'm not aware of anything that does health checks specifically for history service yet. (We don't do this from frontend, we use membership.)

PR #7538, opened 2025-03-27
CHASM: deserialize ChasmNode into tree node value

alexshtin:
## What changed?

CHASM: deserialize `ChasmNode` into tree node `value`. This is only first version with few TODOs to be resolved soon.

## Why?

CHASM worksream.

## How did you test it?

Added unit tests.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -57,11 +57,11 @@ type (
 		nodeName string           // key of this node in parent's children map.
 
 		// Type of attributes controls the type of the node.
-		serializedValue *persistencespb.ChasmNode // serialized component | data | collection with metadata
-		value           any                       // deserialized component | data | collection
+		serializedNode *persistencespb.ChasmNode // serialized component | data | collection with metadata
+		value          any                       // deserialized component | data | collection
 
 		// TODO: synced flag need to be added here and it should be cleared
-		//   when values serializedValue anc value got in-sync.
+		//   when values serializedNode anc value got in-sync.
 		//   And deserialization/serialization can be skipped if synced flag is true.
 
 		// TODO: add other necessary fields here, e.g.
@@ -142,7 +142,7 @@ func NewTree(
 	root := newNode(base, nil, "")
 	if len(serializedNodes) == 0 {
 		// If serializedNodes is empty, it means that this new tree.
-		// Create empty node with nil value and empty serializedValue.
+		// Create empty node with nil value and empty serializedNode.
 		err := root.setValue(nil, fieldTypeComponent)
 		if err != nil {
 			return nil, err
@@ -155,7 +155,7 @@ func NewTree(
 		if err != nil {
 			return nil, err
 		}
-		root.setSerializedValue(nodePath, serializedNode)
+		root.setSerializedNode(nodePath, serializedNode)
 	}
 
 	return root, nil
@@ -178,7 +178,7 @@ func NewEmptyTree(
 	return root
 }
 
-// setValue sets node value field and initialize serializedValue field with empty attributes based on fieldType.
+// setValue sets node value field and initialize serializedNode field with empty attributes based on fieldType.
 func (n *Node) setValue(v any, ft fieldType) error {
 	if err := validateValueType(v); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


chasm/tree.go
```diff
@@ -153,6 +164,210 @@ func NewEmptyTree(
...
+	valueT reflect.Type,
+) error {
+	if err := validateType(valueT); err != nil {
+		return err
+	}
+
+	if n.parent == nil {
+		// Top level node is always component.
+		return n.deserializeComponentNode(valueT)
+	}
```
yycptt, line 256: why do we need this check? root node should also have the metadata available.

chasm/tree.go
```diff
@@ -153,6 +164,210 @@ func NewEmptyTree(
...
+	return nil
+}
+
+func validateValueType(v any) error {
+	return validateType(reflect.TypeOf(v))
+}
+
+func validateType(t reflect.Type) error {
+	// TODO: for component, interface should also be supported.
+	if t.Kind() != reflect.Ptr && t.Elem().Kind() != reflect.Struct {
```
yycptt, line 212: hmmm, `||`?

chasm/tree.go
```diff
@@ -153,6 +164,210 @@ func NewEmptyTree(
...
+	}
+	return nil
+}
+
+func (n *Node) deserializeComponentNode(
+	valueT reflect.Type,
+) error {
+	// TODO: use n.serializedValue.GetComponentAttributes().GetType() instead to support deserialization to interface.
+	valueV := reflect.New(valueT.Elem())
+	if n.serializedValue == nil {
```
yycptt, line 276: When will this be nil? I think the metadata should always be available when this method is invoked.

chasm/tree.go
```diff
@@ -153,6 +164,210 @@ func NewEmptyTree(
...
+		if fieldT.Kind() == reflect.Ptr {
+			continue
+		}
+
+		fieldN := fieldName(valueT.Elem().Field(i))
+
+		switch genericTypePrefix(fieldT) {
+		case chasmFieldTypePrefix:
+			if childNode, found := n.children[fieldN]; found {
+				// TODO: support chasm.Field[interface], type should go from registry
```
yycptt, line 314: My thinking is deserialize is dumb and always deserialize to a concrete value and cache it in the `value` field. The conversion to Interface can be done inside chasm.Field[Interface].Get() implementation. 

Either approach works for me and your's is probably cleaner and I'd like to see how Field.Get() is implemented.
I guess what I want to avoid though is duplicated decoding. If the decoded value is already available, then just return it, because for both component and data there's only one underlying *Struct type.

chasm/tree.go
```diff
@@ -153,6 +164,210 @@ func NewEmptyTree(
...
+}
+
+func fieldName(f reflect.StructField) string {
+	if tagName := f.Tag.Get(fieldNameTag); tagName != "" {
+		return tagName
+	}
+	return f.Name
+}
+
+func (n *Node) setSerializedValue(
```
yycptt, line 225: same comment as the `serializedValue` field. It's feels very confusing to me.

chasm/tree.go
```diff
@@ -44,31 +53,44 @@ type (
...
 		parent   *Node
-		children map[string]*Node
-		nodeName string // key of this node in parent's children map.
+		children map[string]*Node // child name (path segment) -> child node
+		nodeName string           // key of this node in parent's children map.
 
-		persistence *persistencespb.ChasmNode
-		value       any // deserialized component | data | collection
+		// Type of attributes controls the type of the node.
+		serializedValue *persistencespb.ChasmNode // serialized component | data | collection
```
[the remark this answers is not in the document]
alexshtin: renamed to `serializedNode`.

chasm/tree.go
alexshtin: renamed to `setSerializedNode()`.

chasm/tree.go
alexshtin: yes, after I moved to creation to `NewTree` it is always not `nil`. `Data` can be `nil` though.

yycptt (approved):

PR #7539, opened 2025-03-27
disable RSA key length error in test

ychebotarev:
## What changed?

We use 512 key in tests, this generate an error in go 1.24 for some reason.
 `rsa1024min=0` GODEBUG setting suppresses this error.
We already use it in other places in the code.

Also removed t.Parallel()

## Why?

Otherwise go 1.24 generate and error in func test.
t.Parallel() can't be used with t.Setenv

## How did you test it?

By running test with go 1.24

## Potential risks

None.
Though we probably should be using same key length we use in prod.

### DIFF
--- a/tests/testutils/tls.go
+++ b/tests/testutils/tls.go
@@ -133,7 +133,7 @@ func GenerateSelfSignedCAPool(caCert *tls.Certificate) (*x509.CertPool, error) {
 }
 
 func GenerateSelfSignedCA(filePath string) (*tls.Certificate, error) {
-	caCert, err := generateSelfSignedX509CA("undefined", nil, 512)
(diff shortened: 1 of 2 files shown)


fxrlv:
maybe it's better to simply increase it?
https://github.com/temporalio/temporal/blob/b27e02de6d6759ba2b6578351475965ae6e824e1/tests/testutils/tls.go#L136

ychebotarev:
> maybe it's better to simply increase it?
> 
> https://github.com/temporalio/temporal/blob/b27e02de6d6759ba2b6578351475965ae6e824e1/tests/testutils/tls.go#L136

good idea, I will try that

yiminc (approved):
I think we should just move to 2048

PR #7542, opened 2025-03-31
Handle OTEL shutdown timeout

stephanos:
## What changed?

Gracefully handling the case where the OTEL trace provider shutdown times out.

Before this change, it would propagate the "context deadline exceeded" error to fx and it would fail the entire functional test. But now it swallows that error on shutdown.

## Why?

Failure to shut down OTEL should not affect the functional test outcome.

### DIFF
--- a/temporal/fx.go
+++ b/temporal/fx.go
@@ -31,6 +31,7 @@ import (
 	"maps"
 	"os"
 	"sync/atomic"
+	"time"
 
 	"github.com/pborman/uuid"
 	"go.opentelemetry.io/otel"
@@ -978,7 +979,17 @@ var ServiceTracingModule = fx.Options(
 				otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
 					// ignore errors during shutdown
 				}))
-				return tp.Shutdown(ctx)
+
+				shutdownCtx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+				defer cancel()
+
+				err := tp.Shutdown(shutdownCtx)
+				if errors.Is(err, context.DeadlineExceeded) {
+					// Ignore timeouts since it's okay to drop OTEL traces on shutdown.
+					// Either there's no collector, or there are too many traces left to export.
+					return nil
+				}
+				return err
 			}})
 		return tp
 	}),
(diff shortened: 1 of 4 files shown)


temporal/fx.go
```diff
@@ -978,7 +979,12 @@ var ServiceTracingModule = fx.Options(
 				otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
 					// ignore errors during shutdown
 				}))
-				return tp.Shutdown(ctx)
+				ctx, _ = context.WithTimeout(context.Background(), 1*time.Second) // it's okay to drop OTEL traces
+				err := tp.Shutdown(ctx)
+				if errors.Is(err, context.DeadlineExceeded) {
+					return nil // either there's no collector, or there are too many traces to export
+				}
+				return err
```
stephanos, line 987: This is the main change.

temporal/fx.go
```diff
@@ -978,7 +979,13 @@ var ServiceTracingModule = fx.Options(
...
 					// ignore errors during shutdown
 				}))
-				return tp.Shutdown(ctx)
+				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second) // it's okay to drop OTEL traces
+				defer cancel()
+				err := tp.Shutdown(ctx)
+				if errors.Is(err, context.DeadlineExceeded) {
+					return nil // either there's no collector, or there are too many traces to export
+				}
+				return err
```
stephanos, line 988: This is the main change.

temporal/fx.go
```diff
@@ -978,7 +979,13 @@ var ServiceTracingModule = fx.Options(
...
 					// ignore errors during shutdown
 				}))
-				return tp.Shutdown(ctx)
+				shutdownCtx, cancel := context.WithTimeout(context.Background(), 1*time.Second) // it's okay to drop OTEL traces
+				defer cancel()
+				err := tp.Shutdown(shutdownCtx)
+				if errors.Is(err, context.DeadlineExceeded) {
+					return nil // either there's no collector, or there are too many traces to export
+				}
+				return err
```
stephanos, line 988: This is the main change.

temporal/server_test.go
```diff
@@ -48,9 +49,31 @@ import (
...
+	})
+
+	t.Run("without OTEL Collector running", func(t *testing.T) {
+		startAndStopServer(t)
+	})
+}
+
+func startAndStopServer(t *testing.T) {
 	cfg := loadConfig(t)
+	cfg.Global.Metrics.Prometheus.ListenAddress = ":0" // prometheus reporter does not shut down in-between test runs
```
stephanos, line 76: I tried shutting down the Prometheus reporter, but there is no way to do that as it starts the http server inside a goroutine and has no stop/shutdown callback :-/

temporal/fx.go
stephanos, line 1: This is the main change.

stephanos, line 1: This is the main change.

alexshtin (approved):

PR #7544, opened 2025-03-31
Set low request priority for long poll GetWorkflowExecutionHistory requests

prathyushpv:
## What changed?
Set low priority to GetWorkflowHistory requests with WaitNewEvent set to true.
A new config frontend.reducePollWorkflowRequestPriority is added to enable/disable this behaviour.

## Why?
These requests are long poll requests. So we should treat it that way.

## How did you test it?
Running server locally and executing  `temporal workflow show` command.


yiminc (commented):
Lets use PollWorkflowExecutionHistory for long poll, and use GetWorkflowExecutionHistory for non-poll case.

Also, update telemetryUnaryOverrideOperationTag() to only overwrite it to PollXxx if it is long poll.

prathyushpv:
Updated the PR. telemetryUnaryOverrideOperationTag is setting operation tag to to PollWorkflowExecutionHistory if it is long poll already here: https://github.com/temporalio/temporal/blob/f1cedcf2dcf88cbccfe2a734adb054c053897e37/common/rpc/interceptor/telemetry.go#L144-L144

We can override the method name also in telemetryInterceptor but that will spilt this logic between namespaceRateLimiterInterceptor and TelemetryInterceptor. That can be hard to read later.

prathyushpv:
Ideally, we want to change the method name in CallerInfo attached to the request context as well. This is the method name that will be used by rate limiters in other services(eg PersistenceRateLimiter in history service). 

But that requires changes in a few other places and more testing. Otherwise, these requests will get the lowest possible priority. It is fine if we make that change later. As long as total number of frontend requests are limited and persistence QPS values are large enough to accommodate that load, we should not face issues. I will follow up on this change later.

service/frontend/configs/quotas.go
```diff
@@ -36,6 +36,8 @@ import (
 const (
 	// OperatorPriority is used to give precedence to calls coming from web UI or tctl
 	OperatorPriority = 0
+	// PollWorkflowHistoryAPIName is used instead of GetWorkflowExecutionHistory if WaitNewEvent is true in request.
```
yiminc, line 39: move to line 50 to group with the other API name consts.

yiminc (approved):

pdoerner (commented):
Instead of lowering the priority, can we just exclude long-poll GetHistory calls by inspecting the method in the `namespace_rate_limiter` interceptor method? Long-poll requests are already covered by the concurrent limit. We're doing a similar thing in the `concurrent_request_limit` interceptor.

pdoerner (approved):

prathyushpv:
Yeah. If we have to allow high concurrent limit, we want those requests to not block other requests.

PR #7546, opened 2025-04-01
[CHASM] Generate InternalChasmNode.CassandraBlob for CDS

lina-temporal:
## What changed?
- When serializing InternalWorkflowSnapshot/InternalWorkflowMutation, we must set the Cassandra field for the CDS engine as well as our built-in Cassandra engine.

## Why?
- CDS [will expect the Cassandra blob to be set](https://github.com/temporalio/saas-temporal/pull/3249), persisting the whole blob directly.


lina-temporal:
closing - we'll use the separate fields in CDS as well

Issue #4233, opened 2023-04-26
Too high memory usage on history service 


koolay:
## Expected Behavior
Require less memory usage.

## Actual Behavior

container resources:

![image](https://user-images.githubusercontent.com/4273490/234535219-0a0f3858-bb33-4202-8726-46fa96059027.png)

server metrics:

![image](https://user-images.githubusercontent.com/4273490/234550779-036d1636-bfd4-4308-b29d-98000fccfefa.png)

go tool pprof "http://localhost:7936/debug/pprof/heap"

![image](https://user-images.githubusercontent.com/4273490/234534796-fa97c39c-64ed-49d8-94bd-f1150712fa30.png)

## Steps to Reproduce the Problem

  1.  create one thousand workflows

  `temporal-dynamic-config`

 ```yaml
apiVersion: v1
data:
  dynamic_config.yaml: |-
    history.cacheInitialSize:
    - value: "10"
    history.cacheMaxSize:
    - value: "50"
    history.cacheTTL:
    - value: "0.5"
    history.eventsCacheInitialSize:
    - value: "10"
    history.eventsCacheMaxSize:
    - value: "50"
    matching.useOldRouting:
    - value: "false"
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: temporal
    meta.helm.sh/release-namespace: temporal
  creationTimestamp: "2022-11-09T04:02:55Z"
  labels:
    app.kubernetes.io/instance: temporal
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: temporal
    app.kubernetes.io/part-of: temporal
    app.kubernetes.io/version: 1.20.2
    helm.sh/chart: temporal-0.20.2
  name: temporal-dynamic-config
  namespace: temporal
  resourceVersion: "64727569"
  uid: 01925a28-4772-42e9-891e-bdb467bf496f
```

## Specifications

  - Version: temporalio/server:1.20.2
  - Platform: k8s

leididi:
encountered the same problem；
`Name:         temporal-dynamic-config
Namespace:    temporal
Labels:       app.kubernetes.io/instance=temporal
              app.kubernetes.io/managed-by=Helm
              app.kubernetes.io/name=temporal
              app.kubernetes.io/part-of=temporal
              app.kubernetes.io/version=1.20.2
              helm.sh/chart=temporal-0.20.2
Annotations:  meta.helm.sh/release-name: temporal
[...]

PR #7547, opened 2025-04-01
Add active cache eviction for history service

PowerStateFailure:
## What changed?

Add active cache eviction for history service (mutablestate and event caches) implemented as background routine per cache instance that periodically evicts at most N expired elements from cache. The eviction interval and number of elements to be evicted during iteratino (N) can be configured via dynamic config on per-instance base.

The feature is disabled by default.

## Why?

In our use-cases Temporal server is deployed in very restricted environments where it shares resources with different services for different use-cases. We faced that after load spikes Temporal keeps consume memory allocated by caches forever (related issue: #4233), and then we found that Temporal implements only passive memory eviction for caches (the terminology is taken form Redis), i.e. element will be removed from cache only when inserting new element in already full cache. Thus we implement active cache eviction and decided to return it back into upstream.

## How did you test it?

Tested with our internal fork running in several on-premise environments.

## Potential risks

Since the feature is disabled by default, any problems are rarely expected.

### DIFF
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -73,6 +73,18 @@ type Options struct {
 	// TimeSource is an optional clock to use for time-skipping and testing. If this is nil, a real clock will be used.
 	TimeSource clock.TimeSource
 
+	// ActiveEviction creates background worker that will periodically remove expired elements from cache.
+	// Once true, caller must call Close() on cache wen it's no more used to avoid memory leaks.
+	ActiveEviction bool
+
+	// ActiveEvictionInterval sets interval between eviction runs.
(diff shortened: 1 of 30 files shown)


prathyushpv:
This feature is already implemented in https://github.com/temporalio/temporal/pull/7902

PowerStateFailure:
Hello @prathyushpv 
Thanks for reply.
Can u clarify on how to get team attention on review? It seems unclear to me why some PRs gets no reaction for month but others such as #7902 are reviewed almost immediately (not to mention it coincidently shares key concepts with this PR but created much later). We're still having plans on trying to backport our private fixes to upstream, but if they got no attention it will be just waste of time.

PR #7555, opened 2025-04-02
Pause activity by rules in executeActivityRetryTimerTask

ychebotarev:
## What changed?

Add code to pause the activity while processing activity retry task.

## Why?

Part of the workflow rules efforts. This is another place where we need to check rules related to the activity.

## How did you test it?

Add functional test.

## Potential risks

N/A.
## Documentation

N/A.

### DIFF
--- a/service/history/timer_queue_active_task_executor.go
+++ b/service/history/timer_queue_active_task_executor.go
@@ -520,12 +520,6 @@ func (t *timerQueueActiveTaskExecutor) executeActivityRetryTimerTask(
 		return consts.ErrActivityTaskNotFound
 	}
 
-	err = t.processActivityWorkflowRules(ctx, weContext, mutableState, activityInfo)
-	if err != nil {
-		release(nil) // release(nil) so mutable state is not unloaded from cache
-		return err
-	}
-
 	if task.Stamp != activityInfo.Stamp || activityInfo.Paused {
 		// if retry task event is from an old stamp of if activity is paused we should ignore the event.
 		release(nil) // release(nil) so mutable state is not unloaded from cache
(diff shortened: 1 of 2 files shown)


service/history/timer_queue_active_task_executor.go
```diff
@@ -520,6 +520,12 @@ func (t *timerQueueActiveTaskExecutor) executeActivityRetryTimerTask(
 		return consts.ErrActivityTaskNotFound
 	}
 
+	err = t.processActivityWorkflowRules(ctx, weContext, mutableState, activityInfo)
+	if err != nil {
+		release(nil) // release(nil) so mutable state is not unloaded from cache
```
yycptt, line 525: plz remove `release(nil)` here. That should only be used when you know the error won't cause mutable state in memory to be out of sync with DB, which is not the case when update fails.

service/history/timer_queue_active_task_executor.go
```diff
@@ -520,6 +520,12 @@ func (t *timerQueueActiveTaskExecutor) executeActivityRetryTimerTask(
 		return consts.ErrActivityTaskNotFound
 	}
 
+	err = t.processActivityWorkflowRules(ctx, weContext, mutableState, activityInfo)
```
yycptt, line 523: This should only be done after the task passed all the version, stamp, workflow status check below right? at a minimal workflow needs to be running. 
I think L523-533 needs to be  moved after L556.

service/history/timer_queue_active_task_executor.go
```diff
@@ -520,6 +520,12 @@ func (t *timerQueueActiveTaskExecutor) executeActivityRetryTimerTask(
...
+	if err != nil {
+		release(nil) // release(nil) so mutable state is not unloaded from cache
+		return err
+	}
+
 	if task.Stamp != activityInfo.Stamp || activityInfo.Paused {
```
yycptt, line 529: What's the story around actions count for workflow rules. Does that count as an action? 

Asking here because it looks like the logic is relying on the `|| activityInfo.Paused` to return early after updating activity state to paused. But this section returns NotFound error and our cloud aps counting logic will think this task is a duplication and won't count it as an action.

yycptt (changes requested):

service/history/timer_queue_active_task_executor.go
ychebotarev: Moved. But this means we will need to duplicate "if ai.Paused" check.

service/history/timer_queue_active_task_executor.go
ychebotarev: for pause we decided not to count it as actions. Thus - yes, it is correct, it won't count as an action.

service/history/timer_queue_active_task_executor.go
ychebotarev: removed

service/history/timer_queue_active_task_executor.go
```diff
@@ -549,6 +549,18 @@ func (t *timerQueueActiveTaskExecutor) executeActivityRetryTimerTask(
 		return consts.ErrWorkflowCompleted
 	}
 
+	err = t.processActivityWorkflowRules(ctx, weContext, mutableState, activityInfo)
+	if err != nil {
+		return err
+	}
+
+	// task can be paused as the result of processing activity workflow rules, so we need to check again
+	if task.Stamp != activityInfo.Stamp || activityInfo.Paused {
```
yycptt, line 558: will `processActivityWorkflowRules` update the Stamp as well? if not we can only check Paused here I think.

yycptt (approved):

PR #7556, opened 2025-04-02
expire cached entries in history client

alfred-landrum:
## What changed?
In the history client caching redirector, limit the time that a cached client will be used after a membership update signals the shard has moved, defaulting to 30 seconds.

## Why?
The redirector already removed entries that fail with timeouts or shard ownership lost. However, the impact of timeouts can be high, especially when an old shard owners address is unreachable. This manifested as high tail latencies for API calls after history restarts.

## How did you test it?
Added a unit test & in a test setup.

### DIFF
--- a/client/history/caching_redirector.go
+++ b/client/history/caching_redirector.go
@@ -28,10 +28,12 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/api/historyservice/v1"
 	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/goro"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/membership"
@@ -43,6 +45,7 @@ type (
 		shardID    int32
 		address    rpcAddress
 		connection clientConnection
+		staleAt    time.Time
 	}
 
 	// A cachingRedirector is a redirector that maintains a cache of shard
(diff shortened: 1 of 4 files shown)


client/history/caching_redirector.go
```diff
@@ -57,25 +60,46 @@ type (
...
 	}
 )
 
+const cachingRedirectorListener = "cachingRedirectorListener"
+
 func newCachingRedirector(
 	connections connectionPool,
 	historyServiceResolver membership.ServiceResolver,
 	logger log.Logger,
+	staleTTL time.Duration,
```
dnr, line 77: maybe this should be a `dynamicconfig.DurationPropertyFn` to be more dynamic?

client/history/caching_redirector.go
```diff
@@ -57,25 +60,46 @@ type (
...
+		membershipUpdateCh:     make(chan *membership.ChangedEvent, 1),
+		staleTTL:               staleTTL,
 	}
 	r.mu.cache = make(map[int32]cacheEntry)
+
+	if r.staleTTL > 0 {
+		r.goros.Go(func(ctx context.Context) error {
+			r.eventLoop(ctx)
+			return nil
+		})
```
dnr, line 92: make `r.eventLoop` return `error` so this can be just `r.goros.Go(r.eventLoop)`?

client/history/caching_redirector.go
```diff
@@ -57,25 +60,46 @@ type (
...
+	}
+
 	return r
 }
 
+func (r *cachingRedirector) stop() {
```
dnr, line 98: this is never called in non-test environments?

client/history/caching_redirector.go
```diff
@@ -57,25 +60,46 @@ type (
 		}
 
 		connections            connectionPool
+		goros                  goro.Group
 		historyServiceResolver membership.ServiceResolver
 		logger                 log.Logger
+		membershipUpdateCh     chan *membership.ChangedEvent
```
dnr, line 66: does this have to be a field of the struct? it's only used by the goroutine itself. and tests, but maybe there's another way to inject an event?

client/history/caching_redirector.go
```diff
@@ -212,3 +246,47 @@ func maybeHostDownError(opErr error) bool {
...
+		if err := r.historyServiceResolver.RemoveListener(cachingRedirectorListener); err != nil {
+			r.logger.Warn("Error removing listener", tag.Error(err))
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case _ = <-r.membershipUpdateCh:
```
dnr, line 264:
```suggestion
		case <-r.membershipUpdateCh:
```

client/history/caching_redirector.go
alfred-landrum: I'd rather leave it, anything else will be awkward for not much of a benefit.

client/history/caching_redirector.go
alfred-landrum: I don't know of way to accurately call this outside of tests. There's no lifecycle start/stop calls for history clients that I can see, please let me know otherwise.

dnr: Yeah, I think it's fine. I was just asking to confirm

dnr (approved):

PR #7557, opened 2025-04-02
Option to provide a ratelimiter to matching engine

prathyushpv:
## What changed?
Add an option to provide a rate limiter before matching service calls RecordWorkflowTaskStarted and RecordActivityTaskStarted APIs.

## Why?
This rate limiter can be used to limit load on history service.


service/matching/matcher.go
```diff
@@ -88,29 +88,10 @@ var (
...
-		),
-		quotas.NewDefaultOutgoingRateLimiter(
-			config.AdminNamespaceToPartitionDispatchRate,
-		),
-	})
+func newTaskMatcher(config *TaskQueueConfig, fwdr *Forwarder, metricsHandler metrics.Handler, rateLimiter TaskMatcherRateLimiter) *TaskMatcher {
 	return &TaskMatcher{
 		config:                 config,
-		dynamicRateBurst:       dynamicRateBurst,
-		dynamicRateLimiter:     dynamicRateLimiter,
```
prathyushpv, line 112: Why do we have this dynamicRateLimiter in matcher? I don't see it being used anywhere.

service/matching/matching_engine.go
```diff
@@ -2612,6 +2618,18 @@ func (e *matchingEngineImpl) recordWorkflowTaskStarted(
 	pollReq *workflowservice.PollWorkflowTaskQueueRequest,
 	task *internalTask,
 ) (*historyservice.RecordWorkflowTaskStartedResponse, error) {
+	if e.rateLimiter != nil {
+		err := e.rateLimiter.Wait(ctx, quotas.Request{
+			API:        "RecordWorkflowTaskCompleted",
+			Token:      1,
+			Caller:     pollReq.Namespace,
```
dnr, line 2625: just to confirm, this is supposed to be namespace name and not namespace id, right?

prathyushpv: Yes, it should be namespace name.

prathyushpv: Just noticed that forwarded is not populating this field. I added some code above to populate this field.

service/matching/matching_engine.go
```diff
@@ -2612,6 +2630,18 @@ func (e *matchingEngineImpl) recordWorkflowTaskStarted(
 	pollReq *workflowservice.PollWorkflowTaskQueueRequest,
 	task *internalTask,
 ) (*historyservice.RecordWorkflowTaskStartedResponse, error) {
+	if e.rateLimiter != nil {
+		err := e.rateLimiter.Wait(ctx, quotas.Request{
+			API:        "RecordWorkflowTaskStarted",
+			Token:      1,
+			Caller:     pollReq.Namespace,
+			CallerType: headers.CallerTypeMatchingTask,
```
yycptt, line 2638: not sure if a new caller type is necessary. I guess you can treat this in the same way as a PollTask api call, which has the lowest priority among apis but still higher priority than background task processing.

service/matching/matching_engine.go
```diff
@@ -604,6 +610,12 @@ func (e *matchingEngineImpl) PollWorkflowTaskQueue(
 	pollerID := req.GetPollerId()
 	request := req.PollRequest
 	taskQueueName := request.TaskQueue.GetName()
+	ns, err := e.namespaceRegistry.GetNamespaceName(namespace.ID(req.GetNamespaceId()))
```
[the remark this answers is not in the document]
prathyushpv: I can add a check to do this translation if namespace name is not populated already. That will have the same effect right?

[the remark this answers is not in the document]
prathyushpv: I have made the change.

service/matching/matching_engine.go
prathyushpv: Got it. Updated it to callerTypeAPI

service/matching/backlog_manager.go
```diff
@@ -254,6 +254,7 @@ func (c *backlogManagerImpl) completeTask(itask *internalTask, err error) {
 		// We handle this by writing the task back to persistence with a higher taskID.
 		// This will allow subsequent tasks to make progress, and hopefully by the time this task is picked-up
 		// again the underlying reason for failing to start will be resolved.
+		metrics.TaskRewrites.With(c.metricsHandler).Record(1)
```
ShahabT, line 257: let's use c.metricsHandler as it already has a bunch of useful tags in it.

service/matching/pri_backlog_manager.go
```diff
@@ -347,6 +347,7 @@ func (c *priBacklogManagerImpl) respoolTaskAfterError(task *persistencespb.TaskI
 	// This will allow subsequent tasks to make progress, and hopefully by the time this task is picked-up
 	// again the underlying reason for failing to start will be resolved.
 	// Note the task may get written to a different subqueue than it came from.
+	metrics.TaskRewrites.With(c.metricsHandler).Record(1)
```
ShahabT, line 350: same here

service/matching/backlog_manager.go
prathyushpv: I am using the same metrics handler here.

ShahabT: lol sorry did look carefully :D

ShahabT (approved):

PR #7559, opened 2025-04-03
Pause activity in recordactivitytaskstarted

ychebotarev:
## What changed?

Add code to "recordactivitytaskstarted" api to run workflow rules and pause activity if needed.

## Why?

To cover scenario when workflow rule should prevent activity from starting. 

## How did you test it?

Add functional test specifically to this scenario.

### DIFF
--- a/service/history/api/recordactivitytaskstarted/api.go
+++ b/service/history/api/recordactivitytaskstarted/api.go
@@ -182,9 +182,11 @@ func recordActivityTaskStarted(
 		return nil, rejectCodeUndefined, serviceerrors.NewTaskAlreadyStarted("Activity")
 	}
 
-	code, err := processActivityWorkflowRules(shardContext, mutableState, ai)
-	if err != nil || code == rejectCodePaused {
-		return nil, code, err
+	if ai.Stamp == request.Stamp {
+		code, err := processActivityWorkflowRules(shardContext, mutableState, ai)
+		if err != nil || code == rejectCodePaused {
+			return nil, code, err
+		}
 	}
 
 	if ai.Stamp != request.Stamp {
--- a/service/history/api/recordactivitytaskstarted/api.go
+++ b/service/history/api/recordactivitytaskstarted/api.go
@@ -52,10 +52,10 @@ import (
 type rejectCode int32
 
 const (
-	rejectCodeUndefined         rejectCode = iota
-	rejectCodeAccepted          rejectCode = 1
-	rejectCodePaused            rejectCode = 2
-	rejectCodeStartedTransition rejectCode = 3
+	rejectCodeUndefined rejectCode = iota
+	rejectCodeAccepted
+	rejectCodePaused
+	rejectCodeStartedTransition
 )
 
 func Invoke(
@@ -68,7 +68,6 @@ func Invoke(
 
 	var err error
 	response := &historyservice.RecordActivityTaskStartedResponse{}
-	var startedTransition bool
 	var rejectCode rejectCode
 
 	err = api.GetAndUpdateWorkflowWithNew(
@@ -92,17 +91,12 @@ func Invoke(
 				return nil, err
 			}
 
-			startedTransition = false
-			if rejectCode == rejectCodeStartedTransition {
-				startedTransition = true
-			}
-
 			return &api.UpdateWorkflowAction{
 				Noop: false,
 				// Create new wft if a transition started with this activity.
 				// StartDeploymentTransition rescheduled pending wft, but this creates new
 				// one if there is no pending wft.
-				CreateWorkflowTask: startedTransition,
+				CreateWorkflowTask: rejectCode == rejectCodeStartedTransition,
 			}, nil
 		},
 		nil,
--- a/service/history/api/recordactivitytaskstarted/api.go
+++ b/service/history/api/recordactivitytaskstarted/api.go
@@ -34,6 +34,7 @@ import (
 	deploymentspb "go.temporal.io/server/api/deployment/v1"
 	"go.temporal.io/server/api/historyservice/v1"
 	"go.temporal.io/server/api/matchingservice/v1"
+	persistencespb "go.temporal.io/server/api/persistence/v1"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/definition"
 	"go.temporal.io/server/common/metrics"
@@ -74,7 +75,9 @@ func Invoke(
 				return nil, consts.ErrWorkflowCompleted
 			}
 
-			response, startedTransition, err = recordActivityTaskStarted(ctx, shardContext, mutableState, request, matchingClient)
+			response, startedTransition, err = recordActivityTaskStarted(
+				ctx, shardContext, workflowLease.GetContext(), mutableState, request, matchingClient,
+			)
 			if err != nil {
 				return nil, err
 			}
@@ -107,6 +110,7 @@ func Invoke(
 func recordActivityTaskStarted(
 	ctx context.Context,
 	shardContext historyi.ShardContext,
+	weContext historyi.WorkflowContext,
 	mutableState historyi.MutableState,
 	request *historyservice.RecordActivityTaskStartedRequest,
 	matchingClient matchingservice.MatchingServiceClient,
@@ -160,6 +164,11 @@ func recordActivityTaskStarted(
 		// Looks like ActivityTask already started as a result of another call.
 		// It is OK to drop the task at this point.
 		return nil, false, serviceerrors.NewTaskAlreadyStarted("Activity")
+	}
+
+	err = processActivityWorkflowRules(ctx, shardContext, weContext, mutableState, ai)
+	if err != nil {
+		return nil, false, err
 	}
 
 	if ai.Stamp != request.Stamp {
@@ -280,4 +289,50 @@ func getDeploymentVersionForWorkflowId(
 	}
 	current, ramping := worker_versioning.CalculateTaskQueueVersioningInfo(tqData.GetDeploymentData())
 	return worker_versioning.FindDeploymentVersionForWorkflowID(current, ramping, workflowId), nil
+}
+
+func processActivityWorkflowRules(
+	ctx context.Context,
+	shardContext historyi.ShardContext,
+	weContext historyi.WorkflowContext,
+	ms historyi.MutableState,
+	ai *persistencespb.ActivityInfo,
+) error {
+	if ai.Paused {
+		return nil
+	}
+
+	// This is an attempt to reduce the number of workflow rules calls.
+	// We only need to process the first invocation of an activity.
+	// Other invocations should be blocked by either RetryActivity or retryTask.
+	if ai.Attempt > 1 {
+		return nil
+	}
+
+	activityChanged := workflow.ActivityMatchWorkflowRules(ms, shardContext.GetLogger(), ai)
+	if !activityChanged {
+		return nil
+	}
+	if ai.Paused {
+		// need to update activity
+		if err := ms.UpdateActivity(ai.ScheduledEventId, func(activityInfo *persistencespb.ActivityInfo, _ historyi.MutableState) error {
+			activityInfo.StartedEventId = common.EmptyEventID
+			activityInfo.StartedTime = nil
+			activityInfo.RequestId = ""
+			return nil
+		}); err != nil {
+			return err
+		}
+
+		// need to update mutable state
+		err := weContext.UpdateWorkflowExecutionAsActive(
+			ctx,
+			shardContext,
+		)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/service/history/api/recordactivitytaskstarted/api.go
+++ b/service/history/api/recordactivitytaskstarted/api.go
@@ -125,7 +125,7 @@ func Invoke(
 	if rejectCode == rejected_modified {
 		// Rejecting the activity start because activity was modified.
 		// Matching can drop the task. New activity will be scheduled once activity is resumed.
-		errorMessage := fmt.Sprintf("Activity task with this stamp not found: %s", request.Stamp)
+		errorMessage := fmt.Sprintf("Activity task with this stamp not found: %v", request.Stamp)
 		return nil, serviceerror.NewNotFound(errorMessage)
 	}
 
--- a/service/history/api/recordactivitytaskstarted/api.go
+++ b/service/history/api/recordactivitytaskstarted/api.go
@@ -182,11 +182,9 @@ func recordActivityTaskStarted(
 		return nil, rejectCodeUndefined, serviceerrors.NewTaskAlreadyStarted("Activity")
 	}
 
-	if ai.Stamp == request.Stamp {
-		code, err := processActivityWorkflowRules(shardContext, mutableState, ai)
-		if err != nil || code == rejectCodePaused {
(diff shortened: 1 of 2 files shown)


service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -281,3 +290,49 @@ func getDeploymentVersionForWorkflowId(
...
+			activityInfo.StartedEventId = common.EmptyEventID
+			activityInfo.StartedTime = nil
+			activityInfo.RequestId = ""
+			return nil
+		}); err != nil {
+			return err
+		}
+
+		// need to update mutable state
+		err := weContext.UpdateWorkflowExecutionAsActive(
```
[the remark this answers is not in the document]
ychebotarev: yeah, then it is not needed.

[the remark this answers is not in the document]
ychebotarev: actually it is needed. I add a comments in code explaining why.
If activity is paused this API returns error, and in this case api.GetAndUpdateWorkflowWithNew will not call UpdateWorkflowExecutionAsActive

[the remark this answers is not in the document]
yycptt: not following sorry.

> If activity is paused this API returns error

sorry which API? `ms.UpdateActivity` or `weContext.UpdateWorkflowExecutionAsActive`. 
and if error is returned, when in memory mutable state will be cleared and entire recordActivityTaskStarted will be retried.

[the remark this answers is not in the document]
ychebotarev: lets chat offline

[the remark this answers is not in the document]
ychebotarev: discussed. I return "reject code", and in some cases update workflow (there is one existing case, another one - activity was paused)

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -96,24 +116,33 @@ func Invoke(
...
+		return nil, serviceerror.NewNotFound(errorMessage)
+	}
+
 	return response, err
 }
 
 func recordActivityTaskStarted(
 	ctx context.Context,
 	shardContext historyi.ShardContext,
+	weContext historyi.WorkflowContext,
```
yycptt, line 138: this is no longer needed and can be reverted I think.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -48,6 +49,17 @@ import (
 	"go.temporal.io/server/service/history/workflow"
 )
 
+type RejectCode int32
```
yycptt, line 52: nit: doesn't need to be exported.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -48,6 +49,17 @@ import (
...
+
+const (
+	reject_code_undefined       RejectCode = 0
+	activity_accepted           RejectCode = 1
+	rejected_stamp              RejectCode = 2
+	rejected_with_error         RejectCode = 3
```
yycptt, line 58: nit: probably not necessary and can just use default value for RejectCode when returned error is not nil. Caller won't/not supposed to inspect the code when error is returned.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -48,6 +49,17 @@ import (
 	"go.temporal.io/server/service/history/workflow"
 )
 
+type RejectCode int32
+
+const (
+	reject_code_undefined       RejectCode = 0
+	activity_accepted           RejectCode = 1
+	rejected_stamp              RejectCode = 2
```
yycptt, line 57:
```suggestion
	rejectedStampMismatch              RejectCode = 2
```

Should all be camelCase : https://go.dev/doc/effective_go#mixed-caps

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -48,6 +49,17 @@ import (
...
+	rejected_with_error         RejectCode = 3
+	rejected_modified           RejectCode = 4
```
yycptt, line 59: nit: `rejectedPaused` maybe

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -154,12 +183,22 @@ func recordActivityTaskStarted(
...
-		return nil, false, serviceerrors.NewTaskAlreadyStarted("Activity")
+		return nil, rejected_with_error, serviceerrors.NewTaskAlreadyStarted("Activity")
+	}
+
+	activityModified, err := processActivityWorkflowRules(shardContext, mutableState, ai)
+	if err != nil {
+		return nil, rejected_with_error, err
+	}
+
+	if activityModified && ai.Stamp != request.Stamp {
```
yycptt, line 199: what if after modifying the activity the `Stamp` in `ai` matches that in the `request`? 

I don't a stamp check before or inside `processActivityWorkflowRules`

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -154,12 +183,22 @@ func recordActivityTaskStarted(
...
+			return response, activity_accepted, nil
 		}
 
 		// Looks like ActivityTask already started as a result of another call.
 		// It is OK to drop the task at this point.
-		return nil, false, serviceerrors.NewTaskAlreadyStarted("Activity")
+		return nil, rejected_with_error, serviceerrors.NewTaskAlreadyStarted("Activity")
+	}
+
+	activityModified, err := processActivityWorkflowRules(shardContext, mutableState, ai)
```
yycptt, line 194: nit: we can also consider return rejectCode and error from processActivityWorkflowRules. and here simply do:
```
if err != nil || rejectCode != rejectCodeUndefined {
    return rejectCode, err
}
```

tests/activity_api_rules_test.go
```diff
@@ -552,6 +543,126 @@ func (s *ActivityApiRulesClientTestSuite) TestActivityRulesApi_RetryTask() {
...
+
+	// 3. Let namespace config propagate to the history service.
+	// There is no good way to check if the namespace config has propagated to the history service
+	err = util.InterruptibleSleep(ctx, 2*time.Second)
+	s.NoError(err)
+
+	// 4. Start workflow
+	workflowRun := s.createWorkflow(ctx, testRetryTaskWorkflow.WorkflowFuncForPrePause)
+
+	// 5. Wait for activity to be paused by rule. This should happen in the recording activity task started
```
yycptt, line 599: If the rule is there before starting the workflow, activity should not be dispatched to matching in the first place right and there won't be a record task started call.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -48,6 +49,17 @@ import (
 	"go.temporal.io/server/service/history/workflow"
 )
 
+type RejectCode int32
+
+const (
+	reject_code_undefined       RejectCode = 0
```
alexshtin, line 55: Use `iota`, Luke!

service/history/api/recordactivitytaskstarted/api.go
alexshtin: Another good convention is to start all of them with `rejectedCode`. I.e. `rejectedCodeStampMismatch`, `rejectedCodePaused`, etc. It helps intellisense also.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -281,3 +320,38 @@ func getDeploymentVersionForWorkflowId(
 	current, ramping := worker_versioning.CalculateTaskQueueVersioningInfo(tqData.GetDeploymentData())
 	return worker_versioning.FindDeploymentVersionForWorkflowID(current, ramping, workflowId), nil
 }
+
+func processActivityWorkflowRules(
+	shardContext historyi.ShardContext,
+	ms historyi.MutableState,
+	ai *persistencespb.ActivityInfo,
+) (bool, error) {
```
alexshtin, line 328: Use `rejectedCode` instead of bool here too. It is easy to read and you have the enum already.

tests/activity_api_rules_test.go
ychebotarev: it is possible that rule was created after task was scheduled but before it was started.
Adding rule before starting the workflow is the easiest way to test it.

>activity should not be dispatched to matching

We discuss it, and decide that we should dispatch it. Exactly because this way there will be a record task.

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: done

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: renamed

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: removed

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: done

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: renamed

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: done

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: done

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -48,6 +49,15 @@ import (
...
 )
 
+type rejectCode int32
+
+const (
+	rejectCodeUndefined         rejectCode = iota
+	rejectCodeAccepted          rejectCode = 1
+	rejectCodePaused            rejectCode = 2
+	rejectCodeStartedTransition rejectCode = 3
+)
```
yycptt, line 59:
```suggestion
const (
	rejectCodeUndefined         rejectCode = iota
	rejectCodeAccepted
	rejectCodePaused
	rejectCodeStartedTransition
)
```

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -74,11 +85,18 @@ func Invoke(
...
 
-			response, startedTransition, err = recordActivityTaskStarted(ctx, shardContext, mutableState, request, matchingClient)
+			response, rejectCode, err = recordActivityTaskStarted(
+				ctx, shardContext, mutableState, request, matchingClient,
+			)
 			if err != nil {
 				return nil, err
 			}
 
+			startedTransition = false
```
yycptt, line 95: nit: guess we can get rid of of this variable.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -281,3 +312,38 @@ func getDeploymentVersionForWorkflowId(
...
+	}
+
+	// This is an attempt to reduce the number of workflow rules calls.
+	// We only need to process the first invocation of an activity.
+	// Other invocations should be blocked by either RetryActivity or retryTask.
+	if ai.Attempt > 1 {
+		return rejectCodeUndefined, nil
+	}
+
+	ruleMatched := workflow.ActivityMatchWorkflowRules(ms, shardContext.GetLogger(), ai)
```
yycptt, line 332: Since this is before the activity stamp check, we will process the rules even if stamp in `ai` != the one in the `RecordActivityTaskStartedRequest`. Is that intentional?

tests/activity_api_rules_test.go
yycptt: This doesn't have to block the PR but wondering why we made that decision, which means an extra dispatch + write operation to pause the activity?

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -281,3 +312,38 @@ func getDeploymentVersionForWorkflowId(
...
 	return worker_versioning.FindDeploymentVersionForWorkflowID(current, ramping, workflowId), nil
 }
+
+func processActivityWorkflowRules(
+	shardContext historyi.ShardContext,
+	ms historyi.MutableState,
+	ai *persistencespb.ActivityInfo,
+) (rejectCode, error) {
+	if ai.Paused {
+		return rejectCodeUndefined, nil
```
yycptt, line 322: shall we return `rejectCodePaused` here? Otherwise, caller will continue to dispatch the activity task I think.

I guess ideally in this case we also want to prevent mutable state from doing an write operation, i.e. Noop on L101 should be true

Actually just returning rejectCodeUndefined and NotFound error probably works here.

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: removed

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: removed

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: Well that is why I called it 'rejectCodeModified'.
If Paused is set to true - activity was already paused, and we don't want to process the rules.
If we return rejectCodePaused without error - code will still update mutable state, which is not what I want.

If activity was paused its stamp is different, and code will return serviceerror.NewNotFound, and update will not happen.

tests/activity_api_rules_test.go
ychebotarev: Gist - even if activity is not supposed to run we still want to generate scheduled event. Why not to "generate event but don't dispatch"? - there was a good reason, but I don't remember. May be too much dangerous refactoring, and since we need this check anyway it is safer to let that dispatch to happen?

service/history/api/recordactivitytaskstarted/api.go
ychebotarev: after some thinking - yeah, I don't think we should process rules if stamps are different. I will add the check

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -281,3 +306,50 @@ func getDeploymentVersionForWorkflowId(
...
 }
+
+func processActivityWorkflowRules(
+	shardContext historyi.ShardContext,
+	request *historyservice.RecordActivityTaskStartedRequest,
+	ms historyi.MutableState,
+	ai *persistencespb.ActivityInfo,
+) (rejectCode, error) {
+	if ai.Stamp == request.Stamp && ai.Paused {
+		// this shouldn't happen. For now log an error
```
yycptt, line 317: I thought you want to return an error.

tests/activity_api_rules_test.go
yycptt: > "generate event but don't dispatch"

yeah feels like something we should do.  I don't see strong reasons not doing it at least.

Customer may also be confused if they have all their activities paused, but still seeing say actions for dispatching the activity task.

yycptt (approved):
You'll also need to change the standby transfer task logic for activity to stop verification when stamp mismatches or if activity is paused. cc @yux0 @xwduan 

It looks like we don't have the stamp check in active logic either.

PR #7563, opened 2025-04-03
Emit metric for start workflow request deduped

rodrigozhou:
## What changed?

Emit metric for start workflow request deduped.
Fix possible double emitting eager execution denied counter.

### DIFF
--- a/service/history/api/startworkflow/api.go
+++ b/service/history/api/startworkflow/api.go
@@ -117,16 +117,9 @@ func NewStarter(
 		return nil, err
 	}
 
-	metricsHandler := workflow.GetPerTaskQueueFamilyScope(
-		shardContext.GetMetricsHandler(),
-		namespaceEntry.Name(),
-		request.StartRequest.TaskQueue.Name,
-		shardContext.GetConfig(),
-		metrics.WorkflowTypeTag(request.StartRequest.WorkflowType.Name),
-	)
-
 	return &Starter{
-		metricsHandler:             metricsHandler,
+		// metricsHandler is lazily created when needed in Starter.getMetricsHandler
+		metricsHandler:             nil,
 		shardContext:               shardContext,
 		workflowConsistencyChecker: workflowConsistencyChecker,
 		tokenSerializer:            tokenSerializer,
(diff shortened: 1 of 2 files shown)


service/history/api/startworkflow/api.go
```diff
@@ -143,45 +151,35 @@ func (s *Starter) prepare(ctx context.Context) error {
...
+		metrics.WorkflowEagerExecutionCounter.With(s.metricsHandler).Record(1)
 
 		// Override to false to avoid having to look up the dynamic config throughout the different code paths.
 		if !s.shardContext.GetConfig().EnableEagerWorkflowStart(s.namespace.Name().String()) {
-			s.recordEagerDenied(eagerStartDeniedReasonDynamicConfigDisabled)
+			metrics.WorkflowEagerExecutionDeniedCounter.With(s.metricsHandler).
+				Record(1, metrics.ReasonTag(eagerStartDeniedReasonDynamicConfigDisabled))
 			request.RequestEagerExecution = false
-		}
-		if s.request.FirstWorkflowTaskBackoff != nil && s.request.FirstWorkflowTaskBackoff.AsDuration() > 0 {
```
rodrigozhou, line 166: This could be double counting eager execution denied, so I changed to `else if`.

mastermanu: what about the nil check? not needed anymore?

mastermanu (approved):

service/history/api/startworkflow/api.go
rodrigozhou: Not needed, if it's nil, it will return 0.

service/history/api/startworkflow/api.go
```diff
@@ -117,7 +117,16 @@ func NewStarter(
 		return nil, err
 	}
 
+	metricsHandler := workflow.GetPerTaskQueueFamilyScope(
```
yycptt, line 120: nit: I'd do lazy init for the metricsHandler if possible to save some gc considering this is a hot path in the system.

service/history/api/startworkflow/api.go
```diff
@@ -318,6 +316,8 @@ func (s *Starter) handleConflict(
 	request := s.request.StartRequest
 	currentWorkflowRequestIDs := currentWorkflowConditionFailed.RequestIDs
 	if requestIDInfo, ok := currentWorkflowRequestIDs[request.GetRequestId()]; ok {
+		metrics.StartWorkflowRequestDeduped.With(s.metricsHandler).Record(1)
```
yycptt, line 319: I wonder if it's better to emit the metric based on the StartOutcome at a higher level.

rodrigozhou: Hm, if I move to a higher level, I'd have to init metricsHandler at all times, which would be basically not doing lazy init for metricsHandler.
Also, the other cases are successful started, reused and error. With this new metric tracking dedup, we are only missing the distinction between actual start and reused.

yycptt (approved):

PR #7567, opened 2025-04-03
Add metric for paused activities

ychebotarev:
## What changed?

Add metric for paused activities. Metric is emmited every time pause state changed to "true"

## Why?

With introduction of workflow rules pausing activity can become "mass" operation, and no initiated by direct user request.
Thus users need to know about it.

## How did you test it?

Existing tests.

## Potential risks

N/A

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -5564,6 +5564,16 @@ func (ms *MutableStateImpl) UpdateActivity(scheduledEventId int64, updater histo
 		if err != nil {
 			return err
 		}
+
+		if ai.Paused {
+			metrics.PausedActivitiesCounter.With(
+				ms.metricsHandler.WithTags(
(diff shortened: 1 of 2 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -5564,6 +5564,16 @@ func (ms *MutableStateImpl) UpdateActivity(scheduledEventId int64, updater histo
 		if err != nil {
 			return err
 		}
+
+		if ai.Paused {
+			metrics.PausedActivitiesCounter.With(
+				ms.metricsHandler.WithTags(
+					metrics.NamespaceTag(ms.namespaceEntry.Name().String()),
+					metrics.WorkflowTypeTag(ms.GetExecutionInfo().WorkflowTypeName),
+					metrics.ActivityTypeTag(ai.ActivityType.Name),
```
stephanos, line 5573: These two are ignored/dropped in Cloud (due to their cardinality), right?

stephanos (approved):

PR #7570, opened 2025-04-04
Add a system workflow to manage config

yux0:
## What changed?
Add a system workflwo to manage config

## Why?
If the failover version never used, we can update the failover version increment. We should consider add a logic to check the namespace failover history.

## How did you test it?
local test

### DIFF
--- a/service/worker/systemconfig/activities.go
+++ b/service/worker/systemconfig/activities.go
@@ -2,6 +2,7 @@ package systemconfig
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"go.temporal.io/server/common/persistence"
@@ -13,9 +14,7 @@ type (
 		NewFVI     int64
 	}
 
-	UpdateFailoverVersionIncrementOuput struct {
-		CurrentFVI int64
-		NewFVI     int64
+	UpdateFailoverVersionIncrementOutput struct {
 	}
 
 	activities struct {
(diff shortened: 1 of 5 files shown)


service/worker/systemconfig/update_failover_version.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+const (
+	updateFailoverVersionIncrementWorkflowName = "update-failover-version-increment"
+)
+
+type (
+	UpdateFailoverVersionIncrementParams struct {
+		CurrentFVI int64
+		NewFVI     int64
```
hehaifengcn, line 16: some validation that NewFVI should be > 0

service/worker/systemconfig/activities.go
```diff
@@ -0,0 +1,64 @@
...
+		NewFVI     int64
+	}
+
+	activities struct {
+		currentClusterName     string
+		clusterMetadataManager persistence.ClusterMetadataManager
+	}
+)
+
+func (a *activities) UpdateFVI(
```
hehaifengcn, line 27: do we need to restart history and other service while doing the update?

service/worker/systemconfig/activities.go
```diff
@@ -0,0 +1,64 @@
...
+) (UpdateFailoverVersionIncrementOuput, error) {
+	request := &persistence.GetClusterMetadataRequest{
+		ClusterName: a.currentClusterName,
+	}
+	resp, err := a.clusterMetadataManager.GetClusterMetadata(ctx, request)
+	if err != nil {
+		return UpdateFailoverVersionIncrementOuput{}, err
+	}
+	metadata := resp.ClusterMetadata
+	if metadata.FailoverVersionIncrement != input.CurrentFVI {
```
hehaifengcn, line 39: is this check really necessary?

service/worker/systemconfig/activities.go
```diff
@@ -0,0 +1,64 @@
...
+		return UpdateFailoverVersionIncrementOuput{},
+			fmt.Errorf("failover version increment %v does not match input version %v", metadata.FailoverVersionIncrement, input.CurrentFVI)
+	}
+	if metadata.InitialFailoverVersion >= input.NewFVI {
```
hehaifengcn, line 43: add a check that if NewFVI == FailoverVersionIncrement, do nothing.

service/worker/systemconfig/activities.go
```diff
@@ -0,0 +1,64 @@
...
+
+type (
+	UpdateFailoverVersionIncrementInput struct {
+		CurrentFVI int64
+		NewFVI     int64
+	}
+
+	UpdateFailoverVersionIncrementOuput struct {
+		CurrentFVI int64
+		NewFVI     int64
```
hehaifengcn, line 18: not used?

service/worker/systemconfig/activities.go
```diff
@@ -0,0 +1,64 @@
...
+		return UpdateFailoverVersionIncrementOuput{},
+			fmt.Errorf("failover version increment %v is less than initial failover version %v", input.NewFVI, metadata.InitialFailoverVersion)
+	}
+	if !metadata.IsGlobalNamespaceEnabled {
+		return UpdateFailoverVersionIncrementOuput{},
+			fmt.Errorf("please update failover version increment from application yaml file")
+	}
+
+	metadata.FailoverVersionIncrement = input.NewFVI
+	applied, err := a.clusterMetadataManager.SaveClusterMetadata(ctx, &persistence.SaveClusterMetadataRequest{
```
hehaifengcn, line 53: any potential race here (say cluster meta is changed by something else at the same time)?

service/worker/systemconfig/activities.go
yux0: yes

service/worker/systemconfig/activities.go
yux0: Just to be safe

service/worker/systemconfig/activities.go
yux0: no. It is protected by the version.

pglass (approved):

PR #7572, opened 2025-04-04
add PAUSED and PAUSE_REQUESTED to activity state

ychebotarev:
## What changed?

In DescribeWorkflow, for pending activities, introduce PAUSE and PAUSE_REQUESTED states.

## Why?

To clearly distinguish between "activity is paused on the server but running on the worker" vs "activity is paused on the server and not running on the worker"

## How did you test it?

Add unit tests.

## Potential risks

Some existing scripts, that rely on "activity is running means State == STARTED" may stop working.

### DIFF
--- a/service/history/workflow/activity.go
+++ b/service/history/workflow/activity.go
@@ -154,10 +154,12 @@ func GetPendingActivityInfo(
 	// adjust activity state for paused activities
 	if ai.Paused {
 		if p.State == enumspb.PENDING_ACTIVITY_STATE_SCHEDULED {
+			// this state means activity is not running on the worker
 			// if activity is paused on server and not running on worker - mark it as PAUSED
 			p.State = enumspb.PENDING_ACTIVITY_STATE_PAUSED
 		} else if p.State == enumspb.PENDING_ACTIVITY_STATE_STARTED {
-			// if activity is paused on server, but still paused on worker - mark it as PAUSE_REQUESTED
+			// this state means activity is running on the worker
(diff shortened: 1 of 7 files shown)


service/history/workflow/activity.go
```diff
@@ -151,6 +151,17 @@ func GetPendingActivityInfo(
 		}
 	}
 
+	// adjust activity state for paused activities
+	if ai.Paused {
```
drewhoskins-temporal, line 155: lgtm

service/history/workflow/activity.go
```diff
@@ -151,6 +151,17 @@ func GetPendingActivityInfo(
...
+		if p.State == enumspb.PENDING_ACTIVITY_STATE_SCHEDULED {
+			// if activity is paused on server and not running on worker - mark it as PAUSED
+			p.State = enumspb.PENDING_ACTIVITY_STATE_PAUSED
+		} else if p.State == enumspb.PENDING_ACTIVITY_STATE_STARTED {
+			// if activity is paused on server, but still paused on worker - mark it as PAUSE_REQUESTED
```
gow, line 160: This comment is a bit confusing for the first time reader given that the state is `PENDING_ACTIVITY_STATE_STARTED` . Could you reword it to add a bit more clarity?

gow (approved):

service/history/workflow/activity.go
ychebotarev: tnx, I will try to add better comment before merging.

PR #7575, opened 2025-04-05
CHASM: set and use valueState field

alexshtin:
## What changed?

CHASM: set and use `valueState` field.

## Why?

`valueState` indicates if `value` and `serializedNode` are the same and call to `serialize`/`deserialize` methods can exit earlier and save on serialization/deserialization cost.

## How did you test it?

Modified existing unit tests.

## Potential risks

No.

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -49,6 +49,15 @@ var (
 	errComponentNotFound = serviceerror.NewNotFound("component not found")
 )
 
+type valueState uint8
+
+const (
+	valueStateUndefined valueState = iota
+	valueStateSynced
+	valueStateNeedDeserialize
+	valueStateNeedSerialize
+)
+
 type (
 	// Node is the in-memory representation of a persisted CHASM node.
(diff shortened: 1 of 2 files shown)


alexshtin:
> Some U.S. highways, especially in rural areas like Texas and Utah, use 75 mph as a standard speed limit.

chasm/tree.go
```diff
@@ -901,7 +922,7 @@ func (n *Node) IsDirty() bool {
 }
 
 func (n *Node) isValueSynced() bool {
-	if n.value != nil && !n.valueSynced {
+	if !n.valueSynced {
```
yycptt, line 925: With your definition of `valuedSynced` (which is bidirectional, n.value <-> n.serializedNode), we can't use this method in `IsDirty()`. Since a tree with only serializedNode but no decoded valued is not really "Dirty" from DB pov.

yycptt (approved):
👍

PR #7578, opened 2025-04-07
Remove NewSQLAdminDB

stephanos:
## What changed?

Remove duplicate code: three implementations of `CreateAdminDB` were exactly identical to the respective `CreateDB`.

## Why?

DRY.

## How did you test it?

Existing tests.

### DIFF
--- a/common/persistence/sql/sqlplugin/interfaces.go
+++ b/common/persistence/sql/sqlplugin/interfaces.go
@@ -54,8 +54,7 @@ type VersionedBlob struct {
 type (
 	// Plugin defines the interface for any SQL database that needs to implement
 	Plugin interface {
-		CreateDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (DB, error)
-		CreateAdminDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (AdminDB, error)
+		CreateDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (any, error)
 	}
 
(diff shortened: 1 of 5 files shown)


common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -54,8 +54,7 @@ type VersionedBlob struct {
 type (
 	// Plugin defines the interface for any SQL database that needs to implement
 	Plugin interface {
-		CreateDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (DB, error)
-		CreateAdminDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (AdminDB, error)
+		CreateDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (any, error)
```
stephanos, line 57: Unfortunately, Go does not support type parameters here.

carlydf: would it help to put a comment on this function saying that the `any` is the created database? (since that can no longer be inferred from the function signature)

I mean, maybe it could be inferred simply from the fact that this is a create method, but idk just an idea!

carlydf (approved):

common/persistence/sql/sqlplugin/interfaces.go
stephanos: I can add that. I considered using an empty interface, but `DB` is already taken. It's really quite odd that you can just cast your way to the DB interface you need here.

stephanos: Added a `GenericDB` interface with the methods that both share instead, now.

PR #7583, opened 2025-04-07
Modify default replication setting

zigeH:
## What changed?

Modify default replication setting:
normal (default):
QPS: 300
ConcurrentActivityCount: 20
PageCountPerExecution:50
ListWorkflowsPageSize: 1000
Expected speed: 125 wfs/sec

### DIFF
--- a/service/worker/migration/force_replication_workflow.go
+++ b/service/worker/migration/force_replication_workflow.go
@@ -132,8 +132,10 @@ const (
 	taskQueueUserDataReplicationDoneSignalType = "task-queue-user-data-replication-done"
 	taskQueueUserDataReplicationVersionMarker  = "replicate-task-queue-user-data"
 
+	defaultOverallRps                              = 300.0
+	defaultConcurrentActivityCount                 = 20
 	defaultListWorkflowsPageSize                   = 1000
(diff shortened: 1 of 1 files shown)


service/worker/migration/force_replication_workflow.go
```diff
@@ -132,8 +132,10 @@ const (
 	taskQueueUserDataReplicationDoneSignalType = "task-queue-user-data-replication-done"
 	taskQueueUserDataReplicationVersionMarker  = "replicate-task-queue-user-data"
 
+	defaultOverallRps                              = 300.0
+	defaultConcurrentActivityCount                 = 20
 	defaultListWorkflowsPageSize                   = 1000
-	defaultPageCountPerExecution                   = 200
+	defaultPageCountPerExecution                   = 50
 	maxPageCountPerExecution                       = 1000
```
yux0, line 139: Can we maintain those setting in one place? https://github.com/temporalio/saas-control-plane/blob/b9e23f8893b08048b5b1c801477a4e0d7c4edc3d/internal/workflows/xdc/config/settings.go#L86

zigeH: make sense. Will close the pr and move the change to cp only.

PR #7585, opened 2025-04-07
[CHASM] Physical task protos

lina-temporal:
## What changed?
- Added definitions for CHASM physical tasks.

## Why?
- CHASM framework will automatically map logical tasks, as defined by component
authors, onto these two physical task types.


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -450,6 +450,19 @@ message OutboundTaskInfo {
     StateMachineTaskInfo state_machine_info = 8;
 }
 
+message ChasmTransferTaskInfo {
```
lina-temporal, line 453: I also considered adding `chasm_info` as a field off of `TransferTask` (or `OutboundTask`), but this seemed a bit cleaner. For CHASM pure tasks, since we only need the entity key and not any component's path, I think we can just reuse the TimerTaskInfo message.

service/history/tasks/chasm_task.go
```diff
@@ -0,0 +1,111 @@
...
+
+	enumsspb "go.temporal.io/server/api/enums/v1"
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+)
+
+// ChasmTaskPure is a task that applies a batch of state changes to a CHASM
+// entity. All components within the entity's tree will have their pending state
+// changes applied.
+type ChasmTaskPure struct {
+	definition.WorkflowKey // Task interface assumes WorkflowKey.
```
lina-temporal, line 40: Changing this would involve both moving out chasm's `EntityKey` and updating the `Task` proto.. didn't seem worth it, at this point.

proto/internal/temporal/server/api/enums/v1/task.proto
```diff
@@ -75,6 +75,12 @@ enum TaskType {
 
     TASK_TYPE_REPLICATION_SYNC_HSM = 30;
     TASK_TYPE_REPLICATION_SYNC_VERSIONED_TRANSITION = 31;
+
+    // A task that applies a batch of state changes to a CHASM entity.
+    TASK_TYPE_CHASM_PURE = 32;
+
+    // An immediate task generated by a CHASM component.
+    TASK_TYPE_CHASM_TRANSFER = 33;
```
yycptt, line 83: Not necessarily a transfer task right? The same physical task could be in timer/outbound queue as well.

I mean we need to support the case where chasm developers adding delayed tasks with side effect.

proto/internal/temporal/server/api/persistence/v1/executions.proto
yycptt: hmmm if we have a separate definition. How do we know which proto message to decode into after getting a blob from transfer queue?

service/history/tasks/chasm_task.go
yycptt: Agree. In term implementation can continue to used all those old names (workflowKey, workflowContext, etc.)

service/history/tasks/chasm_task.go
```diff
@@ -0,0 +1,111 @@
...
+
+var _ Task = &ChasmTaskTransfer{}
+
+func (t *ChasmTaskTransfer) GetCategory() Category {
+	return CategoryTransfer
+}
+
+func (t *ChasmTaskTransfer) GetType() enumsspb.TaskType {
+	return enumsspb.TASK_TYPE_CHASM_TRANSFER
+}
```
yycptt, line 91: 🤦 sorry I totally forgot about the Category concept we have... I think we will have to parameterize the definition if we want to reuse the same task struct across different queues. 

For PureTask I guess it would be the same thing. We may have outbound timer queue in the future, so not necessarily always CategoryTimer.

I guess I am also fine with having separate TaskType definitions, though feels a bit repetitive to me.

proto/internal/temporal/server/api/persistence/v1/executions.proto
lina-temporal: Yeah, I'm realizing this as I'm doing the deserialization. How about adding a new entry to TransferTaskInfo's `oneof task_details` for `ChasmTaskInfo`, instead? Then I can use the task type to construct the `ChasmTransferTask`.

service/history/tasks/chasm_task.go
lina-temporal: > 🤦 sorry I totally forgot about the Category concept we have... I think we will have to parameterize the definition if we want to reuse the same task struct across different queues.

Sure, like a `ChasmTask[T Category]`, and implement that across Transfer/Timer categories? 

> For PureTask I guess it would be the same thing. We may have outbound timer queue in the future, so not necessarily always CategoryTimer.

👍

proto/internal/temporal/server/api/persistence/v1/executions.proto
yycptt: yeah that could work. Note that we will need the same info in Timer/OutboundTaskInfo as well, so ideally all three should look similar and follow the same pattern.

service/history/tasks/chasm_task.go
yycptt: > like a ChasmTask[T Category]

Yeah generic, or could simply be a task category field in the struct to make things easier. Otherwise I guess serializer implementation needs to have a generic method which is not supported unless serializer itself is also generic?

And I think this category field can be in memory only & doesn't need to be persisted. When decoding a task, we know from which queue is task is loaded and the category is passed into the DeserializeTask method.

lina-temporal: Ah, if a generic wouldn't fit as well into the existing infra, agreed I'll just add it as a struct field. 

> And I think this category field can be in memory only

👍

yycptt: You will need at least one more `destination` field in this in memory task struct to handle outbound tasks. 

The proto def for OutboundTaskInfo already have that field so we don't need it in the proto ChasmTaskInfo.

service/history/tasks/chasm_task.go
```diff
@@ -0,0 +1,118 @@
...
+func (t *ChasmTaskPure) GetCategory() Category {
+	return t.Category
+}
+
+func (t *ChasmTaskPure) GetType() enumsspb.TaskType {
+	return enumsspb.TASK_TYPE_CHASM_PURE
+}
+
+func (t *ChasmTaskPure) GetKey() Key {
+	return NewKey(t.VisibilityTimestamp, t.TaskID)
```
yycptt, line 73: we need the same logic as L97 I think. The pure task can be in transfer queue as well. If, say due to system delay, the next pure task is due now or a time in the past and we want to run it as soon as possible.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -446,8 +455,14 @@ message OutboundTaskInfo {
     // Destination of this task (e.g. protocol+host+port for callbacks).
     // Outbound tasks are grouped by this field (and the namespace ID) when scheduling.
     string destination = 7;
-    // If task addresses a sub-statemachine (e.g. callback), this field will be set.
-    StateMachineTaskInfo state_machine_info = 8;
+
+    oneof task_details {
+        // If task addresses a sub-statemachine (e.g. callback), this field will be set.
+        StateMachineTaskInfo state_machine_info = 8;
```
yycptt, line 461: is this a breaking change?

lina-temporal: Not at the proto level (only for generated Go code, which I've updated):

> Changing a single optional field or extension into a member of a new oneof is binary compatible, however for some languages (notably, Go) the generated code’s API will change in incompatible ways.

and

> implicit: (not recommended) An implicit field has no explicit cardinality label and behaves as follows:
>   * if the field is a message type, it behaves just like an optional field.

 Source:
https://protobuf.dev/programming-guides/proto3/#updating

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -446,8 +455,14 @@ message OutboundTaskInfo {
...
+
+        // If the task addresses a CHASM component, this field will be set.
+        ChasmTaskInfo chasm_task_info = 17;
```
yycptt, line 464: = 9?

yycptt (approved):

proto/internal/temporal/server/api/persistence/v1/executions.proto
lina-temporal: `buf` doesn't like the change, but I think it's because we have our rules set to FILE compatibility. I'm going to temporarily ignore via buf.yaml to merge.

proto/internal/temporal/server/api/persistence/v1/executions.proto
lina-temporal: will fix before merge :)

PR #7588, opened 2025-04-07
Specify reason for versioning-related permission-denied errors

carlydf:
## What changed?
Specify reason for versioning-related permission-denied errors

## Why?
It was confusing for users when one type of versioning was enabled, but if they used an API associated with a different version of versioning, the error still just said "Worker versioning is disabled in this namespace."

### DIFF
--- a/service/frontend/errors.go
+++ b/service/frontend/errors.go
@@ -56,6 +56,7 @@ var (
 	errNamespaceTooLong                                   = serviceerror.NewInvalidArgument("Namespace length exceeds limit.")
 	errWorkflowTypeTooLong                                = serviceerror.NewInvalidArgument("WorkflowType length exceeds limit.")
 	errWorkflowIDTooLong                                  = serviceerror.NewInvalidArgument("WorkflowId length exceeds limit.")
+	errWorkflowRuleIDTooLong                              = serviceerror.NewInvalidArgument("Workflow Rule Id length exceeds limit.")
 	errSignalNameTooLong                                  = serviceerror.NewInvalidArgument("SignalName length exceeds limit.")
 	errTaskQueueTooLong                                   = serviceerror.NewInvalidArgument("TaskQueue length exceeds limit.")
 	errRequestIDTooLong                                   = serviceerror.NewInvalidArgument("RequestId length exceeds limit.")
--- a/service/frontend/errors.go
+++ b/service/frontend/errors.go
@@ -134,7 +134,7 @@ var (
 	errListNotAllowed      = serviceerror.NewPermissionDenied("List is disabled on this namespace.", "")
 	errSchedulesNotAllowed = serviceerror.NewPermissionDenied("Schedules are disabled on this namespace.", "")
 
-	errDeploymentsNotAllowed        = serviceerror.NewPermissionDenied("Deployments are disabled on this namespace.", "")
+	errDeploymentsNotAllowed        = serviceerror.NewPermissionDenied("Deployments (deprecated) are disabled on this namespace.", "")
 	errDeploymentVersionsNotAllowed = serviceerror.NewPermissionDenied("Worker Deployment Versions are disabled on this namespace.", "")
 
 	errBatchAPINotAllowed                = serviceerror.NewPermissionDenied("Batch operation feature are disabled on this namespace.", "")
--- a/service/frontend/errors.go
+++ b/service/frontend/errors.go
@@ -147,7 +147,7 @@ var (
 	errUpdateWorkflowExecutionAsyncAdmittedNotAllowed = serviceerror.NewPermissionDenied("UpdateWorkflowExecution issued asynchronously and waiting on update admitted is not supported.", "")
 	errMultiOperationAPINotAllowed                    = serviceerror.NewPermissionDenied("ExecuteMultiOperation API is disabled on this namespace.", "")
 
-	errWorkerVersioningV1_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v1.0 (Version Set-based) is disabled on this namespace.", "")
+	errWorkerVersioningV1_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v1.0 (Version Set-based, deprecated) is disabled on this namespace.", "")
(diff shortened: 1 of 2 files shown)


service/frontend/errors.go
```diff
@@ -146,7 +147,9 @@ var (
 	errUpdateWorkflowExecutionAsyncAdmittedNotAllowed = serviceerror.NewPermissionDenied("UpdateWorkflowExecution issued asynchronously and waiting on update admitted is not supported.", "")
 	errMultiOperationAPINotAllowed                    = serviceerror.NewPermissionDenied("ExecuteMultiOperation API is disabled on this namespace.", "")
 
-	errWorkerVersioningNotAllowed = serviceerror.NewPermissionDenied("Worker versioning is disabled on this namespace.", "")
+	errWorkerVersioningV1_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v1.0 (Version Set-based) is disabled on this namespace.", "")
```
ShahabT, line 150: maybe also say it's deprecated?
```suggestion
	errWorkerVersioningV1_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v1.0 (Version Set-based, deprecated) is disabled on this namespace.", "")
```

service/frontend/errors.go
```diff
@@ -146,7 +147,9 @@ var (
...
+	errWorkerVersioningV2_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v2.0 (Rules-based) is disabled on this namespace.", "")
```
ShahabT, line 151: Similarly here.
```suggestion
	errWorkerVersioningV2_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v2.0 (Rules-based, deprecated) is disabled on this namespace.", "")
```

ShahabT (approved):

service/frontend/errors.go
```diff
@@ -134,7 +134,8 @@ var (
 	errListNotAllowed      = serviceerror.NewPermissionDenied("List is disabled on this namespace.", "")
 	errSchedulesNotAllowed = serviceerror.NewPermissionDenied("Schedules are disabled on this namespace.", "")
 
-	errDeploymentsNotAllowed = serviceerror.NewPermissionDenied("Deployments are disabled on this namespace.", "")
+	errDeploymentsNotAllowed        = serviceerror.NewPermissionDenied("Deployments are disabled on this namespace.", "")
```
carlydf, line 137:
```suggestion
	errDeploymentsNotAllowed        = serviceerror.NewPermissionDenied("Deployments (deprecated) are disabled on this namespace.", "")
```

service/frontend/errors.go
```diff
@@ -146,7 +147,9 @@ var (
 	errUpdateWorkflowExecutionAsyncAdmittedNotAllowed = serviceerror.NewPermissionDenied("UpdateWorkflowExecution issued asynchronously and waiting on update admitted is not supported.", "")
 	errMultiOperationAPINotAllowed                    = serviceerror.NewPermissionDenied("ExecuteMultiOperation API is disabled on this namespace.", "")
 
-	errWorkerVersioningNotAllowed = serviceerror.NewPermissionDenied("Worker versioning is disabled on this namespace.", "")
+	errWorkerVersioningV1_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v1.0 (Version Set-based, deprecated) is disabled on this namespace.", "")
+	errWorkerVersioningV2_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v2.0 (Rules-based, deprecated) is disabled on this namespace.", "")
```
carlydf, line 151:
```suggestion
	errWorkerVersioningV1_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v0.1 (Version Set-based, deprecated) is disabled on this namespace.", "")
	errWorkerVersioningV2_0NotAllowed         = serviceerror.NewPermissionDenied("Worker versioning v0.2 (Rules-based, deprecated) is disabled on this namespace.", "")
```

service/frontend/workflow_handler.go
```diff
@@ -4422,7 +4422,7 @@ func (wh *WorkflowHandler) GetWorkerTaskReachability(ctx context.Context, reques
 	}
 
 	if !wh.config.EnableWorkerVersioningData(request.Namespace) {
-		return nil, errWorkerVersioningNotAllowed
+		return nil, errWorkerVersioningV2_0NotAllowed
```
Shivs11, line 4425: @carlydf -- would you say that this should be checking the "v2" versioning dynamic config ('frontend.workerVersioningWorkflowAPIs' and not 'frontend.workerVersioningDataAPIs', which is implicitly being checked by `EnableWorkerVersioningData`)

carlydf: `frontend.workerVersioningDataAPIs` was created for v1 and is used for the v1 data APIs:
- `UpdateWorkerBuildIdCompatibility`
- `GetWorkerBuildIdCompatibility`
- `GetWorkerTaskReachability`

`frontend.workerVersioningRuleAPIs` was created for v2 and is used for the v2 data APIs
- `UpdateWorkerVersioningRules`
- `GetWorkerVersioningRules`
in v2, "build id reachability" was queried via DescribeTaskQueue with Enhanced "mode." It [looks like](https://github.com/temporalio/temporal/blob/e8713b63ec054b4c3a2bf7f174c16144f6dbfa56/service/matching/matching_engine.go#L1470-L1471) we don't do any dynamic config capability check for describe task queue enhanced.

`system.enableDeployments` is for v3.1 (deployment and deployment series)

`system.enableDeploymentVersions` is for v3.2 that we have now (worker deployment version and worker deployment)

I thought `frontend.workerVersioningWorkflowAPIs` was used for all versions of worker versioning, not just v2? Like that for all versions of versioning, the "workflow APIs" flag controls whether workflow progress APIs are allowed to send versioning info. But, now that I'm reading the descriptions for the v3 flags, they say:
```
`EnableDeploymentVersions enables deployment versions (versioning v3) in all services,
including deployment-related RPCs in the frontend, deployment version entity workflows in the worker,
and deployment interaction in matching and history.
```
so maybe that list of "all services" includes workflow progress APIs? I think it would probably be easiest to keep `system.enableDeploymentVersions` and `frontend.workerVersioningWorkflowAPIs` in the final state, with the latter enabling version info to be sent in workflow progress APIs, as it does today.

PR #7590, opened 2025-04-08
Fix task queue rate-limit for invalid tasks

carlydf:
## What changed?
Implement `RecycleToken` in `MultiRateLimiter` and add functional tests.

## Why?
`ClockedRateLimiter` is the base rate limiter for many rate limiters in our system, so I made the mistake of assuming that implementing `RecycleToken` there and and then calling `ClockedRateLimiter.RecycleToken` from other wrapper rate limiters (such as `MultiRateLimiter` and `DynamicRateLimiter`) would "just work."

However, a limitation of the recycle token implementation is that another process must be waiting on a token (via `ClockedRateLimiter.Wait`) at the time when `RecycleToken` is called.

`DynamicRateLimiter.Wait` calls `ClockedRateLimiter.Wait`, which means that `DynamicRateLimiter.RecycleToken` can simply call `ClockedRateLimiter.RecycleToken`.

However, `MultiRateLimiter.Wait` does not call the `Wait` method of its sub-rate-limiters; instead it implements a custom `Wait` method using the `Reserve` methods of its sub-rate-limiters. This means the sub-rate-limiters never call `Wait` and never wait for the recycled token to arrive. Giving `MultiRateLimiter` a custom `RecycleToken` implementation that pairs with the custom `MultiRateLimiter.Wait` solves this problem.

## How did you test it?
Tested in a test server using sample workflows, and with functional tests for old and new matcher.
Note that New Matcher did not have this bug, because it does not use `MultiRateLimiter` for task queue rate limiting.

## Is hotfix candidate?
It fixes a bug, but the bug only occurs when many workflows that have backlogged tasks become invalid (ie. via termination or deletion).


common/quotas/multi_rate_limiter_impl.go
```diff
@@ -157,7 +159,28 @@ func (rl *MultiRateLimiterImpl) WaitN(ctx context.Context, numToken int) error {
 	case <-ctx.Done():
 		reservation.CancelAt(time.Now())
 		return ctx.Err()
+	case <-rl.recycleCh:
+		if numToken > 1 {
```
yiminc, line 163: if you got multiple recycle calls, each time is 1 token, would we drop all of them, or would we unblock if numToken of them is recycled? Shall we do numToken-- here?

yiminc: It maybe not relevant, because do we ever wait for more than 1 token at a time?

common/quotas/multi_rate_limiter_impl.go
```diff
@@ -192,10 +215,45 @@ func (rl *MultiRateLimiterImpl) TokensAt(t time.Time) int {
...
+// then later find out (usually via RPC to History) that the task should not be dispatched.
+// If history rpc takes 10ms --> 100 opportunities for the token to be used that second --> 99% recycle probability is ok.
+// If recycle probability is 50% --> need at least 2 opportunities for token to be used --> 500ms history rpc time is ok.
+// Note that the task forwarder rate limit impacts the rate of recycle, which can add inaccuracy.
 func (rl *MultiRateLimiterImpl) RecycleToken() {
-	for _, rateLimiter := range rl.rateLimiters {
-		rateLimiter.RecycleToken()
+	select {
+	case rl.recycleCh <- struct{}{}:
+	default:
```
yiminc, line 257: the token will be dropped if there is no poller waiting at the exact moment when we do recycle, even if a new poller comes just 1ms later. shall we make the recycleCh a buffered chan, even just size 1?

common/quotas/multi_rate_limiter_impl.go
prathyushpv: I remember having this discussion in the old PRs review. At that time, it was decided not to make this change since we only wait for 1 token for this use case.

But looking at this switch case, the `break` below seems to unblock goroutines that are waiting for more than one token. We don't want to unblock them right?

common/quotas/multi_rate_limiter_impl.go
prathyushpv: This is only an issue when we hit the rate limit and tasks are delayed. There should be some pollers waiting when throttling happens.

common/quotas/multi_rate_limiter_impl.go
```diff
@@ -157,7 +159,28 @@ func (rl *MultiRateLimiterImpl) WaitN(ctx context.Context, numToken int) error {
...
+		// to the recycled reservation anymore, and even if we did, Cancel on a reservation that
+		// has fully waited is a no-op, so instead we cancel the current reservation as a proxy.
+		//
+		// Since Cancel() just restores tokens to the rate limiter, cancelling the current 1-token
+		// reservation should have approximately the same effect on the actual rate as cancelling the
+		// recycled reservation.
+		//
+		// If the recycled reservation was for >1 token, cancelling the current 1-token reservation will
+		// lead to a slower actual rate than cancelling the original, so the approximation is conservative.
+		reservation.Cancel()
```
prathyushpv, line 180: Do we still need to cancel here? One token was recycled, and this wait function is giving that token to a poller by returning immediately. Cancelling the reservation could allow one more poller. This effectively allows 2 pollers on 1 recycle call.

common/quotas/multi_rate_limiter_impl.go
carlydf: ohh you're right. In the `ClockedRateLimiter` WaitN function that I implemented Recycle in last fall, the select was inside an infinite `for` loop, so `break` went right back into the select. I just fixed it.

common/quotas/multi_rate_limiter_impl.go
carlydf: We decided not to make it a buffered channel to prevent against ever exceeding the rate limit for valid tasks (even though buffer of 1 would only exceed it a little bit).

common/quotas/multi_rate_limiter_impl.go
carlydf: hm. Your concern seems right, but if the reservation is _not_ cancelled, then my understanding was that the `ReserveN` call from above would still be in effect from the perspective of the rate limiter, which means that a subsequent `ReserveN` would be given a later delay.. Honestly though, `Reservation.Cancel()` is a bit of a black box. It only guarantees that "the effect of the reservation is reversed as much as possible," so I don't feel very confident about my analysis there.

On the other hand, the mental model of:
- Invalid task's process waited for the appropriate amount of time and was given a legitimate token
- Valid task's process calls `ReserveN` to find out how long it should wait before dispatching task
- Invalid task was invalidated after querying History, not dispatched to worker
- `Recycle()` is called, unblocking the valid task's process. 
- The valid task now gets sent to a worker (basically in lieu of the invalid task that got the legitimate "ticket" to do so)

At this point, rate limiter is "even," except for the reservation that the second task's process took:
- 2 reservations
- 1 full wait period
- 1 partial wait period (valid task proc could have waited anywhere from 0ns to just under the full reservation time)
- 1 real task dispatched

I guess, I'm just not sure what the effect of `ReserveN` does to the rate limiter.

carlydf: My tests still pass when I don't cancel the reservation though, so maybe that's safer?

prathyushpv: Ah got it.. So we are using the token from recycle and cancelling the existing reservation for this request. I got a bit confused and thought that we were cancelling the old reservation that was being recycled. This is correct. Sorry about the confusion.

carlydf: no worries, it is confusing!

carlydf: that's why I put the long comment above the "Cancel" but it's still confusing. I'll put the cancel call back now

prathyushpv (approved):

PR #7595, opened 2025-04-10
CHASM: Non-Workflow Mutable State P1 

yycptt:
## What changed?

Changes are mainly on replication side:
- Update mutable state GetCurrentVersion/StartVersion/CloseVersion() methods
- Update mutable state executionInfo.LastEventTaskID. and make sure it's updated even if a transition doesn't generate any events.
- Update state based replication logic to handle no event case (more changes in https://github.com/temporalio/temporal/pull/7700).

## Why?

- CHASM runs may not have events at all. We need to make sure logic continue to work in that case.

## How did you test it?

- Existing tests
- Will have functional tests later when CHASM is ready to ensure things can work e2e.

### DIFF
--- a/common/persistence/versionhistory/version_histories.go
+++ b/common/persistence/versionhistory/version_histories.go
@@ -168,3 +168,12 @@ func SetCurrentVersionHistoryIndex(h *historyspb.VersionHistories, currentVersio
 func GetCurrentVersionHistory(h *historyspb.VersionHistories) (*historyspb.VersionHistory, error) {
 	return GetVersionHistory(h, h.GetCurrentVersionHistoryIndex())
 }
+
+// IsCurrentVersionHistoryEmpty checks if the current VersionHistory is empty.
+func IsCurrentVersionHistoryEmpty(h *historyspb.VersionHistories) (bool, error) {
+	currentVersionHistory, err := GetCurrentVersionHistory(h)
+	if err != nil {
+		return false, err
+	}
+	return IsEmptyVersionHistory(currentVersionHistory), nil
+}
--- a/service/history/ndc/workflow.go
+++ b/service/history/ndc/workflow.go
@@ -76,34 +76,38 @@ func (r *WorkflowImpl) GetVectorClock() (int64, int64, error) {
 			return 0, 0, err
 		}
 	} else {
+		// TODO: If workflow is in zombie state, it should call GetLastWriteVersion() instead of
+		// GetCloseVersion().
+		// Zombie state is handled specially in GetCloseVersion() implementation today, but we should
+		// eventually remove that special handling.
 		version, err = r.mutableState.GetCloseVersion()
 		if err != nil {
 			return 0, 0, err
 		}
 	}
 
-	lastEventTaskID := r.mutableState.GetExecutionInfo().LastEventTaskId
-	return version, lastEventTaskID, nil
+	lastRunningClock := r.mutableState.GetExecutionInfo().LastRunningClock
+	return version, lastRunningClock, nil
 }
 
 func (r *WorkflowImpl) HappensAfter(
 	that Workflow,
 ) (bool, error) {
 
-	thisLastWriteVersion, thisLastEventTaskID, err := r.GetVectorClock()
+	thisLastWriteVersion, thisLastRunningClock, err := r.GetVectorClock()
 	if err != nil {
 		return false, err
 	}
-	thatLastWriteVersion, thatLastEventTaskID, err := that.GetVectorClock()
+	thatLastWriteVersion, thatLastRunningClock, err := that.GetVectorClock()
 	if err != nil {
 		return false, err
 	}
 
 	return WorkflowHappensAfter(
 		thisLastWriteVersion,
-		thisLastEventTaskID,
+		thisLastRunningClock,
 		thatLastWriteVersion,
-		thatLastEventTaskID,
+		thatLastRunningClock,
 	), nil
 }
 
@@ -139,20 +143,20 @@ func (r *WorkflowImpl) SuppressBy(
 	// if the workflow to be suppressed has last write version being remote active
 	//  then turn this workflow into a zombie
 
-	lastWriteVersion, lastEventTaskID, err := r.GetVectorClock()
+	lastWriteVersion, lastRunningClock, err := r.GetVectorClock()
 	if err != nil {
 		return historyi.TransactionPolicyActive, err
 	}
-	incomingLastWriteVersion, incomingLastEventTaskID, err := incomingWorkflow.GetVectorClock()
+	incomingLastWriteVersion, incomingLastRunningClock, err := incomingWorkflow.GetVectorClock()
 	if err != nil {
 		return historyi.TransactionPolicyActive, err
 	}
 
 	if WorkflowHappensAfter(
 		lastWriteVersion,
-		lastEventTaskID,
+		lastRunningClock,
 		incomingLastWriteVersion,
-		incomingLastEventTaskID,
+		incomingLastRunningClock,
 	) {
 		return historyi.TransactionPolicyActive, serviceerror.NewInternal("Workflow cannot suppress workflow by older workflow")
 	}
@@ -285,9 +289,9 @@ func (r *WorkflowImpl) terminateWorkflow(
 
 func WorkflowHappensAfter(
 	thisLastWriteVersion int64,
-	thisLastEventTaskID int64,
+	thisLastRunningClock int64,
 	thatLastWriteVersion int64,
-	thatLastEventTaskID int64,
+	thatLastRunningClock int64,
 ) bool {
 
 	if thisLastWriteVersion != thatLastWriteVersion {
@@ -295,5 +299,5 @@ func WorkflowHappensAfter(
 	}
 
 	// thisLastWriteVersion == thatLastWriteVersion
-	return thisLastEventTaskID > thatLastEventTaskID
+	return thisLastRunningClock > thatLastRunningClock
 }
--- a/service/history/replication/sync_state_retriever.go
+++ b/service/history/replication/sync_state_retriever.go
@@ -522,6 +522,14 @@ func (s *SyncStateRetrieverImpl) getSyncStateEvents(ctx context.Context, workflo
 	if err != nil {
 		return nil, err
 	}
+
+	if versionhistory.IsEmptyVersionHistory(sourceHistory) {
+		return nil, nil
+	}
+
+	// TODO: we may need to handle the case where mutable state only starts to generate events during middle of its execution.
+	// In that case targetVersionHistories maybe empty and sourceVersionHistories is not empty.
+	// The LCA logic doesn't work in that case today.
 	lcaItem, _, err := versionhistory.FindLCAVersionHistoryItemFromItems(targetVersionHistories, sourceHistory.Items)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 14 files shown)


service/history/replication/sync_state_retriever.go
```diff
@@ -476,6 +476,14 @@ func (s *SyncStateRetrieverImpl) getSyncStateEvents(ctx context.Context, workflo
...
 		return nil, err
 	}
+
+	if len(sourceHistory.Items) == 0 {
+		return nil, nil
+	}
+
+	// TODO: we may need to handle the case where mutable state only starts to generate events during middle of its execution.
+	// In that case targetVersionHistories maybe empty and sourceVersionHistories is not empty.
+	// The LCA logic doesn't work in that case today.
```
yux0, line 486: can we use transition history in this case?

service/history/workflow/mutable_state_impl.go
```diff
@@ -938,30 +943,63 @@ func (ms *MutableStateImpl) GetStartVersion() (int64, error) {
 		if err != nil {
 			return 0, err
 		}
-		firstItem, err := versionhistory.GetFirstVersionHistoryItem(versionHistory)
-		if err != nil {
-			return 0, err
+
+		if !versionhistory.IsEmptyVersionHistory(versionHistory) {
```
yux0, line 947: Use IsCurrentVersionHistoryEmpty?

common/persistence/versionhistory/version_histories.go
```diff
@@ -192,3 +192,12 @@ func SetCurrentVersionHistoryIndex(h *historyspb.VersionHistories, currentVersio
 func GetCurrentVersionHistory(h *historyspb.VersionHistories) (*historyspb.VersionHistory, error) {
 	return GetVersionHistory(h, h.GetCurrentVersionHistoryIndex())
 }
+
+// IsCurrentVersionHistoryEmpty checks if the current VersionHistory is empty.
+func IsCurrentVersionHistoryEmpty(h *historyspb.VersionHistories) (bool, error) {
+	currentVersionHistory, err := GetCurrentVersionHistory(h)
```
yux0, line 198: return true if version histories is empty?

service/history/workflow/mutable_state_impl.go
```diff
@@ -938,30 +943,63 @@ func (ms *MutableStateImpl) GetStartVersion() (int64, error) {
...
+		}
+
+		// once the mutable state is closed, the execution state will no longer be updated.
+		// so the last update version is the close version.
+		if ms.executionState.LastUpdateVersionedTransition != nil {
+			return ms.executionState.LastUpdateVersionedTransition.NamespaceFailoverVersion, nil
+		}
+	}
+
+	return common.EmptyVersion, nil
```
yux0, line 1002: is this expected in some case to return empty version?

service/history/workflow/mutable_state_impl.go
```diff
@@ -5988,13 +6026,11 @@ func (ms *MutableStateImpl) closeTransaction(
...
 	}
 
-	if ms.isStateDirty() {
-		if err := ms.closeTransactionUpdateTransitionHistory(
-			transactionPolicy,
-		); err != nil {
-			return closeTransactionResult{}, err
-		}
-		ms.closeTransactionHandleUnknownVersionedTransition()
+	isStateDirty := ms.isStateDirty()
```
yux0, line 6029: move to L6041?

service/history/replication/sync_state_retriever.go
yycptt: no quite following. This method is specifically for dealing with events, and see which events are missing on the target side. Transition history doesn't give any info re. events.

service/history/workflow/mutable_state_impl.go
yycptt: We still need to get the current version history (then to get the first item) if it's not empty.

service/history/workflow/mutable_state_impl.go
yycptt: I feel we have bunch of logic probably trying to handle backward compatibility (by checking for version history is nil) and return EmptyVersion. Like GetCurrentVersion(). If you have any context on that, plz let me know.

I'd say for any new workflows/chasm runs, it's not expected to reach this line (EmptyVersion can still be returned above for local namespace). For very old workflows, i am not sure.

service/history/workflow/mutable_state_impl.go
yycptt: no, it needs to be here.  `closeTransactionPrepareEvents()` and `chasmTree.CloseTransaction()` changes the state of history builder and chasm tree and make them "clean"

I will add a comment here.

service/history/workflow/mutable_state_impl.go
```diff
@@ -938,30 +943,63 @@ func (ms *MutableStateImpl) GetStartVersion() (int64, error) {
...
+			// closing in the current transaction
+			return ms.GetCurrentVersion(), nil
+		}
+
+		// once the mutable state is closed, the execution state will no longer be updated.
+		// so the last update version is the close version.
+		if ms.executionState.LastUpdateVersionedTransition != nil {
+			return ms.executionState.LastUpdateVersionedTransition.NamespaceFailoverVersion, nil
+		}
+	}
```
yycptt, line 1000: This section is after the version history check to reduced the risk as it will not impact workflow (which always have events) in any way.

However, it assumes that if a run has events, then it must generate an event upon closing, which is not necessarily true. I will add a TODO here calling this assumption out. If we do have such cases in the future, we can address it by moving this section before the one for version history.

service/history/workflow/mutable_state_impl.go
```diff
@@ -938,30 +943,63 @@ func (ms *MutableStateImpl) GetStartVersion() (int64, error) {
...
 	}
 
 	return common.EmptyVersion, nil
 }
 
 func (ms *MutableStateImpl) GetCloseVersion() (int64, error) {
-	if ms.IsWorkflowExecutionRunning() {
+	// Do NOT use ms.IsWorkflowExecutionRunning() for the check.
+	// Zombie workflow is not considered running but also not closed.
+	if ms.executionState.State != enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED {
```
yycptt, line 969: I need to double check all the callers. This is breaking ndc workflow right now, which may call it when workflow is in zombie state.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6132,6 +6174,10 @@ func (ms *MutableStateImpl) closeTransactionUpdateTransitionHistory(
 		ms.GetCurrentVersion(),
 	)
 
+	if len(workflowEventsSeq) == 0 {
+		ms.executionInfo.LastUpdateClock = ms.shard.CurrentVectorClock().GetClock()
+	}
```
yycptt, line 6179: hmm I realized I can't do this. this field can't be updated after workflow is closed. 
otherwise the ndc vector clock for an old workflow can be larger than the current workflow if the old workflow is keep getting updated after it's closed. 

I will need to figure out something else...

yycptt: ok I changed the logic to only update the field when workflow is running, and renamed this field to LastRunningClock

service/history/workflow/mutable_state_impl.go
yycptt: Guaranteeing caller not calling this when workflow is zombie will require a big refactoring. For now, I special handled zombie case and returns LastWriteVersion instead.

service/history/ndc/workflow.go
```diff
@@ -163,29 +165,29 @@ func (r *WorkflowImpl) SuppressBy(
...
-		lastEventTaskID,
+		lastRunningClock,
 		incomingLastWriteVersion,
-		incomingLastEventTaskID,
+		incomingLastRunningClock,
 	) {
 		return historyi.TransactionPolicyActive, serviceerror.NewInternal("Workflow cannot suppress workflow by older workflow")
 	}
 
-	// if workflow is in zombie or finished state, keep as is
```
xwduan, line 184: Move those lines seems changed semantic a little bit, what's the reason of moving?

service/history/replication/sync_state_retriever.go
xwduan: Distinguish this case seems straight forward:
target transition history is not empty but target version history is empty?

service/history/workflow/mutable_state_impl.go
```diff
@@ -6261,6 +6320,32 @@ func (ms *MutableStateImpl) closeTransactionHandleUnknownVersionedTransition() {
 	}
 }
 
+func (ms *MutableStateImpl) closeTransactionUpdateLastRunningClock(
+	transactionPolicy historyi.TransactionPolicy,
+	workflowEventsSeq []*persistence.WorkflowEvents,
+) {
+	if transactionPolicy != historyi.TransactionPolicyActive {
```
xwduan, line 6327: Are we replicating the vector clock from active to passive?

service/history/ndc/workflow.go
yycptt: hmm which semantic get changed? The WorkflowHappensAfter logic is only a sanity check that should never be triggered. 

I was thinking the sanity check is unnecessary if we won't really do any mutations here (it should already be done before when putting the workflow into zombie state).

service/history/replication/sync_state_retriever.go
yycptt: Not sure I follow the discussion here...

The TODO comment is just saying that's case is not supported today (maybe we never need to support it) & why/where (LCA implementation on 487 can't handle it). Not it's hard to support or describing how to support it.

service/history/workflow/mutable_state_impl.go
yycptt: Yes the LastRunningClock (was LastEventTaskID), is part of executionInfo and won't be  sanitized during replication. We only check the clock when the failover version are the same, so we know the clock is from the same cluster and comparable.

common/persistence/versionhistory/version_histories.go
yycptt: I reverted my change to all empty version histories. 
I think version histories should never be empty. Even if mutable state have no events at all, the version histories can have one version history when empty items.

service/history/ndc/workflow.go
yycptt: In any case, I have reverted this change. Doesn't feel strong about it anyway.

xwduan (approved):

service/history/workflow/mutable_state_impl.go
```diff
@@ -909,10 +909,15 @@ func (ms *MutableStateImpl) UpdateCurrentVersion(
 }
 
 func (ms *MutableStateImpl) GetCurrentVersion() int64 {
+	// TODO: can we always return ms.currentVersion here?
 	if ms.executionInfo.VersionHistories != nil {
 		return ms.currentVersion
 	}
 
+	if ms.transitionHistoryEnabled && len(ms.executionInfo.TransitionHistory) != 0 {
```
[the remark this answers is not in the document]
yycptt: I think it's not necessary. I guess those checks exists only because we want to support very very old workflows in DB which is created even before VersionHistory & xdc replication is a thing. 
So for almost all workflows, VersionHistories will not a nil and the logic will just return `ms.currentVersion`.

I added those new logic just to be consistent with we already have, but TBH I don't really expect the logic to ever reach them...

PR #7605, opened 2025-04-14
use initialVersionedTransition for Nexus completion token

hai719:
## What changed?
Use initialVersionedTransition as ms versionedTransition for state machine reference in the Nexus completion token.

## Why?
When workflow switches branches, the ms versionedTransition in Nexus completion token may become stale and cannot pass the check when completion callback is triggered.


yycptt (approved):
Is there some unit/functional testing we can add to verify this works?

PR #7606, opened 2025-04-15
CHASM: Generate physical tasks and carry over physical task status

yycptt:
## What changed?

- Generate physical tasks
- Carryover cluster local task status during replication

## Why?

- CHASM task support

## How did you test it?

- Added unit tests

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -25,6 +25,7 @@
 package chasm
 
 import (
+	"cmp"
 	"fmt"
 	"reflect"
 	"time"
@@ -34,10 +35,12 @@ import (
 	"go.temporal.io/api/serviceerror"
 	persistencespb "go.temporal.io/server/api/persistence/v1"
 	"go.temporal.io/server/common/clock"
+	"go.temporal.io/server/common/definition"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/persistence/serialization"
 	"go.temporal.io/server/common/persistence/transitionhistory"
 	"go.temporal.io/server/common/softassert"
+	"go.temporal.io/server/service/history/tasks"
 	"google.golang.org/protobuf/proto"
 )
 
@@ -56,6 +59,11 @@ const (
 	valueStateSynced
 	valueStateNeedDeserialize
 	valueStateNeedSerialize
+)
+
+const (
+	physicalTaskStatusNone int32 = iota
+	physicalTaskStatusCreated
 )
 
 type (
@@ -81,6 +89,12 @@ type (
 		// NOTE: This is a different concept from the IsDirty() method needed by MutableState which means
 		// if the state in memory matches the state in DB.
 		valueState valueState
+
+		// TODO: Consider storing encoded path for the node.
+		// Consider using unique package as well.
+		// Encoded path for different runs of the same Component type are the same.
+		//
+		// encodedPath string
 	}
 
 	// nodeBase is a set of dependencies and states shared by all nodes in a CHASM tree.
@@ -119,12 +133,17 @@ type (
 		// TODO: Add methods needed from MutateState here.
 		GetCurrentVersion() int64
 		NextTransitionCount() int64
+		GetWorkflowKey() definition.WorkflowKey
+		AddTasks(...tasks.Task)
 	}
 
 	// NodePathEncoder is an interface for encoding and decoding node paths.
 	// Logic outside the chasm package should only work with encoded paths.
 	NodePathEncoder interface {
 		Encode(node *Node, path []string) (string, error)
+		// TODO: Return a iterator on node name instead of []string,
+		// so that we can get a node by encoded path without additional
+		// allocation for the decoded path.
 		Decode(encodedPath string) ([]string, error)
 	}
 )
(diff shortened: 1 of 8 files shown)


yycptt:
Exploring another approach for tracking physical pure task status... Closing this one for now. May reopen later.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -61,11 +61,24 @@ message ChasmComponentAttributes {
...
         // Versioned transition of the Entity when the task was created.
         VersionedTransition versioned_transition = 5;
+        // The xth task generated in this versioned transition.
+        // Together with the versioned transition, this is a unique identifier for
+        // this task.
+        int64 versioned_transition_offset = 6;
+        // If a physical task is created for this task in this cluster.
+        // NOTE: this is a cluster-specific field and can not be replicated.
+        // Changes to this field also doesn't require an increase in versioned transition. 
+        int32 physical_task_status = 7;
```
yycptt, line 71: ok I still think this is the easier and more intuitive approach.

The alternative I was exploring was: store a list of timestamps in mutable state (so tree level, and, similar to physical_task_status, a cluster specific field) for which physical pure tasks have been created. New physical pure tasks can generated if the scheduled timestamp is smaller than all timestamps tracked in the list. When processing a physical pure tasks, remove all timestamps less than now from the list. 
Neither fields added in the PR will be needed in this approach. Also no task status carryover logic upon replication.

The main complexity with that approach however is that we need to update this timestamp list whenever a node/task is removed. Also since we don't really know the timestamp corresponds to which node/task, we have to check the entire tree before removing a timestamp (or maintain an in memory ref count for all timestamps). 
If we don't update the list when node/tasks is removed, later at task processing time, we'll have to do an additional write operation to update the list and generate new physical pure tasks even if no chasm component task actually get processed.

So I feel maintaining this list is more complex than doing the carryover logic. The `versioned_transition_offset` can be used for purposes other than carrying over task status. For example, the physical side effect task can now store the offset and point a particular component task instead of duplicating task data again in history queues.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -61,11 +61,24 @@ message ChasmComponentAttributes {
...
 
     // Fully qualified type name of a registered component.
     string type = 1;
-    repeated Task tasks = 3;
+    // Tasks are in their insertion order, 
+    // i.e. by versioned transtion and index_in_transition.
+    repeated Task side_effect_tasks = 2;
+    // Tasks are ordered by their scheduled time, breaking ties by
+    // versioned transition and index_in_transition. 
+    repeated Task pure_tasks = 3;
```
yycptt, line 81: Separated those two tasks type so we can order them differently and make it more efficient for pure task processing. The processing logic doesn't have to loop through all tasks and see if the task is a pure task and is due for processing.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -100,5 +113,5 @@ message ChasmTaskInfo {
 
     // Opaque attached task data. May be nil. Usable by components, not the CHASM
     // framework itself.
-    bytes data = 3;
+    temporal.api.common.v1.DataBlob data = 3;
```
yycptt, line 116: We may have different encoding in the future so using a DataBlob here. 
However, like commented  above, we could store the `versioned_transition_offset` here and avoid double storing task data.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -86,8 +99,8 @@ message ChasmComponentRef {
     // reference was created. Can be used to invalidate this reference.
     VersionedTransition component_last_update_versioned_transition = 2;
 
-    // Full path to the component.
-    repeated string path = 3;
+    // Encoded full path to the component.
+    string path = 3;
```
yycptt, line 103: Makes the generate task logic a bit easier since we already have the encoded path for all updated nodes and thus avoiding creating the string slices for node names.

When processing the task, we need to do an additional decoding step. However, as the TODO commented added in the PathEncoder interface says, we can do that without any additional allocation by having the Decode method return an iterator.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt, line 71: I could use a bool for task status, but int is more flexible and it's var encoded anyway.

chasm/tree.go
```diff
@@ -926,3 +1088,64 @@ func newNode(
...
+}
+
+func taskCategory(
+	task *persistencespb.ChasmComponentAttributes_Task,
+) (tasks.Category, error) {
+	isImmeidate := task.ScheduledTime.AsTime().Equal(TaskScheduledTimeImmediate)
+
+	if task.Destination != "" {
+		if !isImmeidate {
+			return tasks.Category{}, serviceerror.NewInternal(
```
yycptt, line 1140: Need outbound timer queue for this. 
Or maybe we can just run the task in timer queue if there's no strict destination isolation guarantee for the task.

chasm/tree.go
```diff
@@ -668,10 +687,126 @@ func (n *Node) AddTask(
...
+	if err != nil {
+		return err
+	}
+	n.mutation.UpdatedNodes[encodedPath] = firstTaskNode.serializedNode
+	return nil
+}
+
+func (n *Node) walk(
+	visitor func(node *Node) error,
+) error {
```
alexshtin, line 796: I think this can be an iterator too.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -61,11 +61,24 @@ message ChasmComponentAttributes {
...
+        // NOTE: this is a cluster-specific field and can not be replicated.
+        // Changes to this field also doesn't require an increase in versioned transition. 
+        int32 physical_task_status = 7;
     }
 
     // Fully qualified type name of a registered component.
     string type = 1;
-    repeated Task tasks = 3;
+    // Tasks are in their insertion order, 
+    // i.e. by versioned transtion and index_in_transition.
```
alexshtin, line 77: Did you mean `versioned_transition_offset` instead of `index_in_transition`?

chasm/tree.go
```diff
@@ -668,10 +687,126 @@ func (n *Node) AddTask(
 func (n *Node) CloseTransaction() (NodesMutation, error) {
 	defer n.cleanupTransaction()
 
+	if err := n.closeTransactionGenratePhysicalSideEffectTasks(); err != nil {
```
lina-temporal, line 690: (minor) typo - `closeTransactionGenratePhysicalSideEffectTasks` -> `closeTransactionGeneratePhysicalSideEffectTasks`

chasm/tree.go
```diff
@@ -926,3 +1088,67 @@ func newNode(
...
+func carryOverTaskStatus(
+	sourceTasks, targetTasks []*persistencespb.ChasmComponentAttributes_Task,
+	compareFn func(a, b *persistencespb.ChasmComponentAttributes_Task) int,
+) {
+	sourceIdx, targetIdx := 0, 0
+	for sourceIdx < len(sourceTasks) && targetIdx < len(targetTasks) {
+		sourceTask := sourceTasks[sourceIdx]
+		targetTask := targetTasks[targetIdx]
+
+		switch compareFn(sourceTask, targetTask) {
```
lina-temporal, line 1116: Can we comment each of these cases?

yycptt: > if any physical task from a specific versioned transition has been created, the rest must have also been created within the same transaction

That's true for side effect tasks, but not pure tasks. So theoretically yes, I don't need to do this for side effect tasks, and can just mark the status for tasks created in previous VTs as created. But since we need the logic anyway for pure tasks, I can just reuse it for side effect tasks. The logic here also doesn't make any assumption re. how generate physical task works.

The carry over logic here is very similar to merging two sorted list, but instead of merging them, we just copy the task status field from source to target when two tasks matches.
So the cases here are:
- 0: two tasks match, copy the status and advance the pointer in both lists.
- -1: source has a smaller key, meaning the task has been deleted in the target cluster, and we don't need to copy the status. Move on to the next task in the source list.
- 1: source has a larger key, meaning there's a new task inserted in the middle. Nothing to carryover for this new task, but since this is a new task, we make sure its task status is None by reseting the field (since source cluster may have this field set and replicated.)

I will add  ^ as  comment if that helps clear your confusion :)

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt: yeah that was my intention. Will fix.

`index_in_transition` was the original name for that proto field. But linter complains that `in` can't be used in field name 🤦

chasm/tree.go
lina-temporal: Yes, thank you, that clears it up. Yeah, the notes on "-1 for task deleted in target cluster" and "1 for new task in source cluster" are what I'm after.

lina-temporal (approved):

chasm/tree.go
```diff
@@ -49,6 +52,11 @@ var (
 	errComponentNotFound = serviceerror.NewNotFound("component not found")
 )
```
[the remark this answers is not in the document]
yycptt: hmm.. int32 is much easier in this case as the value needs to be stored in a proto message which has type int32.

chasm/tree.go
yycptt: I guess the walk() method itself is already kinda an iterator :)

PR #7607, opened 2025-04-15
Versioning tests: make task queue propagation check more specific, skip flaky v2 cron test

carlydf:
## What changed?
Previously, the test helper confirming that a version had propagated out to each task queue partition only checked that the version existed in the tq user data, not whether the version had the expected routing info. 

I changed that so that it only returns success if the ramping version is ramping in the tq, current is current in the tq, etc.

I also added more checks after each routing config change operation in the failing `TestVersioning3FunctionalSuite/TestUnpinnedWorkflowWithRamp_ToUnversioned/NoTaskForwardNoPollForwardAllowSync` test, so that if there is an issue with task queue propagation that caused the error:
```
versioning_3_test.go:1856: 
        Error Trace:	/home/runner/work/temporal/temporal/tests/versioning_3_test.go:1856
        Error:      	Received unexpected error:
                   	failed to poll workflow task: taskpoller test helper timed out while waiting for the PollWorkflowTaskQueue API response, meaning no workflow task was ever created
        Test:       	TestVersioning3FunctionalSuite/TestUnpinnedWorkflowWithRamp_ToUnversioned/NoTaskForwardNoPollForwardAllowSync
```
We should see it.

## Why?
Timing out task poll makes me think that the task didn't make it into the correct queue, which could be due to a routing problem. Currently, our APIs are consistent, so the server already checks task queue propagation before returning success from `SetCurrentVersion` or `SetRampingVersion`. But I'd like to see these requirements succeed and I can't think of anything else..

Another reason to add this, is that these tests don't always use `SetCurrentVersion` to change the routing config; sometimes the tests do it directly by syncing user data. Which loses the critical step that the activity does for us. `TestUnpinnedQuery` for example, does not use SetCurrent to set current. The stricter requirements for checking user data propagation might help with these similar errors too.
```
   versioning_3_test.go:1938: 
        	Error Trace:	/home/runner/work/temporal/temporal/tests/versioning_3_test.go:1938
        	            				/home/runner/work/temporal/temporal/common/testing/taskpoller/taskpoller.go:468
        	            				/home/runner/work/temporal/temporal/common/testing/taskpoller/taskpoller.go:458
        	            				/home/runner/work/temporal/temporal/common/testing/taskpoller/taskpoller.go:160
        	            				/home/runner/work/temporal/temporal/tests/versioning_3_test.go:1935
        	            				/opt/hostedtoolcache/go/1.23.2/x64/src/runtime/asm_amd64.s:1700
        	Error:      	old deployment should not receive query
        	Test:       	TestVersioning3FunctionalSuite/TestUnpinnedQuery_NoSticky/NoTaskForwardForcePollForwardAllowSync
```

```
versioning_3_test.go:1856: 
        	Error Trace:	/home/runner/work/temporal/temporal/tests/versioning_3_test.go:1856
        	            				/home/runner/work/temporal/temporal/tests/versioning_3_test.go:1863
        	            				/opt/hostedtoolcache/go/1.23.2/x64/src/runtime/asm_amd64.s:1700
        	Error:      	Received unexpected error:
        	            	failed to poll workflow task: taskpoller test helper timed out while waiting for the PollWorkflowTaskQueue API response, meaning no workflow task was ever created
        	Test:       	TestVersioning3FunctionalSuite/TestUnpinnedQuery_Sticky/NoTaskForwardForcePollForwardForceAsync
```

```
        	Error Trace:	/home/runner/work/temporal/temporal/tests/versioning_3_test.go:348
        	            				/home/runner/work/temporal/temporal/tests/versioning_3_test.go:333
        	            				/home/runner/work/temporal/temporal/tests/versioning_3_test.go:287
        	            				/home/runner/work/temporal/temporal/tests/testcore/functional_test_base.go:577
        	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:115
        	Error:      	Received unexpected error:
        	            	context deadline exceeded
        	Test:       	TestVersioning3FunctionalSuite/TestUnpinnedQuery_Sticky/NoTaskForwardNoPollForwardForceAsync
```

### DIFF
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -377,7 +377,7 @@ func (s *Versioning3Suite) testPinnedWorkflowWithLateActivityPoller() {
 			s.NotNil(task)
 			return respondWftWithActivities(tv, tv, false, vbUnpinned, "5"), nil
 		})
-	s.waitForDeploymentDataPropagation(tv, false, tqTypeWf)
+	s.waitForDeploymentDataPropagation(tv, versionStatusInactive, false, tqTypeWf)
 
 	override := tv.VersioningOverridePinned()
 	s.startWorkflow(tv, override)
@@ -542,11 +542,14 @@ func (s *Versioning3Suite) testUnpinnedWorkflowWithRamp(toUnversioned bool) {
 
 	// v1 is current and v2 is ramping at 50%
 
-	// Make sure both TQs are registered in v1 which is the current version. This is to make sure
+	// Make sure both TQs are registered in v1 which will be the current version. This is to make sure
 	// we don't get to ramping while one of the TQs has not yet got v1 as it's current version.
 	// (note that s.setCurrentDeployment(tv1) can pass even with one TQ added to the version)
-	s.waitForDeploymentDataPropagation(tv1, false, tqTypeWf, tqTypeAct)
+	s.waitForDeploymentDataPropagation(tv1, versionStatusInactive, false, tqTypeWf, tqTypeAct)
 	s.setCurrentDeployment(tv1)
+
+	// wait until all task queue partitions know that tv1 is current
+	s.waitForDeploymentDataPropagation(tv1, versionStatusCurrent, false, tqTypeWf, tqTypeAct)
 
 	w2 := worker.New(sdkClient, tv2.TaskQueue().GetName(), worker.Options{
 		BuildID:                 tv2.BuildID(),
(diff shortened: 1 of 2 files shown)


tests/versioning_3_test.go
```diff
@@ -2021,8 +2037,18 @@ func (s *Versioning3Suite) warmUpSticky(
 	)
 }
 
+type versionStatus int
+
```
carlydf, line 2041: one sec, moving this to top of file

carlydf: and removing -1

ShahabT (approved):

PR #7609, opened 2025-04-15
fix cryptic error message for FailedPrecondition failures

Shivs11:
## What changed?

- `FailedPrecondition` Errors from the deployment workflows are returned as `FailedPrecondition` from the deployment client.

## Why?

- Right now, if a user were to call SetCurrent/SetRamping and our internal deployment workflows were to throw a FailedPrecondition error, the deployment client would convert them to `Internal` errors which in-turn would be masked by gRPC interceptors. 
- The error message would be super vague and something like (inside actual):
image

## How did you test it?

- Altered existing tests to check for this.

### DIFF
--- a/service/worker/workerdeployment/util.go
+++ b/service/worker/workerdeployment/util.go
@@ -95,8 +95,8 @@ const (
 	errConflictTokenMismatchType  = "errConflictTokenMismatch"
 	errFailedPrecondition         = "FailedPrecondition"
 
-	ErrRampingVersionDoesNotHaveAllTaskQueues = "New ramping version does not have all the task queues from the previous current version and some missing task queues are active and would become unversioned after setting the new version as the ramping version"
-	ErrCurrentVersionDoesNotHaveAllTaskQueues = "New current version does not have all the task queues from the previous current version and some missing task queues are active and would become unversioned after setting the new version as the current version"
+	ErrRampingVersionDoesNotHaveAllTaskQueues = "New ramping version is missing active task queues from the current version; these would become unversioned after setting the new version as the ramping version"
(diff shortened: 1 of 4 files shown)


service/worker/workerdeployment/util.go
```diff
@@ -93,6 +93,10 @@ const (
 	errVersionHasPollers          = "Version cannot be deleted since it has active pollers."
 	errVersionIsCurrentOrRamping  = "Version cannot be deleted since it is current or ramping."
 	errConflictTokenMismatchType  = "errConflictTokenMismatch"
+	errFailedPrecondition         = "FailedPrecondition"
+
+	ErrRampingVersionDoesNotHaveAllTaskQueues = "New ramping version does not have all the task queues from the previous current version and some missing task queues are active and would become unversioned after this operation"
```
ShahabT, line 98: "after this operation" is slightly inaccurate. More accurate would be "after setting the new version as the current version"

ShahabT (approved):

service/worker/workerdeployment/util.go
carlydf: I worked with ChatGPT to get the error more concise but still specific and came up with this. WDYT? 
`"new ramping version is missing active task queues from the current version; these would become unversioned if the new version is promoted"` ([chat link](https://chatgpt.com/share/67fe93b2-4588-8004-86f6-b95fdd21c741))

carlydf: for the current version error, you might want to say "existing current version" to specify. like
`"new current version is missing active task queues from the existing current version; these would become unversioned if the new version is promoted"`

carlydf: in both cases, idk if it makes more sense to say `proposed ramping version` / `proposed current version` instead of `new` because it's not really `new` yet?

PR #7610, opened 2025-04-15
change to unavailable error to enable retries

jmbarzee:
Change error to retry-able type

## Why?
So the error can result in the request being retryied

## How did you test it?
N/A

## Potential risks
Not much. Should be protected by rate limiting and retry logic

## Documentation
N/A

## Is hotfix candidate?
No


common/persistence/visibility/store/elasticsearch/client/backoff.go
```diff
@@ -0,0 +1,40 @@
+package client
```
rodrigozhou, line 1: What is this file?

rodrigozhou (changes requested):

common/persistence/visibility/store/elasticsearch/client/backoff.go
jmbarzee: Ahahahah! how did git not show me this in my history

rodrigozhou (approved):

PR #7613, opened 2025-04-16
Make SyncBatchSize a local state member for versioning wfs + Flake fix

Shivs11:
## What changed?

- Make `SyncBatchSize` a member of the local state for the two entity workflows used in versioning-0.31 instead of a global variable.
- This was done so that we can use testhooks which would inject custom values for `SyncBatchSize` during testing so that functional tests don't error with `“pg: too many connections"` while running the whole suite.
- A reason why I made `SyncBatchSize` a member of the local state for these workflows is because of @stephanos's helpful comment which mentioned that I should not be altering a global variable in my tests.

## Why?

- Flakes with errors like: `“pg: too many connections"` were being caused because the tests I had written had a large number of open connections to db. These open connections were made because I had a large number of task-queues registered in a single version so that I could see if the code was making the right number of batches as well as preserving the functionality.

## How did you test it?

- Running each of the solved flake 100 times across all DB offerings.
- Running the whole suite.
- Technically, one could have solved this by increasing the total number of connections a DB is okay with at a given point in time but I didn't want to do it because I was not sure about all the potential risks it got along.

### DIFF
--- a/service/worker/workerdeployment/workflow_test.go
+++ b/service/worker/workerdeployment/workflow_test.go
@@ -348,11 +348,11 @@ func (s *WorkerDeploymentSuite) syncUnversionedRampInBatches(totalWorkers int) {
 
 	// Mock the SyncDeploymentVersionUserData activity and expect it to be called totalWorkers times
 	var totalBatches int
-	batchSize := s.workerDeploymentClient.getSyncBatchSize()
-	if totalWorkers%int(batchSize) == 0 {
-		totalBatches = totalWorkers / int(batchSize)
+	batchSize := int(s.workerDeploymentClient.getSyncBatchSize())
+	if totalWorkers%batchSize == 0 {
+		totalBatches = totalWorkers / batchSize
 	} else {
-		totalBatches = totalWorkers/int(batchSize) + 1
+		totalBatches = totalWorkers/batchSize + 1
 	}
 
 	s.env.OnActivity(a.SyncDeploymentVersionUserDataFromWorkerDeployment, mock.Anything, mock.Anything).Times(totalBatches).Return(nil, nil)


service/matching/physical_task_queue_manager.go
```diff
@@ -659,9 +659,14 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeploymentVersion(
 	// we need to update the deployment workflow to tell it about this task queue
 	// TODO: add some backoff here if we got an error last time
 
+	testhook_sync_batch_size := 0
```
stephanos, line 662: Doesn't this need to be 25 still?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -620,6 +620,10 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
 				Data:  data,
 			})
 
+			if d.TesthookTaskQueuesSyncBatchSize > 0 {
+				syncBatchSize = int(d.TesthookTaskQueuesSyncBatchSize)
```
stephanos, line 624: Isn't `syncBatchSize` a global variable? Even in the context of testing it's not the best idea to mutate global state.

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -173,6 +179,10 @@ message RegisterWorkerInWorkerDeploymentArgs {
     temporal.api.enums.v1.TaskQueueType task_queue_type = 2;
     int32 max_task_queues = 3;
     WorkerDeploymentVersion version = 4;
+    
+    // populate testHook values as needed
+    int32 testhook_task_queues_sync_batch_size = 5;
```
stephanos, line 184: I don't think you should embed the testhook concept here. Can't this just be `sync_batch_size`? ie make the special case the regular case. That way the caller can set it to 25 by default, or override it.

service/worker/workerdeployment/version_workflow.go
Shivs11: made a bunch of changes and addressed this! Super helpful!

service/worker/workerdeployment/workflow.go
```diff
@@ -106,8 +106,10 @@ func (d *WorkflowRunner) listenToSignals(ctx workflow.Context) {
 }
 
 func (d *WorkflowRunner) run(ctx workflow.Context) error {
-	if d.State == nil {
-		d.State = &deploymentspb.WorkerDeploymentLocalState{}
+	if d.GetState().GetCreateTime() == nil {
+		if d.State == nil {
+			d.State = &deploymentspb.WorkerDeploymentLocalState{}
+		}
```
Shivs11, line 112: I spent sometime before deciding write stuff like this - let me explain the issue first and why I have made these changes:

Problem: We initialize `d.State` from line 109 right now. I have made `SyncBatchSize` a member of `d.State` since I think this is the most appropriate location for it's existence. Thus, I initialize and pass this value from the client itself. However, before these changes, the client passed `d.State` as `nil` and hence the check here was if it's nil, create and assign the struct as shown. 

Now, this check is no longer correct since I am indeed creating the struct in in the client by passing forward the value for `SyncBatchSize`. Thus, I thought an appropriate way of still keeping the core functionality correct was to check for `d.GetState().GetCreateTime()` and accordingly create/assign this struct.

Additionally: Why do I still have:

```
		if d.State == nil {
			d.State = &deploymentspb.WorkerDeploymentLocalState{}
		}

```

My replay test caught the fact that there could still be workflow histories that are expecting the value of `d.State` to be first initialized before carrying on with assigning the values.

service/worker/workerdeployment/client.go
```diff
@@ -1544,3 +1551,12 @@ func (d *ClientImpl) RegisterWorkerInVersion(
...
 	return nil
 }
+
+func (d *ClientImpl) getSyncBatchSize() int32 {
+	defaultSyncBatchSize := int32(25)
+	if n, ok := testhooks.Get[int](d.testHooks, testhooks.TaskQueuesInDeploymentSyncBatchSize); ok && n > 0 {
+		// In production, the testhook would be set to 0 and never reach here!
+		defaultSyncBatchSize = int32(n)
+	}
+	return defaultSyncBatchSize
```
stephanos, line 1561: nit: I think default is a misnomer here.

```suggestion
	syncBatchSize := int32(25)
	if n, ok := testhooks.Get[int](d.testHooks, testhooks.TaskQueuesInDeploymentSyncBatchSize); ok && n > 0 {
		// In production, the testhook would be set to 0 and never reach here!
		syncBatchSize = int32(n)
	}
	return syncBatchSize
```

service/worker/workerdeployment/version_workflow_test.go
```diff
@@ -55,6 +56,9 @@ func (s *VersionWorkflowSuite) SetupTest() {
 	s.controller = gomock.NewController(s.T())
 	s.env = s.WorkflowTestSuite.NewTestWorkflowEnvironment()
 	s.env.RegisterWorkflow(VersionWorkflow)
+
+	// Initialize an empty ClientImpl to use its helper methods
+	s.workerDeploymentClient = &ClientImpl{}
```
stephanos, line 61: Isn't this just a unit test? Couldn't you just hard-code the batch size?

service/worker/workerdeployment/workflow_test.go
```diff
@@ -344,10 +348,10 @@ func (s *WorkerDeploymentSuite) syncUnversionedRampInBatches(totalWorkers int) {
 
 	// Mock the SyncDeploymentVersionUserData activity and expect it to be called totalWorkers times
 	var totalBatches int
-	if totalWorkers%syncBatchSize == 0 {
-		totalBatches = totalWorkers / syncBatchSize
+	if totalWorkers%int(s.workerDeploymentClient.getSyncBatchSize()) == 0 {
+		totalBatches = totalWorkers / int(s.workerDeploymentClient.getSyncBatchSize())
 	} else {
-		totalBatches = totalWorkers/syncBatchSize + 1
+		totalBatches = totalWorkers/int(s.workerDeploymentClient.getSyncBatchSize()) + 1
```
stephanos, line 354: nit: maybe add `batchSize := s.workerDeploymentClient.getSyncBatchSize()` at the top

service/worker/workerdeployment/version_workflow_test.go
Shivs11: I could but I didn't like that option for the following reason - if someone were to change the `default` value of the syncBatchSize, they shouldn't have their unit-tests failing because the value changed - the change should be seamless in the sense that the tests should be smart enough to incorporate the new default value

stephanos (approved):

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -111,6 +111,9 @@ message VersionLocalState {
     // Key: Task Queue Name
     map<string, TaskQueueFamilyData> task_queue_families = 10;
 
+    // Number of task queues which will be synced in a single batch.
+    int32 sync_batch_size = 11;
```
dnr, line 115: Why does this value have to be persisted in version or deployment state at all? It's only ever overridden with a test hook, right? So just have a global function to get the value (from a test hook or default) where it's needed

service/worker/workerdeployment/version_workflow_test.go
dnr: It doesn't make sense to make the real code more complicated just for that reason. It's totally fine to ask someone to update a unit test when they update a constant.

proto/internal/temporal/server/api/deployment/v1/message.proto
stephanos: I haven't worked with our worker workflows too much, but I assumed that the testhook cannot be injected there? The first version of the PR suggested a global variable, so I suggested this direction instead. If the testhook is (safely) accessible from a worker workflow, then that would indeed be easier.

stephanos: I see that there's an `fx.go` for worker deployment. @Shivs11 could you obtain the testhook instance from there so you don't have to pass it via proto?

PR #7615, opened 2025-04-16
Populate PauseInfo for activities in DescribeWorkflow response

ychebotarev:
## What changed?

* Add proto messages for PauseInfo to ActivityInfo (basically the same as in API)
* Populate PauseInfo for activities paused manually
* Populate PauseInfo for activities paused by the rules
* Populate PauseInfo in DescribeWorkflow response
* Add tests (mostly functional).

## Why?

Part of the activity commands work. If activity is paused - users should be able to get quick answer to "Who/When/Why" questons.

## How did you test it?

Add functional tests

### DIFF
--- a/service/history/workflow/activity_test.go
+++ b/service/history/workflow/activity_test.go
@@ -335,7 +335,7 @@ func (s *activitySuite) TestResetPausedActivityAcceptance() {
 	ai := s.AddActivityInfo()
 
 	prevStamp := ai.Stamp
-	err := PauseActivity(s.mutableState, ai.ActivityId)
+	err := PauseActivity(s.mutableState, ai.ActivityId, nil)
 	s.NoError(err)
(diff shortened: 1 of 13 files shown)


service/history/workflow/activity_test.go
```diff
@@ -335,7 +335,7 @@ func (s *activitySuite) TestResetPausedActivityAcceptance() {
 	ai := s.AddActivityInfo()
 
 	prevStamp := ai.Stamp
-	err := PauseActivity(s.mutableState, ai.ActivityId)
+	err := PauseActivity(s.mutableState, ai.ActivityId, nil)
```
gow, line 338: We should also update one of these tests to include pause info

gow (approved):
Let's also add (or modify existing) unit tests to inlude pause info.

service/history/workflow/activity_test.go
ychebotarev: ok will do that.

PR #7619, opened 2025-04-16
Add Workflow static check to CI

carlydf:
## What changed?
- Add `make workflowcheck` and run it in CI checks
- Fix issues in existing system workflows. These fixes should not cause replay errors, because the ordering of map keys did not change which Temporal history events were generated, nor does switching from `time.Now()` to `workflow.Now()`
## How did you test it?
- Scheduler has replay unit tests that should catch any issues introduced by these changes
- Worker Deployment also has replay unit tests
- Deployment isn't used by any customers, so the lack of replay tests doesn't matter
- Replication workflow does not have replay tests, but the only change there was `time.Now()->workflow.Now()`

## Potential risks
Force replication workflow gets an NDE because `time.Now()` was replaced with `workflow.Now()`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -389,7 +389,7 @@ shell-check:
 
 workflowcheck: $(WORKFLOWCHECK)
 	@printf $(COLOR) "Run workflowcheck for system workflows..."
-	for dir in ./service/worker/*/ ; do \
+	for dir in $(SYSTEM_WORKFLOWS_ROOT)/*/ ; do \
 		echo "Running workflowcheck on $$dir" ; \
 		$(WORKFLOWCHECK) "$$dir" ; \
 	done
--- a/Makefile
+++ b/Makefile
@@ -205,9 +205,10 @@ ACTIONLINT := $(LOCALBIN)/actionlint-$(ACTIONLINT_VER)
 $(ACTIONLINT): | $(LOCALBIN)
 	$(call go-install-tool,$(ACTIONLINT),github.com/rhysd/actionlint/cmd/actionlint,$(ACTIONLINT_VER))
 
-WORKFLOWCHECK := $(LOCALBIN)/workflowcheck
+WORKFLOWCHECK_VER := v0.3.0
+WORKFLOWCHECK := $(LOCALBIN)/workflowcheck-$(WORKFLOWCHECK_VER)
 $(WORKFLOWCHECK): | $(LOCALBIN)
-	$(call go-install-tool,$(WORKFLOWCHECK),go.temporal.io/sdk/contrib/tools/workflowcheck,latest)
+	$(call go-install-tool,$(WORKFLOWCHECK),go.temporal.io/sdk/contrib/tools/workflowcheck,$(WORKFLOWCHECK_VER))
 
 # The following tools need to have a consistent name, so we use a versioned stamp file to ensure the version we want is installed
 # while installing to an unversioned binary name.
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,6 @@ ACTIONLINT := $(LOCALBIN)/actionlint-$(ACTIONLINT_VER)
 $(ACTIONLINT): | $(LOCALBIN)
 	$(call go-install-tool,$(ACTIONLINT),github.com/rhysd/actionlint/cmd/actionlint,$(ACTIONLINT_VER))
 
-WORKFLOWCHECK_VER := latest
 WORKFLOWCHECK := $(LOCALBIN)/workflowcheck
 $(WORKFLOWCHECK): | $(LOCALBIN)
 	$(call go-install-tool,$(WORKFLOWCHECK),go.temporal.io/sdk/contrib/tools/workflowcheck,latest)
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,7 @@ INTEGRATION_TEST_DIRS := $(DB_INTEGRATION_TEST_ROOT) $(DB_TOOL_INTEGRATION_TEST_
 ifeq ($(UNIT_TEST_DIRS),)
 UNIT_TEST_DIRS := $(filter-out $(FUNCTIONAL_TEST_ROOT)% $(FUNCTIONAL_TEST_XDC_ROOT)% $(FUNCTIONAL_TEST_NDC_ROOT)% $(DB_INTEGRATION_TEST_ROOT)% $(DB_TOOL_INTEGRATION_TEST_ROOT)% ./temporaltest%,$(TEST_DIRS))
 endif
+SYSTEM_WORKFLOWS_ROOT := ./service/worker
 
 # Pinning modernc.org/sqlite to this version until https://gitlab.com/cznic/sqlite/-/issues/196 is resolved.
 PINNED_DEPENDENCIES := \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -204,6 +205,10 @@ ACTIONLINT := $(LOCALBIN)/actionlint-$(ACTIONLINT_VER)
 $(ACTIONLINT): | $(LOCALBIN)
 	$(call go-install-tool,$(ACTIONLINT),github.com/rhysd/actionlint/cmd/actionlint,$(ACTIONLINT_VER))
 
+WORKFLOWCHECK := $(LOCALBIN)/workflowcheck
+$(WORKFLOWCHECK): | $(LOCALBIN)
+	$(call go-install-tool,$(WORKFLOWCHECK),go.temporal.io/sdk/contrib/tools/workflowcheck,latest)
```
stephanos, line 210: Can/should we give it a specific version instead? Upgrading dependencies is a pain, but unpredictable behavior changes more so IMO.

stephanos (approved):

Makefile
carlydf: I think you're right! I changed that so it's using v0.3.0

service/worker/scheduler/workflow.go
```diff
@@ -1144,6 +1146,12 @@ func (s *scheduler) updateMemoAndSearchAttributes() {
 	}
 }
 
+func (s *scheduler) deterministicProtoEqual(a, b *schedulepb.ScheduleListInfo) bool {
+	ab, _ := proto.Marshal(a)
+	bb, _ := proto.Marshal(b)
+	return string(ab) == string(bb)
```
dnr, line 1152: This is wrong:
proto.Marshal is not guaranteed to be deterministic unless you use MarshalOptions{Deterministic:true}. But if you're doing that there's no reason not to use proto.Equal, see https://github.com/protocolbuffers/protobuf-go/blob/master/proto/equal.go#L40-L41


(Also, don't convert to string to compare (which does a copy in most cases), use bytes.Equal)

service/worker/scheduler/workflow.go
```diff
@@ -1069,7 +1069,8 @@ func (s *scheduler) updateCustomSearchAttributes(searchAttributes *commonpb.Sear
 	}
 
 	upsertMap := map[string]any{}
-	for key, valuePayload := range searchAttributes.GetIndexedFields() {
+	for _, key := range workflow.DeterministicKeys(searchAttributes.GetIndexedFields()) {
```
dnr, line 1072: The result is going in a map so there's no reason to use DeterministicKeys here

service/worker/scheduler/workflow.go
```diff
@@ -1080,7 +1081,8 @@ func (s *scheduler) updateCustomSearchAttributes(searchAttributes *commonpb.Sear
 
 	//nolint:staticcheck // SA1019 Use untyped version for backwards compatibility.
 	currentSearchAttributes := workflow.GetInfo(s.ctx).SearchAttributes
-	for key, currentValuePayload := range currentSearchAttributes.GetIndexedFields() {
+	for _, key := range workflow.DeterministicKeys(currentSearchAttributes.GetIndexedFields()) {
```
dnr, line 1084: Same thing, the results are going in a map anyway so there's no reason to use DeterministicKeys

service/worker/scheduler/workflow.go
```diff
@@ -1117,7 +1119,7 @@ func (s *scheduler) updateMemoAndSearchAttributes() {
 	if currentInfoPayload == nil ||
 		payload.Decode(currentInfoPayload, &currentInfoBytes) != nil ||
 		currentInfo.Unmarshal(currentInfoBytes) != nil ||
-		!proto.Equal(&currentInfo, newInfo) {
+		!s.deterministicProtoEqual(&currentInfo, newInfo) { // TODO(carlydf): proto.Equal is non-deterministic
```
dnr, line 1122: Changes to this condition can cause nondeterminism errors so if we do need to do this (and I don't think we do) then it should be versioned.

As far as I know proto.Equal is totally fine and correct here. If you disagree could you point to some evidence otherwise?

(Yes there are replay tests but they can't catch all edge cases)

dnr (commented):
I think

service/worker/scheduler/workflow.go
carlydf: proto.Equal calls underlying libraries that trigger the workflow static check to error

service/worker/scheduler/workflow.go
carlydf: The reason to use DeterministicKeys here (and in the other place) is to make the static checker happy. I agree that whether or not the map iteration is done with `range` or `DeterministicKeys` will not affect the overall workflow determinism. That is why I feel this change does not require a patch.

IMO, the reason to use DeterministicKeys is because it's "workflow best practice" to avoid _all_ range over maps, so that workflow authors don't have to think "does the ordering of this range affect my History Events" and then decide. It's better to just always require using DeterministicKeys and then never accidentally use `range` in a place that actually would affect determinism.

service/worker/scheduler/workflow.go
carlydf: Maybe the real fix here is to update the workflow static checker so that it accepts `proto.Equal`

service/worker/scheduler/workflow.go
dnr: workflowcheck (and other linters) are aids for finding bugs, not authorities. If they suggest something semantically equivalent but that makes the code harder to read (and also slower), we don't have to take the suggestion.

I get the point that it reduces mental overhead for someone reading the code, but we could also just disable workflowcheck here with a comment saying "iteration is safe because the results are going into a map".

service/worker/scheduler/workflow.go
carlydf: It looks like we can whitelist certain functions, so we could whitelist proto.Equal. https://github.com/temporalio/sdk-go/tree/master/contrib/tools/workflowcheck

service/worker/scheduler/workflow.go
dnr: workflowcheck is not infallible, it already has a big ignore list and it's almost certainly missing more stuff that's actually deterministic. It should definitely be updated, but in the meantime let's leave this code correct.

service/worker/scheduler/workflow.go
carlydf: I think the reason proto.Equal was flagged is because it called `(reflect.Value).Interface` or something similar. 

I could start by whitelisting proto.Equal in our internal use of it, and also open an issue in go-sdk to ask them to fix the false-positive proto.Equal and false-negative proto.Marshal. And I'll revert the proto.Equal change to this workflow.

```
In some cases, functions that are considered non-deterministic are commonly used in ways that only follow a deterministic code path. For example if a common library function iterates over a map in a rare case that does not apply to the situation, it will be flagged as non-deterministic. A few common cases of this have been force-set as deterministic for common use:

(reflect.Value).Interface - Was considered non-deterministic because deep down in Go internal source, this uses sync.Map (that does map iteration) as a cache of method layouts
runtime.Caller - Was considered non-deterministic because deep down in Go internal source, some runtime source starts a goroutine on lazy GC start when building CGo frames
go.temporal.io/sdk/internal.propagateCancel - Was considered non-deterministic because it starts a goroutine
(*go.temporal.io/sdk/internal.cancelCtx).cancel - Was considered non-deterministic because it iterates over a map
```

dnr: I was only suggesting reverting the changes to the schedule wf (at least the proto.Equal) and adding ignore comments, not the whole PR

PR #7621, opened 2025-04-16
CHASM: implement AddTask

yycptt:
## What changed?

- Implement AddTask functionality for CHASM
- This PR is on top of https://github.com/temporalio/temporal/pull/7606

## How did you test it?

- Unit tests

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -28,6 +28,7 @@ import (
 	"cmp"
 	"fmt"
 	"reflect"
+	"slices"
 	"time"
 
 	commonpb "go.temporal.io/api/common/v1"
@@ -42,6 +43,7 @@ import (
 	"go.temporal.io/server/common/softassert"
 	"go.temporal.io/server/service/history/tasks"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 var (
@@ -104,8 +106,18 @@ type (
 		backend     NodeBackend
 		pathEncoder NodePathEncoder
 		logger      log.Logger
+
+		// Following fields are per transaction states, will get cleaned up
+		// during CloseTransaction().
+
 		// Mutations accumulated so far in this transaction.
 		mutation NodesMutation
+		newTasks map[any][]taskWithAttributes // component value -> task & attributes
+	}
+
+	taskWithAttributes struct {
+		task       any
+		attributes TaskAttributes
 	}
 
 	// NodesMutation is a set of mutations for all nodes rooted at a given node n,
@@ -169,6 +181,7 @@ func NewTree(
 			UpdatedNodes: make(map[string]*persistencespb.ChasmNode),
 			DeletedNodes: make(map[string]struct{}),
 		},
+		newTasks: make(map[any][]taskWithAttributes),
 	}
 
 	root := newNode(base, nil, "")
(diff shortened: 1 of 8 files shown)


chasm/tree_test.go
```diff
@@ -905,6 +908,127 @@ func (s *nodeSuite) TestComponent() {
...
+			name:         "EmptyTask",
+			task:         TestOutboundSideEffectTask{},
+			expectedData: nil,
+		},
+		{
+			name: "StructWithProtoField",
+			task: &TestPureTask{
+				Payload: payload,
+			},
+			expectedData: expectedBlob.GetData(),
```
lina-temporal, line 940: Interesting, this doesn't have to be wrapped in a struct itself? I'm guessing protobuf encodes a struct with a single blob the same way as it would directly encode a blob?

chasm/tree.go
```diff
@@ -1152,3 +1255,58 @@ func taskCategory(
...
+	var blob *commonpb.DataBlob
+	protoMessageFound := false
+	for i := 0; i < taskGoType.NumField(); i++ {
+		fieldV := taskValue.Field(i)
+		if !fieldV.Type().AssignableTo(protoMessageT) {
+			continue
+		}
+
+		if protoMessageFound {
+			return nil, serviceerror.NewInternal("only one proto field allowed in task struct")
```
lina-temporal, line 1296: (minor) since we already have the `taskGoType`, could we also log that in these errors?

lina-temporal (approved):
few small suggestions, LGTM!

chasm/task.go
```diff
@@ -37,10 +37,17 @@ type (
 		Destination   string
 	}
 
-	TaskHandler[C any, T any] interface {
-		Validate(Context, C, T) error
+	SideEffectTaskExecutor[C any, T any] interface {
 		Execute(context.Context, ComponentRef, T) error
```
lina-temporal, line 41: Since we aren't passing a `Context` to plumb through `Now()`, If a side effect task wanted to know the wall clock time, what should it do?

HSM provided `Environment` on both executors:

```
type ImmediateExecutor[T Task] func(ctx context.Context, env Environment, ref Ref, task T) error
type TimerExecutor[T Task] func(env Environment, node *Node, task T) error
```

yycptt: If side effect task executor logic want to know the now() time for a component, it needs to invoke a chasm engine Update/Read method. Then chasm.Context is available in the updateFn or readFn they pass to the Update/Read method. Then from the chasm.Context, the Now() method is available.
(The chasm engine implementation itself needs to be available from the context.Context. https://github.com/yycptt/temporal/commit/163c3ee5203b97144635b87a101404b01725420b#diff-3687673cd41da46293431f19d81ab231ae8e23227baf29a4c9d429c81a064db8R157 )

If the executor just want to know wall clock in general, its implementation (struct) needs to inject a time source itself before registering the task executor to chasm,  like any other dependencies the execution logic may rely on.

chasm/tree_test.go
yycptt: no...  the encoding logic use reflection to find the proto message field within the struct and only encodes that field.

chasm/task.go
```diff
@@ -37,10 +37,17 @@ type (
...
 	}
+
+	PureTaskExecutor[C any, T any] interface {
+		Execute(Context, C, T) error
+	}
+
+	TaskValidator[C any, T any] interface {
+		Validate(Context, C, T) error
```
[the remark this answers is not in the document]
yycptt: I will address this in https://github.com/temporalio/temporal/pull/7636

PR #7622, opened 2025-04-16
Run Cass tests on larger machines

stephanos:
## What changed?

Bumped resources for func tests against Cass.

## Why?

We've seen func tests on Cass time out. This is an attempt to see if bumping the resources fixes the issue.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -429,11 +429,15 @@ jobs:
             persistence_driver: cassandra
             containers: [cassandra, elasticsearch]
             es_version: v7
+            # Cassandra tests need a larger instance
+            runs_on_override: ubuntu-latest-8-cores
           - name: cass_es8
             persistence_type: nosql
             persistence_driver: cassandra
             containers: [cassandra, elasticsearch8]
             es_version: v8
+            # Cassandra tests need a larger instance
+            runs_on_override: ubuntu-latest-8-cores
           - name: sqlite
             persistence_type: sql
             persistence_driver: sqlite
@@ -450,7 +454,7 @@ jobs:
             persistence_type: sql
             persistence_driver: postgres12_pgx
             containers: [postgresql]
-    runs-on: ${{ matrix.runs-on }}
+    runs-on: ${{ matrix.runs_on_override || matrix.runs-on }}
     env:
       TEST_TOTAL_SHARDS: ${{ needs.set-up-single-test.outputs.total_shards }}
       TEST_SHARD_INDEX: ${{ matrix.shard_index }}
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -579,27 +585,32 @@ jobs:
             persistence_driver: elasticsearch
             parallel_flags: ""
             containers: [cassandra, elasticsearch]
+            runs_on: ubuntu-latest-16-cores
```
stephanos, line 588: Bumped up.

.github/workflows/run-tests.yml
```diff
@@ -579,27 +585,32 @@ jobs:
...
           - name: cass_es8
             persistence_type: nosql
             persistence_driver: elasticsearch
             parallel_flags: ""
             containers: [cassandra, elasticsearch8]
+            runs_on: ubuntu-latest-16-cores
```
stephanos, line 594: Bumped up.

.github/workflows/run-tests.yml
```diff
@@ -428,27 +427,33 @@ jobs:
             persistence_driver: cassandra
             containers: [cassandra, elasticsearch]
             es_version: v7
+            runs-on: ubuntu-latest-16-cores
```
stephanos, line 430: Bumped.

.github/workflows/run-tests.yml
```diff
@@ -428,27 +427,33 @@ jobs:
...
           - name: cass_es8
             persistence_type: nosql
             persistence_driver: cassandra
             containers: [cassandra, elasticsearch8]
             es_version: v8
+            runs-on: ubuntu-latest-16-cores
```
stephanos, line 436: Bumped.

.github/workflows/run-tests.yml
```diff
@@ -722,24 +732,29 @@ jobs:
...
             persistence_driver: postgres12
             containers: [postgresql]
+            runs-on: ubuntu-latest
           - name: postgres12_pgx
             persistence_type: sql
             persistence_driver: postgres12_pgx
             containers: [postgresql]
-    runs-on: ubuntu-latest
+            runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runs-on }}
```
stephanos, line 757: ^ Made all explicit to avoid ambiguity.

stephanos:
It took several minutes for the job to obtain an instance of `ubuntu-latest-16-cores`. I don't think this will improve things.

.github/workflows/run-tests.yml
```diff
@@ -413,7 +413,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: ${{ fromJson(needs.set-up-single-test.outputs.runs_on) }}
```
bergundy, line 416: Are we intentionally removing this capability? If so, we should probably remove that whole output and code path.

stephanos: Ah, that's why I didn't put it up for review 2w ago. Okay, I'll need to figure out how to keep the option and provide a default per matrix item then.

ychebotarev (approved):
since that was the only comment and it is resolved - LGFM

PR #7629, opened 2025-04-17
don't emit error logs all the time in deployment version sync summary!

Shivs11:
## What changed?

- WISOTT

## Why?

- We were emitting error logs even on successful calls to `syncSummary` - I don't think we want that!


ShahabT (approved):
A more specific title such as "don't emit error logs all the time in deployment version sync summary!"

prathyushpv (approved):

PR #7638, opened 2025-04-21
check status of docker image build job after triggering it

carlydf:
## What changed?
Check status of docker image build job after triggering it

## Why?
So that we will notice if the OSS image is not being built and pushed successfully

For context, the reason to do this is that last week I noticed that the Long Haul was using a commit from 10 days prior, instead of the most recent commit on main which it should be using.

The reason it was using the commit from 10 days prior is that the Long Haul uses the commit that is in the most recently pushed OSS image. What had happened is that for 10 days, the job that pushes our OSS image to docker hub had been failing, but no one had noticed.

The reason no one noticed is because the job that calls the failing job considered itself to be "successful" as soon as it started the build-push job, vs waiting for the build-push job to complete successfully.
This PR changes that.

## How did you test it?
- Tested run where I forced the [build-push job](https://github.com/temporalio/docker-builds/actions/runs/14581917807) to fail and then the [parent](https://github.com/temporalio/temporal/actions/runs/14581915571) failed
- Fixed the forced bug and saw the [build-push job](https://github.com/temporalio/docker-builds/actions/runs/14581943197) succeed and the [parent job](https://github.com/temporalio/temporal/actions/runs/14581941612) also succeed


.github/workflows/trigger-publish.yml
```diff
@@ -39,4 +39,30 @@ jobs:
           REPO: ${{ toJSON('temporal') }}
           BRANCH: ${{ toJSON(steps.get_branch.outputs.branch) }}
         run: |
+          # Dispatch the workflow
           curl -fL -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "https://api.github.com/repos/$PARENT_REPO/actions/workflows/$WORKFLOW_ID/dispatches" -d '{"ref":'"$PARENT_BRANCH"', "inputs": { "repo":'"$REPO"', "branch":'"$BRANCH"' }}'
+          
+          # Wait a few seconds for the workflow to start
+          sleep 10
```
dnr, line 46: This seems error-prone. And the "get latest run" is racy. Could we do something like:
- get the latest run
- dispatch the workflow
- in a loop:
  - get the last bunch of runs, filter out anything <= the original latest run
  - if we find one, break
  - if we find more than one... can we examine its properties to see if it's the same one we triggered?
- then poll

.github/workflows/trigger-publish.yml
```diff
@@ -39,4 +39,30 @@ jobs:
           REPO: ${{ toJSON('temporal') }}
           BRANCH: ${{ toJSON(steps.get_branch.outputs.branch) }}
         run: |
+          # Dispatch the workflow
```
dnr, line 42: let's factor out:
```
call() {
  curl -fL -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "$@"
}
```
then use it like
```
call "https://api.github.com/repos/$PARENT_REPO/actions/workflows/$WORKFLOW_ID/dispatches" -d '{"ref":'"$PARENT_BRANCH"', "inputs": { "repo":'"$REPO"', "branch":'"$BRANCH"' }}'
```

.github/workflows/trigger-publish.yml
```diff
@@ -39,4 +39,30 @@ jobs:
...
           curl -fL -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "https://api.github.com/repos/$PARENT_REPO/actions/workflows/$WORKFLOW_ID/dispatches" -d '{"ref":'"$PARENT_BRANCH"', "inputs": { "repo":'"$REPO"', "branch":'"$BRANCH"' }}'
+          
+          # Wait a few seconds for the workflow to start
+          sleep 10
+          
+          # Get the latest workflow run for this workflow
+          run_id=$(curl -fL -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "https://api.github.com/repos/$PARENT_REPO/actions/workflows/$WORKFLOW_ID/runs" | jq -r '.workflow_runs[0].id')
```
dnr, line 49: I think we need either `set -euo pipefail` on top so this fails if the call fails, or check whether `run_id` is empty

.github/workflows/trigger-publish.yml
```diff
@@ -39,4 +39,30 @@ jobs:
...
+          
+          # Poll the workflow run status until it completes
+          while true; do
+            status=$(curl -fL -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "https://api.github.com/repos/$PARENT_REPO/actions/runs/$run_id" | jq -r '.status')
+            conclusion=$(curl -fL -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $PAT" "https://api.github.com/repos/$PARENT_REPO/actions/runs/$run_id" | jq -r '.conclusion')
```
dnr, line 54: get both of these with one call, like
```suggestion
  result=$(call "https://api.github.com/repos/$PARENT_REPO/actions/runs/$run_id")
  status=$(echo "$result" | jq -r .status)
  conclusion=$(echo "$result" | jq -r .conclusion)
```

.github/workflows/trigger-publish.yml
```diff
@@ -39,4 +39,30 @@ jobs:
...
+            
+            if [ "$status" = "completed" ]; then
+              if [ "$conclusion" = "success" ]; then
```
dnr, line 57: if you use [[ ]] in bash you don't need quotes (though you can add them back if shellcheck complains)
```suggestion
            if [[ $status = completed ]]; then
              if [[ $conclusion = success ]]; then
```

carlydf:
close since these jobs are handled by release team now

PR #7643, opened 2025-04-22
Use CallerInfo to get namespace name in interceptors

prathyushpv:
## What changed?
Move CallerInfoInterceptor above all other interceptors that get namespace name from the request. Change other interceptors to use namespace name from caller info.

## Why?
Each interceptor does not have to get namespace name from request.

## How did you test it?
Existing tests.


service/frontend/fx.go
```diff
@@ -258,6 +258,8 @@ func GrpcServerOptionsProvider(
 		logger.Fatal("creating gRPC server options failed", tag.Error(err))
 	}
 	unaryInterceptors := []grpc.UnaryServerInterceptor{
+		// CallerInfoInterceptor needs to be before any interceptors that get namespace from requests
+		callerInfoInterceptor.Intercept,
```
yycptt, line 262: hmm this probably needs to get called after the NamespaceValidateIntercept, which populates Namespace field from the task token field if it's not set I think. 

Can you help double check the logic in NamespaceValidateIntercepter?

prathyushpv: That is right, thanks!. I have moved NamespaceValidateIntercept above callerInfoInterceptor.

service/frontend/fx.go
```diff
@@ -258,13 +258,17 @@ func GrpcServerOptionsProvider(
...
 	}
 	unaryInterceptors := []grpc.UnaryServerInterceptor{
+		// CallerInfoInterceptor and namespaceValidatorInterceptor needs to be before any interceptors that get namespace from requests.
+		// namespaceValidatorInterceptor will extract namespace name from task token if it is not present in request.
+		// callerInfoInterceptor will populate this value in CallerInfo in request context.
+		namespaceValidatorInterceptor.NamespaceValidateIntercept,
+		callerInfoInterceptor.Intercept,
 		// Order or interceptors is important
 		// Mask error interceptor should be the most outer interceptor since it handle the errors format
 		// Service Error Interceptor should be the next most outer interceptor on error handling
```
dnr, line 268: These comments are ignored + wrong now. Putting anything before ServiceErrorInterceptor and maskInternalErrorDetailsInterceptor seems wrong since errors from these new ones won't get wrapped correctly.

I think the namespace ones should be after all the error handling ones, probably including rpc.NewFrontendServiceErrorInterceptor. I.e. put callerInfoInterceptor right after namespaceValidatorInterceptor and don't change anything else.

I see that maskInternalErrorDetailsInterceptor does use namespace name, but we can just leave that one as-is and not use the context there.

prathyushpv: yeah that is right. 

I am also thinking that extracting namespace from request struct is better than reading it from the context. to read it from context, it has to construct and delete a struct object. In that case, this change probably does not add any value.

PR #7646, opened 2025-04-23
CHASM: Field.Get implementation

alexshtin:
## What changed?

CHASM: `Field.Get` implementation.

## Why?

Important part of CHASM project.

## How did you test it?

Added unit tests but will add more after `CloseTransaction` is implemented to test it end to wnd.

## Potential risks

No risks.

## Documentation

No.

## Is hotfix candidate?

No.


chasm/field.go
```diff
@@ -58,10 +60,37 @@ func (fi fieldInternal) IsEmpty() bool {
...
+		panic("not implemented")
+	default:
+		err = serviceerror.NewInternal(fmt.Sprintf("unsupported field type: %v", f.Internal.fieldType))
+	}
+
+	if err != nil {
+		var nilT T
+		return nilT, err
+	}
+	f.Internal.value = f.Internal.node.value
```
yycptt, line 92: hmm I feel it's dangerous/error prone to keep two copies of the value. Is the assumption here that the value always have pointer type and two values fields will never go out of sync?

I was originally thinking the Internal.value & field type is only used for newly created field which doesn't have a backing node yet.

alexshtin: Ok, I wasn't sure about it either. I introduced "getter" methods which look in `node` first and then into the `value` itself. It also simplified few other places.

chasm/field.go
```diff
@@ -47,21 +49,64 @@ type Field[T any] struct {
...
+	if fi.node == nil {
+		return fi.ft
+	}
+	return fi.node.fieldType()
+}
+
+func (f Field[T]) Get(chasmContext Context) (T, error) {
+	// If value is not nil, use it.
+	if f.Internal.v != nil {
+		return f.Internal.v.(T), nil
```
yycptt, line 82: just to confirm. For newly created Fields, after closing the transaction, will the `f.Internal.v` field will be reset to nil? so that later Get() will be served by the node?

chasm/field.go
```diff
@@ -47,21 +49,64 @@ type Field[T any] struct {
...
+		panic("not implemented")
+	default:
+		err = serviceerror.NewInternal(fmt.Sprintf("unsupported field type: %v", f.Internal.fieldType()))
+	}
+
+	if err != nil {
+		var nilT T
+		return nilT, err
+	}
+	// TODO: check if f.Internal.node.value is T or implements T?
```
yycptt, line 108: yeah it would be nice to have the check and return a proper error message and not panic.

yycptt (approved):

chasm/field.go
alexshtin: Good point. Yes, I will clear it there, but I am also changing order here. `node` is now taking precedence and will be used if not `nil`.

chasm/field.go
alexshtin: linter won`t pass this anyway.

PR #7648, opened 2025-04-23
Put NexusOperationCancelRequest[Completed|Failed] behind a feature flag

pdoerner:
## What changed?
Putting NexusOperationCancelRequest[Completed|Failed] behind a feature flag.

## Why?
So that we can wait to enable it until it will not cause issues with migration or rollback.

### DIFF
--- a/components/nexusoperations/config.go
+++ b/components/nexusoperations/config.go
@@ -155,21 +155,29 @@ Adding high-cardinality tags (like unique operation names) can significantly inc
 requirements and query complexity. Consider the cardinality impact when enabling these tags.`,
 )
 
+var RecordCancelRequestCompletionEvents = dynamicconfig.NewGlobalBoolSetting(
+	"component.nexusoperations.recordCancelRequestCompletionEvents",
+	false,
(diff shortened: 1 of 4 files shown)


components/nexusoperations/config.go
```diff
@@ -155,21 +155,29 @@ Adding high-cardinality tags (like unique operation names) can significantly inc
 requirements and query complexity. Consider the cardinality impact when enabling these tags.`,
 )
 
+var RecordCancelRequestCompletionEvents = dynamicconfig.NewGlobalBoolSetting(
+	"component.nexusoperations.recordCancelRequestCompletionEvents",
+	false,
+	`Boolean flag to control whether to record NexusOperationCancelRequestComplete and
```
bergundy, line 161:
```suggestion
	`Boolean flag to control whether to record NexusOperationCancelRequestCompleted and 
```

bergundy (approved):
This change is fine but I would change the state machine transitions to be applied as part of the event application so that the state machine states don't diverge in event based replication.

Issue #4094, opened 2023-03-23
[Bug] Version info upgrade notification does not get cleared after cluster version upgrade


bergundy:
Sample output from `tctl admin cluster describe`:

```
"versionInfo": {
  "current": {
   "version": "1.20.0",
   "releaseTime": "2023-02-17T21:48:00Z"
  },
  "recommended": {
   "version": "1.20.0",
   "releaseTime": "2023-02-17T21:48:00Z"
  },
  "alerts": [
   {
    "message": ":ringed_planet: A new release is available!",
    "severity": "Low"
   }
  ],
  "lastUpdateTime": "2023-03-22T08:54:54.838392383Z"
 }
```

Note that the `lastUpdateTime` is after the cluster was upgraded.

smcgivern:
We're seeing something similar:

```json
{
  "supportedClients": {
    "temporal-cli": "\u003c2.0.0",
    "temporal-go": "\u003c2.0.0",
    "temporal-java": "\u003c2.0.0",
    "temporal-php": "\u003c2.0.0",
    "temporal-server": "\u003c2.0.0",
    "temporal-typescript": "\u003c2.0.0",
    "temporal-ui": "\u003c3.0.0"
  },
  "serverVersion": "1.22.0",
  // snip
  "versionInfo": {
    "current": {
      "version": "1.20.3",
      "releaseTime": "2023-05-16T04:25:00Z"
    },
    "recommended": {
      "version": "1.21.2",
      "releaseTime": "2023-07-15T02:00:00Z"
    },
    "alerts": [
      {
        "message": "🪐 A new release is available!",
        "severity": "Low"
      }
    ],
    "lastUpdateTime": "2023-09-08T14:24:09.898917704Z"
  },
  "failoverVersionIncrement": "10",
  "initialFailoverVersion": "1"
}
```

Per https://community.temporal.io/t/cluster-upgrade-from-1-14-4-1-9-0-serverversion-changes-but-current-version-stays-1-14-4/6691/2, we should consider `serverVersion` to be accurate, but the cluster's `versionInfo` hasn't updated, which is confusing.

g4ze:
hey! is this issue still up? if yes, could you pls elaborate a bit more on this issue?

vaibhavyadav-dev:
@yycptt @bergundy I'm Taking this one.
Thanks!

AadiDev005:
[...]

PR #7650, opened 2025-04-23
fixed alert on version match #4094

vaibhavyadav-dev:
## What changed?
Added a condition in `toVersionInfo` (in `version_checker.go`) that removes the version alert when the current version matches the recommended version. This prevents unnecessary alerts like “A new release is available!” from showing up after the server has already been upgraded.

## Why?
To fix [Issue #4094](https://github.com/temporalio/temporal/issues/4094). The version alert was being shown even when both the current and recommended versions were the same, which can be misleading for users who already upgraded their cluster.

## How did you test it?
Ran ```make all``` and ```make test```.

## Potential risks
Low risk — only affects display logic for version alerts. Doesn’t interfere with critical cluster functionality.

## Documentation
No changes to documentation needed. This only affects backend logic for version info output.

## Is hotfix candidate?
No

Fix: issue #4094

### DIFF
--- a/service/frontend/version_checker.go
+++ b/service/frontend/version_checker.go
@@ -25,6 +25,7 @@
 package frontend
 
 import (
+	"strings"
 	"context"
 	"runtime"
 	"sync"
@@ -189,6 +190,11 @@ func (vc *VersionChecker) saveVersionInfo(ctx context.Context, resp *check.Versi
 
 func toVersionInfo(resp *check.VersionCheckResponse) (*versionpb.VersionInfo, error) {
 	for _, product := range resp.Products {
+
+		if strings.TrimSpace(product.Current.Version) == strings.TrimSpace(product.Recommended.Version) {
+			product.Alerts = nil
+		}
+
 		if product.Product == headers.ClientNameServer {
 			return &versionpb.VersionInfo{
 				Current:        convertReleaseInfo(product.Current),


service/frontend/version_checker.go
```diff
@@ -189,6 +190,11 @@ func (vc *VersionChecker) saveVersionInfo(ctx context.Context, resp *check.Versi
 
 func toVersionInfo(resp *check.VersionCheckResponse) (*versionpb.VersionInfo, error) {
 	for _, product := range resp.Products {
+
+		if strings.TrimSpace(product.Current.Version) == strings.TrimSpace(product.Recommended.Version) {
```
bergundy, line 194: I'm wondering about a few things:

1. Why `TrimSpace` is needed.
2. If we can omit the alerts on the server side when we detect that the reported version matches.
3. Do we need to keep `Instructions` if the versions match?

vaibhavyadav-dev: Thanks for pointing it out @bergundy, here what I think about it:

- This ensures we're comparing the actual version values accurately, avoiding mismatches due to formatting, which I know you guys never let this happen, it looks redundant and unnecessary, happy to remove it.

- I agree that there's no need to trigger an alert, when we can optimize the logic on the server side to prevent unnecessary noise; in this case I've tried to do it client side, but optimizing server-side makes sense too.

- we can also consider skipping the instructions in this case, I can include that in the next update if the team aligns on it.

Thanks, I’m excited to keep exploring this further!

bergundy: Let me take this to the team and circle back with you.

PR #7652, opened 2025-04-23
Force UUID on RegisterNamespace()

davewhat:
## What changed?
We allow forcing the UUID of a RegisterNamespace() for the purpose of Emergency Migration.

## Why?
To facility Emergency Migration, an "Imposter" of the original namespace is needed on a different Cell

## How did you test it?
This is a private branch only.


davewhat:
sorry, I wasn't aware this would open on the main repo.

PR #7654, opened 2025-04-23
Add application logic for NexusCancelRequest[Completed|Failed] events

pdoerner:
## What changed?
Added application logic for NexusCancelRequest[Completed|Failed] events

## Why?
To support event-based replication

## How did you test it?
Existing tests

### DIFF
--- a/components/nexusoperations/events.go
+++ b/components/nexusoperations/events.go
@@ -90,7 +90,22 @@ func (d CancelRequestCompletedEventDefinition) Type() enumspb.EventType {
 }
 
 func (d CancelRequestCompletedEventDefinition) Apply(root *hsm.Node, event *historypb.HistoryEvent) error {
-	return nil
+	_, err := transitionOperation(root, event, func(node *hsm.Node, o Operation) (hsm.TransitionOutput, error) {
+		child, err := o.CancelationNode(node)
+		if err != nil {
+			return hsm.TransitionOutput{}, err
+		}
+		if child != nil {
+			return hsm.TransitionOutput{}, hsm.MachineTransition(child, func(c Cancelation) (hsm.TransitionOutput, error) {
+				return TransitionCancelationSucceeded.Apply(c, EventCancelationSucceeded{
+					Time: event.EventTime.AsTime(),
+					Node: child,
+				})
+			})
+		}
+		return hsm.TransitionOutput{}, nil
+	})
+	return err
 }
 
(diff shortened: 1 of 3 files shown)


components/nexusoperations/executors.go
```diff
@@ -668,21 +668,21 @@ func (e taskExecutor) saveCancelationResult(ctx context.Context, env hsm.Environ
...
+									ScheduledEventId: scheduledEventID,
+									RequestedEventId: c.RequestedEventId,
+									Failure:          failure,
+								},
+							}
+							// nolint:revive // We must mutate here even if the linter doesn't like it.
+							e.WorkerMayIgnore = true // For compatibility with older SDKs.
+						})
+					}
 					return TransitionCancelationFailed.Apply(c, EventCancelationFailed{
```
bergundy, line 686: You need an else statement here to avoid double apply.

bergundy: Also need an else statement on the Completed case.

pdoerner: I think we still need to apply the transition here. That's what we are doing for other events at least.

bergundy: I was wrong, adding an event doesn't trigger the `Apply` logic (it probably should but let's not worry about that now).

components/nexusoperations/events.go
```diff
@@ -90,7 +90,22 @@ func (d CancelRequestCompletedEventDefinition) Type() enumspb.EventType {
...
+		}
+		if child != nil {
+			return hsm.TransitionOutput{}, hsm.MachineTransition(child, func(c Cancelation) (hsm.TransitionOutput, error) {
+				return TransitionCancelationSucceeded.Apply(c, EventCancelationSucceeded{
+					Time: event.EventTime.AsTime(),
+					Node: child,
+				})
+			})
+		}
+		return hsm.TransitionOutput{}, nil
```
bergundy, line 106: Maybe invert the condition here to reduce the indentation?

```suggestion
		if child == nil {
			return hsm.TransitionOutput{}, nil
		}
		return hsm.TransitionOutput{}, hsm.MachineTransition(child, func(c Cancelation) (hsm.TransitionOutput, error) {
			return TransitionCancelationSucceeded.Apply(c, EventCancelationSucceeded{
				Time: event.EventTime.AsTime(),
				Node: child,
			})
		})
		}
```

bergundy: Also document that we're okay ignoring if the cancelation state machine is not found during application of this event.

bergundy (approved):

PR #7655, opened 2025-04-23
Handle Transition History update when rebuild workflow

xwduan:
## What changed?

Handle Transition History update when rebuild workflow
## Why?

When workflow is rebuilt from DB, we assign it a new transition, so the replication can continue.


service/history/workflow_rebuilder.go
```diff
@@ -122,9 +127,27 @@ func (r *workflowRebuilderImpl) rebuild(
...
+) {
+	newMutableState.GetExecutionInfo().PreviousTransitionHistory = rebuildSpec.previousTransitionHistory
+	newMutableState.GetExecutionInfo().LastTransitionHistoryBreakPoint = rebuildSpec.lastTransitionHistoryBreakPoint
+	if rebuildSpec.transitionHistory != nil {
+		transitionHistory := rebuildSpec.transitionHistory
+		breakPoint := transitionhistory.CopyVersionedTransition(transitionHistory[len(transitionHistory)-1])
+		// This update is a naive update. It failed to consider the case where the current cluster is not the cluster
+		// that did the last transition. But this should not break the replication protocol.
+		transitionHistory[len(transitionHistory)-1].TransitionCount += 1
+		newMutableState.GetExecutionInfo().TransitionHistory = transitionHistory
```
yycptt, line 146: do you mean `PreviousTransitionHistory` ?

xwduan: No. I mean the TransitionHistory. This is the case where transition history is enabled in current cluster and we are rebuilding.

yycptt: oh I see, so we assume this command is run in the active cluster.

yycptt (approved):

service/history/workflow_rebuilder.go
yycptt: may worth verifying the last version belongs to the current cluster. or create a new VersionedTransitionItem with current version.

yycptt: so I think the story is basically:
- If workflow is corrupted in active cluster. Use rebuild and then trigger replication with admin re-replicate command. (or maybe we can change rebuild to update the workflow as active)
- If workflow is corrupted on standby cluster. Delete the workflow from standby, and just use re-replicate from active side.

service/history/ndc/state_rebuilder.go
```diff
@@ -110,6 +124,115 @@ func (r *StateRebuilderImpl) Rebuild(
...
+	}
+	items := versionhistory.CopyVersionHistoryItems(currentVersionHistory.Items)
+	currentVersionHistory.Items = nil
+
+	// close rebuilt mutable state transaction clearing all generated tasks, etc.
+	_, _, err = rebuiltMutableState.CloseTransactionAsSnapshot(historyi.TransactionPolicyActive)
+	if err != nil {
+		return nil, 0, err
+	}
+	currentVersionHistory.Items = items
```
yycptt, line 202: nit: please help add some comment re. why we are doing this.

yycptt (approved):

PR #7656, opened 2025-04-24
Update-with-Start returns closed workflow's Outcome

stephanos:
**⚠️ Before this is merged, the `StartWorkflowExecutionResponse` API needs to have a new field `Running` added to it.**

## What changed?

When UwS is applied to a closes Workflow where the requested Update (identified by its Update ID) has already been completed, return the Update outcome.

Note that this only applies to the current execution; not any previous execution.

## Why?

Align UwS behavior with Update; a regular Update also returns its outcome from a closed Workflow.

## How did you test it?

New tests.

### DIFF
--- a/service/history/api/multioperation/api.go
+++ b/service/history/api/multioperation/api.go
@@ -128,40 +128,47 @@ func Invoke(
 }
 
 func (mo *multiOp) Invoke(ctx context.Context) (*historyservice.ExecuteMultiOperationResponse, error) {
-	// For workflow id conflict policy terminate-existing, always attempt a start
-	// since that works when the workflow is already running *and* when it's not running.
 	conflictPolicy := mo.startReq.StartRequest.WorkflowIdConflictPolicy
-	if conflictPolicy == enumspb.WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING {
-		resp, err := mo.startAndUpdateWorkflow(ctx)
-		var noStartErr *noStartError
-		switch {
-		case errors.As(err, &noStartErr):
-			// The start request was deduped, no termination is needed.
-			// Continue below by only sending the update.
-		case err != nil:
-			return nil, err
-		default:
-			return resp, nil
-		}
-	}
 
-	runningWorkflowLease, err := mo.getRunningWorkflowLease(ctx)
+	workflowLease, err := mo.getWorkflowLease(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	// Workflow was already started ...
-	if runningWorkflowLease != nil {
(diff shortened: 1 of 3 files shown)


service/history/api/multioperation/api.go
```diff
@@ -128,40 +128,47 @@ func Invoke(
...
+		updateId := mo.updateReq.Request.Request.Meta.GetUpdateId()
+		if outcome, err := workflowLease.GetMutableState().GetUpdateOutcome(ctx, updateId); err == nil {
+			workflowKey := workflowLease.GetContext().GetWorkflowKey()
+			workflowLease.GetReleaseFn()(nil)
+
+			return makeResponse(
+				&historyservice.StartWorkflowExecutionResponse{
+					RunId:   workflowKey.RunID,
+					Started: false, // set explicitly for emphasis
+					// TODO: Running: false,
```
stephanos, line 159: This requires an API change.

service/history/api/multioperation/api.go
```diff
@@ -244,21 +251,10 @@ func (mo *multiOp) getRunningWorkflowLease(ctx context.Context) (api.WorkflowLea
...
-	if !runningWorkflowLease.GetMutableState().IsWorkflowExecutionRunning() {
-		runningWorkflowLease.GetReleaseFn()(nil)
-		return nil, nil
-	}
-
 	return runningWorkflowLease, nil
 }
 
-func (mo *multiOp) allowUpdateWorkflow(
-	ctx context.Context,
```
stephanos, line 261: Unused.

tests/update_workflow_test.go
```diff
@@ -5434,6 +5433,59 @@ func (s *UpdateWorkflowSuite) TestUpdateWithStart() {
...
+						})
+					s.NoError(err)
+
+					// 2nd update-with-start (using the same Update ID but different Request ID)
+					uwsRes = <-sendUpdateWithStart(testcore.NewContext(), startReq, updateReq)
+
+					s.NoError(uwsRes.err)
+					startResp := uwsRes.response.Responses[0].GetStartWorkflow()
+					updateRep := uwsRes.response.Responses[1].GetUpdateWorkflow()
+					s.False(startResp.Started)
```
alexshtin, line 5483: Don't forget to assert on new `status` field here after you add it.

service/history/api/multioperation/api.go
alexshtin: I guess you need to rename it to `status` here too but as soon as it is just a comment it doesn't really matter.

alexshtin (approved):

service/history/api/multioperation/api.go
stephanos: 👍 I'll save the planet by not pushing the comment update now; and will fix it in the next PR.

PR #7660, opened 2025-04-24
Tag non-default task priority metric

stephanos:
## What changed?

Added a custom metric tag for labelling a metric as having a non-default priority level set.

## Why?

To gauge priority usage.

## How did you test it?

Ran `PriorityFairnessSuite` and verified it was true at least once.

### DIFF
--- a/service/matching/task.go
+++ b/service/matching/task.go
@@ -218,6 +218,13 @@ func (task *internalTask) isSyncMatchTask() bool {
 	return task.responseC != nil
 }
 
+func (task *internalTask) isCustomPriority() bool {
+	return task.event != nil &&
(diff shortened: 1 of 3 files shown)


service/matching/task.go
```diff
@@ -218,6 +218,13 @@ func (task *internalTask) isSyncMatchTask() bool {
 	return task.responseC != nil
 }
 
+func (task *internalTask) isCustomPriority() bool {
+	return task.event != nil &&
+		task.event.Data != nil &&
+		task.event.Data.Priority != nil &&
+		task.event.Data.Priority.PriorityKey != 0
```
stephanos, line 225: trying to be uber-defensive

PR #7662, opened 2025-04-25
CHASM: extract struct to separate files

alexshtin:
## What changed?

No code changes. Just file structure changes in `chasm` package.

## Why?

It is better to have structs in separate files as they are getting new methods.

## How did you test it?

N/A


chasm/component_field_option.go
```diff
@@ -0,0 +1,39 @@
+// The MIT License
+//
+// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
yycptt, line 5: plz use 2025 for all new files :)

yycptt (approved):

chasm/component_field_option.go
alexshtin: It is `make copyright`. I need to change `LICENSE` in the root.

PR #7665, opened 2025-04-25
fix cryptic error message on on DeleteVersion

Shivs11:
## What changed?

- WISOTT

## Why?

- Better user-experience. 
- Note: I shall be sending a follow-up PR removing any errors returned to the sdk as`Internal` errors when they are supposed to be returned as `FailedPrecondition`.

## How did you test it?

- Modified tests to check for this specifically.

### DIFF
--- a/service/worker/workerdeployment/client.go
+++ b/service/worker/workerdeployment/client.go
@@ -713,6 +713,12 @@ func (d *ClientImpl) DeleteWorkerDeploymentVersion(
 	if failure := outcome.GetFailure(); failure != nil {
 		if failure.GetApplicationFailureInfo().GetType() == errVersionNotFound {
 			return nil
+		} else if failure.GetCause().GetApplicationFailureInfo().GetType() == ErrVersionIsCurrentOrRamping {
+			return serviceerror.NewFailedPrecondition(ErrVersionIsCurrentOrRamping)
+		} else if failure.GetCause().GetApplicationFailureInfo().GetType() == ErrVersionIsDraining {
+			return serviceerror.NewFailedPrecondition(ErrVersionIsDraining)
+		} else if failure.GetCause().GetApplicationFailureInfo().GetType() == ErrVersionHasPollers {
+			return serviceerror.NewFailedPrecondition(ErrVersionHasPollers)
 		}
 		return serviceerror.NewInternal(failure.Message)
 	}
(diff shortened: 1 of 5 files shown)


service/worker/workerdeployment/client.go
```diff
@@ -713,6 +713,12 @@ func (d *ClientImpl) DeleteWorkerDeploymentVersion(
 	if failure := outcome.GetFailure(); failure != nil {
 		if failure.GetApplicationFailureInfo().GetType() == errVersionNotFound {
 			return nil
```
Shivs11, line 715: note - the reason why I check for `failure.GetCause().GetApplicationFailureInfo().GetType()` and not `failure.GetApplicationFailureInfo().GetType()` is because of the way these errors are thrown. In the case of `failure.GetApplicationFailureInfo().GetType()`, these errors are Activity errors that are *non-retryable* and thus have the information, from which we can parse the error, stored under `.getCause()...`.

service/worker/workerdeployment/util.go
```diff
@@ -88,12 +88,14 @@ const (
...
 	errVersionAlreadyCurrentType  = "errVersionAlreadyCurrent"
-	errVersionIsDraining          = "Version cannot be deleted since it is draining."
 	errVersionNotFound            = "Version not found in deployment"
-	errVersionHasPollers          = "Version cannot be deleted since it has active pollers."
-	errVersionIsCurrentOrRamping  = "Version cannot be deleted since it is current or ramping."
-	errConflictTokenMismatchType  = "errConflictTokenMismatch"
-	errFailedPrecondition         = "FailedPrecondition"
+
+	errConflictTokenMismatchType = "errConflictTokenMismatch"
+	errFailedPrecondition        = "FailedPrecondition"
```
carlydf, line 94: nit: if some of these are public, it's probably worth just making them all public. no need to change it now though, just pointing out.

tests/worker_deployment_version_test.go
```diff
@@ -529,6 +529,9 @@ func (s *DeploymentVersionSuite) TestVersionIgnoresDrainageSignalWhenCurrentOrRa
 }
 
 func (s *DeploymentVersionSuite) TestDeleteVersion_DeleteCurrentVersion() {
+	// Override the dynamic config so that we can verify we don't get any unexpected masked errors.
+	s.OverrideDynamicConfig(dynamicconfig.FrontendMaskInternalErrorDetails, true)
```
carlydf, line 533: nice!

carlydf (approved):

PR #7668, opened 2025-04-28
update workflow task only when the task is updated

hai719:
## What changed?
update workflow task only when the task is updated when applying replication task

## Why?
Always updating workflow task will cause versionedTransition to be cleared in standby side.

## How did you test it?
unit test.

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -7548,7 +7548,7 @@ func (ms *MutableStateImpl) applyUpdatesToUpdateInfos(
 }
 
 func (ms *MutableStateImpl) syncExecutionInfo(current *persistencespb.WorkflowExecutionInfo, incoming *persistencespb.WorkflowExecutionInfo, isSnapshot bool) error {
-	if transitionhistory.Compare(current.WorkflowTaskLastUpdateVersionedTransition, incoming.WorkflowTaskLastUpdateVersionedTransition) < 0 {
+	if transitionhistory.Compare(current.WorkflowTaskLastUpdateVersionedTransition, incoming.WorkflowTaskLastUpdateVersionedTransition) != 0 {
 		ms.workflowTaskManager.UpdateWorkflowTask(&historyi.WorkflowTaskInfo{
 			Version:             incoming.WorkflowTaskVersion,
 			ScheduledEventID:    incoming.WorkflowTaskScheduledEventId,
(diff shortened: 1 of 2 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -6184,7 +6184,7 @@ func (ms *MutableStateImpl) closeTransactionTrackLastUpdateVersionedTransition(
 		ms.executionInfo.UpdateInfos[updateID].LastUpdateVersionedTransition = currentVersionedTransition
 	}
 
-	if ms.workflowTaskUpdated && ms.HasPendingWorkflowTask() {
+	if ms.workflowTaskUpdated {
```
xwduan, line 6187: Will this change makes the update more often?

service/history/workflow/mutable_state_impl.go
```diff
@@ -7548,6 +7548,27 @@ func (ms *MutableStateImpl) applyUpdatesToUpdateInfos(
 }
 
 func (ms *MutableStateImpl) syncExecutionInfo(current *persistencespb.WorkflowExecutionInfo, incoming *persistencespb.WorkflowExecutionInfo, isSnapshot bool) error {
+	if transitionhistory.Compare(current.WorkflowTaskLastUpdateVersionedTransition, incoming.WorkflowTaskLastUpdateVersionedTransition) < 0 {
```
yycptt, line 7551: Need to use `!=` instead of `<` I think. That's what we do in other sync logic as well to handle the case where overall transition history is moving forward but VT for this particular state machine is becoming smaller.

service/history/workflow/mutable_state_impl.go
yycptt: Updates to this field will be more often yes. But we are not creating additional versioned transitions.

hai719: The workflow task will still check the versionedTransition. So no unnecessary update.

service/history/workflow/mutable_state_impl.go
```diff
@@ -7548,6 +7542,40 @@ func (ms *MutableStateImpl) applyUpdatesToUpdateInfos(
...
+
+			SuggestContinueAsNew:   incoming.WorkflowTaskSuggestContinueAsNew,
+			HistorySizeBytes:       incoming.WorkflowTaskHistorySizeBytes,
+			BuildId:                incoming.WorkflowTaskBuildId,
+			BuildIdRedirectCounter: incoming.WorkflowTaskBuildIdRedirectCounter,
+		})
+		workflowTaskVersionUpdated = true
+	}
+
+	if incoming.WorkflowTaskType == enumsspb.WORKFLOW_TASK_TYPE_SPECULATIVE {
```
hai719, line 7568: @yycptt handle speculative task as discussed

yycptt (approved):

PR #7669, opened 2025-04-29
Restore task queue name in proto message

dnr:
## What changed?
Restore the original task queue name in poll proto messages after the call to matching returns.

## Why?
If the frontend does an internal retry, we want the request to be the same as the first time it came in.

## How did you test it?
existing tests


carlydf (commented):
I like the other solution better (copy proto request so that we don't mutate it)

PR #7670, opened 2025-04-29
Don't mutate in matching client

dnr:
## What changed?
Copy matching client proto messages to route to a specific partition instead of mutating the original ones.

## Why?
Avoids validation and other problems on internal retries.

## How did you test it?
existing tests

### DIFF
--- a/client/matching/client.go
+++ b/client/matching/client.go
@@ -84,6 +84,7 @@ func (c *clientImpl) AddActivityTask(
 	ctx context.Context,
 	request *matchingservice.AddActivityTaskRequest,
 	opts ...grpc.CallOption) (*matchingservice.AddActivityTaskResponse, error) {
+	request = common.CloneProto(request)
 	client, err := c.pickClientForWrite(
 		request.GetTaskQueue(),
 		request.GetNamespaceId(),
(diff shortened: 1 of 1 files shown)


carlydf (approved):
this is my preferred solution. LGTM (should we test it?)

client/matching/client.go
```diff
@@ -155,6 +159,15 @@ func (c *clientImpl) PollWorkflowTaskQueue(
 }
 
 func (c *clientImpl) QueryWorkflow(ctx context.Context, request *matchingservice.QueryWorkflowRequest, opts ...grpc.CallOption) (*matchingservice.QueryWorkflowResponse, error) {
+	// use shallow copy since QueryRequest may contain a large payload
```
carlydf, line 162: nit: It could be nice if all of these APIs behaved the same, and we didn't need to think about whether to shallow copy or deep copy the request. We could define a helper "copy" function that basically does a shallow copy of everything except the task queue, and then use it everywhere. Since the task queue is the main thing we need to protect from mutation here.

carlydf (commented):
nit comment, but I still approve as is

client/matching/client.go
dnr: I agree, but they're all separate types so we'd need a bunch of functions. I think we should just generate a shallow copy helper in our proto helpers and use that for all these (and many other places I've wanted a shallow copy)

PR #7671, opened 2025-04-29
Ensure initSystemNamespaces call has a reasonable deadline

jprieto-temporal:
## What changed?

Added a timeout to the context used for initializing system namespaces

## Why?

The context that was used here had no deadline. It is the context tied to the lifetime of the fx application. A slow or non-responsive database can cause this call to hang indefinitely. It is better to let the context deadline expiration cause an error in that case.


rodrigozhou (approved):
Tests failed.

PR #7672, opened 2025-04-29
Fix call to assert.Fail in protoassert

rodrigozhou:
## What changed?

Fix call to `assert.Fail` in `protoassert`

## Why?

Bug fix: instead of getting details of the failed assertion, the test log was reporting a panic due to improper call to `assert.Fail`.

### DIFF
--- a/common/testing/protoassert/assert.go
+++ b/common/testing/protoassert/assert.go
@@ -26,6 +26,8 @@
 package protoassert
 
 import (
+	"fmt"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/assert"
 	"go.temporal.io/api/temporalproto"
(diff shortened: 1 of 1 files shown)


ychebotarev (approved):

common/testing/protoassert/assert.go
```diff
@@ -51,7 +53,7 @@ func ProtoEqual(t assert.TestingT, a proto.Message, b proto.Message) bool {
 		th.Helper()
 	}
 	if diff := cmp.Diff(a, b, protocmp.Transform()); diff != "" {
-		return assert.Fail(t, "Proto mismatch (-want +got):\n", diff)
+		return assert.Fail(t, fmt.Sprintf("Proto mismatch (-want +got):\n%v", diff))
```
dnr, line 56: can't this just be
```suggestion
		return assert.Fail(t, "Proto mismatch (-want +got):\n%v", diff)
```
?
I thought the assert/require functions do a printf internally

rodrigozhou: No, the signature is something like `func Fail(T, failureMessage string, msgAndArgs ...any)`.
It doesn't replace in `failureMessage`. Alternatively, I could add a failure message and a message.

dnr: Yes it does:
https://github.com/stretchr/testify/blob/89cbdd9e7b39eb58896d316a7495597d3aba4371/assert/assertions.go#L298

dnr: Oh sorry, there's a separate failureMessage string, nevermind

PR #7673, opened 2025-04-29
Generate Nexus completion link based on request ID

rodrigozhou:
## What changed?

Store request ID in `CallbackInfo` so we can generate Nexus completion link based on request ID

## Why?

Generate well-defined links for attached callback requests.


service/history/workflow/mutable_state_impl.go
```diff
@@ -598,23 +598,40 @@ func (ms *MutableStateImpl) ChasmTree() historyi.ChasmTree {
...
 		RunId:      ms.executionState.RunId,
+		// Backwards compatibility: this is the default link type.
 		Reference: &commonpb.Link_WorkflowEvent_EventRef{
 			EventRef: &commonpb.Link_WorkflowEvent_EventReference{
 				EventType: enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED,
 			},
 		},
-	})
+	}
+	requestIDInfo := ms.executionState.RequestIds[requestID]
```
bergundy, line 623: I don't think this works across CaN and workflow retries.

rodrigozhou: It doesn't work, if the workflow continued-as-new, retried, or reseted, it will generate a link to the new workflow.

rodrigozhou:
Added tests

service/history/workflow/mutable_state_impl.go
bergundy: Discussed offline, this is okay.

bergundy (approved):

PR #7674, opened 2025-04-29
Generate link in start workflow API

rodrigozhou:
## What changed?

Generate link in start workflow API

## Why?

Generate links to the event resulted from calling `StartWorkflowExecution`.


bergundy (commented):
I would add tests for every case where the link is generated to verify it is the right type. Hard to tell without reading the code very closely.

rodrigozhou:
Added tests.

bergundy (approved):

PR #7675, opened 2025-04-29
[GHA] Validate API and SDK version in server release

rodrigozhou:
## What changed?

Validate the API and SDK versions are tagged versions.

## Why?

Prevent releasing server with untagged depedencies.


bergundy (commented):
Thanks for doing this!

.github/workflows/create-tag.yml
```diff
@@ -73,6 +77,19 @@ jobs:
               echo "tag_no_v=${TAG}" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Validate API and SDK dependencies
+        if: ${{ github.event.inputs.skip_deps_check == 'false' }}
+        run: |
+          SEMVER_RE='^v[0-9]+\.[0-9]+\.[0-9]+$'
```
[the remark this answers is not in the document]
rodrigozhou: It doesn't work.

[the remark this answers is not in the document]
bergundy: Hmmm... maybe a bash oddity.

.github/workflows/create-tag.yml
```diff
@@ -9,6 +9,10 @@ on:
       tag:
         description: "Tag for new version (1.23.4)"
         required: true
+      skip_deps_check:
+        type: boolean
+        description: "Skip dependencies version check"
```
bergundy, line 14: I would add a sentence explaining when it is okay to disable this and make the default `true`.

bergundy (approved):
Please address the comment about the `skep_deps_check` parameter before merging.

.github/workflows/create-tag.yml
rodrigozhou: We wanna make this `true`, ie., _**skip**_ the check by default?

bergundy: No, my mistake. Just a comment would be nice though.

PR #7677, opened 2025-04-29
Send ActivityReset flag to the worker in heartbeat response

ychebotarev:
## What changed?

* Add "reset_activity" and "reset_heartbeat" flags to the activity info
* send "reset_requested" flag to the worker in response to heartbeat request
* if "reset_heartbeat" flag was set - reset heartbeats for new attempt only
* remove "reset_activity" and "reset_heartbeat" flags for the new attempt

Corresponding API PR: https://github.com/temporalio/api/pull/578

## Why?

To notify the workers/SDK that reset happens.

## How did you test it?

Running existing tests.

Note: replication in the next PR


alexshtin (approved):
Would love to see functional test similar to https://github.com/temporalio/temporal/blob/1afbdd27e3e7ff541d7c650bd6e5a7ac5fbfe5b5/tests/activity_test.go#L1171.

gow (approved):

PR #7678, opened 2025-04-30
CHASM: Non-Workflow Mutable State P2 Force Termination

yycptt:
## What changed?

- Support force termination of CHASM runs. 
- Generate retention timers for CHASM runs.
- Get rid of assumptions in workflow.Context and replication that the run is always a Workflow and runs workflow specific logic.

## Why?

- CHASM

## How did you test it?

- Added unit test

### DIFF
--- a/chasm/component.go
+++ b/chasm/component.go
@@ -5,28 +5,38 @@ package chasm
 import (
 	"context"
 	"reflect"
+	"strconv"
+
+	commonpb "go.temporal.io/api/common/v1"
 )
 
 type Component interface {
-	LifecycleState() LifecycleState
+	LifecycleState(Context) LifecycleState
 
-	// TBD: the framework can just put the component in terminated state
-	// component lifecycle state can still be running when getting terminated
-	// but framework will use some rule to block incoming operations
-	// Terminate()
+	Terminate(MutableContext, TerminateComponentRequest) (TerminateComponentResponse, error)
 
 	// we may not need this in the beginning
 	mustEmbedUnimplementedComponent()
 }
 
+type TerminateComponentRequest struct {
+	Identity string
+	Reason   string
+	Details  *commonpb.Payloads
+}
+
+type TerminateComponentResponse struct{}
+
 // Embed UnimplementedComponent to get forward compatibility
 type UnimplementedComponent struct{}
 
-func (UnimplementedComponent) LifecycleState() LifecycleState {
+func (UnimplementedComponent) LifecycleState(Context) LifecycleState {
 	return LifecycleStateUnspecified
 }
 
-// func (UnimplementedComponent) Terminate() {}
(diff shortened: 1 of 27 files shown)


service/history/ndc/workflow.go
yycptt, line 1: cc @yux0 @xwduan plz help review changes in this file.

service/history/ndc/workflow.go
```diff
@@ -216,15 +225,9 @@ func (r *WorkflowImpl) FlushBufferedEvents() error {
 	return nil
 }
 
-func (r *WorkflowImpl) failWorkflowTask(
-	lastWriteVersion int64,
-) (*historypb.HistoryEvent, error) {
+func (r *WorkflowImpl) failWorkflowTask() (*historypb.HistoryEvent, error) {
```
lina-temporal, line 228: (minor) maybe a `softassert` here to make sure MS is backing a workflow? Up to you.

lina-temporal (approved):

chasm/tree.go
```diff
@@ -77,6 +78,15 @@ type (
...
 		//
 		// encodedPath string
+
+		// When terminated is true, regardless of the Lifecycle state of the component,
+		// the component will be considered as closed.
+		//
+		// This right now only applies to the root node and used to update MutableState
+		// executionState and executionStatus and trigger retention timers.
+		// We could consider extending this to sub-components in the future as well, and
+		// block any progress operations on them if they are force terminated.
```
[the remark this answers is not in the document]
yycptt: oh I was thinking maybe the force terminate concept applies not only to root but to sub components as well. I will improve the comment here.

service/history/ndc/workflow.go
yycptt: yeah but I am thinking about adding those asserts to all workflow specific MutableState methods (AddXXXEvents for example), then we will have a better coverage and don't need to worry about breaking tests that are using mockMutableState.

I will create a separate ticket for this.

chasm/component.go
```diff
@@ -5,28 +5,38 @@ package chasm
...
 
 type Component interface {
-	LifecycleState() LifecycleState
+	LifecycleState(Context) LifecycleState
 
-	// TBD: the framework can just put the component in terminated state
-	// component lifecycle state can still be running when getting terminated
-	// but framework will use some rule to block incoming operations
-	// Terminate()
+	Terminate(MutableContext, TerminateComponentRequest) (TerminateComponentResponse, error)
```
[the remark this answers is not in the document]
yycptt: yeah I think we want to make them as stable as possible. A lot of (most?) exported method/functions are variadic functions. Interfaces definitions also have corresponding embedUnimplmented implementation to avoid breaking changes.

We eventually need this when more chasm components are develop, and since it's straight forward to do it for this Terminate method, so think I can just do it from the beginning.

yux0 (approved):

PR #7679, opened 2025-04-30
CHASM: extract generic fields iterator

alexshtin:
## What changed?

CHASM: extract generic fields iterator.
Thanks for the idea @lina-temporal! I believe iterator is better than "walker".

## Why?

To reduce code duplication and improve error messages.

## How did you test it?

Added unit tests.

### DIFF
--- a/chasm/fields_iterator.go
+++ b/chasm/fields_iterator.go
@@ -26,6 +26,7 @@ package chasm
 
 import (
 	"fmt"
+	"iter"
 	"reflect"
 	"strings"
 
@@ -56,7 +57,7 @@ type fieldInfo struct {
 	err  error
 }
 
-func fieldsOf(valueV reflect.Value) func(func(fieldInfo) bool) {
+func fieldsOf(valueV reflect.Value) iter.Seq[fieldInfo] {
 	valueT := valueV.Type()
 	dataFieldFound := false
 	return func(yield func(fi fieldInfo) bool) {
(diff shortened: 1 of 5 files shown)


chasm/fields_iterator.go
```diff
@@ -0,0 +1,128 @@
...
+			fieldT := fieldV.Type()
+			if fieldT == UnimplementedComponentT {
+				continue
+			}
+			fieldN := fieldName(valueT.Elem().Field(i))
+
+			var (
+				fieldK   = fieldKindUnspecified
+				fieldErr error
+			)
```
lina-temporal, line 76: (minor)

```suggestion
				var fieldErr error
				fieldK := fieldKindUnspecified
```

reads smoother to me

chasm/fields_iterator.go
```diff
@@ -0,0 +1,128 @@
...
+					fieldK = fieldKindSubField
+				case chasmCollectionTypePrefix:
+					fieldK = fieldKindSubCollection
+				default:
+					prefix = strings.TrimPrefix(prefix, "*")
+					switch prefix {
+					case chasmFieldTypePrefix:
+						fieldErr = serviceerror.NewInternal(fmt.Sprintf("chasm field must be of type chasm.Field[T] not *chasm.Field[T] in component %s", valueT.String()))
+					case chasmCollectionTypePrefix:
+						fieldErr = serviceerror.NewInternal(fmt.Sprintf("chasm collection must be of type chasm.Collection[T] not *chasm.Collection[T] in component %s", valueT.String()))
```
lina-temporal, line 96: IMO, we should simply check if the prefix `StartsWith` a `*`, and then return the error message given in your `default` block. I don't think the more specific messages here justify the maintenance.

lina-temporal (approved):
Nice refactor! 🚢   few minor comments, take or leave

chasm/fields_iterator.go
alexshtin: I added it because during testing I made this mistake myself. I added `*Field` and it got an error that it is not supported. So I decided to add explicit message for pointers. I refactor it though, and simplified.

PR #7680, opened 2025-04-30
CHASM: children iterator

alexshtin:
## What changed?

CHASM: children iterator: change "walker" pattern to iterator. 

## Why?

More idiomatic go. Plan to use it in few other places.

## How did you test it?

Existing tests.


yycptt (approved):
👍

PR #7684, opened 2025-04-30
Refactor: remove custom helpers from functional tests

alexshtin:
## What changed?

Refactor: remove custom helpers from functional tests.

## Why?

There are standard `s.Eventually` and project wide `historyrequire` package.

## How did you test it?

Run them.

## Potential risks

I am not sure about timing which I put there. I tried to be conservative but if it is too aggressive, times need to be bumped up.

### DIFF
--- a/tests/client_misc_test.go
+++ b/tests/client_misc_test.go
@@ -37,6 +37,7 @@ import (
 	batchpb "go.temporal.io/api/batch/v1"
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
+	historypb "go.temporal.io/api/history/v1"
 	"go.temporal.io/api/serviceerror"
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
 	updatepb "go.temporal.io/api/update/v1"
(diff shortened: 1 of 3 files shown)


tests/client_misc_test.go
```diff
@@ -128,11 +129,13 @@ func (s *ClientMiscTestSuite) TestTooManyChildWorkflows() {
 	future, err := s.SdkClient().ExecuteWorkflow(ctx, options, parentWorkflow)
 	s.NoError(err)
 
-	s.HistoryContainsFailureCausedBy(
-		ctx,
-		parentWorkflowId,
-		enumspb.WORKFLOW_TASK_FAILED_CAUSE_PENDING_CHILD_WORKFLOWS_LIMIT_EXCEEDED,
-	)
+	s.WaitForHistoryEventsSuffix(`
```
ychebotarev, line 132: 1. you replace it with **WaitForHistoryEventsSuffix** here, but later you replace it with **WaitForHistoryEvents**. Any specific reason? I would expect the same function.
2. HistoryContainsFailureCausedBy is shorter, and doesn't require the exact format of the error and details like "limit of 10". Users don't need to know the whole history. What is the benefit of removing it?
3. I suggest to keep it, but rename to something like "WaitForHistoryFailure". may be change signature to match another "WaitForHistoryXXX" functions

alexshtin: 1. Because I run it few times and figured out that it produce different history every time. This is exactly reason why I created those "*Suffix" helpers.
2. and 3. Exact history is more verbose. You see what is expected right away. Not only last event. And it is also generic and doesn't require special helper for every case. You used it once, you know how to use it in all other places. Assert on error message is also nice addition here.

alexshtin: Let's agree to disagree.

ychebotarev (approved):

PR #7688, opened 2025-04-30
Add NamespaceTag for RedirectAPIInvocation

hehaifengcn:

### DIFF
--- a/common/rpc/interceptor/redirection.go
+++ b/common/rpc/interceptor/redirection.go
@@ -218,7 +218,7 @@ func (i *Redirection) handleLocalAPIInvocation(
 ) (_ any, retError error) {
 	scope, startTime := i.BeforeCall(dcRedirectionMetricsPrefix + methodName)
 	defer func() {
-		i.AfterCall(scope, startTime, i.currentClusterName, "", retError)
+		i.AfterCall(scope, startTime, i.currentClusterName, "local", retError)
 	}()
 	return handler(ctx, req)
 }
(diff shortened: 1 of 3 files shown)


common/rpc/interceptor/redirection.go
```diff
@@ -242,7 +242,7 @@ func (i *Redirection) handleLocalAPIInvocation(
 ) (_ any, retError error) {
 	scope, startTime := i.BeforeCall(dcRedirectionMetricsPrefix + methodName)
 	defer func() {
-		i.AfterCall(scope, startTime, i.currentClusterName, retError)
+		i.AfterCall(scope, startTime, i.currentClusterName, "", retError)
```
hehaifengcn, line 245: we don't care about namespace for local invocation.

common/rpc/interceptor/redirection.go
```diff
@@ -296,11 +296,16 @@ func (i *Redirection) AfterCall(
...
 	retError error,
 ) {
 	metricsHandler = metricsHandler.WithTags(metrics.TargetClusterTag(clusterName))
-	metrics.ClientRedirectionRequests.With(metricsHandler).Record(1)
 	metrics.ClientRedirectionLatency.With(metricsHandler).Record(i.timeSource.Now().Sub(startTime))
+
+	if len(namespaceName) != 0 {
+		metricsHandler = metricsHandler.WithTags(metrics.NamespaceTag(namespaceName))
+	}
+	metrics.ClientRedirectionRequests.With(metricsHandler).Record(1)
```
yux0, line 308: Can we remove the counter and depends on the latency metric count?

hehaifengcn: we need counter for namespace label. latency + namespace is not recommended.

yux0: do we also need the namespace tag on error count?

hehaifengcn: I think it wouldn't hurt?

PR #7689, opened 2025-04-30
Remove license header from every file

alexshtin:
## What changed?

Remove license header from every file. Because it is really hard to follow in this PR here is the summary:
1. License header is removed from all `.go` and `.proto` files :fireworks::fireworks::fireworks:. 
2. `LICENSE` file in the root directory has only Temporal and Uber copyrights.
3. 5 other `LICENSE` files added to the packages which have copyrights different from Temporal and Uber: Datadog, Xargin, "Mat Ryer, Tyler Bunnell and contributors".
4. `license_file` flag is removed from all code generation tools.
5. `copyright_file` flag is removed from `go:generate mockgen` directive.
6. All copyright related targets are removed from `Makefile`.
7. Updated Temporal copyright year to 2025 everywhere.

## Why?

I double checked with legal department that it is not needed to have license header in every file. One file per repo is enough. I put all copyrights to the root `LICENSE` file and removed header from all other files. Also updated tools and `Makefile`.


alexshtin:
> Я тебя породил, я тебя и убью!

lina-temporal (approved):

PR #7691, opened 2025-05-01
Add event consecutive check when handling sync state

xwduan:
## What changed?

Add event consecutive check when handling sync state for state based replication
## Why?

More safe guard if source is not behaving correctly when sending events.
## How did you test it?

unit test
## Potential risks

no risk
## Documentation

n/a
## Is hotfix candidate?

yes.

### DIFF
--- a/config/dynamicconfig/development-xdc.yaml
+++ b/config/dynamicconfig/development-xdc.yaml
@@ -38,7 +38,7 @@ frontend.workerVersioningWorkflowAPIs:
   - value: true
 system.enableNexus:
   # enableNexus also controls transition history which can't be enabled now
-  - value: false
+  - value: false 
 component.nexusoperations.callback.endpoint.template:
   - value: http://localhost:7243/api/v1/namespaces/{{.NamespaceName}}/nexus/callback
 component.callbacks.allowedAddresses:
--- a/config/dynamicconfig/development-xdc.yaml
+++ b/config/dynamicconfig/development-xdc.yaml
@@ -37,9 +37,8 @@ frontend.workerVersioningDataAPIs:
 frontend.workerVersioningWorkflowAPIs:
   - value: true
 system.enableNexus:
-  - value: true
-history.enableTransitionHistory:
-  - value: true
+  # enableNexus also controls transition history which can't be enabled now
+  - value: false
 component.nexusoperations.callback.endpoint.template:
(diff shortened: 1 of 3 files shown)


config/dynamicconfig/development-xdc.yaml
```diff
@@ -37,8 +37,9 @@ frontend.workerVersioningDataAPIs:
 frontend.workerVersioningWorkflowAPIs:
   - value: true
 system.enableNexus:
-  # enableNexus also controls transition history which can't be enabled now
-  - value: false 
+  - value: true
+history.enableTransitionHistory:
+  - value: true
 component.nexusoperations.callback.endpoint.template:
   - value: http://localhost:7243/api/v1/namespaces/{{.NamespaceName}}/nexus/callback
```
hai719, line 44: `/api/v1` needs to be removed.

hai719 (approved):

config/dynamicconfig/development-xdc.yaml
yycptt: Guess we can enable it in cass & sql config as well. It will cause nexus to take a different code path. 

I'd suggest to config change in a separate PR though.

yycptt (approved):

config/dynamicconfig/development-xdc.yaml
xwduan: Sure. Will make a separate pr for this.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -401,6 +401,31 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTask(
 			historyEventBatchs = append(historyEventBatchs, sourceEvents)
 		}
 	}
+	checkEventBatchs := func() error {
```
hai719, line 404: nit: These functions are useful. It would be useful to make them as helper functions so they can be used in other places, or combine with existing `validateEventsSlice`

hai719 (approved):

service/history/ndc/workflow_state_replicator.go
xwduan: Yeah, I thought about it, but in this file, we have 3 ways to write history event and each of them require different logic to verify the events..

PR #7692, opened 2025-05-01
Add target cluster to replication tasks

zigeH:
## What changed?

Add a targetCluster field to the replication task and propagate it from the failover workflow all the way to the task generation logic. Include this parameter in the SyncWorkflowStateTask. Finally, update the stream sender to check this parameter and only send the task if target == targetCluster; otherwise, skip transmission.


proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -886,6 +886,7 @@ message GenerateLastHistoryReplicationTasksRequest {
 
     string namespace_id = 1;
     temporal.api.common.v1.WorkflowExecution execution = 2;
+    string target_cluster_id = 3;
```
yux0, line 889: Could we have a list of target clusters?

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -380,6 +380,7 @@ message ReplicationTaskInfo {
     repeated ReplicationTaskInfo task_equivalents = 20;
     history.v1.VersionHistoryItem last_version_history_item = 21;
     bool is_first_task = 22;
+    string target_cluster_id= 23;
```
yux0, line 383: same

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
yux0: the cluster id means the initial failover version. either use it with the IFV or rename to cluster name

service/history/replication/stream_sender.go
```diff
@@ -29,6 +29,7 @@ import (
 	historyi "go.temporal.io/server/service/history/interfaces"
 	"go.temporal.io/server/service/history/shard"
 	"go.temporal.io/server/service/history/tasks"
+	"golang.org/x/exp/slices"
```
yux0, line 32: just use "slices"

service/history/api/replication/generate_task.go
```diff
@@ -41,7 +41,7 @@ func GenerateTask(
 	defer func() { workflowLease.GetReleaseFn()(retError) }()
 
 	mutableState := workflowLease.GetMutableState()
-	replicationTasks, stateTransitionCount, err := mutableState.GenerateMigrationTasks()
+	replicationTasks, stateTransitionCount, err := mutableState.GenerateMigrationTasks(request.GetTargetClusters())
```
yux0, line 44: can you add a sanity check to make sure all target clusters are in the connected cluster list?

service/history/replication/stream_sender.go
zigeH: Go automatically uses slices as the package name

service/history/api/replication/generate_task.go
zigeH: The connected cluster info isn’t exposed here, so I think the sanity check should happen at the workflow level. When this feature is invoked, it's already checked as part of force replication. Once we include this parameter in the replication task, if the target cluster isn't connected, ShouldProcessTask will simply skip the check for that target—so there shouldn’t be any impact.

service/history/replication/stream_sender.go
yux0: This is experiment pkg. use https://pkg.go.dev/slices

yux0 (approved):

PR #7696, opened 2025-05-01
CHASM: Engine Update/ReadComponent implementation

yycptt:
## What changed?
- Implement chasm engine Update/Read Component method

## Why?
- CHASM work stream

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/ref_test.go
+++ b/chasm/ref_test.go
@@ -1,0 +1,71 @@
+package chasm
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"go.temporal.io/server/common/testing/testvars"
+	"go.uber.org/mock/gomock"
+)
+
+type componentRefSuite struct {
+	suite.Suite
+	*require.Assertions
+
+	controller *gomock.Controller
+
+	registry *Registry
+}
+
+func TestComponentRefSuite(t *testing.T) {
+	suite.Run(t, new(componentRefSuite))
+}
+
+func (s *componentRefSuite) SetupTest() {
+	// Do this in SetupSubTest() as well, if we have sub tests in this suite.
+	s.Assertions = require.New(s.T())
+
+	s.controller = gomock.NewController(s.T())
+
+	s.registry = NewRegistry()
+	err := s.registry.Register(newTestLibrary(s.controller))
+	s.NoError(err)
+}
+
+func (s *componentRefSuite) TestShardingKey() {
+	tv := testvars.New(s.T())
+	entityKey := EntityKey{
+		tv.NamespaceID().String(),
+		tv.WorkflowID(),
+		tv.RunID(),
+	}
+	ref := NewComponentRef[*TestComponent](entityKey)
+
+	shardingKey, err := ref.ShardingKey(s.registry)
+	s.NoError(err)
+
+	rc, ok := s.registry.ComponentOf(reflect.TypeFor[*TestComponent]())
+	s.True(ok)
+
+	s.Equal(rc.shardingFn(entityKey), shardingKey)
+	s.Equal(rc.FqType(), ref.archetype)
+}
+
+func (s *componentRefSuite) TestShardID() {
+	tv := testvars.New(s.T())
+	entityKey := EntityKey{
+		tv.NamespaceID().String(),
+		tv.WorkflowID(),
+		tv.RunID(),
+	}
+	ref := NewComponentRef[*TestComponent](entityKey)
+
+	shardID, err := ref.ShardID(s.registry, 10)
+	s.NoError(err)
+
+	// Here we are just checking that the shardID field is populated in the ref.
+	// The actual shardID value is not that important.
+	s.Equal(shardID, ref.shardID)
+}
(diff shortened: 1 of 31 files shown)


service/history/chasm_engine.go
```diff
@@ -0,0 +1,204 @@
...
+		ctx,
+		shardContext,
+	); err != nil {
+		return chasm.ComponentRef{}, err
+	}
+
+	newRef, ok := mutableContext.Ref(component)
+	if !ok {
+		return chasm.ComponentRef{}, serviceerror.NewInternal(
+			fmt.Sprintf("component not found in the new compoenent tree after mutation, componentRef: %+v", ref),
```
lina-temporal, line 109:
```suggestion
			fmt.Sprintf("component not found in the new component tree after mutation, componentRef: %+v", ref),
```

service/history/chasm_engine.go
```diff
@@ -0,0 +1,204 @@
...
+) *ChasmEngine {
+	return &ChasmEngine{
+		entityCache:     entityCache,
+		shardController: shardController,
+		registry:        registry,
+		configs:         configs,
+	}
+}
+
+func (e *ChasmEngine) UpdateComponent(
```
lina-temporal, line 64: Will tasks be calling `UpdateComponent`? Is it a problem if the workflow lease lock is already held (before `g.getExecutionLease`)?

lina-temporal (commented):
just some questions and minor suggestions, otherwise LGTM (I'm assuming it's draft before the tests are ready)

service/history/chasm_engine.go
yycptt: Only side effect task executor can call UpdateComponent. so yeah we need to make sure the mutable state is not locked before calling the side effect task executor. And all verification logic needs to be put into the validateFn field of the ComponentRef.

service/history/chasm_engine.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	defer func() {
+		executionLease.GetReleaseFn()(retError)
+	}()
+
+	mutableState := executionLease.GetMutableState()
+	chasmTree, ok := mutableState.ChasmTree().(*chasm.Node)
+	if !ok {
+		return chasm.ComponentRef{}, serviceerror.NewInternal(
+			fmt.Sprintf("CHASM tree implementation not properly wired up, encountered type: %T", mutableState.ChasmTree()),
```
alexshtin, line 83: Let's agree on `CHASM` vs `chasm` in error messages. I am fine with both, but `CHASM` looks better to me (it is an abbreviation after all).

alexshtin: also
```suggestion
			fmt.Sprintf("CHASM tree implementation not properly wired up, encountered type: %T, expected type: %T", mutableState.ChasmTree(), &chasm.Node{}),
```

service/history/chasm_engine.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		// Always release the lease with nil error since this is a read only operation
+		// So even if it fails, we don't need to clear and reload mutable state.
+		executionLease.GetReleaseFn()(nil)
+	}()
+
+	chasmTree, ok := executionLease.GetMutableState().ChasmTree().(*chasm.Node)
```
alexshtin, line 132: I don't understand this cast. Why not to change `NewContext` to accept `ChasmTree` interface (instead of `*Node`) and add missing methods to `ChasmTree`?

alexshtin (approved):
I believe some unit tests won't hurt here, but because I am going to add e2e tests soon, I guess, you might skip them. `registry.go` had 100% coverage before though :slightly_smiling_face:.

chasm/ref_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (s *componentRefSuite) initAssertions() {
+	// `s.Assertions` (as well as other test helpers which depends on `s.T()`) must be initialized on
+	// both test and subtest levels (but not suite level, where `s.T()` is `nil`).
+	//
+	// If these helpers are not reinitialized on subtest level, any failed `assert` in
+	// subtest will fail the entire test (not subtest) immediately without running other subtests.
+
+	s.Assertions = require.New(s.T())
+}
```
lina-temporal, line 47: imo just simplify to 
```suggestion
func (s *componentRefSuite) SetupSubTest() {
	s.Assertions = require.New(s.T())
}
```

lina-temporal (approved):
LGTM!

service/history/chasm_engine.go
yycptt: +1 for `CHASM` as well.

Updated the message.

chasm/ref_test.go
yycptt: I removed SetupSubTest since there's no sub test in this suite right now. Also added a comment in SetupTest().

service/history/chasm_engine.go
yycptt: > change NewContext to accept ChasmTree interface

That's a cycle dependency. `ChasmTree` is in `historyi` package which depends on `chasm`. 
That interface and noop tree impl will eventually be removed, so I guess I will just go with the simplest type assertion approach now.

PR #7697, opened 2025-05-01
Handle links in Nexus callbacks

rodrigozhou:
## What changed?

Handle links in Nexus callbacks.
Links in `StartWorkflowRequest` will dedup the links that are also in Nexus callbacks.

## Why?

Nexus links are gonna be associated with the Nexus callbacks, and they will follow the callback when there is a continue-as-new or retries.

## How did you test it?

Added tests

### DIFF
--- a/service/history/api/describeworkflow/api.go
+++ b/service/history/api/describeworkflow/api.go
@@ -333,11 +333,14 @@ func buildCallbackInfo(
 	callback callbacks.Callback,
 	outboundQueueCBPool *circuitbreakerpool.OutboundQueueCircuitBreakerPool,
 ) (*workflowpb.CallbackInfo, error) {
-	cbSpec, err := workflow.PersistenceCallbackToAPICallback(callback.Callback)
-	if err != nil || cbSpec.GetNexus() == nil {
-		// Errors only happen for non-nexus callbacks.
+	if callback.Callback.GetNexus() == nil {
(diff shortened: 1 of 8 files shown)


service/history/api/describeworkflow/api.go
```diff
@@ -333,18 +333,9 @@ func buildCallbackInfo(
...
-			Nexus: &commonpb.Callback_Nexus{
-				Url:    variant.Nexus.GetUrl(),
-				Header: variant.Nexus.GetHeader(),
-			},
-		}
-		destination = variant.Nexus.GetUrl()
-	default:
+	cbSpec, err := workflow.PersistenceCallbackToAPICallback(callback.Callback)
+	if err != nil || cbSpec.GetNexus() == nil {
+		// Errors only happen for non-nexus callbacks.
```
bergundy, line 338: Hesitant to make such assumptions in the code. If the function intends to mark something as an ignorable error it should probably just return nil or have a specific error type that you can filter out.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2586,7 +2586,9 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
 		))
 	}
 	for idx, cb := range completionCallbaks {
```
bergundy, line 2588: nit: while you're here can you fix this typo?

PR #7699, opened 2025-05-02
Pass github token to arduino/setup-protoc action

rodrigozhou:
## What changed?

Pass Github token to `arduino/setup-protoc` action.

## Why?

Avoid rate limiting error:
```
Error: Error: API rate limit exceeded for 20.102.199.133. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```


carlydf (approved):
nice!!

PR #7700, opened 2025-05-02
CHASM: Non-Workflow Mutable State P3 State Replicator

yycptt:
## What changed?

- Support no event case in state replicator.

## Why?

- CHASM runs won't have events.

## How did you test it?

- Existing tests
- Unit tests

### DIFF
--- a/service/history/ndc/workflow_state_replicator.go
+++ b/service/history/ndc/workflow_state_replicator.go
@@ -324,7 +324,6 @@ func (r *WorkflowStateReplicatorImpl) ReplicateVersionedTransition(
 	}
 }
 
-//nolint:revive // cognitive complexity 35 (> max enabled 25)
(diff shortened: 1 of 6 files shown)


yycptt (commented):
Functional test are failing because I did something wrong in my previous [PR](https://github.com/temporalio/temporal/pull/7678) I think. Will rebase once that's fixed.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -1290,14 +1296,50 @@ func (r *WorkflowStateReplicatorImpl) backfillHistory(
 	namespaceID namespace.ID,
 	workflowID string,
 	runID string,
-	originalRunID string,
 	mutableState *workflow.MutableStateImpl,
-	lastEventID int64,
```
xwduan, line 1295: This is much cleaner. Thanks.

PR #7705, opened 2025-05-02
Revert generating RequestIdReference links

rodrigozhou:
## What changed?

Revert generating `RequestIdReference` links.

## Why?

PRs #7673 and #7674 introduced `RequestIdReference` links which depends on a new release of sdk-go. Reverting this for now so the main branch won't use an untagged sdk-go version.


ychebotarev (approved):

bergundy:
Only the tests can be reverted, the behavior can stay as is IMHO.

PR #7709, opened 2025-05-03
Add internal pod health check to DeepHealthCheck

laniehei:
## What changed?
-> Part I (History service):
@dnr and others have recently implemented a health check to speed up the history service rollout by explicitly marking it as healthy when the history shards have stabilized after startup instead of waiting an arbitrary time period.

In this PR, I've reused this check in the history service to ensure that the history service has marked the host as healthy. This avoids a false positive because the host isn't yet receiving error metrics, and so it might not be accurate to say it's serving if history shards haven't ceased shuffling. Added health status starting to differentiate from unhealthy host status to use this upstream and help with error propagation and debugging.

(footnote) Further background on the host serving status for documentation and general interest:
The status is set [here](https://github.com/temporalio/temporal/blob/main/service/history/service.go#L82) after it has both joined ringpop, and the shards have stabiliized. It reuses a gRPC history health check that was previously implemented, but was unused. 

-> Part II (Frontend service, aggregator of history.DeepHealthCheck): 
In the frontend service, I've added the default percentage of 5% based on the guidance from @00Sim that this would be quite a high error percentage, accounting for rollouts. I've added this percentage to the NOT_SERVING metric to avoid a combined outage where some pods are completely down, while some are crashlooping. 

## Why?
This PR starts to seek improvement for failure detection in the case of crashloops. 

It's a small improvement as part of the effort to detect crashloopbackoff by not marking pods as healthy until they've been marked as such by the service while starting up. If the pods do not ever "start up", then it's a good indication that something is wrong. 

## How did you test it?
We can test during rollout by ensuring the failure detection is working as expected. 

## Potential risks
If the existing check is not marking pods as healthy, we could break the autodetection code. This is not directly workflow (customer)-impacting. 

## Documentation
N/A

## Is hotfix candidate?
No


common/dynamicconfig/constants.go
```diff
@@ -761,6 +761,11 @@ This config is EXPERIMENTAL and may be changed or removed in a later release.`,
 		0.5,
 		`HistoryHostErrorPercentage is the percentage of hosts that are unhealthy`,
 	)
+	HistoryHostStartingPercentage = NewGlobalFloatSetting(
+		"frontend.historyHostStartingPercentage",
+		0.05,
+		`HistoryHostStartingPercentage is the percentage of hosts that are not yet ready to serve traffic`,
```
dnr, line 767: Maybe say a little more here about what happens when the percentage is above or below this threshold?

common/dynamicconfig/constants.go
```diff
@@ -761,6 +761,11 @@ This config is EXPERIMENTAL and may be changed or removed in a later release.`,
 		0.5,
 		`HistoryHostErrorPercentage is the percentage of hosts that are unhealthy`,
 	)
+	HistoryHostStartingPercentage = NewGlobalFloatSetting(
+		"frontend.historyHostStartingPercentage",
+		0.05,
```
dnr, line 766: nitpicky, but: "percentage" implies a number from 0 to 100. A fraction like 0.05 should be a "fraction".

service/frontend/health_check.go
```diff
@@ -91,15 +94,29 @@ func (h *healthCheckerImpl) Check(ctx context.Context) (enumsspb.HealthState, er
...
 
-	if (failedHostCount / float64(len(hosts))) > h.hostFailurePercentage() {
+	startingHostCountPercentage := startingHostCount / float64(len(hosts))
+	if startingHostCountPercentage > h.hostStartingPercentage() {
+		h.logger.Warn("health check exceeded host starting percentage threshold", tag.NewFloat64("host starting percentage threshold", h.hostStartingPercentage()))
+		return enumsspb.HEALTH_STATE_STARTING, nil
+	}
+
+	failedHostCountPercentage := failedHostCount / float64(len(hosts))
+	if failedHostCountPercentage+startingHostCountPercentage > h.hostFailurePercentage() {
```
dnr, line 118: I'm trying to think if it makes sense to include "starting" in with "failed" here... I think so but not 100% clear. Did you have a reason or example in mind when deciding that?

common/dynamicconfig/constants.go
laniehei: I think this is up to the consumer of the float setting, not the float setting itself.

common/dynamicconfig/constants.go
laniehei: technically the correct word would be [proportion](https://math.stackexchange.com/questions/1350980/correct-term-for-percentage-in-decimal-form) according to the math stack exchange 😅 It looks like we use this notation frequently across the dynamic config. Maybe I'm wrong, but when I see a decimal and the unit is percentage, I assume this notation.

proto/internal/temporal/server/api/enums/v1/cluster.proto
```diff
@@ -38,4 +38,5 @@ enum HealthState {
     HEALTH_STATE_UNSPECIFIED = 0;
     HEALTH_STATE_SERVING = 1;
     HEALTH_STATE_NOT_SERVING = 2;
+    HEALTH_STATE_STARTING = 3;
```
yux0, line 41: not sure if the starting make sense or just use UNKNOWN?

common/dynamicconfig/constants.go
dnr: I said it was nitpicky :)

"proportion" is also great

common/dynamicconfig/constants.go
dnr: But the setting has a single meaning and a specific effect.. the docstring should describe those. And the name should too, ideally. There shouldn't be multiple consumers with different semantics for the same setting.

Anyway, 0.05 isn't "the percentage of hosts that are not yet ready to serve traffic", the percentage of hosts not ready to serve traffic is usually 0%, sometimes 10%, sometimes 50%, etc. The setting is a threshold for some behavior to change based on that value.

service/frontend/health_check.go
dnr: Makes sense to me

proto/internal/temporal/server/api/enums/v1/cluster.proto
laniehei: I could change this to NOT_READY? I suppose this could also be flipped "off" by the service if the host/pod is no longer healthy re: our conversation earlier that we don't just set this on startup

common/dynamicconfig/constants.go
laniehei: changed!

common/dynamicconfig/constants.go
laniehei: I see your point. I've changed the dynamic config documentation and the naming to better reflect its intent.

service/frontend/health_check.go
```diff
@@ -91,15 +94,29 @@ func (h *healthCheckerImpl) Check(ctx context.Context) (enumsspb.HealthState, er
...
+			startingHostCount++
+		case enumsspb.HEALTH_STATE_SERVING:
+			// Do nothing.
 		}
 	}
 	close(receiveCh)
 
-	if (failedHostCount / float64(len(hosts))) > h.hostFailurePercentage() {
+	startingHostCountPercentage := startingHostCount / float64(len(hosts))
+	if startingHostCountPercentage > h.hostStartingPercentage() {
```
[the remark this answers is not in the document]
laniehei: I've change the code such that at least 2 hosts must be failing this check in order to trigger this failure

proto/internal/temporal/server/api/enums/v1/cluster.proto
laniehei: Changed!

service/frontend/health_check.go
```diff
@@ -91,17 +95,49 @@ func (h *healthCheckerImpl) Check(ctx context.Context) (enumsspb.HealthState, er
...
+		numHostsToFail = float64(totalHosts) * proportionOfDeclinedServingHosts
+	}
+	var proportionOfDeclinedServiceHosts float64
+	if numHostsToFail < 2 {
+		correctedProportion := 2.0 / float64(totalHosts)
+		proportionOfDeclinedServiceHosts = math.Max(proportionOfDeclinedServingHosts, correctedProportion)
+	} else {
+		proportionOfDeclinedServiceHosts = proportionOfDeclinedServingHosts
+	}
+	return proportionOfDeclinedServiceHosts
```
yux0, line 142: `if 1 / proportionOfDeclinedServingHosts > totalHost {return 1/ totalHosts} else {return proportionOfDeclinedServingHosts}`

laniehei: I've done you one better :) I definitely overcomplicated the first iteration

service/frontend/health_check.go
```diff
@@ -91,17 +95,40 @@ func (h *healthCheckerImpl) Check(ctx context.Context) (enumsspb.HealthState, er
...
+	proportionOfDeclinedServiceHosts := ensureMinimumProportionOfHosts(h.hostDeclinedServingProportion(), len(hosts))
+
+	hostDeclinedServingProportion := hostDeclinedServingCount / float64(len(hosts))
+	if hostDeclinedServingProportion > proportionOfDeclinedServiceHosts {
+		h.logger.Warn("health check exceeded host declined serving proportion threshold", tag.NewFloat64("host declined serving proportion threshold", proportionOfDeclinedServiceHosts))
+		return enumsspb.HEALTH_STATE_DECLINED_SERVING, nil
+	}
+
+	failedHostCountProportion := failedHostCount / float64(len(hosts))
+	if failedHostCountProportion+hostDeclinedServingProportion > h.hostFailurePercentage() {
```
yux0, line 122: I am a little worry about false positive:
1. double count in failure 
2. the minimum proportion

service/frontend/health_check.go
```diff
@@ -91,17 +95,40 @@ func (h *healthCheckerImpl) Check(ctx context.Context) (enumsspb.HealthState, er
...
+		h.logger.Warn("health check exceeded host failure percentage threshold", tag.NewFloat64("host failure percentage threshold", h.hostFailurePercentage()), tag.NewFloat64("host failure percentage", failedHostCountProportion), tag.NewFloat64("host declined serving percentage", hostDeclinedServingProportion))
 		return enumsspb.HEALTH_STATE_NOT_SERVING, nil
 	}
 
 	return enumsspb.HEALTH_STATE_SERVING, nil
 }
+
+func ensureMinimumProportionOfHosts(proportionOfDeclinedServingHosts float64, totalHosts int) float64 {
+	minimumHostsFailed := 2.0 // We want to ensure that at least 2 fail before we notify the upstream.
```
yux0, line 131: nit: const

service/frontend/health_check.go
yux0: Yes. This is much better.

yux0 (approved):

service/frontend/health_check.go
laniehei: The double count in failure is a good point. If we make changes to the way we mark pods as unhealthy, then this could overlap. Currently, I'm not sure that we could both have persistence latency/errors, and also be in this not serving state, because this internal hostDeclinedServing value is only set on startup of the host (we don't currently flip this back to unhealthy), so no persistence errors exist. 

To the second point, I think that a host not serving traffic at all should factor into our hostFailurePercentage to catch the issue from before where we had some errors, but most pods were crashlooping so we didn't catch this case. If some hosts are failing, it means we have to pick up the slack on other hosts, so it's "failing". WDYT?

PR #7711, opened 2025-05-05
Copy nexus link converter from sdk-go

rodrigozhou:
## What changed?

Copy nexus link converter from sdk-go

### DIFF
--- a/components/nexusoperations/link_converter.go
+++ b/components/nexusoperations/link_converter.go
@@ -1,0 +1,208 @@
+// The MIT License
+//
+// Copyright (c) 2024 Temporal Technologies Inc.  All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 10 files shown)


components/nexusoperations/link_converter.go
```diff
@@ -0,0 +1,207 @@
...
+			url.PathEscape(we.GetRunId()),
+		),
+	}
+
+	switch ref := we.GetReference().(type) {
+	case *commonpb.Link_WorkflowEvent_EventRef:
+		u.RawQuery = convertLinkWorkflowEventEventReferenceToURLQuery(ref.EventRef)
+	case *commonpb.Link_WorkflowEvent_RequestIdRef:
+		u.RawQuery = convertLinkWorkflowEventRequestIdReferenceToURLQuery(ref.RequestIdRef)
+	}
```
rodrigozhou, line 87: @bergundy I'm wondering if we should return an error here if we find an unknown link type. The invert function below returns an error if the Nexus link has an unknown reference type.

bergundy: I would error here, yes.

Just make sure that wherever we use this we are forwards compatible since old servers may be running with newer SDKs.

bergundy: I would just log when these conversions fail.

bergundy (approved):

PR #7714, opened 2025-05-06
Allow dual ES clusters in visibility config

rodrigozhou:
## What changed?

Allow dual ES clusters in visibility config

## Why?

Allow to set a different Elasticsearch cluster as secondary Elasticsearch visibility store.

## How did you test it?

Created two ES clusters as two docker containers, modified the server config with dual visibility enabled, and ran some sample workflow. Verified that visibility records are written to both ES clusters and that the server can read from both of them

## Potential risks

Adding search attributes might be not add to the secondary ES

### DIFF
--- a/common/config/persistence.go
+++ b/common/config/persistence.go
@@ -77,20 +77,6 @@ func (c *Persistence) Validate() error {
 				ErrPersistenceConfig,
 				c.SecondaryVisibilityStore)
 		}
-		if isPrimaryEs && isSecondaryEs {
-			// ElasticSearch config for visibilityStore and secondaryVisibilityStore must be the same except for
-			// `indices.visibility` config key and private fields - this is a restriction due to global ES client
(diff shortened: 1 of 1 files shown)


common/config/persistence.go
```diff
@@ -77,20 +77,6 @@ func (c *Persistence) Validate() error {
 				ErrPersistenceConfig,
 				c.SecondaryVisibilityStore)
 		}
-		if isPrimaryEs && isSecondaryEs {
-			// ElasticSearch config for visibilityStore and secondaryVisibilityStore must be the same except for
-			// `indices.visibility` config key and private fields - this is a restriction due to global ES client
```
jmbarzee, line 82: If we opt for only this option I think we need a better gate here. I worry that we are removing a protection

rodrigozhou: We don't really have a good way to gate this.

jmbarzee (approved):

PR #7715, opened 2025-05-06
versioning entity workflows: enabling auto-restart pt1

Shivs11:
## What changed?

- Worker Deployment entity workflow to now `CAN` after every pending update and signal has been processed.

## Why?

- So that one can avoid NDE errors when server rollbacks to the previous version. This is because CAN'ing effectively reduces the history size as well as the probability of an NDE error!

## How did you test it?

- Current suite of tests are passing locally.
- Updated unit tests as well.

## Potential risks

- NOTE: This is a breaking change and customers currently using our entity workflows will be affected with potential NDE's.


tests/worker_deployment_test.go
```diff
@@ -1603,7 +1603,7 @@ func (s *WorkerDeploymentSuite) verifyDescribeWorkerDeployment(
 	expectedResp *workflowservice.DescribeWorkerDeploymentResponse,
 ) {
 	// relaxed timestamp constraint since the tests make sync calls, which could theoretically take seconds.
-	maxDurationBetweenTimeStamps := 2 * time.Second
+	maxDurationBetweenTimeStamps := 5 * time.Second
```
Shivs11, line 1606: relaxing the timestamp checks a little more since now, after every update, we CAN and this is taking about a second or so. 

I intend on shipping out a change very soon that shall un-revert this change and pass in stricter expected timestamps from the tests itself. Right now, in most tests, the expected value after an operation is conducted is passed as `time.Now()` which is not accurate.

service/worker/workerdeployment/workflow.go
```diff
@@ -877,3 +914,13 @@ func (d *WorkflowRunner) updateMemo(ctx workflow.Context) error {
...
+
+func (d *WorkflowRunner) setStateChanged(ctx workflow.Context) {
+	if !d.stateChanged.value {
+		// Send the value on the channel without blocking.
+		workflow.Go(ctx, func(ctx workflow.Context) {
+			d.stateChanged.value = true
+			d.stateChanged.ch.Send(ctx, nil)
+		})
+	}
+}
```
Shivs11, line 926: Worried our entity workflow could get stuck if I have not implemented the `Send` on channels correctly. The main requirement is to not block for the sends that happen after a state has changed. This is why I have implemented them to be in their own go-routine.

My initial attempt was to not have this snippet of code in a go-routine, but I soon realized that this was a blocking call and was causing timeouts to happen in most tests. I think this works, but feel free to correct me if I am wrong.

cc - @dnr

service/worker/workerdeployment/workflow.go
```diff
@@ -57,28 +69,29 @@ func Workflow(ctx workflow.Context, unsafeMaxVersion func() int, args *deploymen
...
-	forceCAN := false
 
-	selector := workflow.NewSelector(ctx)
-	selector.AddReceive(forceCANSignalChannel, func(c workflow.ReceiveChannel, more bool) {
+	d.signalHandler.signalSelector.AddReceive(forceCANSignalChannel, func(c workflow.ReceiveChannel, more bool) {
+		d.signalHandler.processingSignals++
 		c.Receive(ctx, nil)
-		forceCAN = true
+		d.forceCAN = true
+		d.signalHandler.processingSignals--
```
ShahabT, line 77: let's put these lines in `defer` in case some one later `return`s from the middle.

ShahabT (approved):

PR #7716, opened 2025-05-06
Add subqueue to tdbg tq list-tasks

dnr:
## What changed?
Allow querying other subqueues with tdbg tq list-tasks.

## Why?
Debugging and other operational tasks.


Shivs11 (approved):
curious - how did you test this out? I see we have `task_queue_commands_test.go` under `tdbg` and was gonna suggest maybe putting this flag about there, but not sure if that's still the norm


approving it nevertheless since I don't wanna block + this is only for operational tasks

dnr:
> curious - how did you test this out? I see we have `task_queue_commands_test.go` under `tdbg` and was gonna suggest maybe putting this flag about there, but not sure if that's still the norm

I didn't see any tests for list-tasks, otherwise I would have added it. It seems simple and low risk enough that I don't think it needs an explicit test.

PR #7717, opened 2025-05-06
Spread matching partitions across nodes

dnr:
## What changed?
Add option to use LookupN instead of Lookup for matching partition routing.

## Why?
This can help spread partitions across available nodes better: instead of placing each partition independently, they're always placed on separate nodes, as long as there are enough nodes, up to the batch size.

## How did you test it?
added unit test, will use cicd to test with this setting

## Potential risks
If this dynamic config is changed all at once, it can cause a lot of task queue thrashing. Use the gradual change mechanism to roll it out to a live cluster.


client/clientfactory.go
```diff
@@ -217,12 +218,15 @@ func newServiceKeyResolver(resolver membership.ServiceResolver) *serviceKeyResol
...
-	if err != nil {
-		return "", err
+func (r *serviceKeyResolverImpl) Lookup(key string, n int) (string, error) {
+	hosts := r.resolver.LookupN(key, n+1)
+	if len(hosts) == 0 {
+		return "", membership.ErrInsufficientHosts
+	}
+	if n >= len(hosts) {
+		n %= len(hosts)
 	}
```
ShahabT, line 228: This part could lead to inconsistent hashing when on host goes down/up. Is the assumption that n is comfortably lower than # of host?

common/dynamicconfig/constants.go
```diff
@@ -1291,6 +1291,11 @@ second per poller by one physical queue manager`,
 		60*time.Second,
 		`Timeout for forwarded backlog task (requires new matcher)`,
 	)
+	MatchingSpreadRoutingBatchSize = NewGlobalIntSetting(
+		"matching.spreadRoutingBatchSize",
+		0,
+		`If non-zero, try to spread task queue partitions across matching nodes better, using the given batch size.`,
```
ShahabT, line 1297: It seems the assumption is the batch size will be set to a number that is comfortable lower than the number of matching pods, right? Should we mention that here?

ShahabT: Also should we mention the risk of changing this number after a cluster starts getting traffic?

common/tqid/task_queue_id.go
```diff
@@ -346,8 +347,22 @@ func (p *NormalPartition) Key() PartitionKey {
...
+	if batchSize == 0 {
+		return fmt.Sprintf("%s:%s:%d", p.NamespaceId(), p.RpcName(), p.TaskType()), 0
+	}
+	// We want to use LookupN to spread partitions across available nodes, but LookupN takes O(n)
+	// time and space, so we should limit the n that we pass to it. Reduce the partition id by some
+	// factor and move that factor into the name.
+	batch := p.partitionId / batchSize
+	if batch == 0 {
+		// For the first batch, omit the batch number as if batchSize==0 so that the root
+		// partition doesn't move if we turn on/off spreading.
```
rkannan82: How does batch = 0 imply root partition? Shouldn't we explicitly check if this is a root partition?

client/clientfactory.go
```diff
@@ -217,12 +218,15 @@ func newServiceKeyResolver(resolver membership.ServiceResolver) *serviceKeyResol
...
 }
 
-func (r *serviceKeyResolverImpl) Lookup(key string) (string, error) {
-	host, err := r.resolver.Lookup(key)
-	if err != nil {
-		return "", err
+func (r *serviceKeyResolverImpl) Lookup(key string, n int) (string, error) {
```
rkannan82: IIRC, the caller is passing the index for 'n'. But LookupN should be using the batch size and not the index. Did I misunderstand the logic?

service/matching/matching_engine.go
```diff
@@ -302,6 +302,16 @@ func (e *matchingEngineImpl) listenerKey() string {
...
 func (e *matchingEngineImpl) watchMembership() {
 	self := e.hostInfoProvider.HostInfo().Identity()
+	rc, ok := e.matchingRawClient.(matching.RoutingClient)
+	if !ok {
+		e.logger.Warn("watchMembership found non-routing matching client")
+		return // this should only happen in unit tests
+	}
+	ownedByOther := func(p tqid.Partition) bool {
+		addr, err := rc.Route(p)
+		// don't take action on lookup error
```
rkannan82, line 312: This changes the current behavior, right? Can you clarify the rationale.

client/clientfactory.go
dnr: The key has the batch index (which batch), so the batch size is implicitly part of the key. `n` here is the index within the batch.

common/tqid/task_queue_id.go
dnr: "root partition" is "partitionId == 0", so batch must be 0

service/matching/matching_engine.go
dnr: It's not a change. In the old code, if serviceResolver.Lookup returned an error then no action was taken.

common/tqid/task_queue_id.go
rkannan82: OK. Please note that "root partition id is 0" in the comment.

client/clientfactory.go
rkannan82: Ah, I get it now. This is the optimization you had noted in task_queue_id.go.
--> LookupN takes O(n) time and space, so we should limit the n that we pass to it

It suffices to pass index + 1 to LookupN, because you will anyway discard any host that is > the index.

common/tqid/task_queue_id_test.go
```diff
@@ -227,6 +227,42 @@ func TestInvalidRpcNames(t *testing.T) {
 	}
 }
 
+func TestBatching(t *testing.T) {
```
rkannan82: [Optional] Suggest splitting this into 3 tests. TestWithoutBatching, TestBatching, TestStableKeyForRootPartition.

client/matching/metric_client.go
```diff
@@ -200,3 +201,14 @@ func (c *metricClient) finishMetricsRecording(
 	}
 	metrics.ClientLatency.With(metricsHandler).Record(time.Since(startTime))
 }
+
+func (c *metricClient) Route(p tqid.Partition) (string, error) {
+	// Ideally we wouldn't do a type-check here and require c.client to have
+	// Route, but it would require changing too many types all over the place.
```
rkannan82, line 207: I didn't understand why this requires changing many types. If you can point to an example, that will be helpful.

client/clientfactory.go
rkannan82: [Optional] Just a personal preference ... name the 2nd param as index instead of n and document what this is actually referring to. This way the semantics are clear.  Otherwise, it is not obvious why we do Lookup(n+1)

client/clientfactory.go
```diff
@@ -217,12 +218,15 @@ func newServiceKeyResolver(resolver membership.ServiceResolver) *serviceKeyResol
...
+	hosts := r.resolver.LookupN(key, n+1)
+	if len(hosts) == 0 {
+		return "", membership.ErrInsufficientHosts
+	}
+	if n >= len(hosts) {
```
rkannan82: Since we have some logic now, can you assess if it is worth adding a unit test. Main concern is now the non batching case is also going through all of this new code. I see matching_engine_test has similar mock resolver setup.

client/matching/metric_client.go
dnr: I didn't commit the code when I tried it, but it was starting to infect everything: it requires changing `matchingservice.MatchingServiceClient` to `matching.RoutingClient` in many packages (plus making RoutingClient a superset of MatchingServiceClient instead of just containing Route). It's not necessarily a bad change, but it would be pretty noisy, and it's not necessary since we can just do a type-check here.

Note that this type check is not in the hot path: the hot path just goes from getClientForTaskQueuePartition to Route without any type checks. This is for matchingEngineImpl.watchMembership to work, which is a background process.

client/clientfactory.go
dnr: Sounds good, I added more comments

client/clientfactory.go
dnr: The logic here is just a `%=` which I don't think needs a test. The LookupN change really depends on ringpop. We can verify it there, I'll add a test to membership.

I think the interesting new code is tqid.NormalPartition.RoutingKey, which does split out the batchsSize == 0 case, and does have a few tests.

common/tqid/task_queue_id_test.go
dnr: sure

common/dynamicconfig/constants.go
dnr: I did some simulations and 4 or 8 seem like a good balance without causing too much excessive movement. I also added a note to the dynamic config.

PR #7718, opened 2025-05-06
[Frontend] Log slow gRPC requests

lina-temporal:
## What changed?
A new interceptor has been written and added to frontend's FX chain that logs slow gRPC requests. Requests are logged with their elapsed time, and more importantly, their extracted request IDs/workflow tags, so on-call can quickly determine the problematic IDs.

This is what the log line might look like:
```json
{
  "level": "warn",
  "ts": "2025-05-06T15:34:35.600-0700",
  "msg": "Slow gRPC call for 'temporal.api.workflowservice.v1.WorkflowService/DescribeWorkflowExecution', took 6.000857875s",
  "wf-id": "fc48187c-3df0-4c6e-b898-027ffe3e9066",
  "wf-run-id": "0196a7bc-73ec-7f1b-890d-a0bbbf63ac21",
  "logging-call-at": "/Users/lina/code/temporal/common/rpc/interceptor/slow_request_logger.go:59"
}
```

## Why?
Because these request IDs are often deep within a protobuf request body, they aren't visible through Envoy/Chronicle logs, so although we can often see that a customer is experiencing request-reply latency, we have little else to follow-up on without asking them for more information.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
We could end up with spammy log lines if we have lots of methods which regularly complete over the threshold, that aren't on the ignore list. I based the ignore and threshold on the frontend API dashboards across a few cells.


common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/rpc/interceptor/logtags"
+	"go.temporal.io/server/common/tasktoken"
+	"google.golang.org/grpc"
+)
+
+var (
+	// Requests over this threshold will be logged.
+	SlowRequestThreshold = 5 * time.Second
```
dnr, line 17: should be dynamic config? (also to enable/disable)

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,80 @@
...
+func (i *SlowRequestLoggerInterceptor) Intercept(
+	ctx context.Context,
+	request interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
+	startTime := time.Now()
+
+	defer func() {
+		elapsed := time.Now().Sub(startTime)
```
dnr, line 48:
```suggestion
		elapsed := time.Since(startTime)
```

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, substr := range ignoredMethodSubstrings {
+		if strings.Contains(method, substr) {
+			return
+		}
+	}
+
+	wfTags := i.workflowTags.Extract(request, method)
+	i.logger.Warn(
+		fmt.Sprintf(
+			"Slow gRPC call for '%s', took %s",
```
dnr, line 74: should probably turn these into tags instead of printf?

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,80 @@
...
+	"go.temporal.io/server/common/tasktoken"
+	"google.golang.org/grpc"
+)
+
+var (
+	// Requests over this threshold will be logged.
+	SlowRequestThreshold = 5 * time.Second
+
+	// Certain types of methods are ignored as a rule.
+	ignoredMethodSubstrings = []string{"Poll"}
```
dnr, line 20: GetWorkflowExecutionHistory can be a long poll if WaitNewEvent is set

common/rpc/interceptor/slow_request_logger.go
lina-temporal: Sure, I'll update this to be a dynamic config.

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (i *SlowRequestLoggerInterceptor) logSlowRequest(
+	request interface{},
+	info *grpc.UnaryServerInfo,
+	elapsed time.Duration,
+) {
+	method := info.FullMethod
+
+	// Certain methods aren't useful logged.
+	for _, substr := range ignoredMethodSubstrings {
```
prathyushpv, line 63: We can use a set instead of a list to make it a bit more efficient.

common/rpc/interceptor/slow_request_logger.go
prathyushpv: We do this in rate limiter interceptor to find Poll GetWorkflowExecution requests: https://github.com/temporalio/temporal/pull/7544/files

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,74 @@
...
+
+type SlowRequestLoggerInterceptor struct {
+	logger       log.Logger
+	workflowTags *logtags.WorkflowTags
+	dc           *dynamicconfig.Collection
+}
+
+func NewSlowRequestLoggerInterceptor(
+	logger log.Logger,
+	dc *dynamicconfig.Collection,
```
dnr, line 27: don't take the whole Collection here, take a dynamicconfig.DurationPropertyFn (just `func () time.Duration`). that makes it easier to test and also more efficient (by a tiny bit).

you can use dynamicconfig.GetDurationPropertyFn to create a static one for tests

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+func (i *SlowRequestLoggerInterceptor) Intercept(
+	ctx context.Context,
+	request interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
+	threshold := dynamicconfig.SlowRequestLoggingThreshold.Get(i.dc)()
```
dnr, line 42: Ideally we would do the exclusion before the call to get the dynamic config, to avoid that call if we're excluding anyway

common/rpc/interceptor/slow_request_logger.go
dnr: Just a thought: consider putting the answer to this question (can this be a long poll?) here:
https://github.com/temporalio/temporal/blob/main/common/api/metadata.go#L15-L20

lina-temporal: > We can use a set instead of a list to make it a bit more efficient.

If we did this, then we'd have to have a set of exact method fullnames instead of substrings. I think the API metadata is probably the cleanest, I'll update that.

common/api/metadata.go
```diff
@@ -12,11 +12,16 @@ type (
...
 
+	// Describes if the method supports long-polled requests.
+	Polling int32
+
 	MethodMetadata struct {
 		// Describes the scope of a method (whole cluster or inividual namespace).
 		Scope Scope
 		// Describes what level of access is needed for a method (advisory).
 		Access Access
+		// Describes if long polling is supportede by the method.
```
dnr, line 23:
```suggestion
		// Describes if long polling is supported by the method.
```

common/rpc/interceptor/slow_request_logger.go
```diff
@@ -0,0 +1,69 @@
...
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/rpc/interceptor/logtags"
+	"go.temporal.io/server/common/tasktoken"
+	"google.golang.org/grpc"
+)
+
+// Certain types of methods are ignored as a rule.
+var ignoredMethodSubstrings = []string{"Poll", "GetWorkflowExecutionHistory"}
+
```
dnr, line 18: unused now?
```suggestion
```

dnr (approved):
Nice, thanks!

PR #7723, opened 2025-05-07
Remove cap for dynamic config callback pool

dnr:
## What changed?
Set MaxWorkers for dynamic config subscription callback to 1e9 so there's no effective limit.

## Why?
A batch of subscription callbacks are enqueued (though not called) while holding a lock. If a callback somehow makes a call back into dynamic config subscriptions, then it would have to wait for the batch to finish enqueueing. If there's more than MaxWorkers of those, that would cause a deadlock. We don't need to impose a limit on workers here.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -18,7 +18,7 @@ var (
 		"dynamicconfig.subscriptionCallback",
 		subscriptionCallbackSettings{
 			MinWorkers:   10,
-			MaxWorkers:   100,
+			MaxWorkers:   1e9, // effectively unlimited
 			TargetDelay:  10 * time.Millisecond,
 			ShrinkFactor: 1000, // 10 seconds
 		},


common/dynamicconfig/constants.go
```diff
@@ -18,7 +18,7 @@ var (
 		"dynamicconfig.subscriptionCallback",
 		subscriptionCallbackSettings{
 			MinWorkers:   10,
-			MaxWorkers:   100,
+			MaxWorkers:   1e9, // effectively unlimited
```
prathyushpv, line 21: The why don't we remove this config altogether?

If we don't specify a maximum number of threads, is there value in using AdaptivePool? Can't we just create a new goroutine each time?

dnr: Mainly: because the pool currently requires some max and I wanted to make a minimal change right now.

There's still some value in AdaptivePool: because of the timeout, in cases where callbacks are fast and don't call back into dynamic config (which is the common case), it won't create that many goroutines. Only when they call back into dynamic config and get stuck, then it'll create more goroutines than it otherwise would.

But yeah, you could argue the whole thing is overkill and we could just create new goroutines. It was an attempt to reduce the impact if we need to call something like 1M callbacks at once.

prathyushpv (approved):

PR #7726, opened 2025-05-07
Validate links in completion callbacks

rodrigozhou:
## What changed?
Validate links in completion callbacks

## Why?
Missed validation of links in #7697 

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -5217,6 +5217,9 @@ func (wh *WorkflowHandler) validateWorkflowCompletionCallbacks(
 	}
 
 	for _, callback := range callbacks {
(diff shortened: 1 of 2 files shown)


service/frontend/workflow_handler.go
```diff
@@ -5217,6 +5217,9 @@ func (wh *WorkflowHandler) validateWorkflowCompletionCallbacks(
 	}
 
 	for _, callback := range callbacks {
+		if err := wh.validateLinks(ns, callback.GetLinks()); err != nil {
```
bergundy, line 5220: Now that we support links in both callbacks and outside of them, the validation should limit the total across the two places.

bergundy (approved):
Thanks!

PR #7729, opened 2025-05-08
[CHASM] ExecuteSideEffectTask on Tree

lina-temporal:
## What changed?
- Added ExecuteSideEffectTask to Node, as with ExecutePureTask.

## Why?
- Facilitates the remaining physical task executors.

## How did you test it?
- [ ] built
- [x] added new unit test(s)


lina-temporal:
Once #7855 is approved and lands in `main`, I'll merge main here and that should cut down on the boilerplate noise.

service/history/timer_queue_active_task_executor.go
```diff
@@ -925,6 +929,62 @@ func (t *timerQueueActiveTaskExecutor) processActivityWorkflowRules(
...
+func (t *timerQueueActiveTaskExecutor) executeChasmSideEffectTimerTask(
+	ctx context.Context,
+	task *tasks.ChasmTask,
+) error {
+	ctx, cancel := context.WithTimeout(ctx, taskTimeout)
+	defer cancel()
+
+	if !queues.IsTimeExpired(task, t.Now(), task.GetVisibilityTime()) {
+		return nil
+	}
```
yycptt: This is not expected to happen and we can't drop the task in the case.
Probably add an assertion and return an retryable (unavailable for example) error.

service/history/timer_queue_active_task_executor.go
```diff
@@ -925,6 +929,62 @@ func (t *timerQueueActiveTaskExecutor) processActivityWorkflowRules(
...
+
+	wfCtx, release, err := getWorkflowExecutionContextForTask(ctx, t.shardContext, t.cache, task)
+	if err != nil {
+		return err
+	}
+	defer func() { release(err) }()
+
+	ms, err := loadMutableStateForTimerTask(ctx, t.shardContext, wfCtx, task, t.metricsHandler, t.logger)
```
yycptt, line 945: Can't use `common.FirstEventID` in `GetTimerTaskEventID` because chasm runs may not have events at all.

service/history/timer_queue_active_task_executor.go
```diff
@@ -925,6 +929,62 @@ func (t *timerQueueActiveTaskExecutor) processActivityWorkflowRules(
...
+	if err != nil {
+		return err
+	}
+	if ms == nil {
+		return nil
+	}
```
yycptt: also should never happen once the eventID check is fixed (bypassed essentially).

service/history/timer_queue_active_task_executor.go
yycptt: Actually I don't think we can use this method since we will unlock the mutable state after doing those checks.

Both task generation and zombie check need to be done inside the ComponentRef's validate func, while holding the lock (same as what statemachine_environment is doing today.)
Task (ID) clock check, which doesn't depend on mutable state, can still be here. `outboundQueueActiveTaskExecutor.Execute()` is a good example. 

(Probably construct a validation function here and pass into node.ExecuteSideEffectTask() )

lina-temporal: Done, moved to the validation callback.

service/history/timer_queue_standby_task_executor.go
```diff
@@ -158,6 +160,50 @@ func (t *timerQueueStandbyTaskExecutor) executeChasmPureTimerTask(
...
+
+	return t.processTimer(
+		ctx,
+		task,
+		actionFn,
+		getStandbyPostActionFn(
+			task,
+			t.getCurrentTime,
+			t.config.StandbyTaskMissingEventsDiscardDelay(task.GetType()),
+			t.checkWorkflowStillExistOnSourceBeforeDiscard,
```
yycptt, line 203: hmm can't really use this or at least need to change it's implementation. It's calls DescribeWorkflowExecution() which is specific for workflow. 

Can you add a TODO here and create an issue to track this? No super urgent but need to get it fixed before we run chasm in production

service/history/timer_queue_standby_task_executor.go
```diff
@@ -158,6 +160,50 @@ func (t *timerQueueStandbyTaskExecutor) executeChasmPureTimerTask(
...
+		if err == nil && taskInstance != nil {
+			// If a taskInstance is returned, the task is still valid, and we should keep
+			// it around.
+			return &struct{}{}, nil
+		}
+
+		return nil, err
+	}
+
+	return t.processTimer(
```
yycptt, line 194: Still need to get GetTimerTaskEventID fixed since t.processTimer() is internally using `loadMutableStateForTimerTask`.

and looks like loadMutableStateForTimerTask is still used by active logic as well.

service/history/timer_queue_active_task_executor.go
```diff
@@ -927,6 +929,69 @@ func (t *timerQueueActiveTaskExecutor) processActivityWorkflowRules(
...
+		tgClock := backend.GetExecutionInfo().TaskGenerationShardClockTimestamp
+		if tgClock != 0 && taskID != 0 && taskID < tgClock {
+			return consts.ErrStaleReference
+		}
+
+		return nil
+	}
+
+	engineCtx := chasm.NewEngineContext(ctx, t.chasmEngine)
+	return tree.ExecuteSideEffectTask(
```
yycptt, line 986: nit: hmm is it possible to not make that a method on the chasm Node? It feels very dangerous and error prone to invoke a method on the chasmNode after the mutable state lock is released.

yycptt: I guess maybe the idea is to minimize the things exported from chasm package. If so, let's add a big WARNING on that method saying the method implementation should not access anything through the node. (and ideally pass in pathencoder from here as well, just to make the boundary very clear)

chasm/tree.go
```diff
@@ -2025,3 +2028,155 @@ func (n *Node) ExecutePureTask(baseCtx context.Context, taskInstance any) error
...
+	registrableTask, ok := registry.task(taskType)
+	if !ok {
+		return nil, serviceerror.NewInternalf("unknown task type '%s'", taskType)
+	}
+
+	if registrableTask.isPureTask {
+		return nil, serviceerror.NewInternalf("ValidateSideEffectTask called on a Pure task '%s'", taskType)
+	}
+
+	taskValue, err := deserializeTask(registrableTask, taskInfo.Data)
```
yycptt, line 2056: Should ideally cache the deserialized task I think. maybe add a TODO here.

service/history/timer_queue_active_task_executor.go
lina-temporal: > I guess maybe the idea is to minimize the things exported from chasm package.

Yeah^. I'll add a warning and take in a pathEncoder.

lina-temporal: actually, going to punt on changing pathencoder, since there's already a TODO to defer path deserialization until Ref is accessed.

service/history/timer_queue_standby_task_executor.go
lina-temporal: > Still need to get GetTimerTaskEventID fixed since t.processTimer() is internally using loadMutableStateForTimerTask.

Will fix. 

> and looks like loadMutableStateForTimerTask is still used by active logic as well.

Oops, will fix, thanks.

service/history/timer_queue_standby_task_executor.go
lina-temporal: > hmm can't really use this or at least need to change it's implementation. It's calls DescribeWorkflowExecution() which is specific for workflow.

Given that CHASM entity keys map cleanly back to workflow keys and are still written to DB as workflow executions, what prevents `DescribeWorkflowExecution` from working here?

service/history/timer_queue_standby_task_executor.go
lina-temporal: I updated loadMutableStateForTask to short-circuit its event ID comparison check for tasks without an event ID - seemed like the lowest footprint way to do this.

yycptt (approved):

service/history/timer_queue_standby_task_executor.go
yycptt: It's a workflow specific API and should return notFound if the requested ID is not a Workflow. Or maybe a different error type say InvalidArgument, still in that case needs to at least update the error handling logic in `checkWorkflowStillExistOnSourceBeforeDiscard`

lina-temporal: Gotcha, adding a TODO and a task.

PR #7732, opened 2025-05-08
Include transient and speculative workflow task events in the history

alexshtin:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/service/history/api/get_history_util.go
+++ b/service/history/api/get_history_util.go
@@ -223,6 +223,7 @@ func GetHistory(
 	return executionHistory, nextPageToken, nil
 }
 
+// TODO: need transient support too
 func GetHistoryReverse(
 	ctx context.Context,
 	shardContext historyi.ShardContext,
--- a/service/history/api/getworkflowexecutionhistory/api.go
+++ b/service/history/api/getworkflowexecutionhistory/api.go
@@ -44,22 +44,23 @@ func Invoke(
 
 	isCloseEventOnly := request.Request.GetHistoryEventFilterType() == enumspb.HISTORY_EVENT_FILTER_TYPE_CLOSE_EVENT
 
-	// this function returns the following 7 things,
+	// this function returns the following 8 things:
 	// 1. the current branch token (to use to retrieve history events)
 	// 2. the workflow run ID
 	// 3. the last first event ID (the event ID of the last batch of events in the history)
 	// 4. the last first event transaction id
 	// 5. the next event ID
 	// 6. whether the workflow is running
-	// 7. error if any
-	queryHistory := func(
+	// 7. transient/speculative workflow task info
+	// 8. error if any
+	queryMutableState := func(
 		namespaceUUID namespace.ID,
 		execution *commonpb.WorkflowExecution,
 		expectedNextEventID int64,
 		currentBranchToken []byte,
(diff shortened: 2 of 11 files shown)


service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -44,22 +44,23 @@ func Invoke(
...
+	// this function returns the following 8 things:
 	// 1. the current branch token (to use to retrieve history events)
 	// 2. the workflow run ID
 	// 3. the last first event ID (the event ID of the last batch of events in the history)
 	// 4. the last first event transaction id
 	// 5. the next event ID
 	// 6. whether the workflow is running
-	// 7. error if any
-	queryHistory := func(
+	// 7. transient/speculative workflow task info
```
yiminc, line 54: 😵 , could use a struct

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -180,6 +183,12 @@ func Invoke(
 		continuationToken.PersistenceToken = nil
 	}
 
+	// TODO: remove this after it start to support it
+	if clientName, _ := headers.GetClientNameAndVersion(ctx); clientName == headers.ClientNameCLI || clientName == headers.ClientNameUI {
+		// Pretend that there is transient or speculative WFT if client is CLI or UI because they don't support it.
```
yiminc, line 188: Do you miss a NO in the comment?

service/history/api/get_history_util.go
```diff
@@ -223,6 +223,7 @@ func GetHistory(
 	return executionHistory, nextPageToken, nil
 }
 
+// TODO: need transient support too
```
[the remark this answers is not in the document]
alexshtin: I added a better comment there.

service/history/api/getworkflowexecutionhistory/api.go
alexshtin: It is a func internal to another func. The entire `GetHistory` set of APIs need to be cleaned up and refactored. I don't want to make it more complex here.

service/history/api/get_history_util.go
```diff
@@ -93,7 +93,13 @@ func GetRawHistory(
 	metricsHandler := interceptor.GetMetricsHandlerFromContext(ctx, shardContext.GetLogger()).WithTags(metrics.OperationTag(metrics.HistoryGetHistoryScope))
 	metrics.HistorySize.With(metricsHandler).Record(int64(size))
 
-	if len(nextToken) == 0 && transientWorkflowTaskInfo != nil {
+	// If
+	//    there are no more events in DB (i.e., page token is empty),
+	//    and transient/speculative WFT events are present (i.e., such WFT is in MS),
+	//    and WF is still running (i.e., last event is not completion event),
```
yycptt, line 99: is this for handling the case where transient/speculative workflow task is only scheduled and they are still in ms even when workflow is closed (terminated)?

alexshtin: yes, but not only. History is loaded outside of the WF lock, and during testing I got a case when race between normal WF completion and polling for WF result returned speculative WFT events which were there before completion, but got attached after `WorkflowCompleted` event.

PR #7736, opened 2025-05-08
Version Activation/Deactivation timestamps + sync information with version local state

Shivs11:
## What changed?
- Added two new timestamps to signify when a version first/last receives traffic in the WorkerDeploymentVersionInfo and WorkerDeploymentInfo.
- Added `routing_update_time`, `current_since_time`, `ramping_since_time`, `DrainageInfo` as part of VersionSummary.

## Why?
- Adding the two new timestamps shall allow operators to have more visibility into when a version gets activated/deactivated.
- The reason fields from VersionLocalState were added to the VersionSummary is because the UI, right now, is querying a bunch of different places to get information about a specific version. This might be annoying and the aim of this PR is to make VersionSummary the source of truth for information related to a specific version. A nice cherry on top is that this shall also reduce the number of API calls to get information about a version.

## How did you test it?
- [x] built
- [ ] added new unit test(s)


tests/worker_deployment_test.go
```diff
@@ -373,65 +372,6 @@ func (s *WorkerDeploymentSuite) TestDescribeWorkerDeployment_MultipleVersions_So
 	}, time.Second*10, time.Millisecond*1000)
 }
 
-func (s *WorkerDeploymentSuite) TestDescribeWorkerDeployment_SetCurrentVersion() {
```
Shivs11, line 376: just moved this test along with other setCurrent tests.

tests/worker_deployment_test.go
```diff
@@ -977,27 +1088,55 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_Batching()
 	s.ensureCreateVersionInDeployment(tv)
 	s.ensureCreateVersionWithExpectedTaskQueues(ctx, tv, taskQueues)
 
-	// verify that all the registered task-queues have "__unversioned__" as their current version
```
Shivs11, line 980: was a wrong comment so i fixed it

tests/worker_deployment_test.go
Shivs11, line 1: no new tests have been added to this file - I have mainly refactored `verifyDescribeWorkerDeployment` to now include checks for `VersionSummaries` and these checks have been made stricter than what we had previously. More work can be done to further refactor this file, and that shall come in a different PR.

go.mod
```diff
@@ -177,3 +177,5 @@ require (
 	modernc.org/strutil v1.2.1 // indirect
 	modernc.org/token v1.1.0 // indirect
 )
+
+replace go.temporal.io/api => github.com/temporalio/api-go v1.48.1-0.20250508163833-6d907c430e54
```
carlydf, line 181: revert before merging

service/worker/workerdeployment/workflow.go
```diff
@@ -105,6 +102,25 @@ func (d *WorkflowRunner) listenToSignals(ctx workflow.Context) {
 	d.signalsCompleted = true
 }
 
+func (d *WorkflowRunner) updateVersionSummary(summary *deploymentspb.WorkerDeploymentVersionSummary) {
```
carlydf, line 105:
```suggestion
func (d *WorkflowRunner) updateVersionSummary(newSummary *deploymentspb.WorkerDeploymentVersionSummary) {
```

nit: I feel like calling this parameter something with "new" or "source" in it makes it clear that the input is not what is being updated. Just reading the function signature as is makes me feel like the summary that is passed in is what will be updated.

service/worker/workerdeployment/workflow.go
```diff
@@ -105,6 +102,25 @@ func (d *WorkflowRunner) listenToSignals(ctx workflow.Context) {
...
+	if _, ok := d.State.Versions[summary.GetVersion()]; !ok {
+		d.logger.Error("received summary for a non-existing version, ignoring it", "version", summary.GetVersion())
+		return
+	}
+
+	// Preserve create_time and first_activation_time if they exist in current summary. This is to ensure that if the version
```
carlydf, line 111: Ideally, the version should never update its own first activation time incorrectly, so the summary that arrives here will always be correct. This if clause makes it possible for the Version Local State to be out of sync with the VersionSummary, which is not what we want

service/worker/workerdeployment/workflow.go
```diff
@@ -764,6 +785,22 @@ func (d *WorkflowRunner) syncVersion(ctx workflow.Context, targetVersion string,
...
+	if err == nil {
+		summary := &deploymentspb.WorkerDeploymentVersionSummary{
+			Version:           targetVersion,
+			RoutingUpdateTime: versionUpdateArgs.RoutingUpdateTime,
+			CurrentSinceTime:  versionUpdateArgs.CurrentSinceTime,
+			RampingSinceTime:  versionUpdateArgs.RampingSinceTime,
+		}
+		if activated {
+			summary.FirstActivationTime = versionUpdateArgs.RoutingUpdateTime
+		} else {
```
carlydf, line 799: does !activated always mean deactivated?

service/worker/workerdeployment/workflow.go
Shivs11: > so the summary that arrives here will always be correct
the summary can arrive here in one of two ways: from the version workflow, when it calls `syncState` and from the deployment workflow, when it calls `syncVersion`.

In the long run, we want the deployment workflow to be the source of truth for all things related to these timestamps (hence, we have gotten these timestamps inside the `WorkerDeploymentInfo`). What's happening here is that during a `setCurrent` call (for example), the deployment workflow would first reach here and initialize this value with the appropriate `RoutingUpdateTime`. Eventually, inside the version workflow, it's value would also be updated with the same `RoutingUpdateTime`.

Having said that, I think I agree with you that it's best we prevent a scenario where two values of the same field are actually different in our version workflow and deployment workflow. What I have done is add another helper function, called `syncVersionSummaryFromVersionWorkflow` which replaces the summary for a version in the local state of the workflow. This would ensure that if things were to go out of sync, these calls made from the version workflow would keep things in sync.

One could argue you don't need these two functions and can place separate lines of code inside `syncState` to update fields of the version summary, but I think this is much more readable interface to have and makes things easier for a new reader.

Let me know what you think and thanks for pointing this out!

service/worker/workerdeployment/workflow.go
Shivs11: yes - this function is called with `activated=false` in the following cases: [ note: deactivation is when we remove current/ramping for a version ]

1. When we are unsetting the ramp (removing the ramp, so it's deactivation)
2. Telling previous ramping version it's no longer ramping (removing the ramp, so it's deactivation)
3. Telling previous current version it's no longer current (removing the current, so it's deactivation)

Shivs11: note - this is not called for unversioned syncs since we don't have an unversioned "version"

service/worker/workerdeployment/workflow.go
Shivs11: I think with the latest changes, this concern of yours might have been mitigated - lmk if you still feel the same way and I can change things around 😃

service/worker/workerdeployment/workflow.go
```diff
@@ -94,6 +91,36 @@ func (d *WorkflowRunner) listenToSignals(ctx workflow.Context) {
...
+
+func (d *WorkflowRunner) updateVersionSummary(summary *deploymentspb.WorkerDeploymentVersionSummary) {
+	if _, ok := d.State.Versions[summary.GetVersion()]; !ok {
+		d.logger.Error("received summary for a non-existing version, ignoring it", "version", summary.GetVersion())
+		return
+	}
+
+	// Preserve create_time and first_activation_time if they exist in current summary. This is to ensure that if the version
+	// had already been activated before, we don't override the first activation time by setting it to a wrong value.
+	if existingSummary := d.State.Versions[summary.GetVersion()]; existingSummary != nil && existingSummary.GetCreateTime() != nil {
```
carlydf, line 113:
```suggestion
	if existingSummary := d.State.Versions[summary.GetVersion()]; existingSummary.GetCreateTime() != nil {
```

PR #7737, opened 2025-05-08
Enable transition history in dev env and tests

yycptt:
## What changed?
Enable transition history in dev env and tests

## Why?
- Exercise transition history and state-based replication path more often
- Make chasm development and testing easier since it relies on transition history

## How did you test it?
- [x] built
- [ ] added new unit test(s)


xwduan (approved):

hai719 (approved):

components/nexusoperations/completion.go
```diff
@@ -177,6 +177,12 @@ func CompletionHandler(
 	if errors.As(err, new(*serviceerror.NotFound)) && isRetryableNotFoundErr && ref.WorkflowKey.RunID != "" {
 		// Try again without a run ID in case the original run was reset.
 		ref.WorkflowKey.RunID = ""
+		// VersionedTransition is for a specific run. After reset, the TransitionCount will
+		// start from 1 again. Reset the TransitionCount to 0 here to fallback to old ref
+		// validation logic.
+		ref.StateMachineRef.MutableStateVersionedTransition.TransitionCount = 0
+		ref.StateMachineRef.MachineInitialVersionedTransition.TransitionCount = 0
+		ref.StateMachineRef.MachineLastUpdateVersionedTransition.TransitionCount = 0
```
yycptt, line 185: cc @bergundy

bergundy: Thanks!

Was this caught with test coverage?

Also something to consider finding a good abstraction for in CHASM.

yycptt: yup, `TestNexusOperationAsyncCompletionAfterReset` starts to fail after I enabled TransitionHistory. 

Yeah I will need to think about how to handle it in CHASM world.

PR #7739, opened 2025-05-08
Abstract Catchup Workflow Parameters for Broader Applicability

zigeH:
## What changed?
This PR refactors the WaitCatchup workflow by generalizing its input parameters. The goal is to make the workflow more flexible and reusable across different use cases where similar catch-up logic is needed. e.g. catchup process in passive migration.

## Why?
Screenshot 2025-05-08 at 12 39 23 PM
In the previous setup, we ran catchup on passive replica A after T2 to ensure that cluster C could catch up to replica B’s (the current active) last AckID on A, guaranteeing that C wouldn't miss any data present on A. However, this logic was previously hardcoded to use B the active as catchup target. This PR generalizes the input, allowing the same logic to be reused during migration scenarios—enabling the source passive replica to catch up to the new replica’s ack level when the migration passive is in progress.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
parameter name change. no potential risk


pglass (approved):

service/worker/migration/activities.go
```diff
@@ -93,7 +93,7 @@ type (
 	}
 
 	waitCatchupRequest struct {
-		ActiveCluster string
+		TargetCluster string
 		RemoteCluster string
 		Namespace     string
 	}
```
zigeH, line 99: remoteCluster- > catchupCluster

yux0 (approved):

PR #7740, opened 2025-05-09
move rate limiter to task-queue-partition-manager

Shivs11:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/service/matching/matcher_test.go
+++ b/service/matching/matcher_test.go
@@ -22,6 +22,7 @@ import (
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/payloads"
+	"go.temporal.io/server/common/quotas"
 	"go.temporal.io/server/common/tqid"
 	"go.uber.org/mock/gomock"
 	"google.golang.org/grpc"
@@ -50,6 +51,28 @@ func TestMatcherSuite(t *testing.T) {
 	suite.Run(t, new(MatcherTestSuite))
 }
 
+// attachRateLimiter attaches a rate limiter to the matcher the same way as it is done in task queue partition manager.
+func (t *MatcherTestSuite) attachRateLimiter(taskQueueConfig *taskQueueConfig) quotas.RateLimiter {
+	dynamicRateBurst := quotas.NewMutableRateBurst(
+		defaultTaskDispatchRPS,
+		int(defaultTaskDispatchRPS),
+	)
+	dynamicRateLimiter := quotas.NewDynamicRateLimiter(
+		dynamicRateBurst,
+		defaultTaskDispatchRPSTTL,
+	)
+	limiter := quotas.NewMultiRateLimiter([]quotas.RateLimiter{
+		dynamicRateLimiter,
+		quotas.NewDefaultOutgoingRateLimiter(
+			taskQueueConfig.AdminNamespaceTaskQueueToPartitionDispatchRate,
+		),
+		quotas.NewDefaultOutgoingRateLimiter(
+			taskQueueConfig.AdminNamespaceToPartitionDispatchRate,
+		),
+	})
+	return limiter
+}
+
 func (t *MatcherTestSuite) SetupTest() {
 	t.controller = gomock.NewController(t.T())
 	t.client = matchingservicemock.NewMockMatchingServiceClient(t.controller)
@@ -71,11 +94,11 @@ func (t *MatcherTestSuite) SetupTest() {
 	t.childConfig = tlCfg
 	t.fwdr, err = newForwarder(&t.childConfig.forwarderConfig, t.queue, t.client)
 	t.Assert().NoError(err)
-	t.childMatcher = newTaskMatcher(tlCfg, t.fwdr, metrics.NoopMetricsHandler)
+	t.childMatcher = newTaskMatcher(tlCfg, t.fwdr, metrics.NoopMetricsHandler, t.attachRateLimiter(tlCfg))
 	t.childMatcher.Start()
 
 	t.rootConfig = newTaskQueueConfig(prtn.TaskQueue(), cfg, "test-namespace")
-	t.rootMatcher = newTaskMatcher(t.rootConfig, nil, metrics.NoopMetricsHandler)
+	t.rootMatcher = newTaskMatcher(t.rootConfig, nil, metrics.NoopMetricsHandler, t.attachRateLimiter(t.rootConfig))
 	t.rootMatcher.Start()
 }
 
--- a/service/matching/matching_engine_test.go
+++ b/service/matching/matching_engine_test.go
@@ -992,19 +992,9 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 	s.taskManager.getQueueManagerByKey(dbq).rangeID = initialRangeID
 
 	mgr := s.newPartitionManager(dbq.partition, s.matchingEngine.config)
-	mgrImpl, ok := mgr.(*taskQueuePartitionManagerImpl).defaultQueue.(*physicalTaskQueueManagerImpl)
+	_, ok := mgr.(*taskQueuePartitionManagerImpl).defaultQueue.(*physicalTaskQueueManagerImpl)
 	s.True(ok)
-	mgrImpl.oldMatcher.rateLimiter = quotas.NewRateLimiter(
-		defaultTaskDispatchRPS,
-		defaultTaskDispatchRPS,
-	)
-	mgrImpl.oldMatcher.dynamicRateBurst = &dynamicRateBurstWrapper{
-		MutableRateBurst: quotas.NewMutableRateBurst(
-			defaultTaskDispatchRPS,
-			defaultTaskDispatchRPS,
-		),
-		RateLimiterImpl: mgrImpl.oldMatcher.rateLimiter.(*quotas.RateLimiterImpl),
-	}
+
 	s.matchingEngine.updateTaskQueue(dbq.partition, mgr)
 	mgr.Start()
 
--- a/service/matching/physical_task_queue_manager.go
+++ b/service/matching/physical_task_queue_manager.go
@@ -99,7 +99,6 @@ type (
 	matcherInterface interface {
 		Start()
 		Stop()
-		UpdateRatelimit(rpsPtr float64)
 		Rate() float64
 		Poll(ctx context.Context, pollMetadata *pollMetadata) (*internalTask, error)
 		PollForQuery(ctx context.Context, pollMetadata *pollMetadata) (*internalTask, error)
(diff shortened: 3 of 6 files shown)


service/matching/physical_task_queue_manager.go
```diff
@@ -315,7 +314,11 @@ func (c *physicalTaskQueueManagerImpl) PollTask(
 	// we update the ratelimiter rps if it has changed from the last
 	// value. Last poller wins if different pollers provide different values
 	if rps := pollMetadata.taskQueueMetadata.GetMaxTasksPerSecond(); rps != nil {
-		c.matcher.UpdateRatelimit(rps.Value)
+		if c.priMatcher != nil {
+			c.priMatcher.UpdateRatelimit(rps.Value)
+		} else {
+			c.partitionMgr.UpdateRatelimit(rps.Value)
```
dnr, line 320: it's a little weird for the poll call to go through partition manager and then for this part to go back up.. can partition manager's PollTask method pull out the rate limit on the way down?

then this part only has to do priMatcher. that's weird too, admittedly. but it's closer to how we want it to look in the future when priMatcher can do shared limits, right?

service/matching/matching_engine_test.go
```diff
@@ -1155,6 +1165,209 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
...
+		}
+		go func() {
+			result, _ := pollFunc(float64(maxDispatch), strconv.FormatInt(int64(i), 10))
+			resultChan <- result
+		}()
+
+		currentTime := time.Now()
+		s.Eventually(func() bool {
+			return time.Since(currentTime) > 10*time.Millisecond // Delay to allow the second poller coming in a little later.
+		}, 20*time.Millisecond, 1*time.Millisecond)
```
dnr, line 1287: this is just Sleep?

service/matching/matching_engine_test.go
```diff
@@ -1155,6 +1165,209 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 	s.True(descResp.DescResponse.GetTaskQueueStatus().GetRatePerSecond()*numPartitions >= (defaultTaskDispatchRPS - 1))
 }
 
+func (s *matchingEngineSuite) TestRateLimiterAcrossVersionedQueues() {
+	/*
+		1. Start a versioned poller with maxTasksPerSecond = defaultTaskDispatchRPS
+		2. Start another versioned poller, polling a different version, maxTasksPerSecond = 0
+		3. Add tasks to both these versioned queues and notice no dispatch of tasks.
+		4. Restart the pollers with maxTasksPerSecond = defaultTaskDispatchRPS
+		5. Verify that both the pollers have received tasks.
```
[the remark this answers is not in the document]
Shivs11: Just to play devil’s advocate for a moment — I think testing it this way still effectively validates the same core behaviour, right: whether the rate limit is correctly shared across n versioned pollers. So even if the setup isn’t ideal or realistic in practice (given that it's a unit-test), I believe it still serves the purpose of validating if the behaviour is correct + seems to be easier to implement (maybe?); wdyt?

Moreover, isn't this theoretically possible in production, given that we enforce a last poller wins principle?
> relies on different poll requests coming in with different rps values

service/matching/matching_engine_test.go
Shivs11: yes - however, the linter was complaining that I can't use `Sleep`

service/matching/matching_engine_test.go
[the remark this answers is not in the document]
dnr: We don't really enforce exactly that in production, though. You had to have the test override the implementation to get it to behave the way you want, so it's not even testing the real thing. Yeah, it's a unit test so it's focused on one specific part of the system, so maybe that's okay, but with a little more work I think it could be a better and less fragile test.

service/matching/matching_engine_test.go
dnr: :slightly_frowning_face:  Lint is suggestions. If a linter says something that you think is wrong in the context, then you should do the thing you want to do and disable the linter with a comment explaining why. You should't write more complicated and confusing code to trick it

Shivs11: For some reason I thought it was taboo or plainly just not recommended to ignore the linter - good to understand one can ignore it if need be as I genuinely thought it was a firm wide practice to enforce whatever it said

shall revert!

service/matching/matching_engine_test.go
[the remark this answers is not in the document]
Shivs11: I think the overriding would be required if we were to go down the path you suggested as well, right? The override was mainly done to replace the default multi-rate limiter which has a minute long refresh rate - Given that the versioning crew is a little short on time + I have some other things on my plate that I wanna get to before code-cut off, I think I might leave this as is only in the interest of time if we are aligned on the idea that this at-least tests out the modification done in this PR

I can come back and refactor this (maybe a good idea to get to this when we start tinkering on the changes that shall come with `pri_matcher` + rate-limiter)

service/matching/physical_task_queue_manager.go
Shivs11: I think I agree - I pushed a tweak in my latest commit - lmk if this is what you were thinking!

service/matching/matcher_test.go
```diff
@@ -50,6 +51,20 @@ func TestMatcherSuite(t *testing.T) {
...
+func (t *MatcherTestSuite) attachRateLimiter(taskQueueConfig *taskQueueConfig) quotas.RateLimiter {
+	dynamicRateBurst := quotas.NewMutableRateBurst(
+		defaultTaskDispatchRPS,
+		int(defaultTaskDispatchRPS),
+	)
+	dynamicRateLimiter := quotas.NewDynamicRateLimiter(
+		dynamicRateBurst,
+		defaultTaskDispatchRPSTTL,
+	)
+	return dynamicRateLimiter
```
dnr, line 65: could simplify
```suggestion
	return quotas.NewDynamicRateLimiter(
		quotas.NewMutableRateBurst(
			defaultTaskDispatchRPS,
			int(defaultTaskDispatchRPS),
		),
		defaultTaskDispatchRPSTTL,
	)
```

service/matching/matcher_test.go
```diff
@@ -50,6 +51,20 @@ func TestMatcherSuite(t *testing.T) {
 	suite.Run(t, new(MatcherTestSuite))
 }
 
+// attachRateLimiter attaches a rate limiter to the matcher. This behavior is only implemented in tests,
+// as the matcher relies on the rate limiter that is initialized by the task queue partition manager in production.
+func (t *MatcherTestSuite) attachRateLimiter(taskQueueConfig *taskQueueConfig) quotas.RateLimiter {
```
dnr, line 56: "attachRateLimiter" doesn't "attach" anything, it just returns a new RateLimiter. also it doesn't use anything from the taskQueueConfig or the MatcherTestSuite. so this should just be `func newDefaultRateLimiter() quotasRateLimiter`

service/matching/matching_engine_test.go
```diff
@@ -992,19 +995,26 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 	s.taskManager.getQueueManagerByKey(dbq).rangeID = initialRangeID
 
 	mgr := s.newPartitionManager(dbq.partition, s.matchingEngine.config)
+	tqPTM, ok := mgr.(*taskQueuePartitionManagerImpl)
+	s.True(ok)
 	mgrImpl, ok := mgr.(*taskQueuePartitionManagerImpl).defaultQueue.(*physicalTaskQueueManagerImpl)
 	s.True(ok)
-	mgrImpl.oldMatcher.rateLimiter = quotas.NewRateLimiter(
+
+	// Overriding the rate-limiter in the matcher to have a RateLimiterImpl instead of the default multi-rate limiter.
```
dnr, line 1003: This whole section is getting pretty invasive to the actual code. Is there any way we could have something in the config that makes the actual code set up the rate limiter in a way where it works with the test? Set the refresh interval maybe?

service/matching/task_queue_partition_manager.go
```diff
@@ -85,6 +101,28 @@ func newTaskQueuePartitionManager(
...
+			tqConfig.AdminNamespaceTaskQueueToPartitionDispatchRate,
+		),
+		quotas.NewDefaultOutgoingRateLimiter(
+			tqConfig.AdminNamespaceToPartitionDispatchRate,
+		),
+	})
+
+	pm.dynamicRateBurst = dynamicRateBurst
+	pm.dynamicRateLimiter = dynamicRateLimiter
+	pm.rateLimiter = limiter
```
dnr, line 124: just assign these directly?
```suggestion
	pm.dynamicRateBurst = quotas.NewMutableRateBurst(
		defaultTaskDispatchRPS,
		int(defaultTaskDispatchRPS),
	)
	pm.dynamicRateLimiter = quotas.NewDynamicRateLimiter(
		pm.dynamicRateBurst,
		defaultTaskDispatchRPSTTL,
	)
	pm.rateLimiter = quotas.NewMultiRateLimiter([]quotas.RateLimiter{
		pm.dynamicRateLimiter,
		quotas.NewDefaultOutgoingRateLimiter(
			tqConfig.AdminNamespaceTaskQueueToPartitionDispatchRate,
		),
		quotas.NewDefaultOutgoingRateLimiter(
			tqConfig.AdminNamespaceToPartitionDispatchRate,
		),
	})
```

service/matching/task_queue_partition_manager.go
```diff
@@ -93,6 +131,30 @@ func newTaskQueuePartitionManager(
...
+	}
+	burst := int(math.Ceil(rps))
+
+	minTaskThrottlingBurstSize := pm.config.MinTaskThrottlingBurstSize()
+	if burst < minTaskThrottlingBurstSize {
+		burst = minTaskThrottlingBurstSize
+	}
+
+	pm.dynamicRateBurst.SetRPS(rps)
+	pm.dynamicRateBurst.SetBurst(burst)
```
dnr, line 149: it looks like we can do
```suggestion
	pm.dynamicRateBurst.SetRateBurst(rps, burst)
```
to be more efficient (only lock once)

service/matching/matching_engine_test.go
Shivs11: this is done in my latest commit!

service/matching/task_queue_partition_manager.go
Shivs11: looks like `SetRateBurst` is not part of the `dynamicRateBurst` interface

yycptt (commented):
cc @prathyushpv can you help review and see if there's any impact to the OPS work/impl?

service/matching/task_queue_partition_manager.go
dnr: Nevermind, I see that for MutableRateBurstImpl it doesn't matter

PR #7744, opened 2025-05-09
Url prefix fix 1

SakshiMehta:
## What changed?
Updating CreateRemoteFrontendGRPCConnection to handle schema prefix URL
Screenshot 2025-05-09 at 12 11 52 AM
Screenshot 2025-05-08 at 8 19 49 PM

## Why?

These changes are required so that URL with schema can be used in upsert command

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks


caffalaughrey (approved):
~~LGTM~~
**Edit** - re-reviewed mock resolver, see further down.

common/testing/nettest/rpc_factory_test.go
```diff
@@ -84,3 +264,22 @@ func testDialer(t *testing.T, target string, dial func(rpcFactory *nettest.RPCFa
 func newRPCFactory(dialOptions ...grpc.DialOption) *nettest.RPCFactory {
 	return nettest.NewRPCFactory(nettest.NewListener(nettest.NewPipe()), dialOptions...)
 }
+
+// mockResolver is a custom resolver for DNS URLs
+type mockResolver struct{}
+
+func (r *mockResolver) Build(target resolver.Target, cc resolver.ClientConn, opts resolver.BuildOptions) (resolver.Resolver, error) {
```
caffalaughrey, line 271: Given that [operator_handler.go](https://github.com/temporalio/temporal/blob/main/service/frontend/operator_handler.go) does not implement a resolver, I believe tests using this `mockResolver` could yield a false positive.

caffalaughrey: @SakshiMehta in case the operator upsert command is not behaving in line with these unit tests, we should investigate further the operator_handler.go implementation and whether a separate resolver is needed.

caffalaughrey (changes requested):

common/testing/nettest/rpc_factory_test.go
bergundy: WDYM `operation_handler.go` does not implement a resolver?

common/rpc/rpc.go
```diff
@@ -189,12 +190,20 @@ func getListenIP(cfg *config.RPC, logger log.Logger) net.IP {
...
-		hostname, _, err2 := net.SplitHostPort(rpcAddress)
-		if err2 != nil {
-			d.logger.Fatal("Invalid rpcAddress for remote cluster", tag.Error(err2))
+		target := rpcAddress
+		hostname := ""
+
+		// Try parsing the address as a URI
+		if u, parseErr := url.Parse(rpcAddress); parseErr == nil && u.Scheme != "" && strings.Contains(rpcAddress, "://") {
+			target = strings.TrimPrefix(u.Path, "/") // remove any leading slashes
+		}
```
bergundy, line 200: I would rather only try to parse if needed and avoid swallowing errors if they are expected.

```suggestion
		// Try parsing the address as a URI to support custom gRPC resolvers.
		if strings.Contains(rpcAddress, "://") {
			u, parseErr := url.Parse(rpcAddress)
			if parseErr != nil {
				d.logger.Fatal("Invalid rpcAddress for remote cluster. Unable to parse URL", tag.Error(parseErr), tag.String("rpcAddress", rpcAddress))
			}
			// Use the URL's path as the target as expected by gRPC resolvers such "passthrough".
			target = strings.TrimPrefix(u.Path, "/") // remove any leading slashes
		}
```

common/rpc/rpc.go
```diff
@@ -189,12 +190,20 @@ func getListenIP(cfg *config.RPC, logger log.Logger) net.IP {
...
+
+		hostname, _, err := net.SplitHostPort(target)
+		if err != nil {
+			d.logger.Fatal("Invalid rpcAddress for remote cluster. Unable to extract host:port", tag.Error(err))
```
bergundy, line 204:
```suggestion
			d.logger.Fatal("Invalid rpcAddress for remote cluster. Unable to extract host:port", tag.Error(err), tag.String("hostPort", target))
```

bergundy:
Thanks for the contribution. I've added some small suggestions.

bergundy:
Closing this PR due to lack of followup.

PR #7745, opened 2025-05-09
Add new rate limiter priority to support high/low replication tier

yux0:
## What changed?
Add new rate limiter priority to support high/low replication tier

## Why?
high replication tier has higher priority than the low replication tier. The task processing part is separated. This change allow us throttled low replication tier first.

## How did you test it?
- [x] built
- [ ] added new unit test(s)


service/history/replication/executable_task_converter.go
```diff
@@ -143,6 +143,7 @@ func (e *executableTaskConverterImpl) convertOne(
 			replicationTask,
 		)
 	case enumsspb.REPLICATION_TASK_TYPE_SYNC_VERSIONED_TRANSITION_TASK:
+		// TODO: ask why there is no priority
```
yux0, line 146: @xwduan do you know the context? Could I add link the priority?

service/history/replication/eager_namespace_refresher.go
```diff
@@ -23,7 +21,6 @@ import (
 
 type (
 	EagerNamespaceRefresher interface {
-		UpdateNamespaceFailoverVersion(namespaceId namespace.ID, targetFailoverVersion int64) error
```
yycptt, line 26: hmm interesting. How do we make sure namespace failover version is up to date when processing replication tasks? cc @xwduan

service/worker/migration/activities.go
```diff
@@ -897,13 +897,13 @@ func (a *activities) WaitCatchup(ctx context.Context, params CatchUpParams) erro
 		ActiveCluster: descResp.ReplicationConfig.GetActiveClusterName(),
 	}
 
-	activeAckIDOnShard, err := a.getActiveClusterReplicationStatus(ctx, waitCatchupRequest)
+	activeAckIDOnShard, err := a.getActiveClusterReplicationStatus(ctx, waitCatchupRequest) // from B, A ack level
```
yycptt, line 900: what do those comments mean?

common/headers/caller_info.go
```diff
@@ -33,6 +34,10 @@ var (
 		CallerName: CallerNameSystem,
 		CallerType: CallerTypePreemptable,
 	}
+	SystemOptionalCallerInfo = CallerInfo{
+		CallerName: CallerNameSystem,
+		CallerType: CallerTypeOptional,
+	}
```
yycptt, line 40: I think you want this new priority to be higher priority than CallerTypePreemptable because things like retention timer, delete visibility task are also using CallerTypePreemptable.

service/history/replication/eager_namespace_refresher.go
xwduan: At history based replication, it is not required. For state based, we are using SyncNamespaceFromSourceCluster to check and sync the ns failover version.

service/history/replication/executable_task_converter.go
xwduan: I think it is refactored a little bit: https://github.com/temporalio/temporal/blob/91893f10641bae49f8181235c5e93f61a3c826a8/service/history/replication/executable_sync_versioned_transition_task.go#L58. So you don't need to pass the priority here.

common/headers/caller_info.go
```diff
@@ -11,6 +11,7 @@ const (
 	CallerTypeAPI         = "api"
 	CallerTypeBackground  = "background"
 	CallerTypePreemptable = "preemptable"
+	CallerTypeOptional    = "optional"
```
xwduan, line 14: What does `optional` mean here? May add a comment...

service/worker/migration/activities.go
yux0: will remove. This is note from another dicussion.

common/headers/caller_info.go
yux0: I see. My plan was to keep the current priority. Do you think the mentioned operation can be lower priority than live replication?

yux0: add replication type

service/history/replication/stream_sender.go
```diff
@@ -633,7 +634,7 @@ func (s *StreamSenderImpl) shouldProcessTask(item tasks.Task) bool {
 	return shouldProcessTask
 }
 
-func (s *StreamSenderImpl) getTaskPriority(task tasks.Task) enumsspb.TaskPriority {
+func getTaskPriority(task tasks.Task) enumsspb.TaskPriority {
```
xwduan: Maybe add a comment here: This function should only be used on the source/sender side to get the task priority. Receiver side should get task priority directly from field of task. 

Otherwise we will not be able to change this func in the future.

xwduan (approved):

yycptt (changes requested):
Sorry didn't really mean to request any changes for this PR.
Can you sync with @prathyushpv before landing this? He's also trying to add a new priority level: https://github.com/temporalio/temporal/pull/7901

service/history/replication/stream_sender.go
yux0: I think I will revert this change and pass it into the task converter.

yux0:
> Sorry didn't really mean to request any changes in this PR. Can you sync with @prathyushpv before landing this? He's also trying to add a new priority level: #7901

Synced. I will wait for #7901 and reuse the new background priority level.

PR #7747, opened 2025-05-09
Add metrics for namespace operations

prathyushpv:
## What changed?
Add a new metric `operation` which can track total number of operations executed by a namespace.
This includes certain APIs, all history tasks and attempts to despatch matching tasks to pollers.

This metric includes failed API calls, and all history task processing attempts. This includes retried
history tasks, background history tasks(like retention timers) etc.

## Why?
To track the overall usage of a namespace in the cluster.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/common/rpc/interceptor/telemetry.go
+++ b/common/rpc/interceptor/telemetry.go
@@ -21,6 +21,7 @@ import (
 	"go.temporal.io/server/common/rpc/interceptor/logtags"
 	serviceerrors "go.temporal.io/server/common/serviceerror"
 	"go.temporal.io/server/common/tasktoken"
+	"go.temporal.io/server/service/frontend/configs"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/protobuf/types/known/anypb"
@@ -174,6 +175,10 @@ func (ti *TelemetryInterceptor) UnaryIntercept(
 	}()
 
 	resp, err := handler(ctx, req)
+
+	if configs.IsAPIOperation(info.FullMethod) {
+		metrics.OperationCounter.With(metricsHandler).Record(1)
+	}
 
 	if err != nil {
 		ti.HandleError(req, info.FullMethod, metricsHandler, logTags, err, nsName)
--- a/service/frontend/configs/quotas.go
+++ b/service/frontend/configs/quotas.go
@@ -182,6 +182,21 @@ var (
 	}
 
 	NamespaceReplicationInducingAPIPrioritiesOrdered = []int{0, 1, 2}
+
+	// APIs that are not considered as a namespace operation. Namespace operations are used to track the usage of a namespace.
+	// This includes some APIs, history tasks, etc.
+	operationExcludedAPIs = map[string]struct{}{
+		// Poll requests are not considered as namespace operations. We will count these operations when we try to match a task
+		// from matching service to this request.
+		"PollWorkflowTaskQueue": {},
+		"PollActivityTaskQueue": {},
+
+		// Replication-related APIs are not counted as operations.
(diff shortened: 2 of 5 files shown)


service/frontend/configs/quotas.go
```diff
@@ -182,6 +182,21 @@ var (
...
+		// Poll requests are not considered as namespace operations. We will count these operations when we try to match a task
+		// from matching service to this request.
+		"PollWorkflowTaskQueue": {},
+		"PollActivityTaskQueue": {},
+
+		// Replication-related APIs are not counted as operations.
+		"GetWorkflowExecutionRawHistory":   {},
+		"GetWorkflowExecutionRawHistoryV2": {},
+		"ReapplyEvents":                    {},
+		"SyncWorkflowState":                {},
```
yycptt, line 198: Need full method name here? like temporal.api.workflowservice.xxxx ?

common/rpc/interceptor/telemetry.go
```diff
@@ -175,6 +176,10 @@ func (ti *TelemetryInterceptor) UnaryIntercept(
 
 	resp, err := handler(ctx, req)
 
+	if configs.IsAPIOperation(info.FullMethod) {
+		metrics.OperationCounter.With(metricsHandler).Record(1)
+	}
```
yycptt, line 181: To confirm, this relies on the fact that configs.IsAPIOperation implementation is checking full method name, which contains service name right?  so we are only capturing frontend api calls?

service/frontend/configs/quotas.go
prathyushpv: Yes. Thanks for catching!

common/rpc/interceptor/telemetry.go
prathyushpv: Yes. We are only including APIs that are throttled by the frontend RPS interceptor(excluding Visibility and NamespaceReplicationInducing API). Now that I think about it, shouldn't we also include Visibility and NamespaceReplicationInducing APIs?

service/history/queues/executable.go
```diff
@@ -322,6 +322,7 @@ func (e *executableImpl) Execute() (retErr error) {
 		priorityTaggedProvider := e.metricsHandler.WithTags(metrics.TaskPriorityTag(e.priority.String()))
 		metrics.TaskRequests.With(priorityTaggedProvider).Record(1)
 		metrics.TaskScheduleLatency.With(priorityTaggedProvider).Record(e.scheduleLatency)
+		metrics.OperationCounter.With(e.metricsHandler).Record(1)
```
prathyushpv, line 325: There is a metric tag mismatch from the task_type tag in this handler. Let me fix that.

common/rpc/interceptor/telemetry.go
yycptt: Visibility (on frontend) is using a different backend so it should have it's only limit I guess. Visibility tasks need to load mutable state so they should be included in ops.

Good point on NamespaceReplicationInducingAPIs, I was under the impression that we are counting those apis in both execution and also ns replication rate limiter before. so yeah you are right, we should include them in ops. 

Not super important though given the low rps of those apis.

yycptt (approved):

service/history/queues/executable.go
prathyushpv: Adding an empty task_type to 2 other places is easier than removing it from here. It is not straightforward to remove task_type from this metric handler. It is tricky to remove task_type and still add namespace from the list of metric tags returned by the task. I will create a follow-up to do that and merge this change. For now I have added an empty task_type tag to the other 2 places.

Issue #7631, opened 2025-04-18
Unable to create namespace when s3 archival and internal-frontend is enabled


dlydiard:
## Expected Behavior
Namespace is created when archival is enabled.

## Actual Behavior
> Error: Register namespace operation failed.
> Error Details: rpc error: code = Unknown desc = something went wrong, please retry (6b50ceaa)

## Steps to Reproduce the Problem
```yaml
global:
    membership:
        maxJoinDuration: 30s
        broadcastAddress: '{{ default .Env.POD_IP "0.0.0.0" }}'
    pprof:
        port: 0
        host: ""
    tls:
        internode:
            client:
                serverName: ""
                disableHostVerification: false
                rootCaFiles: []
                rootCaData: []
                forceTLS: false
            server:
                certFile: ""
                keyFile: ""
                clientCaFiles: []
                certData: ""
                keyData: ""
                clientCaData: []
                requireClientAuth: false
            hostOverrides: {}
        frontend:
            client:
                serverName: ""
                disableHostVerification: false
                rootCaFiles: []
                rootCaData: []
                forceTLS: false
            server:
                certFile: ""
                keyFile: ""
                clientCaFiles: []
                certData: ""
                keyData: ""
                clientCaData: []
                requireClientAuth: false
            hostOverrides: {}
        systemWorker:
            certFile: ""
            keyFile: ""
            certData: ""
            keyData: ""
            client:
                serverName: ""
                disableHostVerification: false
                rootCaFiles: []
                rootCaData: []
                forceTLS: false
        remoteClusters: {}
        expirationChecks:
            warningWindow: 0s
            errorWindow: 0s
            checkInterval: 0s
        refreshInterval: 0s
    metrics:
        tags:
            type: '{{ .Env.SERVICES }}'
        excludeTags: {}
        prefix: ""
        perUnitHistogramBoundaries: {}
        m3: null
        statsd: null
        prometheus:
            framework: ""
            listenAddress: 0.0.0.0:9090
            handlerPath: ""
            loggerRPS: 0
            listenNetwork: ""
            timerType: histogram
            defaultHistogramBoundaries: []
            defaultHistogramBuckets: []
            defaultSummaryObjectives: []
            onError: ""
            sanitizeOptions: null
    authorization:
        jwtKeyProvider:
            keySourceURIs:
                - https://dev.auth.test.com/.well-known/jwks.json
            refreshInterval: 10m0s
        permissionsClaimName: https://temporal/permissions
[...]

PR #7748, opened 2025-05-09
Fixes issue #7631 with wrong archiver lookup

PADRESH:
Calling the GetVisibilityArchiver and GetHistoryArchiver functions causes errors when the internal frontend is enabled. We should either check the current configuration or simply call these functions with the InternalFrontendService parameter if an error occurs.

## What changed?
Added fallback logic for archiver retrieval in namespace_handler.go and workflow_handler.go to handle cases when the initial call fails.

## Why?
When the internal frontend is enabled, all visibility and history containers are registered under the InternalFrontendService key. However, some calls in namespace_handler.go and workflow_handler.go still attempt to retrieve archivers using the FrontendService key.

## How did you test it?
- [ v] built
- [v ] run locally and tested manually
- [v ] covered by existing tests
- [ ] added new unit test(s)

## Potential risks
If the archival configuration is incorrect, the system takes additional time to fail/recognize the error.

### DIFF
--- a/common/log/tag/tags.go
+++ b/common/log/tag/tags.go
@@ -281,7 +281,7 @@ func WorkflowTaskFailedCause(workflowTaskFailCause enumspb.WorkflowTaskFailedCau
 
 // WorkflowTaskQueueType returns tag for WorkflowTaskQueueType
 func WorkflowTaskQueueType(taskQueueType enumspb.TaskQueueType) ZapTag {
-	return NewStringerTag("wf-task-queue-type", taskQueueType)
+	return NewStringTag("wf-task-queue-type", taskQueueType.String())
 }
 
 // WorkflowTaskQueueName returns tag for WorkflowTaskQueueName
@@ -632,7 +632,7 @@ func TaskVersion(taskVersion int64) ZapTag {
 }
 
 func TaskType(taskType enumsspb.TaskType) ZapTag {
-	return NewStringerTag("queue-task-type", taskType)
+	return NewStringTag("queue-task-type", taskType.String())
 }
 
 func TaskCategoryID(taskCategoryID int) ZapTag {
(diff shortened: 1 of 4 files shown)


yycptt:
Thanks for the contribution @PADRESH!

It looks like the issue is caused by [`namespaceHandler`](https://github.com/temporalio/temporal/blob/main/service/frontend/namespace_handler.go#L956) and [`workflowHandler`](https://github.com/temporalio/temporal/blob/main/service/frontend/workflow_handler.go#L5390) on frontend hardcoded to use `primitives.FrontendService` when getting the archiver implementation. 

Instead of adding fallback logic, I think we should remove those hardcoded service name and plumb through the actual serviceName to those two components. Similar to what [RegisterBootstrapContainer](https://github.com/temporalio/temporal/blob/main/common/resource/fx.go#L282) is doing. 
`primitives.ServiceName` type is already available in the fx graph and in `HandlerProvider` in service/frontend/fx.go, so only `NewWorkflowHandler` and `newNamespaceHandler` needs to be updated to take in an additional ServiceName argument. 

cc @dnr who has more context on internal frontend.

PADRESH:
@yycptt, thank you for the suggestions. I've committed the changes and tested them locally (kubernetes).

PADRESH:
@yycptt, @dnr What do you think about my last commit? What else can I do to promote the request?

dnr:
This is okay for what it is, but the problem is really that the archival provider/bootstrap/fx stuff is kind of a mess. There's no reason that it should take a service name at all, each fx app gets its own archiverProvider so it will only ever be called with one service name. The "Bootstrap" structs should just be parameter lists or fx.In structs. If you'd like to tackle refactoring this all, that would be even better than this fix, since it would remove a lot of unnecessary code.

Though I think we can accept this in the short term if it fixes the problem.

I ran CI for this PR... note the failures and please fix.

PR #7749, opened 2025-05-09
add rule expiration check

ychebotarev:
## What changed?
Check if rule is expired.

## Why?
We have expiration time. After that time rule shouldn't be applied.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


gow (approved):
Change looks good.
This introduces a new conditional statement. How does the existing tests cover it?

PR #7750, opened 2025-05-09
Add task_priority label to task_dispatch_latency and poll_latency

dnr:
## What changed?
Split poll_latency and task_dispatch_latency metrics by simple priority level.

## Why?
Observability, to have stats on throughput and latency by level.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/pri_matcher.go
+++ b/service/matching/pri_matcher.go
@@ -440,10 +440,16 @@ func (tm *priTaskMatcher) emitDispatchLatency(task *internalTask, forwarded bool
 		return // should not happen but for safety
 	}
 
+	priStr := ""
+	if pri := task.getPriority().GetPriorityKey(); pri > 0 {
+		priStr = strconv.Itoa(int(pri))
+	}
+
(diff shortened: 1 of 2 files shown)


service/matching/pri_matcher.go
```diff
@@ -440,10 +440,16 @@ func (tm *priTaskMatcher) emitDispatchLatency(task *internalTask, forwarded bool
 		return // should not happen but for safety
 	}
 
+	priStr := ""
+	if pri := task.getPriority().GetPriorityKey(); pri > 0 {
```
ShahabT, line 444: is it intentional to use "" for priority 0?

ShahabT (approved):

service/matching/pri_matcher.go
dnr: yes.. do you think it should be 0? I feel like 0 is misleading since it's not "priority 0", it's "use the default priority"

PR #7752, opened 2025-05-10
remove old rules when more space is needed

ychebotarev:
## What changed?
Remove oldest expired workflow rule is more space is needed.
Add per namespace config options to limit the number of rules.

## Why?
Auto-cleanup expired rules. We keep them in case user still want to use them, but we don't want to fail adding new rule.

## How did you test it?
- [ ] built
- [X] added new unit test(s)

### DIFF
--- a/service/frontend/namespace_handler.go
+++ b/service/frontend/namespace_handler.go
@@ -687,6 +687,14 @@ func (d *namespaceHandler) CreateWorkflowRule(
 
 	if config.WorkflowRules == nil {
 		config.WorkflowRules = make(map[string]*rulespb.WorkflowRule)
+	} else {
+		maxRules := d.config.MaxWorkflowRulesPerNamespace(nsName)
+		if len(config.WorkflowRules) > maxRules {
+			d.removeExpiredWorkflowRules(config.WorkflowRules)
+		}
+		if len(config.WorkflowRules) > maxRules {
+			return nil, serviceerror.NewInvalidArgument(fmt.Sprintf("Workflow Rule limit exceeded. Max: %v", maxRules))
+		}
 	}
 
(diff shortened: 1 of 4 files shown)


service/frontend/namespace_handler.go
```diff
@@ -687,6 +687,14 @@ func (d *namespaceHandler) CreateWorkflowRule(
 
 	if config.WorkflowRules == nil {
 		config.WorkflowRules = make(map[string]*rulespb.WorkflowRule)
+	} else {
+		maxRules := d.config.MaxWorkflowRulesPerNamespace(nsName)
+		if len(config.WorkflowRules) > maxRules {
```
gow, line 692: nit: `>=` (here and below). Otherwise we'll have upto max+1 rules per namespace.

service/frontend/namespace_handler.go
```diff
@@ -722,6 +730,29 @@ func (d *namespaceHandler) CreateWorkflowRule(
 	return workflowRule, nil
 }
 
+func (d *namespaceHandler) removeExpiredWorkflowRules(rules map[string]*rulespb.WorkflowRule) {
```
gow, line 733: nit: this method name suggests that it might be removing all expired workflow rules. But we're actually removing just one oldest expired rule. Maybe the name can be something like `removeExpiredWorkflowRules` or `removeOldestExpiredWorkflowRule` etc?

gow: Also, why not remove all expired ones while we are here?

gow (approved):
Please see some comments below.

service/frontend/namespace_handler.go
ychebotarev: renamed to removeOldestExpiredWorkflowRule

PR #7753, opened 2025-05-12
Deprecation proposal: Block build-id as a search attribute

Shivs11:
## What changed?
-  A deprecation proposal to block the usage of build-id as a search attribute. Note, the changes introduced in this PR shall only emit warnings (for now) to analyze how many current users are actually doing this (expected number to be very low).
- Once we have a confirmation that the number of users who are using this are low, I will send a follow-up to error out when buildID is used as a SA.

## Why?
- To prevent users from using pre-defined SA's.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/common/searchattribute/validator.go
+++ b/common/searchattribute/validator.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"fmt"
 
+	"go.temporal.io/server/common/log"
+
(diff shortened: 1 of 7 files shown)


rodrigozhou (changes requested):
I would use logs as a way to measure how many users are using it. I'd suggest creating a new metric for it.

PR #7755, opened 2025-05-12
Block the usage of worker-deployment related SA's

Shivs11:
## What changed?
- Block the usage of worker-deployment related SA's

## Why?
- Don't allow users to use SA's that are supposed to be `predefined` for a feature.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
- The only risk I can think of is that current users using this might face errors. However, this is a pre-release feature going into public-preview and I think it's okay to ask those folks to restart their workflows.


common/searchattribute/validator.go
```diff
@@ -111,6 +111,12 @@ func (v *Validator) Validate(searchAttributes *commonpb.SearchAttributes, namesp
 			)
 		}
 
+		// Don't allow predefined Deployment SA's to be set by the user
```
carlydf, line 114: there's a section a few lines up that does
```
		// user search attribute cannot be a system search attribute
		if _, err = saTypeMap.getType(saFieldName, systemCategory); err == nil {
			if v.suppressErrorSetSystemSearchAttribute(namespace) {
				// if suppressing the error, then just ignore the search attribute
				continue
			}
			return serviceerror.NewInvalidArgument(
				fmt.Sprintf("%s attribute can't be set in SearchAttributes", saFieldName),
			)
		}
```

The Worker Deployment search attributes are also system search attributes. Could they added to a list / given some tag so that they'll be included in the existing "no system search attributes" check?

carlydf (commented):
tests look good. I suggest reusing the existing method for preventing system search attributes from being added by users, if that's possible!

common/searchattribute/validator.go
Shivs11: Discussed offline but pasting this here since someone else might have the same idea of placing this inside the `system` map - I think that may not be a good idea since according to the definition, system SA's seem to be defined as those SA's which have separate fields in the DB. In other words, they would be created every single time for all the workflows a user starts. Placing deployment related SA's over there may not sound like a good idea given that users may not use versioning all the time!

carlydf, line 114:
```suggestion
		// Don't allow these new predefined Deployment SA's to be set by the user
		// Existing predefined SA's should be banned, but may be in use already, so we cannot ban them yet.
		// Plan is to collect metrics on usage of other predefined Deployment SA's to move towards being able
		// to ban their use as Custom Search Attributes.
```

carlydf (approved):

PR #7756, opened 2025-05-12
Add WorkflowTypeName to completion metrics

josesa:
## What changed?
Added the `WorkflowTypeName` as a label on completion metrics. 

## Why?
Being able to track metrics (`operation="CompletionStats"`) for completion and failures (`workflow_success`, `workflow_failed`, ...) per `WorkflowTypeName`.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
In cases where the `WorkflowTypeName` is automatically generated or in general not suitable for usage as a label, it could lead to high cardinality of the metrics.


bergundy:
I don't think we can accept this due to the high cardinality of this label.

We could consider adding a dynamic config to turn this on though.
Curious what others have to say.

josesa:
That was my main concern.
I noticed that there was another metric that was also using this as a label, but I currently I don't recall exactly which: I will try to find it to what is the trigger for it.

In case we prefer to add a configuration, what would be a good example for it?

yycptt:
Task queue tag is controlled by a dynamic config I believe. 
Alternatively, the behavior can be controlled via static config. There is a `ExcludeTags` field in metric config to do this. And I think tags like workflowType, activityType tags are already excluded in our production config given that they are used in other places as well. 
For other oss users, since that tag is already used elsewhere should be fine as well I think.

yycptt (approved):

PR #7757, opened 2025-05-13
Resend parent for standby child completion verification

hai719:
## What changed?
Resend parent for standby child completion verification.

## Why?
In the case that standby fails to verify child completion when parent workflow is not found or not ready, standby should request resend parent from active so standby can finish verification earlier.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
New calls will be made to active. But this will only happen after the new config maxLocalParentWorkflowVerificationDuration. So impacts to active will be low.

### DIFF
--- a/proto/internal/temporal/server/api/historyservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/historyservice/v1/request_response.proto
@@ -556,6 +556,7 @@ message VerifyChildExecutionCompletionRecordedRequest {
     int64 parent_initiated_id = 4;
     int64 parent_initiated_version = 5;
     temporal.server.api.clock.v1.VectorClock clock = 6;
+    bool resend_parent = 7;
 }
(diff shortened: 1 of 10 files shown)


proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -556,6 +556,7 @@ message VerifyChildExecutionCompletionRecordedRequest {
     int64 parent_initiated_id = 4;
     int64 parent_initiated_version = 5;
     temporal.server.api.clock.v1.VectorClock clock = 6;
+    bool resend_parent = 7;
```
yux0, line 559: no strong opinion. But as this API is verifyChild. maybe we should handle the resend logic on the caller side?

yux0 (approved):

PR #7764, opened 2025-05-14
Refactor: remove FunctionalTestSuite base struct

alexshtin:
## What changed?
Refactor: remove `FunctionalTestSuite` base struct.

## Why?
I want to keep functional tests structure simple. My goal is one base struct for all tests. All helper functions and objects should be there.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


tests/testcore/functional_test_base.go
```diff
@@ -76,6 +77,9 @@ type (
 		sdkClient sdkclient.Client
 		worker    sdkworker.Worker
 		taskQueue string
+
+		// TODO (alex): replace with v2
+		taskPoller *taskpoller.TaskPoller
```
dnr, line 82: You don't want to leave it public for convenience?

ychebotarev: lets keep it like the other members (worker, sdkClient)

ychebotarev (approved):

tests/testcore/functional_test_base.go
alexshtin: It is good to guarantee that it is never set outside of initialization code. Yes, I want to keep getters and unexported fields. At least for now. May be will change my mind later :-)

PR #7766, opened 2025-05-14
Refactor archival fx initialization

dnr:
## What changed?
Refactor how archival subsystem gets its dependencies to work better with fx:
- Remove service type maps (archivalProvider is already service-specific)
- Remove "bootstrap" structs and just pass around dependencies

## Why?
The service type maps were causing issues like #7631, and are unnecessary.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/archiver/interface.go
+++ b/common/archiver/interface.go
@@ -8,9 +8,6 @@ import (
 	historypb "go.temporal.io/api/history/v1"
 	workflowpb "go.temporal.io/api/workflow/v1"
 	archiverspb "go.temporal.io/server/api/archiver/v1"
-	"go.temporal.io/server/common/log"
-	"go.temporal.io/server/common/metrics"
-	"go.temporal.io/server/common/persistence"
 	"go.temporal.io/server/common/searchattribute"
 )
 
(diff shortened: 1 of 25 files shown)


common/archiver/interface.go
```diff
@@ -44,14 +40,6 @@ type (
 		NextPageToken  []byte
 	}
 
-	// HistoryBootstrapContainer contains components needed by all history Archiver implementations
-	HistoryBootstrapContainer struct {
```
alexshtin, line 48: Bye bye old friend.

tests/testcore/test_cluster.go
```diff
@@ -225,7 +226,7 @@ func newClusterWithPersistenceTestBaseFactory(t *testing.T, clusterConfig *TestC
 	testBase := tbFactory.NewTestBase(&clusterConfig.Persistence)
 
 	testBase.Setup(clusterMetadataConfig)
-	archiverBase := newArchiverBase(clusterConfig.EnableArchival, logger)
+	archiverBase := newArchiverBase(clusterConfig.EnableArchival, testBase.ExecutionManager, logger)
```
alexshtin, line 229: I have plans around this `archiverBase` too.

alexshtin (approved):
Build is failing. Otherwise great step in the right direction.

common/archiver/interface.go
dnr: Did you add these? Was that before fx?

PR #7768, opened 2025-05-14
functional test to verify update routing with versioning-0.32

Shivs11:
## What changed?
- added two new functional tests that verify whether updates, which are sent as speculative tasks, are being routed properl. - also checks whether the MS has the right versioningInfo while the routing takes place.

## Why?
- making this feature more robust

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -19,6 +19,7 @@ import (
 	nexuspb "go.temporal.io/api/nexus/v1"
 	"go.temporal.io/api/serviceerror"
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
+	updatepb "go.temporal.io/api/update/v1"
 	workflowpb "go.temporal.io/api/workflow/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	"go.temporal.io/sdk/activity"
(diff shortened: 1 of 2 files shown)


tests/versioning_3_test.go
```diff
@@ -63,7 +65,7 @@ const (
 )
 
 type Versioning3Suite struct {
-	testcore.FunctionalTestSuite
+	WorkflowUpdateBaseSuite
```
Shivs11, line 68: this is done so that we can inherit some great helpers defined already in the `update_workflow_suite_base` file!

tests/versioning_3_test.go
```diff
@@ -454,6 +456,198 @@ func (s *Versioning3Suite) testUnpinnedWorkflow(sticky bool) {
...
+	`, events)
+
+	// Since the poller rejected the update, the Worker Deployment Version that completed the last workflow task
+	// of this workflow execution should not have changed.
+	s.verifyWorkflowVersioningState(execution, tv1.DeploymentVersionString(), nil)
+}
+
+func (s *Versioning3Suite) sendUpdateNoError(tv *testvars.TestVars) <-chan *workflowservice.UpdateWorkflowExecutionResponse {
+	s.T().Helper()
+	return s.sendUpdateNoErrorInternal(tv, nil)
```
Shivs11, line 648: method definition inside the update_workflow_suite_base file: https://github.com/temporalio/temporal/blob/49f179736e78a1287db8d7342debac7841419ea6/tests/update_workflow_suite_base.go#L31

PR #7770, opened 2025-05-14
reapply child workflow completion events

hai719:
## What changed?
reapply child workflow completion events (completed, cancelled, etc)

## Why?
During conflict resolution, if parent uses new branch that doesn't include child workflow completion events, then it will not see the events forever.

## How did you test it?
- [x] built
- [x] added new unit test(s)


service/history/ndc/workflow_resetter.go
```diff
@@ -1023,7 +1023,7 @@ func reapplyChildEvents(mutableState historyi.MutableState, event *historypb.His
 	case enumspb.EVENT_TYPE_CHILD_WORKFLOW_EXECUTION_STARTED:
 		childEventAttributes := event.GetChildWorkflowExecutionStartedEventAttributes()
 		ci, childExists := mutableState.GetChildExecutionInfo(childEventAttributes.GetInitiatedEventId())
-		if !childExists {
+		if !childExists || ci.StartedEventId != common.EmptyEventID {
```
xwduan, line 1026: Q: Is this change to avoid duplicate start event?

xwduan (approved):

PR #7773, opened 2025-05-15
Add health signal recording to interceptor

laniehei:
This adds an interceptor to the handler service that logs latency and errors from handlers, and can be reused in other services.

Adds the interceptor to the history service, that takes a health aggregator provider. This provider is also added to the handler such that we can read these values in the DeepHealthCheck handler within the history service. 

A future PR will add this check to the other services.


service/history/handler.go
```diff
@@ -123,6 +125,7 @@ type (
 		TaskQueueManager             persistence.HistoryTaskQueueManager
 		TaskCategoryRegistry         tasks.TaskCategoryRegistry
 		DLQMetricsEmitter            *persistence.DLQMetricsEmitter
+		HealthSignalAggregator       HealthSignalAggregator
```
prathyushpv, line 128: Now we have PersistenceHealthSignal and HealthSignalAggregator in this struct.
How about we name them to make it clear what they aggregate? Something like PersistenceHealthSignal and RpcHealthSignal?

Also, we can use the same persistence.HealthSignalAggregator type for both of these, if possible.

service/history/health_signal_aggregator.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (s *HealthSignalAggregatorImpl) Stop() {
+	if !atomic.CompareAndSwapInt32(&s.status, common.DaemonStatusStarted, common.DaemonStatusStopped) {
+		return
+	}
+	close(s.shutdownCh)
+	s.emitMetricsTimer.Stop()
+}
+
+func (s *HealthSignalAggregatorImpl) Record(method string, latency time.Duration, err error) {
```
prathyushpv, line 112: I see that we are tracking latency for each method. Do we need that?
For persistence, we are emitting per-shard latency. That is because it is not possible to emit shard-id tag in metrics. That will increase the cardinality of that metric beyond what we can handle. 

We are already emitting per method metrics already. So we might not have to run emitMetricsLoop() for history rpc health tracking. 

It would be better if we reuse persistence.HealthSignalAggregator here. We can pass a boolean value to it's constructor to disable emitMetricsLoop(). If we don't care about shard-id and namespace, we can pass empty values to it while calling Record()

service/history/handler.go
```diff
@@ -264,6 +286,11 @@ func (h *Handler) IsActivityTaskValid(ctx context.Context, request *historyservi
 		return nil, h.convertError(err)
 	}
 
+	startTime := h.timeSource.Now()
+	defer func() {
```
prathyushpv, line 290: We can combine this and `metrics.CapturePanic(h.logger, h.metricsHandler, &retError)` above in a single deferred function. It is better to record the start time at the beginning of the function. We can treat the total time spent in this function as the request latency.

service/history/handler.go
laniehei: We can absolutely name it something more descriptive. I would like to push back against reusing the same type, on the basis that this isn't persistence, though. We could talk about moving this to a shared package in the future, or I could do this on this PR if you have a good idea of where it could live.

service/history/handler.go
```diff
@@ -237,6 +254,11 @@ func (h *Handler) IsWorkflowTaskValid(ctx context.Context, request *historyservi
...
+		h.healthSignalAggregator.Record("IsWorkflowTaskValid", time.Since(startTime), retError)
+	}()
```
yux0, line 260: Is it possible to have it as an interceptor?

service/history/health_signal_aggregator.go
```diff
@@ -0,0 +1,198 @@
...
+			s.requestsLock.Unlock()
+
+			for method, count := range requestCounts {
+				methodRPS := int64(float64(count) / emitMetricsInterval.Seconds())
+				s.metricsHandler.Histogram(HistoryServiceRPS.Name(), HistoryServiceRPS.Unit()).Record(methodRPS)
+				if methodRPS > int64(s.perMethodRPSWarnLimit()) {
+					s.logger.Warn("Per method RPS warn limit exceeded",
+						tag.NewStringTag("method", method),
+						tag.NewInt64("rps", methodRPS))
+				}
```
yux0, line 158: what is the reason to emit those metrics?

service/history/health_signal_aggregator.go
laniehei: It looks like we're tracking latency for each method, but the actual aggregation and then subsequent Latency number produced is looking at all of them, regardless of method. 

I've converted the emitMetricsLoop() to be a noop -- you're right on us not needing to emit the metrics because we're already doing this elsewhere where we can consume them. 

I would prefer not to reuse the persistence HealthSignalAggregator, because this is a bit confusing with naming/package placement. Will keep separate for now until we find a good shared spot for this.

service/history/health_signal_aggregator.go
laniehei: Discussed with @prathyushpv and there's no need as we're already doing this. Moved to noop.

service/history/handler.go
laniehei: We would have to pull from prometheus and we risk it being down. Let me know if I'm misunderstanding the suggestion

common/rpc/interceptor/health_check.go
```diff
@@ -0,0 +1,127 @@
...
+	startTime := time.Now()
+	resp, err := handler(ctx, req)
+	elapsed := time.Since(startTime)
+
+	h.historyHealthSignalAggregator.Record(elapsed, err)
+
+	return resp, err
+}
+
+type (
```
yux0, line 42: Is it possible to make this its own file and share between the interceptor and persistence?

common/rpc/interceptor/health_check.go
```diff
@@ -0,0 +1,127 @@
...
+	// HealthSignalAggregator interface for aggregating health signals
+	HealthSignalAggregator interface {
+		Record(latency time.Duration, err error)
+		AverageLatency() float64
+		ErrorRatio() float64
+	}
+
+	// HealthSignalAggregatorImpl implements HealthSignalAggregator
+	HealthSignalAggregatorImpl struct {
```
yux0, line 51: nit: healthSignalAggregatorImpl

service/history/fx.go
```diff
@@ -170,6 +173,31 @@ func TelemetryInterceptorProvider(
...
+	healthSignalAggregator *interceptor.HealthSignalAggregatorImpl,
+) *interceptor.HealthCheckInterceptor {
+	if dynamicconfig.HistoryHealthSignalMetricsEnabled.Get(dynamicCollection)() {
+		return interceptor.NewHealthCheckInterceptor(
+			healthSignalAggregator,
+		)
+	}
+	return interceptor.NewHealthCheckInterceptor(
+		interceptor.NoopHealthSignalAggregator,
+	)
```
yux0, line 199: Is it possible to check HistoryHealthSignalMetricsEnabled so it doesn't require reboot when updated the config.

common/rpc/interceptor/health_check.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// UnaryIntercept implements the gRPC unary interceptor interface
+func (h *HealthCheckInterceptor) UnaryIntercept(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
+	startTime := time.Now()
```
yux0, line 33: maybe exclude DeepHealthCheck API?

common/rpc/interceptor/health_check.go
```diff
@@ -0,0 +1,127 @@
...
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/aggregate"
+	"go.temporal.io/server/common/log"
+	"google.golang.org/grpc"
+)
+
+// HealthCheckInterceptor is a gRPC interceptor that records health metrics
+type HealthCheckInterceptor struct {
+	historyHealthSignalAggregator HealthSignalAggregator
```
yux0, line 16: nit: not specific to history

service/history/handler.go
```diff
@@ -217,6 +225,25 @@ func (h *Handler) DeepHealthCheck(
...
+		metrics.HistoryHostHealthGauge.With(h.metricsHandler).Record(float64(enumsspb.HEALTH_STATE_NOT_SERVING))
+		return &historyservice.DeepHealthCheckResponse{State: enumsspb.HEALTH_STATE_NOT_SERVING}
+	}
+
+	// Check if the RPC error ratio exceeds the threshold
+	if h.historyHealthSignal.ErrorRatio() > h.config.HealthRPCErrorRatio() {
+		metrics.HistoryHostHealthGauge.With(h.metricsHandler).Record(float64(enumsspb.HEALTH_STATE_NOT_SERVING))
+		return &historyservice.DeepHealthCheckResponse{State: enumsspb.HEALTH_STATE_NOT_SERVING}
+	}
+	return nil
```
yux0, line 244: return &historyservice.DeepHealthCheckResponse{State: enumsspb.HEALTH_STATE_SERVING}

service/history/handler.go
```diff
@@ -206,6 +209,11 @@ func (h *Handler) DeepHealthCheck(
 		return &historyservice.DeepHealthCheckResponse{State: enumsspb.HEALTH_STATE_DECLINED_SERVING}, nil
 	}
 
+	rsp := h.checkHistoryHealthSignals()
+	if rsp != nil {
+		return rsp, nil
+	}
+
```
yux0, line 216: Can you check if the L221 should be metrics.HistoryHostHealthGauge.With(h.metricsHandler).Record(float64(enumsspb.HEALTH_STATE_NOT_SERVING))

service/history/fx.go
```diff
@@ -170,6 +173,31 @@ func TelemetryInterceptorProvider(
...
 }
 
+func HealthSignalAggregatorProvider(
+	dynamicCollection *dynamicconfig.Collection,
+	metricsHandler metrics.Handler,
+	logger log.ThrottledLogger,
+) interceptor.HealthSignalAggregator {
+	return interceptor.NewHealthSignalAggregatorImpl(
+		dynamicconfig.PersistenceHealthSignalWindowSize.Get(dynamicCollection)(),
+		dynamicconfig.PersistenceHealthSignalBufferSize.Get(dynamicCollection)(),
```
yux0, line 183: could it use its own config?

common/rpc/interceptor/health.go
```diff
@@ -10,6 +10,7 @@ import (
 	"google.golang.org/grpc"
 )
 
+// HealthInterceptor rejects frontend requests if the host has not been marked as healthy.
```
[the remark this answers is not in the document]
laniehei: If you read the code within the interceptor, it returns a frontend error (notHealthyErr) and is only used in the frontend. https://github.com/temporalio/temporal/blob/lanie/history-rpc-health-aggregator/common/rpc/interceptor/health.go#L39

common/rpc/interceptor/health_check.go
laniehei: where would be a good shared package? Mentioned in an earlier comment that I didn't do this on purpose for simplicity of the PR https://github.com/temporalio/temporal/pull/7773#discussion_r2093650649

common/rpc/interceptor/health_check.go
laniehei: Need to keep exported so that fx has access to it in the history package https://github.com/temporalio/temporal/blob/lanie/history-rpc-health-aggregator/service/history/fx.go#L190

laniehei: lmk if you think fx will keep a pointer to the same aggregator impl if it's passed in as an interface -- fx is a bit of a black box to me

service/history/handler.go
laniehei: oh you're correct -- it should be NotServing, not DeclinedServing. Changed, thanks!

service/history/fx.go
laniehei: We could, but I don't want to overcomplicate before we feel it's necessary! LMK if you have strong opinions

common/rpc/interceptor/health_check.go
yux0: I am thinking a new pkg under common/health

common/rpc/interceptor/health_check.go
yux0: I see. I think as long as the interface is accessible. You might need to declare a new interface for the health signal aggregator for persistence and history if you are going to have a shared package.

service/history/handler.go
yux0: It makes more sense to return the response and check the health state. Return if the health state if not serving

laniehei: but we have more health state to check -- this is just a helper to encapsulate functionality. We don't know if it's serving or not yet

common/rpc/interceptor/health_check.go
laniehei: changed this to be unexported along with the other impl -- it needed an interface, not a struct

service/history/handler.go
yux0: in this case, could you return an error instead?

laniehei: It isn't an error though -- there was no error that occurred, just a lack of signal that anything is wrong

PR #7774, opened 2025-05-15
CHASM: ComponentPointer support

alexshtin:
## What changed?
CHASM: Add basic `ComponentPointer` support to CHASM component.

## Why?
Part of CHASM workstream. Few things are still pending:
1. `RefC` and `RefD` should return error.
2. Implementation of `RefC` and `RefD` needs to be revisited. It is not clear if it is possible to implement w/o deserializing entire component starting from root and all the way down, because `n.value` is needed to compare with passed argument.
3. Dangling pointers protection is not implemented.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/context.go
+++ b/chasm/context.go
@@ -13,8 +13,8 @@ type Context interface {
 
 	// NOTE: component created in the current transaction won't have a ref
 	// this is a Ref to the component state at the start of the transition
-	RefC(Component) (ComponentRef, bool)
-	RefD(proto.Message) (ComponentRef, bool)
+	Ref(Component) (ComponentRef, error)
+	refData(proto.Message) (ComponentRef, error)
 	Now(Component) time.Time
 
 	// Intent() OperationIntent
@@ -31,13 +31,13 @@ type MutableContext interface {
 	// Add more methods here for other storage commands/primitives.
 	// e.g. HistoryEvent
 
-	// Get a RefC for the component
+	// Get a Ref for the component
 	// This ref to the component state at the end of the transition
-	// Same as RefC(Component) method in Context,
+	// Same as Ref(Component) method in Context,
 	// this only works for components that already exists at the start of the transition
 	//
 	// If we provide this method, then the method on the engine doesn't need to
-	// return a RefC
+	// return a Ref
 	// NewRef(Component) (ComponentRef, bool)
 }
 
@@ -63,12 +63,12 @@ func NewContext(
 	}
 }
 
-func (c *ContextImpl) RefC(component Component) (ComponentRef, bool) {
-	return c.root.RefC(component)
+func (c *ContextImpl) Ref(component Component) (ComponentRef, error) {
+	return c.root.Ref(component)
 }
 
-func (c *ContextImpl) RefD(data proto.Message) (ComponentRef, bool) {
-	return c.root.RefD(data)
+func (c *ContextImpl) refData(data proto.Message) (ComponentRef, error) {
+	return c.root.refData(data)
 }
 
 func (c *ContextImpl) Now(component Component) time.Time {
--- a/chasm/context.go
+++ b/chasm/context.go
@@ -13,8 +13,8 @@ type Context interface {
 
 	// NOTE: component created in the current transaction won't have a ref
 	// this is a Ref to the component state at the start of the transition
-	Ref(Component) (ComponentRef, bool)
-	DataRef(data proto.Message) (ComponentRef, bool)
+	RefC(Component) (ComponentRef, bool)
+	RefD(proto.Message) (ComponentRef, bool)
 	Now(Component) time.Time
 
 	// Intent() OperationIntent
@@ -31,13 +31,13 @@ type MutableContext interface {
 	// Add more methods here for other storage commands/primitives.
 	// e.g. HistoryEvent
 
-	// Get a Ref for the component
+	// Get a RefC for the component
 	// This ref to the component state at the end of the transition
-	// Same as Ref(Component) method in Context,
+	// Same as RefC(Component) method in Context,
 	// this only works for components that already exists at the start of the transition
 	//
 	// If we provide this method, then the method on the engine doesn't need to
-	// return a Ref
+	// return a RefC
 	// NewRef(Component) (ComponentRef, bool)
 }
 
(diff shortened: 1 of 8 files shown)


chasm/context.go
```diff
@@ -11,7 +13,8 @@ type Context interface {
 
 	// NOTE: component created in the current transaction won't have a ref
 	// this is a Ref to the component state at the start of the transition
-	Ref(Component) (ComponentRef, bool)
+	RefC(Component) (ComponentRef, bool)
+	RefD(proto.Message) (ComponentRef, bool)
```
yycptt, line 17: I don't think we want to expose the Ref to a Data, for now at least. Component is the most basic unit someone outside the CHASM framework can interact with. 
We probably need a different type of Ref (instead of ComponentRef) if we want to make Data node directly addressable from outside in the future. 

No concern on making this a private method and use it for DataPointer implementation and only expose Ref(Component)

chasm/tree.go
```diff
@@ -354,6 +354,41 @@ func (n *Node) prepareDataValue(
...
+	if n.valueState == valueStateNeedDeserialize {
+		if err := n.deserialize(nil); err != nil {
+			return fmt.Errorf("failed to deserialize data: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (n *Node) findNode(path []string) *Node {
```
yycptt, line 377: I added a `getNodeByPath` method some time ago.

chasm/tree.go
```diff
@@ -727,8 +771,12 @@ func (n *Node) syncSubField(fieldV reflect.Value, fieldN string, nodePath []stri
 		// Field is not empty but tree node is not set. It means this is a new field, and a node must be created.
 		childNode := newNode(n.nodeBase, n, fieldN)
 
-		if err = assertStructPointer(reflect.TypeOf(internal.value())); err != nil {
-			return
+		switch internal.value().(type) {
+		case []string: // Pointer
```
yycptt, line 775: shall we also check if fieldType is pointer here?

chasm/tree.go
```diff
@@ -919,13 +990,39 @@ func unmarshalProto(
...
-func (n *Node) Ref(
+// RefC implements the CHASM Context interface
+func (n *Node) RefC(
 	component Component,
 ) (ComponentRef, bool) {
-	// TODO: Implement this method.
-	// Currently returning an empty reference to unblock tests.
-	return ComponentRef{}, true
+	// TODO: return error
+	_ = n.syncSubComponents()
```
yycptt, line 998: logging the error should be good enough I guess. The same error will be encountered when closing transaction and get returned?

chasm/test_component_test.go
```diff
@@ -20,12 +20,13 @@ type (
...
-		SubData1          Field[*protoMessageType]
-		SubComponents     Map[string, *TestSubComponent1]
-		PendingActivities Map[int, *TestSubComponent1]
+		ComponentData         *protoMessageType
+		SubComponent1         Field[*TestSubComponent1]
+		SubComponent2         Field[*TestSubComponent2]
+		SubData1              Field[*protoMessageType]
+		SubComponents         Map[string, *TestSubComponent1]
+		PendingActivities     Map[int, *TestSubComponent1]
+		SubComponent11Pointer Field[*TestSubComponent11]
```
yycptt, line 29: I am actually debating with myself if we need another `chasm.Ptr` type. I feel the current approach is more flexible and allow Component author to define only one field and have logic like, if there's already a Component in the tree (say from its parent), just points to it, otherwise create a real one. 

If there's need a for making things more explicit we can add `chasm.Ptr` later and it's essentially a Field with type pointer.

chasm/tree.go
```diff
@@ -727,8 +771,12 @@ func (n *Node) syncSubField(fieldV reflect.Value, fieldN string, nodePath []stri
 		// Field is not empty but tree node is not set. It means this is a new field, and a node must be created.
 		childNode := newNode(n.nodeBase, n, fieldN)
 
-		if err = assertStructPointer(reflect.TypeOf(internal.value())); err != nil {
-			return
+		switch internal.value().(type) {
+		case []string: // Pointer
+		default: // Component | Data
+			if err = assertStructPointer(reflect.TypeOf(internal.value())); err != nil {
```
yycptt, line 777: I think it is the case based on my reading of the code but want to double check: If Component or Data field type, it (`T`) needs to be pointer to a struct. For Pointer field type, `T` can be an interface.

chasm/tree.go
yycptt: But yeah returning error is also a good idea and we can define an ErrComponentNotFound constant.

chasm/test_component_test.go
alexshtin: Yeah, I went through the same thought process. It doesn't bring much troubles and, yes, very convenient for component author. I think, if can hide all the complexity of pointers and really make them work exactly as `Field`, then it should stay as `Field` too. If later, we discover some differences due to some internal limitations, then it needs to be explicitly indicated in the type, i.e. `chasm.Ref`.

chasm/tree.go
alexshtin: I am removing mine, but renaming yours to `findNode`. It is a better name.

chasm/tree.go
alexshtin: For some reason I thought that `fieldType` might be not set there. But it should be always set. Changing to switch on `fieldType` because it is much better.

chasm/tree.go
alexshtin: This is a value of pointer itself. And it must be of `[]string` type. It is set internally, so it is always must be the case.

For other call of the `assertStructPointer`, it asserts on actual type of the value which is passed to `deserialize` method, which is always concrete type, even if the field is defined using interface. You can check `TestFieldInterface`.

yycptt: ok I see. The type the `T` in `Field[T]` can always be an interface. But when creating a new Field, it needs to be a concrete type (i.e. pointer to struct) when creating the field (). 

Then I guess what I want to confirm is: do `Component|DataPointerTo()` have the same requirement? Looks like the answer is yes because the Ref() impl only check `= node.value` and node value is always concrete type.

alexshtin: Yes, the answer is "yes" :-).

In 
```go
func ComponentPointerTo[C Component](
	ctx MutableContext,
	c C,
) (Field[C], error) {
```
`C` can be and interface, but `c` will be always an instance of a concrete type which should match the type of what it points to. If they are mismatch, it can be hard to debug, I guess.

yycptt: Got it. Thanks for confirming.

I guess usually the component using Pointer won't actually care or know the concrete type of the component being pointed to. So the use case I have in mind is the component's constructor will just accept an interface and then create a pointer to the passed in interface value, which means `c` is an interface. 
It's easy to support that I think, just need to check for the try to match the underlying value in Ref().

PR #7775, opened 2025-05-15
Refactor: remove yaml config files from functional tests

alexshtin:
## What changed?
Refactor: remove yaml config files from functional tests. ndc and xdc configs are still there and I will do it in following PRs.

## Why?
It is much more convenient to have test configuration in code than in yaml files.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


tests/acquire_shard_test.go
```diff
@@ -102,7 +103,23 @@ func TestAcquireShard_OwnershipLostErrorSuite(t *testing.T) {
...
-	s.FunctionalTestBase.SetupSuiteWithCluster("testdata/acquire_shard_ownership_lost_error.yaml")
+	s.FunctionalTestBase.SetupSuiteWithCluster(testcore.WithFaultInjectionConfig(
+		config.FaultInjection{
+			Targets: config.FaultInjectionTargets{
+				DataStores: map[config.DataStoreName]config.FaultInjectionDataStoreConfig{
+					config.ShardStoreName: {
+						Methods: map[string]config.FaultInjectionMethodConfig{
+							"UpdateShard": {
+								Errors: map[string]float64{
+									"ShardOwnershipLost": 1.0,
```
dnr, line 114: just thinking out loud...

should we have some kind of builder helper for these to simplify the code? e.g.
```
var fi config.FaultInjectionConfig
fi.Add(config.ShardStoreName, "UpdateShard", "ShardOwnershipLost", 1.0)
// fi.SetSeed(config.ShardStoreName, "UpdateShard", 43)
s.FunctionalTestBase.SetupSuiteWithCluster(testcore.WithFaultInjectionConfig(fi)
```

tests/testcore/test_cluster.go
```diff
@@ -79,15 +79,12 @@ type (
 		WorkerConfig           WorkerConfig
 		ESConfig               *esclient.Config
 		MockAdminClient        map[string]adminservice.AdminServiceClient
-		FaultInjection         config.FaultInjection     `yaml:"faultInjection"`
+		FaultInjection         config.FaultInjection
 		DynamicConfigOverrides map[dynamicconfig.Key]any `yaml:"-"`
```
dnr, line 83:
```suggestion
		DynamicConfigOverrides map[dynamicconfig.Key]any
```

tests/testdata/generic_fault_injection.yaml
```diff
@@ -0,0 +1,31 @@
+targets:
```
dnr, line 1: can't we put this one in code too? and enable it with some flag like `-useGenericFI`. with the helper I suggested above it wouldn't be that much code and it would be easy to edit.

common/config/config.go
```diff
@@ -742,3 +742,66 @@ func (k *KeepAliveServerConfig) GetKeepAliveEnforcementPolicy() keepalive.Enforc
...
 	return defaultConfig
 }
+
+func (fi *FaultInjection) WithError(storeName DataStoreName, methodName, errorName string, probability float64) *FaultInjection {
+	if fi == nil {
+		return nil
+	}
+	m := fi.method(storeName, methodName)
+	m.Errors[errorName] = probability
+	fi.Targets.DataStores[storeName].Methods[methodName] = m
```
dnr, line 752: should we use pointers for FaultInjectionMethodConfig so this can be
```suggestion
	fi.method(storeName, methodName).Errors[errorName] = probability
```
?
kinda weird for `method` to set up maps but then return a copied value

common/config/config.go
```diff
@@ -742,3 +742,66 @@ func (k *KeepAliveServerConfig) GetKeepAliveEnforcementPolicy() keepalive.Enforc
...
+	if store.Methods == nil {
+		store.Methods = map[string]FaultInjectionMethodConfig{}
+	}
+	method, ok := store.Methods[methodName]
+	if !ok {
+		method = FaultInjectionMethodConfig{}
+	}
+	if method.Errors == nil {
+		method.Errors = map[string]float64{}
+	}
```
dnr, line 783:
```suggestion
	if !ok {
		method = FaultInjectionMethodConfig{Errors: map[string]float64{}}
	}
```
equivalent?

dnr (approved):
Kill all the yaml :dagger:

common/config/config.go
alexshtin: I gave it a quick look. It will require fixes on caller side which is autogenerated. I might do it separately.

PR #7779, opened 2025-05-15
fixing all occurrences of cryptic errors in the versioning entity wfs

Shivs11:
## What changed?
- WISOTT
- Replace individual error types that fall under `FailedPreCondition` and just check for these in the deployment client. This makes reading code in the deployment client simpler. 
- Reduce `Error` log spamming by the deployment client when it's actually a `PreconditionFailure`. The deployment client should only log `error` when something actually goes wrong in our internal entity workflows.

## Why?
- So that users don't see any cryptic error messages while using this feature.

## How did you test it?
- [x] built
- [ ] added new unit test(s)


service/worker/workerdeployment/workflow.go
```diff
@@ -301,7 +301,7 @@ func (d *WorkflowRunner) handleRegisterWorker(ctx workflow.Context, args *deploy
 		if errors.As(err, &appError) {
 			if appError.Type() == errMaxTaskQueuesInVersionType {
 				return temporal.NewApplicationError(
-					fmt.Sprintf("maximum number of task queues (%d) have been registered in deployment", args.MaxTaskQueues),
+					fmt.Sprintf("cannot add task queue since maximum number of task queues (%d) have been registered in deployment", args.MaxTaskQueues),
```
ShahabT, line 304: let's mention tq name?

service/worker/workerdeployment/workflow.go
```diff
@@ -257,7 +257,7 @@ func (d *WorkflowRunner) addVersionToWorkerDeployment(ctx workflow.Context, args
 	if len(d.State.Versions) >= maxVersions {
 		err := d.tryDeleteVersion(ctx)
 		if err != nil {
-			return temporal.NewApplicationError(fmt.Sprintf("cannot add version, already at max versions %d", maxVersions), errTooManyVersions)
+			return temporal.NewApplicationError(fmt.Sprintf("cannot add version since maximum number of versions (%d) have been registered in the deployment", maxVersions), errTooManyVersions)
```
ShahabT, line 260: let's mention version in the error message?

service/worker/workerdeployment/workflow.go
```diff
@@ -334,16 +334,16 @@ func (d *WorkflowRunner) validateStateBeforeAcceptingRampingUpdate(args *deploym
...
 	if args.Version == d.State.RoutingConfig.CurrentVersion {
 		d.logger.Info("version can't be set to ramping since it is already current")
-		return temporal.NewApplicationError("version can't be set to ramping since it is already current", errVersionAlreadyCurrentType)
+		return temporal.NewApplicationError(fmt.Sprintf("ramping version %s is already current", args.Version), errFailedPrecondition)
 	}
 
 	if _, ok := d.State.Versions[args.Version]; !ok && args.Version != "" && args.Version != worker_versioning.UnversionedVersionId {
 		d.logger.Info("version not found in deployment")
-		return temporal.NewApplicationError("version not found in deployment", errVersionNotFound)
+		return temporal.NewApplicationError("version not found in deployment", errFailedPrecondition)
```
ShahabT, line 346: same: mentioning version string

service/worker/workerdeployment/workflow.go
```diff
@@ -569,11 +569,11 @@ func (d *WorkflowRunner) validateStateBeforeAcceptingSetCurrent(args *deployment
 		return temporal.NewApplicationError("no change", errNoChangeType, d.State.ConflictToken)
 	}
 	if args.ConflictToken != nil && !bytes.Equal(args.ConflictToken, d.State.ConflictToken) {
-		return temporal.NewApplicationError("conflict token mismatch", errConflictTokenMismatchType)
+		return temporal.NewApplicationError("conflict token mismatch", errFailedPrecondition)
 	}
 	if _, ok := d.State.Versions[args.Version]; !ok && args.Version != worker_versioning.UnversionedVersionId {
 		d.logger.Info("version not found in deployment")
-		return temporal.NewApplicationError("version not found in deployment", errVersionNotFound)
+		return temporal.NewApplicationError(fmt.Sprintf("version %s not found in deployment", args.Version), errFailedPrecondition)
```
ShahabT, line 576: I think users would be calling this mostly by passing separate deployment name and build ID. so if we return a error with verison string they can be confused. Maybe we should say something like "version with build ID xyz not found in deployment abc."

service/worker/workerdeployment/workflow.go
```diff
@@ -334,16 +334,16 @@ func (d *WorkflowRunner) validateStateBeforeAcceptingRampingUpdate(args *deploym
 	}
 
 	if args.ConflictToken != nil && !bytes.Equal(args.ConflictToken, d.State.ConflictToken) {
-		return temporal.NewApplicationError("conflict token mismatch", errConflictTokenMismatchType)
+		return temporal.NewApplicationError("conflict token mismatch", errFailedPrecondition)
 	}
 	if args.Version == d.State.RoutingConfig.CurrentVersion {
 		d.logger.Info("version can't be set to ramping since it is already current")
-		return temporal.NewApplicationError("version can't be set to ramping since it is already current", errVersionAlreadyCurrentType)
+		return temporal.NewApplicationError(fmt.Sprintf("ramping version %s is already current", args.Version), errFailedPrecondition)
```
ShahabT, line 341:
```suggestion
		return temporal.NewApplicationError(fmt.Sprintf("requested ramping version %s is already current", args.Version), errFailedPrecondition)
```

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

service/worker/workerdeployment/workflow.go
Shivs11: I think we should do this after public-preview since right now, the update args have `Version` as an attribute which shall eventually change to becoming `BuildID`. 

I don't wanna parse the version string here so if it's ok, can we push this ahead?

ShahabT (approved):

service/worker/workerdeployment/workflow.go
nareshrajkumar866: return temporal.NewApplicationError(fmt.Sprintf("cannot add version %v: maximum number of versions (%d) have been registered in the deployment", newVersion, maxVersions), errTooManyVersions)

PR #7784, opened 2025-05-15
pass target cluster to all migration replication task

zigeH:
## What changed?
pass target cluster to all migration replication task 

## Why?
passive migration will support migrating open and closed wf. So, we would like to pass target cluster to all migration replication task 

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
no potential risk


xwduan (approved):

yux0:
Do you also need to wire up task_serializer?

zigeH:
> Do you also need to wire up task_serializer?

fixed

PR #7787, opened 2025-05-16
Fixing flake caused by the auto-restart versioning feature

Shivs11:
## What changed?
increasing timeouts for breaking test + making the deploymentClient retry on the error returned when an update is admitted but the workflow is closing.

## Why?
- trying to put a band-aid over the breaking func test

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/service/worker/workerdeployment/client.go
+++ b/service/worker/workerdeployment/client.go
@@ -956,8 +956,8 @@ func (d *ClientImpl) update(
 
 	policy := backoff.NewExponentialRetryPolicy(100 * time.Millisecond)
 	isRetryable := func(err error) bool {
-		// Updates admitted as the workflow is closing may result in a ResourceExhausted error, which is considered retryable.
-		return errors.Is(err, errRetry) || (errors.Is(err, consts.ErrWorkflowClosing))
+		// All updates that are admitted as the workflow is closing are considered retryable.
+		return errors.Is(err, errRetry) || err.Error() == consts.ErrWorkflowClosing.Error()
 	}
 
 	var outcome *updatepb.Outcome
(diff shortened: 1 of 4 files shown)


service/worker/workerdeployment/util.go
```diff
@@ -74,6 +74,7 @@ const (
 
 	ErrRampingVersionDoesNotHaveAllTaskQueues = "proposed ramping version is missing active task queues from the current version; these would become unversioned if it is set as the ramping version"
 	ErrCurrentVersionDoesNotHaveAllTaskQueues = "proposed current version is missing active task queues from the current version; these would become unversioned if it is set as the current version"
+	ErrWorkflowClosing                        = "workflow operation can not be applied because workflow is closing"
```
ShahabT, line 77: is it possible to not redefine this and just use the existing constant? https://github.com/temporalio/temporal/blob/91893f10641bae49f8181235c5e93f61a3c826a8/service/history/consts/const.go#L63

Shivs11: sorry heh, did this in a hurry :( 
agreed!

service/worker/workerdeployment/client.go
```diff
@@ -1206,7 +1208,8 @@ func (d *ClientImpl) updateWithStart(
 
 	policy := backoff.NewExponentialRetryPolicy(100 * time.Millisecond)
 	isRetryable := func(err error) bool {
-		return errors.Is(err, errRetry)
+		// Updates admitted as the workflow is closing may result in a ResourceExhausted error, which is considered retryable.
+		return errors.Is(err, errRetry) || (errors.As(err, new(*serviceerror.ResourceExhausted)) && err.Error() == consts.ErrWorkflowClosing.Error())
```
ShahabT, line 1212: if something like below works it'd be preferable because I don't think we care about the ResourceExhausted part.
```suggestion
		return errors.Is(err, errRetry) || errors.Is(err, consts.ErrWorkflowClosing)
```

ShahabT (approved):

PR #7789, opened 2025-05-16
Functional tests: remove ndc/xdc yaml config files

alexshtin:
## What changed?
Functional tests: remove ndc/xdc yaml config files.

## Why?
Last 2 yaml config files for functional test are replaced with corresponding code blocks because it is more convenient.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


tests/ndc/cluster_config.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func clustersConfig() []*testcore.TestClusterConfig {
+	// TODO (alex): I don't know how many of these settings are really needed.
+
+	clusterInformation := map[string]cluster.ClusterInformation{
+		"cluster-a": {
+			Enabled:                true,
+			InitialFailoverVersion: 1,
+			RPCAddress:             "nowhere1:7134",
```
dnr, line 15: I think these are either not needed now, or can be dropped with a tiny bit of work. but no need to do it now.

alexshtin: Simple removal doesn't work, I checked.

PR #7796, opened 2025-05-20
Fix unversioned to versioned transition bug

ShahabT:
## What changed?
The effective behavior did not consider an ongoing transition (unlike effective version) and this caused some tasks getting wrong Directive and be unwontedly dropped.

Plus some other improvements.

## Why?
above

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


service/worker/workerdeployment/workflow.go
```diff
@@ -667,6 +654,22 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
 		d.setDrainageStatus(prevCurrentVersion, enumspb.VERSION_DRAINAGE_STATUS_DRAINING, updateTime)
 	}
 
+	//nolint:staticcheck // deprecated stuff will be cleaned
+	if newCurrentVersion == worker_versioning.UnversionedVersionId && d.State.RoutingConfig.RampingVersion == worker_versioning.UnversionedVersionId {
```
ShahabT, line 658: moving this _after_ the other version becomes current, otherwise, there would be a small discontinuity.

service/history/workflow/mutable_state_impl.go
```diff
@@ -8085,6 +8087,10 @@ func (ms *MutableStateImpl) StartDeploymentTransition(deployment *deploymentpb.D
 		ms.GetExecutionInfo().VersioningInfo = versioningInfo
 	}
 
+	if ms.GetEffectiveDeployment().Equal(deployment) {
+		return serviceerror.NewInternal("start transition should receive a version different from effective version")
```
carlydf, line 8091: can you put the strings of each version in this error?

carlydf: (or do you expect it to be never hit, in which case maybe that's not worth the additional CI run)

service/history/workflow/mutable_state_impl.go
```diff
@@ -8103,17 +8109,20 @@ func (ms *MutableStateImpl) StartDeploymentTransition(deployment *deploymentpb.D
...
-			return err
+		if ms.GetPendingWorkflowTask().Type != enumsspb.WORKFLOW_TASK_TYPE_SPECULATIVE {
+			// sticky queue was just cleared, so the following call only generates
+			// a WorkflowTask, not a WorkflowTaskTimeoutTask.
+			err := ms.taskGenerator.GenerateScheduleWorkflowTaskTasks(ms.GetPendingWorkflowTask().ScheduledEventID)
+			if err != nil {
+				return err
+			}
+		} else {
+			ms.logInfo("start transition did not reschedule pending speculative task")
```
carlydf, line 8125: I think it would be nice to have the version string in this log line as well, but not critical

service/matching/physical_task_queue_manager.go
```diff
@@ -642,8 +642,10 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeploymentVersion(
 		var errTooMany workerdeployment.ErrMaxTaskQueuesInDeployment
 		if errors.As(err, &errTooMany) {
 			c.deploymentRegisterError = errTooMany
+			return errTooMany
 		}
-		return err
+		c.logger.Error("error while registering version", tag.Error(err))
```
carlydf, line 647: also would be nice to have the version string in this error message, although maybe it's already in the wrapped error

service/worker/workerdeployment/workflow.go
carlydf: could this cause an NDE because the order of the `syncVersion` events could be flipped in some cases?

carlydf: Thinking about it more, it would only cause an NDE if a user has previously ramped to unversioned and then set current to unversioned. I think it is very unlikely that a pre-release user has done that, so I'm fine with introducing this change without patching

carlydf (approved):
approved with some minor comments about nice to have info in the log lines

PR #7798, opened 2025-05-22
Worker Version workflow to Continue-as-New after processing ongoing updates

ShahabT:
## What changed?
Similar to the Deployment workflow, now the Version workflow CaNs when there is no pending work to do and state is dirty.

Also, merged the drainage check logic into the Version workflow, not using a child wf anymore.

## Why?
This is to keep wfs in a clean history as mush as possible so there is less concern about forward and backwards compatibility when changing server versions.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


service/worker/workerdeployment/replaytester/replay_test.go
```diff
@@ -1,62 +1,45 @@
...
-}
+//func TestReplays(t *testing.T) {
+//	replayer := worker.NewWorkflowReplayer()
+//	drainageWorkflow := func(ctx workflow.Context, args *deploymentspb.DrainageWorkflowArgs) error {
+//		refreshIntervalGetter := func() any {
+//			return time.Minute
+//		}
+//		visibilityGracePeriodGetter := func() any {
+//			return time.Minute
+//		}
```
Shivs11, line 12: I think rather than removing this, we should probably remove the files that are not relevant anymore - you might have had the same idea but just pasting it here

service/worker/workerdeployment/version_workflow.go
```diff
@@ -708,3 +700,63 @@ func (d *VersionWorkflowRunner) syncSummary(ctx workflow.Context) {
...
+		interval, err = getSafeDurationConfig(ctx, "getDrainageRefreshInterval", d.unsafeRefreshIntervalGetter, defaultVisibilityRefresh)
+	}
+	if err != nil {
+		d.logger.Error("could not calculate drainage refresh interval", tag.Error(err))
+		return
+	}
+	timeSinceLastRefresh := workflow.Now(ctx).Sub(drainage.LastCheckedTime.AsTime())
+	if err = workflow.Sleep(ctx, interval-timeSinceLastRefresh); err != nil {
+		d.logger.Error("error while trying to sleep", tag.Error(err))
+		return
```
Shivs11, line 732: I think we should maybe have a safeguard for this line by checking if it's non-negative:
`interval-timeSinceLastRefresh`

Sure, you could argue that interval is always one of 3/5 minutes and `timeSinceLastRefresh` will always be less than those values in the happy path - but I fear that if a version were to become draining and something were to go wrong, if we restart this workflow and the value of `timeSinceLastRefresh` were to be greater than `interval`, things could maybe get messy if we wanted to sleep.

However, if the logic was to account for the above case by not sleeping at all on some minimum value, I guess this could work - just wanted to put this out and verify if this is something you had thought of

service/worker/workerdeployment/version_workflow.go
```diff
@@ -708,3 +700,63 @@ func (d *VersionWorkflowRunner) syncSummary(ctx workflow.Context) {
...
+		d.VersionState.Version,
+	).Get(ctx, &newInfo)
+	if err != nil {
+		d.logger.Error("could not get version drainage status", tag.Error(err))
+		return
+	}
+
+	if d.VersionState.DrainageInfo == nil {
+		d.VersionState.DrainageInfo = &deploymentpb.VersionDrainageInfo{}
+	}
```
Shivs11, line 750: is this required? AFAIK, this function is only executed once the version workflow enters `draining` state right? In other words, this structure would already be filled when this function is running

service/worker/workerdeployment/version_workflow.go
```diff
@@ -13,54 +14,82 @@ import (
...
-		forceCAN = true
+		d.forceCAN = true
 	})
-	selector.AddReceive(drainageStatusSignalChannel, func(c workflow.ReceiveChannel, more bool) {
-		var newInfo *deploymentpb.VersionDrainageInfo
+	d.signalHandler.signalSelector.AddReceive(drainageStatusSignalChannel, func(c workflow.ReceiveChannel, more bool) {
+		d.signalHandler.processingSignals++
+		defer func() { d.signalHandler.processingSignals-- }()
 
+		var newInfo *deploymentpb.VersionDrainageInfo
```
Shivs11, line 92: this signal check can be removed now right? earlier on, this was required since we had the drainage workflow signalling back to the version workflow the drainage status of the version workflow. now, since the version workflow is conducting these checks periodically within itself, i don't think these signals are required

service/worker/workerdeployment/version_activities.go
```diff
@@ -172,3 +174,17 @@ func (a *VersionActivities) AddVersionToWorkerDeployment(ctx context.Context, in
...
+	logger := activity.GetLogger(ctx)
+	response, err := a.deploymentClient.GetVersionDrainageStatus(ctx, a.namespace, worker_versioning.WorkerDeploymentVersionToString(version))
+	if err != nil {
+		logger.Error("error counting workflows for drainage status", "error", err)
+		return nil, err
+	}
+	return &deploymentpb.VersionDrainageInfo{
+		Status:          response,
+		LastChangedTime: nil, // ignored; whether Status changed will be evaluated by the receiver
+		LastCheckedTime: timestamppb.Now(),
```
Shivs11, line 188: maybe we should only just return the `Status` here and not the other two timestamps (instead of returning the value for one of them and having the comment you have placed)

The caller only needs to see the status from the activity. `LastCheckedTime` will anyways be `timestamppb.Now()` and the logic for `LastChangedTime` is correct on the caller side

service/worker/workerdeployment/client.go
```diff
@@ -700,6 +700,8 @@ func (d *ClientImpl) DeleteWorkerDeploymentVersion(
 	if failure := outcome.GetFailure(); failure != nil {
 		if failure.GetApplicationFailureInfo().GetType() == errVersionNotFound {
 			return nil
+		} else if failure.Message == ErrVersionIsCurrentOrRamping {
+			return serviceerror.NewFailedPrecondition(ErrVersionIsCurrentOrRamping) // non-retryable error to stop multiple activity attempts
```
Shivs11, line 704: based on my comment below, if we move this check into the validator, we might need to tweak this a little bit - moreover, rather than checking for a specific error message like `ErrVersionIsCurrentOrRamping`, we can just do:

`failure.GetApplicationFailureInfo().GetType() == errFailedPrecondition` and return the message stored in the error

service/matching/matching_engine.go
```diff
@@ -1137,10 +1137,28 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		timeSinceLastFanOut := rootPM.TimeSinceLastFanOut()
 		lastFanOutTTL := tqConfig.TaskQueueInfoByBuildIdTTL()
 
-		// TODO bug fix: We cache the same map regardless of VersionSelection or TaskQueueTypes, so if someone queries the Activity Task Queue type of this task queue name, we cache that result and return it even if the next DescribeTQ call is about the WF TQ
+		// TODO bug fix: We cache the last response for each build ID. timeSinceLastFanOut is the last fan out time, that means some enteries in the cache can be more stale if
+		// user is calling this API back-to-back but with different version selection.
+		cacheIsFresh := timeSinceLastFanOut <= lastFanOutTTL
+		missingItemsInCache := false
```
Shivs11, line 1143: these changes don't seem to be related to the version-wf CAN'ing right? any specific reason you included these here or did I miss something?

Shivs11 (changes requested):

service/matching/matching_engine.go
Shivs11: anyways, i think what you have added here is right and it makes sense to me

service/matching/physical_task_queue_manager.go
```diff
@@ -428,6 +429,9 @@ func (c *physicalTaskQueueManagerImpl) DispatchNexusTask(
 }
 
 func (c *physicalTaskQueueManagerImpl) UpdatePollerInfo(id pollerIdentity, pollMetadata *pollMetadata) {
+	if c.queue.Version().IsVersioned() {
+		fmt.Printf("poll info updated in %v\n", c.queue.Version())
+	}
```
Shivs11, line 434: stale comment

ShahabT: yeah fixing it in the next PR. if I had to make other changes to this PR I'll clean this too, otherwise, I clean it in the next PR to save time.

service/worker/workerdeployment/version_activities.go
Shivs11: sure, can agree with that

service/worker/workerdeployment/client.go
Shivs11: the reason it wasn't is because the next check is looking for activity errors specifically - note, the check below is different from what i suggested above

PR #7807, opened 2025-05-26
enable versioning configs for pre-release

Shivs11:
## What changed?
- enable versioning configs for pre-release
- deprecate pre-release versioning configs as well since we should not be supporting them! (shall open a corresponding API PR to remove these API's from the API repo)

## Why?
- versioning

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
- Folks that have been using versioning pre-release (none afaik) will have trouble?


Shivs11:
closing this for now since we don't wanna remove the older API's right now - the PR can be reopened as and when we decide to

PR #7809, opened 2025-05-26
Add Dynamic Config for safe keepalive rollout

xwduan:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -235,6 +235,16 @@ these warning are not emitted if the value is set to 0 or less`,
 		`OperatorRPSRatio is the percentage of the rate limit provided to priority rate limiters that should be used for
 operator API calls (highest priority). Should be >0.0 and <= 1.0 (defaults to 20% if not specified)`,
 	)
+	EnableInterNodeServerKeepAlive = NewGlobalBoolSetting(
+		"system.enableInterNodeServerKeepAlive",
+		false,
+		`enableInterNodeServerKeepAlive is the config to enable keep alive for inter-node connections on server side.`,
+	)
+	EnableInterNodeClientKeepAlive = NewGlobalBoolSetting(
+		"system.enableInterNodeClientKeepAlive",
+		false,
+		`enableInterNodeClientKeepAlive is the config to enable keep alive for inter-node connections on client side.`,
+	)
 	PersistenceQPSBurstRatio = NewGlobalFloatSetting(
 		"system.persistenceQPSBurstRatio",
 		1.0,
--- a/common/resource/fx.go
+++ b/common/resource/fx.go
@@ -41,6 +41,7 @@ import (
 	"go.temporal.io/server/common/searchattribute"
 	"go.temporal.io/server/common/telemetry"
 	"go.temporal.io/server/common/testing/testhooks"
+	"go.temporal.io/server/service/history/configs"
 	"go.uber.org/fx"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health"
@@ -73,6 +74,7 @@ type (
 // See LifetimeHooksModule for detail
 var Module = fx.Options(
 	persistenceClient.Module,
+	dynamicconfig.Module,
 	fx.Provide(HostNameProvider),
 	fx.Provide(TimeSourceProvider),
 	cluster.MetadataLifetimeHooksModule,
@@ -100,6 +102,7 @@ var Module = fx.Options(
 	fx.Provide(FrontendHTTPClientCacheProvider),
 	fx.Provide(PersistenceConfigProvider),
 	fx.Provide(health.NewServer),
+	fx.Provide(ConfigProvider),
 	deadlock.Module,
 	config.Module,
 	testhooks.Module,
@@ -192,6 +195,16 @@ func NamespaceRegistryProvider(
 		dynamicconfig.ForceSearchAttributesCacheRefreshOnRead.Get(dynamicCollection),
 		metricsHandler,
 		logger,
+	)
+}
+
+func ConfigProvider(
+	dc *dynamicconfig.Collection,
+	persistenceConfig *config.Persistence,
+) *configs.Config {
+	return configs.NewConfig(
+		dc,
+		persistenceConfig.NumHistoryShards,
 	)
 }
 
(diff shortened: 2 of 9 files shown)


common/resource/fx.go
```diff
@@ -41,6 +41,7 @@ import (
 	"go.temporal.io/server/common/searchattribute"
 	"go.temporal.io/server/common/telemetry"
 	"go.temporal.io/server/common/testing/testhooks"
+	"go.temporal.io/server/service/history/configs"
```
dnr, line 44: This is in `common`, shared by all services. It shouldn't have a dependency on history service.

common/dynamicconfig/constants.go
```diff
@@ -235,6 +235,16 @@ these warning are not emitted if the value is set to 0 or less`,
 		`OperatorRPSRatio is the percentage of the rate limit provided to priority rate limiters that should be used for
 operator API calls (highest priority). Should be >0.0 and <= 1.0 (defaults to 20% if not specified)`,
 	)
+	EnableInterNodeServerKeepAlive = NewGlobalBoolSetting(
+		"system.enableInterNodeServerKeepAlive",
+		false,
+		`enableInterNodeServerKeepAlive is the config to enable keep alive for inter-node connections on server side.`,
```
dnr, line 241:
```suggestion
	EnableInternodeServerKeepalive = NewGlobalBoolSetting(
		"system.enableInternodeServerKeepalive",
		false,
		`enableInternodeServerKeepalive is the config to enable keep alive for inter-node connections on server side.`,
```

common/dynamicconfig/constants.go
```diff
@@ -235,6 +235,16 @@ these warning are not emitted if the value is set to 0 or less`,
...
+	)
+	EnableInterNodeClientKeepAlive = NewGlobalBoolSetting(
+		"system.enableInterNodeClientKeepAlive",
+		false,
+		`enableInterNodeClientKeepAlive is the config to enable keep alive for inter-node connections on client side.`,
+	)
```
dnr, line 247:
```suggestion
	EnableInternodeClientKeepalive = NewGlobalBoolSetting(
		"system.enableInternodeClientKeepalive",
		false,
		`enableInternodeClientKeepalive is the config to enable keep alive for inter-node connections on client side.`,
	)
```

common/resource/fx.go
```diff
@@ -339,6 +352,7 @@ func RPCFactoryProvider(
 	resolver *membership.GRPCResolver,
 	tracingStatsHandler telemetry.ClientStatsHandler,
 	monitor membership.Monitor,
+	dynamicConfig *configs.Config,
```
dnr, line 355: why do we need history service configs here? this rpc factory is for all services

common/rpc/rpc.go
```diff
@@ -251,6 +258,14 @@ func (d *RPCFactory) dial(hostName string, tlsClientConfig *tls.Config, dialOpti
...
+	if !d.dynamicConfig.EnableInterNodeClientKeepAlive() {
+		// default keepalive settings for clients
+		return grpc.WithKeepaliveParams(keepalive.ClientParameters{
+			Time:                time.Duration(math.MaxInt64),
+			Timeout:             20 * time.Second,
+			PermitWithoutStream: false,
+		})
+	}
 	serviceConfig := d.config.Services[string(serviceName)]
 	return grpc.WithKeepaliveParams(serviceConfig.RPC.ClientConnectionConfig.GetKeepAliveClientParameters())
```
dnr, line 270: maybe write this reversed, like:
```suggestion
	// default keepalive settings for clients
	params := keepalive.ClientParameters{
		Time:                time.Duration(math.MaxInt64),
		Timeout:             20 * time.Second,
		PermitWithoutStream: false,
	}
	if d.dynamicConfig.EnableInterNodeClientKeepAlive() {
		serviceConfig := d.config.Services[string(serviceName)]
		params = serviceConfig.RPC.ClientConnectionConfig.GetKeepAliveClientParameters()
	}
	return grpc.WithKeepaliveParams(params)
```

common/resource/fx.go
```diff
@@ -195,6 +198,16 @@ func NamespaceRegistryProvider(
 	)
 }
 
+func ConfigProvider(
```
dnr, line 201: If this is here (and it shouldn't be), it needs a new name, "ConfigProvider" for something that provides _history_ configs is too generic outside of the context of the history service.

dnr (commented):
Do we really need this? Can't we roll out with the static config changes?

On the server side, the dynamic config isn't dynamic at all, it requires a process restart. On the client side, it would only apply to new connections.

common/resource/fx.go
```diff
@@ -360,7 +364,11 @@ func RPCFactoryProvider(
 		frontendTLSConfig,
 		options,
 		monitor,
-	), nil
+	)
+	factory.EnableInternodeServerKeepalive = enableServerKeepalive
+	factory.EnableInternodeClientKeepalive = enableClientKeepalive
+	logger.Info(fmt.Sprintf("RPC factory created. enableServerKeepalive: %v, enableClientKeepalive: %v", enableServerKeepalive, enableClientKeepalive))
```
dnr, line 370: do we really need this log line? I'd rather drop it unless you think it is very important for debugging (and then maybe debug level?)

common/resource/fx.go
```diff
@@ -360,7 +364,11 @@ func RPCFactoryProvider(
 		frontendTLSConfig,
 		options,
 		monitor,
-	), nil
+	)
+	factory.EnableInternodeServerKeepalive = enableServerKeepalive
+	factory.EnableInternodeClientKeepalive = enableClientKeepalive
```
dnr, line 369: I'm guessing you set these externally instead of passing them to NewFactory to avoid having to change all the tests? that's kind of messy but it's okay since it's just temporary.

common/resource/fx.go
xwduan: This line is temporary and I will remove it when fully rolled out. I want to make sure the dc change is taking effect when rolling out. I think debug level make sense.

common/resource/fx.go
xwduan: yeah, and when removing, we don't have to deal with signature change again.

PR #7818, opened 2025-05-27
Fix nil pointer in user data replication

ShahabT:
## What changed?
ApplyTaskQueueUserDataReplicationEvent was not handling the case when the replication data only contains Versioning V3 config.

## Why?
Above

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None


carlydf (approved):
this makes sense to me -- sounds like you're adding tests

I'm assuming that `mergedUserData` is always non-nil in these two places?

dnr (approved):

PR #7822, opened 2025-05-28
Versioning Metrics: pt1

Shivs11:
## What changed?
- adding helpful metrics for queries so that we can understand customer impact

Metrics added in this PR are:

(Directly related to entity workflows)
- WorkerDeploymentCreated
- WorkerDeploymentVersionCreated
- WorkerDeploymentVersionCreatedManagedByController
- WorkerDeploymentVersionVisibilityQueryCount

(Not directly related to entity workflows)
- WorkflowResetCount
- WorkflowQuerySuccessCount
- WorkflowQueryTimeoutCount
- WorkflowQueryFailureCount
- WorkflowTasksCompleted

## Why?
- vis

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -62,7 +62,7 @@ const (
 	queryTypeOpenSessions     = "__open_sessions"
 	queryTypeWorkflowMetadata = "__temporal_workflow_metadata"
 	queryTypeUserDefined      = "__user_defined"
-	
+
 	newRun      = "new"
 	existingRun = "existing"
 	childRun    = "child"
@@ -415,7 +415,6 @@ func DestinationTag(value string) Tag {
 	}
 }
 
-<<<<<<< HEAD
 func VersioningBehaviorTag(behavior enumspb.VersioningBehavior) Tag {
 	return &tagImpl{versioningBehavior, behavior.String()}
 }
@@ -430,7 +429,8 @@ func QueryTypeTag(queryType string) Tag {
 	}
 	// group all user defined queries into a single tag value
 	return &tagImpl{key: queryType, value: queryTypeUserDefined}
-=======
+}
+
 func VersioningBehaviorBeforeOverrideTag(behavior enumspb.VersioningBehavior) Tag {
 	return &tagImpl{key: behaviorBefore, value: behavior.String()}
 }
@@ -476,5 +476,4 @@ func ToUnversionedTag(version string) Tag {
 		return &tagImpl{key: toUnversioned, value: trueValue}
 	}
 	return &tagImpl{key: toUnversioned, value: falseValue}
->>>>>>> main
 }


common/metrics/metric_defs.go
```diff
@@ -1180,6 +1182,14 @@ var (
 		WithDescription("The number of schedule actions that failed to start"),
 	)
 
+	// Worker Versioning
+	WorkerDeploymentCreated                           = NewCounterDef("worker_deployment_created")
+	WorkerDeploymentVersionCreated                    = NewCounterDef("worker_deployment_version_created")
+	WorkerDeploymentVersionCreatedManagedByController = NewCounterDef("worker_deployment_version_created_managed_by_controller")
+	WorkflowResetCount                                = NewCounterDef("workflow_reset_count")
+	WorkflowQuerySuccessCount                         = NewCounterDef("workflow_query_count")
+	WorkflowQueryFailureCount                         = NewCounterDef("workflow_query_count")
```
Shivs11, line 1191: these may not necessarily be part of "Worker Versioning" but I didn't bother too much - lmk if someone thinks differently

service/history/api/queryworkflow/api.go
```diff
@@ -221,12 +239,31 @@ func Invoke(
...
+			return nil, err
 		default:
 			metrics.QueryRegistryInvalidStateCount.With(scope).Record(1)
 			return nil, consts.ErrQueryEnteredInvalidState
 		}
 	case <-ctx.Done():
+		msResp, err := api.GetMutableState(ctx, shardContext, workflowKey, workflowConsistencyChecker)
+		if err != nil {
+			return nil, err
+		}
```
Shivs11, line 259: note: the reason why this call was not done outside the select case is because I *think* since we give up the MS lock earlier, MS could technically change and its probably safest to fetch the lock back and get the latest MS, so we get the right answer.

ShahabT: I'm not sure this call is free, it'd be great if we don't increase the frequency of the call.

Is it possible to calculate the behavior from `mutableState` before we release the lock in line 182 and just use that behavior in these metrics so we don't have to call GetMutableState?

(theoretically, it's possible that the wf's behavior changes in between these two places, but we do not care for such rare case for this metric).

service/history/api/queryworkflow/api.go
```diff
@@ -345,13 +390,37 @@ func queryDirectlyThroughMatching(
 	nonStickyStartTime := time.Now().UTC()
 	matchingResp, err := matchingClient.QueryWorkflow(ctx, nonStickyMatchingRequest)
 	metrics.DirectQueryDispatchNonStickyLatency.With(metricsHandler).Record(time.Since(nonStickyStartTime))
-	if err != nil {
```
ShahabT, line 348: is this removed on purpose?

service/history/api/queryworkflow/api.go
```diff
@@ -345,13 +390,37 @@ func queryDirectlyThroughMatching(
...
+	if err != nil {
+		if common.IsContextDeadlineExceededErr(err) {
+			metrics.WorkflowQueryTimeoutCount.With(metricsHandler).Record(1, commonTags...)
+		} else {
+			metrics.WorkflowQueryFailureCount.With(metricsHandler).Record(1, commonTags...)
+		}
+		return
+	}
+
+	metrics.WorkflowQuerySuccessCount.With(metricsHandler).Record(1, commonTags...)
```
ShahabT, line 425: Nit: just slightly easier to read:
```suggestion
	if err == nil {
	        metrics.WorkflowQuerySuccessCount.With(metricsHandler).Record(1, commonTags...)
	} else  if common.IsContextDeadlineExceededErr(err) {
		metrics.WorkflowQueryTimeoutCount.With(metricsHandler).Record(1, commonTags...)
	} else {
		metrics.WorkflowQueryFailureCount.With(metricsHandler).Record(1, commonTags...)
	}
```

service/history/timer_queue_active_task_executor.go
```diff
@@ -856,6 +862,8 @@ func (t *timerQueueActiveTaskExecutor) emitTimeoutMetricScopeWithNamespaceTag(
 	namespaceID namespace.ID,
 	operation string,
 	timerType enumspb.TimeoutType,
+	effectiveVersioningBehavior enumspb.VersioningBehavior,
+	firstAttempt int32,
```
ShahabT, line 866:
```suggestion
	taskAttempt int32,
```

service/history/api/queryworkflow/api.go
```diff
@@ -179,6 +179,12 @@ func Invoke(
 	}
 	queryID, completionCh := queryReg.BufferQuery(req.GetQuery())
 	defer queryReg.RemoveQuery(queryID)
+
+	msResp, err := api.GetMutableState(ctx, shardContext, workflowKey, workflowConsistencyChecker)
```
ShahabT, line 183: we have fresh `mutableState`. just use that, no need to make the call.

service/history/api/queryworkflow/api.go
```diff
@@ -191,26 +198,38 @@ func Invoke(
...
+					req.GetQuery().GetQueryType(),
+					err,
+				)
+				return nil, err
 			default:
 				metrics.QueryRegistryInvalidStateCount.With(scope).Record(1)
 				return nil, consts.ErrQueryEnteredInvalidState
 			}
 		case workflow.QueryCompletionTypeUnblocked:
-			msResp, err := api.GetMutableState(ctx, shardContext, workflowKey, workflowConsistencyChecker)
```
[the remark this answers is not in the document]
ShahabT: why not? it can pass `workflow.GetEffectiveVersioningBehavior(msResp.GetVersioningInfo())` instead of the old `behavior`

ShahabT (approved):

PR #7825, opened 2025-05-28
Versioning Metrics pt2: DeploymentTransition + VersioningOverride

Shivs11:
## What changed?
- metrics for DeploymentTransition + VersioningOverride

## Why?
- versioning-0.32

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- a little new to this side of the code involving versioning so a bit scared about doing things wrong, hopefully I am not

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -409,23 +409,29 @@ func VersioningBehaviorAfterOverrideTag(behavior enumspb.VersioningBehavior) Tag
 	return &tagImpl{key: behaviorAfter, value: behavior.String()}
 }
 
+// RunInitiatorTag creates a tag indicating how a workflow run was initiated.
+// It handles both new workflow runs and continuations from previous runs.
+// When attributes is nil (e.g. during AddWorkflowExecutionOptionsUpdatedEvent),
+// it returns a tag indicating an existing run.
 func RunInitiatorTag(prevRunID string, attributes *historypb.WorkflowExecutionStartedEventAttributes) Tag {
-	// UpdateWorkflowExecutionOptions API is the caller
 	if attributes == nil {
 		return &tagImpl{key: runInitiator, value: existingRun}
 	}
 
-	// StartWorkflowExecution API is the caller
 	if prevRunID == "" {
 		return &tagImpl{key: runInitiator, value: newRun}
-	} else if attributes.GetInitiator() == enumspb.CONTINUE_AS_NEW_INITIATOR_WORKFLOW {
-		return &tagImpl{key: runInitiator, value: childRun}
-	} else if attributes.GetInitiator() == enumspb.CONTINUE_AS_NEW_INITIATOR_RETRY {
-		return &tagImpl{key: runInitiator, value: retryRun}
-	} else if attributes.GetInitiator() == enumspb.CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE {
-		return &tagImpl{key: runInitiator, value: cronRun}
 	}
-	return &tagImpl{key: runInitiator, value: existingRun}
+
+	switch attributes.GetInitiator() {
+	case enumspb.CONTINUE_AS_NEW_INITIATOR_WORKFLOW:
+		return &tagImpl{key: runInitiator, value: childRun}
+	case enumspb.CONTINUE_AS_NEW_INITIATOR_RETRY:
+		return &tagImpl{key: runInitiator, value: retryRun}
+	case enumspb.CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE:
+		return &tagImpl{key: runInitiator, value: cronRun}
+	default:
+		return &tagImpl{key: runInitiator, value: existingRun}
+	}
 }
 
 func FromUnversionedTag(version string) Tag {
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	enumspb "go.temporal.io/api/enums/v1"
+	historypb "go.temporal.io/api/history/v1"
 	enumsspb "go.temporal.io/server/api/enums/v1"
 	"go.temporal.io/server/common/locks"
 	"go.temporal.io/server/common/primitives"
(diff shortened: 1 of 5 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -2463,6 +2463,18 @@ func (ms *MutableStateImpl) AddWorkflowExecutionStartedEventWithOptions(
...
 		return nil, err
 	}
+
+	// Versioning Override set on StartWorkflowExecutionRequest
+	if startRequest.GetStartRequest().GetVersioningOverride() != nil {
+		metrics.WorkerDeploymentVersioningOverrideCounter.With(
+			ms.metricsHandler.WithTags(
+				metrics.NamespaceTag(ms.namespaceEntry.Name().String()),
+				metrics.VersioningBehaviorBeforeOverrideTag(enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED),
+				metrics.VersioningBehaviorAfterOverrideTag(startRequest.GetStartRequest().GetVersioningOverride().GetBehavior()), //nolint:staticcheck // SA1019: worker versioning v0.31
```
ShahabT, line 2473: should use the new override fields. you may need a helper to extract the behavior out of the new override fields because it's not as straightforward.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4804,9 +4816,23 @@ func (ms *MutableStateImpl) AddWorkflowExecutionOptionsUpdatedEvent(
 		attachCompletionCallbacks,
 		links,
 	)
+	prevEffectiveVersioningBehavior := ms.GetEffectiveVersioningBehavior()
 	if err := ms.ApplyWorkflowExecutionOptionsUpdatedEvent(event); err != nil {
 		return nil, err
 	}
+
+	// Versioning Override set via UpdateWorkflowExecutionOptionsRequest
+	if versioningOverride != nil {
```
ShahabT, line 4825: the API may be called for other things too. let's only emit the metric if effective behavior OR effective version changes.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2463,6 +2463,18 @@ func (ms *MutableStateImpl) AddWorkflowExecutionStartedEventWithOptions(
...
 	}
+
+	// Versioning Override set on StartWorkflowExecutionRequest
+	if startRequest.GetStartRequest().GetVersioningOverride() != nil {
+		metrics.WorkerDeploymentVersioningOverrideCounter.With(
+			ms.metricsHandler.WithTags(
+				metrics.NamespaceTag(ms.namespaceEntry.Name().String()),
+				metrics.VersioningBehaviorBeforeOverrideTag(enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED),
+				metrics.VersioningBehaviorAfterOverrideTag(startRequest.GetStartRequest().GetVersioningOverride().GetBehavior()), //nolint:staticcheck // SA1019: worker versioning v0.31
+				metrics.VersioningOverrideOnNewWorkflowTag(prevRunID == ""),
```
ShahabT, line 2474: it'd be super cool if instead we could have the following values:
`_new`, `_child`, `_can`, `_retry`, `_cron`. For the last three you can look at the [initiator](https://github.com/temporalio/api/blob/350edebf546cc43c1a545fe665bb99dfbf7c194b/temporal/api/history/v1/message.proto#L52) field.

Add a `_existing` to the list to cover the case of UpdateOptions.

service/history/workflow/mutable_state_impl.go
```diff
@@ -8125,6 +8153,16 @@ func (ms *MutableStateImpl) StartDeploymentTransition(deployment *deploymentpb.D
...
 		}
 	}
+
+	// DeploymentTransition has taken place, so we increment the DeploymentTransition metric
+	metrics.StartDeploymentTransitionCounter.With(
+		ms.metricsHandler.WithTags(
+			metrics.NamespaceTag(ms.namespaceEntry.Name().String()),
+			metrics.EffectiveDeploymentBeforeTransitionTag(worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(preTransitionEffectiveDeployment))),
+			metrics.EffectiveDeploymentAfterTransitionTag(worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(deployment))),
+		),
```
ShahabT, line 8163: I think here we just want two bools: from_unversioned=true/false, to_unversioned=true/false

common/metrics/tags.go
```diff
@@ -38,15 +38,24 @@ const (
...
+	replicationTaskType                 = "replicationTaskType"
+	replicationTaskPriority             = "replicationTaskPriority"
+	versioningBehavior                  = "versioning_behavior"
+	isFirstAttempt                      = "first-attempt"
+	workflowStatus                      = "workflow_status"
+	workflowBehaviorBeforeOverride      = "workflow_behavior_before_override"
+	workflowBehaviorAfterOverride       = "workflow_behavior_after_override"
+	versioningOverrideOnNewWorkflow     = "versioning_override_on_new_workflow"
+	effectiveDeploymentBeforeTransition = "effective_deployment_before_transition"
+	effectiveDeploymentAfterTransition  = "effective_deployment_after_transition"
```
ShahabT, line 50: let's use shorter names otherwise writing queries will be hard. some of the context is passed automatically by the _metric_ name itself.
```suggestion
	behaviorBefore      = "behavior_before"
	behaviorAfter       = "behavior_after"
	runInitiation     = "run_initiation" // see my other comment.
	fromUnversioned = "from_unversioned"
	toUnversioned  = "to_unversioned"
```

common/metrics/tags.go
```diff
@@ -383,3 +400,44 @@ func DestinationTag(value string) Tag {
...
+	// UpdateWorkflowExecutionOptions API is the caller
+	if attributes == nil {
+		return &tagImpl{key: runInitiator, value: existingRun}
+	}
+
+	// StartWorkflowExecution API is the caller
+	if prevRunID == "" {
+		return &tagImpl{key: runInitiator, value: newRun}
+	} else if attributes.GetInitiator() == enumspb.CONTINUE_AS_NEW_INITIATOR_WORKFLOW {
+		return &tagImpl{key: runInitiator, value: childRun}
```
ShahabT, line 427:
```suggestion
		return &tagImpl{key: runInitiator, value: canRun}
```

common/metrics/tags.go
```diff
@@ -40,13 +41,29 @@ const (
...
+	falseValue              = "false"
+	trueValue               = "true"
+	errorPrefix             = "*"
+
+	newRun      = "__new"
+	existingRun = "__existing"
+	childRun    = "__child"
+	canRun      = "__can"
+	retryRun    = "__retry"
+	cronRun     = "__cron"
```
ShahabT, line 66: actually, I don't think we need the `__` prefixes for this values, it's not like these are special or place-holder values, these are all the values that the tag can get.

common/metrics/tags.go
```diff
@@ -383,3 +400,44 @@ func DestinationTag(value string) Tag {
...
+}
+
+func RunInitiatorTag(prevRunID string, attributes *historypb.WorkflowExecutionStartedEventAttributes) Tag {
+	// UpdateWorkflowExecutionOptions API is the caller
+	if attributes == nil {
+		return &tagImpl{key: runInitiator, value: existingRun}
+	}
+
+	// StartWorkflowExecution API is the caller
+	if prevRunID == "" {
```
ShahabT, line 421: do you really need to pass prevRunID? doesn't CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED mean new?

service/history/workflow/mutable_state_impl.go
```diff
@@ -8125,6 +8155,16 @@ func (ms *MutableStateImpl) StartDeploymentTransition(deployment *deploymentpb.D
 			ms.logInfo("start transition did not reschedule pending speculative task")
 		}
 	}
+
+	// DeploymentTransition has taken place, so we increment the DeploymentTransition metric
+	metrics.StartDeploymentTransitionCounter.With(
+		ms.metricsHandler.WithTags(
+			metrics.NamespaceTag(ms.namespaceEntry.Name().String()),
+			metrics.FromUnversionedTag(worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(preTransitionEffectiveDeployment))),
+			metrics.ToUnversionedTag(worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(deployment))),
```
ShahabT, line 8164: why not have `FromUnversionedTag` get a deployment directly? it's nice to not introduce new code that uses strings.

common/metrics/tags.go
```diff
@@ -383,3 +400,50 @@ func DestinationTag(value string) Tag {
...
+
+	switch attributes.GetInitiator() {
+	case enumspb.CONTINUE_AS_NEW_INITIATOR_WORKFLOW:
+		return &tagImpl{key: runInitiator, value: childRun}
+	case enumspb.CONTINUE_AS_NEW_INITIATOR_RETRY:
+		return &tagImpl{key: runInitiator, value: retryRun}
+	case enumspb.CONTINUE_AS_NEW_INITIATOR_CRON_SCHEDULE:
+		return &tagImpl{key: runInitiator, value: cronRun}
+	default:
+		return &tagImpl{key: runInitiator, value: existingRun}
```
ShahabT, line 433: probably you want to add a case for `CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED` -> `new` and return `unknown` for default.

PR #7831, opened 2025-05-30
Fix version not found error type and test flakes

ShahabT:
## What changed?
Some versioning APIs that used to send NotFound error recently changed to send FailedPrecondition errors. 
Some functional tests where able to absorb the NotFound errors but not FailedPrecondition.

## Why?
The NotFound error seems more suited in these cases anyways, so I'm bringing it back.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
No


carlydf (approved):

Shivs11:
I just came across this....shouldn't we be altering our functional tests to account for `preCondition checks` then? I think version not being present is a preCondition error but won't push since fixing things right now is more important

PR #7833, opened 2025-05-30
[metrics] move prometheus specific configs under metrics.prometheus config

gow:
## What changed?
Moved `WithoutUnitSuffix`, `WithoutCounterSuffix` & `RecordTimerInSeconds` configs under `metrics.prometheus` config

## Why?
These configs are specific to prometheus + OTel 

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
N/A

### DIFF
--- a/common/metrics/config.go
+++ b/common/metrics/config.go
@@ -50,22 +50,6 @@ type (
 		// - "milliseconds"
 		// - "bytes"
 		PerUnitHistogramBoundaries map[string][]float64 `yaml:"perUnitHistogramBoundaries"`
-
-		// Following configs are added for backwards compatibility when switching from tally to opentelemetry
-		// All configs should be set to true when using opentelemetry framework to have the same behavior as tally.
-
-		// WithoutUnitSuffix controls the additional of unit suffixes to metric names.
-		// This config only takes effect when using opentelemetry framework.
-		// Note: this config only takes effect when using prometheus via opentelemetry framework
-		WithoutUnitSuffix bool `yaml:"withoutUnitSuffix"`
-		// WithoutCounterSuffix controls the additional of _total suffixes to counter metric names.
-		// This config only takes effect when using opentelemetry framework.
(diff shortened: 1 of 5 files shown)


common/metrics/config.go
```diff
@@ -50,22 +50,6 @@ type (
...
-		// Note: this config only takes effect when using prometheus via opentelemetry framework
-		WithoutUnitSuffix bool `yaml:"withoutUnitSuffix"`
-		// WithoutCounterSuffix controls the additional of _total suffixes to counter metric names.
-		// This config only takes effect when using opentelemetry framework.
-		// Note: this config only takes effect when using prometheus via opentelemetry framework
-		WithoutCounterSuffix bool `yaml:"withoutCounterSuffix"`
-		// RecordTimerInSeconds controls if Timer metric should be emitted as number of seconds
-		// (instead of milliseconds).
-		// This config only takes effect when using opentelemetry framework for both statsd and prometheus.
-		RecordTimerInSeconds bool `yaml:"recordTimerInSeconds"`
```
yycptt, line 68: SInce they are already released, I guess we can't simply removed it. Have to mark them deprecated for one release and remove in the one after I think. Also carryover the those values if set to those new ones under PrometheusConfig.

gow: Reverted

common/metrics/config.go
```diff
@@ -64,7 +64,7 @@ type (
 		WithoutCounterSuffix bool `yaml:"withoutCounterSuffix"`
 		// RecordTimerInSeconds controls if Timer metric should be emitted as number of seconds
 		// (instead of milliseconds).
-		// This config only takes effect when using opentelemetry framework for both statsd and prometheus.
+		// This config only takes effect when using prometheus via opentelemetry framework
```
yycptt, line 67: nit: let's add the same for the other two above as well.

yycptt (approved):

PR #7839, opened 2025-05-30
Add MultiOperation error logging

stephanos:
## What changed?

Added a log for every MultiOperation error in history.

## Why?

Track down a race condition.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

Since it's only logging errors, I expect the log volume to be reasonable/low.

### DIFF
--- a/service/history/api/multioperation/api.go
+++ b/service/history/api/multioperation/api.go
@@ -8,11 +8,13 @@ import (
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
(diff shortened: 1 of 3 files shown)


alexshtin (approved):

service/history/api/multioperation/api.go
```diff
@@ -79,6 +99,39 @@ func Invoke(
 		startReq:           startReq,
 	}
 
+	// TODO(stephanos): remove again
+	defer func() {
+		if retErr != nil {
```
alexshtin, line 104: I would put it behind namespace scoped dynamic config.

stephanos: Done 👍

PR #7840, opened 2025-05-30
Dynamic config generic parse hooks

dnr:
## What changed?
Add interfaces that can be used for types to implement custom dynamic config parsing.

## Why?
It's nice to be able to parse strings into enum values.

## How did you test it?
- [x] added new unit test(s)


common/dynamicconfig/collection.go
```diff
@@ -488,3 +490,26 @@ func mapstructureHookProtoEnum(f, t reflect.Type, data any) (any, error) {
...
+		mth.Func.IsValid() &&
+		mth.Type.NumIn() == 2 &&
+		mth.Type.In(1) == f &&
+		mth.Type.NumOut() == 2 &&
+		mth.Type.Out(0) == t &&
+		mth.Type.Out(1) == errorType {
+
+		out := mth.Func.Call([]reflect.Value{reflect.Zero(t), reflect.ValueOf(data)})
+		if !out[1].IsNil() {
+			return nil, out[1].Interface().(error) // nolint:revive // type checked above
```
stephanos: Where does this error go? Is it logged or returned anywhere? I see in the test that `Get` doesn't return it.

dnr: It gets sometimes-logged here: https://github.com/temporalio/temporal/blob/a5ed8eeebe2ad210778071ec81e87977fcabb6f0/common/dynamicconfig/collection.go#L268

dnr: on second thought I'll check the conversion just in case

PR #7841, opened 2025-05-30
Fix max deployment count check to not block new versions

ShahabT:
## What changed?
The check to enforce number of Worker Deployments remain within the limit for a NS could also prevent adding new versions to existing deployments when user has just enough deployments that reach the limit but not exceed it.

## Why?
Fixing unexpected behavior.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None.

### DIFF
--- a/service/matching/physical_task_queue_manager.go
+++ b/service/matching/physical_task_queue_manager.go
@@ -566,10 +566,13 @@ func (c *physicalTaskQueueManagerImpl) ensureRegisteredInDeploymentVersion(
 		}
 		var errMaxTaskQueuesInVersion workerdeployment.ErrMaxTaskQueuesInVersion
 		var errMaxVersionsInDeployment workerdeployment.ErrMaxVersionsInDeployment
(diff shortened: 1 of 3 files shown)


Shivs11:
>  adding new versions to existing deployment

I did a quick pass and i'm going to do one more soon, but the changes seem to be related to Listing workflow executions; how is/was this preventing adding new versions to deployment?

service/matching/physical_task_queue_manager.go
Shivs11, line 1: >  adding new versions to existing deployment

I did a quick pass and i'm going to do one more soon, but the changes seem to be related to Listing workflow executions; how is/was this preventing adding new versions to deployment?

Shivs11: was wrong - ignore the above comment - reviewing again

Shivs11 (approved):
hopefully fixes the breaking canaries!

PR #7849, opened 2025-06-02
Minor housekeeping items for Versioning

ShahabT:
## What changed?
Renaming old versioning test suite to make it more clear these are old tests.

## Why?
Plus some other minor changes

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/tests/versioning_old_test.go
+++ b/tests/versioning_old_test.go
@@ -37,7 +37,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
-type VersioningIntegSuite struct {
+type VersioningOldSuite struct {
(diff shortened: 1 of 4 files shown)


tests/versioning_old_test.go
```diff
@@ -37,7 +37,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
-type VersioningIntegSuite struct {
+type VersioningOldSuite struct {
```
ychebotarev, line 40: can we rather use "versions"?

ychebotarev: Like 'VersioningV1TestSuite"

PR #7851, opened 2025-06-02
Delete Namespace: allow to delete protected namespace in the cluster which is not in the list of namespace cluster

alexshtin:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


laniehei (approved):
nit: add comment about what this is checking (this namespace has been moved elsewhere so we can ignore this check)

PR #7855, opened 2025-06-03
[CHASM] Export chasm.Engine and bring it into timer queue factory

lina-temporal:
## What changed?
- Exported the chasm.Engine interface
- Generated mocks for chasm.Engine
- Shard-level timer task executor objects now require a chasm.Engine
- Once we have an implementation ready, we can simply provide it via FX

## Why?
- I need this wired in so that SideEffect tasks can add it to the context (I'd rather not leave this as a todo).

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -5862,22 +5862,10 @@ func (ms *MutableStateImpl) truncateRetryableActivityFailure(
 		return activityFailure
 	}
 
-	throttledLogger := log.With(
-		ms.shard.GetThrottledLogger(),
-		tag.WorkflowNamespace(namespaceName),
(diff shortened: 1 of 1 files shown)


chasm/engine.go
```diff
@@ -1,10 +1,12 @@
...
+
 package chasm
 
 import (
 	"context"
 )
 
-type engine interface {
+type Engine interface {
 	newEntity(
```
yycptt, line 10: I think those methods needs to be exported as well or the ChasmEngine struct in history package won't implement this interface.

lina-temporal: Yup, will follow-up with this in the PR that I add the FX provider.

PR #7856, opened 2025-06-03
[CHASM] Validate pure tasks in EachPureTask

lina-temporal:
## What changed?
- Noticed EachPureTask was expected to Validate in the standby validator, fixed that up and added more tests.
## How did you test it?
- [x] built
- [ ] added new unit test(s)


chasm/tree.go
```diff
@@ -1762,6 +1770,15 @@ func (n *Node) EachPureTask(
 				return err
 			}
 
+			valid, err := n.validateTask(ctx, taskValue)
```
yycptt: Shall we do this in standby executor's execute callback func? Seems a bit cleaner to me and there's no double validation for active tasks.

lina-temporal: Yeah, agreed, moved it back into the validation callback.

chasm/tree.go
```diff
@@ -158,10 +158,11 @@ type (
 		Decode(encodedPath string) ([]string, error)
 	}
 
-	// NodeExecutePureTask is intended to be implemented and used within the CHASM
+	// NodePureTask is intended to be implemented and used within the CHASM
 	// framework only.
-	NodeExecutePureTask interface {
+	NodePureTask interface {
```
yycptt, line 163: Can you remind me why we need this interface?

lina-temporal: Sure. It could easily be a Tree interface if we had one in the CHASM package, but as it stands, since `EachPureTask` is defined in `chasm` and `ChasmTree` is in `history/interfaces`, it'd create a circular reference.

PR #7860, opened 2025-06-03
[PoC] Allow different persistence serialization encodings

stephanos:
## What changed?

(1) Separated persistence's serialization from the data encoding type; extracted `Codec` for both JSON and proto3.
(2) Allow overriding the codec via env var, it allows to use JSON instead when working locally.

NOTE: There are still a few places where serialization is done manually or encoding types are hard-coded to proto. I'll follow up on these later, as I want to keep this PR as small as possible.

**No behavior change is expected.**

## Why?

For development/debugging purposes.

Proto is not human-readable, JSON is. When developing locally, it is useful to be able to read the data that goes in/out persistence. Especially when using OTEL for debugging it's useful to be able to read the data.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks

1. It will break internal code using the `Serializer` interface; but that's an easy fix.
2. I don't expect any encoding/decoding issues as the code is (expected to be) the same; and we have plenty of (functional) tests.
3. Makes it difficult to use a _mixed_ encoding scheme. If that ever happens, an encoding type parameter can be added again (only where needed).


common/persistence/serializer_test.go
stephanos, line 1: This test is already included in `serialization/serializer_test.go` 1:1 (someone must have copied it and forgot to delete this one).

common/xdc/ndc_history_resender_test.go
```diff
@@ -512,10 +512,7 @@ func (s *nDCHistoryResenderSuite) TestGetHistory() {
 }
 
 func (s *nDCHistoryResenderSuite) serializeEvents(events []*historypb.HistoryEvent) *commonpb.DataBlob {
-	blob, err := s.serializer.SerializeEvents(events, enumspb.ENCODING_TYPE_PROTO3)
+	blob, err := s.serializer.SerializeEvents(events)
 	s.Nil(err)
-	return &commonpb.DataBlob{
-		EncodingType: enumspb.ENCODING_TYPE_PROTO3,
-		Data:         blob.Data,
-	}
```
stephanos, line 520: Test also succeeds just returning the data blob as-is.

common/persistence/serialization/serializer_test.go
```diff
@@ -49,88 +49,93 @@ func (s *temporalSerializerSuite) SetupTest() {
...
+		name       string
+		serializer Serializer
+	}{
+		{
+			name:       "Default",
+			serializer: NewSerializer(),
 		},
+		{
+			name:       "WithJsonCodec",
+			serializer: NewSerializerWithHelper(NewJsonCodec()),
```
stephanos: This is the only net new test here.

common/persistence/serialization/error.go
stephanos, line 1: I moved the serialization error types to a separate file. Nothing was changed here.

common/persistence/serialization/codec.go
stephanos, line 1: This code was extracted from the (now deleted) serialization helpers.

temporal/server_test.go
```diff
@@ -28,6 +29,13 @@ func TestNewServer(t *testing.T) {
 	startAndStopServer(t)
 }
 
+// TestNewServerWithJsonEncoding verifies that NewServer runs with JSON payload encoding.
+func TestNewServerWithJsonEncoding(t *testing.T) {
+	t.Setenv(serialization.CodecEnvVar, "json")
+
+	startAndStopServer(t)
+}
+
```
stephanos, line 38: Added this as a litmus test to ensure it doesn't crash.

PR #7872, opened 2025-06-06
CHASM: Implement Ref method

yycptt:
## What changed?
- Properly Implement Ref() method and populate all necessary fields
- Refactor ComponentRef and related methods

## Why?
- CHASM workstream

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/context.go
+++ b/chasm/context.go
@@ -13,14 +13,15 @@ type Context interface {
 
 	// NOTE: component created in the current transaction won't have a ref
 	// this is a Ref to the component state at the start of the transition
-	Ref(Component) (ComponentRef, error)
-	refData(proto.Message) (ComponentRef, error)
+	Ref(Component) ([]byte, error)
 	Now(Component) time.Time
 
 	// Intent() OperationIntent
 	// ComponentOptions(Component) []ComponentOption
 
 	getContext() context.Context
+	componentNodePath(Component) ([]string, error)
+	dataNodePath(proto.Message) ([]string, error)
 }
 
 type MutableContext interface {
@@ -63,12 +64,16 @@ func NewContext(
 	}
 }
 
-func (c *ContextImpl) Ref(component Component) (ComponentRef, error) {
+func (c *ContextImpl) Ref(component Component) ([]byte, error) {
 	return c.root.Ref(component)
 }
 
-func (c *ContextImpl) refData(data proto.Message) (ComponentRef, error) {
-	return c.root.refData(data)
+func (c *ContextImpl) componentNodePath(component Component) ([]string, error) {
+	return c.root.componentNodePath(component)
+}
+
+func (c *ContextImpl) dataNodePath(data proto.Message) ([]string, error) {
+	return c.root.dataNodePath(data)
 }
 
 func (c *ContextImpl) Now(component Component) time.Time {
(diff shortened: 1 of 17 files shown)


chasm/context.go
```diff
@@ -13,14 +13,15 @@ type Context interface {
 
 	// NOTE: component created in the current transaction won't have a ref
 	// this is a Ref to the component state at the start of the transition
-	Ref(Component) (ComponentRef, error)
-	refData(proto.Message) (ComponentRef, error)
+	Ref(Component) ([]byte, error)
```
lina-temporal, line 16: Is the intention to make `ComponentRef` opaque to component developers, or is this more about performance?

chasm/field.go
```diff
@@ -44,29 +44,36 @@ func NewComponentField[C Component](
 	}
 }
 
+// TODO: The Component|DataPointerTo() implementation below can't handle the case
+// where Pointer is created in the NewEntity transition, as the tree structure is
+// unknown to the framework yet.
+//
+// To handle that case, we have to store the Component value in the field when
+// the Pointer field is created and resolve the pointer at the end of the transition
+// i.e. when closing the transaction.
```
lina-temporal, line 53: I'm probably reading too much into the TODO, but what about if someone created a new Component, and then mutated multiple Pointer fields to point to that Component? How would we pick which cached Component value gets written (assuming they both resolve to an identical path)?

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -79,17 +79,25 @@ message ChasmComponentRef {
...
+
+// ChasmComponentRef references a specific chasm component.
+message ChasmComponentRef {
+    string namespace_id = 1;
+    string business_id = 2;
+    string entity_id = 3;
+
+    string archetype = 4;
+    
+    VersionedTransition entity_versioned_transition = 5;
```
lina-temporal, line 99: Just curious why this name versus `component_last_update_versioned_transition`?

chasm/field.go
yycptt: Well we only look at the final component value in updated/new Pointer Fields. If that's something we found in the tree then it's valid and resolve to the path for the real node, otherwise we'll just throw an error. 
Of course we need to first go though other updated nodes in the tree so that we can know there this new real component value.

Not sure if that answers your question.

chasm/context.go
yycptt: not really... I just feel in most cases developer call this method to get a ref so that they can pass it to another service and expect that service to callback using the ref. In that case, they will need to serialize the ref anyway so this method can just do that for them. 
I don't really have strong opinion on that TBH, Maybe we will change it back later if people feel it's not good experience.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
yycptt: oh because it is the entity (tree) level VT, not the last update VT for the component. 
The concept of staleness at entity (tree) level. Like you can't update one component in a stale tree when some other components already got updated and persisted, it may cause an inconsistency in the tree. 

However it's worth pointing out that using entity VT might be too strict during conflict resolution case, and I have created a task to track it. Idea is sometimes we can allow the staleness check to pass as long as the component init VT is on the current branch.

PR #7877, opened 2025-06-06
Add monitor in stream sender on sync status

yux0:
## What changed?
Add monitor in stream sender on sync status

## Why?
Receiver send replication status per seconds.
Sender should monitor on this signal and assume the receiver side is stop if it loses this sync status.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/history/replication/stream.go
+++ b/service/history/replication/stream.go
@@ -109,8 +109,7 @@ func livenessMonitor(
 			case <-signalChan:
 				continue
 			default:
-				// No liveness signal received. Stop the stream.
-				logger.Warn("Stream failed to receive sync status from target cluster")
+				logger.Warn("No liveness signal received. Stop the replication stream.")
 				stopStream()
 				return
 			}


service/history/replication/stream_sender.go
```diff
@@ -195,6 +206,23 @@ func (s *StreamSenderImpl) recvEventLoop() (retErr error) {
...
 }
 
+func (s *StreamSenderImpl) recvMonitor() {
+	heartbeatTimeout := time.NewTimer(s.config.ReplicationStreamSyncStatusDuration() * SyncTaskIntervalMultiplier)
+	defer heartbeatTimeout.Stop()
+
+	for !s.shutdownChan.IsShutdown() {
+		select {
+		case <-s.recvSignalChan:
+			heartbeatTimeout.Stop()
```
xwduan: If the timer is already fired, it may cause the go routine leak. Maybe do this:
```
if !timer.Stop() {
	select {
	case <-timer.C: // drain the channel if it fired
	default:
	}
}
```

service/history/replication/stream_sender.go
```diff
@@ -34,7 +34,10 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-const TaskMaxSkipCount int = 1000
+const (
+	TaskMaxSkipCount           int = 1000
+	SyncTaskIntervalMultiplier int = 3
```
xwduan: I think this multiplier is a bit too small, especially when the stream is just started. Receiver side have to wait for sender at least send 1 task before it can send back ack message. `10` maybe safer?

service/history/replication/stream_sender.go
yux0: Updated in both places.

service/history/replication/stream.go
```diff
@@ -89,6 +90,34 @@ func WrapEventLoop(
...
+	signalChan <-chan struct{},
+	timeout time.Duration,
+	shutdownChan channel.ShutdownOnce,
+	stopStream func(),
+	logger log.Logger,
+) {
+	heartbeatTimeout := time.NewTimer(timeout)
+	defer heartbeatTimeout.Stop()
+
+	for !shutdownChan.IsShutdown() {
```
xwduan, line 103: This `for` loop works okay for SyncState Message as we are expecting it every second. But for `stream_receiver`, it might constantly receive replication tasks, and this for loop is constantly doing timer reset for every(almost, as the operation is light) replication task, which I think might not efficient/ideal. Let me know what do you think.

yux0: I am not sure the reset perf. To eliminate the concern, I just move to ticker instead of reset the timer.

xwduan (approved):

PR #7878, opened 2025-06-06
Revert "Include transient and speculative WFT events in GetWorkflowExecutionHistory response (#7732)"

prathyushpv:
This reverts commit 5fc3fb21c4e556750e8dc5c8a2e27f92833ef3b4.

## What changed?
Reverts the change to return transient and speculative WFT events.

## Why?
Started seeing new error like this
"history_events: premature end of stream, expectedLastEventID=4424 but no more events after eventID=4422" in pipelines which was not there before.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ ci-build-misc: \
 	clean-tools \
 	proto \
 	go-generate \
-	buf-breaking \
 	shell-check \
 	goimports \
 	gomodtidy \
--- a/api/history/v1/message.pb.go
+++ b/api/history/v1/message.pb.go
@@ -11,7 +11,6 @@ import (
 	sync "sync"
 	unsafe "unsafe"
 
-	v11 "go.temporal.io/api/enums/v1"
 	v1 "go.temporal.io/api/history/v1"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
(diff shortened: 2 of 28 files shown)


stephanos (approved):

prathyushpv:
Waiting to see if versioning fix solved this issue.

Makefile
```diff
@@ -14,7 +14,6 @@ ci-build-misc: \
 	clean-tools \
 	proto \
 	go-generate \
-	buf-breaking \
```
dnr, line 17: don't do this, add lines here instead:
https://github.com/temporalio/temporal/blob/main/proto/internal/buf.yaml#L14

prathyushpv: I will enable it back after merging this.

dnr: I know, but using that file is a better way to disable the check temporarily:
- it's scoped to specific files instead of everything
- if you forget to re-enable it, someone will notice eventually
- keeps git history of the makefile cleaner

prathyushpv: Got it. Sorry about merging this in a rush. I wanted to have this in today's nightly tests. 
Added a comment here: https://github.com/temporalio/temporal/pull/7892

PR #7880, opened 2025-06-06
Add stubs for matching for RecordWorkerHeartbeat/ListWorkers

ychebotarev:
## What changed?
* Add stubs for matching for RecordWorkerHeartbeat/ListWorkers
* Add routing, based on namespace id
* Route those calls to matching from frontend

## Why?
Part of worker visibility work. Current implementation will be "in-memory", and all worker heartbeats for the same namespace should be on the same service instance.

## How did you test it?
- [X] built

### DIFF
--- a/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
@@ -546,7 +546,7 @@ message ListNexusEndpointsResponse {
 
 message RecordWorkerHeartbeatRequest {
     string namespace_id = 1;
-    temporal.api.workflowservice.v1.RecordWorkerHeartbeatRequest api_request = 3;
+    temporal.api.workflowservice.v1.RecordWorkerHeartbeatRequest heartbeart_request = 2;
 }
 
 message RecordWorkerHeartbeatResponse {
@@ -555,12 +555,10 @@ message RecordWorkerHeartbeatResponse {
 
 message ListWorkersRequest {
     string namespace_id = 1;
-    temporal.api.workflowservice.v1.ListWorkersRequest api_request = 3;
+    temporal.api.workflowservice.v1.ListWorkersRequest list_request = 2;
 }
 
 message ListWorkersResponse {
     repeated temporal.api.worker.v1.WorkerInfo workers_info = 1;
-
-    // Next page token
     bytes next_page_token = 2;
 }
--- a/proto/internal/temporal/server/api/matchingservice/v1/service.proto
+++ b/proto/internal/temporal/server/api/matchingservice/v1/service.proto
@@ -166,6 +166,9 @@ service MatchingService {
 
     // RecordWorkerHeartbeat receive heartbeat request from the worker.
     rpc RecordWorkerHeartbeat (RecordWorkerHeartbeatRequest) returns (RecordWorkerHeartbeatResponse) {}
-    // ListWorkers is a visibility API to list worker status information in a specific namespace.
+
+    // ListWorkers retrieves a list of workers in the specified namespace that match the provided filters.
+    // Supports pagination for large result sets. Returns an empty list if no workers match the criteria.
+    // Returns an error if the namespace doesn't exist.
     rpc ListWorkers (ListWorkersRequest) returns (ListWorkersResponse) {}
 }
--- a/service/matching/handler.go
+++ b/service/matching/handler.go
@@ -540,11 +540,11 @@ func (h *Handler) RecordWorkerHeartbeat(
 	return nil, serviceerror.NewUnimplemented("RecordWorkerHeartbeat is not implemented")
 }
 
-// ListWorkers is a visibility API to list worker status information in a specific namespace.
+// ListWorkers retrieves a list of workers in the specified namespace that match the provided filters.
 func (h *Handler) ListWorkers(
 	context.Context, *matchingservice.ListWorkersRequest,
 ) (*matchingservice.ListWorkersResponse, error) {
-	return nil, serviceerror.NewUnimplemented("RecordWorkerHeartbeat is not implemented")
+	return nil, serviceerror.NewUnimplemented("ListWorkers is not implemented")
 }
 
 func (h *Handler) namespaceName(id namespace.ID) namespace.Name {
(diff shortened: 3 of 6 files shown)


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -541,3 +542,24 @@ message ListNexusEndpointsResponse {
     int64 table_version = 2;
     repeated temporal.server.api.persistence.v1.NexusEndpointEntry entries = 3;
 }
+
+message RecordWorkerHeartbeatRequest {
+    string namespace_id = 1;
+    temporal.api.workflowservice.v1.RecordWorkerHeartbeatRequest api_request = 3;
```
dnr:
```suggestion
    temporal.api.workflowservice.v1.RecordWorkerHeartbeatRequest api_request = 2;
```
?
doesn't matter, just wondering

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -541,3 +542,24 @@ message ListNexusEndpointsResponse {
...
+
+message ListWorkersRequest {
+    string namespace_id = 1;
+    temporal.api.workflowservice.v1.ListWorkersRequest api_request = 3;
+}
+
+message ListWorkersResponse {
+    repeated temporal.api.worker.v1.WorkerInfo workers_info = 1;
+
+    // Next page token
```
dnr: not sure the comment adds anything here :)

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
dnr: also usually it's just called `request`

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -541,3 +542,24 @@ message ListNexusEndpointsResponse {
...
+    temporal.api.workflowservice.v1.RecordWorkerHeartbeatRequest api_request = 3;
+}
+
+message RecordWorkerHeartbeatResponse {
+
+}
+
+message ListWorkersRequest {
+    string namespace_id = 1;
+    temporal.api.workflowservice.v1.ListWorkersRequest api_request = 3;
```
dnr:
```suggestion
    temporal.api.workflowservice.v1.ListWorkersRequest request = 2;
```
?

proto/internal/temporal/server/api/matchingservice/v1/service.proto
```diff
@@ -163,4 +163,9 @@ service MatchingService {
     rpc	DeleteNexusEndpoint(DeleteNexusEndpointRequest) returns (DeleteNexusEndpointResponse) {}
     // List all registered services.
     rpc	ListNexusEndpoints(ListNexusEndpointsRequest) returns (ListNexusEndpointsResponse) {}
+
+    // RecordWorkerHeartbeat receive heartbeat request from the worker.
+    rpc RecordWorkerHeartbeat (RecordWorkerHeartbeatRequest) returns (RecordWorkerHeartbeatResponse) {}
+    // ListWorkers is a visibility API to list worker status information in a specific namespace.
```
dnr: hmm, we might want to be a little careful with this comment: "visibility API" has a very specific meaning. we're not (currently) dealing with the visibility system, though maybe we will later.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -541,3 +542,24 @@ message ListNexusEndpointsResponse {
...
+}
+
+message ListWorkersResponse {
+    repeated temporal.api.worker.v1.WorkerInfo workers_info = 1;
```
dnr, line 562: it may be simpler to wrap the response too:
```suggestion
    temporal.api.workflowservice.v1.ListWorkersResponse response = 1;
```
?

service/matching/handler.go
```diff
@@ -532,6 +533,20 @@ func (h *Handler) ListNexusEndpoints(ctx context.Context, request *matchingservi
...
+	context.Context, *matchingservice.RecordWorkerHeartbeatRequest,
+) (*matchingservice.RecordWorkerHeartbeatResponse, error) {
+	return nil, serviceerror.NewUnimplemented("RecordWorkerHeartbeat is not implemented")
+}
+
+// ListWorkers is a visibility API to list worker status information in a specific namespace.
+func (h *Handler) ListWorkers(
+	context.Context, *matchingservice.ListWorkersRequest,
+) (*matchingservice.ListWorkersResponse, error) {
+	return nil, serviceerror.NewUnimplemented("RecordWorkerHeartbeat is not implemented")
```
dnr:
```suggestion
	return nil, serviceerror.NewUnimplemented("ListWorkers is not implemented")
```

ychebotarev: done

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
ychebotarev: I think we a has the pattern "don't create API response". In other places response is created in the handler, lets keep it that way.

proto/internal/temporal/server/api/matchingservice/v1/service.proto
ychebotarev: ok I will remove it from the comments.

ychebotarev: and add slightly more detailed comments.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
ychebotarev: yep. done

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
ychebotarev: removed

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
ychebotarev: the input parameter will be caller request, and I don't like "request.request". If you look at other request wrappers we are calling them "query_request", "completed_request".
I will rename to "heartbeart_request"

PR #7881, opened 2025-06-06
check if callbacks exist when reapply events

hai719:
## What changed?
Check if callbacks exist when reapply events.

## Why?
If callbacks already exist, we should not reapply event to avoid error. 
If only partial callbacks exist, we cannot handle this case and have to fail.

## How did you test it?
- [x] built
- [x] added new unit test(s)


service/history/ndc/workflow_resetter.go
```diff
@@ -930,11 +930,26 @@ func reapplyEvents(
...
+			callbacks := attr.GetAttachedCompletionCallbacks()
+			if len(callbacks) > 0 {
+				// check if all completion callbacks exist
+				existingCbs := mutableState.GetExistingCompletionCallbackCount(event, attr.GetAttachedRequestId(), callbacks)
+				if existingCbs == len(callbacks) {
+					if attr.GetVersioningOverride() == nil && !attr.GetUnsetVersioningOverride() {
+						// if all completion callbacks exist, and no other updates in the event, we should skip the event
+						continue
+					}
+					return reappliedEvents, serviceerror.NewInternalf("unable to reapply WorkflowExecutionOptionsUpdated event: all %d completion callbacks are already attached but the event contains additional workflow option updates", existingCbs)
```
yycptt: do we expect this to happen? it feels like we need to continue to reapply the versioning part?

service/history/ndc/workflow_resetter.go
```diff
@@ -930,11 +930,26 @@ func reapplyEvents(
 				continue
 			}
 			attr := event.GetWorkflowExecutionOptionsUpdatedEventAttributes()
+			callbacks := attr.GetAttachedCompletionCallbacks()
+			if len(callbacks) > 0 {
```
yycptt: Want to understand if we can only verify whether the requestID exists or not. Is there any case where requestID can be attached but not all of the callbacks?

service/history/ndc/workflow_resetter.go
hai719: If we skip the existing callbacks and continue to reapply the versioning update, this event will be added to history with duplicated callbacks. Will this cause any issue to users?

From existing code, AddWorkflowExecutionOptionsUpdatedEvent is only called with VersioningOverride OR callbacks, not both. Probably there will be future usage of both?

service/history/ndc/workflow_resetter.go
hai719: Good point. I don't see that case. Will update to use requestID.

service/history/ndc/workflow_resetter.go
```diff
@@ -930,11 +930,22 @@ func reapplyEvents(
 				continue
 			}
 			attr := event.GetWorkflowExecutionOptionsUpdatedEventAttributes()
+			callbacks := attr.GetAttachedCompletionCallbacks()
+			requestID := attr.GetAttachedRequestId()
+			if len(callbacks) > 0 && requestID != "" {
+				if mutableState.HasRequestID(requestID) {
```
yycptt, line 936: nit: you can get executionState from mutableState and check the requestID, thus avoid adding a new method. But no strong opinion.

service/history/ndc/workflow_resetter.go
yycptt: I see. Yeah, it's just a bit brittle if it's not a oneof in the proto def. Is it possible to return an error when both are specified when the event is generated? If we only discover it at reapply time, not sure how we can resolve it.

hai719: added check when generating event.

hai719: Synced offline. We should not add that check. WORKFLOW_EXECUTION_OPTIONS_UPDATED can have both of versioningoverride and callbacks.

PR #7889, opened 2025-06-09
Improve error message for aborted Update

stephanos:
## What changed?

Improve the error message for an Update that is aborted by a closing workflow.

## Why?

To make it easier for users to understand what happend; and for devs to debug it easier.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

Very low as the service error _type_ is still the same (NotFound).

### DIFF
--- a/service/history/workflow/update/abort_reason.go
+++ b/service/history/workflow/update/abort_reason.go
@@ -49,10 +49,10 @@ var reasonStateMatrix = map[reasonState]failureError{
 	reasonState{r: AbortReasonRegistryCleared, st: stateAborted}:              {f: nil, err: nil},
 
 	// If the Workflow is completed, then pre-accepted Updates are aborted with non-retryable error.
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateCreated}:               {f: nil, err: WorkflowCompletedErr},
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateProvisionallyAdmitted}: {f: nil, err: WorkflowCompletedErr},
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateAdmitted}:              {f: nil, err: WorkflowCompletedErr},
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateSent}:                  {f: nil, err: WorkflowCompletedErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateCreated}:               {f: nil, err: AbortedByWorkflowClosingErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateProvisionallyAdmitted}: {f: nil, err: AbortedByWorkflowClosingErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateAdmitted}:              {f: nil, err: AbortedByWorkflowClosingErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateSent}:                  {f: nil, err: AbortedByWorkflowClosingErr},
 	// Accepted Updates are failed with special server failure because if a client knows that Update has been accepted,
 	// it expects any following requests to return an Update result (or failure) but not an error.
 	// There can be different types of Update failures coming from worker and a client must handle them anyway.
--- a/service/history/workflow/update/errors_failures.go
+++ b/service/history/workflow/update/errors_failures.go
@@ -8,10 +8,10 @@ import (
 )
 
 var (
-	registryClearedErr       = errors.New("update registry was cleared")
-	WorkflowUpdateAbortedErr = serviceerror.NewUnavailable("workflow update was aborted")
-	WorkflowCompletedErr     = serviceerror.NewNotFound("workflow update was aborted by closing workflow")
-	workflowTaskFailErr      = serviceerror.NewWorkflowNotReady("Unable to perform workflow execution update due to unexpected workflow task failure.")
+	registryClearedErr          = errors.New("update registry was cleared")
+	AbortedByServerErr          = serviceerror.NewUnavailable("workflow update was aborted")
+	AbortedByWorkflowClosingErr = serviceerror.NewNotFound("workflow update was aborted by closing workflow")
+	workflowTaskFailErr         = serviceerror.NewWorkflowNotReady("Unable to perform workflow execution update due to unexpected workflow task failure.")
 )
 
 var (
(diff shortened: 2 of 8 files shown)


service/history/workflow/update/abort_reason.go
```diff
@@ -48,11 +48,11 @@ var reasonStateMatrix = map[reasonState]failureError{
...
-	// If the Workflow is completed, then pre-accepted Updates are aborted with non-retryable ErrWorkflowCompleted error.
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateCreated}:               {f: nil, err: consts.ErrWorkflowCompleted},
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateProvisionallyAdmitted}: {f: nil, err: consts.ErrWorkflowCompleted},
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateAdmitted}:              {f: nil, err: consts.ErrWorkflowCompleted},
-	reasonState{r: AbortReasonWorkflowCompleted, st: stateSent}:                  {f: nil, err: consts.ErrWorkflowCompleted},
+	// If the Workflow is completed, then pre-accepted Updates are aborted with non-retryable error.
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateCreated}:               {f: nil, err: WorkflowCompletedErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateProvisionallyAdmitted}: {f: nil, err: WorkflowCompletedErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateAdmitted}:              {f: nil, err: WorkflowCompletedErr},
+	reasonState{r: AbortReasonWorkflowCompleted, st: stateSent}:                  {f: nil, err: WorkflowCompletedErr},
```
stephanos: This is the change; everything else is just a consequence of that change.

service/history/workflow/update/update.go
```diff
@@ -18,10 +18,6 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 )
 
-var (
-	WorkflowUpdateAbortedErr = serviceerror.NewUnavailable("Workflow Update was aborted.")
```
stephanos, line 22: I moved this to the other errors.

service/history/workflow/update/errors_failures.go
```diff
@@ -8,16 +8,18 @@ import (
 )
 
 var (
-	registryClearedErr  = errors.New("update registry was cleared")
-	workflowTaskFailErr = serviceerror.NewWorkflowNotReady("Unable to perform workflow execution update due to unexpected workflow task failure.")
+	registryClearedErr       = errors.New("update registry was cleared")
+	WorkflowUpdateAbortedErr = serviceerror.NewUnavailable("workflow update was aborted")
+	WorkflowCompletedErr     = serviceerror.NewNotFound("update aborted by closing workflow")
```
stephanos: The new error.

service/history/workflow/update/errors_failures.go
```diff
@@ -8,16 +8,18 @@ import (
 )
 
 var (
-	registryClearedErr  = errors.New("update registry was cleared")
-	workflowTaskFailErr = serviceerror.NewWorkflowNotReady("Unable to perform workflow execution update due to unexpected workflow task failure.")
+	registryClearedErr       = errors.New("update registry was cleared")
+	WorkflowUpdateAbortedErr = serviceerror.NewUnavailable("workflow update was aborted")
```
ychebotarev: Not blocking (though - "readability") but not sure I like naming.
One of them is "worfklow update was aborted"
and another - "worfklow update was aborted because of...."

But form the naming it is not clear that for second one it was aborted.

"WorkflowCompletedErr" - "WorfklowUpdateAbortedWorfklowClosedErr"? Or too long?

stephanos: 👍 Good callout; I can adjust the name.

tests/update_workflow_test.go
```diff
@@ -3039,7 +3040,7 @@ func (s *UpdateWorkflowSuite) TestScheduledSpeculativeWorkflowTask_TerminateWork
 	s.Error(updateResult.err)
 	var notFound *serviceerror.NotFound
 	s.ErrorAs(updateResult.err, &notFound)
-	s.Equal("workflow execution already completed", updateResult.err.Error())
+	s.ErrorContains(updateResult.err, update.WorkflowCompletedErr.Error())
```
ychebotarev: should this also be ErrorAs instead of comparing strings?

stephanos: I still don't know what the best approach for these situations is.

But I don't think it's
```
s.ErrorAs(updateResult.err, &update.AbortedByWorkflowClosingErr)
```
since this will only look at the _type_ of the error (which is `serviceerror.NotFound`); it ignores the error message.

ychebotarev: AbortedByWorkflowClosingErr is of type "error", and thus ErrorAs should work? At least that is how I undrestand it 
* AbortedByWorkflowClosingErr is wrapped into that NotFound
* "errors.As" check wrapped errors
* and I though suite.ErrorAs is a wrapper around "errors.As"

stephanos: Couple of things I tried out now:

- `s.ErrorAs(updateResult.err, &update.AbortedByWorkflowClosingErr)` passes
- however; `s.ErrorAs(updateResult.err, &update.AbortedByServerErr)` _also passes_ despite `AbortedByServerErr` being `serviceerror.Unavailable`; and even worse, it overrides `update.AbortedByServerErr` with `updateResult.err`
- `s.ErrorIs(updateResult.err, update.AbortedByWorkflowClosingErr)` fails
- `s.Equal(updateResult.err, &update.AbortedByWorkflowClosingErr)` fails

So the two assertions are the only way I know how to ensure (1) the type and (2) the message are correct.

ychebotarev: >however; s.ErrorAs(updateResult.err, &update.AbortedByServerErr) also passes despite AbortedByServerErr being serviceerror.Unavailable; and even worse, it overrides update.AbortedByServerErr with updateResult.err

mmm this one is weird. I'm fine to postopen this, but will be nice to figure out at some point. "errorAs" is better then comparing strings.

stephanos: Is it? The docs say

> ErrorAs asserts that at least one of the errors in err's chain matches target, and if so, sets target to that error value.

"matches target" here is only referring to the error type (NotFound and Unavailable have the same error type).

PR #7892, opened 2025-06-10
Enable buf-breaking check

prathyushpv:


Makefile
```diff
@@ -14,6 +14,7 @@ ci-build-misc: \
 	clean-tools \
 	proto \
 	go-generate \
+	buf-breaking \
```
dnr, line 17: Can we put a comment here that says to not remove this, and add ignores to `proto/internal/buf.yaml` instead?

prathyushpv: Sure.

prathyushpv: Added a comment.

dnr (approved):

PR #7895, opened 2025-06-10
Add config flag for replication rate limiter

yux0:
## What changed?
Add config flag for replication rate limiter

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/replication/stream_sender.go
+++ b/service/history/replication/stream_sender.go
@@ -549,7 +549,7 @@ Loop:
 					// continue to send task if wait operation times out.
 				}
 			}
-			if task.Priority == enumsspb.TASK_PRIORITY_LOW {
+			if s.config.ReplicationEnableRateLimit() && task.Priority == enumsspb.TASK_PRIORITY_LOW {
 				nsName, err := s.shardContext.GetNamespaceRegistry().GetNamespaceName(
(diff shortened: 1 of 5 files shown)


service/history/replication/stream_sender.go
```diff
@@ -549,14 +549,15 @@ Loop:
...
+			if s.config.ReplicationEnableRateLimit() && task.Priority == enumsspb.TASK_PRIORITY_LOW {
 				nsName, err := s.shardContext.GetNamespaceRegistry().GetNamespaceName(
 					namespace.ID(item.GetNamespaceID()),
 				)
 				if err != nil {
 					// if there is error, then blindly send the task, better safe than sorry
 					nsName = namespace.EmptyName
 				}
+				rlStartTime := time.Now().UTC()
 				if err := s.ssRateLimiter.Wait(s.server.Context(), quotas.NewRequest(
```
xwduan, line 579: Maybe we can use this to replace the rate limiter inside `stream_sender_flow_controller`

xwduan (approved):

service/history/replication/stream_sender.go
yux0: atm, I think we need both.

PR #7898, opened 2025-06-11
Initial implementation for ListWorkers/RecordWorkerHeartbeat

ychebotarev:
## What changed?
Add initial implementation for
* ListWorkers
* RecordWorkerHeartbeat

I close this in favor of https://github.com/temporalio/temporal/pull/7914

## How did you test it?
- [x] added new unit test(s)

### DIFF
--- a/service/matching/workers/registry.go
+++ b/service/matching/workers/registry.go
@@ -2,18 +2,18 @@ package workers
 
 import (
 	"go.temporal.io/api/serviceerror"
-	workersb "go.temporal.io/api/worker/v1"
+	workerpb "go.temporal.io/api/worker/v1"
 	"go.temporal.io/server/common/log"
 )
 
 type (
 	Registry interface {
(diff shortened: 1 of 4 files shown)


service/matching/workers/registry.go
```diff
@@ -0,0 +1,49 @@
...
+		logger:       logger,
+	}
+}
+
+func (r *registryImpl) RecordWorkerHeartbeat(nsID string, workerHeartbeat *workerpb.WorkerHeartbeat) {
+	if r.workersStore[nsID] == nil {
+		r.workersStore[nsID] = make(map[string]*workerpb.WorkerHeartbeat)
+	}
+
+	r.workersStore[nsID][workerHeartbeat.WorkerInstanceKey] = workerHeartbeat
```
dnr, line 34: This just adds and never deletes? I don't think we can release it like that.. maybe this should be the start of development on a feature branch and not go to main immediately?

ychebotarev: feature is protected by feature flag, per namespace. I want to get to "all components are connected" state asap. We will discuss this

PR #7902, opened 2025-06-11
support background entry removal for workflow cache

alfred-landrum:
## What changed?
This adds support to the lru cache to actively expire entries older than the TTL, by spawning a background goroutine that periodically deletes old entries. A dynamic config, off by default, is added that can enable this feature for the workflow cache.

## Why?
This can reduce the memory usage, and associated Go GC resource usage, for workflow entries that won't be utilized since they are past their TTL.

## How did you test it?
- [x] built
- [x] added new unit test(s)

Also running this in test setups to verify expected memory reduction.

### DIFF
--- a/common/cache/lru.go
+++ b/common/cache/lru.go
@@ -2,12 +2,15 @@ package cache
 
 import (
 	"container/list"
+	"context"
 	"sync"
 	"time"
 
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/common/clock"
+	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/goro"
 	"go.temporal.io/server/common/metrics"
 )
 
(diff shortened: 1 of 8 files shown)


common/cache/lru.go
```diff
@@ -424,3 +444,57 @@ func (c *lru) updateEntryRefCount(entry *entryImpl) {
...
+func (c *lru) Stop() {
+	c.loops.Cancel()
+}
+
+func (c *lru) evictLoop(ctx context.Context) error {
+	ch, t := c.timeSource.NewTimer(c.activeExpiry().LoopInterval)
+	for {
+		select {
+		case <-ch:
+			if c.activeExpiry().Enabled {
```
dnr: tiny optimization: maybe grab the settings once here and pass MaxEntryPerCall into evict?

common/cache/lru.go
```diff
@@ -424,3 +442,54 @@ func (c *lru) updateEntryRefCount(entry *entryImpl) {
...
+			return ctx.Err()
+		}
+	}
+}
+
+func (c *lru) bgEvict(settings dynamicconfig.CacheBackgroundEvictSettings) {
+	now := c.timeSource.Now().UTC()
+
+	// Limit each iteration to scanning MaxEntryPerCall entries, to avoid holding the cache lock for too long.
+	evictToMax := func() (again bool) {
```
prathyushpv, line 470: This is nice!

prathyushpv (approved):

PR #7904, opened 2025-06-11
Fix handling first replication task for state based replication

xwduan:
## What changed?
Use bringLocalEventsUpToSourceCurrentBranch to handle first replication task. 

## Why?
1. Consolidate the event process logic for state based replication task. 
2. Before this change, when handling first replication task, Event is only added to cache before refresh task(need to retrieve the event). In a very rare case, the event is evict from cache which will cause the refresh task fail, and replication task will go to DLQ.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
n/a


service/history/ndc/workflow_state_replicator.go
```diff
@@ -980,36 +894,12 @@ func (r *WorkflowStateReplicatorImpl) bringLocalEventsUpToSourceCurrentBranch(
...
-			lcaItem.GetEventId(),
-			newVersionHistory,
-		)
-		if err != nil {
-			return 0, false, err
-		}
-
-		return newVersionHistoryIndex, true, nil
-	}()
+	index, isNewBranch, err := r.getBranchToAppend(
```
hai719, line 903: if isNewBranch, should we delete the new branch in case of error for non-first replication task case?

xwduan: Added.

hai719 (approved):

PR #7906, opened 2025-06-12
Fix DescribeTaskQueue Stats for versioning

stephanos:
## What changed?

Fix DescribeTaskQueue's stats reporting; versioned activities were counted incorrectly.

## Why?

Bug.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


tests/describe_task_queue_test.go
```diff
@@ -63,38 +66,33 @@ func (s *DescribeTaskQueueSuite) TestAddSingleTaskPerVersion_ValidateStats() {
...
 
-// TODO(stephanos): re-enable this test
-// func (s *DescribeTaskQueueSuite) TestAddSingleTaskPerVersion_ValidateCachedStats_NoMatchingBehaviour() {
-//	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
-//	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
-//	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 500*time.Millisecond)
-//
-//	s.publishConsumeWorkflowTasksValidateStatsCached(1, true)
-// }
-
```
stephanos, line 95: I didn't really understand the benefit of this test. It seems redundant to me?

service/matching/task_queue_partition_manager.go
```diff
@@ -686,7 +686,11 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 			if !found {
 				// still add it as a v2 version because user explicitly asked for the stats, we'd
 				// make sure to load the TQ.
-				versions[PhysicalTaskQueueVersion{buildId: b}] = true
+				dv := worker_versioning.ExternalWorkerDeploymentVersionFromStringV31(b)
```
stephanos, line 689: Is this the right helper to call? There are too many :P

Furthermore, do we need to guard against a parsing issue here? I suspect it happens earlier. Will have to check ...

service/matching/task_queue_partition_manager.go
```diff
@@ -686,7 +686,11 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+				versions[PhysicalTaskQueueVersion{
+					buildId:              dv.BuildId,
+					deploymentSeriesName: dv.DeploymentName,
```
stephanos, line 692: This line is the fix!

tests/describe_task_queue_old_test.go
stephanos, line 1: The old suite only tests unversioned; the new suite tests both!

service/matching/matching_engine.go
```diff
@@ -1294,7 +1294,8 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 				if v.GetVersion() == nil || v.GetVersion().GetDeploymentName() == "" || v.GetVersion().GetBuildId() == "" {
 					continue
 				}
-				buildIds = append(buildIds, worker_versioning.WorkerDeploymentVersionToStringV31(v.GetVersion()))
+				buildId := worker_versioning.WorkerDeploymentVersionToStringV32(v.GetVersion())
```
stephanos, line 1297: Use v3.2 here.

common/worker_versioning/worker_versioning.go
```diff
@@ -772,6 +779,26 @@ func WorkerDeploymentVersionFromStringV31(s string) (*deploymentspb.WorkerDeploy
 	}, nil
 }
 
+func WorkerDeploymentVersionFromStringV32(s string) (*deploymentspb.WorkerDeploymentVersion, error) {
+	if s == UnversionedVersionId {
+		return nil, nil
+	}
```
stephanos, line 785: I ported this over from `WorkerDeploymentVersionFromStringV31`; it's a bit surprising that it returns `nil, nil`, but I didn't want to break any existing contracts/expectations so it can be used as a drop-in replacement.

service/matching/task_queue_partition_manager.go
```diff
@@ -684,9 +684,17 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+				if dv != nil {
+					// Add v3 version.
+					versions[PhysicalTaskQueueVersion{
+						buildId:              dv.BuildId,
+						deploymentSeriesName: dv.DeploymentName,
+					}] = true
+				} else {
+					// Still add v2 version because user explicitly asked for the stats, we'd make sure to load the TQ.
+					versions[PhysicalTaskQueueVersion{buildId: b}] = true
+				}
```
stephanos, line 697: This is the fix.

tests/describe_task_queue_old_test.go
stephanos, line 1: Redundant since `tests/describe_task_queue_test.go` tests this and versioning, too, now.

tests/describe_task_queue_test.go
ShahabT: do we have other test to cover cache behavior?

stephanos: As discussed internally, I'll add a new test for caching as a follow-up.

PR #7911, opened 2025-06-12
Add workflowpb.PostResetOperation to ResetParams in batch operations

gow:
## What changed?
We had missed copying over the `BatchOperationReset.PostResetOperation` to `ResetParams` in the workflow handler. Additionally since this proto message contains `oneof` we need to perform manual protobuf marshal/unmarshal to avoid json serialization/deserialization issues.

## Why?
This parameter is needed to support post reset operations like updating versioning in a batch operation.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
N/A

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -4635,12 +4635,12 @@ func (wh *WorkflowHandler) StartBatchOperation(
 			}
 			resetParams.ResetOptions = encodedResetOptions
 			resetParams.PostResetOperations = make([][]byte, len(op.ResetOperation.PostResetOperations))
-			for _, postResetOperation := range op.ResetOperation.PostResetOperations {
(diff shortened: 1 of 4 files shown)


ShahabT:
@gow thanks for the fix, is this needed for batch reset-with-move to work? if so we need to ensure it's labeled to be included in the OSS release.

carlydf:
> @gow thanks for the fix, is this needed for batch reset-with-move to work? if so we need to ensure it's labeled to be included in the OSS release.

It is required for batch reset-with-move to work

gow:
> The unit tests look good -- are you planning to add a functional test too or just these?

Yes. Adding functional tests as well.

carlydf (approved):
approved pending a functional test!

gow:
> approved pending a functional test!

The functional tests will take some more time. I'll land this one for now and follow up with another PR.

PR #7913, opened 2025-06-13
Allow reading history reverse when mutable state does not have LastFirstEventTxnId set

xwduan:
## What changed?
Allow reading history reverse when mutable state does not have LastFirstEventTxnId set
## Why?
LastFirstEventTxnId maybe not set because of replication.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
No risk.

### DIFF
--- a/common/persistence/history_manager.go
+++ b/common/persistence/history_manager.go
@@ -1083,6 +1083,9 @@ func (m *executionManagerImpl) filterHistoryNodesReverse(
 		if lastNodeID == defaultLastNodeID {
 			lastNodeID = node.NodeID
 		}
+		if lastTransactionID == 0 {
+			m.logger.Warn("lastTransactionID is not set, this should not happen")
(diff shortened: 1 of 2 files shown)


common/persistence/history_manager.go
```diff
@@ -1083,7 +1083,8 @@ func (m *executionManagerImpl) filterHistoryNodesReverse(
 		if lastNodeID == defaultLastNodeID {
 			lastNodeID = node.NodeID
 		}
-		if node.TransactionID != lastTransactionID {
+		if lastTransactionID != 0 && // in the case where the lastTransactionID is not set, we will not compare
```
yycptt, line 1089: is it possible to add a log for the lastTransactionID == 0 case?

yycptt (approved):
We'll need to backfill a functional test for this later.

PR #7914, opened 2025-06-13
Workers registry

ychebotarev:
## What changed?
Add worker registry. 
Now this is functional class, with support for eviction by TTL and by capacity.

## Why?
Properly store workers heartbeat, while also allowing to process them per namespace.
ListWorkers has predicate, that will be used after "query" parting in ListWorkers API call.

## How did you test it?
- [X] added new unit test(s)

## Potential risks
Not used.


service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	workerpb "go.temporal.io/api/worker/v1"
+)
+
+// entry wraps a WorkerHeartbeat along with its namespace and eviction metadata.
+type entry struct {
+	namespace string
```
dnr: maybe we should use namespace id instead? in any case please use the types `namespace.Name` or `namespace.ID` to avoid confusion

service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+func newBucket() *bucket {
+	return &bucket{
+		namespaces: make(map[string]map[string]*entry),
+		order:      list.New(),
+	}
+}
+
+// update inserts or refreshes a WorkerHeartbeat under the given namespace.
+// Returns true if a brand-new entry was created.
+func (b *bucket) update(namespace string, hb *workerpb.WorkerHeartbeat) bool {
```
dnr: use namespace types (and everywhere below)

service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Registry contains all worker heartbeats.
+// It partitions the keyspace into buckets and enforces TTL and capacity.
+// Eviction runs in the background.
+type Registry struct {
+	buckets     []*bucket
+	maxItems    int64
+	TTL         time.Duration
+	MinEvictAge time.Duration
```
dnr: any particular reason these are public?

service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+	maxItems int64,
+	evictionInterval time.Duration,
+) *Registry {
+	m := &Registry{
+		buckets:     make([]*bucket, numBuckets),
+		maxItems:    maxItems,
+		TTL:         ttl,
+		MinEvictAge: minEvictAge,
+		ticker:      time.NewTicker(evictionInterval),
+		hashFn: func(s string) uint32 {
```
dnr: fnv is a pretty weak hash.. why not use `maphash.String`? as a bonus, it's seeded to help avoid dos attacks (forcing everything to the same bucket)

service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+			h := fnv.New32a()
+			h.Write([]byte(s)) //nolint:revive
+			return h.Sum32()
+		},
+		quit: make(chan struct{}),
+	}
+	for i := range m.buckets {
+		m.buckets[i] = newBucket()
+	}
+	go m.evictLoop()
```
dnr: In general most objects with background goroutines support `Start()` and `Stop()` methods and only create goroutines in Start, not the constructor. that's more predictable and makes it easy to integrate with fx lifecycle.

service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+
+// evictLoop periodically triggers TTL and capacity-based eviction.
+func (m *Registry) evictLoop() {
+	for {
+		select {
+		case <-m.ticker.C:
+			m.evictByTTL()
+			m.evictByCapacity()
+		case <-m.quit:
+			m.ticker.Stop()
```
dnr: you don't actually need this anymore since go 1.23, but it's harmless

service/matching/workers/registry.go
```diff
@@ -0,0 +1,228 @@
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	mp := b.namespaces[namespace]
+	if mp == nil {
+		return nil
+	}
+	out := make([]*workerpb.WorkerHeartbeat, 0, len(mp))
+	for _, e := range mp {
+		if predicate(e.hb) {
```
dnr: how do you plan to do limit/paging tokens here? just return the whole list each time and let the caller do limit/offset?

service/matching/workers/registry.go
ychebotarev: changed to namespace.ID

service/matching/workers/registry.go
ychebotarev: changed to namespace.ID

service/matching/workers/registry.go
ychebotarev: nope. Remove capitalization.

service/matching/workers/registry.go
ychebotarev: changed to maphash

service/matching/workers/registry.go
ychebotarev: I was planning to do that separately.
But now I combine that with the other PR, and add lifecycle support.

service/matching/workers/registry.go
ychebotarev: then lets keep it

service/matching/workers/registry.go
ychebotarev: I plan to do pagination in this loop. 
Something like 
```
if token != nil {
  if pos < token {
   continue
  } elif  pos > token + page_size { 
     break
  } else {
    add _to_the _result()
}
```

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+	"go.temporal.io/server/common/namespace"
+	"go.uber.org/fx"
+)
+
+const (
+	defaultBuckets          = 10
+	defaultEntryTTL         = 24 * time.Hour
+	defaultMinEvictAge      = 10 * time.Minute
+	defaultMaxEntries       = 1_000_000
+	defaultEvictionInterval = 1 * time.Hour
```
dnr, line 20: maybe some/all of these should be from dynamic config? (even if you just read it once at startup)

also I'd package these settings into a struct (which makes it very easy to do dynamic config)

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+	lc.Append(fx.Hook{
+		OnStart: func(ctx context.Context) error {
+			m.Start()
+			return nil
+		},
+		OnStop: func(ctx context.Context) error {
+			m.Stop()
+			return nil
+		},
+	})
```
dnr:
```suggestion
	lc.Append(fx.StartStopHook(m.Start, m.Stop))
```

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+	m := &registryImpl{
+		buckets:          make([]*bucket, numBuckets),
+		maxItems:         maxItems,
+		tTL:              ttl,
+		minEvictAge:      minEvictAge,
+		evictionInterval: evictionInterval,
+		hasher:           &maphash.Hash{},
+		quit:             make(chan struct{}),
+	}
+	m.hasher.Reset()
```
dnr: I don't think this is needed?

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+		m.buckets[i] = newBucket()
+	}
+	return m
+}
+
+// bucketFor hashes the namespace to select a bucket.
+func (m *registryImpl) getBucket(nsID namespace.ID) *bucket {
+	m.hasher.Reset()
+	m.hasher.WriteString(nsID.String()) //nolint:revive
+	hs := m.hasher.Sum64()
```
dnr: https://pkg.go.dev/hash/maphash#Hash
> A Hash is not safe for concurrent use by multiple goroutines, but a Seed is. If multiple goroutines must compute the same seeded hash, each can declare its own Hash and call SetSeed with a common Seed.

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+// upsertHeartbeat records or refreshes a WorkerHeartbeat under the given namespace.
+// New entries increment the global counter.
+func (m *registryImpl) upsertHeartbeat(nsID namespace.ID, hb *workerpb.WorkerHeartbeat) {
+	b := m.getBucket(nsID)
+	if newEntry := b.upsertHeartbeat(nsID, hb); newEntry {
+		m.total.Add(1)
+	}
+}
+
+// listWorkers returns all WorkerHeartbeats in a namespace
```
dnr: mismatch with function name

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+// for which predicate(hb) returns true.
+func (m *registryImpl) filterWorkers(
+	nsID namespace.ID,
+	predicate func(*workerpb.WorkerHeartbeat) bool,
+) []*workerpb.WorkerHeartbeat {
+	b := m.getBucket(nsID)
+	b.mu.Lock()
+	defer b.mu.Unlock()
```
dnr: the body of this should probably be moved into a method on `bucket`? the registry shouldn't be locking+manipulating the bucket

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+		order      *list.List // front = oldest, back = newest
+	}
+
+	// registryImpl implements Registry interface. It contains all worker heartbeats.
+	// It partitions the keyspace into buckets and enforces TTL and capacity.
+	// Eviction runs in the background.
+	registryImpl struct {
+		buckets          []*bucket     // buckets for partitioning the keyspace
+		maxItems         int64         // maximum number of entries allowed across all buckets
+		tTL              time.Duration // time after which entries are considered expired
```
dnr:
```suggestion
		ttl              time.Duration // time after which entries are considered expired
```

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+func (m *registryImpl) evictByCapacity() {
+	for m.total.Load() > m.maxItems {
+		removedAny := false
+		threshold := time.Now().Add(-m.minEvictAge)
+		for _, b := range m.buckets {
+			if m.total.Load() <= m.maxItems {
+				return
+			}
+			b.mu.Lock()
+			front := b.order.Front()
```
dnr: there's no fairness across namespaces here? I guess that's fine for this purpose

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// Stop halts background eviction.
+func (m *registryImpl) Stop() {
+	close(m.quit)
+}
+
+func (m *registryImpl) RecordWorkerHeartbeat(nsID namespace.ID, workerHeartbeat *workerpb.WorkerHeartbeat) {
+	m.upsertHeartbeat(nsID, workerHeartbeat)
+}
```
dnr, line 286:
```suggestion
}

```

service/matching/workers/registry_impl.go
```diff
@@ -0,0 +1,278 @@
...
+// evictByCapacity removes entries older than MinEvictAge until under capacity.
+func (m *registryImpl) evictByCapacity() {
+	for m.total.Load() > m.maxItems {
+		removedAny := false
+		threshold := time.Now().Add(-m.minEvictAge)
+		for _, b := range m.buckets {
+			if m.total.Load() <= m.maxItems {
+				return
+			}
+			b.mu.Lock()
```
dnr: move this logic into bucket too? it feels weird that some methods have registryImpl calling bucket to do the work, and some have registryImpl doing the work itself by reaching into bucket. either pattern is okay but let's pick one

service/matching/workers/registry_impl.go
ychebotarev: removed

service/matching/workers/registry_impl.go
ychebotarev: you right, no fairness. I don't want to overcomplicate this atm.

service/matching/workers/registry_impl.go
ychebotarev: moved to the bucket

service/matching/workers/registry_impl.go
ychebotarev: moved to bucket

service/matching/workers/registry_impl.go
ychebotarev: I was planning to do that later, after I add metrics. Yes, at least "max entries" and "evict interval" should be configurable.

service/matching/workers/registry_impl.go
ychebotarev: switched to the seed

dnr (approved):

service/matching/handler.go
```diff
@@ -535,16 +539,32 @@ func (h *Handler) ListNexusEndpoints(ctx context.Context, request *matchingservi
 
 // RecordWorkerHeartbeat receive heartbeat request from the worker.
 func (h *Handler) RecordWorkerHeartbeat(
-	context.Context, *matchingservice.RecordWorkerHeartbeatRequest,
+	_ context.Context, request *matchingservice.RecordWorkerHeartbeatRequest,
 ) (*matchingservice.RecordWorkerHeartbeatResponse, error) {
-	return nil, serviceerror.NewUnimplemented("RecordWorkerHeartbeat is not implemented")
+	nsID := namespace.ID(request.GetNamespaceId())
+	h.workersRegistry.RecordWorkerHeartbeat(nsID, request.GetHeartbeartRequest().GetWorkerHeartbeat())
```
sivagirish81, line 545: @ychebotarev : I am facing some issues in my local build when I try to bring up the server after these changes.
GetWorkerHeartBeat is expected to return a slice of []*v110.WorkerHeartbeat and the workerRegistry.RecordWorkerHeartbeat expects a pointer (*workerpb.WorkerHeartbeat).
I think the workerRegistry implementation needs to be updated

sivagirish81, line 545: @ychebotarev : The worker_heartbeat in RecordWorkerHeartbeatRequest is a repeated field. So the getter is expected to return a slice []*v110.WorkerHeartbeat but the workersRegistry.RecordWorkerHeartbeat expects a pointer to *workerpb.WorkerHeartbeat. I think the workersRegistry.RecordWorkerHeartbeat needs to be changed.

I am facing build issues on my local. There is a type mismatch due to which I am not able to bring up my server. I am not sure if I am missing something here?

PR #7918, opened 2025-06-13
Add DescribeTaskQueue Stats cache test

stephanos:
## What changed?

Added test to verify DescribeTaskQueue Stats are cached.

## Why?

Regression testing.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -1142,25 +1142,41 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 
 		// TODO bug fix: We cache the last response for each build ID. timeSinceLastFanOut is the last fan out time, that means some enteries in the cache can be more stale if
 		// user is calling this API back-to-back but with different version selection.
+		cacheKeyFunc := func(buildId string, taskQueueType enumspb.TaskQueueType) string {
+			return fmt.Sprintf("%s.%s", buildId, taskQueueType.String())
+		}
 		missingItemsInCache := false
-		physicalInfoByBuildId := make(map[string]map[enumspb.TaskQueueType]*taskqueuespb.PhysicalTaskQueueInfo)
+		physicalTqInfos := make(map[string]map[enumspb.TaskQueueType]*taskqueuespb.PhysicalTaskQueueInfo)
 		//nolint:staticcheck // SA1019 deprecated
 		requestedBuildIds, err := e.getBuildIds(req.Versions)
 		if err != nil {
 			return nil, err
 		}
-		for b := range requestedBuildIds {
-			c := rootPM.GetCache(b) // any expired cache entry will return nil
-			if c == nil {
-				missingItemsInCache = true
-				break // once we find a missing item, we can stop checking the cache
+
+		for buildId := range requestedBuildIds {
+			physicalTqInfos[buildId] = make(map[enumspb.TaskQueueType]*taskqueuespb.PhysicalTaskQueueInfo)
+			//nolint:staticcheck // SA1019 deprecated
+			for _, taskQueueType := range req.TaskQueueTypes {
+				cacheKey := cacheKeyFunc(buildId, taskQueueType)
+				cachedInfo := rootPM.GetCache(cacheKey) // any expired cache entry will return nil
+				if cachedInfo == nil {
+					missingItemsInCache = true
+					break // once we find a missing item, we can stop checking the cache
+				}
+				//revive:disable-next-line:unchecked-type-assertion
+				physicalTqInfos[buildId][taskQueueType] = cachedInfo.(*taskqueuespb.PhysicalTaskQueueInfo)
 			}
-			//revive:disable-next-line:unchecked-type-assertion
-			physicalInfoByBuildId[b] = c.(map[enumspb.TaskQueueType]*taskqueuespb.PhysicalTaskQueueInfo)
+			if missingItemsInCache {
+				break // stop checking other build IDs if we already found missing items
+			}
 		}
+
 		if missingItemsInCache {
 			// Fan out to partitions to get the needed info
-			var foundBuildIds []string
+			var foundItems []struct {
+				buildId       string
+				taskQueueType enumspb.TaskQueueType
+			}
 			numPartitions := max(tqConfig.NumWritePartitions(), tqConfig.NumReadPartitions())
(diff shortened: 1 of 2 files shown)


service/matching/matching_engine.go
```diff
@@ -1179,45 +1194,49 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
 					}
 				}
 			}
 
-			for _, b := range foundBuildIds {
-				rootPM.PutCache(b, physicalInfoByBuildId[b])
+			// put the found items into cache
+			for _, item := range foundItems {
+				physicalTqInfo := physicalTqInfos[item.buildId][item.taskQueueType]
+				rootPM.PutCache(cacheKeyFunc(item.buildId, item.taskQueueType), physicalTqInfo)
```
stephanos, line 1234: Changed the cache to use both buildId + task queue type; not just buildId. I'm making the change despite this being deprecated as it's easier to test it this way (and gives more predictable results to users, too).

tests/describe_task_queue_test.go
```diff
@@ -78,12 +95,39 @@ func (s *DescribeTaskQueueSuite) TestAddMultipleTasks_MultiplePartitions_Validat
 	s.publishConsumeWorkflowTasksValidateStats(50, false) // 25 unversioned, 25 versioned
 }
 
-func (s *DescribeTaskQueueSuite) TestAddSingleTaskPerVersion_SinglePartition_ValidateStats() {
-	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
-	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
+// NOTE: Cache _eviction_ is already covered by the other tests.
+func (s *DescribeTaskQueueSuite) TestAddMultipleTasks_MultiplePartitions_ValidateStats_Cached() {
```
stephanos, line 101: This is the only net new test. Everything else is just a refactoring.

service/matching/matching_engine.go
```diff
@@ -1142,25 +1142,41 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
+				if cachedInfo == nil {
+					missingItemsInCache = true
+					break // once we find a missing item, we can stop checking the cache
+				}
+				//revive:disable-next-line:unchecked-type-assertion
+				physicalTqInfos[buildId][taskQueueType] = cachedInfo.(*taskqueuespb.PhysicalTaskQueueInfo)
+			}
+			if missingItemsInCache {
+				break // stop checking other build IDs if we already found missing items
 			}
```
Shivs11, line 1171: I know most of your work on this was mostly refactoring + adding some more functionality, but do you just feel that this function (`DescribeTaskQueue`) is extremely hard to read + daunting? I wonder if we can not have it to be that way 


maybe we can have this go into a helper since it's doing something not directly related to describing the task-queue itself (checking if the cache is warm?)

service/matching/matching_engine.go
```diff
@@ -1142,25 +1142,41 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
-			physicalInfoByBuildId[b] = c.(map[enumspb.TaskQueueType]*taskqueuespb.PhysicalTaskQueueInfo)
 		}
+
 		if missingItemsInCache {
 			// Fan out to partitions to get the needed info
-			var foundBuildIds []string
+			var foundItems []struct {
+				buildId       string
+				taskQueueType enumspb.TaskQueueType
+			}
```
Shivs11, line 1179: nit: any specific reason for making this inline? seems cleaner to define it external to this function (although I get it, the scope is tied to this func)

wdyt?

tests/describe_task_queue_test.go
```diff
@@ -92,31 +136,82 @@ func (s *DescribeTaskQueueSuite) publishConsumeWorkflowTasksValidateStats(workfl
...
+	}
+
+	s.validateDescribeTaskQueue(tqName, expectations, singlePartition)
+}
+
+func (s *DescribeTaskQueueSuite) enqueueWorkflows(count int, tqName string) {
 	deploymentOpts := s.deploymentOptions()
 
-	// manually add worker deployment version
 	_, err := s.GetTestCluster().MatchingClient().SyncDeploymentUserData(
```
Shivs11, line 227: I think we should have a comment over here as to why we are playing with the userData of a task-queue before starting workflow executions - think it took me a second to realize but i think this is just required because a version has to be created before you can kickstart adding tasks to it, right?

tests/describe_task_queue_test.go
Shivs11: looking at the overall suite, I noticed (based solely of the name):

`TestAddSingleTaskPerVersion_ValidateStats` : seems to be testing adding a single task across multiple versions
 `TestAddMultipleTasks_MultiplePartitions_ValidateStats`: no information on the number of "versions" here

was this intended/did I miss something here? are they completely different tests or is the second one just a scale  up of the first one?

Shivs11: the second test also did not seem to use `s.RunTestWithMatchingBehavior(func() {` hence the above question ^

tests/describe_task_queue_test.go
```diff
@@ -92,31 +136,82 @@ func (s *DescribeTaskQueueSuite) publishConsumeWorkflowTasksValidateStats(workfl
 		s.T().Fatal("workflows must be an even number to ensure half of them are versioned and half are unversioned")
 	}
```
Shivs11, line 149: all tests seem to be on a single partition - not sure if this was discussed internally or not, but are there any plans of having more than one partition for these tests?

Shivs11 (approved):

service/matching/matching_engine.go
stephanos: Yup, this whole thing is bad lol. I have a follow-up ticket to address that tech debt.

service/matching/matching_engine.go
stephanos: Usually I'd agree, but given the size of the file (separate type is hard to find) and the limited scope of the struct (only used within this `if` branch), I prefer this approach.

tests/describe_task_queue_test.go
stephanos: Removed "PerVersion" as discussed offline.

PR #7923, opened 2025-06-16
testhooks: add context-wrapping functionality

tdeebswihart:
## What changed?

I've added context-based methods to our testhooks package. This includes:

- `NewInjectedTestHooksImpl` - returns a context and the registry injected into its chain
- `SetCtx` - set a testhooks key on the first TestHooks implementor found in the context chain
- `GetCtx` - get a testhooks value from the first TestHooks implementor found in the context chain
- `CallCtx` - call a testhooks callback value in the first TestHooks implementor found in the context chain

## Why?

This simplifies TestHook usage for modules that have an externally-controllable context. Rather than attach a TestHooks implementation to your production structs, you can instead inject it into the context provided by your tests.

This does not work for temporal's end-to-end or functional tests at the moment, though a gRPC interceptor that injects a registry would not be difficult to write.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
None

### DIFF
--- a/common/testing/testhooks/test_impl.go
+++ b/common/testing/testhooks/test_impl.go
@@ -3,6 +3,7 @@
 package testhooks
 
 import (
+	"context"
 	"sync"
 
 	"go.uber.org/fx"
@@ -30,11 +31,13 @@ type (
 	testHooksImpl struct {
 		m sync.Map
 	}
+
+	contextKey struct{}
 )
 
 // Get gets the value of a test hook from the registry.
 //
-// TestHooks should be used very sparingly, see comment on TestHooks.
+// TestHooks should be used sparingly, see comment on TestHooks.
 func Get[T any](th TestHooks, key Key) (T, bool) {
 	var zero T
 	if th == nil {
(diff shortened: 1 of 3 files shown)


common/testing/testhooks/test_impl.go
```diff
@@ -47,29 +50,71 @@ func Get[T any](th TestHooks, key Key) (T, bool) {
 	return zero, false
 }
 
+// GetCtx gets the value of a test hook from the registry embedded in the
+// context chain.
+//
+// TestHooks should be used sparingly, see comment on TestHooks.
+func GetCtx[T any](ctx context.Context, key Key) (T, bool) {
```
dnr, line 57: Don't you need noop impls for GetCtx and CallCtx?

tdeebswihart: I do, yeah. I paused work on this (hence the `draft` status) until later this week

tdeebswihart: Alright, added. And marked as ready for review

PR #7924, opened 2025-06-16
CHASM: Handle empty version history in execution manager

yycptt:
## What changed?
- Handle empty version history in execution manager

## Why?
- CHASM workstream. CHASM runs may not have events at all.

## How did you test it?
- [x] built
- [x] added new persistence test(s)


lina-temporal (approved):

common/persistence/execution_manager.go
```diff
@@ -1107,11 +1109,33 @@ func getCurrentBranchLastWriteVersion(
 	if err != nil {
```
xwduan, line 1109: Q: Is it guaranteed that empty version history always have current version history?

PR #7928, opened 2025-06-17
Fix QPS calculation in force replication wf

zigeH:
## What changed?
This PR fixes two issues in the QPS calculation logic:
Incorrect accounting of the number of workflow executions processed in each batch.
Incomplete execution of verification select futures, leading to missed QPS updates.

## Why?
Previous QPS calculation was wrong.

## How did you test it?
- [ ] built
- [x] added new unit test(s)


service/worker/migration/force_replication_workflow.go
```diff
@@ -476,6 +477,16 @@ func enqueueReplicationTasks(ctx workflow.Context, workflowExecutionsCh workflow
...
+	for totalPendingTasks > 0 {
+		// Wait for all in-flight activities to complete
+		selector.Select(ctx)
+		totalPendingTasks -= 1
+		if lastActivityErr != nil {
+			return lastActivityErr
+		}
+	}
+
 	for _, future := range futures {
```
yux0: you don't need this anymore?

pglass: yes, i think we don't need the `futures` variable anymore (can remove it)

service/worker/migration/force_replication_workflow_test.go
```diff
@@ -95,7 +95,7 @@ func TestForceReplicationWorkflow(t *testing.T) {
 	assert.True(t, status.TaskQueueUserDataReplicationStatus.Done)
 	assert.Equal(t, "", status.TaskQueueUserDataReplicationStatus.FailureMessage)
 	assert.Equal(t, int64(10), status.TotalWorkflowCount)
-	assert.Equal(t, int64(0), status.ReplicatedWorkflowCount)
+	assert.Equal(t, int64(4), status.ReplicatedWorkflowCount)
```
meiliang86, line 98: Why is this 4? Shouldn't ReplicatedWorkflowCount be equal to TotalWorkflowCount?

service/worker/migration/force_replication_workflow.go
zigeH: agree

service/worker/migration/force_replication_workflow_test.go
zigeH: This test case wasn't well designed — it only includes 4 page tokens, with one workflow per page. I should have set the total to 4 and implemented proper match logic.

service/worker/migration/force_replication_workflow.go
pglass: We can also remove `var futures []workflow.Future` on L402 I think (unless I've overlooked something)

service/worker/migration/force_replication_workflow.go
```diff
@@ -448,11 +448,12 @@ func enqueueReplicationTasks(ctx workflow.Context, workflowExecutionsCh workflow
...
 				pendingVerifyTasks--
 
-				if err := f.Get(ctx, nil); err != nil {
+				var verifyTaskResponse verifyReplicationTasksResponse
+				if err := f.Get(ctx, &verifyTaskResponse); err != nil {
 					lastActivityErr = err
 				} else {
 					// Update replication status
-					params.ReplicatedWorkflowCount += int64(len(workflowExecutions))
+					params.ReplicatedWorkflowCount += int64(verifyTaskResponse.VerifiedWorkflowCount)
```
hehaifengcn, line 454: would it be an issue if verify is disabled?

zigeH: True. either way the calculation is under if EnableVerification is true. (line 434)

service/worker/migration/force_replication_workflow.go
zigeH: removed

meiliang86 (approved):

PR #7929, opened 2025-06-17
[CHASM] fix makeValidationFn parameters

lina-temporal:
## What changed?
- Bug fix to makeValidationFn.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -2250,6 +2250,12 @@ func (n *Node) ExecuteSideEffectTask(
 		return serviceerror.NewInternalf("failed to decode path '%s'", taskInfo.Path)
 	}
 
+	taskValue, err := deserializeTask(registrableTask, taskInfo.Data)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


chasm/tree.go
```diff
@@ -2250,22 +2250,22 @@ func (n *Node) ExecuteSideEffectTask(
 		return serviceerror.NewInternalf("failed to decode path '%s'", taskInfo.Path)
 	}
 
+	taskValue, err := deserializeTask(registrableTask, taskInfo.Data)
+	if err != nil {
+		return err
+	}
+	taskInstance := taskValue.Interface()
```
yycptt: taskValue is already reflect.Value I think, and can be passed directly into makeValidationFn and used in fn.Call()

yycptt (approved):

chasm/tree.go
lina-temporal: thanks, updated to avoid the needless `.Interface()`

PR #7871, opened 2025-06-06
Collect OTEL traces from CI tests

stephanos:
## What changed?

Collecting OTEL traces from functional tests. And packaging them up as an artifact. 

## Why

Help debug flaky tests.

## Testing

You can see the trace artifacts on [this run](https://github.com/temporalio/temporal/actions/runs/15719131802?pr=7871).

PR #7931, opened 2025-06-18
Enable OTEL for sqlite GHA test runs

stephanos:
## What changed?

Follow-up to https://github.com/temporalio/temporal/pull/7871; covers sqlite tests as well.

## Why?

Complete coverage.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -472,7 +472,7 @@ jobs:
           ref: ${{ env.COMMIT }}
 
       - name: Start containerized dependencies
-        if: ${{ toJson(matrix.containers) != '[]' && (inputs.run_single_functional_test != true || (inputs.run_single_functional_test == true && contains(fromJSON(needs.set-up-single-test.outputs.dbs), env.PERSISTENCE_DRIVER))) }}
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -625,7 +625,6 @@ jobs:
           ref: ${{ env.COMMIT }}
 
       - name: Start containerized dependencies
-        if: ${{ toJson(matrix.containers) != '[]' }}
```
stephanos, line 628: Not needed; all jobs have containers. (without OTEL)

.github/workflows/run-tests.yml
```diff
@@ -776,7 +775,6 @@ jobs:
...
```
stephanos, line 779: Not needed; all jobs have containers. (without OTEL)

.github/workflows/run-tests.yml
```diff
@@ -472,7 +472,7 @@ jobs:
           ref: ${{ env.COMMIT }}
 
       - name: Start containerized dependencies
-        if: ${{ toJson(matrix.containers) != '[]' && (inputs.run_single_functional_test != true || (inputs.run_single_functional_test == true && contains(fromJSON(needs.set-up-single-test.outputs.dbs), env.PERSISTENCE_DRIVER))) }}
```
stephanos, line 475: Not needed; all jobs have containers now with the OTEL collector.

PR #7941, opened 2025-06-19
Export OTEL by test suite

stephanos:
## What changed?

Redo of https://github.com/temporalio/temporal/pull/7871.

(1) Instead of collecting OTEL trace across all suites, it tracks it per suite and only exports traces on failure.
(2) It annotates the test's RPCs with the test name.

## Why?

(1) More efficient.
(2) Allows filtering traces by test name.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

Example run with artifacts: https://github.com/temporalio/temporal/actions/runs/15786189984?pr=7941

Screenshot 2025-06-20 at 12 37 20 PM


.github/workflows/run-tests.yml
```diff
@@ -472,11 +472,11 @@ jobs:
           ref: ${{ env.COMMIT }}
 
       - name: Start containerized dependencies
-        if: ${{ inputs.run_single_functional_test != true || (inputs.run_single_functional_test == true && contains(fromJSON(needs.set-up-single-test.outputs.dbs), env.PERSISTENCE_DRIVER)) }}
+        if: ${{ toJson(matrix.containers) != '[]' && (inputs.run_single_functional_test != true || (inputs.run_single_functional_test == true && contains(fromJSON(needs.set-up-single-test.outputs.dbs), env.PERSISTENCE_DRIVER))) }}
```
stephanos, line 475: Reverted back to before.

go.mod
```diff
@@ -50,6 +50,7 @@ require (
 	github.com/uber-go/tally/v4 v4.1.17
 	github.com/urfave/cli v1.22.16
 	github.com/urfave/cli/v2 v2.27.5
+	go.opentelemetry.io/collector/pdata v1.34.0
```
stephanos, line 53: New dependency for trace serialization.

go.mod
```diff
@@ -73,8 +74,8 @@ require (
 	golang.org/x/text v0.24.0
 	golang.org/x/time v0.10.0
 	google.golang.org/api v0.224.0
-	google.golang.org/grpc v1.71.0
-	google.golang.org/protobuf v1.36.5
+	google.golang.org/grpc v1.72.2
+	google.golang.org/protobuf v1.36.6
```
stephanos, line 78: Might as well ...

tests/testcore/functional_test_base.go
```diff
@@ -270,6 +301,9 @@ func (s *FunctionalTestBase) SetupTest() {
 	s.initAssertions()
 	s.setupSdk()
 	s.taskPoller = taskpoller.New(s.T(), s.FrontendClient(), s.Namespace().String())
+	s.testCluster.host.grpcClientInterceptor.Set(func(ctx context.Context) context.Context {
+		return metadata.AppendToOutgoingContext(ctx, "temporal-test-name", s.T().Name())
+	})
```
stephanos, line 310: This is where the test name is injected.

.github/workflows/run-tests.yml
```diff
@@ -385,7 +385,7 @@ jobs:
         if: ${{ !cancelled() }}
         with:
           name: junit-xml--${{github.run_id}}--${{github.run_attempt}}--integration-test
-          path: .testoutput
+          path: ./.testoutput/junit.*.xml
```
stephanos, line 388: Need to be more specific here. This was already including the coverage files accidentally.

tests/testcore/functional_test_base.go
```diff
@@ -234,6 +241,34 @@ func (s *FunctionalTestBase) SetupSuiteWithCluster(options ...TestClusterOption)
 		s.Logger = tl
 	}
 
+	// Initialize the OTEL collector, if OTEL is enabled.
+	// Must be done before the test cluster is created, so that the collector can be used by the test cluster.
+	if otelOutputDir := os.Getenv("TEST_TEMPORAL_OTEL_OUTPUT"); otelOutputDir != "" {
+		// Create an OTEL collector.
+		collector, err := otellogger.Start(s.T())
```
stephanos, line 248: This replaces the one from the docker compose.

.github/workflows/run-tests.yml
```diff
@@ -700,7 +706,7 @@ jobs:
         if: ${{ !cancelled() }}
         with:
           name: junit-xml--${{github.run_id}}--${{steps.get_job_id.outputs.JOB_ID}}--${{github.run_attempt}}--${{matrix.name}}--functional-test-xdc
-          path: .testoutput
+          path: ./.testoutput
```
stephanos, line 709: Just for consistency.

.github/workflows/run-tests.yml
ychebotarev: may be question doesn't make sense. How it was working before? This is a path to a concrete file(s?). Before it was a directory.

.github/workflows/run-tests.yml
ychebotarev: can you go into a bit more details on implications of this? It will run for every test combo?

ychebotarev (approved):

.github/workflows/run-tests.yml
stephanos: Right; this was there before my change in https://github.com/temporalio/temporal/pull/7871. It will start up all containers define by the matrix (e.g. `postgresql`). _Unless_ we use the `run_single_functional_test` option, then it will only start the one requested there.

.github/workflows/run-tests.yml
stephanos: Yes, before it included everything inside the dictionary. Which included`.xml` _as well_ as `.out` (coverage files). But we don't want coverage files here. So I made it more specific.

PR #7942, opened 2025-06-20
Remove encodingType from Serializer interface

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/common/persistence/serialization/serializer.go
+++ b/common/persistence/serialization/serializer.go
@@ -22,65 +22,65 @@ type (
 	// Serializer is used by persistence to serialize/deserialize objects
 	// It will only be used inside persistence, so that serialize/deserialize is transparent for application
 	Serializer interface {
-		SerializeEvents(batch []*historypb.HistoryEvent, encodingType enumspb.EncodingType) (*commonpb.DataBlob, error)
+		SerializeEvents(batch []*historypb.HistoryEvent) (*commonpb.DataBlob, error)
 		DeserializeEvents(data *commonpb.DataBlob) ([]*historypb.HistoryEvent, error)
 
-		SerializeEvent(event *historypb.HistoryEvent, encodingType enumspb.EncodingType) (*commonpb.DataBlob, error)
+		SerializeEvent(event *historypb.HistoryEvent) (*commonpb.DataBlob, error)
 		DeserializeEvent(data *commonpb.DataBlob) (*historypb.HistoryEvent, error)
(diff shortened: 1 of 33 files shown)


common/persistence/serialization/serializer.go
stephanos, line 1: Only meaningful changes are in this file.

Proto3 encoding is hard-coded here now.

common/persistence/serialization/serializer.go
```diff
@@ -22,65 +22,65 @@ type (
 	// Serializer is used by persistence to serialize/deserialize objects
 	// It will only be used inside persistence, so that serialize/deserialize is transparent for application
 	Serializer interface {
```
dnr, line 24: Did we consider making this not an interface and just functions in a package? I can't figure out what it adds. If serialization was going to be controlled by static/dynamic config then yeah, but your proposal is an env var, right?

dnr (approved):

common/persistence/serialization/serializer.go
stephanos: Good question. I haven't considered that, but to your point, it looks like it doesn't provide much value. I see there's `ExecutionMutableStateTaskSuite` which uses the interface to customize the behavior.

Yes, I figured there would be a "codec" parameter that we pass into the Serializer that actually decides how to encode/decode the data. Since changing the codec would only be useful for development purposes, an env variable seemed reasonable to me (as opposed to a configuration field).

I don't think having two implementations of the Serializer would be useful; most methods just defer to the encoding helper function. Some of them massage the input a little beforehand.

dnr: If it needs to hold a reference to a codec or a static/dynamic config, then it has to be a struct or interface. Just a struct would be fine though, and less overhead than an interface. Maybe that's a good idea since we shouldn't trust that os.Getenv is fast

stephanos: This is what I was imagining: https://github.com/temporalio/temporal/pull/7860/files#diff-d5ebdc11437c7249b2a5903af65d1c7859983708bfa430aa2474e92ea0abd6ccR147-R156

I agree, the interface can be removed.

yycptt: We have customized logic for serializing/deserializing internal tasks.

yycptt (approved):

PR #7945, opened 2025-06-21
This PR changes log.With(logger, tags ...) to "upsert" the supplied Tags (based on their keys)

paulnpdev:
## What changed?
this PR changes the semantics of common/log/Logger to apply With(tags) as upserts for duplicate keys, rather than appending duplicates.  This matches the behavior of the TestLogger
## Why?
Having multiple values for the same Tag key in the logs is much more confusing than helpful.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
- It's possible someone is dependent on the existing behavior and may need to adjust.  However, this shouldn't affect functionality.
- Each Logger instance consumes a bit more memory


common/log/zap_logger.go
```diff
@@ -174,6 +186,24 @@ func (l *zapLogger) Skip(extraSkip int) Logger {
 	}
 }
 
+func mergeTags(oldTags, newTags []tag.Tag) (tags []tag.Tag) {
+	if oldTags == nil {
+		return newTags
+	}
+	tagMap := make(map[string]tag.Tag, len(oldTags)+len(newTags))
```
dnr: there's usually only like 1-10 tags and With calls add a small number.. maybe we should just do a linear search instead? using a map like this will shuffle the order, which isn't great since related tags are often passed together.

or else use the map differently: clone the old slice, make a map from key to index, and then lookup and replace or else append?

common/log/zap_logger_test.go
```diff
@@ -93,7 +94,12 @@ func TestDefaultLogger(t *testing.T) {
 	preCaller := caller(1)
 	logger.With(tag.Error(fmt.Errorf("test error"))).Info("test info", tag.WorkflowActionWorkflowStarted)
 
-	// back to normal state
+	// Test tags with duplicate keys are replaced
+	withLogger := With(logger, tag.NewStringTag("xray", "yankee"))
+	withLogger = With(withLogger, tag.NewStringTag("xray", "zulu"))
+	withLogger.Info("Log message with tag")
```
dnr, line 100: what happens if you pass duplicate tags on the Info call itself?

common/log/zap_logger.go
paulnpdev: I went with plan A

common/log/zap_logger_test.go
paulnpdev: the last instance wins (now)

paulnpdev:
ptal

common/log/zap_logger.go
```diff
@@ -172,21 +176,46 @@ func (l *zapLogger) Fatal(msg string, tags ...tag.Tag) {
...
 	return &zapLogger{
-		zl:   l.zl,
-		skip: l.skip + extraSkip,
+		zl:     l.zl,
+		skip:   l.skip + extraSkip,
+		baseZl: l.baseZl,
+	}
+}
+
+// mergeTags MAY update oldTags in place
```
dnr: Is that okay? Won't that affect the tags of the logger that With is called on? Well, not quite, because they're already turned into zap.Fields. But then won't it potentially affect the _next_ call to With?

common/log/zap_logger_test.go
dnr: I mean what happens if this line is
```suggestion
	withLogger.Info("Log message with tag", tag.NewStringTag("xray", "tango"))
```
?

It looks like that just turns the tag into a zap.Field and passes it to zap, which doesn't merge, so it would come out with duplicate tags.

It's starting to feel like this change should really be made in zap. Or maybe we should migrate to slog? But slog [doesn't handle duplicates by default either](https://www.github.com/golang/go/issues/59365)

common/log/zap_logger.go
paulnpdev: I think you are right, it's better to be safe than sorry.
I had in mind that most calls are `logger = with(logger, ...)` but that's certainly not always true.

common/log/zap_logger_test.go
paulnpdev: I actually think that in this case it's ok to leave the duplicate there.  The caller may not even realize that they are using a tag.Key that an earlier logger-provider had applied using With() ... perhaps with actually different semantics for the Value.

paulnpdev: p.s. there's no question that this change belongs in Zap imho

common/log/zap_logger.go
```diff
@@ -172,21 +177,48 @@ func (l *zapLogger) Fatal(msg string, tags ...tag.Tag) {
 	}
 }
 
+// With handles the provided tags as "upserts", replacing any matching keys with new values.
 func (l *zapLogger) With(tags ...tag.Tag) Logger {
+	cloneTags := mergeTags(l.tags, tags)
+	if l.baseZl == nil {
+		l.baseZl = l.zl
+	}
```
dnr, line 190: I don't think this can ever happen?

common/log/zap_logger_test.go
dnr: It feels weird that this removes duplicates sometimes but not always, and that these can now be different:
```
logger.Info("msg", ...)
logger.With(...).Info("msg")
```
but I guess it's still an improvement.

common/log/zap_logger.go
paulnpdev: I don't either.  It was defensive programming because I didn't want the logging code to segv even if someone else introduced a bug.  Too much?

paulnpdev:
Thanks for your great review.  I'd like to hear your response to my (last) PR comments before submitting.  And a brief look at my additional code comments if you don't mind.

common/log/zap_logger.go
dnr: I think I would leave it out myself (having it there makes me doubt my understanding and go searching for where baseZl is nil). But I'm okay either way

dnr:
> Thanks for your great review. I'd like to hear your response to my (last) PR comments before submitting. And a brief look at my additional code comments if you don't mind.

I agree it's good to call out the difference in the comment. Maybe some day it can be changed in zap, but this is fine until then.

PR #7946, opened 2025-06-21
Queries on drained + poller-less version to respond with descriptive error message

Shivs11:
## What changed?
- Queries on drained + poller-less version to respond with descriptive error message
- Also added `VersionStatus` as part of the userData. Detailed explanation as to why I did that is in the file where the changes have been made.

## Why?
- versioning

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
- Version workflow now needs patching since there is an extra sync call every-time a version's drainage status changes.


service/matching/task_queue_partition_manager.go
```diff
@@ -1000,6 +1006,21 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
 			return nil, nil, nil, err
 		}
 
+		// Preventing Query tasks from being dispatched to a drained version with no workers
+		if isQuery {
+			for _, versionData := range deploymentData.GetVersions() {
+				if versionData.GetVersion() != nil && worker_versioning.DeploymentVersionFromDeployment(deployment).Equal(versionData.GetVersion()) {
+					if versionData.GetStatus() == enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINED && len(pm.GetAllPollerInfo()) == 0 {
```
Shivs11, line 1013: Prior to storing status as part of the task-queue's user data, I was making a `DescribeVersion` call which would, theoretically, work. However, I think it could be load intensive on history if a user sends in a bunch of queries on versions that are drained + poller less.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -728,4 +699,107 @@ func (d *VersionWorkflowRunner) updateVersionStatusAfterDrainageStatusChange(new
...
 		d.VersionState.Status = enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_UNSPECIFIED
 	}
+
+	v := workflow.GetVersion(ctx, "Step1", workflow.DefaultVersion, 0)
+	if v != workflow.DefaultVersion {
+		err := d.syncVersionStatusToTaskQueues(ctx)
+		if err != nil {
+			d.logger.Error("failed to sync version status after drainage status change", "error", err)
+		}
+	}
```
Shivs11: patching is required here since every-time the status of a workflow changes from draining/drained, we alter the workflow history by adding in an extra event (did not do that previously)

service/worker/workerdeployment/version_workflow.go
```diff
@@ -728,4 +699,107 @@ func (d *VersionWorkflowRunner) updateVersionStatusAfterDrainageStatusChange(new
...
+		RampPercentage:    d.VersionState.RampPercentage,
+		Status:            d.VersionState.Status,
+	}
+
+	return d.syncVersionDataToTaskQueues(ctx, versionData)
+}
+
+// syncVersionDataToTaskQueues is a helper that syncs the provided version data to all task queues.
+// This function does NOT acquire the workflow lock - the caller is responsible for that.
+func (d *VersionWorkflowRunner) syncVersionDataToTaskQueues(ctx workflow.Context, versionData *deploymentspb.DeploymentVersionData) error {
```
Shivs11, line 742: just pulled this code out into a separate helper - not added anything to it

service/matching/physical_task_queue_manager.go
```diff
@@ -110,6 +110,7 @@ var (
 	errRemoteSyncMatchFailed     = serviceerror.NewCanceled("remote sync match failed")
 	errMissingNormalQueueName    = errors.New("missing normal queue name")
 	errDeploymentVersionNotReady = serviceerror.NewUnavailable("task queue is not ready to process polls from this deployment version, try again shortly")
+	ErrBlackHoledQuery           = "You are trying to query a closed Workflow that is PINNED to Worker Deployment Version %s, but %s is drained and has no pollers to answer the query. Immediately: You can re-deploy Workers in this Deployment Version to take those queries, or you can workflow update-options to change your workflow to AUTO_UPGRADE. For the future: In your infrastructure, consider waiting longer after the last queried timestamp as reported in Describe Deployment before you sunset Workers. Or mark this workflow as AUTO_UPGRADE."
```
carlydf: nit: "blackholed" when used as a verb, is one word https://en.wiktionary.org/wiki/blackholed
```suggestion
	ErrBlackholedQuery           = "You are trying to query a closed Workflow that is PINNED to Worker Deployment Version %s, but %s is drained and has no pollers to answer the query. Immediately: You can re-deploy Workers in this Deployment Version to take those queries, or you can workflow update-options to change your workflow to AUTO_UPGRADE. For the future: In your infrastructure, consider waiting longer after the last queried timestamp as reported in Describe Deployment before you sunset Workers. Or mark this workflow as AUTO_UPGRADE."
```

service/matching/task_queue_partition_manager.go
```diff
@@ -979,6 +984,7 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
 	forwardInfo *taskqueuespb.TaskForwardInfo,
 	runId string,
 	workflowId string,
+	isQuery bool,
```
carlydf, line 987: I think an enum for task type would be better (we probably have one), in case we would want to do something different for Activity, WF, Nexus, or Query task later on.

But also, it's not a breaking change to change this bool to an enum later, if a need ever arises, so if it's non-trivial to change this now, you should just keep it as is.

service/matching/task_queue_partition_manager.go
carlydf: nice!

service/worker/workerdeployment/version_workflow.go
```diff
@@ -511,69 +514,23 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
-				}).Get(ctx, nil)
-			if err != nil {
-				// TODO (Shivam): Compensation functions required to roll back the local state + activity changes.
-				return nil, err
-			}
-		}
+	// sync version information to all the task queues
+	err = d.syncVersionDataToTaskQueues(ctx, versionData)
+	if err != nil {
+		// TODO (Shivam): Compensation functions required to roll back the local state + activity changes.
```
carlydf, line 532: nice consolidation

service/worker/workerdeployment/version_workflow.go
```diff
@@ -591,7 +548,6 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
 		// Version deactivated from current/ramping
```
carlydf, line 544: nit: opportunity to replace `	if wasAcceptingNewWorkflows && !isAcceptingNewWorkflows` with `if newStatus == enumspb.VERSION_STATUS_DRAINING` I think!

service/worker/workerdeployment/version_workflow.go
```diff
@@ -591,7 +548,6 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
 		d.VersionState.LastDeactivationTime = args.RoutingUpdateTime
 		d.startDrainage(ctx, false)
-		state.Status = enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING
 	}
 
 	// started accepting new workflows --> stop drainage child wf if it exists
```
carlydf: same here, you could compare previous status to `newStatus` instead of `if !wasAcceptingNewWorkflows && isAcceptingNewWorkflows`

service/worker/workerdeployment/version_workflow.go
```diff
@@ -728,4 +699,107 @@ func (d *VersionWorkflowRunner) updateVersionStatusAfterDrainageStatusChange(new
...
+			d.logger.Error("failed to sync version status after drainage status change", "error", err)
+		}
+	}
+
+}
+
+// syncVersionStatusToTaskQueues syncs the current version status to all task queues.
+// This function acquires the workflow lock the same way as the update handlers do so that
+// there are no race conditions during task-queue sync. It is called internally when the status
+// of the version becomes draining/drained.
```
carlydf:
```suggestion
// of the version becomes draining or drained.
```
Do you also call it when the version becomes active after it was previously draining or drained?

tests/versioning_3_test.go
```diff
@@ -232,6 +233,90 @@ func (s *Versioning3Suite) TestQueryWithPinnedOverride_Sticky() {
...
 
+func (s *Versioning3Suite) TestPinnedQuery_DrainedVersion_PollersAbsent() {
+	s.RunTestWithMatchingBehavior(
+		func() {
+			s.testPinnedQuery_DrainedVersion(false)
+		},
+	)
+}
+
+func (s *Versioning3Suite) TestPinnedQuery_DrainedVersion_PollersPresent() {
```
carlydf, line 244: can you add a test for the scenario where a version becomes drained, and then becomes activated again (current or ramping), and then the query succeeds?

carlydf (commented):
requested additional test scenario!

service/worker/workerdeployment/version_workflow.go
Shivs11: not really but I can see where the confusion lies; this function was developed with the sole purpose of being called *after* the drainage go-routine changes the version status of a version to draining/drained. All other version changes (such as changing the version from draining -> current/ramping) flow nicely through the function `syncState`, which in turn calls the helper `syncVersionDataToTaskQueues`. 

With that being said, anyone can have the question you raised and I suspect this is because of the way I named my function. I changed it so that one can clearly understand why and when it's called.

service/worker/workerdeployment/version_workflow.go
Shivs11: done

tests/versioning_3_test.go
Shivs11: done!

service/matching/task_queue_partition_manager.go
Shivs11: i tried doing the above while I was drafting the PR but decided against it since a query task, for an example, is actually an internal task which gets created *after* matching understands what queue to dispatch the query task on. References - https://github.com/temporalio/temporal/blob/d127312234d1f56e97660aad71ea60e8cc6f0800/service/matching/physical_task_queue_manager.go#L398

Sure, you could argue we could just create the internal task earlier on and pass the task in this function, but for simplicity's sake as well as to move quicker, I have chosen to come back to this at a later time (if that's okay by you!)

service/worker/workerdeployment/version_workflow.go
Shivs11: thanks!

service/worker/workerdeployment/version_workflow.go
carlydf: ok, so just confirming, when a version becomes active, the task queue will change the status to active in it's VersionInfo also?

carlydf (approved):

service/worker/workerdeployment/version_workflow.go
carlydf: nevermind, the test you added confirms this, no need to explain!

PR #7948, opened 2025-06-23
Small matching cleanups

dnr:
## What changed?
- Consolidate task ids into `writeTaskRequest` instead of separate slice.
- Factor out some common code in `newPhysicalTaskQueueManager`.
- Factor out `taskResponse.err()`.
- Move code out of select in `getTasksPump`.
- Add a `UseCassandraPersistence` test helper.
- Remove some obsolete structs and functions.
- Rename some test functions.
- Update some comments.

## Why?
These are a bunch of small changes from the fairness feature branch that can apply to main immediately, which will make reviewing/merging that branch easier.

## How did you test it?
- [x] built

### DIFF
--- a/service/matching/backlog_manager_test.go
+++ b/service/matching/backlog_manager_test.go
@@ -28,7 +28,7 @@ type BacklogManagerTestSuite struct {
 	ptqMgr     *MockphysicalTaskQueueManager
 }
 
-func TestBacklogManager_Suite(t *testing.T) {
+func TestBacklogManager_Classic_Suite(t *testing.T) {
 	t.Parallel()
 	suite.Run(t, &BacklogManagerTestSuite{newMatcher: false})
 }
--- a/service/matching/backlog_manager_test.go
+++ b/service/matching/backlog_manager_test.go
@@ -28,12 +28,12 @@ type BacklogManagerTestSuite struct {
 	ptqMgr     *MockphysicalTaskQueueManager
 }
 
-func TestBacklogManagerTestSuite(t *testing.T) {
+func TestBacklogManager_Suite(t *testing.T) {
 	t.Parallel()
 	suite.Run(t, &BacklogManagerTestSuite{newMatcher: false})
 }
 
-func TestBacklogManagerWithNewMatcherTestSuite(t *testing.T) {
+func TestBacklogManager_Pri_Suite(t *testing.T) {
 	t.Parallel()
 	suite.Run(t, &BacklogManagerTestSuite{newMatcher: true})
 }
--- a/service/matching/db.go
+++ b/service/matching/db.go
@@ -322,7 +322,6 @@ func (db *taskQueueDB) getTotalApproximateBacklogCount() (total int64) {
 // CreateTasks creates a batch of given tasks for this task queue
 func (db *taskQueueDB) CreateTasks(
 	ctx context.Context,
-	taskIDs []int64,
 	reqs []*writeTaskRequest,
 ) (createTasksResponse, error) {
 	db.Lock()
@@ -337,7 +336,7 @@ func (db *taskQueueDB) CreateTasks(
 	allSubqueues := make([]int, len(reqs))
 	for i, req := range reqs {
 		task := &persistencespb.AllocatedTaskInfo{
-			TaskId: taskIDs[i],
+			TaskId: req.id,
 			Data:   req.taskInfo,
 		}
 		allTasks[i] = task
--- a/service/matching/matching_engine_test.go
+++ b/service/matching/matching_engine_test.go
@@ -132,7 +132,7 @@ func createMockNamespaceCache(controller *gomock.Controller, nsName namespace.Na
 }
 
 // TODO(pri): cleanup; delete this
-func TestMatchingEngine_Suite(t *testing.T) {
+func TestMatchingEngine_Classic_Suite(t *testing.T) {
 	suite.Run(t, &matchingEngineSuite{newMatcher: false})
 }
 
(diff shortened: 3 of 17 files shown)


service/matching/backlog_manager_test.go
```diff
@@ -28,12 +28,12 @@ type BacklogManagerTestSuite struct {
 	ptqMgr     *MockphysicalTaskQueueManager
 }
 
-func TestBacklogManagerTestSuite(t *testing.T) {
+func TestBacklogManager_Suite(t *testing.T) {
```
ychebotarev: just by param name - this one is "TestBacklogManager_OldMatcher"
(or, because all those "old" gets old very quickly - TestBacklogManager_NoPri_Suite?

service/matching/db.go
```diff
@@ -337,7 +336,7 @@ func (db *taskQueueDB) CreateTasks(
 	allSubqueues := make([]int, len(reqs))
 	for i, req := range reqs {
 		task := &persistencespb.AllocatedTaskInfo{
-			TaskId: taskIDs[i],
+			TaskId: req.id,
```
ychebotarev, line 339: why change this?
 I check some callers, and it looks like there is a semantic difference between taskIDs and req.id

service/matching/matching_engine_test.go
```diff
@@ -131,11 +132,11 @@ func createMockNamespaceCache(controller *gomock.Controller, nsName namespace.Na
 }
 
 // TODO(pri): cleanup; delete this
-func TestMatchingEngineSuite(t *testing.T) {
+func TestMatchingEngine_Suite(t *testing.T) {
```
ychebotarev: same, I feel this needs better name.

service/matching/matching_engine_test.go
```diff
@@ -3872,23 +3871,19 @@ func (m *testTaskManager) CreateTasks(
 
 	// Then insert all tasks if no errors
 	for _, task := range request.Tasks {
-		tlm.tasks.Put(task.GetTaskId(), &persistencespb.AllocatedTaskInfo{
-			Data:   task.Data,
-			TaskId: task.GetTaskId(),
-		})
+		tlm.tasks.Put(task.GetTaskId(), common.CloneProto(task))
```
ychebotarev, line 3874: as I understand the difference - it will actually clone TaskInfo. Are we changing it and its causing errors, or this is just to be safe?

service/matching/pri_task_reader.go
```diff
@@ -157,44 +157,43 @@ func (tr *priTaskReader) getTasksPump() {
...
 		case <-ctx.Done():
 			return
-
 		case <-tr.notifyC:
-			if tr.getLoadedTasks() > tr.backlogMgr.config.GetTasksReloadAt() {
-				// Too many loaded already, ignore this signal. We'll get another signal when
-				// loadedTasks drops low enough.
-				continue Loop
-			}
+		}
```
ychebotarev, line 165: I don't get this change. This practically reduce nestings, right? (and removes label, which is always good)

service/matching/task_writer.go
```diff
@@ -122,28 +123,24 @@ func (w *taskWriter) appendTask(
 	}
 }
 
-func (w *taskWriter) allocTaskIDs(count int) ([]int64, error) {
-	result := make([]int64, count)
-	for i := 0; i < count; i++ {
+func (w *taskWriter) allocTaskIDs(reqs []*writeTaskRequest) error {
```
ychebotarev: may be name is outdated. assignTaskIDs?

service/matching/db.go
ychebotarev: ah, I think I got it. They where coming from "allocTaskIDs".
And you also change "allocTaskIDs"

ychebotarev (commented):
mostly nits

service/matching/backlog_manager_test.go
dnr: I hesitate to say "Old" because we may still be stuck with it for quite a while, I'm not sure how fast we can realistically deprecate it. Maybe "Classic" is nicer :)

service/matching/db.go
dnr: yeah, taskIDs is gone. instead of a separate parallel slice I just had it put the id in writeTaskRequest

service/matching/matching_engine_test.go
dnr: it's just to simplify the code. I don't care about performance here, it's tests.

service/matching/matching_engine_test.go
```diff
@@ -3900,7 +3895,7 @@ func (m *testTaskManager) GetTasks(
 	var tasks []*persistencespb.AllocatedTaskInfo
 
 	it := tlm.tasks.Iterator()
-	for it.Next() {
+	for it.Next() && len(tasks) < request.PageSize {
```
dnr, line 3898: this was missing!

service/matching/pri_task_reader.go
dnr: reduce nesting and remove label, yes. otherwise equivalent

service/matching/task_writer.go
dnr: sure.

though I left allocTaskIDBlock since that does "allocate" in some sense

service/matching/task.go
```diff
@@ -80,6 +80,13 @@ type (
 	}
 )
 
+func (res taskResponse) err() error {
```
dnr, line 83: you might ask why there are still two error fields. I think maybe they can be combined but I want to leave it for later

PR #7949, opened 2025-06-23
UwS returns retryable error on workflow close

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,7 +59,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.49.2-0.20250605210325-e57aa581185c
+	go.temporal.io/api v1.50.1-0.20250623222711-4bd288b098bb
 	go.temporal.io/sdk v1.34.0
 	go.temporal.io/version v0.3.0
 	go.uber.org/automaxprocs v1.6.0
--- a/service/history/api/multioperation/api.go
+++ b/service/history/api/multioperation/api.go
@@ -121,18 +121,37 @@ func Invoke(
 	ns := namespaceEntry.Name().String()
 	// TODO(stephan): remove dynamic config again
 	if shardContext.GetConfig().EnableUpdateWithStartRetryOnClosedWorkflowAbort(ns) {
-		if !uws.workflowWasStarted && uws.updateWasAbortedByClosingWorkflow(err) {
+		if uws.noStartButUpdateWasAbortedByClosingWorkflow(err) {
+			testhooks.Call(uws.testHooks, testhooks.UpdateWithStartOnClosingWorkflowRetry)
+
 			uws2, err2 := newUpdateWithStart()
 			if err2 != nil {
 				return nil, err2
 			}
 			res, err = uws2.Invoke(ctx)
+
+			// If the Update-with-Start encountered the same error of a closing workflow again, it will convert
+			// the error to Aborted (which is a retryable error) to allow the client to retry the operation.
+			if shardContext.GetConfig().EnableUpdateWithStartRetryableErrorOnClosedWorkflowAbort(ns) {
+				if uws2.noStartButUpdateWasAbortedByClosingWorkflow(err) {
+					var multiOpsErr *serviceerror.MultiOperationExecution
+					if errors.As(err, &multiOpsErr) {
+						return nil, serviceerror.NewMultiOperationExecution(multiOpsErr.Error(), []error{
+							multiOpsErr.OperationErrors()[0],
+							serviceerror.NewAborted(multiOpsErr.OperationErrors()[1].Error()), // changed from NotFound to Aborted!
+						})
+					}
+				}
+			}
 		}
 	}
 	return res, err
 }
 
-func (uws *updateWithStart) updateWasAbortedByClosingWorkflow(err error) bool {
+func (uws *updateWithStart) noStartButUpdateWasAbortedByClosingWorkflow(err error) bool {
+	if uws.workflowWasStarted {
+		return false
+	}
 	var multiOpsErr *serviceerror.MultiOperationExecution
 	if ok := errors.As(err, &multiOpsErr); !ok {
 		return false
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -59,7 +59,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.49.2-0.20250605210325-e57aa581185c
+	go.temporal.io/api v1.50.1-0.20250623222711-4bd288b098bb
```
stephanos, line 62: This contains the new `serviceerror.NewAborted`.

service/history/api/multioperation/api.go
```diff
@@ -121,18 +121,37 @@ func Invoke(
...
 		}
 	}
 	return res, err
 }
 
-func (uws *updateWithStart) updateWasAbortedByClosingWorkflow(err error) bool {
+func (uws *updateWithStart) noStartButUpdateWasAbortedByClosingWorkflow(err error) bool {
+	if uws.workflowWasStarted {
+		return false
+	}
```
stephanos, line 162: Moved the check in here. (DRY)

tests/update_workflow_test.go
```diff
@@ -5657,6 +5657,11 @@ func (s *UpdateWorkflowSuite) TestUpdateWithStart() {
 			// wait until the update is admitted
 			s.waitUpdateAdmitted(tv)
 
+			s.InjectHook(testhooks.UpdateWithStartOnClosingWorkflowRetry, func() {
+				_, err := s.FrontendClient().StartWorkflowExecution(testcore.NewContext(), startWorkflowReq(tv))
+				s.NoError(err)
+			})
```
stephanos, line 5663: This is the only way AFAIK to test this new behavior.

tests/update_workflow_test.go
```diff
@@ -5675,12 +5680,14 @@ func (s *UpdateWorkflowSuite) TestUpdateWithStart() {
...
 
+			// ensure update-with-start returns retryable error
 			uwsRes := <-uwsCh
 			s.Error(uwsRes.err)
 			errs := uwsRes.err.(*serviceerror.MultiOperationExecution).OperationErrors()
 			s.Len(errs, 2)
-			s.Equal(nil, errs[0])
+			s.Equal("Operation was aborted.", errs[0].Error())
 			s.ErrorContains(errs[1], update.AbortedByWorkflowClosingErr.Error())
+			s.IsType(&serviceerror.Aborted{}, errs[1])
```
stephanos, line 5690: This is new!

service/history/api/multioperation/api.go
```diff
@@ -121,18 +121,37 @@ func Invoke(
 	ns := namespaceEntry.Name().String()
 	// TODO(stephan): remove dynamic config again
 	if shardContext.GetConfig().EnableUpdateWithStartRetryOnClosedWorkflowAbort(ns) {
-		if !uws.workflowWasStarted && uws.updateWasAbortedByClosingWorkflow(err) {
+		if uws.noStartButUpdateWasAbortedByClosingWorkflow(err) {
```
ychebotarev: highly recommend to reduce those nestings. It will be much easier to read. 
I opened in the separate window to just read it.
```
if !shardContext.GetConfig().EnableUpdateWithStartRetryOnClosedWorkflowAbort(ns) ||
 ! uws.noStartButUpdateWasAbortedByClosingWorkflow(err) {
   return res, err
}
```

service/history/api/multioperation/api.go
```diff
@@ -121,18 +121,37 @@ func Invoke(
...
+
 			uws2, err2 := newUpdateWithStart()
 			if err2 != nil {
 				return nil, err2
 			}
 			res, err = uws2.Invoke(ctx)
+
+			// If the Update-with-Start encountered the same error of a closing workflow again, it will convert
+			// the error to Aborted (which is a retryable error) to allow the client to retry the operation.
+			if shardContext.GetConfig().EnableUpdateWithStartRetryableErrorOnClosedWorkflowAbort(ns) {
```
ychebotarev: same here.

if !shardContext.GetConfig().EnableUpdateWithStartRetryableErrorOnClosedWorkflowAbort(ns) || 
! uws2.noStartButUpdateWasAbortedByClosingWorkflow(err) {
   return res, err
}

...

service/history/api/multioperation/api.go
stephanos: I thought you were on an ultra wide screen :D

service/history/api/multioperation/api.go
```diff
@@ -115,24 +116,50 @@ func Invoke(
...
-	// Since - compared to an Update - Update-with-Start has the ability to start a new workflow, the
-	// server will retry the Update-with-Start operation (but only once to keep latency and resource usage low).
-	ns := namespaceEntry.Name().String()
-	// TODO(stephan): remove dynamic config again
-	if shardContext.GetConfig().EnableUpdateWithStartRetryOnClosedWorkflowAbort(ns) {
-		if !uws.workflowWasStarted && uws.updateWasAbortedByClosingWorkflow(err) {
-			uws2, err2 := newUpdateWithStart()
-			if err2 != nil {
-				return nil, err2
+	if err != nil {
```
ychebotarev, line 119: well... :)

not blocking on this, but in general any extra nestings adds unnecessary cognitive load.
flattening (is this a term?) improves readability.
```
if err == nil {
   return res, err
}

...
```
(and you have one more below)

compare:
```

func foo() {
    result, err = tryA()
    if err == nil { return result, nil} 
    
    result, err = tryB()
    if err == nil { return result, nil} 

    result, err = tryB()
    if err == nil { return result, nil} 

    return nil, custom_error
}
```

versus

```
func foo() {
    result, err = tryA()
    if err != nil { 
          result, err = tryB()
          if err != nil { 
               result, err = tryC() 
               if err != nil {
                     return nil, custom_error
               }
          }
     } 
     return result, err
}
```


```

stephanos: I've considered this and what I don't like is the `if err == nil` part. It's surprising and easy to miss/mix up since 99.999% of all checks are `if err != nil`. I'd prefer a little bit of nesting over the surprise there.

stephanos: I also actually like that you can clearly see, from the code structure, that these are special error handling cases that are sequential. The structure tells a story. I say that as someone who also deeply believes in making code as non-nested as possible.

PR #7950, opened 2025-06-24
Matching fairness update persistence and protos

dnr:
## What changed?
- Add pass number fields to some protos.
- Implement pass handling in task store v2.
- Remove TTLs from task store v2.
- Add some error checking to prevent misuse.
- Add support for using LIMIT in the GetTasks query.
- Rename some things.

## Why?
Start implementing matching task fairness.

## How did you test it?
- [x] built

### DIFF
--- a/common/persistence/cassandra/matching_task_store.go
+++ b/common/persistence/cassandra/matching_task_store.go
@@ -34,17 +34,17 @@ func rowTypeTaskInSubqueue(subqueue int) int {
 }
 
 func getTaskTTL(expireTime *timestamppb.Timestamp) int64 {
-	var ttl int64
-	if expireTime != nil && !expireTime.AsTime().IsZero() {
-		expiryTtl := convert.Int64Ceil(time.Until(expireTime.AsTime()).Seconds())
-
-		// 0 means no ttl, we dont want that.
-		// Todo: Come back and correctly ignore expired in-memory tasks before persisting
-		if expiryTtl < 1 {
-			expiryTtl = 1
-		}
-
-		ttl = expiryTtl
+	if expireTime == nil || expireTime.AsTime().IsZero() {
+		return 0
 	}
-	return ttl
+
+	expiryTtl := convert.Int64Ceil(time.Until(expireTime.AsTime()).Seconds())
+
+	// 0 means no ttl, we dont want that.
+	// Todo: Come back and correctly ignore expired in-memory tasks before persisting
+	if expiryTtl < 1 {
+		expiryTtl = 1
+	}
+
+	return expiryTtl
 }
--- a/common/persistence/cassandra/matching_task_store.go
+++ b/common/persistence/cassandra/matching_task_store.go
@@ -1,33 +1,13 @@
-// The MIT License
-//
-// Copyright (c) 2020 Temporal Technologies Inc.  All rights reserved.
-//
-// Copyright (c) 2020 Uber Technologies, Inc.
-//
-// Permission is hereby granted, free of charge, to any person obtaining a copy
-// of this software and associated documentation files (the "Software"), to deal
-// in the Software without restriction, including without limitation the rights
-// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-// copies of the Software, and to permit persons to whom the Software is
-// furnished to do so, subject to the following conditions:
-//
-// The above copyright notice and this permission notice shall be included in
-// all copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-
 package cassandra
 
 import (
+	"time"
+
+	"go.temporal.io/server/common/convert"
 	"go.temporal.io/server/common/log"
 	p "go.temporal.io/server/common/persistence"
 	"go.temporal.io/server/common/persistence/nosql/nosqlplugin/cassandra/gocql"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 func NewMatchingTaskStore(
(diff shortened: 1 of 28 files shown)


proto/internal/temporal/server/api/taskqueue/v1/message.proto
```diff
@@ -41,11 +41,14 @@ message TaskVersionDirective {
 }
 
 message InternalTaskQueueStatus {
+    int64 read_level_pass = 7;
```
ychebotarev: I do remember from the talk why those 2 are close to each other, but for future readers I would add a comment

proto/internal/temporal/server/api/persistence/v1/tasks.proto
```diff
@@ -65,8 +66,16 @@ message SubqueueInfo {
     SubqueueKey key = 1;
 
     // The rest are mutable state for the subqueue:
+    int64 ack_level_pass = 4;
     int64 ack_level = 2;
     int64 approximate_backlog_count = 3;
+
+    // Max read level keeps track of the highest task level ever written, but is only
```
ychebotarev, line 74: those 2 comments are identical.
should this one be "Max read level pass  ..."

common/persistence/data_interfaces.go
```diff
@@ -601,10 +601,12 @@ type (
 		NamespaceID        string
 		TaskQueue          string
 		TaskType           enumspb.TaskQueueType
+		InclusiveMinPass   int64 // FairTaskManager only
 		InclusiveMinTaskID int64
 		ExclusiveMaxTaskID int64
```
ychebotarev, line 609: you don't need ExclusiveMaxPass here?

common/persistence/cassandra/matching_task_store_v2.go
```diff
@@ -379,16 +287,34 @@ func (d *matchingTaskStoreV2) GetTasks(
...
-		request.NamespaceID,
-		request.TaskQueue,
-		request.TaskType,
-		rowTypeTaskInSubqueue(request.Subqueue),
-		request.InclusiveMinTaskID,
-		request.ExclusiveMaxTaskID,
-	).WithContext(ctx)
-	iter := query.PageSize(request.PageSize).PageState(request.NextPageToken).Iter()
+	var query gocql.Query
+	if request.UseLimit {
```
ychebotarev, line 297: suggestion (less repetition):
```
args := []interface{}{
	request.NamespaceID,
	request.TaskQueue,
	request.TaskType,
	rowTypeTaskInSubqueue(request.Subqueue),
	request.InclusiveMinPass,
	request.InclusiveMinTaskID,
}

if request.UseLimit {
	args = append(args, request.PageSize)
	query = d.Session.Query(templateGetTasksQuery_v2_limit, args...)
} else {
	query = d.Session.Query(templateGetTasksQuery_v2, args...)
}

```

common/persistence/cassandra/matching_task_store.go
```diff
@@ -1,43 +1,50 @@
...
+// xxxxxx1x: reserved
+// 00000100: task in subqueue 1
+// nnnnnn00: task in subqueue n, etc.
+func rowTypeTaskInSubqueue(subqueue int) int {
+	return subqueue<<2 | rowTypeTask // nolint:staticcheck
+}
+
+func getTaskTTL(expireTime *timestamppb.Timestamp) int64 {
+	var ttl int64
+	if expireTime != nil && !expireTime.AsTime().IsZero() {
```
ychebotarev: reverse this:
```
if expireTime == nil || expireTime.AsTime().IsZero( { return 0 }
```

less nesting code

proto/internal/temporal/server/api/persistence/v1/tasks.proto
```diff
@@ -14,6 +14,7 @@ import "temporal/server/api/taskqueue/v1/message.proto";
 // task column
 message AllocatedTaskInfo {
     TaskInfo data = 1;
+    int64 pass_number = 3;
```
ychebotarev: one overall comment, and reason I want to keep it open.
Did you consider combining "pass" and "id" into the same structure? (level?)
There is many places where you have this:
```
somethingPass
somethingId
```

Keeping them separately makes it harder to read, easier to make a mistake, etc.
I understand it will not be possible everywhere (in proto to be specific, though I'm not sure - this is not external?)

common/persistence/cassandra/matching_task_store.go
dnr: this is just moving code back that landed in the wrong file, it's not new. but ok

common/persistence/cassandra/matching_task_store_v2.go
dnr: hmm, won't that cause an extra heap allocation? or two actually since the common case will always UseLimit?

common/persistence/data_interfaces.go
dnr: The semantics are different, GetTasks for the fair tasks table has no upper bound. I'll update the comments.

proto/internal/temporal/server/api/persistence/v1/tasks.proto
dnr: Yeah.. I did it this way since several of the "id" fields are already there and I wanted to reuse them instead of having a duplicate id field. I think with a few wrapper functions we can minimize the risk of mistakes. Also it's more efficient to inline it instead of another message with just two fields.

I'm like 80% sure I want to do it this way. Maybe let's reconsider after the whole first batch of PRs?

proto/internal/temporal/server/api/persistence/v1/tasks.proto
dnr: I want the comment to end up as a docstring on both fields (and pop up in ides, etc.) so it can't be ignored if you just look at MaxReadLevelId.

proto/internal/temporal/server/api/persistence/v1/tasks.proto
ychebotarev: I understand the concern with proto, especially if it is already stored somewhere and we need to care about backward compatibility. For those cases we can leave it as is. (or we may still add Level, dupblicate it, and eventually deprecate old field. But I agree this can be done later).

But with the new code, with proto messages that are not stored and we control(like InternalTaskQueueStatus and SubqueueInfo?) - it will benefit future us if its done now. I understand it is annoying to do, but If we delay it will be even worth :)

sometimes it looks inconsistent. Like:
```
 int64 ack_level_pass = 4;
 int64 ack_level = 2;
```
should it be `ack_level_id`? With common structure it will be
`Level ack_level = 2;` - clean, readable

Also in this specific structure - `pass_number` +'task_id' - should it be 'task_pass'? (so together they are "task level"). It is not clear that they are paired. With level it will be:
`Level task_level = 2;` or just 'Level level = 2'

Basically - this will be a great favour for future us, especially less experienced, if we do the right thing now.

common/persistence/cassandra/matching_task_store_v2.go
ychebotarev: It for sure will, and yes - 2 for UseLimit pass. Readability/perf trade off. I don't have opinion, up to you.

dnr: I'll leave it like this, there's a ton of repetition in persistence already, it doesn't seem so bad. And can be changed later easily.

proto/internal/temporal/server/api/persistence/v1/tasks.proto
dnr: I didn't want to rename the existing fields because of compatibility, but actually that only matters for proto-json which we're not using internally, so I'll rename them.

dnr: On the idea of a submessage: messages with the new fields will be persisted in a different table, so it's not about compatibility per se. Though I would like to be able to use the same messages for both tables.

For AllocatedTaskInfo: we have a lot of those in memory and in db so I want to pay attention to efficiency there. I think it makes sense to just inline the two fields.

For SubqueueInfo: this is persisted, but not that many of them, so we could use a new field.

For InternalTaskQueueStatus: yes, could use a new field.

I really hate having to do an extra allocation and chase another pointer just for these values. If we were using gogo it could be non-nullable and embedded, or we could just write custom getters and setters to avoid misuse :(

Let me try it and see how it looks.

dnr: I tried it. Mostly it's about the same readability since we still need wrappers to convert to/from structs. Some parts are slightly clearer. Overall I'd say slight improvement for readability. Adapting the existing code took `6 files changed, 74 insertions(+), 73 deletions(-)`  But it reduces the diff from main slightly.

Overall I'm unhappy about the tradeoff but I'll go with it.

common/persistence/task_manager.go
```diff
@@ -176,7 +176,8 @@ func (m *taskManagerImpl) CreateTasks(
 			return nil, serviceerror.NewUnavailablef("CreateTasks operation failed during serialization. Error : %v", err)
 		}
 		tasks[i] = &InternalCreateTask{
-			TaskId:     task.GetTaskId(),
+			TaskId:     task.TaskId,
```
ychebotarev, line 180: is it possible to also combine this to Level?

dnr: InternalCreateTask isn't a proto so I'm not sure what we could do here. I don't want to use FairLevel in AllocatedTaskInfo for efficiency, and I don't want to use the fairLevel struct from the matching package since that's private. Now that I look here I should rename Pass to TaskPass for consistency. I'll do that.

PR #7951, opened 2025-06-24
[CHASM] Outbound SideEffect task executor

lina-temporal:
## What changed?
- Adds CHASM task executors for the outbound queue.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/service/history/outbound_queue_active_task_executor.go
+++ b/service/history/outbound_queue_active_task_executor.go
@@ -2,18 +2,22 @@ package history
 
 import (
 	"context"
+	"fmt"
 
 	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/service/history/consts"
 	historyi "go.temporal.io/server/service/history/interfaces"
 	"go.temporal.io/server/service/history/queues"
+	"go.temporal.io/server/service/history/tasks"
 	wcache "go.temporal.io/server/service/history/workflow/cache"
 )
 
 type outboundQueueActiveTaskExecutor struct {
 	stateMachineEnvironment
+	chasmEngine chasm.Engine
 }
 
 var _ queues.Executor = &outboundQueueActiveTaskExecutor{}
@@ -23,6 +27,7 @@ func newOutboundQueueActiveTaskExecutor(
 	workflowCache wcache.Cache,
 	logger log.Logger,
 	metricsHandler metrics.Handler,
+	chasmEngine chasm.Engine,
 ) *outboundQueueActiveTaskExecutor {
 	return &outboundQueueActiveTaskExecutor{
 		stateMachineEnvironment: stateMachineEnvironment{
@@ -33,6 +38,7 @@ func newOutboundQueueActiveTaskExecutor(
 				metrics.OperationTag(metrics.OperationOutboundQueueProcessorScope),
 			),
 		},
+		chasmEngine: chasmEngine,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


service/history/outbound_queue_active_task_executor.go
```diff
@@ -79,7 +80,59 @@ func (e *outboundQueueActiveTaskExecutor) Execute(
 		return respond(err)
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, taskTimeout)
+	defer cancel()
```
yycptt: hmmm, no... we can't use the taskTimeout here especially for current hsm framework outbound tasks. It's too short. I think today each hsm outbound task execution logic itself set a timeout (10sec if I remember correctly).

I'd say let's not touch hsm tasks and also set 10s timeout for chasm tasks.

service/history/queues/metrics.go
```diff
@@ -151,11 +151,14 @@ func GetOutboundTaskTypeTagValue(task tasks.Task, isActive bool) string {
...
 	}
 
-	outbound, ok := task.(*tasks.StateMachineOutboundTask)
-	if !ok {
+	switch task := task.(type) {
+	case *tasks.StateMachineOutboundTask:
+		return prefix + "." + task.StateMachineTaskType()
+	case *tasks.ChasmTask:
+		return prefix + "." + task.Info.Type
+	default:
```
yycptt, line 171: Nice! Can we do that for side effect tasks in other queues as well?

service/history/tests/vars.go
```diff
@@ -163,5 +164,6 @@ func NewDynamicConfig() *configs.Config {
 	config.ReplicationEnableUpdateWithNewTaskMerge = dynamicconfig.GetBoolPropertyFn(true)
 	config.EnableWorkflowIdReuseStartTimeValidation = dynamicconfig.GetBoolPropertyFnFilteredByNamespace(true)
 	config.EnableTransitionHistory = dynamicconfig.GetBoolPropertyFn(true)
+	config.OutboundStandbyTaskMissingEventsDiscardDelay = dynamicconfig.GetDurationPropertyFnFilteredByDestination(time.Duration(math.MaxInt64))
```
yycptt: can you give a bit more context on this?

lina-temporal: Yeah, initially I'd had that logic carried over for CHASM tasks, but later moved it back to HSM-specific. I'll remove this.

service/history/queues/metrics.go
lina-temporal: Yup, will do!

service/history/outbound_queue_active_task_executor.go
lina-temporal: Sounds good, I'll make it specific to CHASM tasks.

yycptt: Forgot to remove this here?

service/history/outbound_queue_active_task_executor.go
```diff
@@ -79,7 +80,62 @@ func (e *outboundQueueActiveTaskExecutor) Execute(
...
+	}
+
+	return respond(queues.NewUnprocessableTaskError(fmt.Sprintf("unknown task type '%T'", task)))
+}
+
+func (e *outboundQueueActiveTaskExecutor) executeChasmSideEffectTask(
+	ctx context.Context,
+	task *tasks.ChasmTask,
+) error {
+	ctx, cancel := context.WithTimeout(ctx, taskTimeout)
```
yycptt, line 97: we probably need to use a higher timeout, nexus will eventually use chasm too. Maybe let's do 10s here? Outbound queue has a different worker pool implementation so higher timeout should be fine.

We can change that later as well. no strong opinion.

service/history/outbound_queue_active_task_executor.go
lina-temporal: Fixed, yeah, forgot to stage that

service/history/outbound_queue_active_task_executor.go
lina-temporal: Sure, updated to 10s.

PR #7953, opened 2025-06-24
Matching fairness testTaskManager implementation

dnr:
## What changed?
- Add fairLevel.
- Implement fairness persistence in testTaskManager.
- Partial overlap with https://github.com/temporalio/temporal/pull/7950, that could get merged first.

## Why?
Start implementing matching fairness.

## How did you test it?
- [x] built
- [x] added new unit test(s)


service/matching/ack_manager_test.go
```diff
@@ -225,7 +225,7 @@ func BenchmarkAckManager_CompleteTask(b *testing.B) {
 }
 
 func newTestAckMgr(logger log.Logger) *ackManager {
-	tm := newTestTaskManager(logger)
+	tm := newTestTaskManager(logger, false)
```
ychebotarev: minor. I would rather add `newTestFairTaskManager`.
this will remove that pesky `false`, and your intention will be clear.

proto/internal/temporal/server/api/taskqueue/v1/message.proto
```diff
@@ -41,11 +41,14 @@ message TaskVersionDirective {
 }
 
 message InternalTaskQueueStatus {
+    int64 read_level_pass = 7;
     int64 read_level = 1;
+    int64 ack_level_pass = 8;
```
ychebotarev: guess this needs to be updated once you merge previous PR.

ychebotarev (approved):

proto/internal/temporal/server/api/taskqueue/v1/message.proto
dnr: yup, merging now

service/matching/ack_manager_test.go
dnr: good idea

dnr (commented):
sorry for the rebase, it was the easiest way to merge after the previous one.

I also changed the fairLevel functions to use receivers where possible, seems nicer to read and write.

service/matching/fair_level.go
```diff
@@ -0,0 +1,68 @@
...
+	return fmt.Sprintf("<%d,%d>", a.pass, a.id)
+}
+
+// Returns true if a < b lexicographically.
+func (a fairLevel) less(b fairLevel) bool {
+	return a.pass < b.pass || a.pass == b.pass && a.id < b.id
+}
+
+// Use with treemap
+func fairLevelComparator(aany, bany any) int {
```
ychebotarev: as i can see that treemap in imports (https://github.com/emirpasic/gods/blob/master/maps/treemap/treemap.go)
use generics. Why "any"?

service/matching/fair_level.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	return a
+}
+
+// Returns the next highest fair level.
+func (a fairLevel) inc() fairLevel {
+	return fairLevel{pass: a.pass, id: a.id + 1}
+}
+
+func allocatedTaskFairLevel(t *persistencespb.AllocatedTaskInfo) fairLevel {
```
ychebotarev: nit: `fromProtoAllocatedTask` and `fromProtoFairLevel` a bit more consistant.

service/matching/fair_level.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+	taskqueuespb "go.temporal.io/server/api/taskqueue/v1"
+)
+
+// fairLevel is a pair of "pass" and "id". Pass is a number that can be assigned to cause
+// levels to be ordered to preserve fairness. Id is a unique task id. fairLevels are ordered
+// lexicographically.
+type fairLevel struct {
```
ychebotarev, line 14: do we ever plan to split matching to sub-packages?

ychebotarev (approved):
Please check if it is possible to bump that treemap version and use generics instead of "any". 

Alternatively we can have proper function in fair_level.go, and in the code where you create treemap (if it can't be updated and "any" is required) we can have something like:

```
treemap.NewWith(func(a, b any) { return fairLevelComparator(a.(fairLevel), b..(fairLevel)))
```

service/matching/fair_level.go
dnr: I think we can split out pieces where appropriate. I have no intention of splitting the core right now

service/matching/fair_level.go
dnr: We're using the latest released version, which is 3 years old. The generics change was 2 years ago. We could consider switching to using the main branch instead of a release. Another PR though

service/matching/fair_level.go
dnr: These names are package-scope. `fromProtoAllocatedTask` only tells you what it's coming from, not what it's going to, so that's not good enough. `fairLevelFromAllocatedTask` is borderline too verbose for me, but if you feel strongly I can change to that

PR #7955, opened 2025-06-24
Export OTEL traces per test case

stephanos:
## What changed?

Exporting an OTEL trace file _per test case_ instead of _per test suite_.

I also added a new `MemoryExporter` to make the OTEL collection more efficient and accurate.
Compared to the previous approach, it acts completely in-process/in-memory; not via gRPC.

## Why?

Removes successful test cases from the file output which makes the output smaller and easier to use.

Also removes race condition.

Also more efficient.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

Example:

Screenshot 2025-06-24 at 2 25 17 PM


.github/workflows/run-tests.yml
```diff
@@ -517,10 +517,11 @@ jobs:
       - name: Run functional test
         if: ${{ inputs.run_single_functional_test != true || (inputs.run_single_functional_test == true && contains(fromJSON(needs.set-up-single-test.outputs.dbs), env.PERSISTENCE_DRIVER)) }}
         timeout-minutes: ${{ fromJSON(needs.set-up-single-test.outputs.github_timeout) }} # make sure this is larger than the test timeout in the Makefile
-        run: make OTEL=true functional-test-coverage
```
stephanos, line 520: `OTEL=true` isn't needed anymore; as the OTEL exporter is now hooked up programatically.

common/testing/testtelemetry/collector.go
```diff
@@ -0,0 +1,72 @@
+package testtelemetry
```
stephanos, line 1: Nothing changed here except I renamed the previous `OTELLogger` to `MemoryCollector`; and removed the `WriteFile` part. That lives in `MemoryExporter` now.

tests/testcore/functional_test_base.go
```diff
@@ -412,6 +418,7 @@ func (s *FunctionalTestBase) TearDownCluster() {
 
 // **IMPORTANT**: When overridding this, make sure to invoke `s.FunctionalTestBase.TearDownTest()`.
 func (s *FunctionalTestBase) TearDownTest() {
+	s.exportOTELTraces()
```
stephanos, line 421: This now happens _per test_.

.github/workflows/run-tests.yml
```diff
@@ -517,10 +517,11 @@ jobs:
...
+        run: make functional-test-coverage
         env:
           TEST_ARGS: ${{ needs.set-up-single-test.outputs.single_test_args }}
-          TEST_TEMPORAL_OTEL_OUTPUT: ${{ github.workspace }}/.testoutput
+          TEMPORAL_TEST_OTEL_OUTPUT: ${{ github.workspace }}/.testoutput
```
stephanos, line 523: Renamed to align with the existing env variable `TEMPORAL_TEST_LOG_LEVEL`.

PR #7957, opened 2025-06-24
fix: increase number of attempts if paused activity fails

spkane31:
## What changed?

Increment the attempts for an activity when a paused activity fails

## Why?
Previously this scenario was not incremented

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
None aware of, @ychebotarev would know more potential risks

### DIFF
--- a/tests/activity_api_pause_test.go
+++ b/tests/activity_api_pause_test.go
@@ -180,6 +180,91 @@ func (s *ActivityApiPauseClientTestSuite) TestActivityPauseApi_WhileRunning() {
 	s.NoError(err)
 }
 
+func (s *ActivityApiPauseClientTestSuite) TestActivityPauseApi_IncreaseAttemptsOnFailure() {
+	/*
+	 * 1. Run an activity that runs forever
+	 * 2. Pause the activity
+	 * 3. Send a failure signal to the activity
+	 * 4. Validate activity failed
+	 * 5. Validate number of activity attempts increased
+	 */
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	var startedActivityCount atomic.Int32
+	activityPausedCn := make(chan struct{})
+	activityErr := errors.New("activity-failed-while-paused")
+
+	activityFunction := func() (string, error) {
+		startedActivityCount.Add(1)
+		s.WaitForChannel(ctx, activityPausedCn)
+		return "done!", activityErr
+	}
+
+	workflowFn := s.makeWorkflowFunc(activityFunction)
+
+	s.Worker().RegisterWorkflow(workflowFn)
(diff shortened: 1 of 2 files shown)


gow (approved):
Looks good.

tests/activity_api_pause_test.go
```diff
@@ -180,6 +181,91 @@ func (s *ActivityApiPauseClientTestSuite) TestActivityPauseApi_WhileRunning() {
...
+		require.Equal(t, 1, len(description.PendingActivities))
+		require.True(t, description.PendingActivities[0].Paused)
+		require.Equal(t, int32(2), description.PendingActivities[0].Attempt)
+		require.NotNil(t, description.PendingActivities[0].LastFailure)
+		require.NotNil(t, description.PendingActivities[0].PauseInfo)
+		require.NotNil(t, description.PendingActivities[0].PauseInfo.GetManual())
+		require.Equal(t, testIdentity, description.PendingActivities[0].PauseInfo.GetManual().Identity)
+		require.Equal(t, testReason, description.PendingActivities[0].PauseInfo.GetManual().Reason)
+	}, 5*time.Second, 500*time.Millisecond)
+}
```
ychebotarev, line 301: one more thing - you should let workflow to finish gracefully.
I found that otherwise you may get weird flakyness.

For this test this means:
1. Add a "fail/succeed" flag.
2. After this check (now you sure that number of attempts is 2) set that flag.
3. Unpause an activity
    now on the next run activity will succeed.
4. Let workflow finish, check that it finished successfully. 

All of this exist in other tests (this file, of may be some other).

tests/activity_api_pause_test.go
```diff
@@ -180,6 +181,91 @@ func (s *ActivityApiPauseClientTestSuite) TestActivityPauseApi_WhileRunning() {
...
+		require.NoError(t, err)
+		require.NotNil(t, description)
+		require.Equal(t, 1, len(description.PendingActivities))
+		require.True(t, description.PendingActivities[0].Paused)
+		require.Equal(t, int32(2), description.PendingActivities[0].Attempt)
+		require.NotNil(t, description.PendingActivities[0].LastFailure)
+		require.NotNil(t, description.PendingActivities[0].PauseInfo)
+		require.NotNil(t, description.PendingActivities[0].PauseInfo.GetManual())
+		require.Equal(t, testIdentity, description.PendingActivities[0].PauseInfo.GetManual().Identity)
+		require.Equal(t, testReason, description.PendingActivities[0].PauseInfo.GetManual().Reason)
```
ychebotarev, line 272: add 
`require.Equal(t, int32(1), startedActivityCount.Load())`
to this check.
This is to be sure activity was not called.
Stricktly speaking not required, but why not.

PR #7959, opened 2025-06-24
Add Stats to DescribeWorkerDeploymentVersion

stephanos:
## What changed?

Added task queue stats reporting to DescribeWorkerDeploymentVersion API.

## Why?

To allow inspecting backlog etc of task queues of a particular worker version.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -29,6 +29,7 @@ import (
 	tokenspb "go.temporal.io/server/api/token/v1"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/backoff"
+	"go.temporal.io/server/common/cache"
 	"go.temporal.io/server/common/clock"
 	hlc "go.temporal.io/server/common/clock/hybrid_logical_clock"
 	"go.temporal.io/server/common/cluster"
@@ -134,6 +135,7 @@ type (
 		gaugeMetrics                  gaugeMetrics // per-namespace task queue counters
 		config                        *Config
 		versionChecker                headers.VersionChecker
+		cache                         cache.Cache
 		testHooks                     testhooks.TestHooks
 		// queryResults maps query TaskID (which is a UUID generated in QueryWorkflow() call) to a channel
 		// that QueryWorkflow() will block on. The channel is unblocked either by worker sending response through
@@ -1303,19 +1305,26 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		} else {
 			taskQueueStats := &taskqueuepb.TaskQueueStats{}
 
-			// find all buildIds
+			// determine versions to query
 			var buildIds []string
-			userData, _, err := pm.GetUserDataManager().GetUserData()
-			if err != nil {
-				return nil, err
+			if request.Version != nil {
+				// A particular version was requested.
+				buildIds = []string{worker_versioning.WorkerDeploymentVersionToStringV31(request.Version)}
 			}
-			typedUserData := userData.GetData().GetPerType()[int32(pm.Partition().TaskType())]
-			for _, v := range typedUserData.GetDeploymentData().GetVersions() {
-				if v.GetVersion() == nil || v.GetVersion().GetDeploymentName() == "" || v.GetVersion().GetBuildId() == "" {
-					continue
+			if len(buildIds) == 0 {
+				// No version was requested, so we need to query all versions.
+				userData, _, err := pm.GetUserDataManager().GetUserData()
+				if err != nil {
+					return nil, err
 				}
-				buildId := worker_versioning.WorkerDeploymentVersionToStringV32(v.GetVersion())
-				buildIds = append(buildIds, buildId)
+				typedUserData := userData.GetData().GetPerType()[int32(pm.Partition().TaskType())]
+				for _, v := range typedUserData.GetDeploymentData().GetVersions() {
+					if v.GetVersion() == nil || v.GetVersion().GetDeploymentName() == "" || v.GetVersion().GetBuildId() == "" {
+						continue
+					}
+					buildId := worker_versioning.WorkerDeploymentVersionToStringV32(v.GetVersion())
+					buildIds = append(buildIds, buildId)
+				}
 			}
 
 			// query each partition for stats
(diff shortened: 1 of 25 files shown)


service/matching/matching_engine.go
```diff
@@ -1354,6 +1363,55 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
+
+	cacheKey := worker_versioning.WorkerDeploymentVersionToStringV31(request.Version)
+	if cached := pm.GetCache(cacheKey); cached != nil {
+		return cached.(*matchingservice.DescribeVersionedTaskQueuesResponse), nil
+	}
+
+	resp := &matchingservice.DescribeVersionedTaskQueuesResponse{}
+	for _, tq := range request.VersionTaskQueues {
+		tqResp, err := e.DescribeTaskQueue(ctx,
+			&matchingservice.DescribeTaskQueueRequest{
```
stephanos: I might make these concurrently. Didn't get there yet.

service/matching/matching_engine.go
```diff
@@ -1354,6 +1363,55 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
+		}
+		resp.VersionTaskQueues = append(resp.VersionTaskQueues,
+			&matchingservice.DescribeVersionedTaskQueuesResponse_VersionTaskQueue{
+				Name:  tq.Name,
+				Type:  tq.Type,
+				Stats: tqResp.DescResponse.Stats,
+			})
+	}
+
+	pm.PutCache(cacheKey, resp)
```
stephanos, line 1426: I had considered moving the cache into the `matchingEngineImpl` as part of this PR (it's outstanding tech debt), but that was difficult since it needs to be initialised with the TTL. But the TTL is a namespace-scoped config. So the lifetimes don't match up with the engine.

service/matching/matching_engine.go
```diff
@@ -1354,6 +1363,56 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
+
+	resp := &matchingservice.DescribeVersionedTaskQueuesResponse{}
+	for _, tq := range request.VersionTaskQueues {
+		tqResp, err := e.DescribeTaskQueue(ctx,
+			&matchingservice.DescribeTaskQueueRequest{
+				NamespaceId: request.GetNamespaceId(),
+				DescRequest: &workflowservice.DescribeTaskQueueRequest{
+					TaskQueue: &taskqueuepb.TaskQueue{
+						Name: tq.Name,
+						Kind: enumspb.TaskQueueKind(tq.Type),
```
ShahabT: this seems wrong. In this case it's always NORMAL kind

service/worker/workerdeployment/client.go
```diff
@@ -1350,21 +1360,20 @@ func (d *ClientImpl) record(operation string, retErr *error, args ...any) func()
...
-	for taskQueueName, taskQueueFamilyInfo := range state.TaskQueueFamilies {
-		for taskQueueType := range taskQueueFamilyInfo.TaskQueues {
-			element := &deploymentpb.WorkerDeploymentVersionInfo_VersionTaskQueueInfo{
-				Name: taskQueueName,
-				Type: enumspb.TaskQueueType(taskQueueType),
-				// TODO (Shivam): Add fields here as needed.
-			}
-			taskQueues = append(taskQueues, element)
-		}
+	infos := make([]*deploymentpb.WorkerDeploymentVersionInfo_VersionTaskQueueInfo, 0, len(taskQueueInfos))
```
ShahabT, line 1371: this part will be removed once the deprecated filed is cleaned up, right?

service/matching/matching_engine.go
stephanos: You're right; good catch. I probably auto-completed that. Now I wonder why the tests are passing ...

service/worker/workerdeployment/client.go
```diff
@@ -1382,12 +1391,88 @@ func versionStateToVersionInfo(state *deploymentspb.VersionLocalState) *deployme
...
+		for _, tq := range tqResp.GetVersionTaskQueues() {
+			tqRespMap[tqKey(tq.Name, tq.Type)] = tq
+		}
+
+		// Update stats for existing entries
+		for i, tq := range tqOutputs {
+			if tqRespTQ, ok := tqRespMap[tqKey(tq.Name, tq.Type)]; ok {
+				tqOutputs[i].Stats = tqRespTQ.Stats
+			} else {
+				// Setting empty stats instead of leaving nil (which is only used when not querying for stats).
```
ShahabT: Not sure about this. empty stats will mislead users into thinking the stats are all zero which is a valid stats by itself. It's dangerous if users think backlog is empty while it is actually not.

Options are reporting partial result with some nils in the list, or failing the whole call.

service/matching/matching_engine.go
```diff
@@ -1354,6 +1363,56 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 	return descrResp, nil
 }
 
+func (e *matchingEngineImpl) DescribeVersionedTaskQueues(
```
ShahabT, line 1370: So this API is for saving the tq fan-out by caching the version results, while `DescribeTaskQueue` is caching the per-tq results already, right?

It makes sense, I just wonder if we had to do it now or we could've waited for some more signals that the tq fan-out will be a practical problem. But it's good that we have it, now that we have it.

service/worker/workerdeployment/client.go
stephanos: I was thinking about this; it's not clear to me yet why/when we wouldn't have a result there.

service/matching/matching_engine.go
stephanos: Exactly! Since we have the cache, it was easy to add 🤷

service/worker/workerdeployment/client.go
stephanos: Exactly!

service/worker/workerdeployment/client.go
ShahabT: Yeah, other than random edge cases I cannot think of one either... but in that case, why not just return error for the whole thing? Better to be safe than returning incorrect results.

stephanos: 👍 okidoki

service/matching/matching_engine.go
```diff
@@ -1296,26 +1299,34 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
 		}
-		cacheKey := ".taskQueueStats" // starts with `.` to prevent clashing with build ID cache keys
+
+		var buildIds []string
+		if request.Version != nil {
+			// A particular version was requested. This is only available internally; not user-facing.
+			buildIds = []string{worker_versioning.WorkerDeploymentVersionToStringV31(request.Version)}
+		}
+
+		cacheKey := "dtq_default:" + strings.Join(buildIds, ",")
```
stephanos, line 1311: This needs to be keyed by the requested version or the cache results of all vs one version collide.

service/matching/matching_engine.go
```diff
@@ -1354,6 +1365,66 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
+		if localPM != nil {
+			// If available, query the local partition manager to save a network call.
+			tqResp, err = e.DescribeTaskQueue(ctx, tqReq)
+		} else {
+			// Otherwise, query the other matching service instance.
+			tqResp, err = e.matchingRawClient.DescribeTaskQueue(ctx, tqReq)
+		}
+		if err != nil {
+			return nil, err // some other error, return it
+		}
```
stephanos, line 1416: I realized that our functional test setup doesn't cover this; but in production it would need to actually use the matching client for task queues that aren't on the instance.

Note that `getTaskQueuePartitionManager` AFAIK can also return `nil` when the partition has been _unloaded_ but should be on this instance. I don't know how to check for that; so right now it would make a network call despite that.

service/matching/matching_engine.go
ShahabT: Not for this PR, but could we always put the per-version stats in the cache, keyed by version string so they do not collide? It'd mean the total result here would need to check cache for all individual versions but that should be fine, right?

service/matching/matching_engine.go
ShahabT: makes sense, it seems good to me.

For the future, I wonder we can improve the routing logic inside the client to use the local matching_engine if the target instance ends up being the same.

ShahabT (approved):

service/matching/matching_engine.go
stephanos: 👍 Good idea.

service/matching/matching_engine.go
```diff
@@ -134,6 +136,7 @@ type (
 		gaugeMetrics                  gaugeMetrics // per-namespace task queue counters
 		config                        *Config
 		versionChecker                headers.VersionChecker
+		cache                         cache.Cache
```
dnr, line 139: unused?

service/matching/matching_engine.go
dnr: Hmm, this best-effort thing is a little odd. My thought with the DescribeTaskQueue fanout was: the DescribeTaskQueue call goes to the root. So the root partition should be in the same process. So when asking the root, look up the pm, otherwise do an rpc. The "is it loaded locally" check can theoretically be wrong in both directions, as you point out.

In this case if we're doing a fanout to multiple task queues I think we should just do rpcs.

dnr: Hah, that's what https://github.com/temporalio/temporal/pull/6733 does. But that's way overkill for this.

stephanos: Oh, I didn't realize it can go wrong in _both_ directions? I only assumed that it could falsely state that the tq root isn't on this host (ie unloaded) and then make an avoidable RPC.

I'm trying to assess the impact here and whether I need to patch it for the release asap. Are you saying it's incorrect or unnecessary/overly complex?

service/matching/matching_engine.go
stephanos: Ah, good catch! That's from when I realized this won't work due to the caching TTL being ns scoped.

fix in https://github.com/temporalio/temporal/pull/7975

service/matching/matching_engine.go
dnr: Well, I suppose it could only be wrong in the positive direction for a few seconds, since we have the unload on membership change monitor. The cache ttl is a few seconds so... I would say not a correctness concern.

I just get worried about stuff like this because of the potential inconsistencies: the source of truth for partition owner is membership, not the map in matching engine. The map in matching engine eventually follows membership, hopefully within a few seconds. So I'd rather just go to the source (by doing an rpc that gets routed by membership) than ask the map directly.

In the DescribeTaskQueue case, the rpc to the root _just_ arrived, so that seems safe.

stephanos: 👍 I'll simplify it to always making the rpc.

PR #7961, opened 2025-06-24
Fix data race in Nexus config provider

pdoerner:
## What changed?
Adjusting dynamic config types to address a data race in Nexus config provider that occurred during xdc tests. Ran several times locally to verify fix.

### DIFF
--- a/components/nexusoperations/config.go
+++ b/components/nexusoperations/config.go
@@ -149,7 +149,7 @@ type Config struct {
 	MaxOperationNameLength              dynamicconfig.IntPropertyFnWithNamespaceFilter
 	MaxOperationTokenLength             dynamicconfig.IntPropertyFnWithNamespaceFilter
 	MaxOperationHeaderSize              dynamicconfig.IntPropertyFnWithNamespaceFilter
-	DisallowedOperationHeaders          dynamicconfig.TypedPropertyFn[[]string]
+	DisallowedOperationHeaders          *dynamicconfig.GlobalCachedTypedValue[[]string]
 	MaxOperationScheduleToCloseTimeout  dynamicconfig.DurationPropertyFnWithNamespaceFilter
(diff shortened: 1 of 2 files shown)


components/nexusoperations/workflow/commands_test.go
```diff
@@ -46,12 +46,21 @@ type testContext struct {
...
-	DisallowedOperationHeaders:         dynamicconfig.GetTypedPropertyFn([]string{"request-timeout"}),
+	Enabled:                 dynamicconfig.GetBoolPropertyFn(true),
+	MaxServiceNameLength:    dynamicconfig.GetIntPropertyFnFilteredByNamespace(len("service")),
+	MaxOperationNameLength:  dynamicconfig.GetIntPropertyFnFilteredByNamespace(len("op")),
+	MaxConcurrentOperations: dynamicconfig.GetIntPropertyFnFilteredByNamespace(2),
+	MaxOperationHeaderSize:  dynamicconfig.GetIntPropertyFnFilteredByNamespace(20),
+	DisallowedOperationHeaders: dynamicconfig.NewGlobalCachedTypedValue(dynamicconfig.NewCollection(
+		&dynamicconfig.StaticClient{
+			nexusoperations.DisallowedOperationHeaders.Key(): []string{"request-timeout"},
+		},
```
dnr, line 57: you don't need this if you just do it in the converter, right?

pdoerner: that's true. just wanted to be absolutely sure I guess 🙂

components/nexusoperations/config.go
```diff
@@ -178,7 +178,7 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
 				keys[i] = strings.ToLower(k)
```
dnr, line 178: maybe this is the problem? it shouldn't be modifying the input slice. and definitely shouldn't be modifying `defaultDisallowedOperationHeaders`

PR #7963, opened 2025-06-24
Add query support for ListWorkers

ychebotarev:
## What changed?
Add query support - filtering workers based on "SQL-like WHERE cause" - to ListWorkers.
Only non-metrics fields are supported at the moment:
* WorkerInstanceKey
* WorkerIdentity
* HostName
* TaskQueue
* DeploymentName
* BuildId
* SdkName
* SdkVersion
* StartTime
* LastHeartbeatTime
* Status

## Why?
Users should be ably to search the list of workers based on the pre-defined set of columns.

## How did you test it?
- [ ] built
- [x] added new unit test(s)


service/matching/workers/worker_query_eingine_test.go
dnr, line 1: typo in file name

service/matching/workers/query_errors.go
```diff
@@ -0,0 +1,18 @@
...
+var (
+	malformedSqlQueryErrMessage = "malformed query"
+	notSupportedErrMessage      = "operation is not supported"
+	invalidExpressionErrMessage = "invalid expression"
+)
+
+func NewQueryError(format string, a ...any) error {
+	message := fmt.Sprintf(format, a...)
+	return serviceerror.NewInvalidArgument(message)
+}
```
dnr: we have `serviceerror.NewInvalidArgumentf` now, just use that

service/matching/workers/registry_impl.go
```diff
@@ -285,8 +285,16 @@ func (m *registryImpl) RecordWorkerHeartbeat(nsID namespace.ID, workerHeartbeat
...
+		return m.filterWorkers(nsID, func(_ *workerpb.WorkerHeartbeat) bool { return true }), nil
+	}
+	queryEngine := newWorkerQueryEngine(nsID.String(), query)
 	return m.filterWorkers(nsID, func(heartbeat *workerpb.WorkerHeartbeat) bool {
-		return true
+		result, err := queryEngine.EvaluateWorker(heartbeat)
+		if err != nil {
+			return false
+		}
+		return result
```
dnr:
```suggestion
		return err == nil && result
```

service/matching/workers/registry_impl.go
```diff
@@ -285,8 +285,16 @@ func (m *registryImpl) RecordWorkerHeartbeat(nsID namespace.ID, workerHeartbeat
 	m.upsertHeartbeat(nsID, workerHeartbeat)
 }
 
-func (m *registryImpl) ListWorkers(nsID namespace.ID, _ string, _ []byte) ([]*workerpb.WorkerHeartbeat, error) {
+func (m *registryImpl) ListWorkers(nsID namespace.ID, query string, _ []byte) ([]*workerpb.WorkerHeartbeat, error) {
+	if query == "" {
+		return m.filterWorkers(nsID, func(_ *workerpb.WorkerHeartbeat) bool { return true }), nil
```
dnr: it'd be a little nicer to do it like:
```
  filter := func(_ *workerpb.WorkerHeartbeat) bool { return true }
  if query != "" { filter = ... }
  return m.filterWorkers(nsID, filter), nil
```
so there's just one call to filterWorkers

service/matching/workers/registry_impl.go
```diff
@@ -285,8 +285,16 @@ func (m *registryImpl) RecordWorkerHeartbeat(nsID namespace.ID, workerHeartbeat
...
+	}
+	queryEngine := newWorkerQueryEngine(nsID.String(), query)
```
dnr: hmm, shouldn't the parser be able to return an error?

service/matching/workers/worker_query_engine.go
```diff
@@ -0,0 +1,380 @@
...
+	workerSdkNameColName        = "SdkName"
+	workerSdkVersionColName     = "SdkVersion"
+	workerStartTimeColName      = "StartTime"
+	workerHeartbeatTimeColName  = "HeartbeatTime"
+	workerStatusColName         = "WorkerStatus"
+)
+
+/*
+FilterWorkers filters the list of per-namespace worker heartbeats against the provided query.
+The query should be a valid SQL query with WHERE clause.
```
dnr: for visibility queries the `WHERE` is implied.. don't we want to do the same here?

hmm, it looks like you do add the WHERE, so maybe this comment is wrong?

service/matching/workers/worker_query_engine.go
```diff
@@ -0,0 +1,380 @@
...
+			return hb.WorkerIdentity
+		},
+		workerHostNameColName: func(hb *workerpb.WorkerHeartbeat) string {
+			return hb.HostInfo.HostName
+		},
+		workerTaskQueueColName: func(hb *workerpb.WorkerHeartbeat) string {
+			return hb.TaskQueue
+		},
+		workerDeploymentNameColName: func(hb *workerpb.WorkerHeartbeat) string {
+			return hb.DeploymentVersion.DeploymentName
```
dnr, line 112: can this panic if there is no DeploymentVersion?

service/matching/workers/worker_query_engine.go
```diff
@@ -0,0 +1,380 @@
...
+func (w *workerQueryEngine) EvaluateWorker(hb *workerpb.WorkerHeartbeat) (bool, error) {
+	query, err := prepareQuery(w.query)
+	if err != nil {
+		return false, err
+	}
+
+	whereCause, err := getWhereCause(query)
+	if err != nil {
+		return false, err
+	}
```
dnr: this is parsing the query on every evaluation? shouldn't that happen upfront in newWorkerQueryEngine?

service/matching/workers/worker_query_engine.go
ychebotarev: yes, comment should be "without WHERE ..."

service/matching/workers/worker_query_engine.go
ychebotarev: fixed

service/matching/workers/worker_query_engine.go
ychebotarev: you right, good catch

service/matching/workers/registry_impl.go
ychebotarev: with moving pre-parsing into constructor - yes

service/matching/workers/registry_impl.go
ychebotarev: done

service/matching/workers/query_errors.go
ychebotarev: nice.

service/matching/workers/worker_query_eingine_test.go
ychebotarev: lol tnx.

service/matching/workers/worker_query_engine.go
```diff
@@ -0,0 +1,405 @@
...
+			return hb.WorkerInstanceKey
+		},
+		workerIdentityColName: func(hb *workerpb.WorkerHeartbeat) string {
+			return hb.WorkerIdentity
+		},
+		workerHostNameColName: func(hb *workerpb.WorkerHeartbeat) string {
+			if hb.HostInfo == nil {
+				return ""
+			}
+			return hb.HostInfo.HostName
```
dnr, line 103:
```suggestion
			return hb.HostInfo.GetHostName()
```

service/matching/workers/worker_query_engine.go
```diff
@@ -0,0 +1,405 @@
...
+		},
+		workerTaskQueueColName: func(hb *workerpb.WorkerHeartbeat) string {
+			return hb.TaskQueue
+		},
+		workerDeploymentNameColName: func(hb *workerpb.WorkerHeartbeat) string {
+			if hb.DeploymentVersion == nil {
+				return ""
+			}
+			return hb.DeploymentVersion.DeploymentName
```
dnr, line 112:
```suggestion
			return hb.DeploymentVersion.GetDeploymentName()
```

service/matching/workers/worker_query_engine.go
```diff
@@ -0,0 +1,405 @@
...
+	case *sqlparser.ComparisonExpr:
+		return w.evaluateComparison(e)
+	case *sqlparser.ParenExpr:
+		return w.evaluateExpression(e.Expr)
+	case *sqlparser.RangeCond:
+		return w.evaluateRange(e)
+	case *sqlparser.IsExpr:
+		return false, serviceerror.NewInvalidArgumentf("%s: 'is' expression", notSupportedErrMessage)
+	case *sqlparser.NotExpr:
+		return false, serviceerror.NewInvalidArgumentf("%s: 'not' expression", notSupportedErrMessage)
```
dnr, line 211: this seems trivial to support.. any reason not to? do we think this will be hard when we move to some persistent store?

service/matching/workers/registry_impl.go
```diff
@@ -285,8 +285,18 @@ func (m *registryImpl) RecordWorkerHeartbeat(nsID namespace.ID, workerHeartbeat
 	m.upsertHeartbeat(nsID, workerHeartbeat)
 }
 
-func (m *registryImpl) ListWorkers(nsID namespace.ID, _ string, _ []byte) ([]*workerpb.WorkerHeartbeat, error) {
-	return m.filterWorkers(nsID, func(heartbeat *workerpb.WorkerHeartbeat) bool {
-		return true
-	}), nil
+func (m *registryImpl) ListWorkers(nsID namespace.ID, query string, _ []byte) ([]*workerpb.WorkerHeartbeat, error) {
```
dnr, line 288: I'm still wondering how pagination will work, but ok to merge this for now

service/matching/workers/worker_query_engine.go
ychebotarev: this is just to be the same as our "visibility". But you right, I think it is a bad reason, and trivial to do. I will do that.

PR #7967, opened 2025-06-25
Matching fairness reader, writer, backlog manager

dnr:
## What changed?
- Add fairBacklogManager, fairTaskReader, fairTaskWriter and hook them up to matching.
- Add functional test.

## Why?
Matching task queue fairness.

## How did you test it?
- [x] built
- [ ] added new unit test(s)


service/matching/fairness.md
```diff
@@ -0,0 +1,225 @@
+
+# Fair Task Reader operation
+
+## Definitions
+
+**Task ID**: sequential int64 assigned to tasks. Two tasks never have the same id.
+
+**Pass**: int64 used to "spread out" tasks.
```
stephanos, line 8: Should we mention "stride scheduling" somewhere in the doc?

service/matching/fair_task_reader.go
```diff
@@ -0,0 +1,546 @@
...
+func (tr *fairTaskReader) getOldestBacklogTime() time.Time {
+	tr.lock.Lock()
+	defer tr.lock.Unlock()
+	return tr.backlogAge.oldestTime()
+}
+
+func (tr *fairTaskReader) completeTask(task *internalTask, res taskResponse) {
+	tr.lock.Lock()
+
+	// We might have a race where mergeTasks tries to a task from matcher (because new tasks
```
stephanos:
```suggestion
	// We might have a race where mergeTasks tries to read a task from matcher (because new tasks
```

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+
+	select {
+	case w.appendCh <- req:
+		select {
+		case err := <-ch:
+			metrics.TaskWriteLatencyPerTaskQueue.With(w.backlogMgr.metricsHandler).Record(time.Since(startTime))
+			return err
+		case <-w.backlogMgr.tqCtx.Done():
+			// if we are shutting down, this request will never make
+			// it to cassandra, just bail out and fail this request
```
stephanos: Should this comment be datastore-agnostic?

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+
+		weight, ok := overrideWeights[key]
+		if !ok {
+			weight = task.taskInfo.Priority.GetFairnessWeight()
+		}
+		if weight <= 0.0 {
+			weight = 1.0
+		} else if weight < minWeight {
+			weight = minWeight
+		}
```
stephanos: nit: since we use `max` for `inc` already; might as well use `min` & `max` here, too

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+
+		inc := max(1, int64(strideFactor/weight))
+
+		base := bases[task.subqueue].pass
+		base += int64(rand.Float64() * float64(inc)) // randomize keys within "initial" pass
```
stephanos: nit: I think it'd be helpful to add a brief comment explaining what's going on (e.g. Why add randomization? What's `inc`?)

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+
+		tasks[i].pass = w.counter.GetPass(key, base, inc)
+	}
+}
+
+func (w *fairTaskWriter) initState() error {
+	retryForever := backoff.NewExponentialRetryPolicy(1 * time.Second).
+		WithMaximumInterval(10 * time.Second).
```
stephanos: nit: could extract the two timings as top-level variables to make it easier to spot these

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+			w.pickPasses(reqs, bases)
+			var resp createFairTasksResponse
+			resp, err = w.db.CreateFairTasks(w.backlogMgr.tqCtx, reqs)
+			if err == nil {
+				w.backlogMgr.wroteNewTasks(resp)
+			} else {
+				w.logger.Error("Persistent store operation failure", tag.StoreOperationCreateTask, tag.Error(err))
+				w.backlogMgr.signalIfFatal(err)
+			}
+			unpin(err) // note this must be called after wroteNewTasks!
```
stephanos: nit: regarding the comment; it seems to imply that it would _only_ be needed to be called after `wroteNewTasks` which makes me wonder why it's outside the `if` branch. (I didn't get to pinning yet so maybe it'll become clearer later)

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+		var req *writeTaskRequest
+		select {
+		case <-w.backlogMgr.tqCtx.Done():
+			return
+		case req = <-w.appendCh:
+		}
+
+		// read a batch of requests from the channel
+		reqs = append(reqs[:0], req)
+		reqs = w.getWriteBatch(reqs)
```
stephanos: It look me a moment to put together that the first select is there to unblock after the first request is available; and the second select inside `getWriteBatch` is there to continue fetching as many as are allowed. Would it make sense to move the two `reqs = ` lines inside the `case`?

service/matching/db.go
```diff
@@ -387,6 +448,69 @@ func (db *taskQueueDB) CreateTasks(
...
+
+	for sq, tasks := range newTasks {
+		db.subqueues[sq].ApproximateBacklogCount += int64(len(tasks))
+	}
+
+	for sq, level := range newMaxLevel {
+		db.subqueues[sq].FairMaxReadLevel = fairLevelFromProto(db.subqueues[sq].FairMaxReadLevel).max(level).toProto()
+	}
+
+	resp, err := db.store.CreateTasks(
```
stephanos, line 490: on a related note: we have this trick to squeeze the priority into the row type; should we actually change the schema to include it since we're making a new schema anyway?

stephanos (commented):
First batch of comments :)

service/matching/fair_task_writer.go
```diff
@@ -0,0 +1,250 @@
...
+	var overrideWeights map[string]float32
+
+	for i, task := range tasks {
+		key := task.taskInfo.Priority.GetFairnessKey()
+
+		weight, ok := overrideWeights[key]
+		if !ok {
+			weight = task.taskInfo.Priority.GetFairnessWeight()
+		}
+		if weight <= 0.0 {
```
ychebotarev, line 127: can you add a comment describing this logic?
if weight is <0 then lower bound becomes 1, otherwise minWeight. 
why not just weight = max(minWeight, weight), why we need special code path for negative numbers?

service/matching/fair_task_writer.go
ychebotarev: +1 on that.
If I unwind the code I see this:
```
m.m[key] = max(bases[task.subqueue].pass +int64(rand.Float64() * float64(inc)), m.m[key]+inc)
```
so you move base forward for either inc, of some random offset from "pass". What is the reasoning behind this?

service/matching/fair_task_writer.go
ychebotarev: I would even extract this code and add it into getWriteBatch. It also took me some time to understand what is going on.

service/matching/db.go
dnr: yeah, I considered that. overall I think not:
- we only need one bit for type, and if we need more we can grab another few bits no problem
- adding a new column would (afaik) increase the size of all rows in disk, in memory, etc.
so I think the "hack" is actually totally fine to do for the long term.

service/matching/fair_task_writer.go
dnr: It's for zero, not for negative numbers:
0 (unspecified) => 1
anything else < 0.001 => 0.001

"weight = max(minWeight, weight)" would turn 0 into 0.001, but it should be 1.

service/matching/fair_task_writer.go
dnr: I'm going to take this out for now to just focus on the core.

The rand thing is an attempt to dither the passes so that a key with a high weight gets interleaved among others instead of all going in a block, e.g. A with weight 8 and B, C, D, .. H with weight 1 should go ABACADAEAF... instead of AAAAAAAABCDEF...

But we can put it back later with a specific test for that situation.

service/matching/fair_task_writer.go
dnr: This hasn't really changed from the original taskWriter, I just moved it out of the select to unindent it, and changed it to reuse the slice.

I moved it back into the select and tweaked things a little.. is that better?

service/matching/fair_task_writer.go
dnr: it's sort of the opposite: after getAndPinAckLevels, unpin must definitely be called. _if_ wroteNewTasks is called, then it has to be called before unpin.

I moved it into a new function with defer so the relationship is more clear

service/matching/fairness.md
dnr: yes, that'll be in `## Counting` below (and I'll probably rearrange it some more). this doc is focused on the reader, not the writer yet.

dnr:
I think I fixed the couple issues we noticed yesterday during the review. Two unit tests are reliably failing, but they have to do with approximatebacklogcount in the presence of db errors. It might have to do with retries on failed writes. I'll look into it later but I'm not sure we have to fix it for this PR. It might be easier with some new tests written from scratch, the existing ones are hard to understand. (Actually I think the test could be wrong)

dnr:
Weirdly, the unit tests that were failing ~60% of the time locally passed in CI, but the _functional_ test failed (twice), and I can't get that one to fail locally. I think a more complete unit/integration test with proper fault injection should flush all these out.

service/matching/fair_task_reader.go
```diff
@@ -0,0 +1,560 @@
...
+	tr.lock.Lock()
+
+	// Take the tasks in the matcher plus the tasks that were just written and sort them by level:
+
+	// Get currently loaded tasks. Note these values are *internalTask.
+	merged := tr.outstandingTasks.Select(func(k, v any) bool {
+		_, ok := v.(*internalTask)
+		return ok
+	})
+	// Add the tasks we just read/wrote. Note these values are *AllocatedTaskInfo.
```
stephanos:
```suggestion
	// Collect (1) currently loaded tasks in the matcher plus (2) the tasks we just read/wrote; sorted by level.
	
	// (1) Note these values are *internalTask.
	merged := tr.outstandingTasks.Select(func(k, v any) bool {
		_, ok := v.(*internalTask)
		return ok
	})
	// (2) Note these values are *AllocatedTaskInfo.
```

nit: trying to make it a smidge clearer.

service/matching/fair_task_reader.go
```diff
@@ -0,0 +1,560 @@
...
+
+	if highestLevel.id != 0 {
+		// If we have any tasks at all in memory, set readLevel to the maximum of that set.
+		// Otherwise leave readLevel unchanged.
+		tr.readLevel = highestLevel
+	}
+
+	// If there are remaining tasks in the merged set, they can't fit in memory. If they came
+	// from the tasks we just wrote, ignore them. If they came from matcher, remove them.
+	droppedAnyTasks := false
```
stephanos: nit: how about `evictedAnyTasks`? Would make it a tad clearer that these are essentially "cached". "Drop" sounds a bit more consequential.

service/matching/fair_task_reader.go
```diff
@@ -0,0 +1,560 @@
...
+			task.removeFromMatcher()
+		}
+	}
+
+	internalTasks := make([]*internalTask, len(tasks))
+	for i, t := range tasks {
+		level := fairLevelFromAllocatedTask(t)
+		internalTasks[i] = newInternalTaskFromBacklog(t, tr.completeTask)
+		// After we get to this point, we must eventually call task.finish or
+		// task.finishForwarded, which will call tr.completeTask.
```
stephanos, line 389: This comment confuses me a little; it sounds like it says you have to invoke these methods - but then I don't see it invoking them anywhere.

service/matching/fair_task_reader.go
```diff
@@ -0,0 +1,560 @@
...
+	// If we're at the end, then outstandingTasks is the whole queue so we can set count.
+	if count := tr.knownCountLocked(); count >= 0 {
+		tr.backlogMgr.db.setKnownFairBacklogCount(tr.subqueue, count)
+	}
+
+	// unlock before calling addTaskToMatcher
+	tr.lock.Unlock()
+
+	for _, task := range internalTasks {
+		tr.addTaskToMatcher(task)
```
stephanos, line 415: Given the race condition comment on line 338; is there a chance here that we might add a task again?

service/matching/fair_task_writer.go
dnr: I think that text was written back when Cadence only supported Cassandra :)

service/matching/fair_task_writer.go
dnr: this didn't change from previous versions of taskWriter, but I can clean it up...

service/matching/fair_task_reader.go
dnr: Oh that's much better!

service/matching/fair_task_reader.go
dnr: They're called from matchingEngine or somewhere else after a match. "We" in the sense of "matching service as a whole"

service/matching/fair_task_reader.go
dnr: I don't think so. Here's my reasoning:

Read and write both funnel into mergeTasks. For a given task, at most one call to mergeTasks will add it to outstandingTasks (if it's already there it'll be dropped from the incoming tasks), and that call will also be the one that calls addTaskToMatcher. Essentially, seeing at task not present in outstandingTasks already give it the right(+responsibility) to call addTaskToMatcher for that task.

dnr:
I'm going to merge this now even though I think there's still a bug somewhere (as found by some new testing code not in a PR yet). It'll be easier to rebase the whole branch on the various fixes in main and then focus on the remaining bugs.

PR #7968, opened 2025-06-25
fix: add timestamp to trigger immediately signal

spkane31:
## What changed?
add a timestamp to the `TriggerImmediately` schedule for use instead of now()

## Why?
We can make the wft “more idempotent”: put the desired trigger time in the trigger immediately request itself, so it goes in the signal payload, so all wfts attempt to start the target with the same id.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks

TBD @dnr  @lina-temporal


service/frontend/workflow_handler.go
```diff
@@ -3983,6 +3984,11 @@ func (wh *WorkflowHandler) PatchSchedule(ctx context.Context, request *workflows
 		return nil, err
 	}
 
+	// Set the timestamp to timestamppb.Now if unset
+	if request.Patch.GetTriggerImmediately() != nil && request.Patch.GetTriggerImmediately().TriggerTime == nil {
```
dnr: this has to be done before request.Patch is serialized or it's useless

service/frontend/workflow_handler.go
```diff
@@ -3983,6 +3984,11 @@ func (wh *WorkflowHandler) PatchSchedule(ctx context.Context, request *workflows
...
+		request.Patch.GetTriggerImmediately().TriggerTime = timestamppb.Now()
+	}
```
dnr: just a suggestion:
```suggestion
	if trigger := request.Patch.GetTriggerImmediately(); trigger != nil && trigger.TriggerTime == nil {
		trigger.TriggerTime = timestamppb.Now()
	}
```

dnr (approved):

service/worker/scheduler/testdata/replay_with_trigger_immediately_timestamp.json.gz
dnr, line 1: Assuming this was generated before the version change, I'd remove it since it doesn't reflect any actual version of the code anymore.

service/worker/scheduler/workflow.go
```diff
@@ -409,7 +411,15 @@ func (s *scheduler) processPatch(patch *schedulepb.SchedulePatch) {
...
-		now := s.now()
+		var now time.Time
+		if s.hasMinVersion(TriggerImmediatelyTimestamp) {
+			now = timestamp.TimeValue(trigger.ScheduledTime)
+			if now.IsZero() {
+				now = s.now()
+			}
+		} else {
+			now = s.now()
+		}
```
dnr: could simplify to
```suggestion
		now := s.now()
		if t := timestamp.TimeValue(trigger.ScheduledTime); !t.IsZero() && s.hasMinVersion(TriggerImmediatelyTimestamp) {
			now = t
		}
```
but it looks correct either way

service/worker/scheduler/workflow_test.go
```diff
@@ -1151,16 +1151,28 @@ func (s *workflowSuite) TestTriggerImmediate() {
...
+					s.Equal([]string{"myid-2022-06-01T00:15:00Z"}, s.runningWorkflows(), "Only the scheduled workflow should be running")
+				},
+			},
 			{
 				at: time.Date(2022, 6, 1, 0, 30, 0, 0, time.UTC),
 				f: func() {
 					// this one runs with overridden overlap policy
 					s.env.SignalWorkflow(SignalNamePatch, &schedulepb.SchedulePatch{
 						TriggerImmediately: &schedulepb.TriggerImmediatelyRequest{
+							ScheduledTime: timestamppb.New(time.Date(2022, 6, 1, 0, 30, 0, 0, time.UTC)),
```
dnr, line 1179: should we change the value to check that it's actually used?

I think then we'll need to bump the current version in this test?

spkane31: Changing the ScheduledTime to `ScheduledTime: timestamppb.New(time.Date(2022, 6, 1, 0, 40, 0, 0, time.UTC)),` will cause the test to fail with:

```sh
            	Error Trace:	/Users/seankane/git/github.com/temporalio/temporal/service/worker/scheduler/workflow_test.go:305
            	            				/Users/seankane/git/github.com/temporalio/temporal/service/worker/scheduler/workflow_test.go:1137
            	Error:      	Not equal: 
            	            	expected: 2
            	            	actual  : 1
            	Test:       	TestWorkflow/TestTriggerImmediate
            	Messages:   	started map[string]time.Time{"myid-2022-06-01T00:15:00Z":time.Date(2022, time.June, 1, 0, 15, 0, 0, time.UTC)}
```

I think that shows it's used as we expect. I don't see a way to indicate a test should fail though so I'm not sure what the change here should be

dnr (approved):

PR #7969, opened 2025-06-25
fix missing new run

hai719:
## What changed?
Add SuccessorRunId in ms and use that to get new run when generating sync state result.

## Why?
If the workflow is terminated because of `WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING`, NewExecutionRunId is not set in the mutable state. In this case, Sync state is not able to get the new run info. The new run will not be replicated to the target cluster.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/proto/internal/temporal/server/api/persistence/v1/executions.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/executions.proto
@@ -270,6 +270,9 @@ message WorkflowExecutionInfo {
     // Priority contains metadata that controls relative ordering of task processing
     // when tasks are backed up in a queue.
     temporal.api.common.v1.Priority priority = 104;
+
+    // Run ID of the execution that supersedes this one (via terminate, reset, or continue-as-new).
+    string successor_run_id = 105;
 }
 
 message ExecutionStats {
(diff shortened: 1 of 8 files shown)


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -270,6 +270,9 @@ message WorkflowExecutionInfo {
     // Priority contains metadata that controls relative ordering of task processing
     // when tasks are backed up in a queue.
     temporal.api.common.v1.Priority priority = 104;
+
+    // Run ID of the execution that supersedes this one (via terminate or continue-as-new).
+    string successor_run_id = 105;
```
xwduan, line 275: If this field is also for continue as new, shall we update this field at anywhere the executioninfo.NewExecutionRunId is updated?

hai719: When continue_as_new, it calls weContext.UpdateWorkflowExecutionWithNewAsActive ([code](https://github.com/temporalio/temporal/blob/6f79cf65d580dbbcd335437054a068b77df9b0d6/service/history/api/respondworkflowtaskcompleted/api.go#L586)), successor_run_id will be updated as well.

PR #7970, opened 2025-06-26
Reset backlog age when backlog count becomes 0

ShahabT:
## What changed?
Reset backlog age when backlog count becomes zero.

## Why?
Backlog age metric remains non-zero even after the backlog is empty. This PR fixes that.

## How did you test it?
- [x] built
- [ ] rrun locally and tested manually
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/service/matching/db.go
+++ b/service/matching/db.go
@@ -227,6 +227,7 @@ func (db *taskQueueDB) OldUpdateState(
 	maxReadLevel := db.getMaxReadLevelLocked(subqueueZero)
 	if ackLevel == maxReadLevel {
 		db.subqueues[subqueueZero].ApproximateBacklogCount = 0
+		db.subqueues[subqueueZero].oldestTime = time.Time{} // zero time means no backlog
 	}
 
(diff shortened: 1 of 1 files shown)


service/matching/db.go
```diff
@@ -275,7 +275,7 @@ func (db *taskQueueDB) updateAckLevelAndBacklogStats(subqueue int, newAckLevel i
 	if newAckLevel == db.getMaxReadLevelLocked(subqueue) {
 		// Reset approximateBacklogCount to fix the count divergence issue
 		dbQueue.ApproximateBacklogCount = 0
-		dbQueue.oldestTime = oldestTime
+		dbQueue.oldestTime = time.Time{} // zero time means no backlog
```
dnr: hmm, wouldn't oldestTime have to be zero already here? it comes from `tr.backlogAge.oldestTime()`, which is the zero time if there are no tasks in-mem. if it's not, that suggests maybe there's a bookkeeping bug with `tr.backlogAge`

PR #7972, opened 2025-06-26
TestRunner skips reporting parent test cases

stephanos:
## What changed?

(1) Testrunner won't report parents like `TestCallbacksSuite` from `TestCallbacksSuite/TestWorkflowCallbacks_InvalidArgument`.

(2) Testrunner will still write JUnit file even if missing test case is detected.

## Why?

re (1): Less verbose; reporting parents doesn't add anything.

re (2): JUnit file is still valid + helpful and should be written.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/tools/testrunner/junit.go
+++ b/tools/testrunner/junit.go
@@ -122,9 +122,11 @@ func mergeReports(reports []*junitReport) (*junitReport, error) {
 		return nil, errors.New("no reports to merge")
 	}
 
-	var combined junit.Testsuites
 	var reportingErrs []error
-
+	var combined junit.Testsuites
+	combined.XMLName = reports[0].Testsuites.XMLName
+	combined.Name = reports[0].Testsuites.Name
+	
 	// Collect all test case names into the tree
 	testNameTree := redblacktree.NewWithStringComparator()
 	for _, report := range reports {
@@ -136,11 +138,6 @@ func mergeReports(reports []*junitReport) (*junitReport, error) {
 	}
 
 	for i, report := range reports {
-		if i == 0 {
-			combined.XMLName = report.Testsuites.XMLName
-			combined.Name = report.Testsuites.Name
-		}
-
 		combined.Tests += report.Testsuites.Tests
 		combined.Errors += report.Testsuites.Errors
 		combined.Failures += report.Testsuites.Failures
--- a/tools/testrunner/junit.go
+++ b/tools/testrunner/junit.go
@@ -9,12 +9,14 @@ import (
 	"os"
 	"strings"
 
+	"github.com/emirpasic/gods/trees/redblacktree"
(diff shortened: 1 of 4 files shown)


tools/testrunner/junit.go
```diff
@@ -138,36 +150,56 @@ func mergeReports(reports []*junitReport) (*junitReport, error) {
...
 	}
 
-	return &junitReport{Testsuites: combined}, nil
+	return &junitReport{
+		Testsuites:    combined,
+		reportingErrs: reportingErrs,
+	}, nil
+}
+
+func isLeaf(name string, tree *redblacktree.Tree) bool {
```
ychebotarev, line 193: I would add them all to the list, sort it, and `if L[i-1] is a prefix of L[i] - remove L[i-1]`?

tools/testrunner/junit.go
```diff
@@ -119,16 +121,26 @@ func mergeReports(reports []*junitReport) (*junitReport, error) {
...
+	for _, report := range reports {
+		for _, suite := range report.Testsuites.Suites {
+			for _, tc := range suite.Testcases {
+				testNameTree.Put(tc.Name, struct{}{})
+			}
+		}
+	}
 
 	for i, report := range reports {
+		if i == 0 {
```
ychebotarev: I will keep it outside, like it was before. Less noise inside a loop

ychebotarev (approved):
minor suggestoins

tools/testrunner/junit.go
stephanos: hm, but it's not guaranteed that `L[i-1]` is always the parent. if we have 4 test cases in a suite, only the 1st one would be right after the suite.

for example: `[SuiteA, SuiteA/Test1, SuiteA/Test2, SuiteA/Test3, SuiteA/Test4]`.

stephanos: And for nested tests (ie two levels) you will only know that there's another sub-test once you see the sub-test.

stephanos: I'll merge for now. But feel free to comment, and if we have a better approach, I'll update it.

tools/testrunner/junit.go
```diff
@@ -138,36 +147,56 @@ func mergeReports(reports []*junitReport) (*junitReport, error) {
...
+
+			var missing []string
+			for _, f := range prevFailures {
+				if _, ok := currCases[f]; !ok {
+					missing = append(missing, f)
 				}
 			}
-			if len(casesMissing) > 0 {
-				return nil,
-					fmt.Errorf("expected a rerun of all failures from the previous attempt, missing: %v", casesMissing)
```
stephanos, line 152: This causes us to under-report failed tests since this will skip the write of the report to disk.

PR #7976, opened 2025-06-27
Always issue RPC in DescribeVersionedTaskQueues

stephanos:
## What changed?

Always go through matching client to query `DescribeTaskQueue` endpoint.

## Why?

Ensure ownership is correct; a shortcut to the local partition doesn't account for that.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -1397,7 +1397,7 @@ func (e *matchingEngineImpl) DescribeVersionedTaskQueues(
 						Name: tq.Name,
 						Kind: enumspb.TASK_QUEUE_KIND_NORMAL,
 					},
-					TaskQueueType: enumspb.TaskQueueType(tq.Type),
+					TaskQueueType: tq.Type,
 					ReportStats:   true,
 				},
 				Version: request.Version,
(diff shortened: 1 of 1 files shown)


service/matching/matching_engine.go
```diff
@@ -1389,32 +1389,22 @@ func (e *matchingEngineImpl) DescribeVersionedTaskQueues(
...
-					Kind: enumspb.TASK_QUEUE_KIND_NORMAL,
+		tqResp, err := e.matchingRawClient.DescribeTaskQueue(ctx,
+			&matchingservice.DescribeTaskQueueRequest{
+				NamespaceId: request.GetNamespaceId(),
+				DescRequest: &workflowservice.DescribeTaskQueueRequest{
+					TaskQueue: &taskqueuepb.TaskQueue{
+						Name: tq.Name,
+						Kind: enumspb.TASK_QUEUE_KIND_NORMAL,
+					},
+					TaskQueueType: enumspb.TaskQueueType(tq.Type),
```
dnr: this is already a TaskQueueType, isn't it?
```suggestion
					TaskQueueType: tq.Type,
```

PR #7979, opened 2025-06-27
fix flaky tests in worker-deployment

Shivs11:
## What changed?
- previously, our functional tests to verify if a time field was correctly set by comparing it with a fixed (set by us) defined upper bound. This was becoming too error prone since the upper bound was proving to be small in some cases. I fixed this by now checking if the time fields are in the following bounds: `start-time of the test` <= `time-field` <= `time.now()`
- also tweaked a few things to now pass in the context created inside a `Eventually` function to all the helpers it is using. 

## Why?
- to remove flakes 

## How did you test it?
- [x] built
- [ ] added new unit test(s)


tests/worker_deployment_test.go
```diff
@@ -2217,47 +2222,56 @@ func (s *WorkerDeploymentSuite) tryDeleteVersion(
...
+	s.verifyTimestampEquality(a, expected.GetCurrentVersionChangedTime(), actual.GetCurrentVersionChangedTime())
 }
 
-func (s *WorkerDeploymentSuite) verifyWorkerDeploymentInfo(expected, actual *deploymentpb.WorkerDeploymentInfo, maxDuration time.Duration) {
-	s.Equal(expected.GetName(), actual.GetName())
-	s.verifyTimestampEquality(expected.GetCreateTime(), actual.GetCreateTime(), maxDuration)
-	s.Equal(expected.GetLastModifierIdentity(), actual.GetLastModifierIdentity())
-	s.verifyRoutingConfig(expected.GetRoutingConfig(), actual.GetRoutingConfig(), maxDuration)
+func (s *WorkerDeploymentSuite) verifyWorkerDeploymentInfo(a *require.Assertions, expected, actual *deploymentpb.WorkerDeploymentInfo) {
+
```
stephanos: nit: remove empty line

tests/worker_deployment_test.go
```diff
@@ -2217,47 +2222,56 @@ func (s *WorkerDeploymentSuite) tryDeleteVersion(
...
 }
 
-func (s *WorkerDeploymentSuite) verifyTimestampEquality(expected, actual *timestamppb.Timestamp, maxDuration time.Duration) {
-	s.True((expected == nil) == (actual == nil))
-	if expected != nil {
-		s.True(expected.AsTime().Sub(actual.AsTime()) < maxDuration)
+// verifyTimestampEquality asserts if the actual timestamp is set to an appropriate value. It
+// does this check by checking if the timestamp set is within respectable bounds or is equal to the
+// expected timestamp.
+func (s *WorkerDeploymentSuite) verifyTimestampEquality(a *require.Assertions, expected, actual *timestamppb.Timestamp) {
```
stephanos: `verifyTimestampEquality` is a bit of a misnomer since it doesn't actually verify equality. How about `verifyTimestampWithinRange`?

tests/worker_deployment_test.go
```diff
@@ -2483,12 +2472,17 @@ func (s *WorkerDeploymentSuite) startAndValidateWorkerDeployments(
 		}
 
 		for _, expectedDeploymentSummary := range expectedDeploymentSummaries {
-			deploymentSummaryValidated := false
+			deploymentSummaryFound := false
 			for _, actualDeploymentSummary := range actualDeploymentSummaries {
-				deploymentSummaryValidated = deploymentSummaryValidated ||
-					s.verifyWorkerDeploymentSummary(expectedDeploymentSummary, actualDeploymentSummary)
+				// our assumption is that deployment summaries with the same name are fully ready for checks. May not be true since visibility might take time to update other fields.
```
stephanos:
```suggestion
				// Our assumption is that deployment summaries with the same name are fully ready for checks.
				// May not be true since visibility might take time to update other fields.
```

nit: readability and punctuation

stephanos (approved):

PR #7988, opened 2025-06-30
Exclude Google Cloud Storage gRPC packages

stephanos:
## What changed?

Stripped Google Cloud Storage's gRPC dependencies out for builds via the Makefile.

## Why?

Reduce binary size by 16MB; and reduces build times a little (ie linking).

**Before**

```
-rwxr-xr-x@ 1 stephan  staff   119M Jun 30 11:23 temporal-server
```

**After**

```
-rwxr-xr-x@ 1 stephan  staff   103M Jun 30 11:24 temporal-server
```

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks

Yes, since there are no end-to-end tests, if anyone started using `storage.NewGRPCClient` that would fail silently. I added a (crude) test to catch that.

### DIFF
--- a/common/archiver/gcloud/connector/client_test.go
+++ b/common/archiver/gcloud/connector/client_test.go
@@ -310,7 +310,7 @@ func (s *clientSuite) TestNoGRPCUsage() {
 	packageFiles, err := filepath.Glob("*.go")
 	s.NoError(err)
 
-	var checkedFile bool
+	var checkedClientFile bool
 	for _, file := range packageFiles {
 		if strings.HasSuffix(file, "_test.go") {
 			continue
@@ -321,7 +321,8 @@ func (s *clientSuite) TestNoGRPCUsage() {
 		if strings.Contains(string(content), "NewGRPCClient") {
 			s.T().Errorf("❌ Found forbidden gRPC usage in file: %s", file)
 		}
-		checkedFile = true
+
+		checkedClientFile = checkedClientFile || strings.HasSuffix(file, "client.go")
 	}
-	s.True(checkedFile, "checked at least one file")
+	s.True(checkedClientFile, "checked client.go file")
 }
--- a/common/archiver/gcloud/connector/client_test.go
+++ b/common/archiver/gcloud/connector/client_test.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"errors"
 	"io"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -243,7 +245,6 @@ func (s *clientSuite) TestQuery() {
 }
 
 func (s *clientSuite) TestQueryWithFilter() {
-
 	ctx := context.Background()
 	mockBucketHandleClient := connector.NewMockBucketHandleWrapper(s.controller)
 	mockStorageClient := connector.NewMockGcloudStorageClient(s.controller)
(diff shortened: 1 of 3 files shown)


stephanos:
I just realized that `.goreleaser.yml` doesn't use the Makefile. But CI does.

dnr:
> I just realized that `.goreleaser.yml` doesn't use the Makefile. But CI does.

I think we should get rid of goreleaser and only publish the docker images and the cli. We can still support binaries built other ways, but we don't need to publish them. We had this discussion a while ago and iirc everyone agreed in principle, but we wanted to get some idea of how many people used those binaries.

stephanos:
👍 We use the Makefile for the Docker build, so that would trickle down: https://github.com/temporalio/docker-builds/blob/0e3fc4b7dbd0a37f555ccacd00ae4869932098e7/Makefile#L48

common/archiver/gcloud/connector/client_test.go
```diff
@@ -303,3 +303,26 @@ func newWorkflowIDPrecondition(workflowID string) connector.Precondition {
 		return false
 	}
 }
+
+// Ensures that no code in this package accidentally uses gRPC functions since they are stripped from
+// the binary via `disable_grpc_modules`. This is crude but effective.
+func (s *clientSuite) TestNoGRPCUsage() {
+	packageFiles, err := filepath.Glob("*.go")
```
dnr: should we look in other directories within `common/archiver/gcloud` also?

dnr (approved):

common/archiver/gcloud/connector/client_test.go
```diff
@@ -303,3 +303,26 @@ func newWorkflowIDPrecondition(workflowID string) connector.Precondition {
 		return false
 	}
 }
+
+// Ensures that no code in this package accidentally uses gRPC functions since they are stripped from
+// the binary via `disable_grpc_modules`. This is crude but effective.
+func (s *clientSuite) TestNoGRPCUsage() {
```
ychebotarev, line 309: what will happed if we don't run this test and some1 call NewGRPCClient?

ychebotarev: (point I'm going to - do we actually need it?)

stephanos: Without the test, someone could switch over our usage from JSON to GRPC and - unless they do an end-to-end test (which they should ...) - merge the change and break the functionality (ie it will crash, I believe) without anyone noticing. Our existing tests are only testing a mock of the GCS client.

stephanos: Actually, maybe it wouldn't crash. From the [link in the GH issue](https://pkg.go.dev/cloud.google.com/go/storage#hdr-gRPC_API):

> One major advantage of the gRPC API is that it can use [Direct Connectivity](https://cloud.google.com/vpc-service-controls/docs/set-up-private-connectivity#direct-connectivity), enabling requests to skip some proxy steps and reducing responce latency. Requirements to use Direct Connectivity include:

ychebotarev: >unless they do an end-to-end test (which they should ...) 

yeah, my point. I think we are protecting from something that not just should't, but also will not happen. (for this to happen those who make change need to completely ignore func tests, local e2e testing, other e2e testings).
But I'm fine either way

stephanos: Our functional testing is _not_ end-to-end, ie _not_ talking to GCS. Unless there are instructions somewhere that I missed, the developer has to go _out of their way_ to set up a manual test scenario for this. What I'm saying is, there is no safety net.

PR #7989, opened 2025-06-30
config: enable dropNonRetryableTasks prometheus metric

spkane31:
## What changed?
Removing the dynamic config for `matching.dropNonRetryableTasks` and enable scraping of the metric

## Why?
We want to scrape this metric and alerts for spikes

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
_Potentially_ this metric could have high cardinality but this is not expected

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1244,11 +1244,6 @@ these log lines can be noisy, we want to be able to turn on and sample selective
 		5*time.Second,
 		`TaskQueueInfoByBuildIdTTL serves as a TTL for the cache holding DescribeTaskQueue partition results`,
 	)
-	MatchingDropNonRetryableTasks = NewGlobalBoolSetting(
-		"matching.dropNonRetryableTasks",
-		true,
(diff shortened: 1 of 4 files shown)


common/dynamicconfig/constants.go
```diff
@@ -1246,7 +1246,7 @@ these log lines can be noisy, we want to be able to turn on and sample selective
 	)
 	MatchingDropNonRetryableTasks = NewGlobalBoolSetting(
 		"matching.dropNonRetryableTasks",
-		false,
+		true,
```
Shivs11: was thinking if we really need this dynamic-config, which serves as a gate, at all - now that CDS has given us the green light that they have fixed everything on their end, I don't think we need this as a separate dynamic config

spkane31: Do we want to remove this entirely or leave it as a fallback in case it needs to be turned off in the future?

spkane31: Removed config setting

PR #7995, opened 2025-07-01
[CHASM] Perform access validation in Node.Get

lina-temporal:
Still in draft for:
- Detached/newly-created nodes
- Setting OperationIntent from side effect tasks

## What changed?
- Adds `validateAccess` to `chasm.Node` which performs the access rule check

## Why?
- Prevents writes to closed CHASM trees

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/component.go
+++ b/chasm/component.go
@@ -110,7 +110,7 @@ func newContextWithOperationIntent(
 func operationIntentFromContext(
 	ctx context.Context,
 ) OperationIntent {
-	intent, ok := ctx.Value(engineCtxKey).(OperationIntent)
+	intent, ok := ctx.Value(operationIntentCtxKey).(OperationIntent)
 	if !ok {
 		return OperationIntentUnspecified
 	}
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -36,6 +36,7 @@ var (
 )
 
 var (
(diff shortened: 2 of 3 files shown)


chasm/component.go
```diff
@@ -110,7 +110,7 @@ func newContextWithOperationIntent(
 func operationIntentFromContext(
 	ctx context.Context,
 ) OperationIntent {
-	intent, ok := ctx.Value(engineCtxKey).(OperationIntent)
+	intent, ok := ctx.Value(operationIntentCtxKey).(OperationIntent)
```
yycptt, line 113: oops. Thanks for catching this!

chasm/tree.go
```diff
@@ -36,6 +36,7 @@ var (
 )
 
 var (
+	errAccessCheckFailed    = serviceerror.NewFailedPrecondition("access check failed, CHASM tree is closed for writes")
```
yycptt: hmm return notFound for now (I guess similar to WorkflowExecutionAlreadyCompleted which is notFound as well). otherwise task processing side will stuck and keep retrying unless you catch and convert those errors in task executor.

We certainly need a better story re. those errors, but I don't want worry about them for now.

yycptt (approved):

PR #7997, opened 2025-07-01
Extract fairness key from activity ID

stephanos:
## What changed?

Extract fairness key from activity ID.

## Why?

For testing purposes only.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

It will be removed before merging to `main`; but even if it's not, the identifier was carefully chosen to minimize overlap with users'.

### DIFF
--- a/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
+++ b/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
@@ -1,9 +1,11 @@
 package respondworkflowtaskcompleted
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pborman/uuid"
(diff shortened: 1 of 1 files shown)


service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -423,6 +425,13 @@ func (handler *workflowTaskCompletedHandler) handleCommandScheduleActivity(
 	executionInfo := handler.mutableState.GetExecutionInfo()
 	namespaceID := namespace.ID(executionInfo.NamespaceId)
 
+	// TODO(fairness): remove this again once the SDK allows setting the fairness key
+	fairnessKeyPrefix := "x-temporal-internal-fairness-key-"
+	if strings.HasPrefix(attr.GetActivityId(), fairnessKeyPrefix) {
+		attr.Priority = cmp.Or(attr.Priority, &commonpb.Priority{})
+		attr.Priority.FairnessKey = strings.TrimPrefix(attr.GetActivityId(), fairnessKeyPrefix)
+	}
```
dnr: there's a new CutPrefix:
```suggestion
	const fairnessKeyPrefix = "x-temporal-internal-fairness-key-"
	if key, ok := strings.CutPrefix(attr.GetActivityId(), fairnessKeyPrefix); ok {
		attr.Priority = cmp.Or(attr.Priority, &commonpb.Priority{})
		attr.Priority.FairnessKey = key
	}
```

dnr: I'm not quite sure this is enough though since don't activity ids have to be unique, so we couldn't use the same fairness key twice? or at least twice before the previous one completes?

maybe we need to scan for a delimiter in the middle instead of a prefix?

stephanos: Uh, I didn't know about CutPrefix; nice.

re unique: Oh, let me check that. I ran it locally and I didn't see complaints. The docs only say "Business level activity ID".

stephanos: Okay, I see `GetActivityByActivityID` so at least some parts are expecting it to be unique. I'll make a change.

dnr (approved):

PR #8000, opened 2025-07-02
Route heartbeats from Poll/ShutDown

ychebotarev:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -874,6 +874,22 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
 	pollerID := uuid.New()
 	childCtx := wh.registerOutstandingPollContext(ctx, pollerID, namespaceID.String())
 	defer wh.unregisterOutstandingPollContext(pollerID, namespaceID.String())
+
+	// route heartbeat to the matching service only if the request is valid (all validation checks passed)
+	if request.WorkerHeartbeat != nil {
+		_, err = wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
+			NamespaceId: namespaceID.String(),
+			HeartbeartRequest: &workflowservice.RecordWorkerHeartbeatRequest{
+				Namespace:       request.GetNamespace(),
+				Identity:        request.GetIdentity(),
+				WorkerHeartbeat: request.WorkerHeartbeat,
+			},
+		})
+
+		// drop the heartbeat, it is not needed to process the poll request
+		request.WorkerHeartbeat = nil
+	}
+
 	matchingResp, err := wh.matchingClient.PollWorkflowTaskQueue(childCtx, &matchingservice.PollWorkflowTaskQueueRequest{
 		NamespaceId: namespaceID.String(),
 		PollerId:    pollerID,
@@ -2679,6 +2695,21 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
 	namespaceId, err := wh.namespaceRegistry.GetNamespaceID(namespace.Name(request.GetNamespace()))
 	if err != nil {
 		return nil, err
+	}
+
+	// route heartbeat to the matching service
+	if request.WorkerHeartbeat != nil {
+		_, err = wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
+			NamespaceId: namespaceId.String(),
+			HeartbeartRequest: &workflowservice.RecordWorkerHeartbeatRequest{
+				Namespace:       request.GetNamespace(),
+				Identity:        request.GetIdentity(),
+				WorkerHeartbeat: request.WorkerHeartbeat,
+			},
+		})
+
+		// drop the heartbeat, it is not needed to process the poll request
+		request.WorkerHeartbeat = nil
 	}
 
 	// TODO: update poller info to indicate poller was shut down (pass identity/reason along)


service/frontend/workflow_handler.go
```diff
@@ -874,6 +874,22 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
 	pollerID := uuid.New()
 	childCtx := wh.registerOutstandingPollContext(ctx, pollerID, namespaceID.String())
 	defer wh.unregisterOutstandingPollContext(pollerID, namespaceID.String())
+
+	// route heartbeat to the matching service only if the request is valid (all validation checks passed)
+	if request.WorkerHeartbeat != nil {
```
dnr, line 879: only workflow polls, not activity/nexus polls?

service/frontend/workflow_handler.go
```diff
@@ -874,6 +874,22 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
...
+		_, err = wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
```
dnr: I don't like doing this synchronously, it'll add latency to every poll. can we just fire off a goroutine and do it in the background?

service/frontend/workflow_handler.go
ychebotarev: well...yes for now. May change after todays meeting.

service/frontend/workflow_handler.go
ychebotarev: that was my question in "Potential risks"
Yes we can, will do that.

service/frontend/workflow_handler.go
```diff
@@ -874,6 +874,31 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
...
+	if request.WorkerHeartbeat != nil {
+
+		// route heartbeat to the matching service only if the request is valid (all validation checks passed)
+		go func() {
+			_, err := wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
+				NamespaceId: namespaceID.String(),
+				HeartbeartRequest: &workflowservice.RecordWorkerHeartbeatRequest{
+					Namespace:       request.GetNamespace(),
+					Identity:        request.GetIdentity(),
+					WorkerHeartbeat: request.WorkerHeartbeat,
```
dnr: this is going to race with setting it nil below

service/frontend/workflow_handler.go
```diff
@@ -874,6 +874,31 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
...
+
+	if request.WorkerHeartbeat != nil {
+
+		// route heartbeat to the matching service only if the request is valid (all validation checks passed)
+		go func() {
+			_, err := wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
+				NamespaceId: namespaceID.String(),
+				HeartbeartRequest: &workflowservice.RecordWorkerHeartbeatRequest{
+					Namespace:       request.GetNamespace(),
+					Identity:        request.GetIdentity(),
```
dnr: nit: request can't be nil here so can use direct field access

service/frontend/workflow_handler.go
```diff
@@ -874,6 +874,31 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
...
+					WorkerHeartbeat: request.WorkerHeartbeat,
+				},
+			})
+
+			if err != nil {
+				wh.logger.Error("Failed to record worker heartbeat.",
+					tag.WorkflowTaskQueueName(request.GetTaskQueue().GetName()),
```
dnr: nit: request.TaskQueue can't be nil here so can use direct field access

service/frontend/workflow_handler.go
ychebotarev: ok, this one stays, but will be used as a backup I guess. Just because we already have it.
Main will be Nexus Poll request.
No activity.

service/frontend/workflow_handler.go
ychebotarev: removed

service/frontend/workflow_handler.go
```diff
@@ -874,6 +875,29 @@ func (wh *WorkflowHandler) PollWorkflowTaskQueue(ctx context.Context, request *w
 	pollerID := uuid.New()
 	childCtx := wh.registerOutstandingPollContext(ctx, pollerID, namespaceID.String())
 	defer wh.unregisterOutstandingPollContext(pollerID, namespaceID.String())
+
+	if request.WorkerHeartbeat != nil {
+
+		// route heartbeat to the matching service only if the request is valid (all validation checks passed)
+		go func() {
+			heartbeats := []*workerpb.WorkerHeartbeat{request.WorkerHeartbeat}
```
dnr: we still want to remove it from the request, right?

service/frontend/workflow_handler.go
```diff
@@ -4856,16 +4898,35 @@ func (wh *WorkflowHandler) PollNexusTaskQueue(ctx context.Context, request *work
...
+
 	if len(request.GetIdentity()) > wh.config.MaxIDLengthLimit() {
 		return nil, errIdentityTooLong
 	}
 
-	if err := wh.validateVersioningInfo(request.Namespace, request.WorkerVersionCapabilities, request.TaskQueue); err != nil {
-		return nil, err
+	// route heartbeat to the matching service
+	if len(request.WorkerHeartbeat) > 0 {
+		_, err := wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
```
dnr: should this one be async too? and remove the heartbeat from the request?

ychebotarev: done async

service/frontend/workflow_handler.go
ychebotarev: right. removed

dnr (approved):

PR #8003, opened 2025-07-02
Handle transient namespace rename mismatch

prathyushpv:
## What changed?
As part of deleting a namespace, we rename the namespace with -deleted suffix. This rename happens in two steps. 
1. Change the mapping in namespaces_by_id table from old namespace name to new name.
2. In a batch operation, create a new namespace entry in namespaces table and delete the old entry.

When step 1 is complete and step 2 is not completed, if a client try to get this namespace details by its ID, it will not find the new entry yet. Server will return not found error in that case. This can cause issues with client. Returning unavailable in this case. This inconsistency should resolve when the second step finishes.

## Why?
Returning inconsistent data about namespace state could cause issues in clients.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/common/persistence/cassandra/metadata_store.go
+++ b/common/persistence/cassandra/metadata_store.go
@@ -3,10 +3,8 @@ package cassandra
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"go.temporal.io/api/serviceerror"
-	"go.temporal.io/server/common/backoff"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
 	p "go.temporal.io/server/common/persistence"


common/persistence/cassandra/metadata_store.go
```diff
@@ -323,17 +328,24 @@ func (m *MetadataStore) GetNamespace(
...
-	err = query.Scan(
-		nil,
-		nil,
-		&detail,
-		&detailEncoding,
-		&notificationVersion,
-		&isGlobalNamespace,
-	)
+	policy := backoff.NewConstantDelayRetryPolicy(namespaceStaleRetryDelay).WithMaximumAttempts(namespaceStaleRetryAttempts)
+	err = backoff.ThrottleRetry(func() error {
```
ychebotarev: I don't see why we need this retry. It is not up for this function to keep things in sync. 
Add retry to the code in DeleteNamespace workflow that does DescribeNamespace is case we return Unavailable

ychebotarev: and this one should call both even if first return NotFound. If both return NontFound - return NotFound, and one of them - 500 (serviceerror, retrievable). Client should retry.

yycptt: +1 on retry is unnecessary here. persistence client already retry unavailable error once.


> this one should call both even if first return NotFound?

Why? I think metadata store implementation always update namespace_by_id table first, and namespace ID never changes.

prathyushpv: Yes. Removed retry.

prathyushpv: We have an inconsistency only if the entry is found in `namespaces_by_id` table and not found in `namespaces` table. If entry is not found in namespaces_by_id that, that means namespace does not exist.

PR #8004, opened 2025-07-03
Implemented TaskQueueConfig persistence in UpdateTaskQueueConfig API

sivagirish81:
## What changed?
+ Introduced a new  Update TaskQueueConfigApi.
+ Implemented the corresponding handlers in the frontend service and the matching service.
+ Persistence of TaskQueueConfig from the  UpdateTaskQueueConfigApi.
+ Return the config response as part of the DescribeTaskQueue api.

## Why?
+ This is the first part of the UpdateTaskQueueConfig Implementation.
+ Goal is to persist the TaskQueueConfig and handle nil values in the update request.
+ Next steps would be to attach the corresponding rate limiters with the persisted configs.

## How did you test it?
- [x] built

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -58,7 +58,6 @@ import (
 	"go.temporal.io/server/service/history/api"
 	"go.temporal.io/server/service/worker/deployment"
 	"go.temporal.io/server/service/worker/workerdeployment"
-	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
@@ -1206,7 +1205,7 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 								partitionStats := vii.PhysicalTaskQueueInfo.TaskQueueStats
 								mergedStats = &taskqueuepb.TaskQueueStats{
 									ApproximateBacklogCount: totalStats.ApproximateBacklogCount + partitionStats.ApproximateBacklogCount,
-									ApproximateBacklogAge:   largerBacklogAge(totalStats.ApproximateBacklogAge, partitionStats.ApproximateBacklogAge),
+									ApproximateBacklogAge:   oldestBacklogAge(totalStats.ApproximateBacklogAge, partitionStats.ApproximateBacklogAge),
 									TasksAddRate:            totalStats.TasksAddRate + partitionStats.TasksAddRate,
 									TasksDispatchRate:       totalStats.TasksDispatchRate + partitionStats.TasksDispatchRate,
 								}
@@ -1341,10 +1340,7 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 				}
 				for _, vii := range partitionResp.VersionsInfoInternal {
 					partitionStats := vii.PhysicalTaskQueueInfo.TaskQueueStats
-					taskQueueStats.ApproximateBacklogCount += partitionStats.ApproximateBacklogCount
-					taskQueueStats.ApproximateBacklogAge = largerBacklogAge(taskQueueStats.ApproximateBacklogAge, partitionStats.ApproximateBacklogAge)
-					taskQueueStats.TasksAddRate += partitionStats.TasksAddRate
-					taskQueueStats.TasksDispatchRate += partitionStats.TasksDispatchRate
+					mergeStats(taskQueueStats, partitionStats)
 				}
 			}
 			pm.PutCache(cacheKey, taskQueueStats)
@@ -2927,14 +2923,6 @@ func (e *matchingEngineImpl) reviveBuildId(ns *namespace.Namespace, taskQueue st
 // be processed on the normal queue.
 func stickyWorkerAvailable(pm taskQueuePartitionManager) bool {
 	return pm != nil && pm.HasPollerAfter("", time.Now().Add(-stickyPollerUnavailableWindow))
-}
-
-// largerBacklogAge returns the larger BacklogAge
-func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *durationpb.Duration) *durationpb.Duration {
-	if rootBacklogAge.AsDuration() > currentPartitionAge.AsDuration() {
-		return rootBacklogAge
-	}
-	return currentPartitionAge
 }
 
 func buildRateLimitConfig(update *workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate, updateTime *timestamppb.Timestamp) *taskqueuepb.RateLimitConfig {
--- a/service/matching/matching_engine_test.go
+++ b/service/matching/matching_engine_test.go
@@ -3175,7 +3175,7 @@ func (s *matchingEngineSuite) pollWorkflowTasks(
 
 			// PartitionManager could have been unloaded; fetch the latest copy
 			pgMgr := s.getPhysicalTaskQueueManagerImpl(ptq)
-			s.LessOrEqual(int64(taskCount-tasksPolled), pgMgr.backlogMgr.TotalApproximateBacklogCount())
+			s.LessOrEqual(int64(taskCount-tasksPolled), totalApproximateBacklogCount(pgMgr.backlogMgr))
 		}
 	}
 }
@@ -3197,11 +3197,11 @@ func (s *matchingEngineSuite) addConsumeAllWorkflowTasksNonConcurrently(taskCoun
 
 	// Extract the pgMgr for validating approximateBacklogCounter
 	pgMgr := s.getPhysicalTaskQueueManagerImpl(ptq)
-	s.EqualValues(taskCount*numWorkers, pgMgr.backlogMgr.TotalApproximateBacklogCount())
+	s.EqualValues(taskCount*numWorkers, totalApproximateBacklogCount(pgMgr.backlogMgr))
 
 	s.pollWorkflowTasks(false, workflowType, numPollers, taskCount, ptq, taskQueue, nil)
 
-	s.LessOrEqual(int64(0), pgMgr.backlogMgr.TotalApproximateBacklogCount())
+	s.LessOrEqual(int64(0), totalApproximateBacklogCount(pgMgr.backlogMgr))
 }
 
 func (s *matchingEngineSuite) TestAddConsumeWorkflowTasksNoDBErrors() {
@@ -3256,7 +3256,7 @@ func (s *matchingEngineSuite) resetBacklogCounter(numWorkers int, taskCount int,
 	s.EqualValues(maxTaskId, pqMgr.backlogMgr.getDB().GetMaxReadLevel(0))
 
 	// validate the approximateBacklogCounter
-	s.EqualValues(taskCount*numWorkers, pqMgr.backlogMgr.TotalApproximateBacklogCount())
+	s.EqualValues(taskCount*numWorkers, totalApproximateBacklogCount(pqMgr.backlogMgr))
 
 	// Unload the PQM
 	s.matchingEngine.unloadTaskQueuePartition(partitionManager, unloadCauseForce)
@@ -3288,7 +3288,7 @@ func (s *matchingEngineSuite) resetBacklogCounter(numWorkers int, taskCount int,
 
 	s.EqualValues(0, s.taskManager.getTaskCount(ptq))
 	s.EventuallyWithT(func(collect *assert.CollectT) {
-		require.Equal(collect, int64(0), pqMgr.backlogMgr.TotalApproximateBacklogCount())
+		require.Equal(collect, int64(0), totalApproximateBacklogCount(pqMgr.backlogMgr))
 	}, 4*time.Second, 10*time.Millisecond, "backlog counter should have been reset")
 }
 
@@ -3349,7 +3349,7 @@ func (s *matchingEngineSuite) concurrentPublishAndConsumeValidateBacklogCounter(
 	wg.Wait()
 
 	ptqMgr := s.getPhysicalTaskQueueManagerImpl(ptq)
-	s.LessOrEqual(int64(s.taskManager.getTaskCount(ptq)), ptqMgr.backlogMgr.TotalApproximateBacklogCount())
+	s.LessOrEqual(int64(s.taskManager.getTaskCount(ptq)), totalApproximateBacklogCount(ptqMgr.backlogMgr))
 }
 
 func (s *matchingEngineSuite) TestConcurrentAddWorkflowTasksNoDBErrors() {
@@ -3407,14 +3407,14 @@ func (s *matchingEngineSuite) TestMultipleWorkersLesserNumberOfPollersThanTasksD
 	s.concurrentPublishAndConsumeValidateBacklogCounter(5, 500, 200)
 }
 
-func (s *matchingEngineSuite) TestLargerBacklogAge() {
+func (s *matchingEngineSuite) TestOldestBacklogAge() {
 	firstAge := durationpb.New(100 * time.Second)
 	secondAge := durationpb.New(1 * time.Millisecond)
-	s.Same(firstAge, largerBacklogAge(firstAge, secondAge))
+	s.Same(firstAge, oldestBacklogAge(firstAge, secondAge))
 
 	thirdAge := durationpb.New(5 * time.Minute)
-	s.Same(thirdAge, largerBacklogAge(firstAge, thirdAge))
-	s.Same(thirdAge, largerBacklogAge(secondAge, thirdAge))
+	s.Same(thirdAge, oldestBacklogAge(firstAge, thirdAge))
+	s.Same(thirdAge, oldestBacklogAge(secondAge, thirdAge))
 }
 
 func (s *matchingEngineSuite) TestCheckNexusEndpointsOwnership() {
--- a/service/matching/backlog_manager.go
+++ b/service/matching/backlog_manager.go
@@ -17,6 +17,7 @@ import (
 	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/persistence"
+	"google.golang.org/protobuf/types/known/durationpb"
 )
 
 var (
@@ -38,10 +39,7 @@ type (
 		// It's returned as a hint to the SDK to influence polling behavior (sticky vs normal).
 		BacklogCountHint() int64
 		BacklogStatus() *taskqueuepb.TaskQueueStatus
-		// TotalApproximateBacklogCount returns an estimate of the total size of the backlog in
-		// persistence. It may be off in either direction.
-		TotalApproximateBacklogCount() int64
-		BacklogHeadAge() time.Duration
+		BacklogStatsByPriority() map[int32]*taskqueuepb.TaskQueueStats
 		InternalStatus() []*taskqueuespb.InternalTaskQueueStatus
 
 		// TODO(pri): remove
@@ -175,17 +173,13 @@ func (c *backlogManagerImpl) BacklogCountHint() int64 {
 	return c.taskAckManager.getBacklogCountHint()
 }
 
-func (c *backlogManagerImpl) BacklogHeadAge() time.Duration {
-	return c.taskReader.getBacklogHeadAge()
-}
-
 func (c *backlogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
 	taskIDBlock := rangeIDToTaskIDBlock(c.db.RangeID(), c.config.RangeSize)
 	return &taskqueuepb.TaskQueueStatus{
 		ReadLevel: c.taskAckManager.getReadLevel(),
 		AckLevel:  c.taskAckManager.getAckLevel(),
-		// use getApproximateBacklogCount instead of BacklogCountHint since it's more accurate
-		BacklogCountHint: c.db.getApproximateBacklogCount(subqueueZero),
+		// use getTotalApproximateBacklogCount instead of BacklogCountHint since it's more accurate
+		BacklogCountHint: c.db.getTotalApproximateBacklogCount(),
 		TaskIdBlock: &taskqueuepb.TaskIdBlock{
 			StartId: taskIDBlock.start,
 			EndId:   taskIDBlock.end,
@@ -193,8 +187,14 @@ func (c *backlogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
 	}
 }
 
-func (c *backlogManagerImpl) TotalApproximateBacklogCount() int64 {
-	return c.db.getTotalApproximateBacklogCount()
+func (c *backlogManagerImpl) BacklogStatsByPriority() map[int32]*taskqueuepb.TaskQueueStats {
+	defaultPriority := defaultPriorityLevel(c.config.PriorityLevels())
+	return map[int32]*taskqueuepb.TaskQueueStats{
+		defaultPriority: &taskqueuepb.TaskQueueStats{
+			ApproximateBacklogCount: c.db.getTotalApproximateBacklogCount(),
+			ApproximateBacklogAge:   durationpb.New(c.taskReader.getBacklogHeadAge()),
+		},
+	}
 }
 
 func (c *backlogManagerImpl) InternalStatus() []*taskqueuespb.InternalTaskQueueStatus {
@@ -209,7 +209,7 @@ func (c *backlogManagerImpl) InternalStatus() []*taskqueuespb.InternalTaskQueueS
 			},
 			LoadedTasks:             c.taskAckManager.getBacklogCountHint(),
 			MaxReadLevel:            c.db.GetMaxReadLevel(subqueueZero),
-			ApproximateBacklogCount: c.db.getApproximateBacklogCount(subqueueZero),
+			ApproximateBacklogCount: c.db.getTotalApproximateBacklogCount(),
 		},
 	}
 }
--- a/service/matching/backlog_manager_test.go
+++ b/service/matching/backlog_manager_test.go
@@ -199,13 +199,13 @@ func (s *BacklogManagerTestSuite) TestApproximateBacklogCount_IncrementedByAppen
 		responseCh: make(chan<- error),
 	}
 
-	s.Equal(int64(0), blm.TotalApproximateBacklogCount())
+	s.Equal(int64(0), totalApproximateBacklogCount(blm))
 
 	blm.taskWriter.Start()
 	// Adding tasks to the buffer will increase the in-memory counter by 1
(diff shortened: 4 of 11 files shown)


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -590,4 +590,14 @@ message ListWorkersRequest {
 message ListWorkersResponse {
     repeated temporal.api.worker.v1.WorkerInfo workers_info = 1;
     bytes next_page_token = 2;
-}
\ No newline at end of file
+}
+
+message ConfigureTaskQueueRequest {
```
dnr: oh, I meant that rename comment as a joke, I don't think we should rename it, it's better (more readable, less confusing) to be consistent with the name in workflowservice.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -116,6 +116,9 @@ message DeploymentData {
 // Container for all persistent user data that varies per task queue type within a family.
 message TaskQueueTypeUserData {
     DeploymentData deployment_data = 1;
+
+    // Perist TaskQueueConfig that comes as part of the UpdateTaskQueueConfigApi.
+    TaskQueueConfig task_queue_config = 2;
```
dnr: let's call it `config`. we don't need to type TaskQueue 1000 times

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -142,3 +145,32 @@ message VersionedTaskQueueUserData {
     TaskQueueUserData data = 1;
     int64 version = 2;
 }
+
+message RateLimit {
```
dnr: I'm having second thoughts here... duplicating all of this structure seems like a waste of effort and opportunity for bugs. as you said, we have an HLC value available at the top level, and that's the one we're using to merge anyway... what do you think about basically reusing the public api message types here?

sivagirish81: Yeah that's what I was leaning towards initially as well. Avoid duplication. If we do not see any use-case where we would like to enforce more fine grained LWW in terms of specific task queue configs - we should be fine using the public api fields here and relying on the HLC at top level should suffice.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
sivagirish81: Updated.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
sivagirish81: Yeah. I really like that we have the linter enforced and it did make sense to rename it. Considering it adheres to the linter rules and as far as the matching service is concerned it is still relatable. I don't mind sticking to this but yeah the UpdateTaskQueueConfigRequest does match exactly with the request in the workflowservice and that's the convention we follow right to keep it identical to whatever is being sent from the workflowservice.

sivagirish81: Reverted it to UpdateTaskQueueConfig.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
sivagirish81: Updated it to reuse the public api message types here.

common/api/metadata.go
```diff
@@ -157,6 +157,7 @@ var (
 		"TriggerWorkflowRule":                   {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
 		"RecordWorkerHeartbeat":                 {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
 		"ListWorkers":                           {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
+		"UpdateTaskQueueConfig":                 {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
```
dnr: this is a write api, not readonly

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -590,4 +590,20 @@ message ListWorkersRequest {
...
+
+// (-- api-linter: core::0134::request-resource-required=disabled
+//     aip.dev/not-precedent: UpdateTaskQueueConfigRequest RPC doesn't follow Google API format. --)
+// (-- api-linter: core::0134::request-mask-required=disabled
+//     aip.dev/not-precedent: UpdateTaskQueueConfigRequest RPC doesn't follow Google API format. --)
+// (-- api-linter: core::0134::method-signature=disabled
+//     aip.dev/not-precedent: UpdateTaskQueueConfigRequest RPC doesn't follow Google API format. --)
+message UpdateTaskQueueConfigRequest {
+    string namespace_id = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
```
dnr: isn't this in the request from workflowservice? why do you need it here again?

service/matching/matching_engine.go
```diff
@@ -1352,6 +1352,18 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		}
 	}
 
+	if req.GetReportConfig() {
+		userData, _, err := pm.GetUserDataManager().GetUserData()
+		if err != nil {
+			return nil, err
+		}
+		if userData == nil || userData.GetData() == nil {
+			return descrResp, nil
```
dnr: I don't think we should do an early return here. It's too easy for someone to add another `if req.ReportXX` clause after this one and miss this early return. Let's have all the optional ReportX just add fields if they're set and leave the only non-error return for the end.

service/matching/matching_engine.go
```diff
@@ -1352,6 +1352,18 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		}
 	}
 
+	if req.GetReportConfig() {
+		userData, _, err := pm.GetUserDataManager().GetUserData()
+		if err != nil {
+			return nil, err
+		}
+		if userData == nil || userData.GetData() == nil {
```
dnr: The GetX() wrappers check for nil receivers, so you can just write `if userData.GetData() == nil`

service/matching/matching_engine.go
```diff
@@ -2928,3 +2940,174 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+	if data.PerType[tqType] == nil {
+		data.PerType[tqType] = &persistencespb.TaskQueueTypeUserData{}
+	}
+	if data.PerType[tqType].Config == nil {
+		data.PerType[tqType].Config = &taskqueuepb.TaskQueueConfig{}
+	}
+	return data, tqType
+}
+
+func constructUpdateTaskQueueConfigResponse(
```
dnr: why are we doing so much copying in this function? isn't the big benefit of using the same proto message in persistence that we can just return it as-is?

service/matching/matching_engine.go
```diff
@@ -2928,3 +2940,174 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+	}
+	return &taskqueuepb.RateLimitConfig{
+		RateLimit: &taskqueuepb.RateLimit{
+			RequestsPerSecond: update.GetRateLimit().GetRequestsPerSecond(),
+		},
+		Metadata: &taskqueuepb.ConfigMetadata{
+			Reason:     update.GetReason(),
+			UpdateTime: updateTime,
+		},
+	}
```
dnr: write this with less repetition:

```suggestion
	var rateLimit *taskqueuepb.RateLimitConfig
	if r := update.GetRateLimit(); r != nil {
		rateLimit = &taskqueuepb.RateLimit{RequestsPerSecond: r.RequestsPerSecond}
	}
	return &taskqueuepb.RateLimitConfig{
		RateLimit: rateLimit,
		Metadata: &taskqueuepb.ConfigMetadata{
			Reason:     update.GetReason(),
			UpdateTime: updateTime,
		},
	}
```

service/matching/matching_engine.go
```diff
@@ -2928,3 +2940,174 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+}
+
+func (e *matchingEngineImpl) UpdateTaskqueueConfig(ctx context.Context, request *matchingservice.UpdateTaskQueueConfigRequest) (*matchingservice.UpdateTaskQueueConfigResponse, error) {
+	taskQueueFamily, err := tqid.NewTaskQueueFamily(request.NamespaceId, request.UpdateTaskqueueConfig.GetTaskQueueName())
+	if err != nil {
+		return nil, err
+	}
+	if request.GetUpdateTaskqueueConfig() == nil {
+		return &matchingservice.UpdateTaskQueueConfigResponse{}, err
```
dnr: shouldn't we return the current config here? also write nil instead of err so it's clear this is a success

service/matching/matching_engine.go
```diff
@@ -2928,3 +2940,174 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+		data.PerType = make(map[int32]*persistencespb.TaskQueueTypeUserData)
+	}
+	tqType := int32(taskQueueType)
+	if data.PerType[tqType] == nil {
+		data.PerType[tqType] = &persistencespb.TaskQueueTypeUserData{}
+	}
+	if data.PerType[tqType].Config == nil {
+		data.PerType[tqType].Config = &taskqueuepb.TaskQueueConfig{}
+	}
+	return data, tqType
```
dnr: why return tqtype? it's the same as the input

service/matching/matching_engine.go
```diff
@@ -2928,3 +2940,174 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+			protoTs := hlc.ProtoTimestamp(now)
+			// Update relevant config fields
+			cfg := data.PerType[tqType].Config
+			// Queue Rate Limit
+			updateTaskQueueConfig := request.GetUpdateTaskqueueConfig()
+			queueRateLimit := updateTaskQueueConfig.GetUpdateQueueRateLimit()
+			if queueRateLimit != nil {
+				cfg.QueueRateLimit = buildRateLimitConfig(queueRateLimit, protoTs)
+			} else {
+				cfg.QueueRateLimit = nil
```
dnr: I thought we were going to set a taskqueuepb.RateLimitConfig always, so that we have metadata, even if the config is "no rate limit"?

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
sivagirish81: @dnr : The UpdateTaskQueueConfigRequest operates on a TaskQueue so you had previously mentioned that these requests need to be redirected to the root WF partition in the group for UpdateWorkerVersioningRulesRequest.
So over there the namespace is required at a depth of 1 and specifically a TaskQueue is required at a depth of upto 2 in the request. So that's why I added these extra fields even though I am not using it to ensure it routes to the right WF partition. 
But now when I think about it I am not sure if that is still the right place considering that it also enforces task queue type as workflow.

common/api/metadata.go
sivagirish81: Updated it to AccessWrite.

service/matching/matching_engine.go
sivagirish81: Noted.

service/matching/matching_engine.go
sivagirish81: It's unnecessary. Removed it.

service/matching/matching_engine.go
sivagirish81: Made the changes to ensure that the protos are directly returned from persistence.

service/matching/matching_engine.go
sivagirish81: @dnr : The reason why I chose to not return anything is to signal to the user that no update has happened. Will it be misleading if we send the current config?

Additionally : On this use-case do we return the current config or do we set all the taskqueue configs to nil and revert to worker defaults or dynamic configs. We are receiving an api call from the user with all nil fields? But the problem is there is no taskqueue type so we will need to iterate over the per type map to do this and set all of them to nil. What should be the ideal flow here?

service/matching/matching_engine.go
sivagirish81: Correct. if RateLimitUpdate is present w.r.t to QueueRateLimit or FairnessKeyRateLimitDefault then we apply the update otherwise we retain the existing config and do not apply any updates.
We unset basis the ratelimit field inside RateLimitUpdate if absent unset otherwise set.
Made the changes to reflect this behaviour.

service/matching/matching_engine.go
sivagirish81: Agreed.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
sivagirish81: Made the changes to remove task_queue from the UpdateTaskQueueConfigRequest and updated the api repo to store the TaskQueueName under TaskQueue as a string to ensure routing magic is able to pick it up automatically.

service/matching/matching_engine.go
sivagirish81: Returned the current config in this case.

service/matching/matching_engine.go
```diff
@@ -2928,3 +2937,125 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+	return data
+}
+
+func constructUpdateTaskQueueConfigResponse(
+	taskQueueUserData *persistencespb.TaskQueueUserData,
+	taskQueueType enumspb.TaskQueueType,
+) *matchingservice.UpdateTaskQueueConfigResponse {
+	if taskQueueUserData.GetPerType() == nil {
+		return &matchingservice.UpdateTaskQueueConfigResponse{}
+	}
```
dnr: you can leave out this part, looking up something in a nil map returns "zero, false" so the next case handles it

service/matching/matching_engine.go
```diff
@@ -2928,3 +2937,125 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+	queueRateLimit := updateTaskQueueConfig.GetQueueRateLimit()
+	fairnessKeysRateLimitDefault := updateTaskQueueConfig.GetFairnessKeysRateLimitDefault()
+	return &matchingservice.UpdateTaskQueueConfigResponse{
+		UpdatedTaskqueueConfig: &taskqueuepb.TaskQueueConfig{
+			QueueRateLimit:               queueRateLimit,
+			FairnessKeysRateLimitDefault: fairnessKeysRateLimitDefault,
+		},
+	}
```
dnr: it still looks like we're doing too much work here, can't we simplify the whole thing to this?

```suggestion
	perTypeUserData := taskQueueUserData.GetPerType()[int32(taskQueueType)]
	return &matchingservice.UpdateTaskQueueConfigResponse{
		UpdatedTaskqueueConfig: perTypeUserData.GetConfig(),
	}
```

dnr: but also, you have two calls to this function, and one of them just pulls out the config itself and throws away the UpdateTaskQueueConfigResponse wrapper. so I think the function should just return the config and the one caller that needs to wrap it in another message can do that itself.

service/matching/matching_engine.go
```diff
@@ -2928,3 +2937,110 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+
+func (e *matchingEngineImpl) UpdateTaskqueueConfig(ctx context.Context, request *matchingservice.UpdateTaskQueueConfigRequest) (*matchingservice.UpdateTaskQueueConfigResponse, error) {
+	taskQueueFamily, err := tqid.NewTaskQueueFamily(request.NamespaceId, request.UpdateTaskqueueConfig.GetTaskQueue())
+	if err != nil {
+		return nil, err
+	}
+
+	taskQueueType := request.UpdateTaskqueueConfig.GetTaskQueueType()
+	tqm, _, err := e.getTaskQueuePartitionManager(ctx,
+		taskQueueFamily.TaskQueue(enumspb.TASK_QUEUE_TYPE_WORKFLOW).RootPartition(),
```
sivagirish81, line 2973: Forcing the taskqueuetype to `TASK_QUEUE_TYPE_WORKFLOW` to override api set task queue types to ensure only a root Workflow task partition is always fetched for updating taskqueue config data.

service/matching/matching_engine_test.go
```diff
@@ -3510,6 +3512,50 @@ func (s *matchingEngineSuite) TestPollWorkflowTaskQueueWithRateLimiterError() {
 	s.ErrorIs(err, rateLimiterErr)
 }
 
+// TestUpdateTaskqueueConfig tests the UpdateTaskQueueConfig Api.
```
sivagirish81: Kept the test simple just to check if the value is being persisted in the UpdateTaskQueueConfig api.
Made sure that test is happening for activities as this api at least initially will target updating the ratelimit only for Activity task queues not for workflow task queues - inline with the current behavior of only allowing setting of rate limits against activities via workerOptions.

service/matching/matching_engine.go
sivagirish81: Changed the function to return the config only as it is easier to understand as a function and wrapped the response around the UpdateTaskQueueConfigResponse wherever needed.

proto/internal/temporal/server/api/matchingservice/v1/service.proto
```diff
@@ -176,4 +176,16 @@ service MatchingService {
     // Supports pagination for large result sets. Returns an empty list if no workers match the criteria.
     // Returns an error if the namespace doesn't exist.
     rpc ListWorkers (ListWorkersRequest) returns (ListWorkersResponse) {}
+
+    // UpdateTaskQueueConfig recieves TaskQueue configurations from the frontend call to UpdateTaskQueueConfigApi.
+    // This API is used to persist TaskQueue configurations in the persistence layer.
+    // The TaskQueue configuration in the persistence layer supersedes the in-memory TaskQueue configuration,
+    // as well as the system dynamic configuration.
```
dnr: This sentence "This API is used to persist configuration in the persistence layer" is misleading: it's used to set the configuration, which is persisted, yeah, but more importantly it actually takes effect.

Maybe say something like "This API is used to set the persisted task queue configuration."

Also "supersedes the in-memory TaskQueue configuration" doesn't quite make sense: there is no "in-memory configuration" in general, there's only a rate limit. We could say "The rate limit in the task queue configuration supersedes the rate limit in worker configuration" but I also don't think we need to say anything about that here at all.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -116,6 +116,9 @@ message DeploymentData {
 // Container for all persistent user data that varies per task queue type within a family.
 message TaskQueueTypeUserData {
     DeploymentData deployment_data = 1;
+
+    // Perist TaskQueueConfig that comes as part of the UpdateTaskQueueConfigApi.
```
dnr: typos:
```suggestion
    // Persist TaskQueueConfig that comes as part of the UpdateTaskQueueConfig api.
```

but I don't think this comment really adds anything, the comment on the message says it's for all persisted user data, and it's obvious a TaskQueueConfig

service/frontend/workflow_handler.go
```diff
@@ -6052,3 +6052,27 @@ func (wh *WorkflowHandler) ListWorkers(
 		NextPageToken: resp.GetNextPageToken(),
 	}, nil
 }
+
+// UpdateTaskQueueConfig : updates the indivudal task queue rate limit or fairnessKey Rate limit
```
dnr: I would just leave this out too.. we're inevitably going to forget to update it when we add more fields and then it'll just be confusing

service/matching/matching_engine.go
```diff
@@ -2928,3 +2936,112 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+}
+
+func getTaskQueueConfig(
+	taskQueueUserData *persistencespb.TaskQueueUserData,
+	taskQueueType enumspb.TaskQueueType,
+) *taskqueuepb.TaskQueueConfig {
+	return taskQueueUserData.GetPerType()[int32(taskQueueType)].GetConfig()
+}
+
+func (e *matchingEngineImpl) UpdateTaskqueueConfig(ctx context.Context, request *matchingservice.UpdateTaskQueueConfigRequest) (*matchingservice.UpdateTaskQueueConfigResponse, error) {
```
dnr: names should match

```suggestion
func (e *matchingEngineImpl) UpdateTaskQueueConfig(ctx context.Context, request *matchingservice.UpdateTaskQueueConfigRequest) (*matchingservice.UpdateTaskQueueConfigResponse, error) {
```

service/matching/matching_engine.go
```diff
@@ -2928,3 +2936,112 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+		data.PerType[tqType].Config = &taskqueuepb.TaskQueueConfig{}
+	}
+	return data
+}
+
+func getTaskQueueConfig(
+	taskQueueUserData *persistencespb.TaskQueueUserData,
+	taskQueueType enumspb.TaskQueueType,
+) *taskqueuepb.TaskQueueConfig {
+	return taskQueueUserData.GetPerType()[int32(taskQueueType)].GetConfig()
```
dnr: let's just inline this? it's basically just a map lookup. or at least move the GetData() call into here also 🤷

service/matching/matching_engine.go
```diff
@@ -2928,3 +2936,112 @@ func largerBacklogAge(rootBacklogAge *durationpb.Duration, currentPartitionAge *
...
+			now := hlc.Next(existingClock, e.timeSource)
+			protoTs := hlc.ProtoTimestamp(now)
+			// Update relevant config fields
+			cfg := data.PerType[int32(taskQueueType)].Config
+			updateTaskQueueConfig := request.GetUpdateTaskqueueConfig()
+			// Queue Rate Limit
+			queueRateLimit := updateTaskQueueConfig.GetUpdateQueueRateLimit()
+			if queueRateLimit != nil {
+				cfg.QueueRateLimit = buildRateLimitConfig(queueRateLimit, protoTs)
+			}
```
dnr: small suggestion: for values used if not-nil, declare and check in the same statement:
```suggestion
			if rl := updateTaskQueueConfig.GetUpdateQueueRateLimit(); rl != nil {
				cfg.QueueRateLimit = buildRateLimitConfig(rl, protoTs)
			}
```
and next one too

service/matching/matching_engine_test.go
dnr: This isn't testing the frontend or DescribeTaskQueue (though I admit those are pretty simple). I'd rather see a functional test (the ones in ./tests that start up a test server) that makes an UpdateTaskQueueConfig call and then a DescribeTaskQueue call through the frontend, just to make sure it's all covered (although the functional tests currently don't test routing).

proto/internal/temporal/server/api/matchingservice/v1/service.proto
sivagirish81: Updated to : Set the persisted task queue configuration.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
sivagirish81: Removed as suggested.

service/frontend/workflow_handler.go
sivagirish81: Removed.

service/matching/matching_engine.go
sivagirish81: Updated the name to be consistent across matching and frontend service.

service/matching/matching_engine.go
sivagirish81: Inlined.

service/matching/matching_engine.go
sivagirish81: Implemented for both QueueRateLimit and FairnessKeyRateLimit.

service/matching/matching_engine_test.go
sivagirish81: Wrote a functional test in the tests package to call the UpdateTaskQueueConfig api on an activity task queue and subsequently used describeTaskQueue to describe the task queue and check if the responses match.

service/frontend/workflow_handler.go
```diff
@@ -6052,3 +6052,26 @@ func (wh *WorkflowHandler) ListWorkers(
...
+) (*workflowservice.UpdateTaskQueueConfigResponse, error) {
+	if request == nil {
+		return nil, errRequestNotSet
+	}
+	namespaceName := namespace.Name(request.GetNamespace())
+	namespaceID, err := wh.namespaceRegistry.GetNamespaceID(namespaceName)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := wh.matchingClient.UpdateTaskQueueConfig(ctx, &matchingservice.UpdateTaskQueueConfigRequest{
```
dnr, line 6067: We should do some more validation here before we hand it off:
- prohibit setting rate limit on workflow task queue (at least until we have a distributed rate limiter that can handle sticky queues)
- prohibit negative values
- check maximum length of `reason` fields and anything else that we're going to store in metadata

But let's do them in a follow-up PR.

tests/task_queue_test.go
```diff
@@ -162,3 +163,56 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	describeResp, err := s.FrontendClient().DescribeTaskQueue(testcore.NewContext(), describeReq)
+	s.NoError(err)
+	s.NotNil(describeResp)
+	s.NotNil(describeResp.Config)
+	s.NotNil(describeResp.Config.QueueRateLimit)
+	s.Equal(updateRPS, describeResp.Config.QueueRateLimit.RateLimit.RequestsPerSecond)
+	s.Equal(updateReason, describeResp.Config.QueueRateLimit.Metadata.Reason)
+	s.Equal(updateRPS, describeResp.Config.FairnessKeysRateLimitDefault.RateLimit.RequestsPerSecond)
+	s.Equal(updateReason, describeResp.Config.FairnessKeysRateLimitDefault.Metadata.Reason)
+}
```
dnr, line 218: We should also have a test where we remove a rate limit that was set. Can do later though.

dnr (approved):

PR #8007, opened 2025-07-03
Fix ack level sync on unload for priBacklogManager

dnr:
## What changed?

Classic backlog manager synced ack level from taskReader on unload, not just the copy in taskQueueDB. This fixes priBacklogManager to sync from priTaskReader also.

Also be more careful about updating lastChange in updateAckLevelAndBacklogStats only when values change. This lets us be more accurate about avoiding the write on unload if it's not necessary.

## Why?

This allows a jump in ack level after a gap to be synced on unload instead of having to wait for next time.

This fixes (and reenables) the "cycling behavior" test (which shouldn't have been disabled).

## How did you test it?
- [x] covered by existing tests (newly enabled)

### DIFF
--- a/service/matching/pri_backlog_manager.go
+++ b/service/matching/pri_backlog_manager.go
@@ -120,11 +120,21 @@ func (c *priBacklogManagerImpl) Stop() {
 	// initialized. Also skip if we're stopping due to lost ownership (the update will
 	// fail in that case). Ignore any errors. Don't bother with GC, the next reload will
 	// handle that.
-	if c.initializedError.Ready() && !c.skipFinalUpdate.Load() {
-		ctx, cancel := context.WithTimeout(c.tqCtx, ioTimeout)
-		_ = c.db.SyncState(ctx)
-		cancel()
+	if !c.initializedError.Ready() || c.skipFinalUpdate.Load() {
+		return
 	}
+
+	c.subqueueLock.Lock()
+	for i, r := range c.subqueues {
+		_, ackLevel := r.getLevels()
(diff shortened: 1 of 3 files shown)


stephanos (approved):

service/matching/pri_backlog_manager.go
```diff
@@ -120,11 +120,21 @@ func (c *priBacklogManagerImpl) Stop() {
...
-		_ = c.db.SyncState(ctx)
-		cancel()
+	if !c.initializedError.Ready() || c.skipFinalUpdate.Load() {
+		return
 	}
+
+	c.subqueueLock.Lock()
+	for i, r := range c.subqueues {
+		_, ackLevel := r.getLevels()
+		// oldestTime can be time.Time{} here since countDelta is 0
```
stephanos, line 130: nit: the comment is a little worrying since the behavior of `updateAckLevelAndBacklogStats` could change later. I'd sleep better knowing that the default time value would always be ignored regardless of countDelta.

dnr: True, but I didn't want to add a whole new function just to update ackLevel alone. Would you prefer it that way?

stephanos: I thought in `updateAckLevelAndBacklogStats` it could check for the default value instead. So you could pass in the default value safely for any of the two parameters.

dnr: Isn't that also part of `updateAckLevelAndBacklogStats` behavior that could change later? As the caller the only way to be confident seems to be another function that doesn't even have those parameters

PR #8010, opened 2025-07-04
tdbg for list fairness tasks

stephanos:
## What changed?

Added new flag to tdbg's `list-tasks` command to query the fairness table instead.

## Why?

Allow operation on fairness table, too.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

```
./tdbg taskqueue list-tasks --task-queue-type TASK_QUEUE_TYPE_WORKFLOW --task-queue test --min-pass 1
```

It's late ... I didn't manage to see a result, though. But it didn't fail and I saw the query being executed in the debugger.


common/persistence/cassandra/matching_task_store_v2.go
```diff
@@ -288,7 +287,7 @@ func (d *matchingTaskStoreV2) GetTasks(
 	request *p.GetTasksRequest,
 ) (*p.InternalGetTasksResponse, error) {
 	// Require starting from pass 1.
-	if request.InclusiveMinPass < 1 || request.ExclusiveMaxTaskID != math.MaxInt64 {
```
stephanos, line 291: I tight be too tired, but I don't understand this check and how that fits into querying the tasks here.

schema/cassandra/version.go
```diff
@@ -3,4 +3,4 @@ package cassandra
 // NOTE: whenever there is a new database schema update, plz update the following versions
 
 // Version is the Cassandra database release version
-const Version = "1.12"
+const Version = "1.13"
```
stephanos, line 6: Might as well add the schema change in this PR now.

common/persistence/cassandra/matching_task_store_v2.go
dnr: I'd prefer to leave it. It's a check against misuse of the API: using a fairness request with a classic task store or vice versa, because we conflated both into the same TaskManager interface. They're really two different interfaces that look quite similar. The proper way to do it would be to duplicate the interface and all the request/response structs so we don't even have a ExclusiveMaxTaskID here. But we can do that later. For now I just want errors on misuse.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -382,6 +382,7 @@ message GetTaskQueueTasksRequest {
   int32 batch_size = 6;
   bytes next_page_token = 7;
   int32 subqueue = 8;
+  int64 min_pass = 9;
```
dnr: put this field on line 380 above min_task_id, so we always have <pass, id> next to each other and in that order wherever they appear

service/frontend/admin_handler_test.go
```diff
@@ -153,6 +153,7 @@ func (s *adminHandlerSuite) SetupTest() {
 		s.mockVisibilityMgr,
 		s.mockResource.GetLogger(),
 		s.mockResource.GetTaskManager(),
+		s.mockResource.GetTaskManager(),
```
dnr, line 156: :thinking:  maybe even mockResource should have a FairTaskManager? what implementations do these correspond to, anyway?

tools/tdbg/tdbg_commands.go
```diff
@@ -500,6 +500,10 @@ func newAdminTaskQueueCommands(clientFactory ClientFactory) []*cli.Command {
 					Name:  FlagPrintJSON,
 					Usage: "Print in raw json format",
 				},
+				&cli.Int64Flag{
+					Name:  FlagMinPass,
```
dnr, line 508: This is not really discoverable or understandable unless you already know how everything works... I'd rather see something like `tdbg list-fair-tasks` or at least `tdbg list-tasks --fair` or `tdbg list-tasks --v2`. Those should start from pass 1 by default but there can be a flag to override.

common/persistence/cassandra/matching_task_store_v2.go
stephanos: 👍 I can revert the change and add a comment to explain that here.

stephanos: Actually, why not check for `!= 0` instead?

tools/tdbg/tdbg_commands.go
stephanos: Right; I was going to have a convo around that to figure out the level of discoverability. I like the default of pass 1.

common/persistence/cassandra/matching_task_store_v2.go
dnr: Because of this: https://github.com/temporalio/temporal/blob/main/common/persistence/task_manager.go#L202-L204
I didn't want to mess with that just for this purpose. But we could do it that way with more effort

dnr (approved):
seems fine to me. maybe do a quick manual test if you didn't already?

stephanos:
👍 I'm seeing two tasks in Cass and two tasks in tdbg when running `./tdbg taskqueue list-tasks --fair --task-queue-type TASK_QUEUE_TYPE_WORKFLOW --task-queue test`

dnr:
> 👍 I'm seeing two tasks in Cass and two tasks in tdbg

test more than the page size to check that pagination works?

stephanos:
The CLI or the persistence pagination? Both were copy/pasted or re-used.

stephanos:
https://github.com/temporalio/temporal/pull/8010/commits/ebdaca915c9440f495921ccd28cb9a6ef6703f4c fixes multiple pagination related issues: creates new context per request, uses documented default page size of 10, and actually extracts next page token.

PR #8012, opened 2025-07-04
Fix TestChildWorkflowInheritance_CrossTQ_NoInherit flake

carlydf:
## What changed?
Include "new current version does not have task queues" as a retryable error in the SetCurrentVersion eventually.

## Why?
Because eventually those task queues will arrive (in all tests that use that helper), so this fixes the flake

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -1626,6 +1626,13 @@ func (s *Versioning3Suite) testChildWorkflowInheritance_ExpectNoInherit(crossTq 
 	s.WaitForChannel(ctx, wfStarted)
 	close(wfStarted)
 
+	// wait for pollers to register their task queues in build id 2
+	if crossTq {
+		s.waitForTaskQueueInVersion(tv2Child)
+	} else {
+		s.waitForTaskQueueInVersion(tv2)
+	}
+
 	// make v2 current for both parent and child and unblock the wf to start the child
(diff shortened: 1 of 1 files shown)


tests/versioning_3_test.go
```diff
@@ -1968,7 +1969,7 @@ func (s *Versioning3Suite) setCurrentDeployment(tv *testvars.TestVars) {
 		}
 		_, err := s.FrontendClient().SetWorkerDeploymentCurrentVersion(ctx, req)
 		var notFound *serviceerror.NotFound
-		if errors.As(err, &notFound) {
+		if errors.As(err, &notFound) || (err != nil && err.Error() == workerdeployment.ErrCurrentVersionDoesNotHaveAllTaskQueues) {
```
Shivs11:
```suggestion
		if errors.As(err, &notFound) || errors.Is(err, serviceerror.NewFailedPrecondition(workerdeployment.ErrCurrentVersionDoesNotHaveAllTaskQueues)) {
			return false
		}
```

nit: maybe this looks a little better given that this is a failedPrecondition error

Shivs11 (approved):
approving this since I do think it's a good catch and the comment i have left is a nit given that this is a test

tests/versioning_3_test.go
carlydf: thanks!

PR #8014, opened 2025-07-04
Enable RateLimitedScheduler unconditionally

prathyushpv:
## Summary
- always build RateLimitedScheduler for history queues
- use a noop SchedulerRateLimiter when TaskSchedulerEnableRateLimiter is disabled

## Testing
- `make unit-test` *(fails: go: -race requires cgo; enable cgo by setting CGO_ENABLED=1)*

------
https://chatgpt.com/codex/tasks/task_b_68681cd1002c8330a364e9ce815cad82

### DIFF
--- a/service/history/archival_queue_factory.go
+++ b/service/history/archival_queue_factory.go
@@ -5,6 +5,7 @@ import (
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/quotas"
 	ctasks "go.temporal.io/server/common/tasks"
 	"go.temporal.io/server/common/telemetry"
 	"go.temporal.io/server/service/history/archival"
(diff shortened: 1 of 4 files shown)


common/tasks/rate_limited_scheduler.go
```diff
@@ -79,6 +81,9 @@ func (s *RateLimitedScheduler[T]) Stop() {
 }
 
 func (s *RateLimitedScheduler[T]) wait(task T) {
+	if !s.options.Enabled() {
```
yycptt, line 84: I tend to avoid evaluating dynamic config in hot paths. But I know David made some improvements there so don't know what's the story right now.

prathyushpv: Thanks!

service/history/replication/fx.go
```diff
@@ -273,7 +273,8 @@ func replicationStreamLowPrioritySchedulerProvider(
 		taskQuotaRequestFn,
 		taskMetricsTagsFn,
 		ctasks.RateLimitedSchedulerOptions{
-			EnableShadowMode: config.ReplicationEnableRateLimit(),
+			Enabled:          config.ReplicationEnableRateLimit,
+			EnableShadowMode: config.ReplicationEnableRateLimitShadowMode,
```
yux0, line 277: Thank you for the fix. LGTM

PR #8016, opened 2025-07-05
testTaskManager and matching engine test improvements

dnr:
## What changed?
- Improve testTaskManager (in-memory matching persistence implementation used by unit tests) to behave closer to a real persistence implementation.
  - The main change is that task queue metadata is now stored as-is.
  - One spot where testTaskManager was messing with metadata directly is removed.
  - Add a flag for updateMetadataOnCreateTasks, default true (behave like cassandra).
- Improve fault injection to be more flexible.
- Various small unit test improvements mostly in matchingEngineSuite.
- Disable a few of the reset backlog count tests for priority backlog manager.

## Why?
- The incomplete behavior of testTaskManager was covering up some bugs (e.g. #8015), though some of it cancelled out the other parts.
- Fault injection can be used in future tests.
- The new updateMetadataOnCreateTasks flag can be used in future tests to get better coverage for SQL persistence.

## How did you test it?
Improved and ran unit tests.

### DIFF
--- a/service/matching/matching_engine_test.go
+++ b/service/matching/matching_engine_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -328,7 +329,7 @@ func (s *matchingEngineSuite) PollForTasksEmptyResultTest(callContext context.Co
 		s.NotEmpty(descResp.DescResponse.Pollers[0].GetLastAccessTime())
 		s.Nil(descResp.DescResponse.GetTaskQueueStatus())
 	}
-	s.EqualValues(1, s.taskManager.getQueueManagerByKey(tlID).RangeID())
+	s.EqualValues(1, s.taskManager.getQueueDataByKey(tlID).RangeID())
 }
 
 func (s *matchingEngineSuite) TestOnlyUnloadMatchingInstance() {
@@ -602,7 +603,7 @@ func (s *matchingEngineSuite) TestPollActivityTaskQueues_InternalError() {
 	// add an activity task
 	_, _, err := s.matchingEngine.AddActivityTask(context.Background(), &addRequest)
 	s.NoError(err)
-	s.EqualValues(s.taskManager.getTaskCount(newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)), 1)
+	s.EqualValues(1, s.taskManager.getTaskCount(newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)))
 
 	// task is dropped with no retry; RecordActivityTaskStarted should only be called once
 	s.mockHistoryClient.EXPECT().RecordActivityTaskStarted(gomock.Any(), gomock.Any(), gomock.Any()).
@@ -614,7 +615,7 @@ func (s *matchingEngineSuite) TestPollActivityTaskQueues_InternalError() {
 			Identity:  "identity",
 		},
 	}, metrics.NoopMetricsHandler)
-	s.EqualValues(emptyPollActivityTaskQueueResponse, resp)
+	protoassert.ProtoEqual(s.T(), emptyPollActivityTaskQueueResponse, resp)
 	s.NoError(err)
 }
 
@@ -636,7 +637,7 @@ func (s *matchingEngineSuite) TestPollActivityTaskQueues_DataLossError() {
 	// add an activity task
 	_, _, err := s.matchingEngine.AddActivityTask(context.Background(), &addRequest)
 	s.NoError(err)
-	s.EqualValues(s.taskManager.getTaskCount(newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)), 1)
+	s.EqualValues(1, s.taskManager.getTaskCount(newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)))
 
 	// task is dropped with no retry; RecordActivityTaskStarted should only be called once
 	s.mockHistoryClient.EXPECT().RecordActivityTaskStarted(gomock.Any(), gomock.Any(), gomock.Any()).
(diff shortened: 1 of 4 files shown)


service/matching/matching_engine_test.go
```diff
@@ -3657,36 +3637,46 @@ func newTestTaskQueue(namespaceId string, name string, taskType enumspb.TaskQueu
...
+	return q.rangeID
 }
 
 func (m *testTaskManager) CreateTaskQueue(
 	_ context.Context,
 	request *persistence.CreateTaskQueueRequest,
 ) (*persistence.CreateTaskQueueResponse, error) {
 	tli := request.TaskQueueInfo
-	tlm := m.getQueueManager(tli.Name, tli.NamespaceId, tli.TaskType)
+	tlm := m.getQueueData(tli.Name, tli.NamespaceId, tli.TaskType)
```
ychebotarev, line 3679: nit: should be tqd, here and in other places.

service/matching/matching_engine_test.go
```diff
@@ -3706,51 +3695,42 @@ func (m *testTaskManager) UpdateTaskQueue(
...
-			TaskType:                request.TaskType,
-			Kind:                    enumspb.TASK_QUEUE_KIND_NORMAL,
-			AckLevel:                tlm.ackLevel,
-			ExpiryTime:              nil,
-			LastUpdateTime:          timestamp.TimeNowPtrUtc(),
-			ApproximateBacklogCount: tlm.ApproximateBacklogCount,
-		},
-		RangeID: tlm.rangeID,
+		RangeID:       tlm.rangeID,
+		TaskQueueInfo: common.CloneProto(tlm.info),
```
ychebotarev, line 3727: code in the previous version explicitly set some fields, like "AckLevel: tlm.ackLevel,"
Also "LastUpdateTime: timestamp.TimeNowPtrUtc(),"

Now you just copy whatever you have in tlm.info. Intended?

service/matching/matching_engine_test.go
```diff
@@ -3826,22 +3804,26 @@ func (m *testTaskManager) CreateTasks(
...
 	}
 
-	tlm := m.getQueueManager(taskQueue, namespaceId, taskType)
+	tlm := m.getQueueData(taskQueue, namespaceId, taskType)
 	tlm.Lock()
 	defer tlm.Unlock()
 
+	if tlm.rangeID != rangeID {
+		return nil, &persistence.ConditionFailedError{
+			Msg: fmt.Sprintf("Failed to create task. TaskQueue: %v, taskQueueType: %v, rangeID: %v, db rangeID: %v",
```
ychebotarev: the error is the same as for other cases. I"m not sure how it used, but I would add something that will distinguish it from other errors.
Like "Failed to create task - range mismatch."

service/matching/matching_engine_test.go
dnr: yes but I'm worried about merge conflicts. I probably shouldn't even have renamed getQueueManager but that name bothered me so much. let's clean it all up after merging the fairness branch?

service/matching/matching_engine_test.go
dnr: yes, this is the entire point of this PR: testTaskManager is a database. it absolutely should not look into the blob it's given, it should store it and return it.

service/matching/matching_engine_test.go
dnr: The text doesn't really matter here, but I cleaned it up: I removed the same check from inside the loop and changed the messages to be more clear

service/matching/matching_engine_test.go
```diff
@@ -3657,36 +3637,46 @@ func newTestTaskQueue(namespaceId string, name string, taskType enumspb.TaskQueu
...
+func (q *testQueueData) String() string {
+	var out strings.Builder
+	q.Lock()
+	defer q.Unlock()
+	fmt.Fprintf(&out, "<task queue ")
+	fmt.Fprintf(&out, "rangeID: %d ", q.rangeID)
+	fmt.Fprintf(&out, "info: %s ", q.info.String())
+	fmt.Fprintf(&out, "tasks: %d>", q.tasks.Size())
+	return out.String()
+}
```
Shivs11, line 3666: i assume this is purely present for debugging purposes (search did not show it being used anywhere for now)

service/matching/matching_engine_test.go
```diff
@@ -3657,36 +3637,46 @@ func newTestTaskQueue(namespaceId string, name string, taskType enumspb.TaskQueu
...
+	rangeID  int64
+	info     *persistencespb.TaskQueueInfo
+	tasks    treemap.Map
+	userData *persistencespb.VersionedTaskQueueUserData
+
+	createTaskCount  int
+	getTasksCount    int
+	getUserDataCount int
+	updateCount      int
 }
```
Shivs11, line 3651: just want to make sure I understand this and not questioning your decision to alter this - is the main reason you chose to have `testTaskManager` now store `testQueueData`, rather than `testPhysicalTaskQueueManager`, because you want it to closely resemble the `taskManager` struct?

I can see that this shouldn't have been storing stuff like `testPhysicalTaskQueueManager` in the first place

service/matching/matching_engine_test.go
```diff
@@ -3803,18 +3783,16 @@ func (m *testTaskManager) DeleteTaskQueue(
...
-		if randomNumber < threshold {
-			return true
-		}
+// all calls to addFault should be done before starting to call methods on testTaskManager
+func (m *testTaskManager) addFault(method, err string, fraction float32) {
+	if m.faultInjection == nil {
+		m.faultInjection = make(map[string]float32)
 	}
-	return false
+	m.faultInjection[method+":"+err] = fraction
```
Shivs11, line 3791: def think this approach is better than what I had previously - having said that, it took me a minute to understand what `method+":"+err` was trying to do and had to jump to the functions calling this method to get the "aha" moment

nit: one possible improvement here could be encoding the key as a struct, something like:

```

type faultKey struct {
	method string
	err    string
}

```

Shivs11 (approved):
ty for these changes - looks much better

service/matching/matching_engine_test.go
dnr: It's just renaming testPhysicalTaskQueueManager to testQueueData. The name makes more sense to me, it's just per-queue data for the test manager.

(Actually we should go even further and make it a testTaskStore instead of a testTaskManager, but that can be a later PR)

service/matching/matching_engine_test.go
dnr: yup

service/matching/matching_engine_test.go
dnr: true, but I think I'll just keep it the simple way for now, at least until we have to add another field to "faultKey". the two uses are only four lines apart so it should be easy enough to see

PR #8020, opened 2025-07-07
GetUserDataRetryPolicy should retry forever

dnr:
## What changed?
Fix GetUserDataRetryPolicy to retry forever.

## Why?
The default for retry policies is to give up after 1 minute. For the get user data loop, if we give up then we'll just retry immediately, which is fine, but on error we should follow the backoff schedule.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/user_data_manager_test.go
+++ b/service/matching/user_data_manager_test.go
@@ -459,7 +459,7 @@ func TestUserData_RetriesFetchOnUnavailable(t *testing.T) {
 	m := createUserDataManager(t, controller, tqCfg)
 	m.config.GetUserDataMinWaitTime = 10 * time.Second // wait on success
 	m.config.GetUserDataRetryPolicy = backoff.NewExponentialRetryPolicy(50 * time.Millisecond).
-		WithMaximumInterval(50 * time.Millisecond) // faster retry on failure
+		WithMaximumInterval(50 * time.Millisecond).WithExpirationInterval(backoff.NoInterval) // faster retry on failure
 
 	m.Start()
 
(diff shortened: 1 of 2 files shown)


service/matching/user_data_manager_test.go
```diff
@@ -788,7 +788,7 @@ func TestUserData_Propagation(t *testing.T) {
 		managers[i].config.GetUserDataLongPollTimeout = dynamicconfig.GetDurationPropertyFn(100 * time.Millisecond)
 		managers[i].config.GetUserDataMinWaitTime = 10 * time.Millisecond
 		managers[i].config.GetUserDataReturnBudget = 10 * time.Millisecond
-		managers[i].config.GetUserDataRetryPolicy = backoff.NewExponentialRetryPolicy(100 * time.Millisecond).WithMaximumInterval(1 * time.Second)
+		managers[i].config.GetUserDataRetryPolicy = backoff.NewExponentialRetryPolicy(100 * time.Millisecond).WithMaximumInterval(1 * time.Second).WithExpirationInterval(backoff.NoInterval)
```
carlydf, line 791: why is this `WithMaximumInterval(1 * time.Second)` instead of `WithMaximumInterval(1 * time.Minute)` like the config.go change?

dnr: look a few lines up?
```
// override timeouts to run much faster
```

carlydf: ah sorry that didnt show up in my github diff until I clicked to show more lines. understood!

carlydf (approved):

PR #8027, opened 2025-07-09
Track task queue stats per priority

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


service/matching/backlog_manager.go
```diff
@@ -38,10 +38,11 @@ type (
...
 		BacklogStatus() *taskqueuepb.TaskQueueStatus
-		// TotalApproximateBacklogCount returns an estimate of the total size of the backlog in
-		// persistence. It may be off in either direction.
-		TotalApproximateBacklogCount() int64
-		BacklogHeadAge() time.Duration
+		// ApproxBacklogCountsByPriority returns an estimate of the size of the backlog in persistence, per priority.
+		// It may be off in either direction.
+		ApproxBacklogCountsByPriority() map[int32]int64
+		// BacklogHeadAgeByPriority returns the age of the oldest task in the backlog, per priority.
+		BacklogHeadAgeByPriority() map[int32]time.Duration
```
dnr: what if we made this return a `map[priorityKey]*pb.TaskQueueStats` with count and age filled in?

service/matching/backlog_manager.go
```diff
@@ -175,26 +176,27 @@ func (c *backlogManagerImpl) BacklogCountHint() int64 {
...
 }
 
-func (c *backlogManagerImpl) TotalApproximateBacklogCount() int64 {
-	return c.db.getTotalApproximateBacklogCount()
+func (c *backlogManagerImpl) ApproxBacklogCountsByPriority() map[int32]int64 {
+	return c.db.getApproximateBacklogCounts()
+}
+
+func (c *backlogManagerImpl) BacklogHeadAgeByPriority() map[int32]time.Duration {
+	// This is not per-priority in the classic backlog manager, so return a map with only priority 0.
```
dnr: hmm, priority 0 doesn't exist. maybe we should use the default priority here?

service/matching/task_tracker.go
```diff
@@ -47,12 +47,12 @@ func (cb *circularTaskBuffer) totalTasks() int {
...
-	bucketSize        time.Duration // the duration of each bucket in the buffer
-	numberOfBuckets   int           // the total number of buckets in the buffer
-	totalIntervalSize time.Duration // the number of seconds over which rate of tasks are added/dispatched
-	tasksInInterval   circularTaskBuffer
+	startTime         time.Time                    // time when taskTracker was initialized
+	bucketStartTime   time.Time                    // the starting time of a bucket in the buffer
+	bucketSize        time.Duration                // the duration of each bucket in the buffer
+	numberOfBuckets   int                          // the total number of buckets in the buffer
+	totalIntervalSize time.Duration                // the number of seconds over which rate of tasks are added/dispatched
+	tasksInIntervals  map[int32]circularTaskBuffer // keyed by priority
```
dnr: just to confirm, did you consider keeping a map of taskTrackers instead of putting the map in taskTracker? I didn't think too hard about which is better, but I wonder if we're going to want to keep stats keyed in any other way.. possibly it could be better to keep taskTracker simpler and then making many of them for different aggregations?

but I guess doing it inside is less overhead overall.

dnr: is it okay to have a map with circularTaskBuffers as values instead of pointers, so they get copied?

dnr (commented):
I just skimmed for now, not super-thorough.

I mentioned a while ago that maybe we should have TaskQueueStats keep the absolute timestamp of the oldest instead of the "age", which is relative and constantly changing. should we do that now since we're changing the api?

service/matching/backlog_manager.go
stephanos: Good point; this is one of the places where I was initially using subqueue key instead of priority key.

service/matching/task_tracker.go
stephanos: 👍 I like that. Little bit of extra overhead, true, but should be simpler.

service/matching/physical_task_queue_manager.go
```diff
@@ -82,13 +83,13 @@ type (
 		pollerHistory               *pollerHistory
 		currentPolls                atomic.Int64
 		taskValidator               taskValidator
-		tasksAddedInIntervals       *taskTracker
-		tasksDispatchedInIntervals  *taskTracker
 		deploymentRegistrationCh    chan struct{}
 		deploymentVersionRegistered bool
 		pollerScalingRateLimiter    quotas.RateLimiter
 
-		firstPoll time.Time
```
stephanos, line 91: Not used.

stephanos:
> I mentioned a while ago that maybe we should have TaskQueueStats keep the absolute timestamp of the oldest instead of the "age", which is relative and constantly changing. should we do that now since we're changing the api?

I'd prefer to separate that from this PR. I foresee some discussion there and that would risk the delivery of this change.

service/matching/physical_task_queue_manager.go
```diff
@@ -700,3 +714,55 @@ func (c *physicalTaskQueueManagerImpl) makePollerScalingDecisionImpl(
...
+	// Otherwise, we need to maybe create a new tracker with the write lock.
+	c.taskTrackerMu.Lock()
+	defer c.taskTrackerMu.Unlock()
+	if tracker, ok := intervals[priorityKey]; ok {
+		return tracker // tracker was created while we were waiting for the lock
+	}
+
+	// Initalize all task trackers together; or the timeframes won't line up.
+	c.tasksAddedInIntervals[priorityKey] = newTaskTracker(c.partitionMgr.engine.timeSource)
+	c.tasksDispatchedInIntervals[priorityKey] = newTaskTracker(c.partitionMgr.engine.timeSource)
```
stephanos, line 743: Took me more than an hour to figure this out; an unrelated test was failing and pointed me to this conclusion.

service/matching/db.go
```diff
@@ -308,19 +308,14 @@ func (db *taskQueueDB) updateBacklogStatsLocked(subqueue int, countDelta int64,
...
 }
 
-func (db *taskQueueDB) getApproximateBacklogCount(subqueue int) int64 {
-	db.Lock()
-	defer db.Unlock()
-	return db.subqueues[subqueue].ApproximateBacklogCount
-}
-
-func (db *taskQueueDB) getTotalApproximateBacklogCount() (total int64) {
+func (db *taskQueueDB) getApproximateBacklogCounts() map[int32]int64 {
```
dnr: the change to split types will be later?

service/matching/db.go
```diff
@@ -308,19 +308,14 @@ func (db *taskQueueDB) updateBacklogStatsLocked(subqueue int, countDelta int64,
...
 	db.Lock()
 	defer db.Unlock()
+	counts := make(map[int32]int64)
 	for _, s := range db.subqueues {
-		total += s.ApproximateBacklogCount
+		counts[s.Key.Priority] = s.ApproximateBacklogCount
```
dnr: In the future we may have multiple subqueues for the same priority level (probably)
```suggestion
		counts[s.Key.Priority] += s.ApproximateBacklogCount
```

service/matching/physical_task_queue_manager.go
```diff
@@ -82,13 +83,13 @@ type (
...
 		currentPolls                atomic.Int64
 		taskValidator               taskValidator
-		tasksAddedInIntervals       *taskTracker
-		tasksDispatchedInIntervals  *taskTracker
 		deploymentRegistrationCh    chan struct{}
 		deploymentVersionRegistered bool
 		pollerScalingRateLimiter    quotas.RateLimiter
 
-		firstPoll time.Time
+		taskTrackerMu              sync.RWMutex
```
dnr: we usually use "lock"
```suggestion
		taskTrackerLock            sync.RWMutex
```

service/matching/physical_task_queue_manager.go
```diff
@@ -82,13 +83,13 @@ type (
...
+		tasksAddedInIntervals      map[int32]*taskTracker
+		tasksDispatchedInIntervals map[int32]*taskTracker
```
dnr: I don't think the longer name adds any value
```suggestion
		tasksAdded      map[int32]*taskTracker
		tasksDispatched map[int32]*taskTracker
```

service/matching/physical_task_queue_manager.go
```diff
@@ -700,3 +714,55 @@ func (c *physicalTaskQueueManagerImpl) makePollerScalingDecisionImpl(
 		PollRequestDeltaSuggestion: delta,
 	}
 }
+
+func (c *physicalTaskQueueManagerImpl) getOrCreateTaskTracker(
+	intervals map[int32]*taskTracker,
+	priorityKey int32,
+) *taskTracker {
+	if priorityKey == 0 {
+		priorityKey = defaultPriorityLevel(c.config.PriorityLevels()) // assign default priority key
```
dnr: It feels bad to go through dynamic config all the time here, I wonder if we should store the default level in the ptqm on initialization

service/matching/pri_backlog_manager.go
```diff
@@ -290,38 +298,35 @@ func (c *priBacklogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
...
 			AckLevel:  ackLevel,
 			TaskIdBlock: &taskqueuepb.TaskIdBlock{
 				StartId: currentTaskIDBlock.start,
 				EndId:   currentTaskIDBlock.end,
 			},
 			LoadedTasks:             int64(r.getLoadedTasks()),
-			MaxReadLevel:            c.db.GetMaxReadLevel(i),
-			ApproximateBacklogCount: c.db.getApproximateBacklogCount(i),
+			MaxReadLevel:            c.db.GetMaxReadLevel(subqueueKey),
+			ApproximateBacklogCount: c.db.getApproximateBacklogCounts()[priorityKey],
```
dnr: call this once instead of for each iteration

service/matching/db.go
dnr: actually nevermind.. after reading more I think this should return a slice of counts, one per subqueue. taskQueueDB should work in terms of subqueues only, it shouldn't know anything about their semantics, like priority

service/matching/pri_backlog_manager.go
dnr: see comment in db.. I think db should return these as a slice by subqueue

service/matching/db.go
stephanos: You mean `subqueueType` and `priorityKey` types? Yes, later. I didn't want to risk the delivery of this.

service/matching/db.go
stephanos: Hm, I don't know. I've implemented it, but unless I misunderstood the intention it feels worse now.

The index-based lookups have potential for off-by-one errors (had one already) since priorities are 1-based, the code is more complicated (now you need two loops), and the fact that db knows about priorities is still there and required to encode the priorities into the slice index (otherwise, how would the caller know which count belong to which priority).

Couldn't we just call the method `ApproximateBacklogCountByPriority` and return a map of priorities?

service/matching/backlog_manager.go
```diff
@@ -175,26 +173,28 @@ func (c *backlogManagerImpl) BacklogCountHint() int64 {
...
-}
-
 func (c *backlogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
 	taskIDBlock := rangeIDToTaskIDBlock(c.db.RangeID(), c.config.RangeSize)
 	return &taskqueuepb.TaskQueueStatus{
 		ReadLevel: c.taskAckManager.getReadLevel(),
 		AckLevel:  c.taskAckManager.getAckLevel(),
-		// use getApproximateBacklogCount instead of BacklogCountHint since it's more accurate
-		BacklogCountHint: c.db.getApproximateBacklogCount(subqueueZero),
+		// use getApproximateBacklogCountsBySubqueue instead of BacklogCountHint since it's more accurate
```
dnr:
```suggestion
		// use getTotalApproximateBacklogCount instead of BacklogCountHint since it's more accurate
```

service/matching/db.go
```diff
@@ -308,19 +308,36 @@ func (db *taskQueueDB) updateBacklogStatsLocked(subqueue int, countDelta int64,
...
+	for id, s := range db.subqueues {
+		counts := make([]int64, levels)
+		for p := range levels {
+			if s.Key.Priority == int32(p+1) {
+				counts[p] = s.ApproximateBacklogCount
+			}
+		}
+		result[id] = counts
+	}
+	return result
```
dnr: I meant:

```suggestion
	result := make([]int, len(db.subqueues))
	for id, s := range db.subqueues {
		result[id] = s.ApproximateBacklogCount
	}
	return result
```

and then backlogManager (and variants) can aggregate by priority level from there. I think db shouldn't really know anything about priority or the semantics of subqueues at all. It has to know just one thing, which is the default SubqueueKey, but maybe we can keep it to just that. And possibly even move that knowledge elsewhere later?

service/matching/pri_backlog_manager.go
```diff
@@ -260,20 +261,31 @@ func (c *priBacklogManagerImpl) BacklogCountHint() (total int64) {
...
-		return time.Duration(0)
+	result := make(map[int32]*taskqueuepb.TaskQueueStats)
+	priorityLevels := c.config.PriorityLevels()
+	backlogCounts := c.db.getApproximateBacklogCountsBySubqueue()
+	for idx := range c.subqueues {
+		for priorityIdx := range priorityLevels {
+			priority := priorityIdx + 1 // priority levels start at 1
+			result[priority] = &taskqueuepb.TaskQueueStats{
+				ApproximateBacklogCount: backlogCounts[idx][priorityIdx],
+			}
```
dnr: how about we just look at the priority keys of the subqueues that exist and aggregate by that (for now there will be only one though)? I don't want to return stats for levels that were never used.

(this means we just need to keep the reverse map in priBacklogManager)

service/matching/pri_backlog_manager.go
```diff
@@ -290,7 +302,7 @@ func (c *priBacklogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
 	return &taskqueuepb.TaskQueueStatus{
 		ReadLevel: readLevel,
 		AckLevel:  ackLevel,
-		// use getApproximateBacklogCount instead of BacklogCountHint since it's more accurate
+		// use getApproximateBacklogCountsBySubqueue instead of BacklogCountHint since it's more accurate
```
dnr:
```suggestion
		// use getTotalApproximateBacklogCount instead of BacklogCountHint since it's more accurate
```

service/matching/db.go
stephanos: Ah, I see. That makes sense.

service/matching/pri_backlog_manager.go
```diff
@@ -266,23 +269,18 @@ func (c *priBacklogManagerImpl) BacklogStatsByPriority() map[int32]*taskqueuepb.
...
-			oldestBacklogTime := c.subqueues[idx].getOldestBacklogTime()
-			if oldestBacklogTime.IsZero() {
-				// TODO(pri): returning 0 to match existing behavior, but maybe emptyBacklogAge would
-				// be more appropriate in the future.
-				result[priority].ApproximateBacklogAge = durationpb.New(0)
-			} else {
-				result[priority].ApproximateBacklogAge = durationpb.New(time.Since(oldestBacklogTime))
-			}
+	for subqueueKey, priorityKey := range c.priorityBySubqueue {
+		result[priorityKey] = &taskqueuepb.TaskQueueStats{
```
dnr: ok for now but ideally it should merge (i.e. sum) the counts, in case we have multiple subqueues with the same priority key

stephanos: Right; we had that before. I can add it back now.

PR #8028, opened 2025-07-09
Matching fairness standing backlog test

dnr:
## What changed?
Add unit test for backlog manager that sends tasks through with standing backlogs.

## Why?
Small framework to test concurrent operations on backlog managers.

## How did you test it?
- [x] added new unit test(s)

### DIFF
--- a/.github/.golangci.yml
+++ b/.github/.golangci.yml
@@ -27,15 +27,11 @@ issues:
       linters:
         - forbidigo
     - path: _test\.go|tests/.+\.go|common/testing/
-      text: "return value is not checked"
-      linters:
-        - errcheck
-    - path: _test\.go|tests/.+\.go|common/testing/
       text: "(cyclomatic|cognitive)"  # false positives when using subtests
       linters:
         - revive
     - path: _test\.go|tests/.+\.go|common/testing/
-      text: "(dot-imports|unchecked-type-assertion|unhandled-error)"  # helpful in tests
+      text: "(dot-imports|unchecked-type-assertion)"  # helpful in tests
       linters:
         - revive
     - path: ^tools\/.+\.go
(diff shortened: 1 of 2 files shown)


.github/.golangci.yml
```diff
@@ -22,16 +22,20 @@ issues:
...
+      linters:
+        - errcheck
+    - path: _test\.go|tests/.+\.go|common/testing/
       text: "(cyclomatic|cognitive)"  # false positives when using subtests
       linters:
         - revive
-    - path: _test\.go|tests/.+\.go
-      text: "(dot-imports|unchecked-type-assertion)"  # helpful in tests
+    - path: _test\.go|tests/.+\.go|common/testing/
+      text: "(dot-imports|unchecked-type-assertion|unhandled-error)"  # helpful in tests
```
stephanos: Does `unhandled-error` mean not writing `_ = ...`to drop the error? I think that's still useful even in tests.

common/testing/long.go
```diff
@@ -0,0 +1,23 @@
+package testing
+
+import (
+	"os"
+	"strconv"
+	"testing"
+)
+
+// LongTest calls Skip() on the testing.T or suite unless the environment variable
+// TEMPORAL_TEST_LONG is set to true.
```
stephanos, line 10: Is this more to protect local or CI execution?

.github/.golangci.yml
dnr: I feel like I've seen more bugs introduced by unthinkingly changing code to shut this up than bugs prevented, but okay

common/testing/long.go
dnr: mostly to protect CI, but also nice for local runs if you're not messing with backlog managers.

.github/.golangci.yml
stephanos: did I misunderstand the change? This would ignore `unhandled-error` lints, right? I agree that `_ = ...` is bad, but silently dropping them is worse, no?

dnr: No, that was the change, just avoid the `_ =` in tests and test-related code. But it's fine, I added them back and it passes lint now

dnr:
The remaining failures are `TestQueryNoRecentPoller`, which is weird, but I'll fix it in another PR

PR #8032, opened 2025-07-10
Deprecate ndc_history_resender code

xwduan:
## What changed?
Deprecate ndc_history_resender code

## Why?
It is replaced by Resend Handler which support resend local generated events.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
no risk. The dynamic config is enabled long time in prod.


service/frontend/admin_handler.go
```diff
@@ -1547,63 +1544,7 @@ func (adh *AdminHandler) ResendReplicationTasks(
...
-		request.GetRunId(),
-		resendStartEventID,
-		request.StartVersion,
-		common.EmptyEventID,
-		common.EmptyVersion,
-	); err != nil {
-		return nil, err
-	}
-	return &adminservice.ResendReplicationTasksResponse{}, nil
+	return nil, serviceerror.NewUnimplemented("ResendReplicationTasks is not implemented in AdminHandler")
```
hai719, line 1547: How upgrade/downgrade is handled? How about the s2c?

xwduan: This API is not used in current code base at all. We still have the original poll based replication code, but it is not called at that code path as well. So I believe this API is not used.

hai719 (approved):

PR #8033, opened 2025-07-10
Matching fairness: count-min sketch for counting

dnr:
## What changed?
Add new Counter implementations:
- count-min sketch for high cardinality of keys
- hybrid to switch from map to sketch when number of keys gets high enough

## Why?
Support high cardinality of fairness keys.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


service/matching/counter/hybrid.go
```diff
@@ -0,0 +1,66 @@
...
+	return &hybridCounter{
+		Counter: NewMapCounter(),
+		params:  params,
+		src:     src,
+	}
+}
+
+func (h *hybridCounter) GetPass(key string, base int64, inc int64) int64 {
+	p := h.Counter.GetPass(key, base, inc)
+	if m, ok := h.Counter.(*mapCounter); ok && len(m.m) > h.params.MapLimit {
```
ychebotarev, line 48: hard to read, too much exposure of internal details.
I would add something like "HasCapacity" to Counter interface.
then this code will be
```
if ! h.Counter.HasCapacity() {
   h.migrateToCMS() // I also wouldn't provide m.m
}
```

ychebotarev: I see you added EstimateDistinctKeys.

service/matching/counter/cmsketch.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	return count / s.params.D
+}
+
+func (s *cmSketch) Reseed(now time.Time) {
+	relTime := now.Sub(processStartTime)
+	s.reseeder.iter(relTime, s.seeds)
+}
+
+func (s *cmSketch) getIndexes(k string, indexes []int) {
```
ychebotarev: more like `fillIndexes` or `populateIndexes`. "get" usually imply that something will be returned.

service/matching/counter/reseed.go
```diff
@@ -0,0 +1,40 @@
...
+		src:    src,
+	}
+}
+
+func (r *reseeder) iter(now time.Duration, seeds []uint64) {
+	if r.params.Interval == 0 {
+		return
+	}
+	elapsed := now - r.lastTime
+	num := int(elapsed / r.params.Interval)
```
ychebotarev: nit: or
```
num := min(int(elapsed / r.params.Interval), len(seeds))
```
to avoid that `min(num, len(seeds))` below. I feel this is cleaner, but super nit.

service/matching/counter/cmsketch.go
```diff
@@ -0,0 +1,163 @@
...
+	return &cmSketch{
+		params:   params,
+		seed0:    maphash.MakeSeed(),
+		seeds:    makeSeeds(params.D, src),
+		cells:    make([]int64, params.W*params.D),
+		reseeder: newReseeder(params.Reseed, src),
+	}
+}
+
+func (s *cmSketch) GetPass(key string, base, inc int64) int64 {
```
ychebotarev, line 51: Can we have some testing? I'm not sure I can look at logic and spot any problems.

dnr: Yeah, I wanted to get something out there first.

I'll add some simple tests, but note that it's hard to ensure the correctness of this sort of thing by testing: when I was browsing around for comparable code I saw a go count-min sketch implementation in another project that was broken for multiple years in a way that tests wouldn't catch (reduced randomness)

service/matching/counter/hybrid.go
dnr: If it doesn't open up the map then it can't transition the existing counts. That's not strictly necessary but it's easy to do here so I wanted to do it.

Btw, the future plans for hybridCounter are actually to maintain a map of the top k _in addition_ to the CMS. So there might be more access to implementation details, not less. I don't really consider that a big deal: these are deliberately split out into a small-ish package.

service/matching/counter/reseed.go
dnr: hmm, if I do that then the next line doesn't work. I can move the min onto a new line though

service/matching/counter/hybrid.go
ychebotarev: >If it doesn't open up the map

sorry was not clear. What I mean - you can access that m.m from inside migrateToCMS. You are doing it here because you already get `m = h.Counter.(*mapCounter)`

service/matching/counter/reseed.go
ychebotarev: right. then its fine, lets keep it as is.

service/matching/counter/cmsketch.go
ychebotarev: just a simple basic tests will suffice.

service/matching/counter/hybrid.go
ychebotarev: but again - its not blocking, up to you

PR #8035, opened 2025-07-10
Scheduler: set WFID reuse policy to reject duplicates for scheduled starts

lina-temporal:
## What changed?
- The WFID reuse policy within the scheduler workflow has been updated.

## Why?
- Helps avoid a situation where a runaway WFT, timed out from the history end, still manages to make the `StartWorkflowExecution` request without recording its progress.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- Reuse policy isn't ever written to the scheduler memo, nor would it affect the recorded events, so it shouldn't cause replay problems (or necessitate a version bump)

### DIFF
--- a/service/worker/scheduler/workflow.go
+++ b/service/worker/scheduler/workflow.go
@@ -1315,6 +1315,13 @@ func (s *scheduler) startWorkflow(
 		continuedFailure = nil
 	}
 
+	// Reject duplicates as part of WFID reuse policy when possible, as a measure
+	// against WFT timeouts/failures that lead to non-determinism.
+	reusePolicy := enumspb.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE
+	if start.Manual {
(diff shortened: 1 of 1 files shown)


yiminc (approved):

service/worker/scheduler/workflow.go
```diff
@@ -1315,6 +1315,13 @@ func (s *scheduler) startWorkflow(
 		continuedFailure = nil
 	}
 
+	// Reject duplicates as part of WFID reuse policy when possible, as a measure
+	// against WFT timeouts/failures that lead to non-determinism.
+	reusePolicy := enumspb.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE
+	if start.Manual {
```
dnr, line 1321: This seems like a case where we should differentiate backfill from trigger immediately and only allow duplicate for backfill, so we get the extra protection for trigger immediately?

yiminc: I feel allow duplicate is the right policy for trigger immediately.

PR #8036, opened 2025-07-12
CHASM: Prefix internal ID with archetype

yycptt:
## What changed?
- Prefix internal ID with archetype

## Why?
- Follow the existing approach for separating IDs across archetypes.
- The change only minimize the risk of ID collision across archetype, but can't prevent it from happening entirely as Workflows' ID don't have the prefix and essentially claimed the entire ID space. To really prevent the issue, we need a separate columns in the database table, but that's a very complex change and potentially involves data migration. 

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks

### DIFF
--- a/chasm/engine.go
+++ b/chasm/engine.go
@@ -1,47 +1,8 @@
-//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination engine_mock.go
-
 package chasm
 
 import (
 	"context"
 )
-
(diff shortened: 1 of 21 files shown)


yycptt:
Scheduler will add the prefix in Scheduler's API handler. For other newer archetypes, we will take a different approach than during archetype prefix.

PR #8038, opened 2025-07-14
Matching fairness per-key rate limits

dnr:
## What changed?
- Add support for per-key rate limits that are scaled by weight (so maintaining dispatch in the same order that tasks are read in).
- Refactor simpleLimiter to split parameters and ready time.

## Why?
Support simple per-key limits.

This uses an LRU cache to track recent tasks by fairness key, default size 2000. With target dispatch rate of 1000 tasks/second/partition, that corresponds to at least 2 seconds of keys (equal in the worst case of fully uniform key distribution, which is unlikely). That means that rates >= 0.5 tasks/second/key/partition should be handled accurately (2 tasks/second/key for default 4 partitions). To handle even lower rates, partition count and task size can be tuned.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/service/matching/matcher_data.go
+++ b/service/matching/matcher_data.go
@@ -78,6 +78,18 @@ type taskPQ struct {
 	// Rate limiter for overall queue
 	wholeQueueLimit simpleLimiterParams
 	wholeQueueReady simpleLimiter
+
+	// Rate limiter for individual fairness keys.
+	//
+	// Note that we currently have only one limit for all keys, which is scaled by the key's
+	// weight. If we do this, we can assume that all keys are either at or below their rate
+	// limit at the same time, so that if the head of the queue is at its limit, then the rest
+	// must also be, and so we don't have to "skip over" the head of the queue due to rate
+	// limits. This isn't true in situations where weights have changed in between writing and
+	// reading. We'll handle that situation better in the future.
+	perKeyLimit     simpleLimiterParams
+	perKeyOverrides fairnessWeightOverrides // TODO(fairness): get this from config
+	perKeyReady     map[string]simpleLimiter
 }
 
 func (t *taskPQ) Add(task *internalTask) {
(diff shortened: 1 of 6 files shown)


service/matching/matcher_data.go
```diff
@@ -629,5 +718,19 @@ func (s *simpleLimiter) consume(now int64, tokens int64) {
...
+	if p.never() {
+		return simpleLimiterNever
+	}
+	clippedReady := max(now, int64(ready)+p.burst.Nanoseconds()) - p.burst.Nanoseconds()
+	return simpleLimiter(clippedReady + tokens*p.interval.Nanoseconds())
+}
+
+// clip sets ready to an allowable range based on the given parameters.
+func (ready simpleLimiter) clip(p simpleLimiterParams, now int64, maxTokens int64) simpleLimiter {
+	if p.interval < 0 {
```
stephanos: Should this be `if p.never() {`?

service/matching/matcher_data.go
```diff
@@ -629,5 +718,19 @@ func (s *simpleLimiter) consume(now int64, tokens int64) {
...
+		return simpleLimiterNever
+	}
+	// if ready was set very far in the future (e.g. because the rate was zero), then we can
```
stephanos:
```suggestion
	// If ready was set very far in the future (e.g. because the rate was zero), then we can
```

service/matching/matcher_data.go
```diff
@@ -629,5 +718,19 @@ func (s *simpleLimiter) consume(now int64, tokens int64) {
...
+	// clip it back to now + maxTokens*interval + burst.
+	return min(ready, simpleLimiter(now+maxTokens*p.interval.Nanoseconds()+p.burst.Nanoseconds()))
```
stephanos: Would it be helpful to extract a variable from `now+maxTokens*p.interval.Nanoseconds()+p.burst.Nanoseconds()` to give it a name?

stephanos (approved):

service/matching/matcher_data.go
dnr: sure

service/matching/matcher_data.go
dnr: sure

PR #8043, opened 2025-07-14
Added SQL Support for Fairness

sivagirish81:
Created a new table in SQL schema - tasks_v2 to support Fairness. Added schema changes for MySQL, PostgreSQL, and SQLite. Added new SQL plugin files for matching task queue and task v2. Updated existing SQL plugin files to support the new task v2 structure. Updated tests to cover the new task v2 functionality.

## What changed?
+ Introduced 2 new tables tasks_v2 and task_queue_v2 to support fairness and be consistent with cassandra schema.
+ Added schema changes for MySQL, PostgreSQL, and SQLite. 
+ Added new SQL plugin files for matching task queue and task v2. 
+ Updated existing SQL plugin files to support the new task v2 structure. 
+ Updated tests to cover the new task v2 functionality.

## Why?
+ Updated tasks to store pass value for fairness.
+ Maintain separate tables for fairness and eventually deprecate the V1 schema. 

## How did you test it?
- [ ] built

### DIFF
--- a/common/persistence/sql/sqlplugin/matching_task_queue_v2.go
+++ b/common/persistence/sql/sqlplugin/matching_task_queue_v2.go
@@ -7,7 +7,7 @@ import (
 
 type (
 	// TaskQueuesRow represents a row in task_queues table
-	TaskQueuesRow struct {
+	TaskQueuesRowV2 struct {
 		RangeHash    uint32
 		TaskQueueID  []byte
 		RangeID      int64
@@ -17,7 +17,7 @@ type (
 
 	// TaskQueuesFilter contains the column names within task_queues table that
 	// can be used to filter results through a WHERE clause
-	TaskQueuesFilter struct {
+	TaskQueuesFilterV2 struct {
 		RangeHash                   uint32
 		RangeHashGreaterThanEqualTo uint32
 		RangeHashLessThanEqualTo    uint32
@@ -27,7 +27,7 @@ type (
 		PageSize                    *int
 	}
 
-	UpdateTaskQueueDataRequest struct {
+	UpdateTaskQueueDataRequestV2 struct {
 		NamespaceID   []byte
 		TaskQueueName string
 		Version       int64
@@ -36,15 +36,15 @@ type (
 	}
 
 	// MatchingTaskQueue is the SQL persistence interface for matching task queues
-	MatchingTaskQueue interface {
-		InsertIntoTaskQueues(ctx context.Context, row *TaskQueuesRow) (sql.Result, error)
-		UpdateTaskQueues(ctx context.Context, row *TaskQueuesRow) (sql.Result, error)
+	MatchingTaskQueueV2 interface {
+		InsertIntoTaskQueuesV2(ctx context.Context, row *TaskQueuesRowV2) (sql.Result, error)
+		UpdateTaskQueuesV2(ctx context.Context, row *TaskQueuesRowV2) (sql.Result, error)
 		// SelectFromTaskQueues returns one or more rows from task_queues table
 		// Required Filter params:
 		//  to read a single row: {shardID, namespaceID, name, taskType}
 		//  to range read multiple rows: {shardID, namespaceIDGreaterThan, nameGreaterThan, taskTypeGreaterThan, pageSize}
-		SelectFromTaskQueues(ctx context.Context, filter TaskQueuesFilter) ([]TaskQueuesRow, error)
-		DeleteFromTaskQueues(ctx context.Context, filter TaskQueuesFilter) (sql.Result, error)
-		LockTaskQueues(ctx context.Context, filter TaskQueuesFilter) (int64, error)
+		SelectFromTaskQueuesV2(ctx context.Context, filter TaskQueuesFilterV2) ([]TaskQueuesRowV2, error)
+		DeleteFromTaskQueuesV2(ctx context.Context, filter TaskQueuesFilterV2) (sql.Result, error)
+		LockTaskQueuesV2(ctx context.Context, filter TaskQueuesFilterV2) (int64, error)
 	}
 )
--- a/common/persistence/sql/sqlplugin/matching_task_v2.go
+++ b/common/persistence/sql/sqlplugin/matching_task_v2.go
@@ -1,0 +1,46 @@
+package sqlplugin
+
+import (
+	"context"
+	"database/sql"
+)
+
+type (
+	// TasksRow represents a row in tasks table
+	TasksRowV2 struct {
+		RangeHash    uint32
+		TaskQueueID  []byte
+		TaskID       int64
+		Pass         int64 // pass for tasks (see stride scheduling algorithm for fairness)
+		Data         []byte
+		DataEncoding string
+	}
+
+	// TasksFilter contains the column names within tasks table that
+	// can be used to filter results through a WHERE clause
+	TasksFilterV2 struct {
+		RangeHash          uint32
+		TaskQueueID        []byte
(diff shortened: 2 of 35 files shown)


common/persistence/sql/sqlplugin/matching_task_queue_v2.go
```diff
@@ -0,0 +1,50 @@
+package sqlplugin
+
+import (
+	"context"
+	"database/sql"
+)
+
+type (
+	// TaskQueuesRow represents a row in task_queues table
+	TaskQueuesRowV2 struct {
```
dnr, line 10: Do we have to duplicate these types? For task queues, they're exactly the same as v1, right?

Could we do it more like cassandra where we share all the internal structs like TaskStore and the Internal versions of the requests?

common/persistence/sql/sqlplugin/matching_task_v2.go
```diff
@@ -0,0 +1,46 @@
...
+	"database/sql"
+)
+
+type (
+	// TasksRow represents a row in tasks table
+	TasksRowV2 struct {
+		RangeHash    uint32
+		TaskQueueID  []byte
+		TaskID       int64
+		Pass         int64 // pass for tasks (see stride scheduling algorithm for fairness)
```
dnr, line 14: For consistency with cass, let's call it TaskPass

```suggestion
		TaskPass     int64 // pass for tasks (see stride scheduling algorithm for fairness)
```

Oh, and put it above TaskID to match the order that the columns are declared and the primary key

common/persistence/sql/sqlplugin/matching_task_v2.go
```diff
@@ -0,0 +1,46 @@
...
+		Data         []byte
+		DataEncoding string
+	}
+
+	// TasksFilter contains the column names within tasks table that
+	// can be used to filter results through a WHERE clause
+	TasksFilterV2 struct {
+		RangeHash          uint32
+		TaskQueueID        []byte
+		Pass               int64 // pass for tasks (see stride scheduling algorithm for fairness)
```
dnr, line 24: Min? Max? Inclusive? Exclusive?
Should this be like the TaskID filters where it's optional?

Actually I don't really like this declaration since pass and id are not independent, they're a tuple. If we want a general "filter" with options instead of what we're doing with cassandra where we only really have a lower bound, then we should do it more like:

Define a struct that's `type FairLevel struct { pass, id int64 }` (yes it's the same as the one in matching, but we can duplicate it), then have `InclusiveMinLevel *FairLevel` + `ExclusiveMaxLevel *FairLevel` here

common/persistence/sql/sqlplugin/mysql/task.go
```diff
@@ -57,34 +55,12 @@ task_queue_id = :task_queue_id
 		`tasks(range_hash, task_queue_id, task_id, data, data_encoding) ` +
 		`VALUES(:range_hash, :task_queue_id, :task_id, :data, :data_encoding)`
 
-	deleteTaskQry = `DELETE FROM tasks ` +
-		`WHERE range_hash = ? AND task_queue_id = ? AND task_id = ?`
+	// deleteTaskQry = `DELETE FROM tasks ` +
```
dnr, line 58: if this is unused, just delete it

common/persistence/sql/sqlplugin/mysql/task_v2.go
```diff
@@ -0,0 +1,249 @@
...
+import (
+	"context"
+	"database/sql"
+
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/persistence/sql/sqlplugin"
+)
+
+const (
+	taskQueueCreatePartV2 = `INTO task_queues_v2(range_hash, task_queue_id, range_id, data, data_encoding) ` +
```
dnr, line 12: maybe we can share all or most of the task queue code, and just replace the table name?

schema/mysql/v8/temporal/schema.sql
```diff
@@ -83,6 +83,19 @@ CREATE TABLE tasks (
   PRIMARY KEY (range_hash, task_queue_id, task_id)
 );
 
+-- Stores activity or workflow tasks
+-- Used for fairness scheduling. (pass, task_id) are monotonically increasing.
+CREATE TABLE tasks_v2 (
+  range_hash INT UNSIGNED NOT NULL,
+  task_queue_id VARBINARY(272) NOT NULL,
+  task_id BIGINT NOT NULL,
+  pass BIGINT NOT NULL, -- pass for tasks (see stride scheduling algorithm for fairness)
```
dnr, line 92: put pass above task_id so the columns match the primary key

common/persistence/sql/sqlplugin/mysql/task_v2.go
```diff
@@ -0,0 +1,249 @@
...
+		data_encoding = :data_encoding
+		WHERE
+		range_hash = :range_hash AND
+		task_queue_id = :task_queue_id
+	`
+
+	listTaskQueueRowSelectV2 = `SELECT range_hash, task_queue_id, range_id, data, data_encoding from task_queues_v2 `
+
+	listTaskQueueWithHashRangeQryV2 = listTaskQueueRowSelectV2 +
+		`WHERE range_hash >= ? AND range_hash <= ? AND task_queue_id > ? AND pass > ? ORDER BY task_queue_id ASC LIMIT ?`
```
dnr, line 30: `task_queues_v2` has no pass column?

common/persistence/sql/sqlplugin/mysql/task_v2.go
```diff
@@ -0,0 +1,249 @@
...
+
+// SelectFromTasks reads one or more rows from tasks table
+func (mdb *db) SelectFromTasksV2(
+	ctx context.Context,
+	filter sqlplugin.TasksFilterV2,
+) ([]sqlplugin.TasksRowV2, error) {
+	var err error
+	var rows []sqlplugin.TasksRowV2
+	switch {
+	case filter.PageSize != nil:
```
dnr, line 88: hmm... so we have this general TasksFilter[V2] but we don't actually support all combinations of filters. in fact here we only support exactly one combination. I don't really like this, and I kind of wish this was just set up more like cassandra where the interface just supported one query instead of pretending to be more flexible. but I suppose we shouldn't change it here?

common/persistence/sql/sqlplugin/sqlite/task_v2.go
```diff
@@ -0,0 +1,264 @@
...
+		`WHERE range_hash = ? AND task_queue_id = ?`
+	// *** Task_Queues Table Above ***
+
+	// *** Tasks Below ***
+	getTaskMinMaxQryV2 = `SELECT task_id, data, data_encoding ` +
+		`FROM tasks_v2 ` +
+		`WHERE range_hash = ? ` +
+		`AND task_queue_id = ? ` +
+		`AND (pass, task_id) >= (?, ?) ` +
+		`AND task_id < ? ` +
```
dnr, line 50: I don't think a max for task_id makes sense here, we shouldn't need this query

sivagirish81:
This PR is a work in progress. The test suite still needs to be fixed. Right now I have just replicated the pre-existing tests for sql based on persistence without fairness.

schema/sqlite/version.go
```diff
@@ -1,7 +1,7 @@
 package sqlite
 
 // Version is the SQLite database release version
-const Version = "0.9"
+const Version = "1.0"
```
dnr, line 4: the version after 0.9 is 0.10

PR #8048, opened 2025-07-15
Turn on eager workflow start by default

bergundy:
## Why?

The feature is moving to public preview.


common/dynamicconfig/constants.go
```diff
@@ -189,7 +189,7 @@ config as the other services.`,
 	)
 	EnableEagerWorkflowStart = NewNamespaceBoolSetting(
 		"system.enableEagerWorkflowStart",
-		false,
+		true,
```
yycptt, line 192: nit: plz also remove the overrides in dev configs & functional tests.

yycptt (approved):

PR #8052, opened 2025-07-15
Fix dropped fairness task

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


service/matching/fair_task_reader.go
```diff
@@ -348,7 +348,7 @@ func (tr *fairTaskReader) mergeTasks(tasks []*persistencespb.AllocatedTaskInfo,
 	it := merged.Iterator()
 	var highestLevel fairLevel
 	tasks = tasks[:0] // reuse incoming slice to avoid an allocation
-	for b := 0; it.Next() && b < batchSize; b++ {
+	for b := 0; b < batchSize && it.Next(); b++ {
```
stephanos, line 351: Before, it was possible that a task was consumed from the iterator and then dropped if it happened at the batch size boundary.

dnr (approved):

PR #8053, opened 2025-07-15
Input validation for UpdateTaskQueueConfig api and added additional test

sivagirish81:
## What changed?
+ Added input validation for UpdateTaskQueueConfig api.
+ Added additional test to validate unsetting of rate limits via UpdateTaskQueueConfig api.

## Why?
+ Input validation for UpdateTaskQueueConfig api.

## How did you test it?
- [x] built

## Potential risks
N/A

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -1301,9 +1301,12 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		cacheKey := "dtq_default:" + strings.Join(buildIds, ",")
 		if ts := pm.GetCache(cacheKey); ts != nil {
 			//revive:disable-next-line:unchecked-type-assertion
-			descrResp.DescResponse.Stats = ts.(*taskqueuepb.TaskQueueStats)
+			cachedResp := ts.(*workflowservice.DescribeTaskQueueResponse)
+			descrResp.DescResponse.Stats = cachedResp.Stats
+			descrResp.DescResponse.StatsByPriorityKey = cachedResp.StatsByPriorityKey
 		} else {
 			taskQueueStats := &taskqueuepb.TaskQueueStats{}
+			taskQueueStatsByPriority := make(map[int32]*taskqueuepb.TaskQueueStats)
 
 			// No version was requested, so we need to query all versions.
 			if len(buildIds) == 0 {
@@ -1342,12 +1345,22 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 					return nil, err
 				}
 				for _, vii := range partitionResp.VersionsInfoInternal {
-					partitionStats := vii.PhysicalTaskQueueInfo.TaskQueueStats
-					mergeStats(taskQueueStats, partitionStats)
+					partitionStats := vii.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey
+					for pri, priorityStats := range partitionStats {
+						if _, ok := taskQueueStatsByPriority[pri]; !ok {
+							taskQueueStatsByPriority[pri] = &taskqueuepb.TaskQueueStats{}
+						}
+						mergeStats(taskQueueStats, priorityStats)
+						mergeStats(taskQueueStatsByPriority[pri], priorityStats)
+					}
 				}
 			}
-			pm.PutCache(cacheKey, taskQueueStats)
+			pm.PutCache(cacheKey, &workflowservice.DescribeTaskQueueResponse{
+				Stats:              taskQueueStats,
+				StatsByPriorityKey: taskQueueStatsByPriority,
+			})
 			descrResp.DescResponse.Stats = taskQueueStats
+			descrResp.DescResponse.StatsByPriorityKey = taskQueueStatsByPriority
 		}
 	}
 
@@ -1402,9 +1415,10 @@ func (e *matchingEngineImpl) DescribeVersionedTaskQueues(
 		}
 		resp.VersionTaskQueues = append(resp.VersionTaskQueues,
 			&matchingservice.DescribeVersionedTaskQueuesResponse_VersionTaskQueue{
-				Name:  tq.Name,
-				Type:  tq.Type,
-				Stats: tqResp.DescResponse.Stats,
+				Name:               tq.Name,
+				Type:               tq.Type,
+				Stats:              tqResp.DescResponse.Stats,
+				StatsByPriorityKey: tqResp.DescResponse.StatsByPriorityKey,
 			})
 	}
 
--- a/api/matchingservice/v1/request_response.pb.go
+++ b/api/matchingservice/v1/request_response.pb.go
@@ -4626,12 +4626,16 @@ func (x *DescribeVersionedTaskQueuesRequest_VersionTaskQueue) GetType() v19.Task
 
 // (-- api-linter: core::0123::resource-annotation=disabled --)
 type DescribeVersionedTaskQueuesResponse_VersionTaskQueue struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	Name          string                 `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
-	Type          v19.TaskQueueType      `protobuf:"varint,2,opt,name=type,proto3,enum=temporal.api.enums.v1.TaskQueueType" json:"type,omitempty"`
-	Stats         *v14.TaskQueueStats    `protobuf:"bytes,3,opt,name=stats,proto3" json:"stats,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	state protoimpl.MessageState `protogen:"open.v1"`
+	Name  string                 `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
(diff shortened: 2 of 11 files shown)


service/matching/matching_engine.go
```diff
@@ -2925,16 +2925,17 @@ func stickyWorkerAvailable(pm taskQueuePartitionManager) bool {
 	return pm != nil && pm.HasPollerAfter("", time.Now().Add(-stickyPollerUnavailableWindow))
 }
 
-func buildRateLimitConfig(update *workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate, updateTime *timestamppb.Timestamp) *taskqueuepb.RateLimitConfig {
+func buildRateLimitConfig(update *workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate, updateTime *timestamppb.Timestamp, updateIdentity string) *taskqueuepb.RateLimitConfig {
```
sivagirish81, line 2945: Missed to populate updateIdentity in the configMetaData previously.
Addressed that in this PR.

service/frontend/validators.go
```diff
@@ -17,3 +21,26 @@ func validateExecution(w *commonpb.WorkflowExecution) error {
...
+	if update.RateLimit.GetRequestsPerSecond() < 0 {
+		return serviceerror.NewInvalidArgumentf("RequestsPerSecond for %s rate limit must be non-negative.", label)
+	}
+	if update.RateLimit.GetRequestsPerSecond() > math.MaxFloat32 {
+		return serviceerror.NewInvalidArgumentf("RequestsPerSecond for %s rate limit exceeds the supported maximum.", label)
+	}
+	return nil
+}
+
+func validateStringField(fieldName, value string, maxLen int, required bool) error {
```
sivagirish81: TODO : Missed to declare fieldName as type string.

service/frontend/workflow_handler.go
```diff
@@ -99,6 +99,9 @@ const (
 	errTooManyDeleteDeploymentRequests  = "Too many DeleteWorkerDeployment requests have been issued in rapid succession. Please throttle the request rate to avoid exceeding Worker Deployment resource limits."
 	errTooManyDeleteVersionRequests     = "Too many DeleteWorkerDeploymentVersion requests have been issued in rapid succession. Please throttle the request rate to avoid exceeding Worker Deployment resource limits."
 	errTooManyVersionMetadataRequests   = "Too many UpdateWorkerDeploymentVersionMetadata requests have been issued in rapid succession. Please throttle the request rate to avoid exceeding Worker Deployment resource limits."
+
+	maxReasonLength   = 256 // Maximum length for the reason field in RateLimitUpdate configurations.
```
sivagirish81: Set the maxReasonLength & maxIdentityLength to 256 as a constant for now.
Will  change it to  a constant similar to `MaxIDLengthLimit` from constants dynamicconfig.go.
Will reduce the maxLength to 100 characters itself.

service/frontend/validators.go
```diff
@@ -17,3 +21,26 @@ func validateExecution(w *commonpb.WorkflowExecution) error {
...
+func validateRateLimit(update *workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate, label string) error {
+	if update == nil || update.RateLimit == nil {
+		return nil
+	}
+	if update.RateLimit.GetRequestsPerSecond() < 0 {
+		return serviceerror.NewInvalidArgumentf("RequestsPerSecond for %s rate limit must be non-negative.", label)
+	}
+	if update.RateLimit.GetRequestsPerSecond() > math.MaxFloat32 {
+		return serviceerror.NewInvalidArgumentf("RequestsPerSecond for %s rate limit exceeds the supported maximum.", label)
+	}
```
dnr: is this possible? the field is a float32... I guess if it's infinite? but we don't really have to disallow that, it'll just come out the same as no limit

service/frontend/workflow_handler.go
```diff
@@ -6064,6 +6067,30 @@ func (wh *WorkflowHandler) UpdateTaskQueueConfig(
 	if err != nil {
 		return nil, err
 	}
+	// Validation: prohibit setting rate limit on workflow task queues
+	if request.TaskQueueType == enumspb.TASK_QUEUE_TYPE_WORKFLOW {
+		return nil, serviceerror.NewInvalidArgument("Setting rate limit on workflow task queues is not allowed.")
```
dnr, line 6165: this is okay for now, but as soon as we add any config that does apply to workflows (like fairness weight overrides) we'll have to change this condition

service/frontend/workflow_handler.go
```diff
@@ -6064,6 +6067,30 @@ func (wh *WorkflowHandler) UpdateTaskQueueConfig(
...
+	fairnessKeyRateLimitDefault := request.GetUpdateFairnessKeyRateLimitDefault()
+	// Validate rate limits
+	if err := validateRateLimit(queueRateLimit, "queue"); err != nil {
+		return nil, err
+	}
+	if err := validateRateLimit(fairnessKeyRateLimitDefault, "fairness key"); err != nil {
+		return nil, err
+	}
+	// Validate reason fields
+	if err := validateStringField("UpdateQueueRateLimit.Reason", queueRateLimit.GetReason(), maxReasonLength, false); err != nil {
```
dnr: we should put this call in validateRateLimit so we don't have to repeat this call twice (and maybe forget one)

service/frontend/workflow_handler.go
dnr: I think maxReasonLength is okay as a constant. For schedules notes (similar thing) we use a limit of 1000 (common.ScheduleNotesSizeLimit), so maybe this should be 1000 also.

service/frontend/workflow_handler.go
```diff
@@ -99,6 +99,9 @@ const (
 	errTooManyDeleteDeploymentRequests  = "Too many DeleteWorkerDeployment requests have been issued in rapid succession. Please throttle the request rate to avoid exceeding Worker Deployment resource limits."
 	errTooManyDeleteVersionRequests     = "Too many DeleteWorkerDeploymentVersion requests have been issued in rapid succession. Please throttle the request rate to avoid exceeding Worker Deployment resource limits."
 	errTooManyVersionMetadataRequests   = "Too many UpdateWorkerDeploymentVersionMetadata requests have been issued in rapid succession. Please throttle the request rate to avoid exceeding Worker Deployment resource limits."
+
+	maxReasonLength   = 256 // Maximum length for the reason field in RateLimitUpdate configurations.
+	maxIdentityLength = 256 // Maximum length for the identity field in UpdateRateLimitRequest
```
dnr: elsewhere in this file I see a lot of checks against `wh.config.MaxIDLengthLimit()`, so might as well use that one for identity to be consistent

service/frontend/workflow_handler.go
sivagirish81: Changed maxReasonLength to 1000 and kept it as a constant.

service/frontend/validators.go
sivagirish81: Agreed that it will reflect no limit.
Removed the check for a number greater than MaxFloat32.

service/frontend/workflow_handler.go
sivagirish81: When I override the fairness config I will add the checks over here to make sure that if its a workflow and the queueRateLimit is attempted to being set then I will throw the error.
But if the queueRateLimit is nil and the api tries to set the fairnessKeyRateLimit I will allow it.

service/frontend/workflow_handler.go
sivagirish81: Done.

service/frontend/workflow_handler.go
sivagirish81: Used wh.config.MaxIDLengthLimit() for identity length check to be consistent.

service/frontend/workflow_handler.go
```diff
@@ -6064,6 +6066,23 @@ func (wh *WorkflowHandler) UpdateTaskQueueConfig(
...
+	fairnessKeyRateLimitDefault := request.GetUpdateFairnessKeyRateLimitDefault()
+	// Validate rate limits
+	if err := validateRateLimit(queueRateLimit, "UpdateQueueRateLimit"); err != nil {
+		return nil, err
+	}
+	if err := validateRateLimit(fairnessKeyRateLimitDefault, "UpdateFairnessKeyRateLimitDefault"); err != nil {
+		return nil, err
+	}
+	// Validate identity field
+	if err := validateStringField("Identity", request.GetIdentity(), wh.config.MaxIDLengthLimit(), true); err != nil {
```
dnr: as far as I know, we never _require_ Identity, we just have a max length. we shouldn't require it here either.

service/frontend/validators.go
```diff
@@ -3,6 +3,12 @@ package frontend
 import (
 	"github.com/pborman/uuid"
 	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/api/workflowservice/v1"
+)
+
+const (
+	reason = ".Reason"
```
dnr: you can just use an inline literal string if you want, there's not really any reason to make this a const

dnr (approved):

service/frontend/workflow_handler.go
sivagirish81: Removed the mandatory check for Identity.

service/frontend/validators.go
sivagirish81: Inlined the literal.

PR #8057, opened 2025-07-16
Allow nexus registry to indicate that a not found error is retryable

bergundy:
## Why?

More control for custom registry implementations.


service/frontend/nexus_http_handler.go
```diff
@@ -196,6 +196,13 @@ func (h *NexusHTTPHandler) dispatchNexusTaskByEndpoint(w http.ResponseWriter, r
 		}
 		switch s.Code() {
 		case codes.NotFound:
+			if r, ok := (err.(interface{ Retryable() bool })); ok {
```
pdoerner, line 199: Might consider making this a utility function since I expect we will need to check other gRPC errors for a retryable flag too

bergundy: I can extract that later once there are other use cases.

PR #8058, opened 2025-07-16
Attach RateLimiter using UpdateTaskQueueConfig api configration

sivagirish81:
## What changed?
+ Implemented a new RateLimit Manager interface.
+ Populated effective rate limit field in DescribeTaskQueue api response.
+ Added a test to check whether the api set rate limit is overriding the worker set rate limit

## Why?
+ Configs set via updateTaskQueueConfig api need to override the worker set rate limit.
+ Report effective RPS as the min of api/worker and system dynamic configs.
+ Report source basis the effective RPS.

## How did you test it?
- [x] built

## Potential risks
N/A

### DIFF
--- a/tests/task_queue_test.go
+++ b/tests/task_queue_test.go
@@ -239,7 +239,7 @@ func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
 
 	// Start the activity worker
 	activityWorker := worker.New(s.SdkClient(), activityTaskQueue, worker.Options{
-		// Setting rate limit at worker level
+		// Setting rate limit at worker level (this will be ignored in favor of the limit set through the api)
 		TaskQueueActivitiesPerSecond: workerRPS,
 	})
 	activityWorker.RegisterActivityWithOptions(activityFunc, activity.RegisterOptions{Name: activityName})
@@ -253,7 +253,7 @@ func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
 	defer wfWorker.Stop()
 
 	// Launch workflows
-	for i := 0; i < taskCount; i++ {
+	for i := range taskCount {
 		_, err := s.SdkClient().ExecuteWorkflow(ctx, sdkclient.StartWorkflowOptions{
 			TaskQueue: tv.TaskQueue().GetName(),
 			ID:        fmt.Sprintf("wf-%d", i),


service/matching/ratelimiter_manager_interface.go
sivagirish81, line 1: TODO : Will rename the file to ratelimit_manager_interface.go to be consistent with the implementation.

sivagirish81:
Will need to recheck the functional test added.

tests/task_queue_test.go
```diff
@@ -145,15 +145,15 @@ func (s *TaskQueueSuite) taskQueueRateLimitTest(nPartitions, nWorkers int, timeT
 
 }
 
-func (s *TaskQueueSuite) getBacklogCount(ctx context.Context, tv *testvars.TestVars) int64 {
+func (s *TaskQueueSuite) getBacklogCount(ctx context.Context, tv *testvars.TestVars, taskQueueType int32) int64 {
```
sivagirish81: Will explore avoiding the enhanced api mode of DescribeTaskQueue and instead using the data from the reportStats field for the backlog count.

sivagirish81:
TODO : Fix Race Condition
Explore using different workers and task queues for activity and workflow
Measure the gap between successive activity executions to assert that rate limiter is working.
**Update** : Done.

service/matching/ratelimit_manager_interface.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/server/common/quotas"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+type (
+	// RateLimiterManager interface defines methods to select the appropriate rate limiter for task queues.
+	rateLimitManager interface {
```
dnr: why do we need an interface here? are we mocking this? if not just use the struct directly

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/server/common/quotas"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+type (
+	rateLimitManagerImpl struct {
+		effectiveRPS    float64                 // Min of api/worker set RPS and system defaults.
+		rateLimitSource int32                   // Source of the rate limit, can be set via API, worker or system default.
```
dnr: use the proper enum type

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+)
+
+var _ rateLimitManager = (*rateLimitManagerImpl)(nil)
+
+// Create a new rate limit manager for the task queue partition.
+// `workerRPS` is expected to be nil and will be set lazily via SetWorkerRPS at the time of worker poll.
+// `apiConfigRPS` will be set to nil initially and will get populated once a poll / describeTaskQueue request is called.
+func newRateLimitManager(userDataManager userDataManager,
+	workerRPS *wrapperspb.DoubleValue, apiConfigRPS *wrapperspb.DoubleValue, tqConfig *taskQueueConfig) *rateLimitManagerImpl {
```
dnr: the usual formatting is more like:

```suggestion
func newRateLimitManager(
	userDataManager userDataManager,
	workerRPS *wrapperspb.DoubleValue,
	apiConfigRPS *wrapperspb.DoubleValue,
	tqConfig *taskQueueConfig,
) *rateLimitManagerImpl {
```
either all parameters on the same line, or all on different lines

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		forceRefreshRateOnce sync.Once
+		// rateLimiter that limits the rate at which tasks can be dispatched to consumers
+		rateLimiter quotas.RateLimiter
+	}
+)
+
+var _ rateLimitManager = (*rateLimitManagerImpl)(nil)
+
+// Create a new rate limit manager for the task queue partition.
+// `workerRPS` is expected to be nil and will be set lazily via SetWorkerRPS at the time of worker poll.
```
dnr: if it's going to be nil, why take it as a parameter?

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		),
+	})
+	r.computeEffectiveRPSAndSource(defaultTaskDispatchRPS)
+	return r
+}
+
+// Compute the effective RPS and source based on the task dispatch rate.
+// This method sets the `effectiveRPS` and `rateLimitSource` based on the provided `taskDispatchRate`
+// `taskDispatchRate` can be the rate from API, worker or system defaults.
+func (r *rateLimitManagerImpl) computeEffectiveRPSAndSource(taskDispatchRate float64) {
```
dnr: I'm a little confused.. isn't this supposed to compute the effective rate limit? What is the value that it takes as input?

ok, I read more and it makes sense. but I think this is kind of awkward.

first, using "task dispatch rate" and "rate limit" and "rps" is very confusing. they're all the same thing, use the same naming scheme consistently.

second, you shouldn't reverse-engineer the "source" based on which value it happened to match, the source should come from the control flow: each of those three calls to computeEffectiveRPSAndSource corresponds to a source, so they should pass it in. then the "system" limit from dynamic config can override the api or worker source if it's less than those. do that with control flow too, e.g. `if systemRPS < rps { rps = systemRPS; source = system }`

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+// The source can be API, worker or system default.
+// By default, it returns RATE_LIMIT_SOURCE_SYSTEM (default value).
+func (r *rateLimitManagerImpl) GetSourceForEffectiveRPS() enumspb.RateLimitSource {
+	return enumspb.RateLimitSource(r.rateLimitSource)
+}
+
+// Return the effective RPS for task queues.
+// Whenever the effective RPS is retrieved, it needs to be computed based on the current configuration.
+// It checks for API configured RPS, worker set RPS, and computes the effective rate limit.
+func (r *rateLimitManagerImpl) GetEffectiveRPS(tqType enumspb.TaskQueueType) float64 {
```
dnr: there should be one function that returns the effective rps and its source together. otherwise two calls could be out of sync.. the source could be for a different value than the rps that was retrieved a second ago

dnr: this is a `Get...` call, but it also modifies state. in general you usually want to keep functions to one or the other: called for side effects with no return value, or returns a value but has no side effects. there are exceptions, but generally things are easier to understand if you stick to that.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+// SelectTaskQueueRateLimiter returns the user defined RPS (configured RPS) if exists/system default along with an update flag.
+// Configured RPS may or may not be equal to the effective RPS.
+// If the RPS set via API or workerOptions is lesser than the system defaults,
+// only then configured RPS will be equal to effective RPS.
+func (r *rateLimitManagerImpl) SelectTaskQueueRateLimiter(tqType enumspb.TaskQueueType) (*wrapperspb.DoubleValue, bool) {
+	var configuredRPS *wrapperspb.DoubleValue
+	var updateRequired bool
+	// Try setting RPS from user data (API-configured rate limit).
```
dnr: I'm not quite sure what this is used for yet, but I can see that it's duplicating logic from GetEffectiveRPS. the point of this manager object is to consolidate that logic (which one has precedence, etc.) to exactly one place in the code.

service/matching/ratelimit_manager.go
dnr: you shouldn't be passing around tqType at all, a rateLimitManagerImpl belongs to a partition which already has a type, i.e. the type is known and constant at creation time.

service/matching/task_queue_partition_manager.go
```diff
@@ -381,13 +341,20 @@ func (pm *taskQueuePartitionManagerImpl) PollTask(
...
+	// SelectTaskQueueRateLimiter decides whether the RPS needs to be updated.
+	// Highest priority is given to the rate limit set by the UpdateTaskQueueConfig api call.
+	// Followed by the rate limit set by the poller.
+	// If no rate limit is set, updateRequired will be false and we will continue to use the default rate limit.
+	pm.rateLimitManager.SetWorkerRPS(pollMetadata)
+	taskQueueType := pm.partition.TaskType()
+	effectiveRPSToBeSet, updateRequired := pm.rateLimitManager.SelectTaskQueueRateLimiter(taskQueueType)
+	if updateRequired {
+		pm.rateLimitManager.UpdateRatelimit(effectiveRPSToBeSet.Value)
 	}
```
dnr: this is a pretty clunky interface to notify the rateLimitManager of a new pollMetadata. the rateLimitManager should take care of all this, that's its job. we should just pass it pollMetadata and it does whatever it needs to do.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,209 @@
...
+	"math"
+	"sync"
+
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/server/common/quotas"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+type (
+	rateLimitManagerImpl struct {
```
dnr: how are you handling synchronization here? is rateLimitManagerImpl safe for use from multiple goroutines? does the caller of this have to do their own synchronization?

service/matching/ratelimit_manager_interface.go
sivagirish81: We are not mocking this as of now. So got rid of the interface.

service/matching/ratelimit_manager.go
sivagirish81: Implemented.

service/matching/ratelimit_manager.go
sivagirish81: Noted.

service/matching/ratelimit_manager.go
sivagirish81: I don't see any use case as of now where we might need to set the worker/api RPS at the time of creating a newRateLimitManager. So removed the parameter.

service/matching/ratelimit_manager.go
sivagirish81: Agreed. Fixed to retrieve source and effectiveRPS together.

service/matching/matching_engine_test.go
```diff
@@ -1135,8 +1135,9 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 	s.NotNil(descResp.DescResponse.GetTaskQueueStatus())
 	numPartitions := float64(s.matchingEngine.config.NumTaskqueueWritePartitions("", "", tlType))
 	//nolint:staticcheck // checking deprecated field
-	s.GreaterOrEqual(descResp.DescResponse.GetTaskQueueStatus().GetRatePerSecond()*numPartitions,
-		(defaultTaskDispatchRPS - 1))
+	// The effective task queue rate should be at least 24999 RPS,
+	// given that both system-level rate limits are overridden to 25000.
+	s.GreaterOrEqual(descResp.DescResponse.GetTaskQueueStatus().GetRatePerSecond()*numPartitions, 24999.0)
```
sivagirish81: Initially I was not accounting for partitions while setting the effectiveRateLimit. Due to which this test was breaking. So once I accounted for the partitions while setting the workerRPS and apiRPS the test was fixed.

service/matching/ratelimit_manager.go
sivagirish81: Added a sync mutex to handle data race and synchronization for the rate limit manager.

service/matching/task_queue_partition_manager.go
sivagirish81: Agreed to make the interface more extensible. I handled the updateRateLimit to check if any update is required and accordingly update the rate limit. Therefore, this can be reused independently.

service/matching/ratelimit_manager.go
sivagirish81: Fixed this to store the taskQueueType in the RateLimitManager to avoid passing the value.

service/matching/ratelimit_manager.go
sivagirish81: Removed this post refactoring the code.

service/matching/ratelimit_manager.go
sivagirish81: Fixed the logic to use the source as suggested instead of reverse engineering the rate limit.

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Millisecond)
+
+	var (
+		mu       sync.Mutex
+		runTimes []time.Time
+	)
+
+	const (
+		workerRPS    = 50.0
+		drainTimeout = 30 * time.Second // Setting extremely high drain timeout to prevent flaking
```
sivagirish81: A drain timeout of 30s is still not enough for one of the functional test to pass. Will need to re-evaluate this.
Ideal case the test should not take so long so in most cases we should be good but yeah worst case as observed it is exceeding the limit. Not sure if it is a right thing to increase it even more.
**Update :** Removed the additional tests for different rate limit values which were causing the test to flake as they are not important for checking the api config override.

service/matching/ratelimit_manager.go
sivagirish81: Agreed. Get calls should ideally not make changes to the state. 
Added a new function `GetEffectiveRPSAndSource` that gets both the effectiveRPS and source together. This function is tightly coupled with the `computeEffectiveRPSAndSourceLocked` as we need to always compute the effective RPS before retrieving the values. So inline this call instead the get. Not sure if this is the best way to do it.

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
 
+func (s *TaskQueueSuite) TestTaskQueueApiLimitOverride() {
+	const buffer = 100 * time.Millisecond
+
+	s.Run("RateLimitTest_1.00", func() {
+		expectedGap := time.Second
+		s.runTestTaskQueueAPIRateLimitOverridesWorkerLimit(1.0, 5, expectedGap-buffer, expectedGap+buffer, "RateLimitTest_1.00")
+	})
+
+	s.Run("RateLimitTest_2.00", func() {
```
sivagirish81: Keeping the test simple to just check the override using 1 RPS with 1 partition and not checking the enforcing of different rate limits. There are other tests to check the same.

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	)
+
+	const (
+		workerRPS    = 50.0
+		drainTimeout = 30 * time.Second // Setting extremely high drain timeout to prevent flaking
+		activityName = "trackableActivity"
+	)
+
+	tv := testvars.New(s.T())
+	tv = tv.WithNamespaceName(namespace.Name(fmt.Sprintf("ns-rps-%.2f", apiRPS))) // To isolate state and avoid flaking
```
stephanos: I don't think this has any effect; since you need to use `s.Namespace().String()` anyway. (that's the namespace the test framework sets up for you)

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	// Wait for all activities to complete
+	s.Eventually(func() bool {
+		mu.Lock()
+		defer mu.Unlock()
+		return len(runTimes) >= taskCount
+	}, drainTimeout, 100*time.Millisecond)
+
+	// Validate that API RPS was enforced
+	mu.Lock()
+	defer mu.Unlock()
```
stephanos: This should be unnecessary since you've already checked that all activities completed.

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+		ctx = workflow.WithActivityOptions(ctx, ao)
+		return workflow.ExecuteActivity(ctx, activityName).Get(ctx, nil)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), drainTimeout)
+	defer cancel()
+
+	// Start the activity worker
+	activityWorker := worker.New(s.SdkClient(), activityTaskQueue, worker.Options{
+		TaskQueueActivitiesPerSecond: workerRPS,
```
stephanos, line 243: Maybe add a little comment here that just says `// setting worker limit here`

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	activityFunc := func(context.Context) error {
+		mu.Lock()
+		defer mu.Unlock()
+		runTimes = append(runTimes, time.Now())
+		return nil
+	}
+
+	workflowFn := func(ctx workflow.Context) error {
+		ao := workflow.ActivityOptions{
+			TaskQueue:           activityTaskQueue, // Route activities to a dedicated task queue named `activityTaskQueue`
```
stephanos: Maybe reword the comment to mention _why_ you're doing this instead.

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	s.Run("RateLimitTest_2.00", func() {
+		expectedGap := 500 * time.Millisecond
+		s.runTestTaskQueueAPIRateLimitOverridesWorkerLimit(2.0, 5, expectedGap-buffer, expectedGap+buffer, "RateLimitTest_2.00")
+	})
+}
+
+func (s *TaskQueueSuite) runTestTaskQueueAPIRateLimitOverridesWorkerLimit(apiRPS float32, taskCount int, minGap time.Duration, maxGap time.Duration, activityTaskQueue string) {
+	// Set the burst as 1 to make sure not more than 1 task get's acknowledged at a time.
+	// Helps observe the backlog drain more easily.
+	s.OverrideDynamicConfig(dynamicconfig.MatchingMinTaskThrottlingBurstSize, 1)
```
stephanos: Can you set it to 0 to avoid the special case at the bottom of the test?

tests/task_queue_test.go
```diff
@@ -164,6 +167,129 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+		}, workflowFn)
+		s.NoError(err)
+	}
+
+	// Wait for all activities to complete
+	s.Eventually(func() bool {
+		mu.Lock()
+		defer mu.Unlock()
+		return len(runTimes) >= taskCount
+	}, drainTimeout, 100*time.Millisecond)
```
stephanos: I think a [wait group](https://gobyexample.com/waitgroups) is nicer and simpler here.

tests/task_queue_test.go
sivagirish81: Done.

tests/task_queue_test.go
sivagirish81: Burst is overriden regardless of the value we set here at time of ratelimitupdate.
```
minTaskThrottlingBurstSize := r.config.MinTaskThrottlingBurstSize()
	if burst < minTaskThrottlingBurstSize {
		burst = minTaskThrottlingBurstSize
	}
```
So we will need to enforce the special case at the bottom of the test - ignore the gap between the first and the second task and check for the gap between the remaining tasks.

tests/task_queue_test.go
sivagirish81: Agreed. Removed.

tests/task_queue_test.go
sivagirish81: Replaced with wait group.

tests/task_queue_test.go
sivagirish81: Done.

tests/task_queue_test.go
sivagirish81: Removed the namespace override.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,185 @@
...
+		),
+		quotas.NewDefaultOutgoingRateLimiter(
+			tqConfig.AdminNamespaceToPartitionDispatchRate,
+		),
+	})
+	r.computeEffectiveRPSAndSourceLocked()
+	return r
+}
+
+func (r *rateLimitManager) computeEffectiveRPSAndSourceLocked() {
```
dnr: the naming convention is backwards: "Locked" in a function name means it assumes the lock is already held

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,185 @@
...
+
+// Lazy injection of poll metadata.
+// Internally call UpdateRateLimit to share the same mutex.
+func (r *rateLimitManager) InjectWorkerRPS(meta *pollMetadata) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if meta != nil && meta.taskQueueMetadata != nil {
+		r.workerRPS = meta.taskQueueMetadata.GetMaxTasksPerSecond()
+	}
+	r.UpdateRatelimit()
```
dnr: if we haven't changed anything, don't bother doing any more work
```suggestion
		r.UpdateRatelimit()
	}
```

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Return the effective RPS and its source together.
+// The source can be API, worker or system default.
+// It defaults to the system dynamic config.
+func (r *rateLimitManager) GetEffectiveRPSAndSource() (float64, enumspb.RateLimitSource) {
+	r.computeEffectiveRPSAndSourceLocked()
+	return r.effectiveRPS, r.rateLimitSource
```
dnr, line 129: you can't read these outside the lock

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,185 @@
...
+		return
+	}
+	taskQueueData := userData.GetData()
+	if taskQueueData == nil || taskQueueData.GetPerType() == nil {
+		return
+	}
+	taskQueueTypeData := taskQueueData.GetPerType()[int32(r.taskQueueType)]
+	if taskQueueTypeData == nil || taskQueueTypeData.GetConfig() == nil {
+		return
+	}
```
dnr: use the getters to simplify things:

```suggestion
	if err != nil {
		return
	}
	config := userData.GetData().GetPerType()[int32(r.taskQueueType)].GetConfig()
	if config == nil {
		return
	}
```

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,185 @@
...
+
+type (
+	rateLimitManager struct {
+		mu sync.Mutex
+
+		effectiveRPS    float64                 // Min of api/worker set RPS and system defaults.
+		rateLimitSource enumspb.RateLimitSource // Source of the rate limit, can be set via API, worker or system default.
+		userDataManager userDataManager         // User data manager to fetch user data for task queue configurations.
+		workerRPS       *wrapperspb.DoubleValue // RPS set by worker at the time of polling, if available.
+		apiConfigRPS    *wrapperspb.DoubleValue // RPS set via API, if available.
```
dnr: why use wrapperspb instead of just `*float32`? or to be a little more clever it could just be `float32` with NaN representing "not set".

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,185 @@
...
+	// In this case, the apiConfigRPS will need to be unset.
+	rateLimit := taskQueueTypeData.Config.GetQueueRateLimit()
+	if rateLimit == nil || rateLimit.GetRateLimit() == nil {
+		r.apiConfigRPS = nil
+		return
+	}
+	r.apiConfigRPS = wrapperspb.Double(float64(rateLimit.GetRateLimit().GetRequestsPerSecond()))
+}
+
+// UpdateRatelimit checks and updates the rate limit if changed.
```
dnr: is this called with the lock held or not?

service/matching/ratelimit_manager.go
sivagirish81: Fixed to follow convention.

service/matching/ratelimit_manager.go
sivagirish81: I wanted to use the same lock to update the worker RPS and call the updateRateLimit. The catch here is that regardless of whether the workerRateLimit is being updated we need to make the call to updateRateLimit because the source of the rate limit can now also be the api. 
So I am forcing the updateRateLimit to be called. Within the updateRateLimitLocked I have introduced a check to see if there is any change in the effective rate limit.
So this call is safe from unnecessary updates.

service/matching/ratelimit_manager.go
sivagirish81: Fixed it to read within the lock.

service/matching/ratelimit_manager.go
sivagirish81: Implemented.

service/matching/ratelimit_manager.go
sivagirish81: I wanted to be able to set the workerRPS and apiConfigRPS specifically to nil to indicate no value has been configured. But can explore changing it to float32. I feel a nil will be more clearer to indicate no config set.

service/matching/ratelimit_manager.go
sivagirish81: It is always called with the lock held. Have updated the function name to be reflective of that.

service/matching/ratelimit_manager.go
dnr: Sure, but why not just a pointer to a float32? there are no protos here, why use a proto message?

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	tqConfig *taskQueueConfig,
+	tqType enumspb.TaskQueueType,
+) *rateLimitManager {
+	r := &rateLimitManager{
+		userDataManager: userDataManager,
+		workerRPS:       nil,
+		apiConfigRPS:    nil,
+		config:          tqConfig,
+		rateLimitSource: enumspb.RATE_LIMIT_SOURCE_SYSTEM,
+		taskQueueType:   tqType,
```
dnr: nit: code is easier to read and understand if you use the same name for the same thing consistently. as is, the reader has to associate two different names with this value, taskQueueType and tqType, and think about whether they represent the same thing or not. if you call the parameter taskQueueType also (or call the field tqType), then it's totally obvious you're just copying a value from one place to another.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Create a new rate limit manager for the task queue partition.
+func newRateLimitManager(userDataManager userDataManager,
+	tqConfig *taskQueueConfig,
+	tqType enumspb.TaskQueueType,
+) *rateLimitManager {
+	r := &rateLimitManager{
+		userDataManager: userDataManager,
+		workerRPS:       nil,
+		apiConfigRPS:    nil,
```
dnr: you don't need to initialize fields to nil explicitly, that's the zero value

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+		config:          tqConfig,
+		rateLimitSource: enumspb.RATE_LIMIT_SOURCE_SYSTEM,
```
dnr: you don't need to assign this here, it'll be assigned by r.computeEffectiveRPSAndSource() below

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	)
+	r.rateLimiter = quotas.NewMultiRateLimiter([]quotas.RateLimiter{
+		r.dynamicRateLimiter,
+		quotas.NewDefaultOutgoingRateLimiter(
+			tqConfig.AdminNamespaceTaskQueueToPartitionDispatchRate,
+		),
+		quotas.NewDefaultOutgoingRateLimiter(
+			tqConfig.AdminNamespaceToPartitionDispatchRate,
+		),
+	})
```
dnr: we don't need to do this MultiRateLimiter at all, right? we're already doing the "min of user-set limit and dynamic config limits", which is the same as what the multiratelimiter would do, so we can just use the single DynamicRateLimiter?

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+		r.rateLimitSource = enumspb.RATE_LIMIT_SOURCE_SYSTEM
+	}
+}
+
+// Lazy injection of poll metadata.
+// Internally call UpdateRateLimit to share the same mutex.
+func (r *rateLimitManager) InjectWorkerRPS(meta *pollMetadata) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	nPartitions := float64(r.config.NumReadPartitions())
```
dnr: I see why you changed this, but I think we need to do it differently...

first, the "effective RPS" that we report in DescribeTaskQueue should be overall, not per-partition. so I think we should keep track of everything as overall limits, and then only divide by partitions at the last possible moment (in updateratelimitlocked). but yes, the limits from dynamic config are per-partition. so we can keep track of an additional flag that says whether to divide or not.

second, the logic to do the division is now copied in two locations. it should be factored into a function.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	r.computeEffectiveRPSAndSourceLocked()
+	return r.effectiveRPS, r.rateLimitSource
+}
+
+func (r *rateLimitManager) GetRateLimiter() quotas.RateLimiter {
+	return r.rateLimiter
+}
+
+// TrySetRPSFromUserData attempts to set effectiveRPS from user data.
+func (r *rateLimitManager) TrySetRPSFromUserDataLocked() {
```
dnr: what is going to ensure this gets called if userdata changes? actually we have a callback for this,  you should call something here from taskQueuePartitionManagerImpl.userDataChanged()

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+		return
+	}
+	config := userData.GetData().GetPerType()[int32(r.taskQueueType)].GetConfig()
+	if config == nil {
+		return
+	}
+	// If rate limit is an empty message, it means rate limit could have been unset via API.
+	// In this case, the apiConfigRPS will need to be unset.
+	rateLimit := config.GetQueueRateLimit()
+	if rateLimit == nil || rateLimit.GetRateLimit() == nil {
```
dnr:
```suggestion
	if rateLimit.GetRateLimit() == nil {
```

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	if nPartitions > 0 {
+		r.apiConfigRPS = wrapperspb.Double(float64(rateLimit.GetRateLimit().GetRequestsPerSecond()) / nPartitions)
+		return
+	}
+	// Defaulting to 1 partition in case nPartitions is misconfigured.
+	r.apiConfigRPS = wrapperspb.Double(float64(rateLimit.GetRateLimit().GetRequestsPerSecond()))
+}
+
+// UpdateRatelimit checks and updates the rate limit if changed.
+func (r *rateLimitManager) UpdateRatelimitLocked() {
```
dnr: nit: even though we're all in the same package so it doesn't really matter, let's name methods that are intended to be called by other code with capital letters, and "internal" methods just for this object with lowercase. so this could be `updateRateLimitLocked` (and check various others)

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	if oldRPS == newRPS {
+		// No update required
+		return
+	}
+	burst := int(math.Ceil(newRPS))
+
+	minTaskThrottlingBurstSize := r.config.MinTaskThrottlingBurstSize()
+	if burst < minTaskThrottlingBurstSize {
+		burst = minTaskThrottlingBurstSize
+	}
```
dnr:
```suggestion
	burst = max(burst, r.config.MinTaskThrottlingBurstSize())
```

tests/task_queue_test.go
```diff
@@ -164,6 +166,130 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// All activities completed successfully within drainTimeout
+	case <-ctx.Done():
+		s.FailNow("Timeout waiting for activities to complete")
+	}
```
```suggestion
	s.True(common.AwaitWaitGroup(&wg, drainTimeout), "timeout waiting for activities to complete")
```

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,198 @@
...
+	nPartitions := float64(r.config.NumReadPartitions())
+	if nPartitions > 0 {
+		r.apiConfigRPS = wrapperspb.Double(float64(rateLimit.GetRateLimit().GetRequestsPerSecond()) / nPartitions)
+		return
+	}
+	// Defaulting to 1 partition in case nPartitions is misconfigured.
+	r.apiConfigRPS = wrapperspb.Double(float64(rateLimit.GetRateLimit().GetRequestsPerSecond()))
+}
+
+// UpdateRatelimit checks and updates the rate limit if changed.
```
dnr: update the function comment if you change the name of the function

service/matching/ratelimit_manager.go
sivagirish81: Agreed this was a mistake. The poll metadata had the taskdispatchrate as a proto message which I tried to reuse  -  which is not correct. So changed the implementation to use a pointer to a float32 instead.

service/matching/ratelimit_manager.go
sivagirish81: Noted and implemented.

service/matching/ratelimit_manager.go
sivagirish81: removed the unnecessary initializations.

service/matching/ratelimit_manager.go
sivagirish81: Removed the redundant assignment.

service/matching/ratelimit_manager.go
sivagirish81: Agreed. Made the changes to use the single dynamic rate limiter.

service/matching/ratelimit_manager.go
sivagirish81: Moved this to the taskQueuePartitionManagerImpl.userDataChanged() as suggested.

service/matching/ratelimit_manager.go
sivagirish81: Done.

service/matching/ratelimit_manager.go
sivagirish81: Storing and tracking all the three sources of RPS as the overall queue limit in the rateLimitManager.
Tracking the systemRPS as the 
```
systemRPS := min(
		config.AdminNamespaceTaskQueueToPartitionDispatchRate(),
		config.AdminNamespaceToPartitionDispatchRate(),
	) * r.getNumberOfReadPartitions()
```
This will make sure that the RPS from all 3 sources in the rateLimitManager are on the overallTaskQueue rather than a single partition.
Now the division only happens inside the updateRateLimit at the last moment right before the rate limit update.
This will also solve the problem of reporting the data at DescribeTaskQueue level  - even the systemRPS will be reported correctly as the overall queue limit.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,193 @@
...
+		int(defaultTaskDispatchRPS),
+	)
+	r.dynamicRateLimiter = quotas.NewDynamicRateLimiter(
+		r.dynamicRateBurst,
+		config.RateLimiterRefreshInterval,
+	)
+	// Overall system rate limit will be the min of the two configs that are partition wise times the number of partions.
+	systemRPS := min(
+		config.AdminNamespaceTaskQueueToPartitionDispatchRate(),
+		config.AdminNamespaceToPartitionDispatchRate(),
```
dnr: oh hmm, I guess the advantage of the multiratelimiter was that it would adjust if these were changed.

we can do similar by reading them in computeEffectiveRPSAndSourceLocked each time so that we respond if they're changed (we don't use them often but just for correctness). then we should use subscriptions to get a callback on change, then, but we can do that part in a follow-up PR

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,193 @@
...
+
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/server/common/quotas"
+)
+
+type (
+	rateLimitManager struct {
+		mu sync.Mutex
+
+		effectiveRPS    *float64                // Min of api/worker set RPS and system defaults. Always reflects the overall task queue RPS.
```
dnr: why is this a pointer? we always have a value here, it's not optional

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,193 @@
...
+		rateLimitSource enumspb.RateLimitSource // Source of the rate limit, can be set via API, worker or system default.
+		userDataManager userDataManager         // User data manager to fetch user data for task queue configurations.
+		workerRPS       *float64                // RPS set by worker at the time of polling, if available.
+		apiConfigRPS    *float64                // RPS set via API, if available.
+		systemRPS       *float64                // min of partition level dispatch rates times the number of read partitions.
```
dnr: same here, this one is always present

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,193 @@
...
+	rateLimit := config.GetQueueRateLimit()
+	if rateLimit.GetRateLimit() == nil {
+		r.apiConfigRPS = nil
+		return
+	}
+	val := float64(rateLimit.GetRateLimit().GetRequestsPerSecond())
+	r.apiConfigRPS = &val
+}
+
+func (r *rateLimitManager) getEffectiveRateLimitPartitionWiseLocked() float64 {
```
dnr: do you really need a function for this one line that's called only once?

service/matching/ratelimit_manager.go
sivagirish81: Oh yes missed the part about the dynamicConfig moved it inside computeEffectiveRPSAndSourceLocked to recompute each time.
In the next PR I will try to migrate to the subscriptions similar to how it is done in pri_matcher.

service/matching/ratelimit_manager.go
sivagirish81: Agreed. Changed to float.

service/matching/ratelimit_manager.go
sivagirish81: Agreed changed to float.

service/matching/ratelimit_manager.go
sivagirish81: Removed the function and inlined it.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// Computes the effectiveRPS and its source by evaluating all possible rate limit configurations.
+// - If an API-level RPS is configured, effectiveRPS = min(system default RPS, API-configured RPS)
+// - Else if a worker-level RPS is configured, effectiveRPS = min(system default RPS, worker-configured RPS)
+// - Otherwise, fall back to the system default RPS from dynamic config.
+func (r *rateLimitManager) computeEffectiveRPSAndSourceLocked() {
+
+	var (
+		effectiveRPS    float64
```
sivagirish81: Will initialize effectiveRPS as +Inf.

service/matching/task_queue_partition_manager.go
```diff
@@ -1233,4 +1194,6 @@ func (pm *taskQueuePartitionManagerImpl) userDataChanged() {
 
 	// Do this one in this goroutine.
 	pm.defaultQueue.UserDataChanged()
+	// Update rateLimits if any change is userData.
```
sivagirish81: Will move this to the top of the function.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	// If rate limit is an empty message, it means rate limit could have been unset via API.
+	// In this case, the apiConfigRPS will need to be unset.
+	rateLimit := config.GetQueueRateLimit()
+	if rateLimit.GetRateLimit() == nil {
+		r.apiConfigRPS = nil
+		return
+	}
+	val := float64(rateLimit.GetRateLimit().GetRequestsPerSecond())
+	r.apiConfigRPS = &val
```
dnr, line 163: maybe we could do this to avoid an allocation each time?
```suggestion
	r.apiConfigRPS = &rateLimit.RateLimit.RequestsPerSecond
```
if it's a float64/32 thing, maybe change the types in here? we don't need even 32 bits of precision for any of this

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,188 @@
...
+
+	var (
+		effectiveRPS    float64
+		rateLimitSource enumspb.RateLimitSource
+	)
+	// Overall system rate limit will be the min of the two configs that are partition wise times the number of partions.
+	systemRPS := min(
+		r.config.AdminNamespaceTaskQueueToPartitionDispatchRate(),
+		r.config.AdminNamespaceToPartitionDispatchRate(),
+	) * r.getNumberOfReadPartitions()
```
dnr, line 84: there's a potential issue here if we read getNumberOfReadPartitions() and multiply by that, and then that number changes and we divide by a different number. it's rare for the partition count to change now, so not a big deal, but still worth handling.

when we add subscriptions on these we should probably add a subscription on read partitions too so that we force recalculating everything if it changes.

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,188 @@
...
+	if oldRPS == newRPS {
+		// No update required
+		return
+	}
+	effectiveRPSPartitionWise := r.effectiveRPS / r.getNumberOfReadPartitions()
+	burst := int(math.Ceil(effectiveRPSPartitionWise))
+	burst = max(burst, r.config.MinTaskThrottlingBurstSize())
+	r.dynamicRateBurst.SetRPS(effectiveRPSPartitionWise)
+	r.dynamicRateBurst.SetBurst(burst)
+	r.forceRefreshRateOnce.Do(func() {
```
dnr, line 182: I thought there was a subtle problem here where we would spend our "Once" on the first update from either config or from polls, and then fail to respond to the first update from the other one. I think it turns out that's not really a problem, since the config should come before any polls, and we want the config to take precedence.

anyway, I'm thinking we should get rid of the Once and just always Refresh here: we already checked if the value changed, so we only get here if we have a new value, and then why wouldn't we want it to be applied immediately? the only situation might be if workers disagree and are sending values such that it flips back and forth.

I guess we can leave it for now until we consolidate on just using simpleLimiters?

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// updateRatelimitLocked checks and updates the rate limit if changed.
+func (r *rateLimitManager) updateRatelimitLocked() {
+	// Always check for api configured rate limits before any update to the rate limits.
+	r.trySetRPSFromUserDataLocked()
+	oldRPS := r.effectiveRPS
+	r.computeEffectiveRPSAndSourceLocked()
+	newRPS := r.effectiveRPS
+	if oldRPS == newRPS {
```
dnr, line 173: I think we should compute and compare the value _after_ dividing by partitions, since number of partitions can change also

service/matching/ratelimit_manager.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	val := float64(rateLimit.GetRateLimit().GetRequestsPerSecond())
+	r.apiConfigRPS = &val
+}
+
+// updateRatelimitLocked checks and updates the rate limit if changed.
+func (r *rateLimitManager) updateRatelimitLocked() {
+	// Always check for api configured rate limits before any update to the rate limits.
+	r.trySetRPSFromUserDataLocked()
+	oldRPS := r.effectiveRPS
```
dnr, line 170: on the very first call, what will this be? I suppose 0. I think we should initialize it to -1 or something so that if the configured rps is 0, then we'll see it as a "change" on the first call. otherwise we'll leave the dynamicRateLimiter set to the default (10k)

tests/task_queue_test.go
```diff
@@ -164,6 +167,119 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
 	return "OldMatching_" + ret
 }
 
+func (s *TaskQueueSuite) TestTaskQueueApiLimitOverride() {
+	const buffer = 100 * time.Millisecond
+
+	s.Run("RateLimitTest_1.00", func() {
+		expectedGap := time.Second
+		s.runTestTaskQueueAPIRateLimitOverridesWorkerLimit(1.0, 5, expectedGap-buffer, expectedGap+buffer, "RateLimitTest_1.00")
+	})
```
dnr: why is this using a subtest? are we ever going to run it with different parameters?

tests/task_queue_test.go
```diff
@@ -164,6 +167,119 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+}
+
+func (s *TaskQueueSuite) runTestTaskQueueAPIRateLimitOverridesWorkerLimit(apiRPS float32, taskCount int, minGap time.Duration, maxGap time.Duration, activityTaskQueue string) {
+	// Set the burst as 1 to make sure not more than 1 task get's acknowledged at a time.
+	// Helps observe the backlog drain more easily.
+	s.OverrideDynamicConfig(dynamicconfig.MatchingMinTaskThrottlingBurstSize, 1)
```
dnr: this is the default, why are you overriding it?

tests/task_queue_test.go
```diff
@@ -164,6 +167,119 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	// allowing the first task to dispatch instantly. The second task may also be dispatched quickly,
+	// as the rate limit (e.g., 2 RPS) still allows another token within the same second.
+	// To account for this initial "microburst," we skip validation of the timing between the first two tasks
+	// and begin enforcing the rate limit check from the third task onward.
+	// Always start from 2nd index to avoid the effect of burst for consistency
+	startIdx := 2
+	for i := startIdx; i < len(runTimes); i++ {
+		diff := runTimes[i].Sub(runTimes[i-1])
+		s.GreaterOrEqual(diff, minGap, "Activity ran too quickly between executions")
+		s.LessOrEqual(diff, maxGap, "Activity ran too slowly between executions")
```
dnr: I'm concerned this is going to be flaky in CI, with only a 100ms buffer. maybe we should do it more like, set a rate of 5/second and run 26 activities, check that the time between the first and second is (26-1)/5=5s +/- 1s ? then the fluctuations kind of average out, but we still check the overall rate is the one from the config.

service/matching/ratelimit_manager.go
dnr: (we could add a test for this.. configure the limit be 0 and then ensure we can't get a task within a few seconds)

dnr (commented):
alright, a few small comments. the partition count stuff could be handled in the next pr. what do you think about the others?

service/matching/ratelimit_manager.go
sivagirish81: The dynamic configs are in Float64 and all the rate limiter implementations also use float64. Changing it will involve changing everything to float32. Do we want to do this?
This is why I was considering making changes in the api repo to use double for the request per seconds type to ensure everything is float64.
We will need to do this typecasting as long as we don't confirm to float64/float32 for all requests per second fields. So I was thinking of keeping this as it is.

service/matching/ratelimit_manager.go
sivagirish81: On the very first call when there is no worker set rate limit or api set rate limit the value of effectiveRPS will be the `systemRPS` i.e the system dynamic config itself and not 0 as we are computing effective RPS when we call `newRateLimitManager`. So we will see the change even if a 0 rate limit is set. 

> (we could add a test for this.. configure the limit be 0 and then ensure we can't get a task within a few seconds)

So in this case we will need to additionally override the burst and specifically set it to 0. Followed by which we will need to check if no tasks are getting picked up for a reasonably time. 
Effectively by setting the task queue rate to 0 alone we cannot pause tasks being dispatched from the queue. We will need to override the burst as well. Will add a separate test for this and handle it in the follow up PR.

tests/task_queue_test.go
sivagirish81: Initially I had written a few more tests with different rate limits hence I made it a subtest.
Changed this from a subtest. The test for 0 rate limit will behave differently from this so will try to keep it separate.

tests/task_queue_test.go
sivagirish81: The default behaviour : 
`Burst =  max(int(math.Ceil(effectiveRPSPartitionWise)), r.config.MinTaskThrottlingBurstSize())`
So whenever we set the RPS to 5 the burst will default to the value of at least the configured rate limit regardless of the value configured for the burst.
So if we spawn 25 activity tasks the first 5 tasks will get picked immediately due to the burst.
The rest of the tasks will have the rate limit enforced. So the overall expected time for the tasks to run is expected to be around 4s ((25 - 5)/5) . Changed the test to compare the total execution time with a buffer of 1s.

tests/task_queue_test.go
sivagirish81: Removed the unnecessary override.

service/matching/ratelimit_manager.go
sivagirish81: Will handle this as suggested in the next PR along with switching partition count as well to subscriptions.

service/matching/ratelimit_manager.go
sivagirish81: Agreed. Will handle this in the next PR as suggested.

service/matching/ratelimit_manager.go
dnr: ok, it's fine for now. we can try to clean it up later.

service/matching/ratelimit_manager.go
dnr: something still feels slightly wrong here. how about this scenario:

dynamic config is set to 10 tasks/s. worker does not set rate limit, so InjectWorkerRPS is never called. config and user data are not set, so UserDataChanged is not called. I think that means updateRatelimitLocked will never be called, right? so the rate limit stays at 100k.

the problem seems to be that the "compute" and "set" are separate. why should we "compute" and store a value if we're not going to "set" it in the rate limiter? why can't we always do both? it seems like if we consolidate those things then it'll all get simpler.

tests/task_queue_test.go
```diff
@@ -164,6 +167,109 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+		ctx = workflow.WithActivityOptions(ctx, ao)
+		return workflow.ExecuteActivity(ctx, activityName).Get(ctx, nil)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), drainTimeout)
+	defer cancel()
+
+	// Start the activity worker
+	activityWorker := worker.New(s.SdkClient(), activityTaskQueue, worker.Options{
+		// Setting rate limit at worker level
```
```suggestion
		// Setting rate limit at worker level (this will be ignored in favor of the limit set through the api)
```

tests/task_queue_test.go
```diff
@@ -164,6 +167,109 @@ func (s *TaskQueueSuite) testTaskQueueRateLimitName(nPartitions, nWorkers int, u
...
+	defer activityWorker.Stop()
+
+	// Start the workflow worker
+	wfWorker := worker.New(s.SdkClient(), tv.TaskQueue().GetName(), worker.Options{})
+	wfWorker.RegisterWorkflow(workflowFn)
+	s.NoError(wfWorker.Start())
+	defer wfWorker.Stop()
+
+	// Launch workflows
+	for i := 0; i < taskCount; i++ {
```
```suggestion
	for i := range taskCount {
```

service/matching/ratelimit_manager.go
dnr: ah I see, InjectWorkerRPS is called even if a worker rate limit is not set, so it will get set before any polls happen.

I'd still like to simplify this more but let's do it in the next pr.

dnr (approved):

PR #8059, opened 2025-07-16
add expired tasks state to the metric

ychebotarev:
## What changed?
add "task expire state" tag to the "ExpiredTasksPerTaskQueueCounter" metric.
Possible states - "read", "process", "memory"

## Why?
For fairness, we remove TTL. We expect that this will lead to reading too many expired tasks.
To be sure we want a metrics tag that will allows us to separate tasks that are expired when we read them from other tasks. 

## How did you test it?
- [X] built

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -479,14 +479,18 @@ func ToUnversionedTag(version string) Tag {
 	return &tagImpl{key: toUnversioned, value: falseValue}
 }
 
+var taskExpireStageReadTag = tagImpl{key: taskExpireStage, value: "read"}
+var taskExpireStageMemoryTag = tagImpl{key: taskExpireStage, value: "memory"}
+var taskExpireStageInvalidTag = tagImpl{key: taskExpireStage, value: "invalid"}
+
 func TaskExpireStageReadTag() Tag {
-	return &tagImpl{key: taskExpireStage, value: "read"}
+	return &taskExpireStageReadTag
 }
 
 func TaskExpireStageMemoryTag() Tag {
-	return &tagImpl{key: taskExpireStage, value: "memory"}
+	return &taskExpireStageMemoryTag
 }
 
 func TaskExpireStageValidateTag() Tag {
-	return &tagImpl{key: taskExpireStage, value: "invalid"}
+	return &taskExpireStageInvalidTag
 }
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -479,18 +479,6 @@ func ToUnversionedTag(version string) Tag {
 	return &tagImpl{key: toUnversioned, value: falseValue}
 }
 
-var taskExpireStageReadTag = tagImpl{key: taskExpireStage, value: "read"}
-var taskExpireStageMemoryTag = tagImpl{key: taskExpireStage, value: "memory"}
-var taskExpireStageInvalidTag = tagImpl{key: taskExpireStage, value: "invalid"}
-
-func TaskExpireStageReadTag() Tag {
-	return &taskExpireStageReadTag
-}
-
-func TaskExpireStageMemoryTag() Tag {
-	return &taskExpireStageMemoryTag
-}
-
-func TaskExpireStageValidateTag() Tag {
-	return &taskExpireStageInvalidTag
-}
+var TaskExpireStageReadTag Tag = &tagImpl{key: taskExpireStage, value: "read"}
+var TaskExpireStageMemoryTag Tag = &tagImpl{key: taskExpireStage, value: "memory"}
+var TaskInvalidTag Tag = &tagImpl{key: taskExpireStage, value: "invalid"}
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -41,6 +41,7 @@ const (
 	// See server.api.enums.v1.ReplicationTaskType
 	replicationTaskType     = "replicationTaskType"
 	replicationTaskPriority = "replicationTaskPriority"
+	taskExpireState         = "task_expire"
 	versioningBehavior      = "versioning_behavior"
 	isFirstAttempt          = "first-attempt"
 	workflowStatus          = "workflow_status"
@@ -476,4 +477,8 @@ func ToUnversionedTag(version string) Tag {
 		return &tagImpl{key: toUnversioned, value: trueValue}
 	}
 	return &tagImpl{key: toUnversioned, value: falseValue}
+}
+
+func TaskExpireStateTag(state string) Tag {
+	return &tagImpl{key: taskExpireState, value: state}
 }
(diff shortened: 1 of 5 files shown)


common/metrics/tags.go
```diff
@@ -41,6 +41,7 @@ const (
 	// See server.api.enums.v1.ReplicationTaskType
 	replicationTaskType     = "replicationTaskType"
 	replicationTaskPriority = "replicationTaskPriority"
+	taskExpireState         = "task_expire"
```
dnr: maybe call it `expire_stage`?

common/metrics/tags.go
```diff
@@ -477,3 +478,7 @@ func ToUnversionedTag(version string) Tag {
 	}
 	return &tagImpl{key: toUnversioned, value: falseValue}
 }
+
+func TaskExpireStateTag(state string) Tag {
+	return &tagImpl{key: taskExpireState, value: state}
+}
```
dnr: how about we just define all 2-3 tags that we need statically?

service/matching/physical_task_queue_manager.go
```diff
@@ -338,8 +338,11 @@ func (c *physicalTaskQueueManagerImpl) PollTask(
 		// there. In that case, go back for another task.
 		// If we didn't do this, the task would be rejected when we call RecordXTaskStarted on
 		// history, but this is more efficient.
+
 		if task.event != nil && IsTaskExpired(task.event.AllocatedTaskInfo) {
-			c.metricsHandler.Counter(metrics.ExpiredTasksPerTaskQueueCounter.Name()).Record(1)
+			// task is expired while polling
+			taskExpireStateTag := metrics.TaskExpireStateTag("process")
```
dnr: I'm trying to think through how we might use this...

I'm feeling like there's really only two meaningful values: `read` (expired right when we read it from db) and `memory` (expired later). But then the taskValidator stuff might be worth splitting out (it's not so much "expired" as "obsolete" (workflow or activity was canceled or not needed any more for replication.

So let's say `memory` here, `read` in *_task_reader.go, and `invalid` for when taskValidator.maybeValidate fails?

common/metrics/tags.go
ychebotarev: renamed to "task_expire_stage".

service/matching/physical_task_queue_manager.go
ychebotarev: "task expire while in memory" for me is less descriptive then "task expire while processing"
but ok.

common/metrics/tags.go
ychebotarev: done

service/matching/physical_task_queue_manager.go
dnr: Ok, I admit "memory" is bad because it doesn't really mean anything out of context. But "process" is also bad because it's not "while processing", it's "while buffered in memory". Maybe "buffered" then?

common/metrics/tags.go
```diff
@@ -477,3 +478,15 @@ func ToUnversionedTag(version string) Tag {
 	}
 	return &tagImpl{key: toUnversioned, value: falseValue}
 }
+
+func TaskExpireStageReadTag() Tag {
+	return &tagImpl{key: taskExpireStage, value: "read"}
```
dnr: This will still allocate a new one for every use.. let's just make them global `var`s?

service/matching/physical_task_queue_manager.go
```diff
@@ -382,7 +384,8 @@ func (c *physicalTaskQueueManagerImpl) ProcessSpooledTask(
 ) error {
 	if !c.taskValidator.maybeValidate(task.event.AllocatedTaskInfo, c.queue.TaskType()) {
 		task.finish(nil, false)
-		c.metricsHandler.Counter(metrics.ExpiredTasksPerTaskQueueCounter.Name()).Record(1)
+
+		c.metricsHandler.Counter(metrics.ExpiredTasksPerTaskQueueCounter.Name()).Record(1, metrics.TaskExpireStageMemoryTag())
```
dnr: sorry, I didn't think this through:

so maybeValidate first checks IsTaskExpired (cheap in-memory check) and if not expired, either just returns true or sometimes does an rpc (expensive). I wanted to differentiate between the IsTaskExpired returning false and the rpc returning invalid. but to do that we need to change the signature of maybeValidate. if you want to do that now, great, otherwise let's just remove the "validate" tag and make them all "memory"/"buffer"/whatever

common/metrics/tags.go
ychebotarev: done

service/matching/physical_task_queue_manager.go
ychebotarev: changed.
Not closing this, please take a look at the change.

common/metrics/tags.go
```diff
@@ -477,3 +478,19 @@ func ToUnversionedTag(version string) Tag {
...
+	return &taskExpireStageReadTag
+}
+
+func TaskExpireStageMemoryTag() Tag {
+	return &taskExpireStageMemoryTag
+}
+
+func TaskExpireStageValidateTag() Tag {
+	return &taskExpireStageInvalidTag
+}
```
dnr: can't we just do:

```suggestion
var TaskExpireStageReadTag Tag = &tagImpl{key: taskExpireStage, value: "read"}
var TaskExpireStageMemoryTag Tag = &tagImpl{key: taskExpireStage, value: "memory"}
var TaskExpireStageInvalidTag Tag = &tagImpl{key: taskExpireStage, value: "invalid"}
```

what are the functions adding here?

service/matching/pri_matcher.go
```diff
@@ -187,7 +187,9 @@ func (tm *priTaskMatcher) forwardTask(task *internalTask) error {
 		maybeValid := tm.validator.maybeValidate(task.event.AllocatedTaskInfo, tm.fwdr.partition.TaskType())
 		if !maybeValid {
 			task.finish(nil, false)
-			tm.metricsHandler.Counter(metrics.ExpiredTasksPerTaskQueueCounter.Name()).Record(1)
+
+			// consider this task expired while processing.
```
dnr, line 192: shouldn't we do the same Validate/Memory switch here? and maybe factor it out into a function?

service/matching/pri_matcher.go
```diff
@@ -243,7 +245,13 @@ func (tm *priTaskMatcher) validateTasksOnRoot(lim quotas.RateLimiter, retrier ba
 		if !maybeValid {
 			// We found an invalid one, complete it and go back for another immediately.
 			task.finish(nil, false)
-			tm.metricsHandler.Counter(metrics.ExpiredTasksPerTaskQueueCounter.Name()).Record(1)
+			// not sure what scenario this fits
```
dnr:
```suggestion
```

service/matching/pri_matcher.go
ychebotarev: we do

PR #8061, opened 2025-07-17
batch: add activity reset and update-options

spkane31:
## What changed?
BatchWorkflow now supports resetting activities and updating options.

## Why?
Support future cli work to support `temporal activity {batch,update-options} --query ...`

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
This does not change previous behavior

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.50.1-0.20250707232016-58d5912e8657
+	go.temporal.io/api v1.50.1-0.20250715164317-6157f960f13b
 	go.temporal.io/sdk v1.34.0
 	go.uber.org/automaxprocs v1.6.0
 	go.uber.org/fx v1.23.0
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.50.1-0.20250717214050-28ae23500a33
+	go.temporal.io/api v1.50.1-0.20250717230205-2e8ba1309322
 	go.temporal.io/sdk v1.34.0
 	go.uber.org/automaxprocs v1.6.0
 	go.uber.org/fx v1.23.0
--- a/go.mod
+++ b/go.mod
@@ -182,5 +182,3 @@ require (
 	modernc.org/strutil v1.2.1 // indirect
 	modernc.org/token v1.1.0 // indirect
 )
-
-// replace go.temporal.io/api => ../api-go
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.50.1-0.20250715164317-6157f960f13b
+	go.temporal.io/api v1.50.1-0.20250717214050-28ae23500a33
 	go.temporal.io/sdk v1.34.0
 	go.uber.org/automaxprocs v1.6.0
 	go.uber.org/fx v1.23.0
@@ -183,4 +183,4 @@ require (
 	modernc.org/token v1.1.0 // indirect
 )
 
-replace go.temporal.io/api => ../api-go
+// replace go.temporal.io/api => ../api-go
--- a/go.mod
+++ b/go.mod
@@ -182,3 +182,5 @@ require (
 	modernc.org/strutil v1.2.1 // indirect
 	modernc.org/token v1.1.0 // indirect
 )
+
+replace go.temporal.io/api => ../api-go
--- a/go.mod
+++ b/go.mod
@@ -60,7 +60,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.50.1-0.20250717230205-2e8ba1309322
+	go.temporal.io/api v1.50.1-0.20250718211104-1dc43113346a
 	go.temporal.io/sdk v1.34.0
 	go.uber.org/automaxprocs v1.6.0
 	go.uber.org/fx v1.23.0
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -6052,3 +6052,26 @@ func (wh *WorkflowHandler) ListWorkers(
 		NextPageToken: resp.GetNextPageToken(),
 	}, nil
 }
+
+func (wh *WorkflowHandler) UpdateTaskQueueConfig(
+	ctx context.Context, request *workflowservice.UpdateTaskQueueConfigRequest,
+) (*workflowservice.UpdateTaskQueueConfigResponse, error) {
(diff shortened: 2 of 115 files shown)


go.mod
```diff
@@ -60,7 +60,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.34.0
 	go.opentelemetry.io/otel/sdk/metric v1.34.0
 	go.opentelemetry.io/otel/trace v1.34.0
-	go.temporal.io/api v1.50.1-0.20250715164317-6157f960f13b
+	go.temporal.io/api v1.50.1-0.20250717214050-28ae23500a33
```
bergundy: Please merge the API change before the server PR.

service/worker/batcher/workflow.go
```diff
@@ -42,6 +44,10 @@ const (
 	BatchTypeUpdateOptions = "update_options"
 	// BatchTypePauseActivities is batch type for unpausing activities
 	BatchTypeUnpauseActivities = "unpause_activities"
+	// BatchTypeUpdateOptionsActivities is batch type for updating the options of activities
+	BatchTypeUpdateOptionsActivities = "update_options_activities"
```
bergundy:
```suggestion
	// BatchTypeUpdateActivityOptions is batch type for updating the options of activities
	BatchTypeUpdateActivityOptions = "update_activity_options"
```

service/worker/batcher/workflow.go
```diff
@@ -131,6 +157,10 @@ type (
 		UpdateOptionsParams UpdateOptionsParams
 		// UnpauseActivitiesParams is params only for BatchTypeUnpauseActivities
 		UnpauseActivitiesParams UnpauseActivitiesParams
+		// UpdateOptionsActivitiesParams is params only for BatchTypeUpdateOptionsActivities
+		UpdateOptionsActivitiesParams UpdateOptionsActivitiesParams
```
bergundy: Same here please.

service/frontend/workflow_handler.go
```diff
@@ -4594,24 +4596,81 @@ func (wh *WorkflowHandler) StartBatchOperation(
...
+			updateOptionsActivitiesParams.ActivityType = a.Type
+		case *batchpb.BatchOperationUpdateActivityOptions_MatchAll:
+			if !a.MatchAll {
+				return nil, serviceerror.NewInvalidArgument("Either activity type must be set, or match all should be set to true")
+			}
+			updateOptionsActivitiesParams.MatchAll = true
+		}
+		updateOptionsActivitiesParams.ActivityOptions = op.UpdateOptionsActivitiesOperation.GetActivityOptions()
+		updateOptionsActivitiesParams.UpdateMask = op.UpdateOptionsActivitiesOperation.GetUpdateMask()
+		updateOptionsActivitiesParams.RestoreOriginal = op.UpdateOptionsActivitiesOperation.GetRestoreOriginal()
```
bergundy: This should fail if set along other fields that are mutually exclusive.
Please also add a unit test for this.

service/frontend/workflow_handler.go
```diff
@@ -4594,24 +4596,81 @@ func (wh *WorkflowHandler) StartBatchOperation(
...
+				return nil, serviceerror.NewInvalidArgument("Either activity type must be set, or match all should be set to true")
+			}
+			resetActivitiesParams.MatchAll = true
+		}
+
+		resetActivitiesParams.ResetAttempts = op.ResetActivitiesOperation.ResetAttempts
+		resetActivitiesParams.ResetHeartbeat = op.ResetActivitiesOperation.ResetHeartbeat
+		resetActivitiesParams.Jitter = op.ResetActivitiesOperation.Jitter.AsDuration()
+		resetActivitiesParams.KeepPaused = op.ResetActivitiesOperation.KeepPaused
+		resetActivitiesParams.RestoreOriginalOptions = op.ResetActivitiesOperation.RestoreOriginalOptions
```
bergundy, line 4627: What about the identity?

service/frontend/workflow_handler.go
```diff
@@ -4594,24 +4596,81 @@ func (wh *WorkflowHandler) StartBatchOperation(
...
+				return nil, serviceerror.NewInvalidArgument("Either activity type must be set, or match all should be set to true")
+			}
+			updateOptionsActivitiesParams.ActivityType = a.Type
+		case *batchpb.BatchOperationUpdateActivityOptions_MatchAll:
+			if !a.MatchAll {
+				return nil, serviceerror.NewInvalidArgument("Either activity type must be set, or match all should be set to true")
+			}
+			updateOptionsActivitiesParams.MatchAll = true
+		}
+		updateOptionsActivitiesParams.ActivityOptions = op.UpdateOptionsActivitiesOperation.GetActivityOptions()
```
bergundy: Please validate the activity options here.

service/worker/batcher/activities.go
```diff
@@ -339,6 +339,7 @@ func startTaskProcessor(
 								WorkflowId: workflowID,
 								RunId:      runID,
 							},
+							// QUESTION: do we want to plumb through the identity from cli/ui/sdk ? @yuri/@chetan
```
bergundy: Good question, we should be consistent across all batch jobs with our use of identity.

service/worker/batcher/workflow.go
```diff
@@ -103,6 +109,26 @@ type (
 		Jitter         time.Duration
 	}
 
+	UpdateOptionsActivitiesParams struct {
+		Identity        string
+		ActivityType    string
+		MatchAll        bool
+		ActivityOptions *activitypb.ActivityOptions
+		UpdateMask      *fieldmaskpb.FieldMask
```
bergundy: I _think_ this is fine but you should know that using a nested proto struct in workflow input doesn't use proto JSON serialization, instead it uses plain JSON serialization which may yield unexpected results.

If we were writing this from scratch, the right thing to do here is to define a proto struct for the workflow input and reuse the request structs for each of these actions.
You may want to serialize these proto messages yourself and bypass the SDK's serialization.

spkane31: Chatted w/ Yuri on this. We already use the proto struct for `FieldMask` without custom serialization in other batch commands so we should be fine here.

service/worker/batcher/activities.go
spkane31: Added for these two, the rest of the options do not have an identity in the proto definition. I will add that in a later PR

service/worker/batcher/workflow.go
bergundy: I would test that every field in these message are properly serialized.
Generally, what we've done here is bad practice and should be avoided in our codebase. It's not just `FieldMask`, it's also `ActivityOptions`. Even if we know that all fields are properly serialized today (which I'm not convinced of), there's no guarantee that future fields will.

bergundy (approved):
I would like the input serialization issues to be addressed before merging but otherwise LGTM. Please do not merge before addressing.

service/worker/batcher/workflow.go
gow: fyi: There are known issues with the default protobuf json converter when the message contains `oneoff` fields - https://protobuf.dev/programming-guides/json/. It's recommended to use protojson to explicitly serialize/deserialize in such cases.
In this case you're fine since UpdateActivitiesOptionsParams doesn't contain any protobufs with `oneoff` fields. But that may change in the future if someone adds additional fields to this input.

gow (approved):

spkane31:
> I would like the input serialization issues to be addressed before merging but otherwise LGTM. Please do not merge before addressing.

Created structs for these options for JSON serialization purposes.

service/worker/batcher/workflow.go
```diff
@@ -103,6 +107,52 @@ type (
...
+	UpdateActivitiesOptionsParams struct {
+		Identity        string
+		ActivityType    string
+		MatchAll        bool
+		ActivityOptions *ActivityOptions
+		UpdateMask      *FieldMask
+		RestoreOriginal bool
+	}
+
+	ActivityOptions struct {
```
bergundy, line 119: I think this is suboptimal solution. Instead it would have been better to serialize the proto structs manually or to define the workflow inputs with protos.

PR #8062, opened 2025-07-17
Matching fairness: handle read+write by buffering tasks

dnr:
## What changed?
Buffer tasks that come from a write while a read is pending and process them after the read completes.
Also a bunch of unit test cleanup to make the tests more clear.

## Why?
fairTaskReader could get confused in some concurrent read+write situations and drop tasks that should have ended up in memory.

## How did you test it?
- [x] covered by existing tests, cleaned up tests

### DIFF
--- a/service/matching/fair_task_reader.go
+++ b/service/matching/fair_task_reader.go
@@ -45,6 +45,8 @@ type (
 		ackLevelPins     int64       // pinned when odd. track when writing tasks so that we don't delete just-written tasks
 		atEnd            bool        // whether we believe outstandingTasks represents the entire queue right now
 
+		bufferedTasks []*persistencespb.AllocatedTaskInfo
+
 		// gc state
 		inGC       bool
 		numToGC    int       // counts approximately how many tasks we can delete with a GC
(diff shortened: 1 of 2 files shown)


service/matching/fair_task_reader.go
```diff
@@ -41,9 +41,18 @@ type (
 		loadedTasks      int         // == number of unacked (non-nil) entries in outstandingTasks
 		readLevel        fairLevel   // == highest level in outstandingTasks, or if empty, the level we should read next
 		ackLevel         fairLevel   // inclusive: task exactly at ackLevel _has_ been acked
-		ackLevelPinned   bool        // pinned while writing tasks so that we don't delete just-written tasks
 		atEnd            bool        // whether we believe outstandingTasks represents the entire queue right now
 
+		// Buffer tasks written while a read is pending so we make sure to account for them in
+		// our read level.
+		bufferedTasks []*persistencespb.AllocatedTaskInfo
```
ychebotarev: writtenTasks? incomingTasks? "buffered" is
* overused
* doesn't tell me much about real semantic

service/matching/fair_task_reader.go
```diff
@@ -41,9 +41,18 @@ type (
...
+
+		// Pin ack level while writing tasks so that we don't delete just-written tasks.
+		// Also pin it while reading if we have bufferedTasks, to handle the case of concurrent
+		// reads and writes: if it's pinned by a write while a read is pending, we need to hold
+		// it pinned until bufferedTasks are processed.
+		ackLevelPinnedByWriter bool
```
ychebotarev, line 54: weird. I see this unset in unpinAckLevel.
But I can't find any "unpinAckLevel" call in this PR or in the main branch. Is it in feature branch?

service/matching/fair_task_reader.go
```diff
@@ -165,9 +174,11 @@ func (tr *fairTaskReader) completeTaskLocked(task *internalTask) {
...
-	// shouldReadMoreLocked before it exits.
-	if tr.readPending || !tr.shouldReadMoreLocked() {
+	// If readPending is true, readTasksImpl is running and will check shouldReadMoreLocked
+	// before it exits, so we'll definitely do another read if shouldReadMoreLocked is true.
+	// We also abort here if we're in the middle of a backoff or shutting down.
+	if tr.readPending || !tr.shouldReadMoreLocked() ||
+		tr.backoffTimer != nil || tr.backlogMgr.tqCtx.Err() != nil {
 		return
 	}
 	tr.readPending = true
```
ychebotarev, line 184: here is what is worrying me...

1. Some1 calls maybeReadTasksLocked (so tr.lock.Lock() is set)
2. It starts the go routine with readTasksImpl.
3. maybeReadTasksLocked exits, and eventually tr.lock is released.
4. readTasksImpl progress, and comes to  line 209 (lastErr = tr.readTaskBatch(readLevel, loadedTasks)).
   a. at this moment lock is released
5. Another call started that will lead to maybeReadTasksLocked
6. Another readTasksImpl
7. That another one also comes to line line 209 
  a. at this moment lock is released

So both routines call readTaskBatch, with the same parameters.

Does that make sense?

service/matching/fair_task_reader.go
dnr: how about `newlyWrittenTasks`?

service/matching/fair_task_reader.go
dnr: yes, the call is here: https://github.com/dnr/temporal/blob/fair12/service/matching/fair_backlog_manager.go#L241
all of this code is on the feature branch

service/matching/fair_task_reader.go
dnr: `tr.readPending` is the flag that prevents that: maybeReadTasksLocked sets readPending true before it starts the goroutine (this line). the next maybeReadTasksLocked sees readPending true (line 180) and just returns

ychebotarev: right

service/matching/fair_task_reader.go
ychebotarev: fine, very descriptive

ychebotarev (approved):

PR #8065, opened 2025-07-18
Perform concurrent writes with dual visibility

rodrigozhou:
## What changed?
Perform writes to dual visibility mode concurrently instead of sequentially.

Fixed shadow read context: shadow read requests were using the same context as the main request, so if the main request finishes earlier, the context would close, and fail the shadow request immediately. This PR copies the deadline so it has the same timeout, but it's a new context so it won't be using the main context.

General refactor of dual visibility code.

## Why?
Each write buffers a Elasticsearch request, so executing sequentially makes the overall visibility task latency much higher.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/visibility/visibility_manager_dual.go
+++ b/common/persistence/visibility/visibility_manager_dual.go
@@ -2,6 +2,8 @@ package visibility
 
 import (
 	"context"
+	"errors"
+	"sync"
 
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/namespace"
@@ -14,6 +16,13 @@ type (
 		secondaryVisibilityManager manager.VisibilityManager
 		managerSelector            managerSelector
 		enableShadowReadMode       dynamicconfig.BoolPropertyFn
+
+		// Separate context for the shadow read requests. It ensures the main request context won't
+		// cancel the shadow request if the former completes before the shadow request.
+		// Each shadow request gets a new child context from shadowReadCtx, and we cancel all of them
+		// when visibilityManager closes by calling shadowReadCtxCancel.
+		shadowReadCtx       context.Context
+		shadowReadCtxCancel context.CancelFunc
 	}
 )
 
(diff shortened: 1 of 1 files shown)


common/persistence/visibility/visibility_manager_dual.go
```diff
@@ -14,6 +16,8 @@ type (
 		secondaryVisibilityManager manager.VisibilityManager
 		managerSelector            managerSelector
 		enableShadowReadMode       dynamicconfig.BoolPropertyFn
+		shadowReadCtx              context.Context
```
jmbarzee: I'm honestly shocked that the linter passed. I think this is a strong signal we should improve this pattern

jmbarzee (commented):
Rodrigo and I will probably chat sync to understand constraints he encountered while developing.

common/persistence/visibility/visibility_manager_dual.go
rodrigozhou: Discussed offline: added a comment to explain it

common/persistence/visibility/visibility_manager_dual.go
```diff
@@ -14,6 +16,11 @@ type (
 		secondaryVisibilityManager manager.VisibilityManager
 		managerSelector            managerSelector
 		enableShadowReadMode       dynamicconfig.BoolPropertyFn
+
+		// Separate context for the shadow read requests. It ensures the main request context won't
```
jmbarzee, line 20: This comment is great, but I think we need to better illustrate the reasoning for the next person who has zero context. 

I think we should make it clear that we 
1. Derive shadow contexts outside of the primary context change. 
2. Need a way to globally cancel all running shadow requests.

jmbarzee: Even though I think the pattern is justified, It is uncommon enough that we should be very clear about out motivation

jmbarzee (approved):
Awesome. Thanks for the changes!

PR #8068, opened 2025-07-18
Check if activity is paused or stamp mismatch in processActivityTask()

gow:
## What changed?
Now checking if the activity is paused or if the stamp is outdated before dispatching a matching task.

## Why?
Prevents unnecessary creation and retrying of tasks in matching queues.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
N/A

### DIFF
--- a/service/history/transfer_queue_standby_task_executor.go
+++ b/service/history/transfer_queue_standby_task_executor.go
@@ -164,7 +164,7 @@ func (t *transferQueueStandbyTaskExecutor) processActivityTask(
 		}
 
 		if activityInfo.Stamp != transferTask.Stamp || activityInfo.Paused {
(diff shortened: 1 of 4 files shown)


service/history/transfer_queue_standby_task_executor.go
```diff
@@ -163,6 +163,10 @@ func (t *transferQueueStandbyTaskExecutor) processActivityTask(
 			return nil, nil
 		}
 
+		if activityInfo.Stamp != transferTask.Stamp || activityInfo.Paused {
+			return nil, consts.ErrStaleReference // drop the task
```
yycptt: return nil error, or the ms will be unloaded.

yycptt (approved):

PR #8071, opened 2025-07-21
Add sql connection pool metrics

vanhtuan0409:
## What changed?

Adding sql connection pool metrics from stdlib `Stats`

## Why?

Expose more metrics for config fine tuning

## How did you test it?
- [x] built
- [ ] added new unit test(s)


common/metrics/metric_defs.go
```diff
@@ -27,6 +27,7 @@ const (
 	resourceExhaustedScopeTag   = "resource_exhausted_scope"
 	PartitionTagName            = "partition"
 	PriorityTagName             = "priority"
+	PersistenceDBKindTagName    = "db_kind"
```
bergundy, line 30: Why do users need the db_kind tag? They should know what DB their cluster is running.

common/metrics/metric_defs.go
```diff
@@ -1261,6 +1262,9 @@ var (
 	CassandraSessionRefreshFailures        = NewCounterDef("cassandra_session_refresh_failures")
 	PersistenceSessionRefreshFailures      = NewCounterDef("persistence_session_refresh_failures")
 	PersistenceSessionRefreshAttempts      = NewCounterDef("persistence_session_refresh_attempts")
+	PersistenceSQLMaxOpenConn              = NewGaugeDef("persistence_sql_max_open_conn")
```
bergundy, line 1303: Can we add an `in_use` counter too?

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,54 @@
...
+		return
+	}
+	s := db.Stats()
+	metrics.PersistenceSQLMaxOpenConn.With(r.metrics).Record(float64(s.MaxOpenConnections))
+	metrics.PersistenceSQLOpenConn.With(r.metrics).Record(float64(s.OpenConnections))
+	metrics.PersistenceSQLIdleConn.With(r.metrics).Record(float64(s.Idle))
+}
+
+func (r *DBMetricsReporter) Stop() {
+	close(r.quit)
```
bergundy: How about waiting for the `Start` goroutine to finish before returning here? It should be instantaneous.

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type DBMetricsReporter struct {
+	interval time.Duration
+	handle   *DatabaseHandle
+	metrics  metrics.Handler
+	quit     chan struct{}
+}
+
+func withDBMetricReporter(dbKind DbKind, handle *DatabaseHandle) {
```
bergundy: nit: Can you name this method `newDBMetricsReporter` and explicitly call `Start` on it. Just for symmetry with where `Stop` is called.

bergundy (commented):
Apologies for the late review.

common/metrics/metric_defs.go
vanhtuan0409: because temporal maintain separated sql pool for primary and visibility db. In the scenario of both primary and visibility using sqlplugin, without `db_kind` tag would result in merged total count of connection from both primary and visibility cluster

common/metrics/metric_defs.go
vanhtuan0409: added `in_use` counter in d54181b09a6971f93b2b62e8f07e359eb9afb1ea

common/persistence/sql/sqlplugin/db_metrics.go
vanhtuan0409: added `sync.WaitGroup` in 76e13eaad43d3e73d5309f37a89bad68f686230c. But with an assumption that `Start` is called before

common/persistence/sql/sqlplugin/db_metrics.go
vanhtuan0409: fixed in 4610f784602c23c8e94d753d8fc1bef46debdcd5

common/metrics/metric_defs.go
bergundy: Ah, thanks!

common/metrics/metric_defs.go
bergundy: Thanks!

common/persistence/sql/sqlplugin/db_metrics.go
bergundy: Thanks!

vanhtuan0409:
added concurrency guard in b9a7e9d7e6ef4d55867bbc38848a34b77af2334f

can you check again @bergundy

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Stop signal background reporter to stop
+// and wait for reporter to completely stopped
+// safe to call multiple time
+func (r *DBMetricsReporter) Stop() {
+	if !atomic.CompareAndSwapInt32(&r.stopped, 0, 1) {
+		return
+	}
+	close(r.quit)
+	r.wg.Wait()
```
bergundy: Just slightly more correct, this way any call is ensured to wait for stop to complete.

```suggestion
	if atomic.CompareAndSwapInt32(&r.stopped, 0, 1) {
		close(r.quit)
	}
	r.wg.Wait()
```

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+
+type DBMetricsReporter struct {
+	interval time.Duration
+	handle   *DatabaseHandle
+	metrics  metrics.Handler
+	quit     chan struct{}
+	wg       sync.WaitGroup
+
+	started int32
+	stopped int32
```
bergundy: Can use atomic bools here: https://pkg.go.dev/sync/atomic#Bool

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,86 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-r.quit:
+			return
+		case <-ticker.C:
+			if atomic.LoadInt32(&r.stopped) == 1 {
+				return
+			}
```
bergundy: A but redundant IMHO but no objection to keep this.

common/persistence/sql/sqlplugin/db_metrics.go
bergundy: But also wondering why this would be started multiple times since it's only instantiated in the handle constructor and immediately started there.

common/persistence/sql/sqlplugin/db_handle.go
```diff
@@ -60,6 +62,9 @@ func NewDatabaseHandle(
 		logger:       logger,
 		timeSource:   timeSource,
 	}
+	reporter := newDBMetricReporter(dbKind, handle)
+	handle.reporter = reporter
+	handle.reporter.Start()
```
bergundy: nit:

```suggestion
	handle.reporter = newDBMetricReporter(dbKind, handle)
	handle.reporter.Start()
```

common/persistence/sql/sqlplugin/db_metrics.go
vanhtuan0409: sure enough. I'm keeping this at the moment

common/persistence/sql/sqlplugin/db_metrics.go
vanhtuan0409: I switched it to `atomic.Bool` in 921be964f445d89a0f6ed1c9f35936bc7b524b17

Yeah, `Start` will probably only be called once, but `Stop` might be called multiple times, which I'm not sure about. Just keep it symmetric with the stop.

Before the guard check, the Go race detector complains about a race condition in WaitGroup. I'm adding these to resolve the race condition

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+func newDBMetricReporter(dbKind DbKind, handle *DatabaseHandle) *DBMetricsReporter {
+	reporter := &DBMetricsReporter{
+		interval: time.Minute,
+		handle:   handle,
+		metrics:  handle.metrics.WithTags(metrics.PersistenceDBKindTag(dbKind.String())),
+		quit:     make(chan struct{}),
+		logger:   handle.logger,
+	}
+	reporter.started.Store(false)
+	reporter.stopped.Store(false)
```
bergundy, line 34: No need, the zero value is false.

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	return reporter
+}
+
+// Start run metrics report in background
+// yield control immediately without blocking
+// safe to called multiple time
+func (r *DBMetricsReporter) Start() {
+	if !r.started.CompareAndSwap(false, true) {
+		return
+	}
```
bergundy, line 44: Still don't feel like we need this but 🤷

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	r.wg.Add(1)
+	go r.run()
```
bergundy, line 46:
```suggestion
	r.wg.Go(r.run)
```

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (r *DBMetricsReporter) run() {
+	defer r.wg.Done()
```
bergundy, line 50: No need if you use `wg.Go`.

bergundy (approved):
I've already approved this, at this point it's nitpicking. Feel free to address the last couple of comments if you feel like it. Otherwise, I can merge as is.

common/persistence/sql/sqlplugin/db_metrics.go
vanhtuan0409: oh nice. I just know about this method. Unfortunately, it is only available with go 1.25. Temporal is using go 1.24 instead

vanhtuan0409:
the CI is failing. Do I need to rebase or checking anything @bergundy . Look like some docker service is missing

common/persistence/sql/sqlplugin/db_metrics.go
bergundy: That's not true anymore https://github.com/temporalio/temporal/blob/16b1e29b2a86eaf1a95e026e49e16f31f8c414aa/go.mod#L3

bergundy:
I merged main into the PR branch, lets see if that fixes CI.

vanhtuan0409:
All good. It is great working with you

bergundy:
I'm merging, would have liked some of the small comments to have been addressed but there's nothing critical blocking this right now. Thanks for the contribution.

common/persistence/sql/sqlplugin/db_metrics.go
```diff
@@ -0,0 +1,87 @@
...
+				return
+			}
+			r.report()
+		}
+	}
+}
+
+func (r *DBMetricsReporter) report() {
+	db := r.handle.db.Load()
+	if db == nil {
```
erka, line 69: @bergundy should stats go to zero if this occurs?

bergundy: Hmm... this is just an edge case. I would say it's undefined. Most of the time the gauge will stay at the last value for the host.

PR #8078, opened 2025-07-22
Ability to generate migration task via frontend

yycptt:
## What changed?
- Add capability to generate migration task via frontend.

## Why?
- We don't really have per namespace rate limiting on history client, going through frontend allow us to rate limiting migration traffic per namespace and protect the cluster.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
- Feature is behind a feature flag, so should be safe.
- The feature flag can only be enabled after the change in the admin handler (for forwarding the target cluster field) is deployed.

### DIFF
--- a/service/worker/migration/activities.go
+++ b/service/worker/migration/activities.go
@@ -506,7 +506,8 @@ func (a *activities) GenerateReplicationTasks(ctx context.Context, request *gene
 	}
 
 	generateViaFrontend := a.generateMigrationTaskViaFrontend()
-	for i, we := range request.Executions {
+	for i := startIndex; i < len(request.Executions); i++ {
+		we := request.Executions[i]
 		if err := a.generateWorkflowReplicationTask(
 			ctx,
 			rateLimiter,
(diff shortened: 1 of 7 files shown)


service/worker/migration/activities.go
```diff
@@ -475,27 +499,37 @@ func (a *activities) GenerateReplicationTasks(ctx context.Context, request *gene
...
+		a.logger.Error("force-replication failed to translate namespaceID to name", tag.WorkflowNamespaceID(request.NamespaceID))
+		return err
+	}
 
-				a.logger.Warn("force-replication ignore replication task due to NotFoundServiceError",
-					tag.WorkflowNamespaceID(we.GetNamespaceID()),
-					tag.WorkflowID(we.GetWorkflowID()),
-					tag.WorkflowRunID(we.GetRunID()),
+	generateViaFrontend := a.generateMigrationTaskViaFrontend()
+	for i, we := range request.Executions {
```
hai719: startIndex not used?

yycptt: oh good catch! Thanks! Fixed it and updated the test to cover startIndex as well.

hai719 (approved):

PR #8080, opened 2025-07-22
CHASM Visibility: Component and task processing

yycptt:
## What changed?
- Add CHASM built-in visibility component and task processing logic

## Why?
- Basic chasm visibility work for unblocking Scheduler migration.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/visibility.go
+++ b/chasm/visibility.go
@@ -9,11 +9,11 @@ import (
 )
 
 const (
-	// TODO: add a test for this constant
 	visibilityComponentFqType = "core.vis"
 	visibilityTaskFqType      = "core.visTask"
 )
 
+// CoreLibrary contains built-in components maintained as part of the CHASM framework.
 type CoreLibrary struct {
 	UnimplementedLibrary
 }
--- a/tests/chasm_test.go
+++ b/tests/chasm_test.go
@@ -14,6 +14,7 @@ import (
 	"go.temporal.io/server/chasm/lib/tests"
 	"go.temporal.io/server/common/debug"
 	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/payload"
 	"go.temporal.io/server/common/testing/testvars"
 	"go.temporal.io/server/tests/testcore"
 )
@@ -108,6 +109,19 @@ func (s *ChasmTestSuite) TestPayloadStoreVisibility() {
 	s.Empty(visRecord.StateTransitionCount)
 	s.Empty(visRecord.CloseTime)
 	s.Empty(visRecord.HistoryLength)
+
+	var intVal int
+	p, ok := visRecord.Memo.Fields[tests.TotalCountMemoFieldName]
+	s.True(ok)
+	s.NoError(payload.Decode(p, &intVal))
+	s.Equal(0, intVal)
+	p, ok = visRecord.Memo.Fields[tests.TotalSizeMemoFieldName]
+	s.True(ok)
+	s.NoError(payload.Decode(p, &intVal))
+	s.Equal(0, intVal)
+	var strVal string
+	s.NoError(payload.Decode(visRecord.SearchAttributes.IndexedFields[tests.TestKeywordSAFieldName], &strVal))
+	s.Equal(tests.TestKeywordSAFieldValue, strVal)
 
 	_, err = tests.ClosePayloadStoreHandler(
 		engineContext,
--- a/chasm/lib/tests/handler.go
+++ b/chasm/lib/tests/handler.go
@@ -80,7 +80,8 @@ func NewPayloadStoreHandler(
 			BusinessID:  request.StoreID,
 		},
 		func(mutableContext chasm.MutableContext, _ any) (*PayloadStore, any, error) {
-			return NewPayloadStore(mutableContext), nil, nil
+			store, err := NewPayloadStore(mutableContext)
+			return store, nil, err
 		},
 		nil,
 	)
--- a/chasm/lib/tests/payload.go
+++ b/chasm/lib/tests/payload.go
@@ -6,7 +6,21 @@ import (
 	persistencespb "go.temporal.io/server/api/persistence/v1"
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/common"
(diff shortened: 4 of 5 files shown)


chasm/visibility.go
```diff
@@ -0,0 +1,178 @@
...
+		value, err := field.Get(chasmContext)
+		if err != nil {
+			return nil, err
+		}
+		sa[key] = value
+	}
+	return sa, nil
+}
+
+func (v *Visibility) UpdateSearchAttributes(
```
yycptt: I still need to tweak the interface a little bit and add some generic function to give more strongly typed experience.

chasm/tree.go
```diff
@@ -2399,6 +2510,7 @@ func makeValidationFn(
 
 		// Handle bool result.
 		if !result[0].Bool() {
+			// TODO: this should be not-found error
```
yycptt, line 2495: This feels like a bug to me, and will cause task to keep failing  & retry when task is invalid?

I will fix this in a different PR.

chasm/tree.go
```diff
@@ -2266,21 +2332,38 @@ func (n *Node) ValidateSideEffectTask(
...
 
-	// TODO - cache deserialized task
-	taskValue, err := deserializeTask(registrableTask, taskInfo.Data)
+	nodePath, err := n.pathEncoder.Decode(taskInfo.Path)
 	if err != nil {
-		return nil, err
+		return nil, serviceerror.NewInternalf("failed to decode path '%s'", taskInfo.Path)
+	}
+
+	node, ok := n.findNode(nodePath)
```
yycptt, line 2329: Feels like a bug in the old implementation and we were validating tasks on the wrong node.

chasm/tree.go
```diff
@@ -1313,6 +1357,18 @@ func (n *Node) closeTransactionUpdateComponentTasks(
 		}
 
 		for _, newTask := range newTasks {
+			valid, err := node.validateTask(
```
yycptt, line 1359: Validate new tasks as well. So that only one visibility task is generated even if Visibility component is updated multiple times in a transition.

chasm/context.go
```diff
@@ -27,7 +29,7 @@ type Context interface {
 type MutableContext interface {
 	Context
 
-	AddTask(Component, TaskAttributes, any) error
+	AddTask(Component, TaskAttributes, any)
```
lina-temporal, line 32: 👍  was thinking about making the same change

chasm/tree.go
lina-temporal: Yeah, agree, that should be returning not-found.

chasm/visibility.go
```diff
@@ -0,0 +1,250 @@
...
+	"go.temporal.io/server/common/payload"
+)
+
+const (
+	// TODO: add a test for this constant
+	visibilityComponentFqType = "core.vis"
+	visibilityTaskFqType      = "core.visTask"
+)
+
+type CoreLibrary struct {
```
lina-temporal, line 17:
```suggestion
CoreLibrary registers built-in components maintained as part of the CHASM framework. 
type CoreLibrary struct {
```

chasm/visibility.go
```diff
@@ -0,0 +1,250 @@
...
+		NewRegistrableComponent[*Visibility]("vis"),
+	}
+}
+
+func (b *CoreLibrary) Tasks() []*RegistrableTask {
+	return []*RegistrableTask{
+		NewRegistrableSideEffectTask(
+			"visTask",
+			&visibilityTaskValidator{},
+			&visibilityTaskExecutor{},
```
lina-temporal, line 36: (minor) though I like that these are separate interfaces, I've preferred putting both implementations onto the same struct, since then they can share struct-level helper functions. Not that that matters in this case, of course, so really up to you.

service/history/visibility_queue_task_executor_test.go
```diff
@@ -553,6 +563,198 @@ func (s *visibilityQueueTaskExecutorSuite) TestProcessorDeleteExecution() {
...
+	visTaskData := &persistencespb.ChasmVisibilityTaskData{
+		TransitionCount: taskTransitionCount,
+	}
+	data, err := visTaskData.Marshal()
+	s.NoError(err)
+
+	return &tasks.ChasmTask{
+		WorkflowKey:         key,
+		VisibilityTimestamp: time.Now().UTC(),
+		TaskID:              int64(59),
```
lina-temporal, line 743: From my understanding of the visibility executor, the task ID doesn't really matter (instead, we check against initial VT/last update VT and task's VT). Is that a correct?

tests/chasm_test.go
```diff
@@ -58,4 +62,85 @@ func (s *ChasmTestSuite) TestNewPayloadStore() {
 	s.NoError(err)
 }
 
+func (s *ChasmTestSuite) TestPayloadStoreVisibility() {
```
lina-temporal, line 66: Should we also add test asserts for the stored memo field/search attributes here? Maybe PayloadStore could store a record of its TotalSize/TotalCount?

lina-temporal (approved):
lgtm!! I like the interface to visibility for CHASM. Few questions and minor comments.

chasm/visibility.go
yycptt: Yeah I agree they can be the same struct to share dependencies. Don't feel strongly about it in this case though, as those structs are empty anyway...

service/history/visibility_queue_task_executor_test.go
yycptt: It doesn't matter for chasm related logic, but underlying visibility store uses taskID as the version of each record to prevent old records from overwriting newer ones.

tests/chasm_test.go
yycptt: Yeah... it's a bit tricky in this case because we haven't implemented per-component SA support so only those predefined ones can be used.

So for now I added both TotalSize/Count as memo fields and set an existing pre-defined keyword SA used by Schedule to have coverage on both memo & SA.

chasm/tree.go
yycptt: https://github.com/temporalio/temporal/pull/8134

PR #8081, opened 2025-07-22
improvement: bypass middle-man-struct for batch operations

spkane31:
## What changed?
Remove `BatchParams` to safely serialize proto -> json -> proto and add a `BatchOperation` proto definition

## Why?
Remove custom serialization of proto for json safety.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Could break batching backwards compatibility but tests should catch this

### DIFF
--- a/common/priorities/priority_util.go
+++ b/common/priorities/priority_util.go
@@ -10,11 +10,15 @@ func Merge(
 	base *commonpb.Priority,
 	override *commonpb.Priority,
 ) *commonpb.Priority {
-	if base == nil || override == nil {
-		return cmp.Or(override, base)
+	if base == nil {
+		return override
+	} else if override == nil {
+		return base
 	}
 
 	return &commonpb.Priority{
-		PriorityKey: cmp.Or(override.PriorityKey, base.PriorityKey),
+		PriorityKey:    cmp.Or(override.PriorityKey, base.PriorityKey),
+		FairnessKey:    cmp.Or(override.FairnessKey, base.FairnessKey),
+		FairnessWeight: cmp.Or(override.FairnessWeight, base.FairnessWeight),
 	}
 }
--- a/common/priorities/priority_util_test.go
+++ b/common/priorities/priority_util_test.go
@@ -46,6 +46,30 @@ func TestMerge(t *testing.T) {
 			override: defaultPriority,
 			expected: &commonpb.Priority{PriorityKey: 1},
 		},
+		{
+			name:     "fairness key is overriden",
+			base:     defaultPriority,
+			override: &commonpb.Priority{FairnessKey: "one"},
+			expected: &commonpb.Priority{FairnessKey: "one"},
+		},
+		{
+			name:     "fairness key is not overriden by default value",
+			base:     &commonpb.Priority{FairnessKey: "two"},
+			override: defaultPriority,
+			expected: &commonpb.Priority{FairnessKey: "two"},
+		},
+		{
+			name:     "fairness weight is overriden",
+			base:     defaultPriority,
+			override: &commonpb.Priority{FairnessWeight: 3.0},
+			expected: &commonpb.Priority{FairnessWeight: 3.0},
+		},
+		{
+			name:     "fairness weight is not overriden by default value",
+			base:     &commonpb.Priority{FairnessWeight: 3.0},
+			override: defaultPriority,
+			expected: &commonpb.Priority{FairnessWeight: 3.0},
+		},
 	}
 
 	for _, tc := range testcases {


service/worker/batcher/workflow.go
```diff
@@ -185,7 +176,7 @@ type (
 		// UpdateActivitiesOptionsParams is params only for BatchTypeUpdateActivitiesOptions
 		UpdateActivitiesOptionsParams UpdateActivitiesOptionsParams
 		// ResetActivitiesParams is params only for BatchTypeResetActivities
-		ResetActivitiesParams ResetActivitiesParams
+		ResetActivitiesParams *workflowservicepb.StartBatchOperationRequest_ResetActivitiesOperation
```
dnr: you can't put a proto in a struct and have it serialized correctly by the dataconverter, it'll choose json and serialize the proto using encoding/json, but it needs to use protojson. the workflow argument needs to be proto from the top. see the hack in ResetParams.

service/worker/batcher/fx.go
```diff
@@ -68,6 +69,8 @@ func (s *workerComponent) DedicatedWorkerOptions(ns *namespace.Namespace) *worke
 func (s *workerComponent) Register(registry sdkworker.Registry, ns *namespace.Namespace, _ workercommon.RegistrationDetails) func() {
 	registry.RegisterWorkflowWithOptions(BatchWorkflow, workflow.RegisterOptions{Name: BatchWFTypeName})
 	registry.RegisterActivity(s.activities(ns.Name(), ns.ID()))
+
+	registry.RegisterWorkflowWithOptions(BatchWorkflowProtobuf, workflow.RegisterOptions{Name: BatchWFTypeNameProtobuf})
```
spkane31: does the separate activity have to be registered as well?

service/worker/batcher/workflow.go
```diff
@@ -327,6 +357,54 @@ func validateParams(params BatchParams) error {
 	}
 }
 
+func validateParamsProtobuf(params *batchpb.BatchOperation) error {
```
bergundy: I think it's fine to have this validation here too as long as we make sure to also validate on the frontend.

service/worker/batcher/activities.go
```diff
@@ -198,6 +199,128 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
 	return hbd, nil
 }
 
+// BatchActivityWithProtobuf is an activity for processing batch operations using protobuf as the input type.
+// nolint:revive,cognitive-complexity
+func (a *activities) BatchActivityWithProtobuf(ctx context.Context, batchParams *batchpb.BatchOperation) (HeartBeatDetails, error) {
```
dnr: I was imagining this would share the bulk of the code with the existing activity, i.e. the original activity could convert its struct to a proto and call this one? Are they really that different that we need to duplicate it?

service/frontend/workflow_handler.go
```diff
@@ -4749,6 +4750,357 @@ func (wh *WorkflowHandler) StartBatchOperation(
 	return &workflowservice.StartBatchOperationResponse{}, nil
 }
 
+func (wh *WorkflowHandler) StartBatchOperationWithProtobuf(
```
dnr: This shouldn't be a new workflowservice api, the existing api should just use the proto version of the workflow instead of the struct version. Maybe based on the value of a dynamic config, to avoid the situation that a new batch workflow started by frontend may have to wait a while until new worker is deployed. Or we could just accept the delay for that window.

service/worker/batcher/fx.go
dnr: they're registered by struct, so no

service/worker/batcher/activities.go
```diff
@@ -198,6 +199,128 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
...
+	rateLimit := rate.Limit(rps)
+	burstLimit := int(math.Ceil(rps)) // should never be zero because everything would be rejected
+	rateLimiter := rate.NewLimiter(rateLimit, burstLimit)
+	taskCh := make(chan taskDetail, pageSize)
+	respCh := make(chan error, pageSize)
+	for i := 0; i < a.getOperationConcurrency(int(batchParams.Concurrency)); i++ {
+		go startTaskProcessorProtobuf(ctx, batchParams, taskCh, respCh, rateLimiter, sdkClient, a.FrontendClient, metricsHandler, logger)
+	}
+
+	for {
```
bergundy, line 264: We should eventually fix (not in this PR) this but there's no reason to wait before fetching the next page if we have executors that are free to pick up work.

Would you please open an issue to track?

service/worker/batcher/activities.go
```diff
@@ -222,6 +345,20 @@ func (a *activities) adjustQuery(batchParams BatchParams) string {
 	}
 }
 
+func (a *activities) adjustQueryProtobuf(batchParams *batchpb.BatchOperation) string {
```
bergundy: nit: the logic here can be shared across the two implementations

service/worker/batcher/activities.go
```diff
@@ -222,6 +345,20 @@ func (a *activities) adjustQuery(batchParams BatchParams) string {
...
+	if len(batchParams.Query) == 0 {
+		// don't add anything if query is empty
+		return batchParams.Query
+	}
+
+	switch batchParams.GetOperation().(type) {
+	case *batchpb.BatchOperation_TerminationOperation, *batchpb.BatchOperation_CancellationOperation, *batchpb.BatchOperation_SignalOperation, *batchpb.BatchOperation_UpdateWorkflowExecutionOptionsOperation, *batchpb.BatchOperation_UnpauseActivitiesOperation:
```
bergundy: This should include the two new operations.

service/worker/batcher/activities.go
```diff
@@ -464,6 +601,248 @@ func startTaskProcessor(
...
+							Jitter:                 operation.ResetActivitiesOperation.Jitter,
+							KeepPaused:             operation.ResetActivitiesOperation.KeepPaused,
+							RestoreOriginalOptions: operation.ResetActivitiesOperation.RestoreOriginalOptions,
+						}
+
+						switch ao := operation.ResetActivitiesOperation.GetActivity().(type) {
+						case *batchpb.BatchOperationResetActivities_Type:
+							resetRequest.Activity = &workflowservice.ResetActivityRequest_Type{Type: ao.Type}
+						case *batchpb.BatchOperationResetActivities_MatchAll:
+							resetRequest.Activity = &workflowservice.ResetActivityRequest_MatchAll{MatchAll: true}
```
bergundy, line 784: Let's add a default clause to error out if there's an unmatched type.

service/worker/batcher/activities.go
```diff
@@ -464,6 +601,248 @@ func startTaskProcessor(
...
+							if rp := ao.RetryPolicy; rp != nil {
+								updateRequest.ActivityOptions.RetryPolicy = &commonpb.RetryPolicy{
+									InitialInterval:        rp.InitialInterval,
+									BackoffCoefficient:     rp.BackoffCoefficient,
+									MaximumInterval:        rp.MaximumInterval,
+									MaximumAttempts:        rp.MaximumAttempts,
+									NonRetryableErrorTypes: rp.NonRetryableErrorTypes,
+								}
+							}
+						}
```
bergundy: Better just pass through and not have to remember to propagate new options as they're added

```suggestion
						updateRequest.ActivityOptions = operation.UpdateActivityOptionsOperation.GetActivityOptions()
```

service/worker/batcher/activities.go
```diff
@@ -464,6 +601,248 @@ func startTaskProcessor(
...
+						_, err = frontendClient.UpdateActivityOptions(ctx, updateRequest)
+						return err
+					})
+				// QUESTION seankane (2025-07-18): why do we not have a default case and return an error? @yuri/@chetan
```
bergundy: I think we should have one.

service/worker/batcher/fx.go
bergundy: No because it's instantiated with `s.activities()`

bergundy:
```suggestion
	registry.RegisterWorkflowWithOptions(BatchWorkflow, workflow.RegisterOptions{Name: BatchWFTypeName})
	// Newer version of the batch workflow which was rewritten to accept a proto struct as input.
	registry.RegisterWorkflowWithOptions(BatchWorkflowProtobuf, workflow.RegisterOptions{Name: BatchWFTypeNameProtobuf})
	registry.RegisterActivity(s.activities(ns.Name(), ns.ID()))
```

service/frontend/workflow_handler.go
```diff
@@ -4749,6 +4750,357 @@ func (wh *WorkflowHandler) StartBatchOperation(
...
+				operation.UpdateActivityOptionsOperation.ActivityOptions.RetryPolicy = &commonpb.RetryPolicy{
+					InitialInterval:        rp.InitialInterval,
+					MaximumInterval:        rp.MaximumInterval,
+					BackoffCoefficient:     rp.BackoffCoefficient,
+					NonRetryableErrorTypes: rp.NonRetryableErrorTypes,
+					MaximumAttempts:        rp.MaximumAttempts,
+				}
+			}
+			operation.UpdateActivityOptionsOperation.ActivityOptions.TaskQueue = ao.TaskQueue
+		}
```
bergundy: Same as you did elsewhere but now I'm wondering if we can't just pass in the original request to the batch job and be done with it...

```suggestion
		ao := op.UpdateActivityOptionsOperation.GetActivityOptions()
```

service/frontend/workflow_handler.go
```diff
@@ -4749,6 +4750,358 @@ func (wh *WorkflowHandler) StartBatchOperation(
...
+		Identity:                 identity,
+		RequestId:                uuid.New(),
+		WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_FAIL,
+		WorkflowIdReusePolicy:    enumspb.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE,
+		Memo:                     memo,
+		SearchAttributes:         searchAttributes,
+		Priority:                 &commonpb.Priority{}, // ie default priority
+	}
+
+	// NOTE seankane: this is where the issue is, cannot figure out how to create a new method on historyclient
```
spkane31: How do I add a method onto the history client for the new batch operation? @bergundy @cretz

bergundy: @spkane31 let's sync offline.

service/worker/batcher/activities.go
spkane31: We're duplicating to maintain backwards compatibility. If the workflow is running and then server gets upgraded to the newer version it won't be able to proceed on the previous batch operation

dnr: Yes, there has to be a new activity function, but the two activity functions could share a lot of their code, right? The activities don't have to be deterministic, they just have to be semantically equivalent. But if it's too hard to share then that's okay.

service/worker/batcher/activities.go
spkane31: #8098

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -0,0 +1,45 @@
...
+  string namespace = 1;
+
+  // The query to perform.
+  string query = 2;
+
+  repeated temporal.api.common.v1.WorkflowExecution workflow_executions = 3;
+
+  string reason = 4;
+
+  oneof operation {
```
bergundy: Why not embed the request from workflowservice?

bergundy: This can all be replaced with:
```
message BatchOperationInput {
  string namespace_id = 1;
  temporal.api.workflowservice.v1.StartBatchOperationRequest request = 2;
}
```

service/worker/batcher/workflow.go
```diff
@@ -327,6 +357,54 @@ func validateParams(params BatchParams) error {
 	}
 }
 
+func validateParamsProtobuf(params *batchspb.BatchOperation) error {
```
bergundy: Can we share the same validation logic in the frontend and the workflow?

proto/internal/temporal/server/api/batch/v1/request_response.proto
spkane31: all of `BatchOperation` can be replaced or the `oneof operation` portion? If it's the latter, why would the namespace need to be duplicated?

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -0,0 +1,40 @@
...
+  int64 concurrency = 7;
+
+  int64 attempts_on_retryable_error = 8;
+
+  google.protobuf.Duration activity_heartbeat_timeout = 9;
+
+  repeated string non_retryable_errors = 10;
+
+  string batch_type = 11;
+}
```
bergundy: Why do you need this message? All of these fields are present in `StartBatchOperationRequest`.

service/worker/batcher/workflow.go
```diff
@@ -327,6 +360,111 @@ func validateParams(params BatchParams) error {
 	}
 }
 
+// nolint:revive,cognitive-complexity
+func ValidateBatchOperation(params *batchspb.BatchOperation) error {
```
bergundy: I didn't read this very closely but since we removed validation in workflow_handler, we should make sure that anything that was validated there is ported to this function. Ignore this comment if that's already the case.

service/worker/batcher/activities.go
```diff
@@ -198,6 +200,128 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
...
 }
 
+// BatchActivityWithProtobuf is an activity for processing batch operations using protobuf as the input type.
+// nolint:revive,cognitive-complexity
+func (a *activities) BatchActivityWithProtobuf(ctx context.Context, batchParams *batchspb.BatchOperation) (HeartBeatDetails, error) {
+	logger := a.getActivityLogger(ctx)
+	hbd := HeartBeatDetails{}
+	metricsHandler := a.MetricsHandler.WithTags(metrics.OperationTag(metrics.BatcherScope), metrics.NamespaceTag(batchParams.Namespace))
+
+	if err := a.checkNamespace(batchParams.Namespace); err != nil {
```
bergundy: We should get the namespace name by namespace ID here instead of taking the namespace name from the request to be a bit more correct in case of a namespace rename.

bergundy (commented):
Looks like I reviewed this and forgot to submit.

service/worker/batcher/activities.go
dnr: If this is an issue, we can also just change perNamespaceWorkerManager to recreate workers on namespace name change

spkane31: Done

service/worker/batcher/workflow.go
spkane31: Went back over the functions, everything in workflow_handler that validates / returns errors is encapsulated in this function

proto/internal/temporal/server/api/batch/v1/request_response.proto
spkane31: Removed the second message, keeping fields that are not in the `StartBatchOperationRequest`

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -3,38 +3,24 @@ syntax = "proto3";
...
-  // The query to perform.
-  string query = 2;
-
-  repeated temporal.api.common.v1.WorkflowExecution workflow_executions = 3;
-
-  string reason = 4;
+import "google/protobuf/duration.proto";
 
-  BatchOperationInput input = 5;
+message BatchOperation {
```
bergundy: nit: I would call this `BatchOperationInput` or `BatchOperationRequest`.

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -3,38 +3,24 @@ syntax = "proto3";
...
 
-  google.protobuf.Duration activity_heartbeat_timeout = 9;
+  google.protobuf.Duration activity_heartbeat_timeout = 5;
 
-  repeated string non_retryable_errors = 10;
+  repeated string non_retryable_errors = 6;
 
-  string batch_type = 11;
-}
+  string batch_type = 7;
```
bergundy: Are all of these fields in use? Why do we need to duplicate anything that is present on the request?

service/frontend/workflow_handler_test.go
```diff
@@ -2413,22 +2401,18 @@ func (s *WorkflowHandlerSuite) TestStartBatchOperation_WorkflowExecutions_Signal
 	wh := s.getWorkflowHandler(config)
 	signalPayloads := payloads.EncodeString(signalName)
 	params := &batchspb.BatchOperation{
```
bergundy: You can create the request before this and embed it in "params" instead of duplicating those in the test.

service/worker/batcher/activities.go
```diff
@@ -205,16 +212,16 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
...
-	if err := a.checkNamespace(batchParams.Namespace); err != nil {
+	if err := a.checkNamespaceID(batchParams.NamespaceId); err != nil {
 		metrics.BatcherOperationFailures.With(metricsHandler).Record(1)
 		logger.Error("Failed to run batch operation due to namespace mismatch", tag.Error(err))
 		return hbd, err
 	}
 
 	sdkClient := a.ClientFactory.NewClient(sdkclient.Options{
-		Namespace:     batchParams.Namespace,
+		Namespace:     batchParams.NamespaceId,
```
bergundy: SDK can't target a namespace by ID.

service/worker/batcher/activities.go
```diff
@@ -57,6 +57,13 @@ func (a *activities) checkNamespace(namespace string) error {
 	return nil
 }
 
+func (a *activities) checkNamespaceID(namespaceID string) error {
```
bergundy, line 61: checkNamespace (above) has a TODO to remove special handling, I wonder if we even need to check the namespace ID. I don't have the full context on why this is here, need to understand whether the `a.namespace` is always the system namespace.

service/worker/batcher/activities.go
```diff
@@ -226,10 +233,34 @@ func (a *activities) BatchActivityWithProtobuf(ctx context.Context, batchParams
 		}
 	}
 
-	adjustedQuery := a.adjustQuery(batchParams.Query, batchParams.BatchType)
+	var batchType string
```
bergundy: Don't we have the batch type already in the request?

service/worker/batcher/activities.go
```diff
@@ -626,7 +657,7 @@ func startTaskProcessorProtobuf(
 				err = processTask(ctx, limiter, task,
 					func(workflowID, runID string) error {
 						_, err := frontendClient.SignalWorkflowExecution(ctx, &workflowservice.SignalWorkflowExecutionRequest{
-							Namespace: batchOperation.Namespace,
+							Namespace: batchOperation.Request.Namespace,
```
bergundy: You should resolve the namespace using the ID via the namespace registry (that was the whole point of using the ID).

service/worker/batcher/workflow.go
bergundy: I see additional validation in workflow_handler at the top of the `StartBatchOperation` function.

service/worker/batcher/activities.go
bergundy: We already have batch type defined in proto enums, let's use that.

proto/internal/temporal/server/api/batch/v1/request_response.proto
spkane31: If by request you mean `StartBatchOperationRequest`, none of these are present on that [request](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L1662-L1694). If that isn't the request type you mean, what request has this information?

service/worker/batcher/activities.go
spkane31: Not explicitly, we have it in the `oneof operation {...}` portion of the `StartBatchOperationRequest`. We can use this where we use `batchType`, but the code becomes less re-usable with the non-protobuf version and harder to read with `switch` statements on the operation type.

This comment is also dependent on how many fields in `BatchOperationInput` you want to remove from your earlier comment

proto/internal/temporal/server/api/batch/v1/request_response.proto
spkane31: `activity_heartbeat_timeout` is used to set heartbeat timeout on the batch activity.

`attempts_on_retryable_error` and  `non_retryable_errors` are used by the task processing code.

`rps` is used by the rate limiter in task processing.

`concurrency` is used by task processing but comments say it can be removed in 1.19+ so might safe to remove.

service/worker/batcher/activities.go
spkane31: Disregard, fixed

service/worker/batcher/activities.go
spkane31: `a.namespace.String() != primitives.SystemLocalNamespace` based on my interpretation of this, these activities always run in the system namespace. In which case I think we'd be able to remove them

service/worker/batcher/activities.go
spkane31: If these run in the system namespace do we still have to do that? Can we pass in the activity default namespace?

service/frontend/workflow_handler.go
```diff
@@ -4763,6 +4598,23 @@ func (wh *WorkflowHandler) StartBatchOperation(
 	return &workflowservice.StartBatchOperationResponse{}, nil
 }
 
+func snakeCaseBatchType(batchType enumspb.BatchOperationType) string {
```
spkane31, line 4601: To maintain consistency with how we do these now a few of the batch operation types needed special casing

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -0,0 +1,27 @@
...
+option go_package = "go.temporal.io/server/api/batch/v1;batch";
+
+import "temporal/api/workflowservice/v1/request_response.proto";
+import "temporal/api/enums/v1/batch_operation.proto";
+import "google/protobuf/duration.proto";
+
+message BatchOperationInput { 
+  string namespace_id = 1; 
+
+  double rps = 2;
```
bergundy: Lets remove `rps` as it is a computed property.

service/worker/batcher/activities.go
```diff
@@ -464,6 +611,227 @@ func startTaskProcessor(
...
+				err = processTask(ctx, limiter, task,
+					func(workflowID, runID string) error {
+						unpauseRequest := &workflowservice.UnpauseActivityRequest{
+							Namespace: namespace,
+							Execution: &commonpb.WorkflowExecution{
+								WorkflowId: workflowID,
+								RunId:      runID,
+							},
+							Identity:       operation.UnpauseActivitiesOperation.Identity,
+							ResetAttempts:  !operation.UnpauseActivitiesOperation.ResetAttempts,
```
bergundy: Why is there a not here?

service/frontend/workflow_handler.go
```diff
@@ -4513,216 +4497,67 @@ func (wh *WorkflowHandler) StartBatchOperation(
...
-		operationType = batcher.BatchTypeUpdateActivitiesOptions
-		if op.UpdateActivityOptionsOperation == nil {
-			return nil, serviceerror.NewInvalidArgument("update activity options operation is not set")
-		}
-		if op.UpdateActivityOptionsOperation.GetActivityOptions() != nil && op.UpdateActivityOptionsOperation.GetRestoreOriginal() {
-			return nil, serviceerror.NewInvalidArgument("cannot set both activity options and restore original")
-		}
-		if op.UpdateActivityOptionsOperation.GetActivityOptions() == nil && !op.UpdateActivityOptionsOperation.GetRestoreOriginal() {
-			return nil, serviceerror.NewInvalidArgument("Either activity type must be set, or restore original should be set to true")
-		}
```
bergundy, line 4654: Can you confirm that these validations are covered in `batcher.ValidateBatchOperation`? Same as any other validation that was removed from this function.

service/worker/batcher/activities.go
```diff
@@ -464,6 +611,227 @@ func startTaskProcessor(
...
+		case task := <-taskCh:
+			if isDone(ctx) {
+				return
+			}
+			var err error
+
+			switch operation := batchOperation.Request.Operation.(type) {
+			case *workflowservice.StartBatchOperationRequest_TerminationOperation:
+				err = processTask(ctx, limiter, task,
+					func(workflowID, runID string) error {
```
bergundy: nit: you can refactor this to take a `*commonpb.WorkflowExecution`

service/worker/batcher/activities.go
bergundy: Unaddressed comment.

service/worker/batcher/activities.go
```diff
@@ -464,6 +611,227 @@ func startTaskProcessor(
...
+							UpdateMask:      &fieldmaskpb.FieldMask{Paths: operation.UpdateActivityOptionsOperation.UpdateMask.Paths},
+							RestoreOriginal: operation.UpdateActivityOptionsOperation.RestoreOriginal,
+							Identity:        operation.UpdateActivityOptionsOperation.Identity,
+						}
+
+						switch ao := operation.UpdateActivityOptionsOperation.GetActivity().(type) {
+						case *batchpb.BatchOperationUpdateActivityOptions_Type:
+							updateRequest.Activity = &workflowservice.UpdateActivityOptionsRequest_Type{Type: ao.Type}
+						case *batchpb.BatchOperationUpdateActivityOptions_MatchAll:
+							updateRequest.Activity = &workflowservice.UpdateActivityOptionsRequest_MatchAll{MatchAll: true}
```
bergundy, line 810: Same here, error out if you don't recognize the type.

service/worker/batcher/activities.go
spkane31: good catch, removed

service/frontend/workflow_handler.go
spkane31: Added a couple more sanity checks, afaict everything is copied over.

service/frontend/workflow_handler.go
```diff
@@ -4965,7 +4817,7 @@ func (wh *WorkflowHandler) ListBatchOperations(
 		NextPageToken: request.GetNextPageToken(),
 		Query: fmt.Sprintf("%s = '%s' and %s = '%s'",
 			searchattribute.WorkflowType,
-			batcher.BatchWFTypeName,
+			batcher.BatchWFTypeProtobufName,
```
dnr, line 4820: This may cause some confusion around the deployment transition. Is it worth listing both types temporarily?

service/worker/batcher/activities.go
dnr: `a.namespace` should never be the system namespace anymore, that was for the transition from batcher v1 (system ns) to batcher v2 (user ns).

we don't really need to pass the nsid to the activity, it knows it already.

service/worker/batcher/activities.go
```diff
@@ -198,6 +208,128 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
 	return hbd, nil
 }
 
+// BatchActivityWithProtobuf is an activity for processing batch operations using protobuf as the input type.
+// nolint:revive,cognitive-complexity
+func (a *activities) BatchActivityWithProtobuf(ctx context.Context, batchParams *batchspb.BatchOperationInput) (HeartBeatDetails, error) {
+	logger := a.getActivityLogger(ctx)
+	hbd := HeartBeatDetails{}
```
dnr, line 215: is it worth considering moving heartbeat details to proto too? if we don't do it now we never can :)

service/worker/batcher/activities.go
```diff
@@ -198,6 +208,128 @@ func (a *activities) BatchActivity(ctx context.Context, batchParams BatchParams)
...
+		}
+		hbd.TotalEstimate = estimateCount
+	}
+	rps := a.rps(a.namespace.String())
+	rateLimit := rate.Limit(rps)
+	burstLimit := rps // should never be zero because everything would be rejected
+	rateLimiter := rate.NewLimiter(rateLimit, burstLimit)
+	taskCh := make(chan taskDetail, pageSize)
+	respCh := make(chan error, pageSize)
+	for i := 0; i < a.getOperationConcurrency(int(batchParams.Concurrency)); i++ {
```
dnr, line 260:
```suggestion
	for range a.getOperationConcurrency(int(batchParams.Concurrency)) {
```

service/worker/batcher/activities.go
```diff
@@ -261,22 +407,22 @@ func startTaskProcessor(
...
-					func(workflowID, runID string) error {
+					func(execution *commonpb.WorkflowExecution) error {
 						_, err := frontendClient.SignalWorkflowExecution(ctx, &workflowservice.SignalWorkflowExecutionRequest{
 							Namespace: batchParams.Namespace,
 							WorkflowExecution: &commonpb.WorkflowExecution{
-								WorkflowId: workflowID,
-								RunId:      runID,
+								WorkflowId: execution.WorkflowId,
+								RunId:      execution.RunId,
 							},
```
dnr, line 426:
```suggestion
							WorkflowExecution: execution,
```

service/worker/batcher/activities.go
```diff
@@ -285,22 +431,22 @@ func startTaskProcessor(
...
-					func(workflowID, runID string) error {
+					func(execution *commonpb.WorkflowExecution) error {
 						_, err := frontendClient.DeleteWorkflowExecution(ctx, &workflowservice.DeleteWorkflowExecutionRequest{
 							Namespace: batchParams.Namespace,
 							WorkflowExecution: &commonpb.WorkflowExecution{
-								WorkflowId: workflowID,
-								RunId:      runID,
+								WorkflowId: execution.WorkflowId,
+								RunId:      execution.RunId,
 							},
```
dnr, line 440:
```suggestion
							WorkflowExecution: execution,
```
and same thing repeated several more times

service/worker/batcher/activities.go
```diff
@@ -464,11 +611,238 @@ func startTaskProcessor(
...
+					})
+			case *workflowservice.StartBatchOperationRequest_SignalOperation:
+				err = processTask(ctx, limiter, task,
+					func(execution *commonpb.WorkflowExecution) error {
+						_, err := frontendClient.SignalWorkflowExecution(ctx, &workflowservice.SignalWorkflowExecutionRequest{
+							Namespace: namespace,
+							WorkflowExecution: &commonpb.WorkflowExecution{
+								WorkflowId: execution.WorkflowId,
+								RunId:      execution.RunId,
+							},
```
dnr, line 656:
```suggestion
							WorkflowExecution: execution,
```
and more below

service/frontend/workflow_handler.go
bergundy: Oh I missed this in the review, in person I requested to keep both. Addressed in #8144. We should merge that change before cutting a new server release. I would consider reverting this one until this fix is in.

PR #8084, opened 2025-07-22
Enhance DLQWriter to include namespace state in metrics

simvlad:
## What changed?
- Updated the `WriteTaskToDLQ` method to accept a new parameter `isNamespaceActive` to indicate the state of the namespace (active or passive).
- For replication tasks, the namespace state is always pasive, for history queue tasks determine the state based on LastActiveness, which represents if the task was last executed as active or standby
- Updated related tests to verify the inclusion of the new tag in metrics.

## Why?
To provide better insights into the state of namespaces when tasks are written to the DLQ.

## How did you test it?
- [X] built
- [X] added new unit test(s)


service/history/queues/dlq_writer.go
```diff
@@ -86,7 +87,16 @@ func (q *DLQWriter) WriteTaskToDLQ(
 	if err != nil {
 		return fmt.Errorf("%w: %v", ErrSendTaskToDLQ, err)
 	}
-	metrics.DLQWrites.With(q.metricsHandler).Record(1, metrics.TaskCategoryTag(task.GetCategory().Name()))
+	// "passive" means the namespace is in standby mode and only replicates data
+	namespaceState := "passive"
+	if isNamespaceActive {
+		namespaceState = "active"
+	}
```
yycptt: Is NamespaceStateTag used in other places and do we have any constant defined for its values? It will be ideally if all usages are using the same value.

service/history/queues/executable.go
```diff
@@ -375,6 +375,7 @@ func (e *executableImpl) writeToDLQ(ctx context.Context) error {
 		currentClusterName,
 		tasks.GetShardIDForTask(e.Task, int(numShards)),
 		e.GetTask(),
+		e.lastActiveness,
```
yycptt, line 378: 👍 I almost forget we have this field in executable :)

yycptt (approved):

PR #8089, opened 2025-07-23
make `RampingVersionPercentageChangedTime ` visible

Shivs11:
## What changed?
- Made `RampingVersionPercentageChangedTime` visible to a client (be it the worker-controller or an operator carrying out a `DescribeDeployment` call) from the `RoutingConfig`.

## Why?
- Safe deploys correctness. Also, this is a big bug. I'm very sorry for skipping over this.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None, improves the feature. Tested my changes out as well.

### DIFF
--- a/service/worker/workerdeployment/workflow.go
+++ b/service/worker/workerdeployment/workflow.go
@@ -557,6 +557,7 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
 	d.State.RoutingConfig.RampingDeploymentVersion = worker_versioning.ExternalWorkerDeploymentVersionFromStringV31(newRampingVersion)
 	d.State.RoutingConfig.RampingVersionPercentage = args.Percentage
 	d.State.RoutingConfig.RampingVersionChangedTime = rampingVersionUpdateTime
+	d.State.RoutingConfig.RampingVersionPercentageChangedTime = routingUpdateTime
 	d.State.ConflictToken, _ = routingUpdateTime.AsTime().MarshalBinary()
 	d.State.LastModifierIdentity = args.Identity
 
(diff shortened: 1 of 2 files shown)


carlydf (commented):
looks good -- does this need a replay test?

service/worker/workerdeployment/workflow.go
```diff
@@ -557,6 +557,7 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
 	d.State.RoutingConfig.RampingDeploymentVersion = worker_versioning.ExternalWorkerDeploymentVersionFromStringV31(newRampingVersion)
 	d.State.RoutingConfig.RampingVersionPercentage = args.Percentage
 	d.State.RoutingConfig.RampingVersionChangedTime = rampingVersionUpdateTime
+	d.State.RoutingConfig.RampingVersionPercentageChangedTime = routingUpdateTime
```
Shivs11, line 560: @dnr - pinging ya since I wanted to confirm my hunch/understanding of what will/won't break NDE;

question: Will adding this extra field to the struct `RoutingConfig` make our entity workflows NDE?
context: `d.State.RoutingConfig` is currently being stored as part of the workflow's memo via `UpsertMemo` calls. In other words, we are storing the state present in the routingConfig as a history event.

My initial assumption is no, this wont' cause an NDE since changing/adding inputs does not cause NDE (unless they literally trigger the addition or removal of history events and this does not)

dnr: afaik, changing the contents of a memo update does not count as NDE.

potential caveat: do you ever read the memo from workflow.GetInfo().Memo for this workflow directly? it looks like no. if you did, things could get confused since the memo you read would be different from the memo you set. but if you only set here then I think it's fine.

carlydf (approved):

service/worker/workerdeployment/workflow.go
Shivs11: I also double checked some of the [SDK](https://github.com/temporalio/sdk-go/blob/ce09baa8b3b99d6e36d77465caf08af25210018d/internal/internal_task_handlers.go#L1611) code to see if inputs were being compared and don't think that's the case;

> do you ever read the memo from workflow.GetInfo().Memo for this workflow directly
nope, the memo is only being used for `List` RPC calls and we are fetching the information required in those requests from here

the only other caveat I can think about it the fact that this struct, `RoutingConfig`, is passed as input for every CAN run. I wonder if that is a problem but I don't think so?

Shivs11:
merging this for now since I am fairly confident this won't cause NDE's in our CICD env where we have these wfs running; can also quickly revert if I do notice things might break if @dnr answers to my replied caveat by saying there is a potential 🔥

PR #8093, opened 2025-07-24
Fix race with evicted fair task

dnr:
## What changed?
Handle rare race condition with a task that gets evicted from fairTaskReader in-mem buffer before it was added to the matcher.

## Why?
fix race

## How did you test it?

### DIFF
--- a/service/matching/pri_matcher.go
+++ b/service/matching/pri_matcher.go
@@ -261,7 +261,7 @@ func (tm *priTaskMatcher) validateTasksOnRoot(lim quotas.RateLimiter, retrier ba
 }
 
 func (tm *priTaskMatcher) forwardPolls() {
-	forwarderTask := &internalTask{isPollForwarder: true}
+	forwarderTask := newPollForwarderTask()
 	ctxs := []context.Context{tm.tqCtx}
 	for {
 		res := tm.data.EnqueueTaskAndWait(ctxs, forwarderTask)
@@ -437,7 +437,9 @@ func (tm *priTaskMatcher) OfferNexusTask(ctx context.Context, task *internalTask
 }
 
 func (tm *priTaskMatcher) AddTask(task *internalTask) {
-	task.removeFromMatcher = func() { tm.data.RemoveTask(task) }
+	if !task.setRemoveFunc(func() { tm.data.RemoveTask(task) }) {
+		return
+	}
 	tm.data.EnqueueTaskNoWait(task)
 }
 
--- a/service/matching/task.go
+++ b/service/matching/task.go
@@ -42,6 +42,8 @@ type (
 	// internalTask represents an activity, workflow, query or started (received from another host).
 	// this struct is more like a union and only one of [ query, event, forwarded ] is
 	// non-nil for any given task
+	// TODO(pri): after deprecating classic matcher, we can consolidate backlogCountHint, recycleToken,
+	// and removeFromMatcher into a single *physicalTaskQueueManager field.
 	internalTask struct {
 		event            *genericTaskInfo // non-nil for activity or workflow task that's locally generated
 		query            *queryTaskInfo   // non-nil for a query task that's locally sync matched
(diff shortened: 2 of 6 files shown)


service/matching/task.go
```diff
@@ -268,6 +281,24 @@ func (task *internalTask) fairLevel() fairLevel {
...
+
+// setRemoveFunc sets the function to remove the task from the matcher.
+// It returns true if the task is still valid and the function was set,
+// false if the task was evicted already and should not be added.
+func (task *internalTask) setRemoveFunc(remove func()) bool {
+	return task.removeFromMatcher.CompareAndSwap(&removeFuncNotAddedYet, &remove)
+}
+
+// setEvicted marks the task as evicted. If it was added to a matcher it will be removed.
+func (task *internalTask) setEvicted() {
```
stephanos, line 297: nit: maybe `markEvicted`?

service/matching/task.go
```diff
@@ -268,6 +281,24 @@ func (task *internalTask) fairLevel() fairLevel {
 	return fairLevelFromAllocatedTask(task.event.AllocatedTaskInfo)
 }
 
+// resetMatcherState should be called before adding or re-adding a backlog task to priMatcher.
```
stephanos:
```suggestion
// resetMatcherState must be called before adding or re-adding a backlog task to priMatcher.
```

right?

service/matching/pri_matcher.go
```diff
@@ -437,7 +437,9 @@ func (tm *priTaskMatcher) OfferNexusTask(ctx context.Context, task *internalTask
 }
 
 func (tm *priTaskMatcher) AddTask(task *internalTask) {
-	task.removeFromMatcher = func() { tm.data.RemoveTask(task) }
+	if !task.setRemoveFunc(func() { tm.data.RemoveTask(task) }) {
+		return
```
stephanos: nit: maybe a short comment to explain what's going on?

service/matching/task.go
```diff
@@ -268,6 +281,24 @@ func (task *internalTask) fairLevel() fairLevel {
...
+func (task *internalTask) resetMatcherState() {
+	task.removeFromMatcher.Store(&removeFuncNotAddedYet)
+}
+
+// setRemoveFunc sets the function to remove the task from the matcher.
+// It returns true if the task is still valid and the function was set,
+// false if the task was evicted already and should not be added.
+func (task *internalTask) setRemoveFunc(remove func()) bool {
+	return task.removeFromMatcher.CompareAndSwap(&removeFuncNotAddedYet, &remove)
```
stephanos, line 293: For my own understanding; since we don't (can't?) initialize every instance of `internalTask` with `removeFuncNotAddedYet`, `resetMatcherState` is called to ensure it's set. If a task has a `nil` there, this would return `false` and signal it was already evicted.

stephanos (approved):
Feel free to merge as-is and address the nits later given that it's a busy day today.

service/matching/pri_matcher.go
dnr: I had a comment here and then I decided the comment on setRemoveFunc was enough?

service/matching/task.go
dnr: sure, fixed another too

service/matching/task.go
dnr: We could and I actually did that at first, but reverted it to keep things simpler.

This whole thing is only relevant for backlog tasks going to priMatcher, for others it turns out that the relevant functions are never called, so it doesn't matter.

There's definitely room for improvement in documenting the lifecycle and state transitions of an internalTask. Currently it's basically implicit and spread across the task reader(s), backlog manager(s), matcher(s), matchingEngine, and probably some more places. There are some clear invariants and valid/invalid transitions to extract but they're not enforced right now.

service/matching/task.go
dnr: well, it's not just marking, it's also maybe removing

PR #8094, opened 2025-07-24
Implementation for DescribeWorker API

ychebotarev:
## What changed?
* Add implementation for DescribeWorker API (to both frontend and matching)
* Add feature flag
* Add all quotas/etc.
* Add unit tests
* Add functional tests

## Why?

Part of worker commands work. 
Users should be able to directly get worker info for the worker they need.
Corresponding API PR: https://github.com/temporalio/api/pull/622

## How did you test it?
- [X] add unit tests
- [X] add functional tests

## Potential risks
None, behind feature flag

### DIFF
--- a/service/matching/workers/registry_impl.go
+++ b/service/matching/workers/registry_impl.go
@@ -123,12 +123,12 @@ func (b *bucket) getWorkerHeartbeat(nsID namespace.ID, workerInstanceKey string)
 
 	mp, ok := b.namespaces[nsID]
 	if !ok {
-		return nil, serviceerror.NewNotFoundf("namespace not found: %s", nsID.String())
+		return nil, serviceerror.NewNamespaceNotFound(nsID.String())
 	}
 
 	e, exists := mp[workerInstanceKey]
 	if !exists {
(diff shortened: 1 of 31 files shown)


service/matching/workers/registry_impl.go
```diff
@@ -116,6 +117,23 @@ func (b *bucket) filterWorkers(
 	return out
 }
 
+func (b *bucket) getWorkerHeartbeat(nsID namespace.ID, workerInstanceKey string) (*workerpb.WorkerHeartbeat, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	mp, ok := b.namespaces[nsID]
+	if !ok {
+		return nil, serviceerror.NewNotFoundf("namespace not found: %s", nsID.String())
```
dnr: we actually have a more specific serviceerror.NewNamespaceNotFound(ns). should we use that instead? we have only an id instead of a name but maybe it's okay?

dnr (approved):

service/matching/workers/registry_impl.go
ychebotarev: yeah, good point. It seems we use both, id and name, for NewNamespaceNotFound, so using id is ok.

PR #8099, opened 2025-07-25
Matching task fairness SQL support

dnr:
## What changed?
Add SQL support for matching task fairness.

Based on #8043 with some changes to reduce code duplication (can be applied to cassandra later).

## Why?
Support all persistence implementations.

## How did you test it?
- [x] added new unit test(s)


sivagirish81 (approved):
Looks good to me.
I noticed in Cassandra we have 
> CompleteTasksLessThan - deletes all tasks less than the given task id.

Do we need this in the SQL interface as well or is it not required?

schema/postgresql/v12/temporal/versioned/v1.18/tasks_v2.sql
```diff
@@ -0,0 +1,23 @@
+-- Stores activity or workflow tasks
+-- Used for fairness scheduling. (pass, task_id) are monotonically increasing.
+CREATE TABLE tasks_v2 (
```
sivagirish81, line 3: There is an issue with this schema.
This is an SQL specific schema.

sivagirish81 (changes requested):

schema/postgresql/v12/temporal/versioned/v1.18/tasks_v2.sql
```diff
@@ -0,0 +1,23 @@
...
+-- Stores ephemeral task queue information such as ack levels and expiry times
+CREATE TABLE task_queues_v2 (
+  range_hash BIGINT NOT NULL,
+  task_queue_id BYTEA NOT NULL,
+  --
+  range_id BIGINT NOT NULL,
+  data BYTEA NOT NULL,
+  data_encoding VARCHAR(16) NOT NULL,
+  PRIMARY KEY (range_hash, task_queue_id)
+);
```
sivagirish81, line 23:
```suggestion
-- Stores activity or workflow tasks
-- Used for fairness scheduling. (pass, task_id) are monotonically increasing.
CREATE TABLE tasks_v2 (
  range_hash BIGINT NOT NULL,
  task_queue_id BYTEA NOT NULL,
  pass BIGINT NOT NULL, -- pass for tasks (see stride scheduling algorithm for fairness)
  task_id BIGINT NOT NULL,
  --
  data BYTEA NOT NULL,
  data_encoding VARCHAR(16) NOT NULL,
  PRIMARY KEY (range_hash, task_queue_id, pass, task_id)
);

-- Stores ephemeral task queue information such as ack levels and expiry times
CREATE TABLE task_queues_v2 (
  range_hash BIGINT NOT NULL,
  task_queue_id BYTEA NOT NULL,
  --
  range_id BIGINT NOT NULL,
  data BYTEA NOT NULL,
  data_encoding VARCHAR(16) NOT NULL,
  PRIMARY KEY (range_hash, task_queue_id)
);

```

sivagirish81 (changes requested):

common/persistence/sql/task_user_data.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		return nil, err
+	}
+	return &persistence.InternalGetTaskQueueUserDataResponse{
+		Version:  response.Version,
+		UserData: persistence.NewDataBlob(response.Data, response.DataEncoding),
+	}, nil
+}
+
+func (uds *userDataStore) UpdateTaskQueueUserData(ctx context.Context, request *persistence.InternalUpdateTaskQueueUserDataRequest) error {
```
sivagirish81, line 40: We can move  most parts of the transaction function to a separate function to an applyUserDataUpdate and call it inside the transaction function like this.
```
err = uds.txExecute(ctx, "UpdateTaskQueueUserData", func(tx sqlplugin.Tx) error {
		for taskQueue, update := range request.Updates {
			if err := uds.applyUserDataUpdate(ctx, tx, namespaceID, taskQueue, update); err != nil {
				return err
			}
		}
		return nil
	})
```

schema/postgresql/v12/temporal/versioned/v1.18/tasks_v2.sql
```diff
@@ -0,0 +1,23 @@
...
+CREATE TABLE tasks_v2 (
+  range_hash INT UNSIGNED NOT NULL,
+  task_queue_id VARBINARY(272) NOT NULL,
+  pass BIGINT NOT NULL, -- pass for tasks (see stride scheduling algorithm for fairness)
+  task_id BIGINT NOT NULL,
+  --
+  data MEDIUMBLOB NOT NULL,
+  data_encoding VARCHAR(16) NOT NULL,
+  PRIMARY KEY (range_hash, task_queue_id, pass, task_id)
+);
```
sivagirish81:
```suggestion
-- Stores activity or workflow tasks
-- Used for fairness scheduling. (pass, task_id) are monotonically increasing.
CREATE TABLE tasks_v2 (
  range_hash BIGINT NOT NULL,
  task_queue_id BYTEA NOT NULL,
  pass BIGINT NOT NULL, -- pass for tasks (see stride scheduling algorithm for fairness)
  task_id BIGINT NOT NULL,
  --
  data BYTEA NOT NULL,
  data_encoding VARCHAR(16) NOT NULL,
  PRIMARY KEY (range_hash, task_queue_id, pass, task_id)
);
```
Blocking : This is the correct Postgres Schema. The previous schema was a MySQL specific schema for the tasks_v2 table.

common/persistence/sql/task_user_data.go
dnr: I'd rather not change anything here and just keep to moving code/renaming files wherever possible

schema/postgresql/v12/temporal/versioned/v1.18/tasks_v2.sql
dnr: Ah, I see, the schema.sql is fine but the upgrade is broken?

dnr:
> Looks good to me. I noticed in Cassandra we have
> 
> > CompleteTasksLessThan - deletes all tasks less than the given task id.
> 
> Do we need this in the SQL interface as well or is it not required?

It's in there, it gets translated to DeleteFromTasksV2

dnr:
> It's in there, it gets translated to DeleteFromTasksV2

oh, it's missing limit support. I'll add that.

common/persistence/sql/sqlplugin/matching_task_queue.go
```diff
@@ -40,56 +39,38 @@ type (
...
+	MatchingTaskVersion2 MatchingTaskVersion = 2
+)
+
+var switchTaskQueuesTableV2Cache sync.Map
+
+func SwitchTaskQueuesTable(baseQuery string, v MatchingTaskVersion) string {
+	if v == MatchingTaskVersion1 {
+		return baseQuery
+	} else if v != MatchingTaskVersion2 {
+		return "_invalid_version_"
```
stephanos, line 68: Should this maybe panic instead?

schema/sqlite/v3/temporal/schema.sql
```diff
@@ -93,6 +93,30 @@ CREATE TABLE task_queues (
...
+	pass BIGINT NOT NULL, -- pass for tasks (see stride scheduling algorithm for fairness)
+	task_id BIGINT NOT NULL,
+	--
+	data MEDIUMBLOB NOT NULL,
+	data_encoding VARCHAR(16) NOT NULL,
+	PRIMARY KEY (range_hash, task_queue_id, pass, task_id)
+);
+
+-- Stores ephemeral task queue information such as ack levels and expiry times
+CREATE TABLE task_queues_v2 (
```
stephanos, line 110: The separate task queue table is needed to ensure consistency for v2 tasks and the task queue data?

stephanos (approved):

common/persistence/sql/sqlplugin/matching_task_queue.go
dnr: I considered that but I suppose we do like to avoid panics, and this will clearly fail as a sql statement so it can't really cause any harm.

schema/sqlite/v3/temporal/schema.sql
dnr: Technically for SQL we could use the same task_queues table since we can do cross-table transactions. But it would get pretty confusing: we could combine the metadata for both tables in the same blob. Or we could use a prefix on task_queue_id (but then we might as well use another table).

Basically I think the main thing is: for cassandra we have to keep them separate (the fair tasks and fifo tasks are just separate queues entirely), so the code assumes that, so for sql we should do the same to avoid confusion.

common/persistence/sql/sqlplugin/matching_task_queue.go
stephanos: Since this is a developer-provided value, not a user-provided value, the wrong value would definitely be a clear bug and warrant a panic IMO. Diagnosing an `_invalid_version_` SQL error seems not straightfoward; whereas a panic would be very clear. But it's true that a panic always poses some risk and we want to minimize their use.

Anyway, it's fine as I suspect this will never happen (famous last words).

schema/sqlite/v3/temporal/schema.sql
stephanos: 👍 makes sense

common/persistence/sql/sqlplugin/matching_task_queue.go
dnr: I'll port the same mechanism to cassandra and can change to a panic there

Issue #7939, opened 2025-06-18
Unclear error message on Postgres misconfiguration: 'no usable database connection found'


vitorvezani:
I was trying to install the latest Temporal using the Temporal chart and Postgres as default and visibility stores.
I managed to get the schema job working, but other workloads that connect to Postgres failed with a **generic** error.

## Expected Behavior
Receive a helpful error message that assists in debugging while using Postgres for both default and visibility stores. For example:
When using Temporal chart 0.55 with Temporal server 1.24.2, I got the following error:

```
sql schema version compatibility check failed: pq: Private key has world access. Permissions should be u=rw,g=r (0640) if owned by root, or u=rw (0600), or less
```

This clearly explains the root cause of the failure.

## Actual Behavior
When using chart 0.63.0 and Temporal server 1.27.2, I received this error:

```
sql schema version compatibility check failed: unable to read DB schema version keyspace/database: temporal error: no usable database connection found
```

This message is too generic and doesn't help with debugging.

## Steps to Reproduce the Problem

  1.
  1.
  1.

## Specifications

  - Version: 1.27.2
  - Platform: Kubernetes + SQL

abusutil:
Hi @vitorvezani I have same issue with my temporal cluster (same version).

I have 2 replicas of temporal services, and sometimes, one of them pass in failed state with this error.

I ran Postgres on RDS

```
TEMPORAL_ADDRESS is not set, setting it to 10.1.144.137:7233
2025/06/20 14:40:12 Loading config; env=docker,zone=,configDir=config
2025/06/20 14:40:12 Loading config files=[config/docker.yaml]
[Fx] PROVIDE	fx.Lifecycle <= go.uber.org/fx.New.func1()
[Fx] PROVIDE	fx.Shutdowner <= go.uber.org/fx.(*App).shutdowner-fm()
[Fx] PROVIDE	fx.DotGraph <= go.uber.org/fx.(*App).dotGraph-fm()
[Fx] PROVIDE	*temporal.ServerImpl <= go.temporal.io/server/temporal.NewServerFxImpl()
[Fx] PROVIDE	*temporal.serverOptions <= go.temporal.io/server/temporal.ServerOptionsProvider()
[Fx] PROVIDE	chan interface {} <= go.temporal.io/server/temporal.ServerOptionsProvider()
[Fx] PROVIDE	temporal.synchronizationModeParams <= go.temporal.io/server/temporal.ServerOptionsProvider()
[...]

PR #8100, opened 2025-07-25
Log errors when doing version compatibility check for Cassandra and SQL

simvlad:
## What changed?
Changes version compability checks to pass the real logger, so that the error messages would be seeing in the log

## Why?

Before, the error was generic and there were no logs indicating the real error (see https://github.com/temporalio/temporal/issues/7939):
```
sql schema version compatibility check failed: unable to read DB schema version keyspace/database: temporal error: no usable database connection found
```

With this change, the log would contain the actual error:
```
{"level":"error","ts":"2025-07-25T15:30:49.660-0700","msg":"sql handle: unable to refresh database connection pool","error":"pq: no pg_hba.conf entry for host \"127.0.0.1\", ...
```

The alternative solution would be to return the error, but the design of common/persistence/sql/sqlplugin/db_handle.go suggests we hold the abstraction boundaries, by returning generic errors and logging specific errors.

## How did you test it?
- [X] built

### DIFF
--- a/common/persistence/cassandra/version_checker.go
+++ b/common/persistence/cassandra/version_checker.go
@@ -19,9 +19,10 @@ import (
 func VerifyCompatibleVersion(
 	cfg config.Persistence,
 	r resolver.ServiceResolver,
+	logger log.Logger,
 ) error {
 
-	if err := checkMainKeyspace(cfg, r); err != nil {
+	if err := checkMainKeyspace(cfg, r, logger); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


common/persistence/cassandra/version_checker.go
```diff
@@ -19,37 +19,36 @@ import (
...
-		return CheckCompatibleVersion(*ds.Cassandra, r, cassandraschema.Version)
+		return checkCompatibleVersion(*ds.Cassandra, r, cassandraschema.Version, logger)
 	}
 	return nil
 }
 
-// CheckCompatibleVersion check the version compatibility
-func CheckCompatibleVersion(
+// checkCompatibleVersion check the version compatibility
+func checkCompatibleVersion(
```
yycptt: Need to keep this exported. It's used in internal repos. The internal repo needs to be updated as well

yycptt (approved):
> the design of common/persistence/sql/sqlplugin/db_handle.go suggests we hold the abstraction boundaries, by returning generic errors and logging specific errors.

I was thinking only return the error on `NewDatabaseHandle` which eagerly sets up the connection. Didn't realize we are using no-op logger today and not even logging the real error. This approach LGTM.

PR #8102, opened 2025-07-28
Add subqueueKey and priorityKey types

stephanos:
## What changed?

Added new types `subqueueKey` and `priorityKey`.

## Why?

Instead of int/int32, the types help with documentation and incorrect usage.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/backlog_manager.go
+++ b/service/matching/backlog_manager.go
@@ -193,7 +193,7 @@ func (c *backlogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
 }
 
 func (c *backlogManagerImpl) BacklogStatsByPriority() map[int32]*taskqueuepb.TaskQueueStats {
-	defaultPriority := defaultPriorityLevel(c.config.PriorityLevels())
+	defaultPriority := int32(defaultPriorityLevel(c.config.PriorityLevels()))
 	return map[int32]*taskqueuepb.TaskQueueStats{
 		defaultPriority: &taskqueuepb.TaskQueueStats{
 			ApproximateBacklogCount: c.db.getTotalApproximateBacklogCount(),
--- a/service/matching/config.go
+++ b/service/matching/config.go
@@ -441,6 +441,6 @@ func newTaskQueueConfig(tq *tqid.TaskQueue, config *Config, ns namespace.Name) *
 	}
 }
 
-func defaultPriorityLevel(priorityLevels int32) int32 {
-	return (priorityLevels + 1) / 2
+func defaultPriorityLevel(priorityLevels int32) priorityKey {
+	return priorityKey(priorityLevels+1) / 2
 }
--- a/service/matching/config.go
+++ b/service/matching/config.go
@@ -140,7 +140,7 @@ type (
 		MinTaskThrottlingBurstSize func() int
 		MaxTaskDeleteBatchSize     func() int
 		TaskDeleteInterval         func() time.Duration
-		PriorityLevels             func() int32
+		PriorityLevels             func() priorityKey
 
 		GetUserDataLongPollTimeout dynamicconfig.DurationPropertyFn
 		GetUserDataMinWaitTime     time.Duration
@@ -360,8 +360,8 @@ func newTaskQueueConfig(tq *tqid.TaskQueue, config *Config, ns namespace.Name) *
 		TaskDeleteInterval: func() time.Duration {
 			return config.TaskDeleteInterval(ns.String(), taskQueueName, taskType)
 		},
-		PriorityLevels: func() int32 {
-			return int32(config.PriorityLevels(ns.String(), taskQueueName, taskType))
+		PriorityLevels: func() priorityKey {
+			return priorityKey(config.PriorityLevels(ns.String(), taskQueueName, taskType))
 		},
 		GetUserDataLongPollTimeout: config.GetUserDataLongPollTimeout,
 		GetUserDataMinWaitTime:     1 * time.Second,
@@ -441,6 +441,6 @@ func newTaskQueueConfig(tq *tqid.TaskQueue, config *Config, ns namespace.Name) *
 	}
 }
 
-func defaultPriorityLevel(priorityLevels int32) priorityKey {
+func defaultPriorityLevel(priorityLevels priorityKey) priorityKey {
 	return priorityKey(priorityLevels+1) / 2
 }
--- a/service/matching/db.go
+++ b/service/matching/db.go
@@ -27,7 +27,7 @@ const (
 
 	// Subqueue zero corresponds to "the queue" before migrating metadata to subqueues.
 	// For backwards compatibility, some operations only apply to subqueue zero for now.
-	subqueueZero = subqueueKey(0)
+	subqueueZero = subqueueIndex(0)
 )
 
 type (
@@ -58,10 +58,10 @@ type (
 		subqueues []persistencespb.SubqueueInfo
 	}
 
-	subqueueKey int
+	subqueueIndex int
 
 	createTasksResponse struct {
-		bySubqueue map[subqueueKey]subqueueCreateTasksResponse
+		bySubqueue map[subqueueIndex]subqueueCreateTasksResponse
 	}
 
 	subqueueCreateTasksResponse struct {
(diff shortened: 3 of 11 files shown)


service/matching/backlog_manager.go
```diff
@@ -193,7 +193,7 @@ func (c *backlogManagerImpl) BacklogStatus() *taskqueuepb.TaskQueueStatus {
 }
 
 func (c *backlogManagerImpl) BacklogStatsByPriority() map[int32]*taskqueuepb.TaskQueueStats {
```
stephanos, line 195: I wanted to change the map key type to `priorityKey`, too, but the caller passes the result into a proto-based type which uses int32 - which would require an extra map allocation + for loop to convert.

service/matching/fair_task_writer.go
```diff
@@ -29,8 +29,8 @@ type (
 
 		// state:
 		taskIDBlock        taskIDBlock
-		currentTaskIDBlock taskIDBlock             // copy of taskIDBlock for safe concurrent access via getCurrentTaskIDBlock()
-		counters           map[int]counter.Counter // subqueue -> Counter. only used in taskWriterLoop.
```
stephanos, line 33: Note: Comment is superfluous now.

service/matching/pri_backlog_manager.go
```diff
@@ -52,8 +52,8 @@ type (
 
 		subqueueLock        sync.Mutex
 		subqueues           []*priTaskReader // subqueue index -> fairTaskReader
-		subqueuesByPriority map[int32]int    // priority key -> subqueue index
-		priorityBySubqueue  map[int]int32    // subqueue index -> priority key
```
stephanos, line 56: Note: Comments are superfluous now.

service/matching/db.go
```diff
@@ -404,7 +406,7 @@ func (db *taskQueueDB) CreateTasks(
 		return createTasksResponse{}, nil
 	}
 
-	updates := make(map[int]subqueueCreateTasksResponse)
+	updates := make(map[subqueueKey]subqueueCreateTasksResponse)
 	allTasks := make([]*persistencespb.AllocatedTaskInfo, len(reqs))
 	allSubqueues := make([]int, len(reqs))
```
dnr, line 411: can we do
```suggestion
	allSubqueues := make([]subqueueKey, len(reqs))
```
?

service/matching/db.go
```diff
@@ -424,7 +426,7 @@ func (db *taskQueueDB) CreateTasks(
 	}
 
 	for sq, update := range updates {
-		db.subqueues[sq].ApproximateBacklogCount += int64(len(update.tasks))
+		db.subqueues[int(sq)].ApproximateBacklogCount += int64(len(update.tasks))
```
dnr: that cast is needed? I would have thought not

service/matching/config.go
```diff
@@ -441,6 +441,6 @@ func newTaskQueueConfig(tq *tqid.TaskQueue, config *Config, ns namespace.Name) *
 	}
 }
 
-func defaultPriorityLevel(priorityLevels int32) int32 {
-	return (priorityLevels + 1) / 2
+func defaultPriorityLevel(priorityLevels int32) priorityKey {
```
dnr: isn't it really
```suggestion
func defaultPriorityLevel(priorityLevels priorityKey) priorityKey {
```
?
I mean, "number of levels" is a count of things, not a level itself, but it's just one more than the highest level so I think we can do that

service/matching/fair_backlog_manager.go
```diff
@@ -152,26 +152,27 @@ func (c *fairBacklogManagerImpl) loadSubqueuesLocked(subqueues []persistencespb.
...
-		c.subqueuesByPriority[subqueues[i].Key.Priority] = i
-		c.priorityBySubqueue[i] = subqueues[i].Key.Priority
+		c.subqueuesByPriority[priorityKey(subqueues[i].Key.Priority)] = subqueueKey
+		c.priorityBySubqueue[subqueueKey] = priorityKey(subqueues[i].Key.Priority)
 	}
 }
 
-func (c *fairBacklogManagerImpl) getSubqueueForPriority(priority int32) int {
+func (c *fairBacklogManagerImpl) getSubqueueForPriority(priority priorityKey) subqueueKey {
 	levels := c.config.PriorityLevels()
```
dnr, line 167: I would do
```suggestion
	levels := priorityKey(c.config.PriorityLevels())
```
(based on suggestion above) and then the rest is simpler here

service/matching/fair_backlog_manager.go
```diff
@@ -32,9 +32,9 @@ type (
 		taskWriter *fairTaskWriter
 
 		subqueueLock        sync.Mutex
-		subqueues           []*fairTaskReader // subqueue index -> fairTaskReader
-		subqueuesByPriority map[int32]int     // priority key -> subqueue index
-		priorityBySubqueue  map[int]int32     // subqueue index -> priority key
+		subqueues           []*fairTaskReader           // subqueue index -> fairTaskReader
+		subqueuesByPriority map[priorityKey]subqueueKey // priority key -> subqueue index
+		priorityBySubqueue  map[subqueueKey]priorityKey // subqueue index -> priority key
```
dnr: I suppose these are superfluous...
```suggestion
		subqueuesByPriority map[priorityKey]subqueueKey
		priorityBySubqueue  map[subqueueKey]priorityKey
```

service/matching/pri_backlog_manager.go
```diff
@@ -167,25 +169,25 @@ func (c *priBacklogManagerImpl) loadSubqueuesLocked(subqueues []persistencespb.S
...
-		c.subqueuesByPriority[subqueues[i].Key.Priority] = i
-		c.priorityBySubqueue[i] = subqueues[i].Key.Priority
+		c.subqueuesByPriority[priorityKey(subqueues[i].Key.Priority)] = subqueueKey(i)
+		c.priorityBySubqueue[subqueueKey(i)] = priorityKey(subqueues[i].Key.Priority)
 	}
 }
 
-func (c *priBacklogManagerImpl) getSubqueueForPriority(priority int32) int {
+func (c *priBacklogManagerImpl) getSubqueueForPriority(priority priorityKey) subqueueKey {
 	levels := c.config.PriorityLevels()
```
dnr, line 182: same here (sorry for so much duplicated code)

service/matching/db.go
```diff
@@ -58,8 +58,10 @@ type (
 		subqueues []persistencespb.SubqueueInfo
 	}
 
+	subqueueKey int
```
dnr: sorry if this is too annoying...

I think this should be
```suggestion
	subqueueIndex int
```
we already have a concept called "subqueue key": the proto message that differentiates the subqueues (currently only has priority key but we can put extra splitting stuff there)

service/matching/db.go
stephanos: no, actually!

service/matching/config.go
stephanos: yes, let's!

service/matching/fair_backlog_manager.go
stephanos: I changed `PriorityLevels` to return `priorityKey` instead directly now.

service/matching/db.go
stephanos: It's a good call 👍

service/matching/db.go
stephanos: The only problem is that `CreateTasksRequest` (which this is ultimately passed into) is outside of matching. So I can't use `subqueueIndex` without exporting it. ~Let me see what that looks like if we do that.~ Ah, ofc, circular dependency.

So to change it either the `SubqueueIndex` needs to move to a 3rd package; or we need an extra for loop+allocation to copy the data into `CreateTasksRequest` at the end.

service/matching/config.go
```diff
@@ -441,6 +441,6 @@ func newTaskQueueConfig(tq *tqid.TaskQueue, config *Config, ns namespace.Name) *
 	}
 }
 
-func defaultPriorityLevel(priorityLevels int32) priorityKey {
+func defaultPriorityLevel(priorityLevels priorityKey) priorityKey {
 	return priorityKey(priorityLevels+1) / 2
```
dnr, line 445:
```suggestion
	return (priorityLevels+1) / 2
```

service/matching/db.go
dnr: nah, it's fine. this is already a big improvement

dnr (approved):

dnr:
I'm just going to merge this now to minimize conflicts

PR #8103, opened 2025-07-28
Workflow Task stamp

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


service/history/tasks/activity_retry_timer.go
```diff
@@ -60,7 +60,3 @@ func (r *ActivityRetryTimerTask) GetType() enumsspb.TaskType {
 func (r *ActivityRetryTimerTask) GetStamp() int32 {
 	return r.Stamp
 }
-
-func (r *ActivityRetryTimerTask) SetStamp(stamp int32) {
```
stephanos, line 65: `SetStamp` functions aren't used anywhere.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5011,27 +5010,19 @@ func (ms *MutableStateImpl) updateVersioningOverride(
...
-			// It is protected by timeout on both normal and sticky task queues.
-			// If there is no poller for previous deployment, it will time out,
-			// and will be rescheduled as normal WFT.
-			ms.GetPendingWorkflowTask().Type != enumsspb.WORKFLOW_TASK_TYPE_SPECULATIVE {
-			// Sticky queue was just cleared, so the following call only generates a WorkflowTask, not a WorkflowTaskTimeoutTask.
-			err := ms.taskGenerator.GenerateScheduleWorkflowTaskTasks(ms.GetPendingWorkflowTask().ScheduledEventID)
-			if err != nil {
-				return err
-			}
-		}
```
stephanos, line 5026: Migrated into `reschedulePendingWorkflowTask`.

service/history/workflow/mutable_state_impl_test.go
```diff
@@ -5117,8 +5121,6 @@ func (s *mutableStateSuite) TestHasRequestID() {
 }
 
 func (s *mutableStateSuite) TestHasRequestID_StateConsistency() {
-	s.SetupTest()
```
stephanos, line 5276: Not needed.

service/history/api/recordworkflowtaskstarted/api.go
```diff
@@ -74,6 +74,10 @@ func Invoke(
 				//  - Speculative WFT is lost (ScheduleToStart timeout for speculative WFT will recreate it).
 				return nil, serviceerror.NewNotFound("Workflow task not found.")
 			}
+			if workflowTask.Stamp != mutableState.GetExecutionInfo().GetWorkflowTaskStamp() {
+				// This happens when the workflow task was rescheduled.
+				return nil, serviceerrors.NewObsoleteMatchingTask("workflow task stamp mismatch")
+			}
```
stephanos, line 80: 1/3 places where we invalidate a workflow task with an outdated stamp now.

service/history/timer_queue_standby_task_executor.go
```diff
@@ -422,6 +422,9 @@ func (t *timerQueueStandbyTaskExecutor) executeWorkflowTaskTimeoutTask(
 		if workflowTask == nil {
 			return nil, nil
 		}
+		if timerTask.Stamp != workflowTask.Stamp {
+			return nil, nil
+		}
```
stephanos, line 427: 3/3 places where we invalidate a workflow task with an outdated stamp now.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5011,27 +5013,19 @@ func (ms *MutableStateImpl) updateVersioningOverride(
...
```
stephanos, line 5081: Migrated into `reschedulePendingWorkflowTask`.

service/history/api/recordworkflowtaskstarted/api.go
stephanos: My thinking here is that when deploying a new version, all in-flight tasks and running workflows will have the default stamp, ie zero, and match. Only when the tasks are re-scheduled would they not match.

service/history/tasks/workflow_task.go
```diff
@@ -17,6 +17,7 @@ type (
 		TaskQueue           string
 		ScheduledEventID    int64
 		Version             int64
+		Stamp               int32
```
stephanos, line 21: Is my thinking right that only the workflow transfer and timeout task need the stamp?

service/history/workflow/mutable_state_impl_test.go
```diff
@@ -1056,7 +1059,6 @@ func (s *mutableStateSuite) TestOverride_RedirectFails() {
 }
 
 func (s *mutableStateSuite) TestOverride_BaseDeploymentUpdatedOnCompletion() {
-	s.T().Skip("TODO (Shahab)")
```
stephanos, line 1209: This test works fine AFAICT.

service/history/api/recordworkflowtaskstarted/api.go
```diff
@@ -74,6 +73,10 @@ func Invoke(
 				//  - Speculative WFT is lost (ScheduleToStart timeout for speculative WFT will recreate it).
 				return nil, serviceerror.NewNotFound("Workflow task not found.")
 			}
+			if workflowTask.Stamp != mutableState.GetExecutionInfo().GetWorkflowTaskStamp() {
+				// This happens when the workflow task was rescheduled.
+				return nil, serviceerrors.NewObsoleteMatchingTask("Workflow task stamp mismatch")
```
stephanos, line 78: Only place outside of task processing internals where we invalidate a workflow task with an outdated stamp now.

My thinking here is that when deploying a new version, all in-flight tasks and running workflows will have the default stamp, ie zero, and match. Only when the tasks are re-scheduled would they not match.

service/history/transfer_queue_standby_task_executor_test.go
```diff
@@ -661,6 +661,58 @@ func (s *transferQueueStandbyTaskExecutorSuite) TestProcessWorkflowTask_Success_
...
+	mutableState.GetExecutionInfo().WorkflowTaskStamp = wt.Stamp + 1
+
+	persistenceMutableState := s.createPersistenceMutableState(mutableState, wt.ScheduledEventID, s.version)
+	s.mockExecutionMgr.EXPECT().GetWorkflowExecution(gomock.Any(), gomock.Any()).Return(&persistence.GetWorkflowExecutionResponse{State: persistenceMutableState}, nil)
+	s.mockShard.SetCurrentTime(s.clusterName, now)
+
+	resp := s.transferQueueStandbyTaskExecutor.Execute(context.Background(), s.newTaskExecutable(transferTask))
+	s.ErrorIs(resp.ExecutionErr, consts.ErrStaleReference)
+}
+
```
stephanos, line 714: ... I don't know how _not to_ duplicate those tests over and over.

service/history/api/recordworkflowtaskstarted/api.go
```diff
@@ -74,6 +73,10 @@ func Invoke(
 				//  - Speculative WFT is lost (ScheduleToStart timeout for speculative WFT will recreate it).
 				return nil, serviceerror.NewNotFound("Workflow task not found.")
 			}
+			if workflowTask.Stamp != mutableState.GetExecutionInfo().GetWorkflowTaskStamp() {
```
gow: I'm a bit confused. `workflowTask.Stamp` seems to be [constructed from current mutable state](https://github.com/stephanos/temporal/blob/213014679a4c8b69d1adf1f6aa9cf5881d583925/service/history/workflow/workflow_task_state_machine.go#L1096). Which means this will never evaluate to true?

service/history/tasks/workflow_task.go
gow: For workflow pause workstream we might need it in `UserTimerTask` & `StartChildExecutionTask` as well. But we are not ready for it yet. So we can add it in the future if needed.

service/history/api/recordworkflowtaskstarted/api.go
stephanos: You're absolutely right! 🤦  I'll need to pass it in from Matching instead. Let me fix that.

stephanos:
@gow Alright; check out the 2nd commit I added. There's also a follow-up PR (will be reviewed by matching crew) where I actually use the stamp and have a functional test for it: https://github.com/temporalio/temporal/pull/8396 (see `tests/priority_fairness_test.go`).

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -234,6 +234,8 @@ message RecordWorkflowTaskStartedRequest {
     temporal.api.deployment.v1.Deployment scheduled_deployment = 9;
     // Versioning directive that was sent by history when scheduling the task.
     temporal.server.api.taskqueue.v1.TaskVersionDirective version_directive = 10;
+    // Stamp value from when the workflow task was scheduled. Used to validate the task is still relevant.
+    int32 stamp = 11;
```
yycptt, line 238: ideally should have attempt wherever there's stamp. But we can save that for later :)

yycptt, line 238: Need this field in IsWorkflowTaskValid request as well, and update the caller & handler logic.

service/history/transfer_queue_standby_task_executor.go
```diff
@@ -201,6 +201,9 @@ func (t *transferQueueStandbyTaskExecutor) processWorkflowTask(
 		if wtInfo == nil {
 			return nil, nil
 		}
+		if transferTask.Stamp != wtInfo.Stamp {
+			return nil, consts.ErrStaleReference
```
yycptt, line 205: nit: we probably also need release(nil) here (or just return nil error). I didn't check the logic that invokes this actionFn.

service/history/workflow/task_generator.go
```diff
@@ -419,6 +417,7 @@ func (r *TaskGeneratorImpl) GenerateScheduleWorkflowTaskTasks(
 			EventID:             workflowTask.ScheduledEventID,
 			ScheduleAttempt:     workflowTask.Attempt,
 			Version:             workflowTask.Version,
+			Stamp:               r.mutableState.GetExecutionInfo().GetWorkflowTaskStamp(), // must use current stamp!
```
yycptt: oh maybe provide some more context here. Guess I am still confused on why workflow.Stamp would be wrong here?

service/history/workflow/mutable_state_impl.go
```diff
@@ -8252,6 +8230,36 @@ func (ms *MutableStateImpl) reschedulePendingActivities() error {
...
+	}
+
+	// If requested, increase the stamp ("version") to invalidate the pending non-speculative WFT.
+	// We don't invalidate speculative WFTs because they are very latency sensitive.
+	if invalidatePendingTasks {
+		ms.executionInfo.WorkflowTaskStamp += 1
+	}
+
+	// Reset the attempt; forcing a non-transient workflow task to be scheduled.
+	ms.executionInfo.Attempt = 1
```
yycptt: we should only do this when Stamp is incremented?

service/history/workflow/mutable_state_impl.go
yycptt, line 1: We need to make sure the stamp field is replicated properly in state-based replication as well. Please update the syncExecutionInfo() method in this file and have the Stamp field updated.

service/history/workflow/mutable_state_impl.go
```diff
@@ -4920,14 +4919,16 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionOptionsUpdatedEvent(event *his
...
+	}
 	return nil
 }
 
 func (ms *MutableStateImpl) updateVersioningOverride(
 	override *workflowpb.VersioningOverride,
-) error {
-	previousEffectiveDeployment := ms.GetEffectiveDeployment()
-	previousEffectiveVersioningBehavior := ms.GetEffectiveVersioningBehavior()
+) (requestReschedulePendingWorkflowTask bool, err error) {
```
gow: nit: Prefer avoiding named return values in large and complex functions. It hurts readability since return statements are naked and hard to know what values are being returned.

service/history/workflow/mutable_state_impl.go
```diff
@@ -3024,18 +3024,10 @@ func (ms *MutableStateImpl) ApplyBuildIdRedirect(
...
-		// does not support them.
-		ms.GetPendingWorkflowTask().Type != enumsspb.WORKFLOW_TASK_TYPE_SPECULATIVE {
-		// sticky queue should be cleared by UpdateBuildIdAssignment already, so the following call only generates
-		// a WorkflowTask, not a WorkflowTaskTimeoutTask.
-		err = ms.taskGenerator.GenerateScheduleWorkflowTaskTasks(ms.GetPendingWorkflowTask().ScheduledEventID)
-		if err != nil {
-			return err
-		}
+	// Re-scheduling pending workflow and activity tasks.
+	err = ms.reschedulePendingWorkflowTask(false)
```
gow, line 3063: This seems like a case where we want to invalidate existing tasks? i.e `ms.reschedulePendingWorkflowTask(true)`?

service/history/workflow/mutable_state_impl.go
stephanos: Hm, I 100% agree with that _except here_ because without it, it just says `bool` and you'll have to scan the whole body to find out what the bool is. Alternatives I can think of are (1) add comment (can get out of date more easily) or (2) adding a `var requestReschedulePendingWorkflowTask bool` in the body right at the top. WDYT?

service/history/workflow/mutable_state_impl.go
stephanos: You're right 👍

service/history/workflow/mutable_state_impl.go
stephanos: Using `var` now, looks fine 👍

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
stephanos: I didn't know that API existed; ~who's the caller~? Ah, I see it in Matching.

Added 👍

service/history/transfer_queue_standby_task_executor.go
stephanos: Added in 5a2328373c707067c01d88a32343ff3b10ceeac0

service/history/workflow/mutable_state_impl.go
stephanos: Oh; that's easy to miss if you don't know where to look! Have we considered some kind of test that verifies a field is either replicated or explicitly ignored?

service/history/workflow/mutable_state_impl.go
stephanos: So the reason I don't is I cut scope from this PR. The previous code didn't invalidate pending workflow tasks, so there's no behavior change. The `recordworkflowtaskstarted` API has a dedicated check for buildId.

Why did I declare it out of scope? Yichao asked me how this works with replication and I couldn't tell you since I don't understand the versioning API well enough. So in order to get this merged, I declared it out of scope.

service/history/workflow/mutable_state_impl.go
stephanos: I'm not sure if I got the test right.

service/history/workflow/task_generator.go
stephanos: Alright; I took another look and when I wrote this I didn't realize that `GetWorkflowTaskByID` is actually just returning a struct with the current workflow task's info. I thought it returned the _currently_ scheduled workflow task (I didn't know we don't store that). So I've removed the special case here now!

service/history/workflow/mutable_state_impl.go
yycptt: yeah, I think CGS team is working on some general testing strategy to assert the states are the same on both sides.

yycptt (approved):

service/history/workflow/mutable_state_impl.go
yycptt: Those versioning rescheduling logic is not working with state-based replication yet. That's something replication crew is aware of and will be fixing them soon. cc @yux0 @xwduan

PR #8104, opened 2025-07-28
fix: retryable error when matching service loses ownership

spkane31:
## What changed?
The error type returned by `ListNexusEndpoints` is changed from `FailedPrecondition` -> `Aborted`. 

## Why?
Clients will retry an `Aborted` error type

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
None aware of


common/nexus/endpoint_registry_test.go
```diff
@@ -380,6 +380,80 @@ func TestTableVersionErrorResetsPersistencePagination(t *testing.T) {
 	assert.Equal(t, int64(3), reg.tableVersion)
 }
 
+func TestRetryOnAbortedError(t *testing.T) {
```
bergundy: This is testing the client behavior, the code you changed is in the server.

service/matching/matching_engine.go
bergundy, line 1: Wondering if we want to return Aborted errors here and where we give up retrying failed precondition.
We should never give up retrying, so maybe that should be the fix here.

pdoerner: We convert `FailedPrecondition` to `BadRequest`: https://github.com/temporalio/temporal/blob/47865482ba1ea09db6bc6af201473e33dc20f1c4/common/nexus/failure.go#L208

And the default retry behavior for `BadRequest` is non-retryable: https://github.com/nexus-rpc/sdk-go/blob/ce59b2c82f021201b438cffa1fe52b2e5507039f/nexus/api.go#L257

I think Aborted or Unavailable (both default to `HandlerErrorTypeUnavailable` and are retryable) makes more sense than FailedPrecondition.

common/nexus/endpoint_registry_test.go
pdoerner: I thought we relied on client retries to handle `ShardOwnershipLost` in most cases. Could also add a test to confirm tasks that fail due to these errors are also retried for complete coverage.

spkane31: https://github.com/temporalio/temporal/blob/main/service/history/queues/executable.go#L505-L575

Note for self: this is where the service does the retry portion.

service/matching/matching_engine.go
bergundy: Thanks for the reminder. Let me consider what the right status code would be here.

common/nexus/endpoint_registry_test.go
bergundy: > Note for self: this is where the service does the retry portion.

That's not true, it's here:

https://github.com/temporalio/temporal/blob/ada5d0bbee45150888b63a132be8239fd5a87212/common/util.go#L302

See also the call site here:

https://github.com/temporalio/temporal/blob/ada5d0bbee45150888b63a132be8239fd5a87212/common/nexus/endpoint_registry.go#L281

bergundy: No need to test the client behavior in this PR.

service/history/queues/executable.go
```diff
@@ -476,6 +476,16 @@ func (e *executableImpl) isExpectedRetryableError(err error) (isRetryable bool,
 		return true, consts.ErrNamespaceHandover
 	}
 
+	// Check for serviceerror.Aborted (shard ownership lost, etc.)
```
bergundy: Changing the retry behavior for all task processing without understanding the implications is dangerous.

bergundy (commented):
Please revert changes to the executable logic and delete the client retry tests.

service/history/queues/executable.go
spkane31: removed

bergundy (approved):
I wish we had a reproduction of the bug this is supposed to fix and a test showing that it is fixed.

I still don't understand if we really had a problem, I see we either have a built-in backoff loop or a fallback to persistence in the call sites:

https://github.com/temporalio/temporal/blob/ada5d0bbee45150888b63a132be8239fd5a87212/common/nexus/endpoint_registry.go#L236
https://github.com/temporalio/temporal/blob/ada5d0bbee45150888b63a132be8239fd5a87212/common/nexus/endpoint_registry.go#L356

pdoerner:
We have the retries and fallback in the `endpoint_registry`, yes but that doesn't explain why we had that one Nightly failure where the Nexus operation completed as failed with the table ownership lost message. Even if the start operation call got that error, the task should have been retried since we marked it as retryable. But it's possible that flag wasn't considered properly since it was a FailedPrecondition / BadRequest which is non-retryable by default.

PR #8105, opened 2025-07-28
Nexus gRPC APIs

pdoerner:
## What changed?
Added implementations for workflow service APIs to support Nexus calls over gRPC. Except for `CompleteNexusOperation` which will be coming in a followup.

Depends on https://github.com/temporalio/api/pull/605

## Why?
Required to support non-workflow callers using a Temporal client to interact with Nexus operations.

## How did you test it?
Tests will come in a followup.

### DIFF
--- a/common/api/metadata.go
+++ b/common/api/metadata.go
@@ -161,6 +161,10 @@ var (
 		"UpdateTaskQueueConfig":                 {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
 		"FetchWorkerConfig":                     {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
 		"UpdateWorkerConfig":                    {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
+		"StartNexusOperation":                   {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
+		"RequestCancelNexusOperation":           {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
+		"GetNexusOperationInfo":                 {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
+		"GetNexusOperationResult":               {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
 	}
 	operatorServiceMetadata = map[string]MethodMetadata{
 		"AddSearchAttributes":      {Scope: ScopeNamespace, Access: AccessAdmin, Polling: PollingNone},
--- a/common/rpc/interceptor/redirection.go
+++ b/common/rpc/interceptor/redirection.go
@@ -134,6 +134,11 @@ var (
 		"UpdateTaskQueueConfig": func() any { return &workflowservice.UpdateTaskQueueConfigResponse{} },
 		"FetchWorkerConfig":     func() any { return &workflowservice.FetchWorkerConfigResponse{} },
 		"UpdateWorkerConfig":    func() any { return &workflowservice.UpdateWorkerConfigResponse{} },
+
+		"StartNexusOperation":         func() any { return &workflowservice.StartNexusOperationResponse{} },
+		"RequestCancelNexusOperation": func() any { return &workflowservice.RequestCancelNexusOperationResponse{} },
+		"GetNexusOperationInfo":       func() any { return &workflowservice.GetNexusOperationInfoResponse{} },
+		"GetNexusOperationResult":     func() any { return &workflowservice.GetNexusOperationResultResponse{} },
 	}
 )
 
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -12,12 +12,14 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/nexus-rpc/sdk-go/nexus"
 	"github.com/pborman/uuid"
 	batchpb "go.temporal.io/api/batch/v1"
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	filterpb "go.temporal.io/api/filter/v1"
 	historypb "go.temporal.io/api/history/v1"
+	nexuspb "go.temporal.io/api/nexus/v1"
 	querypb "go.temporal.io/api/query/v1"
 	schedulepb "go.temporal.io/api/schedule/v1"
 	"go.temporal.io/api/serviceerror"
@@ -29,8 +31,10 @@ import (
 	deploymentspb "go.temporal.io/server/api/deployment/v1"
 	"go.temporal.io/server/api/historyservice/v1"
 	"go.temporal.io/server/api/matchingservice/v1"
+	persistencepb "go.temporal.io/server/api/persistence/v1"
 	schedulespb "go.temporal.io/server/api/schedule/v1"
 	taskqueuespb "go.temporal.io/server/api/taskqueue/v1"
+	"go.temporal.io/server/client"
 	"go.temporal.io/server/client/frontend"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/archiver"
@@ -49,6 +53,7 @@ import (
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/namespace/nsreplication"
+	commonnexus "go.temporal.io/server/common/nexus"
 	"go.temporal.io/server/common/payload"
 	"go.temporal.io/server/common/payloads"
 	"go.temporal.io/server/common/persistence"
@@ -91,6 +96,17 @@ var (
 	// Tail room for context deadline to bail out from retry for long poll.
 	longPollTailRoom  = time.Second
 	errWaitForRefresh = serviceerror.NewDeadlineExceeded("waiting for schedule to refresh status of completed workflows")
+
+	errNexusUpstreamTimeout = &nexuspb.HandlerError{
+		ErrorType:     string(nexus.HandlerErrorTypeUpstreamTimeout),
+		RetryBehavior: enumspb.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_RETRYABLE,
+		Failure:       &nexuspb.Failure{Message: "upstream timeout"},
+	}
+	errNexusUnexpectedOutcome = &nexuspb.HandlerError{
+		ErrorType:     string(nexus.HandlerErrorTypeInternal),
+		RetryBehavior: enumspb.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_RETRYABLE,
+		Failure:       &nexuspb.Failure{Message: "unexpected outcome"},
+	}
 )
 
 const (
@@ -130,6 +146,8 @@ type (
 		saMapperProvider                searchattribute.MapperProvider
 		saProvider                      searchattribute.Provider
 		saValidator                     *searchattribute.Validator
+		endpointRegistry                commonnexus.EndpointRegistry
+		clientBean                      client.Bean
 		archivalMetadata                archiver.ArchivalMetadata
 		healthServer                    *health.Server
 		overrides                       *Overrides
@@ -167,6 +185,8 @@ func NewWorkflowHandler(
 	membershipMonitor membership.Monitor,
 	healthInterceptor *interceptor.HealthInterceptor,
 	scheduleSpecBuilder *scheduler.SpecBuilder,
+	endpointRegistry commonnexus.EndpointRegistry,
+	clientBean client.Bean,
 	httpEnabled bool,
 ) *WorkflowHandler {
 	handler := &WorkflowHandler{
@@ -220,6 +240,8 @@ func NewWorkflowHandler(
 		healthInterceptor:   healthInterceptor,
 		scheduleSpecBuilder: scheduleSpecBuilder,
 		outstandingPollers:  collection.NewSyncMap[string, collection.SyncMap[string, context.CancelFunc]](),
+		endpointRegistry:    endpointRegistry,
+		clientBean:          clientBean,
 		httpEnabled:         httpEnabled,
 	}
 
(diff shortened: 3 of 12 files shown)


common/api/metadata.go
```diff
@@ -161,6 +161,10 @@ var (
 		"UpdateTaskQueueConfig":                 {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
 		"FetchWorkerConfig":                     {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
 		"UpdateWorkerConfig":                    {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
+		"StartNexusOperation":                   {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
+		"RequestCancelNexusOperation":           {Scope: ScopeNamespace, Access: AccessWrite, Polling: PollingNone},
+		"GetNexusOperationInfo":                 {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
+		"GetNexusOperationResult":               {Scope: ScopeNamespace, Access: AccessReadOnly, Polling: PollingNone},
```
bergundy: Is this PollingCapable since it can result in a long poll? Not sure if we need to mark it as such or not.

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,413 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+
+		switch v := entry.Endpoint.Spec.GetTarget().GetVariant().(type) {
+		case *persistencepb.NexusEndpointTarget_Worker_:
+			targetNS, err := wh.namespaceRegistry.GetNamespaceByID(namespace.ID(v.Worker.NamespaceId))
+			if err != nil {
+				return nil, nil, err
+			}
+			return targetNS, &taskqueuepb.TaskQueue{Name: v.Worker.GetTaskQueue(), Kind: enumspb.TASK_QUEUE_KIND_NORMAL}, nil
+		default:
+			return nil, nil, serviceerror.NewInvalidArgument("invalid endpoint target")
```
bergundy:
```suggestion
			return nil, nil, serviceerror.NewInvalidArgument("endpoint target must be a worker")
```

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,413 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+
+	targetNS, targetTQ, err := wh.resolveNexusDispatchTarget(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	if targetNS.Name().String() != request.Namespace && !targetNS.ActiveInCluster(wh.clusterMetadata.GetCurrentClusterName()) {
+		// The request dispatch target indicated a namespace which is not active on this cluster. Forward the request.
+		forwardRequest := proto.CloneOf(request)
+		forwardRequest.Namespace = targetNS.Name().String()
```
bergundy: We shouldn't modify the original request, it changes the caller semantics and may confuse authorizers.
We should forward the request as-is while adding some redirect specific headers and handling the fact that the source namespace for Nexus requests may be on a different cluster in the interceptor layer.

Actually, just move anything redirect related to the redirection interceptor.

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,413 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+		}
+		_, remoteClient, err := wh.clientBean.GetRemoteFrontendClient(targetNS.ActiveClusterName())
+		if err != nil {
+			return nil, err
+		}
+		return remoteClient.GetNexusOperationResult(ctx, forwardRequest)
+	}
+
+	// If set, truncate request timeout to min(time.until(ctx.Deadline), request.Wait)
+	timeout := request.GetWait()
```
bergundy: Wait should be completely separate from context deadline. Delete this adjustment of the timeout.

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,413 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+				},
+			}, nil
+		}
+	}
+
+	return &workflowservice.GetNexusOperationResultResponse{
+		Variant: &workflowservice.GetNexusOperationResultResponse_HandlerError{
+			HandlerError: errNexusUnexpectedOutcome,
+		},
+	}, nil
```
bergundy, line 5764: This should just be an internal service error.

service/frontend/workflow_handler.go
```diff
@@ -91,6 +96,17 @@ var (
 	// Tail room for context deadline to bail out from retry for long poll.
 	longPollTailRoom  = time.Second
 	errWaitForRefresh = serviceerror.NewDeadlineExceeded("waiting for schedule to refresh status of completed workflows")
+
+	errNexusUpstreamTimeout = &nexuspb.HandlerError{
+		ErrorType:     string(nexus.HandlerErrorTypeUpstreamTimeout),
+		RetryBehavior: enumspb.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_RETRYABLE,
+		Failure:       &nexuspb.Failure{Message: "upstream timeout"},
+	}
+	errNexusUnexpectedOutcome = &nexuspb.HandlerError{
```
bergundy, line 105: We said we will never return handler errors from the server code to ensure we can track metrics appropriately (upstream timeout seems to be the only exception here).

bergundy (commented):
My comments on GetNexusOperationResult apply to all of the handler methods.

service/frontend/workflow_handler.go
pdoerner: I considered this and actually wrote most of the code but this type of redirection is actually a bit different. For example, if we used that logic to forward the request to the endpoint target namespace, but that namespace does a failover, then we would not forward the request again because of the headers we add to prevent a redirect loop.

It would also mean adjusting the auto-forwarding policy logic and the meaning of the dynamic config which enables auto-forwarding for a namespace. Since from the caller perspective they should not care that their request was forwarded to another cluster.

On top of that, it will make metrics tracking very tricky; we will need custom logic in our namespace validator, caller info, rate limiting, telemetry, etc interceptors to handle these request types and extract the actual target namespace. This approach also has the small optimization of only having to look up the endpoint once, instead of in several interceptors.

It's a good point about the authorizers. I don't know much about our authorizer logic but I thought it would be better actually that the request is authenticated to the target namespace, not just the caller.

service/frontend/workflow_handler.go
pdoerner: How is this not a handler error if the handler returned an unreadable response?

service/frontend/workflow_handler.go
pdoerner: We said that we would return handler errors when we put the responsibility of the error on the worker. Basically in any case where we set the failure-source header to `worker` for the HTTP handler. That would apply in this case. Otherwise we will actually be over counting these errors against our SLAs.

service/frontend/workflow_handler.go
pdoerner: This was discussed on the API PR and we decided we would send the handler the min between the context deadline and the wait duration. Is there some reason this is not the case any longer?

https://github.com/temporalio/api/pull/605#discussion_r2164480760

service/frontend/workflow_handler.go
bergundy: Hmm... I see, you're right.
We should also indicate to the handler that it has responded with an invalid outcome when the worker responds to the frontend but ensure that we unblock the client request.

I'm just not sure about the error string, "unexpected outcome" may be confusing to callers, is this what we do for HTTP? Can we change it to "invalid upstream response"?

service/frontend/workflow_handler.go
bergundy: > but that namespace does a failover, then we would not forward the request again because of the headers we add to prevent a redirect loop.

But here you're already forwarding since you're not in the active cluster, so we should be consistent with other request forwarding logic.

> It would also mean adjusting the auto-forwarding policy logic and the meaning of the dynamic config which enables auto-forwarding for a namespace.

Can you elaborate on that, I'm a bit fuzzy on the details, but it kinda sounds like that's what we want since forwarding should be configurable for all APIs.

> On top of that, it will make metrics tracking very tricky

Isn't that already the case though? We're doing cross namespace calls, and if we want to emit metrics correctly, apply rate limits, etc on the handler side we need to be endpoint aware regardless of whether a request is forwarded.

So basically we need to understand the implication of handling requests by endpoint in WorkflowService from the perspective of interceptors, this may be more involved than what you have here in this PR.

service/frontend/workflow_handler.go
bergundy: I must have missed that comment because that's not what we do in the Nexus SDK. We treat request timeout (which in gRPC would be a context deadline) separately from wait time.
See this code for reference (which BTW has a bug that we should address): https://github.com/nexus-rpc/sdk-go/blob/ce59b2c82f021201b438cffa1fe52b2e5507039f/nexus/server.go#L400

service/frontend/workflow_handler.go
bergundy: Continuing this thread above where `errNexusUnexpectedOutcome` is defined.

service/frontend/workflow_handler.go
```diff
@@ -91,6 +96,17 @@ var (
 	// Tail room for context deadline to bail out from retry for long poll.
 	longPollTailRoom  = time.Second
 	errWaitForRefresh = serviceerror.NewDeadlineExceeded("waiting for schedule to refresh status of completed workflows")
+
+	errNexusUpstreamTimeout = &nexuspb.HandlerError{
+		ErrorType:     string(nexus.HandlerErrorTypeUpstreamTimeout),
+		RetryBehavior: enumspb.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_RETRYABLE,
```
bergundy, line 102: Retryable is already the default, no need to specify it explicitly.

service/frontend/workflow_handler.go
```diff
@@ -91,6 +96,17 @@ var (
...
+		Failure:       &nexuspb.Failure{Message: "upstream timeout"},
+	}
+	errNexusUnexpectedOutcome = &nexuspb.HandlerError{
+		ErrorType:     string(nexus.HandlerErrorTypeInternal),
+		RetryBehavior: enumspb.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_RETRYABLE,
```
bergundy, line 107: This is already the default, not need to specify it.

bergundy (commented):
Looks like we need to understand the implication on interceptors for these requests that target an endpoint instead of a namespace before this can be merged.

service/frontend/workflow_handler.go
pdoerner: So in the most simple case using an endpoint target: the caller and handler namespaces are both active on the cluster which received the request, should we consider that request from the perspective of the caller or the handler? e.g. which frontend rate limit quota do we consume? I was under the impression it was the caller's. Or another example: should that request need to be authorized for the caller namespace or the target or both? Again, I was thinking only the caller (assuming the caller is in the allowed callers list for that endpoint).

About forwarding, the redirection interceptor, auto-forwarding policy, and namespace not active auto-forwarding dynamic config are all working under the assumption that the only redirection that happens is when a request lands on a standby cluster for a namespace. For endpoint targets it is slightly different: the target namespace may not be on the cluster at all. For example, what would be the expected behavior for a namespace that has disabled namespace-not-active auto-forwarding entirely? I think they would still want their completion request sent to the target namespace.

If we would rather only consider the handler namespace for the purposes of authorization, metrics, limits, redirection, etc it actually becomes much easier. I can just add a new interceptor to resolve the endpoint target and put that in the namespace field and put the original caller in some context metadata. Then all our remaining interceptors would work as expected.

service/frontend/workflow_handler.go
pdoerner: In the HTTP handler we just say `empty outcome` but I am fine with `invalid upstream response`, that does make it clearer.

service/frontend/workflow_handler.go
pdoerner: I know it's not strictly necessary but it just makes it a little easier to understand at a glance for anyone tracing this error in the code.

service/frontend/workflow_handler.go
pdoerner: Synced offline. All requests with an endpoint target will go through the HTTP handler.

common/nexus/failure.go
```diff
@@ -183,6 +184,43 @@ func nexusFailureMetadataToPayloads(failure nexus.Failure) (*commonpb.Payloads,
 	}, err
 }
 
+func NexusHandlerErrorToProtoHandlerError(handlerErr *nexus.HandlerError) *nexuspb.HandlerError {
```
bergundy, line 187: This will change soon but for now keep it as is.

common/rpc/interceptor/redirection.go
```diff
@@ -305,3 +316,17 @@ func (i *Redirection) RedirectionAllowed(
...
+		GetTarget() *nexuspb.TaskDispatchTarget
+	})
+	if !ok {
+		return false
+	}
+	switch t.GetTarget().Variant.(type) {
+	case *nexuspb.TaskDispatchTarget_Endpoint:
+		return true
+	}
+	return false
```
bergundy: nit:

```suggestion
	_, ok = t.GetTarget().Variant.(*nexuspb.TaskDispatchTarget_Endpoint)
	return ok
```

service/frontend/workflow_handler.go
bergundy: Lets change in both places.

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,569 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+		return wh.forwardStartNexusOperation(ctx, ns, t.Endpoint, request)
+	case *nexuspb.TaskDispatchTarget_TaskQueue:
+		// dispatch directly
+		return wh.dispatchStartNexusOperation(ctx, ns, t.TaskQueue, request)
+	default:
+		return nil, errNexusUnsupportedDispatch
+	}
+}
+
+func (wh *WorkflowHandler) forwardStartNexusOperation(ctx context.Context, ns *namespace.Namespace, endpoint string, request *workflowservice.StartNexusOperationRequest) (*workflowservice.StartNexusOperationResponse, error) {
```
bergundy:
```suggestion
func (wh *WorkflowHandler) dispatchStartNexusOperationByEndpoint(ctx context.Context, ns *namespace.Namespace, endpoint string, request *workflowservice.StartNexusOperationRequest) (*workflowservice.StartNexusOperationResponse, error) {
```

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,569 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+		Links: links,
+		Variant: &workflowservice.StartNexusOperationResponse_SyncSuccess{
+			SyncSuccess: &workflowservice.StartNexusOperationResponse_Sync{
+				Result: res,
+			},
+		},
+	}, nil
+}
+
+func (wh *WorkflowHandler) dispatchStartNexusOperation(ctx context.Context, ns *namespace.Namespace, taskQueue string, request *workflowservice.StartNexusOperationRequest) (*workflowservice.StartNexusOperationResponse, error) {
```
bergundy:
```suggestion
func (wh *WorkflowHandler) dispatchStartNexusOperationByTaskQueue(ctx context.Context, ns *namespace.Namespace, taskQueue string, request *workflowservice.StartNexusOperationRequest) (*workflowservice.StartNexusOperationResponse, error) {
```

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,569 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+	sizeLimitError := wh.config.BlobSizeLimitError(ns.Name().String())
+	sizeLimitWarn := wh.config.BlobSizeLimitWarn(ns.Name().String())
+	if err := common.CheckEventBlobSizeLimit(
```
bergundy: This should apply to both the HTTP and matching call. No need to forward a request that we know we can't accept.

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,569 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+		return wh.forwardRequestCancelNexusOperation(ctx, ns, t.Endpoint, request)
+	case *nexuspb.TaskDispatchTarget_TaskQueue:
+		// dispatch directly
+		return wh.dispatchRequestCancelNexusOperation(ctx, ns, t.TaskQueue, request)
+	default:
+		return nil, errNexusUnsupportedDispatch
+	}
+}
+
+func (wh *WorkflowHandler) forwardRequestCancelNexusOperation(ctx context.Context, ns *namespace.Namespace, endpoint string, request *workflowservice.RequestCancelNexusOperationRequest) (*workflowservice.RequestCancelNexusOperationResponse, error) {
```
bergundy: Rename this to please.

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,569 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+		return nil, errRequestNotSet
+	}
+	ns, err := wh.namespaceRegistry.GetNamespace(namespace.Name(request.Namespace))
+	if err != nil {
+		return nil, err
+	}
+
+	switch t := request.GetTarget().GetVariant().(type) {
+	case *nexuspb.TaskDispatchTarget_Endpoint:
+		// forward via HTTP
```
bergundy: Not necessarily a forward.

```suggestion
		// dispatch via HTTP
```

service/frontend/workflow_handler.go
```diff
@@ -5180,6 +5202,569 @@ func (wh *WorkflowHandler) RespondNexusTaskFailed(ctx context.Context, request *
...
+		return wh.forwardRequestCancelNexusOperation(ctx, ns, t.Endpoint, request)
+	case *nexuspb.TaskDispatchTarget_TaskQueue:
+		// dispatch directly
```
bergundy:
```suggestion
		// dispatch directly to matching
```

bergundy (approved):
Nothing blocking.
We really need test coverage though before merging to `main`.

PR #8108, opened 2025-07-29
[Pause] - Add WorkflowExecutionInfo.pause_info proto

gow:
## What changed?
Adding proto changes to capture pause information. This contains 2 main objects:
1. WorkflowPauseInfo: A container to capture pause information in a given workflow. Adding this for future compatibility. The intention is to use this container for other workflow level pause information like timeouts, signals, child workflows etc.
2. ActivityPauseInfo: A container to capture pause information for a given activity type. Since individual activity types can be paused, we need to capture the pause state, identity & reason for better debugging.

Note: Sending just the proto changes so that we can work on implementation in parallel.

## Why?
To implement pause activity and eventually pause workflow feature. Need this space to save the state information.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
N/A

### DIFF
--- a/proto/internal/temporal/server/api/persistence/v1/executions.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/executions.proto
@@ -835,7 +835,7 @@ message ActivityPauseInfo {
     // The identity of the actor that paused the activity.
     string identity = 3;
 
-    // The id of the request that paused the activity.
+    // The ID of the request that paused the activity. Duplicates pause requests with the same ID will succeed.
     string request_id = 4;
 
     // The reason for the pause.
--- a/proto/internal/temporal/server/api/persistence/v1/executions.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/executions.proto
@@ -273,6 +273,9 @@ message WorkflowExecutionInfo {
 
     // Run ID of the execution that supersedes this one (via terminate or continue-as-new).
     string successor_run_id = 105;
+
+    // Pause info contains the details of the request to pause the workflow.
+    WorkflowPauseInfo pause_info = 106;
 }
 
 message ExecutionStats {
@@ -815,4 +818,19 @@ message NexusOperationCancellationInfo {
 message ResetChildInfo {
     // If true, the parent workflow should terminate the child before starting it.
     bool should_terminate_and_start = 1;
+}
+
+message WorkflowPauseInfo {
+    // The time when the workflow was paused.
+    google.protobuf.Timestamp paused_time = 1;
+
+    // The identity of the actor that initiated the pause.
+    string identity = 2;
+
+    // If true, all activities are paused.
+    bool pause_all_activities = 3;
+
+    // Map of activity types that are to be paused.
+    // The key is the activity type and the value is true if the activity is to be paused.
+    map<string, bool> activity_types = 4;
 }
--- a/proto/internal/temporal/server/api/persistence/v1/executions.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/executions.proto
@@ -821,30 +821,24 @@ message ResetChildInfo {
 }
 
 message WorkflowPauseInfo {
-    // The time when the workflow was paused.
-    google.protobuf.Timestamp update_time = 1;
-
-    // The identity of the actor that updated the pause info.
-    string identity = 2;
-
     // Map of activity types that are to be paused.
     // The key is the activity type and the value is the activity pause info.
-    map<string, ActivityPauseInfo> activity_pause_infos = 3;
-
-    // If true, all activities are paused.
-    bool pause_all_activities = 4;
+    repeated ActivityPauseInfo activity_pause_infos = 2;
 }
 
 message ActivityPauseInfo {
     // The time when the activity pause info was updated.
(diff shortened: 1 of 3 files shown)


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -811,3 +814,32 @@ message ResetChildInfo {
...
+
+    // The identity of the actor that updated the pause info.
+    string identity = 2;
+
+    // Map of activity types that are to be paused.
+    // The key is the activity type and the value is the activity pause info.
+    map<string, ActivityPauseInfo> activity_pause_infos = 3;
+
+    // If true, all activities are paused.
+    bool pause_all_activities = 4;
```
bergundy: Based on the recent slack discussions, we decided not to add this feature for now.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -811,3 +814,32 @@ message ResetChildInfo {
...
+}
+
+message ActivityPauseInfo {
```
bergundy, line 828: This is different than the proto we discussed in the blueprint. I thought we were aligned there.

bergundy: Consider also adding request_id so we can make retries idempotent to the caller (return success).

bergundy: Also consider avoiding duplicates reasons and identities when a request pauses multiple activity types. Not a blocker IMHO but it might be nicer to visualize this as separate pause requests instead of this map.

```
message ActivityPauseRequest {
    // The time when the activity pause info was updated.
    google.protobuf.Timestamp update_time = 1;

    // If true, the activity is paused.
    bool is_paused = 2;

    // The identity of the actor that updated the activity pause info.
    string identity = 3;

    // The reason for the pause.
    string reason = 4;

    // Idempotency key.
    string request_id = 5;

    repeated string activity_types = 6;
}
```

And do a repeated of these instead. Then the code would just compute the set of paused activities from this struct when needed.

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: I had missed that update. Will remove this for now.

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: It differs by 2 things.
- It's a set instead of list since the access pattern is direct lookups (in both pause & unpause APIs).
- The `type` field is omitted from the info since it's used as a key in the set.

This provides an added benefit of serving as idempotency key. i.e an already paused activity will return success on all subsequent pause requests (even if request_id is different). So, I don't see how request_id might be helpful in this specific case.

Agree on the changes to the `ActivityPauseRequest`. Will include it when I make those changes in the api repo.

gow: Addressed the feedback.
1. Added `activity_type` & `request_id` in activity pause info object.
2. Embedding this object as a list in WorkflowPauseInfo (instead of a map)
3. Removed all other unused fields.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -816,3 +819,26 @@ message ResetChildInfo {
     // If true, the parent workflow should terminate the child before starting it.
     bool should_terminate_and_start = 1;
 }
+
+message WorkflowPauseInfo {
+    // Map of activity types that are to be paused.
+    // The key is the activity type and the value is the activity pause info.
+    repeated ActivityPauseInfo activity_pause_infos = 2;
```
spkane31:
```suggestion
    repeated ActivityPauseInfo activity_pause_infos = 1;
```

spkane31: If this is meant to be a list (per comment below), comment needs fixing

gow: oh! the comment was not updated here. We decided to keep this as a list. Will update the comment.

gow: Updated

proto/internal/temporal/server/api/persistence/v1/executions.proto
gow: Removed.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -816,3 +819,25 @@ message ResetChildInfo {
...
+    // The time when the activity pause info was updated.
+    google.protobuf.Timestamp update_time = 1;
+
+    // The type of the activity.
+    string activity_type = 2;
+
+    // The identity of the actor that paused the activity.
+    string identity = 3;
+
+    // The id of the request that paused the activity.
```
bergundy:
```suggestion
    // The ID of the request that paused the activity. Duplicates pause requests with the same ID will succeed.
    // Otherwise, pausing the same activity will result in a failed precondition error.
```

PR #8111, opened 2025-07-29
Unify DataBlob encoding/decoding

stephanos:
## What changed?

Cleanup of `common/serialization` package:

- unify several DataBlob encode/decode functions into 3 functions (see new codec.go)
- unify function signatures to use `DataBlob` everywhere (instead of split []byte + EncodingType)

## Why?

- less code
- simpler
- consistent
- preparation step to allow changing proto to json encoding via env variable for debugging purposes
  
## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential Risks

See the two callouts for behavior difference.


common/persistence/serialization/codec.go
```diff
@@ -0,0 +1,71 @@
...
+
+		return &commonpb.DataBlob{EncodingType: enumspb.ENCODING_TYPE_PROTO3, Data: data}, nil
+	default:
+		return nil, NewUnknownEncodingTypeError(encoding.String(), enumspb.ENCODING_TYPE_JSON, enumspb.ENCODING_TYPE_PROTO3)
+	}
+}
+
+func ProtoDecode(data *commonpb.DataBlob, result proto.Message) error {
+	if data == nil {
+		return NewDeserializationError(enumspb.ENCODING_TYPE_UNSPECIFIED, errors.New("cannot decode nil"))
```
stephanos, line 50: Note that one helper returned an error (like here), the other one just returned `nil`. I decided to make it as explicit as possible and keep the error response.

common/persistence/serialization/codec.go
```diff
@@ -0,0 +1,69 @@
...
+	"go.temporal.io/server/common/codec"
+	"google.golang.org/protobuf/proto"
+)
+
+func ProtoEncode(m proto.Message) (*commonpb.DataBlob, error) {
+	return encodeBlob(m, enumspb.ENCODING_TYPE_PROTO3)
+}
+
+func encodeBlob(m proto.Message, encoding enumspb.EncodingType) (*commonpb.DataBlob, error) {
+	if m == nil || (reflect.ValueOf(m).Kind() == reflect.Ptr && reflect.ValueOf(m).IsNil()) {
```
stephanos, line 18: One of the helper sets only tested for `nil`, the other one also tested via reflection. To be safe, I keep the more thorough check.

common/persistence/serialization/codec.go
```diff
@@ -0,0 +1,68 @@
...
+	"go.temporal.io/server/common/codec"
+	"google.golang.org/protobuf/proto"
+)
+
+func ProtoEncode(m proto.Message) (*commonpb.DataBlob, error) {
+	return encodeBlob(m, enumspb.ENCODING_TYPE_PROTO3)
+}
+
+func encodeBlob(m proto.Message, encoding enumspb.EncodingType) (*commonpb.DataBlob, error) {
+	if m == nil {
```
stephanos, line 17: One of the helpers only tested for nil, the other one also tested via reflection. I tested it out and the reflection check made no difference (ie the marshaller returned `nil`). So it's not needed.

alexshtin (approved):

common/persistence/serialization/codec.go
```diff
@@ -0,0 +1,68 @@
...
+		return &commonpb.DataBlob{
+			EncodingType: enumspb.ENCODING_TYPE_PROTO3,
+			Data:         data,
+		}, nil
+	default:
+		return nil, NewUnknownEncodingTypeError(encoding.String(), enumspb.ENCODING_TYPE_JSON, enumspb.ENCODING_TYPE_PROTO3)
+	}
+}
+
+func ProtoDecode(data *commonpb.DataBlob, result proto.Message) error {
```
simvlad: nit: the naming suggests that this decodes only ENCODING_TYPE_PROTO3. Maybe rename to decodeBlob and have ProtoDecode ensure that the encoding is  ENCODING_TYPE_PROTO3. Would need to fix that one place where ProtoDecode is called on JSON though.

stephanos: Good point! I might just call it simply `Decode` for now. FWIW, I'll replace this with a swappable impl soon.

stephanos: Actually, on second thought, it is decoding a _proto message_ ... 🤔 but I can see how it's potentially confusing. Anyway, I'll keep it as `Decode` now.

PR #8116, opened 2025-07-29
CHASM: remove workflow specific logic in migration activities

yycptt:
## What changed?
- Remove workflow specific logic in migration activities
- Add `SkipForceReload` flag to admin describeMutableState api to avoid unnecessary mutable state loading.

## Why?
- Migration workflow needs to work with chasm runs as well, not just workflows.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/service/history/ndc_standby_task_util.go
+++ b/service/history/ndc_standby_task_util.go
@@ -7,7 +7,7 @@ import (
 
 	commonpb "go.temporal.io/api/common/v1"
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
-	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/server/api/adminservice/v1"
 	persistencespb "go.temporal.io/server/api/persistence/v1"
 	taskqueuespb "go.temporal.io/server/api/taskqueue/v1"
 	"go.temporal.io/server/client"
@@ -98,16 +98,17 @@ func isWorkflowExistOnSource(
 	if err != nil {
 		return true
 	}
-	_, remoteFrontend, err := clientBean.GetRemoteFrontendClient(remoteClusterName)
+	remoteAdminClient, err := clientBean.GetRemoteAdminClient(remoteClusterName)
 	if err != nil {
 		return true
 	}
-	_, err = remoteFrontend.DescribeWorkflowExecution(ctx, &workflowservice.DescribeWorkflowExecutionRequest{
+	_, err = remoteAdminClient.DescribeMutableState(ctx, &adminservice.DescribeMutableStateRequest{
 		Namespace: namespaceEntry.Name().String(),
 		Execution: &commonpb.WorkflowExecution{
 			WorkflowId: workflowKey.GetWorkflowID(),
 			RunId:      workflowKey.GetRunID(),
 		},
+		SkipForceReload: true,
 	})
 	if err != nil {
 		if common.IsNotFoundError(err) {
--- a/service/worker/migration/activities.go
+++ b/service/worker/migration/activities.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math"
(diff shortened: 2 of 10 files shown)


service/worker/migration/activities.go
```diff
@@ -538,66 +541,6 @@ func (a *activities) GenerateReplicationTasks(ctx context.Context, request *gene
 	return nil
 }
 
-func (a *activities) generateExecutionsToReplicate(
```
yycptt, line 541: Not used.

service/worker/migration/activities.go
```diff
@@ -718,12 +642,13 @@ func (a *activities) checkSkipWorkflowExecution(
 	namespaceID := request.NamespaceID
 	tags := []tag.Tag{tag.WorkflowNamespaceID(namespaceID), tag.WorkflowID(we.WorkflowId), tag.WorkflowRunID(we.RunId)}
 	resp, err := a.historyClient.DescribeMutableState(ctx, &historyservice.DescribeMutableStateRequest{
-		NamespaceId: namespaceID,
-		Execution:   we,
+		NamespaceId:     namespaceID,
+		Execution:       we,
+		SkipForceReload: true,
```
yycptt, line 647: It triggers a mutable state reload before the change.

service/history/ndc_standby_task_util.go
```diff
@@ -98,16 +98,17 @@ func isWorkflowExistOnSource(
 	if err != nil {
 		return true
 	}
-	_, remoteFrontend, err := clientBean.GetRemoteFrontendClient(remoteClusterName)
+	remoteAdminClient, err := clientBean.GetRemoteAdminClient(remoteClusterName)
 	if err != nil {
 		return true
 	}
-	_, err = remoteFrontend.DescribeWorkflowExecution(ctx, &workflowservice.DescribeWorkflowExecutionRequest{
+	_, err = remoteAdminClient.DescribeMutableState(ctx, &adminservice.DescribeMutableStateRequest{
```
yycptt, line 105: Well technically we need to handle backward compatibility here. But in the worst case we are just reloading the mutable state and considering task dropping is already rare, probably not worth introducing a feature flag here.

service/worker/migration/activities.go
```diff
@@ -740,7 +665,7 @@ func (a *activities) checkSkipWorkflowExecution(
 
 	// Zombie workflow should be a transient state. However, if there is Zombie workflow on the source cluster,
 	// it is skipped to avoid such workflow being processed on the target cluster.
-	if resp.GetDatabaseMutableState().GetExecutionState().GetState() == enumsspb.WORKFLOW_EXECUTION_STATE_ZOMBIE {
+	if resp.GetCacheMutableState().GetExecutionState().GetState() == enumsspb.WORKFLOW_EXECUTION_STATE_ZOMBIE {
```
yycptt: `CacheMutableState` field is always populated, no compatibility concern here. 

Old activity impl still uses `DatabaseMutableState` but also won't specify the `SkipForceReload` field, so `DatabaseMutableState` will still be populated.

hai719: Is there a case for [this check (`chasmLease.GetContext().(*workflow.ContextImpl).MutableState != nil`)](https://github.com/temporalio/temporal/blob/cd05266dcc89ecc8174c1d254f34a4f14a1c225a/service/history/api/describemutablestate/api.go#L45) to be false?

yycptt: yeah good point I had wrong impression about this api, I will fix this.

yycptt: Alright fixed. I kept the original behavior, CacheMutableState is only available when it's really in cache and DatabaseMutableState is always available. The new flag only decides if we force clear the cache or not.

hai719 (approved):

PR #8119, opened 2025-07-30
clear drainage status of a rollbacked version workflow

Shivs11:
## What changed?
- When a deactivated (drained/draining) version gets rollbacked to becoming active again, the drainage information that was associated with it will get cleared.
- This is required because when this version were to get deactivated again, we need the code to start the drainage process on it again. Right now, it was not doing that since we were not clearing the drainage information which did not start a fresh drainage tracker for this version!

*NOTE*: This change requires a patch to be present. This is because when this scenario is replicated in production, it will result in two new history calls (syncSummary as well as a CAN of the version workflow). This will specifically happen when the rolled-back version gets deactivated eventually, over [here](https://github.com/temporalio/temporal/blob/ada5d0bbee45150888b63a132be8239fd5a87212/service/worker/workerdeployment/version_workflow.go#L546) - note how previously, since the drainage information was not cleared, [this](https://github.com/temporalio/temporal/blob/ada5d0bbee45150888b63a132be8239fd5a87212/service/worker/workerdeployment/version_workflow.go#L261) was never executed!)

*How did I verify that this requires a patch*:

- Ran the replay test script before the fix introduced in this PR and noticed:
image

- Ran the replay test script after the fix introduced in this PR and noticed:
image

In other words, the total number of our entity workflows went up after introducing this change and thus requires a patch gate!

## Why?
- we found a bug in the versioning entity workflows and this is an effort to fix this!

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- yes, there is a risk of NDE's happening in production but I mitigated that risk by placing the fix against a version gate.


tests/worker_deployment_test.go
```diff
@@ -2019,6 +2019,363 @@ func (s *WorkerDeploymentSuite) verifyTaskQueueVersioningInfo(ctx context.Contex
 	}, time.Second*10, time.Millisecond*1000)
 }
 
+// This test shall first rollback a drained version to a current version. After that, it shall deploy a new version
```
carlydf, line 2022: since this is a really long test, would you mind also adding to this comment something about "this test reproduces a bug we saw in production where the drainage status was not being properly cleared when a draining version is reactivated and then re-deactivated"

service/worker/workerdeployment/version_workflow.go
```diff
@@ -552,6 +552,15 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
 			// First time this version is activated to current/ramping
 			d.VersionState.FirstActivationTime = args.RoutingUpdateTime
 		}
+
+		// Clear drainage information, if present, when a version gets activated.
+		// This handles the rollback scenario where a previously draining/drained version
+		// is reactivated and should have it's drainage information cleared.
+
+		v := workflow.GetVersion(ctx, "Step1", workflow.DefaultVersion, 0)
```
carlydf: does "Step1" have a special meaning that I should understand? it feels like this string is an opportunity to have a more meaningful label for the patch, but if you think Step1 is the best name then I trust you! just curious why you picked that

Shivs11: oh no absolutely not, step1 is a bad name for this and I can make it more meaningful. Changed it in my latest commit.

carlydf (approved):

PR #8120, opened 2025-07-30
first stab at saving histories after each continue-as-new

carlydf:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/service/worker/workerdeployment/replaytester/worker/worker.go
+++ b/service/worker/workerdeployment/replaytester/worker/worker.go
@@ -2,6 +2,10 @@ package main
 
 import (
 	"context"
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/history/v1"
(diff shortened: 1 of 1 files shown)


service/worker/workerdeployment/replaytester/worker/worker.go
```diff
@@ -151,6 +163,42 @@ func main() {
...
+			panic(err)
+		}
+		deploymentWFHistory = append(deploymentWFHistory, event)
+	}
+	// TODO: write history to json file
+
+	versionWFHistory := make([]*history.HistoryEvent, 0)
+	versionWFEvents := c.GetWorkflowHistory(
+		context.Background(),
+		worker_versioning.GenerateVersionWorkflowID(deploymentName, buildId),
```
carlydf, line 187: you might actually need to get the run id of the previous workflow run to do this correctly? not sure

PR #8123, opened 2025-07-30
Auto-approve temporal-cicd[bot] PRs on cloud release branches

lilydoar:
## What changed?

Pull Requests opened on cloud release branches by the temporal-cicd[bot] user will be auto-approved by this GH action.

## Why?

Launchpad creates PRs on cloud release branches as part of the patching process.
The goal is to not require users to manually approve these PRs as they are created.

## Potential risks

Auto approving changes on a release branch is risky.

- Only allows PRs created by the temporal-cicd bot
- Nothing is directly rolled out to production based on release branch changes.
- The new image goes through a full verification process before a user decides to roll out changes.


.github/workflows/auto-approve-cicd-release-pr.yml
```diff
@@ -0,0 +1,17 @@
+name: Auto-approve temporal-cicd[bot] PRs on cloud release branches
+on:
+  pull_request:
+    branches:
+      - cloud/**
+permissions:
+  contents: write
+  pull-requests: write
+jobs:
+  dependabot:
```
dnr: dependabot?

.github/workflows/auto-approve-cicd-release-pr.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'temporal-cicd[bot]' }}
```
dnr, line 12: I'm not really concerned, but just curious, how is this authenticated?

lilydoar: `temporal-cicd[bot]` is the actor when this action is triggered on a PR on a cloud release branch. If someone else opens a PR on a cloud release branch, they'll have to approve the same way they would for any PR

.github/workflows/auto-approve-cicd-release-pr.yml
lilydoar: Thanks, missed renaming that from the article I took this from.

https://jhale.dev/posts/auto-merging-prs/#automatically-enable-auto-merge-for-pull-requests-from-trusted-authors

If you're interested

.github/workflows/auto-approve-cicd-release-pr.yml
dnr: I guess I mean, is it possible for someone to fake that "actor" name by setting their github username or something? Or creating a PR and somehow tricking the bot into taking an action on it? I found this article which describes something like that: https://www.synacktiv.com/en/publications/github-actions-exploitation-dependabot
It looks like that probably only works with dependabot, but it suggests (at least to me) that this whole thing is pretty subtle.

Is it possible to take another approach and have the bot run `gh pr review --approve` or use the API directly, similar to how it creates the PR in the first place, instead of using GHA?

lilydoar: I agree with the concern. User's can't put `[bot]` in their name due to username constraints. GH has not concept of "assuming" an identity so we don't need to worry about someone assuming the role of `temporal-cicd[bot]`.

<https://docs.github.com/en/enterprise-cloud@latest/admin/managing-iam/iam-configuration-reference/username-considerations-for-external-authentication#about-username-normalization>

I would love to run this all through the API. Ideally, launchpad would just call createPR() with auto merge enabled, then call approvePR(). I don't think this works though because the same user who created the PR can't approve it. This GH action works because the approval comes from the `github-actions[bot]` user.

PR #8124, opened 2025-07-30
Extend versioning's replay tester

Shivs11:
## What changed?
- Make the replay tester, used by versioning entity workflows, better.
- The versioning entity workflows CAN on every state change. Previously, the replay tester was simply fetching the last execution of a versioning entity workflow. Thus, we were skipping over a lot of NDE's that could happen if one were to make changes in our workflows.

To fix this, the replay tester:
1. Fetches all the workflow ID's present in the CAN chain. Then, it tests if for each individual run, are there code changes that have caused NDE's.
2. It is important to remember that while we are fetching all the workflows in CAN chain, SDK's replay tester does not test the chaining itself. In other words, it tests the code path from the input that is given at the start of each workflow execution. Thus, there could be a chance that a code change introduced doesn't necessarily add new events into a single workflow run. However, this code change can increase the overall runs of the entity workflows and now the replay tester will see this and understand that this as a cause for NDE.
3. Neatly makes a folder for that run and saves the required workflow histories over there.
4. *Bonus*: I have also made the change to store the raw json file of each workflow history. This would make debugging a whole lotta simpler.

It shall now look like:
image

## Why?
- Robustness.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None


service/worker/workerdeployment/replaytester/testdata/run_1753913370/replay_deployment-workflow_initial.json.gz
Shivs11, line 1: this folder `run_1753913370` basically contains the workflow histories of histories that were present prior to this change.

service/worker/workerdeployment/replaytester/generate_history.sh
```diff
@@ -22,9 +22,82 @@ temporal operator namespace create default
...
+                -r "$run_id" \
+                --output json | \
+                tee "$run_dir/replay_${workflow_name}_run_${run_id}.json" | \
+                gzip -9c > "$run_dir/replay_${workflow_name}_run_${run_id}.json.gz"
+            
+            ((run_index++))
+        fi
+    done
+    
+    # # Save run IDs for reference
```
Shivs11, line 85: helpful if one wanted to compare the runID's being fetched vs the one's present on UI/visibility

carlydf (approved):

PR #8126, opened 2025-07-31
Consolidate cassandra task queue management queries

dnr:
## What changed?
Remove some code duplication for cassandra matching task queue persistence by rewriting v2 to v1 queries.
Switch same scheme for SQL queries to use the v2 table name in the base query and rewriting v1.
Use panic instead of returning an invalid query.

## Why?
Remove code duplication.

## How did you test it?

### DIFF
--- a/common/persistence/cassandra/matching_task_store_queue.go
+++ b/common/persistence/cassandra/matching_task_store_queue.go
@@ -1,0 +1,338 @@
+package cassandra
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/convert"
+	p "go.temporal.io/server/common/persistence"
+	"go.temporal.io/server/common/persistence/nosql/nosqlplugin/cassandra/gocql"
+	"go.temporal.io/server/common/primitives/timestamp"
+)
+
+// CassandraTaskVersion represents the task schema version
+type CassandraTaskVersion int
+
+const (
+	CassandraTaskVersion1 CassandraTaskVersion = 1
+	CassandraTaskVersion2 CassandraTaskVersion = 2
+)
+
+var switchTasksTableV2Cache sync.Map
+
+// SwitchTasksTable switches table names from tasks to tasks_v2 and modifies queries for v2 schema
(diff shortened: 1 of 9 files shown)


common/persistence/cassandra/matching_task_store_queue.go
```diff
@@ -0,0 +1,275 @@
...
+		panic("invalid task schema version") // nolint:forbidigo // hardcoded constants
+	}
+
+	if v1query, ok := switchTasksTableV1Cache.Load(baseQuery); ok {
+		return v1query.(string) // nolint:revive
+	}
+
+	v1query := strings.ReplaceAll(baseQuery, " tasks_v2 ", " tasks ")
+	v1query = strings.ReplaceAll(v1query, " AND pass = 0 ", " ")
+	v1query = strings.ReplaceAll(v1query, "type, pass, task_id", "type, task_id")
```
stephanos, line 42: That's very specific 😬 How confident are we that the tests are catching any (accidental) re-ordering of these?

common/persistence/sql/sqlplugin/matching_task_queue.go
```diff
@@ -59,18 +59,18 @@ const (
...
 	}
-	if v2query, ok := switchTaskQueuesTableV2Cache.Load(baseQuery); ok {
-		return v2query.(string) // nolint:revive
+	if v1query, ok := switchTaskQueuesTableV1Cache.Load(baseQuery); ok {
+		return v1query.(string) // nolint:revive
 	}
-	v2query := strings.ReplaceAll(baseQuery, " task_queues ", " task_queues_v2 ")
-	switchTaskQueuesTableV2Cache.Store(baseQuery, v2query)
-	return v2query
+	v1query := strings.ReplaceAll(baseQuery, " task_queues_v2 ", " task_queues ")
```
stephanos, line 73: That's actually a good idea to flip it 👍

stephanos (approved):

common/persistence/cassandra/matching_task_store_queue.go
dnr: well, if it fails to replace, the query will be left referring to a column that doesn't exist in the table and so it'll fail

stephanos: Right; I suppose all queries are executed in the functional tests and if they aren't, probably in CICD.

PR #7991, opened 2025-06-30
Explain test build tags and env variables

stephanos:
## What changed?
_Describe what has changed in this PR._

## Why?
_Tell your future self why have you made these changes._

## How did you test it?
- [ ] built
- [ ] run locally and tested manually
- [ ] covered by existing tests
- [ ] added new unit test(s)
- [ ] added new functional test(s)

## Potential risks
_Any change is risky. Identify all risks you are aware of. If none, remove this section._

PR #8128, opened 2025-07-31
debugtimeout.Multiplier

stephanos:
## What changed?

Right now, `TEMPORAL_DEBUG` is build tag that multiplies selected timeouts with `100` to give more time to debug an issue when running tests.

This PR 

- changes it from a build tag an env variable
- renames it from `TEMPORAL_DEBUG` to `TEMPORAL_DEBUG_TIMEOUT` (clearer!)
- changes the default from 100 to 10
- makes it configurable 

I also had to change a few `const`s to `var`s now.

## Why?

All other test-specific overrides are env variables.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

Once this is merged, I'll get [docs](https://github.com/temporalio/temporal/pull/7991) merged.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -13,7 +13,6 @@ import (
 	"go.temporal.io/server/common/authorization"
 	"go.temporal.io/server/common/build"
 	"go.temporal.io/server/common/config"
-	"go.temporal.io/server/common/debug"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/headers"
 	"go.temporal.io/server/common/log"
@@ -174,7 +173,6 @@ func buildCLI() *cli.App {
 					tag.NewStringTag("go-version", build.InfoData.GoVersion),
 					tag.NewBoolTag("cgo-enabled", build.InfoData.CgoEnabled),
 					tag.NewStringTag("server-version", headers.ServerVersion),
-					tag.NewBoolTag("debug-mode", debug.Enabled),
 				)
 
 				var dynamicConfigClient dynamicconfig.Client
--- a/common/persistence/persistence-tests/cluster_metadata_manager.go
+++ b/common/persistence/persistence-tests/cluster_metadata_manager.go
@@ -10,9 +10,9 @@ import (
 	"go.temporal.io/api/serviceerror"
 	versionpb "go.temporal.io/api/version/v1"
 	persistencespb "go.temporal.io/server/api/persistence/v1"
-	"go.temporal.io/server/common/debug"
 	p "go.temporal.io/server/common/persistence"
 	"go.temporal.io/server/common/primitives"
+	"go.temporal.io/server/common/testing/debugtimeout"
 )
 
 type (
@@ -36,7 +36,7 @@ func (s *ClusterMetadataManagerSuite) SetupSuite() {
 func (s *ClusterMetadataManagerSuite) SetupTest() {
 	// Have to define our overridden assertions in the test setup. If we did it earlier, s.T() will return nil
 	s.Assertions = require.New(s.T())
-	s.ctx, s.cancel = context.WithTimeout(context.Background(), 30*time.Second*debug.TimeoutMultiplier)
+	s.ctx, s.cancel = context.WithTimeout(context.Background(), 30*time.Second*debugtimeout.Multiplier)
 }
 
 // TearDownTest implementation
--- a/common/persistence/sql/sqlplugin/tests/context.go
+++ b/common/persistence/sql/sqlplugin/tests/context.go
@@ -4,12 +4,12 @@ import (
 	"context"
 	"time"
 
-	"go.temporal.io/server/common/debug"
+	"go.temporal.io/server/common/testing/debugtimeout"
 )
 
-const (
-	executionTimeout  = 2 * time.Second * debug.TimeoutMultiplier
-	visibilityTimeout = 4 * time.Second * debug.TimeoutMultiplier
+var (
+	executionTimeout  = 2 * time.Second * debugtimeout.Multiplier
+	visibilityTimeout = 4 * time.Second * debugtimeout.Multiplier
 )
 
 func newExecutionContext() context.Context {
--- a/common/persistence/tests/execution_mutable_state.go
+++ b/common/persistence/tests/execution_mutable_state.go
@@ -19,12 +19,12 @@ import (
 	persistencespb "go.temporal.io/server/api/persistence/v1"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/convert"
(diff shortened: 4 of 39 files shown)


common/testing/debugtimeout/debugtimeout.go
```diff
@@ -0,0 +1,18 @@
+//go:build test_dep || integration
```
stephanos, line 1: I don't know if we want to include `integration`. We don't use it in OSS. But I know consumers of the Server use it. So we might as well make it easy for them to use this as-is.

common/primitives/constants.go
```diff
@@ -30,3 +25,8 @@ const (
 const (
 	ScheduleWorkflowIDPrefix = "temporal-sys-scheduler:"
 )
+
+var (
+	// DefaultWorkflowTaskTimeout sets the Default Workflow Task timeout for a Workflow
+	DefaultWorkflowTaskTimeout = 10 * time.Second * debugtimeout.Multiplier
+)
```
stephanos, line 32: This is not a const anymore, but it feels const-ish enough to leave it in this file.

cmd/server/main.go
```diff
@@ -174,7 +173,6 @@ func buildCLI() *cli.App {
 					tag.NewStringTag("go-version", build.InfoData.GoVersion),
 					tag.NewBoolTag("cgo-enabled", build.InfoData.CgoEnabled),
 					tag.NewStringTag("server-version", headers.ServerVersion),
-					tag.NewBoolTag("debug-mode", debug.Enabled),
```
stephanos, line 177: Removed as not applicable anymore.

common/testing/debugtimeout/debugtimeout.go
```diff
@@ -0,0 +1,18 @@
...
+	"time"
+)
+
+var Multiplier = func() time.Duration {
+	if v := os.Getenv("TEMPORAL_DEBUG_TIMEOUT"); v != "" {
+		if parsed, err := strconv.ParseInt(v, 10, 64); err == nil {
+			return time.Duration(parsed)
+		}
+	}
+	return time.Duration(10)
```
stephanos, line 17: Right now it's `100`, `10` is probably enough.

common/testing/debugtimeout/debugtimeout.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+	if v == "" {
+		return time.Duration(10)
+	}
+	parsed, err := strconv.ParseInt(v, 10, 64)
+	if err != nil {
+		panic("TEMPORAL_DEBUG_TIMEOUT should not be empty or a valid integer, got: " + v)
+	}
+	return time.Duration(parsed)
+}
```
stephanos, line 26: Compared to the original:

- I changed the behavior to only use `10` when the env var is specified; and stay at `1` otherwise. Our functional tests were actually failing when it was set to `10`.
- I changed it to panic since this is only used in tests and can inform the developer that they're using this wrong.

common/persistence/persistence-tests/cluster_metadata_manager.go
```diff
@@ -36,7 +36,7 @@ func (s *ClusterMetadataManagerSuite) SetupSuite() {
 func (s *ClusterMetadataManagerSuite) SetupTest() {
 	// Have to define our overridden assertions in the test setup. If we did it earlier, s.T() will return nil
 	s.Assertions = require.New(s.T())
-	s.ctx, s.cancel = context.WithTimeout(context.Background(), 30*time.Second*debug.TimeoutMultiplier)
+	s.ctx, s.cancel = context.WithTimeout(context.Background(), 30*time.Second*debugtimeout.Multiplier)
```
paulnpdev, line 39: I think this one and those below are antipatterns.  This IS a test by definition, and doesn't need a multiplier.  It should set the timeout it needs.  If 300 seconds is the right answer here it should be coded that way, it would be clearer.

common/persistence/sql/sqlplugin/tests/context.go
```diff
@@ -4,12 +4,12 @@ import (
...
 
-	"go.temporal.io/server/common/debug"
+	"go.temporal.io/server/common/testing/debugtimeout"
 )
 
-const (
-	executionTimeout  = 2 * time.Second * debug.TimeoutMultiplier
-	visibilityTimeout = 4 * time.Second * debug.TimeoutMultiplier
+var (
+	executionTimeout  = 2 * time.Second * debugtimeout.Multiplier
```
paulnpdev, line 11: ditto.  This code is test-only

common/persistence/tests/execution_mutable_state.go
```diff
@@ -86,7 +86,7 @@ func (s *ExecutionMutableStateSuite) TearDownSuite() {
 
 func (s *ExecutionMutableStateSuite) SetupTest() {
 	s.Assertions = require.New(s.T())
-	s.Ctx, s.Cancel = context.WithTimeout(context.Background(), 30*time.Second*debug.TimeoutMultiplier)
+	s.Ctx, s.Cancel = context.WithTimeout(context.Background(), 30*time.Second*debugtimeout.Multiplier)
```
paulnpdev, line 89: ditto
I'm going to stop mentioning this issue.

common/testing/debugtimeout/debugtimeout.go
paulnpdev: I don't think this is a good decision.  The version in ST results in 10 for all tests. Your approach means fixing tests one at a time to set the ENV var.  

I suspect your functional tests are failing because of what I said earlier: they are using the multiplier *in the tests*.  They should not.  In ST we only use the multiplier to affect "production" timeouts.

also, your scheme is more complicated ... not seeing the value.  Though I like the panic

paulnpdev (commented):
my thoughts

common/testing/debugtimeout/debugtimeout.go
stephanos: I was surprised by this comment so I went back to the original PR descriptions and realized that these two mechanisms seem to look the same at first but have different goals.
ST:

> It should resolve a large number of CI test flakes (and provide a mechanism for fixing others) [...].

OSS:

> Default RPC timeouts work great for production environments. While in server debugging mode it is almost impossible to step over the code with default values.

That would explain your comments about using it in tests.

> I don't think this is a good decision. The version in ST results in 10 for all tests. Your approach means fixing tests one at a time to set the ENV var.

Hm. Nothing is stopping a user from just setting TEMPORAL_DEBUG_TIMEOUT as the default locally or in CI. I think bumping all timeouts in tests by default is very surprising. This makes it explicit.

> also, your scheme is more complicated

Hm. Is this referring to the fact that there is one more state (the default of 1) or sth else? Because other than that it seems the same to me.

Issue #8037, opened 2025-07-14
worker service abnormal restart


tunsuy:
## Steps to Reproduce the Problem

 start temporal worker servcie, log is:

```
TEMPORAL_ADDRESS is not set, setting it to 9.139.60.15:7233
2025/07/14 12:28:05 Loading config; env=docker,zone=,configDir=config
2025/07/14 12:28:05 Loading config files=[config/docker.yaml]
{"level":"info","ts":"2025-07-14T12:28:05.568Z","msg":"Build info.","git-time":"2025-03-27T22:45:52.000Z","git-revision":"a31fd7a9e5db4c28627eb18676558e974e95e1ba","git-modified":false,"go-arch":"amd64","go-os":"linux","go-version":"go1.23.2","cgo-enabled":false,"server-version":"1.27.2","debug-mode":false,"logging-call-at":"/home/runner/work/docker-builds/docker-builds/temporal/cmd/server/main.go:175"}
{"level":"info","ts":"2025-07-14T12:28:05.568Z","msg":"dynamic config changed for the key: worker.historyscannerverifyretention oldValue: nil newValue: { constraints: {} value: true }","logging-call-at":"/home/runner/work/docker-builds/docker-builds/temporal/common/dynamicconfig/file_based_client.go:361"}
{"level":"info","ts":"2025-07-14T12:28:05.568Z","msg":"dynamic config changed for the key: worker.historyscannerdataminage oldValue: nil newValue: { constraints: {} value: 604800000000000 }","logging-call-at":"/home/runner/work/docker-builds/docker-builds/temporal/common/dynamicconfig/file_based_client.go:361"}
{"level":"info","ts":"2025-07-14T12:28:05.568Z","msg":"dynamic config changed for the key: limit.historycount.error oldValue: nil newValue: { constraints: {} value: 102400 }","logging-call-at":"/home/runner/work/docker-builds/docker-builds/temporal/common/dynamicconfig/file_based_client.go:361"}
{"level":"info","ts":"2025-07-14T12:28:05.568Z","msg":"dynamic config changed for the key: history.cachettl oldValue: nil newValue: { constraints: {} value: 600000000000 }","logging-call-at":"/home/runner/work/docker-builds/docker-builds/temporal/common/dynamicconfig/file_based_client.go:361"}
[...]

PR #8133, opened 2025-07-31
Break out of pagination in scavenger on errors

simvlad:
## What changed?
Break out of the loop, when iteration through mutable states fails

## Why?
Previously, we continued to iterate, leading to the panic: https://github.com/temporalio/temporal/issues/8037

## How did you test it?
- [X] added new unit test(s)

### DIFF
--- a/service/worker/scanner/scanner.go
+++ b/service/worker/scanner/scanner.go
@@ -159,10 +159,12 @@ func (s *Scanner) Start() error {
 	}
 
 	var workerTaskQueueNames []string
-	if s.context.cfg.ExecutionsScannerEnabled() {
+	if s.context.cfg.Persistence.DefaultStoreType() != config.StoreTypeSQL && s.context.cfg.ExecutionsScannerEnabled() {
 		s.wg.Add(1)
 		go s.startWorkflowWithRetry(ctx, executionsScannerWFStartOptions, executionsScannerWFTypeName)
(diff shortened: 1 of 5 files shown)


simvlad:
> Can we also enforce the execution scanner workflow won't start when SQL persistence is used? regardless of the feature flag value.

Done!

service/worker/scanner/scanner.go
```diff
@@ -159,10 +159,12 @@ func (s *Scanner) Start() error {
 	}
 
 	var workerTaskQueueNames []string
-	if s.context.cfg.ExecutionsScannerEnabled() {
+	if s.context.cfg.Persistence.DefaultStoreType() != config.StoreTypeSQL && s.context.cfg.ExecutionsScannerEnabled() {
 		s.wg.Add(1)
 		go s.startWorkflowWithRetry(ctx, executionsScannerWFStartOptions, executionsScannerWFTypeName)
 		workerTaskQueueNames = append(workerTaskQueueNames, executionsScannerTaskQueueName)
+	} else if s.context.cfg.ExecutionsScannerEnabled() {
```
yycptt, line 166: nit: plz update the [description](https://github.com/temporalio/temporal/blob/44acb3579860b7648fad2e5678619186c6c16f3b/common/dynamicconfig/constants.go#L2612) for this config accordingly. Thanks!

yycptt (approved):

PR #8135, opened 2025-08-01
TaskQueue Fairness Rate Limit

sivagirish81:
## What changed?
 - PR dependent on : https://github.com/temporalio/temporal/pull/8058 [ Will merge after this PR is reviewed and merged ]
 -  Move the rate limit logic for fairness from priMatcher to taskQueuePartitionManagerLevel
 -  Attach the fairness queue rate limit and the per-key rate limit to the simple rate limiter implementation.

## Why?
- Implementation of UpdateTaskQueueConfig api for fairness tasks.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
N/A


sivagirish81:
Just a heads up — this branch was originally based off a previous PR [#8058](https://github.com/temporalio/temporal/pull/8058) that’s now merged. I’ve merged the latest main into this branch, but you may still see some of those earlier commits showing up in the PR history due to how Git handles ancestry.

The file diff is clean and correct, and since we squash-and-merge, the commit history here shouldn’t affect main. So we can continue reviewing this PR as-is.

I’ve also created a clean branch [#8152](https://github.com/temporalio/temporal/pull/8152) with just the relevant changes cherry-picked, in case you’d prefer to review there instead.

service/matching/config.go
```diff
@@ -380,6 +383,10 @@ func newTaskQueueConfig(tq *tqid.TaskQueue, config *Config, ns namespace.Name) *
 		NumReadPartitions: func() int {
 			return max(1, config.NumTaskqueueReadPartitions(ns.String(), taskQueueName, taskType))
 		},
+		NumReadPartitionsSub: func(cb func(int)) (int, func()) {
+			val, cancel := config.NumTaskqueueReadPartitionsSub(ns.String(), taskQueueName, taskType, cb)
+			return max(1, val), cancel
```
dnr: the `max` here is not really useful, the callback is going to have to do it anyway, so just remove it here and let the callback handle invalid values

service/matching/matcher_data.go
```diff
@@ -550,7 +452,9 @@ func (d *matcherData) findAndWakeMatches() {
 		}
 
 		// check ready time
-		delay := d.tasks.readyTimeForTask(task).delay(now)
+		d.rateLimitManager.mu.Lock()
+		delay := d.rateLimitManager.readyTimeForTask(task).delay(now)
+		d.rateLimitManager.mu.Unlock()
```
dnr: this should not be locking some other object's lock, rateLimitManager should handle its own locking

service/matching/matcher_data_test.go
```diff
@@ -310,9 +320,29 @@ func (s *MatcherDataSuite) TestTaskForward() {
 	s.ErrorIs(fres.forwardErr, someError)
 }
 
+// Sets both QueueRateLimit and FairnessKeysRateLimitDefault for the given task queue type
+func setRPSForTaskQueueUserData(
+	userData *persistencespb.TaskQueueUserData,
+	taskQueueType enumspb.TaskQueueType,
+	rps float32,
+) {
+	// Set user data to 10 tasks/key/sec with burst of 3
```
dnr: burst of 3?

service/matching/matcher_data_test.go
```diff
@@ -310,9 +320,29 @@ func (s *MatcherDataSuite) TestTaskForward() {
...
+	userData.PerType = make(map[int32]*persistencespb.TaskQueueTypeUserData)
+	key := int32(taskQueueType)
+	userData.PerType[key] = &persistencespb.TaskQueueTypeUserData{}
+	tqTypeData := userData.PerType[key]
+	tqTypeData.Config = &taskqueuepb.TaskQueueConfig{}
+	rateLimit := &taskqueuepb.RateLimit{RequestsPerSecond: rps}
+	rateLimitConfig := &taskqueuepb.RateLimitConfig{RateLimit: rateLimit}
+	tqTypeData.Config.QueueRateLimit = rateLimitConfig
+	tqTypeData.Config.FairnessKeysRateLimitDefault = rateLimitConfig
```
dnr: you can just write this with map literals? I think that would be easier to read

service/matching/matcher_data_test.go
```diff
@@ -310,9 +320,29 @@ func (s *MatcherDataSuite) TestTaskForward() {
...
+}
+
 func (s *MatcherDataSuite) TestRateLimitedBacklog() {
-	// 10 tasks/sec with burst of 3
-	s.md.UpdateRateLimit(10, 300*time.Millisecond)
+	s.md.rateLimitManager.mu.Lock()
```
dnr: do not mess with another object's lock

service/matching/matching_engine_test.go
```diff
@@ -1014,6 +1014,10 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 	dbq := newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)
 	mgr := s.newPartitionManager(dbq.partition, s.matchingEngine.config)
 
+	// Set the admin namespace and task queue dispatch rates to 25000.0 to enforce the override of the dynamic config.
+	mgr.GetRateLimitManager().adminNsRate = 25000.0
+	mgr.GetRateLimitManager().adminTqRate = 25000.0
```
dnr: this feels kind of fragile, these fields are protected by a mutex, right?

if we need to do this, maybe we could add a "SetAdminRateForTesting" method to rateLimitManager, and we could even put it in a _test.go file so it only exists in tests

service/matching/pri_matcher.go
```diff
@@ -143,10 +131,7 @@ func (tm *priTaskMatcher) Start() {
 	}
 }
 
-func (tm *priTaskMatcher) Stop() {
-	tm.cancel1()
-	tm.cancel2()
-}
+func (tm *priTaskMatcher) Stop() {}
```
dnr, line 126: let's just remove this entirely?

service/matching/pri_matcher.go
```diff
@@ -498,50 +483,11 @@ func (tm *priTaskMatcher) ReprocessAllTasks() {
...
-
-	tm.data.UpdateRateLimit(rate, burstDuration)
-}
-
 // Rate returns the current dynamic rate setting
 func (tm *priTaskMatcher) Rate() float64 {
-	tm.limiterLock.Lock()
-	defer tm.limiterLock.Unlock()
-	return tm.dynamicRate
+	tm.rateLimitManager.mu.Lock()
```
dnr: do not mess with other object's locks. also I'm guessing we can just remove this?

service/matching/ratelimit_manager.go
```diff
@@ -17,28 +20,66 @@ type (
 		userDataManager userDataManager         // User data manager to fetch user data for task queue configurations.
 		workerRPS       *float64                // RPS set by worker at the time of polling, if available.
 		apiConfigRPS    *float64                // RPS set via API, if available.
+		fairnessRPS     *float64                // fairnessKeyRateLimitDefault set via API, if available
```
dnr: `fairnessRPS` doesn't really mean anything.. if the value is "fairness key default rate limit" then name it that

service/matching/ratelimit_manager.go
```diff
@@ -17,28 +20,66 @@ type (
...
 		systemRPS       float64                 // Min of partition level dispatch rates times the number of read partitions.
 		config          *taskQueueConfig        // Dynamic configuration for task queues set by system.
 		taskQueueType   enumspb.TaskQueueType   // Task queue type
 
+		fairnessEnabled   bool // Whether fairness is enabled for this task queue.
+		newMatcherEnabled bool // Whether the new matcher is enabled for this task queue.
+		isSticky          bool // Whether this is a sticky task queue.
```
dnr: hmm, I wouldn't have thought it would need these bits of info

service/matching/ratelimit_manager.go
```diff
@@ -17,28 +20,66 @@ type (
...
+		// weight. If we do this, we can assume that all keys are either at or below their rate
+		// limit at the same time, so that if the head of the queue is at its limit, then the rest
+		// must also be, and so we don't have to "skip over" the head of the queue due to rate
+		// limits. This isn't true in situations where weights have changed in between writing and
+		// reading. We'll handle that situation better in the future.
+		perKeyLimit                            simpleLimiterParams
+		perKeyReady                            cache.Cache
+		perKeyOverrides                        fairnessWeightOverrides // TODO(fairness): get this from config
+		cancel1, cancel2, cancel3              func()
+		cancelFairnessSub, cancelNewMatcherSub func() // Cancel functions for dynamic config subscriptions
```
dnr: why do two of these have names and three have numbers? be consistent. anyway, it may be simpler to do `[]func()` with so many

service/matching/ratelimit_manager.go
```diff
@@ -48,17 +89,54 @@ func newRateLimitManager(userDataManager userDataManager,
...
+func (r *rateLimitManager) setAdminTqRate(rps float64) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.adminTqRate = rps
+}
+
+func (r *rateLimitManager) setNumReadPartitions(val int) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if val <= 0 {
```
dnr: just use max
```suggestion
	r.numReadPartitions = max(1, val)
```

service/matching/ratelimit_manager.go
```diff
@@ -138,8 +221,14 @@ func (r *rateLimitManager) GetRateLimiter() quotas.RateLimiter {
 func (r *rateLimitManager) UserDataChanged() {
 	r.mu.Lock()
 	defer r.mu.Unlock()
+	// Always check for api configured rate limits before any update to the rate limits.
+	r.trySetRPSFromUserDataLocked()
 	// Immediately recompute and apply rate limit if necessary.
 	r.updateRatelimitLocked()
+	if r.fairnessEnabled || r.newMatcherEnabled {
```
dnr: is this all this is for? just do this all the time and don't worry about what's enabled

service/matching/ratelimit_manager.go
```diff
@@ -154,30 +243,40 @@ func (r *rateLimitManager) trySetRPSFromUserDataLocked() {
...
-	burst := int(math.Ceil(effectiveRPSPartitionWise))
-	burst = max(burst, r.config.MinTaskThrottlingBurstSize())
-	r.dynamicRateBurst.SetRPS(effectiveRPSPartitionWise)
+	// If the effective RPS is zero, we set the burst to zero as well.
+	// This prevents any initial tasks from executing immediately.
+	// Allows pausing of the task queue by setting the RPS to zero.
+	var burst int
+	if newRPS != 0 {
+		// If the effective RPS is non-zero, we can set a burst based on the effective RPS.
+		burst = max(int(math.Ceil(newRPS)), r.config.MinTaskThrottlingBurstSize())
```
dnr, line 248: we should probably work defaultBurstDuration into this calculation to be consistent

service/matching/ratelimit_manager.go
```diff
@@ -186,3 +285,120 @@ func (r *rateLimitManager) updateRatelimitLocked() {
 		r.dynamicRateLimiter.Refresh()
 	})
 }
+
+// UpdateSimpleRateLimit updates the overall queue rate limits for the simpleRateLimiter implementation
+func (r *rateLimitManager) updateSimpleRateLimitLocked(burstDuration time.Duration) {
+	r.computeEffectiveRPSAndSourceLocked()
```
dnr: I'm pretty confused here, I would have thought this would have been computed already and we're just updating the simple limiter params. It feels like the control flow is going in a bunch of directions instead of in one direction and it's hard to follow. This is always called after updateRateLimitLocked, which already does this, right?

I would suggest reorganizing things as follows:

- Have a set of fields in the state that are clearly marked "sources": the dynamic config, the worker rps, and the api config.
- Have some functions that update the sources: dynamic config subscription callbacks, InjectWorkerRPS, and UserDataChanged. They should just update their respective source and then calls:
- Have _one_ function that:
  - combines the sources into the effective values
  - sets the effective values in the state
  - passes those effective values to to everywhere they need to be to take effect

I think it's not that far off, but e.g. the stuff that's happening in this function should be way downstream of the computed values and not recompute anything, and we should not have multiple calls to updateSimpleRateLimitLocked, it should be called from that one single bottleneck. We shouldn't have multiple calls to trySetRPSFromUserDataLocked either.

service/matching/config.go
sivagirish81: Fixed to handle this exclusively in the callback.

service/matching/matcher_data.go
sivagirish81: Moved the locking to rateLimitManager.

service/matching/matcher_data_test.go
sivagirish81: Removed this. Since this is a unit test I avoided checking the propagation of values from the userData into the rateLimitManager and instead forcefully override the effectiveRPS and source.

service/matching/matching_engine_test.go
sivagirish81: Moved this logic to ratelimit_manager_test.go to set the adminNsRate and adminTqRate along with the lock acquired.

service/matching/ratelimit_manager.go
sivagirish81: Changed it to fairness_key_rate_limit_default.

service/matching/ratelimit_manager.go
sivagirish81: Changed it to []func().

service/matching/ratelimit_manager.go
sivagirish81: Done.

service/matching/ratelimit_manager.go
sivagirish81: Removed the checks for fairness and newMatcher. The simpleratelimiter will be created & set at all times from now on regardless of whether we have fairness/newMatcher enabled.

service/matching/pri_matcher.go
sivagirish81: priTaskMatcher implements the matcherInterface so it kind of needs to be there.

service/matching/ratelimit_manager.go
sivagirish81: Removed it and allowing simpleRateLimiter to get created at all times irrespective of fairness/newMatcher enabled.

tests/task_queue_test.go
```diff
@@ -260,6 +243,56 @@ func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
...
+// Burst = 5 i.e max(int(math.Ceil(effectiveRPSPartitionWise)), r.config.MinTaskThrottlingBurstSize())
+// The expected time for all activities to complete is ~ 4 seconds ((25 - 5)/5) +/- 1 second buffer.
+// The first five activities should run immediately, and the rest should be throttled to 5 RPS.
+// The test verifies that the total time taken for all activities to complete is within the expected range
+// To avoid test flakiness, the test uses a buffer of 1 second for the expected total time.
+func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
+	const (
+		apiRPS            = 5.0
+		taskCount         = 25
+		buffer            = 2 * time.Second // High Buffer to account for test flakiness
```
sivagirish81: Had to increase the buffer as it was flaking in the ci. Did not flake with a 1 second buffer on my local machine. Will check once again and if possible revert back to 1s.

service/matching/ratelimit_manager.go
sivagirish81: Reorganized the calls to avoid multiple calls to update rate limits.
+ `computeEffectiveRPSAndSourceLocked`
    + All the subscription callbacks will recompute and set the effectiveRPS and source whenever an incoming change happens.
    + InjectWorkerRPS will recompute the effectiveRPS post any updates to workerRPS.
    + UserDataChanges will recompute the effectiveRPS post any changes to userData
+  `trySetRPSFromUserData` is called only upon UserDataChanges.
+  UpdateRateLimits
   + `updateRatelimitLocked` and `updateSimpleRateLimitLocked` are only called inside `InjectWorkerRPS' i.e at the time of polling and when userDataChanges.
   + `updatePerKeySimpleRateLimitLocked` is only called when userDataChanges as the api is the single source for `fairnessKeyRateLimitDefault`.

service/matching/pri_matcher.go
sivagirish81: pri_matcher implements the matcher interface hence retaining this. Fixed the locking.

service/matching/matcher_data_test.go
sivagirish81: Fixed. The userDataUpdate handles only the RPS that is 10 RPS while the burst is passed as a parameter.

service/matching/matcher_data_test.go
sivagirish81: Kept the locking within rateLimitmanager.

service/matching/matcher_data.go
```diff
@@ -268,6 +211,7 @@ type matcherData struct {
 
 	rateLimitTimer         resettableTimer
 	reconsiderForwardTimer resettableTimer
+	rateLimitManager       *rateLimitManager
```
dnr: put this in the first block of fields, above lock, since it's constant and set at construction time

service/matching/matching_engine_test.go
```diff
@@ -1014,6 +1014,9 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 	dbq := newUnversionedRootQueueKey(namespaceId, tl, enumspb.TASK_QUEUE_TYPE_ACTIVITY)
 	mgr := s.newPartitionManager(dbq.partition, s.matchingEngine.config)
 
+	// Set the admin namespace and task queue dispatch rates to 25000.0 to enforce the override of the dynamic config.
+	mgr.GetRateLimitManager().SetAdminRateForTesting(25000.0)
```
dnr: can you explain more what this is doing? it looks like the config is set to 25000 above so why do we need this?

service/matching/physical_task_queue_manager.go
```diff
@@ -463,7 +456,9 @@ func (c *physicalTaskQueueManagerImpl) AddSpooledTask(task *internalTask) error
 }
 
 func (c *physicalTaskQueueManagerImpl) AddSpooledTaskToMatcher(task *internalTask) {
-	c.priMatcher.AddTask(task)
+	if c.priMatcher != nil {
```
dnr: this check shouldn't be needed: this should only be called by priBacklogManager or fairBacklogManager, where priMatcher will not be nil. if this is getting called when priMatcher is nil then something else is wrong. (and in any case, just ignoring the task is certainly not right)

service/matching/pri_matcher.go
```diff
@@ -46,7 +46,7 @@ type priTaskMatcher struct {
 	adminTqRate float64
 	dynamicRate float64
```
dnr: these were moved to rateLimitManager, right? are these fields still in use?

service/matching/pri_matcher.go
```diff
@@ -46,7 +46,7 @@ type priTaskMatcher struct {
...
 
-	cancel1, cancel2 func()
+	rateLimitManager *rateLimitManager
```
dnr, line 38: I'd move this field in with fwdr and validator, it fits in with those dependencies. in general order struct fields like: dependencies (constant), other constants, then mutable stuff (this one doesn't exactly follow that rule right now)

service/matching/pri_matcher.go
dnr: ah ok

service/matching/pri_matcher.go
```diff
@@ -498,50 +483,9 @@ func (tm *priTaskMatcher) ReprocessAllTasks() {
...
-		perPartitionDynamicRate,
-		tm.adminNsRate,
-		tm.adminTqRate,
-	)
-
-	tm.data.UpdateRateLimit(rate, burstDuration)
-}
-
 // Rate returns the current dynamic rate setting
 func (tm *priTaskMatcher) Rate() float64 {
```
dnr: can we remove this?

service/matching/ratelimit_manager.go
```diff
@@ -113,10 +168,15 @@ func (r *rateLimitManager) InjectWorkerRPS(meta *pollMetadata) {
 			rps = &value
 		}
 	}
+	// Short-circuit if there's no change from the previous worker RPS.
```
dnr: no, let's not add complicated logic to do extra short-circuits. there should be exactly one short-circuit check: when we have the new effective number, compare it to the last, and only update the actual limiters if it changed.

service/matching/ratelimit_manager.go
```diff
@@ -125,10 +185,15 @@ func (r *rateLimitManager) InjectWorkerRPS(meta *pollMetadata) {
 func (r *rateLimitManager) GetEffectiveRPSAndSource() (float64, enumspb.RateLimitSource) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	r.computeEffectiveRPSAndSourceLocked()
 	return r.effectiveRPS, r.rateLimitSource
 }
 
+func (r *rateLimitManager) GetEffectiveRPS() float64 {
```
dnr: remove this. even if you did need priTaskMatcher.Rate(), it should just call GetEffectiveRPSAndSource and ignore the source. but we can get rid of that one too.

service/matching/ratelimit_manager.go
```diff
@@ -138,11 +203,27 @@ func (r *rateLimitManager) GetRateLimiter() quotas.RateLimiter {
...
+	// This avoids unnecessary updates and rate limit recalculations.
+	if (r.apiConfigRPS == nil && oldQueueRateLimit != nil) ||
+		(r.apiConfigRPS != nil && *r.apiConfigRPS != r.effectiveRPS) {
+		r.updateRatelimitLocked()
+		r.updateSimpleRateLimitLocked(defaultBurstDuration)
+	}
+	if (r.fairnessKeyRateLimitDefault == nil && oldFairnessKeyRateLimitDefault != nil) ||
+		(r.fairnessKeyRateLimitDefault != nil && oldFairnessKeyRateLimitDefault != nil && *oldFairnessKeyRateLimitDefault != *r.fairnessKeyRateLimitDefault) {
+		r.updatePerKeySimpleRateLimitLocked(defaultBurstDuration)
+	}
```
dnr: none of this should be here. there are still two calls to updateRatelimitLocked and updateSimpleRateLimitLocked. there should be exactly one, in some function that all changes flow through.

service/matching/ratelimit_manager.go
```diff
@@ -154,30 +235,34 @@ func (r *rateLimitManager) trySetRPSFromUserDataLocked() {
 	}
```
dnr: if config == nil, don't we have to set `r.apiConfigRPS` to nil? but if we just remove this `if config == nil` check, the next few lines will handle that properly

service/matching/ratelimit_manager.go
```diff
@@ -186,3 +271,124 @@ func (r *rateLimitManager) updateRatelimitLocked() {
...
+// UpdateTaskQueueConfig api is the single source for the per-key rate limit.
+// updatePerKeySimpleRateLimitLocked is only called when the user data changes.
+func (r *rateLimitManager) updatePerKeySimpleRateLimitLocked(burstDuration time.Duration) {
+	if r.fairnessKeyRateLimitDefault == nil {
+		return
+	}
+	if r.fairnessKeyRateLimitDefault == nil {
+		r.clearPerKeyRateLimitsLocked()
+		return
+	}
```
dnr: ?  this looks like it's not going to work,

do we have test cases for clearing the whole queue limit from the config, and clearing the per-key limit?

service/matching/ratelimit_manager.go
```diff
@@ -48,17 +82,38 @@ func newRateLimitManager(userDataManager userDataManager,
...
-	if nPartitions <= 0 {
-		// Defaulting to 1 partition if misconfigured
-		nPartitions = 1
-	}
-	return nPartitions
+func (r *rateLimitManager) setAdminNsRate(rps float64) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.adminNsRate = rps
+	r.computeEffectiveRPSAndSourceLocked()
```
dnr: what updates the actual limiters when this changes?

service/matching/matcher_data.go
sivagirish81: Done.

service/matching/matching_engine_test.go
sivagirish81: So since this is a unit test and I don't have a dynamic config client. I wanted to override the value and make it reflect in the rateLimitManager considering it is using subscriptions now.

service/matching/pri_matcher.go
sivagirish81: Removed them.

service/matching/pri_matcher.go
sivagirish81: Noted and implemented.

service/matching/pri_matcher.go
sivagirish81: Removed the Rate() function from both the taskMatcher and the PriMatcher and updated the interface to not have a rate function as it wasn't being used anywhere else.

service/matching/physical_task_queue_manager.go
sivagirish81: This is not required. Removed the check.

service/matching/ratelimit_manager.go
sivagirish81: Removed it and used GetEffectiveRPSAndSource to get the rps.

service/matching/ratelimit_manager.go
sivagirish81: Implemented.

service/matching/matcher_data.go
```diff
@@ -277,58 +221,16 @@ type matcherData struct {
...
-			updates[e.Key().(string)] = clipped
-		}
-	}
-	it.Close()
-
-	// This messes up the LRU order, but we can't avoid that here without adding new
-	// functionality to Cache.
-	for key, clipped := range updates {
-		d.tasks.perKeyReady.Put(key, clipped)
+		rateLimitManager: rateLimitManager,
```
dnr: put these in the same order as the field declarations

service/matching/matcher_data.go
```diff
@@ -702,6 +604,10 @@ func makeSimpleLimiterParams(rate float64, burstDuration time.Duration) simpleLi
 func (p simpleLimiterParams) never() bool   { return p.interval < 0 }
 func (p simpleLimiterParams) limited() bool { return p.interval > 0 }
 
+func (p simpleLimiterParams) Equals(other simpleLimiterParams) bool {
```
dnr: we don't need a whole method for this, we can just use `==` on the structs

service/matching/pri_matcher.go
```diff
@@ -33,20 +33,17 @@ type priTaskMatcher struct {
...
+	rateLimitManager *rateLimitManager
+	metricsHandler   metrics.Handler // namespace metric scope
+	logger           log.Logger
+	numPartitions    func() int // number of task queue partitions
+	markAlive        func()     // function to mark the physical task queue alive
 
 	limiterLock sync.Mutex
-	adminNsRate float64
-	adminTqRate float64
 	dynamicRate float64
```
dnr: can't these be removed too?

service/matching/ratelimit_manager.go
```diff
@@ -3,33 +3,66 @@ package matching
...
+		taskQueueType   enumspb.TaskQueueType // Task queue type
+
+		effectiveRPS float64 // Min of api/worker set RPS and system defaults. Always reflects the overall task queue RPS.
+		systemRPS    float64 // Min of partition level dispatch rates times the number of read partitions.
+
+		// Sources of the effective RPS.
+		workerRPS                   *float64                // RPS set by worker at the time of polling, if available.
+		apiConfigRPS                *float64                // RPS set via API, if available.
+		fairnessKeyRateLimitDefault *float64                // fairnessKeyRateLimitDefault set via API, if available
+		rateLimitSource             enumspb.RateLimitSource // Source of the rate limit, can be set via API, worker or system default.
```
dnr: this is a derived value, not a source. it should be with effectiveRPS and systemRPS, which are also derived values (and give them a little comment heading). actually probably listing sources first and then derived makes more sense.

service/matching/ratelimit_manager.go
```diff
@@ -3,33 +3,66 @@ package matching
...
+
+		timeSource        clock.TimeSource
+		adminNsRate       float64
+		adminTqRate       float64
+		numReadPartitions int
```
dnr: these are sources, they should be with the other sources

service/matching/ratelimit_manager.go
```diff
@@ -3,33 +3,66 @@ package matching
...
+		effectiveRPS float64 // Min of api/worker set RPS and system defaults. Always reflects the overall task queue RPS.
+		systemRPS    float64 // Min of partition level dispatch rates times the number of read partitions.
+
+		// Sources of the effective RPS.
+		workerRPS                   *float64                // RPS set by worker at the time of polling, if available.
+		apiConfigRPS                *float64                // RPS set via API, if available.
+		fairnessKeyRateLimitDefault *float64                // fairnessKeyRateLimitDefault set via API, if available
+		rateLimitSource             enumspb.RateLimitSource // Source of the rate limit, can be set via API, worker or system default.
+
+		timeSource        clock.TimeSource
```
dnr: this is a constant dependency, put it with the other dependencies, not with this state

service/matching/ratelimit_manager.go
```diff
@@ -101,8 +157,22 @@ func (r *rateLimitManager) computeEffectiveRPSAndSourceLocked() {
 	}
 }
 
+func (r *rateLimitManager) computeAndApplyRateLimitLocked() {
+	oldRPS := r.effectiveRPS / float64(r.numReadPartitions)
+	r.computeEffectiveRPSAndSourceLocked()
+	newRPS := r.effectiveRPS / float64(r.numReadPartitions)
```
dnr: you can leave off the division here since you're just comparing them for equality
```suggestion
	oldRPS := r.effectiveRPS
	r.computeEffectiveRPSAndSourceLocked()
	newRPS := r.effectiveRPS
```

service/matching/ratelimit_manager.go
```diff
@@ -138,51 +205,168 @@ func (r *rateLimitManager) GetRateLimiter() quotas.RateLimiter {
...
+	r.dynamicRateLimiter.Refresh()
+}
+
+// UpdateSimpleRateLimit updates the overall queue rate limits for the simpleRateLimiter implementation
+func (r *rateLimitManager) updateSimpleRateLimitLocked(burstDuration time.Duration) {
 	newRPS := r.effectiveRPS
-	if oldRPS == newRPS {
-		// No update required
+	// Always update the rate limit when called, even if RPS hasn't changed
+	// This ensures that the burst duration is properly applied
```
dnr: this comment doesn't make sense? this is only called after checking the rps has changed

service/matching/ratelimit_manager.go
```diff
@@ -138,51 +205,168 @@ func (r *rateLimitManager) GetRateLimiter() quotas.RateLimiter {
...
+	r.perKeyLimit = simpleLimiterParams{}
+}
+
+func (r *rateLimitManager) readyTimeForTask(task *internalTask) simpleLimiter {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.readyTimeForTaskLocked(task)
+}
+
+func (r *rateLimitManager) readyTimeForTaskLocked(task *internalTask) simpleLimiter {
```
dnr:
```suggestion

```
this is just called in one place, right?

service/matching/ratelimit_manager.go
```diff
@@ -138,51 +205,168 @@ func (r *rateLimitManager) GetRateLimiter() quotas.RateLimiter {
...
+	return ready
+}
+
+func (r *rateLimitManager) consumeTokens(now int64, task *internalTask, tokens int64) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.consumeTokensLocked(now, task, tokens)
+}
+
+func (r *rateLimitManager) consumeTokensLocked(now int64, task *internalTask, tokens int64) {
```
dnr:
```suggestion

```
same here?

service/matching/ratelimit_manager_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (s *RateLimitManagerSuite) TestUpdatePerKeySimpleRateLimitLocked_WhenFairnessKeyRateLimitDefaultIsNil() {
+	mockUserDataManager := &mockUserDataManager{}
+	config := newTaskQueueConfig(
+		tqid.UnsafeTaskQueueFamily("test-namespace", "test-task-queue").TaskQueue(enumspb.TASK_QUEUE_TYPE_ACTIVITY),
+		NewConfig(dynamicconfig.NewNoopCollection()),
+		"test-namespace",
+	)
+	rateLimitManager := newRateLimitManager(mockUserDataManager, config, enumspb.TASK_QUEUE_TYPE_ACTIVITY)
+	rateLimitManager.mu.Lock()
```
dnr, line 35: when writing unit tests, it's best to interact with just the public interface of the code instead of its implementation details. that makes the tests less fragile and less likely to break or rot as the implementation changes. sometimes it's unavoidable though. in this case it seems like it should be possible to do this test with just the public interface?

service/matching/ratelimit_manager_test.go
```diff
@@ -0,0 +1,83 @@
...
+	s.True(rateLimitManager.perKeyLimit.limited())
+	// Update the per-key simple rate limit with fairnessKeyRateLimitDefault as nil
+	rateLimitManager.updatePerKeySimpleRateLimitLocked(time.Second)
+	// Verify that clearPerKeyRateLimitsLocked was called
+	// The cache should be replaced with a new empty cache
+	s.Equal(0, rateLimitManager.perKeyReady.Size(), "All per-key ready entries should be cleared")
+	s.False(rateLimitManager.perKeyLimit.limited(), "Per-key limit should be cleared")
+	rateLimitManager.mu.Unlock()
+}
+
```
dnr, line 57:
```suggestion

// Additions to rateLimitManager for use by other unit tests:
```

service/matching/ratelimit_manager_test.go
```diff
@@ -0,0 +1,83 @@
...
+func (r *rateLimitManager) UpdateSimpleRateLimit(burstDuration time.Duration) {
```
dnr: I like the `ForTesting` suffix to make it clear that this isn't the normal interface, it's for tests only:

```suggestion
func (r *rateLimitManager) UpdateSimpleRateLimitForTesting(burstDuration time.Duration) {
```

tests/task_queue_test.go
```diff
@@ -199,75 +278,182 @@ func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
...
 	wfWorker := worker.New(s.SdkClient(), tv.TaskQueue().GetName(), worker.Options{})
 	wfWorker.RegisterWorkflow(workflowFn)
 	s.NoError(wfWorker.Start())
 	defer wfWorker.Stop()
 
-	// Launch workflows
-	for i := range taskCount {
-		_, err := s.SdkClient().ExecuteWorkflow(ctx, sdkclient.StartWorkflowOptions{
+	// Launch workflows under workerSetRPS (2 RPS)
+	for i := 0; i < taskCount; i++ {
```
```suggestion
	for i := range taskCount {
```
:(

tests/task_queue_test.go
```diff
@@ -199,75 +278,182 @@ func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
...
+		TaskQueueType: enumspb.TASK_QUEUE_TYPE_ACTIVITY,
+		UpdateQueueRateLimit: &workflowservice.UpdateTaskQueueConfigRequest_RateLimitUpdate{
+			RateLimit: &taskqueuepb.RateLimit{RequestsPerSecond: float32(apiSetRPS)},
+			Reason:    "test api override",
+		},
+	})
+	s.NoError(err)
+
+	// Launch workflows under API override (1 RPS)
+	for i := 0; i < taskCount; i++ {
```
dnr: range

service/matching/ratelimit_manager.go
```diff
@@ -114,9 +186,8 @@ func (r *rateLimitManager) InjectWorkerRPS(meta *pollMetadata) {
 		}
 	}
 	r.workerRPS = rps
-	// updateRatelimitLocked includes internal logic to determine if an update is needed,
-	// so calling it unconditionally is safe and avoids redundant updates.
-	r.updateRatelimitLocked()
+	r.trySetRPSFromUserDataLocked()
```
sivagirish81: Debug : TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI [ Fails consistently in the SQLite CI tests ]

Test Context : 

- 30 activity tasks first run at 10 RPS workerSetRateLimit.
- Activity tasks completes and total execution time is recorded.
- UpdateTaskQueueConfig api is called and it updates the ApiSetRPS to 5 RPS.
- 30 activity tasks are again enqueued into the same activity task queue.

```
=== Failed
=== FAIL: tests TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI (30.24s)
    functional_test_base.go:343: Running TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI in test shard 1/3
    task_queue_test.go:456: 
        	Error Trace:	/home/runner/work/temporal/temporal/tests/task_queue_test.go:456
        	Error:      	"11.091380681s" is not greater than "21.80354251s"
        	Test:       	TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI
        	Messages:   	API override did not reduce throughput as expected
    functional_test_base.go:403: wrote OTEL traces to /home/runner/work/temporal/temporal/.testoutput/traces.TestTaskQueueSuite-TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI_1754599420.json
```

Expected behavior : First 30 tasks should execute in ~3s and next 30 tasks ~ 6s.
Observed behavior : First 30 tasks should execute in ~11 and next 30 tasks ~ 11. Concerning why the tasks took 11 seconds to execute. It is pretty high. But it is CI and everything is considerably slower. 

Previously when I did not try to pull the latest user data at the time of polling in the sqlite instances I was consistently observing the behavior where both set of tasks execute in a similar times.
After the change : 
```
=== Failed
=== FAIL: tests TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI (22.53s)
    functional_test_base.go:343: Running TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI in test shard 1/3
    task_queue_test.go:459: Rate limit set in Persistence: 5
    task_queue_test.go:479: Completion time for First Activity tasks: 6.261844946s, Second Activity tasks: 9.724101139s
    task_queue_test.go:481:
        	Error Trace:	/home/runner/work/temporal/temporal/tests/task_queue_test.go:481
        	Error:      	"9.724101139s" is not greater than "10.523689892s"
        	Test:       	TestTaskQueueSuite/TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI
        	Messages:   	API override did not reduce throughput as expected
    functional_test_base.go:403: wrote OTEL traces to /home/runner/work/temporal/temporal/.testoutput/traces.TestTaskQueueSuite-TestTaskQueueRateLimit_UpdateFromWorkerConfigAndAPI_1754616100.json
```
This is the debug trace I got afterwards. Then I relaxed the boundaries to 1.5 times the first set of tasks run time with an additional buffer of 2s. Then the test stopped flaking and got fixed.

Will need to see why this is specifically happening for SQLlite.

service/matching/ratelimit_manager.go
sivagirish81: Fixed this to have a central comuteAndApplyRateLimit() function that handles the UpdateRateLimit calls including the check for changes if any.

service/matching/ratelimit_manager.go
sivagirish81: Added a unit test to check this. and fixed the clearPerKeyRateLimitsLocked()

service/matching/ratelimit_manager.go
sivagirish81: Fixed this to call computeAndApplyLocked() in the case of all updates.

service/matching/matcher_data.go
sivagirish81: Noted and implemented.

service/matching/matcher_data.go
sivagirish81: Fixed to use == and removed the method.

service/matching/pri_matcher.go
sivagirish81: Removed limiter lock and dynamicRate as they are no longer being used.

service/matching/ratelimit_manager.go
sivagirish81: Restructured as suggested.

service/matching/ratelimit_manager.go
sivagirish81: Noted and implemented.

service/matching/ratelimit_manager.go
sivagirish81: Done.

service/matching/ratelimit_manager.go
sivagirish81: Removed the division as suggested. 

But now that I think of it. We have a problem in the case of readPartitions. 
It will not work as of now I feel because in the computeAndApplyRateLimitLocked we are checking for changes in the value of  effective RPS -  even if we divide by the partition count we will not be able to catch the change as the updated value for read partitions will be used to compute the old and new RPS signaling no change. 

Need to change the logic to apply rate limits in the case of changes to read Partitions. Will change the logic to track changes to the readPartitions and accordingly update the rate limit.

service/matching/ratelimit_manager.go
sivagirish81: Removed this comment.

tests/task_queue_test.go
sivagirish81: Noted and implemented. Will watch out for these more carefully in the future.

service/matching/ratelimit_manager_test.go
sivagirish81: Yes. I don't think this should be written as a unit test. I think it is better if it is a functional test where the updateTaskQueueConfig resets the value of fairnessKeyRateLimitDefault to nil and we can check the behaviour. Will change this to a functional test soon.

service/matching/pri_matcher.go
```diff
@@ -98,25 +81,22 @@ func newPriTaskMatcher(
...
+		data:             newMatcherData(config, logger, clock.NewRealTimeSource(), fwdr != nil, rateLimitManager),
+		tqCtx:            tqCtx,
+		logger:           logger,
+		metricsHandler:   metricsHandler,
+		partition:        partition,
+		fwdr:             fwdr,
+		validator:        validator,
+		numPartitions:    config.NumReadPartitions,
+		markAlive:        markAlive,
+		rateLimitManager: rateLimitManager,
```
sivagirish81: Will move this right after validator.

service/matching/ratelimit_manager.go
```diff
@@ -15,29 +15,31 @@ type (
...
-		// Sources of the effective RPS.
-		workerRPS                   *float64                // RPS set by worker at the time of polling, if available.
-		apiConfigRPS                *float64                // RPS set via API, if available.
-		fairnessKeyRateLimitDefault *float64                // fairnessKeyRateLimitDefault set via API, if available
-		rateLimitSource             enumspb.RateLimitSource // Source of the rate limit, can be set via API, worker or system default.
-
-		timeSource        clock.TimeSource
-		adminNsRate       float64
-		adminTqRate       float64
 		numReadPartitions int
```
dnr: this is a "source" too (the derived values can change when it changes)

service/matching/ratelimit_manager.go
```diff
@@ -15,29 +15,31 @@ type (
...
+		effectiveRPS    float64                 // Min of api/worker set RPS and system defaults. Always reflects the overall task queue RPS.
+		systemRPS       float64                 // Min of partition level dispatch rates times the number of read partitions.
+		rateLimitSource enumspb.RateLimitSource // Source of the rate limit, can be set via API, worker or system default.
+		// Derived from the `defaultTaskDispatchRPS`.
+		// dynamicRateBurst is the dynamic rate & burst for rate limiter
 		dynamicRateBurst quotas.MutableRateBurst
 		// dynamicRateLimiter is the dynamic rate limiter that can be used to force refresh on new rates.
 		dynamicRateLimiter *quotas.DynamicRateLimiterImpl
 
+		// Dependencies
```
dnr: dependencies should be first in the struct

service/matching/ratelimit_manager.go
dnr: right, I think we need the partition count to be baked into r.effectiveRPS already. so it should track the per-partition rate, and then we can multiply it when reporting the overall rate. or else maybe track both numbers?

sivagirish81:
Will revert to subscriptions for NumReadPartitions once the Dynamic config conversion is updated 
(https://github.com/temporalio/temporal/pull/8180)  to allow subscriptions for TypedConstrainedDefaultSettings.

**Update** : Reverted to use subscriptions.

service/matching/ratelimit_manager.go
sivagirish81: Noted and implemented.

service/matching/ratelimit_manager.go
sivagirish81: **Update : removed the force reload of the config from persistence at the time of InjectWorkerRPS.**
On SQLite I consistently saw that drain times at 10 rps and 5 rps were nearly identical—and occasionally 5 rps was faster. This is possible considering: SQLite has a single-writer concurrency model, and starting/completing activity tasks triggers several small writes (Matching + History). 
For the sake of this test, to test the rate limit without the impact of the db : 
+ I added a sleep inside the activity, so activities don’t complete instantly and create a string of short spaced writes.
Post this the tests started to pass without flaking.

service/matching/ratelimit_manager.go
sivagirish81: **Fix : EffectiveRPS is always maintained partition wise including fairnessKeyRateLimitDefault.
In the getEffectiveRPSAndSource I multiply the effectiveRPS by the number of partitions and return the value.**
This will effectively be able to handle changes in effectiveRPS/fairnessKeyRateLimitDefault cause due to changes in numReadPartitions.

tests/task_queue_test.go
```diff
@@ -206,6 +211,7 @@ func (s *TaskQueueSuite) configureRateLimitAndLaunchWorkflows(
 		defer wg.Done()
 		mu.Lock()
 		*runTimes = append(*runTimes, time.Now())
+		time.Sleep(250 * time.Millisecond) //nolint
```
dnr, line 214: maybe a better way to do this is to set `WorkerActivitiesPerSecond: 4` in worker options?

service/matching/matcher_data_test.go
```diff
@@ -311,8 +312,8 @@ func (s *MatcherDataSuite) TestTaskForward() {
 }
 
 func (s *MatcherDataSuite) TestRateLimitedBacklog() {
-	// 10 tasks/sec with burst of 3
-	s.md.UpdateRateLimit(10, 300*time.Millisecond)
+	s.md.rateLimitManager.SetEffectiveRPSAndSourceForTesting(10.0, enumspb.RATE_LIMIT_SOURCE_API)
+	s.md.rateLimitManager.UpdateSimpleRateLimitForTesting(300 * time.Millisecond)
```
dnr, line 316: slightly weird that this is called UpdateSimpleRateLimitForTesting but the argument is a burst duration, not a rate limit. I see why, it just reads funny. maybe we can adjust the names later.

tests/task_queue_test.go
```diff
@@ -199,75 +285,407 @@ func (s *TaskQueueSuite) TestTaskQueueAPIRateLimitOverridesWorkerLimit() {
...
+	s.OverrideDynamicConfig(dynamicconfig.MatchingEnableFairness, true)
+	// Single partition for simplicity.
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
+	// Fast refresh so limiter state picks up config quickly.
+	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Millisecond)
+
+	// generate a unique base so IDs don't collide across shards/tests
+	base := uuid.NewString()
+	parsePattern := fmt.Sprintf("perkey-wf-%s-%%d", base) // for Sscanf later
```
dnr, line 515: we have the testvars package for stuff like this

dnr (approved):
I added a few more little nitpicks but I think this is good to go! we can address that stuff in followup PRs

tests/task_queue_test.go
sivagirish81: Yes. Will try to incorporate this in the next PR.

service/matching/matcher_data_test.go
sivagirish81: Will improve the naming in the next PR.

tests/task_queue_test.go
sivagirish81: Will change this to use the testvars package in the next PR.

sivagirish81:
Noticed an issue in the CI run.
=== Failed
=== FAIL: tests TestTaskQueueSuite/TestWholeQueueLimit_TighterThanPerKeyDefault_IsEnforced (1.39s)
    functional_test_base.go:343: Running TestTaskQueueSuite/TestWholeQueueLimit_TighterThanPerKeyDefault_IsEnforced in test shard 2/3
    task_queue_test.go:609: 
        	Error Trace:	/home/runner/work/temporal/temporal/tests/task_queue_test.go:609
        	            				/home/runner/work/temporal/temporal/tests/task_queue_test.go:698
        	Error:      	Received unexpected error:
        	            	Workflow execution is already running. WorkflowId: perkey-wf-0, RunId: 0198f7c2-7477-7e97-9246-4dc16e6bd9da.
        	Test:       	TestTaskQueueSuite/TestWholeQueueLimit_TighterThanPerKeyDefault_IsEnforced
    functional_test_base.go:403: wrote OTEL traces to /home/runner/work/temporal/temporal/.testoutput/traces.TestTaskQueueSuite-TestWholeQueueLimit_TighterThanPerKeyDefault_IsEnforced_1756503397.json
    
Will debug this but looks like the workflow ID is being reused and that's why the CI is failing. I will try to raise a follow up PR for this.

PR #8138, opened 2025-08-01
Respect priority levels in rescheduler

prathyushpv:
## What changed?
Right now reshceulder picks task queues randomly in the reschedule loop. This can make it send low priority tasks before sending high priority tasks. It has to process high priority tasks before processing low priority tasks. This change introduces a btree of taskQueueKeys. Tasks queues are sorted in the order of priority inside this btree. This makes it easy for reschduler to iterate in the order of decreasing priority.

## Why?
Better priority enforcement at task scheduler.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,6 +80,8 @@ require (
 	modernc.org/sqlite v1.34.1
 )
 
+require github.com/google/btree v1.1.3
+
 require (
 	cel.dev/expr v0.20.0 // indirect
 	cloud.google.com/go v0.118.3 // indirect; indirect e
--- a/service/history/queues/rescheduler.go
+++ b/service/history/queues/rescheduler.go
@@ -7,6 +7,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/google/btree"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/backoff"
 	"go.temporal.io/server/common/clock"
@@ -66,6 +67,7 @@ type (
 
 		sync.Mutex
 		pqMap          map[TaskChannelKey]collection.Queue[rescheduledExecuable]
+		keyBTree       *btree.BTree // Maintains TaskChannelKeys in priority order
 		numExecutables int
 	}
 )
@@ -88,7 +90,8 @@ func NewRescheduler(
 		timerGate:        timer.NewLocalGate(timeSource),
 		taskChannelKeyFn: scheduler.TaskChannelKeyFn(),
 
-		pqMap: make(map[TaskChannelKey]collection.Queue[rescheduledExecuable]),
+		pqMap:    make(map[TaskChannelKey]collection.Queue[rescheduledExecuable]),
+		keyBTree: btree.New(2), // Degree 2 (minimum degree for B-tree)
 	}
 }
 
@@ -145,11 +148,15 @@ func (r *reschedulerImpl) Reschedule(
 
 	now := r.timeSource.Now()
 	updatedRescheduleTime := false
-	for key, pq := range r.pqMap {
+
+	// Process keys in decreasing order of priority
+	r.keyBTree.Ascend(func(item btree.Item) bool {
+		key := item.(TaskChannelKey)
 		if key.NamespaceID != namespaceID {
-			continue
+			return true // continue iteration
 		}
 
+		pq := r.pqMap[key]
 		updatedRescheduleTime = true
 		// set reschedule time for all tasks in this pq to be now
 		items := make([]rescheduledExecuable, 0, pq.Len())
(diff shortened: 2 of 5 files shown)


service/history/queues/rescheduler.go
```diff
@@ -279,20 +294,21 @@ func (r *reschedulerImpl) getOrCreatePQLocked(
...
+		return collection.NewPriorityQueue(r.rescheduledExecutableCompareLess)
 	}
 
-	return collection.NewPriorityQueueWithItems(r.rescheduledExecuableCompareLess, items)
+	return collection.NewPriorityQueueWithItems(r.rescheduledExecutableCompareLess, items)
 }
 
-func (r *reschedulerImpl) rescheduledExecuableCompareLess(
+func (r *reschedulerImpl) rescheduledExecutableCompareLess(
 	this rescheduledExecuable,
```
yycptt: Can you plz help fix the typo in the struct name as well 🙏

service/history/queues/rescheduler.go
```diff
@@ -66,6 +67,7 @@ type (
 
 		sync.Mutex
 		pqMap          map[TaskChannelKey]collection.Queue[rescheduledExecuable]
+		keyBTree       *btree.BTree // Maintains TaskChannelKeys in priority order
```
yycptt: I think there's a generic version of the bTree and we don't have to do type assertion below.

service/history/queues/scheduler.go
```diff
@@ -257,3 +258,12 @@ func (s *rateLimitedSchedulerImpl) Stop() {
 func (s *rateLimitedSchedulerImpl) TaskChannelKeyFn() TaskChannelKeyFn {
 	return s.baseScheduler.TaskChannelKeyFn()
 }
+
+// Less implements btree.Item interface for TaskChannelKey
+func (t TaskChannelKey) Less(than btree.Item) bool {
+	otherKey := than.(TaskChannelKey)
+	if t.Priority == otherKey.Priority {
+		return t.NamespaceID < otherKey.NamespaceID
```
yycptt: hmm this creates an order across namespaces right? If we have a host level limit on action/operation, then some namespaces may not to be able to process their tasks?

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/go-sql-driver/mysql v1.9.0
 	github.com/gocql/gocql v1.7.0
 	github.com/golang-jwt/jwt/v4 v4.5.2
+	github.com/google/btree v1.1.3
```
dnr: Can we use https://github.com/tidwall/btree instead? I did a little research on btree packages in go and settled on that one for some upcoming changes I'm going to do for matching fairness. It's a fork of google's but with a bunch of performance and other improvements. It should be pretty easy to convert.

(The cds codebase uses both packages, amusingly, but I think it might be nice if we settle on one)

prathyushpv: Thanks for that info! Let me change the code to use https://github.com/tidwall/btree

service/history/queues/scheduler.go
prathyushpv: Right. I will ignore namespace name in this comparison.

service/history/queues/rescheduler.go
prathyushpv: Done

service/history/queues/rescheduler.go
prathyushpv: Yes. made the change.

service/history/queues/scheduler.go
prathyushpv: Done.

go.mod
prathyushpv: I have made the change to use https://github.com/tidwall/btree

service/history/queues/scheduler.go
```diff
@@ -257,3 +257,13 @@ func (s *rateLimitedSchedulerImpl) Stop() {
...
 }
+
+// Less implements comparison for TaskChannelKey for btree
+// Priority is the primary ordering, NamespaceID is the tie-breaker to ensure unique keys
+func (t TaskChannelKey) Less(other TaskChannelKey) bool {
+	if t.Priority != other.Priority {
+		return t.Priority < other.Priority
+	}
+	// Use NamespaceID as tie-breaker to ensure all keys are distinguishable
+	return t.NamespaceID < other.NamespaceID
```
yycptt: Let's probably sync offline on this. I still think this will cause starvation for customers with "bigger" namespaceID when task scheduler encounters host/cell level limit. 

Task scheduler rate limiter are all namespace level today so we are fine from that perspective. 
But the task processing worker pool size may become the issue especially for Visibility queue where each task can take a long time when ES is slow. Then only tasks with small Namespace can be processed.

prathyushpv: Sorry missed this in the previous commit. Fixed this. Made a change to just check the priority value here.

yycptt (approved):

prathyushpv:
I just realized we need a strict total ordering on the B-tree key. So we cannot ignore the namespace ID. Task keys can go missing if there is no strict total ordering. Let me see how I can fix this.

PR #8139, opened 2025-08-01
[CHASM] NewEntity support for deferred pointer resolution

lina-temporal:
## What changed?
- Previously when creating pointers within a CHASM tree, the pointed-to component had to already exist within the tree, as the pointer was resolved immediately. This was cumbersome in some common cases like setting up an initial tree of CHASM components (such as setting up a scheduler tree), where ideally, all components would be created and linked up within the first transaction.
- A new field type, `fieldTypeDeferredPointer`, has been introduced. This field stores the pointer value directly, and resolves the value to the path of the component within the tree during `CloseTransaction` (or fails if the pointer is dangling).

## How did you test it?
- [ ] built
- [x] added new unit test(s)


chasm/tree.go
```diff
@@ -1540,6 +1548,47 @@ func (n *Node) closeTransactionGeneratePhysicalPureTask() error {
 	return nil
 }
 
+// resolveDeferredPointers resolves all deferred pointers in the tree.
+// Returns error if any deferred pointer cannot be resolved, as deferred pointers
+// cannot be persisted after transaction close.
+func (n *Node) resolveDeferredPointers() error {
+	for _, node := range n.andAllChildren() {
```
yycptt, line 1573: not sure I follow, it looks like this can only resolve deferred pointers for components that already exists? I guess otherwise `node.value` will be empty till `syncSubComponents()` is called?

Root is kinda special because we manually assign the `value` field to bootstrap the entire process.

chasm/tree.go
```diff
@@ -829,7 +833,7 @@ func (n *Node) syncSubField(fieldV reflect.Value, fieldN string, nodePath []stri
 				err = serviceerror.NewInternalf("value must be of type []string for the field of pointer type: got %T", internal.value())
 				return
 			}
-		case fieldTypeData, fieldTypeComponent:
+		case fieldTypeData, fieldTypeComponent, fieldTypeDeferredPointer:
```
yycptt: fieldTypeDeferredPointer is skipped above?

chasm/field_test.go
```diff
@@ -148,3 +149,192 @@ func (s *fieldSuite) newTestTree(
...
+
+	s.Equal(fieldTypePointer, rootComponent.SubComponent11Pointer.Internal.fieldType())
+	s.Equal(fieldTypeDeferredPointer, rootComponent.SubComponent11Pointer2.Internal.fieldType())
+
+	_, err = rootNode.CloseTransaction()
+	s.NoError(err)
+
+	// Ensure both pointers have been resolved.
+	s.Equal(fieldTypePointer, rootComponent.SubComponent11Pointer.Internal.fieldType())
+	s.Equal(fieldTypePointer, rootComponent.SubComponent11Pointer2.Internal.fieldType())
```
yycptt: hmm what's the difference between those two pointers and make them `Mixed`?

chasm/field.go
```diff
@@ -57,7 +57,11 @@ func ComponentPointerTo[C Component](
 ) (Field[C], error) {
 	path, err := ctx.componentNodePath(c)
 	if err != nil {
-		return NewEmptyField[C](), err
+		// If we can't resolve the path (e.g., during NewEntity transition),
+		// store the component directly for deferred resolution
+		return Field[C]{
+			Internal: newFieldInternalWithValue(fieldTypeDeferredPointer, c),
+		}, nil
```
yycptt: Can we extend this approach to DataPointer below as well? I feel they should be able to share the same logic or I missed something?

chasm/tree.go
yycptt: I feel we need to call this after `syncSubComponents()` but before `closeTransactionSerializeNodes`. Also `initSerializedNode` for deferred pointers either here or in `syncSubComponents`.

chasm/field.go
```diff
@@ -57,7 +57,11 @@ func ComponentPointerTo[C Component](
 ) (Field[C], error) {
 	path, err := ctx.componentNodePath(c)
```
yycptt: Shall we always go with the deferred pointer approach? Then this function doesn't even need return a error.

chasm/field_test.go
lina-temporal: It's that one component has already been resolved from a previous `CloseTransaction`, and one's still deferred. I found this more convincing as a test since it shows that `CloseTransaction` works on a mix of unresolved and resolved pointers.

chasm/tree.go
lina-temporal: Thanks, will remove deferred from this case.

chasm/field.go
lina-temporal: Yeah, it could be used for data pointers as well, I'll expand it to support those too.

chasm/field.go
lina-temporal: Yep, I like that idea!

chasm/tree.go
lina-temporal: > not sure I follow, it looks like this can only resolve deferred pointers for components that already exists? I guess otherwise node.value will be empty till syncSubComponents() is called?

Ah, yeah, this is since I was testing against pointer fields on root (to newly-added child components further in the tree); the components pointed to were new in the tree, but the component with the pointer field was pre-existing. I'll update the tests as well.

chasm/fields_iterator.go
```diff
@@ -40,10 +40,10 @@ func fieldsOf(valueV reflect.Value) iter.Seq[fieldInfo] {
 		for i := 0; i < valueT.Elem().NumField(); i++ {
 			fieldV := valueV.Elem().Field(i)
 			fieldT := fieldV.Type()
+			fieldN := fieldName(valueT.Elem().Field(i))
 			if fieldT == UnimplementedComponentT {
 				continue
 			}
```
yycptt: nit: is this just for grouping variables together?

chasm/tree.go
```diff
@@ -1096,9 +1099,13 @@ func (n *Node) componentNodePath(
...
 	}
 
-	// It's uncessary to deserialize entire tree as calling this method means
+	// It's unnecessary to deserialize entire tree as calling this method means
 	// caller already have the deserialized value.
 	for path, node := range n.andAllChildren() {
+		if node.fieldType() != fieldTypeComponent {
+			continue
+		}
+
```
yycptt, line 1122: Can you help remove this method (`componentNodePath`) from the chasm Context interface, now that we always use DeferredPointer when creating a pointer? Same for `dataNodePath` below.

And I think we should remove the syncSubComponents() step in those two methods.

chasm/tree.go
```diff
@@ -833,6 +834,8 @@ func (n *Node) syncSubField(fieldV reflect.Value, fieldN string, nodePath []stri
 			if err = assertStructPointer(reflect.TypeOf(internal.value())); err != nil {
 				return
 			}
+		case fieldTypeDeferredPointer:
+			// TODO - validate deferred pointer format here
```
yycptt: not sure what we can validate here... maybe say it the validation is basically the resolve pointer step later?

chasm/tree.go
```diff
@@ -1540,6 +1555,56 @@ func (n *Node) closeTransactionGeneratePhysicalPureTask() error {
...
+			if internal.fieldType() == fieldTypeDeferredPointer && internal.value() != nil {
+				// Must resolve the deferred pointer or fail the transaction.
+				var resolvedPath []string
+				var err error
+
+				switch value := internal.value().(type) {
+				case Component:
+					resolvedPath, err = n.componentNodePath(value)
+				case proto.Message:
+					resolvedPath, err = n.dataNodePath(value)
```
yycptt, line 1599: a bit expensive but should be fine I guess, we can optimize later by tracking all deserialized values if necessary.

chasm/tree.go
```diff
@@ -1160,6 +1171,10 @@ func (n *Node) CloseTransaction() (NodesMutation, error) {
 		return NodesMutation{}, err
 	}
 
+	if err := n.resolveDeferredPointers(); err != nil {
```
yycptt: minor: is it possible to only run this step conditionally if we do find a deferred pointer in syncSubComponents?

chasm/field_internal.go
```diff
@@ -28,14 +28,18 @@ func (fi fieldInternal) isEmpty() bool {
 }
 
 func (fi fieldInternal) value() any {
-	if fi.node == nil {
+	// Deferred pointers are special-cased, since their serialized nodes are
+	// initialized as regular persistable pointers.
+	if fi.node == nil || fi.ft == fieldTypeDeferredPointer {
```
yycptt: nit: you can add comment re. when deferredPointers will have non-nil `fi.node`? Took me some time to realize that's only temporarily after syncSubField and before resolving the deferred pointers

chasm/field_test.go
```diff
@@ -148,3 +149,204 @@ func (s *fieldSuite) newTestTree(
...
+	resolvedComponent, err := sc1.SubComponent2Pointer.Get(ctx)
+	s.NoError(err)
+	s.Equal(sc2, resolvedComponent)
+
+	// TODO - this doesn't resolve, but I've manually verified the tree structure looks correct
+	// TODO
+	// TODO
+	// resolvedData, err := sc1.DataPointer.Get(ctx)
+	// s.NoError(err)
+	// s.Equal(sc2.SubComponent2Data, resolvedData)
```
yycptt, line 258: yeah that indeed looks like a bug to me in Field.Get() impl, should just use `referencedNode` instead of `f.Internal.node`.

Issue #8098, opened 2025-07-25
Remove waits before fetching pages in batch activities


spkane31:
> We should eventually fix (not in this PR) this but there's no reason to wait before fetching the next page if we have executors that are free to pick up work.

https://github.com/temporalio/temporal/blob/main/service/worker/batcher/activities.go#L132-L150

https://github.com/temporalio/temporal/pull/8081/files#r2226888751

PR #8144, opened 2025-08-04
improvement: remove waits before fetching activities

spkane31:
## What changed?
optimize the batch operation processing in `BatchActivity` and `BatchActivityWithProtobuf` by removing the need to wait for entire pages to complete before fetching the next page. 

- Implemented proactive page fetching once a worker becomes available
- common `processWorkflowsWithProactiveFetching` function to reduce code duplication

## Why?
The previous implementation had workers wait for entire pages to complete. This optimization improves resource utilization. The refactoring also eliminates duplicated functions in the `BatchParams` struct and `BatchOperation` protobuf.

Addresses issue #8098.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

The changes maintain backward compatibility.

## Potential risks
While this change improves performance, it does modify the concurrency model of batch processing:

1. **Timing changes**: The optimization changes when pages are fetched relative to task completion, which could expose edge cases in error handling or heartbeat timing
2. **Memory usage**: Pages may be fetched earlier, potentially increasing peak memory usage if the next page is large
3. **Rate limiting interaction**: The more aggressive task scheduling could interact differently with rate limiting, though the same per-worker limits are maintained
4. **Heartbeat behavior**: heartbeats track the progress of an entire page and are applied after an entire page finishes

The changes preserve all existing error handling, retry logic, and rate limiting behavior, but the different execution timing could surface previously hidden race conditions.


service/frontend/workflow_handler.go
```diff
@@ -4815,9 +4815,11 @@ func (wh *WorkflowHandler) ListBatchOperations(
 		Namespace:     request.GetNamespace(),
 		PageSize:      request.PageSize,
 		NextPageToken: request.GetNextPageToken(),
-		Query: fmt.Sprintf("%s = '%s' and %s = '%s'",
+		Query: fmt.Sprintf("(%s = '%s' or %s = '%s') and %s = '%s'",
```
bergundy: Lets just remove the workflow type filter and make sure this is merged ASAP because what we have in main is broken.

service/worker/batcher/activities.go
```diff
@@ -41,6 +41,196 @@ var (
...
+		currentPageToken = resp.NextPageToken
+		hasMorePages = len(currentPageToken) > 0
+		for _, wf := range resp.Executions {
+			executions = append(executions, wf.Execution)
+		}
+	} else {
+		hasMorePages = false
+	}
+
+	// Send initial tasks
```
bergundy: This seems redundant.
Use a buffered channel with `config.concurrency` as the capacity (ensure that it is `>=0` first).

bergundy (commented):
I would use a buffered channel with `config.concurrency` as the capacity (ensure that it is `>=0` first) and structure the code as an event loop (mostly pseudocode):

```
type Page struct {
	executions []*commonpb.WorkflowExecution
	submittedCount int
	successCount int
	errorCount int
	nextPageToken []byte
	wg sync.WaitGroup
	prev, next *Page
}

func (p *Page) hasNext() bool {
	return len(p.nextPageToken) > 0
}

func (p *Page) allSubmitted() bool {
	return p.submittedCount == len(p.Executions)
}

func (p *Page) nextTask() Task {
	return Task{} // TODO fill this in.
}

func (p *Page) done() bool {
	if p.prev != nil && !p.prev.done() {
		return false
	}
	return p.successCount + p.errorCount == len(p.Executions)
}

page := fetchPage() // fetch page of executions if needed here, or initialize from request.GetExecutions().

for {
	if page.hasNext() &&  page.allSubmitted() {
		next := fetchPage()	
		page.next, next.prev, page = next, page, next
	}
	select {
	case taskCh <- page.nextTask():
		page.submittedCount++
	case result := <- respCh:
		for curr := result.page; curr = curr.next; curr != nil {
			// increment stats on curr...
			// update heartbeat details if page and all previous pages are complete...
			// delete all prev page pointers for completed pages to avoid scanning the entire chain repeatedly...
		}
	}
	if page.done() {
		break
	}
}
```

I would also move the heartbeat to a separate goroutine to avoid hitting timeouts when the process slows down for any reason.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,258 @@ var (
...
+	frontendClient workflowservice.WorkflowServiceClient,
+	metricsHandler metrics.Handler,
+	logger log.Logger,
+)
+
+// Task represents a task to be processed (alias for taskDetail for consistency with pseudocode)
+type Task = taskDetail
+
+// Page represents a page of workflow executions to be processed
+type Page struct {
```
bergundy: I know I suggested this name but I realized we shouldn't export it:

```suggestion
type page struct {
```

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,258 @@ var (
...
+	respCh chan taskResponse,
+	rateLimiter *rate.Limiter,
+	sdkClient sdkclient.Client,
+	frontendClient workflowservice.WorkflowServiceClient,
+	metricsHandler metrics.Handler,
+	logger log.Logger,
+)
+
+// Task represents a task to be processed (alias for taskDetail for consistency with pseudocode)
+type Task = taskDetail
```
bergundy: Please don't export and also no need to alias this internal struct, just rename `taskDetail` to `task IMHO.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,258 @@ var (
...
+// nextTask returns the next task to be submitted from this page
+func (p *Page) nextTask(hbd HeartBeatDetails) Task {
+	if p.submittedCount >= len(p.executions) {
+		return Task{} // No more tasks in this page
+	}
+
+	task := Task{
+		execution:  p.executions[p.submittedCount],
+		attempts:   1,
+		hbd:        hbd,
```
bergundy: I think you don't need `hbd` on the task here or `pageNumber`, you can pass a reference to the `page` instead.

spkane31: good call, would not have seen that myself. I think this is a little cleaner here

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+
+	// New event loop using Page-based pager
+	for {
+		// Check if we need to fetch next page
+		if p.hasNext() && p.allSubmitted() {
+			nextPage, err := fetchPage(ctx, sdkClient, config, p.nextPageToken, p.pageNumber+1)
+			if err != nil {
+				metrics.BatcherOperationFailures.With(metricsHandler).Record(1)
+				logger.Error("Failed to fetch next page", tag.Error(err))
+				return HeartBeatDetails{}, err
```
bergundy: The SDK logger should take care of this and it's better to add context to the error so it is easier to debug if it shows up in the UI.

```suggestion
				return HeartBeatDetails{}, fmt.Errorf("Failed to fetch next page: %w", err)
```

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+			// Link pages
+			p.next = nextPage
+			nextPage.prev = p
+			p = nextPage
+
+			// Update heartbeat details for new page (thread-safe)
+			hbdMutex.Lock()
+			hbd.CurrentPage = p.pageNumber
+			hbd.PageToken = p.nextPageToken
+			hbdMutex.Unlock()
```
bergundy: We shouldn't need to lock here. The only access to `hbd` should be in this single goroutine.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+		select {
+		case taskCh <- p.nextTask():
+			// Successfully submitted a task
+			p.submittedCount++
+
+		case result := <-respCh:
+			// Handle task completion result
+			resultPage := result.page
+
+			if resultPage != nil {
```
bergundy: Just assume it can never be `nil` since it's our code that always sets it.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+				// Update counts (thread-safe)
+				hbdMutex.Lock()
```
bergundy: No need to lock.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+				if result.err == nil {
+					resultPage.successCount++
+					hbd.SuccessCount++
```
bergundy: You can't increment hbd until the earlier page has completed processing as we discussed.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+					for curr := resultPage.prev; curr != nil; curr = curr.prev {
+						if curr.successCount+curr.errorCount < len(curr.executions) {
+							allPrevComplete = false
+							break
+						}
+					}
+					if allPrevComplete {
+						// Send immediate heartbeat on page completion (original behavior)
+						hbdMutex.RLock()
+						activity.RecordHeartbeat(ctx, hbd)
```
bergundy: We should heartbeat more frequently to avoid heartbeat timeouts due to slowness during processing. I would use a ticker and heartbeat when the ticker channel is unblocked in this select statement.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+				// Update heartbeat details if this page and all previous pages are complete
+				if resultPage.successCount+resultPage.errorCount == len(resultPage.executions) {
+					allPrevComplete := true
+					for curr := resultPage.prev; curr != nil; curr = curr.prev {
+						if curr.successCount+curr.errorCount < len(curr.executions) {
+							allPrevComplete = false
+							break
+						}
+					}
+					if allPrevComplete {
```
bergundy: Prefer the inverse condition to avoid extra nesting.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+					hbd.SuccessCount++
+				} else {
+					resultPage.errorCount++
+					hbd.ErrorCount++
+				}
+				hbdMutex.Unlock()
+
+				// Update heartbeat details if this page and all previous pages are complete
+				if resultPage.successCount+resultPage.errorCount == len(resultPage.executions) {
+					allPrevComplete := true
```
bergundy: You can use `resultPage.done()` here.

service/worker/batcher/activities.go
```diff
@@ -41,6 +42,248 @@ var (
...
+							break
+						}
+					}
+					if allPrevComplete {
+						// Send immediate heartbeat on page completion (original behavior)
+						hbdMutex.RLock()
+						activity.RecordHeartbeat(ctx, hbd)
+						hbdMutex.RUnlock()
+
+						// Delete all previous page pointers for completed pages to avoid scanning repeatedly
```
bergundy: You just need to break the last prev link, no need to recurse back.

service/worker/batcher/activities.go
```diff
@@ -386,13 +529,14 @@ func (a *activities) getOperationConcurrency(concurrency int) int {
 func startTaskProcessor(
```
bergundy, line 490: No need to change this IMHO, it's never going to be called again and we should delete it once we know nobody is calling the old workflow directly anymore.

service/worker/batcher/activities.go
spkane31: Removed the locks

service/worker/batcher/activities.go
spkane31: added a ticker every 2.5 seconds (1/4th of the default timeout)

service/frontend/workflow_handler.go
```diff
@@ -4815,9 +4815,7 @@ func (wh *WorkflowHandler) ListBatchOperations(
 		Namespace:     request.GetNamespace(),
 		PageSize:      request.PageSize,
 		NextPageToken: request.GetNextPageToken(),
-		Query: fmt.Sprintf("%s = '%s' and %s = '%s'",
-			searchattribute.WorkflowType,
-			batcher.BatchWFTypeProtobufName,
+		Query: fmt.Sprintf("%s = '%s'",
```
bergundy: This change should already be in `main`.

service/worker/batcher/activities.go
```diff
@@ -41,6 +41,223 @@ var (
...
+					}
+				}
+				if allPrevComplete {
+					hbd.SuccessCount += resultPage.successCount
+					hbd.ErrorCount += resultPage.errorCount
+
+					activity.RecordHeartbeat(ctx, hbd)
+
+					resultPage.prev = nil
+				}
```
bergundy: I mentioned this in the previous review, `done()` is recursive already.

```suggestion
				hbd.SuccessCount += resultPage.successCount
				hbd.ErrorCount += resultPage.errorCount
```

bergundy (commented):
Added a couple of last comments.

service/worker/batcher/activities.go
```diff
@@ -41,6 +41,214 @@ var (
...
+
+			// Update heartbeat details if this page and all previous pages are complete
+			if resultPage.done() {
+				hbd.SuccessCount += resultPage.successCount
+				hbd.ErrorCount += resultPage.errorCount
+
+				activity.RecordHeartbeat(ctx, hbd)
+
+				resultPage.prev = nil
+			}
```
bergundy:
```suggestion
			// Find all pages from the current one on that are done, record their stats, and unlink them.
			for page := resultPage; page != nil && page.done(); page = page.next {
				hbd.SuccessCount += page.successCount
				hbd.ErrorCount += page.errorCount
				page.prev = nil
			}
```

bergundy (approved):
Please apply the suggestion before merging.

PR #8147, opened 2025-08-05
Refactor dynamicconfig.FileBasedClient to make reusable

dnr:
## What changed?
Pull out parts of FileBasedConfig to make them reusable.

## Why?
Make it easier to build other Clients.

## How did you test it?
- [x] built

### DIFF
--- a/common/dynamicconfig/client_diff.go
+++ b/common/dynamicconfig/client_diff.go
@@ -1,0 +1,123 @@
+package dynamicconfig
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	enumspb "go.temporal.io/api/enums/v1"
+	enumsspb "go.temporal.io/server/api/enums/v1"
+	"go.temporal.io/server/common/log"
+)
+
+// DiffAndLogConfigs computes the difference between two ConfigValueMaps. The result is
+// returned as a ConfigValueMap that can be merged with old to produce new, except with deleted
+// keys mapped to nil. It also logs the differences to a logger.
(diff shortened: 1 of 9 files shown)


common/dynamicconfig/file_based_client.go
prathyushpv, line 86: Can we use Key.Lower() here?

common/dynamicconfig/client_diff.go
```diff
@@ -0,0 +1,123 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	enumsspb "go.temporal.io/server/api/enums/v1"
+	"go.temporal.io/server/common/log"
+)
+
+// DiffAndLogConfigs computes the difference between two ConfigValueMaps. The result is
+// returned as a ConfigValueMap that can be merged with old to produce new, except with deleted
+// keys mapped to nil. It also logs the differences to a logger.
+func DiffAndLogConfigs(logger log.Logger, oldValues ConfigValueMap, newValues ConfigValueMap) ConfigValueMap {
+	changedMap := make(map[Key][]ConstrainedValue)
```
prathyushpv, line 17: Can we preallocate? `changedMap := make(map[Key][]ConstrainedValue, len(newValues))`

dnr: I don't think we should. Usually very few values change on each reload.

common/dynamicconfig/file_based_client.go
dnr: well, yes, although I'm going to remove Key.Lower() in the next PR :)

also I'd need to cast it back to a string here since the map uses strings. overall it seems not worth it to mess with this since it's going away.

PR #8149, opened 2025-08-05
fix(nexus): add nexus service and operation tags to metrics

spkane31:
## What changed?
added `nexus_{service,operation}` tags to metrics for better observability

## Why?
Metrics were missing these fields making debugging difficult

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
NA

### DIFF
--- a/api/persistence/v1/executions.go-helpers.pb.go
+++ b/api/persistence/v1/executions.go-helpers.pb.go
@@ -929,3 +929,77 @@ func (this *ResetChildInfo) Equal(that interface{}) bool {
 
 	return proto.Equal(this, that1)
 }
+
+// Marshal an object of type WorkflowPauseInfo to the protobuf v3 wire format
+func (val *WorkflowPauseInfo) Marshal() ([]byte, error) {
+	return proto.Marshal(val)
+}
+
+// Unmarshal an object of type WorkflowPauseInfo from the protobuf v3 wire format
+func (val *WorkflowPauseInfo) Unmarshal(buf []byte) error {
+	return proto.Unmarshal(buf, val)
+}
+
+// Size returns the size of the object, in bytes, once serialized
+func (val *WorkflowPauseInfo) Size() int {
+	return proto.Size(val)
+}
+
+// Equal returns whether two WorkflowPauseInfo values are equivalent by recursively
+// comparing the message's fields.
+// For more information see the documentation for
+// https://pkg.go.dev/google.golang.org/protobuf/proto#Equal
+func (this *WorkflowPauseInfo) Equal(that interface{}) bool {
+	if that == nil {
+		return this == nil
+	}
+
+	var that1 *WorkflowPauseInfo
+	switch t := that.(type) {
+	case *WorkflowPauseInfo:
+		that1 = t
+	case WorkflowPauseInfo:
+		that1 = &t
+	default:
+		return false
+	}
+
+	return proto.Equal(this, that1)
+}
+
+// Marshal an object of type ActivityPauseInfo to the protobuf v3 wire format
+func (val *ActivityPauseInfo) Marshal() ([]byte, error) {
+	return proto.Marshal(val)
+}
+
+// Unmarshal an object of type ActivityPauseInfo from the protobuf v3 wire format
+func (val *ActivityPauseInfo) Unmarshal(buf []byte) error {
+	return proto.Unmarshal(buf, val)
+}
+
+// Size returns the size of the object, in bytes, once serialized
+func (val *ActivityPauseInfo) Size() int {
+	return proto.Size(val)
+}
+
+// Equal returns whether two ActivityPauseInfo values are equivalent by recursively
+// comparing the message's fields.
+// For more information see the documentation for
+// https://pkg.go.dev/google.golang.org/protobuf/proto#Equal
+func (this *ActivityPauseInfo) Equal(that interface{}) bool {
+	if that == nil {
+		return this == nil
+	}
+
+	var that1 *ActivityPauseInfo
+	switch t := that.(type) {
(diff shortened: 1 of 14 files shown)


bergundy:
We shouldn't add labels to existing metrics without prior discussion, when we added these label on the frontend, we put them behind a dynamic config to avoid high cardinality issues, which can bring down metrics backends. What's the motivation for adding this?

components/nexusoperations/executors.go
```diff
@@ -218,13 +218,18 @@ func (e taskExecutor) executeInvocationTask(ctx context.Context, env hsm.Environ
...
+		metrics.NamespaceTag(ns.Name().String()),
+		metrics.DestinationTag(endpoint.Endpoint.Spec.GetName()),
+		metrics.OutcomeTag(startCallOutcomeTag(callCtx, rawResult, callErr)),
+		metrics.FailureSourceTag(failureSourceFromContext(ctx)),
+		metrics.NexusServiceTag(args.service),
+		metrics.NexusOperationTag(args.operation),
+	)
+
+	metricsHandler.Counter(OutboundRequestCounter.Name()).Record(1)
+	metricsHandler.Timer(OutboundRequestLatency.Name()).Record(time.Since(startTime))
```
rodrigozhou, line 232: There's another call to export these metrics in this file.

components/nexusoperations/frontend/handler.go
```diff
@@ -524,3 +525,11 @@ func (c *requestContext) shouldForwardRequest(ctx context.Context, header http.H
 		len(c.namespace.ClusterNames()) > 1 &&
 		c.Config.ForwardingEnabledForNamespace(c.namespace.Name().String())
 }
+
+// enrichNexusOperationMetrics enhances metrics with additional Nexus operation context.
+func (c *requestContext) enrichNexusOperationMetrics(service, operation string) {
+	c.metricsHandler = c.metricsHandler.WithTags(
+		metrics.NexusServiceTag(service),
+		metrics.NexusOperationTag(operation),
+	)
```
rodrigozhou: The other `enrich` function checks dynamic configs to add the tags. I think you should do the same.

rodrigozhou (changes requested):

components/nexusoperations/frontend/handler.go
spkane31: added dynamic config via dependency injection

components/nexusoperations/frontend/fx.go
```diff
@@ -31,6 +32,16 @@ func ConfigProvider(coll *dynamicconfig.Collection) *Config {
...
 
+func MetricTagConfigProvider(dc *dynamicconfig.Collection) *dynamicconfig.GlobalCachedTypedValue[*nexusoperations.NexusMetricTagConfig] {
+	return dynamicconfig.NewGlobalCachedTypedValue(
+		dc,
+		nexusoperations.MetricTagConfiguration,
+		func(config nexusoperations.NexusMetricTagConfig) (*nexusoperations.NexusMetricTagConfig, error) {
+			return &config, nil
+		},
+	)
+}
```
rodrigozhou: Instead of having this a separate dependency, include another field in the `Config` object which already stores other dynamic configs.

rodrigozhou (changes requested):

components/nexusoperations/frontend/handler.go
```diff
@@ -524,3 +528,29 @@ func (c *requestContext) shouldForwardRequest(ctx context.Context, header http.H
...
+
+	if conf.IncludeServiceTag {
+		tags = append(tags, metrics.NexusServiceTag(service))
+	}
+
+	if conf.IncludeOperationTag {
+		tags = append(tags, metrics.NexusOperationTag(operation))
+	}
+
+	for _, mapping := range conf.HeaderTagMappings {
```
bergundy, line 549: This config was meant for the handler side, I'm a bit hesitant to also apply it on the caller side but I guess there's no harm in doing so.

components/nexusoperations/frontend/handler.go
```diff
@@ -118,8 +119,10 @@ func (h *completionHandler) CompleteOperation(ctx context.Context, r *nexus.Comp
 			metrics.NamespaceTag(nsName),
 		),
 		requestStartTime: startTime,
+		metricTagConfig:  h.Config.MetricTagConfig,
 	}
 	ctx = rCtx.augmentContext(ctx, r.HTTPRequest.Header)
+	rCtx.enrichNexusOperationMetrics(r.HTTPRequest.Method, r.HTTPRequest.URL.Path, r.HTTPRequest.Header)
```
bergundy, line 125: Those aren't service and operation names. These labels can't be available on this metric.

bergundy (commented):
TBH, I would just not add these labels altogether until we see a reason to include them but at this point the work has already been done so we may as well merge it.
Please see my comment below.

components/nexusoperations/frontend/fx.go
```diff
@@ -28,6 +28,13 @@ func ConfigProvider(coll *dynamicconfig.Collection) *Config {
 		PayloadSizeLimit:              dynamicconfig.BlobSizeLimitError.Get(coll),
 		ForwardingEnabledForNamespace: dynamicconfig.EnableNamespaceNotActiveAutoForwarding.Get(coll),
 		MaxOperationTokenLength:       nexusoperations.MaxOperationTokenLength.Get(coll),
+		MetricTagConfig: dynamicconfig.NewGlobalCachedTypedValue(
```
dnr, line 31: I'm going to remove this soon (finally). If #7052 lands first then you can just get rid of this and use the setting directly. If this lands first, I'll do it in that PR.

PR #8151, opened 2025-08-05
[CHASM] Relax fieldsOf to allow private fields

lina-temporal:
## What changed?
- Relax fieldsOf iterator.

## Why?
- Allow private/locally-cached fields in structs.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


chasm/fields_iterator.go
```diff
@@ -64,7 +67,7 @@ func fieldsOf(valueV reflect.Value) iter.Seq[fieldInfo] {
 					case chasmMapTypePrefix:
 						fieldK = fieldKindSubMap
 					default:
-						fieldErr = serviceerror.NewInternalf("%s.%s: unsupported field type %s: must implement proto.Message, or be chasm.Field[T] or chasm.Map[T]", valueT, fieldN, fieldT)
+						continue // Skip non-CHASM fields.
```
yycptt, line 75: I think we should change logic on L62 as well to allow (and ignore) generic pointer fields that are not SubField or SubMap.

yycptt (commented):
Please also create an issue to track the work for logging ignored fields when registering component.

chasm/fields_iterator.go
```diff
@@ -56,15 +59,16 @@ func fieldsOf(valueV reflect.Value) iter.Seq[fieldInfo] {
 			} else {
 				prefix := genericTypePrefix(fieldT)
 				if strings.HasPrefix(prefix, "*") {
-					fieldErr = serviceerror.NewInternalf("%s.%s: chasm field type %s must not be a pointer", valueT, fieldN, fieldT)
+					// Skip non-CHASM generic pointer fields.
+					continue
```
yycptt: I guess we should still fail if field is a pointer to chasm.Field, or chasm.Map?

yycptt (approved):

chasm/fields_iterator.go
lina-temporal: Oops, yes, will fix before push 👍

PR #8154, opened 2025-08-06
Set `ForceAttemptHTTP2=true` for remote frontend HTTP clients

pdoerner:
## What changed?
Setting `ForceAttemptHTTP2=true` for remote frontend HTTP clients. Local FE clients already set this.

## Why?
ForceAttemptHTTP2 is the default except when providing a custom Dial , DialTLS , DialContext , or TLSClientConfig to prevent breaking older code: https://github.com/golang/go/issues/14275

We set this parameter to true for local clients, so when TLS is enabled and a failover happens, callback requests will fail because the cluster uses a local frontend to make the request with HTTP2 which is then forwarded using HTTP1.x and causes a malformed response.

### DIFF
--- a/common/cluster/frontend_http_client.go
+++ b/common/cluster/frontend_http_client.go
@@ -61,7 +61,10 @@ func (c *FrontendHTTPClientCache) newClientForCluster(targetClusterName string) 
 		if err != nil {
 			return nil, err
 		}
-		client.Transport = &http.Transport{TLSClientConfig: tlsClientConfig}
+		client.Transport = &http.Transport{
+			TLSClientConfig:   tlsClientConfig,
+			ForceAttemptHTTP2: true,
+		}
 		if tlsClientConfig != nil {
 			urlScheme = "https"
 		}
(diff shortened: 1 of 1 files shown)


bergundy (approved):

common/cluster/frontend_http_client.go
```diff
@@ -61,7 +61,10 @@ func (c *FrontendHTTPClientCache) newClientForCluster(targetClusterName string)
 		if err != nil {
 			return nil, err
 		}
-		client.Transport = &http.Transport{TLSClientConfig: tlsClientConfig}
+		client.Transport = &http.Transport{
```
bergundy: Can you set all of the fields we copied from the default transport here?

https://github.com/temporalio/temporal/blob/356eecafdd3fcc57c8c68019c53a27ada803d31b/common/rpc/rpc.go#L291

pdoerner: Sure, makes sense to me.

PR #8162, opened 2025-08-07
Instrument GRPC Dial with 

simvlad:
## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


simvlad:
Will need to figure out cross platform solution for this. Maybe will opt for an easier fix

simvlad:
Looks like we use OpenTelemetry for gRPC here: https://github.com/temporalio/temporal/blob/da90f6248661a4cb6fd463815075e8abd9935c6b/temporal/fx.go#L988 when the tracing is enabled and it has all these:

* grpc_client_connection_duration
* grpc_client_connection_attempts_total
* grpc_client_connection_failures_total

Wondering if we can use that instead of the approach in this PR

common/rpc/grpc.go
```diff
@@ -164,3 +177,74 @@ func addHeadersForResourceExhausted(ctx context.Context, logger log.Logger, err
...
+		// start is the start time for the current attempt
+		var start time.Time
+
+		for {
+			st := cc.GetState()
+
+			switch st {
+			case connectivity.Idle:
+				// No action needed for idle state
+			case connectivity.Connecting:
```
prathyushpv: Can we miss any of these states when `connectivity.Idle->connectivity.Connecting->connectivity.Ready` Transition moves very fast and we do not see `connectivity.Connecting` in this loop? In that case we will miss attempt count and start time.

simvlad:
Ok, I think the best way is to use a custom Dialer for this, but will need to check how to set the tcp keepalive properly

simvlad:
Matched naming with https://github.com/temporalio/temporal/pull/8237

stephanos (approved):
Great tests! 🥇

PR #8163, opened 2025-08-07
[Scheduled Actions] Update Scheduler protos for CHASM

lina-temporal:
## What changed?
- Added protos for the new Scheduler task types.
- Added TODOs for cleanup when the HSM component is removed.

## Why?
- A few fields and messages were made obsolete with the CHASM port.


proto/internal/temporal/server/api/enums/v1/common.proto
```diff
@@ -15,7 +15,8 @@ enum ChecksumFlavor {
     CHECKSUM_FLAVOR_IEEE_CRC32_OVER_PROTO3_BINARY = 1;
 }
 
-// State for the state machine scheduler's Invoker.
+// State for the CHASM scheduler's Invoker.
+// TODO - remove after HSM scheduler is removed (obsolete)
```
bergundy, line 19: Why not just remove the HSM scheduler and start fresh?

lina-temporal: Just to leave it around for quick comparison as the CHASM stuff is merged, I'll be removing it after this initial round of PRs gets merged :).

PR #8164, opened 2025-08-07
[Scheduled Actions] Port HSM utils -> CHASM package

lina-temporal:
## What changed?
- Port utility functions from HSM package to CHASM. Very few changes otherwise. 

## Why?
- We'll need them

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


chasm/lib/scheduler/helper_test.go
```diff
@@ -0,0 +1,64 @@
...
+				{
+					Interval: durationpb.New(defaultInterval),
+					Phase:    durationpb.New(0),
+				},
+			},
+		},
+		Action: &schedulepb.ScheduleAction{
+			Action: &schedulepb.ScheduleAction_StartWorkflow{
+				StartWorkflow: &workflowpb.NewWorkflowExecutionInfo{
+					WorkflowId: "scheduled-wf",
```
bergundy, line 37: Shouldn't this include workflow type and task queue for correctness? I thought we don't allow users to set workflow IDs in schedules (am I wrong).
Not critical because this is just a test util, just wondering if we want to model something closer to a real use case.

chasm/lib/scheduler/spec_processor.go
```diff
@@ -0,0 +1,155 @@
...
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
+	scheduler1 "go.temporal.io/server/service/worker/scheduler"
+	"go.uber.org/fx"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination spec_processor_mock.go
```
bergundy, line 16: Do you need codegen for mocks? Wouldn't it be simpler to just embed the interface and implement just what you want in your tests?

bergundy: Does this also require an interface or can we do with just a struct? I would rather have less indirection in our codebase to enable a simpler debugging experience.

chasm/lib/scheduler/spec_processor_test.go
```diff
@@ -0,0 +1,230 @@
+package scheduler_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/suite"
```
bergundy, line 8: Personally would prefer not using testify suites due to lack of support of parallel tests and difficult to use interface (`s.Equal` doesn't work in subtests).
Not blocking since this is a personal preference.

chasm/lib/scheduler/util.go
```diff
@@ -0,0 +1,56 @@
...
+		nominal.UnixMilli(),
+		actual.UnixMilli(),
+	)
+}
+
+// newTaggedLogger returns a logger tagged with the Scheduler's attributes.
+func newTaggedLogger(baseLogger log.Logger, scheduler *Scheduler) log.Logger {
+	return log.With(
+		baseLogger,
+		tag.NewStringTag("wf-namespace", scheduler.Namespace),
```
bergundy: We should have a predefined tag for namespace.

chasm/lib/scheduler/helper_test.go
lina-temporal: > Shouldn't this include workflow type and task queue for correctness?

I prefer not to include extraneous fields in test code, since it can give the impression those fields are under test when they aren't accessed.

>  I thought we don't allow users to set workflow IDs in schedules (am I wrong).

Yes: https://github.com/temporalio/api/blob/master/temporal/api/schedule/v1/message.proto#L234-L239

Users set the workflow ID, and we append the timestamp.

chasm/lib/scheduler/spec_processor.go
lina-temporal: > Do you need codegen for mocks? Wouldn't it be simpler to just embed the interface and implement just what you want in your tests?

I'd rather stick to a single style of mocking throughout these tests; admittedly, the only time I stub this is to throw an error on `ProcessTimeRange`, else, I do use the `testSpecProcessor`. Still, I'd rather generate a mock than write a struct specifically to act as that stub, I don't think the generated code or annotation to generate it costs us anything.

chasm/lib/scheduler/util.go
yycptt: Are Scheduler.Namespace/ScheduleId in memory fields not managed by CHASM framework?

chasm/lib/scheduler/spec_processor_test.go
```diff
@@ -0,0 +1,230 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	schedulepb "go.temporal.io/api/schedule/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/chasm/lib/scheduler"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/metrics"
+	scheduler1 "go.temporal.io/server/service/worker/scheduler"
```
yycptt: shall we port this as well?

chasm/lib/scheduler/util.go
lina-temporal: Fixed, we had one for SchedulerID as well.

lina-temporal: They're CHASM-managed, yeah - they're just on the state protobuf, which is embedded onto the Scheduler struct. See `SchedulerInternal`.

chasm/lib/scheduler/spec_processor_test.go
lina-temporal: I opted to reference it as long as the rest of the WF scheduler code (which still uses it) is active, but I suppose we could also bring it over in advance and update the WF scheduler. I don't feel strongly about it, wdyt?

yycptt: no strong opinion on this either. Your call.

chasm/lib/scheduler/spec_processor.go
bergundy: It does cost us something, an extra interface and a layer of indirection. That being said, I can see why you may need the interface for testing purposes. Since we are setting a standard here for chasm component authors I would like us to agree on a style for the codebase that can be consistent. I am not the only person who would rather avoid using codegen for mocks.

chasm/lib/scheduler/spec_processor_test.go
bergundy: Same as above, I would like to not recommend devs to use suites, let's decide on this as a team whether we want this style in the codebase.

bergundy:
I'm not blocking the PR since my comments are stylistic and we need team agreement on the style. We may need to revisit the style changes later pending a team discussion.

PR #8165, opened 2025-08-07
[Scheduled Actions] Port HSM components -> CHASM

lina-temporal:
## What changed?
- Port the HSM scheduler components to CHASM. This is most everything related to a component aside from their task executors/validators. 
- I also included the overall scheduler base test suite, since it'll be used throughout the executor tests. 
- Apologies for large review, however, I figured that having all of these in one PR would make their interactions a bit more readable. 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/backfiller.go
+++ b/chasm/lib/scheduler/backfiller.go
@@ -4,9 +4,9 @@ import (
 	"time"
 
 	"github.com/pborman/uuid"
-	legacyschedulespb "go.temporal.io/server/api/schedule/v1"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
 	"go.temporal.io/server/chasm"
-	schedulespb "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
+	"go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
@@ -15,12 +15,11 @@ import (
 type Backfiller struct {
 	chasm.UnimplementedComponent
 
-	*schedulespb.BackfillerInternal
+	*schedulerpb.BackfillerState
 
 	Scheduler chasm.Field[*Scheduler]
 }
 
-// The type of Backfill represented by an invididual Backfiller.
 type BackfillRequestType int
 
 const (
@@ -36,7 +35,7 @@ func newBackfiller(
 ) *Backfiller {
 	id := uuid.New()
 	backfiller := &Backfiller{
-		BackfillerInternal: &schedulespb.BackfillerInternal{
+		BackfillerState: &schedulerpb.BackfillerState{
 			BackfillId:        id,
 			LastProcessedTime: timestamppb.New(ctx.Now(scheduler)),
 		},
@@ -46,8 +45,6 @@ func newBackfiller(
 }
 
 func (b *Backfiller) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
-	// TODO - given that the component is deleted from the Scheduler's Backfillers
-	// map, and has no children, do we need to do anything here?
 	return chasm.LifecycleStateRunning
 }
 
@@ -61,7 +58,7 @@ func (b *Backfiller) RequestType() BackfillRequestType {
 
 type backfillProgressResult struct {
 	// BufferedStarts that should be enqueued to the Invoker.
-	BufferedStarts []*legacyschedulespb.BufferedStart
+	BufferedStarts []*schedulespb.BufferedStart
 
 	// High water mark for when state was last updated.
 	LastProcessedTime time.Time
--- a/chasm/lib/scheduler/generator.go
+++ b/chasm/lib/scheduler/generator.go
@@ -2,7 +2,7 @@ package scheduler
 
 import (
 	"go.temporal.io/server/chasm"
-	schedulespb "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
+	"go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
 )
 
 // The Generator component is responsible for buffering actions according
@@ -11,7 +11,7 @@ import (
 type Generator struct {
 	chasm.UnimplementedComponent
 
-	*schedulespb.GeneratorInternal
+	*schedulerpb.GeneratorState
 
 	Scheduler chasm.Field[*Scheduler]
 }
@@ -20,7 +20,7 @@ type Generator struct {
 // be parented under a Scheduler root node.
 func NewGenerator(ctx chasm.MutableContext, scheduler *Scheduler, invoker *Invoker) *Generator {
 	return &Generator{
-		GeneratorInternal: &schedulespb.GeneratorInternal{
+		GeneratorState: &schedulerpb.GeneratorState{
 			LastProcessedTime: nil,
 		},
 		Scheduler: chasm.ComponentPointerTo(ctx, scheduler),
--- a/chasm/lib/scheduler/invoker.go
+++ b/chasm/lib/scheduler/invoker.go
@@ -5,9 +5,9 @@ import (
 	"time"
 
 	commonpb "go.temporal.io/api/common/v1"
-	legacyschedulespb "go.temporal.io/server/api/schedule/v1"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
 	"go.temporal.io/server/chasm"
-	schedulespb "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
+	"go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
 	"go.temporal.io/server/common/util"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
@@ -16,7 +16,7 @@ import (
 type Invoker struct {
 	chasm.UnimplementedComponent
 
-	*schedulespb.InvokerInternal
+	*schedulerpb.InvokerState
 
 	Scheduler chasm.Field[*Scheduler]
 }
@@ -25,8 +25,8 @@ type Invoker struct {
 // be parented under a Scheduler root component.
 func NewInvoker(ctx chasm.MutableContext, scheduler *Scheduler) *Invoker {
 	return &Invoker{
-		InvokerInternal: &schedulespb.InvokerInternal{
-			BufferedStarts: []*legacyschedulespb.BufferedStart{},
+		InvokerState: &schedulerpb.InvokerState{
+			BufferedStarts: []*schedulespb.BufferedStart{},
 		},
 		Scheduler: chasm.ComponentPointerTo(ctx, scheduler),
 	}
@@ -34,30 +34,26 @@ func NewInvoker(ctx chasm.MutableContext, scheduler *Scheduler) *Invoker {
 
 // EnqueueBufferedStarts adds new BufferedStarts to the invocation queue,
 // immediately kicking off a processing task.
-func (i *Invoker) EnqueueBufferedStarts(ctx chasm.MutableContext, starts []*legacyschedulespb.BufferedStart) {
+func (i *Invoker) EnqueueBufferedStarts(ctx chasm.MutableContext, starts []*schedulespb.BufferedStart) {
 	i.BufferedStarts = append(i.BufferedStarts, starts...)
 
 	// Immediately begin processing the new starts.
-	ctx.AddTask(i, chasm.TaskAttributes{}, &schedulespb.InvokerProcessBufferTask{})
+	ctx.AddTask(i, chasm.TaskAttributes{}, &schedulerpb.InvokerProcessBufferTask{})
 }
 
 type processBufferResult struct {
-	startWorkflows     []*legacyschedulespb.BufferedStart
+	startWorkflows     []*schedulespb.BufferedStart
 	cancelWorkflows    []*commonpb.WorkflowExecution
 	terminateWorkflows []*commonpb.WorkflowExecution
 
 	// discardStarts will be dropped from the Invoker's BufferedStarts without execution.
-	discardStarts []*legacyschedulespb.BufferedStart
+	discardStarts []*schedulespb.BufferedStart
 
 	// Number of buffered starts dropped due to overlap policy during processing.
 	overlapSkipped int64
 
 	// Nunmber of buffered starts dropped from missing the catchup window.
 	missedCatchupWindow int64
-
-	// Number of buffered starts dropped due to the max buffer size having been exceeded.
-	// TODO - set this from Generator/Backfiller
-	bufferDropped int64
 }
 
 // recordProcessBufferResult updates the Invoker's internal state based on result, as well as the
@@ -73,7 +69,7 @@ func (i *Invoker) recordProcessBufferResult(ctx chasm.MutableContext, result *pr
 	}
 
 	// Drop discarded starts, and update requested starts for execution.
-	var starts []*legacyschedulespb.BufferedStart
+	var starts []*schedulespb.BufferedStart
 	for _, start := range i.GetBufferedStarts() {
 		if discards[start.RequestId] {
 			continue
@@ -98,13 +94,13 @@ func (i *Invoker) recordProcessBufferResult(ctx chasm.MutableContext, result *pr
 
 type executeResult struct {
 	// Starts that executed successfully can be removed from the buffer.
-	CompletedStarts []*legacyschedulespb.BufferedStart
+	CompletedStarts []*schedulespb.BufferedStart
 
 	// Starts that failed with a retryable error should be updated and kept in the buffer.
-	RetryableStarts []*legacyschedulespb.BufferedStart
+	RetryableStarts []*schedulespb.BufferedStart
 
 	// Starts that failed with a non-retryable error can be removed from the buffer.
-	FailedStarts []*legacyschedulespb.BufferedStart
+	FailedStarts []*schedulespb.BufferedStart
 
 	CompletedCancels    []*commonpb.WorkflowExecution
 	CompletedTerminates []*commonpb.WorkflowExecution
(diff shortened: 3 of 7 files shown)


chasm/lib/scheduler/backfiller.go
```diff
@@ -0,0 +1,108 @@
...
+) *Backfiller {
+	backfiller := newBackfiller(ctx, scheduler, invoker)
+	backfiller.BackfillerInternal.Request = &schedulespb.BackfillerInternal_TriggerRequest{
+		TriggerRequest: request,
+	}
+	return backfiller
+}
+
+func (b *Backfiller) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
+	// TODO - given that the component is deleted from the Scheduler's Backfillers
```
bergundy: Wondering if this should be the default and if we should force component implementors to implement it.

chasm/lib/scheduler/invoker.go
```diff
@@ -0,0 +1,218 @@
...
+// EnqueueBufferedStarts adds new BufferedStarts to the invocation queue,
+// immediately kicking off a processing task.
+func (i *Invoker) EnqueueBufferedStarts(ctx chasm.MutableContext, starts []*schedulespb.BufferedStart) {
+	i.BufferedStarts = append(i.BufferedStarts, starts...)
+
+	// Immediately begin processing the new starts.
+	ctx.AddTask(i, chasm.TaskAttributes{}, &schedulespb.InvokerProcessBufferTask{})
+}
+
+type processBufferResult struct {
```
bergundy, line 44: nit: do any of these fields need to be exported?

chasm/lib/scheduler/invoker.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	for _, wf := range result.CompletedCancels {
+		canceled[wf.RunId] = true
+	}
+	for _, wf := range result.CompletedTerminates {
+		terminated[wf.RunId] = true
+	}
+
+	// Update Invoker state to remove completed items from their buffers.
+	i.BufferedStarts = util.FilterSlice(i.GetBufferedStarts(), func(start *schedulespb.BufferedStart) bool {
```
bergundy: Any reason not to use `slices.DeleteFunc`?

chasm/lib/scheduler/invoker.go
```diff
@@ -0,0 +1,218 @@
...
+// addTasks adds both ProcessBuffer and Execute tasks as needed. It should be
+// called when completing processing/executing tasks, to drive backoff/retry.
+func (i *Invoker) addTasks(ctx chasm.MutableContext) {
+	totalStarts := len(i.GetBufferedStarts())
+	eligibleStarts := len(i.getEligibleBufferedStarts())
+
+	// Add a ProcessBuffer pure task whenever there are BufferedStarts that are
+	// backing off, or are still pending initial processing.
+	if (totalStarts - eligibleStarts) > 0 {
+		ctx.AddTask(i, chasm.TaskAttributes{
```
bergundy, line 177: Are these tasks automatically deduped if this method is called multiple times?

chasm/lib/scheduler/invoker.go
```diff
@@ -0,0 +1,218 @@
...
+			ScheduledTime: i.processingDeadline(),
+		}, &schedulespb.InvokerProcessBufferTask{})
+	}
+
+	// Add an Execute side effect task whenever there are any eligible actions
+	// pending execution.
+	if len(i.CancelWorkflows) > 0 || len(i.TerminateWorkflows) > 0 || eligibleStarts > 0 {
+		ctx.AddTask(i, chasm.TaskAttributes{}, &schedulespb.InvokerExecuteTask{})
```
bergundy: Same here, what are the dedupe semantics?

chasm/lib/scheduler/scheduler.go
```diff
@@ -0,0 +1,238 @@
...
+		if enumspb.SCHEDULE_OVERLAP_POLICY_ALLOW_ALL == s.resolveOverlapPolicy(policy) {
+			return true
+		}
+	}
+
+	return false
+}
+
+type schedulerActionResult struct {
+	OverlapSkipped      int64
```
bergundy: nit: seems like these fields shouldn't be exported.

chasm/lib/scheduler/scheduler.go
```diff
@@ -0,0 +1,238 @@
...
+	BufferDropped       int64
+	MissedCatchupWindow int64
+	Starts              []*schedulepb.ScheduleActionResult
+}
+
+// recordActionResult updates the Scheduler's customer-facing metadata with execution results.
+func (s *Scheduler) recordActionResult(
+	ctx chasm.MutableContext,
+	result *schedulerActionResult,
+) (struct{}, error) {
```
bergundy: Why return an empty struct here (or an error since it's always nil)?

bergundy (approved):
I'm a bit out of context on both CHASM and the scheduler logic. Asked some questions on the CHASM parts, will defer on to others on the business logic.

chasm/lib/scheduler/backfiller.go
```diff
@@ -0,0 +1,108 @@
...
+			BackfillId:        id,
+			LastProcessedTime: timestamppb.New(ctx.Now(scheduler)),
+		},
+		Scheduler: chasm.ComponentPointerTo(ctx, scheduler),
+		Invoker:   chasm.ComponentPointerTo(ctx, invoker),
+	}
+
+	// Add the backfiller to the scheduler tree, and add a task to kick off backfill
+	// processing.
+	scheduler.Backfillers[id] = chasm.NewComponentField(ctx, backfiller)
```
yycptt: nit: ideally component should not access field in another component directly. They are only exported for the reflection to work.

chasm/lib/scheduler/backfiller.go
yycptt: That's a good question. Right now it's practically only enforced on the root component (the default impl returns Unspecified and will cause failure on CloseTransaction). For non-root, if not implemented, the behavior is the same as always returning StateRunning. 

I don't think we can have a default for root component. So probably make sense to just always enforce implementing that and remove the default LifecycleState impl in UnimplementedComponent?

chasm/lib/scheduler/backfiller.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (b *Backfiller) RequestType() BackfillRequestType {
+	if b.GetTriggerRequest() != nil {
+		return RequestTypeTrigger
+	}
+
+	return RequestTypeBackfill
+}
+
+type backfillProgressResult struct {
```
yycptt, line 59: There used to be a `NextInvocationTime` field here. That's no longer necessary?

chasm/lib/scheduler/generator.go
```diff
@@ -0,0 +1,34 @@
...
+// The Generator component is responsible for buffering actions according
+// to the schedule's specification. Manually requested actions (from an immediate
+// request or backfill) are separately handled in the Backfiller component.
+type Generator struct {
+	chasm.UnimplementedComponent
+
+	*schedulespb.GeneratorInternal
+
+	Scheduler chasm.Field[*Scheduler]
+	Invoker   chasm.Field[*Invoker]
```
yycptt: I wonder if this is necessary or we can just get Invoker through Scheduler. It does incur extra cost to store an extra pointer.

chasm/lib/scheduler/invoker.go
yycptt: It's a bit hard to review when component and task executor implementations are in different PRs. But from the invoker task PR it looks like this will only be called once per transition. 

Also chasm validates all tasks added at the end and only generate tasks for those valid ones. The task validation logic in the other PR seems to suggest that there will be only one valid task at any time.

chasm/lib/scheduler/scheduler.go
```diff
@@ -0,0 +1,238 @@
...
+	invoker := NewInvoker(ctx, sched)
+	generator := NewGenerator(ctx, sched, invoker)
+	sched.Invoker = chasm.NewComponentField(ctx, invoker)
+	sched.Generator = chasm.NewComponentField(ctx, generator)
+
+	return sched
+}
+
+func (s *Scheduler) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
+	return chasm.LifecycleStateRunning
```
yycptt, line 81: hmmm it's always running? There's no operation to, say, stop/terminate the scheduler?

chasm/lib/scheduler/backfiller.go
lina-temporal: Makes sense - maybe it'd be better to have these as methods on *Scheduler, instead of independent constructors (or at least not exported constructors)?

chasm/lib/scheduler/backfiller.go
lina-temporal: Yeah, that's no longer needed. NextInvocationTime was used to facilitate re-generating tasks on demand from the framework. It's removed similarly from other components.

chasm/lib/scheduler/generator.go
lina-temporal: Sure, we could probably go through the scheduler instead. I'll update.

chasm/lib/scheduler/invoker.go
lina-temporal: Nope, will fix and do another pass for unneeded exports.

chasm/lib/scheduler/invoker.go
lina-temporal: +1, and yeah, this is called once per transition.

chasm/lib/scheduler/invoker.go
lina-temporal: In the case of pure tasks (the `InvokerProcessBufferTask`), the CHASM framework's translation to physical (temporal task executor queue) tasks handles deduplicating the logical task into one physical task. Tasks are also validated on transaction close to handle any obsoleted tasks.

Side effect tasks are similar in that they are validated on transaction close, but differ in that they are 1:1 with physical tasks. In this case, validation looks quite symmetric to the condition that adds the task to begin with (from the [invoker PR](https://github.com/temporalio/temporal/pull/8169)):

```go
func (e *InvokerExecuteTaskExecutor) Validate(
	_ chasm.Context,
	invoker *Invoker,
	_ chasm.TaskAttributes,
	_ *schedulespb.InvokerExecuteTask,
) (bool, error) {
	// If another execute task already happened to kick everything off, we don't need
	// this one.
	eligibleStarts := invoker.getEligibleBufferedStarts()
	valid := len(invoker.GetTerminateWorkflows())+
		len(invoker.GetCancelWorkflows())+
		len(eligibleStarts) > 0
	return valid, nil
}
```

.. so as soon as one of those side effect tasks executes, any logical tasks that would be unnecessary would invalidate themselves during transaction close (and when their physical task counterparts attempt execution, they'll fast-succeed from an outdated versioned transition).

chasm/lib/scheduler/scheduler.go
lina-temporal: Yeah, sorry, should have left a TODO here as well, I left it in the Generator PR:

```
		// TODO - add a timer to close this scheduler tree. V1 scheduler waits this
		// period before exiting, whereas we just stop pumping new tasks.
```

Maybe I could do something like have a timestamp field for expiration, like, return closed when `ctx.Now.After(s.ExpiresAt)` (or `RetainedUntil`)? Or I could add a timer task that flips a latch like `s.Expired`, such that the `LifecycleState` flips to closed during the close of the task's transaction? I'm not really sure which is the right approach for CHASM.

chasm/lib/scheduler/scheduler.go
lina-temporal: It's because I'm passing this function directly to [`chasm.UpdateComponent`](https://github.com/temporalio/temporal/blob/main/chasm/engine.go#L177), but I agree, it is unclear. Maybe it'd be better to fix the prototype here, and just wrap its usage on `UpdateComponent` with a little adaptor function.

chasm/lib/scheduler/scheduler.go
lina-temporal: Will fix.

chasm/lib/scheduler/invoker.go
lina-temporal: Nope, will update.

chasm/lib/scheduler/backfiller.go
yycptt: ~sorry what constructor?~

Oh you mean no constructor for Backfiller and make it a method on Scheduler? Yeah that makes sense to me.

chasm/lib/scheduler/scheduler.go
yycptt: hmm interesting. I saw your reply in the "Generate Task" PR. Can you give me a bit more context on the purpose of this Scheduler retention time? Is it just for making sure the retention is at least 7 days regardless of the namespace retention? Or there are other considerations. What if, say, CHASM allows components to specify a minimal retention time.

I am trying to see if there's anything here CHASM framework can take. But yeah a timer task to flip the latch makes sense.

lina-temporal: > Can you give me a bit more context on the purpose of this Scheduler retention time? Is it just for making sure the retention is at least 7 days regardless of the namespace retention? Or there are other considerations. 

Yeah, I think it's primarily a UX thing, so that customers always have a certain window in which they can check on a closed schedule. After the 7 day period, regardless of namespace retention, the schedule will now 404 for `DescribeSchedule` (even if the underlying workflow is still waiting for namespace retention timer). @dnr in case I'm missing another consideration. 

> What if, say, CHASM allows components to specify a minimal retention time.

I'm thinking, how about a core component that implements the retention timer task and latch, used in components like:

```go

type Scheduler struct {
  // ...
  Retention chasm.Field[*chasm.Retention]
}

func (s *Scheduler) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
        // If we have a retention component set, delegate to its state.
        retention, err := s.Retention.Get(ctx)
        if err != nil {
           return retention.LifecycleState(ctx)
        }

	return chasm.LifecycleStateRunning
}

// elsewhere..
if expired { 
  retention := chasm.NewRetention(ctx, 7 * time.Day)
  sched.Retention = chasm.NewComponentField(NewRetention(retention))
}
```

lina-temporal: I prototyped ^ and really liked the result, up for PR: 
https://github.com/temporalio/temporal/pull/8190

lina-temporal: We ended up deciding that the idle-close timer is probably specific enough to scheduler that we're going to keep it in the scheduler package. Will update here instead.

chasm/lib/scheduler/backfiller.go
bergundy: > I don't think we can have a default for root component. So probably make sense to just always enforce implementing that and remove the default LifecycleState impl in UnimplementedComponent?

Yeah, lets do that and also remove Unspecified since that is not a valid state.

chasm/lib/scheduler/generator.go
```diff
@@ -0,0 +1,32 @@
+package scheduler
+
+import (
+	"go.temporal.io/server/chasm"
+	schedulespb "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
```
bergundy: nit: no need to alias the import

chasm/lib/scheduler/invoker.go
bergundy: Seems like we would still benefit from framework support that ensures that there's only a single unique valid task of this type. Otherwise, you'd need to maintain some state transition counter and invalidate tasks that were generated before the last state transition where this task was generated.
It's not a big deal since the logic is idempotent but since we don't (or at least shouldn't) execute the task under a lock, it would be more likely that we get into situations where the same task is executing multiple times.
@yycptt

chasm/lib/scheduler/scheduler.go
```diff
@@ -0,0 +1,275 @@
...
+// tree will consist of 2 or more sub-components:
+// - Generator: buffers actions according to the schedule specification
+// - Invoker: executes buffered actions
+// - Backfiller: buffers actions according to requested backfills
+type Scheduler struct {
+	chasm.UnimplementedComponent
+
+	// Persisted internal state, consisting of state relevant to all components in
+	// the scheduler tree.
+	*schedulespb.SchedulerInternal
```
bergundy: nit: Can we not call this `Internal`? I think `SchedulerState` works better.
Also, can we not embed the struct directly without a pointer?
It should be possible to first construct the component and later load the state with (omitted the reflection logic):

```go
var comp Scheduler
proto.Unmarshal(data, &comp.SchedulerState)
```

chasm/lib/scheduler/scheduler.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Otherwise check and decrement limit.
+	if scheduleState.RemainingActions > 0 {
+		if decrement {
+			scheduleState.RemainingActions--
+
+			// The conflict token is updated because a client might be in the process of
+			// preparing an update request that increments their schedule's RemainingActions
+			// field.
+			s.updateConflictToken()
```
bergundy, line 149: The conflict token could have been calculated from the scheduler state to avoid having to remember to update it every time, which can lead to subtle bugs.

chasm/lib/scheduler/scheduler.go
```diff
@@ -0,0 +1,275 @@
...
+func (s Scheduler) overlapPolicy() enumspb.ScheduleOverlapPolicy {
+	policy := s.Schedule.Policies.OverlapPolicy
+	if policy == enumspb.SCHEDULE_OVERLAP_POLICY_UNSPECIFIED {
+		policy = enumspb.SCHEDULE_OVERLAP_POLICY_SKIP
+	}
+	return policy
+}
+
+func (s Scheduler) resolveOverlapPolicy(overlapPolicy enumspb.ScheduleOverlapPolicy) enumspb.ScheduleOverlapPolicy {
+	if overlapPolicy == enumspb.SCHEDULE_OVERLAP_POLICY_UNSPECIFIED {
```
bergundy: Why don't these methods take a pointer receiver? I'm surprised we don't have lint errors for this.

chasm/lib/scheduler/scheduler_suite_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"go.temporal.io/server/common/testing/testlogger"
+	"go.temporal.io/server/common/testing/testvars"
+	"go.temporal.io/server/service/history/tasks"
+	"go.uber.org/mock/gomock"
+)
+
+// schedulerSuite sets up a suite that has a basic CHASM tree ready
+// for use with the Scheduler library.
+type schedulerSuite struct {
+	suite.Suite
```
bergundy, line 23: Please no more suites :lolsob:

bergundy (commented):
I had mostly nits and stylistic comments. The code looks great.
It's hard to follow the logic with the code being split into these smaller PRs but I'm not sure a single PR would have been better 🤷.

chasm/lib/scheduler/scheduler.go
lina-temporal: Fixed.

chasm/lib/scheduler/scheduler.go
lina-temporal: > nit: Can we not call this Internal? I think SchedulerState works better.

Sure, I think State works better for this now that it's not overloaded with the HSM state. 

>  Also, can we not embed the struct directly without a pointer?

This breaks things that want to check for the `proto.Message` interface:

```
./invoker_tasks.go:132:42: in call to common.CloneProto, T (type schedulerpb.SchedulerInternal) does not satisfy "google.golang.org/protobuf/proto".Message **(method ProtoReflect has pointer receiver)**
```

chasm/lib/scheduler/backfiller.go
yycptt: https://github.com/temporalio/temporal/pull/8250

chasm/lib/scheduler/invoker.go
yycptt: yeah make sense. I have tracked this work item.

PR #8167, opened 2025-08-07
[Scheduled Actions] CHASM Backfill task

lina-temporal:
## What changed?
- Backfill task ported to CHASM. 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/backfiller_tasks.go
+++ b/chasm/lib/scheduler/backfiller_tasks.go
@@ -1,0 +1,237 @@
+package scheduler
+
+import (
+	"fmt"
+	"time"
+
+	"go.temporal.io/api/serviceerror"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
+	"go.uber.org/fx"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type (
+	BackfillerTaskExecutorOptions struct {
+		fx.In
+
+		Config         *Config
+		MetricsHandler metrics.Handler
+		BaseLogger     log.Logger
+		SpecProcessor  SpecProcessor
+	}
+
+	BackfillerTaskExecutor struct {
+		BackfillerTaskExecutorOptions
+	}
+)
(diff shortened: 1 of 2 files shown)


chasm/lib/scheduler/backfiller_tasks.go
```diff
@@ -0,0 +1,237 @@
...
+	return validateTaskHighWaterMark(backfiller.GetLastProcessedTime(), attrs.ScheduledTime)
+}
+
+func (b *BackfillerTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	backfiller *Backfiller,
+	_ chasm.TaskAttributes,
+	_ *schedulespb.BackfillerTask,
+) error {
+	defer func() { backfiller.Attempt++ }()
```
yycptt, line 54: do we need to check Attempt in `Validate()`, I see there's a TODO for that in today's hsm implementation.

chasm/lib/scheduler/backfiller_tasks.go
```diff
@@ -0,0 +1,237 @@
...
+
+	scheduler, err := backfiller.Scheduler.Get(ctx)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Scheduler tree missing node"),
+			err)
+	}
+	logger := newTaggedLogger(b.BaseLogger, scheduler)
```
yycptt, line 62: I am thinking if CHASM framework should provide a logger via maybe the Context and tag the entity key and shardID by default. 

Component has no way to get shardID and looks like have to store the entity key (namespace & business ID) as part of it's state.

chasm/lib/scheduler/backfiller_tasks.go
lina-temporal: I don't believe so because we validate the task's scheduled time versus the `LastProcessedTime` high water mark, and any time we record an attempt, we'd also move `LastProcessedTime` forward as well.

chasm/lib/scheduler/backfiller_tasks.go
lina-temporal: > I am thinking if CHASM framework should provide a logger via maybe the Context and tag the entity key and shardID by default.

I like this! Lemme take a stab at prototyping something.

bergundy (approved):

PR #8168, opened 2025-08-07
[Scheduled Actions] CHASM Generate Task

lina-temporal:
## What changed?
- Port the generate task to CHASM.
- I have a TODO here for adding a timer task to cleanup after the retention period, which would update the scheduler's `LifecycleState`; I meant to discuss this at some point to see if that's the pattern we want in CHASM-world, would love to get thoughts on that in the comments.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/generator_tasks.go
+++ b/chasm/lib/scheduler/generator_tasks.go
@@ -1,0 +1,155 @@
+package scheduler
+
+import (
+	"fmt"
+	"time"
+
+	"go.temporal.io/api/serviceerror"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
+	"go.uber.org/fx"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type (
+	GeneratorTaskExecutorOptions struct {
+		fx.In
+
+		Config         *Config
+		MetricsHandler metrics.Handler
+		BaseLogger     log.Logger
+		SpecProcessor  SpecProcessor
+	}
+
+	GeneratorTaskExecutor struct {
+		GeneratorTaskExecutorOptions
+	}
+)
+
+func NewGeneratorTaskExecutor(opts GeneratorTaskExecutorOptions) *GeneratorTaskExecutor {
+	return &GeneratorTaskExecutor{
+		GeneratorTaskExecutorOptions: opts,
+	}
+}
+
+func (g *GeneratorTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	generator *Generator,
+	_ chasm.TaskAttributes,
+	_ *schedulespb.GeneratorTask,
+) error {
+	scheduler, err := generator.Scheduler.Get(ctx)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Scheduler tree missing node"),
+			err)
+	}
+	logger := newTaggedLogger(g.BaseLogger, scheduler)
+
+	invoker, err := generator.Invoker.Get(ctx)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Scheduler tree missing node"),
+			err)
+	}
+
+	// If we have no last processed time, this is a new schedule.
+	if generator.LastProcessedTime == nil {
+		createdAt := timestamppb.New(ctx.Now(generator))
+		generator.LastProcessedTime = createdAt
+		scheduler.Info.CreateTime = createdAt
+
+		g.logSchedule(logger, "Starting schedule", scheduler)
+	}
+
+	// Process time range between last high water mark and system time.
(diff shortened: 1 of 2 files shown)


chasm/lib/scheduler/generator_tasks.go
```diff
@@ -0,0 +1,155 @@
...
+	// Enqueue newly-generated buffered starts.
+	if len(result.BufferedStarts) > 0 {
+		invoker.EnqueueBufferedStarts(ctx, result.BufferedStarts)
+	}
+
+	// Write the new high water mark.
+	generator.LastProcessedTime = timestamppb.New(result.LastActionTime)
+
+	_, inRetention := g.getRetentionExpiration(ctx, scheduler, result.NextWakeupTime)
+	if inRetention {
```
yycptt: chasm framework should be able to block the task with the access rule?

lina-temporal: ~~I agree, but what about closing off the scheduler (e.g., so it no longer shows in the frontend)?~~

Oops sorry saw your other comment afterwards. I replied on the scheduler PR. Yeah, scheduler retention is a little bit different, we hold it open so it still shows up on the schedule page for 7 days before we let the WF close and do its regular cleanup (which is probably what the CHASM cleanup is more analogous to).

yycptt: Let's continue the discussion in https://github.com/temporalio/temporal/pull/8165#discussion_r2268010797. Resolving this one.

chasm/lib/scheduler/generator_tasks.go
```diff
@@ -0,0 +1,155 @@
...
+		// We're in retention, no need for another buffer task.
+		//
+		// TODO - add a timer to close this scheduler tree. V1 scheduler waits this
+		// period before exiting, whereas we just stop pumping new tasks.
+		return nil
+	}
+
+	// No more tasks if we're paused.
```
bergundy, line 109: Shouldn't this be the first check before processing the time range?

bergundy: @yycptt seems like we're implementing paused in "user-land" here and we already know we want framework support for pausing. Can we add this functionality to the framework and get ahead?

chasm/lib/scheduler/generator_tasks.go
```diff
@@ -0,0 +1,155 @@
...
+		ScheduledTime: result.NextWakeupTime,
+	}, &schedulespb.GeneratorTask{})
+
+	return nil
+}
+
+func (g *GeneratorTaskExecutor) logSchedule(logger log.Logger, msg string, scheduler *Scheduler) {
+	// Log spec as json since it's more readable than the Go representation.
+	specJson, _ := protojson.Marshal(scheduler.Schedule.Spec)
+	policiesJson, _ := protojson.Marshal(scheduler.Schedule.Policies)
```
bergundy: nit: can we avoid serialization if the logger is not enabled for debug?

chasm/lib/scheduler/generator_tasks.go
yycptt: Synced offline. yes we will look into adding pause into the framework later on. Scheduler will continue to do it in user land due to resource and time constraint.

chasm/lib/scheduler/generator_tasks.go
lina-temporal: Yep, will fix.

chasm/lib/scheduler/generator_tasks.go
```diff
@@ -0,0 +1,155 @@
...
+		//
+		// Once the idle timer expires, we close the component.
+		ctx.AddTask(scheduler, chasm.TaskAttributes{
+			ScheduledTime: idleDuration,
+		}, &schedulespb.SchedulerIdleTask{})
+		return nil
+	}
+
+	// No more tasks if we're paused.
+	if scheduler.Schedule.State.Paused {
```
bergundy, line 110: Shouldn't we prevent processing altogether if we are paused? E.g. this check should be at the top of this function?

chasm/lib/scheduler/scheduler_tasks.go
```diff
@@ -0,0 +1,31 @@
...
+	return nil
+}
+
+func (r *SchedulerIdleTaskExecutor) Validate(
+	ctx chasm.Context,
+	scheduler *Scheduler,
+	_ chasm.TaskAttributes,
+	_ *schedulespb.SchedulerIdleTask,
+) (bool, error) {
+	return !scheduler.Closed, nil
```
bergundy, line 50: If what I remember about the semantics is true, the scheduler can turn from idle to not idle, if that's the case, the task should only be valid if the scheduler's calculated idle time is the same as this task's deadline and it has not been closed yet.

chasm/lib/scheduler/generator_tasks.go
lina-temporal: The processing makes sure that the high watermark keeps moving forward while paused, skipping time ranges as needed. There's a comment for this in `SpecProcessorImpl.ProcessTimeRange`, but I'll note that here, too.

chasm/lib/scheduler/scheduler_tasks.go
lina-temporal: Good catch, thanks!

PR #8169, opened 2025-08-08
[Scheduled Actions] CHASM Invoker tasks

lina-temporal:
## What changed?
- Port Invoker tasks to CHASM.
- Invoker is now mercifully split into two (still large) parts: `invokerExecuteTask` and `invokerProcessBufferTask`
- I wanted to do the 1:1 port from what HSM had before I started adding in CHASM visibility and workflow watcher; I'll be following with both of those. 
- I'm considering just getting rid of the local rate limiting hooks - I don't think we plan to make use of them at this point. 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/invoker_tasks.go
+++ b/chasm/lib/scheduler/invoker_tasks.go
@@ -1,0 +1,610 @@
+package scheduler
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+	schedulepb "go.temporal.io/api/schedule/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/server/api/historyservice/v1"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/resource"
+	"go.temporal.io/server/common/util"
+	scheduler1 "go.temporal.io/server/service/worker/scheduler"
+	"go.uber.org/fx"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type (
+	InvokerTaskExecutorOptions struct {
+		fx.In
+
+		Config         *Config
+		MetricsHandler metrics.Handler
+		BaseLogger     log.Logger
+		HistoryClient  resource.HistoryClient
+		FrontendClient workflowservice.WorkflowServiceClient
+	}
+
+	InvokerExecuteTaskExecutor struct {
+		InvokerTaskExecutorOptions
+	}
+
+	InvokerProcessBufferTaskExecutor struct {
+		InvokerTaskExecutorOptions
+	}
+
+	// Per-task context.
+	invokerTaskExecutorContext struct {
+		context.Context
+
+		actionsTaken int
+		maxActions   int
+	}
+
+	rateLimitedError struct {
+		// The requested interval to delay processing by rescheduilng.
+		delay time.Duration
+	}
+)
+
+const (
+	// Lower bound for the deadline in which buffered actions are dropped.
+	startWorkflowMinDeadline = 5 * time.Second
+
+	// Because the catchup window doesn't apply to a manual start, pick a custom
+	// execution deadline before timing out a start.
+	manualStartExecutionDeadline = 1 * time.Hour
+
+	// Upper bound on how many times starting an individual buffered action should be retried.
+	InvokerMaxStartAttempts = 10 // TODO - dial this up/remove it
+)
+
+var (
+	errRetryLimitExceeded       = errors.New("retry limit exceeded")
+	_                     error = &rateLimitedError{}
+)
+
+func NewInvokerExecuteTaskExecutor(opts InvokerTaskExecutorOptions) *InvokerExecuteTaskExecutor {
+	return &InvokerExecuteTaskExecutor{
+		InvokerTaskExecutorOptions: opts,
+	}
+}
+
+func NewInvokerProcessBufferTaskExecutor(opts InvokerTaskExecutorOptions) *InvokerProcessBufferTaskExecutor {
+	return &InvokerProcessBufferTaskExecutor{
+		InvokerTaskExecutorOptions: opts,
+	}
+}
+
+func (e *InvokerExecuteTaskExecutor) Validate(
+	_ chasm.Context,
+	invoker *Invoker,
+	_ chasm.TaskAttributes,
+	_ *schedulespb.InvokerExecuteTask,
+) (bool, error) {
+	// If another execute task already happened to kick everything off, we don't need
+	// this one.
+	eligibleStarts := invoker.getEligibleBufferedStarts()
+	valid := len(invoker.GetTerminateWorkflows())+
+		len(invoker.GetCancelWorkflows())+
+		len(eligibleStarts) > 0
+	return valid, nil
+}
+
+func (e *InvokerExecuteTaskExecutor) Execute(
+	ctx context.Context,
+	invokerRef chasm.ComponentRef,
+	_ chasm.TaskAttributes,
+	_ *schedulespb.InvokerExecuteTask,
+) error {
+	var result executeResult
+
+	invoker, err := chasm.ReadComponent(
+		ctx,
+		invokerRef,
+		func(i *Invoker, _ chasm.Context, _ struct{}) (*Invoker, error) {
+			return i, nil
+		},
+		struct{}{},
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
+			err)
+	}
+
+	schedulerRef := chasm.ComponentRef{
+		EntityKey: invokerRef.EntityKey,
+	}
+	scheduler, err := chasm.ReadComponent(
+		ctx,
+		schedulerRef,
+		func(s *Scheduler, _ chasm.Context, _ struct{}) (*Scheduler, error) {
+			return s, nil
+		},
+		struct{}{},
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
+			err)
+	}
+	logger := newTaggedLogger(e.BaseLogger, scheduler)
+
+	// If we have nothing to do, we can return without any additional writes.
+	eligibleStarts := invoker.getEligibleBufferedStarts()
+	if len(invoker.GetTerminateWorkflows())+
+		len(invoker.GetCancelWorkflows())+
+		len(eligibleStarts) == 0 {
+		return nil
+	}
+
+	// Terminate, cancel, and start workflows. The result struct contains the
+	// complete outcome of all requests executed in a single batch.
+	ictx := e.newInvokerTaskExecutorContext(ctx, scheduler)
+	result = result.Append(e.terminateWorkflows(ictx, logger, scheduler, invoker.GetTerminateWorkflows()))
+	result = result.Append(e.cancelWorkflows(ictx, logger, scheduler, invoker.GetCancelWorkflows()))
+	sres, startResults := e.startWorkflows(ictx, logger, scheduler, eligibleStarts)
+	result = result.Append(sres)
+
+	// Record completed executions on the Invoker (internal state).
+	_, _, err = chasm.UpdateComponent(
+		ctx,
+		invokerRef,
+		(*Invoker).recordExecuteResult,
+		&result,
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to update component state"),
+			err)
+	}
+
+	// Record action results on the Scheduler (user-facing metrics).
+	_, _, err = chasm.UpdateComponent(
+		ctx,
+		schedulerRef,
+		(*Scheduler).recordActionResult,
+		&schedulerActionResult{Starts: startResults},
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to update component state"),
+			err)
+	}
+
+	return nil
+}
+
+// takeNextAction increments the context's actionTaken counter, returning true if
+// the action should be executed, and false if the task should instead yield.
+func (c *invokerTaskExecutorContext) takeNextAction() bool {
+	taken := c.actionsTaken
+	c.actionsTaken++
+	return taken < c.maxActions
+}
+
+// cancelWorkflows does a best-effort attempt to cancel all workflow executions provided in targets.
+func (e *InvokerExecuteTaskExecutor) cancelWorkflows(
+	ctx invokerTaskExecutorContext,
+	logger log.Logger,
+	scheduler *Scheduler,
+	targets []*commonpb.WorkflowExecution,
+) (result executeResult) {
+	for _, wf := range targets {
+		if !ctx.takeNextAction() {
+			break
+		}
+
+		err := e.cancelWorkflow(ctx, scheduler, wf)
+		if err != nil {
+			logger.Error("Failed to cancel workflow", tag.Error(err), tag.WorkflowID(wf.WorkflowId))
+			e.MetricsHandler.Counter(metrics.ScheduleCancelWorkflowErrors.Name()).Record(1)
+		}
+
+		// Cancels are only attempted once.
+		result.CompletedCancels = append(result.CompletedCancels, wf)
+	}
+	return
+}
+
+// terminateWorkflows does a best-effort attempt to terminate all workflow executions provided in targets.
+func (e *InvokerExecuteTaskExecutor) terminateWorkflows(
+	ctx invokerTaskExecutorContext,
+	logger log.Logger,
+	scheduler *Scheduler,
+	targets []*commonpb.WorkflowExecution,
+) (result executeResult) {
+	for _, wf := range targets {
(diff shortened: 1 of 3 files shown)


chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	schedulerRef := chasm.ComponentRef{
+		EntityKey: invokerRef.EntityKey,
+	}
+	scheduler, err := chasm.ReadComponent(
+		ctx,
+		schedulerRef,
+		func(s *Scheduler, _ chasm.Context, _ struct{}) (*Scheduler, error) {
+			return s, nil
```
yycptt: shall we have a deep copy here? same Q for the ReadComponent logic for invoker above.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
+			err)
+	}
+
+	schedulerRef := chasm.ComponentRef{
+		EntityKey: invokerRef.EntityKey,
+	}
```
yycptt: Please always use chasm.NewComponentRef(). I should probably hide the EntityKey field and provide it as a method instead.

chasm/lib/scheduler/invoker_tasks.go
yycptt: Also, can/should we return both scheduler and invoker in one transition?

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+		&result,
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to update component state"),
+			err)
+	}
+
+	// Record action results on the Scheduler (user-facing metrics).
+	_, _, err = chasm.UpdateComponent(
```
yycptt: Same Q here. We should do only one write operation here.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+		NamespaceId: scheduler.NamespaceId,
+		TerminateRequest: &workflowservice.TerminateWorkflowExecutionRequest{
+			Namespace:           scheduler.Namespace,
+			WorkflowExecution:   &commonpb.WorkflowExecution{WorkflowId: target.WorkflowId},
+			Reason:              "terminated by schedule overlap policy",
+			Identity:            scheduler.identity(),
+			FirstExecutionRunId: target.RunId,
+		},
+	}
+	_, err := e.HistoryClient.TerminateWorkflowExecution(ctx, request)
```
yycptt, line 566: Why terminate & cancel goes through history client instead frontend like start workflow does? (related to Action counting?)

chasm/lib/scheduler/invoker_tasks.go
bergundy: +1 to both. Wouldn't the invoker have a pointer to the scheduler already?

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+		struct{}{},
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
+			err)
+	}
+	logger := newTaggedLogger(e.BaseLogger, scheduler)
+
+	// If we have nothing to do, we can return without any additional writes.
```
bergundy: Shouldn't this be done at the framework level when we validate the task?

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+		len(eligibleStarts) == 0 {
+		return nil
+	}
+
+	// Terminate, cancel, and start workflows. The result struct contains the
+	// complete outcome of all requests executed in a single batch.
+	ictx := e.newInvokerTaskExecutorContext(ctx, scheduler)
+	result = result.Append(e.terminateWorkflows(ictx, logger, scheduler, invoker.GetTerminateWorkflows()))
+	result = result.Append(e.cancelWorkflows(ictx, logger, scheduler, invoker.GetCancelWorkflows()))
+	sres, startResults := e.startWorkflows(ictx, logger, scheduler, eligibleStarts)
```
bergundy: Can we run as many of these requests concurrently? Is there a reason to run sequentially?

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	return nil
+}
+
+// takeNextAction increments the context's actionTaken counter, returning true if
+// the action should be executed, and false if the task should instead yield.
+func (c *invokerTaskExecutorContext) takeNextAction() bool {
+	taken := c.actionsTaken
+	c.actionsTaken++
```
bergundy: Shouldn't we only take the next action if we can? Seems like we will overcount here.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
+			err)
+	}
+
+	// Make sure we have something to start. If not, we can clear the buffer and
+	// complete without adding more tasks.
+	executionInfo := scheduler.Schedule.Action.GetStartWorkflow()
+	if executionInfo == nil || len(invoker.GetBufferedStarts()) == 0 {
```
bergundy: That seems like it should be an error, `GetStartWorkflow` should never be nil.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,610 @@
...
+	"go.temporal.io/server/api/historyservice/v1"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/resource"
+	"go.temporal.io/server/common/util"
+	scheduler1 "go.temporal.io/server/service/worker/scheduler"
```
bergundy: Can we rename this import to something more informative?

yycptt:
> I agree that if we have at least one request that wasn't rate limited, we should record that but otherwise, if the backoff duration is short enough, we could consider bubbling up the error.

I am thinking that the framework can define a ErrTaskYield with a backoff duration for this case. Also allowing each task to have in-memory fields to temporarily record the progress without doing a write operation on the component (similar to the RetentionTimer task today where we record what stage the task is currently at).

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: > Can we run as many of these requests concurrently? Is there a reason to run sequentially?

The way this is structured doesn't suggest it, but `ProcessBuffer` won't ever actually have workflows to terminate, cancel, and start, simultaneously in the same run (only 1 of the 3). The way terminate/cancel work is that `Execute` makes the `CancelWorkflow`/`TerminateWorkflow` call, which triggers the workflow watcher callback mechanism, updating the Scheduler's `RunningWorkflows` list. `ProcessBuffer` looks at that list when determining whether to issue cancels, terminates, or starts, so once the callback finishes its transition, `ProcessBuffer` will begin to issue starts instead of cancels.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Sure, will do.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Will fix.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Yeah, will fix on both counts.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Yeah, it's done as part of Validate as well, I suppose this is just redundant. I'll remove it.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: In practice, it's just used to break out of the for loop, but it is a bit sloppy. I'll fix it.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Existing implementation allows for it, but I agree. Hmm, I wonder if there was a time when we didn't have validation on `Schedule.Action`?

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: It's because although `startWorkflow` needs to go through frontend for metering's intercepting to count the start, these two are free to go directly to history and skip the metering stack.

lina-temporal:
> Seems like any time we're getting rate limited, we're creating more state transitions where in some cases it may be more reasonable to let the task processing stack retry the current task. I agree that if we have at least one request that wasn't rate limited, we should record that but otherwise, if the backoff duration is short enough, we could consider bubbling up the error.

Yeah, I agree with that. 

> I am thinking that the framework can define a ErrTaskYield with a backoff duration for this case.

👍  Neat, I like that component authors can control the backoff themselves.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+				compiledSpec:       s.compiledSpec,
+			}
+
+			return struct{}{}, nil
+		},
+		nil,
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
```
yycptt: hmm not really an internal error right, can be a transient error like timeout when loading the component from DB say.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+			i.recordExecuteResult(ctx, &result)
+			s.recordActionResult(&schedulerActionResult{starts: startResults})
+
+			return struct{}{}, nil
+		},
+		nil,
+	)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to update component state"),
```
yycptt: ditto

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+func (e *InvokerProcessBufferTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	invoker *Invoker,
+	_ chasm.TaskAttributes,
+	_ *schedulespb.InvokerProcessBufferTask,
+) error {
+	scheduler, err := invoker.Scheduler.Get(ctx)
+	if err != nil {
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal("Failed to read component"),
```
yycptt: Inside a transaction it's probably fine for now to treat this as a internal error since all component are loaded. 
If we do partial load in the future, ^ assumption may not hold.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+func (r *rateLimitedError) Error() string {
+	return fmt.Sprintf("rate limited for %s", r.delay)
+}
+
+func (e *InvokerExecuteTaskExecutor) newInvokerTaskExecutorContext(
+	ctx context.Context,
+	scheduler *Scheduler,
+) invokerTaskExecutorContext {
+	tweakables := e.Config.Tweakables(scheduler.Namespace)
+	maxActions := tweakables.MaxActionsPerExecution
```
yycptt, line 631: what's the default value here? now that it's a transfer task, the timeout a lot tighter than an activity, 3s I remember. Maybe something like 5?

yycptt: I will see if I can find some time to allow storing some temp state in the task struct.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+			}
+
+			s, err := i.Scheduler.Get(ctx)
+			if err != nil {
+				return struct{}{}, err
+			}
+			scheduler = &Scheduler{
+				SchedulerInternal:  common.CloneProto(s.SchedulerInternal),
+				cacheConflictToken: s.cacheConflictToken,
+				compiledSpec:       s.compiledSpec,
```
bergundy, line 140: Can you confirm and add a comment saying that the compiled spec is never mutated, only replaced.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+
+	logger := newTaggedLogger(e.BaseLogger, scheduler)
+
+	// Terminate, cancel, and start workflows. The result struct contains the
+	// complete outcome of all requests executed in a single batch.
+	ictx := e.newInvokerTaskExecutorContext(ctx, scheduler)
+	result = result.Append(e.terminateWorkflows(ictx, logger, scheduler, invoker.GetTerminateWorkflows()))
+	result = result.Append(e.cancelWorkflows(ictx, logger, scheduler, invoker.GetCancelWorkflows()))
+	sres, startResults := e.startWorkflows(ictx, logger, scheduler, invoker.getEligibleBufferedStarts())
+	result = result.Append(sres)
```
bergundy, line 163: Could you please document some of what you said in this comment here?

> Can we run as many of these requests concurrently? Is there a reason to run sequentially?

The way this is structured doesn't suggest it, but ProcessBuffer won't ever actually have workflows to terminate, cancel, and start, simultaneously in the same run (only 1 of the 3). The way terminate/cancel work is that Execute makes the CancelWorkflow/TerminateWorkflow call, which triggers the workflow watcher callback mechanism, updating the Scheduler's RunningWorkflows list. ProcessBuffer looks at that list when determining whether to issue cancels, terminates, or starts, so once the callback finishes its transition, ProcessBuffer will begin to issue starts instead of cancels.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+}
+
+// terminateWorkflows does a best-effort attempt to terminate all workflow executions provided in targets.
+func (e *InvokerExecuteTaskExecutor) terminateWorkflows(
+	ctx invokerTaskExecutorContext,
+	logger log.Logger,
+	scheduler *Scheduler,
+	targets []*commonpb.WorkflowExecution,
+) (result executeResult) {
+	for _, wf := range targets {
```
bergundy, line 246: This should be able to run concurrently.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -0,0 +1,603 @@
...
+func (e *InvokerExecuteTaskExecutor) startWorkflows(
+	ctx invokerTaskExecutorContext,
+	logger log.Logger,
+	scheduler *Scheduler,
+	starts []*legacyschedulespb.BufferedStart,
+) (result executeResult, startResults []*schedulepb.ScheduleActionResult) {
+	metricsWithTag := e.MetricsHandler.WithTags(
+		metrics.StringTag(metrics.ScheduleActionTypeTag, metrics.ScheduleActionStartWorkflow))
+
+	for _, start := range starts {
```
bergundy, line 285: This should be able to run concurrently.

chasm/lib/scheduler/invoker_tasks.go
bergundy: Please make it an error.

chasm/lib/scheduler/invoker_tasks.go
bergundy: Can you put a comment in the code explaining this detail?
Later on we should discuss if we can enforce metering in the history service directly, not for this PR.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Fixed

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: I adjusted the default max action count per execution to 5, esp since we plan to support yielding to retry a single task instead of needing a new task per batch.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Confirmed, added a comment.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Sure - were you thinking of anything particularly fancier than a couple goroutines in a waitgroup? I think that's probably fine since our max action count is always :) going to be "fairly low".

lina-temporal: Done with a simple wg/mutex.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -121,15 +127,15 @@ func (e *InvokerExecuteTaskExecutor) Execute(
 		invokerRef,
 		func(i *Invoker, ctx chasm.Context, _ any) (struct{}, error) {
```
yycptt, line 128: nit: re. the `struct{}`, the intention of that is for the method to return something, i.e. the scheduler created here. and this method doesn't have to capture the variable defined on L115.

yycptt (approved):
It's a bit hard to review but I think I do see all the comments get addressed. 👍

PR #8171, opened 2025-08-08
Use handle for dynamicconfig.Key

dnr:
**Depends on #8147, review after that**

## What changed?
Use unique.Handle for dynamicconfig.Keys instead of strings.

## Why?
Force keys to always be lowercase so we don't have to convert at lookup time, and also speed up lookups by interning.

## How did you test it?


prathyushpv (approved):
Nice!

PR #8172, opened 2025-08-08
allow nil/unversioned ramping version if current version is already nil/unversioned

carlydf:
## What changed?
allow nil/unversioned ramping version if current version is already nil/unversioned

## Why?
So that people can revert a task queue back to unversioned state without needing to set a 0% ramped non-nil Ramping Version.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


service/worker/workerdeployment/workflow.go
```diff
@@ -409,7 +409,8 @@ func (d *WorkflowRunner) validateStateBeforeAcceptingRampingUpdate(args *deploym
 		return temporal.NewApplicationError("conflict token mismatch", errFailedPrecondition)
 	}
 	//nolint:staticcheck // SA1019: worker versioning v0.31
-	if args.Version == d.State.GetRoutingConfig().GetCurrentVersion() {
+	if args.Version == d.State.GetRoutingConfig().GetCurrentVersion() &&
+		!(args.Version == worker_versioning.UnversionedVersionId && args.Percentage == 0) {
 		d.logger.Info("version can't be set to ramping since it is already current")
 		return temporal.NewApplicationError(fmt.Sprintf("requested ramping version %s is already current", args.Version), errFailedPrecondition)
```
yiminc, line 419: what does return error mean?

service/worker/workerdeployment/workflow.go
```diff
@@ -908,14 +909,17 @@ func (d *WorkflowRunner) syncVersion(ctx workflow.Context, targetVersion string,
 func (d *WorkflowRunner) syncUnversionedRamp(ctx workflow.Context, versionUpdateArgs *deploymentspb.SyncVersionStateUpdateArgs) error {
 	activityCtx := workflow.WithActivityOptions(ctx, defaultActivityOptions)
 
-	// DescribeVersion activity to get all the task queues in the current version
+	// DescribeVersion activity to get all the task queues in the current version, or the ramping version if current is nil
+	version := d.State.RoutingConfig.CurrentVersion //nolint:staticcheck // SA1019: worker versioning v0.31
+	if version == worker_versioning.UnversionedVersionId {
+		version = d.State.RoutingConfig.RampingVersion //nolint:staticcheck // SA1019: worker versioning v0.31
```
yiminc, line 962: why are we suppressing this lint role, and why is it specific to versioning v0.31?

carlydf: `RampingVersion` is deprecated. It's the "string representation" of a Worker Deployment Version. Externally we have switched to using a `WorkerDeploymentVersion{DeploymentName, BuildId}` struct, but unfortunately there was not enough time before Public Preview to remove the deprecated string from being used inside the system workflow.

We may need to wait until CHASM implementation to move away from the deprecated string representation of a version in a clean way.

I'm suppressing the linter error for using a deprecated field, and the deprecated field is related to versioning v0.31 (Replay 2025 Pre-release)

service/worker/workerdeployment/workflow.go
carlydf: Normally, we don't allow someone to set a version as the Ramping Version if that version is already the Current Version.

But, people want to be able to restore their Worker Deployment to "factory settings" which means `Current Version == Ramping Version == nil/unversioned && RampPercentage == 0`.

So, in the specific case where the requested Ramping Version is nil/unversioned and the requested RampPercentage is 0, this PR changes the code to allow Current Version == Ramping Version

service/worker/workerdeployment/workflow.go
Shivs11, line 962: if someone were to set a nil ramp with 0 percentage when ramping version is "" and current version is _unversioned_ (factory settings), we right now don't seem to have checks preventing this and this line that you have added shall equate version = '' (the empty string)

this shall further break further down the road when the activity tries describing a version

carlydf: I'll write a test for this situation and make sure it works! I think you're right about this bug. But IMO the correct place to handle it is in the update validator. Setting a nil ramp with 0 percent should be considered a no-op when ramping version is "" or when ramping version is "__unversioned__"

carlydf: wait, my test does test that situation

PR #8180, opened 2025-08-08
Support subscriptions to settings with constrained defaults

dnr:
## What changed?
Fill in support for subscriptions to dynamic config values with constrained defaults.

## Why?
We'd like to use this combination of functionality.

## How did you test it?
- [x] added new unit test(s)


dnr:
> LGTM, but it's hard to review with little context on this code.

I agree.. do you have any ideas for how to share the context better? I can explain in a zoom call? Maybe we need to nominate an official co-owner

PR #8182, opened 2025-08-09
Add support for protos in chasm libs

bergundy:
## What changed?

Added support for defining protos in chasm libs.

## Why?

Keep everything local to the library.

## How did you test it?
- [x] built


chasm/lib/activity/proto/v1/test.proto
bergundy, line 1: This is here just to demo, it should be reverted before merged.

chasm/lib/workflow/proto/v1/test.proto
bergundy, line 1: This is here just to demo, it should be reverted before merged.

chasm/lib/buf.yaml
```diff
@@ -0,0 +1,12 @@
+# buf rules for CHASM protos
```
lina-temporal, line 1: Minor, but does this need to be a separate config from the rest (maybe for excludes)? I'd assume that we'd mostly be using all of the same lints server-wide.

yycptt (approved):

bergundy:
CI will fail until we add some CHASM lib protos. I'll keep the PR open until the CHASM scheduler needs this.

PR #8185, opened 2025-08-12
Produce workflow_duration metric on completion

simvlad:
## What changed?
This PR produces the metric workflow_duration, when the workflow execution completes.

## Why?
Currently there is no metric that captures the duration of the workflow execution. It's also valuable to have the duration broken down by task queue, namespace, workflow type, which this PR enables

## How did you test it?
- [X] built
- [X] added new unit test(s)


service/history/workflow/transaction_impl.go
```diff
@@ -23,6 +25,9 @@ type (
 		namespaceState   string
 		workflowTypeName string
 		status           enumspb.WorkflowExecutionStatus
+		state            enumsspb.WorkflowExecutionState
```
yycptt: not sure why we need this. `Status` alone can tell if workflow is closed or not.

service/history/workflow/metrics.go
```diff
@@ -77,19 +77,26 @@ func emitMutableStateStatus(
...
 	)
 
-	switch status {
+	if completion.state == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED &&
+		completion.executionTime != nil && completion.closeTime != nil {
+
+		executionTime := completion.executionTime.AsTime()
+		closeTime := completion.closeTime.AsTime()
+		if closeTime.After(executionTime) {
+			metrics.WorkflowDuration.With(handler).Record(closeTime.Sub(executionTime))
```
yycptt: just double checking, it's intentional to use execution time instead of start time here?

yycptt: Per the design doc, this new metric should only be emitted for Workflows but not other archetypes. I think we should do the same for workflow success/cancel etc. metric below as well. 

You can check if mutable state is for a Workflow using the IsWorkflow() method of the MutableState interface, but I think some changes is needed to pass that information through the call chain.

PR #8186, opened 2025-08-12
Revert PR 8081 changes in service/frontend directory only

spkane31:
This reverts the batch operation changes from PR 8081 (#8081) but only for files in the service/frontend directory, restoring the original BatchParams-based implementation.

## What changed?
Revert #8081 frontend changes.

## Why?
This change is not backwards compatible, in the future we can turn on the protobuf workflow

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


rodrigozhou (approved):

service/frontend/workflow_handler.go
```diff
@@ -4824,7 +4965,9 @@ func (wh *WorkflowHandler) ListBatchOperations(
 		Namespace:     request.GetNamespace(),
 		PageSize:      request.PageSize,
 		NextPageToken: request.GetNextPageToken(),
-		Query: fmt.Sprintf("%s = '%s'",
+		Query: fmt.Sprintf("%s = '%s' and %s = '%s'",
+			searchattribute.WorkflowType,
+			batcher.BatchWFTypeName,
```
bergundy:
```suggestion
		Query: fmt.Sprintf("%s = '%s'",
```

bergundy: Let's make sure that we can capture any workflow type here for future compat.

PR #8189, opened 2025-08-12
[CHASM Scheduler] Move scheduler protobufs to scheduler/proto package

lina-temporal:
## What changed?
- CHASM scheduler protos are moved to live alongside the scheduler implementation code, within the `chasm` package.

## Why?
- See #8182. Sending this PR in advance, as that PR asserts protobufs were generated. 

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -0,0 +1,74 @@
...
+
+option go_package = "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb;schedulerpb";
+
+import "temporal/api/common/v1/message.proto";
+import "temporal/api/schedule/v1/message.proto";
+
+import "google/protobuf/timestamp.proto";
+
+// CHASM scheduler top-level state.
+message SchedulerInternal {
```
bergundy, line 13: Why not call this `Scheduler` or `SchedulerState` (same for the other structs)? Is it because we embed it as an exported field?

lina-temporal: It could be `SchedulerState` now, but it previously wasn't since that collided with the state machine enum.

PR #8190, opened 2025-08-13
[CHASM] Retention component

lina-temporal:
## What changed?
- Added a new reusable CHASM component, `Retention`, which simply updates its own LifecycleState to a given terminal state after a given duration elapses.

## Why?
- Schedules make use of this sort of timer to hold the workflow open for a period before the workflow actually closes, and the namespace-level workflow retention period takes effect. This is used to keep a completed schedule visible in the UI/frontend API for a period of time after it is completed. 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/retention/retention.go
+++ b/chasm/lib/retention/retention.go
@@ -1,0 +1,55 @@
+package retention
+
+import (
+	"time"
+
+	"go.temporal.io/api/serviceerror"
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+	"go.temporal.io/server/chasm"
+)
+
+// Retention timer component. This component can be used to close out a CHASM
+// component after a period of time has elapsed. This is a separate timer from the
+// namespace-level workflow/mutable state retention (which takes effect *after* the workflow has
+// closed, e.g., by this component).
+type Retention struct {
+	chasm.UnimplementedComponent
+
+	*persistencespb.RetentionData
+}
+
+// NewRetention creates a new Retention component whose LifecycleState will flip
+// to the given terminalState after the given duration has elapsed.
+func NewRetention(
+	ctx chasm.MutableContext,
+	duration time.Duration,
+	terminalState chasm.LifecycleState,
+) (*Retention, error) {
+	if !terminalState.IsClosed() {
+		return nil,
+			serviceerror.NewInternalf("terminalState isn't a 'Closed' state: %v", terminalState)
+	}
+
+	r := &Retention{
+		RetentionData: &persistencespb.RetentionData{
+			TerminalState: int64(terminalState),
+		},
+	}
+
+	// Adds a timer task to flip the latch to closed after the retention period
+	// elapses.
+	scheduledTime := ctx.Now(r).Add(duration)
+	ctx.AddTask(r, chasm.TaskAttributes{
+		ScheduledTime: scheduledTime,
+	}, &persistencespb.RetentionTask{})
+
+	return r, nil
+}
+
+func (r *Retention) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
+	if r.Closed {
+		return chasm.LifecycleState(r.TerminalState)
+	}
+
+	return chasm.LifecycleStateRunning
+}
--- a/proto/internal/temporal/server/api/persistence/v1/chasm.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/chasm.proto
@@ -101,3 +101,18 @@ message ChasmComponentRef {
     repeated string component_path = 6;
     VersionedTransition component_initial_versioned_transition = 7;
 }
+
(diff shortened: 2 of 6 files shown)


chasm/lib/retention/retention.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// Retention timer component. This component can be used to close out a CHASM
+// component after a period of time has elapsed. This is a separate timer from the
+// namespace-level workflow/mutable state retention (which takes effect *after* the workflow has
+// closed, e.g., by this component).
+type Retention struct {
+	chasm.UnimplementedComponent
+
+	*persistencespb.RetentionData
```
bergundy, line 18: This should live in the lib dir.

chasm/lib/retention/retention.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// NewRetention creates a new Retention component whose LifecycleState will flip
+// to the given terminalState after the given duration has elapsed.
+func NewRetention(
+	ctx chasm.MutableContext,
+	duration time.Duration,
+	terminalState chasm.LifecycleState,
```
bergundy, line 26: Can we make this a proto enum or have the equivalent enum defined in protos to avoid just storing it as an int?

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -101,3 +101,18 @@ message ChasmComponentRef {
...
+// TODO TODO TODO - move into retention/proto/v1 as soon as Roey's PR lands
+// 
+// Data for the Retention component.
+message RetentionData {
+    // True once the retention timer has fired and the component is now in its
+    // terminal state.
+    bool closed = 1;    
+
+    // The desired terminal chasm.LifecycleState (either Completed or Failed).
+    int64 terminal_state = 2;
```
bergundy, line 114: Should be an enum.

chasm/lib/retention/retention.go
bergundy: Why actually take this? Wouldn't we always want the terminal states to be closed states?

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -101,3 +101,18 @@ message ChasmComponentRef {
     repeated string component_path = 6;
     VersionedTransition component_initial_versioned_transition = 7;
 }
+
+// TODO TODO TODO - move into retention/proto/v1 as soon as Roey's PR lands
+// 
+// Data for the Retention component.
+message RetentionData {
```
bergundy, line 108: We can just call this retention IMHO, no need to use suffixes like Info and Data.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -101,3 +101,18 @@ message ChasmComponentRef {
...
+    // True once the retention timer has fired and the component is now in its
+    // terminal state.
+    bool closed = 1;
```
bergundy, line 111: This shouldn't be `closed` IMHO because an entity would always be in a closed state before the retention period kicks in.
It's more like `deleted` or `is_tombstone`.

chasm/lib/retention/retention.go
```diff
@@ -0,0 +1,55 @@
...
+	ctx.AddTask(r, chasm.TaskAttributes{
+		ScheduledTime: scheduledTime,
+	}, &persistencespb.RetentionTask{})
+
+	return r, nil
+}
+
+func (r *Retention) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
+	if r.Closed {
+		return chasm.LifecycleState(r.TerminalState)
```
bergundy, line 51: Shouldn't the timer impact the parent's lifecycle state?

chasm/lib/retention/retention.go
lina-temporal: To terminate eventually as either failed or succeeded.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
lina-temporal: > This shouldn't be closed IMHO because an entity would always be in a closed state before the retention period kicks in.

That's not true of how this is used for Scheduler.

chasm/lib/retention/retention.go
lina-temporal: > Shouldn't the timer impact the parent's lifecycle state?

Yes, the idea is that the parent component can delegate (on top of any of its own regular lifecycle logic) to this component.

yycptt (commented):
I think we can just define an interface and a timer task based on that interface without defining a separate component for this. 

The timer task can store the terminal state and work on any components that has, say, get and set lifecycle state.

lina-temporal:
Closing, we don't plan to reuse this capability across other features.

PR #8191, opened 2025-08-13
Change default for `component.nexusoperations.recordCancelRequestCompletionEvents`

pdoerner:
## What changed?
Changed default for `component.nexusoperations.recordCancelRequestCompletionEvents` to `true`

## Why?
Flag was added to ensure backwards compatibility. Now that 1.28 is released, can change the default. Flag will be removed after 1.29 is released.


bergundy (approved):
Approving since it seems that the last open source release should be compatible with this change.

PR #8196, opened 2025-08-14
Record various activity metrics

simvlad:
## What changed?
Changes activity metrics

### Deprecated Metrics

- **activity_e2e_latency** → Deprecated, as it measures individual attempts rather than true end-to-end latency, and replaced with clearer **activity_start_to_close_latency**

### New & Replacement Metrics

- **activity_start_to_close_latency**: Measures the latency from activity start to close (per attempt).
- **activity_schedule_to_close_latency**: Measures true end-to-end duration, including retries and backoff.
- **activity_success**: Counts the number of succeeded activities. Aligned with the existing **workflow_success** counter.
- **activity_fail**: Counts final failures for activities. Similar to **workflow_failed,** although it doesn’t include retries.
- **activity_timeout**: Incremented on the final activity timeout (including ScheduleToStartTimeout), tagged by `timeout_type`. Aligned with the existing **workflow_timeout** counter.
- **activity_task_fail**: Counts failures for activities including retries. Note that we don’t need to capture the number of retries, as this metric represents this number well.
- **activity_task_timeout**: Incremented on the activity attempt timeout (including ScheduleToStartTimeout), tagged by `timeout_type`.
- **activity_cancel:** Incremented when an activity is cancelled ****

## Why?
ActivityE2ELatency is inaccurate, since it measures only individual attempts. The other metrics are not recorded yet.

## How did you test it?
- [X] built
- [X] added new unit test(s)

Ran the server locally and checked the metrics in Prometheus:

Test 1: Activity with 5 attempts (1s each, 1s gap, all time out)

- activity_task_timeout = 5
- activity_timeout = 1
- activity_success = 0
- activity_fail = 0 (counted as timeout instead)
- activity_task_fail = 0 (counted as timeout instead)
- activity_schedule_to_close_latency_sum = 9s
- activity_end_to_end_latency_sum = 1s (deprecated) 
- activity_start_to_close_latency_sum = 1s

Test 2: Basic helloworld activity

- activity_success = 1

Test 3: Activity with 5 attempts (1s each, 1s gap, all fail except for the last one)
- activity_task_fail = 4
- activity_success = 1

### DIFF
--- a/service/history/timer_queue_active_task_executor.go
+++ b/service/history/timer_queue_active_task_executor.go
@@ -301,7 +301,7 @@ func (t *timerQueueActiveTaskExecutor) processSingleActivityTimeoutTask(
 		namespace.Name(mutableState.GetNamespaceEntry().Name()),
 		ai.TaskQueue,
 		workflow.ActivityCompletionMetrics{
-			State:              workflow.ActivityStateTimeout,
+			Status:             workflow.ActivityStatusTimeout,
 			AttemptStartedTime: timestamp.TimeValue(ai.StartedTime),
 			FirstScheduledTime: timestamp.TimeValue(ai.FirstScheduledTime),
 			Closed:             retryState != enumspb.RETRY_STATE_IN_PROGRESS,
--- a/service/history/timer_queue_active_task_executor.go
+++ b/service/history/timer_queue_active_task_executor.go
@@ -307,6 +307,9 @@ func (t *timerQueueActiveTaskExecutor) processSingleActivityTimeoutTask(
 			Closed:             retryState != enumspb.RETRY_STATE_IN_PROGRESS,
 			TimerType:          timerSequenceID.TimerType,
 		},
+		metrics.OperationTag(metrics.TimerActiveTaskActivityTimeoutScope),
+		metrics.WorkflowTypeTag(mutableState.GetWorkflowType().GetName()),
+		metrics.ActivityTypeTag(ai.ActivityType.GetName()),
 	)
 
 	if retryState == enumspb.RETRY_STATE_IN_PROGRESS {
--- a/service/history/timer_queue_active_task_executor.go
+++ b/service/history/timer_queue_active_task_executor.go
@@ -296,6 +296,19 @@ func (t *timerQueueActiveTaskExecutor) processSingleActivityTimeoutTask(
 		return result, nil
 	}
 
+	workflow.RecordActivityCompletionMetrics(
+		t.shardContext,
+		namespace.Name(mutableState.GetNamespaceEntry().Name()),
+		ai.TaskQueue,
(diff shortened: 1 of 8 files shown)


service/history/timer_queue_active_task_executor.go
```diff
@@ -296,6 +296,19 @@ func (t *timerQueueActiveTaskExecutor) processSingleActivityTimeoutTask(
...
+	workflow.RecordActivityCompletionMetrics(
+		t.shardContext,
+		namespace.Name(mutableState.GetNamespaceEntry().Name()),
+		ai.TaskQueue,
+		workflow.ActivityCompletionMetrics{
+			State:              workflow.ActivityStateTimeout,
+			AttemptStartedTime: timestamp.TimeValue(ai.StartedTime),
+			FirstScheduledTime: timestamp.TimeValue(ai.FirstScheduledTime),
+			Closed:             retryState != enumspb.RETRY_STATE_IN_PROGRESS,
+			TimerType:          timerSequenceID.TimerType,
```
nishkrishnan, line 308: no activity_type or workflow_type tags here?

simvlad: Good catch. Done:


activity_task_timeout{activityType="OrderProcessingActivity", instance="host.docker.internal:8000", job="services", namespace="default", operation="TimerActiveTaskActivityTimeout", service_name="history", taskqueue="timer", timeout_type="StartToClose", workflowType="SampleTimerWorkflow"}

service/history/workflow/metrics.go
```diff
@@ -131,3 +134,74 @@ func GetPerTaskQueueFamilyScope(
...
+		shard.GetConfig(),
+		tags...,
+	)
+
+	if !metricsState.AttemptStartedTime.IsZero() {
+		latency := time.Since(metricsState.AttemptStartedTime)
+		// ActivityE2ELatency is deprecated due to its inaccurate naming. It captures the attempt duration instead of an end-to-end duration as its name suggests. For now record both metrics
+		metrics.ActivityE2ELatency.With(metricsHandler).Record(latency)
+		metrics.ActivityStartToCloseLatency.With(metricsHandler).Record(latency)
+	}
```
yycptt: so start_to_close latency will be emitted in the case of timeout as well? That doesn't feel right to me as it's basically the timeout value and not the actual start to close latency from user side.

service/history/workflow/metrics.go
```diff
@@ -131,3 +134,74 @@ func GetPerTaskQueueFamilyScope(
 		tags...,
 	)
 }
+
+type ActivityExecutionState int
```
yycptt: nit: Shall we call this `Status` instead of `State`, to be consistent with the definition of workflow status. WorkflowState means something else.

yycptt (approved):

PR #8198, opened 2025-08-14
Reconstruct failure for forwarded Nexus completion requests

pdoerner:
## What changed?
When forwarding a `CompleteNexusOperation` HTTP request that contains a failure, the completion will be reconstructed instead of reusing the original request body.

## Why?
The Nexus SDK reads and closes the HTTP request body when the operation state is `failed` or `canceled` so we cannot reuse it for the forwarded request. For `successful` operations, the SDK just passes on the result content in the form of a `nexus.LazyValue` which we can forward directly since it is not read or closed.

## How did you test it?
new functional xdc tests

### DIFF
--- a/components/nexusoperations/frontend/handler.go
+++ b/components/nexusoperations/frontend/handler.go
@@ -264,10 +264,42 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 		}
 	}
 
-	forwardReq, err := http.NewRequestWithContext(ctx, r.HTTPRequest.Method, forwardURL, r.HTTPRequest.Body)
-	if err != nil {
-		h.Logger.Error("failed to construct forwarding HTTP request", tag.Operation(apiName), tag.WorkflowNamespace(rCtx.namespace.Name().String()), tag.Error(err))
-		return nexus.HandlerErrorf(nexus.HandlerErrorTypeInternal, "internal error")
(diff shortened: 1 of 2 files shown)


components/nexusoperations/frontend/handler.go
```diff
@@ -337,6 +368,19 @@ func isMediaTypeJSON(contentType string) bool {
...
 }
 
+// Copies HTTP request headers to Nexus headers except those starting with content- since those will be added by the client.
+func httpHeaderToNexusHeader(httpHeader http.Header) nexus.Header {
+	header := nexus.Header{}
+	for k, v := range httpHeader {
+		lowerK := strings.ToLower(k)
+		if strings.HasPrefix(lowerK, "content-") {
+			// Nexus headers can only have single values, ignore multiple values.
+			header[lowerK[len("content-"):]] = v[0]
```
bergundy: Why do we only take the content headers? Isn't it supposed to be the other way around?

components/nexusoperations/frontend/handler.go
bergundy, line 1: We should try out the new completion client and server once it's ready. It will simplify this code significantly.

components/nexusoperations/frontend/handler.go
pdoerner: Ah yeah I got that backwards. Good catch

PR #8204, opened 2025-08-18
Pass through original HTTP headers for forwarded Nexus requests

pdoerner:
## What changed?
When forwarding Nexus Start/Cancel requests, the original HTTP headers will be passed through without sanitization.

## Why?
Some headers that are still needed for the forwarded request may be sanitized during original request processing (e.g. authorization information headers).

## How did you test it?
existing tests


bergundy:
Worth testing that stripped headers are visible after forwarding.

PR #8207, opened 2025-08-18
[Scheduled Actions] Use the Execution returned from PollMutableState when calling GetWorkflowExecutionHistory

lina-temporal:
## What changed?
- In WatchWorkflow, we'll now use the Execution returned as a result from `PollMutableState`, instead of the Execution we used as part of the `PollMutableState` request.

## Why?
- We have a likely race condition if a workflow starts and completes during the `PollMutableState` call, where our originally-requested Execution is no longer the latest.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/worker/scheduler/activities.go
+++ b/service/worker/scheduler/activities.go
@@ -162,8 +162,13 @@ func (a *activities) tryWatchWorkflow(ctx context.Context, req *schedulespb.Watc
 
 	// get last event from history
 	histReq := &workflowservice.GetWorkflowExecutionHistoryRequest{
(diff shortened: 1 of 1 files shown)


carlydf (approved):
would it be doable / worth it to write an integration test that triggers this race condition by having an execution start and complete really quickly? I'm thinking probably not, because you'd need to control time in order to make sure that this activity yields to the CPU at exactly the right spot to trigger the race..

could this be a candidate for @stephanos testing efforts?

PR #8214, opened 2025-08-19
add command to dump dynamic config values

### LINKED ISSUE #421: Add admin command to dump dynamic config value[s]
Currently it is hard to know the current cluster configuration. Dumping all dynamic config values or at least a specific one would be very useful for troubleshooting.

vaibhavyadav-dev:
## What changed?
Added an admin command that dump all dynamic config values,

## Why?
Operators currently cannot easily inspect runtime dynamic configuration; dumping values aids troubleshooting, incident response, and verification after config changes.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

Closes #421

### DIFF
--- a/proto/internal/temporal/server/api/adminservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/adminservice/v1/request_response.proto
@@ -5,6 +5,7 @@ option go_package = "go.temporal.io/server/api/adminservice/v1;adminservice";
 
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
 
 import "temporal/api/enums/v1/common.proto";
 import "temporal/api/enums/v1/task_queue.proto";
@@ -601,4 +602,10 @@ message ForceUnloadTaskQueuePartitionRequest {
 
 message ForceUnloadTaskQueuePartitionResponse {
   bool was_loaded = 1;
+}
+
+message GetConfigurationsRequest {}
+message GetConfigurationsResponse{
+  string cluster_name = 1;
+  map<string, google.protobuf.Value> config_values = 2;
 }
--- a/proto/internal/temporal/server/api/adminservice/v1/service.proto
+++ b/proto/internal/temporal/server/api/adminservice/v1/service.proto
@@ -161,4 +161,6 @@ service AdminService {
     rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {}
 
     rpc ForceUnloadTaskQueuePartition (ForceUnloadTaskQueuePartitionRequest) returns (ForceUnloadTaskQueuePartitionResponse) {}
+
+    rpc GetConfigurations(GetConfigurationsRequest) returns (GetConfigurationsResponse) {}
 }
--- a/service/frontend/admin_handler.go
+++ b/service/frontend/admin_handler.go
@@ -2,12 +2,15 @@ package frontend
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"maps"
 	"math"
 	"net"
+	"reflect"
+	"regexp"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -61,6 +64,7 @@ import (
 	"go.temporal.io/server/service/worker/dlq"
 	"google.golang.org/grpc/health"
 	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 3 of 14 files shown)


dnr:
This isn't dumping all dynamic config values, it's only dumping the frontend config, and only for one particular (nonexistent) namespace. To do this properly you'd need to add support in the dynamicconfig.Client interface, but I'm not sure we want to do that. We could support getting the value of a single key by name without changing the interface.

What are you actually trying to use this for?

bergundy:
@dnr, @vaibhavyadav-dev joined the community slack and volunteered to contribute. I gave him [this](https://github.com/temporalio/temporal/issues/421) issue since I knew there's been a longstanding ask to add this API.

100% agree with the feedback so far. I can guide from here.

proto/internal/temporal/server/api/adminservice/v1/service.proto
```diff
@@ -161,4 +161,6 @@ service AdminService {
     rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {}
 
     rpc ForceUnloadTaskQueuePartition (ForceUnloadTaskQueuePartitionRequest) returns (ForceUnloadTaskQueuePartitionResponse) {}
+
+    rpc GetConfigurations(GetConfigurationsRequest) returns (GetConfigurationsResponse) {}
```
bergundy:
```suggestion
    rpc GetConfiguration(GetConfigurationRequest) returns (GetConfigurationResponse) {}
```

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -602,3 +603,9 @@ message ForceUnloadTaskQueuePartitionRequest {
 message ForceUnloadTaskQueuePartitionResponse {
   bool was_loaded = 1;
 }
+
+message GetConfigurationsRequest {}
+message GetConfigurationsResponse{
```
bergundy:
```suggestion
message GetConfigurationRequest {}
message GetConfigurationResponse {
```

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -602,3 +603,9 @@ message ForceUnloadTaskQueuePartitionRequest {
...
+  string cluster_name = 1;
+  map<string, google.protobuf.Value> config_values = 2;
```
bergundy:
```suggestion
repeated HostConfig host_config = 1;
```

And add a `HostConfig` message that will embed structured config via a [struct](https://protobuf.dev/reference/protobuf/google.protobuf/#struct):

```proto
message HostConfig {
  // The host that this configuration was loaded from.
  string hostname = 1;
  // The raw dynamic config loaded with all constraints.
  google.protobuf.Struct raw_dynamic_config = 2;
}
```

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
dnr: Please make all this stuff have the string "DynamicConfig" in it. There are two types of configuration and we should be explicit to avoid confusion. (I would even prefer GetDynamicConfig to GetDynamicConfiguration since that's how it's referred to all over the code.)

dnr:
> You are going to want to also add a `ListKeys()` method to the dynamic config client.
> 
> @dnr do you think this is a reasonable request?

I would rather not. I don't think we should add methods there without a very good reason, and I don't feel like this is it. I think we should support a single key query to start and then see if it satisfies people's use case. (Tbh I'm skeptical this is even useful at all but I guess it doesn't hurt.)

The problem is that it'll break the build for anyone with a custom Client until they add support. Also just ListKeys plus calling GetValue on each key is racy (a key may be added or removed between the List and the Get). So you want to get a snapshot of the whole state, but that may be expensive.

If we really do need to do that, we should add it as optional (i.e. interface-assert it when needed) to not interfere with existing Clients.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
vaibhavyadav-dev: okay

bergundy:
> If we really do need to do that, we should add it as optional (i.e. interface-assert it when needed) to not interfere with existing Clients.

Curious if there are known client implementations that we would be concerned with here or if this is mostly speculation.
Agree that the interface assert would be cleaner.
100% can go with a `Snapshot()` method if needed. I think that would be a reasonable approach.
I'm thinking that this API would be good for debugging, I would want this to evolve to a point where we could potentially spot misaligned configurations between the nodes and it may come in handy for when we migrate between self-hosted and cloud.

bergundy:
Lets make `GetConfigurationRequest` take a `repeated string dynamic_config_keys` and make the response include the `repeated HostConfig` as I've mentioned before but only for the current host, no need to fan out for now.

You will need to copy the definition of `Constraints` from `common/dynamicconfig/client.go` to `proto/internal/temporal/server/api/config/v1/message.proto` along with `HostConfig` and `ConstrainedValue` detailed below:

```proto
message ConstrainedValue {
  Constraints constraints = 1;
  google.protobuf.Struct value = 2;
}

message HostConfig {
  // The host that this configuration was loaded from.
  string hostname = 1;
  // The raw dynamic config loaded with all constraints.
  map<string, ConstrainedValue> dynamic_config = 2;
}
```

bergundy:
Sorry if I mislead you, each config name may have multiple constrained values, this would be a better representation:


```proto
message ConstrainedValue {
  Constraints constraints = 1;
  google.protobuf.Struct value = 2;
}

message ConstrainedValues {
  repeated ConstrainedValue items = 1;
}

message HostConfig {
  // The host that this configuration was loaded from.
  string hostname = 1;
  // The raw dynamic config loaded with all constraints.
  map<string, ConstrainedValues> dynamic_config = 2;
}
```

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -603,3 +604,37 @@ message ForceUnloadTaskQueuePartitionRequest {
...
+}
+
+message ConstrainedValue {
+  Constraints constraints = 1;
+  google.protobuf.Struct value = 2;
+}
+
+message ConstrainedValues {
+  repeated ConstrainedValue items = 1;
+}
```
bergundy: You can put this in `proto/internal/temporal/server/api/dynamicconfig/v1/message.proto`.

proto/internal/temporal/server/api/adminservice/v1/service.proto
```diff
@@ -161,4 +161,6 @@ service AdminService {
     rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {}
 
     rpc ForceUnloadTaskQueuePartition (ForceUnloadTaskQueuePartitionRequest) returns (ForceUnloadTaskQueuePartitionResponse) {}
+
+    rpc GetDynamicConfigurations(GetDynamicConfigurationsRequest) returns (GetDynamicConfigurationsResponse) {}
```
bergundy:
```suggestion
    rpc GetHostConfiguration(GetHostConfigurationRequest) returns (GetHostConfigurationResponse) {}
```

service/frontend/admin_handler.go
```diff
@@ -2180,6 +2184,138 @@ func (adh *AdminHandler) getDLQWorkflowID(
...
 
+func (adh *AdminHandler) GetDynamicConfigurations(
+	ctx context.Context,
+	req *adminservice.GetDynamicConfigurationsRequest,
+) (*adminservice.GetDynamicConfigurationsResponse, error) {
+	keys := req.GetDynamicConfigKeys()
+	pageSize := 100
+	var nextPageToken []byte
+
+	namespaces := make([]string, 0)
```
bergundy: You don't need to get the list of namespaces to implement this feature.

service/frontend/admin_handler.go
```diff
@@ -2180,6 +2184,138 @@ func (adh *AdminHandler) getDLQWorkflowID(
...
+	}
+
+	dynamicConfig := make(map[string]*adminservice.ConstrainedValues, len(keys))
+	for _, key := range keys {
+		constrainedValues := adh.getConstrainedValueForKey(key, namespaces)
+		if len(constrainedValues) > 0 {
+			dynamicConfig[key] = &adminservice.ConstrainedValues{
+				Items: constrainedValues,
+			}
+		}
```
bergundy: This would just be:

```
dcClient.GetValue(key)
```

Just make sure you have access to a `dynamicconfig.Client` instance in this handler.

vaibhavyadav-dev:
Closing this PR; opening a fresh one with a cleaner commit history.
New [PR](https://github.com/temporalio/temporal/pull/8676)

PR #8215, opened 2025-08-19
Convert Nexus HandlerError back to serviceerror for gRPC APIs

pdoerner:
## What changed?
For Nexus gRPC APIs, when the Nexus client returns a HandlerError that is the fault of the server, that error is converted back to a `serviceerror`

## Why?
So that metrics and SLAs are tracked appropriately

### DIFF
--- a/common/nexus/failure.go
+++ b/common/nexus/failure.go
@@ -351,53 +351,24 @@ func ConvertGRPCError(err error, exposeDetails bool) error {
 }
 
 // ConvertHandlerError converts a nexus.HandlerError returned by a Nexus client into a gRPC serviceerror.
-func ConvertHandlerError(err *nexus.HandlerError, exposeDetails bool) error {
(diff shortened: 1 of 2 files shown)


common/nexus/failure.go
```diff
@@ -338,6 +350,56 @@ func ConvertGRPCError(err error, exposeDetails bool) error {
 	return err
 }
 
+// ConvertHandlerError converts a nexus.HandlerError returned by a Nexus client into a gRPC serviceerror.
+func ConvertHandlerError(err *nexus.HandlerError, exposeDetails bool) error {
```
bergundy: Do we need `exposeDetails` here? I think we already control what to expose in the HTTP handler. Let's keep the control there.

pdoerner: I left it off originally then thought you would ask me to add it

PR #8216, opened 2025-08-19
Upgrade Go SDK to 1.35.0

bergundy:
Had to change WorkerDeploymentOptions and VersioningOverride to work with new SDK.

Changed the worker setup for the versioning internal workflow replay tests, so I generated a new set of workflow histories to test with. Normally we generate new workflow histories only when we've made a change to the workflow definitions that we want to test (ie. a new patch), but since the operations to generate the workflow histories had to change slightly, I think it makes sense to generate a fresh set of histories to replay-test with next time there is a change.


service/worker/workerdeployment/replaytester/worker/worker.go
```diff
@@ -78,63 +84,66 @@ func main() {
...
 	})
 	if err != nil {
 		log.Fatalln("Unable to set ramping version", err)
 	}
-	verifyDeployment(dHandle, "__unversioned__", deploymentName+".1.0", client.WorkerDeploymentVersionDrainageStatusUnspecified)
+	verifyDeployment(dHandle, "", build1, 1, client.WorkerDeploymentVersionDrainageStatusUnspecified)
 
-	// Unset the ramping version
+	// Set the ramp percent to 0
+	// TODO(carlydf): Once we allow it, test setting ramping version to nil while current version is also nil
```
carlydf, line 96: v1.35.0 of the SDK removes the ability to "unset" / "remove" the ramping version, you can only set it to nil (which now means unversioned) with a ramp percent of 0 for the same effect as what we previously called "unset."

We don't allow the ramping version of a worker deployment to be the same as the current version, so when the current version is nil, it is not allowed to "unset" the ramping version by setting it to nil, 0%

https://github.com/temporalio/temporal/pull/8172 fixes that issue (customers have requested to be able to have current version == ramping version == nil when ramp percent is 0, so that they can reset their Worker Deployment to "unversioned" settings)

service/worker/workerdeployment/replaytester/worker/worker.go
```diff
@@ -184,23 +193,52 @@ func main() {
...
+		if rv.BuildId != expectedRampingVersionBuildId {
+			log.Fatalln(fmt.Sprintf("Ramping version build id is %s not %s", rv.BuildId, expectedRampingVersionBuildId))
+		}
+	} else {
+		if expectedRampingVersionBuildId != "" {
+			log.Fatalln("Ramping version is empty, expected build id ", expectedRampingVersionBuildId)
+		}
+	}
+
+	if rp := describeResponse.Info.RoutingConfig.RampingVersionPercentage; rp != expectedRampPercentage {
```
carlydf, line 225: improved error messages here only because I ran into some dumb errors and wanted better messages

tests/versioning_3_test.go
```diff
@@ -1464,12 +1456,18 @@ func (s *Versioning3Suite) testChildWorkflowInheritance_ExpectInherit(crossTq bo
 	// v1 is current for both parent and child
 	s.setCurrentDeployment(tv1)
 
-	run, err := s.SdkClient().ExecuteWorkflow(ctx, sdkclient.StartWorkflowOptions{
+	startOpts := sdkclient.StartWorkflowOptions{
 		ID:                  tv1.WorkflowID(),
 		TaskQueue:           tv1.TaskQueue().GetName(),
-		VersioningOverride:  sdkOverride,
+		VersioningOverride:  nil,
```
carlydf, line 1461: If I make this value be the nil value of `var sdkOverride *sdkclient.PinnedVersioningOverride`, it causes a nil pointer panic [here](https://github.com/temporalio/sdk-go/blob/v1.35.0/internal/internal_workflow_execution_options.go#L137) in the Go SDK. A plain untyped nil as used above works. Is that a problem that should be fixed @Sushisource or is that expected?

PR #8217, opened 2025-08-20
Wire up api health monitor component

yux0:
## What changed?
Wire up api health monitor component

## Why?
The health monitor component did not wire correctly in fx

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/common/rpc/interceptor/health_check.go
+++ b/common/rpc/interceptor/health_check.go
@@ -40,6 +40,12 @@ type (
 	}
 )
 
+var excludedAPIsForHealthSignal = map[string]struct{}{
+	"DeepHealthCheck":             {},
+	"PollMutableState":            {},
+	"PollWorkflowExecutionUpdate": {},
+}
+
 // NewHealthCheckInterceptor creates a new health check interceptor
 func NewHealthCheckInterceptor(healthSignalAggregator HealthSignalAggregator) *HealthCheckInterceptor {
 	return &HealthCheckInterceptor{
@@ -60,10 +66,9 @@ func (h *HealthCheckInterceptor) UnaryIntercept(
 
 	// Skip health check recording for specific methods
 	methodName := api.MethodName(info.FullMethod)
-	if methodName != "DeepHealthCheck" { // Do not try to measure a tool with itself.
+	if _, ok := excludedAPIsForHealthSignal[methodName]; !ok {
 		h.healthSignalAggregator.Record(elapsed, err)
 	}
-
 	return resp, err
 }
 
--- a/service/history/fx.go
+++ b/service/history/fx.go
@@ -57,6 +57,7 @@ var Module = fx.Options(
 	fx.Provide(RateLimitInterceptorProvider),
 	fx.Provide(HealthSignalAggregatorProvider),
 	fx.Provide(HealthCheckInterceptorProvider),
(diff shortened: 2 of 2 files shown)


common/rpc/interceptor/health_check.go
```diff
@@ -40,6 +40,12 @@ type (
 	}
 )
 
+var excludedAPIsForHealthSignal = map[string]struct{}{
```
laniehei, line 44: nit: also mention we're using this in the UnaryIntercept package to make obvious to future maintainers

common/rpc/interceptor/health_check.go
```diff
@@ -40,6 +40,12 @@ type (
...
+	"DeepHealthCheck":             {},
+	"PollMutableState":            {},
+	"PollWorkflowExecutionUpdate": {},
```
prathyushpv, line 47: There are a few other Poll API like PollWorkflowTaskQueue, PollActivityTaskQueue etc. We have to cover those as well right? Polling APIs are marked by this field in method metadata: https://github.com/temporalio/temporal/blob/370f618c7035f2865823857606d8ddeffc36dd43/common/api/metadata.go#L24-L24

GetWorkflowExecutionHistory is a an API call that can either be polling or non-polling. To exclude polling GetWorkflowExecutionHistory, you will have to do something like : https://github.com/temporalio/temporal/blob/370f618c7035f2865823857606d8ddeffc36dd43/common/rpc/interceptor/telemetry.go#L135-L135

yux0: PollWorkflowTaskQueue and PollActivityTaskQueue are not in history role.

yux0: I guess I have to special handle GetWorkflowExecutionHistory. Is there any plan to split GetWorkflowExecutionHistory to get and poll two APIs?

prathyushpv: No, I don't think we have a plan to split it into two APIs. Thanks for making the change.

service/history/fx.go
```diff
@@ -185,13 +185,19 @@ func HealthSignalAggregatorProvider(
...
 func HealthCheckInterceptorProvider(
-	dynamicCollection *dynamicconfig.Collection,
 	healthSignalAggregator interceptor.HealthSignalAggregator,
 ) *interceptor.HealthCheckInterceptor {
 	return interceptor.NewHealthCheckInterceptor(
 		healthSignalAggregator,
 	)
 }
+
+func HistoryAdditionalInterceptorsProvider(
```
prathyushpv, line 195: I could not find where this provided interceptor is used in history service. Shouldn't we add this interceptor to this struct?: https://github.com/temporalio/temporal/blob/370f618c7035f2865823857606d8ddeffc36dd43/service/fx.go#L37-L37 

I might be missing how you added this interceptor to the service.

yux0: I added into the additional server interceptors param. We only do it in history service for now. If we decided to do this for other service roles. We declare one in GrpcServerOptionsParams.

prathyushpv (approved):

service/history/fx.go
prathyushpv: Got it. Thank you!

PR #8222, opened 2025-08-20
Customizable DataBlob codec

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


common/persistence/sql/common.go
```diff
@@ -13,41 +13,44 @@ import (
...
 	"go.temporal.io/server/common/persistence"
+	"go.temporal.io/server/common/persistence/serialization"
 	"go.temporal.io/server/common/persistence/sql/sqlplugin"
 )
 
 // TODO: Rename all SQL Managers to Stores
 type SqlStore struct {
-	Db     sqlplugin.DB
-	logger log.Logger
+	DB         sqlplugin.DB
```
stephanos, line 22: Linter made me do this.

common/persistence/serialization/blob.go
```diff
@@ -1,108 +0,0 @@
...
-}
-
-func QueueStateToBlob(info *persistencespb.QueueState) (*commonpb.DataBlob, error) {
-	return ProtoEncode(info)
-}
-
-func QueueStateFromBlob(data *commonpb.DataBlob) (*persistencespb.QueueState, error) {
-	result := &persistencespb.QueueState{}
-	return result, Decode(data, result)
-}
```
stephanos, line 108: ^ moved all this into `serializer.go`.

common/persistence/serialization/serializer.go
```diff
@@ -11,94 +11,90 @@ import (
...
+var DefaultDecoder Decoder = NewSerializerWithEncoding(enumspb.ENCODING_TYPE_UNSPECIFIED)
+
 type (
-	// Serializer is used by persistence to serialize/deserialize objects
-	// It will only be used inside persistence, so that serialize/deserialize is transparent for application
-	Serializer interface {
+	// Encoder is used to encode objects to DataBlobs.
+	Encoder interface {
 		SerializeEvents(batch []*historypb.HistoryEvent) (*commonpb.DataBlob, error)
-		DeserializeEvents(data *commonpb.DataBlob) ([]*historypb.HistoryEvent, error)
```
stephanos, line 23: Moved deserialize methods to own interface.

common/persistence/serialization/serializer.go
```diff
@@ -120,17 +116,23 @@ type (
...
 		Marshal() ([]byte, error)
 	}
 )
 
-// NewSerializer returns a PayloadSerializer
+// NewSerializer returns a Serializer that picks the encoding type based on an environment variable.
+// If none is set, it defaults to the Proto3 codec. See EncodingTypeFromEnv for details.
 func NewSerializer() Serializer {
-	return &serializerImpl{}
+	return NewSerializerWithEncoding(EncodingTypeFromEnv())
```
stephanos, line 130: The benefit from this approach that it'll always be the one from the env which I think is beneficial since there's no question of whether or not the env var takes effect.

common/persistence/history_branch_util.go
```diff
@@ -73,29 +78,25 @@ func (u *HistoryBranchUtilImpl) NewHistoryBranch(
 func (u *HistoryBranchUtilImpl) ParseHistoryBranchInfo(
 	branchToken []byte,
 ) (*persistencespb.HistoryBranch, error) {
-	return serialization.HistoryBranchFromBlob(&commonpb.DataBlob{Data: branchToken, EncodingType: enumspb.ENCODING_TYPE_PROTO3})
```
stephanos, line 76: So this was problematic since it hard-coded the encoding type as we are not storing it. I changed the interface instead to accept `[]byte` and use the encoding type of the current serializer. This way it can also be JSON, if that's configured.

service/history/handler.go
```diff
@@ -1808,9 +1809,10 @@ func (h *Handler) ReapplyEvents(ctx context.Context, request *historyservice.Rea
 	}
 
 	// deserialize history event object
+	eventsBlob := request.GetRequest().GetEvents()
 	historyEvents, err := h.payloadSerializer.DeserializeEvents(&commonpb.DataBlob{
-		EncodingType: enumspb.ENCODING_TYPE_PROTO3,
-		Data:         request.GetRequest().GetEvents().GetData(),
+		EncodingType: cmp.Or(eventsBlob.GetEncodingType(), enumspb.ENCODING_TYPE_PROTO3),
```
stephanos, line 1774: Changed this so it uses the stored encoding type, but if not defined falls back to the hard-coded one. It was too risky too remove this entirely since I don't know if it's safe to assume that all have an encoding type specified.

common/persistence/cassandra/errors_test.go
```diff
@@ -228,7 +228,7 @@ func (s *cassandraErrorsSuite) TestExtractCurrentWorkflowConflictError_Success()
 			},
 		},
 	}
-	blob, err := serialization.WorkflowExecutionStateToBlob(workflowState)
+	blob, err := serialization.NewSerializer().WorkflowExecutionStateToBlob(workflowState)
```
yycptt, line 231: nit: maybe we need a default serializer as well.

common/persistence/cassandra/errors.go
```diff
@@ -203,7 +203,7 @@ func extractCurrentWorkflowConflictError(
 		binary, _ := conflictRecord["execution_state"].([]byte)
 		encoding, _ := conflictRecord["execution_state_encoding"].(string)
 		executionState := &persistencespb.WorkflowExecutionState{}
-		if state, err := serialization.WorkflowExecutionStateFromBlob(p.NewDataBlob(binary, encoding)); err == nil {
+		if state, err := serialization.DefaultDecoder.WorkflowExecutionStateFromBlob(p.NewDataBlob(binary, encoding)); err == nil {
```
yycptt, line 212: should it use the serializer passed into the mutable state store?

common/persistence/cassandra/mutable_state_task_store.go
```diff
@@ -507,7 +509,7 @@ func (d *MutableStateTaskStore) PutReplicationTaskToDLQ(
 	request *p.PutReplicationTaskToDLQRequest,
 ) error {
 	task := request.TaskInfo
-	datablob, err := serialization.ReplicationTaskInfoToBlob(task)
+	datablob, err := d.serializer.ReplicationTaskInfoToBlob(task)
```
yycptt, line 512: cc @temporalio/cgs we should remove those old replication DLQ logic.

common/persistence/client/factory.go
```diff
@@ -197,7 +197,7 @@ func (f *factoryImpl) NewExecutionManager() (persistence.ExecutionManager, error
 		return nil, err
 	}
 
-	result := persistence.NewExecutionManager(store, f.serializer, f.eventBlobCache, f.logger, f.config.TransactionSizeLimit)
+	result := persistence.NewExecutionManager(store, f.serializer, serialization.NewTaskSerializer(f.serializer), f.eventBlobCache, f.logger, f.config.TransactionSizeLimit)
```
yycptt: hmm why we need to construct another TaskSerializer (and have a separate interface for that) instead of using f.serializer directly?

common/persistence/sql/sqlplugin/tests/history_current_execution.go
```diff
@@ -323,7 +323,7 @@ func (s *historyCurrentExecutionSuite) newRandomCurrentExecutionRow(
 			},
 		},
 	}
-	executionStateBlob, _ := serialization.WorkflowExecutionStateToBlob(executionState)
+	executionStateBlob, _ := serialization.NewSerializer().WorkflowExecutionStateToBlob(executionState)
```
yycptt, line 326: same, should use the serializer passed into persistence I think.

common/persistence/sql/execution_util.go
```diff
@@ -1191,7 +1190,7 @@ func workflowExecutionStateFromCurrentExecutionsRow(
 	row *sqlplugin.CurrentExecutionsRow,
 ) (*persistencespb.WorkflowExecutionState, error) {
 	if len(row.Data) > 0 && row.DataEncoding != "" {
-		return serialization.WorkflowExecutionStateFromBlob(p.NewDataBlob(row.Data, row.DataEncoding))
+		return serialization.DefaultDecoder.WorkflowExecutionStateFromBlob(p.NewDataBlob(row.Data, row.DataEncoding))
```
yycptt: make it a method and use `m.serializer` ?

common/persistence/sql/common.go
```diff
@@ -13,19 +13,22 @@ import (
...
+	serializer serialization.Serializer
 }
 
 func NewSqlStore(db sqlplugin.DB, logger log.Logger) SqlStore {
 	return SqlStore{
-		DB:     db,
-		logger: logger,
+		DB:         db,
+		logger:     logger,
+		serializer: serialization.NewSerializer(),
```
yycptt: why this serializer is hard coded and not passed down from caller, e.g. `NewSQLExecutionStore`?

common/persistence/xdc_cache.go
```diff
@@ -99,8 +99,7 @@ func NewEventsBlobCache(
 				Pin: false,
 			},
 		),
-		logger:     logger,
-		serializer: serialization.NewSerializer(),
```
yycptt, line 103: I think the `serializer` field is still being used

temporal/fx.go
```diff
@@ -584,6 +585,7 @@ func ApplyClusterMetadataConfigProvider(
 	logger = log.With(logger, tag.ComponentMetadataInitializer)
 	metricsHandler = metricsHandler.WithTags(metrics.ServiceNameTag(primitives.ServerService))
 	clusterName := persistenceClient.ClusterName(svc.ClusterMetadata.CurrentClusterName)
+	serializer := serialization.NewSerializer()
```
yycptt: do we need to get this serializer from fx graph?

temporal/server_impl.go
```diff
@@ -151,6 +152,7 @@ func initSystemNamespaces(
 ) error {
 	clusterName := persistenceClient.ClusterName(currentClusterName)
 	metricsHandler = metricsHandler.WithTags(metrics.ServiceNameTag(primitives.ServerService))
+	serializer := serialization.NewSerializer()
```
yycptt: same Q here

common/persistence/serialization/task_serializer.go
```diff
@@ -14,16 +14,28 @@ import (
...
+		DeserializeReplicationTask(replicationTask *persistencespb.ReplicationTaskInfo) (tasks.Task, error)
+	}
+	taskSerializer struct {
+		serializer Serializer
 	}
 )
 
-func NewTaskSerializer() *TaskSerializer {
-	return &TaskSerializer{}
+func NewTaskSerializer(serializer Serializer) TaskSerializer {
```
yycptt: Can you provide more context on this change?

common/persistence/cassandra/errors_test.go
stephanos: For testing-only or prod code as well?

It's true that right now there is only one serializer per process. But that could change?

I could certainly remove a lot of function parameters and fields if we had a global one; but I also want to generally avoid global constructs even if they are practically singletons right now. WDYT?

common/persistence/cassandra/errors.go
stephanos: I could pass it all the way down; but unless it's for testing purposes I'm not sure there's a big win there? The decoder only reacts to the data, ie it cannot be customized (as of right now). So in that sense it works the same way as the code before.

common/persistence/client/factory.go
stephanos: The `TaskSerializer` and `Serializer` are completely different. I think it makes sense to keep them separate.

`TaskSerializer`

```
SerializeTask(task tasks.Task) (*commonpb.DataBlob, error)
DeserializeTask(category tasks.Category, blob *commonpb.DataBlob) (tasks.Task, error)
[...]
```

`Serializer`

```
SerializeEvents(batch []*historypb.HistoryEvent) (*commonpb.DataBlob, error)
DeserializeEvents(data *commonpb.DataBlob) ([]*historypb.HistoryEvent, error)
[...]
```

The `TaskSerializer` is more specialized. See my other comment below.

common/persistence/sql/sqlplugin/tests/history_current_execution.go
stephanos: This seems to be a test suite (despite the name of the file). Give that, do you still think it should be passed in?

common/persistence/sql/common.go
stephanos: Ah, I tried to only use `NewSerializer()` in tests and fx. I must have missed this one.

common/persistence/xdc_cache.go
stephanos: Good catch; addressed in https://github.com/temporalio/temporal/pull/8222/commits/212bebdab32cedb66e499e842f447a9448c4222a!

common/persistence/serialization/task_serializer.go
stephanos: Sure! So before you'll see calls to `TimerTaskInfoFromBlob(blob)` etc in `TaskSerializer`. But these are now _methods_ on `Serializer`. So to invoke them now, you need a `Serializer`. Therefore it's `s.serializer.TimerTaskInfoFromBlob(blob)` now. And that's why we need the new parameter here.

common/persistence/sql/common.go
stephanos: Addressed in https://github.com/temporalio/temporal/pull/8222/commits/9abba24b1501be95c5ff0847c05d3c3f243f7497

temporal/server_impl.go
stephanos: I might be confused about the fx setup; but at this point we don't have fx yet, do we?

temporal/fx.go
stephanos: Maybe I'm a bit confused about the fx graph, I find it hard to follow. I've made this change, not sure it addresses your concern. https://github.com/temporalio/temporal/pull/8222/commits/8acd951cb5f2528f81aee7d09f2c97580bb45108

common/persistence/sql/execution_util.go
stephanos: Also addressed in https://github.com/temporalio/temporal/commit/9abba24b1501be95c5ff0847c05d3c3f243f7497; method wasn't quite feasible but it's passed as an argument now.

temporal/fx.go
stephanos: Also noticed one more; addressed in https://github.com/temporalio/temporal/pull/8222/commits/045e533c2c6e2db75e3f05843e5de529d3dd7512

common/persistence/client/factory.go
stephanos: I've added https://github.com/temporalio/temporal/pull/8222/commits/491108e18719ac3fa1d2e99a3840a2b28099e999 to only allocate it once now, though

common/persistence/cassandra/errors.go
yycptt: ah I see now, the default decode uses whatever encoding type in the blob.

common/persistence/cassandra/errors_test.go
yycptt: nvm, I had a wrong understanding re. the default decoder.

common/persistence/sql/sqlplugin/tests/history_current_execution.go
yycptt: ah ok, there's no upper layer for the test, lol

temporal/server_impl.go
yycptt: yeah your are right. I thought serializer is part of the top level module. I was wrong.

temporal/fx.go
yycptt: oh if we do that we should just make serializer a top level module, otherwise it's very confusing as it's provided again later in the common/resource module.

I am fine either way (revert the 045e533 and just create one inline, sorry!) or make it top level.

common/persistence/client/factory.go
yycptt: hmm I can see they are different, but not like "completely" different, they are still doing serialization/deserialization logic. 

I don't have strong opinion, just passing both Serializer and TaskSerializer together in bunch of places for the difference seems unnecessary to me.

stephanos: Okay, I see. I agree now that merging them isn't a bad idea. But at the same time; this PR is already long and old 😬 as long as I'm not making the code significantly worse I'd prefer to defer that. there are only a few uses of `TaskSerializer`

temporal/fx.go
stephanos: done 👍 https://github.com/temporalio/temporal/pull/8222/commits/21e1f5d270a5b9230fededa057b3f2018b68f0aa

common/persistence/serialization/serializer.go
```diff
@@ -56,6 +58,8 @@ type (
 		ArchivalTaskInfoToBlob(info *persistencespb.ArchivalTaskInfo) (*commonpb.DataBlob, error)
 		OutboundTaskInfoToBlob(info *persistencespb.OutboundTaskInfo) (*commonpb.DataBlob, error)
 		QueueStateToBlob(info *persistencespb.QueueState) (*commonpb.DataBlob, error)
+		SerializeTask(task tasks.Task) (*commonpb.DataBlob, error)
+		SerializeReplicationTask(task tasks.Task) (*persistencespb.ReplicationTaskInfo, error)
```
stephanos, line 62: Merged previous `TaskSerializer` and `ReplicationTaskSerializer` interface in here.

common/persistence/serialization/serializer.go
```diff
@@ -130,6 +136,50 @@ func NewSerializer() Serializer {
...
+	case tasks.CategoryIDReplication:
+		return deserializeReplicationTask(t, blob)
+	case tasks.CategoryIDArchival:
+		return deserializeArchivalTask(t, blob)
+	case tasks.CategoryIDOutbound:
+		return deserializeOutboundTask(t, blob)
+	default:
+		return nil, serviceerror.NewInternalf("Unknown task category: %v", category)
+	}
+}
```
stephanos, line 181: Moved these over from `task_serializer.go`.

common/persistence/serialization/task_serializers.go
```diff
@@ -14,101 +14,52 @@ import (
...
-	}
+// ReplicationTaskSerializer is an interface for serializing and deserializing replication tasks.
+// It is now part of the Encoder and Decoder interfaces.
+type ReplicationTaskSerializer interface {
+	SerializeReplicationTask(task tasks.Task) (*persistencespb.ReplicationTaskInfo, error)
+	DeserializeReplicationTask(replicationTask *persistencespb.ReplicationTaskInfo) (tasks.Task, error)
 }
 
-func (s *taskSerializer) serializeTransferTask(
+func serializeTransferTask(
```
stephanos, line 17: Made all of these in this file functions, ie helpers. Renamed file to `task_serializers` (added `s`) to make that a bit clearer.

common/persistence/serialization/task_serializers.go
```diff
@@ -14,101 +14,43 @@ import (
...
-type (
-	TaskSerializer interface {
-		ReplicationTaskSerializer
-		SerializeTask(task tasks.Task) (*commonpb.DataBlob, error)
-		DeserializeTask(category tasks.Category, blob *commonpb.DataBlob) (tasks.Task, error)
-	}
-	ReplicationTaskSerializer interface {
-		SeralizeReplicationTask(task tasks.Task) (*persistencespb.ReplicationTaskInfo, error)
-		DeserializeReplicationTask(replicationTask *persistencespb.ReplicationTaskInfo) (tasks.Task, error)
-	}
```
stephanos: Moved this one to `service/history/replication` as its the only place the interface is used/needed.

common/persistence/serialization/task_serializers.go
```diff
@@ -14,101 +14,43 @@ import (
...
-	taskSerializer struct {
-		serializer Serializer
-	}
```
stephanos: Removed this.

common/persistence/serialization/serializer.go
```diff
@@ -170,19 +210,23 @@ func (t *serializerImpl) DeserializeStrippedEvents(data *commonpb.DataBlob) ([]*
...
-	//nolint:exhaustive
 	switch data.EncodingType {
 	case enumspb.ENCODING_TYPE_PROTO3:
 		// Discard unknown fields to improve performance. StrippedHistoryEvents is usually deserialized from HistoryEvent
 		// which has extra fields that are not needed for this message.
 		err = proto.UnmarshalOptions{
 			DiscardUnknown: true,
 		}.Unmarshal(data.Data, events)
+	case enumspb.ENCODING_TYPE_JSON:
+		err = temporalproto.CustomJSONUnmarshalOptions{
```
yycptt, line 221: TIL we have this package.

service/history/replication/raw_task_converter.go
```diff
@@ -66,13 +67,15 @@ func NewSourceTaskConverter(
 	historyEngine historyi.Engine,
 	namespaceCache namespace.Registry,
 	serializer serialization.Serializer,
+	replicationTaskSerializer TaskSerializer,
```
yycptt, line 70: looks like this can be removed now?

common/persistence/serialization/task_serializers.go
```diff
@@ -14,90 +14,43 @@ import (
...
-		return s.deserializeReplicationTasks(blob)
-	case tasks.CategoryIDArchival:
-		return s.deserializeArchivalTasks(blob)
-	case tasks.CategoryIDOutbound:
-		return s.deserializeOutboundTask(blob)
-	default:
-		return nil, serviceerror.NewInternalf("Unknown task category: %v", category)
-	}
+// TaskSerializer is an interface for serializing and deserializing tasks.
+type TaskSerializer interface {
```
yycptt: nit: do we still need this interface in this case?

service/history/replication/raw_task_converter.go
stephanos: That one actually refers to `replication`'s own `TaskSerializer` from the package's `fx.go`.

common/persistence/serialization/task_serializers.go
stephanos: 👍 removed

PR #8223, opened 2025-08-20
Proof-of-Concept: Stuck Workflow Visibility

spkane31:
## What changed?
proof-of-concept for adding an SA after successive WFT failures OR WFT timeouts.

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


common/dynamicconfig/constants.go
```diff
@@ -978,6 +978,12 @@ to allow waiting on the "Accepted" lifecycle stage.`,
 		`FrontendEnableWorkerVersioningRuleAPIs enables worker versioning in workflow progress APIs.`,
 	)
 
+	FrontendEnableReportedProblemsSearchAttribute = NewNamespaceBoolSetting(
```
bergundy: This should be a system level config not a frontend one.

common/searchattribute/defs.go
```diff
@@ -92,6 +92,13 @@ const (
 	// execution. It is updated at workflow task completion when the server gets the
 	// behavior (`auto_upgrade` or `pinned`) from the SDK. Empty for unversioned workflows.
 	TemporalWorkflowVersioningBehavior = "TemporalWorkflowVersioningBehavior"
+
+	// TemporalReportedProblems is a search attribute that stores the information about problems
+	// the workflow has encountered in making progress. It is updated after successive workflow task
+	// failures with the last workflow task failure cause. After the workflow task is completed
+	// successfully, the search attribute is removed. Format of a single problem:
+	// "property:reportedProblem=<problem_type> property:reportedCause=<cause>".
```
bergundy:
```suggestion
	// "category=<problem_type> cause=<cause>".
```

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -297,6 +300,39 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
...
+					tag.WorkflowNamespaceID(namespaceEntry.ID().String()))
+
+				clearedPayload, err := searchattribute.EncodeValue(nil, enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST)
+				if err != nil {
+					return nil, err
+				}
+
+				fmt.Println("RespondWorkflowTaskCompleted:Invoke clearedPayload", clearedPayload)
+
+				if _, err := ms.AddUpsertWorkflowSearchAttributesEvent(currentWorkflowTask.ScheduledEventID, &commandpb.UpsertWorkflowSearchAttributesCommandAttributes{
```
bergundy: This wouldn't be done via a history event, and I would generally move all of this logic to the workflow task state machine and mutable state impl.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5872,54 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	// This is not guaranteed to be accurate because of ordering non-determinism. Needs to be fixed.
+	if proto.Equal(exeInfo.SearchAttributes[searchattribute.TemporalReportedProblems], reportedProblemsPayload) {
+		return nil // unchanged
+	}
+
+	ms.updateSearchAttributes(map[string]*commonpb.Payload{searchattribute.TemporalReportedProblems: reportedProblemsPayload})
+	return ms.taskGenerator.GenerateUpsertVisibilityTask()
+}
+
+func (ms *MutableStateImpl) removeReportedProblemsSearchAttribute() error {
```
bergundy: You would need a method to remove a specific cause, there could potentially be more than one in the future.
Also there could be multiple causes for the same category so you would probably need to maintain a mapping of problems (causes) to categories.

service/history/api/respondworkflowtaskcompleted/api.go
bergundy: I see there's a method already on mutable state to remove the search attribute.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5872,54 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	fmt.Println("removeReportedProblemsSearchAttribute")
+	reportedProblemsPayload, err := searchattribute.EncodeValue(nil, enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST)
+	if err != nil {
+		return err
+	}
+
+	exeInfo := ms.executionInfo
+	if exeInfo.SearchAttributes == nil {
+		exeInfo.SearchAttributes = make(map[string]*commonpb.Payload, 1)
```
```suggestion
		return nil
```

bergundy: And this condition can go to the top of the function.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5872,54 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	}
+
+	if proto.Equal(exeInfo.SearchAttributes[searchattribute.TemporalReportedProblems], reportedProblemsPayload) {
+		return nil
+	}
+
+	ms.updateSearchAttributes(map[string]*commonpb.Payload{searchattribute.TemporalReportedProblems: reportedProblemsPayload})
```
bergundy: You should just delete from the map.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5872,54 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	return ms.taskGenerator.GenerateUpsertVisibilityTask()
```
bergundy, line 5965: I think this method is idempotent and only a single task would be created but worth validating that assumption or ensuring that it works this way.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -810,10 +811,19 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskFailedEvent(
 		)
 	}
 
+	// One pathway for TemporalReportedProblems search attribute is here.
 	if err := m.ApplyWorkflowTaskFailedEvent(); err != nil {
 		return nil, err
 	}
 
+	// Set TemporalReportedProblems search attribute when reaching 2 consecutive failures.
+	if m.ms.executionInfo.WorkflowTaskAttempt == 2 {
```
bergundy: I think this should go in the `ApplyWorkflowTaskFailedEvent()` method (same for timeout below) to be applied in a passive cluster as well.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -907,6 +925,84 @@ func (m *workflowTaskStateMachine) failWorkflowTask(
...
+	req := &persistence.ReadHistoryBranchReverseRequest{
+		ShardID:                m.ms.shard.GetShardID(),
+		BranchToken:            branchToken,
+		MaxEventID:             maxEventID,
+		PageSize:               128, // small page; paginate if needed
+		LastFirstTransactionID: txID,
+		NextPageToken:          nil,
+	}
+	for {
+		resp, err := m.ms.shard.GetExecutionManager().ReadHistoryBranchReverse(ctx, req)
```
bergundy: You definitely do not want to incur persistence reads here. All of the work you do should be strictly using data stored in mutable state.

service/history/workflow/mutable_state_impl.go
spkane31: From my reading of the code: the method itself is not idempotent but the elasticsearch processor does deduplication so only a single visibility update is actually processed.

service/history/workflow/mutable_state_impl.go
spkane31: This method needs work because there can be multiple values in the list and we're only interested in removing certain ones, your comment elsewhere alluded to that.

service/history/api/respondworkflowtaskcompleted/api.go
spkane31: That is a private method, are there issues around creating new public methods on MutableState for just this one usecase?

service/history/workflow/workflow_task_state_machine.go
spkane31: Both `ApplyWorkflowTaskFailedEvent` and `ApplyWorkflowTaskTimedOutEvent` call this function directly. Putting this code here reduces duplication.

service/history/workflow/mutable_state_impl.go
spkane31: > Also there could be multiple causes for the same category so you would probably need to maintain a mapping of problems (causes) to categories.

I think we could use the same methodology we use to create the tag but in reverse. Find the last failure, which gives the cause and problem, then remove those two from the list.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5876,54 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+
+	exeInfo := ms.executionInfo
+	if exeInfo.SearchAttributes == nil {
+		exeInfo.SearchAttributes = make(map[string]*commonpb.Payload, 1)
+	}
+
+	// This is not guaranteed to be accurate because of ordering non-determinism. Needs to be fixed.
+	if proto.Equal(exeInfo.SearchAttributes[searchattribute.TemporalReportedProblems], reportedProblemsPayload) {
+		return nil
+	}
```
gow: Yes. You need to sort and compare the two.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5876,54 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+
+	ms.updateSearchAttributes(map[string]*commonpb.Payload{searchattribute.TemporalReportedProblems: reportedProblemsPayload})
+	return ms.taskGenerator.GenerateUpsertVisibilityTask()
+}
+
+func (ms *MutableStateImpl) RemoveReportedProblemsSearchAttribute(
+	reportedProbs ...string,
```
gow: `reportedProbs` is unused. Maybe not needed?

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -297,6 +297,12 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 		if err != nil {
 			return nil, err
 		}
+
+		if handler.config.EnableReportedProblemsSearchAttribute(nsName) {
+			if err := ms.RemoveReportedProblemsSearchAttribute("category=WorkflowTaskFailed", "cause=WorkflowWorkerUnhandledFailure"); err != nil {
```
gow: The param is string literal (also seems unused). Do you anticipate this method being called from other locations with a different value? If not, maybe cleanup the interface?

spkane31: For now I can clean this up, in the future if we expand what `TemporalReportedProblems` can have then we'll need to know exactly what to remove. That was the purpose of the values in the interface

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -907,6 +914,20 @@ func (m *workflowTaskStateMachine) failWorkflowTask(
 	}
 	m.retainWorkflowTaskBuildIdInfo(failWorkflowTaskInfo)
 	m.UpdateWorkflowTask(failWorkflowTaskInfo)
+
+	if failWorkflowTaskInfo.Attempt == 2 {
```
yycptt: oh so this means we only care about the failure of the first attempt? 
Then theoretically this should work with event based replication as well I think, because the first failed/timeout event will get replicated and applied on the target side.

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -297,6 +297,12 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 		if err != nil {
 			return nil, err
 		}
+
+		if handler.config.EnableReportedProblemsSearchAttribute(nsName) {
```
bergundy: Let's manage this search attribute always in `workflow_task_state_machine.go` so it's not spread across the entire codebase.

service/history/historybuilder/event_store.go
bergundy, line 1: Unrelated change?

service/history/timer_queue_active_task_executor.go
```diff
@@ -398,6 +398,7 @@ func (t *timerQueueActiveTaskExecutor) executeWorkflowTaskTimeoutTask(
 	scheduleWorkflowTask := false
 	switch task.TimeoutType {
 	case enumspb.TIMEOUT_TYPE_START_TO_CLOSE:
+		// WFT Timed Out path is here, add TemporalReportedProblems search attribute
```
bergundy: Same, keep the search attribute logic contained to the state machine.

service/history/workflow/mutable_state_impl.go
```diff
@@ -237,6 +237,10 @@ type (
 
 		// Tracks all events added via the AddHistoryEvent method that is used by the state machine framework.
 		currentTransactionAddedStateMachineEventTypes []enumspb.EventType
+
+		// TODO seankane: store the failure and cause of the last workflow task failure/timedout event
```
bergundy: This should go here FTR: https://github.com/temporalio/temporal/blob/5aedf512ec80d11a5d4b14546083915bb2c23186/proto/internal/temporal/server/api/persistence/v1/executions.proto#L102

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5876,105 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	if !ok {
+		return false
+	}
+
+	// Compare as multisets (element counts), order does not matter but counts must match
+	if len(keywordListA) != len(keywordListB) {
+		return false
+	}
+
+	counts := make(map[string]int, len(keywordListA))
```
bergundy: Technically this can be a `map[string]struct{}` since you should never have duplicates in the list.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5876,105 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	// This is not guaranteed to be accurate because of ordering non-determinism. Needs to be fixed.
+	if areKeywordListsEqual(exeInfo.SearchAttributes[searchattribute.TemporalReportedProblems], reportedProblemsPayload) {
+		return nil
+	}
+
+	ms.updateSearchAttributes(map[string]*commonpb.Payload{searchattribute.TemporalReportedProblems: reportedProblemsPayload})
+	return ms.taskGenerator.GenerateUpsertVisibilityTask()
+}
+
+func areKeywordListsEqual(a, b *commonpb.Payload) bool {
```
bergundy: nit: This helper can go somewhere in the visibility or common folders.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5876,105 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+		exeInfo.SearchAttributes = make(map[string]*commonpb.Payload, 1)
+	}
+
+	// This is not guaranteed to be accurate because of ordering non-determinism. Needs to be fixed.
+	if areKeywordListsEqual(exeInfo.SearchAttributes[searchattribute.TemporalReportedProblems], reportedProblemsPayload) {
+		return nil
+	}
+
+	ms.updateSearchAttributes(map[string]*commonpb.Payload{searchattribute.TemporalReportedProblems: reportedProblemsPayload})
+	return ms.taskGenerator.GenerateUpsertVisibilityTask()
```
bergundy, line 5945: Not for this PR (CC @yycptt, @rodrigozhou)

Since we already have the `visibilityUpdated` flag on mutable state, we shouldn't have to call this method ever.
Arguably, all of this checking of whether the search attribute changed or not should be done in `updateSearchAttributes`.

service/history/interfaces/mutable_state.go
```diff
@@ -362,5 +362,8 @@ type (
 
 		HasRequestID(requestID string) bool
 		SetSuccessorRunID(runID string)
+
+		UpdateReportedProblemsSearchAttribute(reportedProblemCategory, reportedCause string) error
+		RemoveReportedProblemsSearchAttribute() error
```
bergundy: You should only call this from `workflow_task_state_machine`, which takes the `Impl`, no need to add to the interface.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5872,6 +5876,105 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
 	return ms.taskGenerator.GenerateUpsertVisibilityTask()
 }
 
+func (ms *MutableStateImpl) UpdateReportedProblemsSearchAttribute(
+	reportedProblemCategory, reportedCause string,
```
bergundy: You already know what the category and the cause is, they were previously set.

```suggestion
```

service/history/workflow/workflow_task_state_machine.go
bergundy: Can you confirm that this gets triggered on the second time a task fails and the SDK drops the response? Ideally via a functional test.

bergundy: Let's also make the attempt number configurable (see comment above).

common/dynamicconfig/constants.go
```diff
@@ -276,6 +276,12 @@ operator API calls (highest priority). Should be >0.0 and <= 1.0 (defaults to 20
 		`How many extra goroutines can be created per root.`,
 	)
 
+	EnableReportedProblemsSearchAttribute = NewNamespaceBoolSetting(
```
bergundy: Make this an IntSetting, where `0` is turned off and anything above that would count the number of failed attempts that would trigger setting the search attribute. Then change the config name to `NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute` or something similar.

service/history/workflow/workflow_task_state_machine.go
bergundy: Can you also verify that this doesn't get set on transient workflow task failures?

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+	s.scheduleToCloseTimeout = 30 * time.Minute
+	s.startToCloseTimeout = 15 * time.Minute
+
+	s.activityRetryPolicy = &temporal.RetryPolicy{
+		InitialInterval:    s.initialRetryInterval,
+		BackoffCoefficient: 1,
+	}
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) makeWorkflowFunc(activityFunction ActivityFunctions) WorkflowFunction {
```
bergundy: Can you simplify this? There's no need to dynamically define this workflow. It doesn't even need an activity.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+		var ret string
+		err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+			ActivityID:             "activity-id",
+			DisableEagerExecution:  true,
+			StartToCloseTimeout:    s.startToCloseTimeout,
+			ScheduleToCloseTimeout: s.scheduleToCloseTimeout,
+			RetryPolicy:            s.activityRetryPolicy,
+		}), activityFunction).Get(ctx, &ret)
+
+		if !s.shouldFail.Load() {
```
bergundy: Logic is reversed here.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING, description.WorkflowExecutionInfo.Status)
+	}, 5*time.Second, 500*time.Millisecond)
+
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		wfHistory := s.GetHistory(s.Namespace().String(), &commonpb.WorkflowExecution{
+			WorkflowId: workflowRun.GetID(),
+			RunId:      workflowRun.GetRunID(),
+		})
+		require.GreaterOrEqual(t, len(wfHistory), 1)
+	}, 5*time.Second, 500*time.Millisecond)
```
bergundy: Redundant

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+	// Check if the search attributes are not empty and has TemporalReportedProblems
+	s.EventuallyWithT(func(t *assert.CollectT) {
```
bergundy, line 86: Also verify that the attempt number is >= 2 on the workflow task

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -867,6 +867,13 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskTimedOutEvent(
 			workflowTask.StartedEventID,
 			enumspb.TIMEOUT_TYPE_START_TO_CLOSE,
 		)
+
+		if m.ms.LastWorkflowTaskFailureCategory == "" {
+			m.ms.LastWorkflowTaskFailureCategory = "WorkflowTaskTimedOut"
+		}
+		if m.ms.LastWorkflowTaskFailureCause == "" {
+			m.ms.LastWorkflowTaskFailureCause = enumspb.TIMEOUT_TYPE_START_TO_CLOSE.String()
```
bergundy: This would show `TimeoutTypeStartToClose`, which is not a good cause IMHO, we would want some cause that explains better that this is workflow task related, e.g. `WorkflowTaskStartToCloseTimeOut` or just `WorkflowTaskTimedOut`.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -907,6 +914,20 @@ func (m *workflowTaskStateMachine) failWorkflowTask(
...
+	if failWorkflowTaskInfo.Attempt == 2 {
+		fmt.Println("failWorkflowTask: LastWorkflowTaskFailureCategory", m.ms.LastWorkflowTaskFailureCategory)
+		fmt.Println("failWorkflowTask: LastWorkflowTaskFailureCause", m.ms.LastWorkflowTaskFailureCause)
+
+		err := m.ms.UpdateReportedProblemsSearchAttribute(
+			m.ms.LastWorkflowTaskFailureCategory,
+			m.ms.LastWorkflowTaskFailureCause,
+		)
+		if err != nil {
+			return fmt.Errorf("failWorkflowTask: updateReportedProblemsSearchAttribute error: %v", err)
```
bergundy: Verify that we properly handle the wrapped error. We may not be able to wrap if we want to maintain the status code (e.g. INTERNAL).

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+		require.NoError(t, err)
+		require.Equal(t, 1, len(queriedWorkflows.Executions))
+
+		queriedWorkflows, err = s.SdkClient().ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+			Namespace: s.Namespace().String(),
+			Query:     "TemporalReportedProblems IN ('category=WorkflowTaskFailed', 'cause=WorkflowWorkerUnhandledFailure')",
+			PageSize:  100,
+		})
+		require.NoError(t, err)
+		require.Equal(t, 1, len(queriedWorkflows.Executions))
```
bergundy: Just check that `TemporalReportedProblems = 'category=WorkflowTaskFailed' AND TemporalReportedProblems = 'cause=WorkflowWorkerUnhandledFailure'` returns true.
I think that should work, otherwise, you can split into two separate queries. The rest of the queries are testing the visibility system, not this feature.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+	// Validate the workflow completed successfully and the search attribute is removed
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED, description.WorkflowExecutionInfo.Status)
+		require.Nil(t, description.WorkflowExecutionInfo.SearchAttributes.IndexedFields[searchattribute.TemporalReportedProblems])
+	}, 5*time.Second, 500*time.Millisecond)
+}
+
+type WFTTimedOutReportedProblemsTestSuite struct {
```
bergundy: Just make one suite for this file, not need to have 2.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+			// Sleep for a duration that's longer than the workflow task timeout (2 seconds)
+			// but shorter than the default timeout (10 seconds) to ensure we hit the timeout
+			// nolint:forbidigo // required to induce start to close timeout
+			time.Sleep(5 * time.Minute)
+			workflow.GetLogger(ctx).Info("t.WFTTimedOutReportedProblemsTestSuite: makeWorkflowFunc: done sleeping")
+		}
+
+		return nil
+	}
+}
```
bergundy: Same as above, no need for the activity or this function builder.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING, description.WorkflowExecutionInfo.Status)
+
+		wfHistory := s.GetHistory(s.Namespace().String(), &commonpb.WorkflowExecution{
+			WorkflowId: workflowRun.GetID(),
+			RunId:      workflowRun.GetRunID(),
+		})
+		require.GreaterOrEqual(t, len(wfHistory), 1)
+	}, 5*time.Second, 500*time.Millisecond)
```
bergundy: Redundant.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+		require.NotNil(t, description.WorkflowExecutionInfo.SearchAttributes.IndexedFields[searchattribute.TemporalReportedProblems])
+
+		// Decode the search attribute in keyword list format
+		searchValBytes := description.WorkflowExecutionInfo.SearchAttributes.IndexedFields[searchattribute.TemporalReportedProblems]
+		searchVal, err := searchattribute.DecodeValue(searchValBytes, enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST, false)
+		require.NoError(t, err)
+		require.NotEmpty(t, searchVal)
+		require.Equal(t, "category=WorkflowTaskFailed", searchVal.([]string)[0])
+		require.Equal(t, "cause=WorkflowWorkerUnhandledFailure", searchVal.([]string)[1])
+	}, 5*time.Second, 500*time.Millisecond)
```
bergundy, line 98: Wouldn't this take longer than 5 seconds to happen with the two attempts?

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+		wfHistory := s.GetHistory(s.Namespace().String(), &commonpb.WorkflowExecution{
+			WorkflowId: workflowRun.GetID(),
+			RunId:      workflowRun.GetRunID(),
+		})
+		require.GreaterOrEqual(t, len(wfHistory), 1)
+	}, 5*time.Second, 500*time.Millisecond)
+
+	// Check if the search attributes are not empty and has TemporalReportedProblems
+	s.EventuallyWithT(func(t *assert.CollectT) {
```
bergundy: Technically redundant because you're already doing a visibility search.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,206 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	activityFunction := func() (string, error) {
+		return "done!", nil
+	}
+
+	workflowFn := func(ctx workflow.Context) error {
+		var ret string
+		err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
```
bergundy: no need for the activity.

bergundy (commented):
Overall LGTM.

Remove the print statements before merging.

service/history/workflow/mutable_state_impl.go
yycptt: if we only have one visibility task type (besides deletion task) then yeah I agree, mutable state can automatically generate one when closing the transaction.

service/history/workflow/mutable_state_impl.go
yycptt: That method is idempotent in the sense that we collapse those task into one when closing mutable state transaction: https://github.com/temporalio/temporal/blob/dc709a2de1d53ee099efc558977e6b5338f199de/service/history/workflow/mutable_state_impl.go#L7378

service/history/workflow/workflow_task_state_machine.go
yycptt: > Can you confirm that this gets triggered on the second time a task fails

I thought it's triggered on the first failure, no? I mean the comparison is done after incrementing the attempt. Not sure what's behavior we are looking for here. 

> make the attempt number configurable

If it's configurable (or anything other than 2), then it will only work with state-based replication. That's probably fine, just want to call that out.

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -297,6 +297,12 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
...
+			if err := ms.RemoveReportedProblemsSearchAttribute(); err != nil {
```
yycptt: I think there are some places where we force reset the workflow tasks' attempt to 1 (e.g. failover or buffered events ). Do we need to clear the SA as well in those places?

service/history/workflow/workflow_task_state_machine.go
spkane31: Changing to `return err`, this was just for debugging use

tests/workflow_task_reported_problems_test.go
spkane31: For two attempts it only takes ~3 seconds, this test is passing as is.

common/dynamicconfig/constants.go
```diff
@@ -282,6 +282,12 @@ operator API calls (highest priority). Should be >0.0 and <= 1.0 (defaults to 20
 		`How many extra goroutines can be created per root.`,
 	)
 
+	NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute = NewNamespaceIntSetting(
+		"system.numConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute",
+		0,
+		`NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute is the number of consecutive workflow task problems to trigger the TemporalReportedProblems search attribute.`,
```
bergundy:
```suggestion
		`NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute is the number of consecutive workflow task problems to trigger the TemporalReportedProblems search attribute.
Setting this to 0 (the default) prevents the search attribute from being set when a problem is detected, and unset when the problem is resolved.`,
```

bergundy: Also please call this search attribute and feature out in the release notes draft document for 1.30.

common/searchattribute/compare.go
```diff
@@ -0,0 +1,57 @@
+package searchattribute
+
+import (
+	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+// AreKeywordListPayloadsEqual compares two payloads assumed to be of type KeywordList.
+// Returns true if both are nil or decode to equal multisets (order-insensitive, counts match).
```
bergundy: Counts don't matter, please fix this, it's only the unique set values that matter.
I wonder if you really need to compare search attributes directly given that you know what the current problems are, you can do a more tailored comparison to what was there at the beginning of the transaction (either by caching the values, or by hydrating them from search attributes).

schema/elasticsearch/visibility/versioned/v10/upgrade.sh
bergundy, line 1: @rodrigozhou @robholland ideally the new elasticsearch tool will do a better job at this.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -867,6 +875,13 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskTimedOutEvent(
 			workflowTask.StartedEventID,
 			enumspb.TIMEOUT_TYPE_START_TO_CLOSE,
 		)
+
+		if m.ms.executionInfo.LastWorkflowTaskFailure == nil {
+			m.ms.executionInfo.LastWorkflowTaskFailure = &persistencespb.WorkflowExecutionInfo_LastWorkflowTaskTimedOutType{
+				LastWorkflowTaskTimedOutType: enumspb.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE,
```
bergundy: Above we set this to START_TO_CLOSE, I would extract that value to a variable and set in both places.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -907,6 +922,14 @@ func (m *workflowTaskStateMachine) failWorkflowTask(
 	}
 	m.retainWorkflowTaskBuildIdInfo(failWorkflowTaskInfo)
 	m.UpdateWorkflowTask(failWorkflowTaskInfo)
+
+	consecutiveFailuresRequired := m.ms.config.NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute(m.ms.GetNamespaceEntry().Name().String())
+	if consecutiveFailuresRequired > 0 && failWorkflowTaskInfo.Attempt == int32(consecutiveFailuresRequired) {
```
bergundy: You're going to want to set this if the dynamic config is set for the first time, or the config value is decreased, or the cause changed.

```suggestion
	if consecutiveFailuresRequired > 0 && failWorkflowTaskInfo.Attempt >= int32(consecutiveFailuresRequired) {
```

bergundy: And now I'm going to ask that you cover all of these cases with tests :)

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		panic("forced-panic-to-fail-wft")
+	}
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) workflowWithActivity(ctx workflow.Context) (string, error) {
```
bergundy, line 53: Explain why we need the activity (to trigger transition from sticky to normal queue).
Can be anything else that would generate a second workflow task.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+	var ret string
+	err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ActivityID:             "activity-id",
+		DisableEagerExecution:  true,
+		StartToCloseTimeout:    s.startToCloseTimeout,
+		ScheduleToCloseTimeout: s.scheduleToCloseTimeout,
```
bergundy: Better not set schedule to close timeout, it's a way to add flakiness to tests when CI is slow.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) workflowWithActivity(ctx workflow.Context) (string, error) {
+	var ret string
+	err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ActivityID:             "activity-id",
+		DisableEagerExecution:  true,
+		StartToCloseTimeout:    s.startToCloseTimeout,
```
bergundy: No need to parameterize this IMHO, just hardcode it, it has no relevance to the test.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) workflowWithActivity(ctx workflow.Context) (string, error) {
+	var ret string
+	err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ActivityID:             "activity-id",
+		DisableEagerExecution:  true,
```
bergundy: Why? This isn't required for the test.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) workflowWithActivity(ctx workflow.Context) (string, error) {
+	var ret string
+	err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		ActivityID:             "activity-id",
```
bergundy: Why set this? It's not required and taking focus from the core test logic.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		DisableEagerExecution:  true,
+		StartToCloseTimeout:    s.startToCloseTimeout,
+		ScheduleToCloseTimeout: s.scheduleToCloseTimeout,
+		RetryPolicy:            s.activityRetryPolicy,
```
bergundy: Redundant.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	workflowOptions := sdkclient.StartWorkflowOptions{
+		ID:        testcore.RandomizeStr("wf_id-" + s.T().Name()),
+		TaskQueue: s.TaskQueue(),
+	}
+
+	workflowRun, err := s.SdkClient().ExecuteWorkflow(ctx, workflowOptions, s.simpleWorkflow)
+	s.NoError(err)
+
+	// Make sure the workflow has started and had an activity task scheduled and finished
```
bergundy: Comment is inaccurate, this workflow doesn't have an activity.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING, description.WorkflowExecutionInfo.Status)
+	}, 5*time.Second, 500*time.Millisecond)
```
bergundy: The fact that `ExecuteWorkflow` returned above indicates that the workflow is running. Delete this redundant code.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// Check if the search attributes are not empty and has TemporalReportedProblems
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
```
bergundy: Use the recommended `DescribeWorkflow` which has typed search attributes instead.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		require.NoError(t, err)
+		require.Equal(t, 1, len(queriedWorkflows.Executions))
+
+		queriedWorkflows, err = s.SdkClient().ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+			Namespace: s.Namespace().String(),
+			Query:     "TemporalReportedProblems IN ('category=WorkflowTaskFailed', 'cause=WorkflowWorkerUnhandledFailure')",
+			PageSize:  100,
+		})
+		require.NoError(t, err)
+		require.Equal(t, 1, len(queriedWorkflows.Executions))
```
bergundy: I would verify that the execution workflow ID is the expect workflow ID just to verify that this isn't some sideeffect from some other test.

Alternatively, add `WorkflowId = ...` to the query, that will give you built-in isolation in the tests.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+func TestWFTFailureReportedProblemsTestSuite(t *testing.T) {
+	s := new(WFTFailureReportedProblemsTestSuite)
+	suite.Run(t, s)
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) SetupTest() {
+	s.FunctionalTestBase.SetupTest()
+
+	s.tv = testvars.New(s.T()).WithTaskQueue(s.TaskQueue()).WithNamespaceName(s.Namespace())
```
bergundy: I don't see this used in the tests. Just remove it.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// Check if the search attributes are queryable
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		queriedWorkflows, err := s.SdkClient().ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+			Namespace: s.Namespace().String(),
+			Query:     "TemporalReportedProblems IS NOT NULL",
+			PageSize:  100,
+		})
+		require.NoError(t, err)
+		require.Equal(t, 1, len(queriedWorkflows.Executions))
```
bergundy: Can just delete this, you're not testing the visibility system, the fact that the second query returns a list of size 1 is enough to give you the coverage you need.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED, description.WorkflowExecutionInfo.Status)
+		require.Nil(t, description.WorkflowExecutionInfo.SearchAttributes.IndexedFields[searchattribute.TemporalReportedProblems])
+	}, 5*time.Second, 500*time.Millisecond)
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_SetAndClear_FailAfterActivity() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	s.OverrideDynamicConfig(dynamicconfig.NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute, 2)
```
bergundy: Maybe set this as part of setup?

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	workflowOptions := sdkclient.StartWorkflowOptions{
+		ID:        testcore.RandomizeStr("wf_id-" + s.T().Name()),
+		TaskQueue: s.TaskQueue(),
+	}
+
+	workflowRun, err := s.SdkClient().ExecuteWorkflow(ctx, workflowOptions, s.simpleWorkflow)
+	s.NoError(err)
+
+	// Make sure the workflow has started and had an activity task scheduled and finished
```
bergundy: I'm going to repeat the comment from above here, but I won't do that for the rest of the comments that need to be pasted for this test case, please apply the rest of the comments above.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+	// Unblock the workflow
+	s.shouldFail.Store(false)
+
+	var out string
+	err = workflowRun.Get(ctx, &out)
+
+	s.NoError(err)
+
+	// Validate the workflow completed successfully and the search attribute is removed
+	s.EventuallyWithT(func(t *assert.CollectT) {
```
bergundy: No need to use `Eventually` here, describe is a consistent API and you've already waited for the workflow to complete above.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED, description.WorkflowExecutionInfo.Status)
```
bergundy: No need to test this, it's obviously complete if you waited for it to complete.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		require.Equal(t, 1, len(queriedWorkflows.Executions))
+	}, 5*time.Second, 500*time.Millisecond)
+
+	// Unblock the workflow
+	s.shouldFail.Store(false)
+
+	var out string
+	err = workflowRun.Get(ctx, &out)
+
+	s.NoError(err)
```
bergundy: nit: consider

```suggestion
	s.NoError(workflowRun.Get(ctx, &out))
```

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// Validate the workflow completed successfully and the search attribute is removed
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED, description.WorkflowExecutionInfo.Status)
+		require.Nil(t, description.WorkflowExecutionInfo.SearchAttributes.IndexedFields[searchattribute.TemporalReportedProblems])
+	}, 5*time.Second, 500*time.Millisecond)
+}
```
bergundy: I would add a visibility query here with an eventually block just to be sure.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+	if err != nil {
+		return "", err
+	}
+	if s.shouldFail.Load() {
+		panic("forced-panic-to-fail-wft")
+	}
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_SetAndClear_QueryFailure() {
```
bergundy: You're not querying the workflow, this test doesn't provide any additional coverage.
I would say it's redundant anyways.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_SetAndClear_FailAfterActivity() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	s.OverrideDynamicConfig(dynamicconfig.NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute, 2)
+	s.shouldFail.Store(true)
+
+	s.Worker().RegisterWorkflow(s.simpleWorkflow)
```
bergundy:
```suggestion
	s.Worker().RegisterWorkflow(s.workflowWithActivity)
```

bergundy: And invoke the right one below.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+	// Unblock the workflow
+	s.shouldFail.Store(false)
+
+	var out string
+	err = workflowRun.Get(ctx, &out)
+
+	s.NoError(err)
+
+	// Validate the workflow completed successfully and the search attribute is removed
+	s.EventuallyWithT(func(t *assert.CollectT) {
```
bergundy: No need for this eventually block.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		description, err := s.SdkClient().DescribeWorkflowExecution(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED, description.WorkflowExecutionInfo.Status)
```
bergundy: It's obviously completed if `workflowRun.Get` returned.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+		require.Nil(t, description.WorkflowExecutionInfo.SearchAttributes.IndexedFields[searchattribute.TemporalReportedProblems])
+	}, 5*time.Second, 500*time.Millisecond)
```
bergundy: Add a visibility check (see comment below)

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func getBytesForString(str string) int {
+	return len([]byte(str)) + int(unsafe.Sizeof(str))
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) TestSizeOfLastFailureAttributes() {
```
bergundy: Make this a unit test instead.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,360 @@
...
+	// maxsize is 2 KB
+	maxSize := 2 * 1024
+	for _, category := range []string{
+		"WorkflowTaskFailed",
+		"WorkflowTaskTimedOut",
+	} {
+		for causeInt := range enumspb.WorkflowTaskFailedCause_name {
+			enumFromInt := enumspb.WorkflowTaskFailedCause(causeInt)
+			bytes := getBytesForString(fmt.Sprintf("category=%s,cause=%s", category, enumFromInt.String()))
+			require.LessOrEqual(s.T(), bytes, maxSize)
```
bergundy: Why do you need to convert to bytes? Just use `len` on the string.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,224 @@
...
+func (s *WorkflowTaskReportedProblemsReplicationSuite) checkReportedProblemsSearchAttribute(
+	admin adminservice.AdminServiceClient,
+	client sdkclient.Client,
+	workflowID, runID string,
+	namespace string,
+	expectedCategory, expectedCause string,
+	shouldExist bool,
+) {
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := client.DescribeWorkflowExecution(context.Background(), workflowID, runID)
```
bergundy: use `DescribeWorkflow` instead please.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,224 @@
...
+		workflowRun.GetID(),
+		workflowRun.GetRunID(),
+		ns,
+		"WorkflowTaskFailed",
+		"WorkflowWorkerUnhandledFailure",
+		true,
+	)
+
+	// start worker2 in cluster1
+	worker2 := sdkworker.New(standbyClient, taskQueue, sdkworker.Options{})
```
bergundy: No need for this second worker, you're not failing over here.

tests/xdc/workflow_task_reported_problems_test.go
bergundy, line 1: I had to stop and didn't fully review this file, I may have more comments when I do another round.

tests/workflow_task_reported_problems_test.go
spkane31: This test isn't useful anymore, I wrote this to validate the size of the search attribute would be small in our scenarios. Removing

common/searchattribute/compare.go
```diff
@@ -0,0 +1,57 @@
+package searchattribute
+
+import (
+	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+// AreKeywordListPayloadsEqual compares two payloads assumed to be of type KeywordList.
+// Returns true if both are nil or decode to equal multisets (order-insensitive).
```
bergundy: This isn't a multiset anymore.

common/searchattribute/compare.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Convert to sets to compare unique values only
+	setA := make(map[string]bool, len(keywordListA))
+	for _, v := range keywordListA {
+		setA[v] = true
+	}
+	setB := make(map[string]bool, len(keywordListB))
+	for _, v := range keywordListB {
+		setB[v] = true
+	}
```
bergundy: nit: use `map[string]struct{}` for sets in Go.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5882,6 +5882,56 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
 	return ms.taskGenerator.GenerateUpsertVisibilityTask()
 }
 
+func (ms *MutableStateImpl) UpdateReportedProblemsSearchAttribute() error {
+	var reportedProblems []string
+	switch wftFailure := ms.executionInfo.LastWorkflowTaskFailure.(type) {
+	case *persistencespb.WorkflowExecutionInfo_LastWorkflowTaskFailureCause:
+		reportedProblems = []string{
+			"category=WorkflowTaskFailed",
+			fmt.Sprintf("cause=%s", wftFailure.LastWorkflowTaskFailureCause.String()),
```
bergundy: I asked before, not sure how this was lost, don't just use `String()` here, you want the output to be `WorkflowTaskFailedCauseX`. (same thing below for timeout).

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -808,6 +810,12 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskFailedEvent(
 			forkEventVersion,
 			binaryChecksum,
 		)
+
+		if event != nil {
+			m.ms.executionInfo.LastWorkflowTaskFailure = &persistencespb.WorkflowExecutionInfo_LastWorkflowTaskFailureCause{
+				LastWorkflowTaskFailureCause: cause,
+			}
+		}
```
bergundy, line 820: I don't think your code covers the case where the last failure changes from timeout to failure and back, or when the failure reason changes from one cause to the other, can you please confirm that this works as expected please?

tests/workflow_task_reported_problems_test.go
bergundy: I see this is resolved but no action was taken.

tests/workflow_task_reported_problems_test.go
bergundy: I meant that describe is consistent, so if the workflow completed, you don't need an eventually block to check the result.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,185 @@
...
+		panic("forced-panic-to-fail-wft")
+	}
+	return "done!", nil
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
+	return "done!", nil
+}
+
+// workflowWithActivity is used to trigger the transition from a sticky task queue to a non-sticky task queue
```
bergundy: Please explain why, it's not clear that the activity does this.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,185 @@
...
+	s.shouldFail.Store(true)
+
+	s.Worker().RegisterWorkflow(s.simpleWorkflowWithShouldFail)
+
+	workflowOptions := sdkclient.StartWorkflowOptions{
+		ID:        testcore.RandomizeStr("wf_id-" + s.T().Name()),
+		TaskQueue: s.TaskQueue(),
+	}
+
+	workflowRun, err := s.SdkClient().ExecuteWorkflow(ctx, workflowOptions, s.simpleWorkflowWithShouldFail)
```
bergundy, line 170:
```suggestion
	workflowRun, err := s.SdkClient().ExecuteWorkflow(ctx, workflowOptions, s.workflowWithActivity)
```

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}, 5*time.Second, 500*time.Millisecond)
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_SetAndClear_FailAfterActivity() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	s.shouldFail.Store(true)
+
+	s.Worker().RegisterWorkflow(s.simpleWorkflowWithShouldFail)
```
bergundy, line 163:
```suggestion
	s.Worker().RegisterWorkflow(s.workflowWithActivity)
```

tests/xdc/workflow_task_reported_problems_test.go
bergundy, line 1: Please also test that the mutable state fields are replicated if the task failure cause changes.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	workflowRun, err := activeSDKClient.ExecuteWorkflow(ctx, workflowOptions, s.simpleWorkflow)
+	s.NoError(err)
+
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := activeSDKClient.DescribeWorkflow(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING, description.Status)
+	}, 5*time.Second, 500*time.Millisecond)
```
bergundy: I wish I didn't have to call out every location where this redundant check was made. Please remove and try not to repeat.

common/searchattribute/compare.go
```diff
@@ -0,0 +1,57 @@
...
+	if a == nil && b == nil {
+		return true
+	}
+	// If exactly one is nil, they're not equal
+	if (a == nil) != (b == nil) {
+		return false
+	}
+
+	decodedA, err := DecodeValue(a, enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST, false)
+	if err != nil {
```
bergundy: I just realized this but we should error out instead of swallowing the error here.

common/searchattribute/compare.go
```diff
@@ -0,0 +1,57 @@
+package searchattribute
+
+import (
+	commonpb "go.temporal.io/api/common/v1"
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+// AreKeywordListPayloadsEqual compares two payloads assumed to be of type KeywordList.
+// Returns true if both are nil or decode to equal sets (order-insensitive).
+func AreKeywordListPayloadsEqual(a, b *commonpb.Payload) bool {
```
bergundy: I would add something to the function name that clarifies that this is comparing them as sets.

service/frontend/namespace_handler_test.go
```diff
@@ -386,6 +386,7 @@ func (s *namespaceHandlerCommonSuite) TestCapabilities() {
 	s.config.EnableEagerWorkflowStart = dc.GetBoolPropertyFnFilteredByNamespace(false)
 	s.config.EnableUpdateWorkflowExecution = dc.GetBoolPropertyFnFilteredByNamespace(false)
 	s.config.EnableUpdateWorkflowExecutionAsyncAccepted = dc.GetBoolPropertyFnFilteredByNamespace(false)
+	s.config.NumConsecutiveWorkflowTaskProblemsToTriggerSearchAttribute = dc.GetIntPropertyFnFilteredByNamespace(0)
```
bergundy: Put `s.False` in line `385` and then turn it on here and assert that it is set to true.
That's how the rest of the capabilities are tested.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5882,6 +5882,121 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+	switch wftFailure := ms.executionInfo.LastWorkflowTaskFailure.(type) {
+	case *persistencespb.WorkflowExecutionInfo_LastWorkflowTaskFailureCause:
+		reportedProblems = []string{
+			"category=WorkflowTaskFailed",
+			fmt.Sprintf("cause=WorkflowTaskFailedCause%s", wftFailure.LastWorkflowTaskFailureCause.String()),
+		}
+	case *persistencespb.WorkflowExecutionInfo_LastWorkflowTaskTimedOutType:
+		reportedProblems = []string{
+			"category=WorkflowTaskTimedOut",
+			fmt.Sprintf("cause=WorkflowTaskTimedOutCause%s", wftFailure.LastWorkflowTaskTimedOutType.String()),
```
bergundy, line 5912: I'm actually wondering if this is ever going to be anything but StartToClose but what you have here is fine.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5882,6 +5882,121 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+			tag.WorkflowNamespaceID(ms.executionInfo.NamespaceId),
+			tag.WorkflowID(ms.executionInfo.WorkflowId),
+			tag.WorkflowRunID(ms.executionState.RunId),
+			tag.Error(err))
+		return []string{"<decode-error>"}
+	}
+
+	problems, ok := decoded.([]string)
+	if !ok {
+		ms.logger.Warn("TemporalReportedProblems payload decoded to unexpected type for logging",
```
bergundy: This would be a critical error, not a warning.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5882,6 +5882,121 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+
+// decodeReportedProblems safely decodes a keyword list payload to []string
+func (ms *MutableStateImpl) decodeReportedProblems(p *commonpb.Payload) []string {
+	if p == nil {
+		return nil
+	}
+
+	decoded, err := searchattribute.DecodeValue(p, enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST, false)
+	if err != nil {
+		ms.logger.Warn("Failed to decode TemporalReportedProblems payload for logging",
```
bergundy: This would be a critical error, not a warning.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5882,6 +5882,121 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+			tag.WorkflowNamespaceID(ms.executionInfo.NamespaceId),
+			tag.WorkflowID(ms.executionInfo.WorkflowId),
+			tag.WorkflowRunID(ms.executionState.RunId),
+			tag.Error(err))
+		return []string{"<decode-error>"}
```
bergundy: I would use a softassert here and return an error. Or just panic TBH, this is unexpected and shouldn't happen ever.

service/history/workflow/mutable_state_impl.go
```diff
@@ -5882,6 +5882,121 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+		return err
+	}
+
+	exeInfo := ms.executionInfo
+	if exeInfo.SearchAttributes == nil {
+		exeInfo.SearchAttributes = make(map[string]*commonpb.Payload, 1)
+	}
+
+	existingPayload := exeInfo.SearchAttributes[searchattribute.TemporalReportedProblems]
+	if searchattribute.AreKeywordListPayloadsEqual(existingPayload, reportedProblemsPayload) {
```
bergundy: If you're going to decode them below, why not decode the existing one and compare the decoded values?
You can simplify the entire path by just decoding what's existing and use `slices.Equal`. You know what the values look like because you insert them.

tests/workflow_task_reported_problems_test.go
bergundy: Still don't see this in the test.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	workflowRun, err := s.SdkClient().ExecuteWorkflow(ctx, workflowOptions, s.simpleWorkflowWithShouldFail)
+	s.NoError(err)
+
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := s.SdkClient().DescribeWorkflow(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+		require.NoError(t, err)
+		require.NotNil(t, description.TypedSearchAttributes)
+		_, ok := description.TypedSearchAttributes.GetKeywordList(temporal.NewSearchAttributeKeyKeywordList(searchattribute.TemporalReportedProblems))
```
bergundy, line 176: nit: you can merge this with the eventually block below for after the attempt is greater than or equal to 2.

tests/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,263 @@
...
+	s.NoError(workflowRun.Get(ctx, &out))
+
+	description, err := s.SdkClient().DescribeWorkflow(ctx, workflowRun.GetID(), workflowRun.GetRunID())
+	s.NoError(err)
+	s.NotNil(description.TypedSearchAttributes)
+	_, ok := description.TypedSearchAttributes.GetKeywordList(temporal.NewSearchAttributeKeyKeywordList(searchattribute.TemporalReportedProblems))
+	s.False(ok)
+}
+
+func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_FailureChanges() {
```
bergundy: Verify that the search attribute is first set to timeout, then ask the workflow to change the failure, and verify that the search attribute changes.
The first attempt doesn't even trigger the search attribute.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,396 @@
...
+	namespace string,
+	expectedCategory, expectedCause string,
+	shouldExist bool,
+) {
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		description, err := client.DescribeWorkflow(context.Background(), workflowID, runID)
+		require.NoError(t, err)
+
+		if shouldExist {
+			require.NotNil(t, description.TypedSearchAttributes)
```
bergundy: I think TypedSearchAttributes will never be nil so you can just remove the `NotNil` assertions everywhere. You don't check `NotNil` in the else block here BTW.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -0,0 +1,396 @@
...
+		workflowRun.GetID(),
+		workflowRun.GetRunID(),
+		ns,
+		"",
+		"",
+		false,
+	)
+}
+
+func (s *WorkflowTaskReportedProblemsReplicationSuite) TestWFTFailureReportedProblems_DynamicConfigChanges() {
```
bergundy, line 199: Same as the functional tests, you need to allow the search attribute to be set once first.

bergundy (commented):
Also missing a change from one failure cause to another. You only covered timeout to failure.

service/history/workflow/mutable_state_impl.go
spkane31: In the TimedOut scenario I don't think it will be.

service/history/workflow/mutable_state_impl.go
spkane31: I made this future proof in the scenario where we add other types of reported problems to this search attribute in methods other than this one. I'll remove this and then I can remove the `AreKeywordListPayloadsEqual` function too

tests/workflow_task_reported_problems_test.go
spkane31: I don't think this is possible, the worker doesn't communicate failures to the server after the second failure so the server won't have a way to update the search attribute.

spkane31: We discussed this offline, the SDKs don't report failures after the first attempt fails so there's no way to update the search attribute with the latest failure. If the SDK behavior changes in the future we can add support for this.

common/searchattribute/compare.go
spkane31: Ended up removing this entirely and simplifying with `slices.Equal`

tests/xdc/workflow_task_reported_problems_test.go
spkane31: same as [above](https://github.com/temporalio/temporal/pull/8223#discussion_r2412078240)

service/history/workflow/mutable_state_impl.go
```diff
@@ -5897,6 +5898,133 @@ func (ms *MutableStateImpl) updatePauseInfoSearchAttribute() error {
...
+
+	// Just remove the search attribute entirely for now
+	ms.updateSearchAttributes(map[string]*commonpb.Payload{searchattribute.TemporalReportedProblems: nil})
+	return ms.taskGenerator.GenerateUpsertVisibilityTask()
+}
+
+// logReportedProblemsChange logs changes to the TemporalReportedProblems search attribute
+func (ms *MutableStateImpl) logReportedProblemsChange(oldPayload, newPayload *commonpb.Payload) {
+	oldProblems := ms.decodeReportedProblems(oldPayload)
+	newProblems := ms.decodeReportedProblems(newPayload)
```
bergundy: nit: you can avoid decoding this when called from `UpdateReportedProblemsSearchAttribute`, you can consider decoding it in `RemoveReportedProblemsSearchAttribute` and saving a few CPU cycles.

service/history/workflow/workflow_task_state_machine.go
bergundy: Discussed offline, this isn't possible since the SDKs will avoid sending completions after the first task attempt.
Could be nice change the DC to set the attribute on the first failure and verify that the change is eventually reflected when the reason changes.

PR #8226, opened 2025-08-21
Use better string splitting techniques where possible

picatz:
## What changed?

This PR aims to avoid usage of [`strings.Split`](https://pkg.go.dev/strings#Split) where possible in favor of better string splitting techniques, speficially: [`strings.SplitN`](https://pkg.go.dev/strings#SplitN) and [`strings.SplitSeq`](https://pkg.go.dev/strings#SplitSeq) where appropriate. 

There was also a [`strings.Fields`](https://pkg.go.dev/strings#Fields) change I made to use [`strings.FieldsSeq`](https://pkg.go.dev/strings#FieldsSeq) instead, and another for S3 to use the [`path`](https://pkg.go.dev/path) package instead of [`strings.Split`](https://pkg.go.dev/strings#Split).

## Why?

[`strings.SplitN`](https://pkg.go.dev/strings#SplitN) and [`strings.SplitSeq`](https://pkg.go.dev/strings#SplitSeq) are often better options in many cases, and can be _partially_ detected using [`modernize`](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize):
> `stringsseq`: replace Split in "for range strings.Split(...)" by go1.24's more efficient `SplitSeq`, or `Fields` with `FieldSeq`.

## How did you test it?
- [X] built
- [x] added new unit test(s)

## Potential risks

There are lots of potentially subtle behaviors from the `strings.Split` (and `strings.Fields`) usage that should be accounted for. If our existing tests don't cover those subtleties, there's risk for introducing an unintended bug. More intricate handling/parsing previously using the `strings` package should get extra attention from reviewers. I've attempted to break up my changes into logical commit chunks to aid in review / help spot potentially concerning changes.

### DIFF
--- a/common/persistence/data_interfaces.go
+++ b/common/persistence/data_interfaces.go
@@ -1385,34 +1385,16 @@ func BuildHistoryGarbageCleanupInfo(namespaceID, workflowID, runID string) strin
 
 // SplitHistoryGarbageCleanupInfo returns workflow identity information
 func SplitHistoryGarbageCleanupInfo(info string) (namespaceID, workflowID, runID string, err error) {
-	// Expect format: namespaceID:workflowID:runID, but workflowID may contain ':'
-	var (
-		count int
-		first string
-		prev  string
-		mid   strings.Builder
-	)
-	for part := range strings.SplitSeq(info, ":") {
-		count++
-		if count == 1 {
-			first = part
-			continue
-		}
-		if count > 2 {
-			if mid.Len() > 0 {
-				mid.WriteByte(':')
-			}
-			mid.WriteString(prev)
-		}
-		prev = part
+	// Expect format: namespaceID:workflowID:runID, but workflowID may contain ':' so we
+	// take everything between the first and last ':' as workflowID.
+	first := strings.IndexByte(info, ':')
+	last := strings.LastIndexByte(info, ':')
+	if first < 0 || first == last { // need at least two ':' to have 3 parts
+		return "", "", "", fmt.Errorf("not able to split info for %s", info)
 	}
-	// workflowID can contain ':' so we require at least 3 parts overall
-	if count < numItemsInGarbageInfo {
-		return "", "", "", fmt.Errorf("not able to split info for  %s", info)
-	}
-	namespaceID = first
-	runID = prev
-	workflowID = mid.String()
+	namespaceID = info[:first]
+	workflowID = info[first+1 : last]
+	runID = info[last+1:]
 	return
 }
 
--- a/common/persistence/data_interfaces.go
+++ b/common/persistence/data_interfaces.go
@@ -1385,15 +1385,34 @@ func BuildHistoryGarbageCleanupInfo(namespaceID, workflowID, runID string) strin
 
 // SplitHistoryGarbageCleanupInfo returns workflow identity information
 func SplitHistoryGarbageCleanupInfo(info string) (namespaceID, workflowID, runID string, err error) {
-	ss := strings.Split(info, ":")
-	// workflowID can contain ":" so len(ss) can be greater than 3
-	if len(ss) < numItemsInGarbageInfo {
+	// Expect format: namespaceID:workflowID:runID, but workflowID may contain ':'
+	var (
+		count int
+		first string
+		prev  string
+		mid   strings.Builder
+	)
+	for part := range strings.SplitSeq(info, ":") {
+		count++
+		if count == 1 {
+			first = part
+			continue
+		}
+		if count > 2 {
+			if mid.Len() > 0 {
+				mid.WriteByte(':')
+			}
+			mid.WriteString(prev)
+		}
+		prev = part
+	}
+	// workflowID can contain ':' so we require at least 3 parts overall
+	if count < numItemsInGarbageInfo {
 		return "", "", "", fmt.Errorf("not able to split info for  %s", info)
 	}
-	namespaceID = ss[0]
-	runID = ss[len(ss)-1]
-	workflowEnd := len(info) - len(runID) - 1
-	workflowID = info[len(namespaceID)+1 : workflowEnd]
+	namespaceID = first
+	runID = prev
+	workflowID = mid.String()
 	return
 }
 
--- a/common/persistence/sql/sqlplugin/sqlite/visibility.go
+++ b/common/persistence/sql/sqlplugin/sqlite/visibility.go
@@ -10,21 +10,6 @@ import (
 )
 
 var (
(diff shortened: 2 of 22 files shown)


service/worker/scheduler/calendar.go
```diff
@@ -404,13 +416,19 @@ func makeRange(s, field, def string, minVal, maxVal int, parseMode parseMode) ([
...
 		var err error
 		step := 1
 		hasStep := false
 		if strings.Contains(part, "/") {
-			skipParts := strings.Split(part, "/")
-			if len(skipParts) != 2 {
+			// We only allow a single slash introducing an integer step. Reject inputs
+			// with more than one slash (e.g. "3/5/7") so tests receive the
+			// canonical "too many slashes" error instead of a strconv parse error.
+			if strings.Count(part, "/") > 1 { // faster than re-splitting pieces later
```
simvlad: nit: a bit easier to read:
```
slashes := strings.Contains(part, "/")
if slashes > 1 {
    return nil, fmt.Errorf("%s has too many slashes", field)
}
if slashes == 1 {
    ...
}
```

service/worker/scheduler/calendar.go
```diff
@@ -404,13 +416,19 @@ func makeRange(s, field, def string, minVal, maxVal int, parseMode parseMode) ([
...
+				return nil, fmt.Errorf("%s has too many slashes", field)
+			}
+			skipParts := strings.SplitN(part, "/", 2)
+			if len(skipParts) != 2 || skipParts[1] == "" { // second part must exist and be non-empty
```
simvlad: length of skipParts will always be 2 here. So this is really `skipParts[1] == ""` check and "has too many slashes" error is not accurate

service/worker/scheduler/calendar.go
```diff
@@ -270,14 +270,26 @@ func parseCronString(c string) (*schedulepb.StructuredCalendarSpec, *schedulepb.
...
 	// split fields
 	cal := schedulepb.CalendarSpec{Comment: comment}
-	fields := strings.Fields(c)
-	switch len(fields) {
+	// Use FieldsSeq to avoid building an unbounded slice; we only accept 5–7 fields.
+	const maxCronFields = 7
+	var toks [maxCronFields]string
+	n := 0
+	for tok := range strings.FieldsSeq(c) {
+		if n < maxCronFields {
```
simvlad, line 278:
```
if n == maxCronFields {
    // More than 7 fields → invalid.
    return nil, nil, "", errors.New("CronString does not have 5-7 fields")
}
toks[n] = tok
n++
```

service/frontend/nexus_handler.go
```diff
@@ -126,7 +126,7 @@ func (c *operationContext) augmentContext(ctx context.Context, header nexus.Head
 		func() string { return c.method },
 	)
 	if userAgent, ok := header[headerUserAgent]; ok {
-		parts := strings.Split(userAgent, clientNameVersionDelim)
+		parts := strings.SplitN(userAgent, clientNameVersionDelim, 2)
```
simvlad: I think it is worth to point out all those Split -> SplitN checks with later check for the N parts are more permissive, since they only enforce the lower bound on the number of parts (extra parts are coalesced into last part). Seems fine in this context though.

Same applies to few other places in this PR

common/persistence/data_interfaces.go
```diff
@@ -1385,15 +1385,34 @@ func BuildHistoryGarbageCleanupInfo(namespaceID, workflowID, runID string) strin
 
 // SplitHistoryGarbageCleanupInfo returns workflow identity information
 func SplitHistoryGarbageCleanupInfo(info string) (namespaceID, workflowID, runID string, err error) {
-	ss := strings.Split(info, ":")
-	// workflowID can contain ":" so len(ss) can be greater than 3
-	if len(ss) < numItemsInGarbageInfo {
+	// Expect format: namespaceID:workflowID:runID, but workflowID may contain ':'
+	var (
```
simvlad: This can be a bit simpler:
```
	first := strings.IndexByte(info, ':')
	last := strings.LastIndexByte(info, ':')

	if first < 0 || first == last {
		return "", "", "", fmt.Errorf("not able to split info for %s", info)
	}

	namespaceID = info[:first]
	runID = info[last+1:]
	workflowID = info[first+1 : last]
	return
```

common/persistence/sql/sqlplugin/sqlite/visibility.go
```diff
@@ -10,6 +10,21 @@ import (
...
+	//
+	// Actual values are already size-limited at write time by search-attribute
+	// validation in the Frontend using dynamic config:
+	//   - per-value size: common/dynamicconfig/constants.go → SearchAttributesSizeOfValueLimit
+	//     default: 2*1024 bytes (config key: "frontend.searchAttributesSizeOfValueLimit")
+	//   - total map size: common/dynamicconfig/constants.go → SearchAttributesTotalSizeLimit
+	//     default: 40*1024 bytes (config key: "frontend.searchAttributesTotalSizeLimit")
+	//   - number of keys: common/dynamicconfig/constants.go → SearchAttributesNumberOfKeysLimit
+	//     default: 100 (config key: "frontend.searchAttributesNumberOfKeysLimit")
+	// With a 2KiB per-value cap and a 3-byte separator, a pathological value can
```
simvlad: What if someone changes defaults?

service/worker/scheduler/calendar.go
picatz: 🤔 If I understand correctly, this case is already handled by the existing code?

service/frontend/nexus_handler.go
picatz: Good call out. For these nexus cases specifically I added a more strict check with `strings.Count` to help ensure it retains the same behavior as before. Extra parts aren't coalesced.

common/persistence/sql/sqlplugin/sqlite/visibility.go
picatz: If the default changed I think it would break some of the assumptions made for this value for sure. 🤔 

I'm not totally sure how to wire this together properly. Pulling in `dynammicconfig` doesn't seem right. **Since I'm less concerned about this package, I'm going to restore the original behavior.**

I need to understand how/if we should pull in `dynammicconfig` or have some other lever to configure this.

service/worker/scheduler/calendar.go
simvlad: I think it's just a bit easier to read. Not a big deal though.

PR #8231, opened 2025-08-22
Adding minimum timeout require in system workflow

yux0:
## What changed?
Adding minimum timeout require in system workflow

## Why?
The system workflow needs to have sufficient time to execute the defer logic.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/worker/migration/handover_workflow.go
+++ b/service/worker/migration/handover_workflow.go
@@ -14,6 +14,7 @@ const (
 	minimumAllowedLaggingSeconds  = 5
 	maximumAllowedLaggingSeconds  = 120
 	maximumHandoverTimeoutSeconds = 30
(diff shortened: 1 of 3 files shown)


service/worker/migration/handover_workflow.go
```diff
@@ -58,11 +59,22 @@ type (
 	}
 )
 
+func NamespaceHandoverWorkflowV2(ctx workflow.Context, params NamespaceHandoverParams) (retErr error) {
+	workflowInfo := workflow.GetInfo(ctx)
+	if workflowInfo.WorkflowRunTimeout > 0 {
+		return temporal.NewNonRetryableApplicationError(
+			"Workflow run timeout is insufficient for handover process",
```
hai719: use more relevant message?

PR #8232, opened 2025-08-22
Adding make fairsim

yiminc:
## What changed?
Add build target fairsim in makefile and update README.

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


fairsim
yycptt, line 1: Should add it to .gitignore.

PR #8233, opened 2025-08-22
Reset replication state when active cluster changed

yux0:
## What changed?
Reset replication state when active cluster changed

## Why?
When the active cluster changes, we treat it a failover happens, graceful or forceful. It should clean the state.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/frontend/namespace_handler_test.go
+++ b/service/frontend/namespace_handler_test.go
@@ -754,6 +754,7 @@ func (s *namespaceHandlerCommonSuite) TestUpdateNamespace_ChangeActiveClusterWit
 			ReplicationConfig: &persistencespb.NamespaceReplicationConfig{
 				ActiveClusterName: clusterName2,
 				Clusters:          []string{clusterName1, clusterName2},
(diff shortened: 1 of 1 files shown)


service/frontend/namespace_handler.go
```diff
@@ -513,6 +513,7 @@ func (d *namespaceHandler) UpdateNamespace(
 		if updateReplicationConfig.GetActiveClusterName() != "" {
 			activeClusterChanged = true
 			replicationConfig.ActiveClusterName = updateReplicationConfig.GetActiveClusterName()
+			replicationConfig.State = enumspb.REPLICATION_STATE_NORMAL
```
hai719, line 516: Does this introduce any implicit changes to the config? It looks like it could potentially conflict with line 510.
Would it make sense to move this logic to the caller

hai719: synced offline. It doesn't allow update both active cluster and the replication state. So this should be ok.

hai719 (approved):

xwduan (approved):

PR #8237, opened 2025-08-25
gRPC connection metrics

stephanos:
## What changed?

Added gRPC connection metrics.

## Why?

To gain insights into gRPC connections for operational purposes.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

I'll be looking at the metric in our nightly pipeline to verify.

### DIFF
--- a/common/metrics/grpc_stats.go
+++ b/common/metrics/grpc_stats.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"sync/atomic"
 
-	"go.temporal.io/server/common/log"
 	"google.golang.org/grpc/stats"
 )
 
@@ -13,16 +12,13 @@ type ServerStatsHandler stats.Handler
 var _ stats.Handler = (*grpcStatsHandler)(nil)
 
 type grpcStatsHandler struct {
-	mh Handler
-	l  log.Logger
-
+	mh          Handler
 	activeConns atomic.Int64
 }
 
-func NewServerStatsHandler(mh Handler, logger log.Logger) ServerStatsHandler {
+func NewServerStatsHandler(mh Handler) ServerStatsHandler {
 	return &grpcStatsHandler{
 		mh: mh,
-		l:  logger,
 	}
 }
 
--- a/common/metrics/grpc_stats.go
+++ b/common/metrics/grpc_stats.go
@@ -1,0 +1,56 @@
+package metrics
+
+import (
+	"context"
+	"sync/atomic"
+
+	"go.temporal.io/server/common/log"
+	"google.golang.org/grpc/stats"
+)
+
+type ServerStatsHandler stats.Handler
+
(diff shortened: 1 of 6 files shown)


common/metrics/metric_defs.go
```diff
@@ -595,6 +595,18 @@ var (
 		"service_error_with_type",
 		WithDescription("The number of all service request errors by error type."),
 	)
+	ServiceConnAccepted = NewCounterDef(
+		"service_grpc_conn_accepted",
+		WithDescription("Number of gRPC connections accepted by the service."),
```
lina-temporal: Can we clarify that these metrics are for TCP-level or stream-level connection statistics? I notice that the underlying package doesn't make the distinction, which is frustrating. From my read of the code, it should correlate with a single TCP-level connection ([`ConnBegin` generated here](https://github.com/grpc/grpc-go/blob/3074bcdd8f3c52e4b5cc9527592d475780b062a8/server.go#L1044), callsite [here](https://github.com/grpc/grpc-go/blob/3074bcdd8f3c52e4b5cc9527592d475780b062a8/server.go#L962-L965) )

lina-temporal (approved):

common/metrics/metric_defs.go
stephanos: 👍 I also had to double-check when I did this work. Yes, it corresponds to a TCP-level connection. I'll expand the description.

common/metrics/grpc_stats.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"sync/atomic"
+
+	"go.temporal.io/server/common/log"
+	"google.golang.org/grpc/stats"
+)
+
+type ServerStatsHandler stats.Handler
+
+var _ stats.Handler = (*grpcStatsHandler)(nil)
```
chaptersix, line 12: why is this needed? _ implies it's not used. Is there a side effect for this?

common/metrics/grpc_stats.go
```diff
@@ -0,0 +1,56 @@
...
+	mh Handler
+	l  log.Logger
+
+	activeConns atomic.Int64
+}
+
+func NewServerStatsHandler(mh Handler, logger log.Logger) ServerStatsHandler {
+	return &grpcStatsHandler{
+		mh: mh,
+		l:  logger,
```
chaptersix: is it worth retaining a reference to the logger if it's not being used?

common/metrics/grpc_stats.go
stephanos: So this is a Go idiom; it's used to ensure a struct implements an interface. (since Go has structural typing)

common/metrics/grpc_stats.go
stephanos: 👍 Good catch; I was using it before and hadn't remove it again.

PR #8238, opened 2025-08-26
GetWorkflowExecutionHistory long poll soft timeout

stephanos:
## What changed?

Added a "soft timeout" (language used by Workflow Update) to `GetWorkflowExecutionHistory` long polls.

## Why?

We don't want to terminate the long poll connection but instead keep it alive by sending a response back just before the timeout. The idea is that this will prevent connections from opening/terminating repeatedly (ie connection churn). 

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

I was only able to verify this manually by forcing a timeout in the server and verifying that instead of a deadline exceeded I saw a result. I'll assume this will work since the existing code already tried doing just exactly that, but it didn't do it well.

### DIFF
--- a/common/contextutil/deadline.go
+++ b/common/contextutil/deadline.go
@@ -1,0 +1,37 @@
+package contextutil
+
+import (
+	"context"
+	"time"
+)
+
+var noop = func() {}
+
+// WithDeadlineBuffer creates a child context with desired timeout.
+// If buffer is non-zero, then child context timeout will be
+// the minOf(parentCtx.Deadline()-buffer, maxTimeout). Use this
+// method to create child context when childContext cannot use
+// all of parent's deadline but instead there is a need to leave
+// some time for parent to do some post-work
+func WithDeadlineBuffer(
+	parent context.Context,
+	timeout time.Duration,
+	buffer time.Duration,
+) (context.Context, context.CancelFunc) {
+	if parent.Err() != nil {
+		return parent, noop
+	}
+
+	deadline, hasDeadline := parent.Deadline()
+
+	if !hasDeadline {
+		return context.WithTimeout(parent, timeout)
+	}
+
+	remaining := time.Until(deadline) - buffer
+	if remaining < timeout {
+		// Cap the timeout to the remaining time minus buffer.
+		timeout = max(0, remaining)
+	}
+	return context.WithTimeout(parent, timeout)
+}
--- a/common/metrics/grpc_test.go
+++ b/common/metrics/grpc_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/suite"
 	metricsspb "go.temporal.io/server/api/metrics/v1"
 	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/testing/rpctest"
 	"go.uber.org/mock/gomock"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
@@ -36,7 +37,7 @@ func (s *grpcSuite) TearDownTest() {}
 func (s *grpcSuite) TestMetadataMetricInjection() {
 	logger := log.NewMockLogger(s.controller)
 	ctx := context.Background()
-	ssts := newMockServerTransportStream()
+	ssts := rpctest.NewMockServerTransportStream("/temporal.test/MetadataMetricInjection")
 	ctx = grpc.NewContextWithServerTransportStream(ctx, ssts)
 	anyMetricName := "any_metric_name"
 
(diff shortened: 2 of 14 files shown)


service/matching/pri_matcher.go
```diff
@@ -383,15 +384,10 @@ func (tm *priTaskMatcher) syncOfferTask(
 ) (any, error) {
 	ctxs := []context.Context{ctx, tm.tqCtx}
 
-	if returnNoPollerErr {
-		if deadline, ok := ctx.Deadline(); ok && tm.fwdr == nil {
```
stephanos, line 387: Dropped `ctx.Deadline()` since that's what the helper already checks for.

common/metrics/grpc_test.go
```diff
@@ -36,7 +37,7 @@ func (s *grpcSuite) TearDownTest() {}
 func (s *grpcSuite) TestMetadataMetricInjection() {
 	logger := log.NewMockLogger(s.controller)
 	ctx := context.Background()
-	ssts := newMockServerTransportStream()
```
stephanos, line 39: Moved this to a separate package for re-use.

common/rpc/interceptor/frontend_service_error.go
```diff
@@ -0,0 +1,60 @@
...
+	ResourceExhaustedCauseHeader = "X-Resource-Exhausted-Cause"
+
+	// ResourceExhaustedScopeHeader is added to rpc response if request returns ResourceExhausted error.
+	ResourceExhaustedScopeHeader = "X-Resource-Exhausted-Scope"
+)
+
+// NewFrontendServiceErrorInterceptor returns a gRPC interceptor that has two responsibilities:
+//  1. Mask certain internal service error details.
+//  2. Propagate resource exhaustion details via gRPC headers.
+func NewFrontendServiceErrorInterceptor(
```
stephanos, line 26: Moved this here from `common/rpc/grpc.go` and merged helper into it.

common/rpc/interceptor/frontend_service_error.go
```diff
@@ -0,0 +1,60 @@
...
+		if err == nil {
+			return resp, nil
+		}
+
+		switch serviceErr := err.(type) {
+		case *serviceerrors.ShardOwnershipLost:
+			err = serviceerror.NewUnavailable("shard unavailable, please backoff and retry")
+		case *serviceerror.DataLoss:
+			err = serviceerror.NewUnavailable("internal history service error")
+		case *serviceerror.ResourceExhausted:
```
yiminc, line 45: does history/matching service not needing this header?

stephanos: I don't think they do since this is the frontend adding extra headers for the user's benefit.

The behavior hasn't changed from [current main](https://github.com/temporalio/temporal/pull/8238/files/c80a12cd0dad0ac8bd9905b47be794f88adbd2ed#diff-ec0b9acffb8fa1b542fc69dba7b5f969da8cc7a84ad9707d0575e3c274e545e2L125) there.

service/history/api/get_workflow_util.go
```diff
@@ -183,8 +186,12 @@ func GetOrPollMutableState(
 		if err != nil {
 			return nil, err
 		}
-		timer := time.NewTimer(shardContext.GetConfig().LongPollExpirationInterval(namespaceRegistry.Name().String()))
-		defer timer.Stop()
+
+		// Send back response just before caller context would time out.
+		longPollInterval := shardContext.GetConfig().LongPollExpirationInterval(namespaceRegistry.Name().String())
+		longPollCtx, cancel := contextutil.WithDeadlineBuffer(ctx, longPollInterval, longPollSoftTimeout)
+		defer cancel()
```
stephanos, line 193: This is the main change.

common/contextutil/deadline.go
```diff
@@ -0,0 +1,40 @@
...
+
+var noop = func() {}
+
+// WithDeadlineBuffer creates a child context with desired timeout.
+// If buffer is non-zero, then child context timeout will be
+// the minOf(parentCtx.Deadline()-buffer, maxTimeout). Use this
+// method to create child context when childContext cannot use
+// all of parent's deadline but instead there is a need to leave
+// some time for parent to do some post-work
+func WithDeadlineBuffer(
```
stephanos, line 16: Moved this here out of Matching - with some minor modifications.

service/history/api/get_workflow_util.go
```diff
@@ -183,8 +186,12 @@ func GetOrPollMutableState(
 		if err != nil {
 			return nil, err
 		}
-		timer := time.NewTimer(shardContext.GetConfig().LongPollExpirationInterval(namespaceRegistry.Name().String()))
-		defer timer.Stop()
+
+		// Send back response just before caller context would time out.
+		longPollInterval := shardContext.GetConfig().LongPollExpirationInterval(namespaceRegistry.Name().String())
```
stephanos, line 191: Default is 20s.

service/history/api/get_workflow_util.go
```diff
@@ -239,7 +246,7 @@ func GetOrPollMutableState(
 				if expectedNextEventID < response.GetNextEventId() || response.GetWorkflowStatus() != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING {
 					return response, nil
 				}
-			case <-timer.C:
+			case <-longPollCtx.Done():
 				return response, nil
 			case <-ctx.Done():
 				return nil, ctx.Err()
```
yycptt: this can be removed I think.

PR #8252, opened 2025-08-27
CHASM: Cache deserilized tasks

yycptt:
## What changed?
- Cache deserialized tasks
- For pure tasks, the decoded version is cached in the chasm tree
- For side effect tasks, the decoded version is cached in the task itself. Additional changes on how side effect tasks are implemented is required to use the same approach pure tasks. Will follow up later on this. 

## Why?
- Avoid decoding the task on every Validate and Execute call.

## How did you test it?
- [x] built
- [x] added new unit test(s)


chasm/tree.go
```diff
@@ -1363,6 +1366,9 @@ func (n *Node) closeTransactionUpdateComponentTasks(
 				validationErr = err
 				return false
 			}
+			if !valid {
+				delete(n.taskValueCache, existingTask.Data)
```
lina-temporal: Should we also delete the cached task if validation errors out?

yycptt: That's a good point! I added logic to invalidate tasks cached in tasks.ChasmTask{} on validation & execution error (and panic). For `taskValueCache`, it's part of mutable state and we clear entire mutable state on error, so don't really need to do anything here I believe.

PR #8253, opened 2025-08-27
CHASM: Improve path encoder usage

yycptt:
## What changed?
- Minimized the # of chasm component path encoding & decoding
- Cache encoded path for a component.

## Why?
- Avoid unnecessary path encoding/decoding
- Simplify code

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -78,11 +78,12 @@ type (
 		// if the state in memory matches the state in DB.
 		valueState valueState
 
-		// TODO: Consider storing encoded path for the node.
-		// Consider using unique package as well.
-		// Encoded path for different runs of the same Component type are the same.
+		// Cached encoded path for this node.
+		// Use encodedPath() method to retrieve the encoded path.
 		//
-		// encodedPath string
+		// TODO: Consider using unique package here.
+		// Encoded path for different runs of the same Component type are the same.
+		cachedEncodedPath *string
 
 		// When terminated is true, regardless of the Lifecycle state of the component,
 		// the component will be considered as closed.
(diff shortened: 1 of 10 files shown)


chasm/tree.go
```diff
@@ -1312,119 +1311,72 @@ func (n *Node) closeTransactionSerializeNodes() error {
...
-		// This also prevents standby logic from updating component tasks, since the condition
-		// will never be true.
-		lastUpdateVT := node.serializedNode.GetMetadata().LastUpdateVersionedTransition
-		if transitionhistory.Compare(lastUpdateVT, nextVersionedTransition) != 0 {
-			continue
-		}
+		// First update component logical tasks.
+		//
+		// The conditions excludes replication logic (applyMutation/Snapshot) which sets
+		// valueState to valueStateNeedDeserialize.
```
lina-temporal, line 1333: For my understanding, we're doing this so replicated states are the only way logical tasks can change in the standby's tree component, and for physical tasks, the standby executor can run the validate logic on its own, right?

chasm/tree.go
```diff
@@ -1312,119 +1311,72 @@ func (n *Node) closeTransactionSerializeNodes() error {
 	return nil
 }
 
-//nolint:revive // cognitive complexity 28 (> max enabled 25)
 func (n *Node) closeTransactionUpdateComponentTasks(
```
lina-temporal, line 1314: Really nice cleanup in the logical task funcs!

lina-temporal (approved):

chasm/tree.go
yycptt: Right, standby cluster only get logic task update from replication stack (ApplyMutation/Snapshot). This method is still used by standby cluster to generate physical tasks though (starting from L1355 below).

> the standby executor can run the validate logic on its own

yes. Physically tasks once generated is basically decoupled from logical tasks and can run & complete on its own.

PR #8254, opened 2025-08-27
Add `--allow-no-pollers` option for SetCurrentVersion and SetRampingVersion

carlydf:
## What changed?
Add `--allow-no-pollers` option for SetCurrentVersion and SetRampingVersion

## Why?
Users want to be able to opt-in to setting versions that have not yet had any pollers to current or ramping

## How did you test it?
- [x] built
- [ ] added new unit test(s)


service/worker/workerdeployment/client.go
```diff
@@ -638,25 +681,52 @@ func (d *ClientImpl) SetRampingVersion(
...
+			deploymentName,
+			versionObj.GetBuildId(),
+			&updatepb.Request{
+				Input: &updatepb.Input{Name: SetRampingVersion, Args: updatePayload},
+				Meta:  &updatepb.Meta{UpdateId: updateID, Identity: identity},
+			},
+			identity,
+			requestID,
+			d.getSyncBatchSize(),
+		)
```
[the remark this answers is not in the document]
carlydf: we don't allow unsetting the current version with an empty version string, so this is irrelevant

tests/worker_deployment_test.go
```diff
@@ -1990,6 +1990,50 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_Unversione
 	s.verifyTaskQueueVersioningInfo(ctx, currentVars.TaskQueue(), currentVars.DeploymentVersionString(), worker_versioning.UnversionedVersionId, 75)
 }
 
+func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentCurrentVersion_NoPollers() {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60) // verify task queue versioning info is taking a lonng time :/
```
Shivs11: any idea on why this happens?

tests/worker_deployment_test.go
```diff
@@ -2880,16 +2934,22 @@ func (s *WorkerDeploymentSuite) setAndVerifyRampingVersionUnversionedOption(
...
+}
+
+func (s *WorkerDeploymentSuite) setCurrentVersionUnversionedOption(ctx context.Context, tv *testvars.TestVars, unversioned bool, previousCurrent string, ignoreMissingTaskQueues, allowNoPollers, ensureSystemWorkflowsExist bool, expectedError string) {
 	version := tv.DeploymentVersionString()
-	if unversioned {
-		version = worker_versioning.UnversionedVersionId
-		s.ensureCreateDeployment(tv)
-	} else {
-		s.ensureCreateVersionInDeployment(tv)
+	if !allowNoPollers && ensureSystemWorkflowsExist {
```
Shivs11, line 3012: not sure i see the reason why we should have `ensureSystemWorkflowsExist` (I might be missing some context) - can we not just simplify things by only checking `!allowNoPollers`? In other words, don't we always have to ensure systemWorkflows exist when allowNoPollers == false?

tests/worker_deployment_test.go
```diff
@@ -1990,6 +1990,50 @@ func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_Unversione
...
+	s.setCurrentVersionAllowNoPollersOption(ctx, tv, worker_versioning.UnversionedVersionId, true, allowNoPollers, false, expectedErr)
+
+	// let a poller arrive with that version
+	s.pollFromDeployment(ctx, tv)
+
+	// that poller's task queue should have the current versioning info
+	s.verifyTaskQueueVersioningInfo(ctx, tv.TaskQueue(), tv.DeploymentVersionString(), "", 0)
+}
+
+func (s *WorkerDeploymentSuite) TestSetWorkerDeploymentRampingVersion_NoPollers() {
```
Shivs11, line 2049: I wonder: Do we wanna have a slightly complicated end-to-end test to ensure that this newly added feature works as intended?

I think these tests should be there, but what I was thinking is just having a test that builds on top of this (also ok to just build on top of this)

- do a describe on the worker deployment workflow: This could be important because we now have added code to separately add a version to the local state of the deployment and it would be nice to verify if we do indeed see this version in the output.
- delete this version and we could then verify if the version has been removed from the local state? Same reason as above since we are playing with the local state of the deployment workflow from within these API calls.

tests/worker_deployment_test.go
```diff
@@ -2012,10 +2056,17 @@ func (s *WorkerDeploymentSuite) verifyTaskQueueVersioningInfo(ctx context.Contex
...
+			a.Equal(expectedPercentage, tqDesc.GetVersioningInfo().GetRampingVersionPercentage())
+		} else {
+			a.Equal(expectedCurrentVersion, worker_versioning.UnversionedVersionId)
+			a.Equal(expectedRampingVersion, worker_versioning.UnversionedVersionId)
+			a.Equal(expectedPercentage, 0)
+		}
 		a.Nil(err)
-		a.Equal(expectedCurrentVersion, tqDesc.GetVersioningInfo().GetCurrentVersion()) //nolint:staticcheck // SA1019: old worker versioning
-		a.Equal(expectedRampingVersion, tqDesc.GetVersioningInfo().GetRampingVersion()) //nolint:staticcheck // SA1019: old worker versioning
-		a.Equal(expectedPercentage, tqDesc.GetVersioningInfo().GetRampingVersionPercentage())
```
Shivs11: confused: wouldn't this function without these changes work over here? I had a peek at some of the matching code and don't seem to see changes directly related to the API

Shivs11 (approved):
these changes LGTM overall and don't wanna block rn - however, would be nice to get that test in which i described below

tests/worker_deployment_test.go
carlydf: For some of the allow-no-pollers tests I want there to be no system workflows and also !allow-no-pollers so that I can check the error message is what we expect

tests/worker_deployment_test.go
carlydf: I moved the `a.Nil(err)` check to after the expected versions check because when it was being checked before, the test would eventually fail with "context timed out" error, when really the reason it timed out was due to one of the `a.Equal` checks failing. It was easier to debug the failing tests when I could see what the actual mismatch was, instead of just "context deadline exceeded".

You're right, I don't think the other changes are needed though. I made those changes because when a task queue is totally unversioned (nil ramp and nil current version), the VersioningInfo is nil, which makes the expected ramping and current versions `""` instead of `__unversioned__`. As a test writer, I felt like it was easier to reason about the test if I put "expected unversioned" instead of "expect empty string" .. but tbh now that I think about it more, I think that's the wrong approach. Will change!

tests/worker_deployment_test.go
carlydf: I think it's because the user data propagation step occurs after the worker polls (second to last line of the test), which means the last two lines of the test take a while to complete. It could be interesting to use a separate context for the last operation so that we are more specific about which parts are taking which amounts of time ... 60s is a lot and does make me a bit uncomfortable. I'll try that out

tests/worker_deployment_test.go
carlydf: I added a describe call! I don't think we need to do a delete version -- the describe call confirms that the workflow state is the same as if the version was created by a poller

tests/worker_deployment_test.go
carlydf: reverted and all still works!

PR #8257, opened 2025-08-28
Reduce PollWorkflowExecution priority

prathyushpv:
## What changed?
Reduce PollWorkflowExecution API priority to 5.

## Why?
A spike in PollWorkflowExecution requests should not block other Poll* APIs stalling workflow progress.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/service/frontend/configs/quotas.go
+++ b/service/frontend/configs/quotas.go
@@ -150,8 +150,9 @@ var (
 		"/temporal.api.workflowservice.v1.WorkflowService/RecordWorkerHeartbeat":              4,
 		"/temporal.api.workflowservice.v1.WorkflowService/FetchWorkerConfig":                  4,
 		"/temporal.api.workflowservice.v1.WorkflowService/UpdateWorkerConfig":                 4,
-		// GetWorkflowExecutionHistory with WaitNewEvent set to true is a long poll API. Consider it as any other poll API.
-		PollWorkflowHistoryAPIName: 4,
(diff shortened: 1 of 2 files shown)


service/frontend/configs/quotas.go
```diff
@@ -150,8 +150,9 @@ var (
 		"/temporal.api.workflowservice.v1.WorkflowService/RecordWorkerHeartbeat":              4,
 		"/temporal.api.workflowservice.v1.WorkflowService/FetchWorkerConfig":                  4,
 		"/temporal.api.workflowservice.v1.WorkflowService/UpdateWorkerConfig":                 4,
-		// GetWorkflowExecutionHistory with WaitNewEvent set to true is a long poll API. Consider it as any other poll API.
-		PollWorkflowHistoryAPIName: 4,
+		// GetWorkflowExecutionHistory with WaitNewEvent set to true is a long poll API.
+		// Treat as long-poll but lower priority (5) so spikes don’t block Poll* APIs.
+		PollWorkflowHistoryAPIName: 5,
```
yycptt: move to the next section?

prathyushpv: Technically, it belongs in this section with other Poll* operations since it is also a Poll request.

prathyushpv: Yeah makes sense to group all priority 5 ops in a section. Made the change

yycptt: hmm yeah that also makes sense.

PR #8259, opened 2025-08-29
refactor: replace urfave/cli v1 with v2

erka:
## What changed?

There are two versions of github.com/urfave/cli. This PR migrates it to v2 version.

## Why?

For some reason, the project uses both versions of the same dependency, which is not ideal.

## How did you test it?

- [x] built


bergundy:
Thanks for the contribution. We are planning on moving to `cobra` and figure there is no point in making this incremental change even if it's relatively low risk. I am closing this PR for now.

PR #8262, opened 2025-08-29
Improve ChildWF_CrossTQ flake - retry on missing task queues error

carlydf:
## What changed?
I was not detecting the retryable "task queues not found in proposed current version" error correctly, which meant that the SetCurrent operation was not retried, even when it should have been retried.

Basically, this "fix" didn't actually work because I was not detecting the error correctly: https://github.com/temporalio/temporal/commit/8939899df70b20a4570f9311afd3558d00616e13

## Why?
To reduce flakiness

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -2045,7 +2045,7 @@ func (s *Versioning3Suite) setCurrentDeployment(tv *testvars.TestVars) {
 		}
 		_, err := s.FrontendClient().SetWorkerDeploymentCurrentVersion(ctx, req)
 		var notFound *serviceerror.NotFound
-		if errors.As(err, &notFound) || strings.Contains(err.Error(), workerdeployment.ErrCurrentVersionDoesNotHaveAllTaskQueues) {
+		if errors.As(err, &notFound) || (err != nil && strings.Contains(err.Error(), workerdeployment.ErrCurrentVersionDoesNotHaveAllTaskQueues)) {
 			return false
 		}
 		s.NoError(err)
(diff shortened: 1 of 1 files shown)


tests/versioning_3_test.go
```diff
@@ -1947,6 +2009,23 @@ func (s *Versioning3Suite) waitForDeploymentVersionCreation(
 				}
 			}
 			a.True(found)
+			if found {
```
stephanos: I think you can remove this `if` since you already have `a.True`

tests/versioning_3_test.go
```diff
@@ -1947,6 +2009,23 @@ func (s *Versioning3Suite) waitForDeploymentVersionCreation(
...
+				versionResp, err := s.FrontendClient().DescribeWorkerDeploymentVersion(ctx,
+					&workflowservice.DescribeWorkerDeploymentVersionRequest{
+						Namespace:         s.Namespace().String(),
+						DeploymentVersion: tv.ExternalDeploymentVersion(),
+					})
+				a.Nil(err)
+				if err == nil {
```
stephanos: Same here I think. I'm assuming `a.Nil` will make the eventually attempt abort, right?

tests/versioning_3_test.go
```diff
@@ -1926,31 +1926,6 @@ func (s *Versioning3Suite) TestSyncDeploymentUserData_Update() {
 	}, data)
 }
 
-func (s *Versioning3Suite) waitForDeploymentVersionCreation(
```
carlydf, line 1929: not used

PR #8267, opened 2025-09-02
TaskQueue Fairness Rate Limit redo

stephanos:
Reverts temporalio/temporal#8263 which reverted https://github.com/temporalio/temporal/pull/8135

So this is https://github.com/temporalio/temporal/pull/8135 plus a fix for flaky tests.

Comparing the results of the test runs (4 each) for the original (1st commit) and fix (2nd commit), I can only conclude that the issue has been addressed.

### DIFF
--- a/service/matching/task_queue_partition_manager.go
+++ b/service/matching/task_queue_partition_manager.go
@@ -132,6 +132,7 @@ func (pm *taskQueuePartitionManagerImpl) GetRateLimitManager() *rateLimitManager
 // Stop does not unload the partition from matching engine. It is intended to be called by matching engine when
 // unloading the partition. For stopping and unloading a partition call unloadFromEngine instead.
 func (pm *taskQueuePartitionManagerImpl) Stop(unloadCause unloadCause) {
+	pm.rateLimitManager.Stop()
 	pm.versionedQueuesLock.Lock()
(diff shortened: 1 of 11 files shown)


service/matching/task_queue_partition_manager.go
```diff
@@ -132,9 +132,9 @@ func (pm *taskQueuePartitionManagerImpl) GetRateLimitManager() *rateLimitManager
 // Stop does not unload the partition from matching engine. It is intended to be called by matching engine when
 // unloading the partition. For stopping and unloading a partition call unloadFromEngine instead.
 func (pm *taskQueuePartitionManagerImpl) Stop(unloadCause unloadCause) {
-	pm.rateLimitManager.Stop()
 	pm.versionedQueuesLock.Lock()
 	defer pm.versionedQueuesLock.Unlock()
+	pm.rateLimitManager.Stop()
```
stephanos, line 137: The change.

carlydf (approved):
Approved! Really nice job pinning this down 🙏 

Nit: Could you change the title / add something to the title to indicate "revert but moved `pm.rateLimitManager.Stop()` after `pm.versionedQueuesLock.Lock()`" or "revert but with fix"

PR #8270, opened 2025-09-02
Migrate tdbg to Code Gen from temporal/cli

chaptersix:
first generated output

just get rid of lint errors from generated code

fix more lint errors

tidy

progress

add to gitingore

rm debug bin

corrections to commands.yml and a helper script to compare the command output

updated commands

## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+    description: A command-line tool for Temporal server debugging
+    summary: Run admin operation on Temporal server
+    has-init: true
+    options:
+      - name: address
+        type: string
+        description: Host:port for Temporal frontend service.
+        env: TEMPORAL_CLI_ADDRESS
+
+      - name: namespace
```
bergundy, line 142: This should be set on a per command basis, some commands, such as DLQ are not namespace specific.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: string
+        description: Temporal Workflow namespace.
+        default: default
+        env: TEMPORAL_CLI_NAMESPACE
+        aliases:
+          - n
+
+      - name: context-timeout
```
bergundy, line 150: I would copy the `--command-timeout` flag from `temporal` CLI for consistency.

cmd/tools/tdbg/main.go
```diff
@@ -1,14 +1,21 @@
...
 
 	"go.temporal.io/server/tools/tdbg"
 )
 
 func main() {
-	app := tdbg.NewCliApp()
-	if err := app.Run(os.Args); err != nil {
-		os.Exit(1)
+	// check env for ENABLE_TDBG_V2.
+	if os.Getenv("ENABLE_TDBG_V2") == "true" {
```
bergundy, line 12: I would rather not support both versions and instead deprecate `tdbg` in favor of a `temporal-debug` command, which is more inline with the CLI extension format we are invisioning.
All of this is despite `tdbg` not having any backwards compatibility guarantees.
We can keep `tdbg` around for another release or two just for extra safety and issue a big warning that it is deprecated.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
+# This document is automatically parsed.
+# Follow these rules.
+
+# IN-HOUSE STYLE
```
bergundy, line 4: Arguably we shouldn't just blindly copy the rules from the CLI, can you instead put a link to the CLI rules here?

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+#     Avoid using 'for example' unless there's no other way to introduce code.
+
+# * Code, flags, and keys:
+#   * Demonstrate at least one example invocation of the command in every long description.
+#   * Include the most commonly used patterns in long descriptions so users don't
+#     have to call help at multiple invocation levels.
+#   * Avoid deprecated period-delineated versions of environment-specific keys.
+#     * Yes:
+#           ```
+#           temporal env set \
```
bergundy, line 38: This example doesn't apply here, it's for `temporal` CLI.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+# - name: The name of the option set. (string)
+#   options: A list of options. Same template as Command options above (Option[])
+
+# Options that are similar but slightly different don't need to be in option sets.
+# Reserve option sets for when the behavior of the option is the same across commands.
+
+# * Include a new-line after each option set.
+
+commands:
+  - name: tdbg
```
bergundy, line 132: `temporal-debug`

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+
+      - name: context-timeout
+        type: int
+        description: Timeout in seconds for RPC calls.
+        default: 5
+        env: TEMPORAL_CONTEXT_TIMEOUT
+        aliases:
+          - ct
+
+      - name: yes
```
bergundy, line 158: This should only be present in commands that present a prompt.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+commands:
+  - name: tdbg
+    description: A command-line tool for Temporal server debugging
+    summary: Run admin operation on Temporal server
+    has-init: true
+    options:
+      - name: address
+        type: string
+        description: Host:port for Temporal frontend service.
+        env: TEMPORAL_CLI_ADDRESS
```
bergundy, line 140: If we are using the CLI conventions here, we should leverage the `envconfig` package: https://github.com/temporalio/sdk-go/tree/master/contrib/envconfig

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+          - ct
+
+      - name: yes
+        type: bool
+        description: Automatically confirm all prompts.
+
+      - name: tls-cert-path
+        type: string
+        description: Path to x509 certificate.
+        env: TEMPORAL_CLI_TLS_CERT
```
bergundy, line 165: Same here, use envconfig.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+
+      - name: tls-key-path
+        type: string
+        description: Path to private key.
+        env: TEMPORAL_CLI_TLS_KEY
```
bergundy, line 170: Same here, use envconfig.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+# Reserve option sets for when the behavior of the option is the same across commands.
+
+# * Include a new-line after each option set.
+
+commands:
+  - name: tdbg
+    description: A command-line tool for Temporal server debugging
+    summary: Run admin operation on Temporal server
+    has-init: true
+    options:
```
bergundy, line 136: Copy these from the CLI: https://github.com/temporalio/cli/blob/a29a1eea9fb18df5ffab49bdf138e8ff2bf52134/temporalcli/commandsgen/commands.yml#L163-L234

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: string-enum
+        description: When to use color auto, always, never.
+        default: auto
+        enum-values:
+          - auto
+          - always
+          - never
+
+  - name: tdbg workflow
+    summary: Run admin operations on a workflow
```
bergundy, line 197: The CLI documentation conventions is to capitalize proper nouns. Please see https://github.com/temporalio/cli/blob/a29a1eea9fb18df5ffab49bdf138e8ff2bf52134/temporalcli/commandsgen/commands.yml#L3102 for reference.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        - workflow replicate
+        - workflow delete
+      tags:
+        - workflow
+        - tdbg
+    aliases:
+      - w
+
+  - name: tdbg workflow import
+    summary: Import workflow history to database
```
bergundy, line 218: Please fix the grammar while you're here.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+    description: This command will import workflow history to the database.
+    options:
+      - name: workflow-id
+        type: string
+        description: Workflow ID to import history for.
+        required: true
+      - name: run-id
+        type: string
+        description: Run ID to import history for.
```
bergundy, line 227: nit: this can be an option-set.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+      - name: min-event-id
+        type: int
+        description: Minimum event ID to be included in the history.
+      - name: max-event-id
+        type: int
+        description: Maximum event ID to be included in the history.
+        default: 9223372036854775807
+      - name: min-event-version
+        type: int
+        default: 0
```
bergundy, line 253: Wouldn't 0 be the implicit default?

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+    options:
+      - name: workflow-id
+        type: string
+        description: Workflow ID to replicate.
+        required: true
+      - name: run-id
+        type: string
+        description: Run ID to replicate.
+
+  - name: tdbg workflow delete
```
bergundy, line 315: Do we need this considering that we have `temporal workflow delete` @yycptt? Will this delete things immediately without a background task and touching visibility? If so, worth calling out how the commands are different.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+
+# Options that are similar but slightly different don't need to be in option sets.
+# Reserve option sets for when the behavior of the option is the same across commands.
+
+# * Include a new-line after each option set.
+
+commands:
+  - name: tdbg
+    description: A command-line tool for Temporal server debugging
+    summary: Run admin operation on Temporal server
```
bergundy, line 134: I would put a big disclaimer here saying that the API of this CLI is considered experimental and can change without notice for any given release.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        - workflow describe
+        - workflow refresh-tasks
+        - workflow rebuild
+        - workflow replicate
+        - workflow delete
+      tags:
+        - workflow
+        - tdbg
+    aliases:
+      - w
```
bergundy, line 215: Let's not repeat past mistakes and add all of these command aliases. We explicitly decided against this in `temporal` CLI since it hurts readability. We were often seeing cryptic snippets in slack messages such as `tctl w s -w abc` and made a hard rule to avoid this.

tools/tdbg/commandsgen/commands.yml
bergundy: You can leverage option sets for this.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+      - name: shard-id
+        type: int
+        description: The ID of the shard to describe.
+        required: true
+
+  - name: tdbg shard list-tasks
+    summary: List tasks for a shard and task category
+    description: This command will list tasks for the given shard ID and task category.
+    options:
+      - name: more
```
bergundy, line 360: nit: I would call this `--fetch-all` or something more descriptive.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: string
+        description: >
+          Inclusive min task fire timestamp. Optional for timer tasks. Not allowed for transfer, replication, visibility tasks.
+          Supported formats: '2006-01-02T15:04:05+07:00', raw UnixNano, or time range (e.g. '15m', '1h').
+      - name: max-visibility-timestamp
+        type: string
+        description: >
+          Exclusive max task fire timestamp. Required for timer tasks. Not allowed for transfer, replication, visibility tasks.
+          Supported formats: '2006-01-02T15:04:05+07:00', raw UnixNano, or time range (e.g. '15m', '1h').
+      - name: print-json
```
bergundy, line 391: Use the CLI's standard formatting primitives.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+      - hh
+  - name: tdbg history-host describe
+    summary: Describe internal information of history host
+    description: This command will describe the internal information of a history host.
+    aliases:
+      - d
+    options:
+      - name: workflow-id
+        type: string
+        description: The ID of the workflow.
```
bergundy, line 449: Needs an explanation of how this is used and that it is mutually exclusive with shard-id.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: string
+        description: The ID of the namespace.
+      - name: workflow-id
+        type: string
+        description: The ID of the workflow.
+      - name: number-of-shards
+        type: int
+        description: Number of shards for the Temporal cluster (check config for `numHistoryShards`).
+
+  - name: tdbg taskqueue
```
bergundy, line 474: Rename to `task-queue` for consistency with the CLI.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+      - name: min-task-id
+        type: int
+        description: Minimum task ID. Default is -12346.
+      - name: max-task-id
+        type: int
+        description: Maximum task ID.
+      - name: subqueue
+        type: int
+        description: Subqueue to query. Default is 0.
+      - name: print-json
```
bergundy, line 516: Same here, use the CLI's output conventions.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: bool
+        description: Print in raw JSON format.
+      - name: fair
+        type: bool
+        description: Query fairness tasks.
+      - name: min-pass
+        type: int
+        description: Minimum pass for fairness tasks. Default is 1.
+
+  - name: tdbg taskqueue describe-task-queue-partition
```
bergundy, line 526: No need to stutter

```suggestion
  - name: tdbg task-queue describe-partition
```

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+    summary: Describe information related to a task queue partition
+    description: This command will describe the partition information for a given task queue.
+    options:
+      - name: namespace-id
```
bergundy, line 530: Arguably all of our commands should work with either namespace or namespace-id and both should be in an option set.
This would be a bigger change but it would be worth adding while we're touching this.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: string
+        description: Namespace ID.
+        default: default
+      - name: task-queue
+        type: string
+        description: Task Queue name. Required.
+      - name: task-queue-type
+        type: string
+        description: Task Queue type (activity, workflow, nexus).
+        default: TASK_QUEUE_TYPE_WORKFLOW
```
bergundy, line 540: Description is inconsistent with the default. Can we simplify to take the lower case string form? `activity`, `workflow`, `nexus`?

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        description: Partition ID. Default is 0.
+      - name: sticky-name
+        type: string
+        description: Sticky Name for a task queue partition.
+      - name: build-ids
+        type: string[]
+        description: List of Build IDs.
+      - name: unversioned
+        type: bool
+        description: Whether the task queue partition is unversioned. Default is true.
```
bergundy, line 552: I don't think the default would be true here. Arguably this flag is redundant if `--build-ids` is empty.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+      - name: all-active
+        type: bool
+        description: Whether to include all active task queue versions. Default is true.
```
bergundy, line 555: Same the default will not be true here.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+
+  - name: tdbg taskqueue force-unload-task-queue-partition
```
bergundy, line 557:
```suggestion
  - name: tdbg task-queue force-unload-partition
```

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        type: string
+        description: Namespace ID.
+        default: default
+      - name: task-queue
+        type: string
+        description: Task Queue name. Required.
+      - name: task-queue-type
+        type: string
+        description: "Task Queue type: activity, workflow, nexus(experimental)."
+        default: TASK_QUEUE_TYPE_WORKFLOW
```
bergundy, line 571: Same as the comment above.

tools/tdbg/commandsgen/commands.yml
```diff
@@ -0,0 +1,892 @@
...
+        description: Page size to use when reading messages from DB (v2 only).
+        default: 10
+      - name: last-message-id
+        type: int
+        description: The upper boundary of messages to operate on. If not provided, all messages will be operated on. However, you will be prompted for confirmation unless the --yes flag is also provided.
+      - name: dlq-type
+        type: string-enum
+        description: Type of DLQ to manage.
+        required: true
+        enum-values:
```
bergundy, line 678: I think this is missing `outbound`

bergundy (commented):
At a high level, this is looking more and more like a new command and I would spend more time designing the interface for this command instead of just changing the parsing framework which doesn't provide much value.

This command can leverage a lot of functionality exposed by the CLI and seems like it would be a perfect candidate for a CLI extension. That being said, we can make incremental progress on it.

I would recommend we bring the high level discussion items up with the team and restart this work by writing a design document listing the changes we are proposing and the general approach. After that we can focus on porting one command at a time to ease the review process.

tools/tdbg/commandsgen/commands.yml
chaptersix: fair point, I assume you mean just a perma link to this correspondence file in temporalio/cli?

chaptersix:
> At a high level, this is looking more and more like a new command and I would spend more time designing the interface for this command instead of just changing the parsing framework which doesn't provide much value.
> 
> This command can leverage a lot of functionality exposed by the CLI and seems like it would be a perfect candidate for a CLI extension. That being said, we can make incremental progress on it.
> 
> I would recommend we bring the high level discussion items up with the team and restart this work by writing a design document listing the changes we are proposing and the general approach. After that we can focus on porting one command at a time to ease the review process.

Sounds good to me. I had a hard time checking for consistency myself but I didn't want to have a PR for every command if the expectation was for this to be complete within the sprint. Seems like it takes about a day to get a PR reviewed / merged. creating a new `temporal-debug` command would make it way easier to port over time.

chaptersix:
Closing this PR but will keep the work around to reference later.

bergundy:
Just putting a summary of what we'd concluded for when we do decide to take this on:

1. The server CLI should be structured as CLI extensions, once those are available, leveraging envconfig, formatting tools, and YAML based codegen.
2. The CLI flags should only apply to the commands where they're relevant.
3. Any command should take either namespace name or namespace ID.
4. Flags should be settable before or after any positional argument, including in a parent command (following CLI practices).
5. Leveraging envconfig will deprecate some of the connection args defined in the tool and allow the tool to connect to separate server instances.
6. Flags should be aligned with the CLI (for example `--workflow-id` should be aliased to `-w` instead of `--wid`).
7. We should get rid of all of the command aliases. We explicitly decided against this in temporal CLI since it hurts readability. We were often seeing cryptic snippets in slack messages such as `tctl w s -w abc` and made a hard rule to avoid this.
8. The command should be renamed to `temporal-debug` to qualify as an extension and to avoid breaking users that rely on the `tdbg`s API.
9. The command should clearly note that its public API is experimental and subject to change at any given time.
10. We will want clear examples for all of the commands available.
11. We need to fix the grammar for all of the command and flag descriptions.
12. Capitalize proper nouns everywhere.
13. All of the comments in this PR should be addressed.

PR #8273, opened 2025-09-02
Find last message in DLQ before starting merge operation

prathyushpv:
## What changed?
Add last message id in the DLQ ListQueues response. Use this last message id in merge DLQ command if it is not provided by the user.

## Why?
tdbg dlq merge command has an optional argument --last-message-id. If this is not provided, the system workflow created for this task will merge/purge messages until the DLQ is empty.
This can create a loop that will not end. If the underlying issue that pushed the messages to DLQ is still around, the newly merged tasks will again get added to DLQ. The system workflow will proceed to merge this task again. This can go in an infinite loop.

## How did you test it?
- [x] built
- [x] added new unit test(s)


tools/tdbg/dlq_v2_service.go
```diff
@@ -240,9 +240,34 @@ func (ac *DLQV2Service) PurgeMessages(c *cli.Context) error {
...
+		if lastMessageID < persistence.FirstQueueMessageID {
+			return fmt.Errorf(
+				"--%s must be at least %d but was %d",
+				FlagLastMessageID,
+				persistence.FirstQueueMessageID,
+				lastMessageID,
+			)
+		}
+	} else {
+		_, _ = fmt.Fprint(c.App.Writer, "Warning: No last message ID provided. Using ListQueues to find the last message ID.\n")
```
stephanos: Is "Warning" the right word here? Should it be "Note" maybe?

tools/tdbg/dlq_v2_service.go
```diff
@@ -294,6 +319,43 @@ func (ac *DLQV2Service) getLastMessageID(c *cli.Context, action string) (int64,
...
+			}
+		}
+
+		if len(resp.NextPageToken) == 0 {
+			break
+		}
+		nextPageToken = resp.NextPageToken
+	}
+
+	// Queue not found, it's empty
```
stephanos: do you mean "Queue not found or it's empty"?

stephanos (approved):

tools/tdbg/dlq_v2_service.go
prathyushpv: Yeah make sense. Let me change that.

tools/tdbg/dlq_v2_service.go
prathyushpv: Queue not found means queue is empty in this case. The queue is not created until the first item is pushed into it. I will make the comment clear.

PR #8276, opened 2025-09-03
Add dc to expose Authorizer errors

pdoerner:
## What changed?
Added a new dynamic config to control whether the frontend authorization interceptor should expose errors returned by the `Authorizer` component. Previously (now default), if the `Authorizer` returned an error, the intercepter would always return a generic `PermissionDenied` error without details.

## Why?
Authorizers may fail with errors that should not be translated into `PermissionDenied`

### DIFF
--- a/common/authorization/interceptor.go
+++ b/common/authorization/interceptor.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/headers"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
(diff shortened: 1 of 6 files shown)


common/authorization/interceptor.go
```diff
@@ -222,6 +225,9 @@ func (a *Interceptor) Authorize(ctx context.Context, claims *Claims, ct *CallTar
 	if err != nil {
 		metrics.ServiceErrAuthorizeFailedCounter.With(mh).Record(1)
 		a.logger.Error("Authorization error", tag.Error(err))
+		if a.exposeAuthorizerErrors != nil && a.exposeAuthorizerErrors() {
```
bergundy, line 228: This should never be `nil` right?

bergundy (approved):

common/authorization/interceptor.go
bergundy: I would rather us fail fast and crash than have a situation where the flag is not properly passed to the authorizer.

pdoerner: It was in some functional tests. Shouldn't ever be in practice.

PR #8278, opened 2025-09-03
Implement SetWorkerDeploymentManager

carlydf:
## What changed?
Implement SetWorkerDeploymentManager

## Why?
To help coordinate writes to a worker controller by different clients without overwriting each others changes

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
Replay error is a risk because this touches wf code, but this is all new code paths that only are executed if the options are new, so I think it's ok


Shivs11 (approved):
LGTM apart from the NDE bug the AI caught!

PR #8279, opened 2025-09-03
[Scheduled Actions] Record schedule blob size metrics

lina-temporal:
## What changed?
- Added a new bytes histogram for V1 scheduler data.

## Why?
- We want to measure how much traffic we're sending in terms  scheduler writes,
since we plan to move most of that traffic from the event WAL to the history WAL.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
- Introduces an extra serialization step in the scheduler's loop to record metrics.


common/metrics/metric_defs.go
```diff
@@ -1240,6 +1240,10 @@ var (
 		"schedule_action_dropped",
 		WithDescription("The number of schedule actions that failed to start"),
 	)
+	ScheduleSize = NewBytesHistogramDef(
```
bergundy: I thought we were interested in the data rate. Wouldn't it be better to represent this is a bytes counter for every write operation (e.g. when we record an input, a result, a failure, a memo)?

lina-temporal: We are interested in the rate, yeah, but those places we mutate scheduler state (processing a watcher result, processing a signal) don't correlate very precisely with a transaction close/write, since it all happens as part of the workflow's `run` loop. I did consider that maybe the run loop is the wrong place to be recording this; maybe we'd get higher fidelity out of the mutable state statistics (and passed along `isSchedule` or something.. eh..).

lina-temporal: Thought about this some more. Since we're mostly interested in customer payload size throughput, maybe the better option is a `Counter` incremented for each signal (new start argument payload)/workflow result (new completion/failure payload). wdyt?

yycptt: 👍 Counter sounds good to me.

PR #8283, opened 2025-09-04
[Scheduled Actions] Clean up test suites' addedTasks checks

lina-temporal:
## What changed?
- Cleaned up s.addedTasks within test suites

## Why?
- In preparation for adding the Visibility component, which itself also adds tasks (breaking some of the existing asserts).
- I manually verified that the backfiller is properly generating rescheduled tasks and that they validate in CloseTransaction.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/backfiller_tasks_test.go
+++ b/chasm/lib/scheduler/backfiller_tasks_test.go
@@ -241,13 +241,9 @@ func (s *backfillerTasksSuite) runTestCase(c *backfillTestCase) {
 	// Either type of request will spawn a Backfiller and schedule an immediate pure task.
 	_, err = s.node.CloseTransaction()
 	s.NoError(err)
-	s.GreaterOrEqual(1, len(s.addedTasks))
-	task, ok := s.addedTasks[0].(*tasks.ChasmTaskPure)
-	s.True(ok)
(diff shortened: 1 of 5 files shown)


chasm/lib/scheduler/backfiller_tasks_test.go
```diff
@@ -241,13 +241,9 @@ func (s *backfillerTasksSuite) runTestCase(c *backfillTestCase) {
 	// Either type of request will spawn a Backfiller and schedule an immediate pure task.
 	_, err = s.node.CloseTransaction()
 	s.NoError(err)
-	s.GreaterOrEqual(1, len(s.addedTasks))
-	task, ok := s.addedTasks[0].(*tasks.ChasmTaskPure)
-	s.True(ok)
-	s.Equal(chasm.TaskScheduledTimeImmediate, task.GetVisibilityTime())
+	s.True(s.hasTask(&tasks.ChasmTaskPure{}, chasm.TaskScheduledTimeImmediate))
```
yycptt, line 244: Not related to this PR, but I don't think transfer queue supports pure task. That combination doesn't really make sense I think and the pure task logic can be executed in the same transaction instead.

yycptt (approved):

yycptt (commented):
Approved this PR but we need to sync offline re. immediate pure task.

PR #8286, opened 2025-09-05
Unconditionally return next workflow task in response to a task completed request

rkannan82:
## What changed?
The workflow task completion handler returns the next workflow task as part of the response even if the request has not asked for it.

This behavior is controlled by a dynamic config: EnableReturnNewWorkflowTaskUnconditionally
It is disabled by default. Therefore this change by itself is a noop.

## Why?
To standardize on a single worfkflow path in the history service. Always return the next task.

Context: The SDK can set the return_new_workflow_task to false when worker caching is disabled. This is because even if we returned the next task, the worker has to make an RPC to get the full history anyway.

With this change, we will incur an unnecessary extra RPC to fetch the history. But this is acceptable given production workloads are supposed to have caching enabled.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
This change by itself is a noop as it is guarded behind a flag (disabled).
Backwards compatibility: The SDK is expected to handle the case where it receives a new workflow task in the response when it does not have the fully history. So technically this should not break. The only concern is someone using a very old sdk version that does not handle this well.

For this, we can do 1 or more of these:
1. Add a metric to track who is still setting the request option to false, and vet out the impact radius.
2. Enable this flag in couple of SDK tests.
3. Gradually rollout the flag change. If we see customer reports, quickly disable it.

JIRA: OSS-2055, OSS-2057.

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1424,6 +1424,12 @@ timeout timer when execution timeout is specified when starting a workflow.`,
 		`EnableUpdateWorkflowModeIgnoreCurrent controls whether to enable the new logic for updating closed workflow execution
 by mutation using UpdateWorkflowModeIgnoreCurrent`,
 	)
+	EnableReturnNewWorkflowTaskUnconditionally = NewGlobalBoolSetting(
+		"history.enableNewWorkflowTaskUnconditionally",
+		false,
+		`enableNewWorkflowTaskUnconditionally controls whether the history service should return the next workflow task as part
+		of the response even if the SDK has set return_new_workflow_task to false. See OSS-2055.`,
+	)
 	EnableTransitionHistory = NewGlobalBoolSetting(
 		"history.enableTransitionHistory",
 		false,
--- a/service/history/api/respondworkflowtaskcompleted/api.go
+++ b/service/history/api/respondworkflowtaskcompleted/api.go
@@ -337,6 +337,7 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 	//   and admitted updates are lost. Uncomment this check when durable admitted is implemented
 	//   or updates stay in the registry after WFT is failed.
 	hasBufferedEventsOrMessages := ms.HasBufferedEvents() // || updateRegistry.HasOutgoingMessages(false)
+
 	if err := namespaceEntry.VerifyBinaryChecksum(request.GetBinaryChecksum()); err != nil {
 		wtFailedCause = newWorkflowTaskFailedCause(
 			enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_BINARY,
@@ -496,7 +497,6 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 			ms.GetDeploymentTransition() != nil {
 
 			newWorkflowTaskType = enumsspb.WORKFLOW_TASK_TYPE_NORMAL
-
 		} else if updateRegistry.HasOutgoingMessages(true) {
 			// There shouldn't be any sent updates in the registry because
 			// all sent but not processed updates were rejected by server.
@@ -506,10 +506,9 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 		}
 	}
 
-	bypassTaskGeneration := request.GetReturnNewWorkflowTask() && wtFailedCause == nil
-	// TODO (alex-update): All current SDKs always set ReturnNewWorkflowTask to true
-	//  which means that server always bypass task generation if WFT didn't fail.
-	//  ReturnNewWorkflowTask flag needs to be removed.
+	// By pass normal task generation flow and return the next workflow task as part of the response.
+	// It is efficient to bypass unless the task had failed.
+	bypassTaskGeneration := handler.shouldBypassTaskGeneration(req, wtFailedCause)
 
 	if newWorkflowTaskType == enumsspb.WORKFLOW_TASK_TYPE_SPECULATIVE && !bypassTaskGeneration {
 		// If task generation can't be bypassed (i.e. WFT has failed),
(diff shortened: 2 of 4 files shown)


common/dynamicconfig/constants.go
```diff
@@ -1424,6 +1424,12 @@ timeout timer when execution timeout is specified when starting a workflow.`,
 		`EnableUpdateWorkflowModeIgnoreCurrent controls whether to enable the new logic for updating closed workflow execution
 by mutation using UpdateWorkflowModeIgnoreCurrent`,
 	)
+	EnableReturnNewWorkflowTaskUnconditionally = NewGlobalBoolSetting(
+		"history.enableNewWorkflowTaskUnconditionally",
+		false,
+		`enableNewWorkflowTaskUnconditionally controls whether the history service should return the next workflow task as part
+		of the response even if the SDK has set return_new_workflow_task to false. See OSS-2055.`,
```
yycptt: plz do not put internal task tracking number in oss repo.

rkannan82: Removed here and also from api.go.

common/dynamicconfig/constants.go
```diff
@@ -1429,6 +1429,13 @@ timeout timer when execution timeout is specified when starting a workflow.`,
 		true,
 		`EnableUpdateWorkflowModeIgnoreCurrent controls whether to enable the new logic for updating closed workflow execution
 by mutation using UpdateWorkflowModeIgnoreCurrent`,
+	)
+	EnableReturnNewWorkflowTaskUnconditionally = NewGlobalBoolSetting(
+		"history.enableNewWorkflowTaskUnconditionally",
```
carlydf: I know it's verbose, but I would include the string "return" in this name, like `history.enableReturnNewWorkflowTaskUnconditionally`. Especially because that string is what people will interact with when setting this value, not the `EnableReturnNewWorkflowTaskUnconditionally` variable name

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -506,10 +506,9 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 		}
 	}
 
-	bypassTaskGeneration := request.GetReturnNewWorkflowTask() && wtFailedCause == nil
-	// TODO (alex-update): All current SDKs always set ReturnNewWorkflowTask to true
-	//  which means that server always bypass task generation if WFT didn't fail.
-	//  ReturnNewWorkflowTask flag needs to be removed.
+	// By pass normal task generation flow and return the next workflow task as part of the response.
```
carlydf, line 509: nit: 
```suggestion
	// Bypass normal task generation flow and return the next workflow task as part of the response.
```

common/dynamicconfig/constants.go
carlydf: Also I don't think the comment about "This is to have a single server side implementation" is necessary as part of the description of the config.

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -1061,3 +1057,21 @@ func (handler *WorkflowTaskCompletedHandler) clearStickyTaskQueue(ctx context.Co
 	}
 	return nil
 }
+
+// Determines whether to bypass the normal task generation flow and return the next workflow task
+// directly in the response. By default, returns true unless the task had failed.
```
carlydf: >  By default, returns true unless the task had failed.

This is not true yet, right? Because the default value for your dynamic config is false

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -1061,3 +1057,21 @@ func (handler *WorkflowTaskCompletedHandler) clearStickyTaskQueue(ctx context.Co
...
+func (handler *WorkflowTaskCompletedHandler) shouldBypassTaskGeneration(
+	request *historyservice.RespondWorkflowTaskCompletedRequest,
+	wtFailedCause *workflowTaskFailedCause,
+) bool {
+	unconditionallyEnable := handler.config.EnableReturnNewWorkflowTaskUnconditionally()
+
+	// TODO: Stop honoring the client option once we roll out the unconditionallyEnable flag.
+	// Today, the SDK can set this to false when the worker is not configured to cache history.
```
carlydf: by "this" in "set this to false" do you mean the old `return_new_workflow_task` flag?

common/dynamicconfig/constants.go
rkannan82: Done.

service/history/api/respondworkflowtaskcompleted/api.go
rkannan82: Yea, I removed that statement. The original intent was to have the comment reflect after we rollout the changes.

service/history/api/respondworkflowtaskcompleted/api.go
rkannan82: Yes. I added the field name to the comment now.

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -506,10 +506,9 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 		}
 	}
 
-	bypassTaskGeneration := request.GetReturnNewWorkflowTask() && wtFailedCause == nil
-	// TODO (alex-update): All current SDKs always set ReturnNewWorkflowTask to true
-	//  which means that server always bypass task generation if WFT didn't fail.
-	//  ReturnNewWorkflowTask flag needs to be removed.
+	// By pass normal task generation flow and return the next workflow task as part of the response.
+	// It is efficient to bypass unless the task had failed.
+	bypassTaskGeneration := handler.shouldBypassTaskGeneration(req, wtFailedCause)
```
yycptt, line 511: not sure I agree with the variable and method name as L525 - L543 are overriding this bool flag as well. 

Not blocking but please help see if they can be merged somehow.

common/dynamicconfig/constants.go
```diff
@@ -1429,6 +1429,12 @@ timeout timer when execution timeout is specified when starting a workflow.`,
 		true,
 		`EnableUpdateWorkflowModeIgnoreCurrent controls whether to enable the new logic for updating closed workflow execution
 by mutation using UpdateWorkflowModeIgnoreCurrent`,
+	)
+	EnableReturnNewWorkflowTaskUnconditionally = NewGlobalBoolSetting(
+		"history.enableReturnNewWorkflowTaskUnconditionally",
+		false,
```
yycptt, line 1435: I'd probably enable it by default for development (`config/dynamicconfig`) and all functional tests (`test/testcore/dynamic_config_overrides.go`)

yycptt (approved):

PR #8292, opened 2025-09-05
[Scheduled Actions] CHASM scheduled actions visibility support

lina-temporal:
## What changed?
- Visibility writes are added to the Scheduler components.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/generator_tasks.go
+++ b/chasm/lib/scheduler/generator_tasks.go
@@ -77,9 +77,10 @@ func (g *GeneratorTaskExecutor) Execute(
 	result, err := g.SpecProcessor.ProcessTimeRange(scheduler, t1, t2, scheduler.overlapPolicy(), "", false, nil)
 	if err != nil {
 		// An error here should be impossible, send to the DLQ.
-		logger.Error("error processing time range", tag.Error(err))
+		msg := "failed to process a time range"
+		logger.Error(msg, tag.Error(err))
 		return fmt.Errorf("%w: %w",
-			serviceerror.NewInternalf("failed to process a time range"),
+			serviceerror.NewInternal(msg),
 			err)
 	}
 
@@ -90,6 +91,16 @@ func (g *GeneratorTaskExecutor) Execute(
 
 	// Write the new high water mark.
 	generator.LastProcessedTime = timestamppb.New(result.LastActionTime)
+
+	// Update visibility, since future action times may have changed.
+	err = scheduler.UpdateVisibility(ctx, g.SpecProcessor, nil)
+	if err != nil {
+		msg := "failed to update visibility"
+		logger.Error(msg, tag.Error(err))
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal(msg),
+			err)
+	}
 
 	// Check if the schedule has gone idle.
 	idleTimeTotal := g.Config.Tweakables(scheduler.Namespace).IdleTime
--- a/chasm/lib/scheduler/invoker_tasks.go
+++ b/chasm/lib/scheduler/invoker_tasks.go
@@ -34,6 +34,7 @@ type (
 		Config         *Config
 		MetricsHandler metrics.Handler
 		BaseLogger     log.Logger
+		SpecProcessor  SpecProcessor
 
 		HistoryClient resource.HistoryClient
 
@@ -175,6 +176,12 @@ func (e *InvokerExecuteTaskExecutor) Execute(
 
 			i.recordExecuteResult(ctx, &result)
 			s.recordActionResult(&schedulerActionResult{starts: startResults})
+
+			// Update visibility, since RecentActions may have been updated.
+			err = s.UpdateVisibility(ctx, e.SpecProcessor, nil)
+			if err != nil {
+				return struct{}{}, err
+			}
 
 			return struct{}{}, nil
 		},
--- a/chasm/lib/scheduler/scheduler.go
+++ b/chasm/lib/scheduler/scheduler.go
@@ -1,13 +1,19 @@
 package scheduler
 
 import (
+	"bytes"
 	"fmt"
 	"time"
 
+	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	schedulepb "go.temporal.io/api/schedule/v1"
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
+	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/payload"
+	"go.temporal.io/server/common/primitives/timestamp"
+	"go.temporal.io/server/common/searchattribute"
 	"go.temporal.io/server/common/util"
 	"go.temporal.io/server/service/worker/scheduler"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -29,6 +35,8 @@ type Scheduler struct {
 	Invoker     chasm.Field[*Invoker]
 	Backfillers chasm.Map[string, *Backfiller] // Backfill ID => *Backfiller
 
+	Visibility chasm.Field[*chasm.Visibility]
+
 	// Locally-cached state, invalidated whenever cacheConflictToken != ConflictToken.
 	cacheConflictToken int64
 	compiledSpec       *scheduler.CompiledSpec // compiledSpec is only ever replaced whole, not mutated.
@@ -37,6 +45,12 @@ type Scheduler struct {
 const (
 	// How many recent actions to keep on the Info.RecentActions list.
 	recentActionCount = 10
+
+	// Item limit per spec field on the ScheduleInfo memo.
+	listInfoSpecFieldLimit = 10
+
+	// Field in which the schedule's memo is stored.
+	visibilityMemoFieldInfo = "ScheduleInfo"
 )
 
 // NewScheduler returns an initialized CHASM scheduler root component.
@@ -66,9 +80,13 @@ func NewScheduler(
 	}
 
 	invoker := NewInvoker(ctx, sched)
-	generator := NewGenerator(ctx, sched, invoker)
 	sched.Invoker = chasm.NewComponentField(ctx, invoker)
+
+	generator := NewGenerator(ctx, sched, invoker)
 	sched.Generator = chasm.NewComponentField(ctx, generator)
+
+	visibility := chasm.NewVisibility(ctx)
+	sched.Visibility = chasm.NewComponentField(ctx, visibility)
 
 	return sched
 }
@@ -289,4 +307,166 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
 			s.Info.RunningWorkflows = append(s.Info.RunningWorkflows, start.StartWorkflowResult)
 		}
 	}
+}
+
+// UpdateVisibility updates the schedule's visibility record, including memo
+// and search attributes. customSearchAttributes is optional, and custom search
+// attributes will be left as-is with it unset.
+//
+// See mergeCustomSearchAttributes for how custom search attributes are merged.
+func (s *Scheduler) UpdateVisibility(
+	ctx chasm.MutableContext,
+	specProcessor SpecProcessor,
+	customSearchAttributes *commonpb.SearchAttributes,
+) error {
+	needsTask := false // Set to true if we need to write anything to Visibility.
+	visibility, err := s.Visibility.Get(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Update the schedule's search attributes. This includes both Temporal-managed
+	// fields (paused state), as well as upserts for customer-specified fields.
+	upsertAttrs := make(map[string]any)
+	currentAttrs, err := visibility.GetSearchAttributes(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Only attempt to merge additional search attributes if any are given, otherwise
+	// we'll unset existing attributes.
+	if customSearchAttributes != nil &&
+		len(customSearchAttributes.GetIndexedFields()) > 0 {
+		mergeCustomSearchAttributes(
+			currentAttrs,
+			customSearchAttributes.GetIndexedFields(),
+			upsertAttrs,
+		)
+	}
+
+	// Update Paused status.
+	var currentPaused bool
+	currentPausedPayload, ok := currentAttrs[searchattribute.TemporalSchedulePaused]
+	if ok {
+		err = payload.Decode(currentPausedPayload, &currentPaused)
+		if err != nil {
+			return err
+		}
+	}
+	if !ok || currentPaused != s.Schedule.State.Paused {
+		upsertAttrs[searchattribute.TemporalSchedulePaused] = s.Schedule.State.Paused
+	}
+
+	if len(upsertAttrs) > 0 {
+		err = visibility.UpsertSearchAttributes(ctx, upsertAttrs)
+		if err != nil {
+			return err
+		}
+		needsTask = true
+	}
+
+	newInfo, err := s.GetListInfo(ctx, specProcessor)
+	if err != nil {
+		return err
+	}
+	newInfoPayload, err := newInfo.Marshal()
+	if err != nil {
+		return err
+	}
+	currentMemo, err := visibility.GetMemo(ctx)
+	if err != nil {
+		return err
+	}
+	currentInfoPayload := currentMemo[visibilityMemoFieldInfo]
+
+	// Update visibility if the memo is out-of-date or absent.
+	if currentInfoPayload == nil ||
+		!bytes.Equal(currentInfoPayload.Data, newInfoPayload) {
+		newMemo := map[string]any{visibilityMemoFieldInfo: newInfoPayload}
+		err = visibility.UpsertMemo(ctx, newMemo)
+		if err != nil {
+			return err
+		}
+		needsTask = true
+	}
+
+	if needsTask {
+		visibility.GenerateTask(ctx)
+	}
+
+	return nil
+}
+
+// GetListInfo returns the ScheduleListInfo, used as the visibility memo, and to
+// answer List queries.
+func (s *Scheduler) GetListInfo(
+	ctx chasm.Context,
+	specProcessor SpecProcessor,
+) (*schedulepb.ScheduleListInfo, error) {
+	spec := common.CloneProto(s.Schedule.Spec)
+
+	// Clear fields that are too large/not useful for the list view.
+	spec.TimezoneData = nil
+
+	// Limit the number of specs and exclusions stored on the memo.
+	spec.ExcludeStructuredCalendar = util.SliceHead(spec.ExcludeStructuredCalendar, listInfoSpecFieldLimit)
+	spec.Interval = util.SliceHead(spec.Interval, listInfoSpecFieldLimit)
+	spec.StructuredCalendar = util.SliceHead(spec.StructuredCalendar, listInfoSpecFieldLimit)
+
+	futureActionTimes, err := s.getFutureActionTimes(ctx, specProcessor)
+	if err != nil {
+		return nil, err
+	}
+
+	return &schedulepb.ScheduleListInfo{
+		Spec:              spec,
+		WorkflowType:      s.Schedule.Action.GetStartWorkflow().GetWorkflowType(),
+		Notes:             s.Schedule.State.Notes,
+		Paused:            s.Schedule.State.Paused,
+		RecentActions:     util.SliceTail(s.Info.RecentActions, recentActionCount),
+		FutureActionTimes: futureActionTimes,
+	}, nil
+}
+
+// getFutureActionTimes returns up to min(`recentActionCount`, `RemainingActions`)
+// future action times. Future action times that precede the schedule's UpdateTime
+// are not included.
+func (s *Scheduler) getFutureActionTimes(
+	ctx chasm.Context,
+	specProcessor SpecProcessor,
+) ([]*timestamppb.Timestamp, error) {
+	generator, err := s.Generator.Get(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	nextTime := func(t time.Time) (time.Time, error) {
+		res, err := specProcessor.GetNextTime(s, t)
+		return res.Next, err
+	}
+
+	count := recentActionCount
+	if s.Schedule.State.LimitedActions {
+		count = min(int(s.Schedule.State.RemainingActions), recentActionCount)
+	}
+	out := make([]*timestamppb.Timestamp, 0, count)
+	t := timestamp.TimeValue(generator.LastProcessedTime)
+	for len(out) < count {
+		t, err = nextTime(t)
+		if err != nil {
+			return nil, err
+		}
+		if t.IsZero() {
+			break
+		}
+
+		if s.Info.UpdateTime.AsTime().After(t) {
+			// Skip action times whose nominal times are prior to the schedule's update time.
+			continue
+		}
+
+		out = append(out, timestamppb.New(t))
+	}
+
+	return out, nil
 }
--- a/chasm/lib/scheduler/scheduler_test.go
+++ b/chasm/lib/scheduler/scheduler_test.go
@@ -1,0 +1,80 @@
+package scheduler_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/service/history/tasks"
+)
+
+type schedulerTestSuite struct {
+	schedulerSuite
+}
+
+func TestSchedulerSuite(t *testing.T) {
+	suite.Run(t, &schedulerTestSuite{})
+}
+
+func (s *schedulerTestSuite) TestGetListInfo() {
+	ctx := s.newMutableContext()
+	now := s.timeSource.Now()
+
+	listInfo, err := s.scheduler.GetListInfo(ctx, s.specProcessor)
+	s.NoError(err)
+
+	// Should return a populated info block.
+	s.NotNil(listInfo)
+	s.NotNil(listInfo.Spec)
+	s.NotNil(listInfo.WorkflowType)
+	s.NotEmpty(listInfo.FutureActionTimes)
+
+	for _, timestamp := range listInfo.FutureActionTimes {
+		s.True(timestamp.AsTime().After(now))
+	}
+}
+
+func (s *schedulerTestSuite) TestUpdateVisibility_TaskGeneration() {
+	ctx := s.newMutableContext()
+
+	// Custom search attributes should generate a visibility task.
+	customAttrs := &commonpb.SearchAttributes{
+		IndexedFields: map[string]*commonpb.Payload{
+			"custom_field": {Data: []byte("test_value")},
+		},
+	}
+
+	err := s.scheduler.UpdateVisibility(ctx, s.specProcessor, customAttrs)
+	s.NoError(err)
+
+	_, err = s.node.CloseTransaction()
+	s.NoError(err)
+
+	s.True(s.hasTask(&tasks.ChasmTask{}, chasm.TaskScheduledTimeImmediate))
+}
+
+func (s *schedulerTestSuite) TestUpdateVisibility_NoTaskWhenUnchanged() {
+	ctx := s.newMutableContext()
+
+	// Set up Visibility to already be up-to-date.
+	err := s.scheduler.UpdateVisibility(ctx, s.specProcessor, nil)
+	s.NoError(err)
+
+	_, err = s.node.CloseTransaction()
+	s.NoError(err)
+	s.True(s.hasTask(&tasks.ChasmTask{}, chasm.TaskScheduledTimeImmediate))
+
+	// TODO - fix this with unit testing task generation hooks. Manually verified
+	// for now.
+	s.addedTasks = make([]tasks.Task, 0)
+	// s.node.RefreshTasks()
+
+	// No custom search attributes and no changes should not generate a task.
+	err = s.scheduler.UpdateVisibility(ctx, s.specProcessor, nil)
+	s.NoError(err)
+
+	_, err = s.node.CloseTransaction()
+	s.NoError(err)
+	s.Empty(s.addedTasks)
+}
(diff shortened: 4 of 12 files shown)


chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+// attributes will be left as-is with it unset.
+//
+// See mergeCustomSearchAttributes for how custom search attributes are merged.
+func (s *Scheduler) UpdateVisibility(
+	ctx chasm.MutableContext,
+	specProcessor SpecProcessor,
+	customSearchAttributes *commonpb.SearchAttributes,
+) error {
+	needsTask := false // Set to true if we need to write anything to Visibility.
+	visibility, err := s.Visibility.Get(ctx)
```
bergundy: Not related to this PR but are we ever expecting an error getting a component to be an error that the application needs to handle? What reasons would be for this failure? Can we just make the method panic instead?

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	upsertAttrs := make(map[string]any)
+	currentAttrs, err := visibility.GetSearchAttributes(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Only attempt to merge additional search attributes if any are given, otherwise
+	// we'll unset existing attributes.
+	if customSearchAttributes != nil &&
+		len(customSearchAttributes.GetIndexedFields()) > 0 {
```
bergundy: nit: This is enough because proto getters work with nils.

```suggestion
	if len(customSearchAttributes.GetIndexedFields()) > 0 {
```

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+		mergeCustomSearchAttributes(
+			currentAttrs,
+			customSearchAttributes.GetIndexedFields(),
+			upsertAttrs,
+		)
+	}
+
+	// Update Paused status.
+	var currentPaused bool
```
bergundy: This is exactly why I didn't want components to use imperative logic to update their search attributes. This logic should be part of the framework.

bergundy: If we can afford it timeline-wise I'd much prefer if we coded up a better abstraction in the framework.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	}
+	currentMemo, err := visibility.GetMemo(ctx)
+	if err != nil {
+		return err
+	}
+	currentInfoPayload := currentMemo[visibilityMemoFieldInfo]
+
+	// Update visibility if the memo is out-of-date or absent.
+	if currentInfoPayload == nil ||
+		!bytes.Equal(currentInfoPayload.Data, newInfoPayload) {
```
bergundy: Proto marshal is not deterministic by default, this comparison will give you false negatives. You will want to unmarshal and use `proto.Equal` instead.

Another reason why component authors shouldn't implement all of this logic.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
 		}
 	}
 }
+
+// UpdateVisibility updates the schedule's visibility record, including memo
+// and search attributes. customSearchAttributes is optional, and custom search
+// attributes will be left as-is with it unset.
+//
+// See mergeCustomSearchAttributes for how custom search attributes are merged.
+func (s *Scheduler) UpdateVisibility(
```
bergundy: I don't like this approach, it assumes that the component author will remember to update visibility anytime that an attribute changes. This should be the framework's responsibility and should be done automatically when a transaction is closed.

chasm/lib/scheduler/util.go
```diff
@@ -67,3 +69,33 @@ func (j jsonStringer) String() string {
...
+func mergeCustomSearchAttributes(
+	currentAttrs map[string]*commonpb.Payload,
+	customAttrs map[string]*commonpb.Payload,
+	upsertAttrs map[string]any,
+) {
+	// Insert new and updated values.
+	for key, newPayload := range customAttrs {
+		oldPayload, alreadySet := currentAttrs[key]
+
+		if !alreadySet || !bytes.Equal(oldPayload.Data, newPayload.Data) {
```
bergundy: As mentioned before, you cannot rely on byte equality for this comparison.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+		err = visibility.UpsertMemo(ctx, newMemo)
+		if err != nil {
+			return err
+		}
+		needsTask = true
+	}
+
+	if needsTask {
+		visibility.GenerateTask(ctx)
+	}
```
yycptt: This is unnecessary, Visibility component automatically generates a task when it's updated.

chasm/lib/scheduler/scheduler.go
yycptt: When we support partial read in the future, there could be some transient error returned here. For now, it will only be serialization errors.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	currentPausedPayload, ok := currentAttrs[searchattribute.TemporalSchedulePaused]
+	if ok {
+		err = payload.Decode(currentPausedPayload, &currentPaused)
+		if err != nil {
+			return err
+		}
+	}
+	if !ok || currentPaused != s.Schedule.State.Paused {
+		upsertAttrs[searchattribute.TemporalSchedulePaused] = s.Schedule.State.Paused
+	}
```
yycptt: This is more like a "refresh visibility" approach. Do we have a like PauseSchedule method where we can update the field?

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	newInfoPayload, err := newInfo.Marshal()
+	if err != nil {
+		return err
+	}
+	currentMemo, err := visibility.GetMemo(ctx)
+	if err != nil {
+		return err
+	}
+	currentInfoPayload := currentMemo[visibilityMemoFieldInfo]
```
yycptt: nit: can use the generic chasm.GetMemo() function which returns a strongly typed value instead of Payload

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+
+	// Update visibility if the memo is out-of-date or absent.
+	if currentInfoPayload == nil ||
+		!bytes.Equal(currentInfoPayload.Data, newInfoPayload) {
+		newMemo := map[string]any{visibilityMemoFieldInfo: newInfoPayload}
+		err = visibility.UpsertMemo(ctx, newMemo)
```
yycptt: UpsertMemo will perform the encoding to Payload type, should just use `newInfo` here.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +308,165 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+
+	// Update Paused status.
+	var currentPaused bool
+	currentPausedPayload, ok := currentAttrs[searchattribute.TemporalSchedulePaused]
+	if ok {
+		err = payload.Decode(currentPausedPayload, &currentPaused)
+		if err != nil {
+			return err
+		}
+	}
```
yycptt: nit: can use the generic chasm.GetSearchAttribute() method here as well and you don't need to do the decoding.

chasm/lib/scheduler/util.go
```diff
@@ -67,3 +69,33 @@ func (j jsonStringer) String() string {
...
+			upsertAttrs[key] = newPayload
+		}
+	}
+
+	// Remove keys missing from customAttrs.
+	for key := range currentAttrs {
+		_, keep := customAttrs[key]
+		if !keep {
+			// Key isn't in the new map, delete it.
+			upsertAttrs[key] = nil
```
yycptt: Will this work? At least when I was coding the visibility.UpsertSearchAttributes() method I wasn't thinking of using that as a way for deletion.

chasm/lib/scheduler/scheduler.go
yycptt: In general I agree. Also acknowledging that imperative logic means component author needs to remember update those attributes. 

We discussed about the field tag approach which addresses this concern and also the SA registration problem but decided to do that as a follow up to unblock the Scheduler migration and deliver value sooner. It's more of a ROI & priority discussion for the entire OSS team.

chasm/lib/scheduler/scheduler.go
bergundy: If it's a transient, unexpected error, it should be okay to panic IMHO. That would be better than complicating the user interface. There's never really a way for a developer to react to this situation.

chasm/lib/scheduler/util.go
bergundy: This is how it worked in the SDKs before we had typed search attributes.

bergundy (commented):
Based on @yycptt's comments the abstraction isn't too bad today.
Seems like you don't need to manually generate a task, serialize and deserialize payloads, etc...

If we could also eliminate the need to compare search attributes and do incremental updates only to fields that changed, that would be good enough for me for merging this PR and deferring the declarative APIs for later.

Ideally we wouldn't even need to do that, if you could just recompute the desired visibility state and have the visibility component understand what needs to be updated (if anything at all) that would be much much better than what we have today.
Is that doable?

common/metrics/metric_defs.go
```diff
@@ -1241,10 +1241,6 @@ var (
 		"schedule_action_success",
 		WithDescription("The number of schedule actions that were successfully taken by a schedule"),
 	)
-	ScheduleActionAttempt = NewCounterDef(
-		"schedule_action_attempt",
-		WithDescription("The number of schedule actions attempts"),
-	)
```
lina-temporal, line 1259: I'm pretty sure these were from another PR that might still be pending.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -540,9 +541,14 @@ func (e *InvokerExecuteTaskExecutor) startWorkflow(
...
+	// realTime may be slightly past the time of the action's first scheduled WFT.
+	realTime := time.Now()
+	desiredTime := start.ActualTime
+	e.MetricsHandler.Timer(metrics.ScheduleActionDelay.Name()).Record(realTime.Sub(desiredTime.AsTime()))
+
 	return &schedulepb.ScheduleActionResult{
-		ScheduleTime: start.ActualTime,
-		ActualTime:   timestamppb.New(time.Now()),
+		ScheduleTime: desiredTime,
+		ActualTime:   timestamppb.New(realTime),
```
lina-temporal: also from another PR (nexus-watcher). Ignore, will fix before merge.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +306,75 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	p, err := payload.Encode(paused)
+	if err != nil {
+		return err
+	}
+
+	// "Pause" is the only Temporal-managed search attribute for schedules, so it is
+	// updated here ad-hoc, instead of via the SearchAttributesProvider interface.
+	pauseAttr := make(map[string]*commonpb.Payload)
+	pauseAttr[searchattribute.TemporalSchedulePaused] = p
+	return vis.SetSearchAttributes(ctx, pauseAttr) // merges with custom search attributes
```
lina-temporal: I did it this way because it seemed that to do it in the SA provider, I'd have to manually first Get the existing SAs and merge the paused status. Plus, schedule search attributes rarely change.

chasm/lib/scheduler/generator_tasks.go
```diff
@@ -88,8 +91,15 @@ func (g *GeneratorTaskExecutor) Execute(
...
 	}
 
-	// Write the new high water mark.
+	// Write the new high water mark and future action times.
 	generator.LastProcessedTime = timestamppb.New(result.LastActionTime)
+	if err := g.updateFutureActionTimes(ctx, generator, scheduler); err != nil {
+		msg := "failed to update future action times"
+		logger.Error(msg, tag.Error(err))
+		return fmt.Errorf("%w: %w",
+			serviceerror.NewInternal(msg),
```
bergundy: Should this be a non retryable error to ensure this task doesn't stay in the queue?

chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -35,6 +35,9 @@ message SchedulerState {
 message GeneratorState {
     // High water mark.
     google.protobuf.Timestamp last_processed_time = 3;
+
+    // A list of upcoming times an action will be triggered.
+    repeated google.protobuf.Timestamp future_action_times = 4;
```
bergundy, line 40: Does this need to be persisted? Can't it be calculated on the fly when calculating the memo?

chasm/lib/scheduler/scheduler.go
bergundy: You should put this in the SA provider. The visibility component is for custom search attributes. The way that you've implemented it, you'll need to update the this attribute when a user modifies the schedule, and you're deviating from the design and how we want to structure the other CHASM components.

chasm/lib/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,38 @@
+package scheduler_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type schedulerTestSuite struct {
```
bergundy: Please, we asked not to add more suites into the codebase.

chasm/lib/scheduler/scheduler.go
lina-temporal: >  The way that you've implemented it, you'll need to update the this attribute when a user modifies the schedule,

Why would I have to do that? `SetSearchAttributes` does a merge, the SA provider does not, from my read of it (it looks to be a full replace). Because these change infrequently, and independently, the ad-hoc call seemed more appropriate than writing an SA provider that has to run every transaction and have its own manual merge.

chasm/lib/scheduler/proto/v1/message.proto
lina-temporal: No, it can't be, because the Memo is computed in the context of a component (not a task executor), and therefore it doesn't have the wired-in dependencies that computing this relies upon (SpecProcessor).

chasm/lib/scheduler/scheduler_test.go
lina-temporal: So we're just going to start redoing scheduler's tests, for a package where it's a well-established pattern, before it's even landed to match that..? Also, that *isn't* the agreement we came to as a team, as far as I can remember. I'd asked specifically during that meeting if there was an issue with test suites themselves, and you'd said that it was about the way that an improper `testing.T` gets propagated to subtests improperly.

chasm/lib/scheduler/scheduler.go
lina-temporal: Chatted about this offline - the idea is for `SearchAttributes` provider to provide the Temporal-provided attributes, and `SetSearchAttributes` to be for custom search attributes. Will update

chasm/lib/scheduler/scheduler_test.go
bergundy: We said we wouldn't use suites for new tests. It's just impossible to not user the suite's `T()` and embedded assertions. But I know you're in a time crunch so we can leave this as tech debt.

chasm/lib/scheduler/generator_tasks.go
lina-temporal: Agreed, it should - I updated it here, I also did a pass of the rest of the scheduler package for similar changes (~~will open a separate PR with those momentarily~~ actually it's small enough I'm going to include it here).

chasm/lib/scheduler/scheduler_test.go
lina-temporal: I did end up replacing it without suites, I already had the helpers written from the Nexus completion tests.

chasm/lib/scheduler/proto/v1/message.proto
bergundy: I'm not sure I fully understand why this needs to be on the generator. You already keep the next actions in the invoker, can't you use that?

chasm/lib/scheduler/scheduler.go
```diff
@@ -599,3 +635,64 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+
+	newInfoPayload, err := newInfo.Marshal()
+	if err != nil {
+		// Unable to marshal list info. Return nil to skip memo update.
+		// Error will be logged once loggers are available in CHASM.
+		return nil
+	}
+
+	return map[string]chasm.VisibilityValue{
+		visibilityMemoFieldInfo: chasm.VisibilityValueByteSlice(newInfoPayload),
```
bergundy: This approach may cause extra visibility writes because proto marshalling is not deterministic and you may be returning a different byte slice every time you marshal.

I would suggest removing `VisibilityValueByteSlice` from the library so others don't fall into this trap and adding a `VisiblityValueProtoMessage` that will use `proto.Equal` for the comparison.

chasm/lib/scheduler/scheduler.go
```diff
@@ -599,3 +635,64 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+	}
+}
+
+// GetListInfo returns the ScheduleListInfo, used as the visibility memo, and to
+// answer List queries.
+func (s *Scheduler) GetListInfo(
```
bergundy: Don't use `Get` for getter names in Go and seems like you don't need to export this anyways.

```suggestion
func (s *Scheduler) listInfo(
```

chasm/lib/scheduler/scheduler.go
```diff
@@ -599,3 +635,64 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+
+	// Clear fields that are too large/not useful for the list view.
+	spec.TimezoneData = nil
+
+	// Limit the number of specs and exclusions stored on the memo.
+	spec.ExcludeStructuredCalendar = util.SliceHead(spec.ExcludeStructuredCalendar, listInfoSpecFieldLimit)
+	spec.Interval = util.SliceHead(spec.Interval, listInfoSpecFieldLimit)
+	spec.StructuredCalendar = util.SliceHead(spec.StructuredCalendar, listInfoSpecFieldLimit)
+
+	generator, err := s.Generator.Get(ctx)
```
bergundy: We're making `Get` panic on error because we realize that library authors can't handle the errors anyways, you won't have to worry about unprocessable errors here.

chasm/lib/scheduler/backfiller_tasks.go
```diff
@@ -61,17 +59,15 @@ func (b *BackfillerTaskExecutor) Execute(
 
 	scheduler, err := backfiller.Scheduler.Get(ctx)
 	if err != nil {
-		return fmt.Errorf("%w: %w",
-			serviceerror.NewInternal("scheduler tree missing node"),
-			err)
+		b.baseLogger.Error("scheduler tree missing node", tag.Error(err))
```
bergundy: Won't task executor errors be logged already by the task processing stack?

chasm/lib/scheduler/scheduler.go
```diff
@@ -599,3 +635,64 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
 	current := s.generateConflictToken()
 	return bytes.Equal(current, token)
 }
+
+// SearchAttributes returns the Temporal-managed key values for visibility.
+func (s *Scheduler) SearchAttributes(chasm.Context) []chasm.SearchAttributeKeyValue {
+	return []chasm.SearchAttributeKeyValue{
```
bergundy, line 610: You will likely need to also emit the execution status here but that's still pending discussion.

chasm/lib/scheduler/scheduler.go
```diff
@@ -599,3 +635,64 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+		chasm.SearchAttributeTemporalSchedulePaused.Value(s.Schedule.State.Paused),
+		chasm.SearchAttributeTemporalNamespaceDivision.Value(scheduler.NamespaceDivision),
```
bergundy: As we mentioned on slack, you can fix this on the read side.

chasm/lib/scheduler/backfiller_tasks.go
```diff
@@ -61,17 +59,15 @@ func (b *BackfillerTaskExecutor) Execute(
...
+		return ErrUnprocessable
```
bergundy: You can use `NewUnprocessableTaskError` from `service/history/queues` here and include all of the error details.

chasm/lib/scheduler/backfiller_tasks.go
```diff
@@ -61,17 +59,15 @@ func (b *BackfillerTaskExecutor) Execute(
...
 	}
 	logger := newTaggedLogger(b.baseLogger, scheduler)
 
 	invoker, err := scheduler.Invoker.Get(ctx)
 	if err != nil {
-		return fmt.Errorf("%w: %w",
-			serviceerror.NewInternal("scheduler tree missing node"),
-			err)
+		logger.Error("scheduler tree missing node", tag.Error(err))
+		return ErrUnprocessable
```
bergundy: Comments from above apply here too (and below).

chasm/lib/scheduler/scheduler.go
```diff
@@ -51,9 +53,18 @@ type Scheduler struct {
...
+	listInfoSpecFieldLimit = 10
+
+	// Field in which the schedule's memo is stored.
+	visibilityMemoFieldInfo = "ScheduleInfo"
 )
 
-var ErrConflictTokenMismatch = serviceerror.NewFailedPrecondition("mismatched conflict token")
+var (
+	ErrConflictTokenMismatch = errors.New("mismatched conflict token")
+	ErrUnprocessable         = errors.New("unprocessable schedule")
```
bergundy: You need to make sure that any errors you return as part of API responses have the correct status code attached and that task failures are marked as unprocessable if they cannot be retried.

I didn't go over all of the locations where these errors are returned but you should do a scan and verify that you are returning the correct status codes everywhere.

chasm/lib/scheduler/scheduler.go
```diff
@@ -121,7 +134,19 @@ func CreateScheduler(
 		req.FrontendRequest.InitialPatch,
 	)
 
-	// TODO - use visibility component to update SAs
+	// Update visibility with custom attributes.
+	visibility, err := sched.Visibility.Get(ctx)
+	if err != nil {
+		return nil, nil, ErrUnprocessable
```
bergundy: I would use internal status errors here but this is also fine because unknown errors get translated to that by default AFAIR.

chasm/lib/scheduler/scheduler.go
lina-temporal: I'll rename it, but it needs to be exported so that I can unit test it.

chasm/lib/scheduler/backfiller_tasks.go
lina-temporal: Yeah, I suppose so. This can be simplified.

chasm/lib/scheduler/proto/v1/message.proto
lina-temporal: This is for actions that haven't yet been buffered yet, which means they're still speculative (whereas Invoker has buffered actions already slated to be driven to completion, modulus overlap policy).

chasm/lib/scheduler/generator.go
```diff
@@ -32,6 +32,12 @@ func NewGenerator(ctx chasm.MutableContext, scheduler *Scheduler, invoker *Invok
 	return generator
 }
 
+// Generate immediately kicks off a new GeneratorTask. Used after updating the
+// schedule specification.
+func (g *Generator) Generate(ctx chasm.MutableContext) {
```
bergundy, line 37: nit: does this need to be exported?

chasm/lib/scheduler/generator_tasks.go
```diff
@@ -61,6 +63,12 @@ func (g *GeneratorTaskExecutor) Execute(
 		g.logSchedule(logger, "starting schedule", scheduler)
 	}
 
+	// If the high water mark is earlier than when a schedule was updated, we must skip any actions that hadn't
+	// yet been processed.
+	if scheduler.Info.GetUpdateTime().AsTime().After(generator.LastProcessedTime.AsTime()) {
```
bergundy, line 68: Is this change related to visibility?

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -513,6 +517,26 @@ func (e *InvokerProcessBufferTaskExecutor) startWorkflowDeadline(
 	return start.ActualTime.AsTime().Add(timeout)
 }
 
+// startWorkflowSearchAttributes returns the search attributes to be applied to
+// workflows kicked off. Includes custom search attributes and Temporal-managed.
+func startWorkflowSearchAttributes(
+	scheduler *Scheduler,
+	nominal time.Time,
+) *commonpb.SearchAttributes {
```
bergundy, line 525: nit:

```suggestion
func (s *Scheduler) startWorkflowSearchAttributes(
	nominal time.Time,
) *commonpb.SearchAttributes {
```

chasm/lib/scheduler/scheduler.go
```diff
@@ -52,9 +54,19 @@ type Scheduler struct {
...
+	listInfoSpecFieldLimit = 10
+
+	// Field in which the schedule's memo is stored.
+	visibilityMemoFieldInfo = "ScheduleInfo"
 )
 
-var ErrConflictTokenMismatch = serviceerror.NewFailedPrecondition("mismatched conflict token")
+var (
+	ErrConflictTokenMismatch = serviceerror.NewFailedPrecondition("mismatched conflict token")
+	ErrClosed                = serviceerror.NewInvalidArgument("schedule closed")
```
bergundy, line 67: This is a better fit since it's a timing issue

```suggestion
	ErrClosed                = serviceerror.NewFailedPrecondition("schedule closed")
```

chasm/lib/scheduler/scheduler.go
```diff
@@ -582,3 +604,63 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+	return map[string]chasm.VisibilityValue{
+		visibilityMemoFieldInfo: chasm.VisibilityValueByteSlice(payload),
+	}
+}
+
+// ListInfo returns the ScheduleListInfo, used as the visibility memo, and to
+// answer List queries.
+func (s *Scheduler) ListInfo(
+	ctx chasm.Context,
+) (*schedulepb.ScheduleListInfo, error) {
```
bergundy, line 645: This doesn't need to return an error anymore.

chasm/lib/scheduler/scheduler.go
```diff
@@ -582,3 +604,63 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+		// Error will be logged once loggers are available in CHASM.
+		return nil
+	}
+
+	payload, err := newInfo.Marshal()
+	if err != nil {
+		// Unable to marshal list info. Return nil to skip memo update.
+		// Error will be logged once loggers are available in CHASM.
+		return nil
+	}
```
bergundy, line 631: Unrelated to this PR but both places where this could error out previously will not be relevant once the the next visibility change is checked in making the logger comment irrelevant.

chasm/lib/scheduler/scheduler.go
```diff
@@ -582,3 +604,63 @@ func (s *Scheduler) validateConflictToken(token []byte) bool {
...
+	ctx chasm.Context,
+) map[string]chasm.VisibilityValue {
+	newInfo, err := s.ListInfo(ctx)
+	if err != nil {
+		// Unable to retrieve list info. Return nil to skip memo update.
+		// Error will be logged once loggers are available in CHASM.
+		return nil
+	}
+
+	payload, err := newInfo.Marshal()
```
bergundy, line 626: Use deterministic marshaling to avoid excessive visibility writes please.

chasm/lib/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,30 @@
...
+	listInfo, err := scheduler.ListInfo(ctx)
+	require.NoError(t, err)
+
+	// Should return a populated info block.
+	require.NotNil(t, listInfo)
+	require.NotNil(t, listInfo.Spec)
+	require.NotEmpty(t, listInfo.Spec.Interval)
+	protorequire.ProtoEqual(t, listInfo.Spec.Interval[0], scheduler.Schedule.Spec.Interval[0])
+	require.NotNil(t, listInfo.WorkflowType)
+	require.NotEmpty(t, listInfo.FutureActionTimes)
```
bergundy, line 28: nit: this is redundant given the comparison in the following line

chasm/lib/scheduler/spec_processor.go
```diff
@@ -36,6 +36,9 @@ type (
 			manual bool,
 			limit *int,
 		) (*ProcessedTimeRange, error)
+
+		// GetNextTime provides a peek at the next time in the spec following 'after'.
+		GetNextTime(scheduler *Scheduler, after time.Time) (legacyscheduler.GetNextTimeResult, error)
```
bergundy, line 41: nit: Don't prefix getters in Go with `Get`.

```suggestion
		NextTime(scheduler *Scheduler, after time.Time) (legacyscheduler.GetNextTimeResult, error)
```

PR #8293, opened 2025-09-05
[Scheduled Actions] Record ScheduleActionDelay in CHASM scheduler, remove some unused metric definitions

lina-temporal:
## What changed?
- Did a pass on metrics for the CHASM scheduler and added the missing delay metric.
- A few of the old schedule metrics aren't used anywhere at all, so I've removed them.

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1216,10 +1216,6 @@ var (
 		"schedule_action_success",
 		WithDescription("The number of schedule actions that were successfully taken by a schedule"),
 	)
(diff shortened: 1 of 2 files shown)


common/metrics/metric_defs.go
bergundy, line 1: Sounds like these deleted metrics provide useful information, we could consider record these metrics in the new scheduler implementation. Worth talking to @dnr to understand why these definitions are here but not implemented.

I would just track this for later consideration.

PR #2977, opened 2022-06-09
Add elasticsearch cli tool

gnz00:
**What changed?**
- A new elasticsearch-cli-tool that replaces curl invocations
- `ClusterPutSettings` and `Ping` methods added to the ES client wrappers

**Why?**
Currently, there are scripts referenced in the dockerfiles and the helm charts that rely on curl to create ES indices and poll for ES availability, e.g.

https://github.com/temporalio/docker-builds/blob/main/docker/auto-setup.sh#L245-L263
https://github.com/temporalio/docker-builds/blob/main/docker/auto-setup.sh#L265-L278
https://github.com/temporalio/temporal/blob/master/Makefile#L372-L377
https://github.com/temporalio/helm-charts/blob/master/templates/server-job.yaml#L281
https://github.com/temporalio/helm-charts/blob/master/templates/server-job.yaml#L288-L289
https://github.com/temporalio/helm-charts/blob/master/templates/server-deployment.yaml#L76

It would be better if there was a tool that leveraged the built-in temporal ES auth providers (https://github.com/temporalio/temporal/pull/1011) and a place to put common ES operations

**How did you test it?**
- Verified the tool using a new AWS Opensearch domain running Elasticsearch 7.10
- Ran the test suites

**Potential risks**
- An additional tool to maintain
- May be a better place for this
- Curl supports AWSv4 signature signing, but this would allow other providers to be implemented and keep things consistent with the core temporal ES auth provider implementation

**Is hotfix candidate?**
No

PR #8296, opened 2025-09-08
Add Elasticsearch CLI tool

robholland:
## What changed?

- Revive and update original PR #2977 for Elasticsearch CLI tool
- Add ClusterPutSettings method to CLIClient interface using raw HTTP requests
- Implement temporal-elasticsearch-tool with setup and ping commands
- Update Makefile to build and use the new tool in install-schema-es targets
- Replace curl calls with proper CLI tool in Elasticsearch schema installation
- Support AWS authentication (static, environment, aws-sdk-default)
- Update logging to use current Temporal tag-based logging system
- Add comprehensive README with usage examples

The tool provides:
- setup-schema: Create cluster settings, index templates, and indices
- ping: Test Elasticsearch connectivity and authentication

Replaces manual curl invocations with a proper CLI tool that leverages Temporal's built-in Elasticsearch auth providers and provides better error handling and logging.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks

None, net new.

---

> [!NOTE]
> Introduces `temporal-elasticsearch-tool` for ES schema/index management and ping, extends ES client APIs, embeds ES schema, and updates Makefile to use the tool.
> 
> - **Tools**:
>   - New `temporal-elasticsearch-tool` CLI with commands: `setup-schema`, `update-schema`, `create-index`, `drop-index`, `ping`; supports AWS auth and uses embedded schema files.
>   - Adds entrypoint `cmd/tools/elasticsearch`, README, and basic tests.
> - **Elasticsearch Client**:
>   - Extends `CLIClient` with `ClusterPutSettings`, `IndexPutTemplate`, `IndexPutMapping`, `Ping` and implements them (v7) using raw requests where needed; allows custom HTTP client from config (e.g., AWS-signed).
> - **Schema**:
>   - Embeds ES v7 cluster settings and index template (`schema.Embedded...` accessors).
> - **Build/Makefile**:
>   - Adds build target and binary cleanup for `temporal-elasticsearch-tool`; updates `install-schema-es` and `install-schema-xdc` to use the CLI instead of curl.
>   - Includes binary in `.goreleaser.yml`; excludes it in `.dockerignore`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b37864809ac95a427e74b2a424852992aac2442b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tools/elasticsearch/handler.go
```diff
@@ -0,0 +1,110 @@
+package elasticsearch
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/urfave/cli/v2"
```
chaptersix, line 9: should you use cobra instead? I believe we'll be using it going forward for the cli extension project.

robholland: Not sure, the other tools are still `urfave`. I think better to keep the consistent and then update them all as/when personally, but don't feel strongly.

bergundy: Let's stay with `urfave` for consistency for now. But you're right @chaptersix that we will likely move to `cobra` soon. We would need to think through whether we want to use `cobra` as is or via the command gen tool that the CLI offers.

Makefile
```diff
@@ -518,20 +523,19 @@ install-schema-postgresql12: temporal-sql-tool
...
-install-schema-es-secondary:
+install-schema-es-secondary: temporal-elasticsearch-tool
 	@printf $(COLOR) "Install Elasticsearch schema..."
-	curl --fail -X PUT "http://127.0.0.1:8200/_cluster/settings" -H "Content-Type: application/json" --data-binary @./schema/elasticsearch/visibility/cluster_settings_v7.json --write-out "\n"
-	curl --fail -X PUT "http://127.0.0.1:8200/_template/temporal_visibility_v1_template" -H "Content-Type: application/json" --data-binary @./schema/elasticsearch/visibility/index_template_v7.json --write-out "\n"
-# No --fail here because create index is not idempotent operation.
-	curl -X PUT "http://127.0.0.1:8200/temporal_visibility_v1_secondary" --write-out "\n"
+	./temporal-elasticsearch-tool -e http://127.0.0.1:8200 setup \
+		--settings-file ./schema/elasticsearch/visibility/cluster_settings_v7.json \
+		--template-file ./schema/elasticsearch/visibility/index_template_v7.json \
```
bergundy: Let's make these files included in the binary, no need to specify them, they are specific to the server version.

Also make sure to remove the curl commands from install-scema-xdc please.

cmd/tools/elasticsearch/main.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"os"
+
+	"go.temporal.io/server/tools/elasticsearch"
+)
+
+func main() {
+	if err := elasticsearch.RunTool(os.Args); err != nil {
+		os.Exit(1)
```
bergundy, line 11: Shouldn't this also print the error?

tools/elasticsearch/README.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+The options are optional. For example, if you only want to create the index you can provide just the `--index` argument. This allows you to choose which setup operations can fail by combining them with `--fail` to ignore errors.
+
+```
+export ES_SERVER=http://127.0.0.1:9200
+export ES_USER=$USER
+export ES_PWD=$PWD
+
+# ESv6
```
bergundy: We only support schema v7, this is different from the ES version. Note that we removed this from auto-setup recently too since it was tripping users up who were trying to use ES8.

https://github.com/temporalio/docker-builds/commit/38db15d48e42178075601eb225b37309265b865b

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,145 @@
...
+func BuildCLIOptions() *cli.App {
+
+	app := cli.NewApp()
+	app.Name = "temporal-elasticsearch-tool"
+	app.Usage = "Command line tool for temporal elasticsearch operations"
+	app.Version = "0.0.1"
+
+	logger := log.NewCLILogger()
+
+	app.Flags = []cli.Flag{
```
bergundy, line 56: Assuming that the env vars are for compatibility with docker-compose / auto-setup?

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,145 @@
...
+			Name:    CLIFlagAWSCredentials,
+			Value:   "",
+			Usage:   "AWS credentials provider (supported ['static', 'environment', 'aws-sdk-default'])",
+			EnvVars: []string{"ES_AWS_CREDENTIALS"},
+		},
+		&cli.StringFlag{
+			Name:    CLIFlagAWSToken,
+			Value:   "",
+			Usage:   "AWS token for use with 'static' AWS credentials provider",
+			EnvVars: []string{"ES_AWS_TOKEN"},
```
bergundy: Would this be different from any other AWS credentials given to the process? Do we need. the `ES_` prefix?

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,145 @@
...
+	CLIFlagESVersion       = CLIOptESVersion
+	CLIFlagESURL           = CLIOptESURL + ", e"
+	CLIFlagESUsername      = CLIOptESUsername + ", u"
+	CLIFlagESPassword      = CLIOptESPassword + ", p"
+	CLIFlagAWSToken        = CLIOptAWSToken + ", t"
+	CLIFlagVisibilityIndex = CLIOptVisibilityIndex + ", i"
+	CLIFlagAWSCredentials  = CLIOptAWSCredentials + ", aws"
+	CLIFlagFailSilently    = CLIOptFailSilently
+	CLIFlagSettingsFile    = CLIOptSettingsFile
+	CLIFlagTemplateFile    = CLIOptTemplateFile
```
bergundy: Any reason not to inline these? It'll make the readability of the command definitions so much better.

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,145 @@
...
+		&cli.BoolFlag{
+			Name:  schema.CLIOptQuiet,
+			Usage: "don't log errors to stderr",
+		},
+	}
+
+	app.Commands = []*cli.Command{
+		{
+			Name:    "setup-schema",
+			Aliases: []string{"setup"},
```
bergundy: Let's not add aliases, better to have one way of doing things IMHO. I can already imagine users going down the wrong path when debugging issues: "oh you ran `setup`, I ran `setup-schema` and it worked for me.

tools/elasticsearch/tasks.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read template file: %w", err)
+	}
+
+	success, err := task.esClient.IndexPutTemplate(context.TODO(), templateName, string(body))
+	if err != nil {
+		return task.handleOperationResult("template creation failed", err, false)
+	} else if !success {
+		return task.handleOperationResult("template creation failed without error", errors.New("acknowledged=false"), false)
+	}
```
rodrigozhou, line 84: Does this handle the case the template already exists?

rodrigozhou (commented):
It seems this tool is not handling upgrades.

tools/elasticsearch/main.go
robholland: No, true, fixed.

cmd/tools/elasticsearch/main.go
robholland: The error is printed via `RunTool`, as is the seutp for the other tools.

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,145 @@
...
+var osExit = os.Exit
+
+// root handler for all cli commands
+func cliHandler(c *cli.Context, handler func(c *cli.Context, logger log.Logger) error, logger log.Logger) {
+	quiet := c.Bool(schema.CLIOptQuiet)
+	err := handler(c, logger)
+	if err != nil && !quiet {
+		osExit(1)
+	}
+}
```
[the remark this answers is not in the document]
robholland: The logging is handled further down the stack (bit of a weird pattern, but that's what is done by the other tools also).

tools/elasticsearch/tasks.go
```diff
@@ -0,0 +1,135 @@
...
+
+// handleOperationResult handles the result of an operation, optionally failing silently
+func (task *SetupTask) handleOperationResult(msg string, err error, success bool) error {
+	if !task.config.FailSilently {
+		task.logger.Error(msg, tag.Error(err))
+		return err
+	}
+	task.logger.Warn(msg, tag.Error(err))
+	return nil
+}
```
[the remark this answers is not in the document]
robholland: Fixed in https://github.com/temporalio/temporal/pull/8296/commits/084e50a99f89772a2707a93987915e1dc11a7a14

tools/elasticsearch/main.go
robholland: Yep.

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,172 @@
...
```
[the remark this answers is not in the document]
robholland: This is how the other tools behave also, so not going to investigate that as part of this PR. I believe the intent is that if --quiet mode is given the tool should not report errors to the calling process.

tools/elasticsearch/main.go
robholland: They are used in the handler.go also.

tools/elasticsearch/tasks.go
robholland: Running `setup-schema` repeatedly won't report an error, it will just update it.

schema/embed.go
```diff
@@ -46,3 +46,21 @@ func PathsByDB(dbName string) []string {
...
+	data, err := assets.ReadFile("elasticsearch/visibility/cluster_settings_v7.json")
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
+
+// ElasticsearchIndexTemplate returns the embedded index template for Elasticsearch v7 (latest version)
+func ElasticsearchIndexTemplate() (string, error) {
+	data, err := assets.ReadFile("elasticsearch/visibility/versioned/v9/index_template_v7.json")
```
rodrigozhou: Every schema update bumps the version. Can you use the symlink? Or go through the file tree and identify the highest version at runtime.

tools/elasticsearch/tasks.go
```diff
@@ -0,0 +1,159 @@
...
+
+// setupTemplate handles template configuration
+func (task *SetupTask) setupTemplate() error {
+	config := task.config
+	if len(config.TemplateContent) == 0 {
+		task.logger.Info("Skipping template creation, no embedded template content")
+		return nil
+	}
+
+	success, err := task.esClient.IndexPutTemplate(context.TODO(), templateName, config.TemplateContent)
```
rodrigozhou, line 79: Updating the template doesn't updates the index mappings.
The `upgrade-schema` should update the mappings for the index.

robholland: Added `update-index` for this.

robholland: Sorry, I mean I added `--index` argument to `update-schema` that does this.

tools/elasticsearch/handler.go
```diff
@@ -0,0 +1,223 @@
...
+			VisibilityIndex: indexName,
+			FailSilently:    cli.Bool(CLIOptFailSilently),
+		},
+	}
+
+	// If index is specified, update index mappings; otherwise update template
+	if indexName != "" {
+		return task.RunIndexUpdate()
+	}
+	return task.RunTemplateUpgrade()
```
rodrigozhou: I think we should do both, update the template and the index. Otherwise, creating new indices will not have the right mappings.

Also, based on the `README` you wrote, I think you meant to always upgrade the template as well.

```suggestion
	if err := task.RunTemplateUpgrade(); err != nil {
		return err
	}

	// If index is specified, update index mappings; otherwise update template
	if indexName != "" {
		return task.RunIndexUpdate()
	}
```

rodrigozhou (changes requested):

tools/elasticsearch/handler.go
robholland: Yes, README behaviour was the intention, fixed now.

tools/elasticsearch/main.go
```diff
@@ -0,0 +1,179 @@
...
+				},
+				&cli.BoolFlag{
+					Name:  CLIFlagFailSilently,
+					Usage: "fail silently on HTTP errors",
+				},
+			},
+			Action: func(c *cli.Context) error {
+				cliHandler(c, dropIndex, logger)
+				return nil
+			},
```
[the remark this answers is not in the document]
robholland: That's not true, urfave understands that the flag is set due to the environment variable being present.

PR #8299, opened 2025-09-08
Retry reloading Visibility Manager search attributes cache when cold

rodrigozhou:
## What changed?
If the visibility manager's search attribute cache is cold (ie., never loaded yet), and reading cluster metadata failed (`Unavailable` error), then retry randomly within 5 seconds instead of the default 20 seconds.

## Why?
If the cache is cold and there are custom search attributes, then reading or writing workflows with custom search attributes would fail for 20 seconds. Retrying within 5 seconds to minimize the impact.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/common/searchattribute/manager.go
+++ b/common/searchattribute/manager.go
@@ -93,6 +93,7 @@ func (m *managerImpl) needRefreshCache(saCache cache, forceRefreshCache bool, no
 }
 
 func (m *managerImpl) refreshCache(forceRefreshCache bool, now time.Time) (cache, error) {
+	//nolint:revive // cache value is always of type `cache`
 	saCache := m.cache.Load().(cache)
 	if !m.needRefreshCache(saCache, forceRefreshCache, now) {
 		return saCache, nil
@@ -100,6 +101,7 @@ func (m *managerImpl) refreshCache(forceRefreshCache bool, now time.Time) (cache
 
 	m.cacheUpdateMutex.Lock()
 	defer m.cacheUpdateMutex.Unlock()
+	//nolint:revive // cache value is always of type `cache`
 	saCache = m.cache.Load().(cache)
 	if !m.needRefreshCache(saCache, forceRefreshCache, now) {
 		return saCache, nil


common/searchattribute/manager.go
```diff
@@ -114,12 +92,26 @@ func (m *managerImpl) needRefreshCache(saCache cache, forceRefreshCache bool, no
...
+		return saCache, nil
+	}
+
+	return m.refreshCacheLocked(saCache, now)
+}
+
+func (m *managerImpl) refreshCacheLocked(saCache cache, now time.Time) (cache, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), cacheRefreshTimeout)
+	defer cancel()
+	ctx = headers.SetCallerInfo(ctx, headers.SystemBackgroundHighCallerInfo)
```
yiminc: We should use higher priority (CallerTypeOperator or CallerTypeAPI) if the cache is cold.

common/searchattribute/manager.go
```diff
@@ -128,14 +120,21 @@ func (m *managerImpl) refreshCache(saCache cache, now time.Time) (cache, error)
 			// NotFound means cluster metadata was never persisted and custom search attributes are not defined.
 			// Ignore the error.
 			saCache.expireOn = now.Add(cacheRefreshInterval)
+			err = nil
 		case *serviceerror.Unavailable:
-			// If persistence is Unavailable, ignore the error and use existing cache for cacheRefreshIfUnavailableInterval.
-			saCache.expireOn = now.Add(cacheRefreshIfUnavailableInterval)
-		default:
-			return saCache, err
+			if saCache.dbVersion == 0 {
```
yiminc, line 132: If cache is cold, 1) default refresh interval should be much shorter, 1s or even shorter than that. 2) we should retry on all errors if cache is code, not just Unavailable error, like resource exhausted or timeout errors.

rodrigozhou: Reduced the interval to 1 second.
If there are any other error besides `NotFound` or `Unavailable`, it will retry immediately no matter if the cache is cold or not. This is the existing logic, and I didn't change.

PR #8307, opened 2025-09-09
Unified query converter for Visibility

rodrigozhou:
## What changed?
Unify Visibility query converters.

The new query converter is the struct `query.QueryConverter[ExprT]`. The generic parameter specify the output type of the converter (in ES is `elastic.Query`, in SQL is `sqlparser.Expr`).

The query converter has a `query.StoreQueryConverter[ExprT]`. This is the Visibility store specific implementation for building the output query.

For example, the store needs to implement `BuildAndExpr(exprs)` to tell the query converter how to build the `AND` expression. In SQL, it's `*sqlparser.AndExpr{exprs}`. In Elasticsearch, it's `elastic.NewBoolQuery().Filter(exprs)`.

The store query converter for SQL needs a plugin query converter called `sqlplugin.VisibilityQueryConverter` which is implemented by each plugin (MySQL, PostgreSQL, SQLite). This interface is needed to build DB specific syntax for `KeywordList` and `Text` searches as well as to build the final `SELECT` statements.

Added dynamic config `system.VisibilityEnableUnifiedQueryConverter` that acts as switch between the legacy and the unified query converters.

## Why?
There are two query converters: one for Elasticsearch, and another for SQL.
They are somewhat similar, and have quite a few code duplications for validation, modifying the query, etc.
Unifying the query converter ensures that we provide the exact same behavior across different Visibility stores.

## How did you test it?
- [x] built
- [x] added new unit test(s)

I've added a lot of unit tests for the new code with coverage close to 100%.

Note that there has been basically no changes to functional tests, which suggests that the change works well.

## Potential risks
There might be some edge cases that either query converters that this unified query converter might deal differently.

### DIFF
--- a/service/worker/scheduler/query.go
+++ b/service/worker/scheduler/query.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/persistence/visibility/store/elasticsearch"
 	"go.temporal.io/server/common/persistence/visibility/store/query"
@@ -37,13 +38,42 @@ func newFieldNameAggInterceptor(
 	}
 }
 
+type saAggInterceptor struct {
+	names map[string]struct{}
+}
+
+var _ query.SearchAttributeInterceptor = (*saAggInterceptor)(nil)
+
+func newSaAggInterceptor() *saAggInterceptor {
+	return &saAggInterceptor{
+		names: make(map[string]struct{}),
+	}
+}
+
+func (i *saAggInterceptor) Intercept(col *query.SAColumn) error {
+	i.names[col.Alias] = struct{}{}
+	return nil
+}
+
 func ValidateVisibilityQuery(
 	namespaceName namespace.Name,
 	saNameType searchattribute.NameTypeMap,
 	saMapperProvider searchattribute.MapperProvider,
+	enableUnifiedQueryConverter dynamicconfig.BoolPropertyFn,
 	queryString string,
 ) error {
-	fields, err := getQueryFields(namespaceName, saNameType, saMapperProvider, queryString)
+	var fields []string
+	var err error
+	if enableUnifiedQueryConverter() {
+		fields, err = getQueryFields(
+			namespaceName,
+			saNameType,
+			saMapperProvider,
+			queryString,
+		)
+	} else {
+		fields, err = getQueryFieldsLegacy(namespaceName, saNameType, saMapperProvider, queryString)
+	}
 	if err != nil {
 		return err
 	}
@@ -58,6 +88,36 @@ func ValidateVisibilityQuery(
 }
 
 func getQueryFields(
+	namespaceName namespace.Name,
+	saNameType searchattribute.NameTypeMap,
+	saMapperProvider searchattribute.MapperProvider,
+	queryString string,
+) ([]string, error) {
+	saMapper, err := saMapperProvider.GetMapper(namespaceName)
+	if err != nil {
+		return nil, err
+	}
+	saInterceptor := newSaAggInterceptor()
+	queryConverter := query.NewNilQueryConverter(
+		namespaceName,
+		saNameType,
+		saMapper,
+	).WithSearchAttributeInterceptor(saInterceptor)
+	_, err = queryConverter.Convert(queryString)
+	if err != nil {
+		var converterErr *query.ConverterError
+		if errors.As(err, &converterErr) {
+			return nil, converterErr.ToInvalidArgument()
+		}
+		return nil, err
+	}
+	if !queryConverter.SeenNamespaceDivision() {
+		delete(saInterceptor.names, searchattribute.TemporalNamespaceDivision)
+	}
+	return expmaps.Keys(saInterceptor.names), nil
+}
+
+func getQueryFieldsLegacy(
 	namespaceName namespace.Name,
 	saNameType searchattribute.NameTypeMap,
 	saMapperProvider searchattribute.MapperProvider,
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -4087,6 +4087,7 @@ func (wh *WorkflowHandler) ListSchedules(
 			namespaceName,
 			saNameType,
 			wh.saMapperProvider,
+			wh.config.VisibilityEnableUnifiedQueryConverter,
 			request.Query,
 		); err != nil {
 			return nil, err
--- a/service/worker/scheduler/query_test.go
+++ b/service/worker/scheduler/query_test.go
@@ -4,7 +4,9 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/persistence/visibility/store/query"
 	"go.temporal.io/server/common/searchattribute"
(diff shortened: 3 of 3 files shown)


common/persistence/visibility/store/query/interceptors.go
```diff
@@ -0,0 +1,13 @@
+package query
+
+type SearchAttributeInterceptor interface {
```
rkannan82, line 9: Suggest a comment on the semantics of this interceptor, and what is it used for.

common/persistence/visibility/store/query/nil_converter.go
```diff
@@ -0,0 +1,91 @@
+package query
+
+import (
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/searchattribute"
+)
+
+type nilStoreQueryConverter struct{}
+
+var _ StoreQueryConverter[any] = (*nilStoreQueryConverter)(nil)
```
rkannan82, line 14: Where is this StoreQueryConverter defined?

common/persistence/visibility/store/query/util.go
```diff
@@ -2,11 +2,133 @@ package query
...
 
+type (
+	// UnsafeSQLString don't escape the string value; unlike sqlparser.SQLVal.
+	// This is used for building string known to be safe.
+	UnsafeSQLString struct {
+		sqlparser.Expr
+		Val string
+	}
+
+	ColName struct {
```
rkannan82: Should we call this as Column instead of ColName? Similarly, SAColumn instead of SAColName?
Rationale: this struct encapsulates a column and associated methods.

common/persistence/visibility/store/query/util.go
```diff
@@ -2,11 +2,133 @@ package query
...
+		sqlparser.Expr
+		Name string
+	}
+
+	SAColName struct {
+		sqlparser.Expr
+		Alias     string
+		FieldName string
```
rkannan82, line 32: Comment that FieldName is the actual column name in the table.

common/persistence/visibility/store/query/util.go
```diff
@@ -22,3 +144,22 @@ func ParseExecutionDurationStr(durationStr string) (time.Duration, error) {
...
 }
+
+func ReduceExprs(
+	reduceFunc func(left, right sqlparser.Expr) sqlparser.Expr,
+	exprs ...sqlparser.Expr,
+) sqlparser.Expr {
+	if len(exprs) == 0 {
+		return nil
+	}
+	for n := len(exprs); n > 1; n = (n + 1) / 2 {
```
rkannan82: Just curious as to why we prefer a pairwise reduction compared to a single left fold approach like this:
The pairwise is useful if we want to parallelize the operations, but I don't see that happening here.

```
result := exprs[0]
for i := 1; i < len(exprs); i++ {
    result = reduceFunc(result, exprs[i])
}
```

common/persistence/visibility/store/sql/query_converter_legacy_factory.go
```diff
@@ -8,7 +8,7 @@ import (
 	"go.temporal.io/server/common/searchattribute"
 )
 
-func NewQueryConverter(
+func NewQueryConverterLegacy(
```
rkannan82, line 11: Shouldn't this change and visibility_store.go (below) be part of this parent pull request?
https://github.com/temporalio/temporal/pull/8333:

common/persistence/visibility/store/query/nil_converter.go
rkannan82, line 1: Would it be useful to add a file comment that says:  this is a dummy converter that is used for collecting query metadata via interceptors.

service/worker/scheduler/query.go
```diff
@@ -37,13 +38,44 @@ func newFieldNameAggInterceptor(
 	}
 }
 
+type saAggInterceptor struct {
+	names map[string]bool
```
rkannan82: Suggest a comment stating what this is storing (key and value).
Reason for asking is: The implementation is adding an entry to the map when a column was referenced in the query. It is not clear why we need the bool though. As in, can the value be just an empty struct?

common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -31,6 +31,8 @@ type (
 	// Plugin defines the interface for any SQL database that needs to implement
 	Plugin interface {
 		CreateDB(dbKind DbKind, cfg *config.SQL, r resolver.ServiceResolver, l log.Logger, mh metrics.Handler) (GenericDB, error)
+
+		GetVisibilityQueryConverter() VisibilityQueryConverter
```
rkannan82, line 35: Why do we have this Visibility prefix here?

common/persistence/sql/sqlplugin/postgresql/query_converter.go
rkannan82, line 1: Why isn't this code branched from common/persistence/visibility/store/sql/query_converter_postgresql.go.
Same for other files. I was expecting fewer diff.

common/persistence/visibility/store/query/util.go
rodrigozhou: Yes, no good reason besides keeping the tree height smaller. Just personal preference, nothing against doing left fold.

common/persistence/visibility/store/sql/query_converter_legacy_factory.go
rodrigozhou: This PR is stacked on top of #8333, that's why that PR changes is also reflected here.

service/worker/scheduler/query.go
rodrigozhou: Yes, sure. We don't use the bool value, this is more meant to be a set, but Go doesn't have a native set type.

common/persistence/sql/sqlplugin/interfaces.go
rodrigozhou: Just to be explicit that this query converter inside the `Plugin` is for Visibility. The `Plugin` interface is used to represent the whole DB implementation for both the core and Visibility stores.

common/persistence/sql/sqlplugin/postgresql/query_converter.go
rodrigozhou: The code is branched from those files actually, but since it's a new file, not renaming, git is showing just as new code.

common/persistence/visibility/store/query/nil_converter.go
rodrigozhou: It's defined in `common/persistence/visibility/store/query/converter.go`

common/persistence/visibility/store/query/util.go
rkannan82: - In this impl, what is the significant of tree height?
- The left fold impl is quite simple. That's why I suggested it in favor of this logic. You can make the final call.

common/persistence/visibility/store/query/util.go
```diff
@@ -2,11 +2,136 @@ package query
...
 
+type (
+	// UnsafeSQLString don't escape the string value; unlike sqlparser.SQLVal.
+	// This is used for building string known to be safe.
+	UnsafeSQLString struct {
+		sqlparser.Expr
+		Val string
+	}
+
+	ColumnName struct {
```
rkannan82, line 22: I thought I left a comment earlier on calling this as Column instead of ColumnName. Similarly, SAColumn. 
Rationale was it encapsulates more than just a name.

common/persistence/visibility/store/query/converter.go
```diff
@@ -0,0 +1,713 @@
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination converter_mock.go
```
rkannan82, line 1: Delete this?

common/persistence/visibility/store/query/converter.go
```diff
@@ -0,0 +1,713 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/primitives"
+	"go.temporal.io/server/common/searchattribute"
+	"go.temporal.io/server/common/sqlquery"
+)
+
+type (
+	StoreQueryConverter[ExprT any] interface {
```
rkannan82, line 29: Suggest adding some comments to this interface and the method. For the methods, I think it suffices to give an example of what the input is and what the output is. This will also help me verify the impl.

Sorry, I missed reviewing this file earlier because the diff was collapsed :)

common/persistence/visibility/store/query/converter.go
```diff
@@ -0,0 +1,713 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return queryParams, nil
+}
+
+func (c *QueryConverter[ExprT]) convertWhereString(queryString string) (*QueryParams[ExprT], error) {
+	where := strings.TrimSpace(queryString)
+	if where != "" &&
```
rkannan82, line 196: Should we return an error if the queryString has where clause in it?

common/persistence/visibility/store/query/converter.go
```diff
@@ -0,0 +1,713 @@
...
+
+	//nolint:revive // type cast is guaranteed to be a *sqlparser.Select
+	selectStmt, _ := stmt.(*sqlparser.Select)
+	return c.convertSelectStmt(selectStmt)
+}
+
+func (c *QueryConverter[ExprT]) convertSelectStmt(
+	sel *sqlparser.Select,
+) (*QueryParams[ExprT], error) {
+	// if sel.OrderBy != nil {
```
rkannan82, line 218: delete?

common/persistence/visibility/store/query/converter.go
```diff
@@ -0,0 +1,713 @@
...
+	if sel.Where == nil {
+		sel.Where = &sqlparser.Where{
+			Type: sqlparser.WhereStr,
+			Expr: nil,
+		}
+	}
+
+	res := &QueryParams[ExprT]{}
+	if sel.Where.Expr != nil {
+		// // Wrap the WHERE clause in parenthesis to make sure the added conditions below (namespace and
```
rkannan82: delete?

common/persistence/visibility/store/query/converter.go
```diff
@@ -0,0 +1,713 @@
...
+			"%s: 'GROUP BY' clause supports only a single field",
+			NotSupportedErrMessage,
+		)
+	}
+	for k := range sel.GroupBy {
+		colName, err := c.convertColName(sel.GroupBy[k])
+		if err != nil {
+			return nil, err
+		}
+		if colName.FieldName != searchattribute.ExecutionStatus {
```
rkannan82: Why only ExecutionStatus is allowed? What about other search attributes?

common/persistence/sql/sqlplugin/interfaces.go
rkannan82: Sorry, still unclear. Are you saying for a core store, we need to add another API called GetCoreQueryConverter to this interface?

rkannan82 (approved):

service/worker/scheduler/query.go
rkannan82: Could you just use an empty struct as value instead? is that an idiomatic way to represent a set?

common/persistence/sql/sqlplugin/postgresql/query_converter.go
rkannan82: Let me sync up with you offline on this. For example, could we have done this in 2 steps:
1. Copy the file. Noithing to review.
2. Make the changes? This shows just the diff?

I am just worried there is a lot of code.

common/persistence/sql/sqlplugin/interfaces.go
rodrigozhou: Discussed offline

common/persistence/visibility/store/query/converter.go
rodrigozhou: What do you mean?

common/persistence/visibility/store/query/converter.go
rodrigozhou: This is to auto-generate the `converter_mock.go` file which is used in tests.

common/persistence/visibility/store/query/converter.go
rodrigozhou: We only allow group by ExecutionStatus at this moment because of perf concerns.

common/persistence/visibility/store/query/util.go
rodrigozhou: `ColumnName` just encapsulate a column name. I can rename `SAColumnName` to `SAColumn`.

common/persistence/visibility/store/query/converter.go
rodrigozhou: I added a comment that we want to uncomment these lines in a future release.

common/persistence/visibility/store/query/converter.go
rodrigozhou: Added comments

common/persistence/visibility/store/query/util.go
rodrigozhou: Changed to left fold.

common/persistence/visibility/store/query/converter.go
rkannan82: Optional. Externalize the allowed attributes in a constant with a comment and use it here? Then it is obvious.

common/persistence/visibility/store/query/converter.go
rkannan82: This code assumes the string does not contain where clause. We implicitly add it inside this function. So was asking if we should validate the string does not have it.

rodrigozhou: If it has the `where` string, the parser will fail.

Issue #8298, opened 2025-09-08
Make Nexus work out of the box with zero config


bergundy:
## Start with

- Allow `temporal://system` URLs always to be attached as callbacks.
- Dispatch `temporal://system` callback URLs internally, automatically (see link below).

## Then

- Add a token [here](https://github.com/temporalio/api/blob/0286a6924ef7bd2d31ef1a1141d5e50bd43bb3c6/temporal/api/common/v1/message.proto#L171).
 
## Then update the SPEC and Nexus SDKs

- In the Nexus SPEC.md, call out that the `Nexus-Callback-Token` is a standard header and that it is required as part of the `StartOperation` request.
- In the Nexus SDKs (start with Go), add a `CallbackToken []byte (or string TBD)` field in `StartOpertionOptions` (called details, context) in the other SDKs.

## Once the server can handle these new URLs, switch to using them

- Behind a dynamic config, which can be deprecated in a few releases.
- Default the server's callback URL to `temporal://system`.
  - The callback URL template, must be provided when making "external" calls but "worker" endpoints should use the `temporal://system` callback URL.
- Figure out how to make this work in Cloud.

## Reference
https://github.com/temporalio/temporal/blob/16b1e29b2a86eaf1a95e026e49e16f31f8c414aa/components/callbacks/fx.go#L45
https://github.com/temporalio/temporal/blob/16b1e29b2a86eaf1a95e026e49e16f31f8c414aa/components/nexusoperations/fx.go#L102-L111

chaptersix:
PR: #8308 for first task.

PR #8308, opened 2025-09-09
Support Internal Nexus Callback Routing 

chaptersix:
## What changed?
This PR introduces two key features to enable out-of-the-box Nexus functionality with zero configuration:

1. **Always allow `temporal://system` callback URLs**: The `component.callbacks.allowedAddresses` configuration now automatically allows `temporal://system` URLs regardless of configured address rules. This special scheme+host combination bypasses normal URL validation and is always permitted.

2. **New UseSystemCallbackURL toggle**: Added `component.nexusoperations.useSystemCallbackURL` dynamic config that controls callback URL generation for worker targets:
   - When `false` (default): Uses the existing callback URL template approach
   - When `true`: Worker targets automatically get `temporal://system`
   - External targets continue to use template-generated URLs regardless of this setting
   - NOTE: if this toggle is `true` HTTPCallerProvider must be able to route internal requests based on theSource and Token headers. It should also override the path to `/nexus/callback`. See the HTTPCallerProvider code changes here. 

3. **Callback URL Validation**  has been refactored. The `AddressMatchRules` struct now includes a `Validate` method that handles the special `temporal://system` case and applies configured rules for other URLs.
4. **Callback Source Header**: will no longer be populated for external Targets.

## Why?
Once `component.nexusoperations.useSystemCallbackURL` is enabled by default in a future release, these changes will allow Self-Hosted Temporal Servers to support Nexus with zero dynamic configuration
See this issue: https://github.com/temporalio/temporal/issues/8298

## How did you test it?
- [x] built
- [x] added new unit test(s)


components/callbacks/config.go
```diff
@@ -53,14 +54,27 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
 	[]AddressMatchRule(nil),
 	`The per-namespace list of addresses that are allowed for callbacks and whether secure connections (https) are required.
 URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation.
-Default is no address rules, meaning all callbacks will be rejected. Any invalid entries are ignored. Each entry is a map with possible values:
+Default is no address rules, meaning all callbacks will be rejected, unless they are configured with the "temporal://system. Any invalid entries are ignored. Each entry is a map with possible values:
```
chaptersix:
```suggestion
Default is no address rules, meaning all callbacks will be rejected, unless they are configured with the "temporal://system". Any invalid entries are ignored. Each entry is a map with possible values:
```

components/callbacks/config.go
```diff
@@ -87,8 +101,20 @@ func allowedAddressConverter(val any) ([]AddressMatchRule, error) {
 		configs = append(configs, AddressMatchRule{
 			Regexp:        re,
 			AllowInsecure: e.AllowInsecure,
+			schemes:       []string{"http", "https"},
+			secureSchemes: []string{"https"},
```
chaptersix: I could use a constant as well. not sure where to define those. this package or somewhere else

components/callbacks/config.go
```diff
@@ -53,14 +54,27 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
...
 	AllowInsecure bool
+	schemes       []string
+	secureSchemes []string
+}
+
+func (a AddressMatchRule) MatchHost(host string) bool {
+	return a.Regexp.MatchString(host)
+}
+
+func (a AddressMatchRule) SchemeAllowed(scheme string) bool {
```
chaptersix: could these functions return errors to replace the errors returned by the workflow_handler?

service/frontend/workflow_handler.go
```diff
@@ -5261,12 +5261,12 @@ func (wh *WorkflowHandler) validateCallbackURL(ns namespace.Name, rawURL string)
...
 	}
-	if !(u.Scheme == "http" || u.Scheme == "https") {
+	if !(u.Scheme == "http" || u.Scheme == "https" || u.Scheme == "temporal") {
 		return status.Errorf(codes.InvalidArgument, "invalid url: unknown scheme: %v", u)
 	}
 	for _, cfg := range wh.config.CallbackEndpointConfigs(ns.String()) {
-		if cfg.Regexp.MatchString(u.Host) {
-			if u.Scheme == "http" && !cfg.AllowInsecure {
+		if cfg.MatchHost(u.Host) {
+			if !cfg.SchemeAllowed(u.Scheme) {
```
pdoerner: I don't think you need this check since you are already checking the known schemes above on L5264

components/callbacks/config.go
```diff
@@ -53,16 +54,30 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
 	[]AddressMatchRule(nil),
 	`The per-namespace list of addresses that are allowed for callbacks and whether secure connections (https) are required.
 URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation.
-Default is no address rules, meaning all callbacks will be rejected. Any invalid entries are ignored. Each entry is a map with possible values:
+Default is no address rules, meaning all callbacks will be rejected, unless they are configured with the "temporal://system". Any invalid entries are ignored. Each entry is a map with possible values:
```
pdoerner: I would remove the `unless` and just make it clear that `temporal://system` is always allowed, regardless of what the user sets for this config. Later, we can make `temporal://system` the default value for the config.

components/callbacks/config.go
```diff
@@ -89,6 +104,14 @@ func allowedAddressConverter(val any) ([]AddressMatchRule, error) {
 			AllowInsecure: e.AllowInsecure,
 		})
 	}
+	//used to indicate the callback should be routed internally
+	// other metadata from the client request will be used to route the request to the
+	// correct namespace
+	re, _ := regexp.Compile("^temporal://system$")
```
pdoerner: Since this is effectively a constant, we shouldn't be calling `Compile` every time the dynamic config converter is called.

pdoerner (commented):
I don't think you need to worry about whether to allow insecure connections for `temporal://system` since those internally-routed requests will always use our internal local frontend client which should have already been configured with the correct secure/insecure setting.

service/frontend/workflow_handler.go
chaptersix: SchemeAllowed also checks insecure vs not secure. it's configured on a per endpoint basis. I'd prefer to return an error from here so I can return an error for if the scheme is not support outright or if insecure is just not allowed for the given match. I could just make another method that check if the schema itself is valid.

components/callbacks/config.go
chaptersix: updated but other reviewers please comment.

components/callbacks/config.go
```diff
@@ -52,17 +53,40 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
 	allowedAddressConverter,
 	[]AddressMatchRule(nil),
 	`The per-namespace list of addresses that are allowed for callbacks and whether secure connections (https) are required.
-URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation.
-Default is no address rules, meaning all callbacks will be rejected. Any invalid entries are ignored. Each entry is a map with possible values:
+URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation. URL: "temporal://system" is always allowed.
```
chaptersix:
```suggestion
URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation. URL: "temporal://system" is always valid.
```

components/callbacks/config.go
```diff
@@ -52,17 +53,40 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
 	allowedAddressConverter,
 	[]AddressMatchRule(nil),
 	`The per-namespace list of addresses that are allowed for callbacks and whether secure connections (https) are required.
-URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation.
-Default is no address rules, meaning all callbacks will be rejected. Any invalid entries are ignored. Each entry is a map with possible values:
+URLs are checked against each in order when starting a workflow with attached callbacks and only need to match one to pass validation. URL: "temporal://system" is always valid.
+Default is no address rules. Any invalid entries are ignored. Each entry is a map with possible values:
```
bergundy: Could you just merge the two statements saying that `tempral://system` is always valid and the default is no address rules just to clarify a bit? Also mention that this config is required for external endpoint targets.

components/callbacks/config.go
```diff
@@ -52,17 +53,40 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
...
 		Wildcards, '*', are supported and can match any number of characters (e.g. '*' matches everything, 'prefix.*.domain' matches 'prefix.a.domain' as well as 'prefix.a.b.domain').
 	 - "AllowInsecure":bool (optional, default=false) indicates whether https is required`)
 
+// allowedSchema contains all schema both insecure and secure
+var allowedSchemes = []string{"http", "https", "temporal"}
+
+// allowedSecurSchema contains only secure schemas
+var allowedSecureSchemes = []string{"https", "temporal"}
+
+var matchTemporalSystem, _ = regexp.Compile("^temporal://system$")
```
bergundy: Use `MustCompile` please, we don't want to ignore errors accidentally.

components/callbacks/config.go
```diff
@@ -89,6 +113,12 @@ func allowedAddressConverter(val any) ([]AddressMatchRule, error) {
 			AllowInsecure: e.AllowInsecure,
 		})
 	}
+	// used to indicate the callback should be routed internally
+	// other metadata from the client request will be used to route the request to the
+	// correct namespace
+	configs = append(configs, AddressMatchRule{
```
bergundy: nit: it would be best to put `temporal://system` first since it will likely be the most frequently used.

service/frontend/workflow_handler.go
```diff
@@ -5261,12 +5262,12 @@ func (wh *WorkflowHandler) validateCallbackURL(ns namespace.Name, rawURL string)
...
 	}
-	if !(u.Scheme == "http" || u.Scheme == "https") {
+	if !callbacks.IsSchemeAllowed(u.Scheme) {
 		return status.Errorf(codes.InvalidArgument, "invalid url: unknown scheme: %v", u)
 	}
 	for _, cfg := range wh.config.CallbackEndpointConfigs(ns.String()) {
-		if cfg.Regexp.MatchString(u.Host) {
-			if u.Scheme == "http" && !cfg.AllowInsecure {
+		if cfg.MatchHost(u.Host) {
+			if !cfg.CheckSchemeSecureLevel(u.Scheme) {
```
bergundy: I would change this signature to `cfg.Allow(url) error`.
This will allow you to match `temporal://system` properly because you need to verify both the scheme and the host together and the regexp based validation doesn't work here.

chaptersix: awesome, I was thinking about going this direction.

components/nexusoperations/executors_test.go
```diff
@@ -39,6 +39,13 @@ var endpointEntry = &persistencespb.NexusEndpointEntry{
 	Endpoint: &persistencespb.NexusEndpoint{
 		Spec: &persistencespb.NexusEndpointSpec{
 			Name: "endpoint",
+			Target: &persistencespb.NexusEndpointTarget{
+				Variant: &persistencespb.NexusEndpointTarget_External_{
+					External: &persistencespb.NexusEndpointTarget_External{
+						Url: "temporal://system",
+					},
+				},
+			},
```
chaptersix, line 49: this should likely be changed to use a worker target

components/nexusoperations/executors.go
```diff
@@ -89,6 +89,33 @@ type taskExecutor struct {
 	TaskExecutorOptions
 }
 
+func buildCallbackURL(callbackTemplate string, ns *namespace.Namespace, endpoint *persistencespb.NexusEndpointEntry) (string, error) {
+	switch target := endpoint.GetEndpoint().GetSpec().GetTarget().GetVariant().(type) {
+	case *persistencespb.NexusEndpointTarget_Worker_:
+		return fmt.Sprintf("temporal://system/namespaces/%s/nexus/callback", ns.Name().String()), nil
```
chaptersix: better solution?

chaptersix: likely would not work for SASS.

tests/nexus_workflow_test.go
```diff
@@ -2260,7 +2260,11 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
 	callerWF := func(ctx workflow.Context) error {
 		c := workflow.NewNexusClient(endpointName, svc.Name)
 		fut := c.ExecuteOperation(ctx, op, nil, workflow.NexusOperationOptions{})
-		return fut.Get(ctx, nil)
+		if err := fut.Get(ctx, nil); err != nil {
+			fmt.Println("future not resolved: ", err.Error())
+			return err
+		}
+		return nil
```
chaptersix:
```suggestion
		return fut.Get(ctx, nil)
```

components/callbacks/config.go
```diff
@@ -52,17 +56,58 @@ var AllowedAddresses = dynamicconfig.NewNamespaceTypedSettingWithConverter(
...
 
+// allowedSchema contains all schemes both insecure and secure
+var allowedSchemes = []string{"http", "https", "temporal"}
+
+// allowedSecurSchema contains only secure schemes
+var allowedSecureSchemes = []string{"https", "temporal"}
+
+const (
+	temporalSystemURI = "temporal://system"
+)
```
chaptersix:
```suggestion
```

components/nexusoperations/executors.go
```diff
@@ -89,6 +89,35 @@ type taskExecutor struct {
 	TaskExecutorOptions
 }
 
+func buildCallbackURL(callbackTemplate string, ns *namespace.Namespace, endpoint *persistencespb.NexusEndpointEntry) (string, error) {
+	switch target := endpoint.GetEndpoint().GetSpec().GetTarget().GetVariant().(type) {
+	case *persistencespb.NexusEndpointTarget_Worker_:
+
+		return fmt.Sprintf("temporal://system/%s",
```
chaptersix: does `allowSystemCallbackURL` need to be checked here? If the HTTPProvider does not support this scheme then the callbacks will fail to route.

components/callbacks/config.go
```diff
@@ -47,22 +51,74 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
 	}
 }
 
+var AllowSystemCallbackURL = dynamicconfig.NewNamespaceBoolSetting(
```
chaptersix: Should this be a global bool?

chaptersix: I believe so. changing it to global

components/nexusoperations/executors.go
chaptersix: https://github.com/temporalio/saas-temporal/blob/80c6f2e442f00ecb4939bc00044f3a4a863b9568/nexus/integration/http_caller_provider_factory.go#L114-L114
looks like we could just change this to pull the namespace from the token and change the host validation.

chaptersix: the reason I'm appending the callback URL is the frontend was returning a 404 because there's not a handler registered on the root path. If we'd rather not append the callback URL I could setup the frontend to accept requests at the root path.

components/callbacks/config.go
```diff
@@ -47,22 +51,75 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
 	}
 }
 
+var AllowSystemCallbackURL = dynamicconfig.NewGlobalBoolSetting(
+	"component.callbacks.allowSystemCallbackURL",
+	false,
+	`The global feature toggle that controls whether or not client workers can use "temporal://system" URL. 
+	- false (default): do not allow callback URLS to use the "temporal://system" URL
+	- true: allow callback URLS to use the "temporal://system" URL
+	- The default will switch to true in future releases.`)
```
bergundy: This config should be in `components/nexusoperations/config.go` and be called `UseSystemCallbackURL` (`components.nexusoperations.useSystemCallbackURL`).

The config should control whether the executor uses the system callback URL for worker targets, by default it would use the callback URL template.

components/callbacks/config.go
```diff
@@ -47,22 +51,75 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+
+const (
+	temporalScheme = "temporal"
+	systemHost     = "system"
+)
+
+// allowedSecurSchema contains only secure schemes
+var allowedSecureSchemes = []string{"https", "temporal"}
+
+func IsSchemeAllowed(scheme string, enableSystem bool) bool {
```
bergundy: No need for this, just move the logic for validating the scheme into `AddressMatchRule`.

components/callbacks/config.go
```diff
@@ -47,22 +51,75 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
 }
 
+// Allow assumes the scheme has already been checked with IsSchemeAllowed and will return:
+// 1. true, nil if the provided url matches the rule and passed validation
+// for the given rule.
+// 2. false, nil if the URL does not match the rule.
+// 3. It false, error if there is a match and the URL fails validation
+func (a AddressMatchRule) Allow(u *url.URL) (bool, error) {
+	if a.Regexp == nil {
+
```
bergundy: nit: Remove this empty line.

components/callbacks/config.go
```diff
@@ -47,22 +51,75 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+var allowedSecureSchemes = []string{"https", "temporal"}
+
+func IsSchemeAllowed(scheme string, enableSystem bool) bool {
+	if scheme == temporalScheme && !enableSystem {
+		return false
+	}
+	return slices.Contains(allowedSchemes, scheme)
+}
+
 type AddressMatchRule struct {
```
bergundy, line 97: Technically this could be an interface with two implementations: `systemAddressMatchRule` and `httpAddressMatchRule`, it would be a bit cleaner, but I'll leave that up to you to decide.

components/nexusoperations/executors.go
bergundy: Yes, we need to check the dynamic config (see my comment on the naming of that config).

components/nexusoperations/executors.go
```diff
@@ -116,24 +145,13 @@ func (e taskExecutor) executeInvocationTask(ctx context.Context, env hsm.Environ
...
-		NamespaceID:   ns.ID().String(),
-	})
-	if err != nil {
-		return fmt.Errorf("failed to format callback URL: %w", err)
+		return err
 	}
-	callbackURL := builder.String()
+
+	// Set this value on the parent context so that our custom HTTP caller can mutate it since we cannot access response headers directly.
+	ctx = context.WithValue(ctx, commonnexus.FailureSourceContextKey, &atomic.Value{})
```
bergundy: This is done on `callCtx` now, you haven't resolve your merge conflicts properly.

service/frontend/workflow_handler.go
```diff
@@ -5240,19 +5241,24 @@ func (wh *WorkflowHandler) validateCallbackURL(ns namespace.Name, rawURL string)
 	if len(rawURL) > wh.config.CallbackURLMaxLength(ns.String()) {
 		return status.Errorf(codes.InvalidArgument, "invalid url: url length longer than max length allowed of %d", wh.config.CallbackURLMaxLength(ns.String()))
 	}
+	rules := wh.config.CallbackEndpointConfigs(ns.String())
+	return allowCallbackURL(rawURL, rules, wh.config.AllowSystemCallbackURL())
```
bergundy: This was not what we discussed, always allow system callback URLs on the handler side.

tests/nexus_workflow_test.go
```diff
@@ -2289,7 +2289,7 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
 		require.Len(ct, resp.Callbacks, 1)
 		require.Equal(ct, enumspb.CALLBACK_STATE_FAILED, resp.Callbacks[0].State)
 		require.NotNil(ct, resp.Callbacks[0].LastAttemptFailure)
-		require.Equal(ct, resp.Callbacks[0].LastAttemptFailure.Message, "handler error (NOT_FOUND): workflow execution already completed")
+		require.Equal(ct, "handler error (NOT_FOUND): workflow execution already completed", resp.Callbacks[0].LastAttemptFailure.Message)
```
bergundy, line 2435: Thanks!

tests/testcore/dynamic_config_overrides.go
```diff
@@ -65,6 +65,7 @@ var (
 		dynamicconfig.FrontendEnableWorkerVersioningWorkflowAPIs.Key():      true,
 		dynamicconfig.RefreshNexusEndpointsMinWait.Key():                    1 * time.Millisecond,
 		nexusoperations.RecordCancelRequestCompletionEvents.Key():           true,
-		callbacks.AllowedAddresses.Key():                                    []any{map[string]any{"Pattern": "*", "AllowInsecure": true}},
+		callbacks.AllowedAddresses.Key():                                    []any{map[string]any{}},
```
bergundy: Just don't set this.

tests/xdc/nexus_request_forwarding_test.go
```diff
@@ -78,7 +78,8 @@ func (s *NexusRequestForwardingSuite) SetupSuite() {
 		dynamicconfig.FrontendGlobalNamespaceNamespaceReplicationInducingAPIsRPS.Key(): 1000,
 		dynamicconfig.RefreshNexusEndpointsMinWait.Key():                               1 * time.Millisecond,
 		dynamicconfig.FrontendNexusRequestHeadersBlacklist.Key():                       dynamicconfig.GetTypedPropertyFn(re),
-		callbacks.AllowedAddresses.Key():                                               []any{map[string]any{"Pattern": "*", "AllowInsecure": true}},
+		callbacks.AllowedAddresses.Key():                                               []any{map[string]any{}},
```
bergundy: Remove this override altogether please.

bergundy (commented):
We're getting there, seems like we had a small misunderstanding.

tests/testcore/dynamic_config_overrides.go
chaptersix: Currently the way It's configured is there must be at least 1 element in array for it the validation to actually run. There's a for loop that runs over the slice, if there are zero elements then any URL passed in, including system url will be invalid. 
I'll probably just change the interface to make it fit better.

chaptersix: the setup function that runs during server startup does not run during the test setup.

chaptersix:
> We're getting there, seems like we had a small misunderstanding.

yeah, it seems like I inverted what requires the dynamic config.

components/nexusoperations/executors.go
chaptersix:
```suggestion
```

components/nexusoperations/executors.go
```diff
@@ -208,6 +226,7 @@ func (e taskExecutor) executeInvocationTask(ctx context.Context, env hsm.Environ
 	if callTimeout < e.Config.MinOperationTimeout(ns.Name().String()) {
 		callErr = ErrOperationTimeoutBelowMin
 	} else {
+		// this is the nexus sdk client pointing at the internal frontend
```
chaptersix:
```suggestion
```

components/callbacks/config.go
```diff
@@ -50,30 +54,90 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+// for the given rule.
+// 2. false, nil if the URL does not match the rule.
+// 3. It false, error if there is a match and the URL fails validation
+func (a AddressMatchRule) Allow(u *url.URL) (bool, error) {
+	if a.Regexp == nil {
+		if u.Host == systemHost && u.Scheme == temporalScheme {
+			return true, nil
+		}
+		return false, nil
+	}
```
bergundy: Can remove this now.

components/callbacks/config.go
```diff
@@ -50,30 +54,90 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+	return status.Errorf(codes.InvalidArgument, "invalid url: url does not match any configured callback address: %v", u)
+}
+
 type AddressMatchRule struct {
 	Regexp        *regexp.Regexp
 	AllowInsecure bool
 }
 
-func allowedAddressConverter(val any) ([]AddressMatchRule, error) {
+// Allow assumes the scheme has already been checked with IsSchemeAllowed and will return:
```
bergundy: Can remove this now.

components/callbacks/config_test.go
```diff
@@ -37,3 +38,217 @@ func Test_addressPatternToRegexp(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			u, err := url.Parse(tt.args.rawURL)
+			if err != nil {
+				t.Fatalf("failed to parse url: %v", err)
+			}
+			rules := AddressMatchRules{Rules: tt.args.rules}
+			err = rules.Validate(u)
+			if (err != nil) != tt.wantErr {
```
bergundy: It would be better for the test case to say what error is expected.
You could instead of `wantErr` have a `validateErr: func(t *testing.T, err error)` method on each case.
I'd also encourage you to use `require.ErrorsAs` and `require.Equal` from `testify`.

components/nexusoperations/config.go
```diff
@@ -73,6 +73,17 @@ ScheduleNexusOperation commands with a "nexus_header" field that exceeds this li
 Uses Go's len() function on header keys and values to determine the total size.`,
 )
 
+var UseSystemCallbackURL = dynamicconfig.NewGlobalBoolSetting(
```
bergundy, line 84: Very nice!

components/nexusoperations/executors.go
```diff
@@ -89,6 +89,46 @@ type taskExecutor struct {
...
+	ns *namespace.Namespace,
+	endpoint *persistencespb.NexusEndpointEntry,
+) (string, error) {
+	target := endpoint.GetEndpoint().GetSpec().GetTarget().GetVariant()
+	if !useSystemCallback {
+		return buildCallbackFromTemplate(callbackTemplate, ns)
+	}
+	switch target.(type) {
+	case *persistencespb.NexusEndpointTarget_Worker_:
+		return fmt.Sprintf("temporal://system/%s",
```
bergundy: This should just be `temporal://system` with no formatting.

Let's just change the callback implementation to resolve the namespace name from the namespace ID that's embedded in the token and construct the route appropriately.

components/callbacks/config.go
```diff
@@ -50,30 +54,90 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+	temporalScheme = "temporal"
+	systemHost     = "system"
+)
+
+type AddressMatchRules struct {
+	Rules []AddressMatchRule
+}
+
+func (a AddressMatchRules) Validate(u *url.URL) error {
+	if u.Host == systemHost && u.Scheme == temporalScheme {
```
bergundy: I would validate that the URL is simply `temporal://system` with no path, query, etc...

You may want to just take a string here an construct the URL inside this function.

common/routing/route_test.go
```diff
@@ -110,3 +110,23 @@ func TestNewRoute(t *testing.T) {
 		assert.Equal(t, QualifiedWorkflow{Namespace: "TEST-NAMESPACE", WorkflowID: "TEST-WORKFLOW-ID"}, params)
 	})
 }
+func TestNewRouteNoVariable(t *testing.T) {
+	t.Parallel()
+
+	route := routing.NewBuilder[string]().
+		Constant("nexus", "callback").
+		Build()
```
bergundy: You would want to test the existing route, but I would just not TBH... it's overkill IMHO.

components/callbacks/config.go
```diff
@@ -50,30 +55,81 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+// 2. false, nil if the URL does not match the rule.
+// 3. It false, error if there is a match and the URL fails validation
+func (a AddressMatchRule) Allow(u *url.URL) (bool, error) {
+	if !a.Regexp.MatchString(u.Host) {
+		return false, nil
+	}
+	if a.AllowInsecure {
+		return true, nil
+	}
+	if !slices.Contains(allowedSecureSchemes, u.Scheme) {
```
bergundy: I would just keep this to `== "https"`, it's easier to follow.

common/routing/route_test.go
chaptersix: yeah, I just didn't see any tests for when there's not a route var assigned. Just wanted to confirm it'd worked correctly. 

Overkill using the route builder or to test the external route? 
Constant would be find with me but I went with the existing pattern of defining the route.

components/callbacks/fx.go
```diff
@@ -25,6 +27,47 @@ var Module = fx.Module(
 	fx.Invoke(RegisterExecutor),
 )
 
+var systemCallbackPath = "/" + nexus.RouteCompletionCallbackNoIdentifier.Path("")
+
+func routeInternally(r *http.Request,
```
chaptersix: Once you are happy with the general approach, how do you suggest we break this out? Maybe we have a different HTTPCallerProviderProvider for internally routed requests?

chaptersix: I'm a little rusty on DI.

common/routing/route_test.go
bergundy: Ah, I see... This is just covering the routing package, not the newly defined Nexus routes.

TBH, I would just define this as a constant, I think this whole route abstraction is overkill but up to you.

If you plan on using this abstraction for this simple constant, you may want to use an empty struct as the type argument for this route since there is no string value (`routing.NewBuilder[struct{}]()`) but 🤷.

components/callbacks/config.go
```diff
@@ -50,30 +55,78 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+// allowedSchemes contains all schemes, both insecure and secure
+var allowedSchemes = []string{"http", "https"}
+
+type AddressMatchRules struct {
+	Rules []AddressMatchRule
+}
+
+func (a AddressMatchRules) Validate(u *url.URL) error {
+	// Exact match only; no path, query, or fragment allowed for system URL
+	if u.String() == nexus.SystemCallbackURL {
```
bergundy: Or just accept a string instead of a URL and save the back and forth conversion for the common case. Negligible and not blocking this PR.

components/callbacks/config.go
```diff
@@ -50,30 +55,78 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+		return nil
+	}
+	if !slices.Contains(allowedSchemes, u.Scheme) {
```
bergundy: nit: Could simplify to just compare the schemes but this is fine too.
I typically prefer readability where possible, it's easier to understand the code if the condition was inline IMHO: `u.Scheme != "http" && u.Scheme != "https"`.
These layers tend to add up and you find yourself jumping around multiple times to find how something works.

components/callbacks/fx.go
bergundy: Just move this to a separate file, e.g. `request.go` or something similar, just to avoid non DI code from cluttering the DI code.

components/callbacks/fx.go
bergundy, line 1: We should stop sending the source header for external targets (safe to change since those are experimental), we should assume that for an external target, the handler doesn't have any use for this header.

Then, on the callback path (in this file), if we don't know the caller cluster when iterating the metadata, and the caller set the source header, we should use the local frontend client instead of the default HTTP client to make the call.
This will be more resilient in situations where a cluster was removed from the group of clusters and the namespace was migrated since the initial start call was made.

components/callbacks/fx.go
```diff
@@ -25,6 +27,47 @@ var Module = fx.Module(
...
+			}
+			r.URL.Scheme = frontendClient.Scheme
+			r.URL.Host = frontendClient.Address
+			r.URL.Path = systemCallbackPath
+			r.Host = frontendClient.Address
+
+			return frontendClient.Do(r)
+		}
+	}
+	return nil, errors.New("could not find frontend client for request")
```
bergundy: See my comment above and how we've handled this below with a warning, just instead of falling back to the default client, fall back to the local client.

components/nexusoperations/build_callback_url_test.go
```diff
@@ -0,0 +1,61 @@
+package nexusoperations
```
bergundy, line 1: Typically we prefer using `<package>_test` for test packages to avoid testing internal implementation details.

components/callbacks/fx.go
```diff
@@ -46,6 +89,14 @@ func HTTPCallerProviderProvider(
...
+			if r.URL.String() == nexus.SystemCallbackURL {
+				return routeInternally(r,
+					clusterMetadata,
+					httpClientCache,
+					localClient,
+					logger,
+				)
+			}
 
 			callbackSource := r.Header.Get(callbackSourceHeader)
```
bergundy: Some of this logic can be merged with the `routeInternally` function.
AFAICT, the only difference is how we set the URL path.

common/routing/route_test.go
chaptersix: Sounds good, I'll use a constant.

components/callbacks/config.go
chaptersix: fair point. made sense when I was validating the temporal scheme as well but overkill now.

components/nexusoperations/build_callback_url_test.go
chaptersix: interesting, I thought units test are for testing small implementation details. part of the reason I pulled out this logic info an un-exported function is to make is easier to test. Far less mocking.

chaptersix: and I don't think the nexus executors would be the proper place to have an exported function for building the callback url.

components/callbacks/fx.go
```diff
@@ -8,14 +8,10 @@ import (
 	"go.temporal.io/server/common/cluster"
 	"go.temporal.io/server/common/collection"
 	"go.temporal.io/server/common/log"
-	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/service/history/queues"
 	"go.uber.org/fx"
 )
 
-// Header key used to identify callbacks that originate from and target the same cluster.
-const callbackSourceHeader = "source"
```
chaptersix, line 17: I believe this was a bug. The header set in in the Client is: `Nexus-Callback-Source`. @bergundy could you confirm?

bergundy: This is correct `Nexus-Callback-Source` on the StartOperation request should be delivered back as a callback header in the callback request as `Source`.

See:

https://github.com/nexus-rpc/api/blob/main/SPEC.md#request-headers

common/nexus/routes.go
```diff
@@ -27,3 +27,9 @@ var RouteCompletionCallback = routing.NewBuilder[string]().
 	StringVariable("namespace", func(namespace *string) *string { return namespace }).
 	Constant("nexus", "callback").
 	Build()
+
+// RouteCompletionCallbackNoIdentifier is an HTTP route for completing a Nexus operation via callback.
+// Unlike  RouteCompletionCallback it does not require any identifying information to be provided in the
+// URL path. Instead the handler for this path is expected to look up identifying information through
+// via the callback token in the headers.
+var RouteCompletionCallbackNoIdentifier = "/nexus/callback"
```
bergundy: nit: Can call this `Path` instead of `Route` to distinguish this from the `Route`s that are of a different type.

components/callbacks/config.go
```diff
@@ -50,30 +54,79 @@ func ConfigProvider(dc *dynamicconfig.Collection) *Config {
...
+}
+
+func (a AddressMatchRules) Validate(rawURL string) error {
+	// Exact match only; no path, query, or fragment allowed for system URL
+	if rawURL == nexus.SystemCallbackURL {
+		return nil
+	}
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return err
```
bergundy: Make sure that this is an invalid argument, and not an internal error.

```suggestion
			return status.Errorf(codes.InvalidArgument, "invalid callback url: %v", err)
```

components/callbacks/request.go
```diff
@@ -0,0 +1,84 @@
...
+const callbackSourceHeader = "source"
+
+func routeRequest(
+	r *http.Request,
+	clusterMetadata cluster.Metadata,
+	httpClientCache *cluster.FrontendHTTPClientCache,
+	defaultCilent *http.Client,
+	localClient *common.FrontendHTTPClient,
+	logger log.Logger,
+
```
bergundy: nit:

```suggestion
```

components/callbacks/request.go
```diff
@@ -0,0 +1,84 @@
...
+				r,
+				httpClientCache,
+				localClient,
+				logger,
+				clusterMetadata.GetCurrentClusterName(),
+				clusterName,
+			)
+		}
+	}
+	// this should not happen
```
bergundy: This can happen if the cluster was removed from the group, and we should do the same request path manipulation that we do in `execRequest` here.
Also change the log level to warning.

components/nexusoperations/config.go
```diff
@@ -81,6 +81,17 @@ ScheduleNexusOperation commands with a "nexus_header" field that exceeds this li
...
+	"component.nexusoperations.useSystemCallbackURL",
+	false,
+	`UseSystemCallbackURL is a global feature toggle that controls how the executor generates callback URLs for worker targets in Nexus Operations.
+When set to true, the executor will use the fixed system callback URL ("temporal://system") for all worker targets, instead of generating URLs from the callback URL template.
```
bergundy: nit: Keep lines under 120 characters.

tests/nexus_workflow_test.go
```diff
@@ -2581,8 +2657,9 @@ func (s *NexusWorkflowTestSuite) sendNexusCompletionRequest(
 
 	res, err := http.DefaultClient.Do(req)
 	require.NoError(t, err)
-	_, err = io.ReadAll(res.Body)
+	by, err := io.ReadAll(res.Body)
 	require.NoError(t, err)
+	fmt.Println(string(by))
```
bergundy:
```suggestion
```

bergundy (commented):
Had some comments, I still need to review the test but wanted to go ahead and submit.

components/callbacks/request.go
chaptersix: should I log anything in this case?

components/callbacks/config.go
chaptersix: it's actually pretty hard to hit this condition. Any string that that doesn't match the scheme://host will just be added to the path. I added a check for the host being populated though. 

I assume what you meant by "make sure this is an invalid argument, and not an internal error." is add a unit test.

tests/nexus_workflow_test.go
```diff
@@ -673,10 +677,23 @@ func (s *NexusWorkflowTestSuite) TestNexusOperationAsyncCompletion() {
 
 	// Send an invalid completion request and verify that we get an error that the namespace in the URL doesn't match the namespace in the token.
 	invalidCallbackUrl := "http://" + s.HttpAPIAddress() + "/" + commonnexus.RouteCompletionCallback.Path(invalidNamespace)
-	res, snap = s.sendNexusCompletionRequest(ctx, s.T(), invalidCallbackUrl, completion, callbackToken)
+
+	// Make the request directly to check the response body
+	req, err := nexus.NewCompletionHTTPRequest(ctx, invalidCallbackUrl, completion)
```
bergundy: Instead you can use something like `readAndReplaceBody` in `sendNexusCompletionRequest`:
https://github.com/nexus-rpc/sdk-go/blob/ce59b2c82f021201b438cffa1fe52b2e5507039f/nexus/client.go#L419

tests/nexus_workflow_test.go
```diff
@@ -1222,8 +1239,12 @@ func (s *NexusWorkflowTestSuite) TestNexusOperationAsyncCompletionErrors() {
 	})
 
 	s.Run("NamespaceNotFound", func() {
+		// Generate a token with a non-existent namespace ID
+		tokenWithBadNamespace, err := s.generateValidCallbackToken("namespace-doesnt-exist-id", testcore.RandomizeStr("workflow"), uuid.NewString())
+		s.NoError(err)
+
 		publicCallbackUrl := "http://" + s.HttpAPIAddress() + "/" + commonnexus.RouteCompletionCallback.Path("namespace-doesnt-exist")
```
bergundy: Want to do a test for the new route too?
You can run each of these cases for the static and parameterized routes.

chaptersix:
> In general it looks like we are missing functional test coverage for the new `/nexus/callback` route. Correct me if I'm wrong.

explicitly yes but any test with a worker target for the callback exercises that new endpoint. I'll add a few tests to explicitly test the new endpoint.

components/callbacks/request.go
```diff
@@ -0,0 +1,88 @@
...
+	// we need this check while there is a config to toggle
+	// systemURL usage
+	if r.URL.String() == nexus.SystemCallbackURL {
+		r.URL.Path = nexus.PathCompletionCallbackNoIdentifier
+	}
+	r.URL.Scheme = frontendClient.Scheme
+	r.URL.Host = frontendClient.Address
+	r.Host = frontendClient.Address
+	return frontendClient.Do(r)
+}
```
bergundy, line 71: You missed manipulating the URL.
I just make it all take a single path with fallbacks.

```suggestion
func routeRequest(
	r *http.Request,
	clusterMetadata cluster.Metadata,
	httpClientCache *cluster.FrontendHTTPClientCache,
	defaultCilent *http.Client,
	localClient *common.FrontendHTTPClient,
	logger log.Logger,
) (*http.Response, error) {
	// This source header is populated in nexusoperations/executors (via the ClientProvider) for worker targets
	// if this header is not populated then we assume it's and external target.
	if r.Header == nil || r.Header.Get(callbackSourceHeader) == "" {
		return defaultCilent.Do(r)
	}
	// If we got here, we assume that the endpoint in the original call was a worker target, and we should route
	// internally, either to a local frontend, or one of the other connected clusters' frontends.
	var frontendClient *common.FrontendHTTPClient
	callbackSource := r.Header.Get(callbackSourceHeader)
	for clusterName, clusterInfo := range clusterMetadata.GetAllClusterInfo() {
		if callbackSource == clusterInfo.ClusterID {
			fec, err := httpClientCache.Get(clusterName)
			if err != nil {
				logger.Warn(
					"HTTPCallerProvider unable to get FrontendHTTPClient for callback target cluster. Using local HTTP Client.",
					tag.SourceCluster(clusterMetadata.GetCurrentClusterName()),
					tag.TargetCluster(clusterName),
					tag.Error(err),
				)
				frontendClient = localClient
			} else {
				frontendClient = fec
			}
			break
		}
	}
	if frontendClient == nil {
		// This can happen when a cluster is disconnected.
		logger.Warn(
			"HTTPCallerProvider unable to find the target cluster. Using local HTTP Client.",
			tag.SourceCluster(clusterMetadata.GetCurrentClusterName()),
		)
		frontendClient = localClient
	}

	if r.URL.String() == nexus.SystemCallbackURL {
		r.URL.Path = nexus.PathCompletionCallbackNoIdentifier
	}
	r.URL.Scheme = frontendClient.Scheme
	r.URL.Host = frontendClient.Address
	r.Host = frontendClient.Address
	return frontendClient.Do(r)
}
```

bergundy (approved):
Approved with one blocking suggestion. I'm just going to apply it myself.

components/callbacks/request.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	if r.URL.String() == nexus.SystemCallbackURL {
+		r.URL.Path = nexus.PathCompletionCallbackNoIdentifier
+	}
+	r.URL.Scheme = frontendClient.Scheme
+	r.URL.Host = frontendClient.Address
+	r.Host = frontendClient.Address
+	return frontendClient.Do(r)
+}
```
bergundy, line 71:
```suggestion
func routeRequest(
	r *http.Request,
	clusterMetadata cluster.Metadata,
	httpClientCache *cluster.FrontendHTTPClientCache,
	defaultClient *http.Client,
	localClient *common.FrontendHTTPClient,
	logger log.Logger,
) (*http.Response, error) {
	// This source header is populated in nexusoperations/executors (via the ClientProvider) for worker targets
	// if this header is not populated then we assume it's and external target.
	if r.Header == nil || r.Header.Get(callbackSourceHeader) == "" {
		return defaultClient.Do(r)
	}
	// If we got here, we assume that the endpoint in the original call was a worker target, and we should route
	// internally, either to a local frontend, or one of the other connected clusters' frontends.
	var frontendClient *common.FrontendHTTPClient
	callbackSource := r.Header.Get(callbackSourceHeader)
	for clusterName, clusterInfo := range clusterMetadata.GetAllClusterInfo() {
		if callbackSource == clusterInfo.ClusterID {
			if clusterMetadata.GetCurrentClusterName() == clusterName {
				frontendClient = localClient
			} else {
				fec, err := httpClientCache.Get(clusterName)
				if err != nil {
					logger.Warn(
						"HTTPCallerProvider unable to get FrontendHTTPClient for callback target cluster. Using local HTTP Client.",
						tag.SourceCluster(clusterMetadata.GetCurrentClusterName()),
						tag.TargetCluster(clusterName),
						tag.Error(err),
					)
					frontendClient = localClient
				} else {
					frontendClient = fec
				}
			}
			break
		}
	}
	if frontendClient == nil {
		// This can happen when a cluster is disconnected.
		logger.Warn(
			"HTTPCallerProvider unable to find the target cluster. Using local HTTP Client.",
			tag.SourceCluster(clusterMetadata.GetCurrentClusterName()),
		)
		frontendClient = localClient
	}

	if r.URL.String() == nexus.SystemCallbackURL {
		r.URL.Path = nexus.PathCompletionCallbackNoIdentifier
	}
	r.URL.Scheme = frontendClient.Scheme
	r.URL.Host = frontendClient.Address
	r.Host = frontendClient.Address
	return frontendClient.Do(r)
}
```

PR #8309, opened 2025-09-09
fix: correct typo on logs

leopic:
## What changed?
Fixing minor typo on the word namespace.

## Why?
Because it was wrong and now it's right.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)
- [x] I didn't 😬

## Potential risks
Not sure there are any.


common/persistence/visibility/visiblity_manager_metrics.go
```diff
@@ -122,7 +122,7 @@ func (m *visibilityManagerMetrics) ListWorkflowExecutions(
 		m.logger.Warn("List query exceeded threshold",
 			tag.NewDurationTag("duration", elapsed),
 			tag.NewStringTag("visibility-query", request.Query),
-			tag.NewStringerTag("namepsace", request.Namespace),
+			tag.NewStringerTag("Namespace", request.Namespace),
```
yycptt: make it lower case?

yycptt (approved):
Thanks for the contribution!

common/persistence/visibility/visiblity_manager_metrics.go
leopic: Updated!

PR #8310, opened 2025-09-09
Add metrics for data loss events

prathyushpv:
## What changed?
Add metrics for dataloss events that containe workflowID, runID and namespaceID.
This metric will only be emitted if config `system.enableDataLossMetrics` is enabled.

## Why?
To find affected workflow runs if there are any dataloss errors.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1027,6 +1027,11 @@ var (
 		"dlq_message_count",
 		WithDescription("The number of messages currently in DLQ."),
 	)
+	DataLossCounter = NewCounterDef(
+		"data_loss_errors",
+		WithDescription("Total number of data loss errors encountered. This is a high cardinality metrics that is only"+
+			"emitted when system.enableDataLossMetrics is enabled. Only enable this if metrics system can handle it's cardinality"),
+	)
 	ReadNamespaceErrors                     = NewCounterDef("read_namespace_errors")
 	RateLimitedTaskRunnableWaitTime         = NewTimerDef("rate_limited_task_runnable_wait_time")
 	CircuitBreakerExecutableBlocked         = NewCounterDef("circuit_breaker_executable_blocked")
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1029,8 +1029,8 @@ var (
 	)
 	DataLossCounter = NewCounterDef(
 		"data_loss_errors",
-		WithDescription("Total number of data loss errors encountered. This is a high cardinality metrics that is only"+
-			"emitted when system.enableDataLossMetrics is enabled. Only enable this if metrics system can handle it's cardinality"),
+		WithDescription("Total number of data loss errors encountered. This is a high cardinality metrics that has namespace, workflowID and runID tags."+
+			"It is only emitted when system.enableDataLossMetrics is enabled. Only enable this if metrics system can handle it's cardinality"),
 	)
 	ReadNamespaceErrors                     = NewCounterDef("read_namespace_errors")
 	RateLimitedTaskRunnableWaitTime         = NewTimerDef("rate_limited_task_runnable_wait_time")
--- a/common/persistence/client/fx.go
+++ b/common/persistence/client/fx.go
@@ -52,6 +52,7 @@ type (
 		Logger                             log.Logger
 		HealthSignals                      persistence.HealthSignalAggregator
 		DynamicRateLimitingParams          DynamicRateLimitingParams
+		EnableDataLossMetrics              dynamicconfig.BoolPropertyFn
 	}
 
 	FactoryProviderFn func(NewFactoryParams) Factory
@@ -74,6 +75,7 @@ var Module = fx.Options(
 	fx.Provide(HealthSignalAggregatorProvider),
 	fx.Provide(persistence.NewDLQMetricsEmitter),
 	fx.Provide(EventBlobCacheProvider),
+	fx.Provide(EnableDataLossMetricsProvider),
 )
 
 func ClusterNameProvider(config *cluster.Config) ClusterName {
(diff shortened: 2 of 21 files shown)


common/metrics/metric_defs.go
```diff
@@ -1027,6 +1027,11 @@ var (
 		"dlq_message_count",
 		WithDescription("The number of messages currently in DLQ."),
 	)
+	DataLossCounter = NewCounterDef(
+		"data_loss_errors",
+		WithDescription("Total number of data loss errors encountered. This is a high cardinality metrics that is only"+
```
yycptt: nit: maybe mentioned the tags here?

prathyushpv: right. Added that.

prathyushpv:
> Ideally all persistence api should emit this metric on dataloss, but looks like we are just applying it to selected apis right now. so even if we lack codegen today for persistence clients it should still be doable and the complexity should be similar to what you have in this PR?

Yes. That makes sense. I have moved it to persistence rate limited clients. There are a few places in getWorkflowExecution history where we validate history events and return dataloss outside persistence layer. I have kept the code to emit metrics from those places.

common/persistence/client/fx.go
```diff
@@ -52,6 +52,7 @@ type (
 		Logger                             log.Logger
 		HealthSignals                      persistence.HealthSignalAggregator
 		DynamicRateLimitingParams          DynamicRateLimitingParams
+		EnableDataLossMetrics              dynamicconfig.BoolPropertyFn
```
yycptt: nit: may want to define a separate type for fx injection...

common/persistence/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"go.temporal.io/server/common/metrics"
+)
+
+const (
+	WorkflowIDTagName = "workflow_id"
+	RunIDTagName      = "run_id"
+	ErrorTagName      = "error"
+)
```
yycptt, line 11: nit: is it intentional to define them here instead of in `metrics/tag.go`?

service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -242,13 +244,31 @@ func Invoke(
 				}
 				// GetHistory func will not return empty history. Log workflow details if that is not the case
 				if len(history.Events) == 0 {
+					ns, err := shardContext.GetNamespaceRegistry().GetNamespaceName(namespaceID)
+					if err != nil {
+						shardContext.GetLogger().Error("failed to get namespace name from namespace ID for emitting data loss metric",
+							tag.WorkflowNamespaceID(namespaceID.String()))
+					}
```
yycptt: namespace name should be available from the request? `request.Request.GetNamespace()`

service/history/api/get_history_util.go
```diff
@@ -35,7 +36,28 @@ func GetRawHistory(
...
+					execution.GetWorkflowId(),
+					execution.GetRunId(),
+					"GetRawHistory",
+					retError,
+				)
+			}
+		}
+	}()
+
 	shardID := common.WorkflowIDToHistoryShard(namespaceID.String(), execution.GetWorkflowId(), shardContext.GetConfig().NumberOfShards)
```
yycptt: not related to this PR but shardID can be retrieved from shardContext directly

service/history/api/get_history_util.go
```diff
@@ -35,7 +36,28 @@ func GetRawHistory(
...
-) ([]*commonpb.DataBlob, []byte, error) {
+) (_ []*commonpb.DataBlob, _ []byte, retError error) {
+	defer func() {
+		var dataLossErr *serviceerror.DataLoss
+		if errors.As(retError, &dataLossErr) {
+			ns, err := shardContext.GetNamespaceRegistry().GetNamespaceName(namespaceID)
+			if err != nil {
+				shardContext.GetLogger().Error("failed to get namespace name from namespace ID for emitting data loss metric",
+					tag.WorkflowNamespaceID(namespaceID.String()))
+			}
```
yycptt: nit: I think callers (the api handlers) can specify namespace name.

service/history/api/get_history_util.go
```diff
@@ -233,10 +268,37 @@ func GetHistoryReverse(
...
-) (*historypb.History, []byte, int64, error) {
+) (history *historypb.History, token []byte, lastFirstTransactionID int64, retError error) {
+	ns, err := shardContext.GetNamespaceRegistry().GetNamespaceName(namespaceID)
+	if err != nil {
+		return nil, nil, 0, err
+	}
+
+	defer func() {
+		var dataLossErr *serviceerror.DataLoss
+		if errors.As(retError, &dataLossErr) {
```
yycptt: shall we check for SerializationError here as well?

common/persistence/persistence_metrics_clients_test.go
yycptt, line 1: 👍

yycptt (approved):

common/persistence/metrics.go
prathyushpv: Yeah, I did not want any other metrics to accidently use these tags, or someone else to later think that these are valid tags for metrics. That is why I defined them separately here.

service/history/api/getworkflowexecutionhistory/api.go
prathyushpv: Thanks! changed code to use that.

service/history/api/get_history_util.go
prathyushpv: do you mean pass both name and ID from the caller?  We need namespace ID anyways in caller and this function.

common/persistence/client/fx.go
prathyushpv: Done!

service/history/api/get_history_util.go
prathyushpv: Yes. We are doing that in other function. Added that.

PR #8317, opened 2025-09-12
Add service specific dial options in rpc factory

awln-temporal:
### What changed?
This change introduces options for custom dial options per Service Name. No behavior will change in OSS temporal.

### How did you test it?
Link to a test plan if any -
- [X] Unit Tests
- [ ] Staging
- [X] End to End Tests


common/rpc/rpc.go
```diff
@@ -237,7 +240,8 @@ func (d *RPCFactory) createInternodeGRPCConnection(hostName string, serviceName
 			return nil
 		}
 	}
-	c := d.dial(hostName, tlsClientConfig, d.getClientKeepAliveConfig(serviceName))
+	additionalDialOptions := append([]grpc.DialOption{}, d.serviceDialOptions[serviceName]...)
```
yycptt: we need to have a story when someone specifies `serviceDialOptions` that uses FrontendService as the key. `createInternodeGRPCConnection` is only used for history & matching client I think.

yycptt: nit: probably also need to handle `d.serviceDialOptions` == nil case.

common/rpc/rpc.go
```diff
@@ -68,20 +69,22 @@ func NewFactory(
 	frontendHTTPPort int,
 	frontendTLSConfig *tls.Config,
 	dialOptions []grpc.DialOption,
+	serviceDialOptions map[primitives.ServiceName][]grpc.DialOption,
```
yycptt: nit: personal preference. Feel free to ignore.
```suggestion
	commonDialOptions []grpc.DialOption,
	perServiceDialOptions map[primitives.ServiceName][]grpc.DialOption,
```

yycptt:
Please do not link internal tickets/PRs in OSS

common/rpc/rpc.go
awln-temporal: Yea, I will add the same behavior for FrontendServices. As for the nil case, golang handles nil variadic objects, there shouldn't be any issues if serviceDialOption is not set.

common/resource/fx.go
```diff
@@ -385,7 +392,7 @@ func FrontendHTTPClientCacheProvider(
 	return cluster.NewFrontendHTTPClientCache(metadata, tlsConfigProvider)
 }
 
-func getFrontendConnectionDetails(
+func GetFrontendConnectionDetails(
```
yycptt: nit: no longer need to export this?

awln-temporal: yea, will update along with unit tests

PR #8324, opened 2025-09-15
Changed nexus executor failure source to use correct context.

fretz12:
## What changed?
- Changed failure source to use correct context.
- Handled request timeouts avoid exposing internal info to user. 

## Why?
In the nexus executor, the ctx when setting the failure source is not using the call ctx, but the original request context. This should be switched to use the call ctx for consistency. 
During executor request timeouts, we should not expose internal info to the user. To address it, we are setting a new error with a message that doesn't contain  such info.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/components/nexusoperations/executors.go
+++ b/components/nexusoperations/executors.go
@@ -32,6 +32,7 @@ import (
 
 var ErrOperationTimeoutBelowMin = errors.New("remaining operation timeout is less than required minimum")
 var ErrInvalidOperationToken = errors.New("invalid operation token")
+var errRequestTimedOut = errors.New("request timed out")
 
 // ClientProvider provides a nexus client for a given endpoint.
 type ClientProvider func(ctx context.Context, namespaceID string, entry *persistencespb.NexusEndpointEntry, service string) (*nexus.HTTPClient, error)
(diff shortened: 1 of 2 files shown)


components/nexusoperations/executors.go
```diff
@@ -408,27 +409,33 @@ func (e taskExecutor) handleStartOperationError(env hsm.Environment, node *hsm.N
 	var handlerErr *nexus.HandlerError
 	var opFailedErr *nexus.OperationError
 
-	if errors.As(callErr, &opFailedErr) {
+	switch {
```
fretz12, line 412: FYI refactored to address `max-control-nesting: control flow nesting exceeds 5`

components/nexusoperations/executors.go
```diff
@@ -408,27 +409,33 @@ func (e taskExecutor) handleStartOperationError(env hsm.Environment, node *hsm.N
...
+	case errors.As(callErr, &opFailedErr):
 		return handleOperationError(node, operation, opFailedErr)
-	} else if errors.As(callErr, &handlerErr) {
+	case errors.As(callErr, &handlerErr):
+		// The StartOperation request got an unexpected response that is not retryable, fail the operation.
+		// Although Failure is nullable, Nexus SDK is expected to always populate this field
 		if !handlerErr.Retryable() {
```
bergundy:
```suggestion
	case errors.As(callErr, &handlerErr) && !handlerErr.Retryable():
		// The StartOperation request got a non-retryable handler error, fail the operation.
```

bergundy: This suggestion won't apply as is because of the extra closing brace below but GH wouldn't let me add a suggestion on deleted lines.

fretz12: Used your suggestion.

PR #8325, opened 2025-09-15
Use locally installed protoc plugins

stephanos:
## What changed?

~Enforce `temporal/cmd/tools/protogen` to set the PATH to point to the locally installed tools in `.bin`.~

Enforce `temporal/cmd/tools/protogen` to use locally installed protoc plugins.

## Why?

Without this, it will use whatever is on the PATH; which might not be the same as in `.bin`.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

Without this change, I get different `.pb.go` files emitted as my global proto-gen-go is used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,10 +214,8 @@ WORKFLOWCHECK := $(LOCALBIN)/workflowcheck-$(WORKFLOWCHECK_VER)
 $(WORKFLOWCHECK): | $(LOCALBIN)
 	$(call go-install-tool,$(WORKFLOWCHECK),go.temporal.io/sdk/contrib/tools/workflowcheck,$(WORKFLOWCHECK_VER))
 
-# The following tools need to have a consistent name, so we use a versioned stamp file to ensure the version we want is installed
-# while installing to an unversioned binary name.
 GOIMPORTS_VER := v0.36.0
-GOIMPORTS := $(LOCALBIN)/goimports
+GOIMPORTS := $(LOCALBIN)/goimports-$(GOIMPORTS_VER)
 $(STAMPDIR)/goimports-$(GOIMPORTS_VER): | $(STAMPDIR) $(LOCALBIN)
 	$(call go-install-tool,$(GOIMPORTS),golang.org/x/tools/cmd/goimports,$(GOIMPORTS_VER))
 	@touch $@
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -307,10 +307,8 @@ protoc: $(PROTOGEN) $(MOCKGEN) $(GOIMPORTS) $(PROTOC_GEN_GO) $(PROTOC_GEN_GO_GRP
 		-root=$(ROOT) \
 		-proto-out=$(PROTO_OUT) \
 		-proto-root=$(PROTO_ROOT) \
-		-protogen=$(PROTOGEN) \
 		-api-binpb=$(API_BINPB) \
-		-goimports=$(GOIMPORTS) \
-		-mockgen=$(MOCKGEN) \
+		-tool-bin-dir=$(ROOT)/$(LOCALBIN) \
```
stephanos: Seems easier to just point at a single tools dir here. Especially since proto-gen-go is picked up through PATH; not an explicit path.

Makefile
```diff
@@ -200,7 +200,7 @@ $(BUF): | $(LOCALBIN)
 
 GO_API_VER = $(shell go list -m -f '{{.Version}}' go.temporal.io/api \
 	|| (echo "failed to fetch version for go.temporal.io/api" >&2))
-PROTOGEN := $(LOCALBIN)/protogen-$(GO_API_VER)
+PROTOGEN := $(LOCALBIN)/protogen
```
bergundy: Why? All of the other external tools append their version to the binary.

Note that the api repo also has a tool called protogen.

Makefile
bergundy: But those tools are installed as binaries with the version name appended to them, we need to pass those to protogen for it to work properly.

stephanos: Well, `goimports` and `mockgen` and the protoc plugins are already installed without a version in their file name. 

Screenshot 2025-09-15 at 11 03 00 AM

stephanos: Let me see what api-go's protogen's flags are ...

Makefile
stephanos: > Why?

I don't think it's completely unreasonable to use the existing `.bin` dir (where non-versioned binaries are placed) and point the tool at that directory?

> Note that the api repo also has a tool called protogen.

Right; this _is_ that tool.

stephanos:
I've changed it from using PATH to explicitly stating every binary. TBH, I think the PATH approach was easier/simpler; but this approach - while more complex - is more explicit. Another benefit is that accidentally using an old version is now impossible.

bergundy (approved):
Thanks!

PR #8326, opened 2025-09-15
Add tests to improve coverage for worker heartbeat implementation.

rkannan82:
## What changed?
Tests for the following cases:
1. Cache eviction logic.
2. ListWorkers API with query predicate.

## Why?
Improve coverage,

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None.

### DIFF
--- a/service/matching/workers/registry_impl_test.go
+++ b/service/matching/workers/registry_impl_test.go
@@ -112,6 +112,62 @@ func TestEvictByCapacity(t *testing.T) {
 	assert.LessOrEqual(t, m.total.Load(), maxItems, "total counter should not exceed maxItems")
 }
 
+// Tests the critical edge case where evictByCapacity() cannot evict any entries because they're all
+// protected by minEvictAge. This verifies the "break" logic prevents infinite loops during memory
+// pressure.
+func TestEvictByCapacityWithMinAgeProtection(t *testing.T) {
+	maxItems := int64(2)
+	minEvictAge := 5 * time.Second
+	m := newRegistryImpl(1, time.Hour, minEvictAge, maxItems, time.Hour)
+	defer m.Stop()
+
+	// Add 3 entries (over capacity) - all will be "new" (< minEvictAge)
+	for i := 1; i <= 3; i++ {
+		key := fmt.Sprintf("worker%d", i)
+		hb := &workerpb.WorkerHeartbeat{WorkerInstanceKey: key}
+		m.upsertHeartbeats("ns", []*workerpb.WorkerHeartbeat{hb})
+	}
+
+	// Verify we're over capacity
+	assert.Equal(t, int64(3), m.total.Load(), "should have 3 entries initially")
+	assert.Greater(t, m.total.Load(), maxItems, "should be over capacity")
+
+	// Attempt eviction - should break because all entries are too new
+	m.evictByCapacity()
+
+	// All entries should still be there (protected by minEvictAge)
+	workers := m.filterWorkers("ns", alwaysTrue)
+	assert.Len(t, workers, 3, "all entries should be protected by minEvictAge")
+	assert.Equal(t, int64(3), m.total.Load(), "should still exceed maxItems due to protection")
+}
+
+// Tests that entries can be evicted once they exceed minEvictAge.
+func TestEvictByCapacityAfterMinAge(t *testing.T) {
+	maxItems := int64(2)
+	minEvictAge := 100 * time.Millisecond // Short age for test
+	m := newRegistryImpl(1, time.Hour, minEvictAge, maxItems, time.Hour)
+	defer m.Stop()
+
+	// Add 3 entries (over capacity)
+	for i := 1; i <= 3; i++ {
+		key := fmt.Sprintf("worker%d", i)
+		hb := &workerpb.WorkerHeartbeat{WorkerInstanceKey: key}
+		m.upsertHeartbeats("ns", []*workerpb.WorkerHeartbeat{hb})
+	}
+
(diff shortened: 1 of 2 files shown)


service/matching/workers/registry_test.go
```diff
@@ -195,6 +195,112 @@ func TestRegistryImpl_ListWorkers(t *testing.T) {
...
+		expectedCount   int
+		expectedWorkers []string // WorkerInstanceKeys
+		expectError     bool
+	}{
+		{
+			name: "valid query - basic filtering",
+			setup: func(r *registryImpl) {
+				r.upsertHeartbeats("namespace1", []*workerpb.WorkerHeartbeat{
+					{WorkerInstanceKey: "worker1", TaskQueue: "queue1"},
+					{WorkerInstanceKey: "worker2", TaskQueue: "queue2"},
```
stephanos, line 216: nit: can/should `TaskQueue` be dropped here if it's not relevant to the test?

service/matching/workers/registry_test.go
```diff
@@ -195,6 +195,112 @@ func TestRegistryImpl_ListWorkers(t *testing.T) {
...
+		{
+			name: "invalid query - malformed SQL",
+			setup: func(r *registryImpl) {
+				r.upsertHeartbeats("namespace1", []*workerpb.WorkerHeartbeat{
+					{WorkerInstanceKey: "worker1"},
+				})
+			},
+			nsID:        "namespace1",
+			query:       "invalid SQL syntax here",
+			expectError: true,
```
stephanos: it: would it be better to be able to provide an error substring that the error has to match?

service/matching/workers/registry_impl_test.go
```diff
@@ -112,6 +112,62 @@ func TestEvictByCapacity(t *testing.T) {
...
+		key := fmt.Sprintf("worker%d", i)
+		hb := &workerpb.WorkerHeartbeat{WorkerInstanceKey: key}
+		m.upsertHeartbeats("ns", []*workerpb.WorkerHeartbeat{hb})
+	}
+
+	// Verify we're over capacity
+	assert.Equal(t, int64(3), m.total.Load(), "should have 3 entries initially")
+	assert.Greater(t, m.total.Load(), maxItems, "should be over capacity")
+
+	// Attempt eviction - should break because all entries are too new
```
stephanos, line 136: nit: the "break" terminology feels too close to the implementation details

service/matching/workers/registry_impl_test.go
```diff
@@ -112,6 +112,62 @@ func TestEvictByCapacity(t *testing.T) {
...
+	// Add 3 entries (over capacity)
+	for i := 1; i <= 3; i++ {
+		key := fmt.Sprintf("worker%d", i)
+		hb := &workerpb.WorkerHeartbeat{WorkerInstanceKey: key}
+		m.upsertHeartbeats("ns", []*workerpb.WorkerHeartbeat{hb})
+	}
+
+	// Wait for entries to exceed minEvictAge
+	// TODO: Use mock time for more reliable tests.
+	time.Sleep(300 * time.Millisecond)
```
stephanos: I'd use testify's `eventually` (lots of code samples in the tests) to verify that the assertions eventually pass.

stephanos (approved):
Looks good! The `sleep` is the only part I'd definitely like to see addressed.

service/matching/workers/registry_impl_test.go
rkannan82: I have changed to inject a clock instead. Although we can use eventually here, since this is purely time-dependent logic, using a mock clock will make the test fast and deterministic. Let me know if this is OK.

service/matching/workers/registry_test.go
rkannan82: I added a test with compound query (multiple conditions) now :)

service/matching/workers/registry_impl_test.go
stephanos: It might be better to use this: https://victoriametrics.com/blog/go-synctest/

It (prob) requires no code changes to the production code.

rkannan82: synctest works, but this is an experimental feature. So we need to enable it in the test invocations ...
GOEXPERIMENT=synctest go test ./service/matching/workers -v

I have a revised implementation that is much closer to the original impl, except it uses TimeSource.NewTimer instead of time.NewTicker. The common/cache/cache.go also uses this API. So I suggest that we go with this for now. Let me know if you feel otherwise.

rkannan82: Thanks Stephan for pointing out this is GA in Go version 1.25. Turns out the API has changed, but they kept the old API for now, and only that API requires the goexperiement flag.

service/matching/workers/registry_impl_test.go
```diff
@@ -112,6 +113,65 @@ func TestEvictByCapacity(t *testing.T) {
...
+
+		// Add 3 entries (over capacity)
+		for i := 1; i <= 3; i++ {
+			key := fmt.Sprintf("worker%d", i)
+			hb := &workerpb.WorkerHeartbeat{WorkerInstanceKey: key}
+			m.upsertHeartbeats("ns", []*workerpb.WorkerHeartbeat{hb})
+		}
+
+		// Virtual time advance - instant with synctest!
+		time.Sleep(300 * time.Millisecond)
```
stephanos, line 164: I see the linter is still unhappy here. since it doesn't know about synctest, you'll have to add a comment to override it here: `//nolint:forbidigo`.

PR #8328, opened 2025-09-15
Remove scan API from visibility

rodrigozhou:
## What changed?
Removing scan API from Visibility

## Why?
Visibility Scan API has been marked as deprecated since v1.27.0.
`Scan` in Visibility with SQL DB is the same as the `List` API. With Elasticsearch, it provides a search without an order of the results, and we don't see any reason to use it instead of the `List` API.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
Users will be required to migrate to `List` API prior to upgrading Temporal.

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -2538,40 +2538,24 @@ func (wh *WorkflowHandler) ListArchivedWorkflowExecutions(ctx context.Context, r
 	}, nil
 }
 
-// ScanWorkflowExecutions is a visibility API to list large amount of workflow executions in a specific namespace without order.
+// ScanWorkflowExecutions _was_ a Visibility API to list large amount of workflow executions in a specific namespace without order.
+// It has since been deprecated in favor of `ListWorkflowExecutions` and rewritten to use `ListWorkflowExecutions` internally.
+// Deprecated: Use `ListWorkflowExecutions`
 func (wh *WorkflowHandler) ScanWorkflowExecutions(ctx context.Context, request *workflowservice.ScanWorkflowExecutionsRequest) (_ *workflowservice.ScanWorkflowExecutionsResponse, retError error) {
 	defer log.CapturePanic(wh.logger, &retError)
 
-	if request == nil {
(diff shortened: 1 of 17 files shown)


service/frontend/workflow_handler.go
```diff
@@ -2538,40 +2538,22 @@ func (wh *WorkflowHandler) ListArchivedWorkflowExecutions(ctx context.Context, r
 	}, nil
 }
 
-// ScanWorkflowExecutions is a visibility API to list large amount of workflow executions in a specific namespace without order.
+// Deprecated: Use `ListWorkflowExecutions`
```
bergundy, line 2543:
```suggestion
// ScanWorkflowExecutions _was_ a visibility API to list large amount of workflow executions in a specific namespace without order.
// It has since been deprecated in favor of `ListWorkflowExecutions` and rewritten to use `ListWorkflowExecutions` internally.
// Deprecated: Use `ListWorkflowExecutions`
```

rodrigozhou:
`Scan` calls `List` only when using SQL visibility. With ES, `Scan` has its own implementation that's different from `List`. As it's in our [documentation](https://docs.temporal.io/tctl-v1/workflow#scan), `Scan` differs from `List` that it's not guaranteed to have any order.

It's not a performance issue, it's just that we don't want to keep two APIs that does very similar things. In SQL it's already the same, and ES we don't see any reason to have the `Scan` API.

bergundy:
> `Scan` calls `List` only when using SQL visibility. With ES, `Scan` has its own implementation that's different from `List`. As it's in our [documentation](https://docs.temporal.io/tctl-v1/workflow#scan), `Scan` differs from `List` that it's not guaranteed to have any order.
> 
> It's not a performance issue, it's just that we don't want to keep two APIs that does very similar things. In SQL it's already the same, and ES we don't see any reason to have the `Scan` API.

SGTM, just put this in the PR description and make sure it makes it to the release notes.

bergundy (approved):
Please add my suggestion but otherwise LGTM.

PR #8330, opened 2025-09-16
CHASM: Auto sync visibility SA and memo from root component

yycptt:
## What changed?
- Add Visibility SearchAttribute and Memo provider interface. If a root component implements those interfaces, and also uses a Visibility component, CHASM framework will automatically invoke those two interfaces and sync SA & memo. 

## Why?
- This saves component author from calling UpdateSA/Memo() method whenever some attributes get updated. They can focus on updating the component state and CHASM automatically perform the updates to visibility

## How did you test it?
- [x] built
- [x] added new unit test(s)


chasm/context.go
```diff
@@ -100,3 +102,7 @@ func (c *MutableContextImpl) AddTask(
 ) {
 	c.root.AddTask(component, attributes, payload)
 }
+
+func (c *MutableContextImpl) ToImmutable() Context {
```
bergundy: Why is this needed? `MutableContext` is already a `Context` by embedding it.

chasm/library_core.go
```diff
@@ -0,0 +1,26 @@
...
+		NewRegistrableComponent[*Visibility]("vis"),
+	}
+}
+
+func (b *CoreLibrary) Tasks() []*RegistrableTask {
+	return []*RegistrableTask{
+		NewRegistrableSideEffectTask(
+			"visTask",
+			&visibilityTaskValidator{},
+			&visibilityTaskExecutor{},
```
bergundy: This could be the same struct.
Not for this PR: Maybe it should just be one interface for both Execute and Validate? Is there ever a case for splitting them?

chasm/tree.go
```diff
@@ -249,7 +258,14 @@ func (n *Node) SetRootComponent(
...
 	root := n.root()
 	root.value = rootComponent
-	root.valueState = valueStateNeedSerialize
+	root.setValueState(valueStateNeedSerialize)
+}
+
+func (n *Node) setValueState(state valueState) {
+	n.valueState = state
+	if state == valueStateNeedSerialize {
+		n.isActiveStateDirty = true
```
bergundy, line 325: So is this always true only if `state == valueStateNeedSerialize`? If so, do we need another boolean? Can this be a method instead? I'm not very familiar with the code, but is there an assumption that the state would change again but we want to leave `isActiveStateDirty = true`?

chasm/tree.go
```diff
@@ -1260,31 +1273,96 @@ func (n *Node) closeTransactionHandleRootLifecycleChange() (bool, error) {
...
+	}
+
+	rootComponent, err := n.Component(immutableContext, ComponentRef{})
+	if err != nil {
+		return err
+	}
+
+	saProvider, ok := rootComponent.(VisibilitySearchAttributesProvider)
+	if ok {
+		updated, err := visibility.SetSearchAttributes(mutableContext, saProvider.SearchAttributes(immutableContext))
```
bergundy: I don't think this is the desirable approach, there's no need to duplicate the data in both the root and the visibility components.

bergundy (commented):
I wonder if the visibility component is even needed here. Seems like it should be enough to let the root component implement the provider interfaces. We could get the memo and search attributes from the root component before and after a transaction and do a comparison at the end instead of duplicating the data.

chasm/context.go
yycptt: It's for the framework to know that the context is a read only context so it won't mark the component as dirty. 
It's useful for doing a read only op on a child component inside a write op on the parent component.

chasm/library_core.go
yycptt: yeah they could be the same struct, just personal preference here I guess.

Re. one interface v.s. two interfaces. It's mainly because the executor method for pure task and side effect task are different, but validation method are the same, so I separate the interface for validation and execution. I think it's good to define small interfaces in general, but no really strong opinion here.

chasm/tree.go
yycptt: It's basically tracking if any node is ever accessed (active logic by definition) with a mutable context in a transaction.

So yeah it's only true when a node's state is set to valueStateNeedSerialize, and the flag remains true for the entire transaction.

chasm/tree.go
yycptt: Updated the implementation as we discussed offline. Component SA and memo are no longer double stored.

yycptt:
> if the visibility component is even needed here

Synced offline. Visibility component is for custom search attribute/memo. And the new interfaces are for component SA/Memo.

chasm/context.go
bergundy: Ah I see. So do you expect library authors to ever use this method? In any case document it, and consider not exporting it.

chasm/context.go
bergundy, line 1: Not related to this PR but let's not use `Impl` as a convention. If you look at Go's `context.go` implementation, they use an unexported `emptyCtx`, `backgroundCtx`, etc... we should do the same and use best practices for Go.
Also really missing documentation here.

chasm/context.go
bergundy: Typically if you have to explain something in a review, it's better to put the explanation in the code for posterity.

chasm/visibility.go
```diff
@@ -13,29 +13,20 @@ const (
...
 
-// CoreLibrary contains built-in components maintained as part of the CHASM framework.
-type CoreLibrary struct {
-	UnimplementedLibrary
+// VisibilitySearchAttributesProvider if implemented by the root Component,
+// allows the CHASM framework to automatically determine, at the end of
+// a transaction, if a visibility task needs to be generated to update the
+// visibility record with the returned search attributes.
+type VisibilitySearchAttributesProvider interface {
+	SearchAttributes(Context) map[string]any
```
bergundy: Let's merge with this interface but we may want a type safe version of this later.

chasm/library_core.go
bergundy: I just saw that for scheduler we only have one struct implementing both so I would just merge.

chasm/visibility.go
```diff
@@ -57,10 +48,42 @@ func NewVisibility(
...
+
+	visibility.generateTask(mutableContext)
 	return visibility
 }
 
+func NewVisibilityWithData(
+	mutableContext MutableContext,
+	searchAttributes map[string]any,
+	memo map[string]any,
+) (*Visibility, error) {
```
bergundy, line 60: I'm a bit confused about this constructor.
Is this for user provided search attributes and memo or is it for system ones too?
If it's just an alternative to the Provider interfaces, how about we remove it for now instead of leaving it around and confusing library authors?
We can always bring this back later IMHO.
Also would be okay to not export this method and leave the code around in case we want to use it but would rather not have dead code.

chasm/visibility.go
```diff
@@ -169,13 +223,14 @@ func GetSearchAttribute[T any](
...
-func UpsertSearchAttribute[T ~int | ~int32 | ~int64 | ~string | ~bool | ~float64 | ~[]byte](
-	chasmContext MutableContext,
+func SetSearchAttribute[T ~int | ~int32 | ~int64 | ~string | ~bool | ~float64 | ~[]byte](
+	mutableContext MutableContext,
 	visibility *Visibility,
 	name string,
 	value T,
 ) {
-	upsertVisibilityPayload(chasmContext, &visibility.SA, visibility, name, value)
+	// TODO: only update if value is different and return bool to indicate if updated
```
bergundy: E.g. this would be an optimization?

service/history/visibility_queue_task_executor.go
```diff
@@ -398,6 +398,33 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
...
+	if err != nil {
+		return err
+	}
+	if saProvider, ok := rootComponent.(chasm.VisibilitySearchAttributesProvider); ok {
+		for key, value := range saProvider.SearchAttributes(visTaskContext) {
+			p, err := payload.Encode(value)
+			if err != nil {
+				// This should never happen as we validated the value can be encoded when
+				// closing the transaction. See chasm/tree.go closeTransactionForceUpdateVisibility().
+				return err
```
bergundy: Provide context with the error.

```suggestion
				// This should never happen as we validated the value can be encoded when
				// closing the transaction. See chasm/tree.go closeTransactionForceUpdateVisibility().
				return fmt.Errorf("unexpected error when encoding search attribute: %w", err)
```

service/history/visibility_queue_task_executor.go
```diff
@@ -398,6 +398,33 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
...
+			searchattributes[key] = p
+		}
+	}
+	if memoProvider, ok := rootComponent.(chasm.VisibilityMemoProvider); ok {
+		for key, value := range memoProvider.Memo(visTaskContext) {
+			p, err := payload.Encode(value)
+			if err != nil {
+				// This should never happen as we validated the value can be encoded when
+				// closing the transaction. See chasm/tree.go closeTransactionForceUpdateVisibility().
+				return err
```
bergundy: Same here.

chasm/tree.go
bergundy: So why do we need another boolean? Will state ever change from `valueStateNeedSerialize` (I think it should only transition to `valueStateSynced` and in both cases you will necessarily have a dirty active state)?

chasm/tree.go
```diff
@@ -1262,30 +1307,91 @@ func (n *Node) closeTransactionHandleRootLifecycleChange() (bool, error) {
...
-		if !ok {
-			return serviceerror.NewInternalf("expected visibility component for component with type %s, but got %T", visibilityComponentFqType, visComponent)
+	}
+	if visibilityNode == nil {
+		return nil
+	}
+
+	needUpdate := rootLifecycleChanged
+
+	mutableContext := NewMutableContext(context.Background(), n)
```
bergundy: Is there a good reason to construct a `context.Context` object every time here? You can just pass in nil or a static context object.
Looks like the only reason we hold a context object is to pass in an intent.
I also don't see a reason for a library author to ever create context objects directly but can't come up with a way to hide this method since the non-chasm part of the codebase needs it.
I'll need to think about it some more, it's late.

service/history/visibility_queue_task_executor.go
bergundy: Also probably best to make sure this task isn't retried by returning a specific error.

chasm/tree.go
```diff
@@ -1262,30 +1307,91 @@ func (n *Node) closeTransactionHandleRootLifecycleChange() (bool, error) {
...
+
+	rootComponent, err := n.Component(immutableContext, ComponentRef{})
+	if err != nil {
+		return err
+	}
+
+	saProvider, ok := rootComponent.(VisibilitySearchAttributesProvider)
+	if ok {
+		newSearchAttributes := saProvider.SearchAttributes(immutableContext)
+		if !maps.Equal(n.currentSearchAttributes, newSearchAttributes) {
```
bergundy: This doesn't work with keyword lists:

```
func TestFoo(t *testing.T) {
	m1 := map[string]any{"a": []string{"a", "b", "c"}}
	m2 := map[string]any{"a": []string{"a", "b", "c"}}
	require.True(t, maps.Equal(m1, m2))
}


// panic: runtime error: comparing uncomparable type []string [recovered, repanicked
```

chasm/tree.go
```diff
@@ -1262,30 +1307,91 @@ func (n *Node) closeTransactionHandleRootLifecycleChange() (bool, error) {
...
+			for k, v := range newSearchAttributes {
+				if _, err := payload.Encode(v); err != nil {
```
bergundy: If you take the typed search attributes from the Go SDK and use those for this interface, you will exchange this expensive runtime check with a compile time check and increase safety for all library authors.
The downside is that it will be harder to use this for custom search attributes, maybe we need a different experience for that. I was thinking that the visibility component would be responsible for that.

https://pkg.go.dev/go.temporal.io/sdk@v1.36.0/temporal#SearchAttributes

chasm/tree.go
```diff
@@ -1262,30 +1307,91 @@ func (n *Node) closeTransactionHandleRootLifecycleChange() (bool, error) {
...
+
+			needUpdate = true
+		}
+		n.currentSearchAttributes = newSearchAttributes
+	}
+
+	memoProvider, ok := rootComponent.(VisibilityMemoProvider)
+	if ok {
+		newMemo := memoProvider.Memo(immutableContext)
+		if !maps.Equal(n.currentMemo, newMemo) {
```
bergundy: Same here, you're going to need comparable types as values or this will panic.
I'm too tired to come up with good suggestions.

chasm/context.go
yycptt: yeah I do expect component authors to use it. Will add comment on Context & MutableContext interfaces and this method.

chasm/library_core.go
yycptt: ~~I am still thinking this as mostly a personal preference of the implementation and not introduced in this PR. fwiw, this impl is here before scheduler code is checked in.~~

~~But yeah if there's a compelling reason or strong opinion on merging them I can do it.~~

nvm, I merged

chasm/tree.go
yycptt: > why do we need another boolean

valueState is on a single node and this flag is for the entire tree to avoid scanning the tree every time.

> state ever change from valueStateNeedSerialize

It will no longer change in a transaction. When closing transaction, after value is serialized, it will be changed to synced.

chasm/context.go
yycptt: yeah I agree and will change the name. 

> they use an unexported emptyCtx, backgroundCtx

yeah they use a unexported struct but then they have to return the Context interface from constructors like context.Background(), which is also against best practice of taking interface & return struct.

but guess I will just follow their approach here and return an interface.

chasm/visibility.go
yycptt: sorry merge which two interfaces? 

SA and Memo provider? But there's no guarantee that a component will always use both right.

chasm/visibility.go
yycptt: Visibility component is only for custom search attributes, i can name the parameters better. 

It's mostly for making component author's life a bit easier especially when creating a new entity which usually already have user provided customer SA and memo available. so author can just invoke this one method to init the visibility component, instead of first creating an empty one and then call another method to add the data.

chasm/tree.go
yycptt: The context is associated with current chasm tree and when it's used later to generate task, the task will be added to a newTask field for this tree. so it can't be nil or a static object.

> library author to ever create context objects directly

Correct library author never creates Context objects, they just use framework constructed ones. 
Here's this is framework logic constructing a Context and use it to invoke component method. This is not user/component code.

chasm/tree.go
yycptt: ah thanks for catching! Missed keyword list case, will find a way to fix.

chasm/tree.go
yycptt: yeah agree. let me study ^. 

~~I am thinking of defining a type constraint for the returned value, and we can guarantee that those values can be encoded.~~

~~type VisibilityValue interface {~~
~~int | int32 | int64 | float64 | bool | string | []string | time.Time~~
~~}~~

nvm can't really use type constraints as map value.

Update: Resolving the comment as it no longer applies with the new approach.

chasm/context.go
yycptt: If it's ok, I will do this renaming and adding comments to Context/MutableContext in separate PR, want to keep this PR focused on visibility work.

chasm/visibility.go
yycptt: Code removed.

service/history/visibility_queue_task_executor.go
yycptt: Resolving the comment as it no longer applies with the new approach.

chasm/tree.go
yycptt: Resolving the comment as it no longer applies with the new approach.

chasm/tree.go
yycptt: Resolving the comment as it no longer applies with the new approach.

chasm/context.go
bergundy: Hmm... you probably have use cases in mind that I don't. Waiting for the comments.

chasm/tree.go
bergundy: SGTM, please consider moving any information from this PR thread to code comments.

chasm/context.go
bergundy: SGTM.

chasm/visibility.go
bergundy: I meant merge the PR.

chasm/visibility.go
bergundy: Okay, I get it, would be nice to document this and a `custom` or something to the parameter names for clarification.

chasm/tree.go
bergundy: Still not sure why we need a `context.Context` object in the CHASM context, you shouldn't be able to invoke methods that require `context.Context` where you hold a lock in application code.

chasm/visibility.go
```diff
@@ -5,37 +5,27 @@ import (
...
 
-// CoreLibrary contains built-in components maintained as part of the CHASM framework.
-type CoreLibrary struct {
-	UnimplementedLibrary
+// VisibilitySearchAttributesProvider if implemented by the root Component,
+// allows the CHASM framework to automatically determine, at the end of
+// a transaction, if a visibility task needs to be generated to update the
+// visibility record with the returned search attributes.
+type VisibilitySearchAttributesProvider interface {
+	SearchAttributes(Context) map[string]VisibilityValue
```
bergundy, line 23: Mentioned this on slack too, but I'm not sure why we can't use the typed search attribute definitions from the SDK.

chasm/visibility.go
```diff
@@ -57,10 +45,42 @@ func NewVisibility(
...
-	visibility.GenerateTask(mutableContext)
+
+	visibility.generateTask(mutableContext)
 	return visibility
 }
 
+func NewVisibilityWithData(
+	mutableContext MutableContext,
+	searchAttributes map[string]*commonpb.Payload,
+	memo map[string]*commonpb.Payload,
```
bergundy: Maybe call this `userSearchAttributes` or `customSearchAttributes` and `userMemo`?

chasm/visibility.go
```diff
@@ -45,8 +35,6 @@ type Visibility struct {
 
 	SA   Map[string, *commonpb.Payload]
 	Memo Map[string, *commonpb.Payload]
```
bergundy: Seems like overkill to use `chasm.Map` for search attributes and arguably for memo too. The total memo size shouldn't exceed the size of a payload AFAIR.

chasm/tree.go
yycptt: yeah application code has no way to get and use that context.  
It's more like a placeholder for later partial read work where the framework potentially needs to go to persistence and load some component states. Let me actually change it to context.TODO() here and add comments.

chasm/visibility.go
yycptt: ah, ok :)

chasm/visibility.go
yycptt: Synced offline, looks like the overall size will always be small, so partial write doesn't worth it. Will switch to just one payload.

chasm/context.go
yycptt: A use case is the current implementation of SetSA/Memo method of the visibility component. It needs to first read the payload for a given key, do a comparison, and then decide if the payload needs to be overwritten or not. If a mutable context is used through out the method, then all payloads will be marked as dirty. 

But as we discussed that using a Map there will be an overkill, I actually no longer need this new method. Will remove and add it back later when there's a use case.

chasm/tree.go
```diff
@@ -1262,30 +1306,82 @@ func (n *Node) closeTransactionHandleRootLifecycleChange() (bool, error) {
...
+		}
+		n.currentMemo = newMemo
+	}
+
+	if !needUpdate {
+		return nil
+	}
+
+	var visibilityNode *Node
 	for _, child := range n.children {
```
lina-temporal, line 1435: (minor) this might be useful extracted as a  `getNodesByType(fqType string)` helper.

chasm/tree.go
```diff
@@ -1740,7 +1838,35 @@ func (n *Node) ApplyMutation(
...
+		return err
+	}
+
+	// For replication case, we only update the search attributes and memo
+	// but not force updating the visibility component itself to generate a task.
+	//
+	// This is because the visibility component is already force updated in the active
+	// cluster and that forced update will be replicated as well. Standby cluster
+	// only needs to track the current SA and memo to prevent generating an unnecessary
+	// visibility component update & task if there is a failover.
```
lina-temporal, line 1957: Just for my understanding of how this works with task validation/replication, I get that we're relying on comparing the VT to guarantee up-to-dateness. But what happens if the incoming mutations we apply end up invalidating standby's `VisibilityTask` (VersionedTransition advances), and active-side never ends up creating another task before we fail over? I suppose we'd just create one as soon as the newly-promoted active cluster closed a transaction and noticed `currentSA`/`currentMemo` were unsynced with ES, either way.

bergundy (approved):
We went over all of the comments. Feel free to merge when they are addressed or submit followup PRs.

chasm/visibility.go
```diff
@@ -71,18 +91,23 @@ func (v *Visibility) GetSearchAttributes(
 	return v.getPayloadMap(chasmContext, v.SA)
 }
 
-func (v *Visibility) UpsertSearchAttributes(
+func (v *Visibility) SetSearchAttributes(
```
lina-temporal, line 93: Should we do payload size limit checking within the Visibility component, or within CHASM components? I think the pertinent part is, if we care about individual components having to manage their own payload size limits [which may be lower than the system total for whatever reason], or if CHASM should just have a blanket size limit all components can rely on.

chasm/tree.go
yycptt: > what happens if the incoming mutations we apply end up invalidating standby's VisibilityTask (VersionedTransition advances)

That means a new (logical) Visibility task is generated and will be replicated as well. The VT and logical task are updated in the same transaction so they will also be replicated together. Not sure if that answers your question.

chasm/visibility.go
yycptt: Good point! I read the code and looks like today we are doing those validations on frontend to prevent unnecessary forwarding of big payload. I guess we want to continue doing that and that will be outside CHASM framework.

Let's actually chat about this in next sync meeting.

chasm/tree.go
yycptt: introduced node.IsComponent() in another PR, going to use that here.

chasm/visibility.go
yycptt: Added `custom` to those variable names.

chasm/visibility.go
yycptt: Synced offline. Will keep this interface for now and for memo. Once SA registration is supported, this interface will be updated.

PR #8331, opened 2025-09-16
Add timeout buffering when dispatching nexus task.

fretz12:
## What changed?
- Added timeout buffering when dispatching nexus task. Exposed as a dynamic config.
- Added new timeout status to dispatch nexus response.
- Renamed nexus operation timeout config to request timeout

## Why?
As the nexus start/cancel operations span through the executor to handler to matching, we should buffer the timeout so the spanned calls timeout in the right order. Buffering takes place in matching, and we explicitly indicate in the dispatch response if dispatch times out. Exposing this buffer value as a dynamic config gives users the choice to tweak it if needed.  Nexus operation timeout config was renamed to correct typos and more closely match what it represents.

## How did you test it?
- [X] built
- [X] added new unit test(s)


service/frontend/nexus_handler.go
```diff
@@ -423,13 +423,6 @@ func (h *nexusHandler) StartOperation(
 	// RPC.
 	response, err := h.matchingClient.DispatchNexusTask(ctx, request)
 	if err != nil {
-		if common.IsContextDeadlineExceededErr(err) {
-			oc.metricsHandler = oc.metricsHandler.WithTags(metrics.OutcomeTag("handler_timeout"))
```
bergundy, line 436: The default outcome would be internal_error which is technically correct but I wonder if we would be better off knowing what the internal error was.

Let's not solve it in this PR though, it's note just about internal timeouts.

service/matching/matching_engine.go
```diff
@@ -2221,12 +2221,37 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
...
-	// this remote host's result can be returned directly
-	if resp != nil || err != nil {
-		return resp, err
+	// Buffer the deadline so we can still respond with timeout if we hit the deadline while dispatching
+	var cancel context.CancelFunc
+	if deadline, ok := ctx.Deadline(); ok {
+		timeout := time.Until(deadline)
+		ctx, cancel = contextutil.WithDeadlineBuffer(ctx, timeout, time.Second)
+		defer cancel()
+	}
```
bergundy: No need to get the original context's deadline, `WithDeadlineBuffer` already does this for you.
You can replace this with:

```suggestion
	ctx, cancel = contextutil.WithDeadlineBuffer(ctx, matching.DefaultTimeout, time.Second)
	defer cancel()
```

service/matching/matching_engine.go
```diff
@@ -2221,12 +2221,37 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
...
+
+	select {
+	case r := <-dispatchCh:
+		// if we get a response or error it means that the Nexus task was handled by forwarding to another matching host
+		// this remote host's result can be returned directly
+		if r.resp != nil || r.err != nil {
+			return r.resp, err
+		}
+	case <-ctx.Done():
+		return &matchingservice.DispatchNexusTaskResponse{Outcome: &matchingservice.DispatchNexusTaskResponse_RequestTimeout{}}, nil
```
bergundy: You will already be getting a context error if `pm.DispatchNexusTask` times out with the given context deadline.

```suggestion

	resp, err := pm.DispatchNexusTask(ctx, taskID, request)
	
	
	if ctx.Err() != nil {
		// The context deadline has expired with the given buffer, return an explicit timeout to the caller to indicate this is an issue with the worker and not the server.
		return &matchingservice.DispatchNexusTaskResponse{Outcome: &matchingservice.DispatchNexusTaskResponse_RequestTimeout{}}, nil
	}
	// if we get a response or error it means that the Nexus task was handled by forwarding to another matching host
	// this remote host's result can be returned directly
	if resp != nil || err != nil {
		return resp, err	
```

service/matching/matching_engine.go
```diff
@@ -2250,7 +2275,7 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
 			Response: result.successfulWorkerResponse.GetRequest().GetResponse(),
 		}}, nil
 	case <-ctx.Done():
-		return nil, ctx.Err()
+		return &matchingservice.DispatchNexusTaskResponse{Outcome: &matchingservice.DispatchNexusTaskResponse_RequestTimeout{}}, nil
```
bergundy: Nice, but it would be worth adding a comment why this conversion is done.

service/matching/matching_engine.go
fretz12: Cool, `matching.DefaultTimeout` was what I needed. Addressed.

service/matching/matching_engine.go
fretz12: Ah, I though pm.DispatchNexusTask would be blocking potentially longer, but looking through the stack I see it manages ctx timeout. Fixed.

service/matching/matching_engine.go
fretz12: Added.

service/matching/matching_engine.go
```diff
@@ -2221,8 +2222,18 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
 	}
 
 	taskID := uuid.New()
+
+	// Buffer the deadline so we can still respond with timeout if we hit the deadline while dispatching
+	ctx, cancel := contextutil.WithDeadlineBuffer(ctx, matching.DefaultTimeout, time.Second)
```
MonkeyCode-AI, line 2227: 在 `service/matching/matching_engine.go` 中，建议添加注释来解释为什么选择 1 秒作为缓冲时间，以及这个值是如何确定的。这有助于其他开发者理解这个实现决策的背景。
```suggestion
// Buffer the deadline so we can still respond with timeout if we hit the deadline while dispatching.
// Using a 1-second buffer to ensure we have enough time to return a proper timeout response.
ctx, cancel := contextutil.WithDeadlineBuffer(ctx, matching.DefaultTimeout, time.Second)
```

service/frontend/nexus_handler.go
```diff
@@ -442,6 +435,13 @@ func (h *nexusHandler) StartOperation(
 		err := h.convertOutcomeToNexusHandlerError(t)
 		return nil, err
 
+	case *matchingservice.DispatchNexusTaskResponse_RequestTimeout:
```
MonkeyCode-AI, line 438: 在 `service/frontend/nexus_handler.go` 中，建议统一处理错误的逻辑。当前的错误处理在 StartOperation 和 CancelOperation 中有重复的代码。可以考虑将公共部分提取到一个辅助函数中以减少重复。

MonkeyCode-AI (commented):
总体来说，这次 PR 的实现很清晰，很好地解决了调度 Nexus 任务时的超时问题。代码结构良好，测试也做了相应调整。

有几点小建议：
1. 在匹配引擎中添加了缓冲时间，但没有注释说明为什么选择 1 秒作为缓冲时间。建议添加注释说明这个值的选择原因。
2. 前端处理程序中有一些重复的错误处理代码，可以考虑提取到一个公共函数中。

除此之外，PR 的实现符合预期，测试也通过了。

fretz12:
> Looks great, can you also confirm that the timing works as expected. This is going to be difficult to achieve in a functional test without introducing flakiness so just manual verification should suffice here. There are ways to test parts of this via unit tests, we can discuss those offline.
> 
> 1. The request-timeout header reflects the original request's timeout minus the one second buffer.
> 2. The request times out roughly a second earlier than it did before this change.

Unit tests added, including that using synctest

common/nexus/failure.go
```diff
@@ -22,6 +23,8 @@ const (
 	FailureSourceHeaderName = "Temporal-Nexus-Failure-Source"
 	// FailureSourceWorker indicates the failure originated from outside the server (e.g. bad request or on the Nexus worker).
 	FailureSourceWorker = "worker"
+	// NexusTimeoutBuffer is the amount of time to leave as a buffer when dispatching a Nexus task
+	NexusTimeoutBuffer = time.Second
```
bergundy: I wouldn't put this in `failure.go`, you can keep this in the matching code IMHO.

service/frontend/nexus_handler.go
```diff
@@ -432,13 +432,6 @@ func (h *nexusHandler) StartOperation(
 	// RPC.
 	response, err := h.matchingClient.DispatchNexusTask(ctx, request)
```
bergundy, line 435: I realized that we should look at the current context timeout and avoid dispatching the request if we can't give enough buffer (let's say <= 1.5 seconds, a second of buffer and 500 millis for the handler to process).
It'd be the caller's fault so let's fail with a retryable bad request error to allow the caller to adjust its configuration and avoid failing the operation completely.

service/matching/matching_engine_test.go
```diff
@@ -3547,6 +3550,93 @@ func (s *matchingEngineSuite) TestPollWorkflowTaskQueueWithRateLimiterError() {
...
+		sleepTime time.Duration
+		assertion func(s *matchingEngineSuite, response *matchingservice.DispatchNexusTaskResponse, err error)
+	}
+
+	testCases := []testCase{
+		{
+			name:      "deadline_exceeded",
+			sleepTime: ctxTimeout - cnexus.NexusTimeoutBuffer,
+			assertion: func(s *matchingEngineSuite, response *matchingservice.DispatchNexusTaskResponse, err error) {
+				s.NoError(err)
```
bergundy: You don't want to use `s.T()` for assertions, you need to use the `T` instance from the subtest to properly attribute the failure to the subtest.
This is a common pitfall when using the testify suites.

service/matching/matching_engine_test.go
```diff
@@ -3547,6 +3550,93 @@ func (s *matchingEngineSuite) TestPollWorkflowTaskQueueWithRateLimiterError() {
...
+				s.NotNil(response)
+				s.NotNil(response.Outcome)
+
+				_, ok := response.Outcome.(*matchingservice.DispatchNexusTaskResponse_RequestTimeout)
+				s.True(ok)
```
bergundy: Use `require.NotNil(t, response.GetRequestTimeout()` it will do all of the nil checks and casts for you.
(The `t` is related to the comment above).

service/matching/matching_engine_test.go
```diff
@@ -3547,6 +3550,93 @@ func (s *matchingEngineSuite) TestPollWorkflowTaskQueueWithRateLimiterError() {
...
+				s.NotNil(response.Outcome)
+
+				_, ok := response.Outcome.(*matchingservice.DispatchNexusTaskResponse_Response)
+				s.True(ok)
+			},
+		},
+	}
+
+	testFn := func(s *matchingEngineSuite, tc testCase) {
+		synctest.Test(s.T(), func(t *testing.T) {
```
bergundy: Nice to see `synctest` in action!

tests/nexus_workflow_test.go
```diff
@@ -2249,7 +2249,7 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
 	svc := nexus.NewService("test")
 
 	handlerWF := func(ctx workflow.Context, _ nexus.NoValue) (nexus.NoValue, error) {
-		return nil, workflow.Sleep(ctx, 1*time.Second)
+		return nil, workflow.Sleep(ctx, 2*time.Second)
```
bergundy: Why is this required? The workflow isn't part of the `StartNexusOperation` request duration, it is run asynchronously and completed via a callback.

tests/nexus_workflow_test.go
```diff
@@ -2271,7 +2271,7 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
 
 	run, err := s.SdkClient().ExecuteWorkflow(ctx, client.StartWorkflowOptions{
 		TaskQueue:          taskQueue,
-		WorkflowRunTimeout: 200 * time.Millisecond,
+		WorkflowRunTimeout: 1200 * time.Millisecond,
```
bergundy: Not sure I understand why this change is required.

common/nexus/failure.go
fretz12: refactored to matching_engine

service/matching/matching_engine_test.go
fretz12: Good call, fixed

service/matching/matching_engine_test.go
fretz12: done

service/matching/matching_engine_test.go
fretz12: Indeed!

tests/nexus_workflow_test.go
fretz12: hmm.. isn't this designed to timeout the WF? Since I bumped up the WorkflowRunTimeout (see below), this had to follow too. Otherwise `s.Error(wfErr)` line 2279 wouldn't be satisifed.

tests/nexus_workflow_test.go
fretz12: This was defining the nexus_handler StartOperation timeout and was insufficient now given the new buffering

service/frontend/nexus_handler.go
fretz12: Added this check. I don't have a good way to test this, but open to ideas if you have any.

service/frontend/nexus_handler.go
```diff
@@ -427,18 +432,19 @@ func (h *nexusHandler) StartOperation(
 		return nil, nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "input exceeds size limit")
 	}
 
+	if deadline, hasDeadline := ctx.Deadline(); hasDeadline && time.Until(deadline) < nexusDispatchPreflightTimeout {
```
bergundy: Awesome, there are some SDK tests that might fail because of this. We would need to modify those or make this configurable. Not blocking the PR as this is not technically meant for users to invoke directly.

service/matching/matching_engine_test.go
```diff
@@ -3547,6 +3549,91 @@ func (s *matchingEngineSuite) TestPollWorkflowTaskQueueWithRateLimiterError() {
...
+	}
+
+	testCases := []testCase{
+		{
+			name:      "deadline_exceeded",
+			sleepTime: ctxTimeout - NexusTimeoutBuffer,
+			assertion: func(t *testing.T, response *matchingservice.DispatchNexusTaskResponse, err error) {
+				require.NoError(t, err)
+				require.NotNil(t, response)
+				require.NotNil(t, response.Outcome)
```
bergundy: As I mentioned before, this is redundant. Applies below too. See the implementation of `GetRequestTimeout()`.

service/matching/matching_engine.go
```diff
@@ -2221,8 +2224,20 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
...
 
 	taskID := uuid.New()
+
+	// Buffer the deadline so we can still respond with timeout if we hit the deadline while dispatching
+	ctx, cancel := contextutil.WithDeadlineBuffer(ctx, matching.DefaultTimeout, NexusTimeoutBuffer)
+	defer cancel()
+
 	resp, err := pm.DispatchNexusTask(ctx, taskID, request)
 
+	if ctx.Err() != nil {
```
bergundy: Hmm... now that I think of it, this is before we even give a poll to be matched with a task. This should be a context deadline error that will be translated to an internal error.

bergundy: Explain in a comment that this request comes in with a buffer of 500 millis and should be enough to handle a request to dispatch internally to the partition manager.

service/matching/matching_engine_test.go
```diff
@@ -3547,6 +3549,91 @@ func (s *matchingEngineSuite) TestPollWorkflowTaskQueueWithRateLimiterError() {
...
+	const ctxTimeout = 2 * time.Second
+	type testCase struct {
+		name      string
+		sleepTime time.Duration
+		assertion func(t *testing.T, response *matchingservice.DispatchNexusTaskResponse, err error)
+	}
+
+	testCases := []testCase{
+		{
+			name:      "deadline_exceeded",
```
bergundy: Add a third case where the dispatch to the partition manager succeeds but waiting on the channel time out please.

tests/nexus_workflow_test.go
bergundy: This timer isn't part of the request flow and is not subject to the changes in this PR.

tests/nexus_workflow_test.go
```diff
@@ -2271,7 +2271,7 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
 
 	run, err := s.SdkClient().ExecuteWorkflow(ctx, client.StartWorkflowOptions{
 		TaskQueue:          taskQueue,
-		WorkflowRunTimeout: 200 * time.Millisecond,
+		WorkflowRunTimeout: 1700 * time.Millisecond,
```
bergundy: Hmm... is this because we truncate the request timeout to the workflow run timeout duration?
This test seems flaky to me, if the CI server is busy, we may not be able to deliver the Nexus StartOperation call.

Let's change this test to:
- The handler workflow waits on a signal for completion
- The caller workflow doesn't use a `WorkflowRunTimeout` at all
- The test waits for the handler workflow to be started and then terminates the caller workflow and signals the handler workflow to unblock

service/matching/matching_engine.go
fretz12: Added the comment.

service/matching/matching_engine_test.go
fretz12: Fixed

service/matching/matching_engine_test.go
fretz12: Added

tests/nexus_workflow_test.go
fretz12: Changed accordingly (I think)

tests/nexus_workflow_test.go
fretz12: Changed test behavior as specified below.

service/frontend/nexus_handler.go
fretz12: TBD on discussion w/ team

fretz12: @bergundy - when you get a chance, guide me what needs to be done based on our standup discussion

bergundy: Instead of enforcing the extra buffer on the handler side (on top of the 1 second buffer that we give for the worker), we need to avoid making calls from the executor if there's not enough time to do so.
There's a config for that already:

https://github.com/temporalio/temporal/blob/b9ae7789a8bf37d1a5c253296658b8a369b592fb/components/nexusoperations/config.go#L19 (note that there's a typo in the dynamic config name, let's change the default to be 1.5 seconds).

fretz12: As discussed, renamed operation timeout to request timeout, and exposed the dispatch task timeout as a dynamic config.

service/matching/matching_engine.go
bergundy: The error should be bubbled out as is, we didn't have a chance to dispatch to the worker. There's no need to translate the error at all, it's just an internal error.

bergundy (commented):
One small comment, but we're almost there.

service/matching/matching_engine.go
fretz12: Ok, I understand what you're saying in the first comment now. Fixed accordingly.

service/matching/matching_engine.go
```diff
@@ -2221,8 +2222,24 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
...
+	ctx, cancel := contextutil.WithDeadlineBuffer(ctx, matching.DefaultTimeout, e.config.MinDispatchTaskTimeout(ns.Name().String()))
+	defer cancel()
+
 	resp, err := pm.DispatchNexusTask(ctx, taskID, request)
 
+	// If we get a context error, that means we didn't manage to dispatch the task to a remote host
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
```
bergundy: You can just return the error here, not need for this special case.

```suggestion
```

tests/nexus_workflow_test.go
```diff
@@ -2248,8 +2249,14 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
...
 	svc := nexus.NewService("test")
 
+	completeHandlerCh := make(chan struct{})
+	wgHandlerStarted := sync.WaitGroup{}
+	wgHandlerStarted.Add(1)
+
 	handlerWF := func(ctx workflow.Context, _ nexus.NoValue) (nexus.NoValue, error) {
-		return nil, workflow.Sleep(ctx, 1*time.Second)
+		wgHandlerStarted.Done()
+		<-completeHandlerCh
```
bergundy: You should wait on a Temporal signal, not a channel here. These constructs block the workflow task and are not safe to use in workflows.

tests/nexus_workflow_test.go
```diff
@@ -2270,13 +2277,16 @@ func (s *NexusWorkflowTestSuite) TestNexusCallbackAfterCallerComplete() {
...
 	run, err := s.SdkClient().ExecuteWorkflow(ctx, client.StartWorkflowOptions{
-		TaskQueue:          taskQueue,
-		WorkflowRunTimeout: 200 * time.Millisecond,
+		TaskQueue: taskQueue,
 	}, callerWF)
 	s.NoError(err)
 
-	wfErr := run.Get(ctx, nil)
-	s.Error(wfErr)
+	wgHandlerStarted.Wait()
```
bergundy: It would be best to avoid blocking the test here and timing out the suite.
You can wait on the wait group with a timeout or do poll on the handler workflow ID to be present (DescribeWorkflow not returning an error).

See https://github.com/temporalio/temporal/blob/17f22dd4ad0c5c424bec708de6d4cb4521526f42/common/util.go#L131

bergundy (commented):
Had a couple more small comments, should be easy to fix.

tests/nexus_workflow_test.go
fretz12: Cleaned up this test per our session

tests/nexus_workflow_test.go
fretz12: Changed to wait on signal

PR #8334, opened 2025-09-16
[Scheduled Actions] Record initial total idle time on the idle task itself

lina-temporal:
## What changed?
- Recorded the idle expiration time into the SchedulerIdleTask.

## Why?
- The CHASM scheduler's idle expiration time is controlled by dynamic config. If this value changes after the idle task has been pushed, it would have inadvertedly invalidated itself. Now, DC changes are handled gracefully.

## How did you test it?
- [ ] built
- [x] added new unit test(s)


chasm/lib/scheduler/scheduler_idle_tasks_test.go
```diff
@@ -0,0 +1,156 @@
...
+	schedulerSuite
+	executor *scheduler.SchedulerIdleTaskExecutor
+}
+
+func TestIdleTasksSuite(t *testing.T) {
+	suite.Run(t, &idleTasksSuite{})
+}
+
+func (s *idleTasksSuite) SetupTest() {
+	s.SetupSuite()
```
bergundy: Why do we need to set up the suite on every test? If this is supposed to be used for test setup, the method should be called `SetupTest`.

chasm/lib/scheduler/scheduler_idle_tasks_test.go
```diff
@@ -0,0 +1,156 @@
...
+	// Note: executor will be created per test case with custom config
```
bergundy: So why is it a member of the class? I don't see this field is ever accessed.

chasm/lib/scheduler/scheduler_idle_tasks_test.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type idleValidateTestCase struct {
+	ConfigIdleTime           time.Duration
+	TaskIdleTimeTotal        time.Duration
+	ScheduledTime            time.Time
+	SchedulerClosed          bool
+	IdleMatchesScheduledTime bool
+	SetupScheduler           func(*scheduler.Scheduler, chasm.Context)
+	ExpectedValid            bool
```
bergundy: nit: don't expose fields that don't need to be exposed.

chasm/lib/scheduler/generator_tasks.go
```diff
@@ -102,7 +103,9 @@ func (g *GeneratorTaskExecutor) Execute(
 		// Once the idle timer expires, we close the component.
 		ctx.AddTask(scheduler, chasm.TaskAttributes{
 			ScheduledTime: idleExpiration,
-		}, &schedulerpb.SchedulerIdleTask{})
+		}, &schedulerpb.SchedulerIdleTask{
```
bergundy, line 106: Not to say that we have a problem today, but I would love to confirm this.
Can you please confirm what happens a scheduler is created with a schedule that doesn't have _any_ future runs? Will it just hang there and never close? Is the schedule rejected? What if that happens on update?

I typically like to make idle hooks like this more "declarative" because it's too easy to forget to trigger them.
For example this would be a great candidate for a hook that is triggered before transaction close.

chasm/lib/scheduler/scheduler_idle_tasks_test.go
```diff
@@ -0,0 +1,156 @@
...
+		ScheduledTime:     now,
+		SetupScheduler: func(sched *scheduler.Scheduler, ctx chasm.Context) {
+			// Make scheduler not idle by setting it as paused
+			sched.Schedule.State.Paused = true
+		},
+		ExpectedValid: false,
+	})
+}
+
+func (s *idleTasksSuite) TestValidate_IdleExpirationChanged() {
```
bergundy: Seems redundant to test this now that we know for sure that we don't look at the config value but 🤷

chasm/lib/scheduler/scheduler_tasks.go
```diff
@@ -35,15 +35,21 @@ func (r *SchedulerIdleTaskExecutor) Execute(
...
 	scheduler *Scheduler,
-	task chasm.TaskAttributes,
-	_ *schedulerpb.SchedulerIdleTask,
+	taskAttrs chasm.TaskAttributes,
+	task *schedulerpb.SchedulerIdleTask,
 ) (bool, error) {
 	idleTimeTotal := r.Config.Tweakables(scheduler.Namespace).IdleTime
-	idleExpiration, isIdle := scheduler.getIdleExpiration(ctx, idleTimeTotal, time.Time{})
 
+	// Validate total idle time. If the time has been increased, we'll use the longer
```
bergundy: But this doesn't reschedule a new timer... So you just have to accept the time that was recorded when the task was generated (what I would do because it is simple) or you have to deal with this and when you execute the task, you reschedule a new task with the updated duration.

chasm/lib/scheduler/scheduler_idle_tasks_test.go
lina-temporal: Earlier it'd been in use but no longer, I'll clean it up.

chasm/lib/scheduler/scheduler_tasks.go
lina-temporal: Yeah, I guess there's no good way to bump a physical task along once it's created. I'll just simplify this and have it always go off of the `IdleTimeTotal` on the task itself.

chasm/lib/scheduler/generator_tasks.go
lina-temporal: > Not to say that we have a problem today, but I would love to confirm this. Can you please confirm what happens a scheduler is created with a schedule that doesn't have any future runs?

Sure, yeah. This isn't validated up-front. Instead, when the Scheduler's GeneratorTask runs, it'll detect that the schedule's entirely out of actions (`getNextTime` in `ProcessTimeRange` stops returning results, specifically), and add the idle task to close out the scheduler. During this time, an update could come in and set a new spec. If the new spec has actions, then the idle task invalidates itself, and the schedule comes to life. If the new spec is also 'empty', then we'll add another idle task (the customer reset their own idle timer). 

>  For example this would be a great candidate for a hook that is triggered before transaction close.

Sure, it's just not very straightforward since the close conditions are deep within the Generator's task, essentially.

chasm/lib/scheduler/scheduler_idle_tasks_test.go
lina-temporal: You're right, fixed here, elsewhere.

chasm/lib/scheduler/scheduler_idle_tasks_test.go
lina-temporal: fixed.

chasm/lib/scheduler/scheduler_tasks.go
```diff
@@ -35,15 +35,15 @@ func (r *SchedulerIdleTaskExecutor) Execute(
...
 	ctx chasm.Context,
 	scheduler *Scheduler,
-	task chasm.TaskAttributes,
-	_ *schedulerpb.SchedulerIdleTask,
+	taskAttrs chasm.TaskAttributes,
+	task *schedulerpb.SchedulerIdleTask,
 ) (bool, error) {
-	idleTimeTotal := r.Config.Tweakables(scheduler.Namespace).IdleTime
+	idleTimeTotal := task.IdleTimeTotal.AsDuration()
 	idleExpiration, isIdle := scheduler.getIdleExpiration(ctx, idleTimeTotal, time.Time{})
```
bergundy, line 42: I wonder if it'd be safer to store the task schedule time in the scheduler state and compare that. The scheduler would have to reset that field when needed.

PR #8335, opened 2025-09-16
nexus: propagate headers on cancel op

spkane31:
## What changed?
Propagate nexus headers on CancelOperation

## Why?
proper header handling

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
NA


components/nexusoperations/executors_test.go
```diff
@@ -658,6 +659,21 @@ func TestProcessCancelationTask(t *testing.T) {
...
 			},
 		},
+		{
+			name:            "success with headers",
+			requestTimeout:  time.Hour,
+			destinationDown: false,
+			header:          map[string]string{"key": "value"},
+			onCancelOperation: func(ctx context.Context, service, operation, token string, options nexus.CancelOperationOptions) error {
+				require.Equal(t, "value", options.Header["key"])
+				return nil
```
bergundy: This isn't using the subtest's `t` and can cause issues in test output and reporting.
You can instead do this:

```suggestion
				if options.Header["key"] != "value" {
					return nexus.HandlerErrorf(nexus.HandlerErrorBadRequest, `"key" header is not equal to "value"`)
				}
				return nil
```

bergundy:
LGTM apart from a single comment. Thanks.

PR #8342, opened 2025-09-17
fix invalid activity task token version check

hai719:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -658,6 +658,7 @@ message SyncActivityRequest {
     google.protobuf.Duration retry_maximum_interval = 23;
     int32 retry_maximum_attempts = 24;
     double retry_backoff_coefficient = 25;
+    int64 started_event_version = 26;
```
yycptt: nit: probably just call it `started_version` or `start_version` since we may not really have an event for it.

service/history/api/recordactivitytaskheartbeat/api.go
```diff
@@ -73,7 +73,9 @@ func Invoke(
 			if !isRunning ||
 				ai.StartedEventId == common.EmptyEventID ||
 				(token.GetScheduledEventId() != common.EmptyEventID && token.Attempt != ai.Attempt) ||
-				(token.GetVersion() != common.EmptyVersion && token.Version != ai.Version) {
+				(token.GetVersion() != common.EmptyVersion &&
```
yycptt: Shall we introduce a new field in task token? Quite confusing to repurpose an existing one...

service/history/workflow/mutable_state_impl.go
```diff
@@ -7051,6 +7056,7 @@ func (ms *MutableStateImpl) updatePendingEventIDs(
 	for scheduledEventID, startedEventID := range scheduledIDToStartedID {
 		if activityInfo, ok := ms.GetActivityInfo(scheduledEventID); ok {
 			activityInfo.StartedEventId = startedEventID
+			activityInfo.StartedEventVersion = ms.GetCurrentVersion()
```
yycptt: hmm why do we need to reset the start version here? I think it should already be set.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
hai719: update.

service/history/api/recordactivitytaskheartbeat/api.go
hai719: added.

service/history/workflow/mutable_state_impl.go
hai719: removed. Thanks for pointing out.

service/history/api/activity_util.go
```diff
@@ -52,3 +54,25 @@ func GetActivityScheduledEventID(
...
+			return true
+		}
+	}
+	if token.GetScheduledEventId() != common.EmptyEventID && token.Attempt != ai.Attempt {
+		return true
+	}
+	if token.GetStartVersion() != common.EmptyVersion && ai.GetStartVersion() != common.EmptyVersion {
+		return token.GetStartVersion() != ai.GetStartVersion()
+	}
+	if token.GetVersion() != common.EmptyVersion && token.GetVersion() != ai.GetVersion() {
```
yycptt, line 74: nit: plz add a comment and say this is for backward compatibility and we shouldn't actually check Version here.

service/history/api/activity_util.go
```diff
@@ -52,3 +54,25 @@ func GetActivityScheduledEventID(
 	}
 	return activityInfo.ScheduledEventId, nil
 }
+
+func IsActivityTaskNotFoundForToken(
+	token *tokenspb.Task,
+	ai *persistencespb.ActivityInfo,
+	isCompletedByID *bool,
```
yycptt, line 61: isCompletedByID is same as token.ScheduledEventId == 0?

service/history/api/activity_util.go
```diff
@@ -52,3 +54,25 @@ func GetActivityScheduledEventID(
...
+) bool {
+	if isCompletedByID == nil || !*isCompletedByID {
+		if ai.StartedEventId == common.EmptyEventID {
+			return true
+		}
+	}
+	if token.GetScheduledEventId() != common.EmptyEventID && token.Attempt != ai.Attempt {
```
yycptt, line 68: nit: I feel `token.GetScheduledEventId() != common.EmptyEventID` is not really necessary since it's the same as notCompletedByID. But I understand this is existing logic.

service/history/api/activity_util.go
hai719: Synced offline. These two are same, but for tasks other than recordactivitytaskcompleted, using token.ScheduledEventId would result in different logic.

service/history/api/activity_util.go
hai719: added.

PR #8344, opened 2025-09-17
extend softassert package

stephanos:
## What changed?

~Adds the Antithesis SDK as a dependency and two example use cases.~

~Note that the Antithesis SDK is behind a build tag.~

Extends the `softassert` package to (1) allow log tags and (2) add "Sometimes" assertion. This is both to align the with the Antithesis SDK's needs.

## Why?

Allow Antithesis SDK compatibility.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


.github/.golangci.yml
```diff
@@ -16,7 +16,16 @@ linters:
     - forbidigo
     - exhaustive
     - godox
+    - depguard
   settings:
+    depguard:
```
stephanos, line 21: Ensures that Antithesis SDK isn't used directly by accident.

.github/.golangci.yml
```diff
@@ -16,7 +16,16 @@ linters:
...
+      rules:
+        main:
+          deny:
+            - pkg: "github.com/antithesishq/antithesis-sdk-go/assert"
+              desc: "use internal antithesis package instead"
+          files:
+            - $all
```
stephanos: Ensures that Antithesis SDK isn't used directly by accident.

service/history/workflow/update/util.go
```diff
@@ -93,7 +98,10 @@ func (i *instrumentation) oneOf(counterName string) {
 }
 
 func (i *instrumentation) stateChange(updateID string, from, to state) {
-	i.log.Debug("update state change",
+	softassert.Sometimes(
```
stephanos, line 101: Any better ideas for naming?

.github/.golangci.yml
spkane31: Curious: why doesn't this fail the current PR? Shouldn't you require an `exclude: common/softassert` type of statement?

spkane31: Nvm, see there's a nolint below. Could we do a file exception instead?

common/softassert/softassert.go
```diff
@@ -6,7 +8,7 @@ import (
 )
 
 // That performs a soft assertion by logging an error if the given condition is false.
-// It is meant to indicate a condition that are always expected to be true.
+// It is meant to indicate a condition is always expected to be true.
```
spkane31, line 9: Already in here but "soft assertion" and "always expected to be true" seem contradictory to me.  What is the use case for a soft assertion?

spkane31 (approved):

common/softassert/softassert.go
spkane31: Better phrased as: is this our wrapper of antithesis' `sometimes` ? Or is this something different

go.mod
```diff
@@ -10,6 +10,7 @@ retract (
 require (
 	cloud.google.com/go/storage v1.51.0
 	github.com/Masterminds/sprig/v3 v3.3.0
+	github.com/antithesishq/antithesis-sdk-go v0.5.1-0.20250924165633-f60b0222f1b6
```
bergundy: Don't they have a stable release? We shouldn't have unstable dependencies in the server codebase.

common/softassert/softassert.go
stephanos: Right; that's a long long story. The `soft` comes from the desire to _not_ fail hard but let outside observers (such as testlogger, CICD and Antithesis) raise an alarm when it sees those.

stephanos: You'll see that `That` maps to Antithesis' `Always` (ie always expected to be true).

stephanos: Does the "Example" section in the comment address "What is the use case for a soft assertion?"?

go.mod
stephanos: No, they created this for just for us for now - I expect them to push a proper release once we report back.

Keep in mind that this has absolutely zero effect (ie is dead code eliminated) as it's in build tagged source code and the linter prevents you from using it. I'm also counting 18 unstable dependencies in go.mod 🙈

Is this a blocker?

stephanos: Keep in mind that this has absolutely zero effect (ie is dead code eliminated) as it's behind a build flagged source code and the linter prevents you from using it.

.github/.golangci.yml
stephanos: It seems I can actually omit the nolint as the entire file is behind a build tag so the linter ignores it already.

go.mod
stephanos: Thinking about this more; I actually don't need to add the Antithesis-specific parts anymore if we extend `common/softassert/softassert.go` as described here. I can just swap out the implementation on my end. Let me do that ...

PR #8345, opened 2025-09-18
Fail health check if no hosts are available

laniehei:
## What changed?
Instead of returning healthy, return unhealthy if no hosts available.

## Why?
 It's a fair assertion that if there are no hosts available, the service is not healthy.

## How did you test it?
- [x ] built
- [ x] run locally and tested manually
- [ ] added new unit test(s)

## Potential risks
There could be a false positive edge case that I'm not considering, but unlikely.


yux0 (approved):

yux0:
Is it possible to add a unit test?

PR #8346, opened 2025-09-18
fips: make defaults fips140-3 compatible

pjjw:
## What changed?
Changes to some hash functions and increases to test fixture key sizes to enable functionality under `GODEBUG=fips140=only,tlsmlkem=0`

## Why?
We need to run the Temporal server in a FIPS-compliant environment, and it'd be nice to see the tests pass first.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
Still calling the manifest hash "md5" even though it's not. Given its usage, this avoids any need for a migration.


tomelliot16:
Is there any movement on this?

PR #8351, opened 2025-09-18
tests: expand testrunner support to alert on panics or dataraces

spkane31:
## What changed?
Expanding `testrunner` tool to add alerts for panics and data races. Sample table with an injected data race (and real panic):

<table><tr><th><th>Tests</th><th>Passed ✅</th><th>Skipped ⏭️</th><th>Failed ❌</th></tr><tr><td>JUnit Test Report</td><td>721 ran</td><td>340 passed</td><td>370 skipped</td><td>11 failed</td></tr></table>
<table><tr><th>Test</th><th>Result</th></tr><tr><td colspan="2"><strong>JUnit Test Report</strong></td></tr><tr><td>functional-test (crash)</td><td>❌ failure</td></tr><tr><td>io/server/tests.TestActivityApiBatchResetClientTestSuite/TestActivityBatchReset_DataRace</td><td>❌ failure</td></tr><tr><td>io/server/tests.TestTaskQueueStats_Pri_Suite/TestNoTasks_ValidateStats</td><td>❌ failure</td></tr><tr><td>io/server/tests.TestVersioning3FunctionalSuite/TestUnpinnedQuery_NoSticky/ForceTaskForwardForcePollForwardForceAsync</td><td>❌ failure</td></tr><tr><td>io/server/tests.TestVersioningFunctionalSuite/TestDispatchQueryOld/ForceTaskForwardForcePollForwardForceAsync</td><td>❌ failure</td></tr><tr><td>io/server/tests.TestActivityApiBatchResetClientTestSuite/TestActivityBatchReset_DataRace (retry 1)</td><td>❌ failure</td></tr><tr><td>io/server/tests.TestActivityApiBatchResetClientTestSuite/TestActivityBatchReset_DataRace (retry 2)</td><td>❌ failure</td></tr><tr><td>func1.DATA RACE: Data race detected — in go.temporal.io/server/tests.(*ActivityApiBatchResetClientTestSuite).TestActivityBatchReset_DataRace.func1</td><td>❌ failure</td></tr><tr><td>PANIC: runtime error: invalid memory address or nil pointer dereference — in TestTaskQueueStats_Pri_Suite/TestAddMultipleTasks_MultiplePartitions_ValidateStats_Cached</td><td>❌ failure</td></tr><tr><td>func1.DATA RACE: Data race detected — in go.temporal.io/server/tests.(*ActivityApiBatchResetClientTestSuite).TestActivityBatchReset_DataRace.func1</td><td>❌ failure</td></tr><tr><td>func1.DATA RACE: Data race detected — in go.temporal.io/server/tests.(*ActivityApiBatchResetClientTestSuite).TestActivityBatchReset_DataRace.func1</td><td>❌ failure</td></tr></table>

Source: https://github.com/temporalio/temporal/actions/runs/17870152469?pr=8351

## Why?
This change will help the temporal team find bugs sooner.

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
NA


tools/testrunner/junit_test.go
```diff
@@ -115,6 +115,46 @@ func TestMergeReports_MissingRerun(t *testing.T) {
...
+	require.NotNil(t, tc1.Failure)
+	require.Equal(t, "PANIC", tc1.Failure.Message)
+	require.Contains(t, tc1.Failure.Data, "panic: This is a panic")
+	require.Contains(t, tc1.Failure.Data, "Detected in tests:")
+	require.Contains(t, tc1.Failure.Data, "TestPanicExample")
+
+	// Ensure totals updated at the top level.
+	require.Equal(t, 2, j.Failures)
+	require.Equal(t, 2, j.Tests)
+}
```
stephanos, line 138: I wonder if it would be easier to leverage `requireReportEquals` to just compare the file output verbatim?

(`TestMergeReports_MultipleReports` is not a good example, it should probably also do that, too)

tools/testrunner/testrunner.go
```diff
@@ -57,7 +57,7 @@ func (a *attempt) run(ctx context.Context, args []string) (string, error) {
 	if err := cmd.Run(); err != nil {
 		return output.String(), err
 	}
-	return "", nil
+	return output.String(), nil
```
stephanos: could do 
```
err := cmd.Run()
return output.String(), err
```

tools/testrunner/testrunner.go
```diff
@@ -212,6 +213,8 @@ func (r *runner) runTests(ctx context.Context, args []string) {
 
 		// Run tests.
 		stdout, err := currentAttempt.run(ctx, args)
+		// Extract prominent alerts from this attempt's output.
+		r.alerts = append(r.alerts, parseAlerts(stdout)...)
```
stephanos, line 215: I forgot that the test-runner collects the entire output in memory. I am a bit concerned about memory usage since it would claim memory the tests and dependencies can't use. Could be solved with a stream-based log parser, but I also don't want to blow up scope. Let's keep that in mind for the future.

tools/testrunner/log_test.go
```diff
@@ -21,3 +21,40 @@ func TestParseTestTimeouts(t *testing.T) {
...
+func TestParseAlerts_DataRaceAndPanic(t *testing.T) {
+	input, err := os.ReadFile("testdata/alerts-input.log")
+	require.NoError(t, err)
+
+	alerts := parseAlerts(string(input))
+	require.NotEmpty(t, alerts)
+
+	// Expect at least one data race and one panic alert.
+	var hasRace, hasPanic bool
+	for _, a := range alerts {
```
stephanos: Should the test be more specific in checking there are 3 alerts (right?) and assert them directly. The loop-based approach makes it a bit indirect.

tools/testrunner/log.go
```diff
@@ -65,3 +66,285 @@ func testOnlyStacktrace(stacktrace string) string {
...
+	return isTestResultBoundary(line)
+}
+
+func shouldStopOnTestBoundary(line string, _ int, _ int) bool {
+	return isTestResultBoundary(line)
+}
+
+// dedupeAlerts removes duplicate alerts (e.g., repeated across retries) based
+// on kind and details while preserving the first-seen order.
+func dedupeAlerts(alerts []alert) []alert {
```
stephanos: Would it make sense to do this in the JUnit merge part? That way it would be alert-agnostic and seems like a good place for this functionality.

tools/testrunner/log.go
```diff
@@ -65,3 +66,285 @@ func testOnlyStacktrace(stacktrace string) string {
...
+	alertKindDataRace alertKind = "DATA RACE"
+	alertKindPanic    alertKind = "PANIC"
+	alertKindFatal    alertKind = "FATAL"
+)
+
+// alert captures a prominent issue detected from stdout/stderr of test runs.
+type alert struct {
+	Kind    alertKind
+	Summary string
+	Details string
```
stephanos, line 82: At first I thought we should keep things simple and not include the detailed logs in the report, but it doesn't seem that complicated and very helpful; so seems good to me now.

tools/testrunner/log.go
```diff
@@ -65,3 +66,285 @@ func testOnlyStacktrace(stacktrace string) string {
...
+	}
+	log.Printf("%s\nFULL ALERT DETAILS:\n%s", banner, banner)
+	for idx, a := range alerts {
+		if len(a.Tests) > 0 {
+			log.Printf("-- Alert %d: [%s] %s --\nDetected in tests:\n\t%s\n\n%s", idx+1, a.Kind, a.Summary, strings.Join(a.Tests, "\n\t"), a.Details)
+		} else {
+			log.Printf("-- Alert %d: [%s] %s --\n%s", idx+1, a.Kind, a.Summary, a.Details)
+		}
+	}
+	log.Printf("%s\n", banner)
```
stephanos: Should/could some of these be `log.Println`? For example, not seeing a newline in the for loop anywhere and I wonder if that looks right.

Could consider adding this function to the tests to verify.

bergundy:
For the generated report. I am wondering if it's the best approach to add these panics as failed tests.
If we do this, some tools that examine junit reports, would show that some runs have more tests than others.
As long as we put the new failures in separate junit report files, this should be fine.

Also, could you show the test path when you print the panic? e.g. instead of:

> PANIC: test panic [recovered, repanicked] — in TestShowPanic

Show:

`PANIC: test panic [recovered, repanicked] — in io/server/tools/testrunner.TestShowPanic`

tools/testrunner/testrunner.go
spkane31: I had this thought, I'll create a tracking issue for it.

tools/testrunner/log_test.go
spkane31: Caught a bug here, there should only be two. The data race wasn't being parsed correctly.

tools/testrunner/log.go
spkane31: Removing this for now

tools/testrunner/junit_test.go
spkane31: I thought that too but the implementation was a little funky and I thought this was cleaner.

tools/testrunner/log_test.go
stephanos: Okay; but then why not just assert alerts[0] and alerts[1]? I don't see the need for the loop when there's two alerts. Would be much simpler to read and follow IMO.

tools/testrunner/junit_test.go
stephanos: Why's that? If you compare it with an empty file at first, you then copy the expected file in there and the next run should work, no? Assuming the output is deterministic (I'd expect it to).

stephanos (approved):
Just minor notes on testing; nothing blocking.

PR #8354, opened 2025-09-19
Reduce dynamic config client log level to INFO to address noisy debug logging

chaptersix:
## What changed?
Added functionality to override the dynamic config client log level to INFO for all services in the functional tests. 

### Why?
The dynamic config client introduced in #7052 logs "No such key in dynamic config, using default" at DEBUG level every time a dynamic config key is checked without an override. This creates extremely noisy logs during testing as it fires for many dynamic config checks, making it difficult to see relevant test output and debug information.

This approach:
- Only affects the dynamic config client logging, not the main service logs
- Only applicable to tests not production code

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/testcore/onebox.go
+++ b/tests/testcore/onebox.go
@@ -108,14 +108,15 @@ type (
 		captureMetricsHandler            *metricstest.CaptureHandler
 		hostsByProtocolByService         map[transferProtocol]map[primitives.ServiceName]static.Hosts
 
-		onGetClaims           func(*authorization.AuthInfo) (*authorization.Claims, error)
-		onAuthorize           func(context.Context, *authorization.Claims, *authorization.CallTarget) (authorization.Result, error)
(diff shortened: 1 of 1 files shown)


tests/testcore/onebox.go
```diff
@@ -210,6 +211,7 @@ func newTemporal(t *testing.T, params *TemporalParams) *TemporalImpl {
 		dcClient:                         dynamicconfig.NewMemoryClient(),
 		// If this doesn't build, make sure you're building with tags 'test_dep':
 		testHooks:                testhooks.NewTestHooksImpl(),
+		dynamicConfigLogger:      log.NewZapLogger(log.BuildZapLogger(log.Config{Level: "info"})),
```
stephanos, line 218: Maybe a short comment here why `info` is chosen.

stephanos (approved):
Awesome!

PR #8358, opened 2025-09-22
enable iotamixing linter

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


common/membership/ringpop/service_resolver.go
```diff
@@ -48,7 +48,9 @@ const (
 	minRefreshInternal     = time.Second * 4
 	defaultRefreshInterval = time.Second * 10
 	replicaPoints          = 100
+)
 
+const (
```
stephanos, line 53: One and only finding flagged by `iotamixing`. Safe to change as `refreshMode`s are only used internally and not persisted.

yycptt (approved):

PR #8359, opened 2025-09-22
Add metrics to track the health of worker registry

rkannan82:
## What changed?
Added 2 metrics to debug any memory related issues with the registry cache.

Metrics to answer the following questions:
- Utilization ratio. Indicates memory pressure as this gets closer to 1.
- Are we not able to evict entries when we hit capacity limits?

Note: If needed, we can later add another metric to track the number of entries per namespace. This can be useful to know if the growth is coming from specific namespace.

## Why?
Debug memory issues related to this new feature.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None.

### DIFF
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1110,6 +1110,25 @@ var (
 		WithDescription("Count of tasks that hit a transient error during match or forward and are retried immediately"),
 	)
 
+	// ----------------------------------------------------------------------------------------------------------------
+	// Matching Workers Registry
+	MatchingRegistryEntriesMetric = NewGaugeDef(
+		"matching_registry_entries",
+		WithDescription("Number of entries in the registry, broken down by namespace."),
+	)
+	MatchingRegistryEvictionBlockedByAgeMetric = NewGaugeDef(
+		"matching_registry_eviction_blocked_by_age",
+		WithDescription(
+			"Set if entries could not be evicted due to minEvictAge policy in a given eviction iteration. "+
+				"Reset once a subsequent eviction succeeds.",
+		),
+	)
+	MatchingRegistryCapacityUtilizationMetric = NewGaugeDef(
+		"matching_registry_capacity_utilization",
+		WithDescription("Tracks the ratio of total entries to maxItems."),
+	)
+	// ----------------------------------------------------------------------------------------------------------------
+
 	// Versioning and Reachability
 	ReachabilityExitPointCounter = NewCounterDef("reachability_exit_point_count")
 
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1111,16 +1111,16 @@ var (
 	)
 
 	// ----------------------------------------------------------------------------------------------------------------
-	// Matching Workers Registry
-	MatchingRegistryEvictionBlockedByAgeMetric = NewGaugeDef(
-		"matching_registry_eviction_blocked_by_age",
+	// Matching service: Metrics to track the health of worker registry.
+	WorkerRegistryEvictionBlockedByAgeMetric = NewGaugeDef(
+		"worker_registry_eviction_blocked_by_age",
 		WithDescription(
 			"Set if entries could not be evicted due to minEvictAge policy in a given eviction iteration. "+
 				"Reset once a subsequent eviction succeeds.",
 		),
 	)
-	MatchingRegistryCapacityUtilizationMetric = NewGaugeDef(
-		"matching_registry_capacity_utilization",
+	WorkerRegistryCapacityUtilizationMetric = NewGaugeDef(
+		"worker_registry_capacity_utilization",
 		WithDescription("Tracks the ratio of total entries to maxItems."),
 	)
 	// ----------------------------------------------------------------------------------------------------------------
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1112,10 +1112,6 @@ var (
 
 	// ----------------------------------------------------------------------------------------------------------------
 	// Matching Workers Registry
-	MatchingRegistryEntriesMetric = NewGaugeDef(
-		"matching_registry_entries",
-		WithDescription("Number of entries in the registry, broken down by namespace."),
-	)
 	MatchingRegistryEvictionBlockedByAgeMetric = NewGaugeDef(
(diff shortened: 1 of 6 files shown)


Shivs11:
bugbot

Shivs11:
bugbot run

service/matching/workers/registry_impl.go
```diff
@@ -21,6 +22,19 @@ const (
...
+	entriesMetric = "entries"
+
+	// Set if entries could not be evicted due to minEvictAge policy in a given eviction iteration.
+	// Note: The metric is reset once a subsequent eviction succeeds.
+	evictionBlockedByAgeMetric = "eviction_blocked_by_age"
+
+	// Tracks the ratio of total entries to maxItems.
+	capacityUtilizationMetric = "capacity_utilization"
+)
+
```
Shivs11: I think we could maybe place these new metric definitions in this file: common/metrics/metric_defs.go

Most user required metrics are placed there imho

service/matching/workers/registry_impl.go
```diff
@@ -228,6 +246,36 @@ func (m *registryImpl) upsertHeartbeats(nsID namespace.ID, heartbeats []*workerp
...
+	m.recordEntriesMetric(nsID)
+	m.recordUtilizationMetric()
+}
+
+// recordEntriesMetric records the current number of entries for a given namespace.
+func (m *registryImpl) recordEntriesMetric(nsID namespace.ID) {
+	count := len(m.filterWorkers(nsID, func(_ *workerpb.WorkerHeartbeat) bool { return true }))
+	m.metricsHandler.Gauge(entriesMetric).Record(
+		float64(count),
+		metrics.NamespaceTag(nsID.String()),
```
Shivs11: I think this value which you are injecting for a NamespaceTag should be a NamespaceName instead of an ID

If you wanna use the ID, you should use `metrics.NamespaceIDTag` 

I think it's better to use the NamespaceName over here imo since operators would wanna see the namespace name rather than the ID when squinting at these metrics

service/matching/workers/registry_impl_test.go
```diff
@@ -170,11 +233,90 @@ func TestEvictByCapacityAfterMinAge(t *testing.T) {
...
+// and overall capacity utilization across multiple namespaces.
+func TestMultiNamespaceMetrics(t *testing.T) {
+	maxItems := int64(10)
+	captureHandler := metricstest.NewCaptureHandler()
+	capture := captureHandler.StartCapture()
+	defer captureHandler.StopCapture(capture)
+
+	m := newRegistryImpl(
+		2, time.Hour, 0, maxItems, time.Hour, captureHandler,
+	)
```
Shivs11: nit: maybe a helper here since I spotted this in an earlier function too?

service/matching/workers/registry_impl.go
rkannan82: Ah good point!

service/matching/workers/registry_impl.go
rkannan82: Ah this is much better now. Thanks.

service/matching/workers/registry_impl_test.go
rkannan82: Since this is just creating the object, suggest to keep it as is.

common/metrics/metric_defs.go
```diff
@@ -1110,6 +1110,21 @@ var (
 		WithDescription("Count of tasks that hit a transient error during match or forward and are retried immediately"),
 	)
 
+	// ----------------------------------------------------------------------------------------------------------------
+	// Matching Workers Registry
+	MatchingRegistryEvictionBlockedByAgeMetric = NewGaugeDef(
```
Shivs11: any reason why you chose prepending the metric names with matching? We could have another registry inside of matching which could lead to confusion for an operator in the near future imho 

how does something like _WorkerRegistryEvictionBlockedByAge_ sound to ya?

common/metrics/metric_defs.go
```diff
@@ -1110,6 +1110,21 @@ var (
...
+		"matching_registry_eviction_blocked_by_age",
+		WithDescription(
+			"Set if entries could not be evicted due to minEvictAge policy in a given eviction iteration. "+
+				"Reset once a subsequent eviction succeeds.",
+		),
+	)
+	MatchingRegistryCapacityUtilizationMetric = NewGaugeDef(
+		"matching_registry_capacity_utilization",
+		WithDescription("Tracks the ratio of total entries to maxItems."),
+	)
```
Shivs11, line 1125: hmm, I quite liked your idea of adding the previously included metric of the per namespace entries - more often than not, it's one particular namespace you are interested in when you are an on-call operator 

I think we should def include it in a subsequent PR, if not for this one - wdyt?
I know you mentioned it added a bit more complexity than you like in this one so I shall leave the decision to you as to when we should add it

service/matching/workers/registry_impl.go
```diff
@@ -228,6 +233,25 @@ func (m *registryImpl) upsertHeartbeats(nsID namespace.ID, heartbeats []*workerp
 	b := m.getBucket(nsID)
 	newEntries := b.upsertHeartbeats(nsID, heartbeats)
 	m.total.Add(newEntries)
+	m.recordUtilizationMetric()
+}
+
+// recordUtilizationMetric records the overall capacity utilization ratio.
+func (m *registryImpl) recordUtilizationMetric() {
+	utilization := float64(m.total.Load()) / float64(m.maxItems)
+	metrics.MatchingRegistryCapacityUtilizationMetric.With(m.metricsHandler).Record(utilization)
```
Shivs11: just asking for confirmation - this metrics handler shall include the namespace name tags, right?

common/metrics/metric_defs.go
rkannan82: Yea, I will add the per ns usage breakdown in a separate PR as this requires a bit of refactoring of the the bookkeeping logic.

service/matching/workers/registry_impl.go
rkannan82: No these metrics are not per namespace. This is an aggregate usage of that node.

common/metrics/metric_defs.go
rkannan82: Done.

Shivs11 (approved):
ty!

PR #8360, opened 2025-09-23
Fix replication task skip count calculation

xwduan:
## What changed?
Fix replication task skip count calculation.

## Why?
If we skip the task due to namespace clusters check, we should also count as skipped task. So empty task will be sent to passive side to update watermark.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
no risk.

### DIFF
--- a/service/history/replication/stream_sender.go
+++ b/service/history/replication/stream_sender.go
@@ -529,7 +529,7 @@ Loop:
 			continue Loop
 		}
 		if !s.shouldProcessTask(item) {
-			continue
+			continue Loop
 		}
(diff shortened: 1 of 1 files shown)


service/history/replication/stream_sender.go
```diff
@@ -526,11 +522,15 @@ Loop:
 			}); err != nil {
 				return err
 			}
+			skipCount = 0
```
xwduan, line 525: This is the efficiency improvement, so we will not send too many empty message to stream. No correctness issue.

PR #8367, opened 2025-09-23
Fix Single Test Run Failures

chaptersix:
## What changed?
Add condition for getting the job id during test runs.

## Why?
Some jobs will fail to find the local action since the checkout does not fire. Adding the same condition to to get job id should skip the step for single test runs.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -512,6 +512,7 @@ jobs:
         with:
           path: ~/.cache/go-build
           key: go-${{ runner.os }}${{ runner.arch }}-build-${{ env.COMMIT }}
+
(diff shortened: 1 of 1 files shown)


chaptersix:
https://github.com/temporalio/temporal/actions/runs/17956902324/job/51070407513
Single test run pass

.github/workflows/run-tests.yml
```diff
@@ -513,6 +513,7 @@ jobs:
           path: ~/.cache/go-build
           key: go-${{ runner.os }}${{ runner.arch }}-build-${{ env.COMMIT }}
       - name: Get job ID
```
stephanos, line 516: nit: can we add a newline above this?

PR #8370, opened 2025-09-23
visibility(chasm): enable querying w/out Temporal prefix

spkane31:
## What changed?
When resolving search attributes allow users to query for `TemporalXXX` search attributes w/out the `Temporal` prefix. The server will default to the user namespace search attributes, then system/predefined search attributes, then system/predefined `TemporalXXX` search attributes.

## Why?
This will unblock future chasm workstreams and allow users to create search attributes that might conflict with existing system/predefined SAs.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
...

### DIFF
--- a/service/history/queues/dlq_writer.go
+++ b/service/history/queues/dlq_writer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
@@ -20,6 +21,7 @@ type (
 		metricsHandler    metrics.Handler
 		logger            log.SnTaggedLogger
 		namespaceRegistry namespace.Registry
+		enqueueMutex      sync.Map // map[persistence.QueueKey]*sync.Mutex for per-queue locking
 	}
 	// QueueWriter is a subset of persistence.HistoryTaskQueueManager.
 	QueueWriter interface {
@@ -77,13 +79,21 @@ func (q *DLQWriter) WriteTaskToDLQ(
 		}
 	}
 
-	resp, err := q.dlqWriter.EnqueueTask(ctx, &persistence.EnqueueTaskRequest{
-		QueueType:     queueKey.QueueType,
-		SourceCluster: queueKey.SourceCluster,
-		TargetCluster: queueKey.TargetCluster,
-		Task:          task,
-		SourceShardID: sourceShardID,
-	})
+	resp, err := func() (*persistence.EnqueueTaskResponse, error) {
+		// Acquire a process-level lock for this specific DLQ to prevent concurrent writes
+		// from multiple shards causing CAS conflicts in the persistence layer.
+		mu := q.getQueueMutex(queueKey)
+		mu.Lock()
+		defer mu.Unlock()
+
+		return q.dlqWriter.EnqueueTask(ctx, &persistence.EnqueueTaskRequest{
+			QueueType:     queueKey.QueueType,
+			SourceCluster: queueKey.SourceCluster,
+			TargetCluster: queueKey.TargetCluster,
+			Task:          task,
+			SourceShardID: sourceShardID,
+		})
+	}()
 	if err != nil {
 		return fmt.Errorf("%w: %v", ErrSendTaskToDLQ, err)
 	}
@@ -116,4 +126,16 @@ func (q *DLQWriter) WriteTaskToDLQ(
 		namespaceTag,
 	)
 	return nil
+}
+
+// getQueueMutex returns a per-queue mutex, creating it if it doesn't exist.
+// This provides process-level locking to serialize concurrent writes to the same queue.
+func (q *DLQWriter) getQueueMutex(queueKey persistence.QueueKey) *sync.Mutex {
+	if mu, ok := q.enqueueMutex.Load(queueKey); ok {
+		return mu.(*sync.Mutex) //nolint:revive
+	}
+
+	newMutex := &sync.Mutex{}
+	actual, _ := q.enqueueMutex.LoadOrStore(queueKey, newMutex)
+	return actual.(*sync.Mutex) //nolint:revive
 }
--- a/service/history/queues/dlq_writer_test.go
+++ b/service/history/queues/dlq_writer_test.go
@@ -3,7 +3,10 @@ package queues_test
 import (
 	"context"
 	"errors"
+	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -13,11 +16,13 @@ import (
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/metrics/metricstest"
 	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/persistence"
 	"go.temporal.io/server/service/history/queues"
 	"go.temporal.io/server/service/history/queues/queuestest"
 	"go.temporal.io/server/service/history/tasks"
 	"go.temporal.io/server/service/history/tests"
 	"go.uber.org/mock/gomock"
+	"golang.org/x/sync/errgroup"
 )
 
 type (
@@ -27,11 +32,14 @@ type (
 	}
 	logRecorder struct {
 		log.SnTaggedLogger
+		mu      sync.Mutex
 		records []logRecord
 	}
 )
 
 func (l *logRecorder) Warn(msg string, tags ...tag.Tag) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
 	l.records = append(l.records, logRecord{msg: msg, tags: tags})
 }
 
@@ -127,4 +135,159 @@ func TestDLQWriter_Ok(t *testing.T) {
 	assert.Equal(t, "transfer", recordings[0].Tags[metrics.TaskCategoryTagName])
 	namespaceStateTag := metrics.NamespaceStateTag("active")
 	assert.Equal(t, "active", recordings[0].Tags[namespaceStateTag.Key])
+}
+
+func TestDLQWriter_ConcurrentWrites(t *testing.T) {
+	t.Parallel()
+
+	// This test verifies that the DLQ writer serializes concurrent writes to the same queue
+	// using a process-level lock, preventing CAS conflicts in the persistence layer.
+	queueWriter := &queuestest.FakeQueueWriter{}
+	ctrl := gomock.NewController(t)
+	namespaceRegistry := namespace.NewMockRegistry(ctrl)
+	namespaceRegistry.EXPECT().GetNamespaceByID(gomock.Any()).Return(&namespace.Namespace{}, nil).AnyTimes()
+	logger := &logRecorder{SnTaggedLogger: log.NewTestLogger()}
+	metricsHandler := metricstest.NewCaptureHandler()
+	writer := queues.NewDLQWriter(queueWriter, metricsHandler, logger, namespaceRegistry)
+
+	const numConcurrentWrites = 50
+	var g errgroup.Group
+	var concurrentAccessCount atomic.Int32
+	var maxConcurrentAccess atomic.Int32
+
+	// Create tasks that will write to the same DLQ (same category, source, target)
+	testTasks := make([]*tasks.WorkflowTask, numConcurrentWrites)
+	for i := 0; i < numConcurrentWrites; i++ {
+		testTasks[i] = &tasks.WorkflowTask{
+			WorkflowKey: definition.WorkflowKey{
+				NamespaceID: string(tests.NamespaceID),
+				WorkflowID:  tests.WorkflowID,
+				RunID:       tests.RunID,
+			},
+		}
+	}
+
+	// Wrap the queue writer to track concurrent access
+	queueWriter.EnqueueTaskFunc = func(ctx context.Context, request *persistence.EnqueueTaskRequest) (*persistence.EnqueueTaskResponse, error) {
+		// Increment concurrent access counter
+		current := concurrentAccessCount.Add(1)
+
+		// Track max concurrent access
+		for {
+			maxWrites := maxConcurrentAccess.Load()
+			if current <= maxWrites || maxConcurrentAccess.CompareAndSwap(maxWrites, current) {
+				break
+			}
+		}
+
+		// Simulate some work that could cause race conditions.
+		time.Sleep(10 * time.Millisecond) //nolint:forbidigo
+
+		// Decrement counter
+		concurrentAccessCount.Add(-1)
+
+		return &persistence.EnqueueTaskResponse{Metadata: persistence.MessageMetadata{ID: 0}}, nil
+	}
+
+	// Launch concurrent writes to the same DLQ
+	for i := 0; i < numConcurrentWrites; i++ {
+		task := testTasks[i]
+		g.Go(func() error {
+			err := writer.WriteTaskToDLQ(
+				context.Background(),
+				"source-cluster",
+				"target-cluster",
+				1, // same shard ID
+				task,
+				true,
+			)
+			require.NoError(t, err)
+			return nil
+		})
+	}
+
+	require.NoError(t, g.Wait())
+
+	// Verify all writes succeeded
+	require.Len(t, queueWriter.EnqueueTaskRequests, numConcurrentWrites)
+
+	// The key assertion: with the lock in place, we should never have more than 1 concurrent access
+	assert.Equal(t, int32(1), maxConcurrentAccess.Load(),
+		"Expected serialized access (max 1 concurrent), but got %d concurrent accesses. "+
+			"This indicates the lock is not working properly.", maxConcurrentAccess.Load())
+}
+
+func TestDLQWriter_ConcurrentWritesDifferentQueues(t *testing.T) {
+	t.Parallel()
+
+	// This test verifies that concurrent writes to DIFFERENT queues can proceed in parallel
+	queueWriter := &queuestest.FakeQueueWriter{}
+	ctrl := gomock.NewController(t)
+	namespaceRegistry := namespace.NewMockRegistry(ctrl)
+	namespaceRegistry.EXPECT().GetNamespaceByID(gomock.Any()).Return(&namespace.Namespace{}, nil).AnyTimes()
+	logger := &logRecorder{SnTaggedLogger: log.NewTestLogger()}
+	metricsHandler := metricstest.NewCaptureHandler()
+	writer := queues.NewDLQWriter(queueWriter, metricsHandler, logger, namespaceRegistry)
+
+	const numConcurrentWrites = 50
+	const numQueues = 5
+	var g errgroup.Group
+	var concurrentAccessCount atomic.Int32
+	var maxConcurrentAccess atomic.Int32
+
+	// Wrap the queue writer to track concurrent access
+	queueWriter.EnqueueTaskFunc = func(ctx context.Context, request *persistence.EnqueueTaskRequest) (*persistence.EnqueueTaskResponse, error) {
+		current := concurrentAccessCount.Add(1)
+
+		for {
+			maxValue := maxConcurrentAccess.Load()
+			if current <= maxValue || maxConcurrentAccess.CompareAndSwap(maxValue, current) {
+				break
+			}
+		}
+
+		// Simulate some work that could cause race conditions.
+		time.Sleep(10 * time.Millisecond) //nolint:forbidigo
+		concurrentAccessCount.Add(-1)
+
+		return &persistence.EnqueueTaskResponse{Metadata: persistence.MessageMetadata{ID: 0}}, nil
+	}
+
+	// Launch concurrent writes to DIFFERENT target clusters (different DLQs)
+	for i := 0; i < numConcurrentWrites; i++ {
+		index := i
+		g.Go(func() error {
+			task := &tasks.WorkflowTask{
+				WorkflowKey: definition.WorkflowKey{
+					NamespaceID: string(tests.NamespaceID),
+					WorkflowID:  tests.WorkflowID,
+					RunID:       tests.RunID,
+				},
+			}
+			// Use different target clusters to create different queue keys
+			targetCluster := "target-cluster-" + string(rune('A'+index%numQueues))
+			err := writer.WriteTaskToDLQ(
+				context.Background(),
+				"source-cluster",
+				targetCluster,
+				1,
+				task,
+				true,
+			)
+			require.NoError(t, err)
+			return nil
+		})
+	}
+
+	require.NoError(t, g.Wait())
+
+	// Verify all writes succeeded
+	require.Len(t, queueWriter.EnqueueTaskRequests, numConcurrentWrites)
+
+	// Since these are different queues, they should be able to execute concurrently
+	// We expect to see more than 1 concurrent access, but less than or equal to numQueues.
+	assert.Greater(t, maxConcurrentAccess.Load(), int32(1),
+		"Expected concurrent access to different queues (> 1), but got %d.", maxConcurrentAccess.Load())
+	assert.LessOrEqual(t, maxConcurrentAccess.Load(), int32(numQueues),
+		"Expected less than %d concurrent accesses, but got %d.", numQueues, maxConcurrentAccess.Load())
 }
--- a/service/history/queues/queuestest/fake_queue_writer.go
+++ b/service/history/queues/queuestest/fake_queue_writer.go
@@ -2,25 +2,40 @@ package queuestest
 
 import (
 	"context"
+	"sync"
 
 	"go.temporal.io/server/common/persistence"
 	"go.temporal.io/server/service/history/queues"
 )
 
+// EnqueueTaskFunc is a function type for custom EnqueueTask behavior in tests
+type EnqueueTaskFunc func(context.Context, *persistence.EnqueueTaskRequest) (*persistence.EnqueueTaskResponse, error)
+
 // FakeQueueWriter is a [queues.QueueWriter] which records the requests it receives and returns the provided errors.
 type FakeQueueWriter struct {
+	mu                  sync.Mutex
 	EnqueueTaskRequests []*persistence.EnqueueTaskRequest
 	EnqueueTaskErr      error
 	CreateQueueErr      error
+	// EnqueueTaskFunc allows tests to provide custom behavior for EnqueueTask calls.
+	// If set, this function is called instead of the default behavior.
+	EnqueueTaskFunc EnqueueTaskFunc
 }
 
 var _ queues.QueueWriter = (*FakeQueueWriter)(nil)
 
 func (d *FakeQueueWriter) EnqueueTask(
-	_ context.Context,
+	ctx context.Context,
 	request *persistence.EnqueueTaskRequest,
 ) (*persistence.EnqueueTaskResponse, error) {
+	// Protect the slice append from concurrent access
+	d.mu.Lock()
 	d.EnqueueTaskRequests = append(d.EnqueueTaskRequests, request)
+	d.mu.Unlock()
+
+	if d.EnqueueTaskFunc != nil {
+		return d.EnqueueTaskFunc(ctx, request)
+	}
 	return &persistence.EnqueueTaskResponse{Metadata: persistence.MessageMetadata{ID: 0}}, d.EnqueueTaskErr
 }
 


common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -48,37 +49,68 @@ func NewValuesInterceptor(
...
 }
 
-// TODO: this is invoked for non-ES validation code flow. Needs refactoring
-func (ni *nameInterceptor) Name(name string, usage query.FieldNameUsage) (string, error) {
-	fieldName := name
-	if searchattribute.IsMappable(name) {
-		mapper, err := ni.searchAttributesMapperProvider.GetMapper(ni.namespace)
-		if err != nil {
-			return "", err
+func resolveSearchAttributeAlias(
```
bergundy: Does this need to be implemented per visibility store implementation?
It would be ideal if this function was defined once.

common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -48,37 +49,68 @@ func NewValuesInterceptor(
...
+	name string, ns namespace.Name,
+	mapperProvider searchattribute.MapperProvider,
+	saTypeMap searchattribute.NameTypeMap,
+) (string, enumspb.IndexedValueType, error) {
+	// 1. Skip mapping to custom search attribute if the name has "Temporal" prefix
+	if strings.HasPrefix(name, "Temporal") {
+		// Check if it's a system/predefined search attribute with Temporal prefix
+		if saType, err := saTypeMap.GetType(name); err == nil {
```
bergundy: We also need to support querying `BuildIds` as `TemporalBuildIds`.

common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -48,37 +49,68 @@ func NewValuesInterceptor(
...
 	}
 
-	fieldType, err := ni.searchAttributesTypeMap.GetType(fieldName)
+	// 3. If it exists as a system or pre-defined attribute, map it
+	if saType, err := saTypeMap.GetType(name); err == nil {
+		return name, saType, nil
+	}
+	if saType, err := saTypeMap.GetType(fmt.Sprintf("Temporal%s", name)); err == nil {
+		return fmt.Sprintf("Temporal%s", name), saType, nil
+	}
```
bergundy: nit:

```suggestion
	prefixedName := "Temporal" + name
	if saType, err := saTypeMap.GetType(prefixedName); err == nil {
		return prefixedName, saType, nil
	}
```

common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -48,37 +49,68 @@ func NewValuesInterceptor(
...
-func (ni *nameInterceptor) Name(name string, usage query.FieldNameUsage) (string, error) {
-	fieldName := name
-	if searchattribute.IsMappable(name) {
-		mapper, err := ni.searchAttributesMapperProvider.GetMapper(ni.namespace)
-		if err != nil {
-			return "", err
+func resolveSearchAttributeAlias(
+	name string, ns namespace.Name,
+	mapperProvider searchattribute.MapperProvider,
+	saTypeMap searchattribute.NameTypeMap,
```
bergundy: Can you confirm that this includes both system and predefined attributes?

common/persistence/visibility/store/elasticsearch/query_interceptors.go
bergundy: I see that you do the inverse of this below.
It'd probably be best if `saTypeMap` normalized the access instead of doing that in this function.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,222 @@
...
+	return "done!", nil
+}
+
+func (ts *WorkflowAliasSearchAttributeTestSuite) createWorkflow(ctx context.Context, workflowFn WorkflowFunction, prefix string, scheduleByID string) sdkclient.WorkflowRun {
+	workflowOptions := sdkclient.StartWorkflowOptions{
+		ID:        testcore.RandomizeStr(prefix + "-" + ts.T().Name()),
+		TaskQueue: ts.TaskQueue(),
+	}
+	if scheduleByID != "" {
+		scheduleByIDKey := temporal.NewSearchAttributeKeyKeyword("CustomerSA")
```
bergundy:
```suggestion
		scheduleByIDKey := temporal.NewSearchAttributeKeyKeyword("CustomSA")
```

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,222 @@
...
+		},
+	}), ts.ActivityFunc).Get(ctx, nil)
+	return err
+}
+
+func (ts *WorkflowAliasSearchAttributeTestSuite) ActivityFunc() (string, error) {
+	return "done!", nil
+}
+
+func (ts *WorkflowAliasSearchAttributeTestSuite) createWorkflow(ctx context.Context, workflowFn WorkflowFunction, prefix string, scheduleByID string) sdkclient.WorkflowRun {
```
bergundy: Just call the parameter `customSA` because it is not a scheduledByID search attribute. I admit to being confused what it is that you are indexing and querying here.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,222 @@
...
+	s := new(WorkflowAliasSearchAttributeTestSuite)
+	suite.Run(t, s)
+}
+
+func (ts *WorkflowAliasSearchAttributeTestSuite) SetupTest() {
+	ts.FunctionalTestBase.SetupTest()
+}
+
+func (ts *WorkflowAliasSearchAttributeTestSuite) WorkflowFunc(ctx workflow.Context) error {
+	err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
```
bergundy: You don't need an activity in the workflow, it just adds load in CI.
Just make it an empty workflow that immediately returns.

tests/workflow_alias_search_attribute_test.go
bergundy: I typically avoid creating these testsuite level helpers because they cannot be safely used in subtests.
`ts.NoError` is unsafe to use within a subtest or an `EventuallyWithT` block because it is not associated with the correct instance of `testing.T`.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,222 @@
...
+	ts.NoError(err)
+	ts.NotNil(getWFResponse)
+	ts.Equal(workflowRun1.GetID(), getWFResponse.GetWorkflowExecutionInfo().Execution.WorkflowId)
+	ts.Equal(workflowRun1.GetRunID(), getWFResponse.GetWorkflowExecutionInfo().Execution.RunId)
+
+	getWFResponse, err = ts.SdkClient().DescribeWorkflowExecution(ctx, workflowRun2.GetID(), workflowRun2.GetRunID())
+	ts.NoError(err)
+	ts.NotNil(getWFResponse)
+	ts.Equal(workflowRun2.GetID(), getWFResponse.GetWorkflowExecutionInfo().Execution.WorkflowId)
+	ts.Equal(workflowRun2.GetRunID(), getWFResponse.GetWorkflowExecutionInfo().Execution.RunId)
```
bergundy: Redundant, don't add code that doesn't add value to the test. Tests are already verbose (and should be so), we don't want to make them harder to follow than they need to be.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Use Eventually pattern to wait for visibility store to be updated
+	ts.Eventually(
+		func() bool {
+			resp, err := ts.SdkClient().ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+				Namespace: ts.Namespace().String(),
+				Query:     "WorkflowId = '" + workflowRun1.GetID() + "' OR WorkflowId = '" + workflowRun2.GetID() + "'",
+			})
+			ts.NoError(err)
+			ts.NotNil(resp)
```
bergundy: This will cause the test to panic, you need to use `EventuallyWithT` if you want to add assertions, and you'll need to use the right `T` for those assertions, not the one in the suite.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Namespace: ts.Namespace().String(),
+				Query:     "ScheduledById IS NULL",
+			})
+			ts.NoError(err)
+			ts.NotNil(resp)
+			return len(resp.GetExecutions()) == 2
+		},
+		testcore.WaitForESToSettle,
+		100*time.Millisecond,
+	)
```
bergundy: Not really understanding the intention of this test.
The unit tests are very clear, I would copy that pattern here.

common/persistence/visibility/store/elasticsearch/query_interceptors.go
spkane31: [Confirmed here](https://github.com/temporalio/temporal/blob/main/common/searchattribute/name_type_map.go#L65-L87)

common/persistence/visibility/store/elasticsearch/query_interceptors.go
spkane31: Moved to single implementation

common/persistence/visibility/store/sql/query_converter_test.go
```diff
@@ -634,7 +634,7 @@ func (s *queryConverterSuite) TestConvertColName() {
 	}
 
 	for _, tc := range tests {
-		s.Run(tc.name, func() {
+		s.T().Run(tc.name, func(t *testing.T) {
```
rodrigozhou: Why this change?

common/searchattribute/defs.go
```diff
@@ -238,7 +238,8 @@ func IsReserved(name string) bool {
 	return strings.HasPrefix(name, ReservedPrefix)
 }
 
-// IsMappable returns true if name can have be mapped tho the alias.
+// IsMappable returns true if name can be mapped to the alias.
+// Mappable search attributes are those that are defined by the user.
```
rodrigozhou:
```suggestion
// Mappable search attributes are those that can be defined by the user.
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+	name string,
+	ns namespace.Name,
+	mapper searchattribute.Mapper,
+	saTypeMap searchattribute.NameTypeMap,
+) (string, enumspb.IndexedValueType, error) {
+	if strings.HasPrefix(name, "Temporal") {
+		if name == "TemporalBuildIds" {
+			saType, err := saTypeMap.GetType(searchattribute.BuildIds)
+			if err != nil {
+				return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", searchattribute.BuildIds)
```
rodrigozhou: This error goes to the user, so if the user typed "TemporalBuildIds", it's better for the user to see "TemporalBuildIds" in their error.
```suggestion
				return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", name)
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+	"go.temporal.io/server/common/searchattribute"
+)
+
+func ResolveSearchAttributeAlias(
+	name string,
+	ns namespace.Name,
+	mapper searchattribute.Mapper,
+	saTypeMap searchattribute.NameTypeMap,
+) (string, enumspb.IndexedValueType, error) {
+	if strings.HasPrefix(name, "Temporal") {
```
rodrigozhou:
```suggestion
	if strings.HasPrefix(name, searchattribute.ReservedPrefix) {
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+		if name == "TemporalBuildIds" {
```
rodrigozhou: Why do you need to check for `BuildIds`? What about other reserved search attributes without `Temporal` prefix?

common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -50,35 +50,10 @@ func NewValuesInterceptor(
...
-				if !isCustom {
-					// ScheduleId is a fake SA -- convert to WorkflowId
-					fieldName = searchattribute.WorkflowID
-				}
-			}
-		}
-	}
-
-	fieldType, err := ni.searchAttributesTypeMap.GetType(fieldName)
+	mapper, _ := ni.searchAttributesMapperProvider.GetMapper(ni.namespace)
```
rodrigozhou: Avoid suppressing errors.
```suggestion
	mapper, err := ni.searchAttributesMapperProvider.GetMapper(ni.namespace)
	if err != nil {
		return "", err
	}
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+		return resolveResult{}, false
+	}
+
+	fieldName, err := mapper.GetFieldName(name, ns.String())
+	if err == nil {
+		fieldType, err := saTypeMap.GetType(fieldName)
+		if err == nil {
+			return resolveResult{fieldName: fieldName, fieldType: fieldType}, true
+		}
+		return resolveResult{}, false
```
rodrigozhou: Can you add a comment here why not returning an error?

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	var internalErr serviceerror.Internal
+	if errors.Is(err, &internalErr) {
+		return resolveResult{err: err}, true
+	}
```
rodrigozhou: When does it return an internal error?

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+		saType, err := saTypeMap.GetType(searchattribute.WorkflowID)
+		if err == nil {
+			return resolveResult{fieldName: searchattribute.WorkflowID, fieldType: saType}, true
+		}
+	}
+
+	// If the mapper returns an InvalidArgument error for a mappable field,
+	// and the error message is "mapper error", then the mapper is authoritative
+	// for this namespace and we should propagate the error instead of falling back
+	if errors.As(err, &invalidArgument) && searchattribute.IsMappable(name) && invalidArgument.Message == "mapper error" {
```
rodrigozhou: It seems that you only call this function if `searchattribute.IsMappable(name)` returns true, so it's redundant here.

rodrigozhou: I'd avoid hard-coded `"mapper error"`. Also, where does it even return this string? I couldn't find it.

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+		return resolveResult{err: err}, true
```
rodrigozhou: Why is it returning `true`?

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return resolveResult{}, false
```
rodrigozhou: It's much more common to handle the errors inside the `if` block. What about writing like this?

```suggestion
	fieldName, err := mapper.GetFieldName(name, ns.String())
	if err != nil {
		var internalErr serviceerror.Internal
		if errors.Is(err, &internalErr) {
			return resolveResult{err: err}, false
		}
		
		var invalidArgument *serviceerror.InvalidArgument
		if errors.As(err, &invalidArgument)
			if name == searchattribute.ScheduleID {
				fieldName = searchattribute.WorkflowID
			} else {
				return resolveResult{err: err}, false
			}
		}
	}

	fieldType, err := saTypeMap.GetType(fieldName)
	if err != nil {
		return resolveResult{err: err}, false
	}
	return resolveResult{fieldName: fieldName, fieldType: fieldType}, true
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func tryDirectAndPrefixedLookup(name string, saTypeMap searchattribute.NameTypeMap) (resolveResult, bool) {
+	if saType, err := saTypeMap.GetType(name); err == nil {
+		return resolveResult{fieldName: name, fieldType: saType}, true
+	}
+
+	prefixedName := fmt.Sprintf("Temporal%s", name)
```
rodrigozhou:
```suggestion
	prefixedName := searchattribute.ReservedPrefix + name
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+		return result.fieldName, result.fieldType, result.err
+	}
+
+	if name == searchattribute.ScheduleID {
+		saType, err := saTypeMap.GetType(searchattribute.WorkflowID)
+		if err != nil {
+			return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", searchattribute.ScheduleID)
+		}
+		return searchattribute.WorkflowID, saType, nil
+	}
```
rodrigozhou: Isn't this already handled inside `tryVisibilityMapper`? Why do you need it here again?

common/persistence/visibility/store/elasticsearch/query_interceptors.go
spkane31: I don't think this is the right way to handle this, we only want to return the error if `searchattribute.IsMappable` is true, which is handled inside the `ResolveSearchAttributeAlias`

common/persistence/visibility/store/query/resolve.go
spkane31: @bergundy brought up this one. This would be new behavior so do we want to make the same change for all reserved SAs w/out `Temporal`?

common/persistence/visibility/store/sql/query_converter_test.go
spkane31: This ensures that the test is using the correct `t *testing.T`, making any tests failures correctly populate which test fails

common/persistence/visibility/store/query/resolve.go
spkane31: This was from the test mapper, updating the test mapper to align with the prod mapper

common/persistence/visibility/store/query/resolve.go
spkane31: This was changed so not needed anymore

common/persistence/visibility/store/sql/query_converter_test.go
rodrigozhou: `s.Run` should handle it internally, you shouldn't need to call `s.T().Run` explicitly.

common/persistence/visibility/store/elasticsearch/query_interceptors.go
rodrigozhou: This is getting the mapper. No matter if the `name` is mappable or not, this should never return an error unless there's an issue.

common/persistence/visibility/store/query/resolve.go
spkane31: Fixed

common/persistence/visibility/store/sql/query_converter_test.go
spkane31: Reverted

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,131 @@
...
+	fieldName, err := mapper.GetFieldName(name, ns.String())
+	if err != nil {
+		var internalErr serviceerror.Internal
+		if errors.Is(err, &internalErr) {
+			// Internal errors indicate system-level issues that should always be propagated
+			// and not fall back to direct lookup. These could come from the MapperProvider.GetMapper()
+			// call (e.g., namespace registry failures, persistence layer issues) rather than
+			// the mapper methods themselves.
+			return resolveResult{err: err}, true
+		}
```
rodrigozhou: Where do we return an internal error in `GetFieldName`?

spkane31: Removed, apologies thought I had done that in a previous commit.

common/persistence/visibility/store/query/resolve.go
rodrigozhou: I think we should either allow everything or not allow at all.
I don't see why `BuildIds` should be a special case.

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,122 @@
...
+			if err != nil {
+				return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", name)
+			}
+			return searchattribute.BuildIds, saType, nil
+		}
+		saType, err := saTypeMap.GetType(name)
+		if err != nil {
+			return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", name)
+		}
+		return name, saType, nil
```
rodrigozhou: This is cleaner.

```suggestion
		fieldName := name
		if name[len(searchattribute.ReservedPrefix):] == searchattribute.BuildIds {
			fieldName = searchattribute.BuildIds
		}
		saType, err := saTypeMap.GetType(fieldName)
		if err != nil {
			return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", name)
		}
		return fieldName, saType, nil
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,119 @@
...
+	return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, NewConverterError("invalid search attribute: %s", name)
+}
+
+type resolveResult struct {
+	fieldName string
+	fieldType enumspb.IndexedValueType
+	err       error
+}
+
+func tryVisibilityMapper(name string, ns namespace.Name, mapper searchattribute.Mapper, saTypeMap searchattribute.NameTypeMap) (resolveResult, bool) {
```
bergundy: Put the error as the second return type as is common in Go. The boolean can go into the `resolveResult` struct.

bergundy: Or just return `NewConverterError("invalid search attribute: %s", name)` in the error and drop the bool entirely.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,118 @@
...
+	if customSA != "" {
+		customSAKey := temporal.NewSearchAttributeKeyKeyword("CustomSA")
+		workflowOptions.TypedSearchAttributes = temporal.NewSearchAttributes(
+			customSAKey.ValueSet(customSA),
+		)
+	}
+	return s.SdkClient().ExecuteWorkflow(ctx, workflowOptions, workflowFn)
+}
+
+func (s *WorkflowAliasSearchAttributeTestSuite) TestWorkflowAliasSearchAttribute() {
```
bergundy, line 73: What are you testing here?
You should test that if a workflow has a `TemporalFoo` search attribute you can query it with `Foo` and that if `Foo` is a custom search attribute, the query overrides it.

bergundy: You can use one of the versioning search attributes for this test. Make sure the scenario you create generates these attributes.

spkane31: Added

common/persistence/visibility/store/query/resolve.go
spkane31: simplified

tests/workflow_alias_search_attribute_test.go
bergundy: What did you add?

spkane31: 🤦 I removed my changes when I reverted previous commit...

spkane31: adding it back in a few moments

spkane31: Updated to create workflows with visibility attributes, then using those to query both `TemporalWorkflowVersioningBehavior` and `WorkflowVersioningBehavior` ensuring they return the same thing. Added a second test to create custom SA `WorkflowVersioningBehavior` and ensure that comes back as expected.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,190 @@
...
+	s.FunctionalTestBase.SetupTest()
+
+	s.Worker().RegisterWorkflow(s.workflowFunc)
+}
+
+func (s *WorkflowAliasSearchAttributeTestSuite) workflowFunc(ctx workflow.Context) (string, error) {
+	return "done!", nil
+}
+
+// func (s *WorkflowAliasSearchAttributeTestSuite) createWorkflow(
```
bergundy: Please remove commented out code before merging.

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,190 @@
...
+func (s *WorkflowAliasSearchAttributeTestSuite) TestWorkflowAliasSearchAttribute() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	tv := testvars.New(s.T())
+
+	_, err := s.createWorkflow(ctx, "wf_id_1", tv, nil)
+	require.NoError(s.T(), err)
+	_, err = s.createWorkflow(ctx, "wf_id_2", tv, nil)
+	require.NoError(s.T(), err)
```
bergundy: Why do you need two workflows?

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,190 @@
...
+		sa1       string
+		sa2       string
+		predicate string
+	}{
+		{sa1: searchattribute.BuildIds, sa2: "TemporalBuildIds", predicate: "IS NOT NULL"},
+		{sa1: searchattribute.TemporalWorkflowVersioningBehavior, sa2: "WorkflowVersioningBehavior", predicate: "= 'Pinned'"},
+		{sa1: searchattribute.TemporalWorkerDeploymentVersion, sa2: "WorkerDeploymentVersion", predicate: "IS NOT NULL"},
+		{sa1: searchattribute.TemporalWorkerDeployment, sa2: "WorkerDeployment", predicate: "IS NOT NULL"},
+		{sa1: searchattribute.ScheduleID, sa2: searchattribute.WorkflowID, predicate: "IS NOT NULL"},
+	}
```
bergundy: Instead of all of this, can you simply test that both `TemporalWorkflowVersioningBehavior = 'Pinned'` and `WorkflowVersioningBehavior = 'Pinned'` return your execution?

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (s *WorkflowAliasSearchAttributeTestSuite) TestWorkflowAliasSearchAttribute() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	tv := testvars.New(s.T())
+
+	_, err := s.createWorkflow(ctx, "wf_id_1", tv, nil)
+	require.NoError(s.T(), err)
```
bergundy: Would you terminate this workflow just to have clean slate after the test is run?

tests/workflow_alias_search_attribute_test.go
```diff
@@ -0,0 +1,190 @@
...
+	sa := &commonpb.SearchAttributes{
+		IndexedFields: map[string]*commonpb.Payload{
+			"WorkflowVersioningBehavior": payload.EncodeString("user-defined"),
+		},
+	}
+
+	_, err = s.createWorkflow(ctx, "wf_id_1", tv, sa)
+	require.NoError(s.T(), err)
+	_, err = s.createWorkflow(ctx, "wf_id_2", tv, sa)
+	require.NoError(s.T(), err)
```
bergundy: Redundant.

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,122 @@
...
+			if strings.Contains(errMsg, "invalid alias") || // From GetFieldName() when alias doesn't exist in mapper
+				strings.Contains(errMsg, "invalid field") || // From GetAlias() when field name doesn't exist in mapper
+				strings.Contains(errMsg, "no mapping defined") { // From test mappers and some implementations when field is unknown
+				// Mapper doesn't handle this field, allow fallback to direct/prefixed lookup
+				return resolveResult{}, nil
+			}
+			// Other InvalidArgument errors suggest mapper tried to handle the field but failed
+			// due to validation issues, malformed data, etc. These should be propagated.
+			return resolveResult{}, err
+		}
```
bergundy: Seems like a hack. Was this ported from a previous implementation? @rodrigozhou can you take a look please?

spkane31: Here are my thoughts: I agree this is hacky, but I don't think the `InvalidArgument` specifies between "mapper doesn't handle this field" vs "mapper had a validation error". We can add a backlog item to do one of two things:
1. Add structured information to the `GetAlias` and `GetFieldName` to return `(string, bool, error)` w/ the bool indicating whether the mapper handles that field.
2. Add an `ErrFieldNotMapped` error that can be returned from the mapper and understood by the `tryVisibilityMapper` function

Thoughts @rodrigozhou

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if searchattribute.IsMappable(name) {
+		// First check if the visibility mapper can handle this field (e.g., custom search attributes)
+		if result, err := tryVisibilityMapper(name, ns, mapper, saTypeMap); err == nil && result.fieldName != "" {
+			return result.fieldName, result.fieldType, err
+		} else if err != nil {
+			// If there was an error from the mapper, return it
+			return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, err
+		}
```
rodrigozhou: nit
```suggestion
		if result, err := tryVisibilityMapper(name, ns, mapper, saTypeMap); err != nil {
			// If there was an error from the mapper, return it
			return "", enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, err
		} else if result.fieldName != "" {
			return result.fieldName, result.fieldType, nil
		}
```

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,122 @@
...
+		// This fallback only applies when the visibility mapper doesn't handle the field
+		if name == searchattribute.ScheduleID {
+			// First check if ScheduleID exists as a custom search attribute
+			if _, err := saTypeMap.GetType(searchattribute.ScheduleID); err != nil {
+				// ScheduleID is not defined, transform to WorkflowID if WorkflowID exists
+				saType, err := saTypeMap.GetType(searchattribute.WorkflowID)
+				if err == nil {
+					return searchattribute.WorkflowID, saType, nil
+				}
+				// If WorkflowID is not in type map, continue with normal flow
```
rodrigozhou: `WorkflowID` is always defined
```suggestion
				// ScheduleID is not defined, transform to WorkflowID
				saType, _ := saTypeMap.GetType(searchattribute.WorkflowID)
				return searchattribute.WorkflowID, saType, nil
```

common/persistence/visibility/store/query/resolve.go
rodrigozhou: I don't like this. I don't see anywhere in the code that return `"invalid alias"` for example. Neither of the existing implementation do this, so I'm not sure why this is needed.

rodrigozhou:
@spkane31 As I'm working in the new unified query converter, I gave a shot implementing this alias resolution [here](https://github.com/temporalio/temporal/blob/9c821f6d6ff956c0a5d1919af03af082987d9dee/common/persistence/visibility/store/query/converter.go#L476-L523), but I'm not sure if it's exactly the same behavior as you want in this PR.
I've added unit tests [here](https://github.com/temporalio/temporal/blob/9c821f6d6ff956c0a5d1919af03af082987d9dee/common/persistence/visibility/store/query/converter_test.go#L1907-L2069), although I'm not sure if it covers everything.

common/persistence/visibility/store/query/resolve.go
```diff
@@ -0,0 +1,97 @@
...
+		fieldName, fieldType := tryVisibilityMapper(name, ns, mapper, saTypeMap)
+		if fieldName != "" {
+			return fieldName, fieldType, nil
+		}
+
+		// Handle ScheduleID → WorkflowID transformation, but only if ScheduleID is not defined as a custom search attribute
+		// This fallback only applies when the visibility mapper doesn't handle the field
+		if name == searchattribute.ScheduleID {
+			// First check if ScheduleID exists as a custom search attribute
+			if _, err := saTypeMap.GetType(searchattribute.ScheduleID); err != nil {
```
rodrigozhou: This checks seems unnecessary. If it were a custom search attribute, `tryVisibilityMapper` would returned the correct value.

rodrigozhou (approved):

PR #8371, opened 2025-09-23
Add versioning behavior tag to activity metrics

ShahabT:
## What changed?
Adding new `versioning_behavior` tag to the newly added activity metrics so we can compare activity failure and timeout rates between users using versioning vs the ones who don't.

## Why?
above.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None. new tag has only 3 values (low cardinality)


Shivs11:
thank you Shahab!

PR #8372, opened 2025-09-23
[Nexus] Support internal callback completion routing to CHASM

lina-temporal:
## What changed?
* The Callbacks component now detects Nexus URLs of the form `temporal://internal/chasm`, which are routed to CHASM components through a new `ChasmInvocation`.
  * CHASM routing information is serialized into a header, the URL is left opaque. 
* A new CHASM interface has been added, `chasm.NexusCompletionHandler`, which components can implement to receive Nexus callbacks.
* A CHASM helper method, `GetNexusCallback`, has been added that returns a `Callback` protobuf routing to the given component.
* The frontend's callback argument validator has been patched to allow "temporal" schemes unconditionally. Scheduler starts workflows through the frontend, in order to be counted by metering, and so frontend must allow requests with internal-facing callbacks.
   * I believe @chaptersix 's changes also touch this area, we may not need my interim patch there.
 
## Why?
* CHASM Scheduler needs to find out about workflow completions, and we want to do that with Nexus completion callbacks. 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
* The funkiness is in the frontend handler - see the TODO. I don't think it presents a security risk, since a customer would only be able to forge completions within their own namespace (since this validation takes place within `StartWorkflowExecution`'s frontend flow). But it would be nice to validate that it was only coming from internal pods.

### DIFF
--- a/api/persistence/v1/executions.pb.go
+++ b/api/persistence/v1/executions.pb.go
@@ -1517,6 +1517,7 @@ type ReplicationTaskInfo struct {
 	LastVersionHistoryItem *v13.VersionHistoryItem `protobuf:"bytes,21,opt,name=last_version_history_item,json=lastVersionHistoryItem,proto3" json:"last_version_history_item,omitempty"`
 	IsFirstTask            bool                    `protobuf:"varint,22,opt,name=is_first_task,json=isFirstTask,proto3" json:"is_first_task,omitempty"`
 	TargetClusters         []string                `protobuf:"bytes,23,rep,name=target_clusters,json=targetClusters,proto3" json:"target_clusters,omitempty"`
+	IsForceReplication     bool                    `protobuf:"varint,24,opt,name=is_force_replication,json=isForceReplication,proto3" json:"is_force_replication,omitempty"`
 	unknownFields          protoimpl.UnknownFields
 	sizeCache              protoimpl.SizeCache
 }
@@ -1682,6 +1683,13 @@ func (x *ReplicationTaskInfo) GetTargetClusters() []string {
 		return x.TargetClusters
 	}
 	return nil
+}
+
+func (x *ReplicationTaskInfo) GetIsForceReplication() bool {
+	if x != nil {
+		return x.IsForceReplication
+	}
+	return false
 }
 
 // visibility_task_data column
@@ -4688,7 +4696,7 @@ const file_temporal_server_api_persistence_v1_executions_proto_rawDesc = "" +
 	"\x05stamp\x18\x11 \x01(\x05R\x05stamp\x1a\\\n" +
 	"\x19CloseExecutionTaskDetails\x12?\n" +
 	"\x1ccan_skip_visibility_archival\x18\x01 \x01(\bR\x19canSkipVisibilityArchivalB\x0e\n" +
-	"\ftask_detailsJ\x04\b\x0e\x10\x0f\"\x83\b\n" +
+	"\ftask_detailsJ\x04\b\x0e\x10\x0f\"\xb5\b\n" +
 	"\x13ReplicationTaskInfo\x12!\n" +
 	"\fnamespace_id\x18\x01 \x01(\tR\vnamespaceId\x12\x1f\n" +
 	"\vworkflow_id\x18\x02 \x01(\tR\n" +
@@ -4710,7 +4718,8 @@ const file_temporal_server_api_persistence_v1_executions_proto_rawDesc = "" +
 	"\x10task_equivalents\x18\x14 \x03(\v27.temporal.server.api.persistence.v1.ReplicationTaskInfoR\x0ftaskEquivalents\x12m\n" +
 	"\x19last_version_history_item\x18\x15 \x01(\v22.temporal.server.api.history.v1.VersionHistoryItemR\x16lastVersionHistoryItem\x12\"\n" +
 	"\ris_first_task\x18\x16 \x01(\bR\visFirstTask\x12'\n" +
-	"\x0ftarget_clusters\x18\x17 \x03(\tR\x0etargetClustersJ\x04\b\t\x10\n" +
+	"\x0ftarget_clusters\x18\x17 \x03(\tR\x0etargetClusters\x120\n" +
+	"\x14is_force_replication\x18\x18 \x01(\bR\x12isForceReplicationJ\x04\b\t\x10\n" +
 	"J\x04\b\n" +
 	"\x10\vJ\x04\b\f\x10\rJ\x04\b\x0e\x10\x0f\"\x99\x04\n" +
 	"\x12VisibilityTaskInfo\x12!\n" +
--- a/api/replication/v1/message.pb.go
+++ b/api/replication/v1/message.pb.go
@@ -1899,6 +1899,7 @@ type VersionedTransitionArtifact struct {
 	NewRunInfo               *NewRunInfo                                   `protobuf:"bytes,4,opt,name=new_run_info,json=newRunInfo,proto3" json:"new_run_info,omitempty"`
 	IsFirstSync              bool                                          `protobuf:"varint,5,opt,name=is_first_sync,json=isFirstSync,proto3" json:"is_first_sync,omitempty"`
 	IsCloseTransferTaskAcked bool                                          `protobuf:"varint,6,opt,name=is_close_transfer_task_acked,json=isCloseTransferTaskAcked,proto3" json:"is_close_transfer_task_acked,omitempty"`
+	IsForceReplication       bool                                          `protobuf:"varint,7,opt,name=is_force_replication,json=isForceReplication,proto3" json:"is_force_replication,omitempty"`
 	unknownFields            protoimpl.UnknownFields
 	sizeCache                protoimpl.SizeCache
 }
@@ -1982,6 +1983,13 @@ func (x *VersionedTransitionArtifact) GetIsFirstSync() bool {
 func (x *VersionedTransitionArtifact) GetIsCloseTransferTaskAcked() bool {
 	if x != nil {
 		return x.IsCloseTransferTaskAcked
+	}
+	return false
+}
+
+func (x *VersionedTransitionArtifact) GetIsForceReplication() bool {
+	if x != nil {
+		return x.IsForceReplication
 	}
 	return false
 }
@@ -2172,7 +2180,7 @@ const file_temporal_server_api_replication_v1_message_proto_rawDesc = "" +
 	"\fnamespace_id\x18\x06 \x01(\tR\vnamespaceId\x12\x1f\n" +
 	"\vworkflow_id\x18\a \x01(\tR\n" +
 	"workflowId\x12\x15\n" +
-	"\x06run_id\x18\b \x01(\tR\x05runIdJ\x04\b\x01\x10\x02J\x04\b\x02\x10\x03J\x04\b\x03\x10\x04J\x04\b\x04\x10\x05\"\xf2\x04\n" +
+	"\x06run_id\x18\b \x01(\tR\x05runIdJ\x04\b\x01\x10\x02J\x04\b\x02\x10\x03J\x04\b\x03\x10\x04J\x04\b\x04\x10\x05\"\xa4\x05\n" +
 	"\x1bVersionedTransitionArtifact\x12\x9f\x01\n" +
 	"'sync_workflow_state_mutation_attributes\x18\x01 \x01(\v2G.temporal.server.api.replication.v1.SyncWorkflowStateMutationAttributesH\x00R#syncWorkflowStateMutationAttributes\x12\x9f\x01\n" +
 	"'sync_workflow_state_snapshot_attributes\x18\x02 \x01(\v2G.temporal.server.api.replication.v1.SyncWorkflowStateSnapshotAttributesH\x00R#syncWorkflowStateSnapshotAttributes\x12E\n" +
@@ -2180,7 +2188,8 @@ const file_temporal_server_api_replication_v1_message_proto_rawDesc = "" +
 	"\fnew_run_info\x18\x04 \x01(\v2..temporal.server.api.replication.v1.NewRunInfoR\n" +
 	"newRunInfo\x12\"\n" +
 	"\ris_first_sync\x18\x05 \x01(\bR\visFirstSync\x12>\n" +
-	"\x1cis_close_transfer_task_acked\x18\x06 \x01(\bR\x18isCloseTransferTaskAckedB\x12\n" +
+	"\x1cis_close_transfer_task_acked\x18\x06 \x01(\bR\x18isCloseTransferTaskAcked\x120\n" +
+	"\x14is_force_replication\x18\a \x01(\bR\x12isForceReplicationB\x12\n" +
 	"\x10state_attributesB5Z3go.temporal.io/server/api/replication/v1;repicationb\x06proto3"
 
 var (
--- a/common/persistence/serialization/task_serializer.go
+++ b/common/persistence/serialization/task_serializer.go
@@ -1386,6 +1386,7 @@ func (s *TaskSerializer) replicationSyncVersionedTransitionTaskToProto(
 		LastVersionHistoryItem: syncVersionedTransitionTask.LastVersionHistoryItem,
 		IsFirstTask:            syncVersionedTransitionTask.IsFirstTask,
 		TargetClusters:         syncVersionedTransitionTask.TargetClusters,
+		IsForceReplication:     syncVersionedTransitionTask.IsForceReplication,
 		TaskEquivalents:        taskInfoEquivalents,
 	}, nil
 }
@@ -1424,6 +1425,7 @@ func (s *TaskSerializer) replicationSyncVersionedTransitionTaskFromProto(
 		TaskEquivalents:        taskEquivalents,
 		IsFirstTask:            syncVersionedTransitionTask.IsFirstTask,
 		TargetClusters:         syncVersionedTransitionTask.TargetClusters,
+		IsForceReplication:     syncVersionedTransitionTask.IsForceReplication,
 	}, nil
 }
 
--- a/proto/internal/temporal/server/api/persistence/v1/executions.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/executions.proto
@@ -373,6 +373,7 @@ message ReplicationTaskInfo {
     history.v1.VersionHistoryItem last_version_history_item = 21;
     bool is_first_task = 22;
     repeated string target_clusters= 23;
+    bool is_force_replication = 24;
 }
 
 // visibility_task_data column
--- a/proto/internal/temporal/server/api/replication/v1/message.proto
+++ b/proto/internal/temporal/server/api/replication/v1/message.proto
@@ -248,4 +248,5 @@ message VersionedTransitionArtifact {
     NewRunInfo new_run_info = 4;
     bool is_first_sync = 5;
     bool is_close_transfer_task_acked = 6;
+    bool is_force_replication = 7;
 }
--- a/service/frontend/fx.go
+++ b/service/frontend/fx.go
@@ -211,6 +211,7 @@ func GrpcServerOptionsProvider(
 	maskInternalErrorDetailsInterceptor *interceptor.MaskInternalErrorDetailsInterceptor,
 	slowRequestLoggerInterceptor *interceptor.SlowRequestLoggerInterceptor,
 	customInterceptors []grpc.UnaryServerInterceptor,
+	customStreamInterceptors []grpc.StreamServerInterceptor,
 	metricsHandler metrics.Handler,
 ) GrpcServerOptions {
 	kep := keepalive.EnforcementPolicy{
@@ -272,6 +273,9 @@ func GrpcServerOptionsProvider(
 
 	streamInterceptor := []grpc.StreamServerInterceptor{
 		telemetryInterceptor.StreamIntercept,
+	}
+	if len(customStreamInterceptors) > 0 {
+		streamInterceptor = append(streamInterceptor, customStreamInterceptors...)
 	}
 
 	grpcServerOptions = append(
--- a/service/fx.go
+++ b/service/fx.go
@@ -38,14 +38,15 @@ type (
 	GrpcServerOptionsParams struct {
 		fx.In
 
-		Logger                 log.Logger
-		RpcFactory             common.RPCFactory
-		RetryableInterceptor   *interceptor.RetryableInterceptor
-		TelemetryInterceptor   *interceptor.TelemetryInterceptor
-		RateLimitInterceptor   *interceptor.RateLimitInterceptor
-		TracingStatsHandler    telemetry.ServerStatsHandler
-		MetricsStatsHandler    metrics.ServerStatsHandler
-		AdditionalInterceptors []grpc.UnaryServerInterceptor `optional:"true"`
+		Logger                       log.Logger
+		RPCFactory                   common.RPCFactory
+		RetryableInterceptor         *interceptor.RetryableInterceptor
+		TelemetryInterceptor         *interceptor.TelemetryInterceptor
+		RateLimitInterceptor         *interceptor.RateLimitInterceptor
+		TracingStatsHandler          telemetry.ServerStatsHandler
+		MetricsStatsHandler          metrics.ServerStatsHandler
+		AdditionalInterceptors       []grpc.UnaryServerInterceptor  `optional:"true"`
+		AdditionalStreamInterceptors []grpc.StreamServerInterceptor `optional:"true"`
 	}
 )
 
@@ -121,7 +122,7 @@ func GrpcServerOptionsProvider(
 	params GrpcServerOptionsParams,
 ) []grpc.ServerOption {
 
-	grpcServerOptions, err := params.RpcFactory.GetInternodeGRPCServerOptions()
+	grpcServerOptions, err := params.RPCFactory.GetInternodeGRPCServerOptions()
 	if err != nil {
 		params.Logger.Fatal("creating gRPC server options failed", tag.Error(err))
 	}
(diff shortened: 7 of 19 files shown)


chasm/nexus_completion.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/nexus-rpc/sdk-go/nexus"
+	commonpb "go.temporal.io/api/common/v1"
+)
+
+const (
+	// Header name for the CHASM ComponentRef.
+	NexusComponentRefHeader = "X-CHASM-Component-Ref"
+
+	// Base URL for Nexus->CHASM callbacks.
+	NexusCompletionHandlerURL = "temporal://internal/chasm"
```
bergundy:
```suggestion
	NexusCompletionHandlerURL = "temporal://internal"
```

chasm/nexus_completion.go
```diff
@@ -0,0 +1,51 @@
...
+	"encoding/base64"
+	"errors"
+
+	"github.com/nexus-rpc/sdk-go/nexus"
+	commonpb "go.temporal.io/api/common/v1"
+)
+
+const (
+	// Header name for the CHASM ComponentRef.
+	NexusComponentRefHeader = "X-CHASM-Component-Ref"
```
bergundy: No need, just put this in the callback token header.

components/callbacks/executors.go
```diff
@@ -144,17 +147,30 @@ func (e taskExecutor) loadInvocationArgs(
...
-			nexusInvokable.completion, err = target.GetNexusCompletion(ctx, callback.GetRequestId())
+			nexusInvokable.completion = completion
 			nexusInvokable.workflowID = ref.WorkflowKey.WorkflowID
 			nexusInvokable.runID = ref.WorkflowKey.RunID
 			nexusInvokable.attempt = callback.Attempt
 			invokable = nexusInvokable
-			if err != nil {
-				return err
-			}
 		case *persistencespb.Callback_Hsm:
```
bergundy, line 175: Can just remove this branch and the rest of the code it references.

components/callbacks/executors.go
```diff
@@ -144,17 +147,30 @@ func (e taskExecutor) loadInvocationArgs(
...
+				return err
+			}
+
+			// CHASM internal callbacks make use of Nexus as their callback delivery
+			// mechanism, but with a special prefix.
+			if strings.HasPrefix(variant.Nexus.Url, chasm.NexusCompletionHandlerURL) {
+				chasmInvokable := chasmInvocation{}
+				chasmInvokable.nexus = variant.Nexus
+				chasmInvokable.attempt = callback.Attempt
+				invokable = chasmInvokable
```
bergundy: You're going to want to set the completion on here too...

Would you also please just set all of the fields here inline with the struct definition?
Could just do:
```
invokable = chasmInvocation{
  nexus: variant.Nexus,
  ...
}
```

components/callbacks/executors.go
```diff
@@ -144,17 +147,30 @@ func (e taskExecutor) loadInvocationArgs(
...
+				break
+			}
+
 			// variant struct is immutable and ok to reference without copying
 			nexusInvokable := nexusInvocation{}
```
bergundy: Same here, would be nice to just inline the whole field assignment (unrelated to your change).

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,73 @@
...
+	ref, err := chasm.DeserializeComponentRef(decodedRef)
+	if err != nil {
+		return invocationResultFail{fmt.Errorf("failed to deserialize CHASM ComponentRef: %v", err)}
+	}
+
+	// Attempt to access the component and call its invocation method. We execute
+	// this similarly as we would a pure task (holding an exclusive lock), as the
+	// assumption is that the accessed component will be recording (or generating a
+	// task) based on this result.
+	_, err = e.ChasmEngine.UpdateComponent(ctx, ref, func(ctx chasm.MutableContext, component chasm.Component) error {
```
bergundy: This doesn't do an RPC, you're going to have to add an RPC to the shard that owns the execution of the original caller.
We can then optimize it so if the ref is something that we can resolve locally we would bypass the RPC call.

service/frontend/workflow_handler.go
```diff
@@ -5245,6 +5245,11 @@ func (wh *WorkflowHandler) validateCallbackURL(ns namespace.Name, rawURL string)
 	if err != nil {
 		return err
 	}
+	// Internally routed.
+	// TODO - validate requests involving this URL come from a history pod.
+	if u.Scheme == "temporal" {
```
bergundy: @chaptersix will take care of this, he's modifying this code right now.

bergundy: We need to validate that the caller is from within the cluster if it tries to attach a callback with an internal URL.

components/callbacks/chasm_invocation.go
bergundy: The RPC would be `historyservice.CompleteNexusOperation`, you'll want to update that method to invoke `ChasmEngine.UpdateComponent` and take a different form of completion.

Feel free to rename the `completion` field to `hsm_completion` to disambiguate between HSM and CHASM refs.

bergundy: Let's also not change the Nexus RPC stuff for now since we are not using that SDK for delivering the completions.

We can change it later when the SDK supports pluggable transports if we want to but that's probably not critical.

bergundy (changes requested):

chasm/nexus_completion.go
lina-temporal: When I synced up with @chaptersix , we decided to use `SystemCallbackURL` so it goes through the same handler as other internal callbacks, which I think makes sense. However, if we don't have a unique URL nor a unique CHASM header, how would we distinguish this as a CHASM request in the callback executor?

components/callbacks/executors.go
lina-temporal: Sure, I'll open a separate PR to remove the Callback_Hsm invocation stuff.

components/callbacks/executors.go
lina-temporal: Oops, yeah, cleaned these up.

chasm/nexus_completion.go
```diff
@@ -0,0 +1,51 @@
...
+	commonpb "go.temporal.io/api/common/v1"
+	tokenspb "go.temporal.io/server/api/token/v1"
+)
+
+const (
+	// Header name for the CHASM ComponentRef.
+	NexusComponentRefHeader = "X-CHASM-Component-Ref"
+
+	// Base URL for Nexus->CHASM callbacks.
+	NexusCompletionHandlerURL = "temporal://system/nexus/callback/chasm"
```
lina-temporal: I'm still planning to get rid of this and use the constant `SystemCallbackURL`, but this branch is stacked on top of others, so I'll do that after others have merged in.

chasm/nexus_completion.go
bergundy: We are thinking of deprecating the callback headers for the most part, which is why I would ask that you use a single token field.

chasm/nexus_completion.go
bergundy: I would consider using `temporal://internal` to avoid having to unpack the callback tokens.

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// RPC to History for cross-shard completion delivery.
+	_, err = e.HistoryClient.CompleteNexusOperationChasm(ctx, request)
+	if err != nil {
+		return invocationResultRetry{fmt.Errorf("failed to complete Nexus operation: %v", err)}
+	}
+
+	return invocationResultOK{}
+}
```
bergundy, line 82: Are these internal errors ever exposed to users or is this all internal?

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,155 @@
...
+	attempt    int32
+	completion nexus.OperationCompletion
+	requestID  string
+}
+
+var ErrUnimplementedHandler = serviceerror.NewUnimplemented("component does not implement NexusCompletionHandler")
+
+func (c chasmInvocation) WrapError(result invocationResult, err error) error {
+	if failure, ok := result.(invocationResultFail); ok {
+		return queues.NewUnprocessableTaskError(failure.err.Error())
```
bergundy, line 33: This will prevent the task from being retried, is that what you are trying to do here? Doesn't seem like it...

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	switch op := c.completion.(type) {
+	case *nexus.OperationCompletionSuccessful:
+		links, err := convertNexusLinksToAPILinks(op.Links)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert Nexus links: %v", err)
+		}
+
+		payload, err := io.ReadAll(op.Reader)
```
bergundy: Use `proto.Unmarshal` into `common.Payload` message here. And consider that the payload may be nil.

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,155 @@
...
+		Ref:         ref,
+		RequestId:   c.requestID,
+	}
+
+	switch op := c.completion.(type) {
+	case *nexus.OperationCompletionSuccessful:
+		links, err := convertNexusLinksToAPILinks(op.Links)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert Nexus links: %v", err)
+		}
```
bergundy: Better just log a warning here and ignore the error, it's just nice to have metadata that shouldn't fail the completion.
(same thing below).

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,155 @@
...
+	case *nexus.OperationCompletionUnsuccessful:
+		links, err := convertNexusLinksToAPILinks(op.Links)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert Nexus links: %v", err)
+		}
+
+		req = &historyservice.CompleteNexusOperationChasmRequest{
+			Completion: token,
+			State:      string(op.State),
```
bergundy: Do you need the state? It should be enough to just have the outcome oneof to distinguish between the different states.

components/callbacks/executors.go
```diff
@@ -144,17 +146,33 @@ func (e taskExecutor) loadInvocationArgs(
...
+			// CHASM internal callbacks make use of Nexus as their callback delivery
+			// mechanism, but with a special prefix.
+			if _, ok := variant.Nexus.Header[chasm.NexusComponentRefHeader]; ok {
+				invokable = chasmInvocation{
+					nexus:      variant.Nexus,
+					attempt:    callback.Attempt,
+					completion: completion,
+					requestID:  callback.RequestId,
+				}
+				break
```
bergundy: nit: I find an else block easier to follow.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -1139,6 +1139,35 @@ message ListTasksResponse {
     temporal.server.api.adminservice.v1.ListHistoryTasksResponse response = 1;
 }
 
+message CompleteNexusOperationChasmRequest {
+    // TODO - We're adding support for routing via business_id/archetype, use that here
```
bergundy: You can add `execution_id` to the routing options, `workflow_id` will also work for now. I think we are happier with the name `execution_id` than `business_id`.

proto/internal/temporal/server/api/historyservice/v1/service.proto
```diff
@@ -337,9 +337,16 @@ service HistoryService {
 
     // Complete an async Nexus Operation using a completion token. The completion state could be successful, failed, or
     // canceled.
+    //
+    // Deprecated. Will be renamed to CompleteNexusOperationHsm in a future release.
```
bergundy, line 341: Would you please add a new method while you're here and start using it while leaving the old handler in tact?

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -101,3 +101,23 @@ message ChasmComponentRef {
     repeated string component_path = 6;
     VersionedTransition component_initial_versioned_transition = 7;
 }
+
+// ChasmNexusCompletion includes details about a completed Nexus operation.
+message ChasmNexusCompletionInfo {
+    // Operation state - may only be successful / failed / canceled.
+    string state = 1;
```
bergundy: As mentioned the `oneof` is enough.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -101,3 +101,23 @@ message ChasmComponentRef {
     repeated string component_path = 6;
     VersionedTransition component_initial_versioned_transition = 7;
 }
+
+// ChasmNexusCompletion includes details about a completed Nexus operation.
+message ChasmNexusCompletionInfo {
```
bergundy: `Info` is redundant IMHO.

service/frontend/workflow_handler.go
bergundy: You need to merge your code with main (or rebase if you prefer :))

service/history/handler.go
```diff
@@ -2421,6 +2422,57 @@ func (h *Handler) CompleteNexusOperation(ctx context.Context, request *historyse
...
+	}
+
+	// Attempt to access the component and call its invocation method. We execute
+	// this similarly as we would a pure task (holding an exclusive lock), as the
+	// assumption is that the accessed component will be recording (or generating a
+	// task) based on this result.
+	_, err := h.chasmEngine.UpdateComponent(ctx, ref, func(ctx chasm.MutableContext, component chasm.Component) error {
+		handler, ok := component.(chasm.NexusCompletionHandler)
+		if !ok {
+			return serviceerror.NewUnimplementedf("component '%T' does not implement NexusCompletionHandler", component)
```
bergundy, line 2400: Will the caller consider this a retryable error? We should verify that it doesn't because the callback executor will just spin indefinitely.

chasm/nexus_completion.go
bergundy: A good reason to use `temporal://internal` callbacks is because they will trip a separate circuit breaker in the outbound queue.
We should also consider running these callbacks on the immediate "transfer" queue instead and avoid using circuit breakers altogether but let's add an issue for this, we are going to refactor callbacks to use CHASM soon and can do this work as part of that.

components/callbacks/chasm_invocation.go
lina-temporal: This is internal-facing, for operators who are log-diving.

components/callbacks/chasm_invocation.go
lina-temporal: Yes, that's what I'm trying to do; the errors I've marked as 'fail' explicitly are non-retryable (e.g., fails validation checks).

service/frontend/workflow_handler.go
lina-temporal: I'll do that now that the stacked PRs are merged.

service/history/handler.go
lina-temporal: Good point, yeah the ChasmInvocation shouldn't be retrying on this.

proto/internal/temporal/server/api/historyservice/v1/service.proto
lina-temporal: Not sure I follow - like updating all of the existing code to start using a `CompleteNexusOperationHsm`?

proto/internal/temporal/server/api/persistence/v1/chasm.proto
lina-temporal: It's to disambiguate against the token, `ChasmNexusCompletion`; I could suffix that instead, maybe?

chasm/nexus_completion.go
lina-temporal: Moved to the `commonnexus.CallbackTokenHeader` field. Relies on `temporal://internal` being unique to CHASM.

components/callbacks/chasm_invocation.go
lina-temporal: Updated, got rid of the deprecated fields on all of the new NexusChasm stuff.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
lina-temporal: Added a TODO, I'll fix this after I merge back from main

components/callbacks/chasm_invocation.go
bergundy: I see. That's fine. I think you just don't want to wrap in a destination down error though to avoid triggering the circuit breaker.

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return invocationResultOK{}
+}
+
+func (c chasmInvocation) getHistoryRequest(
+	ref *persistencespb.ChasmComponentRef,
+) (*historyservice.CompleteNexusOperationChasmRequest, error) {
+	var req *historyservice.CompleteNexusOperationChasmRequest
+
+	token := &tokenspb.NexusOperationChasmCompletion{
```
bergundy: nit: call this variable `completion`?

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,130 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert failure type: %v", err)
+		}
+
+		req = &historyservice.CompleteNexusOperationChasmRequest{
+			Completion: token,
+			Outcome: &historyservice.CompleteNexusOperationChasmRequest_Failure{
+				Failure: apiFailure,
+			},
+			CloseTime: timestamppb.New(op.CloseTime),
```
bergundy, line 127: nit: you could just set the outcome here and initialize the request outside of the switch statement.

proto/internal/temporal/server/api/historyservice/v1/service.proto
bergundy: Not to start using it, but have it defined already. You need two versions to safely point the client to use the new API.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
bergundy: Suffix how?

proto/internal/temporal/server/api/token/v1/message.proto
```diff
@@ -85,3 +86,13 @@ message NexusOperationCompletion {
     // Allows completing a started operation after a workflow has been reset.
     string request_id = 5;
 }
+
+// A completion token for a Nexus operation started from a CHASM component.
+message NexusOperationChasmCompletion {
```
bergundy: Hmm... can we merge this into the existing `NexusOperationCompletion` and deprecate the original fields?

You can just add a `component_ref` field and it would work well.

Not critical but it would simplify the decoding later.

service/history/handler.go
```diff
@@ -2421,6 +2422,54 @@ func (h *Handler) CompleteNexusOperation(ctx context.Context, request *historyse
...
+		}
+	default:
+		return nil, serviceerror.NewUnimplemented("unhandled Nexus operation outcome")
+	}
+
+	// Attempt to access the component and call its invocation method. We execute
+	// this similarly as we would a pure task (holding an exclusive lock), as the
+	// assumption is that the accessed component will be recording (or generating a
+	// task) based on this result.
+	_, err := h.chasmEngine.UpdateComponent(ctx, ref, func(ctx chasm.MutableContext, component chasm.Component) error {
```
bergundy, line 2397: Use `chasm.UpdateComponent` and avoid casting. If you merge the current main changes, you'll see that the chasm engine should already be on the request context.

bergundy (commented):
Didn't have a lot left to add. This is very close to ready.

service/history/handler.go
lina-temporal: > Use chasm.UpdateComponent and avoid casting

I don't think I can do that here in a way that keeps this generic. I'm using the `NexusCompletionHandler` interface which doesn't bundle in `Component`, therefore, if I try to accept a `NexusCompletionHandler` here, it won't meet the generic requirements of implementing `Component`'s `LifecycleState`.

proto/internal/temporal/server/api/token/v1/message.proto
lina-temporal: Sure, moved to that.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
lina-temporal: Renamed to `ChasmNexusCompletion`.

components/callbacks/chasm_invocation.go
lina-temporal: I thought so too, but `CloseTime` is independent on both `CompletionSuccess`/`CompletionFailure` variants, so you have to type assert for that as well anyways.

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,130 @@
...
+
+var ErrUnimplementedHandler = serviceerror.NewUnimplemented("component does not implement NexusCompletionHandler")
+
+func (c chasmInvocation) WrapError(result invocationResult, err error) error {
+	if failure, ok := result.(invocationResultFail); ok {
+		return queues.NewUnprocessableTaskError(failure.err.Error())
+	}
+
+	if retry, ok := result.(invocationResultRetry); ok {
+		return queues.NewDestinationDownError(retry.err.Error(), err)
```
bergundy:
```suggestion
		return err
```

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return err
+}
+
+func (c chasmInvocation) Invoke(ctx context.Context, ns *namespace.Namespace, e taskExecutor, task InvocationTask) invocationResult {
```
bergundy, line 48: Thinking about this some more, we shouldn't expose these errors. They are user visible. I suggest that instead we follow this pattern.

```go
referenceID := uuid.NewString()
e.Logger.Error("failed to decode CHASM ComponentRef", tag.Error(err), tag.NewStringTag("reference-id", referendID))
return invocationResultFail{fmt.Errorf("internal error, reference-id: %v", referenceID)}
```

Same for `invocationResultRetry`

components/callbacks/executors_test.go
```diff
@@ -443,3 +447,292 @@ func newRoot(t *testing.T) *hsm.Node {
...
+		setupHistoryClient func(*testing.T, *gomock.Controller) *historyservicemock.MockHistoryServiceClient
+		completion         nexus.OperationCompletion
+		headerValue        string
+		expectedError      error
+		assertOutcome      func(*testing.T, callbacks.Callback)
+	}{
+		{
+			name: "success-with-successful-operation",
+			setupHistoryClient: func(t *testing.T, ctrl *gomock.Controller) *historyservicemock.MockHistoryServiceClient {
+				client := historyservicemock.NewMockHistoryServiceClient(ctrl)
```
bergundy, line 482: Unrelated to this PR but I couldn't help myself. Really the whole mock and codegen here is redundant, this is so much simpler:

```go
type MockHistoryClient struct {
  historyservice.HistoryServiceClient
  HandleCompleteNexusOperationChasm func(ctx context.Context, req *historyservice.CompleteNexusOperationChasmRequest, opts ...grpc.CallOption) (*historyservice.CompleteNexusOperationChasmResponse, error)
}
```

components/callbacks/chasm_invocation.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+type chasmInvocation struct {
+	nexus      *persistencespb.Callback_Nexus
+	attempt    int32
+	completion nexus.OperationCompletion
+	requestID  string
+}
+
+var ErrUnimplementedHandler = serviceerror.NewUnimplemented("component does not implement NexusCompletionHandler")
```
bergundy: Is this used?

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -1142,6 +1142,27 @@ message ListTasksResponse {
     temporal.server.api.adminservice.v1.ListHistoryTasksResponse response = 1;
 }
 
+message CompleteNexusOperationChasmRequest {
+    option (routing) = {
+        namespace_id: "completion.component_ref.namespace_id"
+        // TODO - fix after merge
+        // execution_id: "completion.component_ref.business_id"
```
bergundy:
```suggestion
        // business_id: "completion.component_ref.business_id"
```

proto/internal/temporal/server/api/persistence/v1/chasm.proto
```diff
@@ -101,3 +101,18 @@ message ChasmComponentRef {
     repeated string component_path = 6;
     VersionedTransition component_initial_versioned_transition = 7;
 }
+
+// ChasmNexusCompletion includes details about a completed Nexus operation.
+message ChasmNexusCompletion {
```
bergundy, line 106:
```suggestion
// NexusCompletion includes details about a completed Nexus operation.
message NexusCompletion {
```

service/history/handler.go
bergundy: That sounds like a bug in CHASM. You shouldn't need to inspect the type. There's no reason that `chasmEngine.UpdateComponent` would work and `chasm.UpdateComponent(ctx)` would not work.

components/callbacks/chasm_invocation.go
lina-temporal: Nope, got rid of it.

components/callbacks/chasm_invocation.go
lina-temporal: Updated to not wrap in DestinationDown.

service/history/handler.go
lina-temporal: What component type would you expect I use here? `NexusCompletionHandler` won't work since it isn't a `Component`.

proto/internal/temporal/server/api/persistence/v1/chasm.proto
lina-temporal: Every other type in this protobuf is prefixed with `Chasm`, I'd rather keep that consistent or change all at once.

bergundy: It's not CHASM specific, you can move this somewhere else but not blocking the PR.

service/history/handler.go
bergundy: That's what I expect, that you can use the interface instead of a concrete type.

PR #8374, opened 2025-09-24
Added metrics to record activity payload events.

fretz12:
## What changed?
Added counter metrics to record activity payload size during activity start, completion, failure, and heartbeats.

## Why?
We need these metrics to better understand how activity payloads impact CDS in prod so we can design standalone activities correctly.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


common/metrics/metric_defs.go
```diff
@@ -853,6 +853,9 @@ var (
 	ActivityCancel                                       = NewCounterDef("activity_cancel")
 	ActivityTaskTimeout                                  = NewCounterDef("activity_task_timeout", WithDescription("Number of activity task timeouts (including retries)."))
 	ActivityTimeout                                      = NewCounterDef("activity_timeout", WithDescription("Number of terminal activity timeouts."))
+	ActivityPayloadSize                                  = NewCounterDef("activity_payload_size", WithDescription("Size of activity start/complete payloads in bytes."))
```
fretz12: Don't have a good way to test these besides manually in local dev. Open to ideas.

bergundy: We have tests for mutable state and adding history events that will cover most of this.
I'm okay with not adding automated testing for it though.

bergundy: Let's do all in one metric and classify what the payload is in with the operation label.

service/history/historybuilder/history_builder.go
```diff
@@ -282,6 +282,12 @@ func (b *HistoryBuilder) AddActivityTaskScheduledEvent(
 ) *historypb.HistoryEvent {
 	event := b.EventFactory.CreateActivityTaskScheduledEvent(workflowTaskCompletedEventID, command)
 	event, _ = b.EventStore.add(event)
+
+	if resultPayload, err := command.Input.Marshal(); err == nil && resultPayload != nil {
```
bergundy: It's typically recommended to use `Size()` which may avoid marshaling if possible. (Not sure it does here but the code would be easier to follow).

common/metrics/metric_defs.go
fretz12: Consolidated.

service/history/historybuilder/history_builder.go
fretz12: Love it. Switched to use Size().

PR #8379, opened 2025-09-25
Fairness weight override

stephanos:
## What changed?

API PR : https://github.com/temporalio/api/pull/626

Adds the ability to set/unset fairness weight overrides by fairness key via `UpdateTaskQueueConfig` API.

From the API:

> // Fairness weight for a task can come from multiple sources for
    // flexibility. From highest to lowest precedence:
    // 1. Weights for a small set of keys can be overridden in task queue
    //    configuration with an API.
    // 2. It can be attached to the workflow/activity in this field.
    // 3. The default weight of 1.0 will be used.
    //
    // Weight values are clamped to the range [0.001, 1000].

## Why?

To allow users to fine-tune their fairness key behavior via the API. 

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
N/A

### DIFF
--- a/tests/task_queue_test.go
+++ b/tests/task_queue_test.go
@@ -864,7 +864,7 @@ func (s *TaskQueueSuite) runActivitiesWithPriorities(
 		_, err := s.TaskPoller().PollAndHandleWorkflowTask(
 			tv,
 			func(task *workflowservice.PollWorkflowTaskQueueResponse) (*workflowservice.RespondWorkflowTaskCompletedRequest, error) {
-				s.Equal(3, len(task.History.Events))
+				s.Len(task.History.Events, 3)
 
 				var idx int
 				_, scanErr := fmt.Sscanf(task.WorkflowExecution.WorkflowId, parsePattern, &idx)
--- a/service/frontend/validators_test.go
+++ b/service/frontend/validators_test.go
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestValidateFairnessWeightUpdate(t *testing.T) {
@@ -15,14 +15,14 @@ func TestValidateFairnessWeightUpdate(t *testing.T) {
 		}
 		unset := []string{}
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 	})
 
 	t.Run("unset overrides", func(t *testing.T) {
 		set := map[string]float32{}
 		unset := []string{"z"}
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 	})
 
 	t.Run("enforce max number of overrides", func(t *testing.T) {
@@ -32,13 +32,13 @@ func TestValidateFairnessWeightUpdate(t *testing.T) {
 		unset := []string{"z"}
 
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		err = validateFairnessWeightUpdate(set, unset, 2)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		err = validateFairnessWeightUpdate(set, unset, 1)
-		assert.ErrorContains(t, err, "too many fairness weight overrides in request: got 2, maximum 1")
+		require.ErrorContains(t, err, "too many fairness weight overrides in request: got 2, maximum 1")
 	})
 
 	t.Run("reject too long key in `set`", func(t *testing.T) {
@@ -47,14 +47,14 @@ func TestValidateFairnessWeightUpdate(t *testing.T) {
 		}
 		unset := []string{}
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.ErrorContains(t, err, "fairness key length exceeds limit")
+		require.ErrorContains(t, err, "fairness key length exceeds limit")
 	})
 
 	t.Run("reject too long key in `unset`", func(t *testing.T) {
 		set := map[string]float32{"a": 1.0}
 		unset := []string{strings.Repeat("abcdefg", 10)}
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.ErrorContains(t, err, "fairness key length exceeds limit")
+		require.ErrorContains(t, err, "fairness key length exceeds limit")
 	})
 
 	t.Run("reject negative weight", func(t *testing.T) {
@@ -63,7 +63,7 @@ func TestValidateFairnessWeightUpdate(t *testing.T) {
 		}
 		unset := []string{}
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.ErrorContains(t, err, "invalid fairness weight weight for key \"a\": must be greater than zero")
+		require.ErrorContains(t, err, "invalid fairness weight weight for key \"a\": must be greater than zero")
 	})
 
 	t.Run("reject overlap between `set` and `unset`", func(t *testing.T) {
@@ -72,6 +72,6 @@ func TestValidateFairnessWeightUpdate(t *testing.T) {
 		}
 		unset := []string{"a"}
 		err := validateFairnessWeightUpdate(set, unset, 10)
-		assert.ErrorContains(t, err, "fairness weight override key \"a\" present in both set and unset lists")
+		require.ErrorContains(t, err, "fairness weight override key \"a\" present in both set and unset lists")
 	})
 }
--- a/service/matching/fairness_util_test.go
+++ b/service/matching/fairness_util_test.go
@@ -3,7 +3,7 @@ package matching
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMergeFairnessWeightOverrides(t *testing.T) {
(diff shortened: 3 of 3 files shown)


service/frontend/workflow_handler.go
```diff
@@ -6022,23 +6022,34 @@ func (wh *WorkflowHandler) UpdateTaskQueueConfig(
 	if err != nil {
 		return nil, err
 	}
+
 	// Validation: prohibit setting rate limit on workflow task queues
 	if request.TaskQueueType == enumspb.TASK_QUEUE_TYPE_WORKFLOW {
 		return nil, serviceerror.NewInvalidArgument("Setting rate limit on workflow task queues is not allowed.")
 	}
```
sivagirish81, line 6012:
```suggestion
        queueRateLimit := request.GetUpdateQueueRateLimit()
	// Validation: prohibit setting queue rate limit on workflow task queues
	// Allow fairness-related settings on workflow task queues
	if request.TaskQueueType == enumspb.TASK_QUEUE_TYPE_WORKFLOW {
		if queueRateLimit != nil {
			return nil, serviceerror.NewInvalidArgument("Setting queue rate limit on workflow task queues is not allowed.")
		}
	}
```
Currently, in the frontend routing for UpdateTaskQueueConfig, we block updates on workflow task queues because setting queue-level rate limits are not allowed for them. However, now that the fairness-related config changes have been merged, this restriction can be revised.
Since there are no restrictions on setting fairness key weight overrides or default fairness key rate limits for workflow task queues, we can refine the check. Specifically, requests that attempt to configure a queue rate limit on a workflow task queue can be rejected, but requests that update fairness-related configs can be allowed.

service/frontend/workflow_handler.go
```diff
@@ -6022,23 +6022,34 @@ func (wh *WorkflowHandler) UpdateTaskQueueConfig(
...
-	queueRateLimit := request.GetUpdateQueueRateLimit()
-	fairnessKeyRateLimitDefault := request.GetUpdateFairnessKeyRateLimitDefault()
+
 	// Validate rate limits
+	queueRateLimit := request.GetUpdateQueueRateLimit()
```
sivagirish81, line 6015: This can be moved up as suggested previously.

service/frontend/workflow_handler.go
```diff
@@ -6022,23 +6022,34 @@ func (wh *WorkflowHandler) UpdateTaskQueueConfig(
...
 		return nil, err
 	}
+
 	// Validate identity field
 	if err := validateStringField("Identity", request.GetIdentity(), wh.config.MaxIDLengthLimit(), false); err != nil {
 		return nil, err
 	}
+
+	// Validate Fairness Weight Updates
+	setFairnessWeightOverrides := request.GetSetFairnessWeightOverrides()
```
sivagirish81, line 6030: Suggestion: Consider renaming `set_fairness_weight_overrides` to `fairness_weight_overrides` in the api PR. Since this is part of UpdateTaskQueueConfig, the context already makes it clear that the fields are being updated, so the `set_` prefix may not be necessary. This change would make the field name less verbose and improve readability, especially for the getter function.

service/frontend/workflow_handler.go
stephanos: Thank you; excellent catch!

common/dynamicconfig/constants.go
```diff
@@ -849,6 +849,11 @@ of Timeout and if no activity is seen even after that the connection is closed.`
 		true,
 		`FrontendEnableSchedules enables schedule-related RPCs in the frontend`,
 	)
+	FrontendMaxFairnessWeightOverrideConfigLimit = NewGlobalIntSetting(
```
dnr: We don't need another config for this, just use the max overrides. We're never going to configure it separately.

service/frontend/validators.go
```diff
@@ -39,3 +41,47 @@ func validateStringField(fieldName string, value string, maxLen int, required bo
...
+	total := len(set) + len(unset)
+	if total > maxConfigLimit {
+		return serviceerror.NewInvalidArgumentf(
+			"too many fairness weight overrides in request: got %d, maximum %d",
+			total, maxConfigLimit,
+		)
+	}
+
+	for k, w := range set {
+		if strings.TrimSpace(k) == "" {
```
dnr: Why TrimSpace? The actual code that uses them doesn't do that, it just uses the string as-is. This should do the same.

service/frontend/validators.go
```diff
@@ -39,3 +41,47 @@ func validateStringField(fieldName string, value string, maxLen int, required bo
...
+			return serviceerror.NewInvalidArgument(
+				"fairness weight override key must not be empty")
+		}
+		if w <= 0 {
+			return serviceerror.NewInvalidArgumentf(
+				"invalid fairness weight weight for key %q: must be greater than zero", k,
+			)
+		}
+	}
```
dnr, line 73: We should use the existing code in `common/priorities` to validate (and refactor it if necessary). This is missing length validation.

service/matching/fairness_util.go
```diff
@@ -24,3 +24,29 @@ func getEffectiveWeight(overrides fairnessWeightOverrides, pri *commonpb.Priorit
...
+func mergeFairnessWeightOverrides(
+	existing map[string]float32,
+	set map[string]float32,
+	unset []string,
+	maxFairnessKeyWeightOverrides int,
+) (map[string]float32, error) {
+	res := make(map[string]float32, len(existing))
+	for k, v := range existing {
+		res[k] = v
+	}
```
dnr:
```suggestion
	res := maps.Clone(existing)
```

service/matching/fairness_util.go
```diff
@@ -24,3 +24,29 @@ func getEffectiveWeight(overrides fairnessWeightOverrides, pri *commonpb.Priorit
...
+
+	for _, k := range unset {
+		delete(res, k)
+	}
+
+	for k, w := range set {
+		res[k] = w
+	}
```
dnr:
```suggestion
	maps.Copy(res, set)
```

common/dynamicconfig/constants.go
stephanos: 👍 I had the same inclination, but wasn't sure if that had been discussed between the two of you before. I'll consolidate them.

service/frontend/validators.go
stephanos: True. Is it too late to fix that? Seems problematic to me to allow fairness keys with just whitespace. I doubt any user _wants_ that behavior.

common/priorities/priority_util.go
```diff
@@ -12,9 +12,9 @@ const (
 )
 
 var (
-	ErrInvalidPriority       = serviceerror.NewInvalidArgument("PriorityKey can't be negative")
-	ErrFairnessKeyLength     = serviceerror.NewInvalidArgument("FairnessKey length exceeds limit")
-	ErrInvalidFairnessWeight = serviceerror.NewInvalidArgument("FairnessWeight can't be negative")
+	ErrInvalidPriority       = serviceerror.NewInvalidArgument("priority key can't be negative")
+	ErrFairnessKeyLength     = serviceerror.NewInvalidArgument("fairness key length exceeds limit")
+	ErrInvalidFairnessWeight = serviceerror.NewInvalidArgument("fairness weight can't be negative")
```
stephanos: This way they can be used in multiple contexts; without losing clarity IMO.

service/frontend/service.go
```diff
@@ -78,6 +78,8 @@ type Config struct {
 	ShutdownDrainDuration                                             dynamicconfig.DurationPropertyFn
 	ShutdownFailHealthCheckDuration                                   dynamicconfig.DurationPropertyFn
 
+	MaxFairnessWeightOverrideConfigLimit dynamicconfig.IntPropertyFn
```
dnr: this is never initialized?

service/frontend/validators.go
```diff
@@ -5,6 +5,7 @@ import (
 	commonpb "go.temporal.io/api/common/v1"
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/api/workflowservice/v1"
+	priorities "go.temporal.io/server/common/priorities"
```
dnr:
```suggestion
	"go.temporal.io/server/common/priorities"
```

service/frontend/validators.go
```diff
@@ -39,3 +40,53 @@ func validateStringField(fieldName string, value string, maxLen int, required bo
...
+
+	for k, w := range set {
+		if k == "" {
+			return serviceerror.NewInvalidArgument(
+				"fairness weight override key must not be empty")
+		}
+		if err := priorities.ValidateFairnessKey(k); err != nil {
+			return err
+		}
+		if w <= 0 {
```
dnr: could move this into `priorities` too to reuse the code+error? either way

service/frontend/validators.go
```diff
@@ -39,3 +40,53 @@ func validateStringField(fieldName string, value string, maxLen int, required bo
...
+				"fairness weight override key must not be empty")
+		}
+		if err := priorities.ValidateFairnessKey(k); err != nil {
+			return err
+		}
+	}
+
+	for k := range set {
+		for _, u := range unset {
+			if k == u {
```
dnr: switch the loop order and do a map lookup in set, then it's O(U) instead of O(S*U)

service/matching/matching_engine.go
```diff
@@ -3040,18 +3040,41 @@ func (e *matchingEngineImpl) UpdateTaskQueueConfig(ctx context.Context, request
...
 			// Fairness Queue Rate Limit
 			if fkrl := updateTaskQueueConfig.GetUpdateFairnessKeyRateLimitDefault(); fkrl != nil {
 				cfg.FairnessKeysRateLimitDefault = buildRateLimitConfig(fkrl, protoTs, updateIdentity)
 			}
+
+			// Fairness Weight Overrides
+			if len(updateTaskQueueConfig.GetSetFairnessWeightOverrides()) > 0 ||
+				len(updateTaskQueueConfig.GetUnsetFairnessWeightOverrides()) > 0 {
+				maxFairnessKeyWeightOverrides := e.config.MaxFairnessKeyWeightOverrides(
+					request.NamespaceId,
```
dnr: dynamic config takes a namespace name, not id. we should prevent this with types, though we can't do much about protos 😞

dnr: but, why not use `tqm.config.MaxFairnessKeyWeightOverrides()` here? you shouldn't need to re-bind the dc to ns+tq+type

stephanos:
> Please don't force-push after comments, I can't use "Changes since your last review"

So I'm using force push after something was reviewed for two purposes: (1) rebasing and (2) correcting a commit I pushed after the review and before the next review. I suspect (1) is breaking your workflow?

dnr:
> So I'm using force push after something was reviewed for two purposes: (1) rebasing and (2) correcting a commit I pushed after the review and before the next review. I suspect (1) is breaking your workflow?

Unfortunately yes. I hate it, it's 100% a github limitation, other code review tools are not this inept. But to speed up reviews, yeah, just merge main instead of rebase? I think (2) is safe since it probably just records the last-commented-on commit, so if you preserve that one then it should be ok.

service/matching/matching_engine.go
stephanos: Ah, I inherited that code and didn't notice. I'll fix it.

Okay, `taskQueuePartitionManager` is ofc the interface and I don't have access to the config. I could add a new method ...

stephanos:
I never realized it was messing up the "Changes since your last review"; thanks for letting me know. I can see how that's annoying. I'll have to add merge instead of rebase to my workflow.

service/matching/task_queue_partition_manager_interface.go
```diff
@@ -63,5 +63,6 @@ type (
 		PutCache(key any, value any)
 		GetCache(key any) any
 		GetRateLimitManager() *rateLimitManager
+		GetConfig() *taskQueueConfig
```
dnr, line 66: I'd really like to get rid of these interfaces some day...

dnr (approved):

tests/task_queue_test.go
```diff
@@ -659,16 +549,36 @@ func (s *TaskQueueSuite) TestPerKeyRateLimit_Default_IsEnforcedAcrossThreeKeys()
...
+	fairnessKeysWithWeight := map[string]float32{"A": 1.0, "B": 1.0, "C": 1.0}
 
 	tv := testvars.New(s.T())
 
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
-	perKeyTimes, _ := s.setupPerKeyRateLimitWorkflow(ctx, tv, keys, tasksPerKey, wholeQueueRPS, perKeyRPS)
+	// Fast refresh so ratelimiter picks up config quickly.
+	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Millisecond)
```
dnr: Just a couple more thoughts, don't need to fix right now:

Why does this dc affect how the rate limiter picks up config? That should be immediate now with subscriptions?
Either the comment and setting are wrong or something is very unintuitive and should be cleaned up.

tests/task_queue_test.go
```diff
@@ -831,3 +829,119 @@ func (s *TaskQueueSuite) TestUpdateUnsetAndDescribeTaskQueueConfig() {
 	s.Equal(unsetReasonFairness, describeResp.Config.FairnessKeysRateLimitDefault.Metadata.Reason)
 	s.Equal(updateIdentity, updateResp.Config.FairnessKeysRateLimitDefault.Metadata.UpdateIdentity)
 }
+
+// removed: inlined where needed
```
dnr: this comment doesn't make sense on its own

tests/task_queue_test.go
stephanos: Good question; I suspect a copy/paste case. I've removed it now, let's see if it still succeeds.

service/matching/fairness_util.go
stephanos: This caused a panic actually since before it handled `existing` being `nil` but `Clone` returns `nil` instead. Saw it in the func tests now.

dnr: sorry, try this: https://github.com/temporalio/temporal/blob/main/common/util/util.go#L60-L61

stephanos: Good to know we have that - I've changed it to return the `set` now in that case. Added a test as well.

PR #8380, opened 2025-09-25
chore: remove unused min func

stemlaud:
## What changed?

Go 1.21 adds built-in functions min and max.   https://tip.golang.org/doc/go1.21
Therefore, we can directly remove our own implementations and use the built-in function with the same names. 

## Why?

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks


stemlaud:
Hi, If there's anything need I to do, please feel free to let me know.  @yiminc @xwduan

PR #8381, opened 2025-09-25
flakes: add slack reporting on flaky tests

spkane31:
## What changed?
Run our flaky test report on a cadence and send results in Slack

## Why?
Help oss team take action on our flaky alerts

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
NA

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -252,7 +252,7 @@ jobs:
           key: go-${{ runner.os }}${{ runner.arch }}-build-${{ env.COMMIT }}
 
       - name: Run unit tests
-        timeout-minutes: 15
+        timeout-minutes: 20
         run: make unit-test-coverage
         env:
           UNIT_TEST_DIRS: ${{ inputs.unit_test_directory }}
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -2441,6 +2441,56 @@ that task will be sent to DLQ.`,
 		true,
 		`ReplicationEnableRateLimit is the feature flag to enable replication global rate limiter`,
 	)
+	ReplicationStreamSenderErrorRetryWait = NewGlobalDurationSetting(
+		"history.ReplicationStreamSenderErrorRetryWait",
+		1*time.Second,
+		`ReplicationStreamSenderErrorRetryWait is the initial retry wait when we see errors in sending replication tasks`,
+	)
+	ReplicationStreamSenderErrorRetryBackoffCoefficient = NewGlobalFloatSetting(
+		"history.ReplicationStreamSenderErrorRetryBackoffCoefficient",
+		1.2,
+		`ReplicationStreamSenderErrorRetryBackoffCoefficient is the retry wait backoff time coefficient`,
+	)
+	ReplicationStreamSenderErrorRetryMaxInterval = NewGlobalDurationSetting(
+		"history.ReplicationStreamSenderErrorRetryMaxInterval",
+		3*time.Second,
+		`ReplicationStreamSenderErrorRetryMaxInterval is the retry wait backoff max duration`,
+	)
+	ReplicationStreamSenderErrorRetryMaxAttempts = NewGlobalIntSetting(
+		"history.ReplicationStreamSenderErrorRetryMaxAttempts",
+		80,
+		`ReplicationStreamSenderErrorRetryMaxAttempts is the max retry attempts for sending replication tasks`,
+	)
+	ReplicationStreamSenderErrorRetryExpiration = NewGlobalDurationSetting(
+		"history.ReplicationStreamSenderErrorRetryExpiration",
+		3*time.Minute,
+		`ReplicationStreamSenderErrorRetryExpiration is the max retry duration for sending replication tasks`,
+	)
+	ReplicationExecutableTaskErrorRetryWait = NewGlobalDurationSetting(
+		"history.ReplicationExecutableTaskErrorRetryWait",
+		1*time.Second,
+		`ReplicationExecutableTaskErrorRetryWait is the initial retry wait when we see errors in executing replication tasks`,
+	)
+	ReplicationExecutableTaskErrorRetryBackoffCoefficient = NewGlobalFloatSetting(
+		"history.ReplicationExecutableTaskErrorRetryBackoffCoefficient",
+		1.2,
+		`ReplicationExecutableTaskErrorRetryBackoffCoefficient is the retry wait backoff time coefficient`,
+	)
+	ReplicationExecutableTaskErrorRetryMaxInterval = NewGlobalDurationSetting(
+		"history.ReplicationExecutableTaskErrorRetryMaxInterval",
+		5*time.Second,
+		`ReplicationExecutableTaskErrorRetryMaxInterval is the retry wait backoff max duration`,
+	)
+	ReplicationExecutableTaskErrorRetryMaxAttempts = NewGlobalIntSetting(
+		"history.ReplicationExecutableTaskErrorRetryMaxAttempts",
+		80,
+		`ReplicationExecutableTaskErrorRetryMaxAttempts is the max retry attempts for executing replication tasks`,
+	)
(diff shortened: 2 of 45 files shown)


tools/flakes/README.md
```diff
@@ -0,0 +1,56 @@
...
+- `--slack-webhook`: Slack webhook URL for notifications
+- `--slack-message-type`: Type of message ('success' or 'failure')
+- `--run-id`: GitHub Actions run ID
+- `--ref-name`: Git branch name (required for failure messages)
+- `--sha`: Git commit SHA (required for failure messages)
+
+## Output Files
+
+The tool generates several report files:
+- `flaky.txt`: Markdown table of flaky tests
```
stephanos, line 46: Should it be `flaky.md`? and then `flaky_slack.txt` becomes just `flaky.txt`.

tools/flakes/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "flakes"
+version = "0.1.0"
+description = "Add your description here"
```
stephanos: Should we delete this line or fill it in?

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
+name: Flaky Tests Report
+
+on:
+  schedule:
+    # Run on Tuesdays and Fridays at noon Eastern time (5 PM UTC)
+    - cron: '0 17 * * 2,5'
+  workflow_dispatch:
+    inputs:
+      output_file:
```
stephanos: I don't see this being used.

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
...
+        default: 'out.json'
+
+permissions:
+  contents: read
+  actions: read
+
+jobs:
+  flaky-tests-report:
+    runs-on: ubuntu-latest
+
```
stephanos: nit: usually I've seen this written without a newline here.

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
...
+              echo "### Failure Point"
+              echo ""
+              echo "Summary generation failed"
+              echo ""
+            } >> "$GITHUB_STEP_SUMMARY"
+          fi
+          
+          echo "Please check the workflow logs for more details." >> "$GITHUB_STEP_SUMMARY"
+
+
```
stephanos: nit: use just one newline?

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
...
+              cat tools/flakes/retry.txt
+              echo ""
+            } >> "$GITHUB_STEP_SUMMARY"
+          fi
+          
+          if [ "$crash_count" -eq 0 ] && [ "$flaky_count" -eq 0 ] && [ "$retry_count" -eq 0 ] && [ "$timeout_count" -eq 0 ]; then
+            echo "🎉 **No test failures found in the last 7 days!**" >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+
```
stephanos: nit: use just one newline?

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
```
stephanos: Can we pin that? I find you get more unpleasant surprises when using "latest".

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
...
+
+      - name: Install Python dependencies
+        run: |
+          cd tools/flakes
+          uv sync
+
+      - name: Install tringa
+        run: |
+          uv tool install git+https://github.com/spkane31/tringa@spk/skip-xml-parsing-errors --force
```
stephanos: Hm, I wonder if we should have a temporalio fork instead. It's generally concerning to point to any employee's personal GitHub account.

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,399 @@
...
+
+      - name: Run bash script to generate output file
+        id: generate-output-file
+        env:
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+        run: |
+          set -euo pipefail
```
stephanos, line 53: Generally I think it's best to keep logic in scripts and invoke them from CI. Makes them easier to run locally, too. Could we extract the two scripts - this and the next?

stephanos: I just looked around; maybe under `develop/github`? Not sure tbh.

.github/workflows/flaky-tests-report.yml
spkane31: Had this thought too. I think we can also move the functionality into this script long term which would provide more customizability long-term

.github/workflows/flaky-tests-report.yml
spkane31: Moved all of this into the Python script

tools/flakes/pyproject.toml
spkane31: Filled it in

tools/flakes/README.md
spkane31: `flaky.txt` is created by `tringa` so I opted for the `_slack` suffix to clarify this is part of the message details.

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: Run tringa
+        id: generate-output-file
+        env:
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+        run: |
+          set -euo pipefail
+          
+          # Clean up any existing files
+          rm -rf out.json timeout.txt retry.txt flaky.txt crash.txt
```
stephanos: nit: there would never be any files here, right? I'll leave it up to you if you want to delete it or keep it in the Python script.

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,96 @@
...
+          echo "📄 Full output file:"
+          cat out.json
+
+      - name: Run Python script to process flaky tests
+        id: process-flaky-tests
+        run: |
+          echo "Starting Python script processing..."
+          echo "Current directory: $(pwd)"
+          
+          echo "Running: cd tools/flakes && uv run main.py --file ../../out.json --github-summary --slack-webhook ${{ secrets.SLACK_WEBHOOK }} --run-id ${{ github.run_id }}"
```
stephanos: nit: I'd remove the echo details as they are likely going to be out-of-sync eventually.

.github/workflows/flaky-tests-report.yml
```diff
@@ -0,0 +1,96 @@
...
+        if: steps.process-flaky-tests.outcome == 'success'
+        with:
+          name: flaky-tests-reports-${{ github.run_number }}
+          path: |
+            out.json
+            tools/flakes/crash.txt
+            tools/flakes/flaky.txt
+            tools/flakes/flaky_slack.txt
+            tools/flakes/retry.txt
+            tools/flakes/timeout.txt
```
stephanos: nit: could consider emitting these to `/out` and then just use `tools/flakes/out/*`. Makes cleanup easier, too.

stephanos (approved):
I only now notice that the script isn't actually using `tringa` itself and could also be in Go like all other scripts we have. Let's label that tech debt for now and go ahead. We could throw an AI agent at the conversion.

.github/workflows/flaky-tests-report.yml
tdeebswihart: `set -x` is your friend here: it will echo exactly what's executed as it's a shell option

PR #8384, opened 2025-09-25
Use a concrete type for metrics.Tag

BenEddy:
## What changed?

Changes `metrics.Tag` from an interface to a concrete type. 

## Why?

To reduce allocations and GC pressure. Tags are often created in hot paths as part of metrics instrumentation, and interface boxing can trigger heap allocations.

Here's a benchmark for `metrics.OperationTag("test")`
```
BenchmarkTags-14    	1000000000	       0.3070 ns/op	       0 B/op	       0 allocs/op
BenchmarkTags-14    	69621046	        17.08 ns/op	      32 B/op	       1 allocs/op
```

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -73,22 +73,14 @@ const (
 	unknownRun  = "unknown"
 )
 
-// Tag is an interface to define metrics tags
+// Tag is a struct to define metrics tags
 type Tag struct {
-	key   string
-	value string
-}
-
-func (v Tag) Key() string {
-	return v.key
-}
-
-func (v Tag) Value() string {
-	return v.value
+	Key   string
+	Value string
 }
 
 func (v Tag) String() string {
-	return fmt.Sprintf("tag{key: %q, value: %q}", v.Key(), v.Value())
+	return fmt.Sprintf("tag{key: %q, value: %q}", v.Key, v.Value)
 }
 
 // NamespaceTag returns a new namespace tag. For timers, this also ensures that we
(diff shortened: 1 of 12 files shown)


common/metrics/tags.go
```diff
@@ -74,28 +74,21 @@ const (
 )
 
 // Tag is an interface to define metrics tags
-type Tag interface {
-	Key() string
-	Value() string
+type Tag struct {
+	key   string
+	value string
```
bergundy: Just make these exported fields and get rid of the redundant methods IMHO.

BenEddy: @bergundy Sounds good, I wasn't sure if `metrics.Tag` API compatibility was a concern for third party metrics handlers.

bergundy: Ah... I think you are right, the metrics provider interface is public.

bergundy: Let me double check that.

bergundy: Okay so this whole API is public and changing `Tag` from an `interface` to a `struct` is technically a breaking change.

Users may have assumed that they can create their own tag instances and those would not be accepted any more.

In practice, I doubt that this is likely to break anyone but I'm uncomfortable making this breaking change without discussing with the team.

bergundy: I will bring this up

bergundy: We talked and decided it is safe to break this API as long as we call it out in the release notes. I will send you a link to a draft document you can put this in.

bergundy (approved):
Thanks!

PR #8385, opened 2025-09-25
Static softassert message

stephanos:
## What changed?

Changed callers of softasserts that used a dynamic message to use a static message; and pass rest in through tags.

NOTE: I've experimented with https://github.com/quasilyte/go-ruleguard to enforce it; but that is too brittle.

## Why?

It makes grouping in logs (and Antithesis) much easier as the identifier is static.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/common/softassert/softassert.go
+++ b/common/softassert/softassert.go
@@ -1,3 +1,14 @@
+// softassert provides utilities for performing assertions.
+//
+// Unlike traditional assertions that may terminate the program or panic, soft assertions log an
+// error message when a condition is not met but allow the program to continue running. The runner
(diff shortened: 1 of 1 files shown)


common/log/tag/values.go
```diff
@@ -126,6 +126,7 @@ var (
 	ComponentVisibilityHandler         = component("visibility-handler")
 	ComponentNamespaceReplication      = component("namespace-replication")
 	ComponentPersistence               = component("persistence")
+	ComponentWorkflowUpdate            = component("workflow-update")
```
stephanos, line 129: DRY

PR #8387, opened 2025-09-25
Bump unit test timeout

stephanos:
## What changed?

~Removed crash reporting from unit tests.~

Bumped unit test job timeout.

## Why?

~Unit tests don't emit JUnit reports; so when they fail it's reported that they "crashed".~

Make sure test-runner has enough time for re-runs.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -259,13 +259,6 @@ jobs:
           TEST_ARGS: ${{ needs.set-up-single-test.outputs.single_test_args }}
           TEST_TIMEOUT: ${{ needs.set-up-single-test.outputs.test_timeout }}
 
-      - name: Generate crash report
-        if: failure()
-        run:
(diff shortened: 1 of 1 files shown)


simvlad:
> Unit tests don't emit JUnit reports

Why is that? https://github.com/temporalio/temporal/blob/main/Makefile#L452 seems to produce the report, but not per attempt

stephanos:
@simvlad you're actually right, thank you for catching that! I looked in the wrong place. The [unit test crash I observed](https://github.com/temporalio/temporal/actions/runs/18021616540/job/51280244853#step:6:1593) was due to a timeout since the unit test was re-run multiple times.

It says `DONE 6756 tests, 80 skipped, 23 failures in 359.125s`, so about 6min. x3 is over 15min. Let me change this PR to allow for for 20min instead.

simvlad (approved):
👍

PR #8388, opened 2025-09-25
softassert serviceerror.Internal helper

stephanos:
## What changed?

Added a new helper `UnexpectedInternalErr` to combine logging and error construction in one go.

## Why?

Observed use case in CHASM code and this collapses two separate statements (softassert and serviceerror.NewInternal) into one - which reduces friction.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


chasm/tree.go
```diff
@@ -804,7 +830,12 @@ func (n *Node) stringToMapKey(nodeName string, key string, keyT reflect.Type) (r
 		keyV = reflect.ValueOf(b)
 	default:
 		err = fmt.Errorf("unsupported type %s of kind %s: supported key types: %s", keyT.String(), keyT.Kind().String(), mapKeyTypes)
-		softassert.Fail(n.logger, err.Error())
+		softassert.Fail(
+			n.logger,
+			"unsupported CHASM map key type",
+			"",
+			tag.Error(err),
```
stephanos, line 837: TODO: is this right?

common/softassert/serviceerror.go
```diff
@@ -0,0 +1,35 @@
...
+	// using a unique tag name to avoid potential conflicts with other tags
+	errorDetailsTagName = "softassert-error-details"
+)
+
+// UnexpectedInternalErr creates a serviceerror.Internal from a static message and error details.
+// The error message follows the format "<staticMessage>: <optionalErr>", if optionalErr is not nil.
+//
+// It also logs the error using softassert.Fail.
+// `optionalErr` is added as a tag with key "softassert-error-details".
+func UnexpectedInternalErr(logger log.Logger, staticMessage string, optionalErr error, tags ...tag.Tag) error {
```
stephanos, line 21: I had `optionalErr error` as `errorDetails string` at first; but I feel the error is better as it's clearer - even if it requires one extra `fmt.Errorf`. Happy to discuss!

chasm/field_test.go
```diff
@@ -318,6 +318,9 @@ func (s *fieldSuite) TestUnresolvableDeferredPointerError() {
 	s.nodeBackend.EXPECT().GetWorkflowKey().Return(tv.Any().WorkflowKey()).AnyTimes()
 	s.nodeBackend.EXPECT().AddTasks(gomock.Any()).AnyTimes()
 
+	s.logger.(*testlogger.TestLogger).
+		Expect(testlogger.Error, "failed to resolve deferred pointer during transaction close")
```
stephanos, line 322: This test seems to intentionally trigger this so we need to tell the testlogger.

lina-temporal (approved):
Discussed during standup, 👍 , I like that we can quickly get a lot of assertions in for the internal errors we're already generating.

PR #8389, opened 2025-09-25
Abstract workflow verifier for force replication workflow

xwduan:
- **Abstract workflow verifier for force replication workflow**
- **fix**

## What changed?
Abstract workflow verifier for force replication workflow

## Why?
To allow user plugin custom verification logic

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
no risk.

### DIFF
--- a/service/worker/migration/activities.go
+++ b/service/worker/migration/activities.go
@@ -115,7 +115,18 @@ type (
 		forceReplicationMetricsHandler   metrics.Handler
 		namespaceReplicationQueue        persistence.NamespaceReplicationQueue
 		generateMigrationTaskViaFrontend dynamicconfig.BoolPropertyFn
+		workflowVerifier                 WorkflowVerifier
 	}
+
+	WorkflowVerifier func(
+		ctx context.Context,
+		request *verifyReplicationTasksRequest,
+		remoteAdminClient adminservice.AdminServiceClient,
+		localAdminClient adminservice.AdminServiceClient,
+		ns *namespace.Namespace,
+		we *commonpb.WorkflowExecution,
+		mu *adminservice.DescribeMutableStateResponse,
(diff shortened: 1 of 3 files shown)


service/worker/migration/activities.go
```diff
@@ -709,10 +720,11 @@ func (a *activities) verifySingleReplicationTask(
 
 	switch err.(type) {
 	case nil:
-		a.forceReplicationMetricsHandler.WithTags(metrics.NamespaceTag(request.Namespace)).Counter(metrics.VerifyReplicationTaskSuccess.Name()).Record(1)
-		return verifyResult{
-			status: verified,
-		}, nil
+		result, err := a.workflowVerifier(ctx, request, remotAdminClient, a.adminClient, ns, we, mu)
+		if err != nil && result.status == verified {
```
yux0: err == nil?

service/worker/migration/activities.go
```diff
@@ -700,7 +711,7 @@ func (a *activities) verifySingleReplicationTask(
 ) (verifyResult, error) {
 	s := time.Now()
 	// Check if execution exists on remote cluster
-	_, err := remotAdminClient.DescribeMutableState(ctx, &adminservice.DescribeMutableStateRequest{
+	mu, err := remotAdminClient.DescribeMutableState(ctx, &adminservice.DescribeMutableStateRequest{
```
yux0, line 714: is it possible to move 714-748 into this verifier?

xwduan: Synced offline. line 714-748 contains some logic that could be shared across different verifier i..e handling the scenario where the workflow is not found. This pr is to provide ability to override the verification when the workflow is found.

yux0 (approved):

PR #8391, opened 2025-09-26
Add functional test to verify heartbeat is registered via PollNexusTaskQueue API.

rkannan82:
## What changed?
Heartbeat registration can happen as part of PollNexusTaskQueue API as well. This is an optimization to piggy back on the long poll instead of sending a separate heartbeat.

In addition, made changes to existing test to verify the heartbeat result.

## Why?
Increase coverage.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None.

### DIFF
--- a/tests/worker_registry_test.go
+++ b/tests/worker_registry_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"go.temporal.io/api/serviceerror"
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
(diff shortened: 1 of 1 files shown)


tests/worker_registry_test.go
```diff
@@ -36,15 +38,26 @@ func (s *WorkerRegistryTestSuite) TestWorkerRegistry_DescribeWorker() {
...
+	// Record heartbeat for 2 workers
+	hbResp, err := s.FrontendClient().RecordWorkerHeartbeat(ctx, &workflowservice.RecordWorkerHeartbeatRequest{
 		Namespace: s.Namespace().String(),
 		WorkerHeartbeat: []*workerpb.WorkerHeartbeat{
-			{WorkerInstanceKey: "worker1", TaskQueue: "taskQueue1"},
-			{WorkerInstanceKey: "worker2", TaskQueue: "taskQueue2"},
+			{
+				WorkerInstanceKey:   "worker1",
+				TaskQueue:           "taskQueue1",
+				TotalStickyCacheHit: 1,
```
Shivs11: not going to block on this but just pasting this as a good practice - you should consider using the testvars package in situations where you may realize you are repeating test strings like "worker1", "unknown-namespace", etc

tests/worker_registry_test.go
```diff
@@ -61,26 +75,62 @@ func (s *WorkerRegistryTestSuite) TestWorkerRegistry_DescribeWorker() {
...
 	s.ErrorAs(err, &namespaceNotFound)
 
-	resp, err := s.FrontendClient().DescribeWorker(ctx, &workflowservice.DescribeWorkerRequest{
-		Namespace:         s.Namespace().String(),
-		WorkerInstanceKey: "worker1",
-	})
-	s.NoError(err)
-	s.NotNil(resp)
+	// Test success case - verify worker1 heartbeat data
+	{
```
Shivs11, line 88: !!!! I had no idea you could just create mini scopes like this in go - cool!

tests/worker_registry_test.go
```diff
@@ -110,6 +184,58 @@ func (s *WorkerRegistryTestSuite) TestWorkerRegistry_ListWorkers() {
...
+		_, _ = s.FrontendClient().PollNexusTaskQueue(ctx, &workflowservice.PollNexusTaskQueueRequest{
+			Namespace:       s.Namespace().String(),
+			TaskQueue:       &taskqueuepb.TaskQueue{Name: "nexusTaskQueue"},
+			Identity:        "nexus-worker",
+			WorkerHeartbeat: []*workerpb.WorkerHeartbeat{heartbeat},
+		})
+	}()
+
+	// Use Eventually to verify heartbeat registration
+	s.Eventually(func() bool {
```
Shivs11: I think you should use EventuallyWithT in this case - the reason being that your current block of code is waiting on a condition that is eventually consistent *and* you are also conducting some assertions within the eventually block

If you use EventuallyWithT, I think you would be able to get rid of the print statements and can verify assertions quite easily

tests/worker_registry_test.go
rkannan82: Done.

PR #8393, opened 2025-09-27
Convert HistoryManager error logs into softassert

stephanos:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks


prathyushpv:
Thanks for making this change.
We can also add this soft assert in this function: https://github.com/temporalio/temporal/blob/b0cf1b02ed6072ca306bc4dd315cd6c92e1b6bfc/common/persistence/persistence_metric_clients.go#L1413-L1413

This can catch any data loss errors returned from the persistence layer.

stephanos:
After a brief discussion, decided to not add it to persistence as that's an internal error that we'll tag where it's created instead.

PR #8395, opened 2025-09-29
Use only history length for force replication

yux0:
## What changed?
Use only history length for force replication

## Why?
With state based replication, only the history length generate extra loads to replication history events. The history events QPS is based on the number of history event batches.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/worker/migration/activities.go
+++ b/service/worker/migration/activities.go
@@ -401,7 +401,7 @@ func (a *activities) generateWorkflowReplicationTask(
 	// in such situation.
 	count := min(stateTransitionCount, historyLength)
 	if a.enableHistoryRateLimiter() {
-		count = historyLength / largeHistoryLength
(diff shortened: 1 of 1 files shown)


service/worker/migration/activities.go
```diff
@@ -137,6 +138,8 @@ const (
 	notVerified verifyStatus = 0
 	verified    verifyStatus = 1
 	skipped     verifyStatus = 2
+
+	largeHistoryLength = 1000
```
xwduan, line 142: What does largeHistoryLength mean here? And why do we want to divide by it at line 404?

yux0: I don't think it makes sense to acquire tokens as the same amount of history events. So. I think it should be ok to reserve 1 token per 1000 events.

PR #8396, opened 2025-09-29
Update Workflow/Activity priority

stephanos:
## What changed?

Allow updating priority of Workflow and Activity.

Based on

- https://github.com/temporalio/api/pull/610
- https://github.com/temporalio/temporal/pull/8103

## Why?

Users want to change the priority after starting the workflow/activity.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -156,9 +156,6 @@ type (
 
 	// NodesMutation is a set of mutations for all nodes rooted at a given node n,
 	// including the node n itself.
-	//
-	// TODO: Return tree size changes in NodesMutation as well. MutateState needs to
-	// track the overall size of itself and terminate workflow if it exceeds the limit.
 	NodesMutation struct {
 		UpdatedNodes map[string]*persistencespb.ChasmNode // encoded node path -> chasm node
 		DeletedNodes map[string]struct{}
--- a/common/persistence/visibility/store/elasticsearch/visibility_store.go
+++ b/common/persistence/visibility/store/elasticsearch/visibility_store.go
@@ -705,9 +705,7 @@ func (s *VisibilityStore) GetListFieldSorter(fieldSorts []elastic.Sorter) ([]ela
 		return defaultSorter, nil
 	}
 	res := make([]elastic.Sorter, len(fieldSorts)+1)
-	for i, fs := range fieldSorts {
-		res[i] = fs
-	}
+	copy(res, fieldSorts)
 	// RunID is explicit tiebreaker.
 	res[len(res)-1] = elastic.NewFieldSort(sadefs.RunID).Desc()
 
@@ -1052,9 +1050,7 @@ func (s *VisibilityStore) parseCountGroupByResponse(
 				return fmt.Errorf("unable to parse 'doc_count' field: %w", err)
 			}
 			groupValues := make([]*commonpb.Payload, len(groupByFields))
-			for i := range bucketValues {
-				groupValues[i] = bucketValues[i]
-			}
+			copy(groupValues, bucketValues)
 			response.Groups = append(
 				response.Groups,
 				&workflowservice.CountWorkflowExecutionsResponse_AggregationGroup{
--- a/service/history/consts/const.go
+++ b/service/history/consts/const.go
@@ -91,6 +91,8 @@ var (
 	ErrChildExecutionNotFound = serviceerror.NewNotFound("Pending child execution not found.")
 	// ErrWorkflowNotReady is error indicating workflow mutable state is missing necessary information for handling the request
 	ErrWorkflowNotReady = serviceerror.NewWorkflowNotReady("Workflow state is not ready to handle the request.")
+	// ErrWorkflowTaskNotFound is error indicating workflow task is not found, either due to it has already completed or attempt/stamp mismatch
+	ErrWorkflowTaskNotFound = serviceerror.NewNotFound("workflow task not found.")
 	// ErrWorkflowTaskNotScheduled is error indicating workflow task is not scheduled yet.
 	ErrWorkflowTaskNotScheduled = serviceerror.NewWorkflowNotReady("Workflow task is not scheduled yet.")
 	// ErrNamespaceHandover is error indicating namespace is in handover state and cannot process request.
--- a/service/history/queues/executable.go
+++ b/service/history/queues/executable.go
@@ -43,7 +43,6 @@ type (
 		ctasks.Task
 		tasks.Task
 
-		Attempt() int
 		GetTask() tasks.Task
 		GetPriority() ctasks.Priority
 		GetScheduledTime() time.Time
@@ -111,10 +110,7 @@ type (
 		tasks.Task
 
 		sync.Mutex
-		state          ctasks.State
-		priority       ctasks.Priority // priority for the current attempt
-		lowestPriority ctasks.Priority // priority for emitting metrics across multiple attempts
-		attempt        int
+		state ctasks.State
 
 		executor            Executor
 		scheduler           Scheduler
@@ -131,6 +127,8 @@ type (
 		dlqWriter           *DLQWriter
 
 		readerID                   int64
+		attempt                    int
+		priority                   ctasks.Priority
 		scheduledTime              time.Time
 		scheduleLatency            time.Duration
 		attemptNoUserLatency       time.Duration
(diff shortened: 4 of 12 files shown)


service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -166,11 +166,10 @@ func recordActivityTaskStarted(
 
 	if ai.Stamp != request.Stamp {
 		// activity has changes before task is started.
-		// ErrActivityStampMismatch is the error to indicate that requested activity has mismatched stamp
 		errorMessage := fmt.Sprintf(
 			"Activity task with this stamp not found. Id: %s,: type: %s, current stamp: %d",
 			ai.ActivityId, ai.ActivityType.Name, ai.Stamp)
-		return nil, rejectCodeUndefined, serviceerror.NewNotFound(errorMessage)
+		return nil, rejectCodeUndefined, serviceerrors.NewObsoleteMatchingTask(errorMessage)
```
stephanos, line 172: This seems much more consistent and a better error type, too.

service/history/api/updateactivityoptions/api.go
```diff
@@ -202,10 +203,48 @@ func mergeActivityOptions(
...
+		mergeInto.Priority.FairnessWeight = mergeFrom.Priority.FairnessWeight
+	}
+
 	if mergeInto.RetryPolicy == nil {
 		mergeInto.RetryPolicy = &commonpb.RetryPolicy{}
 	}
 
+	if _, ok := updateFields["retryPolicy"]; ok {
+		mergeInto.RetryPolicy = mergeFrom.RetryPolicy
+	}
```
stephanos, line 247: This seems like a reasonable addition to me.

tests/priority_fairness_test.go
```diff
@@ -348,3 +354,195 @@ func unfairness(vs []int) float64 {
...
+		Execution: &commonpb.WorkflowExecution{
+			WorkflowId: tv.WorkflowID(),
+			RunId:      startResp.GetRunId(),
+		},
+	})
+	s.NoError(err)
+	s.NotNil(descResp.GetPendingActivities())
+	s.Len(descResp.GetPendingActivities(), 1)
+	s.ProtoEqual(updatedPriority, descResp.GetPendingActivities()[0].GetPriority())
+	s.ProtoEqual(updatedPriority, descResp.GetPendingActivities()[0].GetActivityOptions().GetPriority())
```
stephanos, line 723: Other activity options appear to always match the activity's fields 1:1.

service/history/api/updateactivityoptions/api.go
```diff
@@ -148,6 +148,7 @@ func processActivityOptionsUpdate(
 		ScheduleToStartTimeout: ai.ScheduleToStartTimeout,
 		StartToCloseTimeout:    ai.StartToCloseTimeout,
 		HeartbeatTimeout:       ai.HeartbeatTimeout,
+		Priority:               ai.Priority,
```
dnr: hmm, this is mergeInto, so mergeActivityOptions will write directly into ai.Priority?
shouldn't we expand it like RetryPolicy? or use common.CloneProto (on both)?

service/history/api/updateactivityoptions/api.go
dnr: I guess so. I was originally a little confused at the semantics, but it seems reasonable, to allow updating a whole sub-object without listing all the fields

service/history/api/updateworkflowoptions/api.go
```diff
@@ -125,6 +133,11 @@ func getOptionsFromMutableState(ms historyi.MutableState) *workflowpb.WorkflowEx
 		}
 		opts.VersioningOverride = override
 	}
+	if priority := ms.GetExecutionInfo().GetPriority(); priority != nil {
+		if prio, ok := proto.Clone(priority).(*commonpb.Priority); ok {
+			opts.Priority = prio
+		}
```
dnr:
```suggestion
		opts.Priority = common.CloneProto(priority)
```

dnr: (and change the one on line 130 also)

service/history/api/updateworkflowoptions/api.go
```diff
@@ -155,5 +188,10 @@ func mergeWorkflowExecutionOptions(
 		}
 		mergeInto.VersioningOverride = mergeFrom.GetVersioningOverride()
 	}
+
+	if _, ok := updateFields["versioningOverride"]; ok {
```
dnr: move this one before the ones that do sub-fields so they're consistent? (even though it doesn't matter in this case)

service/history/api/updateworkflowoptions/api.go
```diff
@@ -138,10 +151,30 @@ func mergeWorkflowExecutionOptions(
...
-		mergeInto.VersioningOverride = mergeFrom.GetVersioningOverride()
+
+	// ==== Priority
+
+	if _, ok := updateFields["priority"]; ok {
+		mergeInto.Priority = mergeFrom.GetPriority()
+	}
+
+	if _, ok := updateFields["priority.priorityKey"]; ok {
+		mergeInto.Priority = cmp.Or(mergeInto.Priority, &commonpb.Priority{})
```
dnr: hmm, maybe this should do an `if` like the updateactivityoptions one to avoid the extra allocation? (I'm not sure it does an extra allocation but it looks like it)

if you want a generic helper to initialize a pointer if it's nil, I'd be okay with that. I've wanted it in various places but it feels a little un-go-like so I've just done if statements or local helpers

service/matching/forwarder.go
```diff
@@ -130,8 +130,8 @@ func (fwdr *Forwarder) ForwardTask(ctx context.Context, task *internalTask) erro
 				ScheduleToStartTimeout: expirationDuration,
 				ForwardInfo:            fwdr.getForwardInfo(task),
 				VersionDirective:       task.event.Data.GetVersionDirective(),
-				Priority:               task.event.Data.GetPriority(),
 				Stamp:                  task.event.Data.GetStamp(),
+				Priority:               task.event.Data.GetPriority(),
```
dnr, line 134: keep them in the same order as the declaration?

tests/priority_fairness_test.go
```diff
@@ -348,3 +354,195 @@ func unfairness(vs []int) float64 {
...
+	s.Equal(2, addActivityTaskCount, "Expected 2 activity tasks to be dispatched to matching")
+	for _, rec := range snap[metrics.ClientFailures.Name()] {
+		for key, val := range rec.Tags {
+			if key == metrics.ErrorTypeTagName && val == fmt.Sprintf("%T", serviceerror.ObsoleteMatchingTask{}) {
+				obsoleteActivityTaskCount++
+			}
+		}
+	}
+	s.Equal(1+obsoleteWorkflowTaskCount, obsoleteActivityTaskCount, "Expected 1 activity task to be obsolete")
+}
```
dnr, line 758: nice test, clear and thorough. the metrics stuff doesn't add any flakiness, I hope?

stephanos: I don't expect any; it's all in-memory and only checked in latency insensitive places AFAICT

service/history/api/updateworkflowoptions/api.go
stephanos: let's go with a simple `if` 👍

service/history/api/updateactivityoptions/api.go
stephanos: good call 👍  (there is no `ai.RetryPolicy`, though)

service/history/api/updateworkflowoptions/api.go
```diff
@@ -109,8 +111,14 @@ func MergeAndApply(
 	if mergedOpts.GetVersioningOverride() == nil {
 		unsetOverride = true
 	}
-	_, err = ms.AddWorkflowExecutionOptionsUpdatedEvent(mergedOpts.GetVersioningOverride(), unsetOverride, "", nil, nil)
-	if err != nil {
+	if _, err = ms.AddWorkflowExecutionOptionsUpdatedEvent(
```
dnr: I forgot if we talked about this: if a wf Priority changes, and there are pending activities that got some or all of their Priority fields from the wf's Priority, then we should reschedule those activities too, right? does this do that? if not, should we do it in a followup PR?

dnr (approved):
I noticed this is on a feature branch.. should it be retargetted to main now?

stephanos:
@dnr Yup, I can retarget it once I merge the API change. Speaking of which, that's not approved yet: https://github.com/temporalio/api/pull/610

service/history/api/updateworkflowoptions/api.go
stephanos: It's been a while since I looked/worked on this, but that sounds like a gap. I don't think it'll do that.

dnr: Yeah, currently it doesn't look like it will. We should add that

PR #8397, opened 2025-09-29
Server Visibility config and preallocate custom search attributes

rodrigozhou:
## What changed?
Add a top-level Visibility config in the server config file.
Option to overwrite the default number of preallocated custom search attributes in the server config (`visibility.persistenceCustomSearchAttributes`).
Register the preallocated custom search attributes when using custom Visibility store.
The config can be changed at any point: increasing the number of custom search attributes will register additional custom search attributes in the cluster metadata; decreasing the number of custom search attributes is no-op.

Eg.: the following config would set the max number of Keyword search attributes to 20, Text search attributes to 0, and leave all other types with default max number set in `common/searchattributes/defs.go#defaultNumDBCustomSearchAttributes`.
```
visibility:
	persistenceCustomSearchAttributes:
		Keyword: 20
		Text: 0
```

Note that you also need to modify the schema and make the appropriate changes to support the limits set in the config.

## Why?
Be able to add more custom search attributes when using SQL Visibility store.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/common/searchattribute/defs.go
+++ b/common/searchattribute/defs.go
@@ -200,7 +200,7 @@ var (
 	dbCustomSearchAttributeFieldNameRE = func() map[enumspb.IndexedValueType]*regexp.Regexp {
 		res := map[enumspb.IndexedValueType]*regexp.Regexp{}
 		for t := range defaultNumDBCustomSearchAttributes {
-			res[t] = regexp.MustCompile(fmt.Sprintf(`^%s\d\d$`, t.String()))
+			res[t] = regexp.MustCompile(fmt.Sprintf(`^%s(0[1-9]|[1-9][0-9])$`, t.String()))
 		}
 		return res
 	}()
@@ -247,12 +247,12 @@ func GetSqlDbColName(name string) string {
 }
 
 func GetDBIndexSearchAttributes(
-	overwrite map[enumspb.IndexedValueType]int,
+	override map[enumspb.IndexedValueType]int,
 ) *persistencespb.IndexSearchAttributes {
 	csa := map[string]enumspb.IndexedValueType{}
 	for saType, defaultNumAttrs := range defaultNumDBCustomSearchAttributes {
 		numAttrs := defaultNumAttrs
-		if value, ok := overwrite[saType]; ok {
+		if value, ok := override[saType]; ok {
 			numAttrs = value
 		}
 		for i := range numAttrs {
--- a/common/searchattribute/defs_test.go
+++ b/common/searchattribute/defs_test.go
@@ -1,8 +1,6 @@
 package searchattribute
 
 import (
-	"fmt"
-	"math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -45,7 +43,7 @@ func TestGetDbIndexSearchAttributesDefault(t *testing.T) {
 }
 
 func TestGetDbIndexSearchAttributes(t *testing.T) {
-	overwrite := map[enumspb.IndexedValueType]int{
+	override := map[enumspb.IndexedValueType]int{
 		// Bool: no overwrite, it should use default value
 		enumspb.INDEXED_VALUE_TYPE_DATETIME:     1,
 		enumspb.INDEXED_VALUE_TYPE_DOUBLE:       2,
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/fatih/color v1.18.0
 	github.com/go-faker/faker/v4 v4.6.0
 	github.com/go-jose/go-jose/v4 v4.0.5
+	github.com/go-playground/validator/v10 v10.27.0
```
bergundy: This added a bunch of new dependencies that we don't need. Let's just write 5 lines of code to check what's required and other constraints instead of adding more dependency bloat.

bergundy (commented):
I don't have the context to know why we need to make this customizable.

rodrigozhou:
> I don't have the context to know why we need to make this customizable.

This feature has been asked by some users, although I can't find any open tickets. Maybe someone in the community or Slack.

go.mod
rodrigozhou: Reverted this change.

common/config/config.go
```diff
@@ -573,6 +581,25 @@ type (
...
+		// If any type is not specified, it will pre-allocate the default number of custom search
+		// attributes for the type defined in the map defaultNumDbCustomSearchAttributes in
+		// common/searchattribute/defs.go.
+		// Modifying the number of pre-allocated custom search attributes:
+		// - if you increase a number, it will pre-allocate additional custom search attributes to match
+		//   the desired number;
+		// - if you decrease a number, it will not delete the existing custom search attributes, ie., it
+		//   is no-op.
+		// This config only applies to SQL or custom Visibility stores.
+		PersistenceCustomSearchAttributes map[string]int `yaml:"persistenceCustomSearchAttributes" validate:"persistence_custom_search_attributes"`
```
bergundy, line 600: I was going to propose that we get rid of the validator dependency altogether and just hardcode the validation logic.
It would be easier to follow without all of the reflection.

common/searchattribute/defs_test.go
```diff
@@ -1,19 +1,107 @@
...
+}
+
+func TestIsCustomSearchAttributeFieldName(t *testing.T) {
+	allTypes := make([]enumspb.IndexedValueType, 0, len(enumspb.IndexedValueType_name))
+	for t := range enumspb.IndexedValueType_name {
+		allTypes = append(allTypes, enumspb.IndexedValueType(t))
+	}
+	for range 100 {
+		tp := allTypes[rand.Intn(len(allTypes))]
+		idx := rand.Intn(1000)
```
bergundy: Why do you need randomness in this test? Just test what you want explicitly.

common/searchattribute/defs.go
```diff
@@ -186,36 +187,23 @@ var (
...
+		enumspb.INDEXED_VALUE_TYPE_DATETIME:     3,
+		enumspb.INDEXED_VALUE_TYPE_KEYWORD:      10,
+		enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST: 3,
+		enumspb.INDEXED_VALUE_TYPE_TEXT:         3,
 	}
+
+	dbCustomSearchAttributeFieldNameRE = func() map[enumspb.IndexedValueType]*regexp.Regexp {
+		res := map[enumspb.IndexedValueType]*regexp.Regexp{}
+		for t := range defaultNumDBCustomSearchAttributes {
+			res[t] = regexp.MustCompile(fmt.Sprintf(`^%s\d\d$`, t.String()))
```
bergundy: Technically this allows `00`. You might want to change this to `(0[1-9]|[1-9]\d)`.

common/searchattribute/defs_test.go
```diff
@@ -1,19 +1,107 @@
...
+		"KeywordList01": enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
+		"KeywordList02": enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
+		"KeywordList03": enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
+	}
+	got := GetDBIndexSearchAttributes(nil)
+	require.Equal(t, want, got.CustomSearchAttributes)
+}
+
+func TestGetDbIndexSearchAttributes(t *testing.T) {
+	overwrite := map[enumspb.IndexedValueType]int{
```
bergundy: nit: override.

bergundy (approved):

common/searchattribute/defs_test.go
rodrigozhou: No particular reason.

common/config/config.go
rodrigozhou: Maybe, but the config loader is kind of generic, and if I were to implement a validation function for the visibility struct, I would have to traverse the config to find the struct and validate it if found. In this case, using the validator module does help.

PR #8398, opened 2025-09-29
CHASM client codegen

bergundy:
## What changed?

Added a protoc plugin to generate client for internal CHASM requests to the history service.
A CHASM library can now define its own set of gRPC services and register a handler for those.

I had to modify the history client files to be generic, which broke the mock generation, so I had to hand code the tests there, which IMHO is actually nicer than the way they were with mocks.

Also opted out of mock generation for CHASM services, there's just not enough value there.

**NOTE**: Instead of creating three separate clients, I generated a single layered client, with a constructor that is DI friendly. There doesn't seem to be a good enough reason to break this out into separate clients or make client construction configurable (for now).

**NOTE**: We can also get rid of the genrpcwrappers script eventually and use the protoc plugin approach for all services, I decided not to do that here though to reduce scope.

For a preview of the generated file see:
- https://github.com/bergundy/temporal/blob/3a89614a9097eca9061b5f347c01eda735534bbe/chasm/lib/activity/proto/v1/service.proto
- https://github.com/bergundy/temporal/blob/3a89614a9097eca9061b5f347c01eda735534bbe/chasm/lib/activity/gen/activitypb/v1/service_client.pb.go
 
## Why?

Keep all CHASM library functionality contained.

## How did you test it?
- [ ] built


common/backoff/retry.go
```diff
@@ -97,6 +97,66 @@ func ThrottleRetryContext(
 	return ctx.Err()
 }
 
+// ThrottleRetryContext is a context and resource aware version of Retry.
+// Context timeout/cancellation errors are never retried, regardless of IsRetryable.
+// Resource exhausted error will be retried using a different throttle retry policy, instead of the specified one.
+// TODO: allow customizing throttle retry policy and what kind of error are categorized as throttle error.
+func ThrottleRetryContextWithReturn[T any](
```
fretz12, line 104: Is this supposed to be checked in as part of chasm codegen or this addresses something else?

proto/internal/temporal/server/api/routing/v1/extension.proto
```diff
@@ -0,0 +1,18 @@
...
+package temporal.server.api.routing.v1;
+
+option go_package = "go.temporal.io/server/api/routing/v1;routing";
+
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.MethodOptions { optional RoutingOptions routing = 7234; }
+
+message RoutingOptions {
+  // Requested routed by task token or workflow ID may also specify how to
```
fretz12: nit: typo `Request`

common/backoff/retry.go
bergundy: It's used in the generated code.

PR #8399, opened 2025-09-30
CHASM: Improve sync sub component logic

yycptt:
## What changed?
- Improve sync sub component logic so that syncing only happens when the node is dirty. Instead of always sync when node is deserialized. 

## Why?
- Allow doing multiple transitions and sync sub component step per transaction.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -48,8 +48,8 @@ var (
 //
 // The enum value below is defined in increasing order of "dirtiness".
 // - NeedDeserialize: Value is not even deserialized yet.
-// - Synced: Value is deserialized, but are in sync with both serializedNode and children.
-// - NeedSerialize: Value is deserialized, but data part is not in sync with serializedNode, but the tree structure has been synced.
+// - Synced: Value is deserialized and in sync with both serializedNode and children.
+// - NeedSerialize: Value is deserialized, the child tree structure is synced, but the value is not in sync with serializedNode.
 // - NeedSyncStructure: Value is deserialized, neither data nor tree structure is synced.
 //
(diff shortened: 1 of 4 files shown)


chasm/tree.go
```diff
@@ -41,13 +41,31 @@ var (
...
+// valueState is an in-memory indicator of the dirtiness of a deserialized node value.
+// The dirtiness has two parts:
+//  1. If the data part of the value is in sync with the serializedNode field.
+//  2. For component node, if the structure of the component is in sync with the children field.
+//
+// The enum value below is defined in increasing order of "dirtiness".
+// - NeedDeserialize: Value is not even deserialized yet.
+// - Synced: Value is deserialized, but are in sync with both serializedNode and children.
+// - NeedSerialize: Value is deserialized, but data part is not in sync with serializedNode, but the tree structure has been synced.
+// - NeedSyncStructure: Value is deserialized, neither data nor tree structure is synced.
```
lina-temporal:
```suggestion
// - Synced: Value is deserialized and in sync with both serializedNode and children.
// - NeedSerialize: Value is deserialized, the child tree structure is synced, but the value is not in sync with serializedNode.
```

minor reflow suggestion ^

chasm/tree.go
```diff
@@ -41,13 +41,31 @@ var (
 	errTaskNotValid      = serviceerror.NewNotFound("task is no longer valid")
 )
 
+// valueState is an in-memory indicator of the dirtiness of a deserialized node value.
```
lina-temporal, line 44: Thank you for the docs on this!

lina-temporal (approved):

PR #8400, opened 2025-09-30
CHASM: Sync tree structure when processing pure tasks

yycptt:
## What changed?
- Fix pure task processing logic and sync tree structure after processing tasks from a node

## Why?
- Pure task execution logic can potentially delete a node with valid pure task in which case pure tasks from the deleted node should not be executed.

## How did you test it?
- [x] built
- [x] added new unit test(s)


chasm/tree.go
```diff
@@ -2102,53 +2102,94 @@ func isComponentTaskExpired(
...
+			err := node.prepareComponentValue(ctx)
 			if err != nil {
 				return err
 			}
 
-			taskAttributes := TaskAttributes{
-				ScheduledTime: task.ScheduledTime.AsTime(),
-				Destination:   task.Destination,
+		TaskLoop:
+			for _, task := range componentAttr.GetPureTasks() {
```
lina-temporal: I think moving `TaskLoop` to a separate private function would give a quick readability win.

lina-temporal (approved):

chasm/tree.go
yycptt: 👍 makes sense. Moved to a separate function and simplified the logic.

PR #8401, opened 2025-09-30
CHSAM: Support Immediate Pure Task

yycptt:
## What changed?
- Support Immediate Pure Task and run them in the same transaction as they are generated.
- Depends on https://github.com/temporalio/temporal/pull/8399

## Why?
- Currently immediate pure task is the only way for component to run logic that depends on things that are not part of component states (e.g. logger, metrics handler, other util structs)

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -122,8 +122,9 @@ type (
 		// mutation field captures all user state changes (those will be replicated)
 		mutation NodesMutation
 		// systemMutation field captures all cell specific system changes (those will NOT be replicated)
-		systemMutation NodesMutation
-		newTasks       map[any][]taskWithAttributes // component value -> task & attributes
+		systemMutation     NodesMutation
+		newTasks           map[any][]taskWithAttributes // component value -> task & attributes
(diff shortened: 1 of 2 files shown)


chasm/tree.go
```diff
@@ -1176,6 +1178,16 @@ func (n *Node) AddTask(
 	taskAttributes TaskAttributes,
 	task any,
 ) {
+	rt, ok := n.registry.taskFor(task)
+	if ok && rt.isPureTask && taskAttributes.IsImmediate() {
+		// Those tasks will be executed in the current transaction.
+		n.immediatePureTasks[component] = append(n.immediatePureTasks[component], taskWithAttributes{
```
lina-temporal, line 1189: Shouldn't this be on `nodeBase`/root node only? The loop in `executeImmediatePureTasks` only goes through tasks on the root.

lina-temporal: (Ah, `nodeBase` is itself embedded, nvm)

lina-temporal (approved):

PR #8403, opened 2025-09-30
Chasm archetype cluster metadata

awln-temporal:
## What changed?
Adds Archetype search attributes to ClusterMetadata to save all pre-allocated field name to type mappings in the Cassandra backed ClusterMetadata store.

## Why?
Required for CHASM Visibility support. Component authors will register a component with tagged attribute values for querying (List, Count APIs for now), and executions of the registered component will be tagged with these attributes.

Archetype name used for querying is left out of this schema change, may require further discussion.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


rodrigozhou (changes requested):
After some recent discussions, we won't need these changes anymore.

PR #8406, opened 2025-09-30
Add BestEffort field to CompleteHistoryTaskRequest

prathyushpv:
## What changed?
Add a field to CompleteHistoryTaskRequest that suggest that this request can be ignored by persistence layer if needed.
Cassandra's MutableStateTaskStore will ignore the request if BestEffort is set to true.

## Why?
This allows persistence layer to ignore requests if it can negatively impact underlying storage.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/cassandra/mutable_state_task_store.go
+++ b/common/persistence/cassandra/mutable_state_task_store.go
@@ -243,6 +243,10 @@ func (d *MutableStateTaskStore) CompleteHistoryTask(
 	ctx context.Context,
 	request *p.CompleteHistoryTaskRequest,
 ) error {
+	// Ignore the request if it is best effort
+	if request.BestEffort {
+		return nil
+	}
 	switch request.TaskCategory.ID() {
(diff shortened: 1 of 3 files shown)


common/persistence/cassandra/mutable_state_task_store.go
```diff
@@ -243,6 +243,10 @@ func (d *MutableStateTaskStore) CompleteHistoryTask(
 	ctx context.Context,
 	request *p.CompleteHistoryTaskRequest,
 ) error {
+	// Ignore the request if it is best effort
+	if request.BestEffort {
+		return nil
+	}
```
yycptt, line 249: what about sql implementation?

prathyushpv: I thought SQL might not have the tombstone issue from deletions. But there could be other issues from large number of deletion in SQL. Added the same logic there as well. Thanks!

PR #8408, opened 2025-09-30
Added namespace tag to activity payload size metrics.

fretz12:
## What changed?
Added namespace tag to activity payload size metrics.

## Why?
Gives us better searchability when investigating payload size.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/historybuilder/history_builder.go
+++ b/service/history/historybuilder/history_builder.go
@@ -279,13 +279,16 @@ func (b *HistoryBuilder) AddWorkflowTaskFailedEvent(
 func (b *HistoryBuilder) AddActivityTaskScheduledEvent(
 	workflowTaskCompletedEventID int64,
 	command *commandpb.ScheduleActivityTaskCommandAttributes,
+	targetNamespace namespace.Name,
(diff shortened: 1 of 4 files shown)


service/history/historybuilder/history_builder.go
```diff
@@ -279,13 +279,16 @@ func (b *HistoryBuilder) AddWorkflowTaskFailedEvent(
 func (b *HistoryBuilder) AddActivityTaskScheduledEvent(
 	workflowTaskCompletedEventID int64,
 	command *commandpb.ScheduleActivityTaskCommandAttributes,
+	targetNamespace namespace.Name,
```
bergundy: nit: it's not a "target" namespace, it's just the namespace of the workflow that this event belongs to.
Consider renaming to `ns` or something.

bergundy (approved):
Approved with a nit. Thanks!

service/history/historybuilder/history_builder.go
fretz12: Renamed to ns.

PR #8411, opened 2025-10-01
Added Chasm RPC handler foundation and interceptor.

fretz12:
## What changed?
Added Chasm RPC handler foundation and interceptor. Added test as use case demo.

## Why?
The Chasm RPC handler will be used as the abstraction layer for future work. A grpc interceptor was added to intercept requests to process boilerplate code so the handlers don't have to regurgitate it. Please look at the added test as it demonstrates how the RPC handler is registered and a request gets processed.

## How did you test it?
- [X] built
- [X] added new unit test(s)

## Potential risks
There is a [future ticket](https://temporalio.atlassian.net/browse/ACT-69) to refactor all history RPC calls to start using the interceptor as to avoid boilerplate code. This does need a deeper discussion with the team and we also need to figure out how to identify the incoming RPC request so we can extract namespace etc. properly.


chasm/library.go
```diff
@@ -27,6 +29,10 @@ func (UnimplementedLibrary) Tasks() []*RegistrableTask {
 	return nil
 }
 
+// RegisterServices Registers the gRPC calls to the handlers of the library.
+func (UnimplementedLibrary) RegisterServices(_ *grpc.Server) {
```
fretz12, line 33: FYI next phase to be called by Register in the registry

fretz12:
Still working through some proto mockgen issues that's causing a check to fail, but shouldn't affect functionality

chasm/interceptors.go
```diff
@@ -0,0 +1,51 @@
...
+
+var _ grpc.UnaryServerInterceptor = (*ChasmRequestInterceptor)(nil).Intercept
+
+func (i *ChasmRequestInterceptor) Intercept(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (resp interface{}, retError error) {
+	// TODO https://temporalio.atlassian.net/browse/ACT-69 is the work to have existing RPC calls utilize this
```
bergundy: Generally prefer not to put references to internal Jiras in this codebase which is open to the public.

chasm/lib/tests/proto/v1/service.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package temporal.server.chasm.lib.tests.proto.v1;
+
+option go_package = "go.temporal.io/server/chasm/lib/tests/gen/testspb;testspb";
+
+import "chasm/lib/tests/proto/v1/request_response.proto";
+
+service TestService {
+    rpc Test(TestRequest) returns (TestResponse) {}
```
bergundy: Once my PR is merged, you'll want to put a routing directive on this. But let's see what gets merged first and address.

bergundy (commented):
Overall looks great.
You are missing the code that takes all of the registered library handlers and calls them when we create the service.

chasm/interceptors.go
fretz12: oops, removed it

bergundy (approved):

PR #8412, opened 2025-10-01
Optimize WorkflowTask Timeouts

prathyushpv:
## What changed?
Optimize workflow task timeouts by completing them when the workflow task finishes.
This optimization is enabled when the dynamic config system.enableDeleteHistoryTasksOnUpdate is turned on.
It is disabled by default, as it can cause performance degradation in cassandra based persistence.

## Why?
A large fraction of all timer tasks processed is workflow task timeouts. In most of the cases, these are no-ops.
We can complete these tasks when the corresponding workflow task finishes. This will reduce the number of timer tasks processes.

## How did you test it?
- [x] built
- [x] added new unit test(s)


common/dynamicconfig/constants.go
```diff
@@ -1385,6 +1385,12 @@ If value less or equal to 0, will fall back to HistoryPersistenceMaxQPS`,
 Fields: Enabled, RefreshInterval, LatencyThreshold, ErrorThreshold, RateBackoffStepSize, RateIncreaseStepSize, RateMultiMin, RateMultiMax.
 See DynamicRateLimitingParams comments for more details.`,
 	)
+	EnableDeleteTasksOnWorkflowUpdate = NewGlobalBoolSetting(
```
yiminc: consider rename to EnableBestEffortDeleteTaskAfterComplete, so we could apply to activity task as well.

prathyushpv: This name is already general enough. It means we will delete the tasks listed for deletion in MutableState when UpdateWorkflowExecution is called in the persistence manager. We can add any tasks to that list, including activity timeouts.
It makes sense to add BestEffort to the name to make it clear.

common/persistence/execution_manager.go
```diff
@@ -226,6 +230,35 @@ func (m *executionManagerImpl) UpdateWorkflowExecution(
 	}
 }
 
+// deleteHistoryTasks iterates over provided task keys and completes them when the dynamic config
+// history.enableDeleteTasksOnWorkflowUpdate is enabled. Completion is best-effort and failures are logged.
+func (m *executionManagerImpl) deleteHistoryTasks(
```
yycptt, line 235: nit: I thought those logic & the flag won't be in oss, but no strong opinion.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -680,6 +681,18 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskCompletedEvent(
...
+	if workflowTask != nil {
+		if t := workflowTask.ScheduleToStartTimeoutTask; t != nil && !t.InMemory &&
+			t.VisibilityTimestamp.Sub(workflowTask.ScheduledTime) < maxWorkflowTaskTimeoutToDelete {
+			m.ms.DeleteTasks[t.GetCategory()] = append(m.ms.DeleteTasks[t.GetCategory()], t.GetKey())
+		}
+		if t := workflowTask.StartToCloseTimeoutTask; t != nil && !t.InMemory &&
+			t.VisibilityTimestamp.Sub(workflowTask.StartedTime) < maxWorkflowTaskTimeoutToDelete {
+			m.ms.DeleteTasks[t.GetCategory()] = append(m.ms.DeleteTasks[t.GetCategory()], t.GetKey())
+		}
+	}
```
yycptt: I remember there's a deleteWorkflowTask() method in this field, can probably move the logic there and have cases like failed task covered as well.

service/history/interfaces/workflow_task_info.go
```diff
@@ -48,6 +49,11 @@ type WorkflowTaskInfo struct {
 	// - when persisting transient/speculative WFT, the right Build ID is used in the WFT started event
 	// Deprecated.
 	BuildId string
+
+	// References to timeout tasks for this workflow task. These are used to directly complete timeout tasks when the
+	// workflow task completes.
+	ScheduleToStartTimeoutTask *tasks.WorkflowTaskTimeoutTask
+	StartToCloseTimeoutTask    *tasks.WorkflowTaskTimeoutTask
```
yycptt, line 56: When will those fields be cleared?

yycptt: Actually I am bit confused on how this works, is this struct stored in memory somewhere, I am under the impression that it's always generated on the fly.

common/persistence/data_interfaces.go
```diff
@@ -346,7 +346,8 @@ type (
 		NewBufferedEvents         []*historypb.HistoryEvent
 		ClearBufferedEvents       bool
 
-		Tasks map[tasks.Category][]tasks.Task
+		Tasks       map[tasks.Category][]tasks.Task
+		DeleteTasks map[tasks.Category][]tasks.Key
```
yycptt: you may want to explain why this field doesn't exist in persistence_interface.go. Maybe even better to let each persistence implementation decide if they want to honor task deletions. 

Probably should also have BestEffort in the field name or at least comment that.

service/history/interfaces/workflow_task_info.go
prathyushpv: Yeah, I just realized it is generated on the fly. Let me fix that.

common/persistence/data_interfaces.go
prathyushpv: hmm. Do you mean that we add DeleteTasks in persistence_interface.go and let the persistence store decide what to do with it?

service/history/workflow/workflow_task_state_machine.go
prathyushpv: Yes. Made the change. Added this logic for failed workflow task as well.

common/persistence/execution_manager.go
```diff
@@ -226,6 +230,35 @@ func (m *executionManagerImpl) UpdateWorkflowExecution(
...
+	}
+	for category, keys := range toDelete {
+		for _, key := range keys {
+			if err := m.persistence.CompleteHistoryTask(ctx, &CompleteHistoryTaskRequest{
+				ShardID:      shardID,
+				TaskCategory: category,
+				TaskKey:      key,
+				BestEffort:   true,
+			}); err != nil {
+				m.logger.Warn("Failed to delete history task after workflow update",
```
yycptt, line 252: nit: tag shardID and ideally workflow key as well.

service/history/workflow/task_generator.go
```diff
@@ -421,6 +421,8 @@ func (r *TaskGeneratorImpl) GenerateScheduleWorkflowTaskTasks(
 			Version:             workflowTask.Version,
 		}
 		r.mutableState.AddTasks(wttt)
+		key := wttt.GetKey()
+		r.mutableState.SetWorkflowTaskScheduleToStartTimeoutTaskKey(&key)
```
yycptt: task key doesn't exist here, it's only set later in shardContext.

The original approach should work where we store the task struct (pointer), and later in shard context the key will be updated in the same object. it's also very hacky but at least should work.

common/persistence/execution_manager.go
prathyushpv: Added tags.

service/history/workflow/task_generator.go
prathyushpv: Ok. Updated it to store task struct pointer.

PR #8417, opened 2025-10-02
Include failover history when replicating namespace.

robholland:
## What changed?
Include FailoverHistory when replicating a namespace.

## Why?
While we were updating the FailoverHistory when FailoverVersion was changed, we didn't send FailoverHistory on the initial replication task.

## How did you test it?
- [x] built
- [x] added new unit test(s)

---

> [!NOTE]
> Adds propagation of `FailoverHistory` during namespace replication on create and update, with conversion to persistence format and updated tests.
> 
> - **Namespace Replication**:
>   - Include `FailoverHistory` in `ReplicationConfig` when creating a namespace (`replication_task_executor.go`).
>   - On updates where `FailoverVersion` increases, persist `ReplicationConfig.FailoverHistory`.
>   - Add `ConvertFailoverHistoryToPersistenceProto` to map API types to persistence types.
> - **Tests**:
>   - Update/create tests to assert `FailoverHistory` is persisted on initial replication and during failover updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 889f8657180eb3d10242ac23c887079f554ce94b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


robholland:
Note that this won't fix existing passive clusters which have a blank history as it wasn't sent on replication setup. We could add a check alongside the FailoverVersion comparison that FailoverHistory is up to date, but I'm not clear on whether that is worth the trade-off vs just ensuring that all future replicas are correct.

robholland:
Added a check to update FailoverHistory for affected existing replicas.

common/namespace/nsreplication/replication_task_executor.go
```diff
@@ -295,6 +296,9 @@ func (h *taskExecutorImpl) handleNamespaceUpdateReplicationTask(
 		request.Namespace.FailoverVersion = task.GetFailoverVersion()
 		request.Namespace.FailoverNotificationVersion = notificationVersion
 		request.Namespace.ReplicationConfig.FailoverHistory = ConvertFailoverHistoryToPersistenceProto(task.GetFailoverHistory())
+	} else if len(resp.Namespace.ReplicationConfig.FailoverHistory) == 0 && len(task.GetFailoverHistory()) > 0 {
+		recordUpdated = true
+		request.Namespace.ReplicationConfig.FailoverHistory = ConvertFailoverHistoryToPersistenceProto(task.GetFailoverHistory())
```
yux0: I think the change in namespace creation path handles this issue. This logic could be hard to explain as we overwrite the failover history with a lower version. Could we change this to an error log message to see if this ever happen ?

robholland: As discussed on Slack I've removed the out of band update, so just passing failover history on creation now.

PR #8418, opened 2025-10-02
fix: SQLite connection pool resource leak in conn_pool.go

ArchishmanSengupta:
## What changed?
Fixed a `resource leak` in the SQLite connection pool by implementing proper connection cleanup when reference count reaches 0. 

The changes include:
1. Added logger support to the connection pool for error reporting
2. Fixed reference count update bug in the `Allocate` method where the incremented entry wasn't being stored back to the pool
3. Implemented proper db connection cleanup in the `Close` method when reference count drops to 0
4. Added error logging for failed connection closures

## Why?
The previous implementation had a critical resource leak where SQLite db connections were never properly closed. 

The TODO comment indicated this was a known issue that caused connections to persist for the entire application duration, leading to:
1. Memory leaks from unclosed DB connections
2. Potential file descriptor exhaustion
3. Resource waste in long-running applications
4. Risk of "database is locked" errors due to lingering connections
5. The reference count bug in Allocate also meant that shared connections weren't being tracked correctly, potentially causing premature or incorrect cleanup attempts.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/sql/sqlplugin/sqlite/conn_pool.go
+++ b/common/persistence/sql/sqlplugin/sqlite/conn_pool.go
@@ -80,7 +80,7 @@ func (cp *connPool) Close(cfg *config.SQL) {
 
 	e.refCount--
 	cp.pool[dsn] = e
-	
+
 	if e.refCount == 0 {
 		if e.db != nil {
 			if closeErr := e.db.Close(); closeErr != nil {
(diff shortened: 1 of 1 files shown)


ArchishmanSengupta:
@robholland could you please take a look at this?

prathyushpv:
Thanks for your contribution! But in this specific case, we do not want the database connection to close when the reference count reaches 0. This will delete the in-memory sqlite DB and cause data loss. This is intentional that we don't close the connection in this case. 

We close the connection for MySQL and PostgreSQL when the reference count reaches 0.

PR #8420, opened 2025-10-02
try adding eventually

chaptersix:
## What changed?
Added some log statements to the TestMaxBufferedEventsLimit functional test to help debug why the test is failing. I also added require.Eventually blocks to repeatedly check the conditions. 
## Why?
This tests has been very flaky but I can't seem to reproduce the failure in single tests runs or locally. Logs are too verbose to glean anything meaningful from them. 

I also tried visualizing the traces and nothing seemed to stick out. (Could just be me). Hopefully these changes either fix the issue or make it easier to see what's happening.

## How did you test it?
- [x] built

### DIFF
--- a/tests/max_buffered_event_test.go
+++ b/tests/max_buffered_event_test.go
@@ -14,6 +14,7 @@ import (
 	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/workflow"
+	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/payloads"
 	"go.temporal.io/server/tests/testcore"
 )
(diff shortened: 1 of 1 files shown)


tests/max_buffered_event_test.go
```diff
@@ -151,15 +158,21 @@ func (s *MaxBufferedEventSuite) TestBufferedEventsMutableStateSizeLimit() {
...
 		sigCount := 0
+		s.Logger.Info("reading signals",
+			// not the actual workflow ID
+			tag.WorkflowID("TestBufferedEventsMutableStateSizeLimit"))
 		for sigCh.ReceiveAsync(nil) {
 			sigCount++
 		}
+		s.Logger.Info("workflow completed",
+			// not the actual workflow ID
+			tag.WorkflowID("TestBufferedEventsMutableStateSizeLimit"))
```
chaptersix, line 169: @stephanos is there a better way to tag this?

stephanos: Tag is good 👍 Why not make it debug; since we know we have debug logs for tests enabled. This way production stays clean.

PR #8422, opened 2025-10-02
Skip sending newRuninfo when new run is not created in current cluster

xwduan:
## What changed?
Skip sending newRuninfo when new run is not created in current cluster

## Why?
If new run is not created at current cluster, it means the transaction is not happened at current cluster, so the current cluster should no include new run in the current run replication task.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
no risk.

### DIFF
--- a/service/history/replication/sync_state_retriever.go
+++ b/service/history/replication/sync_state_retriever.go
@@ -334,9 +334,7 @@ func (s *SyncStateRetrieverImpl) getNewRunInfo(ctx context.Context, namespaceId 
 		return nil, err
 	}
 	clusterMetadata := s.shardContext.GetClusterMetadata()
-	newRunCluster := clusterMetadata.ClusterNameForFailoverVersion(true, startVersion)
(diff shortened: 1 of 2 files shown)


service/history/replication/sync_state_retriever.go
```diff
@@ -326,6 +326,20 @@ func (s *SyncStateRetrieverImpl) getNewRunInfo(ctx context.Context, namespaceId
...
+	startVersion, err := mutableState.GetStartVersion()
+	if err != nil {
+		return nil, err
+	}
+	clusterMetadata := s.shardContext.GetClusterMetadata()
+	newRunCluster := clusterMetadata.ClusterNameForFailoverVersion(true, startVersion)
+	currentCluster := clusterMetadata.GetCurrentClusterName()
+	if currentCluster != newRunCluster {
+		return nil, nil
+	}
```
yux0, line 339: nit: if !clusterMetadata.IsVersionFromSameCluster(clusterMetadata.GetClusterID(), startVersion) ?

PR #8423, opened 2025-10-02
Removed history handler start wait group.

fretz12:
## What changed?
Removed history handler start wait group.

## Why?
The history handler currently blocks serving RPC handling until the required modules are initialized via a start wait group. However, the RPC handlers aren't registered with the server until after the modules are initialized, making the wait group redundant.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


bergundy (approved):
LGTM since we saw that we don't start the RPC handler until the waitgroup is done anyways.

PR #8424, opened 2025-10-03
[Scheduled Actions] Nexus-based workflow completions (watcher)

lina-temporal:
- Adds the Nexus-based workflow watcher to Scheduler V2.
- Scheduler V2 now maintains a mapping of request IDs -> workflow IDs, recorded within the same transaction as the buffered start, that can be used to identify workflow completion results.
  - Workflow ID generation is moved from within `startWorkflow` to within `SpecProcessor` (earlier, but identical algorithm/results), so that it can be recorded at the same time as the BufferedStart.
- `LastCompletionState`  is added as a CHASM field (separate document) to keep track of payload results.
- A Nexus completion callback is attached to workflows started via Invoker.

### DIFF
--- a/chasm/lib/scheduler/scheduler.go
+++ b/chasm/lib/scheduler/scheduler.go
@@ -296,13 +296,37 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
 	s.Info.OverlapSkipped += result.overlapSkipped
 	s.Info.MissedCatchupWindow += result.missedCatchupWindow
 
-	if len(result.starts) > 0 {
-		s.Info.RecentActions = util.SliceTail(append(s.Info.RecentActions, result.starts...), recentActionCount)
+	// Filter any actions being recorded that may have already been recorded/completed.
+	newActions := util.FilterSlice(result.starts, func(incoming *schedulepb.ScheduleActionResult) bool {
+		wid := incoming.StartWorkflowResult.WorkflowId
+		recentIdx := slices.IndexFunc(s.Info.RecentActions, func(existing *schedulepb.ScheduleActionResult) bool {
+			recorded := existing.StartWorkflowResult.WorkflowId == wid
+			return recorded
+		})
+
+		return recentIdx < 0
+	})
+	if len(newActions) > 0 {
+		s.Info.RecentActions = util.SliceTail(append(s.Info.RecentActions, newActions...), recentActionCount)
 	}
 
+	// Update RunningWorkflows.
 	for _, start := range result.starts {
-		if start.StartWorkflowResult != nil {
+		completed := start.StartWorkflowStatus >= enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED
+
+		if !completed {
+			// Add running workflows to the list (duplicates are acceptable for ALLOW_ALL policy)
 			s.Info.RunningWorkflows = append(s.Info.RunningWorkflows, start.StartWorkflowResult)
+		} else {
+			// Remove completed workflows by WorkflowId+RunId
+			wid := start.StartWorkflowResult.WorkflowId
+			rid := start.StartWorkflowResult.RunId
+			runningIdx := slices.IndexFunc(s.Info.RunningWorkflows, func(wfe *commonpb.WorkflowExecution) bool {
+				return wfe.WorkflowId == wid && wfe.RunId == rid
+			})
+			if runningIdx >= 0 {
+				s.Info.RunningWorkflows = slices.Delete(s.Info.RunningWorkflows, runningIdx, runningIdx+1)
+			}
 		}
 	}
 }
@@ -398,17 +422,14 @@ func (s *Scheduler) recordCompletedAction(
 	})
 
 	// Update the RecentActions entry's status.
-	found := false
-	for _, action := range s.Info.RecentActions {
-		if action.StartWorkflowResult.WorkflowId == workflowID {
-			action.StartWorkflowStatus = workflowStatus
-			found = true
-			break
-		}
-	}
+	idx := slices.IndexFunc(s.Info.RecentActions, func(action *schedulepb.ScheduleActionResult) bool {
+		return action.StartWorkflowResult.WorkflowId == workflowID
+	})
 
-	// If we didn't find an entry in RecentActions, add one.
-	if !found {
+	if idx >= 0 {
+		s.Info.RecentActions[idx].StartWorkflowStatus = workflowStatus
+	} else {
+		// If we didn't find an entry in RecentActions, add one.
 		if scheduleTime.IsZero() {
 			// We're completing a workflow that wasn't in BufferedStarts, RunningWorkflows,
 			// or RecentActions, but *did* have a request ID entry. That shouldn't be possible.
--- a/chasm/lib/scheduler/scheduler_nexus_completion_test.go
+++ b/chasm/lib/scheduler/scheduler_nexus_completion_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/suite"
+	"github.com/stretchr/testify/require"
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	failurepb "go.temporal.io/api/failure/v1"
@@ -15,34 +15,103 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-type schedulerNexusCompletionSuite struct {
-	schedulerSuite
-}
-
-func TestSchedulerNexusCompletionSuite(t *testing.T) {
-	suite.Run(t, &schedulerNexusCompletionSuite{})
-}
-
 type nexusCompletionTestCase struct {
-	SetupInvoker   func(*scheduler.Invoker)
-	SetupScheduler func(*scheduler.Scheduler)
-	Completion     *persistencespb.ChasmNexusCompletion
-
-	ExpectPaused          bool
-	ExpectStatus          enumspb.WorkflowExecutionStatus
-	ExpectInRecentActions bool
-	ExpectNoOp            bool // For idempotent case
-	ValidateInvoker       func(*testing.T, *scheduler.Invoker)
+	name              string
+	setupInvoker      func(*scheduler.Invoker)
+	setupScheduler    func(*scheduler.Scheduler)
+	completion        *persistencespb.ChasmNexusCompletion
+	expectPaused      bool
+	expectStatus      enumspb.WorkflowExecutionStatus
+	expectNoOp        bool
+	validateInvoker   func(*testing.T, *scheduler.Invoker)
+	validateScheduler func(*testing.T, *scheduler.Scheduler)
 }
 
-func (s *schedulerNexusCompletionSuite) TestHandleNexusCompletion_Success() {
-	s.runTestCase(&nexusCompletionTestCase{
-		SetupInvoker: func(invoker *scheduler.Invoker) {
+func executeNexusCompletion(t *testing.T, tc nexusCompletionTestCase) {
+	sched, ctx, node := setupSchedulerForTest(t)
+
+	invoker, err := sched.Invoker.Get(ctx)
+	require.NoError(t, err)
+
+	if tc.setupInvoker != nil {
+		tc.setupInvoker(invoker)
+	}
+	if tc.setupScheduler != nil {
+		tc.setupScheduler(sched)
+	}
+
+	initialRunningWorkflows := len(sched.Info.RunningWorkflows)
+	initialRecentActions := len(sched.Info.RecentActions)
+	initialLastCompletion, err := sched.LastCompletionResult.Get(ctx)
+	require.NoError(t, err)
+
+	err = sched.HandleNexusCompletion(ctx, tc.completion)
+	require.NoError(t, err)
+
+	_, err = node.CloseTransaction()
+	require.NoError(t, err)
+
+	if tc.expectNoOp {
+		require.Len(t, sched.Info.RunningWorkflows, initialRunningWorkflows)
+		require.Len(t, sched.Info.RecentActions, initialRecentActions)
+		currentLastCompletion, _ := sched.LastCompletionResult.Get(ctx)
+		require.Equal(t, initialLastCompletion, currentLastCompletion)
+		return
+	}
+
+	lastCompletion, err := sched.LastCompletionResult.Get(ctx)
+	require.NoError(t, err)
+	require.NotNil(t, lastCompletion)
+
+	if tc.completion.GetSuccess() != nil {
+		require.NotNil(t, lastCompletion.GetSuccess())
+	} else if tc.completion.GetFailure() != nil {
+		require.NotNil(t, lastCompletion.GetFailure())
+	}
+
+	require.Equal(t, tc.expectPaused, sched.Schedule.State.Paused)
+	if tc.expectPaused {
+		require.NotEmpty(t, sched.Schedule.State.Notes)
+		require.Contains(t, sched.Schedule.State.Notes, "wf-1")
+	}
+
+	for _, wf := range sched.Info.RunningWorkflows {
+		require.NotEqual(t, "wf-1", wf.WorkflowId)
+	}
+
+	if tc.expectStatus != enumspb.WORKFLOW_EXECUTION_STATUS_UNSPECIFIED {
+		found := false
+		for _, action := range sched.Info.RecentActions {
+			if action.StartWorkflowResult.WorkflowId == "wf-1" {
+				require.Equal(t, tc.expectStatus, action.StartWorkflowStatus)
+				found = true
+				break
+			}
+		}
+		require.True(t, found)
+	}
+
+	require.Empty(t, invoker.WorkflowID(tc.completion.RequestId))
+
+	if tc.validateInvoker != nil {
+		tc.validateInvoker(t, invoker)
+	}
+	if tc.validateScheduler != nil {
+		tc.validateScheduler(t, sched)
+	}
+}
+
+// TestHandleNexusCompletion_Success verifies that a successful workflow completion
+// is properly recorded with the success payload and workflow status is updated.
+func TestHandleNexusCompletion_Success(t *testing.T) {
+	tc := nexusCompletionTestCase{
+		name: "successful completion",
+		setupInvoker: func(invoker *scheduler.Invoker) {
 			invoker.RequestIdToWorkflowId = map[string]string{
 				"req-1": "wf-1",
 			}
 		},
-		SetupScheduler: func(sched *scheduler.Scheduler) {
+		setupScheduler: func(sched *scheduler.Scheduler) {
 			sched.Info.RunningWorkflows = []*commonpb.WorkflowExecution{
 				{WorkflowId: "wf-1", RunId: "run-1"},
 			}
@@ -56,27 +125,31 @@ func (s *schedulerNexusCompletionSuite) TestHandleNexusCompletion_Success() {
 				},
 			}
 		},
-		Completion: &persistencespb.ChasmNexusCompletion{
+		completion: &persistencespb.ChasmNexusCompletion{
 			RequestId: "req-1",
 			Outcome: &persistencespb.ChasmNexusCompletion_Success{
 				Success: &commonpb.Payload{Data: []byte("success-data")},
 			},
 			CloseTime: timestamppb.New(time.Now()),
 		},
-		ExpectPaused:          false,
-		ExpectStatus:          enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED,
-		ExpectInRecentActions: true,
-	})
+		expectPaused: false,
+		expectStatus: enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED,
+	}
+
+	executeNexusCompletion(t, tc)
 }
 
-func (s *schedulerNexusCompletionSuite) TestHandleNexusCompletion_Failure() {
-	s.runTestCase(&nexusCompletionTestCase{
-		SetupInvoker: func(invoker *scheduler.Invoker) {
+// TestHandleNexusCompletion_Failure verifies that a failed workflow completion
+// is properly recorded with the failure payload and workflow status is updated.
+func TestHandleNexusCompletion_Failure(t *testing.T) {
+	tc := nexusCompletionTestCase{
+		name: "failed completion",
+		setupInvoker: func(invoker *scheduler.Invoker) {
 			invoker.RequestIdToWorkflowId = map[string]string{
 				"req-1": "wf-1",
 			}
 		},
-		SetupScheduler: func(sched *scheduler.Scheduler) {
+		setupScheduler: func(sched *scheduler.Scheduler) {
 			sched.Info.RunningWorkflows = []*commonpb.WorkflowExecution{
 				{WorkflowId: "wf-1", RunId: "run-1"},
 			}
@@ -90,7 +163,7 @@ func (s *schedulerNexusCompletionSuite) TestHandleNexusCompletion_Failure() {
 				},
 			}
 		},
-		Completion: &persistencespb.ChasmNexusCompletion{
+		completion: &persistencespb.ChasmNexusCompletion{
 			RequestId: "req-1",
 			Outcome: &persistencespb.ChasmNexusCompletion_Failure{
 				Failure: &failurepb.Failure{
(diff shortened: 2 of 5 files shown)


chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +306,143 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
 }
+
+var _ chasm.NexusCompletionHandler = &Scheduler{}
+
+func executionStatusFromFailure(failure *failurepb.Failure) enumspb.WorkflowExecutionStatus {
+	switch failure.FailureInfo.(type) {
+	case *failurepb.Failure_CanceledFailureInfo:
+		return enumspb.WORKFLOW_EXECUTION_STATUS_CANCELED
+	case *failurepb.Failure_TimeoutFailureInfo:
+		return enumspb.WORKFLOW_EXECUTION_STATUS_TERMINATED
```
lina-temporal: this is a typo

service/frontend/workflow_handler.go
```diff
@@ -5250,7 +5250,7 @@ func (wh *WorkflowHandler) validateCallbackURL(ns namespace.Name, rawURL string)
 	if u.Scheme == "temporal" {
 		return nil
 	}
-	if !(u.Scheme == "http" || u.Scheme == "https") {
+	if u.Scheme != "http" && u.Scheme != "https" {
```
lina-temporal: unintentional, just a merge casualty from a bunch of stacked changes

chasm/lib/scheduler/backfiller_tasks_test.go
```diff
@@ -241,7 +240,7 @@ func (s *backfillerTasksSuite) runTestCase(c *backfillTestCase) {
 	// Either type of request will spawn a Backfiller and schedule an immediate pure task.
 	_, err = s.node.CloseTransaction()
 	s.NoError(err)
-	s.True(s.hasTask(&tasks.ChasmTaskPure{}, chasm.TaskScheduledTimeImmediate))
+	// s.True(s.hasTask(&tasks.ChasmTaskPure{}, chasm.TaskScheduledTimeImmediate))
```
bergundy:
```suggestion
```

chasm/lib/scheduler/invoker.go
```diff
@@ -169,6 +174,57 @@ func (i *Invoker) recordExecuteResult(ctx chasm.MutableContext, result *executeR
...
+func (i *Invoker) recordCompletedAction(
+	ctx chasm.MutableContext,
+	closeTime time.Time,
+	requestID string,
+) (scheduleTime time.Time) {
+	// Clean up the RequestID map, since we're done with the request.
+	delete(i.RequestIdToWorkflowId, requestID)
+
+	// Check if the action is still in BufferedStarts, clear it out.
+	i.BufferedStarts = slices.DeleteFunc(i.BufferedStarts, func(start *schedulespb.BufferedStart) bool {
```
bergundy: nit: `DeleteFunc` will not break on first match. You could change this to `IndexFunc` and `Delete`.

chasm/lib/scheduler/invoker.go
```diff
@@ -169,6 +174,57 @@ func (i *Invoker) recordExecuteResult(ctx chasm.MutableContext, result *executeR
...
+	})
+
+	// If we don't have any starts buffered, recording actions is a no-op from
+	// Invoker POV.
+	if len(i.BufferedStarts) < 1 {
+		return
+	}
+
+	// Update DesiredTime on the first start for metrics.
+	i.BufferedStarts[0].DesiredTime = timestamppb.New(closeTime)
```
bergundy: Not sure I follow this logic, can you add some more information in this comment?

chasm/lib/scheduler/invoker_execute_task_test.go
```diff
@@ -92,6 +92,10 @@ func (s *invokerExecuteTaskSuite) TestExecuteTask_Basic() {
 	s.mockFrontendClient.EXPECT().
 		StartWorkflowExecution(gomock.Any(), gomock.Any()).
 		Times(2).
+		// Do(func(req *workflowservice.StartWorkflowExecutionRequest) {
+		// 	// Expect callbacks to be set.
+		// 	s.NotEmpty(req.CompletionCallbacks)
+		// }).
```
bergundy:
```suggestion
```

chasm/lib/scheduler/invoker.go
```diff
@@ -169,6 +174,57 @@ func (i *Invoker) recordExecuteResult(ctx chasm.MutableContext, result *executeR
...
+
+	// We add an immediate ProcessBufferTask if we have any queued-up starts, since
+	// a completed action will bump the next buffered action to "ready". The regular
+	// addTasks logic won't apply, since getEligibleBufferedStarts only looks at the
+	// current state of the BufferedStarts queue (it doesn't recompute eligible).
+	ctx.AddTask(i, chasm.TaskAttributes{
```
bergundy: I don't understand the logic here but also it seems to complicate the flow that these tasks can be scheduled in multiple places compared to potentially having a single location where these decisions are made.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -513,38 +537,55 @@ func (e *InvokerExecuteTaskExecutor) startWorkflow(
...
 	request := &workflowservice.StartWorkflowExecutionRequest{
-		Namespace:                scheduler.Namespace,
-		WorkflowId:               workflowID,
-		WorkflowType:             requestSpec.WorkflowType,
-		TaskQueue:                requestSpec.TaskQueue,
-		Input:                    requestSpec.Input,
-		WorkflowExecutionTimeout: requestSpec.WorkflowExecutionTimeout,
-		WorkflowRunTimeout:       requestSpec.WorkflowRunTimeout,
-		WorkflowTaskTimeout:      requestSpec.WorkflowTaskTimeout,
+		CompletionCallbacks:      []*commonpb.Callback{callback},
```
bergundy, line 542: You're going to want to add links here too, not blocking this PR but it's something to discuss with the team.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -513,38 +537,55 @@ func (e *InvokerExecuteTaskExecutor) startWorkflow(
...
+		request.LastCompletionResult = &commonpb.Payloads{
+			Payloads: []*commonpb.Payload{outcome.Success},
+		}
+	}
+
 	result, err := e.FrontendClient.StartWorkflowExecution(ctx, request)
 	if err != nil {
 		return nil, err
 	}
+	actualStartTime := time.Now()
```
bergundy, line 575: Technically this may not be the actual workflow start time if this is a retried request but you don't have a way to know this. I think you still would want to use some time utility from CHASM and never use `time.Time` so we can support time skipping in the framework later.

chasm/lib/scheduler/invoker.go
```diff
@@ -41,6 +42,10 @@ func NewInvoker(ctx chasm.MutableContext, scheduler *Scheduler) *Invoker {
 func (i *Invoker) EnqueueBufferedStarts(ctx chasm.MutableContext, starts []*schedulespb.BufferedStart) {
 	i.BufferedStarts = append(i.BufferedStarts, starts...)
 
+	for _, start := range starts {
+		i.RequestIdToWorkflowId[start.RequestId] = start.WorkflowId
```
bergundy, line 46: nit: Probably not worth keeping the map if you can search the array. The storage of this additional map is likely more costly than the bit of CPU to iterate through the array.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -292,10 +309,17 @@ func (e *InvokerExecuteTaskExecutor) startWorkflows(
 			break
 		}
 
+		// Check if this start is already in RecentActions. If so, we crashed
+		// after starting a workflow, but before recording the result.
+		if scheduler.isActionCompleted(start.WorkflowId) {
+			logger.Warn("skipping already-completed workflow", tag.WorkflowID(start.WorkflowId))
```
bergundy: This is expected to happen

```suggestion
			logger.Info("skipping already-completed workflow", tag.WorkflowID(start.WorkflowId))
```

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -292,10 +309,17 @@ func (e *InvokerExecuteTaskExecutor) startWorkflows(
...
+			continue
+		}
+
 		// Run all starts concurrently.
 		newCtx := ctx.Clone()
 		wg.Go(func() {
```
bergundy, line 321: nit: You can use MapConcurrent here if you structure your result object to be a slice of `(startResult, error)` "tuples": https://github.com/temporalio/temporal/blob/ae5720ac1bd207e1f5f7707633607e970a2c3879/common/util/util.go#L82

chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -77,3 +84,12 @@ message BackfillerState {
     // needs to back off before retrying to fill.
     int64 attempt = 8;
 }
+
+// CHASM scheduler retains the payload data for the last completed workflow
+// (either failure or success). 
+message LastCompletionState {
```
bergundy: nit: `LastCompletionResult`?

chasm/lib/scheduler/scheduler.go
```diff
@@ -25,6 +30,10 @@ type Scheduler struct {
 	// the scheduler tree.
 	*schedulerpb.SchedulerState
 
+	// Last success/failure payloads, stored on this separate data node
+	// to minimize write traffic.
+	LastCompletionState chasm.Field[*schedulerpb.LastCompletionState]
```
bergundy: Looks like you always mutate other scheduler fields when you update this state, doesn't seem like this is saving any write traffic but I agree that payloads should typically live in their own dedicated fields.
But if you do this for outputs, do it for inputs too.

chasm/lib/scheduler/scheduler.go
```diff
@@ -170,6 +180,12 @@ func (s *Scheduler) getCompiledSpec(specBuilder *scheduler.SpecBuilder) (*schedu
 	return s.compiledSpec, nil
 }
 
+// GetWorkflowID returns the Workflow ID given as part of the request spec.
+// During start generation, nominal time is suffixed to this ID.
+func (s *Scheduler) GetWorkflowID() string {
```
bergundy: Don't name getters in Go with a `Get` prefix: https://go.dev/doc/effective_go#Getters.

chasm/lib/scheduler/invoker.go
```diff
@@ -169,6 +174,57 @@ func (i *Invoker) recordExecuteResult(ctx chasm.MutableContext, result *executeR
 	i.addTasks(ctx)
 }
 
+// GetWorkflowID returns the workflow ID associated with the given request, or an
+// empty string if not found.
+func (i *Invoker) GetWorkflowID(requestID string) string {
```
bergundy: Don't name getters in Go with a `Get` prefix: https://go.dev/doc/effective_go#Getters.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +306,143 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+// recordCompletedAction ensures that the given action is recorded in
+// RecentActions and cleaned up from other state.
+func (s *Scheduler) recordCompletedAction(
+	ctx chasm.MutableContext,
+	scheduleTime time.Time,
+	workflowID string,
+	workflowStatus enumspb.WorkflowExecutionStatus,
+) {
+	// Clear out closed workflows from RunningWorkflows.
+	s.Info.RunningWorkflows = slices.DeleteFunc(s.Info.RunningWorkflows, func(wf *commonpb.WorkflowExecution) bool {
```
bergundy, line 419: Why do you need to duplicate `RunningWorkflows` and `invoker.BufferedStarts`? Seems like those can be collapsed into a single representation.

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +306,143 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+		// We don't evaluate RunId here, since the action may have retried.
+		return wf.WorkflowId == workflowID
+	})
+
+	// Update the RecentActions entry's status.
+	found := false
+	for _, action := range s.Info.RecentActions {
```
bergundy: I wonder if it'd be worth putting the recent actions on the invoker and avoiding mutating the scheduler at all.

bergundy: nit: `slices.IndexFunction` is more readable here IMHO.

chasm/lib/scheduler/scheduler_nexus_completion_test.go
```diff
@@ -0,0 +1,318 @@
...
+				{
+					StartWorkflowResult: &commonpb.WorkflowExecution{
+						WorkflowId: "wf-1",
+						RunId:      "run-1",
+					},
+					StartWorkflowStatus: enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING,
+				},
+			}
+		},
+		Completion: &persistencespb.ChasmNexusCompletion{
```
bergundy: Would have been better if this didn't have `Chasm` in the name (mentioned in the other PR too).

chasm/lib/scheduler/scheduler_nexus_completion_test.go
```diff
@@ -0,0 +1,318 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	failurepb "go.temporal.io/api/failure/v1"
+	schedulepb "go.temporal.io/api/schedule/v1"
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	"go.temporal.io/server/chasm/lib/scheduler"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type schedulerNexusCompletionSuite struct {
```
bergundy: We agreed as a team not to add more test suites for unit tests.

chasm/lib/scheduler/invoker_tasks.go
bergundy: Will you be making up the start time for this already-completed workflow if needed?

chasm/lib/scheduler/scheduler.go
bergundy, line 419: In the complete-before-start case, you'll be appending to `RunningWorkflows` and never resolving those running workflows.

bergundy: You need better handling plus tests for this edge case.

bergundy: Looks like you just need to change the append in `recordActionResult` to first verify there's no workflow with that ID that is already completed and only delete from `RunningWorkflows` if you have both a start and a close.
If the workflow is already in `RunningWorkflows` on start, you'll want to delete it if it is already completed.

But really if you see my comment above, I think it's a mistake to duplicate the invoker buffered starts into the scheduler fields.

chasm/lib/scheduler/scheduler_nexus_completion_test.go
```diff
@@ -0,0 +1,318 @@
...
+			},
+			CloseTime: timestamppb.New(time.Now()),
+		},
+		ExpectPaused:          false,
+		ExpectStatus:          enumspb.WORKFLOW_EXECUTION_STATUS_CANCELED,
+		ExpectInRecentActions: true,
+	})
+}
+
+func (s *schedulerNexusCompletionSuite) TestHandleNexusCompletion_CompletionBeforeStart() {
```
bergundy: Doesn't cover the case I mentioned above.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: > Technically this may not be the actual workflow start time if this is a retried request but you don't have a way to know this. 

I don't follow. Retried in the sense that the client here retries? Why would this not be accurate in that case?

> I think you still would want to use some time utility from CHASM and never use time.Time so we can support time skipping in the framework later.

This isn't run within a CHASM transaction, so I can't use CHASM's utility, unless I were to do an arbitrary read of a component.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: Yes:

```
			ActualTime:   timestamppb.New(scheduleTime), // best guess, as we're recording complete before start was recorded
```

chasm/lib/scheduler/invoker.go
lina-temporal: Yeah, will update.

chasm/lib/scheduler/invoker.go
lina-temporal: Yeah, taking a second look, this can still go through the regular Invoker.addTasks flow, since a completed action will only affect starts at `Attempt = 0` (pending to kick off) anyways. Simplified. I also dropped the early return since it wasn't saving much.

chasm/lib/scheduler/invoker.go
lina-temporal: Yeah, with the planned rework of `BufferedStarts`, I won't need this separate map anyways.

chasm/lib/scheduler/invoker.go
lina-temporal: Fixed.

chasm/lib/scheduler/proto/v1/message.proto
lina-temporal: Sure, fixed.

chasm/lib/scheduler/invoker.go
lina-temporal: Fixed.

chasm/lib/scheduler/invoker_tasks.go
lina-temporal: I liked this and I took a look at switching it over, but `MapConcurrent` will fail fast after the first error, which won't work here.

chasm/lib/scheduler/scheduler.go
lina-temporal: I think this'll change when we collapse `RecentActions`/`RunningWorkflows` into `BufferedStarts` on Invoker.

chasm/lib/scheduler/scheduler.go
lina-temporal: (we've talked offline) Yeah, I agree, but I'd rather make the change to unify them separately.

lina-temporal: Yeah, will fix this case and add a test. I agree on simplifying BufferedStarts, though, and plan to fast follow on that (right after the API handler PR).

chasm/lib/scheduler/scheduler_nexus_completion_test.go
lina-temporal: Added a bunch of new tests for specifically recording completed actions.

chasm/lib/scheduler/scheduler_nexus_completion_test.go
lina-temporal: Fixed

chasm/lib/scheduler/scheduler_nexus_completion_test.go
lina-temporal: I'll open a PR after to rename it.

chasm/lib/scheduler/invoker.go
bergundy: Okay, I will wait for the refactor in a separate PR.

chasm/lib/scheduler/invoker.go
```diff
@@ -169,6 +174,58 @@ func (i *Invoker) recordExecuteResult(ctx chasm.MutableContext, result *executeR
...
+	// Update DesiredTime on the first pending start for metrics. DesiredTime is used
+	// to drive action latency between buffered starts (the time it takes between
+	// completing one start and kicking off the next). We set that on the first start
+	// pending execution.
+	idx = slices.IndexFunc(i.BufferedStarts, func(start *schedulespb.BufferedStart) bool {
+		return start.Attempt == 0
+	})
+	if idx >= 0 {
+		i.BufferedStarts[idx].DesiredTime = timestamppb.New(closeTime)
+	}
```
bergundy, line 220: I admit to not following this logic completely but I skimmed the scheduler workflow implementation and I _think_ the logic is copied over from there. It's a bit confusing because I think the use of buffered starts is different there and it looks at the first buffered start (index 0).

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -140,6 +142,19 @@ func (e *InvokerExecuteTaskExecutor) Execute(
 				compiledSpec:       s.compiledSpec,
 			}
 
+			lcs, err := s.LastCompletionResult.Get(ctx)
+			if err != nil {
+				return struct{}{}, err
```
bergundy, line 147: I suggested creating a `type NoValue = *struct{}` in the `chasm` package, like we did in `nexus`.
I feel like it's nicer to use `func() (chasm.NoValue, error) { return nil, err }` than the empty struct approach.
Alternatively we can add a `ReadComponentRetErr` helper. But feel like the former option is nicer.
Anyways, definitely not blocking the PR, just something to consider for the future.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -140,6 +142,19 @@ func (e *InvokerExecuteTaskExecutor) Execute(
...
+			}
+			lastCompletionState = common.CloneProto(lcs)
```
bergundy, line 149: I think you can avoid cloning if this state is immutable but probably better safe than sorry 🤷

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -140,6 +142,19 @@ func (e *InvokerExecuteTaskExecutor) Execute(
 				compiledSpec:       s.compiledSpec,
 			}
 
+			lcs, err := s.LastCompletionResult.Get(ctx)
```
bergundy, line 145: I see you call this last completion result and state interchangeably, consider settling on a single term.

chasm/lib/scheduler/invoker_tasks.go
bergundy: It works:

> if you structure your result object to be a slice of (startResult, error) "tuples"

chasm/lib/scheduler/invoker_tasks.go
bergundy: > I don't follow. Retried in the sense that the client here retries? Why would this not be accurate in that case?

If the history node crashes or you fail to commit the start result and have to reprocess this task, you may have already started this workflow in a previous attempt. To get the actual start time, we would need to return it in the `StartWorkflowExecutionResponse`.

> This isn't run within a CHASM transaction, so I can't use CHASM's utility, unless I were to do an arbitrary read of a component.

Hmm... yeah, the "right" thing to do here would be to use a timestamp returned in the start response if one was available. Since we don't have that you could use the framework's time when you record the result back into the component's state. You can ignore this for now though, it's not blocking the PR but please track.

chasm/lib/scheduler/scheduler.go
```diff
@@ -280,13 +296,174 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+			if runningIdx >= 0 {
+				s.Info.RunningWorkflows = slices.Delete(s.Info.RunningWorkflows, runningIdx, runningIdx+1)
+			}
 		}
 	}
 }
+
+var _ chasm.NexusCompletionHandler = &Scheduler{}
+
+func executionStatusFromFailure(failure *failurepb.Failure) enumspb.WorkflowExecutionStatus {
```
bergundy, line 336: You're missing terminated.

chasm/lib/scheduler/scheduler.go
```diff
@@ -280,13 +296,174 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	}
+
+	// Handle last completed/failed status and payloads.
+	//
+	// TODO - also record payload sizes once we have metrics wired into CHASM context.
+	var wfStatus enumspb.WorkflowExecutionStatus
+	switch outcome := info.Outcome.(type) {
+	case *persistencespb.ChasmNexusCompletion_Failure:
+		wfStatus = executionStatusFromFailure(outcome.Failure)
+		s.LastCompletionResult = chasm.NewDataField(ctx, &schedulerpb.LastCompletionResult{
```
bergundy, line 372: If there are multiple concurrent runs, how do you decide what's the "last completion"? What if there's a backfill? What if there's overlap and the first started execution completed after the second?

chasm/lib/scheduler/scheduler.go
```diff
@@ -280,13 +296,174 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+		})
+	case *persistencespb.ChasmNexusCompletion_Success:
+		wfStatus = enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED
+		s.LastCompletionResult = chasm.NewDataField(ctx, &schedulerpb.LastCompletionResult{
+			Outcome: &schedulerpb.LastCompletionResult_Success{
+				Success: outcome.Success,
+			},
+		})
+	default:
+		wfStatus = enumspb.WORKFLOW_EXECUTION_STATUS_FAILED
```
bergundy, line 385: Why is this the default? I would fail with a bad request error.

chasm/lib/scheduler/scheduler.go
```diff
@@ -280,13 +296,174 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	}
+
+	// Handle pause-on-failure.
+	if wfStatus != enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED &&
+		s.Schedule.Policies.PauseOnFailure && !s.Schedule.State.Paused {
```
bergundy, line 390: Same here, do you always want to pause? Even if the workflow is a backfill or not the most recent one?

chasm/lib/scheduler/invoker.go
lina-temporal: Yeah, it's similar from there. I had to update it here because our BufferedStarts are already more stateful than in V1 (namely, `Attempt = 0` means they're pending..). So this effectively does the same thing as V1, yeah. 

Since it doesn't make sense in an `ALLOW_ALL` context, maybe we can just lift this to a field onto Invoker itself (single field), and not touch BufferedStarts at all..

chasm/lib/scheduler/scheduler.go
lina-temporal: At the moment, and this includes for V1, it simply means 'the last thing we recorded a result for'. That even includes backfill results, intertwined with automation results, AFAICT, so.. yeah, it's sort of concurrently strange. Something I'm open to changing.. a bit surprised we haven't seen feedback on this.

chasm/lib/scheduler/scheduler.go
lina-temporal: Intentional - terminated is hidden from users (at least, per V1 :) ).

chasm/lib/scheduler/scheduler.go
lina-temporal: Yeah, you're right. Will fix in a small follow-up.

PR #8426, opened 2025-10-03
Customizable serialization

stephanos:
## What changed?

Added `TEMPORAL_TEST_DATA_ENCODING` to change DataBlob encoding from "proto3" to "json".

## Why?

Observability and debugging. The ability to see payloads decoded in debugger and OTEL traces is valuable.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

End users should never use this. The env var name therefore includes `TEST_`.

### DIFF
--- a/api/matchingservice/v1/request_response.go-helpers.pb.go
+++ b/api/matchingservice/v1/request_response.go-helpers.pb.go
@@ -560,6 +560,80 @@ func (this *CancelOutstandingPollResponse) Equal(that interface{}) bool {
 	return proto.Equal(this, that1)
 }
 
+// Marshal an object of type CancelOutstandingWorkerPollsRequest to the protobuf v3 wire format
+func (val *CancelOutstandingWorkerPollsRequest) Marshal() ([]byte, error) {
+	return proto.Marshal(val)
+}
+
+// Unmarshal an object of type CancelOutstandingWorkerPollsRequest from the protobuf v3 wire format
+func (val *CancelOutstandingWorkerPollsRequest) Unmarshal(buf []byte) error {
+	return proto.Unmarshal(buf, val)
+}
+
+// Size returns the size of the object, in bytes, once serialized
+func (val *CancelOutstandingWorkerPollsRequest) Size() int {
+	return proto.Size(val)
+}
+
+// Equal returns whether two CancelOutstandingWorkerPollsRequest values are equivalent by recursively
+// comparing the message's fields.
+// For more information see the documentation for
+// https://pkg.go.dev/google.golang.org/protobuf/proto#Equal
+func (this *CancelOutstandingWorkerPollsRequest) Equal(that interface{}) bool {
+	if that == nil {
+		return this == nil
+	}
+
+	var that1 *CancelOutstandingWorkerPollsRequest
+	switch t := that.(type) {
+	case *CancelOutstandingWorkerPollsRequest:
+		that1 = t
+	case CancelOutstandingWorkerPollsRequest:
+		that1 = &t
+	default:
+		return false
+	}
+
+	return proto.Equal(this, that1)
+}
+
+// Marshal an object of type CancelOutstandingWorkerPollsResponse to the protobuf v3 wire format
+func (val *CancelOutstandingWorkerPollsResponse) Marshal() ([]byte, error) {
+	return proto.Marshal(val)
+}
+
+// Unmarshal an object of type CancelOutstandingWorkerPollsResponse from the protobuf v3 wire format
+func (val *CancelOutstandingWorkerPollsResponse) Unmarshal(buf []byte) error {
+	return proto.Unmarshal(buf, val)
+}
+
+// Size returns the size of the object, in bytes, once serialized
+func (val *CancelOutstandingWorkerPollsResponse) Size() int {
+	return proto.Size(val)
+}
+
+// Equal returns whether two CancelOutstandingWorkerPollsResponse values are equivalent by recursively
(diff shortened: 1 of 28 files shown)


yycptt:
Is this PR same as https://github.com/temporalio/temporal/pull/8222?

stephanos:
@yycptt I left the actual customization in this one and will rebase it once https://github.com/temporalio/temporal/pull/8222 is merged - that one only contains the refactorings now. The bulk of the changes are in https://github.com/temporalio/temporal/pull/8222 so it's easier to complete, merge and review them separately IMO.

temporal/server_test.go
```diff
@@ -38,45 +50,114 @@ func TestNewServerWithOTEL(t *testing.T) {
...
 }
 
-func startAndStopServer(t *testing.T) {
+// TestNewServerWithJSONEncoding verifies that NewServer works when JSON encoding is enabled.
+func TestNewServerWithJSONEncoding(t *testing.T) {
+	t.Setenv(serialization.SerializerDataEncodingEnvVar, enumspb.ENCODING_TYPE_JSON.String())
+	runAndTestServer(t)
+}
+
+func runAndTestServer(t *testing.T) {
```
stephanos, line 68: I expanded the test to actually process a workflow in order to verify things are working end-to-end with JSON encoding.

temporal/server_test.go
```diff
@@ -138,9 +228,13 @@ func (d *errorLogDetector) logUnexpected(operation string, msg string, tags []ta
...
 }
 
-func (d *errorLogDetector) Debug(string, ...tag.Tag) {}
+func (d *errorLogDetector) Debug(msg string, tags ...tag.Tag) {
+	d.logger.Debug(msg, tags...)
+}
 
-func (d *errorLogDetector) Info(string, ...tag.Tag) {}
+func (d *errorLogDetector) Info(msg string, tags ...tag.Tag) {
+	d.logger.Info(msg, tags...)
```
stephanos, line 236: It was difficult to debug issues without the server logs being logged. I don't see why they should be silenced.

common/dynamicconfig/constants.go
```diff
@@ -2262,11 +2261,6 @@ When the this config is zero or lower we will only update shard info at most onc
 		false,
 		`EmitShardLagLog whether emit the shard lag log`,
 	)
-	DefaultEventEncoding = NewNamespaceStringSetting(
-		"history.defaultEventEncoding",
-		enumspb.ENCODING_TYPE_PROTO3.String(),
-		`DefaultEventEncoding is the encoding type for history events`,
-	)
```
stephanos, line 2304: Drive-by change: this is not actually used anywhere.

temporal/server_test.go
```diff
@@ -177,10 +274,17 @@ func (d *errorLogDetector) Warn(msg string, tags ...tag.Tag) {
 }
 
 func (d *errorLogDetector) Error(msg string, tags ...tag.Tag) {
```
stephanos, line 276: Tangent: I'm not sure the `errorLogDetector` is super valuable; an actual end-to-end test that we're doing right now seems much better as we can actually confirm it works. Checking for error logs seems brittle.

service/history/api/get_history_util.go
```diff
@@ -135,6 +135,14 @@ func GetRawHistory(
 			rawHistory = append(rawHistory, blob)
 		}
 	}
+
+	// Ensure all raw history is proto3 encoded since data may be stored in other formats during testing.
+	// In production (proto3 encoding), this returns the input unchanged.
+	rawHistory, err = serialization.ReencodeEventBlobsAsProto3(shardContext.GetPayloadSerializer(), rawHistory)
+	if err != nil {
+		return nil, nil, err
+	}
```
stephanos, line 144: So this is annoying, but I don't have a better idea on how to address this. It has to be proto3 for the worker.

common/persistence/serialization/serializer.go
```diff
@@ -18,14 +18,20 @@ import (
...
 )
 
-// DefaultDecoder is here for convenience to skip the need to create a new Serializer when only decodig is needed.
-// It does not need an encoding type; as it will use the one defined in the DataBlob.
-var r Serializer = &serializerImpl{encodingType: enumspb.ENCODING_TYPE_UNSPECIFIED}
-var DefaultDecoder Decoder = r
+var (
+	// DefaultDecoder is here for convenience to skip the need to create a new Serializer when only decoding is needed.
+	// It does not need an encoding type; as it will use the one defined in the DataBlob.
+	defaultSerializer Serializer = &serializerImpl{encodingType: enumspb.ENCODING_TYPE_UNSPECIFIED}
```
yycptt, line 24: nit: shall we make sure `encodingType` is always specified?

service/history/api/get_history_util.go
yycptt: I am a bit confused. I thought we always return decoded history to SDK.

stephanos: This is due to the optimization from https://github.com/temporalio/temporal/pull/7342 where we don't decode history events. But here, if we've made them be JSON, we have to _re-encode_ them now as proto. This is ofc only happening if the env var is set; ie not in prod.

yycptt: oh so it's not about SDK, it's the proto deserializing trick we used in that optimization. 

hmmm @prathyushpv is using same trick in couple of more places: https://github.com/temporalio/temporal/pull/8829

You guys might want to coordinate.

PR #8427, opened 2025-10-03
softassert serviceerror.DataLoss helper

stephanos:
## What changed?

Added a new helper `UnexpectedDataLoss` to combine logging and error construction in one go.

And applied it everywhere since I assume every DataLoss is _unexpected_.

## Why?

Just like https://github.com/temporalio/temporal/pull/8388 it combines the two affordances of `softassert.Fail` and constructing an error.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/api/getworkflowexecutionhistory/api.go
+++ b/service/history/api/getworkflowexecutionhistory/api.go
@@ -250,8 +250,8 @@ func Invoke(
 				if len(history.Events) == 0 {
 					dataLossErr := softassert.UnexpectedDataLoss(
 						shardContext.GetLogger(),
-						"GetHistory returned empty history",
-						errors.New("no events in workflow history"),
+						"no events in workflow history",
+						nil,
 						tag.WorkflowNamespaceID(namespaceID.String()),
 						tag.WorkflowNamespace(request.GetRequest().GetNamespace()),
 						tag.WorkflowID(execution.GetWorkflowId()),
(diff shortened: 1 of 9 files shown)


service/history/api/getworkflowexecutionhistory/api.go
```diff
@@ -247,14 +248,15 @@ func Invoke(
 				}
 				// GetHistory func will not return empty history. Log workflow details if that is not the case
 				if len(history.Events) == 0 {
-					shardContext.GetLogger().Error(
+					dataLossErr := softassert.UnexpectedDataLoss(
```
prathyushpv, line 251: In this case, data loss error message will have both string `GetHistory returned empty history` and `no events in workflow history` right? But we only need the second string in error. Also, does the third argument need to be type error instead of string? If we are just using it to get the error string.

stephanos: > In this case, data loss error message will have both string GetHistory returned empty history and no events in workflow history right? But we only need the second string in error.

So the result will be:

- `"GetHistory returned empty history"` is logged.
- `"no events in workflow history"` is a log tag.
- the DataLoss error's message is `"GetHistory returned empty history: no events in workflow history"` (maybe that's a bit redundant)

I've updated the function docs to make that clearer.

stephanos: > But we only need the second string in error. Also, does the third argument need to be type error instead of string? If we are just using it to get the error string.

I went back and forth on whether it should be a string or an error. I thought error makes it clearer that it's part of the error that's being created.

stephanos: I've removed one of the messages now.

prathyushpv: Makes sense. But `GetHistory returned empty history` does not follow the go convention that error messages are all small case. Should we remove the first part from error string?

stephanos: I've made it `"no events in workflow history"` now.

prathyushpv (approved):

PR #8430, opened 2025-10-03
Ns handover error

michaely520:
## What changed?
Errors that occur prior to the telemetry interceptor are not captured, meaning there is no log, the request isn't counted, and the error isn't counted. I've extracted the telemetry interceptors error handling into a shared obj that I've wired into the NamespaceHandover error handling for now. An alternative approach was to move the TelemetryInterceptor earlier in the chain, but this would have caused double counting on redirects, and adding another dimension to our metrics (pre/post redirect) here to address the double counting would cause other classes of issues

As of now, other interceptors that are added prior to the TelemetryInterceptor are also at risk of this. We can wrap PreTelemetry errors and have a new interceptor handle these if we want to more generically address this issue - view my commit history for "Experimental" (which I reverted) to see an example of this.

## Why?
To specifically track metrics/logs on NamespaceHandover failures.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
Potential metrics loss if for some reason telemetry metric error handling is not wired correctly after the refactor.


service/frontend/fx.go
```diff
@@ -75,6 +75,7 @@ var Module = fx.Options(
 	fx.Provide(NamespaceLogInterceptorProvider),
 	fx.Provide(NamespaceHandoverInterceptorProvider),
 	fx.Provide(RedirectionInterceptorProvider),
+	fx.Provide(ErrorHandlerProvider),
 	fx.Provide(TelemetryInterceptorProvider),
```
yux0, line 79: Can you add a comment here to make sure the TelemetryInterceptor is after NamespaceHandoverInterceptor?

common/rpc/interceptor/namespace_handover.go
```diff
@@ -32,6 +32,7 @@ type (
 		nsCacheRefreshInterval dynamicconfig.DurationPropertyFn
 		metricsHandler         metrics.Handler
 		logger                 log.Logger
+		requestErrorHandler    *RequestErrorHandler
```
yux0: nit: ErrorHandler

yux0 (approved):

PR #8437, opened 2025-10-07
[Scheduled Actions] Manage CHASM backfiller lifecycle state with a closed flag

lina-temporal:
## What changed?
- Scheduler Backfillers are now closed and deleted by virtue of their LifecycleState transitioning to completed, instead of being deleted directly from the map.

## Why?
-  More idiomatic to CHASM, and also works around an edge case where we delete the node that generated the task deleting it.

### DIFF
--- a/chasm/lib/scheduler/proto/v1/message.proto
+++ b/chasm/lib/scheduler/proto/v1/message.proto
@@ -83,8 +83,14 @@ message BackfillerState {
     // needs to back off before retrying to fill.
     int64 attempt = 8;
 
-    // Set to true after the backfill is complete, used for lifecycle management.
-    bool closed = 9;
+    // Set to closed after the backfill is complete.
+    BackfillerLifecycle lifecycle = 9;
+}
+
+enum BackfillerLifecycle {
+    BACKFILLER_LIFECYCLE_UNSPECIFIED = 0;
+    BACKFILLER_LIFECYCLE_OPEN = 1;
+    BACKFILLER_LIFECYCLE_CLOSED = 2;
 }
 
 // CHASM scheduler retains the payload data for the last completed workflow
--- a/chasm/lib/scheduler/backfiller.go
+++ b/chasm/lib/scheduler/backfiller.go
@@ -38,6 +38,7 @@ func newBackfiller(
 		BackfillerState: &schedulerpb.BackfillerState{
 			BackfillId:        id,
 			LastProcessedTime: timestamppb.New(ctx.Now(scheduler)),
+			Lifecycle:         schedulerpb.BACKFILLER_LIFECYCLE_OPEN,
 		},
 		Scheduler: chasm.ComponentPointerTo(ctx, scheduler),
 	}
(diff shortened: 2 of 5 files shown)


chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -76,4 +76,7 @@ message BackfillerState {
     // Attempt count, incremented when the buffer is full and the Backfiller
     // needs to back off before retrying to fill.
     int64 attempt = 8;
+
+    // Set to true after the backfill is complete, used for lifecycle management.
+    bool closed = 9;
```
bergundy: I typically don't like to use booleans for lifecycles (or almost any other API), if you add a third stage later, you would either want to deprecate this field or add another boolean, which would allow you to represent invalid states.
I would use an enum for this.

lina-temporal:
> I'm slightly worried that this would allow introducing bugs later with code that loops over backfillers and forgets to inspect the `Closed` flag. 

I wouldn't think of that so much in terms of the `closed` flag, as much as that other components need to be mindful of referenced component lifecycle state, which I think is reasonable. Additionally, if they try to write to a closed component, CHASM will throw an error.

> Also wondering how CHASM handles completed components, it's not supposed to delete them AFAIU. 

~~This is presently how it's implemented, with closing being the intended way to clear out a component~~ (edit: this only applies to the root component). That's why I add the separate 'lingering'/idle state for the scheduler root component - so that it stays "active" w.r.t. the CHASM framework, while still being visible to `Describe` calls.  

> For example the callback component, when attached to a workflow, would be closed but we keep it around because it is useful for debugging purposes.

Sure, but suppose the workflow and the callback were both modeled as a (single) CHASM tree, which I think is where we want to get to ultimately. The callback's internal state may go from scheduled -> running -> completed {success, failure}, but the CHASM-facing lifecycle would always be 'running', precisely because we want to keep it around. When the parent workflow node closes, we'd also end up closing the callback child node. 

I get that I'm describing the current behavior but as an application developer it makes sense to me; I tell the framework that from its perspective, I want to keep referencing my node (even if my internal state means I don't actually plan to do any more work for it). Maybe a takeaway here is to delineate the closed lifecycle state into "closed" and "deleted" (or "lingering" and "closed").

chasm/lib/scheduler/proto/v1/message.proto
lina-temporal: Sure, I'll update this.

chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -82,6 +82,15 @@ message BackfillerState {
     // Attempt count, incremented when the buffer is full and the Backfiller
     // needs to back off before retrying to fill.
     int64 attempt = 8;
+
+    // Set to closed after the backfill is complete.
+    BackfillerLifecycle lifecycle = 9;
+}
+
+enum BackfillerLifecycle {
```
bergundy, line 90: Please call this status to be consistent with other CHASM components we are writing right now.

```suggestion
enum BackfillerStatus {
```

bergundy (commented):
I can see how having a status would be nice to allow a component to run in standalone mode. I don't think the backfiller is ever going to be run in this mode so I would consider it very much redundant.
I still want to make sure that CHASM as a framework doesn't auto-delete close components as that is not the right behavior for many components.
I was thinking that the application would have to delete the components from the map and not make it the framework's responsibility but I could be convinced otherwise if it was opt-in behavior.

lina-temporal:
>  I was thinking that the application would have to delete the components from the map and not make it the framework's responsibility but I could be convinced otherwise if it was opt-in behavior.

Yeah, I took a second look at it, and I still preferred how it was prior to this PR. I chatted with Yichao and I don't think we need this change, since the bug in question occurs when multiple pure tasks are being executed, and one of them deletes the node that still has tasks left. That won't happen here. Closing.

PR #8442, opened 2025-10-07
Handle permission denied error in Nexus HandleScheduleCommand

nikki-dag:
## What changed?
Nexus HandleScheduleCommand handles PermissionDenied by failing the WFT.

## Why?
We want to distinguish between Nexus endpoint not found and caller namespace unauthorized errors.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

---

> [!NOTE]
> Handle `PermissionDenied` from endpoint lookup by failing the workflow task and add a targeted unit test.
> 
> - **Workflow command handling** (`components/nexusoperations/workflow/commands.go`):
>   - On `GetByName` error `PermissionDenied`, return `FailWorkflowTaskError` with cause `WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES` and an authorization message, instead of propagating the error.
> - **Tests** (`components/nexusoperations/workflow/commands_test.go`):
>   - Extend fake endpoint registry to simulate `PermissionDenied` for a specific endpoint name.
>   - Add test "caller namespace unauthorized" asserting WFT failure and no history events.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7585f5a9406e095ee4ef0878519376cbe7cea656. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/nexusoperations/workflow/commands.go
+++ b/components/nexusoperations/workflow/commands.go
@@ -63,6 +63,11 @@ func (ch *commandHandler) HandleScheduleCommand(
 				}
 			}
 			// Ignore, and let the operation fail when the task is executed.
+		} else if errors.As(err, new(*serviceerror.PermissionDenied)) {
+			return workflow.FailWorkflowTaskError{
+				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
+				Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.ID(), attrs.Endpoint),
+			}
 		} else {
 			return err
 		}
--- a/components/nexusoperations/workflow/commands.go
+++ b/components/nexusoperations/workflow/commands.go
@@ -66,7 +66,7 @@ func (ch *commandHandler) HandleScheduleCommand(
 		} else if errors.As(err, new(*serviceerror.PermissionDenied)) {
 			return workflow.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
-				Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.ID(), attrs.Endpoint),
+				Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.Name(), attrs.Endpoint),
 			}
 		} else {
 			return err
--- a/components/nexusoperations/workflow/commands_test.go
+++ b/components/nexusoperations/workflow/commands_test.go
@@ -59,7 +59,9 @@ var defaultConfig = &nexusoperations.Config{
 func newTestContext(t *testing.T, cfg *nexusoperations.Config) testContext {
 	endpointReg := nexustest.FakeEndpointRegistry{
 		OnGetByName: func(ctx context.Context, namespaceID namespace.ID, endpointName string) (*persistencespb.NexusEndpointEntry, error) {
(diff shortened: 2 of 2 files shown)


components/nexusoperations/workflow/commands.go
```diff
@@ -63,6 +63,11 @@ func (ch *commandHandler) HandleScheduleCommand(
 				}
 			}
 			// Ignore, and let the operation fail when the task is executed.
+		} else if errors.As(err, new(*serviceerror.PermissionDenied)) {
```
pdoerner, line 66: To be totally sure we won't break OSS Nexus users and for safety, we should put this behind a dynamic config flag, similar to `EndpointNotFoundAlwaysNonRetryable`

components/nexusoperations/workflow/commands.go
```diff
@@ -63,6 +63,11 @@ func (ch *commandHandler) HandleScheduleCommand(
...
+			return workflow.FailWorkflowTaskError{
```
pdoerner, line 67: If the caller is legitimately unauthorized, won't this will cause the workflow task to fail and retry forever? Is that the desired behavior?

components/nexusoperations/workflow/commands.go
```diff
@@ -63,6 +63,11 @@ func (ch *commandHandler) HandleScheduleCommand(
...
+				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
+				Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.ID(), attrs.Endpoint),
```
pdoerner: Does `GetByName` return more than one type of `PermissionDenied` error? Should the message here be based on the message in the returned error?

bergundy: It would have been better to use `ns.Name()`, users don't know their internal namespace IDs typically.

pdoerner (commented):
I am nervous about blindly retrying all `PermissionDenied` errors. Callers that are legitimately unauthorized (not due to registry propagation delays), will see their workflows get stuck continually retrying the workflow task instead of just failing the operation and allowing the workflow to continue to make progress.

components/nexusoperations/workflow/commands.go
nikki-dag: AFAICT, the current `GetByName` OSS implementation never returns `PermissionDenied`, it ignores the passed caller namespace id.
https://github.com/temporalio/temporal/blob/main/common/nexus/endpoint_registry.go#L154-L166

Do you still think there is a risk for OSS users?

components/nexusoperations/workflow/commands.go
nikki-dag: We decided this would be more in line with the Temporal philosophy and gathered feedback from customers:
https://temporaltechnologies.slack.com/archives/C03SSH1GG0P/p1759424090596619?thread_ts=1757035642.898789&cid=C03SSH1GG0P

components/nexusoperations/workflow/commands.go
nikki-dag: OSS Temporal `GetByName` never returns `PermissionDenied`
Cloud Temporal  `GetByName` doesnt return `PermissionDenied` yet. But will start returning along with this change.

> Should the message here be based on the message in the returned error?

Let me see if I can fix this, but NotFound suffers from the same problem. In the case of Temporal Cloud, we can also return NotFound if the caller namespace is not found on the cell, but we don't expose that reason from the OSS.

PR #8451, opened 2025-10-08
Remove legacy batch path

spkane31:
## What changed?
Remove legacy batch pathway, this has been migrated to use an internal protobuf and not need protobuf -> struct -> protobuf serialization

## Why?
Deadcode

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
We've checked other internal locations for this dependency but we haven't found any, there's a chance we missed some.

### DIFF
--- a/service/worker/batcher/fx.go
+++ b/service/worker/batcher/fx.go
@@ -14,8 +14,7 @@ import (
 )
 
 const (
-	// BatchWFTypeName is the workflow type
-	BatchWFTypeName         = "temporal-sys-batch-workflow"
+	// BatchWFTypeProtobufName is the workflow type
 	BatchWFTypeProtobufName = "temporal-sys-batch-workflow-protobuf"
 	NamespaceDivision       = "TemporalBatcher"
 )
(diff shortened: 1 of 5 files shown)


service/worker/batcher/fx.go
```diff
@@ -15,9 +15,8 @@ import (
 
 const (
 	// BatchWFTypeName is the workflow type
-	BatchWFTypeName         = "temporal-sys-batch-workflow"
-	BatchWFTypeProtobufName = "temporal-sys-batch-workflow-protobuf"
-	NamespaceDivision       = "TemporalBatcher"
+	BatchWFTypeName   = "temporal-sys-batch-workflow-protobuf"
+	NamespaceDivision = "TemporalBatcher"
```
bergundy: This is fine but we will be stuck with the `-protobuf` name. You could consider registering the workflow with both names and removing the `-protobuf` suffix after the next release.

spkane31: Want to double check my plan, I can register both and have new workflows go to the `temporal-sys-batch-workflow`. The worker will still pick up remaining workflows and all new workflows, then we can remove the `*-protobuf` in the next release. Correct in that plan?

PR #8452, opened 2025-10-08
Remove Nexus endpoint not found always retryable dc

pdoerner:
## What changed?
Removed dynamic config `component.nexusoperations.endpointNotFoundAlwaysNonRetryable`

## Why?
Decided that continually failing and retrying the workflow task is more in line with Temporal philosophy.

## How did you test it?
- [x] built
- [ ] added new unit test(s)


bergundy (approved):
Thanks!

PR #8456, opened 2025-10-08
Add more softassert.UnexpectedInternalErr

stephanos:
## What changed?

Replace a few `serviceerror.NewInternal` with `softassert.UnexpectedInternalErr`.

## Why?

Catch these issues in CICD and other testing.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/api/startworkflow/api.go
+++ b/service/history/api/startworkflow/api.go
@@ -255,7 +255,7 @@ func (s *Starter) prepareNewWorkflow(workflowID string) (*creationParams, error)
 	if s.requestEagerStart() && workflowTaskInfo == nil {
 		return nil, softassert.UnexpectedInternalErr(
 			s.shardContext.GetLogger(),
-			"mutable state did not have a started workflow task",
+			"unexpected error: mutable state did not have a started workflow task",
 			nil,
 		)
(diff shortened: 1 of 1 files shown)


service/history/api/startworkflow/api.go
```diff
@@ -252,7 +253,11 @@ func (s *Starter) prepareNewWorkflow(workflowID string) (*creationParams, error)
 
 	workflowTaskInfo := mutableState.GetStartedWorkflowTask()
 	if s.requestEagerStart() && workflowTaskInfo == nil {
-		return nil, serviceerror.NewInternal("unexpected error: mutable state did not have a started workflow task")
+		return nil, softassert.UnexpectedInternalErr(
+			s.shardContext.GetLogger(),
```
spkane31, line 257: only comment, do we rely on error strings anywhere (we shouldn't imo) but if so we shouldn't change this

spkane31 (approved):

service/history/api/startworkflow/api.go
stephanos: I don't think so; but I can add it back, doesn't hurt.

PR #8459, opened 2025-10-09
add forbid require.Assertions lint rule

chaptersix:
## What changed?
Added a lint rule to forbid the embedding of require.Assertions in testify test suites. 

## Why?
When a test suite uses embedded `require.Assertions` on a test suite failures are not correctly attributed to subtests. Instead developers should use the require package directly and pass in the appropriate context.  

## How did you test it?
- [x] built


chaptersix:
image

PR #8470, opened 2025-10-10
Optimize activity task timeouts

prathyushpv:
## What changed?
Optimize activity task timeouts by removing the corresponding timeout timers when activities are scheduled/started/heartbeat recorded/completed etc.

## Why?
This will reduce the number of timeout timers in the timer queue.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/tests/activity_test.go
+++ b/tests/activity_test.go
@@ -893,25 +893,20 @@ func (s *ActivityTestSuite) TestActivityHeartbeatTimeout_MultipleTimers() {
 	// cleaned up when heartbeat timeouts are refreshed/cancelled during activity execution.
 	s.OverrideDynamicConfig(dynamicconfig.EnableBestEffortDeleteTasksOnWorkflowUpdate, true)
 
-	id := "functional-heartbeat-multiple-timers-test"
-	wt := "functional-heartbeat-multiple-timers-test-type"
-	tl := "functional-heartbeat-multiple-timers-test-taskqueue"
-	identity := "worker1"
-	activityName := "activity_heartbeat_timer"
+	tv := testvars.New(s.T())
 
-	workflowType := &commonpb.WorkflowType{Name: wt}
-	taskQueue := &taskqueuepb.TaskQueue{Name: tl, Kind: enumspb.TASK_QUEUE_KIND_NORMAL}
+	activityName := "activity_heartbeat_timer"
 
 	request := &workflowservice.StartWorkflowExecutionRequest{
 		RequestId:           uuid.New(),
 		Namespace:           s.Namespace().String(),
-		WorkflowId:          id,
-		WorkflowType:        workflowType,
-		TaskQueue:           taskQueue,
+		WorkflowId:          tv.WorkflowID(),
+		WorkflowType:        tv.WorkflowType(),
+		TaskQueue:           tv.TaskQueue(),
 		Input:               nil,
 		WorkflowRunTimeout:  durationpb.New(100 * time.Second),
 		WorkflowTaskTimeout: durationpb.New(1 * time.Second),
-		Identity:            identity,
+		Identity:            tv.WorkerIdentity(),
 	}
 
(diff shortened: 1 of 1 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -1836,6 +1849,9 @@ func (ms *MutableStateImpl) UpdateActivityProgress(
 			metrics.OperationTag(metrics.HistoryRecordActivityTaskHeartbeatScope),
 			metrics.NamespaceTag(ms.namespaceEntry.Name().String()))
 	}
+
+	// Delete Heartbeat timeout task since heartbeat has been received
+	ms.deleteActivityTimeoutTask(ai.ScheduledEventId, enumspb.TIMEOUT_TYPE_HEARTBEAT, ai)
```
yiminc, line 1854: we need a new heartbeat timeout task, make sure we don't delete the new one.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6159,6 +6180,118 @@ func (ms *MutableStateImpl) GetWorkflowTaskStartToCloseTimeoutTask() *tasks.Work
...
 }
 
+// StoreActivityTimeoutTask stores an activity timeout task reference in memory.
+func (ms *MutableStateImpl) StoreActivityTimeoutTask(
+	task *tasks.ActivityTimeoutTask,
+) {
+	timeoutTasks, exists := ms.activityTimeoutTasks[task.EventID]
+	if !exists {
+		timeoutTasks = &activityTimeoutTasks{}
+		ms.activityTimeoutTasks[task.EventID] = timeoutTasks
```
yiminc, line 6190: is task.EventID the activity scheduledEventID?

service/history/workflow/mutable_state_impl.go
```diff
@@ -6159,6 +6180,118 @@ func (ms *MutableStateImpl) GetWorkflowTaskStartToCloseTimeoutTask() *tasks.Work
...
+
+	timeoutTasks, exists := ms.activityTimeoutTasks[scheduledEventID]
+	if !exists {
+		return
+	}
+
+	var task *tasks.ActivityTimeoutTask
+
+	switch timeoutType {
+	case enumspb.TIMEOUT_TYPE_SCHEDULE_TO_START:
```
yiminc, line 6227: what about Start_to_close and Schedule_to_close?

service/history/workflow/mutable_state_impl.go
prathyushpv: Yes. I have added a test TestActivityHeartbeatTimeout_MultipleTimers below to make sure subsequent heartbeat timers fire. But I just realized that we ignore the request to delete tasks here.

service/history/workflow/mutable_state_impl.go
prathyushpv: Yes. It is scheduledEventID: https://github.com/temporalio/temporal/blob/add064848f9e9ecab6917d00819bcc3e69edcb36/service/history/workflow/timer_sequence.go#L251-L251

service/history/workflow/mutable_state_impl.go
prathyushpv: Those two timers are deleted in recordActivityTimeoutTasksForDeletion() function below. That happens when pending activity is deleted from the mutable state. This function is only called when an activity task is started or a heartbeat is received.

service/history/workflow/mutable_state_impl.go
yiminc: Not sure I understand this `But I just realized that we ignore the request to delete tasks here.`?

prathyushpv: I meant that I added a test to make sure subsequent timers are fired. But the test is not doing anything since persistence layer will ignore our request to delete timers.

yiminc (approved):

PR #8474, opened 2025-10-11
[Scheduled Actions] CHASM API Handlers

lina-temporal:
## What changed?
- Added an RPC service to the CHASM Scheduler which implements its CRUD operations.
- The intention is for the frontend handler to attempt to route to CHASM
operations in history, and in the event of a not-found, let frontend retry its
old (V1 scheduler) path.
- These assume frontend's validation. As we move towards ripping out the old
scheduler code, we'll start moving all of that existing logic over into the CHASM
world, but for now, let's keep it as-is for as light a possible touch on existing
paths.
- This is based on `standalone-activity` just to get the API boilerplate stuff.

### DIFF
--- a/service/frontend/namespace_handler.go
+++ b/service/frontend/namespace_handler.go
@@ -513,6 +513,7 @@ func (d *namespaceHandler) UpdateNamespace(
 		if updateReplicationConfig.GetActiveClusterName() != "" {
 			activeClusterChanged = true
 			replicationConfig.ActiveClusterName = updateReplicationConfig.GetActiveClusterName()
+			replicationConfig.State = enumspb.REPLICATION_STATE_NORMAL
 		}
 	}
 
@@ -884,7 +885,7 @@ func (d *namespaceHandler) createResponse(
 	replicationConfigResult := &replicationpb.NamespaceReplicationConfig{
 		ActiveClusterName: replicationConfig.ActiveClusterName,
 		Clusters:          clusters,
-		State:             replicationConfig.State,
+		State:             replicationConfig.GetState(),
 	}
 
 	var failoverHistory []*replicationpb.FailoverStatus
--- a/service/frontend/namespace_handler_test.go
+++ b/service/frontend/namespace_handler_test.go
@@ -679,7 +679,7 @@ func (s *namespaceHandlerCommonSuite) TestUpdateNamespace_UpdateActiveClusterWit
 			ReplicationConfig: &persistencespb.NamespaceReplicationConfig{
 				ActiveClusterName: clusterName2,
 				Clusters:          []string{clusterName1, clusterName2},
-				State:             enumspb.REPLICATION_STATE_HANDOVER,
+				State:             enumspb.REPLICATION_STATE_NORMAL,
 				FailoverHistory: []*persistencespb.FailoverStatus{
 					{
 						FailoverTime:    timestamppb.New(update1Time),
@@ -754,6 +754,7 @@ func (s *namespaceHandlerCommonSuite) TestUpdateNamespace_ChangeActiveClusterWit
 			ReplicationConfig: &persistencespb.NamespaceReplicationConfig{
 				ActiveClusterName: clusterName2,
 				Clusters:          []string{clusterName1, clusterName2},
+				State:             enumspb.REPLICATION_STATE_NORMAL,
 				FailoverHistory: []*persistencespb.FailoverStatus{
 					{
 						FailoverTime:    timestamppb.New(update1Time),
@@ -865,6 +866,7 @@ func (s *namespaceHandlerCommonSuite) TestUpdateNamespace_UpdateActiveCluster_Li
 			ReplicationConfig: &persistencespb.NamespaceReplicationConfig{
 				ActiveClusterName: clusterName2,
 				Clusters:          []string{clusterName1, clusterName2},
+				State:             enumspb.REPLICATION_STATE_NORMAL,
 				FailoverHistory:   sizeLimitedFailoverHistory,
 			},
 			ConfigVersion:               0,
@@ -1147,6 +1149,7 @@ func (s *namespaceHandlerCommonSuite) TestUpdateLocalNamespace_AllAttrSet() {
 			ReplicationConfig: &persistencespb.NamespaceReplicationConfig{
 				ActiveClusterName: activeClusterName,
 				Clusters:          []string{activeClusterName},
+				State:             enumspb.REPLICATION_STATE_NORMAL,
 			},
 			ConfigVersion:               1,
 			FailoverNotificationVersion: 0,
@@ -1666,6 +1669,7 @@ func (s *namespaceHandlerCommonSuite) TestFailoverGlobalNamespace_NotMaster() {
 			ReplicationConfig: &persistencespb.NamespaceReplicationConfig{
 				ActiveClusterName: clusterName2,
 				Clusters:          []string{clusterName1, clusterName2},
+				State:             enumspb.REPLICATION_STATE_NORMAL,
 				FailoverHistory: []*persistencespb.FailoverStatus{
 					{
 						FailoverTime:    timestamppb.New(update1Time),


chasm/lib/activity/frontend.go
bergundy, line 1: Don't commit this file or any of the others from the standalone-activity branch, this needs to stay in a feature branch for now and is unrelated to your work.

chasm/lib/scheduler/proto/v1/request_response.proto
```diff
@@ -0,0 +1,118 @@
+syntax = "proto3";
+
+package temporal.server.chasm.lib.scheduler.proto.v1;
+
+option go_package = "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb;schedulerpb";
+
+import "temporal/api/schedule/v1/message.proto";
+import "temporal/api/common/v1/message.proto";
+
+message CreateScheduleRequest {
```
bergundy, line 9: Wondering why not take the frontend request and embed it here instead of duplicating all of the fields?

chasm/lib/scheduler/scheduler.go
```diff
@@ -64,15 +66,51 @@ func NewScheduler(
...
+// Create initializes a new Scheduler for CreateSchedule requests.
+func Create(
+	ctx chasm.MutableContext,
+	req *schedulerpb.CreateScheduleRequest,
+) (*Scheduler, *schedulerpb.CreateScheduleResponse, error) {
+	// TODO: namespace name should be resolved from namespace_id via namespace registry
+	sched := NewScheduler(
+		ctx,
+		"", // TODO: should be namespace name
+		req.NamespaceId,
```
bergundy, line 117: Can't the scheduler get this information from CHASM? The execution / component ref should include all of this information.

chasm/lib/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,386 @@
...
+	return chasm.NewMutableContext(context.Background(), e.node)
+}
+
+// closeTransaction closes the CHASM transaction and verifies no error.
+func (e *testEnv) closeTransaction() {
+	_, err := e.node.CloseTransaction()
+	require.NoError(e.t, err)
+}
+
+func TestDescribe_Basic(t *testing.T) {
```
bergundy: Why not do this in functional tests? That will be much more beneficial.

cmd/tools/protoc-gen-go-chasm/main.go
bergundy, line 1: You will need to copy this file into your branch.

chasm/lib/scheduler/proto/v1/request_response.proto
bergundy: I would like to establish this as a pattern where we simply embed the frontend request unless there's anything that the frontend adds to that request. Frontend should typically just do validation and pass the request along to history.

chasm/lib/scheduler/scheduler.go
```diff
@@ -39,6 +41,8 @@ const (
 	recentActionCount = 10
 )
 
+var ErrConflictTokenMismatch = errors.New("conflict token mismatch")
```
bergundy: You're going to want to make this a failed precondition grpc error so it doesn't get treated as an internal error.

chasm/lib/scheduler/proto/v1/request_response.proto
lina-temporal: Sure, I like this pattern far more!

chasm/lib/activity/frontend.go
lina-temporal: Yeah, I'm aware, I'll rebase this back off main now that the other RPC changes have landed there.

chasm/lib/scheduler/scheduler.go
lina-temporal: Will fix.

chasm/lib/scheduler/scheduler.go
lina-temporal: CHASM does have all of this information, but I don't think that components have a way to access it, particularly in the Create handler where the backing component/its Ref aren't set yet.

chasm/lib/scheduler/scheduler_test.go
lina-temporal: Sure, yeah, I can move these to functional.

chasm/lib/scheduler/proto/v1/request_response.proto
```diff
@@ -0,0 +1,62 @@
...
+
+message DeleteScheduleResponse {
+    temporal.api.workflowservice.v1.DeleteScheduleResponse response = 1;
+}
+
+message DescribeScheduleRequest {
+    temporal.api.workflowservice.v1.DescribeScheduleRequest request = 1;
+
+    // Internal namespace ID (UUID).
+    string namespace_id = 2;
```
bergundy: Really nitpicking here but I would put this in field first 🤷

chasm/lib/scheduler/proto/v1/service.proto
```diff
@@ -0,0 +1,35 @@
...
+package temporal.server.chasm.lib.scheduler.proto.v1;
+
+option go_package = "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb;schedulerpb";
+
+import "chasm/lib/scheduler/proto/v1/request_response.proto";
+import "temporal/server/api/routing/v1/extension.proto";
+
+service SchedulerService {
+    rpc CreateSchedule(CreateScheduleRequest) returns (CreateScheduleResponse) {
+        option (temporal.server.api.routing.v1.routing).namespace_id = "namespace_id";
```
bergundy: This is optional and the default is already `namespace_id`.

chasm/lib/scheduler/scheduler.go
bergundy: Hmmm... we should make it available IMHO. Not blocking this PR though, but let's address before going live and creating persistence record variations that we have to support forever.

chasm/lib/scheduler/proto/v1/request_response.proto
```diff
@@ -0,0 +1,62 @@
+syntax = "proto3";
+
+package temporal.server.chasm.lib.scheduler.proto.v1;
+
+option go_package = "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb;schedulerpb";
+
+import "temporal/api/workflowservice/v1/request_response.proto";
+
+message CreateScheduleRequest {
+    temporal.api.workflowservice.v1.CreateScheduleRequest request = 1;
```
bergundy: I want to establish a pattern where we call these fields `frontend_request` and in the response` frontend_response` just to clarify a bit more what they are and avoid the `request.Request` stutter in the code.

chasm/lib/scheduler/scheduler.go
```diff
@@ -64,15 +69,53 @@ func NewScheduler(
...
+		ctx,
+		"", // TODO: should be namespace name
+		req.NamespaceId,
+		req.Request.ScheduleId,
+		req.Request.Schedule,
+		req.Request.InitialPatch,
+	)
+	return sched, &schedulerpb.CreateScheduleResponse{
+		Response: &workflowservice.CreateScheduleResponse{
+			ConflictToken: []byte{},
```
bergundy: Don't we need to provide a conflict token here?

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +333,86 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+func (s *Scheduler) GenerateConflictToken() []byte {
+	token := make([]byte, 8)
+	binary.LittleEndian.PutUint64(token, uint64(s.ConflictToken))
+	return token
+}
+
+func (s *Scheduler) ValidateConflictToken(token []byte) bool {
+	current := s.GenerateConflictToken()
+	return bytes.Equal(current, token)
+}
```
bergundy: Can be private... (along with all of the other fields on this struct)

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +333,86 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+	}, nil
+}
+
+// Patch applies a patch to the schedule for PatchSchedule requests.
+func (s *Scheduler) Patch(
+	ctx chasm.MutableContext,
+	req *schedulerpb.PatchScheduleRequest,
+) (*schedulerpb.PatchScheduleResponse, error) {
+	// Handle paused status.
+	// TODO - use SetPaused from visibility for this
```
bergundy: Not sure I follow this comment...

chasm/lib/scheduler/scheduler.go
```diff
@@ -290,3 +333,86 @@ func (s *Scheduler) recordActionResult(result *schedulerActionResult) {
...
+func (s *Scheduler) Update(
+	ctx chasm.MutableContext,
+	req *schedulerpb.UpdateScheduleRequest,
+) (*schedulerpb.UpdateScheduleResponse, error) {
+	if !s.ValidateConflictToken(req.Request.ConflictToken) {
+		return nil, ErrConflictTokenMismatch
+	}
+
+	s.Schedule = common.CloneProto(req.Request.Schedule)
+	// TODO - also merge custom search attributes here
```
bergundy: You just need to update the visibility component here with the request's search attributes, it does the merging for you already.

chasm/lib/scheduler/scheduler.go
```diff
@@ -64,15 +69,53 @@ func NewScheduler(
...
+		}
+
+		for _, backfill := range patch.BackfillRequest {
+			s.NewRangeBackfiller(ctx, backfill)
+		}
+	}
+}
+
+// Create initializes a new Scheduler for CreateSchedule requests.
+func Create(
```
bergundy: I would call this method `CreateScheduler` or even better, use `NewScheduler` for this this.

lina-temporal: Updated to `CreateScheduler`.

chasm/lib/scheduler/scheduler.go
lina-temporal: Removed, it's outdated

chasm/lib/scheduler/scheduler.go
lina-temporal: Updated

chasm/lib/scheduler/scheduler.go
lina-temporal: Fixed

chasm/lib/scheduler/proto/v1/request_response.proto
lina-temporal: Updated

chasm/lib/scheduler/scheduler.go
bergundy: https://github.com/temporalio/temporal/pull/8533

PR #8478, opened 2025-10-13
Move NewMatcher and EnableFairness evaluation to the partition manager

majiru:
## What changed?
Moved the NewMatcher and EnableFairness evaluation to the partition manager.

## Why?
David noted that having this within the physical task queue leads to scenarios where we wind up with some queues with it enabled and some that don't, leading to test flakiness and crashes.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/task_queue_partition_manager.go
+++ b/service/matching/task_queue_partition_manager.go
@@ -118,12 +118,11 @@ func newTaskQueuePartitionManager(
 		pm.unloadFromEngine(unloadCauseConfigChange)
 	}
 
-	tqConfig.EnableFairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(unload)
+	var fairness bool
+	fairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(unload)
 	// Fairness is disabled for sticky queues for now so that we can still use TTLs.
-	if partition.Kind() == enumspb.TASK_QUEUE_KIND_STICKY {
-		tqConfig.EnableFairness = false
-	}
-	if tqConfig.EnableFairness {
+	tqConfig.EnableFairness = fairness && partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY
+	if fairness {
 		tqConfig.NewMatcher = true
 	} else {
 		tqConfig.NewMatcher, pm.cancelNewMatcherSub = tqConfig.NewMatcherSub(unload)


service/matching/physical_task_queue_manager.go
```diff
@@ -230,12 +221,7 @@ func newPhysicalTaskQueueManager(
 		return pqMgr, nil
 	}
 
-	newMatcher, pqMgr.cancelMatcherSub = config.NewMatcher(func(bool) {
-		// unload on change to NewMatcher so that we can reload with the new setting:
-		pqMgr.UnloadFromPartitionManager(unloadCauseConfigChange)
-	})
-
-	if newMatcher {
+	if config.NewMatcher {
```
dnr: there's no code in the middle so this can be an else if now
```suggestion
  } else if config.NewMatcher {
```

service/matching/physical_task_queue_manager.go
```diff
@@ -178,15 +176,8 @@ func newPhysicalTaskQueueManager(
...
-	var fairness, newMatcher bool
 	// Fairness is disabled for sticky queues for now so that we can still use TTLs.
-	if !isSticky {
-		fairness, pqMgr.cancelFairnessSub = config.EnableFairness(func(bool) {
-			// unload on change so that we can reload with the new setting:
-			pqMgr.UnloadFromPartitionManager(unloadCauseConfigChange)
-		})
-	}
-	if fairness {
+	if !isSticky && config.EnableFairness {
```
dnr: we don't need the sticky check anymore here since we do it at the partition level, right?
```suggestion
	if config.EnableFairness {
```

service/matching/task_queue_partition_manager.go
```diff
@@ -111,6 +114,16 @@ func newTaskQueuePartitionManager(
...
 
+	unload := func(bool) {
+		pm.unloadFromEngine(unloadCauseConfigChange)
+	}
+
+	// Fairness is disabled for sticky queues for now so that we can still use TTLs.
+	if partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY {
+		tqConfig.EnableFairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(unload)
+	}
+	tqConfig.NewMatcher, pm.cancelNewMatcherSub = tqConfig.NewMatcherSub(unload)
```
dnr: I'm not sure if it's worth it, but I wonder if we should do it like:
```
fairness, cancel = config.EnableFairnessSub()
config.EnableFairness = fairness && kind != STICKY
if fairness {
  config.newMatcher = true
} else {
  config.newMatcher, cancel = config.NewMatcherSub()
}
```
i.e. the fairness setting always implies newMatcher even on sticky, and we can skip the second subscription if fairness is on?

service/matching/task_queue_partition_manager.go
```diff
@@ -111,6 +114,21 @@ func newTaskQueuePartitionManager(
...
+	unload := func(bool) {
+		pm.unloadFromEngine(unloadCauseConfigChange)
+	}
+
+	tqConfig.EnableFairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(unload)
+	// Fairness is disabled for sticky queues for now so that we can still use TTLs.
+	if partition.Kind() == enumspb.TASK_QUEUE_KIND_STICKY {
+		tqConfig.EnableFairness = false
+	}
+	if tqConfig.EnableFairness {
```
dnr: the idea was that even for sticky, EnableFairness would still imply NewMatcher even without the explicit setting. this way, if EnableFairness is true and NewMatcher is false, then sticky gets old matcher.

it doesn't really matter since both should be set at once in the config, but I thought we could make it explicit in code.

dnr (approved):

PR #8480, opened 2025-10-13
Set IsCloseTransferTaskAcked and IsForceReplication on replication

michaely520:
## What changed?

1. Impl to write IsCloseTransferTaskAcked and IsForceReplication flags on replication stream.
2. Impl for a replication stream interceptor, which fully intercepts client and server and stores calls in-memory for us to assert against. Also writes to log files so we can reason about/follow logs without touching server code (this impl does not immediately log to file for perf reasons).
3. Accompanying tests (unit and functional)

## Why?
This is part of a larger change to optimize passive task processing by no-oping side-effects of task processing depending on if we can deem that the task has already been processed on the parent. Will expand further in the next PR.

## How did you test it?
- [x] built
- [x] added new unit test(s)

`michaely520 ~/projects/temporal % go test -v -tags test_dep ./tests/xdc -run TestStreamBasedReplicationTestSuite/EnableTransitionHistory/TestCloseTransferTaskAckedReplication -timeout 10m -count=1`

### DIFF
--- a/service/history/replication/raw_task_converter.go
+++ b/service/history/replication/raw_task_converter.go
@@ -712,7 +712,6 @@ func (c *syncVersionedTransitionTaskConverter) convert(
 
 	syncStateResult.VersionedTransitionArtifact.IsCloseTransferTaskAcked = c.isCloseTransferTaskAcked(mutableState)
 	syncStateResult.VersionedTransitionArtifact.IsForceReplication = taskInfo.IsForceReplication
-
 	// do not access mutable state after this point
 
 	err = c.replicationCache.Update(taskInfo.RunID, targetClusterID, syncStateResult.VersionedTransitionHistory, currentHistoryCopy.Items)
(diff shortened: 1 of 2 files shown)


service/history/replication/raw_task_converter.go
```diff
@@ -708,6 +709,10 @@ func (c *syncVersionedTransitionTaskConverter) convert(
 	if err != nil {
 		return nil, err
 	}
+
+	if taskInfo.IsForceReplication {
+		syncStateResult.VersionedTransitionArtifact.IsCloseTransferTaskAcked = c.isCloseTransferTaskAcked(mutableState)
```
xwduan: After thinking a bit more, `IsCloseTransferTaskAcked` probably is not appropriate. I understand that `IsCloseTransferTaskAcked` is more generic and can be used for other purpose. But the in order to correctly use this field, the user has to read this line of code to understand that this field can only be true when the task is force replication. 

I think we have 2 options:
1. we can send both `isForceReplication` and `isForceReplication` to target
2. use a more specific name i.e. `shouldSkipGeneratingCloseTransferTask`

michaely520: discussed offline, will go with the two flag approach

xwduan (approved):
We also need to change SyncWorkflowStateTask. But they can be in a different pr.

Issue #8436, opened 2025-10-06
Temporal does not support setting postgresql sslmode=verify-ca


jtamagnan:
## Expected Behavior

I wanted to connect to my postgresql database using `sslmode:verify-ca`. I tried to set:

```
                    connectAttributes:
                      sslmode: "verify-ca"
```

but it failed.

## Actual Behavior

This is unsupported and leads to errors like:
```
panic: duplicate connection attr: sslmode:require, sslmode:verify-ca
```
when `cfg.TLS.Enabled == true`

or:
```
panic: duplicate connection attr: sslmode:require, sslmode:verify-ca
```

when `cfg.TLS.Enabled == false`

The [code](https://github.com/temporalio/temporal/blob/main/common/persistence/sql/sqlplugin/postgresql/session/session.go#L93) that causes this is pretty easy to read:
```go
	if cfg.TLS != nil && cfg.TLS.Enabled {
		if !cfg.TLS.EnableHostVerification {
			parameters.Set(sslMode, sslModeRequire)
		} else {
			parameters.Set(sslMode, sslModeFull)
		}

		...
	} else {
		parameters.Set(sslMode, sslModeNoop)
	}

	for k, v := range cfg.ConnectAttributes {
		key := strings.TrimSpace(k)
		value := strings.TrimSpace(v)
		if parameters.Get(key) != "" {
			panic(fmt.Sprintf("duplicate connection attr: %v:%v, %v:%v",
				key,
				parameters.Get(key),
				key, value,
			))
		}
```

regardless of what set of options we put in `sslmode` will always be set to one of `require`, `disable`, or `verify-full`. This means that if we have a configuration like:
```
                    connectAttributes:
                      sslmode: "verify-ca"
```
we'll run into issues.

## Steps to Reproduce the Problem

Set a config like:
```
                    connectAttributes:
                      sslmode: "verify-ca"
```

## Specifications

  - Version: [Master](https://github.com/temporalio/temporal/blob/main/common/persistence/sql/sqlplugin/postgresql/session/session.go#L93)
  - Platform: Self-hosted

jtamagnan:
A targeted approach for postgresql may look something like this. I'd be happy to open a PR and try to usher it through if we think this would be a desireable solution. Another solution may be to not set the sslmode if it would be disabled or to let the sslmode be overwritten

<details>

<summary> Diff </summary>

```diff
diff --git a/common/auth/tls.go b/common/auth/tls.go
[...]

PR #8481, opened 2025-10-13
Allow overriding all postgres connection attributes

jtamagnan:
## What changed?

User provided connection attributes take precedent over ones set automatically the code.

## Why?

This fixes https://github.com/temporalio/temporal/issues/8436 by allowing users to set `sslmode` through `connectAttributes`.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks

Users may override values that they didn't intend to, where the server would fail to start up before.

### DIFF
--- a/common/persistence/sql/sqlplugin/postgresql/session/session.go
+++ b/common/persistence/sql/sqlplugin/postgresql/session/session.go
@@ -1,6 +1,7 @@
 package session
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
@@ -54,7 +55,11 @@ func createConnection(
 	d driver.Driver,
 	resolver resolver.ServiceResolver,
 ) (*sqlx.DB, error) {
-	db, err := d.CreateConnection(buildDSN(cfg, resolver))
+	dsn, err := buildDSN(cfg, resolver)
+	if err != nil {
+		return nil, err
+	}
+	db, err := d.CreateConnection(dsn)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


jtamagnan:
Thanks for the review! Really excited to get this PR in.

> Thanks for the contribution. Instead of what you've done here, let's change the `if` in line 96 to first check `parameters.Get(sslMode)` and avoid setting the mode if a user provided a specific mode.

I'm sorry. I'm a bit confused by this. On line 96 `parameters.Get(sslMode)` should always be `""` since we haven't set it yet. Do you mean checking `cfg.ConnectAttributes[sslMode]`?

bergundy:
> Thanks for the review! Really excited to get this PR in.
> 
> > Thanks for the contribution. Instead of what you've done here, let's change the `if` in line 96 to first check `parameters.Get(sslMode)` and avoid setting the mode if a user provided a specific mode.
> 
> I'm sorry. I'm a bit confused by this. On line 96 `parameters.Get(sslMode)` should always be `""` since we haven't set it yet. Do you mean checking `cfg.ConnectAttributes[sslMode]`?

Yes, that's what I meant. Sorry for the confusion.

common/persistence/sql/sqlplugin/postgresql/session/session.go
```diff
@@ -122,5 +104,26 @@ func buildDSNAttr(cfg *config.SQL) url.Values {
...
+				parameters.Set(sslMode, sslModeFull)
+			} else {
+				parameters.Set(sslMode, sslModeRequire)
+			}
+		}
+
+		if parameters.Get(sslCA) == "" && cfg.TLS.CaFile != "" {
+			parameters.Set(sslCA, cfg.TLS.CaFile)
+		}
+		if parameters.Get(sslKey) == "" && cfg.TLS.KeyFile != "" && parameters.Get(sslCert) == "" && cfg.TLS.CertFile != "" {
```
prathyushpv: Thanks for the change! Can we split this condition into two if blocks, so that the user can override `KeyFile` or `CertFile`  without affecting the other?

bergundy: I'll improve this, the idea was that we don't set one without the other to maintain the previous behavior.

prathyushpv (approved):

jtamagnan:
@bergundy @prathyushpv thank you both so much for taking this across the finish line! I really appreciate it!

PR #8483, opened 2025-10-14
Consolidate default priority handling

dnr:
## What changed?
- Set default priority on all internalTasks after creation.
- Combine effective priority and poll forwarder flag into one value.
- For simplicity, default priority and level count are set at partition load time.

## Why?
Tasks with unset priority key were not actually getting the default value when read from backlog. This sets the default priority on all sync, backlog, query, and nexus tasks. (Priority is not meaningful for sync match tasks, and nexus has no priority yet, but they're set anyway for consistency.) This will also make it easier to adjust the priority for other purposes, e.g. prioritizing the old backlog for migrating persistence.

## How did you test it?
- [x] built
- [ ] added new unit test(s)
- [x] modified functional test

### DIFF
--- a/service/matching/physical_task_queue_manager.go
+++ b/service/matching/physical_task_queue_manager.go
@@ -766,9 +766,11 @@ func (c *physicalTaskQueueManagerImpl) getOrCreateTaskTracker(
 	priorityKey priorityKey,
 ) *taskTracker {
 	// priorityKey could be zero here if we're tracking dispatched tasks (i.e. called from PollTask)
-	// and the the poll was forwarded so we have a "started" task. We don't return the priority with
-	// the started task info so it's not available here. Use the default priority to avoid confusion
+	// and the poll was forwarded so we have a "started" task. We don't return the priority with the
+	// started task info so it's not available here. Use the default priority to avoid confusion
 	// even though it may not be accurate.
+	// TODO: either return priority with the started task, or do this tracking on the node where the
+	// match happened, so we have the right value here.
 	if priorityKey == 0 {
 		priorityKey = c.config.DefaultPriorityKey
 	}


service/matching/matcher_data.go
```diff
@@ -14,7 +14,10 @@ import (
 	"go.temporal.io/server/common/util"
 )
 
-const invalidHeapIndex = -13 // use unusual value to stand out in panics
+const (
+	invalidHeapIndex      = -13     // use unusual value to stand out in panics
+	pollForwarderPriority = 1000000 // lower than any other priority
```
majiru: Maybe an unwarranted concern, but do we have any way to enforce that the pollForwarderPriority is lower than any other priority? Does it make sense to clamp the PriorityLevels we pull from the config and/or put a upper bound at the API level?

service/matching/physical_task_queue_manager.go
```diff
@@ -766,8 +765,12 @@ func (c *physicalTaskQueueManagerImpl) getOrCreateTaskTracker(
 	intervals map[priorityKey]*taskTracker,
 	priorityKey priorityKey,
 ) *taskTracker {
+	// priorityKey could be zero here if we're tracking dispatched tasks (i.e. called from PollTask)
+	// and the the poll was forwarded so we have a "started" task. We don't return the priority with
```
majiru: Small inconsequential typo
```suggestion
	// and the poll was forwarded so we have a "started" task. We don't return the priority with
```

service/matching/matcher_data.go
dnr: I can add a check. It should never be anywhere near that high but it's true there is no check now

PR #8484, opened 2025-10-14
No-op refreshing tasks if flag set

michaely520:
## What changed?

Followup to https://github.com/temporalio/temporal/pull/8480 

Optimize passive side to not generate close transfer tasks if the flag is set in the replication stream.

## Why?
In the case where we are force replicating a large number of closed child workflows, each replicated workflow will generate a close transfer task, which requires a side-effect verification against its parent workflow on the passive resulting in massive lock contention. This optimization works by ensuring that if the flag is set on the replication message, then it means the child has fully been processed on the active and the parent already has the child result in its mutable state, meaning no need to verify on passive side.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

Log is emitted correctly in functional test from prior PR. From test:
`2025-10-14T13:08:33.488-0700	info	Skipping workflow close task generation - close transfer task already acked on active cluster	{"cluster-name": "standby_mnomr", "host": "127.0.0.1:55628", "shard-id": 1, "address": "127.0.0.1:55628", "wf-namespace-id": "a43d78f4-b7b5-4940-b8ee-c3156f8016aa", "wf-id": "test-replication-d9b7cd87-b5ae-409c-ab94-532cb5cea69c", "wf-run-id": "0199e456-a3d1-73d4-a44d-1c2ffaf387cf", "logging-call-at": "/Users/michaely520/projects/temporal/service/history/workflow/task_refresher.go:120"}`

### DIFF
--- a/service/history/workflow/task_generator.go
+++ b/service/history/workflow/task_generator.go
@@ -194,7 +194,6 @@ func (r *TaskGeneratorImpl) GenerateWorkflowCloseTasks(
 
 	var closeTasks []tasks.Task
 
-	// Only add the close transfer task if it hasn't already been acked on the active cluster.
 	if !skipCloseTransferTask {
 		closeExecutionTask := &tasks.CloseExecutionTask{
 			// TaskID, Visiblitytimestamp is set by shard


service/history/ndc/workflow_state_replicator.go
```diff
@@ -1159,6 +1159,7 @@ func (r *WorkflowStateReplicatorImpl) applySnapshotWhenWorkflowNotExist(
 	sourceCluster string,
 	newRunInfo *replicationspb.NewRunInfo,
 	isStateBased bool,
+	isCloseTransferTaskAcked bool,
```
xwduan: nit: skipGenerateCloseTransferTask

service/history/workflow/task_generator.go
```diff
@@ -179,20 +185,30 @@ func (r *TaskGeneratorImpl) GenerateWorkflowStartTasks(
...
-		// TaskID, Visiblitytimestamp is set by shard
-		WorkflowKey:      r.mutableState.GetWorkflowKey(),
-		Version:          closeVersion,
-		DeleteAfterClose: deleteAfterClose,
-	}
-	closeTasks := []tasks.Task{
-		closeExecutionTask,
+	var closeTasks []tasks.Task
+
+	// Only add the close transfer task if it hasn't already been acked on the active cluster.
```
xwduan: nit: this comments maybe just:
`Only add the close transfer task if not skipped` or just remove the comment? I think we don't need to expose active/passive info here.

xwduan (approved):

PR #8485, opened 2025-10-14
Add MSPointer to CHASM framework

pdoerner:
## What changed?
Added a new special `MSPointer` type to the CHASM framework

## Why?
To support components invoking methods from the underlying mutable state. Specifically, Nexus callbacks will need to get the Nexus completion from the mutable state.

## How did you test it?
updated unit tests


chasm/lib/workflow/workflow.go
```diff
@@ -1,9 +1,46 @@
...
 	// Archetype for today's workflow implementation.
 	// This value is NOT persisted today, and ok to be changed.
 	Archetype chasm.Archetype = "Workflow"
 )
+
+type Workflow struct {
+	chasm.UnimplementedComponent
+
+	// State of the workflow is managed by mutable_state_impl, not CHASM engine, so this will always be empty.
+	State proto.Message
```
bergundy:
```suggestion
	// State of the workflow is managed by mutable_state_impl, not CHASM engine, so this will always be empty.
	*emptypb.Empty
```

chasm/lib/workflow/workflow.go
```diff
@@ -1,9 +1,46 @@
...
+
+	// MSPointer is a special in-memory field for getting mutable state access.
+	MSPointer chasm.MSPointer
```
bergundy, line 21: No other component is allowed to have this pointer and this is really just a transitional artifact (even if the transition may take years).

service/history/workflow/mutable_state_impl.go
```diff
@@ -2568,6 +2577,12 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 			ms.executionState.RunId, execution.GetRunId())
 	}
 
+	if ms.chasmEnabled() {
+		// Initialize chasm tree once for new workflows.
+		mutableContext := chasm.NewMutableContext(context.TODO(), ms.chasmTree.(*chasm.Node))
```
bergundy: @yycptt again us needing to make up these `context.Context` objects.
@pdoerner I guess you can use `context.Background()` for this but put a comment explaining why.

chasm/lib/workflow/workflow.go
bergundy: Ideally the CHASM framework would panic if any other component tried to use this hack.

chasm/lib/scheduler/scheduler_suite_test.go
```diff
@@ -68,6 +68,7 @@ func (s *schedulerSuite) SetupSuite() {
 	s.nodeBackend.EXPECT().GetCurrentVersion().Return(int64(1)).AnyTimes()
 	s.nodeBackend.EXPECT().UpdateWorkflowStateStatus(gomock.Any(), gomock.Any()).Return(true, nil).AnyTimes()
 	s.nodeBackend.EXPECT().GetWorkflowKey().Return(tv.Any().WorkflowKey()).AnyTimes()
+	s.nodeBackend.EXPECT().IsWorkflow().Return(false).AnyTimes()
```
bergundy, line 71: @lina-temporal, @yycptt this is what you get when you use these tight mocks, every little behavior change needs to be propagated to every test in the codebase.

bergundy (approved):
Had some comments but non of them are blocking.

PR #8486, opened 2025-10-14
Export a guage metric for plugins configured in a worker

rkannan82:
## What changed?
As part of the worker heartbeat, the matching service exports a metric per namespace, plugin name.

Metric: worker_plugin_name
Type: gauge
Dimensions: namespace, plugin_name

This is an optional feature; flag protected and disabled by default.
Flag: matching.enableWorkerPluginMetrics

## Why?
To understand what plugins are being enabled.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
High cardinality: #namespaces x #plugin.
To protect against this, we have guarded this metric behind a dynamic config and disabled it by default.
matching.enableWorkerPluginMetrics. This will be turned in setups that have a scalable implementation of metric collection.

### DIFF
--- a/service/matching/handler.go
+++ b/service/matching/handler.go
@@ -552,8 +552,9 @@ func (h *Handler) RecordWorkerHeartbeat(
 	_ context.Context, request *matchingservice.RecordWorkerHeartbeatRequest,
 ) (*matchingservice.RecordWorkerHeartbeatResponse, error) {
 	nsID := namespace.ID(request.GetNamespaceId())
+	nsName := h.namespaceName(nsID)
 
-	h.workersRegistry.RecordWorkerHeartbeats(nsID, request.GetHeartbeartRequest().GetWorkerHeartbeat())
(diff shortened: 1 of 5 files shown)


common/dynamicconfig/constants.go
```diff
@@ -1339,6 +1339,13 @@ second per poller by one physical queue manager`,
 		2000,
 		"Cache size for fairness key rate limits.",
 	)
+	MatchingEnableWorkerPluginMetrics = NewGlobalBoolSetting(
+		"matching.enableWorkerPluginMetrics",
+		false,
+		`MatchingEnableWorkerPluginMetrics controls whether to export worker plugin metrics. 
+The metric has 2 dimensions: namespace_id and plugin_name. Disabled by default as this is
```
ShahabT, line 1361: we don't use namespace_id in any config or metric AFAIK, and it's not readable, can we change it to namespace name?

rkannan82: Done.

PR #8491, opened 2025-10-15
Add CHASM Visibility registration and task processing

awln-temporal:
## What changed?
Adds CHASM search attribute provider and mapper implementation. To use CHASM search attributes, embed your component with ComponentSearchAttributesProvider and register the search attributes to the CHASM Registry.

To define a component search attribute, define a SearchAttribute as a package/global scoped variable (CHASM search attributes should reference exported variables like `SearchAttributeFieldInt01`, `SearchAttributeFieldDateTime01`:
```
var (
	TestComponentStartTimeSearchAttribute = NewSearchAttributeTime(testComponentStartTimeSAKey, SearchAttributeFieldDateTime01)

	_ VisibilitySearchAttributesProvider = (*TestComponent)(nil)
	_ VisibilityMemoProvider             = (*TestComponent)(nil)
	_ VisibilitySearchAttributesMapper   = (*TestComponent)(nil)
)
```

To update SearchAttributes during an execution of a CHASM archetype, implement the provider `SearchAttributes` in the root component, and call the `NewValue` method on a SearchAttribute within the Provider method to generate a new SearchAttributeValue. On SearchAttribute updates during CHASM transactions, the framework will detect changes and submit a Visibility task for persistence.
```
type TestComponent struct {
	UnimplementedComponent

	...
	Visibility Field[*Visibility]
}


func (tc *TestComponent) SearchAttributes(ctx Context) []SearchAttributeValue {
	return []SearchAttributeValue{
             TestComponentStartTimeSearchAttribute.NewValue(time)
        }
}
```

To register the search attribute with the CHASM Registry, `WithSearchAttributes` must be passed as a RegistrableComponentOption to the root component of the library. This is required to support Visibility queries.

```
		chasm.NewRegistrableComponent[*Scheduler]("scheduler", WIthSearchAttributes([]SearchAttributeDefinition{TestComponentStartTimeSearchAttribute}),
		chasm.NewRegistrableComponent[*Generator]("generator"),
		chasm.NewRegistrableComponent[*Invoker]("invoker"),
		chasm.NewRegistrableComponent[*Backfiller]("backfiller"),
	}
```

## Why?
Required to support CHASM Search Attributes. 

## How did you test it?
- [X] built
- [X] added new unit test(s)

## Potential risks
TBD

### DIFF
--- a/schema/elasticsearch/visibility/index_template_v7.json
+++ b/schema/elasticsearch/visibility/index_template_v7.json
@@ -1,1 +1,1 @@
-./versioned/v9/index_template_v7.json
+./versioned/v11/index_template_v7.json
--- a/schema/elasticsearch/visibility/versioned/v11/index_template_v7.json
+++ b/schema/elasticsearch/visibility/versioned/v11/index_template_v7.json
@@ -1,0 +1,158 @@
+{
+  "order": 0,
+  "index_patterns": ["temporal_visibility_v1*"],
+  "settings": {
+    "index": {
+      "number_of_shards": "1",
+      "number_of_replicas": "0",
+      "auto_expand_replicas": "0-2",
+      "search.idle.after": "365d",
+      "sort.field": ["CloseTime", "StartTime", "RunId"],
+      "sort.order": ["desc", "desc", "desc"],
+      "sort.missing": ["_first", "_first", "_first"]
+    }
+  },
+  "mappings": {
+    "dynamic": "false",
+    "properties": {
+      "NamespaceId": {
+        "type": "keyword"
+      },
+      "TemporalNamespaceDivision": {
+        "type": "keyword"
+      },
+      "WorkflowId": {
+        "type": "keyword"
+      },
+      "RunId": {
+        "type": "keyword"
+      },
+      "WorkflowType": {
+        "type": "keyword"
+      },
+      "StartTime": {
+        "type": "date_nanos"
+      },
+      "ExecutionTime": {
+        "type": "date_nanos"
+      },
+      "CloseTime": {
+        "type": "date_nanos"
+      },
+      "ExecutionDuration": {
+        "type": "long"
+      },
+      "ExecutionStatus": {
+        "type": "keyword"
+      },
+      "TaskQueue": {
+        "type": "keyword"
+      },
+      "TemporalChangeVersion": {
+        "type": "keyword"
+      },
+      "BatcherNamespace": {
+        "type": "keyword"
+      },
+      "BatcherUser": {
+        "type": "keyword"
+      },
+      "BinaryChecksums": {
+        "type": "keyword"
+      },
+      "HistoryLength": {
+        "type": "long"
+      },
+      "StateTransitionCount": {
+        "type": "long"
+      },
+      "TemporalScheduledStartTime": {
+        "type": "date_nanos"
+      },
+      "TemporalScheduledById": {
+        "type": "keyword"
+      },
+      "TemporalSchedulePaused": {
+        "type": "boolean"
+      },
+      "HistorySizeBytes": {
+        "type": "long"
+      },
+      "BuildIds": {
+        "type": "keyword"
+      },
+      "ParentWorkflowId": {
+        "type": "keyword"
+      },
+      "ParentRunId": {
+        "type": "keyword"
+      },
+      "RootWorkflowId": {
+        "type": "keyword"
+      },
+      "RootRunId": {
+        "type": "keyword"
+      },
+      "TemporalPauseInfo": {
+        "type": "keyword"
+      },
+      "TemporalWorkerDeploymentVersion": {
+        "type": "keyword"
+      },
+      "TemporalWorkflowVersioningBehavior": {
+        "type": "keyword"
+      },
+      "TemporalWorkerDeployment": {
+        "type": "keyword"
+      },
+			"TemporalReportedProblems": {
+					"type": "keyword"
+			},
+      "TemporalBool01": {
+        "type": "boolean"
+      },
+      "TemporalBool02": {
+        "type": "boolean"
+      },
+      "TemporalDatetime01": {
+        "type": "date_nanos"
+      },
+      "TemporalDatetime02": {
+        "type": "date_nanos"
+      },
+      "TemporalDouble01": {
+        "type": "scaled_float",
+        "scaling_factor": 10000
+      },
+      "TemporalDouble02": {
+        "type": "scaled_float",
+        "scaling_factor": 10000
+      },
+      "TemporalInt01": {
+        "type": "long"
+      },
+      "TemporalInt02": {
+        "type": "long"
+      },
+      "TemporalKeyword01": {
+        "type": "keyword"
+      },
+      "TemporalKeyword02": {
+        "type": "keyword"
+      },
+      "TemporalKeyword03": {
+        "type": "keyword"
+      },
+      "TemporalKeyword04": {
+        "type": "keyword"
+      },
+      "TemporalKeywordList01": {
+        "type": "keyword"
+      },
+      "TemporalKeywordList02": {
+        "type": "keyword"
+      }
+    }
+  },
+  "aliases": {}
+}
--- a/schema/elasticsearch/visibility/versioned/v11/upgrade.sh
+++ b/schema/elasticsearch/visibility/versioned/v11/upgrade.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+# Prerequisites:
+#   - jq
+#   - curl
+
+# Input parameters.
+: "${ES_SCHEME:=http}"
+: "${ES_SERVER:=127.0.0.1}"
+: "${ES_PORT:=9200}"
+: "${ES_USER:=}"
+: "${ES_PWD:=}"
+: "${ES_VERSION:=v7}"
+: "${ES_VIS_INDEX_V1:=temporal_visibility_v1_dev}"
+: "${AUTO_CONFIRM:=}"
+: "${SLICES_COUNT:=auto}"
+
+es_endpoint="${ES_SCHEME}://${ES_SERVER}:${ES_PORT}"
+
+echo "=== Step 0. Sanity check if Elasticsearch index is accessible ==="
+
+if ! curl --silent --fail --user "${ES_USER}":"${ES_PWD}" "${es_endpoint}/${ES_VIS_INDEX_V1}/_stats/docs" --write-out "\n"; then
+    echo "Elasticsearch index ${ES_VIS_INDEX_V1} is not accessible at ${es_endpoint}."
+    exit 1
+fi
+
+echo "=== Step 1. Add new builtin search attributes ==="
+
+new_mapping='
+{
+  "properties": {
+    "TemporalBool01": {
+      "type": "boolean"
+    },
+    "TemporalBool02": {
+      "type": "boolean"
+    },
+    "TemporalDatetime01": {
+      "type": "date_nanos"
+    },
+    "TemporalDatetime02": {
+      "type": "date_nanos"
+    },
+    "TemporalDouble01": {
+      "type": "scaled_float",
+      "scaling_factor": 10000
+    },
+    "TemporalDouble02": {
+      "type": "scaled_float",
+      "scaling_factor": 10000
+    },
+    "TemporalInt01": {
+      "type": "long"
+    },
+    "TemporalInt02": {
+      "type": "long"
+    },
+    "TemporalKeyword01": {
+      "type": "keyword"
+    },
+    "TemporalKeyword02": {
+      "type": "keyword"
+    },
+    "TemporalKeyword03": {
+      "type": "keyword"
+    },
+    "TemporalKeyword04": {
+      "type": "keyword"
+    },
+    "TemporalKeywordList01": {
+      "type": "keyword"
+    },
+    "TemporalKeywordList02": {
+      "type": "keyword"
+    }
+  }
+}
+'
+
+if [ -z "${AUTO_CONFIRM}" ]; then
+    read -p "Add new builtin search attributes to the index ${ES_VIS_INDEX_V1}? (N/y)" -n 1 -r
+    echo
+else
+    REPLY="y"
+fi
+if [ "${REPLY}" = "y" ]; then
+    curl --silent --fail --user "${ES_USER}":"${ES_PWD}" -X PUT "${es_endpoint}/${ES_VIS_INDEX_V1}/_mapping" -H "Content-Type: application/json" --data-binary "$new_mapping" | jq
+    # Wait for mapping changes to go through.
+    until curl --silent --user "${ES_USER}":"${ES_PWD}" "${es_endpoint}/_cluster/health/${ES_VIS_INDEX_V1}" | jq --exit-status '.status=="green" | .'; do
+        echo "Waiting for Elasticsearch index ${ES_VIS_INDEX_V1} become green."
+        sleep 1
+    done
+fi
--- a/schema/embed.go
+++ b/schema/embed.go
@@ -58,7 +58,7 @@ func ElasticsearchClusterSettings() (string, error) {
 
 // ElasticsearchIndexTemplate returns the embedded index template for Elasticsearch v7 (latest version)
 func ElasticsearchIndexTemplate() (string, error) {
-	data, err := assets.ReadFile("elasticsearch/visibility/versioned/v9/index_template_v7.json")
+	data, err := assets.ReadFile("elasticsearch/visibility/versioned/v11/index_template_v7.json")
 	if err != nil {
 		return "", err
 	}
--- a/schema/mysql/v8/version.go
+++ b/schema/mysql/v8/version.go
@@ -6,4 +6,4 @@ package v8
 const Version = "1.18"
 
 // VisibilityVersion is the MySQL visibility database release version
-const VisibilityVersion = "1.10"
+const VisibilityVersion = "1.11"
--- a/schema/mysql/v8/visibility/schema.sql
+++ b/schema/mysql/v8/visibility/schema.sql
@@ -168,3 +168,55 @@ CREATE FULLTEXT INDEX by_text_03  ON custom_search_attributes (Text03);
 CREATE INDEX by_keyword_list_01   ON custom_search_attributes (namespace_id, (CAST(KeywordList01 AS CHAR(255) ARRAY)));
 CREATE INDEX by_keyword_list_02   ON custom_search_attributes (namespace_id, (CAST(KeywordList02 AS CHAR(255) ARRAY)));
 CREATE INDEX by_keyword_list_03   ON custom_search_attributes (namespace_id, (CAST(KeywordList03 AS CHAR(255) ARRAY)));
+
+CREATE TABLE chasm_search_attributes (
+  namespace_id      CHAR(64)        NOT NULL,
+  run_id            CHAR(64)        NOT NULL,
+  _version          BIGINT          NOT NULL DEFAULT 0,
+  search_attributes JSON            NULL,
+
+  -- Pre-allocated CHASM search attributes
+  TemporalBool01            BOOLEAN         GENERATED ALWAYS AS (search_attributes->"$.TemporalBool01"),
+  TemporalBool02            BOOLEAN         GENERATED ALWAYS AS (search_attributes->"$.TemporalBool02"),
+  TemporalDatetime01        DATETIME(6)     GENERATED ALWAYS AS (
+    CONVERT_TZ(
+      REGEXP_REPLACE(search_attributes->>"$.TemporalDatetime01", 'Z|[+-][0-9]{2}:[0-9]{2}$', ''),
+      SUBSTR(REPLACE(search_attributes->>"$.TemporalDatetime01", 'Z', '+00:00'), -6, 6),
+      '+00:00'
+    )
+  ),
+  TemporalDatetime02        DATETIME(6)     GENERATED ALWAYS AS (
+    CONVERT_TZ(
+      REGEXP_REPLACE(search_attributes->>"$.TemporalDatetime02", 'Z|[+-][0-9]{2}:[0-9]{2}$', ''),
+      SUBSTR(REPLACE(search_attributes->>"$.TemporalDatetime02", 'Z', '+00:00'), -6, 6),
+      '+00:00'
+    )
+  ),
+  TemporalDouble01          DECIMAL(20, 5)  GENERATED ALWAYS AS (search_attributes->"$.TemporalDouble01"),
+  TemporalDouble02          DECIMAL(20, 5)  GENERATED ALWAYS AS (search_attributes->"$.TemporalDouble02"),
+  TemporalInt01             BIGINT          GENERATED ALWAYS AS (search_attributes->"$.TemporalInt01"),
+  TemporalInt02             BIGINT          GENERATED ALWAYS AS (search_attributes->"$.TemporalInt02"),
+  TemporalKeyword01         VARCHAR(255)    GENERATED ALWAYS AS (search_attributes->>"$.TemporalKeyword01"),
+  TemporalKeyword02         VARCHAR(255)    GENERATED ALWAYS AS (search_attributes->>"$.TemporalKeyword02"),
+  TemporalKeyword03         VARCHAR(255)    GENERATED ALWAYS AS (search_attributes->>"$.TemporalKeyword03"),
+  TemporalKeyword04         VARCHAR(255)    GENERATED ALWAYS AS (search_attributes->>"$.TemporalKeyword04"),
+  TemporalKeywordList01     JSON            GENERATED ALWAYS AS (search_attributes->"$.TemporalKeywordList01"),
+  TemporalKeywordList02     JSON            GENERATED ALWAYS AS (search_attributes->"$.TemporalKeywordList02"),
+
+  PRIMARY KEY (namespace_id, run_id)
+);
+
+CREATE INDEX by_temporal_bool_01           ON chasm_search_attributes (namespace_id, TemporalBool01);
+CREATE INDEX by_temporal_bool_02           ON chasm_search_attributes (namespace_id, TemporalBool02);
+CREATE INDEX by_temporal_datetime_01       ON chasm_search_attributes (namespace_id, TemporalDatetime01);
+CREATE INDEX by_temporal_datetime_02       ON chasm_search_attributes (namespace_id, TemporalDatetime02);
+CREATE INDEX by_temporal_double_01         ON chasm_search_attributes (namespace_id, TemporalDouble01);
+CREATE INDEX by_temporal_double_02         ON chasm_search_attributes (namespace_id, TemporalDouble02);
+CREATE INDEX by_temporal_int_01            ON chasm_search_attributes (namespace_id, TemporalInt01);
+CREATE INDEX by_temporal_int_02            ON chasm_search_attributes (namespace_id, TemporalInt02);
+CREATE INDEX by_temporal_keyword_01        ON chasm_search_attributes (namespace_id, TemporalKeyword01);
+CREATE INDEX by_temporal_keyword_02        ON chasm_search_attributes (namespace_id, TemporalKeyword02);
+CREATE INDEX by_temporal_keyword_03        ON chasm_search_attributes (namespace_id, TemporalKeyword03);
+CREATE INDEX by_temporal_keyword_04        ON chasm_search_attributes (namespace_id, TemporalKeyword04);
+CREATE INDEX by_temporal_keyword_list_01   ON chasm_search_attributes (namespace_id, (CAST(TemporalKeywordList01 AS CHAR(255) ARRAY)));
+CREATE INDEX by_temporal_keyword_list_02   ON chasm_search_attributes (namespace_id, (CAST(TemporalKeywordList02 AS CHAR(255) ARRAY)));
--- a/schema/mysql/v8/visibility/versioned/v1.11/add_chasm_search_attributes.sql
+++ b/schema/mysql/v8/visibility/versioned/v1.11/add_chasm_search_attributes.sql
@@ -1,0 +1,51 @@
+CREATE TABLE chasm_search_attributes (
+  namespace_id      CHAR(64)        NOT NULL,
+  run_id            CHAR(64)        NOT NULL,
+  _version          BIGINT          NOT NULL DEFAULT 0,
+  search_attributes JSON            NULL,
+
+  -- Pre-allocated CHASM search attributes
+  TemporalBool01            BOOLEAN         GENERATED ALWAYS AS (search_attributes->"$.TemporalBool01"),
+  TemporalBool02            BOOLEAN         GENERATED ALWAYS AS (search_attributes->"$.TemporalBool02"),
+  TemporalDatetime01        DATETIME(6)     GENERATED ALWAYS AS (
+    CONVERT_TZ(
+      REGEXP_REPLACE(search_attributes->>"$.TemporalDatetime01", 'Z|[+-][0-9]{2}:[0-9]{2}$', ''),
+      SUBSTR(REPLACE(search_attributes->>"$.TemporalDatetime01", 'Z', '+00:00'), -6, 6),
+      '+00:00'
+    )
+  ),
+  TemporalDatetime02        DATETIME(6)     GENERATED ALWAYS AS (
+    CONVERT_TZ(
+      REGEXP_REPLACE(search_attributes->>"$.TemporalDatetime02", 'Z|[+-][0-9]{2}:[0-9]{2}$', ''),
+      SUBSTR(REPLACE(search_attributes->>"$.TemporalDatetime02", 'Z', '+00:00'), -6, 6),
+      '+00:00'
+    )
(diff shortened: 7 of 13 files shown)


chasm/registrable_component.go
```diff
@@ -56,6 +57,32 @@ func WithShardingFn(
...
+func WithSearchAttributes(searchAttributes []*SearchAttribute) RegistrableComponentOption {
+	return func(rc *RegistrableComponent) {
+		if len(searchAttributes) == 0 {
+			return
+		}
+
+		aliasToKey := make(map[string]string)
+
+		for _, sa := range searchAttributes {
+			alias := sa.GetAlias()
```
bergundy: Alias is optional.

chasm/lib/tests/payload.go
```diff
@@ -148,12 +148,12 @@ func (s *PayloadStore) LifecycleState(
...
 func (s *PayloadStore) SearchAttributes(
 	_ chasm.Context,
-) map[string]chasm.VisibilityValue {
+) []*chasm.SearchAttribute {
 	// TODO: UpsertSearchAttribute as well when CHASM framework supports Per-Component SearchAttributes
 	// For now, we just update a random existing pre-defined SA to make sure the logic works.
-	return map[string]chasm.VisibilityValue{
-		TestKeywordSAFieldName: chasm.VisibilityValueString(TestKeywordSAFieldValue),
-	}
+	attr := chasm.NewSearchAttributeKeywordByAlias(TestKeywordSAFieldName, TestKeywordSAFieldName)
```
bergundy: This should be a global `var` so it can be registered with the library.

chasm/lib/tests/payload.go
```diff
@@ -148,12 +148,12 @@ func (s *PayloadStore) LifecycleState(
...
+	attr.SetValue(TestKeywordSAFieldValue)
+	return []*chasm.SearchAttribute{&attr.SearchAttribute}
```
bergundy: You would want the experience to be `return []chasm.SearchAttributeKeyValue{attr.Value("some-value")}`

bergundy (commented):
Started reviewing but I think we still don't agree on the API, so let's discuss that before and agree before implementing anything.

chasm/lib/tests/payload.go
awln-temporal: EDIT: Below code block is out of date, refer to comments below for updated implementation

From the latest revision, the current experience is for Component authors to update via
```
type CustomComponent struct {
  UnimplementedComponent
  // Must embed to upsert search attributes
  ComponentSearchAttributesProvider
}

func (c *CustomComponent) updateComponent() {
  ...
  c.UpsertSearchAttributes(SEARCH_ATTRIBUTE_KEY.newValue(TEST_VALUE))
  ...
}
```
the embedded ComponentSearchAttributesProvider will provide the default implementation of VisibilitySearchAttributesProvider and VisibilityMemoProvider

chasm/lib/tests/payload.go
awln-temporal: moved to a global var

chasm/registrable_component.go
awln-temporal: Alias and Field name are both required fields right now. We can make either alias or field name optional when they are the same (eg. System search attributes), if that is preferred. As of now, this is supported via the constructor specifying the field name explicitly, where the attribute can explicitly map from same alias -> field name.

chasm/lib/tests/payload.go
awln-temporal: UPDATE: Removed ComponentSearchAttributesProvider embedded struct, and following pattern for component authors to implement the SearchAttributes method like
```
func (tc *TestComponent) SearchAttributes(ctx Context) []SearchAttributeValue {
	return []SearchAttributeValue{
		testComponentStartTimeSearchAttribute.NewValue(tc.ComponentData.GetStartTime().AsTime()),
	}
}
```

chasm/registrable_component.go
```diff
@@ -66,3 +115,25 @@ func (rc RegistrableComponent) fqType() string {
 	}
 	return fullyQualifiedName(rc.library.Name(), rc.componentType)
 }
+
+// GetAlias returns the search attribute alias for the given field name.
+func (rc *RegistrableComponent) GetAlias(field string) (string, error) {
```
bergundy:
```suggestion
func (rc *RegistrableComponent) SearchAttributeAlias(field string) (string, error) {
```

chasm/registrable_component.go
```diff
@@ -66,3 +115,25 @@ func (rc RegistrableComponent) fqType() string {
...
+	if !ok {
+		return "", serviceerror.NewInvalidArgument(
+			fmt.Sprintf("RegistrableComponent name %s has no alias defined for field name %s", rc.fqType(), field),
+		)
+	}
+	return alias, nil
+}
+
+// GetField returns the search attribute field name for the given alias.
+func (rc *RegistrableComponent) GetField(alias string) (string, error) {
```
bergundy:
```suggestion
func (rc *RegistrableComponent) SearchAttributeField(alias string) (string, error) {
```

chasm/search_attribute.go
```diff
@@ -0,0 +1,336 @@
...
+	SearchAttributeFieldKeywordList01 = newSearchAttributeFieldKeywordList(1)
+	SearchAttributeFieldKeywordList02 = newSearchAttributeFieldKeywordList(2)
+)
+
+type (
+	SearchAttribute interface {
+		getAlias() string
+		getField() string
+		getValueType() enumspb.IndexedValueType
+		mustEmbedSearchAttributeDefinition()
```
bergundy:
```suggestion
		definition() searchAttributeDefinition
```

chasm/registrable_component.go
```diff
@@ -56,6 +67,44 @@ func WithShardingFn(
...
+
+		for _, sa := range searchAttributes {
+			alias := sa.getAlias()
+			field := sa.getField()
+			valueType := sa.getValueType()
+
+			rc.aliasToField[alias] = field
+			rc.fieldToAlias[field] = alias
+			rc.saTypeMap[field] = valueType
+		}
```
rodrigozhou: You cannot create these maps before validating. It's possible to create an invalid mapping that passes the validation. Example: `Foo -> Field1, Foo -> Field2, Bar -> Field1`

chasm/search_attribute.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// GetValueType returns the indexed value type.
+func (s *searchAttributeDefinition) getValueType() enumspb.IndexedValueType {
+	return s.valueType
+}
+
+func (s *searchAttributeDefinition) mustEmbedSearchAttributeDefinition() {}
+
+func NewSearchAttributeBool(alias string, boolField SearchAttributeFieldBool) *SearchAttributeBool {
```
bergundy: Don't return pointers, these are just data bundle structs and are not supposed to be mutable.

chasm/search_attribute.go
```diff
@@ -0,0 +1,346 @@
...
+	return &SearchAttributeBool{
+		searchAttributeDefinition: searchAttributeDefinition{
+			alias:     alias,
+			field:     boolField.field,
+			valueType: enumspb.INDEXED_VALUE_TYPE_BOOL,
+		},
+	}
+}
+
+func NewSearchAttributeBoolByField(field string) *SearchAttributeBool {
```
bergundy: Don't expose these constructors, create vars for all of the predefined fields that we want to use as we discussed.

chasm/search_attribute.go
bergundy, line 1: Please document all of the exported members so chasm library authors understand how to use this library.

chasm/search_attribute.go
```diff
@@ -0,0 +1,346 @@
...
+	searchAttributeDefinition struct {
+		// alias refers to the user defined name of the search attribute
+		alias string
+		// field refers to a fully formed schema field, which is either a Predefined or CHASM search attribute
+		field     string
+		valueType enumspb.IndexedValueType
+	}
+
+	SearchAttributeKeyValue struct {
+		Field string
```
bergundy, line 95: Can you also add the alias here for debugging? Component authors will be more familiar with the alias than the mapped field.

chasm/test_component_test.go
```diff
@@ -95,9 +102,9 @@ func (tc *TestComponent) Fail(_ MutableContext) {
 }
 
 // SearchAttributes implements VisibilitySearchAttributesProvider interface.
-func (tc *TestComponent) SearchAttributes(_ Context) map[string]VisibilityValue {
-	return map[string]VisibilityValue{
-		testComponentStartTimeSAKey: VisibilityValueTime(tc.ComponentData.GetStartTime().AsTime()),
+func (tc *TestComponent) SearchAttributes(_ Context) []SearchAttributeKeyValue {
+	return []SearchAttributeKeyValue{
+		testComponentStartTimeSearchAttribute.Value(tc.ComponentData.GetStartTime().AsTime()),
```
bergundy: Can you add a search attribute that is mapped to an existing non-chasm pre-defined field for completeness of this test?

chasm/visibility.go
```diff
@@ -31,6 +29,11 @@ type VisibilityMemoProvider interface {
 	Memo(Context) map[string]VisibilityValue
 }
 
+type VisibilitySearchAttributesMapper interface {
+	GetAlias(field string) (string, error)
+	GetField(alias string) (string, error)
```
bergundy:
```suggestion
	SearchAttributeAlias(field string) (string, error)
	SearchAttributeField(alias string) (string, error)
```

chasm/search_attribute.go
```diff
@@ -0,0 +1,346 @@
...
+	case enumspb.INDEXED_VALUE_TYPE_DATETIME:
+		return "TemporalDatetime" + suffix
+	case enumspb.INDEXED_VALUE_TYPE_DOUBLE:
+		return "TemporalDouble" + suffix
+	case enumspb.INDEXED_VALUE_TYPE_INT:
+		return "TemporalInt" + suffix
+	case enumspb.INDEXED_VALUE_TYPE_KEYWORD:
+		return "TemporalKeyword" + suffix
+	case enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST:
+		return "TemporalKeywordList" + suffix
```
bergundy: I think you can replace this switch statement with the `valueType.String()` method.

bergundy (commented):
Overall looks great.

chasm/search_attribute.go
awln-temporal: the only reason i left it public was a test case was creating and validating some search attribute. In this case, do you think it makes sense to create some separate test_constants file that has this search attribute, or should we follow the pattern of having a test file contain all its needed state and keep it public?

chasm/search_attribute.go
awln-temporal: I didn't see a huge benefit to using `valueType.String()`, since there technically isn't any coupling with the IndexedValueType and the schema column names either. Those values are also hardcoded, so I left this hardcoded as well, let me know your thoughts.

chasm/search_attribute.go
awln-temporal: is the suggestion to call `searchAttribute.definition().getAlias()` or `searchAttribute.definition().Alias` when looping through `WithSearchAttributes`, and also make similar calls during test case verification? I kind of like the current style a little better since it's more concise, is it generally better style to have data fields and accessors within a struct (not expose getters in interface in Go)?

chasm/search_attribute.go
awln-temporal: nvm converted to use valueType.String()

chasm/search_attribute.go
bergundy: You don't need methods if you can access members directly.

Also don't name getters with a `get` prefix in Go: https://go.dev/doc/effective_go#Getters.

chasm/search_attribute.go
bergundy: Create the vars in this package. There are none that are test specific.

chasm/search_attribute.go
bergundy: I saw that's what we do in other visibility code.

chasm/visibility.go
rodrigozhou: Isn't it redundant to have functions here prefixed with `SearchAttribute` since the interface is already called `VisibilitySearchAttributesMapper`? I think `GetAlias/GetFieldName` are in this case.

chasm/registrable_component.go
rodrigozhou: Also replied in the other comment, I think the only reason to have the `SearchAttribute` prefix in this function name is because it's making `RegistrableComponent` implement the `VisibilitySearchAttributesMapper` interface, which I'm not sure I like this approach.

I'd rather if `RegistrableComponent` had an actual var for the mapper. In fact, I'm wondering if the interface is even needed.

chasm/visibility.go
```diff
@@ -31,6 +32,45 @@ type VisibilityMemoProvider interface {
...
+func (v *VisibilitySearchAttributesMapper) GetFieldValueType(field string) (enumspb.IndexedValueType, error) {
+	if v == nil {
+		return enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, serviceerror.NewInvalidArgument("visibility search attributes mapper is not registered")
+	}
+	saType, ok := v.saTypeMap[field]
+	if !ok {
+		return enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED, serviceerror.NewInvalidArgument(fmt.Sprintf("visibility search attributes mapper has no registered field %s", field))
+	}
+	return saType, nil
+}
```
bergundy: Why do you need to export these methods? And please don't use `Get` as a prefix for getters in Go.

chasm/search_attribute.go
```diff
@@ -0,0 +1,340 @@
...
+	_ SearchAttribute = (*searchAttributeDefinition)(nil)
+	_ SearchAttribute = (*SearchAttributeBool)(nil)
+	_ SearchAttribute = (*SearchAttributeDateTime)(nil)
+	_ SearchAttribute = (*SearchAttributeInt)(nil)
+	_ SearchAttribute = (*SearchAttributeDouble)(nil)
+	_ SearchAttribute = (*SearchAttributeKeyword)(nil)
+	_ SearchAttribute = (*SearchAttributeKeywordList)(nil)
+)
+
+type (
```
bergundy, line 77: nit: don't group all of the types together. The practice in Go is to put the type definition above its constructor followed by all of the methods.

chasm/search_attribute.go
```diff
@@ -0,0 +1,340 @@
...
+	SearchAttribute interface {
+		definition() searchAttributeDefinition
+	}
+
+	searchAttributeDefinition struct {
+		// alias refers to the user defined name of the search attribute
+		Alias string
+		// field refers to a fully formed schema field, which is either a Predefined or CHASM search attribute
+		Field     string
+		ValueType enumspb.IndexedValueType
```
bergundy: Do these fields need to be exported? They're defined on a struct that isn't exported.

chasm/lib/tests/payload.go
```diff
@@ -24,6 +24,8 @@ const (
 )
 
 var (
+	TestKeywordSearchAttribute = chasm.NewSearchAttributeKeyword(TestKeywordSAFieldName, chasm.SearchAttributeFieldKeyword01)
```
bergundy: We need another search attribute to test aliasing.
Also, don't use `TemporalScheduledById` as the mapped attribute.

You're going to want:
1. Define `SearchAttributeTemporalScheduledById = newSearchAttributeKeywordByField(searchattribute.TemporalScheduledById)` in `chasm/search_attribute.go`
2. Define a separate search attribute that will map to `SearchAttributeFieldKeyword01` for testing aliasing
3. Register both with the component
4. Emit both from the component
5. Test both in functional tests

bergundy (commented):
This is already very close.
Please document all exported fields from the `chasm` package.

chasm/visibility.go
awln-temporal: On the read path, custom search attributes are mapped to field names in the visibility store. If translation happens in either some FrontEnd service handler or Visibility Store that isn't in the `chasm` package, we might need to keep these as exported? Or, can we expect that these translations would happen within the `chasm` package?

I'll rename the methods as well.

chasm/visibility.go
bergundy: Don't use `Get` for getters in Go please.

chasm/registrable_component.go
```diff
@@ -56,6 +61,40 @@ func WithShardingFn(
...
+
+		for _, sa := range searchAttributes {
+			alias := sa.definition().alias
+			field := sa.definition().field
+			valueType := sa.definition().valueType
+
+			if _, ok := rc.searchAttributesMapper.aliasToField[alias]; ok {
+				//nolint:forbidigo
+				panic(fmt.Sprintf("registrable component validation error: search attribute alias %s is already defined", alias))
```
rodrigozhou: nit
```suggestion
				panic(fmt.Sprintf("registrable component validation error: search attribute alias %q is already defined", alias))
```

chasm/registrable_component.go
```diff
@@ -56,6 +61,40 @@ func WithShardingFn(
...
+			}
+			if _, ok := rc.searchAttributesMapper.fieldToAlias[field]; ok {
+				//nolint:forbidigo
+				panic(fmt.Sprintf("registrable component validation error: search attribute field %s is already defined", field))
```
rodrigozhou: nit
```suggestion
				panic(fmt.Sprintf("registrable component validation error: search attribute field %q is already defined", field))
```

chasm/lib/tests/payload.go
```diff
@@ -15,15 +14,15 @@ const (
...
 const (
-	TestKeywordSAFieldName  = searchattribute.TemporalScheduledById
-	TestKeywordSAFieldValue = "test-keyword-value"
+	PayloadTotalCountSAAlias = "PayloadTotalCount"
+	PayloadTotalSizeSAAlias  = "PayloadTotalSize"
 )
 
 var (
+	PayloadTotalCountSearchAttribute = chasm.NewSearchAttributeInt(PayloadTotalCountSAAlias, chasm.SearchAttributeFieldInt01)
+	PayloadTotalSizeSearchAttribute  = chasm.NewSearchAttributeInt(PayloadTotalSizeSAAlias, chasm.SearchAttributeFieldInt02)
```
bergundy, line 25: Can you use one attribute that is an alias to an existing pre-defined field for completeness? I feel like you're not understanding what I'm asking.

chasm/search_attribute.go
```diff
@@ -0,0 +1,358 @@
...
+
+	SearchAttributeFieldKeyword01 = newSearchAttributeFieldKeyword(1)
+	SearchAttributeFieldKeyword02 = newSearchAttributeFieldKeyword(2)
+	SearchAttributeFieldKeyword03 = newSearchAttributeFieldKeyword(3)
+	SearchAttributeFieldKeyword04 = newSearchAttributeFieldKeyword(4)
+
+	SearchAttributeFieldKeywordList01 = newSearchAttributeFieldKeywordList(1)
+	SearchAttributeFieldKeywordList02 = newSearchAttributeFieldKeywordList(2)
+
+	SearchAttributeTemporalScheduledByID = newSearchAttributeKeywordByField(searchattribute.TemporalScheduledById)
```
bergundy: I mean use this attribute please for the test component.

chasm/search_attribute.go
```diff
@@ -0,0 +1,358 @@
...
+func (s searchAttributeDefinition) definition() searchAttributeDefinition {
+	return s
+}
+
+// SearchAttributeBool is a search attribute for a boolean value.
+type SearchAttributeBool struct {
+	searchAttributeDefinition
+}
+
+func NewSearchAttributeBool(alias string, boolField SearchAttributeFieldBool) SearchAttributeBool {
```
bergundy, line 182: Document these constructors as well please.

chasm/visibility.go
```diff
@@ -31,6 +32,41 @@ type VisibilityMemoProvider interface {
 	Memo(Context) map[string]VisibilityValue
 }
 
+type VisibilitySearchAttributesMapper struct {
```
bergundy, line 36: Does this type need to be exported? If so, please document it and all of its methods.

bergundy (approved):
Approved with comments. Overall this looks great. Please add the missing docstrings and test coverage before merging.

chasm/lib/tests/payload.go
awln-temporal: apologies, there was a reference to use a "predefined" field (not TemporalInt01 etc) in test_component_test.go. I also added a reference to the predefined field `SearchAttributeTemporalScheduledByID`.

As for whether to include it in Payload.go, there was a previous TODO to track `Count` and `Size` as search attributes, and I wasn't sure that adding a reference to `TemporalScheduledById` made sense within the scope of the Payload component. This component only tracks the current count and size of payloads, and a map from the PayloadKey to its Payload blob. What would be the values provided to the search attribute?

I agree that functional tests should cover this case, I just wasn't sure if it was acceptable style to add a search attribute that doesn't seem related to the component. Added a reference in payload.go to this.

chasm/search_attribute.go
rodrigozhou: Also, don't we want to define those vars for all the reserved search attributes (`TemporalScheduledByID`, `BuildIds`, etc.)?

chasm/visibility.go
```diff
@@ -31,6 +32,45 @@ type VisibilityMemoProvider interface {
...
+	aliasToField map[string]string
+	fieldToAlias map[string]string
+	saTypeMap    map[string]enumspb.IndexedValueType
+}
+
+// Alias returns the alias for a given field.
+func (v *VisibilitySearchAttributesMapper) Alias(field string) (string, error) {
+	if v == nil {
+		return "", serviceerror.NewInvalidArgument("visibility search attributes mapper is not registered")
```
rodrigozhou:
```suggestion
		return "", serviceerror.NewInvalidArgument("visibility search attributes mapper not defined")
```

chasm/visibility.go
```diff
@@ -31,6 +32,45 @@ type VisibilityMemoProvider interface {
...
+	if !ok {
+		return "", serviceerror.NewInvalidArgument(fmt.Sprintf("visibility search attributes mapper has no registered field %s", field))
+	}
+	return alias, nil
+}
+
+// Field returns the field for a given alias.
+func (v *VisibilitySearchAttributesMapper) Field(alias string) (string, error) {
+	if v == nil {
+		return "", serviceerror.NewInvalidArgument("visibility search attributes mapper is not registered")
```
rodrigozhou:
```suggestion
		return "", serviceerror.NewInvalidArgument("visibility search attributes mapper not defined")
```

chasm/visibility_value.go
```diff
@@ -17,6 +19,7 @@ type VisibilityValueInt int
 
 func (v VisibilityValueInt) MustEncode() *commonpb.Payload {
 	p, _ := payload.Encode(int(v))
+	searchattribute.SetMetadataType(p, enumspb.INDEXED_VALUE_TYPE_INT)
 	return p
 }
```
rodrigozhou, line 25: Do we need the `int` and `int32` types? cc: @bergundy @yycptt

common/searchattribute/defs.go
```diff
@@ -279,6 +279,18 @@ func IsPreallocatedCSAFieldName(name string, valueType enumspb.IndexedValueType)
 	return re != nil && re.MatchString(name)
 }
 
+var chasmSearchAttributePattern = regexp.MustCompile(`^Temporal(Bool|Datetime|Int|Double|Keyword|KeywordList)(0[1-9]|[1-9][0-9])$`)
```
rodrigozhou: I'd add the `Text` type here too, even though there are no `Text` type search attributes yet.

common/persistence/visibility/store/sql/visibility_store.go
```diff
@@ -471,7 +471,12 @@ func (s *VisibilityStore) processRowSearchAttributes(
...
 		tp, err := saTypeMap.GetType(name)
 		if err != nil {
-			return nil, err
+			// TODO: CHASM search attributes are not in the typeMap and SQL only stores raw values (no metadata).
+			// The Encode() call below will fail to add type metadata, causing decode issues.
+			if !searchattribute.IsChasmSearchAttribute(name) {
+				return nil, err
+			}
+			continue
 		}
```
rodrigozhou, line 480: Wouldn't this make more sense?


```suggestion
		// TODO: CHASM search attributes are not in the typeMap and SQL only stores raw values (no metadata).
		// The Encode() call below will fail to add type metadata, causing decode issues.
		if searchattribute.IsChasmSearchAttribute(name) {
			continue
		}
		tp, err := saTypeMap.GetType(name)
		if err != nil {
			return nil, err
		}
```

common/persistence/visibility/store/sql/visibility_store.go
rodrigozhou, line 470: I suppose you can't just skip the CHASM search attributes based on this comment.

common/searchattribute/encode.go
```diff
@@ -28,11 +28,14 @@ func Encode(searchAttributes map[string]interface{}, typeMap *NameTypeMap) (*com
 		saType := enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED
 		if typeMap != nil {
 			saType, err = typeMap.getType(saName, customCategory|predefinedCategory)
-			if err != nil {
+			if err != nil && !IsChasmSearchAttribute(saName) {
 				lastErr = err
 				continue
 			}
```
rodrigozhou, line 34: Ditto, I think it's better that you check if it's a CHASM search attribute before checking the type. Also, in this code, if it's a CHASM search attribute, `saType` will be `UNSPECIFIED`, so you'd be setting the wrong type for CHASM search attributes.

common/searchattribute/encode.go
```diff
@@ -59,7 +62,11 @@ func Decode(
 			var err error
 			saType, err = typeMap.getType(saName, customCategory|predefinedCategory)
 			if err != nil {
-				lastErr = err
+				// If the search attribute name is not in typeMap but the payload has type metadata,
+				// we can still decode it (e.g., for CHASM search attributes).
+				if _, hasTypeMetadata := saPayload.Metadata[MetadataType]; !hasTypeMetadata {
+					lastErr = err
+				}
```
rodrigozhou, line 69: ditto
```suggestion
				// If the search attribute name is not in typeMap but the payload has type metadata,
				// we can still decode it (e.g., for CHASM search attributes).
				saType = saPayload.Metadata[MetadataType]
				if saType == enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED {
					lastErr = err
				}
```

common/searchattribute/encode.go
rodrigozhou, line 52: Please add unit tests for changes you did.

chasm/visibility_value.go
bergundy: We just need one type per anything that is allowed in visibility.

rodrigozhou: So, we just need `int64`.

common/searchattribute/encode.go
awln-temporal: If `UNSPECIFIED`, `SetMetadataType` will return early. To be clear, this is just temp code that will need to change when the query path is implemented.


```
func SetMetadataType(p *commonpb.Payload, t enumspb.IndexedValueType) {
	if t == enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED {
		return
	}

	_, isValidT := enumspb.IndexedValueType_name[int32(t)]
	if !isValidT {
		panic(fmt.Sprintf("unknown index value type %v", t))
	}
	p.Metadata[MetadataType] = []byte(enumspb.IndexedValueType(t).String())
}
```

common/persistence/visibility/store/sql/visibility_store.go
awln-temporal: I can change it, I just didn't think this portion was too relevant since I'll be changing this entire section when working on the Query path.

common/searchattribute/encode.go
awln-temporal: this isn't necessary, `DecodeValue` handles this case

```
func DecodeValue(
	value *commonpb.Payload,
	t enumspb.IndexedValueType,
	allowList bool,
) (any, error) {
	if t == enumspb.INDEXED_VALUE_TYPE_UNSPECIFIED {
		var err error
		t, err = enumspb.IndexedValueTypeFromString(string(value.Metadata[MetadataType]))
		if err != nil {
			return nil, fmt.Errorf("%w: %v", ErrInvalidType, t)
		}
	}

	switch t {
	case enumspb.INDEXED_VALUE_TYPE_BOOL:
		return decodeValueTyped[bool](value, allowList)
	case enumspb.INDEXED_VALUE_TYPE_DATETIME:
		return decodeValueTyped[time.Time](value, allowList)
	case enumspb.INDEXED_VALUE_TYPE_DOUBLE:
		return decodeValueTyped[float64](value, allowList)
	case enumspb.INDEXED_VALUE_TYPE_INT:
		return decodeValueTyped[int64](value, allowList)
	case enumspb.INDEXED_VALUE_TYPE_KEYWORD:
		return validateStrings(decodeValueTyped[string](value, allowList))
	case enumspb.INDEXED_VALUE_TYPE_TEXT:
		return validateStrings(decodeValueTyped[string](value, allowList))
	case enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST:
		return validateStrings(decodeValueTyped[[]string](value, false))
	default:
		return nil, fmt.Errorf("%w: %v", ErrInvalidType, t)
	}
}
```

common/persistence/visibility/store/sql/visibility_store.go
awln-temporal: aka revert this change

chasm/search_attribute.go
awln-temporal: all "predefined" ones? What about system ones? Should we just allow them to be added as needed?

rodrigozhou: The system search attributes are set by Visibility itself, no one else can set them.

common/searchattribute/encode.go
rodrigozhou: This is in the write path, not query path. What I mean is this code will always set the type for CHASM search attribute as UNSPECIFIED, and as you pointed out, the function will just ignore, ie., the CHASM search attributes will never have its type set in the metadata.
Or did I miss something?

common/persistence/visibility/store/sql/visibility_store.go
awln-temporal: This is temporary, we won't skip CHASM search attributes once we consolidate the type maps on the query path.

common/searchattribute/encode.go
awln-temporal: Encode is on the read path. I suppose either way, the MetadataType is somewhat ignored on the write path, and we basically reconstruct the MetadataType solely from the `saTypeMap` right now? Either way, without merging the `saTypeMap` with the CHASM component's `saTypeMap`, current functionality won't work. On the next PR to support querying, we can re-evaluate how we want to support setting the MetadataType. This portion of code is mainly to get functional tests to succeed.

common/searchattribute/encode.go
awln-temporal: there are unit test validations for the changes. Previously, key2 would have failed decoding due it missing in the saTypeMap

chasm/visibility.go
```diff
@@ -31,6 +32,45 @@ type VisibilityMemoProvider interface {
...
+}
+
+// Alias returns the alias for a given field.
+func (v *VisibilitySearchAttributesMapper) Alias(field string) (string, error) {
+	if v == nil {
+		return "", serviceerror.NewInvalidArgument("visibility search attributes mapper not defined")
+	}
+	alias, ok := v.fieldToAlias[field]
+	if !ok {
+		return "", serviceerror.NewInvalidArgument(fmt.Sprintf("visibility search attributes mapper has no registered field %s", field))
```
rodrigozhou: nit
```suggestion
		return "", serviceerror.NewInvalidArgument(fmt.Sprintf("visibility search attributes mapper has no registered field %q", field))
```

chasm/visibility.go
```diff
@@ -31,6 +32,45 @@ type VisibilityMemoProvider interface {
...
+}
+
+// Field returns the field for a given alias.
+func (v *VisibilitySearchAttributesMapper) Field(alias string) (string, error) {
+	if v == nil {
+		return "", serviceerror.NewInvalidArgument("visibility search attributes mapper not defined")
+	}
+	field, ok := v.aliasToField[alias]
+	if !ok {
+		return "", serviceerror.NewInvalidArgument(fmt.Sprintf("visibility search attributes mapper has no registered alias %s", alias))
```
rodrigozhou: ditto

rodrigozhou (approved):

common/searchattribute/encode.go
rodrigozhou: We need to revert this change. This is swallowing errors when the custom search attribute is not defined in the `typeMap`.

The correct behavior is: if the `typeMap` input is not nil, `Decode` checks everything against `typeMap`; otherwise, it uses the payload metadata if available.

awln-temporal: just fmi, in which cases would the `typeMap` be nil, and we would then be permitted to decode the value based off the payload `Metadata`? I would have assumed we would've thrown an error when translating from alias to field upstream, but it seems that isn't the case.

PR #8499, opened 2025-10-16
chasm(callback): implement backoff and invocation executor

spkane31:
## What changed?
Adding an `InvocationTaskExecutor` and `BackoffTaskExecutor` to `chasm/lib/callback`

## Why?
Second step of migrating callback from HSM -> CHASM

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
None, this is not integrated.

### DIFF
--- a/chasm/lib/callback/component.go
+++ b/chasm/lib/callback/component.go
@@ -15,6 +15,9 @@ const (
 	Archetype chasm.Archetype = "Callback"
 )
 
+var _ chasm.Component = (*Callback)(nil)
+var _ chasm.StateMachine[callbackspb.CallbackStatus] = (*Callback)(nil)
+
 // Callback represents a callback component in CHASM.
 type Callback struct {
 	chasm.UnimplementedComponent
@@ -104,26 +107,26 @@ func (c *Callback) loadInvocationArgs(
 func (c *Callback) saveResult(
 	ctx chasm.MutableContext,
 	result invocationResult,
-) (struct{}, error) {
+) (chasm.NoValue, error) {
 	switch r := result.(type) {
 	case invocationResultOK:
 		err := TransitionSucceeded.Apply(ctx, c, EventSucceeded{Time: ctx.Now(c)})
-		return struct{}{}, err
+		return nil, err
 	case invocationResultRetry:
 		err := TransitionAttemptFailed.Apply(ctx, c, EventAttemptFailed{
 			Time:        ctx.Now(c),
 			Err:         r.err,
 			RetryPolicy: r.retryPolicy,
 		})
-		return struct{}{}, err
+		return nil, err
 	case invocationResultFail:
 		err := TransitionFailed.Apply(ctx, c, EventFailed{
 			Time: ctx.Now(c),
 			Err:  r.err,
 		})
-		return struct{}{}, err
+		return nil, err
 	default:
-		return struct{}{}, queues.NewUnprocessableTaskError(
+		return nil, queues.NewUnprocessableTaskError(
 			fmt.Sprintf("unrecognized callback result %v", result),
 		)
 	}
--- a/chasm/lib/callback/metrics.go
+++ b/chasm/lib/callback/metrics.go
@@ -3,10 +3,10 @@ package callback
 import "go.temporal.io/server/common/metrics"
 
 var RequestCounter = metrics.NewCounterDef(
-	"chasm_callback_outbound_requests",
-	metrics.WithDescription("The number of CHASM outbound callback requests made by the history service."),
+	"callback_outbound_requests",
+	metrics.WithDescription("The number of outbound callback requests made by the history service."),
 )
 var RequestLatencyHistogram = metrics.NewTimerDef(
-	"chasm_callback_outbound_latency",
+	"callback_outbound_latency",
 	metrics.WithDescription("Latency histogram of CHASM outbound callback requests made by the history service."),
 )
--- a/chasm/lib/callback/statemachine.go
+++ b/chasm/lib/callback/statemachine.go
@@ -29,9 +29,9 @@ type EventRescheduled struct{}
 var TransitionRescheduled = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_BACKING_OFF},
 	callbackspb.CALLBACK_STATUS_SCHEDULED,
-	func(cb *Callback, mctx chasm.MutableContext, event EventRescheduled) error {
+	func(cb *Callback, ctx chasm.MutableContext, event EventRescheduled) error {
 		cb.NextAttemptScheduleTime = nil
-		mctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{Attempt: cb.Attempt})
+		ctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{Attempt: cb.Attempt})
 		return nil
 	},
 )
@@ -46,7 +46,7 @@ type EventAttemptFailed struct {
 var TransitionAttemptFailed = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_SCHEDULED},
 	callbackspb.CALLBACK_STATUS_BACKING_OFF,
-	func(cb *Callback, mctx chasm.MutableContext, event EventAttemptFailed) error {
+	func(cb *Callback, ctx chasm.MutableContext, event EventAttemptFailed) error {
 		cb.recordAttempt(event.Time)
 		// Use 0 for elapsed time as we don't limit the retry by time (for now).
 		nextDelay := event.RetryPolicy.ComputeNextDelay(0, int(cb.Attempt), event.Err)
@@ -60,7 +60,7 @@ var TransitionAttemptFailed = chasm.NewTransition(
 				},
 			},
 		}
-		mctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{})
+		ctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{})
 		return nil
 	},
 )
@@ -74,7 +74,7 @@ type EventFailed struct {
 var TransitionFailed = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_SCHEDULED},
 	callbackspb.CALLBACK_STATUS_FAILED,
-	func(cb *Callback, mctx chasm.MutableContext, event EventFailed) error {
+	func(cb *Callback, ctx chasm.MutableContext, event EventFailed) error {
 		cb.recordAttempt(event.Time)
 		cb.LastAttemptFailure = &failurepb.Failure{
 			Message: event.Err.Error(),
@@ -84,7 +84,7 @@ var TransitionFailed = chasm.NewTransition(
 				},
 			},
 		}
-		mctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{})
+		ctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{})
 		return nil
 	},
 )
@@ -97,10 +97,10 @@ type EventSucceeded struct {
 var TransitionSucceeded = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_SCHEDULED},
 	callbackspb.CALLBACK_STATUS_SUCCEEDED,
-	func(cb *Callback, mctx chasm.MutableContext, event EventSucceeded) error {
+	func(cb *Callback, ctx chasm.MutableContext, event EventSucceeded) error {
 		cb.recordAttempt(event.Time)
 		cb.LastAttemptFailure = nil
-		mctx.AddTask(cb, chasm.TaskAttributes{}, &callbackspb.InvocationTask{})
+		ctx.AddTask(cb, chasm.TaskAttributes{}, &callbackspb.InvocationTask{})
 		return nil
 	},
 )


chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+) error {
+	var ns *namespace.Namespace
+	// var invoker *Invoker
+	var callback *Callback
+
+	// Read the invoker component and load invocation arguments.
+	_, err := chasm.ReadComponent(
+		ctx,
+		invokerRef,
+		func(c *Callback, ctx chasm.Context, _ any) (struct{}, error) {
```
bergundy: Just return what you need here instead of cloning the entire proto IMHO. I typically would rather not cloning components for use outside of the chasm contexts.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (e *InvocationTaskExecutor) Execute(
+	ctx context.Context,
+	invokerRef chasm.ComponentRef,
+	taskAttributes chasm.TaskAttributes,
+	task *callbackspb.InvocationTask,
+) error {
+	var ns *namespace.Namespace
+	// var invoker *Invoker
```
bergundy: Leftover comment.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+	var callback *Callback
```
bergundy: This can be the return type from `ReadComponent`.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+	return nil
+}
+
+func (e *InvocationTaskExecutor) Validate(
+	ctx chasm.Context,
+	callback *Callback,
+	_ chasm.TaskAttributes,
+	_ *callbackspb.InvocationTask,
+) (bool, error) {
+	return callback.Status == callbackspb.CALLBACK_STATUS_SCHEDULED, nil
```
bergundy: We should also validate the attempt matches, copy the attempt to the invocation task when scheduled. This would prevent duplicate tasks from being considered valid in some cases.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+// generateInvocationTask creates an InvocationTask and TaskAttributes based on the callback variant.
+// This is the CHASM port of HSM's RegenerateTasks logic from statemachine.go:79-100.
+// The destination (scheme + host) is extracted from the URL and used for task routing,
+// matching HSM's pattern where InvocationTask stores destination as u.Scheme + "://" + u.Host.
+func (e *BackoffTaskExecutor) generateInvocationTask(callback *Callback) (*callbackspb.InvocationTask, chasm.TaskAttributes, error) {
+	switch variant := callback.Callback.GetVariant().(type) {
+	case *callbackspb.Callback_Nexus:
+		// Parse URL to extract scheme and host, matching HSM's behavior
+		// from statemachine.go:86-90
+		u, err := url.Parse(variant.Nexus.Url)
```
bergundy: @chaptersix we should follow up and change this to use the destination in the token in case the URL is `temporal://system`.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+) (bool, error) {
+	// Validate that the callback is in BACKING_OFF state
+	return callback.Status == callbackspb.CALLBACK_STATUS_BACKING_OFF, nil
+}
+
+// generateInvocationTask creates an InvocationTask and TaskAttributes based on the callback variant.
+// This is the CHASM port of HSM's RegenerateTasks logic from statemachine.go:79-100.
+// The destination (scheme + host) is extracted from the URL and used for task routing,
+// matching HSM's pattern where InvocationTask stores destination as u.Scheme + "://" + u.Host.
+func (e *BackoffTaskExecutor) generateInvocationTask(callback *Callback) (*callbackspb.InvocationTask, chasm.TaskAttributes, error) {
```
bergundy: I would change this to a function that just generates the attributes. The task is easy to generate.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+		u, err := url.Parse(variant.Nexus.Url)
+		if err != nil {
+			return nil, chasm.TaskAttributes{}, fmt.Errorf("failed to parse URL: %w", err)
+		}
+
+		// Extract destination (scheme + host) for task routing
+		destination := u.Scheme + "://" + u.Host
+
+		return &callbackspb.InvocationTask{
+				Url: variant.Nexus.Url,
```
bergundy: I don't think you need the URL since you have the destination?

chasm/lib/callback/proto/v1/message.proto
```diff
@@ -54,19 +59,23 @@ enum CallbackStatus {
...
+message Nexus {
+    // Callback URL.
+    // (-- api-linter: core::0140::uri=disabled
+    //     aip.dev/not-precedent: Not respecting aip here. --)
+    string url = 1;
+    // Header to attach to callback request.
+    map<string, string> header = 2;
+}
+
+message InvokerState {
```
bergundy: What is this?

chasm/lib/callback/component.go
```diff
@@ -1,22 +1,49 @@
...
 	// Persisted internal state
 	*callbackspb.CallbackState
 
-	status *callbackspb.CallbackStatus
+	// Interface to retrieve Nexus operation completion data
+	CanGetNexusCompletion chasm.Field[CanGetNexusCompletion]
+
+	// Fields from HSM's nexusInvocation struct (nexus_invocation.go:35-40)
+	// These hold the invocation context needed for the HTTP request
+	completion nexusrpc.OperationCompletion
```
bergundy: Don't hold this here, create a separate struct for that as we did in the HSM version.

chasm/lib/callback/component.go
```diff
@@ -39,10 +67,172 @@ func (c *Callback) LifecycleState(_ chasm.Context) chasm.LifecycleState {
...
+// This is the CHASM port of nexusInvocation.Invoke() from nexus_invocation.go:63-131.
+// It performs the same HTTP request logic but returns CHASM's CallbackStatus enum
+// instead of HSM's invocationResult types (invocationResultOK, invocationResultRetry,
+// invocationResultFail).
+//
+// The return value mapping:
+// - HSM invocationResultOK -> CHASM CALLBACK_STATUS_SUCCEEDED
+// - HSM invocationResultRetry -> CHASM CALLBACK_STATUS_BACKING_OFF
+// - HSM invocationResultFail -> CHASM CALLBACK_STATUS_FAILED
+func (c *Callback) invoke(
```
bergundy: This shouldn't be a method on the component. I know we did this pattern in scheduler but I do not want us to reference components in general outside of the chasm context, it's too error prone and hard to reason about.

chasm/lib/callback/executor.go
```diff
@@ -0,0 +1,219 @@
...
+		return fmt.Errorf("failed to read component: %w", err)
+	}
+
+	callCtx, cancel := context.WithTimeout(
+		context.Background(),
+		e.Config.RequestTimeout(callback.NamespaceId, taskAttributes.Destination),
+	)
+	defer cancel()
+
+	result := callback.invoke(callCtx, ns, e, taskAttributes, task)
```
bergundy: You're missing the chasm variant that @lina-temporal recently added.

chasm/lib/callback/proto/v1/message.proto
```diff
@@ -33,6 +33,11 @@ message CallbackState {
 
     // Request ID that added the callback.
     string request_id = 9;
+
+    string workflow_id = 10;
+    string run_id = 11;
+
+    string namespace_id = 12;
```
bergundy: What is this for?

spkane31: Namespace is used to get request timeout from dynamic config on a per-namespace basis

chasm/lib/callback/proto/v1/message.proto
spkane31: Removed

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,279 @@
...
+
+type callbackInvokable interface {
+	// Invoke executes the callback logic and returns the invocation result.
+	Invoke(ctx context.Context, ns *namespace.Namespace, e InvocationTaskExecutor, task *callbackspb.InvocationTask) invocationResult
+	// WrapError provides each variant the opportunity to wrap the error returned by the task executor for, e.g. to
+	// trigger the circuit breaker.
+	WrapError(result invocationResult, err error) error
+}
+
+func (e InvocationTaskExecutor) executeInvocationTask(
```
bergundy: You're going to want to call this `Execute` but you'll figure that out when you register it in the library.

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,279 @@
...
+		e.Config.RequestTimeout(ns.Name().String(), task.Destination),
+	)
+	defer cancel()
+
+	result := invokable.Invoke(callCtx, ns, e, task)
+	saveErr := e.saveResult(ctx, ref, result)
+	return invokable.WrapError(result, saveErr)
+}
+
+func (e InvocationTaskExecutor) loadInvocationArgs(
```
bergundy: nit: I would write this as a function that you pass to `chasm.ReadComponent` to save the closure.

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,279 @@
...
+				return nil, queues.NewUnprocessableTaskError(
+					fmt.Sprintf("unprocessable callback variant: %v", variant),
+				)
+			}
+		},
+		nil,
+	)
+}
+
+func (e InvocationTaskExecutor) saveResult(
```
bergundy: Same here, write this as a function that you pass to `UpdateComponent`.

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (e *BackoffTaskExecutor) Validate(
+	ctx chasm.Context,
+	callback *Callback,
+	_ chasm.TaskAttributes,
+	_ *callbackspb.BackoffTask,
+) (bool, error) {
+	// Validate that the callback is in BACKING_OFF state
+	return callback.Status == callbackspb.CALLBACK_STATUS_BACKING_OFF, nil
```
bergundy: Let's also check that the attempt matches.

chasm/lib/callback/component.go
```diff
@@ -12,11 +12,17 @@ const (
...
+//
+// This is the CHASM port of HSM's nexusInvocation struct from nexus_invocation.go:25-32.
 type Callback struct {
 	chasm.UnimplementedComponent
 
 	// Persisted internal state
 	*callbackspb.CallbackState
+
+	// Interface to retrieve Nexus operation completion data
+	CanGetNexusCompletion chasm.Field[CanGetNexusCompletion]
```
bergundy: Call this `CompletionSource` or something more informative?

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,519 @@
...
+					Attempt:     1,
+					WorkflowId:  "workflow-id",
+					RunId:       "run-id",
+					NamespaceId: "namespace-id",
+				},
+			}
+			tc.setupCallback(callback)
+
+			// Set up the CanGetNexusCompletion field to return our mock
+			tree.SetRootComponent(callback)
```
bergundy: Why do you need to construct a tree? I don't see the tree being used here.

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,519 @@
...
+
+			// Create ComponentRef
+			ref := chasm.NewComponentRef[*Callback](chasm.EntityKey{
+				NamespaceID: "namespace-id",
+				BusinessID:  "workflow-id",
+				EntityID:    "run-id",
+			})
+
+			// Create context with engine
+			ctx := chasm.NewEngineContext(context.Background(), mockEngine)
```
bergundy, line 257: Note to self. We need better testing utilities in CHASM for this.

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,519 @@
...
+
+			// Verify no error and correct status was set
+			require.NoError(t, err)
+			require.Equal(t, tc.expectedStatus, callback.Status)
+		})
+	}
+}
+
+// Test the invocation result types
+func TestInvocationResultTypes(t *testing.T) {
```
bergundy: Redundant.

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,519 @@
...
+							Nexus: &callbackspb.Callback_Nexus{
+								Url: tc.nexusURL,
+							},
+						},
+					},
+					Attempt: 1,
+				},
+			}
+			tc.setupCallback(callback)
+			tree.SetRootComponent(callback)
```
bergundy: Same here, where is the tree used?

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,519 @@
...
+
+			// Setup completion getter
+			completion, err := nexusrpc.NewOperationCompletionSuccessful(nil, nexusrpc.OperationCompletionSuccessfulOptions{})
+			require.NoError(t, err)
+			completionGetter := &mockNexusCompletionGetter{
+				completion: completion,
+			}
+			// Set the CanGetNexusCompletion field using reflection
+			setFieldValue(&callback.CanGetNexusCompletion, CanGetNexusCompletion(completionGetter))
```
bergundy: This helper needs to be in `chasm/chasmtest`, not blocking the PR.

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,519 @@
...
+		assertOutcome         func(*testing.T, *Callback)
+	}{
+		{
+			name: "success",
+			caller: func(r *http.Request) (*http.Response, error) {
+				return &http.Response{StatusCode: 200, Body: http.NoBody}, nil
+			},
+			expectedMetricOutcome: "status:200",
+			setupCallback: func(cb *Callback) {
+				cb.Status = callbackspb.CALLBACK_STATUS_SCHEDULED
```
bergundy: Might as well set the status up for all test cases.

chasm/lib/callback/executors_test.go
bergundy, line 1: You are missing a bunch of test cases that exist in the HSM implementation.

chasm/lib/callback/proto/v1/message.proto
bergundy: Callback is not guaranteed to be attached to a workflow.
Wait for https://github.com/temporalio/temporal/pull/8533 to be merged and use `ExecutionKey()` instead.

chasm/lib/callback/proto/v1/tasks.proto
```diff
@@ -4,10 +4,17 @@ package temporal.server.chasm.lib.callbacks.proto.v1;
...
+
+    // The attempt number for this invocation.
+    int32 attempt = 2;
 }
 
-message BackoffTask {}
\ No newline at end of file
+message BackoffTask {
+    // deadline is the time at which the backoff period ends.
+    google.protobuf.Timestamp deadline = 1;
```
bergundy: Don't need this, it's part of the task attributes.

chasm/lib/callback/proto/v1/tasks.proto
```diff
@@ -4,10 +4,17 @@ package temporal.server.chasm.lib.callbacks.proto.v1;
...
 option go_package = "go.temporal.io/server/chasm/lib/callbacks/gen/callbackspb;callbackspb";
 
+import "google/protobuf/timestamp.proto";
+
 message InvocationTask {
-    // The base URL for nexus callbacks.
-    // Will have other meanings as more callback use cases are added.
-    string url = 1;
+    // The destination for callbacks. Can be a URL for nexus callbacks or temporal:// for internal callbacks.
+    string destination = 1;
```
bergundy: Don't need this, it's part of the task attributes.

chasm/lib/callback/proto/v1/tasks.proto
bergundy: And please add attempt for verification.

chasm/lib/callback/executors_test.go
spkane31: Is `chasm/chasmtest` the yet-to-be-developed library for writing tests with chasm?

chasm/lib/callback/executors_test.go
spkane31: I think this is leftover from a previous test implementation that did use the tree for operations, I changed this to use mocks.

chasm/lib/callback/proto/v1/tasks.proto
spkane31, line 1: Changes to this are triggering our proto breaking change CI check, do we have a way to force skip that check?

chasm/lib/callback/component.go
```diff
@@ -12,11 +12,17 @@ const (
 	Archetype chasm.Archetype = "Callback"
 )
 
+// Callback represents a callback component in CHASM.
+//
+// This is the CHASM port of HSM's nexusInvocation struct from nexus_invocation.go:25-32.
```
bergundy:
```suggestion
```

bergundy: The comment was incorrect.

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,277 @@
...
+	component *Callback,
+	chasmCtx chasm.Context,
+	ctx context.Context,
+) (callbackInvokable, error) {
+	target, err := component.CompletionSource.Get(chasmCtx)
+	if err != nil {
+		return nil, err
+	}
+
+	completion, err := target.GetNexusCompletion(ctx, component.RequestId)
```
bergundy: Ah I see that the context object is used here, we should be able to use the chasm context here, that's going to be tech debt we need to track. CC @yycptt.

chasm/lib/callback/executors.go
bergundy: Well... now that you've refactored this function to work within a lock, you can just make it a method of the Callback struct. Same thing for `saveResult`. You don't actually need `e` here.

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,277 @@
...
+		)
+	}
+}
+
+func (e InvocationTaskExecutor) saveResult(
+	component *Callback,
+	ctx chasm.MutableContext,
+	result invocationResult,
+) (struct{}, error) {
+	switch result.(type) {
```
bergundy: Go through the statemachine transitions here.

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,277 @@
...
+		InvocationTaskExecutorOptions: InvocationTaskExecutorOptions{
+			Config:         e.Config,
+			MetricsHandler: e.MetricsHandler,
+			Logger:         e.Logger,
+		},
+	}
+
+	// Convert the CHASM task to the internal BackoffTask type
+	// Note: BackoffTask proto is empty, deadline comes from NextAttemptScheduleTime in callback
+	backoffTask := &callbackspb.BackoffTask{
```
bergundy: What is this for?

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,277 @@
...
+	task *callbackspb.BackoffTask,
+) error {
+	// Create a taskExecutor wrapper with the same options
+	executor := InvocationTaskExecutor{
+		InvocationTaskExecutorOptions: InvocationTaskExecutorOptions{
+			Config:         e.Config,
+			MetricsHandler: e.MetricsHandler,
+			Logger:         e.Logger,
+		},
+	}
```
bergundy: Why create another executor?

chasm/lib/callback/executors.go
```diff
@@ -0,0 +1,277 @@
...
+	ctx chasm.Context,
+	callback *Callback,
+	taskAttr chasm.TaskAttributes,
+	task *callbackspb.BackoffTask,
+) (bool, error) {
+	// Validate that the callback is in BACKING_OFF state
+	return callback.Status == callbackspb.CALLBACK_STATUS_BACKING_OFF && callback.Attempt == task.Attempt, nil
+}
+
+func (e InvocationTaskExecutor) executeBackoffTask(
```
bergundy: This can just be `BackoffTaskExecutor.Execute`.

chasm/lib/callback/proto/v1/tasks.proto
bergundy: Yes, you can disable in `buf.yaml`.

chasm/lib/callback/executors.go
spkane31: This makes for an awkward (imo): 
```go
	invokable, err := chasm.ReadComponent(
		ctx,
		ref,
		(*Callback).loadInvocationArgs,
		ctx,
	)
```
any issues with that?

chasm/lib/buf.yaml
```diff
@@ -7,6 +7,7 @@ breaking:
     - WIRE
   ignore:
     - chasm/lib/scheduler/proto/v1/message.proto
+    - chasm/lib/callback/proto/v1/tasks.proto
```
bergundy, line 11: We should have a TODO to remove these after merging this PR. I'd be fine to leave the scheduler one though and let @lina-temporal handle that.

chasm/lib/callback/component.go
```diff
@@ -51,3 +59,72 @@ func (c *Callback) recordAttempt(ts time.Time) {
...
+		return nil, queues.NewUnprocessableTaskError(
+			fmt.Sprintf("unprocessable callback variant: %v", variant),
+		)
+	}
+}
+
+func (c *Callback) saveResult(
+	ctx chasm.MutableContext,
+	result invocationResult,
+) (struct{}, error) {
```
bergundy: Not blocking this PR but we need to add a `chasm.NoValue` type that is `*struct{}` and just return `nil` from methods that don't have a return value.

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,542 @@
...
+
+			// Verify the invokable was created successfully
+			require.NoError(t, err)
+			require.NotNil(t, invokable)
+		})
+	}
+}
+
+// Test saveResult transitions
+func TestSaveResult(t *testing.T) {
```
bergundy: This is redundant IMHO, just test the entire executor.

chasm/lib/callback/executors_test.go
```diff
@@ -0,0 +1,542 @@
...
+			} else {
+				// Verify the outcome if no error
+				tc.assertOutcome(t, callback)
+			}
+		})
+	}
+}
+
+// Test loadInvocationArgs with ComponentRef
+func TestLoadInvocationArgs(t *testing.T) {
```
bergundy: Not blocking: This is redundant IMHO.

If you really want to test load and save (which I think isn't required) move the tests to `component_test.go`.

chasm/lib/callback/executors_test.go
bergundy: Still not seeing all of the tests ported here.

chasm/lib/callback/metrics.go
```diff
@@ -0,0 +1,12 @@
+package callback
+
+import "go.temporal.io/server/common/metrics"
+
+var RequestCounter = metrics.NewCounterDef(
+	"chasm_callback_outbound_requests",
+	metrics.WithDescription("The number of CHASM outbound callback requests made by the history service."),
+)
+var RequestLatencyHistogram = metrics.NewTimerDef(
+	"chasm_callback_outbound_latency",
```
bergundy:
```suggestion
	"callback_outbound_requests",
	metrics.WithDescription("The number of outbound callback requests made by the history service."),
)
var RequestLatencyHistogram = metrics.NewTimerDef(
	"callback_outbound_latency",
```

chasm/lib/callback/metrics.go
```diff
@@ -0,0 +1,12 @@
...
+	metrics.WithDescription("Latency histogram of CHASM outbound callback requests made by the history service."),
```
bergundy, line 11:
```suggestion
	metrics.WithDescription("Latency histogram of outbound callback requests made by the history service."),
```

chasm/lib/callback/metrics.go
spkane31: This would conflict with the existing HSM metric, do we want to emit these metrics with the same name or differentiate the two?

chasm/lib/callback/metrics.go
spkane31: same as above, did this to clarify the difference between the HSM and CHASM version.

chasm/lib/callback/component.go
spkane31: #8566

chasm/lib/callback/executors_test.go
spkane31: | HSM | CHASM |
| - | - |
|  config_test.go | config_test.go (copied directly) |
| TestValidTransitions | TestValidTransitions (copied) |
| TestCompareState | **I don't think this test applies, is there a `chasm.CompareState` equivalence?** |
| TestProcessInvocationTaskNexus_Outcomes | TestExecuteInvocationTaskNexus_Outcomes (renamed) |
| TestProcessInvocationTaskHsm_Outcomes | **Not relevant right?** |
| TestProcessBackoffTask | TestProcessBackoffTask |
| TestProcessInvocationTaskChasm_Outcomes | TestExecuteInvocationTaskChasm_Outcomes (added now) |

chasm/lib/callback/executors_test.go
spkane31: Removed

chasm/lib/callback/executors_test.go
spkane31: removed

chasm/lib/callback/component.go
bergundy: I added this type FTR.

chasm/lib/callback/metrics.go
bergundy: Same, it's a transparent change.

chasm/lib/callback/statemachine.go
```diff
@@ -74,7 +74,7 @@ type EventFailed struct {
 var TransitionFailed = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_SCHEDULED},
 	callbackspb.CALLBACK_STATUS_FAILED,
-	func(cb *Callback, ctx chasm.MutableContext, event EventFailed) error {
+	func(cb *Callback, mctx chasm.MutableContext, event EventFailed) error {
```
bergundy: nit: rename back to `ctx`? also below please.

PR #8508, opened 2025-10-20
Consolidated and refactored activity request validations.

fretz12:
## What changed?
Consolidated and refactored activity request validations to the chasm package.
Added additional tests for the validator.

## Why?
Existing workflow activities and standalone activities should share the same validation code. Standalone activities also directly process the frontend request and therefore has additional fields to validate compared to embedded activities.

## How did you test it?
- [X] built
- [X] added new unit test(s)

## Potential risks
This refactors the existing stack, but the validation is basically exactly copied over and all relevant tests are passing. The standalone start activity request is now cloned before sanitizing the request attributes to preserve idempotent behavior during retries, but can potentially impact performance if there are large inputs.

### DIFF
--- a/service/history/api/command_attr_validator.go
+++ b/service/history/api/command_attr_validator.go
@@ -99,26 +99,34 @@ func (v *CommandAttrValidator) ValidateActivityScheduleAttributes(
 		attributes.RetryPolicy = &commonpb.RetryPolicy{}
 	}
 
+	opts := &activitypb.ActivityOptions{
+		TaskQueue:              attributes.TaskQueue,
+		ScheduleToCloseTimeout: attributes.GetScheduleToCloseTimeout(),
+		ScheduleToStartTimeout: attributes.GetScheduleToStartTimeout(),
+		StartToCloseTimeout:    attributes.GetStartToCloseTimeout(),
+		HeartbeatTimeout:       attributes.GetHeartbeatTimeout(),
+		RetryPolicy:            attributes.RetryPolicy,
+	}
+
 	err := activity.ValidateAndNormalizeActivityAttributes(
 		activityID,
 		activityType,
 		v.getDefaultActivityRetrySettings,
 		v.maxIDLengthLimit,
 		namespaceID,
-		&activitypb.ActivityOptions{
-			TaskQueue:              attributes.TaskQueue,
-			ScheduleToCloseTimeout: attributes.GetScheduleToCloseTimeout(),
-			ScheduleToStartTimeout: attributes.GetScheduleToStartTimeout(),
-			StartToCloseTimeout:    attributes.GetStartToCloseTimeout(),
-			HeartbeatTimeout:       attributes.GetHeartbeatTimeout(),
-			RetryPolicy:            attributes.RetryPolicy,
-		},
+		opts,
 		attributes.GetPriority(),
 		runTimeout)
 
 	if err != nil {
 		return failedCause, err
 	}
+
+	attributes.ScheduleToCloseTimeout = opts.ScheduleToCloseTimeout
+	attributes.ScheduleToStartTimeout = opts.ScheduleToStartTimeout
+	attributes.StartToCloseTimeout = opts.StartToCloseTimeout
+	attributes.HeartbeatTimeout = opts.HeartbeatTimeout
+	attributes.RetryPolicy = opts.RetryPolicy
 
 	return enumspb.WORKFLOW_TASK_FAILED_CAUSE_UNSPECIFIED, nil
 }


chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// StandaloneActivityModifiedAttributes Modified attributes after validating standalone activity specific attributes.
+// It is up to the caller to decide how to handle these modified attributes.
+type StandaloneActivityModifiedAttributes struct {
+	requestID                 string
+	searchAttributesUnaliased *commonpb.SearchAttributes // Unaliased search attributes after validation
+}
+
+func NewRequestAttributesValidator(
```
dandavison: I believe all public functions should have a docstring.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+
+	v.options.HeartbeatTimeout = timestamp.MinDurationPtr(v.options.GetHeartbeatTimeout(), v.options.GetStartToCloseTimeout())
+
+	return nil
+}
+
+// ValidateStandaloneActivity validates the standalone activity specific attributes. It will return the modified
+// attributes that are not idempotent, letting the caller decide how to handle them.
```
dandavison: "idempotent" usually refers to a function. What does it mean for an attribute to be idempotent?

chasm/lib/activity/frontend.go
```diff
@@ -38,6 +69,53 @@ func (h *frontendHandler) StartActivityExecution(ctx context.Context, req *workf
...
+		req.Options.RetryPolicy = &commonpb.RetryPolicy{}
+	}
+
+	validator := NewRequestAttributesValidator(
+		req.ActivityId,
+		activityType,
+		dynamicconfig.DefaultActivityRetryPolicy.Get(h.dc),
+		dynamicconfig.MaxIDLengthLimit.Get(h.dc)(),
+		namespaceID,
+		req.Options,
```
dandavison, line 107: `ValidateAndAdjustTimeouts` is going to mutate the timeouts on the passed-in `req.Options`. Is this really what we want to do, given that the request struct is reused during retries?

chasm/lib/activity/frontend.go
```diff
@@ -38,6 +69,53 @@ func (h *frontendHandler) StartActivityExecution(ctx context.Context, req *workf
...
+		return nil, err
+	}
+
+	if modifiedAttributes.requestID != "" {
+		req.RequestId = modifiedAttributes.requestID
+	}
+
+	if modifiedAttributes.searchAttributesUnaliased != nil && modifiedAttributes.searchAttributesUnaliased != req.SearchAttributes {
+		// Since searchAttributesUnaliased is not idempotent, we need to clone the request so that in case of retries,
+		// the field is set to the original value.
```
dandavison: same q about meaning of idempotent.

and nit... :)

```suggestion
		// Since searchAttributesUnaliased is not idempotent, we need to clone the request so that in the case of retries,
		// the field is set to the original value.
```

chasm/lib/activity/frontend.go
```diff
@@ -38,6 +69,53 @@ func (h *frontendHandler) StartActivityExecution(ctx context.Context, req *workf
...
+		&StandaloneActivityAttributes{
+			namespaceName:    req.GetNamespace(),
+			requestID:        req.RequestId,
+			searchAttributes: req.SearchAttributes,
+			saMapperProvider: h.saMapperProvider,
+			saValidator:      h.saValidator,
+		},
+	)
+
+	err = validator.ValidateAndAdjustTimeouts(durationpb.New(0))
```
dandavison: This function must always be called before making use of the validator, but a user of this API might forget to call it. I made a suggestion in my review comment.

fretz12:
> Nice improvement -- the existing validation code was messy and complex and it's good to encapsulate it.
> 
> I have some suggestions for improving this further. Here's the API this PR introduces:
> 
> **Workflow Activity**
> 
> ```go
> validator := activity.NewRequestAttributesValidator(...)
> err := validator.ValidateAndAdjustTimeouts(runTimeout)
> activityOptions := validator.GetActivityOptions()
> // use activityOptions to mutate ScheduleActivityTaskCommandAttributes
> ```
> 
> **Standalone Activity**
> 
> ```go
> validator := NewRequestAttributesValidator(...)
> err = validator.ValidateAndAdjustTimeouts(durationpb.New(0))
> modifiedAttributes, err := validator.ValidateStandaloneActivity()
> // use modifiedAttributes to mutate StartActivityExecutionRequest
> ```
> 
> Two issues with this that I'd like to improve are
> 
> 1. Someone could forget to call `ValidateAndAdjustTimeouts` before calling `GetActivityOptions` or `ValidateStandaloneActivity`
> 2. We're mutating the timeouts on the passed-in `req.Options` in the Standalone case; I am not sure this mutation is correct/desirable given that the request is re-used by retries.
> 
> My first idea is how about we change the API to something like
> 
> ```go
> type ValidatedActivityRequest struct {
>     Options                   *activitypb.ActivityOptions
>     RequestID                 string
>     SearchAttributesUnaliased *commonpb.SearchAttributes
> }
> 
> func ValidateActivityRequest(
>     activityID string,
>     activityType string,
>     // ...
>     runTimeout *durationpb.Duration,
>     standaloneAttrs *StandaloneActivityAttributes,
> ) (*ValidatedActivityRequest, error)
> ```
> 
> Then `ValidateActivityRequest` will take care of cloning input structs when necessary and adjusting the timeouts, and both code paths can use the returned struct of validated data as needed (e.g. update their own request / task attribute structs).

@dandavison - good questions, and the proto mutation is a thorn that needs improvement. We mutably sanitize the incoming request as that's what the current codebase does.

On your concerns:

> 1. Someone could forget to call `ValidateAndAdjustTimeouts` before calling `GetActivityOptions` or `ValidateStandaloneActivity`

I'm ok with that, separating out the mutated fields to a new return struct and letting the caller decide what to do. I'd rename `ValidatedActivityRequest` perhaps `NormalizedActivityRequest` as to convey the intent of mutation.

> 2. We're mutating the timeouts on the passed-in `req.Options` in the Standalone case; I am not sure this mutation is correct/desirable given that the request is re-used by retries.
> 
Yea def a concern with the retries. If you look at StartActivityExecution in frontend.go, the request is essentially forwarded directly over to the history handler.go rpc. If we truly want to not affect retries, we should reallly clone the request with mutated fields before forwarding. The input can potentially be big. Based on your experience, do you think cloning the request will cause performance issues?


UPDATE: per discussion, we'll clone the whole request if anything needs to be mutated. The safe thing to do for predictable retries.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,255 @@
...
+
+// ValidateAndAdjustTimeouts validates the activity request attributes and adjusts the ActivityOptions timeout based on
+// the following rules. This validation is shared by both standalone and embedded activities.
+// runTimeout is the workflow run timeout. Set to durationpb.New(0) if not applicable.
+// 1. If ScheduleToClose is set, fill in missing ScheduleToStart and StartToClose from ScheduleToClose
+// 2. If StartToClose is set but ScheduleToClose is not set, set ScheduleToClose to runTimeout, and fill in missing ScheduleToStart from runTimeout
+// 3. If neither ScheduleToClose nor StartToClose is set, return error
+// 4. Ensure all timeouts do not exceed runTimeout if runTimeout is set (>0)
+// 5. Ensure HeartbeatTimeout does not exceed StartToClose
+func (v *RequestAttributesValidator) ValidateAndAdjustTimeouts(runTimeout *durationpb.Duration) error {
```
fretz12: I need to add input payload size validator

chasm/lib/activity/frontend.go
```diff
@@ -23,12 +30,36 @@ type frontendHandler struct {
...
+	dc *dynamicconfig.Collection,
+	saMapperProvider searchattribute.MapperProvider,
+	saProvider searchattribute.Provider,
+	visibilityMgr manager.VisibilityManager) FrontendHandler {
 	return &frontendHandler{
 		client:            client,
 		namespaceRegistry: namespaceRegistry,
+		dc:                dc,
+		saMapperProvider:  saMapperProvider,
+		saValidator: searchattribute.NewValidator(
```
bergundy: This should be injected as a separate provider, add it to `common/resource/fx.go`. This will prevent the need to reconstruct this everywhere it is needed.

chasm/lib/activity/frontend.go
```diff
@@ -38,6 +69,53 @@ func (h *frontendHandler) StartActivityExecution(ctx context.Context, req *workf
 		return nil, err
 	}
 
+	activityType := ""
+	if req.ActivityType != nil {
```
bergundy: You don't need to add nil checks when validating protos, use `req.GetActivityType().GetName()`. The getters give you the default value if the field is nil/unset.

chasm/lib/activity/frontend.go
bergundy: Based on recent slack discussions, we were saying that we don't want to use this `ActivityOptions` struct because we may want to make more fields in the start request be considered as options. Let's flatten this out after we change the API. (Same for describe BTW).

chasm/lib/activity/validator.go
fretz12: Outdated, but added docstring to all public funcs

chasm/lib/activity/frontend.go
fretz12: Outdated, but I refactored to explicitly return modified attributes, as we discussed. 
Also, the SA frontend request is now cloned before mutation to ensure idempotency during retries.

chasm/lib/activity/frontend.go
fretz12: Refactored to explicitly return modified attributes, as we discussed.

chasm/lib/activity/frontend.go
fretz12: Outdated, but refactored to single func call.

chasm/lib/activity/validator.go
fretz12: Done

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +31,142 @@ type FrontendHandler interface {
...
+
+	// Update the command attributes with the adjusted timeouts
+	if modifiedAttributes != nil {
+		req.Options.ScheduleToCloseTimeout = modifiedAttributes.ScheduleToCloseTimeout
+		req.Options.ScheduleToStartTimeout = modifiedAttributes.ScheduleToStartTimeout
+		req.Options.StartToCloseTimeout = modifiedAttributes.StartToCloseTimeout
+		req.Options.HeartbeatTimeout = modifiedAttributes.HeartbeatTimeout
+	}
+
+	modifiedSaAttributes, err := ValidateStandaloneActivity(
```
dandavison: You're using "SA" here to mean Standalone Activity, but earlier in the PR and elsewhere in the codebase it means "Search Attribute".

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,333 @@
...
+func populateTimeouts(
+	activityID string,
+	activityType string,
+	runTimeout *durationpb.Duration,
+	options *activitypb.ActivityOptions,
+	modifiedAttributes *ModifiedActivityRequestAttributes,
+) error {
+	ScheduleToCloseSet := options.GetScheduleToCloseTimeout().AsDuration() > 0
+	ScheduleToStartSet := options.GetScheduleToStartTimeout().AsDuration() > 0
+	StartToCloseSet := options.GetStartToCloseTimeout().AsDuration() > 0
```
dandavison: These variable names are capitalized but shouldn't be.

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +31,142 @@ type FrontendHandler interface {
...
 	resp, err := h.client.StartActivityExecution(ctx, &activitypb.StartActivityExecutionRequest{
 		NamespaceId:     namespaceID.String(),
-		FrontendRequest: req,
+		FrontendRequest: modifiedReq,
 	})
 
 	return resp.GetFrontendResponse(), err
 }
+
+func (h *frontendHandler) validateAndPopulateListRequest(
```
dandavison: This function name should contain "Start" instead of "List" right?

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +31,142 @@ type FrontendHandler interface {
...
+	dc                *dynamicconfig.Collection
+	logger            log.Logger
+	metricsHandler    metrics.Handler
 	namespaceRegistry namespace.Registry
+	saMapperProvider  searchattribute.MapperProvider
+	saValidator       *searchattribute.Validator
 }
 
-func NewFrontendHandler(client activitypb.ActivityServiceClient, namespaceRegistry namespace.Registry) FrontendHandler {
+func NewFrontendHandler(
```
dandavison, line 41: Needs a docstring

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +31,142 @@ type FrontendHandler interface {
...
+				dynamicconfig.VisibilityAllowList.Get(dc),
+			),
+			dynamicconfig.SuppressErrorSetSystemSearchAttribute.Get(dc),
+		),
 	}
 }
 
+// StartActivityExecution initiates a standalone activity execution in the specified namespace.
+// It validates the request, resolves the namespace ID, applies default configurations,
+// and forwards the request to the activity service client.
```
dandavison:
```suggestion
// and forwards the request to the activity service handler.
```

chasm/lib/activity/validator_test.go
```diff
@@ -0,0 +1,413 @@
...
+const (
+	defaultActivityID       = "test-activity-id"
+	defaultActivityType     = "test-activity-type"
+	defaultTaskQueue        = "test-task-queue"
+	defaultMaxIDLengthLimit = 1000
+	defaultNamespaceID      = "default"
+)
+
+var (
+	defaultActvityOptions = activitypb.ActivityOptions{
```
dandavison: typo in this variable name

dandavison (commented):
Not a real review but a few trivial comments here!

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,333 @@
...
+// Returns the modified/normalized timeout attributes or an error if validation fails.
+func ValidateActivityRequestAttributes(
+	activityID string,
+	activityType string,
+	getDefaultActivityRetrySettings dynamicconfig.TypedPropertyFnWithNamespaceFilter[retrypolicy.DefaultRetrySettings],
+	maxIDLengthLimit int,
+	namespaceID namespace.ID,
+	options *activitypb.ActivityOptions,
+	priority *commonpb.Priority,
+	runTimeout *durationpb.Duration) (*ModifiedActivityRequestAttributes, error) {
```
dandavison: "Modified" doesn't feel like the right word to me in the return types `ModifiedActivityRequestAttributes` and `ModifiedStandaloneActivityRequestAttributes`,
partly since the returned values might not actually differ from the input, and partly because there's a mismatch between "validate" and "modified".

I think 
```
func ValidateActivityRequestAttributes(...) (*ValidatedActivityRequestAttributes, error)
```

could be good.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,333 @@
...
+	blobSizeLimitWarn dynamicconfig.IntPropertyFnWithNamespaceFilter,
+	inputSizeBytes int,
+	logger log.Logger,
+	maxIDLengthLimit int,
+	namespaceName string,
+	requestID string,
+	searchAttributes *commonpb.SearchAttributes,
+	saMapperProvider searchattribute.MapperProvider,
+	saValidator *searchattribute.Validator) (*ModifiedStandaloneActivityRequestAttributes, error) {
+	modifiedAttributes := &ModifiedStandaloneActivityRequestAttributes{}
```
dandavison: nit: formatting

```suggestion
	saValidator *searchattribute.Validator,
) (*ModifiedStandaloneActivityRequestAttributes, error) {
	modifiedAttributes := &ModifiedStandaloneActivityRequestAttributes{}
```

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +31,142 @@ type FrontendHandler interface {
...
-func NewFrontendHandler(client activitypb.ActivityServiceClient, namespaceRegistry namespace.Registry) FrontendHandler {
+func NewFrontendHandler(
+	client activitypb.ActivityServiceClient,
+	dc *dynamicconfig.Collection,
+	logger log.Logger,
+	metricsHandler metrics.Handler,
+	namespaceRegistry namespace.Registry,
+	saMapperProvider searchattribute.MapperProvider,
+	saProvider searchattribute.Provider,
+	visibilityMgr manager.VisibilityManager) FrontendHandler {
```
dandavison: nit (here and elsewhere): I believe the existing codebase formatting style is:

```suggestion
	saProvider searchattribute.Provider,
	visibilityMgr manager.VisibilityManager,
) FrontendHandler {
```

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +31,142 @@ type FrontendHandler interface {
...
 		NamespaceId:     namespaceID.String(),
-		FrontendRequest: req,
+		FrontendRequest: modifiedReq,
 	})
 
 	return resp.GetFrontendResponse(), err
 }
+
+func (h *frontendHandler) validateAndPopulateListRequest(
+	req *workflowservice.StartActivityExecutionRequest, namespaceID namespace.ID) (*workflowservice.StartActivityExecutionRequest, error) {
```
dandavison:
```suggestion
func (h *frontendHandler) validateStartRequest(
	req *workflowservice.StartActivityExecutionRequest,
	namespaceID namespace.ID,
) (*workflowservice.StartActivityExecutionRequest, error) {
```

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,333 @@
...
+	retrypolicy.EnsureDefaults(retryPolicy, defaultActivityRetrySettings)
+	return retrypolicy.Validate(retryPolicy)
+}
+
+func populateTimeouts(
+	activityID string,
+	activityType string,
+	runTimeout *durationpb.Duration,
+	options *activitypb.ActivityOptions,
+	modifiedAttributes *ModifiedActivityRequestAttributes,
```
dandavison: nit: More standard to make this a method on `ModifiedActivityRequestAttributes`?

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func validateInputSize(
+	activityID string,
+	activityType string,
+	blobSizeLimitError dynamicconfig.IntPropertyFnWithNamespaceFilter,
+	blobSizeLimitWarn dynamicconfig.IntPropertyFnWithNamespaceFilter,
+	inputSize int,
+	logger log.Logger,
+	namespaceName string) error {
```
dandavison:
```suggestion
	namespaceName string,
) error {
```

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,333 @@
...
+		if inputSize > sizeErrorLimit {
+			return common.ErrBlobSizeExceedsLimit
+		}
+	}
+
+	return nil
+}
+
+func validateAndUnaliasSearchAttributes(
+	modifiedAttributes *ModifiedStandaloneActivityRequestAttributes,
```
dandavison: Make it a method? I don't personally think the method name needs to encode what it does (i.e. doesn't need to contain "unalias")

chasm/lib/activity/validator.go
dandavison: The codebase has similar methods like [`tqid.NormalizeAndValidate`]( https://github.com/temporalio/temporal/blob/main/common/tqid/task_queue_validator.go#L54) so maybe

```go
func (v *ValidatedActivityRequestAttributes) normalizeAndValidateTimeouts(...) error
```

chasm/lib/activity/validator.go
dandavison: Seeing as we end up doing

```go
req.SearchAttributes = modifiedSaAttributes.searchAttributesUnaliased
```

I think `validateAndNormalize` works as a name here too right? I.e. this unaliasing can be seen as a form of normalization.

dandavison (approved):
LGTM. The main thing of course is that we don't change the behavior or performance on the workflow activity code path.

I made a bunch of style suggestions -- hopefully this is appropriate as a way for us to converge on a consistent style seeing as we're both ramping up on the codebase!

chasm/lib/activity/frontend.go
fretz12: Added

chasm/lib/activity/frontend.go
fretz12: I do like the readability, but we should automate this. [gofumpt](https://github.com/mvdan/gofumpt) seems to do the job, but do you know of a better way?
We could use this as a prehook commit.

chasm/lib/activity/frontend.go
fretz12: Probably should add this to lint rule. Need buy in from entire OSS team, we should bring up during standup.

chasm/lib/activity/frontend.go
fretz12: refactored

chasm/lib/activity/frontend.go
fretz12: Fixed

chasm/lib/activity/frontend.go
fretz12: Yes, autocomplete madness

chasm/lib/activity/frontend.go
fretz12: simplified

chasm/lib/activity/frontend.go
fretz12: renamed.

chasm/lib/activity/validator.go
fretz12: I see your point. However, `ValidatedActivityRequestAttributes` is ambiguous too as it seems to indicate the attributes that have been validated but not necessarily mutated. Since mutation is code critical, I'd like a more active verb. How about `NormalizedActivityRequestAttributes` or `SanitizedActivityRequestAttributes`?

chasm/lib/activity/frontend.go
fretz12: fixed.

chasm/lib/activity/validator.go
fretz12: I like `normalizeAndValidateTimeouts`

chasm/lib/activity/validator.go
fretz12: Fixed. From copy & paste.

chasm/lib/activity/validator.go
fretz12: done

chasm/lib/activity/validator.go
fretz12: done

chasm/lib/activity/validator.go
fretz12: Yup, `validateAndNormalize` FTW

chasm/lib/activity/validator_test.go
fretz12: fixed

chasm/lib/activity/frontend.go
```diff
@@ -22,26 +29,129 @@ type FrontendHandler interface {
...
+// before mutation to preserve the original for retries.
+// 3. Sends the request to the history activity service.
+//
+// Parameters:
+//   - ctx: Context for the request, used for cancellation and deadlines
+//   - req: StartActivityExecutionRequest containing activity details, options, and metadata
+//
+// Returns:
+//   - *workflowservice.StartActivityExecutionResponse: Response containing activity execution details
+//   - error: Any error encountered during namespace resolution, validation, or execution
```
bergundy: nit: I don't find this type of comment especially helpful. I'd rather have comments that explains what type of errors I can expect and other semantics. For example validation of the request is important to note here.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,337 @@
+package activity
+
+import (
+	"fmt"
+
+	"github.com/pborman/uuid"
```
bergundy: Don't use this package, use `google/uuid` please.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,337 @@
...
+	activityID string,
+	activityType string,
+	getDefaultActivityRetrySettings dynamicconfig.TypedPropertyFnWithNamespaceFilter[retrypolicy.DefaultRetrySettings],
+	maxIDLengthLimit int,
+	namespaceID namespace.ID,
+	options *activitypb.ActivityOptions,
+	priority *commonpb.Priority,
+	runTimeout *durationpb.Duration,
+) (*ModifiedActivityRequestAttributes, error) {
+	if err := tqid.NormalizeAndValidate(options.TaskQueue, "", maxIDLengthLimit); err != nil {
```
bergundy, line 44: This mutates the task queue, you might as well mutate the other attributes in this method and name it accordingly.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil, serviceerror.NewInvalidArgumentf("ActivityId exceeds length limit. ActivityId=%s ActivityType=%s Length=%d Limit=%d",
+			activityID, activityType, len(activityID), maxIDLengthLimit)
+	}
+	if len(activityType) > maxIDLengthLimit {
+		return nil, serviceerror.NewInvalidArgumentf("ActivityType exceeds length limit. ActivityId=%s ActivityType=%s Length=%d Limit=%d",
+			activityID, activityType, len(activityType), maxIDLengthLimit)
+	}
+
+	// Only attempt to deduce and fill in unspecified timeouts only when all timeouts are non-negative.
+	if err := timestamp.ValidateAndCapProtoDuration(options.GetScheduleToCloseTimeout()); err != nil {
```
bergundy: Same here, this also mutates.

chasm/lib/activity/validator.go
```diff
@@ -0,0 +1,337 @@
...
+		runTimeout,
+		options,
+		modifiedAttributes); err != nil {
+		return nil, err
+	}
+
+	return modifiedAttributes, nil
+}
+
+func validateActivityRetryPolicy(
```
bergundy, line 79: This also mutates and should be named accordingly.

common/log/tag/tags.go
```diff
@@ -914,6 +914,16 @@ func ActivityInfo(activityInfo interface{}) ZapTag {
 	return NewAnyTag("activity-info", activityInfo)
 }
 
+// ActivityID returns tag for a standalone activity ID
```
bergundy, line 917:
```suggestion
// ActivityID returns a tag for a standalone activity ID
```

common/log/tag/tags.go
```diff
@@ -914,6 +914,16 @@ func ActivityInfo(activityInfo interface{}) ZapTag {
...
+func ActivityID(id string) ZapTag {
+	return NewStringTag("activity-id", id)
+}
+
+// ActivitySize returns tag for a standalone activity size
```
bergundy:
```suggestion
// ActivitySize returns a tag for a standalone activity size
```

chasm/lib/activity/frontend.go
fretz12: Agreed, that seems to be accepted practice in go

chasm/lib/activity/validator.go
fretz12: Agreed, I copied that from the workflow start code and it felt wrong.

chasm/lib/activity/validator.go
fretz12: hmm good catch. Seems like there's way more mutation going on than expected. Since that's the case, I'm kinda agree to just allow input mutation, and explicitly rename + doc it warning others to clone the input if desired. @dandavison - what do you think?

chasm/lib/activity/validator.go
fretz12: good catch, pending decision above.

chasm/lib/activity/validator.go
fretz12: Pending decision above.

common/log/tag/tags.go
fretz12: fixed

common/log/tag/tags.go
fretz12: fixed

chasm/lib/activity/validator.go
dandavison: Sure, since the existing workflow activity validation mutates the input then let's follow that pattern, with names making it clear.

(Obviously we do also need to watch out for genuine bugs that would be caused by some of these mutations to data structures that are being resubmitted in retries, i.e. where the mutation is not idempotent)

fretz12: Changed back to using mutation, renamed func, and updated docstring to have a clear warning on mutation.

PR #8511, opened 2025-10-20
batch: register proto workflow for proto-qualified and non-proto-qualified workflow type.

spkane31:
## What changed?
Register the `BatchWorkflowProtobuf` as both `BatchWFTypeName` and `BatchWFTypeNameProtobuf`, picking up the legacy batch operation in addition to the new one.

## Why?
Now in the version after 1.30 the frontend can be changed to call the new workflow function.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None in this iteration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,6 +232,11 @@ $(STAMPDIR)/gomajor-$(GOMAJOR_VER): | $(STAMPDIR) $(LOCALBIN)
 	@touch $@
 $(GOMAJOR): $(STAMPDIR)/gomajor-$(GOMAJOR_VER)
 
+ERRORTYPE_VER := v0.0.7
+ERRORTYPE := $(LOCALBIN)/errortype
+$(ERRORTYPE): | $(LOCALBIN)
+	$(call go-install-tool,$(ERRORTYPE),fillmore-labs.com/errortype,$(ERRORTYPE_VER))
+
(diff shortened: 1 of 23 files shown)


service/worker/batcher/fx.go
```diff
@@ -67,7 +67,7 @@ func (s *workerComponent) DedicatedWorkerOptions(ns *namespace.Namespace) *worke
 }
 
 func (s *workerComponent) Register(registry sdkworker.Registry, ns *namespace.Namespace, _ workercommon.RegistrationDetails) func() {
-	registry.RegisterWorkflowWithOptions(BatchWorkflow, workflow.RegisterOptions{Name: BatchWFTypeName})
+	registry.RegisterWorkflowWithOptions(BatchWorkflowProtobuf, workflow.RegisterOptions{Name: BatchWFTypeName})
```
bergundy, line 72:
```suggestion
	// Register the batch workflow with both the proto-qualified and unqualified types.
	// TODO(spkane31): Remove the proto-qualified type and call the unqualified type from the frontend after the 1.30 release.
	registry.RegisterWorkflowWithOptions(BatchWorkflowProtobuf, workflow.RegisterOptions{Name: BatchWFTypeName})
```

spkane31: added manually

PR #8515, opened 2025-10-21
Validate activity task attempt count

yiminc:
## What changed?

## Why?

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks

### DIFF
--- a/proto/internal/temporal/server/api/persistence/v1/executions.proto
+++ b/proto/internal/temporal/server/api/persistence/v1/executions.proto
@@ -342,6 +342,7 @@ message TransferTaskInfo {
     // It monotonically increments when the activity options are changed.
     // It is used to check if activity related tasks are still relevant to  their corresponding state machine.
     int32 stamp = 17;
+    int32 attempt = 19;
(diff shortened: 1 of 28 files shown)


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -342,6 +342,7 @@ message TransferTaskInfo {
     // It monotonically increments when the activity options are changed.
     // It is used to check if activity related tasks are still relevant to  their corresponding state machine.
     int32 stamp = 17;
+    int32 attempt = 19;
```
dnr, line 345: Since we just added a stamp, and stamp serves the exact same purpose (invalidate an obsolete task), why not just increment the stamp on each attempt?

yiminc: We won't be able to tell if a stamp mismatch is because it is stale or because it is from backlog tasks generated before this change. We would need to know what is the original stamp.

PR #8521, opened 2025-10-22
Proto changes for async deployment info propagation

Shivs11:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/proto/internal/temporal/server/api/deployment/v1/message.proto
+++ b/proto/internal/temporal/server/api/deployment/v1/message.proto
@@ -22,6 +22,7 @@ message WorkerDeploymentVersion {
 
 // The source of truth for this data is in the WorkerDeployment entity workflows, which is
 // synced to all TQs whenever the source changes.
+// Deprecated.
 message DeploymentVersionData {
     // Nil means unversioned.
     WorkerDeploymentVersion version = 1;
@@ -44,6 +45,12 @@ message DeploymentVersionData {
     float ramp_percentage = 5;
 
     // Status of the Worker Deployment Version.
+    temporal.api.enums.v1.WorkerDeploymentVersionStatus status = 6;
+}
+
+// Information that a TQ should know about a particular Deployment Version. This info is not part of
+// RoutingConfig and hence not protected by the revision number.
+message WorkerDeploymentVersionData {
     temporal.api.enums.v1.WorkerDeploymentVersionStatus status = 6;
 }
 
--- a/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
@@ -340,6 +340,10 @@ message GetTaskQueueUserDataResponse {
 message SyncDeploymentUserDataRequest {
     string namespace_id = 1;
     string task_queue = 2;
+    // Required, unless deprecated fields are used.
+    // (-- api-linter: core::0203::required=disabled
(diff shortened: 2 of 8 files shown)


proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -101,10 +101,16 @@ message DeploymentData {
...
+    // Remove after `AsyncSetCurrentAndRamping` workflow version is irreversibly enabled.
+    repeated temporal.server.api.deployment.v1.DeploymentVersionData versions = 2 [deprecated = true];
 
     // Present if the task queue's ramping version is unversioned.
-    temporal.server.api.deployment.v1.DeploymentVersionData unversioned_ramp_data = 3;
+    // Remove after `AsyncSetCurrentAndRamping` workflow version is irreversibly enabled.
+    temporal.server.api.deployment.v1.DeploymentVersionData unversioned_ramp_data = 3 [deprecated = true];
+
+    // Routing and version membership data for all worker deployments that this task queue belongs to.
+    map<string, WorkerDeploymentData> deployments_data = 4;
```
Shivs11, line 113: can you clarify what exactly would the key in this map would represent?  My humble guess is that it would be the version but would love to be sure.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -355,15 +356,23 @@ message SyncDeploymentUserDataRequest {
...
+        temporal.server.api.deployment.v1.DeploymentVersionData update_version_data = 6 [deprecated = true];
         // The version whose data should be cleaned from the task queue.
-        temporal.server.api.deployment.v1.WorkerDeploymentVersion forget_version = 7;
+        temporal.server.api.deployment.v1.WorkerDeploymentVersion forget_version = 7 [deprecated = true];
     }
+
+    // Absent means no change.
+    // Ignored by the task queue if new revision number is not greater that what it has.
+    temporal.api.deployment.v1.RoutingConfig update_routing_config = 10;
+    map<string, temporal.server.api.deployment.v1.WorkerDeploymentVersionData> upsert_versions_data = 11;
```
Shivs11, line 374: can you clarify what string would represent here? any particular reason of moving the routingConfig related information to it's own field now?

proto/internal/temporal/server/api/deployment/v1/message.proto
```diff
@@ -47,6 +48,11 @@ message DeploymentVersionData {
     temporal.api.enums.v1.WorkerDeploymentVersionStatus status = 6;
 }
 
+// Information that a TQ should know about a particular Deployment Version
+message WorkerDeploymentVersionData {
+    temporal.api.enums.v1.WorkerDeploymentVersionStatus status = 6;
+}
```
Shivs11: I think we should include the version here

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
Shivs11: Looking at the nested comment present inside "WorkerDeploymentData", it seems to me that this string would then just be the revision number while the key present in "WorkerDeploymentData" would probably be the version.

To me, we should be adding the revision number per versionID instead of per WorkerDeployment, no?

Shivs11 (commented):
some clarifications required

PR #8522, opened 2025-10-22
Add callback token support for Nexus operations

chaptersix:


common/nexus/nexusrpc/completion.go
```diff
@@ -121,6 +127,10 @@ func (c *OperationCompletionSuccessful) applyToHTTPRequest(request *http.Request
 	}
 	request.Header.Set(headerOperationState, string(nexus.OperationStateSucceeded))
 
+	if c.CallbackToken != "" {
+		encodedToken := base64.StdEncoding.EncodeToString([]byte(c.CallbackToken))
+		request.Header.Set("Nexus-Callback-Token", encodedToken)
+	}
```
chaptersix, line 133:
```suggestion
	if c.CallbackToken != "" {
		encodedToken := base64.StdEncoding.EncodeToString([]byte(c.CallbackToken))
		request.Header.Set("Nexus-Callback-Token", encodedToken)
	}
```
this should likely be handled by the handler that receives the callback. should be `token` in completion requests

chaptersix:
will need to look back at the executor that calls Start Operation and the completion handler to make sure they are handling this correctly.

chaptersix:
Will re-open if needed

PR #8496, opened 2025-10-16
State machine for CHASM activities

dandavison:
## What changed?
Adds a state machine defining valid transitions for a CHASM activity.

The state machine does not include transitions related to cancel and pause states: we can add those when we implement those features.

image

## Why?
Part of foundations for CHASM activity implementation / Standalone Activity.

## How did you test it?
- Added new unit test

---

> [!NOTE]
> Introduces an Activity state machine with core execution transitions and updates the chasm Transition API to take (SM, Context, Event) parameter order.
> 
> - **Activity state machine** (`chasm/lib/activity/statemachine.go`):
>   - Implements `activity.Activity` as `chasm.StateMachine[activitypb.ActivityExecutionStatus]` with `State()`/`SetState()`.
>   - Defines transitions:
>     - `UNSPECIFIED -> SCHEDULED`
>     - `SCHEDULED -> STARTED`
>     - `STARTED -> COMPLETED | FAILED`
>     - `SCHEDULED | STARTED -> TERMINATED | TIMED_OUT`
> - **API change** (`chasm/statemachine.go`):
>   - Reorders `Transition.apply` signature to `apply(SM, MutableContext, E)` and updates `NewTransition`/`Apply` accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 59cd16886c581cdcdcd790b6f888a4ba79f24617. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8523, opened 2025-10-22
Refactor chasm statemachine helper

bergundy:
## What changed?

- Change the signature of the apply function to take the context as a second argument to allow passing in method pointers.
- Rename the "state" concept to "status" as that is what we are converging on when representing these status enums.

### DIFF
--- a/chasm/lib/callback/component.go
+++ b/chasm/lib/callback/component.go
@@ -17,6 +17,8 @@ type Callback struct {
 
 	// Persisted internal state
 	*callbackspb.CallbackState
+
+	status *callbackspb.CallbackStatus
 }
 
 func NewCallback(
@@ -39,11 +41,11 @@ func (c *Callback) LifecycleState(_ chasm.Context) chasm.LifecycleState {
 	return chasm.LifecycleStateRunning
 }
 
-func (c *Callback) State() callbackspb.CallbackStatus {
+func (c *Callback) StateMachineState() callbackspb.CallbackStatus {
 	return c.Status
 }
 
-func (c *Callback) SetState(status callbackspb.CallbackStatus) {
+func (c *Callback) SetStateMachineState(status callbackspb.CallbackStatus) {
 	c.Status = status
 }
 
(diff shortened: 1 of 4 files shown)


dandavison:
I don't think I'm in favor of renaming "state" to "status".
- The status field on Chasm components is just one example of a field which might use the statemachine helper.
- "state" seems to me the appropriate name for the state concept in a state machine library. Using the same name for the concept in the library as for the field in the client of the library is a tight coupling that isn't desirable IMO.

(I made the same comment [here](https://github.com/temporalio/temporal/pull/8496#discussion_r2437016770))

bergundy:
We are discussing the change in terminology internally. Will wait with this PR until we finalize.

bergundy:
> I don't think I'm in favor of renaming "state" to "status".
> 
> * The status field on Chasm components is just one example of a field which might use the statemachine helper.
> * "state" seems to me the appropriate name for the state concept in a state machine library. Using the same name for the concept in the library as for the field in the client of the library is a tight coupling that isn't desirable IMO.
> 
> (I made the same comment [here](https://github.com/temporalio/temporal/pull/8496#discussion_r2437016770))

I missed your previous comment, I'm with you.

bergundy:
I'm merging this already but @dandavison feel free to keep the discussion open if you feel like the new method names are not what you were expecting.

chasm/lib/callback/component.go
```diff
@@ -17,6 +17,8 @@ type Callback struct {
 
 	// Persisted internal state
 	*callbackspb.CallbackState
+
+	status *callbackspb.CallbackStatus
```
spkane31, line 21: Why do you need this change? `CallbackStatus` is a part of the `CallbackState` field? Feels redundant

bergundy: Ah oops, I think I somehow had an old version of this file

bergundy: This field accidentally popped back into the code. @spkane31 will you remove as part of your executors PR?

PR #8524, opened 2025-10-22
Proto changes for async deployment info propagation

ShahabT:
## What changed?
Adding internal proto changes to support routing config revision number and async operations.

## Why?
^

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
@@ -368,12 +368,10 @@ message SyncDeploymentUserDataRequest {
     // Ignored by the task queue if new revision number is not greater that what it has.
     temporal.api.deployment.v1.RoutingConfig update_routing_config = 10;
     // Optional map of build id to upsert version data.
-    // Ignored if `update_routing_config` is present and has an outdated revision number.
     // (-- api-linter: core::0203::required=disabled
     //     aip.dev/not-precedent: Not following Google API format --)
     map<string, temporal.server.api.deployment.v1.WorkerDeploymentVersionData> upsert_versions_data = 11;
     // List of build ids to forget from task queue.
-    // Ignored if `update_routing_config` is present and has an outdated revision number.
     repeated string forget_versions = 12;
 }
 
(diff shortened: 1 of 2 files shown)


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -355,15 +359,29 @@ message SyncDeploymentUserDataRequest {
...
         // The version whose data should be cleaned from the task queue.
-        temporal.server.api.deployment.v1.WorkerDeploymentVersion forget_version = 7;
+        temporal.server.api.deployment.v1.WorkerDeploymentVersion forget_version = 7 [deprecated = true];
     }
+
+    // Absent means no change.
+    // Ignored by the task queue if new revision number is not greater that what it has.
+    temporal.api.deployment.v1.RoutingConfig update_routing_config = 10;
+    // Optional map of build id to upsert version data.
+    // Ignored if `update_routing_config` is present and has an outdated revision number.
```
ShahabT: @Shivs11 probably best to ignore the whole request when it has out dated routing info. But that doesn't entirely protect us from applying out dated version data. I think we should be fine for now, but if that becomes an issue we can add separate counters for each version data too.

ShahabT: discussed offline, best to not mix them together.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -101,10 +101,17 @@ message DeploymentData {
...
     // routed to the Current Version. If there is no Current Version, any poller with UNVERSIONED
     // (or unspecified) WorkflowVersioningMode will receive the tasks.
-    repeated temporal.server.api.deployment.v1.DeploymentVersionData versions = 2;
+    // Remove after `AsyncSetCurrentAndRamping` workflow version is irreversibly enabled.
+    repeated temporal.server.api.deployment.v1.DeploymentVersionData versions = 2 [deprecated = true];
 
     // Present if the task queue's ramping version is unversioned.
-    temporal.server.api.deployment.v1.DeploymentVersionData unversioned_ramp_data = 3;
+    // Remove after `AsyncSetCurrentAndRamping` workflow version is irreversibly enabled.
+    temporal.server.api.deployment.v1.DeploymentVersionData unversioned_ramp_data = 3 [deprecated = true];
```
Shivs11, line 109: thinking a bit more about this, do we really wanna remove this? I don't see another way to represent unversioned ramps/current version in TQ unless I have missed something here....

Shivs11: Sure, one could find this by going over each of the routingConfig's present for this TQ and then realizing if the most recently updated config has an unversioned ramp/current or not but that seems like some more work compared to this nice structure we had in place earlier

ShahabT: If we stored routing info like this PR from the beginning we wouldn't need this field. All info that we need is in the routing infos, better not to store repetitive data in different format to make our read easier because that could cause sources of discrepancy later on.

ShahabT:
this is going to merge in #8553

PR #8531, opened 2025-10-23
CHASM: Best effort pure task deletion

yycptt:
## What changed?
- CHASM: Best effort pure task deletion

## Why?
- Performance improvement, prevent invalid physical pure tasks from firing.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/tree_test.go
+++ b/chasm/tree_test.go
@@ -2164,10 +2164,10 @@ func (s *nodeSuite) TestCloseTransaction_NewComponentTasks() {
 		TestOutboundSideEffectTask{},
 	)
 
-	// TODO: delete task
-	// s.nodeBackend.EXPECT().DeleteCHASMPureTasks(s.timeSource.Now().UTC()).Times(1)
 	mutation, err := root.CloseTransaction()
 	s.NoError(err)
+
+	s.Equal(s.timeSource.Now().UTC(), s.nodeBackend.LastDeletePureTaskCall())
 
 	rootAttr := mutation.UpdatedNodes[""].GetMetadata().GetComponentAttributes()
 	s.Len(rootAttr.SideEffectTasks, 1) // Only one valid side effect task.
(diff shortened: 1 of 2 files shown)


chasm/tree_test.go
```diff
@@ -2138,6 +2142,8 @@ func (s *nodeSuite) TestCloseTransaction_NewComponentTasks() {
 		TestOutboundSideEffectTask{},
 	)
 
+	// TODO: delete task
+	// s.nodeBackend.EXPECT().DeleteCHASMPureTasks(s.timeSource.Now().UTC()).Times(1)
```
prathyushpv: Isn't this deletion needed?

service/history/workflow/mutable_state_impl_test.go
```diff
@@ -5314,3 +5314,77 @@ func (s *mutableStateSuite) TestHasRequestID_EmptyExecutionState() {
...
+				},
+				{
+					VisibilityTimestamp: now.Add(time.Minute),
+					Category:            tasks.CategoryTimer,
+				},
+			}
+			s.mutableState.BestEffortDeleteTasks = make(map[tasks.Category][]tasks.Key)
+
+			s.mutableState.DeleteCHASMPureTasks(tc.maxScheduledTime)
+			s.Len(s.mutableState.chasmPureTasks, tc.expectedRemaining)
```
prathyushpv, line 5431: Does it make sense to check the scheduled time of remaining tasks to make sure that the expected tasks are deleted?

chasm/tree_test.go
```diff
@@ -2361,6 +2367,9 @@ func (s *nodeSuite) TestCloseTransaction_ApplyMutation_PureTasks() {
 	err = root.ApplyMutation(incomingMutation)
 	s.NoError(err)
 
+	// TODO: delete task
```
prathyushpv: If this call is not made, will it get removed?

chasm/tree_test.go
yycptt: Good catch! Thanks. We've switch to a handwritten mock, so all invocation will just go through by default. I've added the assertion.

service/history/workflow/mutable_state_impl_test.go
yycptt: Make sense. Added the check.

chasm/tree_test.go
yycptt: Same situation as your first comment. Added the assertion.

PR #8533, opened 2025-10-23
Add chasm.Context.ExecutionKey() method

bergundy:
## What changed?

- Add chasm.Context.ExecutionKey() method
- Refactor `context.go` to use better struct names and unexport concepts that should not be exported
- Add some docstrings

### DIFF
--- a/chasm/lib/scheduler/helper_test.go
+++ b/chasm/lib/scheduler/helper_test.go
@@ -95,7 +95,6 @@ func setupSchedulerForTest(t *testing.T) (*scheduler.Scheduler, chasm.MutableCon
 			TransitionCount:          1,
 		}
 	}
-	nodeBackend.HandleIsWorkflow = func() bool { return false }
 
 	node := chasm.NewEmptyTree(registry, timeSource, nodeBackend, nodePathEncoder, logger)
 	ctx := chasm.NewMutableContext(context.Background(), node)
--- a/chasm/lib/scheduler/scheduler_suite_test.go
+++ b/chasm/lib/scheduler/scheduler_suite_test.go
@@ -68,7 +68,7 @@ func (s *schedulerSuite) SetupTest() {
 		HandleCurrentVersionedTransition: func() *persistencespb.VersionedTransition {
 			return &persistencespb.VersionedTransition{
 				NamespaceFailoverVersion: 1,
-				TransitionCount:          2,
+				TransitionCount:          1,
 			}
 		},
 	}
--- a/chasm/node_backend_mock.go
+++ b/chasm/node_backend_mock.go
@@ -13,8 +13,8 @@ import (
 )
 
 // MockNodeBackend is a lightweight manual mock for the NodeBackend interface.
-// Methods may be stubbed by assigning the corresponding Handle fields. Call history
-// is recorded in the struct fields (thread-safe).
+// Methods may be stubbed by assigning the corresponding Handle fields. Update call history is recorded in the struct
+// fields (thread-safe).
 type MockNodeBackend struct {
 	// Optional function overrides. If nil, methods return zero-values.
 	HandleGetExecutionState          func() *persistencespb.WorkflowExecutionState
@@ -115,6 +115,24 @@ func (m *MockNodeBackend) UpdateWorkflowStateStatus(
 	m.mu.Unlock()
 
 	return false, nil
+}
+
+func (m *MockNodeBackend) LastUpdateWorkflowState() enumsspb.WorkflowExecutionState {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if len(m.UpdateCalls) == 0 {
+		return enumsspb.WORKFLOW_EXECUTION_STATE_UNSPECIFIED
+	}
+	return m.UpdateCalls[len(m.UpdateCalls)-1].State
+}
+
+func (m *MockNodeBackend) LastUpdateWorkflowStatus() enumspb.WorkflowExecutionStatus {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	if len(m.UpdateCalls) == 0 {
+		return enumspb.WORKFLOW_EXECUTION_STATUS_UNSPECIFIED
+	}
+	return m.UpdateCalls[len(m.UpdateCalls)-1].Status
 }
 
 func (m *MockNodeBackend) IsWorkflow() bool {
--- a/chasm/tree_test.go
+++ b/chasm/tree_test.go
@@ -1824,6 +1824,10 @@ func (s *nodeSuite) TestCloseTransaction_LifecycleChange() {
 	chasmCtx := NewMutableContext(context.Background(), node)
 	_, err := node.Component(chasmCtx, ComponentRef{componentPath: rootPath})
 	s.NoError(err)
+	_, err = node.CloseTransaction()
+	s.NoError(err)
+	s.Equal(enumsspb.WORKFLOW_EXECUTION_STATE_RUNNING, s.nodeBackend.LastUpdateWorkflowState())
+	s.Equal(enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING, s.nodeBackend.LastUpdateWorkflowStatus())
 
 	// Test force terminate case
 	_, err = node.Component(chasmCtx, ComponentRef{componentPath: rootPath})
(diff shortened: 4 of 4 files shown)


chasm/context.go
```diff
@@ -14,7 +14,11 @@ type Context interface {
 	// NOTE: component created in the current transaction won't have a ref
 	// this is a Ref to the component state at the start of the transition
 	Ref(Component) ([]byte, error)
+	// Now returns the current time in the context of the given component.
+	// In a context of a transaction, this time must be used to allow for framework support of pause and time skipping.
 	Now(Component) time.Time
+	// ExecutionKey returns the execution key for the execution the context is operating on.
+	ExecutionKey() EntityKey
```
yycptt, line 19: I wonder when this will be used? Inside task executor logic?

chasm/context.go
```diff
@@ -40,53 +47,81 @@ type MutableContext interface {
...
 }
 
-type ContextImpl struct {
+type immutableCtx struct {
 	// The context here is not really used today.
 	// But it will be when we support partial loading later,
 	// and the framework potentially needs to go to persistence to load some fields.
 	ctx context.Context
 
+	executionKey EntityKey
```
yycptt, line 54: nit: can we get it from nodeBackend on demand? I think most cases caller has a componentRef and key is available there and don't need to get it from the Context.

yycptt (approved):
Thanks for the renaming and new docstring!

chasm/context.go
bergundy: Every component we're writing is storing this information. We can just not do that anymore.

chasm/context.go
bergundy: It's meant to be accessible from within an engine method.

chasm/context.go
yycptt: hmm I think I see it now, if the request doesn't have runID specified, then it's not available in the Ref. And in pure task executor there's no access to Ref to ns, entity, runID are not available. ok, makes sense.

chasm/context.go
yycptt: I still think most Context doesn't really need this, chasm engine methods should already know the key from either the ComponentRef or the mutable state cache. Probably maybe I don't see your use case. 
Anyway I think you don't need to fix all the mocks in unit test if you get it in when the method is called.

bergundy: I will fix the mocks, yup.

bergundy: See here where we are storing our own ID and namespace ID: https://github.com/temporalio/temporal/blob/11e54a1b9ae638e89b2161d6e50fdb0497a7306a/chasm/lib/scheduler/scheduler.go#L69-L70

chasm/node_backend_mock.go
```diff
@@ -0,0 +1,145 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	enumsspb "go.temporal.io/server/api/enums/v1"
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+	"go.temporal.io/server/common/definition"
+	"go.temporal.io/server/common/nexus/nexusrpc"
+	"go.temporal.io/server/service/history/tasks"
+)
+
+// MockNodeBackend is a lightweight manual mock for the NodeBackend interface.
+// Methods may be stubbed by assigning the corresponding Handle fields. Call history
```
yycptt:
```suggestion
// Methods may be stubbed by assigning the corresponding Handle fields. Update call history
```

chasm/node_pure_task_mock.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+	"sync"
+)
+
+// MockNodePureTask is a lightweight manual mock for the NodePureTask interface.
+// Methods may be stubbed by assigning the corresponding Handle fields. Call history
+// is recorded in the struct fields (thread-safe).
+type MockNodePureTask struct {
+	HandleExecutePureTask  func(baseCtx context.Context, taskAttributes TaskAttributes, taskInstance any) (bool, error)
```
yycptt, line 12: I think Execute only returns an error

chasm/lib/scheduler/helper_test.go
```diff
@@ -87,17 +85,17 @@ func setupSchedulerForTest(t *testing.T) (*scheduler.Scheduler, chasm.MutableCon
...
+	nodeBackend.HandleIsWorkflow = func() bool { return false }
+	nodeBackend.HandleCurrentVersionedTransition = func() *persistencespb.VersionedTransition {
+		return &persistencespb.VersionedTransition{
+			NamespaceFailoverVersion: 1,
+			TransitionCount:          1,
+		}
 	}
-	nodeBackend.EXPECT().CurrentVersionedTransition().Return(currentVT).AnyTimes()
-	nodeBackend.EXPECT().AddTasks(gomock.Any()).Return().AnyTimes()
+	nodeBackend.HandleIsWorkflow = func() bool { return false }
```
yycptt: nit: not needed and duplicates L91

chasm/lib/scheduler/helper_test.go
```diff
@@ -87,17 +85,17 @@ func setupSchedulerForTest(t *testing.T) (*scheduler.Scheduler, chasm.MutableCon
...
-	nodeBackend.EXPECT().GetCurrentVersion().Return(int64(1)).AnyTimes()
-	nodeBackend.EXPECT().UpdateWorkflowStateStatus(gomock.Any(), gomock.Any()).Return(true, nil).AnyTimes()
-	nodeBackend.EXPECT().GetWorkflowKey().Return(tv.Any().WorkflowKey()).AnyTimes()
-	nodeBackend.EXPECT().IsWorkflow().Return(false).AnyTimes()
-	currentVT := &persistencespb.VersionedTransition{
-		NamespaceFailoverVersion: 1,
-		TransitionCount:          1,
+	nodeBackend.HandleNextTransitionCount = func() int64 { return 2 }
+	nodeBackend.HandleGetCurrentVersion = func() int64 { return 1 }
+	nodeBackend.HandleGetWorkflowKey = tv.Any().WorkflowKey
```
yycptt, line 90: nit: maybe tv.Any().WorkflowKey can be the default?

chasm/lib/scheduler/scheduler_suite_test.go
```diff
@@ -65,28 +60,18 @@ func (s *schedulerSuite) SetupTest() {
...
-		TransitionCount:          1,
+	s.nodeBackend = &chasm.MockNodeBackend{
+		HandleNextTransitionCount: func() int64 { return 2 },
+		HandleGetCurrentVersion:   func() int64 { return 1 },
+		HandleGetWorkflowKey:      tv.Any().WorkflowKey,
+		HandleIsWorkflow:          func() bool { return false },
+		HandleCurrentVersionedTransition: func() *persistencespb.VersionedTransition {
+			return &persistencespb.VersionedTransition{
+				NamespaceFailoverVersion: 1,
+				TransitionCount:          2,
```
yycptt:
```suggestion
				TransitionCount:          1,
```

chasm/tree_test.go
```diff
@@ -1879,83 +1820,51 @@ func (s *nodeSuite) TestCloseTransaction_EmptyNode() {
...
-	s.NoError(err)
-
 	// Test force terminate case
 	_, err = node.Component(chasmCtx, ComponentRef{componentPath: rootPath})
 	s.NoError(err)
 	node.terminated = true
-	s.nodeBackend.EXPECT().UpdateWorkflowStateStatus(
-		enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED,
-		enumspb.WORKFLOW_EXECUTION_STATUS_TERMINATED,
-	).Return(true, nil).Times(1)
```
yycptt, line 1906: I do want to validate those Update calls and the state & status parameter.

chasm/tree_test.go
```diff
@@ -2465,55 +2347,34 @@ func (s *nodeSuite) TestCloseTransaction_ApplyMutation_PureTasks() {
...
+
 	// Then terminate the node and verify only that node will be in the mutation.
 	err = node.Terminate(TerminateComponentRequest{})
 	s.NoError(err)
 	s.True(node.terminated)
 
-	s.nodeBackend.EXPECT().UpdateWorkflowStateStatus(
-		enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED,
-		enumspb.WORKFLOW_EXECUTION_STATUS_TERMINATED,
-	).Return(true, nil).Times(1)
```
yycptt, line 2515: plz assert this.

chasm/tree_test.go
```diff
@@ -2618,14 +2477,9 @@ func (s *nodeSuite) testComponentTree() *Node {
 }
 
 func (s *nodeSuite) TestExecuteImmediatePureTask() {
-	tv := testvars.New(s.T())
+	// tv := testvars.New(s.T())
```
yycptt: nit: remove

chasm/node_pure_task_mock.go
bergundy: I checked, it returns bool. Not sure that was intended. I would also like to not export this helper, it only has one place where it's used that doesn't justify this. Also, library authors should not use these lower level APIs.

PR #8536, opened 2025-10-24
Validate activity task stamps before reuse

yiminc:
## Summary
- add the activity task stamp to `IsActivityTaskValidRequest` and generated API stubs
- require history's activity validation logic to match the stored stamp
- plumb the stamp from matching task validation and cover the new behavior with tests
- increment activity stamps for each retry attempt and extend tests for paused retries

## Testing
- `GOTOOLCHAIN=local go test -tags test_dep ./service/history/workflow -run TestRetryActivity` *(fails: go.mod requires go >= 1.25.0 and local toolchain is go1.24.3)*

------
https://chatgpt.com/codex/tasks/task_b_68faffae045c832c9d8220bcbe50bac7

### DIFF
--- a/service/history/transfer_queue_standby_task_executor.go
+++ b/service/history/transfer_queue_standby_task_executor.go
@@ -564,10 +564,14 @@ func (t *transferQueueStandbyTaskExecutor) pushActivity(
 		return nil
 	}
 
+	activityTask, ok := task.(*tasks.ActivityTask)
+	if !ok {
+		return serviceerror.NewInternal("task is not an ActivityTask")
+	}
 	pushActivityInfo := postActionInfo.(*activityTaskPostActionInfo)
 	return t.transferQueueTaskExecutorBase.pushActivity(
 		ctx,
-		task.(*tasks.ActivityTask),
+		activityTask,
 		pushActivityInfo.activityTaskScheduleToStartTimeout,
 		pushActivityInfo.versionDirective,
 		pushActivityInfo.priority,
--- a/service/history/workflow/task_generator.go
+++ b/service/history/workflow/task_generator.go
@@ -552,6 +552,7 @@ func (r *TaskGeneratorImpl) GenerateActivityTasks(
 		TaskQueue:        activityInfo.TaskQueue,
 		ScheduledEventID: activityInfo.ScheduledEventId,
 		Version:          activityInfo.Version,
+		Stamp:            activityInfo.Stamp,
 	})
 
 	return nil


service/history/api/isactivitytaskvalid/api.go
```diff
@@ -49,14 +49,15 @@ func Invoke(
...
+	stamp int32,
 ) (bool, error) {
 	mutableState := workflowLease.GetMutableState()
 	if !mutableState.IsWorkflowExecutionRunning() {
 		return false, consts.ErrWorkflowCompleted
 	}
 
 	ai, ok := mutableState.GetActivityInfo(scheduledEventID)
-	if ok && ai.StartedEventId == common.EmptyEventID {
+	if ok && ai.StartedEventId == common.EmptyEventID && ai.GetStamp() == stamp {
```
bergundy, line 60: Won't this fail for existing activity tasks that were generated without a stamp?

yiminc: Stamp was added to AI the same time they were added to the matching tasks. For tasks before that, the Stamp from AI would be 0 and same for the tasks. But I could add a defensive check to enforce only when both stamp are > 0.

yiminc:
> The way that this is implemented feels error prone to me. It's too easy to mutate the activity and forget to update the stamp. I skimmed the locations where the activity is updated in the code and I think we may have some edge cases that are not covered (need to double check). I would consider updating the stamp in `MutableStateImpl.UpdateActivity` for any field change that warrants updating it.

The purpose of this validation is to avoid building up backlog with invalid tasks. There isn't much of negative consequence to mis-update the stamp for any update to activity, as long as the most common case of retry has the stamp updated that would serve the purpose. But if there is better way of doing this, we could improve later.

service/history/api/isactivitytaskvalid/api.go
bergundy: Ah okay, so this should be fine.

bergundy:
> > The way that this is implemented feels error prone to me. It's too easy to mutate the activity and forget to update the stamp. I skimmed the locations where the activity is updated in the code and I think we may have some edge cases that are not covered (need to double check). I would consider updating the stamp in `MutableStateImpl.UpdateActivity` for any field change that warrants updating it.
> 
> The purpose of this validation is to avoid building up backlog with invalid tasks. There isn't much of negative consequence to mis-update the stamp for any update to activity, as long as the most common case of retry has the stamp updated that would serve the purpose. But if there is better way of doing this, we could improve later.

We should do better for when we port this to CHASM.

service/history/api/isactivitytaskvalid/api.go
yiminc: we decided to revert this because the additional check will cause the first task always be valid and that could cause head-of-queue blocking issue on standby side.

PR #8540, opened 2025-10-24
Added standalone activity chasm tasks and E2E functionality.

fretz12:
## What changed?
Added standalone activity Chasm tasks. Added handling of start activity and e2e implementation of standalone activity start execution with existing services. Updated protos related to standalone activities. 

## Why?
The Chasm tasks are needed to kick off standalone activity execution via the existing services. Proto changes needed to so that the component ref can be passed and handled via service stack.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/api/historyservice/v1/request_response.pb.go
+++ b/api/historyservice/v1/request_response.pb.go
@@ -1618,8 +1618,9 @@ type RecordActivityTaskStartedRequest struct {
 	ScheduledDeployment *v16.Deployment `protobuf:"bytes,10,opt,name=scheduled_deployment,json=scheduledDeployment,proto3" json:"scheduled_deployment,omitempty"`
 	// Versioning directive that was sent by history when scheduling the task.
 	VersionDirective *v112.TaskVersionDirective `protobuf:"bytes,12,opt,name=version_directive,json=versionDirective,proto3" json:"version_directive,omitempty"`
-	// Reference to the Chasm component for activity execution (if applicable). The component will have all the
-	// necessary data, deprecating the need for the separate request fields used here.
+	// Reference to the Chasm component for activity execution (if applicable). For standalone activities, all necessary
+	// start information is carried within this component, obviating the need to use the fields that apply to embedded
+	// activities with the exception of version_directive.
 	ComponentRef  []byte `protobuf:"bytes,13,opt,name=component_ref,json=componentRef,proto3" json:"component_ref,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
--- a/api/matchingservice/v1/request_response.pb.go
+++ b/api/matchingservice/v1/request_response.pb.go
@@ -744,8 +744,9 @@ type AddActivityTaskRequest struct {
 	ForwardInfo      *v18.TaskForwardInfo      `protobuf:"bytes,11,opt,name=forward_info,json=forwardInfo,proto3" json:"forward_info,omitempty"`
 	Stamp            int32                     `protobuf:"varint,12,opt,name=stamp,proto3" json:"stamp,omitempty"`
 	Priority         *v11.Priority             `protobuf:"bytes,13,opt,name=priority,proto3" json:"priority,omitempty"`
-	// Reference to the Chasm component for activity execution (if applicable). The component will have all the
-	// necessary data, deprecating the need for the separate request fields used here.
+	// Reference to the Chasm component for activity execution (if applicable). For standalone activities, all
+	// necessary start information is carried within this component, obviating the need to use the fields that apply to
+	// embedded activities.
 	ComponentRef  []byte `protobuf:"bytes,14,opt,name=component_ref,json=componentRef,proto3" json:"component_ref,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache


chasm/lib/activity/activity.go
```diff
@@ -56,23 +63,31 @@ func NewStandaloneActivity(
...
 			UserMetadata: request.UserMetadata,
 		}),
 		Visibility: chasm.NewComponentField(ctx, visibility),
-	}, nil
+	}
+
+	ctx.AddTask(
+		activity,
+		chasm.TaskAttributes{},
+		&activitypb.ActivityStartExecuteTask{
```
bergundy: I think this is a better name.

```suggestion
		&activitypb.ActivityDispatchTask{
```

chasm/lib/activity/activity.go
```diff
@@ -56,23 +63,31 @@ func NewStandaloneActivity(
...
+			Attempt: 1,
```
bergundy: This should always be taken from the attempt data field even if we know it's 1 here just for completeness IMHO.

chasm/lib/activity/activity.go
```diff
@@ -56,23 +63,31 @@ func NewStandaloneActivity(
...
 		RequestData: chasm.NewDataField(ctx, &activitypb.ActivityRequestData{
 			Input:        request.Input,
 			Header:       request.Header,
 			UserMetadata: request.UserMetadata,
 		}),
 		Visibility: chasm.NewComponentField(ctx, visibility),
-	}, nil
+	}
+
+	ctx.AddTask(
```
bergundy: This call should be made through the state machine transition, not called directly here.

chasm/lib/activity/activity.go
```diff
@@ -82,6 +97,91 @@ func NewEmbeddedActivity(
...
+}
+
+// HandleRecordActivityTaskStarted processes the start of an activity task execution. It transitions the activity to
+// started state, updates the component fields, and returns the updated activity component.
+func HandleRecordActivityTaskStarted(
+	ctx context.Context,
+	activityRef chasm.ComponentRef,
+	versionDirective *taskqueuespb.TaskVersionDirective,
+	workerIdentity string,
+) (*Activity, error) {
```
bergundy: You should return the full task here not the activity instance. Typically chasm components should not be accessed outside of the `UpdateComponent` and `ReadComponent` scopes. That could introduce subtle bugs where their fields are mutated and the author would not be aware that the changes are not committed.

chasm/lib/activity/activity.go
```diff
@@ -82,6 +97,91 @@ func NewEmbeddedActivity(
...
+			attempt.LastWorkerIdentity = workerIdentity
+
+			if versionDirective := versionDirective.GetDeploymentVersion(); versionDirective != nil {
+				attempt.LastDeploymentVersion = &deploymentpb.WorkerDeploymentVersion{
+					BuildId:        versionDirective.GetBuildId(),
+					DeploymentName: versionDirective.GetDeploymentName(),
+				}
+			}
+
+			return a, nil
```
bergundy: There is a data race here. You are returning an object that can never be accessed outside of a lock and may be mutate by concurrent updates while the lock is held.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+type ActivityStartTaskExecutorOptions struct {
+	fx.In
+
+	MatchingClient resource.MatchingClient
+}
+
+type ActivityStartTaskExecutor struct {
+	MatchingClient resource.MatchingClient
```
bergundy: nit: you can just embed the options and you'll have everything available directly to the executor.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,76 @@
...
+	"go.uber.org/fx"
+)
+
+type ActivityStartTaskExecutorOptions struct {
+	fx.In
+
+	MatchingClient resource.MatchingClient
+}
+
+type ActivityStartTaskExecutor struct {
```
bergundy: Ideally we would not export anything that doesn't need to be exported.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,76 @@
...
+	return true, nil
+}
+
+func (e *ActivityStartTaskExecutor) Execute(
+	ctx context.Context,
+	activityRef chasm.ComponentRef,
+	_ chasm.TaskAttributes,
+	_ *activitypb.ActivityStartExecuteTask,
+) error {
+	activityState, err := GetActivityState(ctx, activityRef)
```
bergundy: Same issue here. Construct the matching request under a lock.

chasm/task.go
```diff
@@ -22,6 +22,25 @@ type (
...
+		// Validate determines whether a task should proceed with execution based on the current context, component
+		// state, task attributes, and task data.
+		//
+		// This function serves as a gate to prevent unnecessary task execution in several scenarios:
+		// 1. Standby cluster deduplication: When state is replicated to standby clusters, tasks are also replicated.
+		//    Validate allows standby clusters to check if a task was already completed on the active cluster and
+		//    skip execution if so (e.g., checking if an activity already transitioned from scheduled to started state).
+		// 2. Task obsolescence: Tasks can become irrelevant when state changes invalidate them (e.g., when a scheduler
+		//    is updated to run at a different time, making the previously scheduled task invalid for the new state).
+		// 3. Optimization: For pure tasks that can run in a single transaction, Validate is called before execution
```
bergundy: This is the same as 2.

chasm/task.go
```diff
@@ -22,6 +22,25 @@ type (
 	}
 
 	TaskValidator[C any, T any] interface {
+		// Validate determines whether a task should proceed with execution based on the current context, component
```
bergundy, line 25: Thanks!!

chasm/task.go
```diff
@@ -22,6 +22,25 @@ type (
...
+		// 3. Optimization: For pure tasks that can run in a single transaction, Validate is called before execution
+		//    to avoid unnecessary work.
+		//
+		// The framework automatically calls Validate at key points, such as after closing transactions, to check all
+		// generated tasks before they execute.
+		//
+		// Returns:
+		// - (true, nil) if the task is valid and should be executed
+		// - (false, nil) if the task should be silently dropped (it's no longer relevant)
+		// - (false, error) if validation fails with an error
```
bergundy:
```suggestion
		// - (anything, error) if validation fails with an error
```

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -101,6 +107,61 @@ func Invoke(
...
+	if err != nil {
+		return nil, err
+	}
+
+	lastHeartbeat, err := activity.GetActivityLastHeartbeat(ctx, activityRef)
+	if err != nil {
+		return nil, err
+	}
+
+	options := updatedActivity.GetActivityOptions()
```
bergundy: There's no point in locking the activity 4 times, create one function that is executed once under a lock instead and returns `RecordActivityTaskStartedResponse`.

Also, we mentioned writing the code in a way that would work for a workflow CHASM activity. Please make sure we use the `Store` interface (we should rename it but the concept should still be there).

service/history/handler.go
```diff
@@ -350,6 +350,8 @@ func (h *Handler) RecordActivityTaskStarted(ctx context.Context, request *histor
 		return nil, h.convertError(err)
 	}
 
+	ctx = chasm.NewEngineContext(ctx, h.chasmEngine)
```
bergundy: I thought the interceptor we put in there should already do this...

tests/standalone_activity_test.go
```diff
@@ -35,15 +38,30 @@ func (s *standaloneActivityTestSuite) TestStartActivityExecution() {
...
-		Namespace:  s.Namespace().String(),
-		ActivityId: testcore.RandomizeStr(t.Name()),
-		ActivityType: &commonpb.ActivityType{
-			Name: "TestActivity",
+	activityID := testcore.RandomizeStr(t.Name())
+	activityType := &commonpb.ActivityType{
+		Name: "test-activity-type",
+	}
+	input := &commonpb.Payloads{
+		Payloads: []*commonpb.Payload{
```
bergundy: There should be a helper to create this payload.

chasm/lib/activity/activity.go
fretz12: Moved.

chasm/lib/activity/activity.go
fretz12: Renamed

chasm/lib/activity/activity.go
fretz12: Done

chasm/lib/activity/activity_tasks.go
fretz12: Made private, along with ActivityStartTaskExecutorOptions

chasm/lib/activity/activity_tasks.go
fretz12: Refactored accordingly

chasm/task.go
fretz12: combined with 2

chasm/task.go
fretz12: fixed

chasm/lib/activity/activity.go
fretz12: I understand the importance of doing all component r/w atomically within the update func now. Refactored.

chasm/lib/activity/activity.go
fretz12: Understood now, refactored to return `RecordActivityTaskStartedResponse`

service/history/api/recordactivitytaskstarted/api.go
fretz12: Understood now, refactored.

chasm/lib/activity/activity_tasks.go
fretz12: Understood now, refactored.

service/history/handler.go
fretz12: Currently it only injects the engine if its a RPC handler registered in chasm. This path is still outside of chasm. 
```
func (i *ChasmRequestInterceptor) Intercept(
	ctx context.Context,
	req interface{},
	info *grpc.UnaryServerInfo,
	handler grpc.UnaryHandler,
) (resp interface{}, retError error) {
	if strings.HasPrefix(info.FullMethod, chasmRequestPrefix) {
		defer metrics.CapturePanic(i.logger, i.metricsHandler, &retError)

		ctx = NewEngineContext(ctx, i.engine)
	}

	return handler(ctx, req)
}

```

I can however, just set the engine outside the prefix check. Should do no harm, unless you think that's unwise. Let me know.

tests/standalone_activity_test.go
fretz12: refactored to a helper func.

service/history/handler.go
bergundy: Populate the context for all handlers, and probably the panic capture is also applicable to all of them, but that can be done separately and I wouldn't do that in the feature branch.

bergundy: Just the context should be set for all handlers now, there's no harm in that.

chasm/lib/activity/activity.go
```diff
@@ -56,23 +67,27 @@ func NewStandaloneActivity(
...
 			Input:        request.Input,
 			Header:       request.Header,
 			UserMetadata: request.UserMetadata,
 		}),
 		Visibility: chasm.NewComponentField(ctx, visibility),
-	}, nil
+	}
+
+	// Standalone activities point to themselves as the store.
+	activity.Store = chasm.Field[ActivityStore](chasm.ComponentPointerTo(ctx, activity))
```
bergundy: I'm wondering if there's value in actually referencing `self` here, there's a storage cost to storing this pointer, we could instead check if there is not a component ref to the store and fallback to `self`.

chasm/lib/activity/activity.go
```diff
@@ -82,18 +97,111 @@ func NewEmbeddedActivity(
...
-	store, err := a.Store.Get(ctx)
+// HandleRecordActivityTaskStarted processes the start of an activity task execution. It transitions the activity to
+// started state, updates the component fields, and returns the updated activity component.
+func HandleRecordActivityTaskStarted(
+	ctx context.Context,
+	activityRef chasm.ComponentRef,
+	versionDirective *taskqueuespb.TaskVersionDirective,
+	workerIdentity string,
+) (*historyservice.RecordActivityTaskStartedResponse, error) {
+	response, _, err := chasm.UpdateComponent(
```
bergundy: Put the `UpdateComponent` call outside of this function and save the closure.

service/history/handler.go
fretz12: Changed to inject engine for all handlers.

chasm/lib/activity/proto/v1/tasks.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package temporal.server.chasm.lib.activity.proto.v1;
+
+option go_package = "go.temporal.io/server/chasm/lib/activity/gen/activitypb;activitypb";
+
+message ActivityDispatchTask {
```
dandavison, line 7: The name that the codebase uses for the history tasks whose executor calls `matching.AddFooTaskRequest` is "transfer task", not "dispatch task". So this should be called `ActivityTransferTask`.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,82 @@
...
+	"context"
+
+	"go.temporal.io/server/api/matchingservice/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"
+	"go.temporal.io/server/common/resource"
+	"go.uber.org/fx"
+)
+
+type activityDispatchTaskExecutorOptions struct {
```
dandavison, line 12: ~Following from my comment below, this should be called `activityTransferTaskExecutorOptions` and `activityTransferTaskExecutor` etc~

chasm/lib/activity/proto/v1/tasks.proto
fretz12: I don't have a huge preference. Transfer or dispatch both work well. @bergundy had suggested "dispatch".... I'll let him weight in.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,82 @@
...
+	_ chasm.TaskAttributes,
+	task *activitypb.ActivityDispatchTask,
+) (bool, error) {
+	attempt, err := activity.Attempt.Get(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	// TODO make sure we handle resets when we support them, as they will reset the attempt count
+	if activity.State() != activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED || task.Attempt != attempt.Count {
```
bergundy: I typically like to model this using the state machine transitions, basically checking if the machine can transition to Started (using the `Possible()` method) but this is also fine.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,82 @@
...
+	activityRef chasm.ComponentRef,
+	_ chasm.TaskAttributes,
+	_ *activitypb.ActivityDispatchTask,
+) error {
+	request, err := chasm.ReadComponent(
+		ctx,
+		activityRef,
+		func(a *Activity, ctx chasm.Context, activityRef chasm.ComponentRef) (*matchingservice.AddActivityTaskRequest, error) {
+
+			protoRef, err := chasm.ComponentRefToProtoRef(activityRef, nil)
```
bergundy: Get the updated ref here with `ctx.Ref(a)`

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,82 @@
...
+func (e *activityDispatchTaskExecutor) Execute(
+	ctx context.Context,
+	activityRef chasm.ComponentRef,
+	_ chasm.TaskAttributes,
+	_ *activitypb.ActivityDispatchTask,
+) error {
+	request, err := chasm.ReadComponent(
+		ctx,
+		activityRef,
+		func(a *Activity, ctx chasm.Context, activityRef chasm.ComponentRef) (*matchingservice.AddActivityTaskRequest, error) {
```
bergundy: This can be a method on the Activity struct instead of a closure.

chasm/lib/activity/proto/v1/tasks.proto
bergundy: I think dispatch makes sense because the task is to dispatch a task to a worker via matching, you'll just want to name the executor accordingly.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -42,6 +44,12 @@ func Invoke(
 	workflowConsistencyChecker api.WorkflowConsistencyChecker,
 	matchingClient matchingservice.MatchingServiceClient,
 ) (resp *historyservice.RecordActivityTaskStartedResponse, retError error) {
+	if activityRefProto := request.GetComponentRef(); activityRefProto != nil {
+		activityRef := chasm.ProtoRefToComponentRef(activityRefProto)
+
+		return activity.HandleRecordActivityTaskStarted(
```
bergundy: Repeating myself here but you can just change this to `chasm.UpdateComponent` with a method call on the `Activity` struct.

fretz12: Sure I can do that, but curious what's the reasoning behind this? So it's more explicit to chasm callers that there's a component mutation?

chasm/lib/activity/activity.go
fretz12: Valid point on efficiency, however the trade off is anywhere the store is being accessed we need if-else code. If you think memory cost is expensive we can go with nil -> self, else from a code perspective storing self would be cleaner.

chasm/lib/activity/activity_tasks.go
fretz12: Love it, makes it more readable. Changed.

chasm/lib/activity/activity_tasks.go
fretz12: Don't think I'm following. ctx.Ref(a) returns []byte, but here I don't want to do the serialization yet, but have it done together with the proto model when the request gets sent out.

chasm/lib/activity/proto/v1/tasks.proto
dandavison: OK let's discuss offline. I might be missing something but it seems to me to be exactly the same as the existing "transfer task" concept, in which case it would be confusing to use a new name.

chasm/lib/activity/activity_tasks.go

service/history/api/recordactivitytaskstarted/api.go
fretz12: Refactored.

chasm/lib/activity/activity.go
fretz12: Refactored

chasm/lib/activity/activity.go
bergundy: The cost of storing another reference is too high but we can potentially optimize this on the CHASM framework side. CC @yycptt, this is coming up again similarly to how the parent pointer optimization came up.

chasm/lib/activity/activity.go
```diff
@@ -82,18 +105,109 @@ func NewEmbeddedActivity(
...
+		}
+	}
+
 	store, err := a.Store.Get(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if store == nil {
+		return nil, errors.New("activity store is nil")
```
bergundy: I would just assume that this isn't nil and let the code panic, it'll end up in the same outcome. No need to be defensive here.

chasm/lib/activity/activity_tasks.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type activityDispatchTaskExecutorOptions struct {
+	fx.In
+
+	MatchingClient resource.MatchingClient
+}
+
+type activityDispatchTaskExecutor struct {
+	opts activityDispatchTaskExecutorOptions
```
bergundy, line 19: nit: if you embed the options, you can use `e.MatchingClient` directly.

chasm/lib/activity/library.go
```diff
@@ -33,3 +36,13 @@ func (l *library) Components() []*chasm.RegistrableComponent {
 		chasm.NewRegistrableComponent[*Activity]("activity"),
 	}
 }
+
+func (l *library) Tasks() []*chasm.RegistrableTask {
+	return []*chasm.RegistrableTask{
+		chasm.NewRegistrableSideEffectTask(
+			"startActivity",
+			l.ActivityStartTaskExecutor,
+			l.ActivityStartTaskExecutor,
```
bergundy: Is it dispatch or start? The executor and task names should be consistent.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -42,6 +44,26 @@ func Invoke(
 	workflowConsistencyChecker api.WorkflowConsistencyChecker,
 	matchingClient matchingservice.MatchingServiceClient,
 ) (resp *historyservice.RecordActivityTaskStartedResponse, retError error) {
+	if activityRefProto := request.GetComponentRef(); activityRefProto != nil {
+		activityRef := chasm.ProtoRefToComponentRef(activityRefProto)
+
+		response, _, err := chasm.UpdateComponent(
+			ctx,
+			activityRef,
+			(*activity.Activity).HandleRecordActivityTaskStarted,
```
bergundy: nit:

```suggestion
			(*activity.Activity).RecordActivityTaskStarted,
```

chasm/lib/activity/activity.go
```diff
@@ -82,18 +105,109 @@ func NewEmbeddedActivity(
...
 	}
-	if err := store.PopulateRecordActivityTaskStartedResponse(ctx, res); err != nil {
+
+	requestData, err := a.RequestData.Get(ctx)
+	if err != nil {
 		return err
 	}
-	// ...
+
+	options := a.GetActivityOptions()
```
bergundy: We said we're getting rid of the use of `ActivityOptions` and flattening them out, didn't we?

bergundy: Also @dandavison for the API changes.

chasm/lib/activity/activity.go
```diff
@@ -82,18 +105,109 @@ func NewEmbeddedActivity(
 ) {
 }
 
-func (a *Activity) PopulateRecordActivityTaskStartedResponse(ctx chasm.Context, res *historyservice.RecordActivityTaskStartedResponse) error {
+func (a *Activity) createAddActivityTaskRequest(_ chasm.Context, activityRef chasm.ComponentRef) (*matchingservice.AddActivityTaskRequest, error) {
+	protoRef, err := chasm.ComponentRefToProtoRef(activityRef, nil)
```
bergundy: My previous comment was lost, don't pass in the ref, use `ctx.Ref(a)` here instead.

fretz12: My response on last comment:
Don't think I'm following. ctx.Ref(a) returns []byte, but here I don't want to do the serialization yet, but have it done together with the proto model when the request gets sent out.

chasm/lib/activity/activity.go
fretz12: Ok, for now I can just represent nil->self. I'll add a TODO once we have parent pointers optimized.

chasm/lib/activity/activity_tasks.go
fretz12: Yea I tried 
```
type activityDispatchTaskExecutor struct {
	activityDispatchTaskExecutorOptions
}
```
but kept running into `bad result 1: cannot provide parameter objects: *activity.activityDispatchTaskExecutor embeds a dig.In"`
Am I doing something wrong?

chasm/lib/activity/proto/v1/tasks.proto
dandavison: Resolved -- @bergundy explained to me offline why it's not right to think of the new task as a "transfer chasm activity task".

chasm/lib/activity/activity.go
fretz12: Done

chasm/lib/activity/activity.go
fretz12: No longer relevant since we use nil for standalone activities.

chasm/lib/activity/activity.go
fretz12: Flattened. WIll refactor again once API changes are in.

chasm/lib/activity/library.go
fretz12: Missed it. Renamed.

service/history/api/recordactivitytaskstarted/api.go
fretz12: Renamed.

chasm/lib/activity/activity.go
fretz12: Per conversation, it's imporant to get the latest component ref as the one passed in may be outdated.

client/history/client.go
```diff
@@ -12,6 +12,7 @@ import (
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/api/historyservice/v1"
 	replicationspb "go.temporal.io/server/api/replication/v1"
+	"go.temporal.io/server/chasm"
```
fretz12: Getting a lint error: ../chasm/archetype.go:1: : import cycle not allowed in test (typecheck)
Re: We need to support routing by a chasm bytes ref.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -307,6 +307,9 @@ message RecordActivityTaskStartedRequest {
     reserved 11;
     // Versioning directive that was sent by history when scheduling the task.
     temporal.server.api.taskqueue.v1.TaskVersionDirective version_directive = 12;
+    // Reference to the Chasm component for activity execution (if applicable). The component will have all the
+    // necessary data, deprecating the need for the separate request fields used here.
```
dandavison: I don't understand this comment:

```
// The component will have all the
// necessary data, deprecating the need for the separate request fields used here.
```

fretz12: How about this?
```
// Reference to the Chasm component for activity execution (if applicable).
// For standalone activities, all required start information is carried within
// this component, obviating the need to use the above fields. 
```

chasm/lib/activity/activity.go
```diff
@@ -82,18 +108,120 @@ func NewEmbeddedActivity(
...
+	}
+
+	activityRef, err := chasm.DeserializeComponentRef(activityRefBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	response := &historyservice.RecordActivityTaskStartedResponse{}
+	if store == nil {
+		if err := a.PopulateRecordActivityTaskStartedResponse(ctx, activityRef.EntityKey, response); err != nil {
```
bergundy, line 166: You can just get the entity key (we renamed that to execution key) from the context. You'll need to update your branch with main though.

chasm/lib/activity/activity.go
```diff
@@ -82,18 +108,120 @@ func NewEmbeddedActivity(
...
+	if versionDirective := params.VersionDirective.GetDeploymentVersion(); versionDirective != nil {
+		attempt.LastDeploymentVersion = &deploymentpb.WorkerDeploymentVersion{
+			BuildId:        versionDirective.GetBuildId(),
+			DeploymentName: versionDirective.GetDeploymentName(),
+		}
+	}
+
 	store, err := a.Store.Get(ctx)
+	if err != nil {
+		return nil, err
```
bergundy, line 150: Wouldn't this return a not found error?

chasm/lib/activity/activity_tasks.go
bergundy: Hmm... might be something in fx that we need to work around. Forget it for now.

chasm/lib/activity/library.go
```diff
@@ -33,3 +36,13 @@ func (l *library) Components() []*chasm.RegistrableComponent {
 		chasm.NewRegistrableComponent[*Activity]("activity"),
 	}
 }
+
+func (l *library) Tasks() []*chasm.RegistrableTask {
+	return []*chasm.RegistrableTask{
+		chasm.NewRegistrableSideEffectTask[*Activity, *activitypb.ActivityDispatchTask](
+			"startActivity",
```
bergundy:
```suggestion
			"dispatchActivity",
```

chasm/lib/activity/activity.go
fretz12: Actually it just returns the store as nil. If you look inside Store.Get, it short circuits if node is nil and returns the value which is nil: 
```
	// If node is nil, then there is nothing to deserialize from, return value (even if it is also nil).
	if f.Internal.node == nil {
		if f.Internal.v == nil {
			return nilT, nil
		}
		vT, isT := f.Internal.v.(T)
		if !isT {
			return nilT, serviceerror.NewInternalf("internal value doesn't implement %s", reflect.TypeFor[T]().Name())
		}
		return vT, nil
	}
```

chasm/lib/activity/activity.go
```diff
@@ -26,14 +34,22 @@ type Activity struct {
...
-	// the workflow).
+	// Pointer to an implementation of the "store". for a workflow activity this would be a parent pointer back to
+	// the workflow. For a standalone activity this would be nil.
+	// TODO: revisit a standalone activity pointing to itself once we handle storing it more efficiently.
 	// TODO: figure out better naming.
 	Store chasm.Field[ActivityStore]
 }
 
+// RecordActivityTaskStartedParams holds parameters for RecordActivityTaskStarted method when recording activity
+// task started.
```
```suggestion
// RecordActivityTaskStartedParams holds parameters for RecordActivityTaskStarted
```

chasm/lib/activity/activity.go
```diff
@@ -47,6 +63,7 @@ func (a Activity) LifecycleState(_ chasm.Context) chasm.LifecycleState {
 	}
 }
 
+// NewStandaloneActivity creates a new activity component and launches associated tasks to start execution.
```
```suggestion
// NewStandaloneActivity creates a new activity component and adds associated tasks to start execution.
```

chasm/lib/activity/activity.go
fretz12: Cool, that'll be helpful. I put a TODO here since @dandavison and I plan to rebase on main after this PR

chasm/lib/activity/library.go
fretz12: renamed

chasm/lib/activity/activity.go
```diff
@@ -82,18 +108,120 @@ func NewEmbeddedActivity(
...
 
-func (a *Activity) PopulateRecordActivityTaskStartedResponse(ctx chasm.Context, res *historyservice.RecordActivityTaskStartedResponse) error {
+func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID string) (*matchingservice.AddActivityTaskRequest, error) {
+	// Get latest component ref and unmarshal into proto ref
+	componentRef, err := ctx.Ref(a)
+	if err != nil {
+		return nil, err
+	}
+
+	// Note: No need to set the vector clock here, as the components track version conflicts for read/write
```
dandavison, line 117: An alternative way to say this would be

```suggestion
	// No need to set vector clock; staleness checks are performed using
	// the versioned transition contained in the component ref
```

(There are various other places where we use component ref VT and not vector clock, so probably no comment needed here)

chasm/lib/activity/activity.go
```diff
@@ -82,18 +108,121 @@ func NewEmbeddedActivity(
...
+	if store == nil {
+		// TODO Get entity key from from context once we rebase on main
+		if err := a.PopulateRecordActivityTaskStartedResponse(ctx, activityRef.EntityKey, response); err != nil {
+			return nil, err
+		}
+	} else {
+		if err := store.PopulateRecordActivityTaskStartedResponse(ctx, activityRef.EntityKey, response); err != nil {
+			return nil, err
+		}
+	}
```
dandavison, line 173: WDYT about a comment explaining why store would/would not be nil?

chasm/lib/activity/proto/v1/tasks.proto
```diff
@@ -0,0 +1,10 @@
...
+    // The current attempt number for this activity execution.
+    int32 attempt = 1;
```
dandavison, line 10: Out of interest why is this one field special? Will it be used at a point where we do not hold the full component tree?

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
dandavison: Great! Although -- maybe "above" isn't accurate because `version_directive` is still relevant I believe.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -130,6 +130,9 @@ message AddActivityTaskRequest {
     temporal.server.api.taskqueue.v1.TaskForwardInfo forward_info = 11;
     int32 stamp = 12;
     temporal.api.common.v1.Priority priority = 13;
+    // Reference to the Chasm component for activity execution (if applicable). The component will have all the
+    // necessary data, deprecating the need for the separate request fields used here.
```
dandavison: (same discussion as above)

chasm/lib/activity/activity.go
fretz12: Instead of repeating ourselves at multiple places, suggest we doc the component ref about VT vs vec clock.

chasm/lib/activity/activity.go
fretz12: I put a comment on the Store Field itself and a TODO if we decide to keep standalone self referencing. I feel that should be enough so we don't have to keep repeating ourselves whenever we need this check.

chasm/lib/activity/proto/v1/tasks.proto
fretz12: Since task validation/exec happen outside of a lock, we need to guard against any concurrent operations where the originally intended task may be outdated. See the Validate check of the task. I can add a docstring here to clarify.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
fretz12: Good point, I'll rephrase `obviating the need to use the fields that apply to embedded activities with the exception of version_directive.`

fretz12: I think version_directive will actually originate from the chasm start request once we support it properly. We can revisit it then.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
fretz12: Added same doc.

PR #8542, opened 2025-10-24
add Namespace State to the metrics at partition manager creation time

majiru:
## What changed?
add Namespace State to the metrics at partition manager creation time

## Why?
This is the allow us to easier filter out out metrics from passive clusters.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

Marked as draft until more manual testing is done.

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -427,7 +427,7 @@ func (e *matchingEngineImpl) getTaskQueuePartitionManager(
 
 	tqConfig := newTaskQueueConfig(partition.TaskQueue(), e.config, namespaceEntry.Name())
 	tqConfig.loadCause = loadCause
-	logger, throttledLogger, metricsHandler := e.loggerAndMetricsForPartition(namespaceEntry, e.clusterMeta.GetCurrentClusterName(), partition, tqConfig)
+	logger, throttledLogger, metricsHandler := e.loggerAndMetricsForPartition(namespaceEntry, partition, tqConfig)
 	onFatalErr := func(cause unloadCause) { newPM.unloadFromEngine(cause) }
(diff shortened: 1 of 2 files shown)


service/matching/matching_engine.go
```diff
@@ -471,26 +470,34 @@ func (e *matchingEngineImpl) getTaskQueuePartitionManager(
 }
 
 func (e *matchingEngineImpl) loggerAndMetricsForPartition(
-	nsName namespace.Name,
+	nsEntry *namespace.Namespace,
+	clusterName string,
```
ShahabT: nit: you have access to `e` inside this func so no need to pass this.

majiru: Thanks, not sure how I missed that.

service/matching/matching_engine.go
```diff
@@ -471,26 +470,33 @@ func (e *matchingEngineImpl) getTaskQueuePartitionManager(
...
 	metricsHandler := metrics.GetPerTaskQueuePartitionIDScope(
 		e.metricsHandler,
-		nsName.String(),
+		nsName,
 		partition,
 		tqConfig.BreakdownMetricsByTaskQueue(),
 		tqConfig.BreakdownMetricsByPartition(),
 		metrics.OperationTag(metrics.MatchingTaskQueuePartitionManagerScope),
-	)
+	).WithTags(metrics.NamespaceStateTag(nsState))
```
dnr, line 499: GetPerTaskQueuePartitionIDScope accepts a varargs list of tags so you don't need to call WithTags here:
```suggestion
metrics.OperationTag(metrics.MatchingTaskQueuePartitionManagerScope),
metrics.NamespaceStateTag(nsState))
```

PR #8544, opened 2025-10-25
CHASM: implement UseExisting ID conflict policy

yycptt:
## What changed?
- Implement ID conflict policy UseExisting
- Introduce chasm.ExecutionAlreadyStartedError, same semantic as serviceerror.WorkflowExecutionAlreadyStarted.

## Why?
- UseExisting conflict will be used by components we are building.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/api/persistence/v1/chasm.pb.go
+++ b/api/persistence/v1/chasm.pb.go
@@ -213,8 +213,9 @@ func (*ChasmNodeMetadata_PointerAttributes) isChasmNodeMetadata_Attributes() {}
 
 type ChasmComponentAttributes struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
-	// Fully qualified type name of a registered component.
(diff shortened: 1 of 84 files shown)


chasm/errors.go
```diff
@@ -0,0 +1,21 @@
+package chasm
+
+type ExecutionAlreadyStartedError struct {
```
bergundy, line 3: As mentioned on our zoom meeting, this will need to be a user facing error that we will want to expose to users. That's pending some naming discussions though. Keep as is for now though until we finalize those decisions.

bergundy (approved):

yycptt:
~~Will update this PR after https://github.com/temporalio/api/pull/664 is landed.~~

Decided to keep chasm.ExecutionAlreadyStartedError and each archetype will do the conversion in api handler to its own service error definition.

PR #8547, opened 2025-10-25
CHASM: Track node size changes

yycptt:
## What changed?
- Track chasm nodes size changes in mutable state

## Why?
- The size is used for observability and ms cache eviction

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -162,7 +162,7 @@ type (
 		// Running approximate total size of mutable state fields (except buffered events) when written to DB in bytes.
 		// Buffered events are added to this value when calling GetApproximatePersistedSize.
 		approximateSize int
-		chasmNodeSizes  map[string]int // chasm node path -> size in bytes
+		chasmNodeSizes  map[string]int // chasm node path -> key + node size in bytes
 		// Total number of tomestones tracked in mutable state
 		totalTombstones int
 		// Buffer events from DB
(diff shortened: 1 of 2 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -528,7 +530,9 @@ func NewMutableStateFromDB(
 			return nil, err
 		}
 		for key, node := range dbRecord.ChasmNodes {
-			mutableState.approximateSize += len(key) + node.Size()
+			nodeSize := node.Size()
+			mutableState.approximateSize += len(key) + nodeSize
+			mutableState.chasmNodeSizes[key] = nodeSize
```
prathyushpv: wouldn't it be better if we store `len(key) + nodeSize` in `chasmNodeSizes`. That way we can use this size in `closeTransaction` below. Right now, `closeTransaction` is not subtracting key size from `ms.approximateSize`.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6445,12 +6449,19 @@ func (ms *MutableStateImpl) closeTransaction(
...
-	// TODO: update approximateSize once chasm.NodesMutation returns size change as well.
 	chasmNodesMutation, err := ms.chasmTree.CloseTransaction()
 	if err != nil {
 		return closeTransactionResult{}, err
 	}
+	for nodePath := range chasmNodesMutation.DeletedNodes {
+		ms.approximateSize -= ms.chasmNodeSizes[nodePath]
+		delete(ms.chasmNodeSizes, nodePath)
+	}
+	for nodePath, node := range chasmNodesMutation.UpdatedNodes {
```
prathyushpv, line 6799: Is newly created nodes persisted through `chasmNodesMutation.UpdatedNodes`? in that case, we will have to track the key size as well. Is it possible to get the size here?

service/history/workflow/mutable_state_impl.go
yycptt: Good point. Included key size in the map as well now.

service/history/workflow/mutable_state_impl.go
yycptt: Good catch. Fixed.

PR #8553, opened 2025-10-27
worker-versioning GA: revision number to handle async workflow inconsistencies.

Shivs11:
## What changed?
- This PR adds revision number mechanics to handle task dispatch inconsistencies that could arise since our versioning API's are becoming eventually-consistent.

## Why?
- Making our versioning API's eventually consistent.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- There are risks but they are gated by a dynamic config.

---

> [!NOTE]
> Adds revision-number semantics to task dispatch/versioning and introduces a new per-deployment TQ user-data schema, updating APIs, persistence, and server logic (DC-gated) with comprehensive tests.
> 
> - **APIs (proto changes)**:
>   - Add `revision_number` to `taskqueue.v1.TaskVersionDirective` and propagate `task_dispatch_revision_number` in History `Record*TaskStarted` requests.
>   - Matching `SyncDeploymentUserDataRequest`: new fields `deployment_name`, `update_routing_config`, `upsert_versions_data`, `forget_versions`; response includes `routing_config_changed`. Deprecate old `update_version_data`/`forget_version` oneof.
>   - Deployment API: add `WorkerDeploymentVersionData`; mark `DeploymentVersionData` deprecated where applicable.
> - **Persistence (task queue user data)**:
>   - Extend `DeploymentData` with `deployments_data` (map of deployment → `WorkerDeploymentData` holding `RoutingConfig` and per-build version data). Deprecate legacy `versions` and `unversioned_ramp_data` fields.
> - **Server logic**:
>   - Matching/History: compute target version using revision numbers; start workflow deployment transitions using revision-aware decisions; support mixed old/new schemas when calculating current/ramping.
>   - Matching: new helpers to migrate/clean old-format entries, apply routing-config updates atomically, and prevent query blackholes using status.
>   - Task dispatch carries and records revision number; internal task struct/plumbing updated.
> - **Dynamic config**:
>   - Add `system.useRevisionNumberForWorkerVersioning` (plumbed through History/Matching) to gate new behavior.
> - **Tests**:
>   - Add/expand unit and functional tests covering new schema, routing-config updates, revision-number behavior, and propagation.
> - **Misc**:
>   - Describe/Stats paths updated to read both schemas; minor build/deps update (api-go replace).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e84f4845067edbb8f65d2b7bac1f579b620db41c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/interfaces/mutable_state.go
+++ b/service/history/interfaces/mutable_state.go
@@ -286,6 +286,8 @@ type (
 
 		AddTasks(tasks ...tasks.Task)
 		PopTasks() map[tasks.Category][]tasks.Task
+		DeleteCHASMPureTasks(maxScheduledTime time.Time)
+
 		SetUpdateCondition(int64, int64)
 		GetUpdateCondition() (int64, int64)
 
--- a/api/historyservice/v1/request_response.pb.go
+++ b/api/historyservice/v1/request_response.pb.go
@@ -2605,8 +2605,10 @@ type IsActivityTaskValidRequest struct {
 	Execution        *v14.WorkflowExecution `protobuf:"bytes,2,opt,name=execution,proto3" json:"execution,omitempty"`
 	Clock            *v17.VectorClock       `protobuf:"bytes,3,opt,name=clock,proto3" json:"clock,omitempty"`
 	ScheduledEventId int64                  `protobuf:"varint,4,opt,name=scheduled_event_id,json=scheduledEventId,proto3" json:"scheduled_event_id,omitempty"`
-	unknownFields    protoimpl.UnknownFields
-	sizeCache        protoimpl.SizeCache
+	// Stamp represents the internal “version” of the activity options and can/will be changed with Activity API.
+	Stamp         int32 `protobuf:"varint,5,opt,name=stamp,proto3" json:"stamp,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *IsActivityTaskValidRequest) Reset() {
@@ -2663,6 +2665,13 @@ func (x *IsActivityTaskValidRequest) GetClock() *v17.VectorClock {
 func (x *IsActivityTaskValidRequest) GetScheduledEventId() int64 {
 	if x != nil {
 		return x.ScheduledEventId
+	}
+	return 0
+}
+
+func (x *IsActivityTaskValidRequest) GetStamp() int32 {
+	if x != nil {
+		return x.Stamp
 	}
 	return 0
 }
@@ -10087,12 +10096,13 @@ const file_temporal_server_api_historyservice_v1_request_response_proto_rawDesc 
 	"\"RespondActivityTaskCanceledRequest\x12!\n" +
 	"\fnamespace_id\x18\x01 \x01(\tR\vnamespaceId\x12j\n" +
 	"\x0ecancel_request\x18\x02 \x01(\v2C.temporal.api.workflowservice.v1.RespondActivityTaskCanceledRequestR\rcancelRequest:\x1f\x92\xc4\x03\x1b2\x19cancel_request.task_token\"%\n" +
-	"#RespondActivityTaskCanceledResponse\"\x94\x02\n" +
+	"#RespondActivityTaskCanceledResponse\"\xaa\x02\n" +
 	"\x1aIsActivityTaskValidRequest\x12!\n" +
 	"\fnamespace_id\x18\x01 \x01(\tR\vnamespaceId\x12G\n" +
 	"\texecution\x18\x02 \x01(\v2).temporal.api.common.v1.WorkflowExecutionR\texecution\x12?\n" +
 	"\x05clock\x18\x03 \x01(\v2).temporal.server.api.clock.v1.VectorClockR\x05clock\x12,\n" +
-	"\x12scheduled_event_id\x18\x04 \x01(\x03R\x10scheduledEventId:\x1b\x92\xc4\x03\x17*\x15execution.workflow_id\"8\n" +
+	"\x12scheduled_event_id\x18\x04 \x01(\x03R\x10scheduledEventId\x12\x14\n" +
+	"\x05stamp\x18\x05 \x01(\x05R\x05stamp:\x1b\x92\xc4\x03\x17*\x15execution.workflow_id\"8\n" +
 	"\x1bIsActivityTaskValidResponse\x12\x19\n" +
 	"\bis_valid\x18\x01 \x01(\bR\aisValid\"\xfb\x02\n" +
 	"\x1eSignalWorkflowExecutionRequest\x12!\n" +
--- a/chasm/engine.go
+++ b/chasm/engine.go
@@ -176,7 +176,7 @@ func UpdateWithNewEntity[C Component, I any, O1 any, O2 any](
 //   - consider remove ComponentRef from the return value and allow components to get
 //     the ref in the transition function. There are some caveats there, check the
 //     comment of the NewRef method in MutableContext.
-func UpdateComponent[C Component, R []byte | ComponentRef, I any, O any](
+func UpdateComponent[C any, R []byte | ComponentRef, I any, O any](
 	ctx context.Context,
 	r R,
 	updateFn func(C, MutableContext, I) (O, error),
@@ -207,7 +207,7 @@ func UpdateComponent[C Component, R []byte | ComponentRef, I any, O any](
 	return output, newSerializedRef, err
 }
 
-func ReadComponent[C Component, R []byte | ComponentRef, I any, O any](
+func ReadComponent[C any, R []byte | ComponentRef, I any, O any](
 	ctx context.Context,
 	r R,
 	readFn func(C, Context, I) (O, error),
@@ -234,14 +234,7 @@ func ReadComponent[C Component, R []byte | ComponentRef, I any, O any](
 	return output, err
 }
 
-type PollComponentRequest[C Component, I any, O any] struct {
-	Ref         ComponentRef
-	PredicateFn func(C, Context, I) bool
-	OperationFn func(C, MutableContext, I) (O, error)
-	Input       I
-}
-
-func PollComponent[C Component, R []byte | ComponentRef, I any, O any, T any](
+func PollComponent[C any, R []byte | ComponentRef, I any, O any, T any](
 	ctx context.Context,
 	r R,
 	predicateFn func(C, Context, I) (T, bool, error),
--- a/chasm/lib/callback/chasm_invocation.go
+++ b/chasm/lib/callback/chasm_invocation.go
@@ -59,32 +59,32 @@ func (c chasmInvocation) Invoke(
 	// Get back the base64-encoded ComponentRef from the header.
 	encodedRef, ok := c.nexus.GetHeader()[commonnexus.CallbackTokenHeader]
 	if !ok {
-		return invocationResultFail{logInternalError(e.Logger, "callback missing token", nil)}
+		return invocationResultFail{logInternalError(e.logger, "callback missing token", nil)}
 	}
 
 	decodedRef, err := base64.RawURLEncoding.DecodeString(encodedRef)
 	if err != nil {
-		return invocationResultFail{logInternalError(e.Logger, "failed to decode CHASM ComponentRef", err)}
+		return invocationResultFail{logInternalError(e.logger, "failed to decode CHASM ComponentRef", err)}
 	}
 
 	// Validate that the bytes are a valid ChasmComponentRef
 	ref := &persistencespb.ChasmComponentRef{}
 	err = proto.Unmarshal(decodedRef, ref)
 	if err != nil {
-		return invocationResultFail{logInternalError(e.Logger, "failed to unmarshal CHASM ComponentRef", err)}
+		return invocationResultFail{logInternalError(e.logger, "failed to unmarshal CHASM ComponentRef", err)}
 	}
 
 	request, err := c.getHistoryRequest(decodedRef)
 	if err != nil {
-		return invocationResultFail{logInternalError(e.Logger, "failed to build history request: %v", err)}
+		return invocationResultFail{logInternalError(e.logger, "failed to build history request: %v", err)}
 	}
 
 	// RPC to History for cross-shard completion delivery.
-	_, err = e.HistoryClient.CompleteNexusOperationChasm(ctx, request)
+	_, err = e.historyClient.CompleteNexusOperationChasm(ctx, request)
 	if err != nil {
-		msg := logInternalError(e.Logger, "failed to complete Nexus operation: %v", err)
+		msg := logInternalError(e.logger, "failed to complete Nexus operation: %v", err)
 		if isRetryableRPCResponse(err) {
-			return invocationResultRetry{err: msg, retryPolicy: e.Config.RetryPolicy()}
+			return invocationResultRetry{err: msg, retryPolicy: e.config.RetryPolicy()}
 		}
 		return invocationResultFail{msg}
 	}
--- a/chasm/lib/callback/executors.go
+++ b/chasm/lib/callback/executors.go
@@ -23,7 +23,14 @@ type HTTPCallerProvider func(queues.NamespaceIDAndDestination) HTTPCaller
 
 func NewInvocationTaskExecutor(opts InvocationTaskExecutorOptions) *InvocationTaskExecutor {
 	return &InvocationTaskExecutor{
-		InvocationTaskExecutorOptions: opts,
+		config:             opts.Config,
+		namespaceRegistry:  opts.NamespaceRegistry,
+		metricsHandler:     opts.MetricsHandler,
+		logger:             opts.Logger,
+		httpCallerProvider: opts.HTTPCallerProvider,
+		httpTraceProvider:  opts.HTTPTraceProvider,
+		historyClient:      opts.HistoryClient,
+		chasmEngine:        opts.ChasmEngine,
 	}
 }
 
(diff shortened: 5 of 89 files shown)


common/worker_versioning/worker_versioning.go
```diff
@@ -524,27 +526,73 @@ func ValidateVersioningOverride(override *workflowpb.VersioningOverride) error {
...
+	currentVersionRoutingConfig *deploymentpb.RoutingConfig,
+	rampingVersionRoutingConfig *deploymentpb.RoutingConfig,
 	workflowId string,
-) *deploymentspb.WorkerDeploymentVersion {
-	ramp := ramping.GetRampPercentage()
-	rampingVersion := ramping.GetVersion()
-	if ramp <= 0 {
+) (*deploymentspb.WorkerDeploymentVersion, int64) {
+	// Find most recent currentDeploymentVersion by comparing old vs new format timestamps
+	var finalCurrentDep *deploymentspb.WorkerDeploymentVersion
```
ShahabT, line 590: Nit: better to call new vars based on the latest terminology: `finalCurrentVersion`

common/worker_versioning/worker_versioning.go
```diff
@@ -524,27 +526,73 @@ func ValidateVersioningOverride(override *workflowpb.VersioningOverride) error {
...
+	var finalCurrentRev int64
+
+	oldCurrentTime := current.GetRoutingUpdateTime().AsTime()
+	newCurrentTime := currentVersionRoutingConfig.GetCurrentVersionChangedTime().AsTime()
+
+	if newCurrentTime.After(oldCurrentTime) || newCurrentTime.Equal(oldCurrentTime) {
```
ShahabT, line 598: Need to update the versioning Info in describe task queue to take new format into account too.

Nit: You might want to move this comparison to CalculateTaskQueueVersioningInfo so it returns nil `current` or nil `currentVersionRoutingConfig`.

common/worker_versioning/worker_versioning.go
```diff
@@ -593,7 +642,73 @@ func CalculateTaskQueueVersioningInfo(deployments *persistencespb.DeploymentData
...
+
+	if deployments.GetDeploymentsData() != nil {
+
+		for _, deploymentInfo := range deployments.GetDeploymentsData() {
+			routingConfig := deploymentInfo.GetRoutingConfig()
+			if routingConfig == nil {
+				continue
+			}
+
+			// fmt.Println("The revision number is ", routingConfig.GetRevisionNumber())
```
ShahabT: place holder comment for cleanups

common/worker_versioning/worker_versioning.go
```diff
@@ -593,7 +642,73 @@ func CalculateTaskQueueVersioningInfo(deployments *persistencespb.DeploymentData
...
+			// fmt.Println("The ramping version is ", routingConfig.GetRampingDeploymentVersion())
+
+			// Choose current/ramping based on the deployment having the most recent routingConfig update time.
+			if t := routingConfig.GetCurrentVersionChangedTime().AsTime(); t.After(routingConfigLatestCurrentVersion.GetCurrentVersionChangedTime().AsTime()) {
+				routingConfigLatestCurrentVersion = routingConfig
+			}
+
+			if t := routingConfig.GetRampingVersionChangedTime().AsTime(); t.After(routingConfigLatestRampingVersion.GetRampingVersionChangedTime().AsTime()) {
+				routingConfigLatestRampingVersion = routingConfig
+			} else if t := routingConfig.GetRampingVersionPercentageChangedTime().AsTime(); t.After(routingConfigLatestRampingVersion.GetRampingVersionPercentageChangedTime().AsTime()) {
```
ShahabT: I think we should only use percentage change time because that include the other one too:
```
    // Last time ramping version percentage was changed.
	// If ramping version is changed, this is also updated, even if the percentage stays the same.
	RampingVersionPercentageChangedTime
```

service/matching/matching_engine.go
```diff
@@ -1887,7 +1887,8 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
 	if err != nil {
 		return nil, err
 	}
-	if req.Deployment == nil && req.GetOperation() == nil {
+	// TODO (Shivam): Please fix this error message and condition check.
+	if (req.Deployment == nil && req.GetOperation() == nil) && (req.GetDeploymentName() == "" && req.GetUpdateRoutingConfig() == nil && req.GetUpsertVersionsData() == nil) {
```
ShahabT:
```suggestion
	if req.Deployment == nil && req.GetOperation() == nil && req.GetDeploymentName() == "" {
```

service/matching/matching_engine.go
```diff
@@ -1972,6 +1975,42 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
...
+
+				rc := req.GetUpdateRoutingConfig()
+				tqWorkerDeploymentData := deploymentData.GetDeploymentsData()[req.GetDeploymentName()]
+
+				applyUpdatesToDeploymentData := rc == nil
+				if rc != nil {
+					if rc.GetRevisionNumber() >= tqWorkerDeploymentData.GetRoutingConfig().GetRevisionNumber() {
+						changed = true
+						// Update routing config when newer or equal revision is provided
+						tqWorkerDeploymentData.RoutingConfig = rc
```
ShahabT: when this we sent routing config and it has higher revision number we need to set routing_config_changed=true in the response. wf uses that flag to optimize waiting for propagation.

service/matching/matching_engine.go
```diff
@@ -1972,6 +1975,42 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
...
+						applyUpdatesToDeploymentData = true
+					}
+				}
+
+				if applyUpdatesToDeploymentData {
```
ShahabT: It seems I forgot to remove these comments from the proto files: 
> Ignored if `update_routing_config` is present and has an outdated revision number.

Based on our last discussion the req.GetUpsertVersionsData() and req.GetForgetVersions() should be applied regardless of the routing config situation

service/history/interfaces/mutable_state.go
```diff
@@ -355,7 +355,9 @@ type (
 		// activities.
 		// If there is a pending workflow task that is not started yet, it'll be rescheduled after
 		// transition start.
-		StartDeploymentTransition(deployment *deploymentpb.Deployment) error
+		StartDeploymentTransition(deployment *deploymentpb.Deployment, revisionNumber int64) error
+		GetRevisionNumber() int64
```
ShahabT: Better to call these `Get/SetVersioningRevisionNumber` because in mutable state there are other concepts close to this that can make ppl confused.

service/matching/task_queue_partition_manager.go
```diff
@@ -993,87 +1025,147 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
...
+				isIndependentPinnedActivity = true
+			}
+
+			// 2. Check the new deployment data format.
+			if deploymentsData != nil && directive.GetDeploymentVersion() != nil {
+				workerDeploymentData = deploymentsData[directive.GetDeploymentVersion().GetDeploymentName()]
+				if workerDeploymentData != nil && workerDeploymentData.GetVersions() != nil && workerDeploymentData.GetVersions()[directive.GetDeploymentVersion().GetBuildId()] == nil {
+					isIndependentPinnedActivity = true
+				}
+			}
```
ShahabT: I think this should go into HasDeploymentVersion. that function should cover both new and old formats.

service/matching/task_queue_partition_manager.go
```diff
@@ -993,87 +1025,147 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
...
 			}
 			if forwardInfo == nil {
 				// Task is not forwarded, so it can be spooled if sync match fails.
 				// Spool queue and sync match queue is the same for pinned workflows.
-				return pinnedQueue, pinnedQueue, userDataChanged, nil
+				return pinnedQueue, pinnedQueue, userDataChanged, 0, nil
 			} else {
 				// Forwarded from child partition - only do sync match.
-				return nil, pinnedQueue, userDataChanged, nil
+				return nil, pinnedQueue, userDataChanged, 0, nil
```
ShahabT, line 1067: I think in these cases we still return the revision number of the activity task queue.

service/matching/task_queue_partition_manager.go
```diff
@@ -640,14 +650,27 @@ func (pm *taskQueuePartitionManagerImpl) LegacyDescribeTaskQueue(includeTaskQueu
...
-		current, ramping := worker_versioning.CalculateTaskQueueVersioningInfo(perTypeUserData.GetDeploymentData())
-		info := &taskqueuepb.TaskQueueVersioningInfo{
-			//nolint:staticcheck // SA1019: [cleanup-wv-3.1]
-			CurrentVersion:           worker_versioning.WorkerDeploymentVersionToStringV31(current.GetVersion()),
-			CurrentDeploymentVersion: worker_versioning.ExternalWorkerDeploymentVersionFromVersion(current.GetVersion()),
-			UpdateTime:               current.GetRoutingUpdateTime(),
+		current, ramping, currentVersionRoutingConfig, rampingVersionRoutingConfig := worker_versioning.CalculateTaskQueueVersioningInfo(perTypeUserData.GetDeploymentData())
+
+		var info *taskqueuepb.TaskQueueVersioningInfo
+		if current != nil {
```
ShahabT: this might not work correctly based on current implementation of CalculateTaskQueueVersioningInfo because it can return non-nil `current` while the latest current version is in `currentVersionRoutingConfig`

service/matching/task_queue_partition_manager.go
ShahabT: ok maybe we don't care. then we need to document that pinned activities do not return revision number.

tests/task_queue_stats_test.go
```diff
@@ -431,7 +433,7 @@ func (s *TaskQueueStatsSuite) validateTaskQueueStatsByType(
 	halfExpectation.MaxExtraTasks /= 2
 
 	s.validateDescribeTaskQueueWithEnhancedMode(ctx, tqName, tqType, halfExpectation)
-	s.validateDescribeWorkerDeploymentVersion(ctx, tqName, tqType, halfExpectation)
+	// s.validateDescribeWorkerDeploymentVersion(ctx, tqName, tqType, halfExpectation)
```
ShahabT: do we want to disable this?

tests/versioning_3_test.go
```diff
@@ -61,19 +61,23 @@ const (
...
+	useV32               bool
+	useNewDeploymentData bool
 }
 
 func TestVersioning3FunctionalSuite(t *testing.T) {
 	t.Parallel()
-	suite.Run(t, &Versioning3Suite{useV32: true})
-	suite.Run(t, &Versioning3Suite{useV32: false})
+	// suite.Run(t,&Versioning3Suite{useV32: true})
+	// suite.Run(t, &Versioning3Suite{useV32: false}) // May not need this since these tests are for pre-release users.
```
ShahabT: let's enable these

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -209,8 +210,17 @@ func recordActivityTaskStarted(
...
+
+		// We start a transition if one of the following conditions are met:
+		// 1. The workflow will be dispatching to the same deployment as the activity but has not yet.
+		// 2. The workflow TQ is lagging behind the activity TQ, with respect to the current version of a deployment.
+
+		// Note: We use > instead of >= because a non-backlogged activity task could have the same revision number as the MS and that should not commence a transition.
+		// Note: Revision number mechanics are only involved if the dynamic config is enabled.
+		useRevisionNumber := shardContext.GetConfig().UseRevisionNumberForWorkerVersioning(namespaceName)
+		if pollerDeployment.Equal(worker_versioning.DeploymentFromDeploymentVersion(wftDepVer)) ||
+			(useRevisionNumber && pollerDeployment.GetSeriesName() == wftDepVer.GetDeploymentName() && request.TaskDispatchRevisionNumber > wftDepRevNum) {
```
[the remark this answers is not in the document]
Shivs11: you are sadly wrong.

service/matching/matching_engine.go
```diff
@@ -1884,10 +1897,13 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
...
 ) (*matchingservice.SyncDeploymentUserDataResponse, error) {
 	taskQueueFamily, err := tqid.NewTaskQueueFamily(req.NamespaceId, req.GetTaskQueue())
+	applyUpdatesToRoutingConfig := false
+
 	if err != nil {
 		return nil, err
 	}
-	if req.Deployment == nil && req.GetOperation() == nil {
+	// TODO (Shivam): Please fix this error message and condition check.
+	if req.Deployment == nil && req.GetOperation() == nil && req.GetDeploymentName() == "nil" {
```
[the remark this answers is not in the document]
Shivs11: good, you passed the test.

common/worker_versioning/worker_versioning.go
```diff
@@ -692,18 +720,19 @@ func CalculateTaskQueueVersioningInfo(deployments *persistencespb.DeploymentData
 			}
 
 			// Choose current/ramping based on the deployment having the most recent routingConfig update time.
-			if t := routingConfig.GetCurrentVersionChangedTime().AsTime(); t.After(routingConfigLatestCurrentVersion.GetCurrentVersionChangedTime().AsTime()) &&
-				HasDeploymentVersion(deployments, DeploymentVersionFromDeployment(DeploymentFromExternalDeploymentVersion(routingConfig.GetCurrentDeploymentVersion()))) {
+			if t := routingConfig.GetCurrentVersionChangedTime().AsTime(); t.After(routingConfigLatestCurrentVersion.GetCurrentVersionChangedTime().AsTime()) {
```
Shivs11: @ShahabT  - I don't think we should have the `HasDeploymentVersion` check here. The reason I came to this conclusion was because I realized that in the case where a user uses the new deploymentData format to ramp/set current as`unversioned`, this check will always fail and we don't want that.

I propose to remove the check here since my main idea was to push this responsibility on the shoulders of the deployment manager which should check the version before syncing it in the TQ's. The only issue with this idea is that some user could, in theory, move the task-queue from the version after a sync is done.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -198,7 +198,8 @@ func recordActivityTaskStarted(
 		// The workflow transition happens only if the workflow task of the same execution would go
 		// to the poller deployment. Otherwise, it means the activity is independently versioned, we
 		// allow it to start without affecting the workflow.
-		wftDepVer, err := getDeploymentVersionForWorkflowId(ctx,
+
+		wftDepVer, wftDepRevNum, err := getDeploymentVersionAndRevisionNumberForWorkflowID(ctx,
```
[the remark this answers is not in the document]
Shivs11: this is just wrong?

common/worker_versioning/worker_versioning.go
```diff
@@ -308,6 +308,11 @@ func FindDeploymentVersion(deployments *persistencespb.DeploymentData, v *deploy
 
 //nolint:staticcheck
 func HasDeploymentVersion(deployments *persistencespb.DeploymentData, v *deploymentspb.WorkerDeploymentVersion) bool {
+	// Represents unversioned workers. Discussion: Should unversioned workers be considered to be in a worker deployment version by default?
```
ShahabT: clean up comment

service/matching/matching_engine.go
```diff
@@ -2021,6 +2021,44 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
...
+				old deployment data format under the same deployment.
+				*/
+
+				if applyUpdatesToRoutingConfig {
+					//nolint:staticcheck // SA1019 deprecated versions will clean up later
+					oldVersions := deploymentData.GetVersions()
+					dst := make([]*deploymentspb.DeploymentVersionData, 0, len(oldVersions))
+					for _, dv := range oldVersions {
+						if dv.GetVersion().GetDeploymentName() != req.GetDeploymentName() {
+							dst = append(dst, dv)
```
ShahabT: need to also add it to tqWorkerDeploymentData.Versions

service/matching/matching_engine.go
ShahabT, line 1965: first send the deployment name in the old format then check new format for the deployment name and unset the ramping version and its timestamps

ShahabT, line 1988: after doing old stuff, check the new format for the given deployment name and depending on this version being ramping or current, clean up the fields in new format routing config.

service/matching/matching_engine.go
```diff
@@ -1986,6 +1985,7 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
 					deploymentData.Versions = append(deploymentData.Versions, vd)
 				}
 			} else if v := req.GetForgetVersion(); v != nil {
+				// TODO (Shivam): Remove versions from both places.
```
ShahabT: yup need to remove from both places

service/matching/matching_engine.go
```diff
@@ -2021,6 +2021,44 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
 					delete(tqWorkerDeploymentData.Versions, buildID)
```
ShahabT: need to remove from old format too.

Also, when all versions of the deployment are removed, clean up the whole deployment data from the upper map.

service/matching/matching_engine.go
```diff
@@ -3186,3 +3211,120 @@ func (e *matchingEngineImpl) UpdateTaskQueueConfig(
...
+//
+//nolint:staticcheck // SA1019 deprecated versions will clean up later
+func migrateOldFormatVersions(
+	deploymentData *persistencespb.DeploymentData,
+	deploymentName string,
+	workerDeploymentData *persistencespb.WorkerDeploymentData,
+) {
+
+	oldVersions := deploymentData.GetVersions()
+	dst := make([]*deploymentspb.DeploymentVersionData, 0, len(oldVersions))
```
ShahabT, line 3226: should we initialize this with `deploymentData.Versions` in order to not drop possible versions already in the list?

ShahabT (approved):
feel free to merge after addressing my comment

PR #8557, opened 2025-10-28
Add Priority tag to ApproximateBacklogCount

majiru:
## What changed?
Add Priority tag to ApproximateBacklogCount

## Why?
Gives us some more visibility into the backlog count for each subqueue.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/pri_matcher.go
+++ b/service/matching/pri_matcher.go
@@ -475,7 +475,7 @@ func (tm *priTaskMatcher) poll(
 ) (*internalTask, error) {
 	start := time.Now()
 	pollWasForwarded := false
-	var priority int32 = 0
+	var priority int32
 
 	defer func() {
 		// TODO(pri): can we consolidate all the metrics code below?


service/matching/db.go
```diff
@@ -729,10 +730,10 @@ func (db *taskQueueDB) emitBacklogGaugesLocked() {
 		return
 	}
 
-	var approximateBacklogCount, totalLag int64
+	var totalLag int64
 	var oldestTime time.Time
 	for _, s := range db.subqueues {
-		approximateBacklogCount += s.ApproximateBacklogCount
+		metrics.ApproximateBacklogCount.With(db.metricsHandler).Record(float64(s.ApproximateBacklogCount), metrics.PriorityTag(locks.Priority(s.Key.Priority)))
```
dnr: use the same tag as the other matching metrics, `metrics.TaskPriorityTagName`. see other uses of that, e.g. https://github.com/temporalio/temporal/blob/main/service/matching/pri_matcher.go#L447

dnr: don't assume s.Key.Priority is unique across subqueues. s.Key is, but we may allow multiple subqueues for the same priority key in the future

service/matching/db.go
```diff
@@ -745,7 +749,13 @@ func (db *taskQueueDB) emitBacklogGaugesLocked() {
 		}
 	}
 
-	metrics.ApproximateBacklogCount.With(db.metricsHandler).Record(float64(approximateBacklogCount))
+	for _, v := range count {
+		priStr := ""
+		if v > 0 {
+			priStr = strconv.FormatInt(v, 10)
+		}
```
dnr: I think it's about time to add a helper for this, maybe in the metrics package like other tag wrappers?

Not blocking this PR though

service/matching/db.go
```diff
@@ -729,10 +730,13 @@ func (db *taskQueueDB) emitBacklogGaugesLocked() {
...
 
-	var approximateBacklogCount, totalLag int64
+	var totalLag int64
 	var oldestTime time.Time
+	count := make(map[int32]int64)
 	for _, s := range db.subqueues {
-		approximateBacklogCount += s.ApproximateBacklogCount
+		c := count[s.Key.Priority]
+		c += s.ApproximateBacklogCount
+		count[s.Key.Priority] = c
```
dnr:
```suggestion
		count[s.Key.Priority] += s.ApproximateBacklogCount
```
should work, no?

service/matching/db.go
```diff
@@ -745,7 +749,13 @@ func (db *taskQueueDB) emitBacklogGaugesLocked() {
 		}
 	}
 
-	metrics.ApproximateBacklogCount.With(db.metricsHandler).Record(float64(approximateBacklogCount))
+	for _, v := range count {
+		priStr := ""
+		if v > 0 {
+			priStr = strconv.FormatInt(v, 10)
```
dnr: the priority is the key, not the value

service/matching/db.go
```diff
@@ -745,7 +749,13 @@ func (db *taskQueueDB) emitBacklogGaugesLocked() {
...
+		}
+		metrics.ApproximateBacklogCount.With(db.metricsHandler).Record(float64(v), metrics.StringTag(metrics.TaskPriorityTagName, priStr))
+	}
 	if oldestTime.IsZero() {
 		metrics.ApproximateBacklogAgeSeconds.With(db.metricsHandler).Record(0)
```
dnr, line 753: it feels weird to have ApproximateBacklogCount be per-priority but ApproximateBacklogAgeSeconds not be. maybe we should split that one too? I'm not sure so not blocking this one

dnr: but also, if we do split this one, I think we should do it as a new metric that is the absolute time of the oldest task, not a relative time (and then deprecate this one). it's easier to deal with downstream.

service/matching/db.go
```diff
@@ -745,7 +749,13 @@ func (db *taskQueueDB) emitBacklogGaugesLocked() {
 		}
 	}
 
-	metrics.ApproximateBacklogCount.With(db.metricsHandler).Record(float64(approximateBacklogCount))
+	for _, v := range count {
+		priStr := ""
+		if v > 0 {
+			priStr = strconv.FormatInt(v, 10)
+		}
+		metrics.ApproximateBacklogCount.With(db.metricsHandler).Record(float64(v), metrics.StringTag(metrics.TaskPriorityTagName, priStr))
```
dnr: look down at line 826, we need to match this for emitZeroBacklogGauges

service/matching/db.go
majiru: Makes sense, my preference would be to do this in another PR.

service/matching/db.go
majiru: I added a wrapper for this, however we did have one wrapper for how it's used in history so I came up with a name I don't super love, happy to change it.

service/matching/db.go
majiru: Yup, I don't know why I thought that didn't work for concrete values like this...

service/matching/db.go
majiru: I think I have this fixed, emitting a zero per priority level in emitZeroBacklogGauges.

PR #8558, opened 2025-10-28
Add CHASM visibility task handling of alias to field mapping for custom search attributes

awln-temporal:
Add CHASM visibility task handling of alias to field mapping for custom search attributes. The Visibility component stores aliased custom search attributes and their corresponding payloads.

On the WRITE path, Visibility Store CURRENTLY assumes search attributes are schema field names, and does not apply any mappings from alias -> field. This is because FrontEnd "Unaliases" to its field names before sending to History. 

## What changed?
Before calling the visibility manager `UpsertWorkflowExecution`, the search attributes need to be mapped from their aliases to their field names.

## Why?
Support Visibility component compatibility with current Visibility Store implementation.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/api/historyservice/v1/request_response.pb.go
+++ b/api/historyservice/v1/request_response.pb.go
@@ -65,9 +65,11 @@ type RoutingOptions struct {
 	// Request will be routed by resolving the namespace ID and the workflow ID from this task token to a given shard.
 	TaskToken string `protobuf:"bytes,6,opt,name=task_token,json=taskToken,proto3" json:"task_token,omitempty"`
 	// Request will be routed by resolving the namespace ID and the workflow ID from the first task info element.
-	TaskInfos     string `protobuf:"bytes,7,opt,name=task_infos,json=taskInfos,proto3" json:"task_infos,omitempty"`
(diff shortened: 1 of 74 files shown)


service/history/visibility_queue_task_executor.go
```diff
@@ -389,13 +389,33 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
...
+	searchattributes := make(map[string]*commonpb.Payload)
+
+	aliasedSearchAttributes, err := visComponent.GetSearchAttributes(visTaskContext)
+	if err != nil {
+		return err
 	}
+
+	for alias, value := range aliasedSearchAttributes {
+		fieldName, err := searchAttributesMapper.GetFieldName(alias, namespaceEntry.Name().String())
+		if err != nil {
```
bergundy, line 413: I thought we said we would log these errors since search attributes can be removed and we don't want to fail the visibility task for this.

awln-temporal: yea, this should not fail the task executor, will address this in a follow up

PR #8560, opened 2025-10-28
Mutable state changes for workflow pause.

gow:
## What changed?
**Note**: This depends on https://github.com/temporalio/api/pull/653. Sending it for some early feedback.
- Made changes to `WorkflowPauseInfo` (in WorkflowExecutionInfo). Reusing unused proto fields.
- Added mutablestate.IsWorkflowExecutionPaused()
- Added mutablestate.AddWorkflowExecutionPausedEvent()

## Why?
- These changes are needed to implement pause/unpause features.

## How did you test it?
- [x] built
- [x] added new unit test(s)

---

> [!NOTE]
> Adds workflow-level pause/unpause with new `WorkflowPauseInfo` fields, client endpoints, history event handling, state/status validation for `PAUSED`, and tests; removes activity-level pause info.
> 
> - **Protocol/API**:
>   - Redefine `persistence.v1.WorkflowPauseInfo` to `{pause_time, identity, reason, request_id}`; remove `ActivityPauseInfo` and related helpers.
>   - Bump dependency `go.temporal.io/api` and adjust generated code indexes.
>   - Update RPC metadata, redirection maps, quotas, and log tags for `PauseWorkflowExecution`/`UnpauseWorkflowExecution`.
> - **Frontend Clients**:
>   - Add `PauseWorkflowExecution` and `UnpauseWorkflowExecution` to `client_impl`, `metric_client`, `retryable_client`, and mocks.
> - **History/State**:
>   - Add `WorkflowExecutionPaused` event creation (`Create/Add...PausedEvent`) with buffering rules (paused allowed to buffer; unpaused not buffered).
>   - Implement `ApplyWorkflowExecutionPausedEvent`: set status `PAUSED`, populate `executionInfo.PauseInfo`, invalidate pending activities and workflow task via stamps.
>   - Rebuilder applies paused event; state transition validation supports `PAUSED` across CREATED/RUNNING/ZOMBIE.
> - **Validation/Tests**:
>   - Allow `WORKFLOW_EXECUTION_STATUS_PAUSED` in validators; expand unit tests for create/update state/status and paused behavior.
> - **Tooling**:
>   - `buf.yaml` breaking ignore for `executions.proto`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 951c9c4a53473a319afc5e2d62ac3cc1c85f3770. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -2859,9 +2859,8 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionPausedEvent(event *historypb.H
 	// Invalidate pending workflow task by incrementing the persisted stamp.
 	// This ensures subsequent task dispatch detects the change.
 	if ms.HasPendingWorkflowTask() {
-		workflowTask := ms.GetPendingWorkflowTask()
-		workflowTask.Stamp = workflowTask.Stamp + 1
-		ms.workflowTaskManager.UpdateWorkflowTask(workflowTask)
+		ms.executionInfo.WorkflowTaskStamp += 1
+		ms.workflowTaskManager.UpdateWorkflowTask(ms.GetPendingWorkflowTask())
 	}
 	return nil
 }
--- a/service/history/workflow/workflow_task_state_machine.go
+++ b/service/history/workflow/workflow_task_state_machine.go
@@ -1025,7 +1025,6 @@ func (m *workflowTaskStateMachine) UpdateWorkflowTask(
 	m.ms.executionInfo.WorkflowTaskHistorySizeBytes = workflowTask.HistorySizeBytes
 	m.ms.executionInfo.WorkflowTaskBuildId = workflowTask.BuildId
 	m.ms.executionInfo.WorkflowTaskBuildIdRedirectCounter = workflowTask.BuildIdRedirectCounter
-	m.ms.executionInfo.WorkflowTaskStamp = workflowTask.Stamp
 
 	m.ms.workflowTaskUpdated = true
 


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -831,23 +831,15 @@ message ResetChildInfo {
...
+    // The identity of the actor that paused the workflow.
+    string identity = 2;
 
-    // The identity of the actor that paused the activity.
-    string identity = 3;
+    // The reason for pausing the workflow.
+    string reason = 3;
 
-    // The ID of the request that paused the activity. Duplicates pause requests with the same ID will succeed.
+    // A unique identifier for this pause request (for idempotency checks)
```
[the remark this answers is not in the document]
gow: None of these fields were ever used. So it's safe to reuse them.

common/persistence/workflow_state_status_validator.go
```diff
@@ -62,7 +63,7 @@ func ValidateUpdateWorkflowStateStatus(
 
 	// validate workflow state & status
 	if (state == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && status == enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING) ||
-		(state != enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING) {
+		(state != enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && (status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING && status != enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED)) {
```
spkane31: I forget where our discussions landed, Is this only for the first iteration or are we guaranteeing this for the future?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2214,6 +2214,10 @@ func (ms *MutableStateImpl) IsWorkflowExecutionRunning() bool {
 	}
 }
 
+func (ms *MutableStateImpl) IsWorkflowExecutionPaused() bool {
```
spkane31: Do we want to update the `IsWorkflowExecutionRunning` definition too? Are we saying that a paused WF is still running, there's some ambiguity there.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2783,6 +2787,57 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 	return nil
 }
 
+func (ms *MutableStateImpl) PauseWorkflowExecution(
+	identity string,
+	reason string,
+	requestId string,
+) error {
+	ms.executionState.Status = enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED
```
yycptt: plz use ms.UpdateWorkflowStateStatus, and update sanity checks around workflow status changes.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2783,6 +2787,57 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+	// Mark the event as 'worker may ignore' so that older SDKs can safely ignore it.
+	event.WorkerMayIgnore = true
+	if err := ms.applyWorkflowExecutionPausedEvent(event); err != nil {
+		return err
+	}
+	return nil
+}
+
+// applyWorkflowExecutionPausedEvent applies the paused event to the mutable state. It updates the workflow execution status to paused and sets the pause info.
+func (ms *MutableStateImpl) applyWorkflowExecutionPausedEvent(event *historypb.HistoryEvent) error {
```
yycptt: need to update mutable_state_rebuilder.go as well.

yycptt: do we plan to reapply this event on conflict resolution?

service/history/workflow/mutable_state_impl.go
gow: Yes. I see the ambiguity. The problem is 
- IsWorkflowExecutionRunning(): defines the execution 'state' based on the state `ms.executionState.State` (not the same as status)
- IsWorkflowExecutionPaused(): is defining the execution 'status' based on the status (`ms.executionState.Status`).

I'll rename this to `IsWorkflowExecutionStatusPaused` to add more clarity.

service/history/workflow/mutable_state_impl.go
gow: Done.

service/history/workflow/mutable_state_impl.go
gow: Yes. We do. Updated mutable_state_rebuilder.go.
I think we don't need to do anything special to handle these events during reset operation (since reset points cannot be in between pause & unpause events)

service/history/workflow/mutable_state_impl.go
```diff
@@ -2794,6 +2798,55 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+	}
+
+	// Invalidate pending workflow task. This will cause a
+	if ms.HasPendingWorkflowTask() {
+		pendingWorkflowTask := ms.GetPendingWorkflowTask()
+		pendingWorkflowTask.Stamp = pendingWorkflowTask.Stamp + 1
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
gow: PauseWorkflowExecution() call was left over in a test by mistake. Fixed it.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2794,6 +2798,51 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+		ai.Stamp = ai.Stamp + 1
+	}
+
+	// Invalidate pending workflow task by incrementing the persisted stamp.
+	// This ensures subsequent task dispatch detects the change.
+	if ms.HasPendingWorkflowTask() {
+		ms.executionInfo.WorkflowTaskStamp += 1
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
gow: Interesting. I'm not sure if this is strictly required. There would be max limits to reason, identity & request_id (validated in the pause API).
Will bump the approximateSize here to stick to the convention.

common/persistence/workflow_state_status_validator.go
```diff
@@ -40,9 +41,15 @@ func ValidateCreateWorkflowStateStatus(
...
-	if (state == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && status == enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING) ||
-		(state != enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING) {
-		return serviceerror.NewInternalf("Create workflow with invalid state: %v or status: %v", state, status)
+	if state == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED {
+		if status == enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING {
+			return serviceerror.NewInternalf("Create workflow with invalid state: %v or status: %v", state, status)
+		}
+	} else {
+		if status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING &&
+			status != enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED {
```
yycptt: do we ever create new execution in paused state?

service/history/historybuilder/event_store.go
```diff
@@ -306,6 +306,13 @@ func (b *EventStore) bufferEvent(
 		enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_COMPLETED:
 		return false
 
+	// Unpaused event should not be buffered since it transitions the workflow from paused to running status.
+	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_UNPAUSED:
+		return false
+	// A paused workflow event *should be* allowed to be buffered since we want to accept any inflight workflow task completion.
+	case enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_PAUSED:
+		return true
```
yycptt, line 314: why not fail existing workflow task, flush buffer events and then add the paused event (in the same batch).

service/history/interfaces/mutable_state.go
```diff
@@ -212,6 +212,9 @@ type (
 		TaskQueueScheduleToStartTimeout(name string) (*taskqueuepb.TaskQueue, *durationpb.Duration)
 
 		IsWorkflowExecutionRunning() bool
+		IsWorkflowExecutionStatusPaused() bool
```
yycptt: it can also be done with `ms.GetExecutionState().Status == enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED`, new method isn't really necessary.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2801,6 +2798,54 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+	reason string,
+	requestID string,
+) (*historypb.HistoryEvent, error) {
+	opTag := tag.WorkflowActionWorkflowPaused
+	if err := ms.checkMutability(opTag); err != nil {
+		return nil, err
+	}
+	event := ms.hBuilder.AddWorkflowExecutionPausedEvent(identity, reason, requestID)
+	// Mark the event as 'worker may ignore' so that older SDKs can safely ignore it.
+	event.WorkerMayIgnore = true
```
yycptt: shall we make it part it hBuilder.AddWorkflowExecutionPausedEvent?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2801,6 +2798,54 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 	return nil
 }
 
+func (ms *MutableStateImpl) AddWorkflowExecutionPausedEvent(
```
yycptt, line 2816: do we need to first check is workflow is already in paused state or not? or that will be done in the api handler?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2801,6 +2798,54 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+	// Invalidate all the pending activities. Do not mark individual activities as paused.
+	for _, ai := range ms.GetPendingActivityInfos() {
+		ai.Stamp = ai.Stamp + 1
+	}
+
+	// Invalidate pending workflow task by incrementing the persisted stamp.
+	// This ensures subsequent task dispatch detects the change.
+	if ms.HasPendingWorkflowTask() {
+		ms.executionInfo.WorkflowTaskStamp += 1
+	}
```
yycptt, line 2864: hmm we can't change those fields directly, there are lots of other related fields that need to be updated together, or they won't be persisted/replicated.
For activities, use for example `ms.UpdateActivity()`. For workflow task, you could use  `workflowTaskManager.UpdateWorkflowTask` or just set `ms.workflowTaskUpdated` to true.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2801,6 +2798,54 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+func (ms *MutableStateImpl) ApplyWorkflowExecutionPausedEvent(event *historypb.HistoryEvent) error {
+	// Update workflow status.
+	if _, err := ms.UpdateWorkflowStateStatus(ms.executionState.GetState(), enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED); err != nil {
+		return err
+	}
+	// Set pause info in mutable state.
+	ms.executionInfo.PauseInfo = &persistencespb.WorkflowPauseInfo{
+		PauseTime: timestamppb.New(event.GetEventTime().AsTime()),
+		Identity:  event.GetWorkflowExecutionPausedEventAttributes().GetIdentity(),
+		Reason:    event.GetWorkflowExecutionPausedEventAttributes().GetReason(),
```
yycptt, line 2842: those two fields are basically for returning the information as part of describe workflow execution response? Wonder if they need to be in mutable state given they are already stored in the paused event.

service/history/workflow/mutable_state_state_status.go
```diff
@@ -24,12 +24,12 @@ func setStateStatus(
 	case enumsspb.WORKFLOW_EXECUTION_STATE_CREATED:
 		switch state {
 		case enumsspb.WORKFLOW_EXECUTION_STATE_CREATED:
-			if status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING {
+			if status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING && status != enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED {
```
yycptt: same Q re. if a new execution can be in paused state.

service/history/workflow/mutable_state_impl.go
yycptt, line 2816: Support for unpausing will be done in a follow up PR?

common/persistence/workflow_state_status_validator.go
gow: Currently no. And there are no plans to allow it. Will update it.

service/history/historybuilder/event_store.go
gow: We made a decision to allow in-flight workflow tasks to complete (mainly to support local activities). So this became necessary.

service/history/interfaces/mutable_state.go
gow: There are some follow-up PRs where I'm using this. They appear inside a if statement with multiple conditionals. So it comes handy. I'll remove it from here and add it in the PR where i'm using it so that it becomes clear.

service/history/workflow/mutable_state_impl.go
gow: Yes - https://github.com/temporalio/temporal/pull/8605

gow: Yes. Just to keep the PRs smaller and easy to review.

service/history/workflow/mutable_state_impl.go
gow: No objection to that. Only reason I put it here was for visibility. If this detail is buried deep inside the builder it may be difficult to discover.
But in the new world of vibe coding, discoverability might not really be an issue :)

service/history/workflow/mutable_state_impl.go
gow: Yes. Those fields would be returned in describe APIs.
This is mainly for optimization. Since history can continue to grow when the workflow is paused, there may be a lot of events accumulated after the paused event. If one describes such a workflow, we would have to iterate through events.
There would be validations and limits on these fields (enforced in API). So caching them in mutable state seemed ok.

service/history/workflow/mutable_state_impl.go
gow: will update.

service/history/workflow/mutable_state_state_status.go
gow: No. Will update it.

service/history/workflow/mutable_state_impl.go
yycptt: > If one describes such a workflow, we would have to iterate through events

sorry why we need to iterate through events?

gow: Because there is no index to the paused event. Maybe I can just store the current paused event index and load it up in describe?

common/persistence/workflow_state_status_validator.go
gow: Status changes are guaranteed for the future.

service/history/workflow/mutable_state_impl.go
gow: updated.

common/persistence/workflow_state_status_validator.go
gow: done

service/history/workflow/mutable_state_impl.go
gow: moved it.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -1022,6 +1022,7 @@ func (m *workflowTaskStateMachine) UpdateWorkflowTask(
 	m.ms.executionInfo.WorkflowTaskHistorySizeBytes = workflowTask.HistorySizeBytes
 	m.ms.executionInfo.WorkflowTaskBuildId = workflowTask.BuildId
 	m.ms.executionInfo.WorkflowTaskBuildIdRedirectCounter = workflowTask.BuildIdRedirectCounter
+	m.ms.executionInfo.WorkflowTaskStamp = workflowTask.Stamp
```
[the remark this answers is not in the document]
gow: @stephanos why do we have [this comment](https://github.com/temporalio/temporal/blob/d3e4c8125d495a8c165401ceb93b1591c582e34b/service/history/workflow/workflow_task_state_machine.go#L1030)?
A workflow stamp changes not just when we reschedule a workflow task but when we want to invalidate a workflow task (without rescheduling it). Could you please review this part of the code and let me know if updating the stamp results in something bad? If not I can fix the comment.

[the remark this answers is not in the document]
stephanos: When I wrote the comment I assumed that updating the stamp is not a valid use case as AFAICT from my use case the stamp should never be updated; instead a new task with a new stamp should be generated. Let me read more about the PR to understand why that would not be true.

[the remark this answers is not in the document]
gow: That assumption would not be true in the cases where we just want to invalidate the current scheduled task without generating a new one.

In my case, when a workflow is paused I want to invalidate the current task. I will reschedule a new task when the workflow is unpaused later.

service/history/workflow/mutable_state_impl.go
yycptt: yeah that's what in mine mind. you'll need both the eventID and eventBatchID to retrieve the event it. I don't have strong opinion on this though given it will be empty for most workflows.

gow: There is also the complexity of buffered event. We won't have the event ID generated at the time of pause. So I'd have to record it at the time of flushing the event.
Probably it's simpler the way it is currently.

service/history/workflow/workflow_task_state_machine.go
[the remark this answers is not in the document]
stephanos: Gotcha! In that case feel free to remove the comment 👍

common/persistence/workflow_state_status_validator.go
```diff
@@ -40,9 +41,14 @@ func ValidateCreateWorkflowStateStatus(
 	}
 
 	// validate workflow state & status
-	if (state == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && status == enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING) ||
-		(state != enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED && status != enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING) {
-		return serviceerror.NewInternalf("Create workflow with invalid state: %v or status: %v", state, status)
+	if state == enumsspb.WORKFLOW_EXECUTION_STATE_COMPLETED {
+		if status == enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING || status == enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED {
```
yycptt, line 45: looks like we don't need this either since workflow won't be created as paused.

yycptt (approved):

common/persistence/workflow_state_status_validator.go
gow: Yes. This check ensures that it remains like that in the future. So, I'll leave it here.

yycptt: ah ok, makes sense. I mis-read it before.

PR #8564, opened 2025-10-29
Introduce a chasm entity for worker

rkannan82:
## What changed?
Added a new chasm component to track lifecycle of a worker and its state.

Key files:
- worker.go: Chasm component.
- statemachine.go: State transition logic.
- tasks.go: Task executors
- proto/v1/message.proto: Worker state definition.

See README.md for details.

Overview:
State: Consists of run-time configuration, activities that are bound to it, etc.

Lifecycle: A worker refers to a worker process and can be in one of 3 states: Active, Inactive, CleanedUp. 
- Active->Active: When a worker is running, it will periodically send a heartbeat to the server to renew its lease and keep it alive.
- Active->Inactive: When the server stops receiving a heartbeat for some time period, it will transition it to Inactive state. It will also request the history server to reschedule activities that were bound to this worker. The rationale is that most likely the worker process has died.
- Inactive->Active: But it is possible that the worker is alive, but due to network partition, it was not able to renew its lease. In this case, when the the worker eventually connects to the server, the server will transition it back into Active state.
- Inactive->CleanedUp: After some pre-defined grace period, the server will transition the worker from inactive to CleanedUp state. This is a terminal state. If the server receives a heartbeat in this state, it will be rejected and treated as an error.

## Why?
To support worker liveness and activity cancellation.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None. Feature is not yet integrated.


chasm/lib/workersession/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+	return TransitionHeartbeatReceived.Apply(ctx, ws, EventHeartbeatReceived{
+		Time:          time.Now(),
+		LeaseDeadline: leaseDeadline,
+	})
+}
+
+// scheduleLeaseExpiry schedules a timer task that will fire when the lease expires.
+func scheduleLeaseExpiry(ctx chasm.MutableContext, ws *WorkerSession, leaseDeadline time.Time) {
+	expiryTask := &workersessionpb.LeaseExpiryTask{
+		LeaseExpirationTime: timestamppb.New(leaseDeadline),
```
bergundy: No need to record this on the task, it's already on the attributes.

chasm/lib/workersession/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+// EventHeartbeatReceived is triggered when a heartbeat is received from the worker.
+type EventHeartbeatReceived struct {
+	Time          time.Time
+	LeaseDeadline time.Time
+}
+
+// TransitionHeartbeatReceived handles heartbeat reception, extending the lease and scheduling a new timeout.
+var TransitionHeartbeatReceived = chasm.NewTransition(
+	[]workersessionpb.WorkerSessionStatus{workersessionpb.WORKER_SESSION_STATUS_ACTIVE},
+	workersessionpb.WORKER_SESSION_STATUS_ACTIVE,
```
bergundy: I wonder if we could transition back from expired to active but what you have now is fine IMHO.

chasm/lib/workersession/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+
+// EventCleanupCompleted is triggered when cleanup is finished.
+type EventCleanupCompleted struct {
+	Time time.Time
+}
+
+// TransitionCleanupCompleted handles cleanup completion, marking session as terminated.
+var TransitionCleanupCompleted = chasm.NewTransition(
+	[]workersessionpb.WorkerSessionStatus{workersessionpb.WORKER_SESSION_STATUS_EXPIRED},
+	workersessionpb.WORKER_SESSION_STATUS_TERMINATED,
```
bergundy: Probably call this CLEANED_UP or CLOSED, terminated has a different meaning in Temporal.

chasm/lib/workersession/proto/v1/message.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
+
+package temporal.server.chasm.lib.workersession.proto.v1;
+
+option go_package = "go.temporal.io/server/chasm/lib/workersession/gen/workersessionpb;workersessionpb";
+
+import "google/protobuf/timestamp.proto";
+
+// Metadata about the worker session.
+message WorkerSessionState {
```
bergundy: You're going to want to keep track of all of the activities here so you can notify all of their state machines when they time out and for the activity timer to be able to verify that the worker confirmed ownership of the activity.

chasm/lib/workersession/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"go.temporal.io/server/chasm"
+	workersessionpb "go.temporal.io/server/chasm/lib/workersession/gen/workersessionpb/v1"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	// Time to wait before cleaning up an expired session.
+	expiredSessionCleanupDelay = 10 * time.Minute
```
bergundy: This should be a dynamic config

chasm/lib/workersession/statemachine.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// RecordHeartbeat processes a heartbeat by applying the heartbeat received transition.
+func RecordHeartbeat(ctx chasm.MutableContext, ws *WorkerSession, leaseDeadline time.Time) error {
```
bergundy: Here's where you're going to want to get a diff of all of the activities from the worker.

chasm/lib/workersession/component.go
```diff
@@ -0,0 +1,48 @@
+package workersession
+
+import (
+	"go.temporal.io/server/chasm"
+	workersessionpb "go.temporal.io/server/chasm/lib/workersession/gen/workersessionpb/v1"
+)
+
+const (
+	Archetype chasm.Archetype = "WorkerSession"
```
bergundy: Wonder if we want to call this WorkerSession or just Worker.

bergundy (commented):
This looks great!

chasm/lib/workersession/proto/v1/message.proto
rkannan82: Ack. Will be adding the activity related code in a separate PR.

chasm/lib/workersession/statemachine.go
rkannan82: Ack. Will be adding the activity related code in a separate PR.

chasm/lib/workersession/component.go
rkannan82: Renamed to Worker. No need to introduce a new terminology.

chasm/lib/workersession/statemachine.go
rkannan82: Duh!

chasm/lib/workersession/statemachine.go
rkannan82: Changed to CLEANED_UP. Also renamed expired to INACTIVE to be consistent with active state.

chasm/lib/worker/statemachine.go
```diff
@@ -0,0 +1,88 @@
...
+	LeaseDeadline time.Time
+}
+
+// TransitionHeartbeatReceived handles heartbeat reception, extending the lease and scheduling a new timeout.
+var TransitionHeartbeatReceived = chasm.NewTransition(
+	[]workerpb.WorkerStatus{workerpb.WORKER_STATUS_ACTIVE},
+	workerpb.WORKER_STATUS_ACTIVE,
+	func(ctx chasm.MutableContext, w *Worker, event EventHeartbeatReceived) error {
+		// Store client-provided lease deadline.
+		w.LeaseExpirationTime = timestamppb.New(event.LeaseDeadline)
```
yiminc: how does client config the heartbeat timeout? should server control that interval?

chasm/lib/worker/statemachine.go
```diff
@@ -0,0 +1,88 @@
...
+
+// EventHeartbeatReceived is triggered when a heartbeat is received from the worker.
+type EventHeartbeatReceived struct {
+	Time          time.Time
+	LeaseDeadline time.Time
+}
+
+// TransitionHeartbeatReceived handles heartbeat reception, extending the lease and scheduling a new timeout.
+var TransitionHeartbeatReceived = chasm.NewTransition(
+	[]workerpb.WorkerStatus{workerpb.WORKER_STATUS_ACTIVE},
```
yiminc: should we allow resurrect an inactive worker if a new heartbeat is received?

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,113 @@
...
+	// If no lease deadline set, lease expiry is not valid.
+	if component.LeaseExpirationTime == nil {
+		return false
+	}
+
+	// Timer is valid if its deadline is >= the current lease deadline.
+	// (i.e., it hasn't been superseded by a newer heartbeat).
+	taskDeadline := attrs.ScheduledTime
+	componentDeadline := component.LeaseExpirationTime.AsTime()
+	return !taskDeadline.Before(componentDeadline)
```
yiminc: nit: it would be easer to read if use componentDeadline.Before(taskDeadline)

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,113 @@
...
+	component *Worker,
+	attrs chasm.TaskAttributes,
+	task *workerpb.WorkerCleanupTask,
+) error {
+	// Only clean up if worker is inactive.
+	if component.Status != workerpb.WORKER_STATUS_INACTIVE {
+		return nil // Not inactive, nothing to clean up.
+	}
+
+	e.logger.Info("Cleaning up inactive worker")
```
yiminc: would this include worker identity info?

chasm/lib/worker/statemachine.go
yiminc: I think we have to handle this case, for situation where worker is not able to connect for sometime and reconnect after that.

chasm/lib/workersession/statemachine.go
rkannan82: After some thought, changing from active to inactive is better. This way SDK does not have to handle it. The only case that SDK needs to handle is inactive -> cleaned up state. The grace period should be large enough to avoid this scenario.

chasm/lib/workersession/statemachine.go
rkannan82: Done.

chasm/lib/worker/statemachine.go
rkannan82: Yes, client sends the lease duration, and server computes it. Just that this happens in the handler code (child PR). The chasm component itself takes the deadline.

chasm/lib/worker/statemachine.go
rkannan82: After some thought, changing from active to inactive is better. This way SDK does not have to handle it. The only case that SDK needs to handle is inactive -> cleaned up state. The grace period should be large enough to avoid this scenario.

chasm/lib/worker/tasks.go
rkannan82: Added a workerid tag and used it for logging.

rkannan82: Added a workerid tag and used it for logging.

chasm/lib/worker/statemachine.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// updateWorkerLease is a shared helper that updates the worker's lease and schedules expiry.
+func updateWorkerLease(ctx chasm.MutableContext, w *Worker, leaseDeadline time.Time) {
+	w.LeaseExpirationTime = timestamppb.New(leaseDeadline)
+	scheduleLeaseExpiry(ctx, w, leaseDeadline)
+}
+
+// scheduleLeaseExpiry schedules a timer task that will fire when the lease expires.
+func scheduleLeaseExpiry(ctx chasm.MutableContext, w *Worker, leaseDeadline time.Time) {
```
bergundy: nit: these can be method of the Worker struct.

chasm/lib/worker/library.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (l *Library) Tasks() []*chasm.RegistrableTask {
+	return []*chasm.RegistrableTask{
+		chasm.NewRegistrablePureTask(
+			"LeaseExpiryTask",
+			l.leaseExpiryTaskExecutor,
+			l.leaseExpiryTaskExecutor,
+		),
+		chasm.NewRegistrablePureTask(
+			"WorkerCleanupTask",
```
bergundy: No need to stutter and I see that we are using camel case for the other task names.

```suggestion
			"expiry",
			l.leaseExpiryTaskExecutor,
			l.leaseExpiryTaskExecutor,
		),
		chasm.NewRegistrablePureTask(
			"cleanup",
```

chasm/lib/worker/README.md
bergundy, line 1: Nice touch adding a README to your library.

chasm/lib/worker/statemachine.go
```diff
@@ -0,0 +1,91 @@
...
+		ScheduledTime: leaseDeadline,
+		Destination:   "", // Empty means local execution.
+	}
+
+	ctx.AddTask(w, taskAttrs, expiryTask)
+}
+
+// EventHeartbeatReceived is triggered when a heartbeat is received from the worker.
+type EventHeartbeatReceived struct {
+	Time          time.Time
```
bergundy: Don't put time on the events, use `ctx.Now(w)` to get the time from the framework. This is critical to allow us to implement time skipping in the framework later.

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Execute is called when a lease expiry timer fires.
+func (e *LeaseExpiryTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	worker *Worker,
+	attrs chasm.TaskAttributes,
+	task *workerstatepb.LeaseExpiryTask,
+) error {
+	// Validate that this lease expiry is still relevant.
+	if !e.isLeaseExpiryTaskValid(worker, attrs) {
```
bergundy: Don't need to validate a pure task, validation should happen under the same lock.

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,121 @@
...
+	worker *Worker,
+	attrs chasm.TaskAttributes,
+) bool {
+	// If worker is not active, no point in processing the least expiry task.
+	// A previous lease expiry must have already transitioned it to inactive.
+	if worker.Status != workerstatepb.WORKER_STATUS_ACTIVE {
+		return false
+	}
+
+	if worker.LeaseExpirationTime == nil {
```
bergundy: Can this ever happen? If it can please document, otherwise remove the condition.

chasm/lib/worker/library.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func (l *Library) Name() string {
+	return "worker"
+}
+
+func (l *Library) Components() []*chasm.RegistrableComponent {
+	return []*chasm.RegistrableComponent{
+		chasm.NewRegistrableComponent[*Worker]("Worker"),
```
bergundy: I see you defined an archetype name and did not use it here.

chasm/lib/worker/statemachine.go
```diff
@@ -0,0 +1,91 @@
...
+	scheduleLeaseExpiry(ctx, w, leaseDeadline)
+}
+
+// scheduleLeaseExpiry schedules a timer task that will fire when the lease expires.
+func scheduleLeaseExpiry(ctx chasm.MutableContext, w *Worker, leaseDeadline time.Time) {
+	expiryTask := &workerstatepb.LeaseExpiryTask{}
+
+	taskAttrs := chasm.TaskAttributes{
+		ScheduledTime: leaseDeadline,
+		Destination:   "", // Empty means local execution.
```
bergundy: nit: this is redundant.

chasm/lib/worker/README.md
```diff
@@ -0,0 +1,126 @@
...
+- **Lease**: Period for which a worker should be considered alive
+- **Heartbeat**: Periodic signal from worker to extend its lease
+- **Activity Binding**: Association between workers and their assigned activities
+
+## Worker Lifecycle
+
+### States
+
+```
+ACTIVE ──────► INACTIVE ──────► CLEANED_UP
```
bergundy: This will be auto generated for you with @dandavison's tool BTW.

chasm/lib/worker/executors.go
```diff
@@ -0,0 +1,121 @@
...
+	ctx chasm.Context,
+	worker *Worker,
+	attrs chasm.TaskAttributes,
+	task *workerstatepb.LeaseExpiryTask,
+) (bool, error) {
+	return e.isLeaseExpiryTaskValid(worker, attrs), nil
+}
+
+// isLeaseExpiryTaskValid checks if this lease expiry task is valid or if the lease has been renewed.
+func (e *LeaseExpiryTaskExecutor) isLeaseExpiryTaskValid(
```
bergundy: nit: this can be a worker method.

chasm/lib/worker/worker.go
```diff
@@ -0,0 +1,87 @@
...
+// WorkerID returns the unique identifier for this worker.
+func (w *Worker) WorkerID() string {
+	if w.WorkerHeartbeat == nil {
+		return ""
+	}
+	return w.WorkerHeartbeat.WorkerInstanceKey
+}
+
+// RecordHeartbeat processes a heartbeat, updating worker state and extending the lease.
+func (w *Worker) RecordHeartbeat(ctx chasm.MutableContext, heartbeat *workerpb.WorkerHeartbeat, leaseDuration time.Duration) error {
```
bergundy: No need to export these methods as they will not be called from outside of the library.

```suggestion
func (w *Worker) workerID() string {
	if w.WorkerHeartbeat == nil {
		return ""
	}
	return w.WorkerHeartbeat.WorkerInstanceKey
}

// RecordHeartbeat processes a heartbeat, updating worker state and extending the lease.
func (w *Worker) recordHeartbeat(ctx chasm.MutableContext, heartbeat *workerpb.WorkerHeartbeat, leaseDuration time.Duration) error {
```

chasm/lib/worker/worker.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Apply appropriate state transition based on current status
+	switch w.Status {
+	case workerstatepb.WORKER_STATUS_ACTIVE:
+		return TransitionActiveHeartbeat.Apply(ctx, w, EventHeartbeatReceived{
+			Time:          time.Now(),
+			LeaseDeadline: leaseDeadline,
+		})
+	case workerstatepb.WORKER_STATUS_INACTIVE:
+		// Handle worker resurrection after network partition
```
bergundy: This isn't just due to network partition, it can happen when a worker is overloaded, e.g. has large GC freeze. Basically we don't know why...

chasm/lib/worker/statemachine.go
```diff
@@ -0,0 +1,91 @@
...
+	workerstatepb.WORKER_STATUS_CLEANED_UP,
+	func(w *Worker, ctx chasm.MutableContext, event EventCleanupCompleted) error {
+		return nil
+	},
+)
+
+// TransitionWorkerResurrection handles worker reconnection when in INACTIVE state.
+// This is a special case for when the same worker process reconnects after network partition.
+// Note: Any activities associated with this worker may have already been rescheduled.
+var TransitionWorkerResurrection = chasm.NewTransition(
```
bergundy: nit: You can merge this transition with `TransitionActiveHeartbeat` if you want, which may save you some lines of code but what you have is also fine.

bergundy: nit: This would be called `TransitionResurrected` to be consistent with the other transition names.

chasm/lib/worker/worker.go
```diff
@@ -0,0 +1,87 @@
...
+type Worker struct {
+	chasm.UnimplementedComponent
+
+	// Persisted state.
+	*workerstatepb.WorkerState
+}
+
+// NewWorker creates a new Worker component with ACTIVE status.
+func NewWorker() *Worker {
+	return &Worker{
```
bergundy, line 45: You will immediately want to invoke the "active" transition to schedule a timer task here (or this can happen in the RPC handler).

chasm/lib/worker/worker.go
```diff
@@ -0,0 +1,87 @@
...
+func (w *Worker) SetStateMachineState(status workerstatepb.WorkerStatus) {
+	w.Status = status
+}
+
+// WorkerID returns the unique identifier for this worker.
+func (w *Worker) WorkerID() string {
+	if w.WorkerHeartbeat == nil {
+		return ""
+	}
+	return w.WorkerHeartbeat.WorkerInstanceKey
```
bergundy: Use getters for protos:

```suggestion
	return w.GetWorkerHeartbeat().GetWorkerInstanceKey()
```

chasm/lib/worker/worker.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// RecordHeartbeat processes a heartbeat, updating worker state and extending the lease.
+func (w *Worker) RecordHeartbeat(ctx chasm.MutableContext, heartbeat *workerpb.WorkerHeartbeat, leaseDuration time.Duration) error {
+	w.WorkerHeartbeat = heartbeat
+
+	// Calculate lease deadline
+	leaseDeadline := time.Now().Add(leaseDuration)
```
bergundy: Pass the duration into the state machine transition or use `ctx.Now(w)` instead or `time.Now()`.

common/log/tag/tags.go
```diff
@@ -301,6 +301,11 @@ func WorkerBuildId(buildId string) ZapTag {
 	return NewStringTag("worker-build-id", buildId)
 }
 
+// WorkerID returns tag for a worker
```
bergundy: AFAIC, you can put this log tag in the library. I don't think we are going to use it elsewhere.

common/dynamicconfig/constants.go
```diff
@@ -459,6 +459,11 @@ Deleted Redirect Rules will be kept in the DB (with DeleteTimestamp). After this
 		`PollerHistoryTTL is the time to live for poller histories in the pollerHistory cache of a physical task queue. Poller histories are fetched when
 		requiring a list of pollers that polled a given task queue.`,
 	)
+	InactiveWorkerCleanupDelay = NewNamespaceDurationSetting(
```
bergundy: Put the library's dynamic configs in the library.

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Validate checks if cleanup is still needed.
+func (e *WorkerCleanupTaskExecutor) Validate(
+	ctx chasm.Context,
+	worker *Worker,
+	attrs chasm.TaskAttributes,
+	task *workerstatepb.WorkerCleanupTask,
+) (bool, error) {
+	// Only valid if worker is inactive.
+	return worker.Status == workerstatepb.WORKER_STATUS_INACTIVE, nil
```
bergundy: You're going to want to avoid multiple cleanup tasks from being considered valid. They could be if you're resurrecting a worker. The simplest way of achieving this is to store the time the statemachine considers itself inactive and compare that to the task attributes schedule time.

chasm/lib/worker/executors.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func NewWorkerCleanupTaskExecutor(logger log.Logger) *WorkerCleanupTaskExecutor {
+	return &WorkerCleanupTaskExecutor{
+		logger: logger,
+	}
+}
+
+// Execute is called to clean up inactive workers.
+func (e *WorkerCleanupTaskExecutor) Execute(
```
bergundy: This will not be a pure task when you have to make RPCs into all of the activity state machines to notify them of completion.

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	if worker.LeaseExpirationTime == nil {
+		return false
+	}
+
+	scheduledLeaseExpirationTime := attrs.ScheduledTime
+	workerLeaseExpirationTime := worker.LeaseExpirationTime.AsTime()
+	// If the lease has been renewed past the scheduled expiry time, this task is no longer valid.
+	return !workerLeaseExpirationTime.After(scheduledLeaseExpirationTime)
```
bergundy: AFAIU the timestamps should match for this task to be considered valid.

chasm/lib/worker/tasks.go
```diff
@@ -0,0 +1,121 @@
...
+	if worker.Status != workerstatepb.WORKER_STATUS_ACTIVE {
+		return false
+	}
+
+	if worker.LeaseExpirationTime == nil {
+		return false
+	}
+
+	scheduledLeaseExpirationTime := attrs.ScheduledTime
+	workerLeaseExpirationTime := worker.LeaseExpirationTime.AsTime()
```
bergundy:
```suggestion
	workerLeaseExpirationTime := worker.GetLeaseExpirationTime().AsTime()
```

chasm/lib/worker/tasks.go
rkannan82: Yes, it should match since the timer and state are updated atomically with same value. So I changed it to equal.
At that time, it seemed logical to do the > check as I was looking at this code in isolation.

chasm/lib/worker/statemachine.go
rkannan82: I removed time fields from other events also. My intent was for future bookkeeping. But realized that chasm can record this automatically when a transition is created or executed.

chasm/lib/worker/executors.go
rkannan82: WorkerCleanupTask just transitions the state to cleaned_up (terminal state). This is my thinking. When lease expires, the LeaseExpiryTaskExecutor will do 2 things:
1. Add RescheduleActivityTask: This will not be a pure task as it notifies all activities.
2. Add WorkerCleanupTask: This is a pure task that just changes the state to cleaned_up.

What do you think?

chasm/lib/worker/statemachine.go
rkannan82: So I was debating this ... Problem is cyclic dependency (only logical and not a real golang issue).
- worker and tasks.go use statemachine to create state transitions.
- If we move scheduleLeaseExpiry to worker, then statemachine will depend on worker.

chasm/lib/worker/tasks.go
rkannan82: Ahha, good catch! Yea, we need to do something similar to how we handle lease expiry.

chasm/lib/worker/worker.go
rkannan82: Done.

chasm/lib/worker/statemachine.go
rkannan82: Renamed for now.

chasm/lib/worker/worker.go
rkannan82: Ack. It is being done from RPC handler in this [PR](https://github.com/temporalio/temporal/pull/8603)

chasm/lib/worker/library.go
rkannan82: Done.

chasm/lib/worker/tasks.go
rkannan82: It will be a bug if this happens. I added an error log. Should I panic instead of returning false?

chasm/lib/worker/README.md
rkannan82: Ack.

common/log/tag/tags.go
rkannan82: Thought I would centralize the tag related utils.

chasm/lib/worker/tasks.go
rkannan82: Ended up using equals (based on Roey's comment).

chasm/lib/worker/config.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"go.temporal.io/server/common/dynamicconfig"
+)
+
+var (
+	// InactiveWorkerCleanupDelay is the time to wait before cleaning up inactive workers after their lease expires.
+	InactiveWorkerCleanupDelay = dynamicconfig.NewNamespaceDurationSetting(
+		"worker.inactiveWorkerCleanupDelay",
+		60*time.Minute,
+		`InactiveWorkerCleanupDelay is the time to wait before cleaning up inactive workers after their lease expires.`,
```
bergundy: nit:

```suggestion
		"worker.inactiveCleanupDelay",
		60*time.Minute,
		`Time to wait before cleaning up inactive workers after their lease expires.`,
```

chasm/lib/worker/executors.go
```diff
@@ -0,0 +1,125 @@
...
+// Execute is called when a lease expiry timer fires.
+func (e *LeaseExpiryTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	worker *Worker,
+	attrs chasm.TaskAttributes,
+	task *workerstatepb.LeaseExpiryTask,
+) error {
+	// Calculate cleanup delay from dynamic config.
+	namespaceID := ctx.ExecutionKey().NamespaceID
+	cleanupDelay := e.config.InactiveWorkerCleanupDelay(namespaceID)
```
bergundy: You're going to need the namespace name here, not the ID. This will require injecting the namespace registry into the executor and resolving the ID to a name.

chasm/lib/worker/statemachine.go
bergundy: It already depends on the worker by referencing it here.

chasm/lib/worker/executors_test.go
bergundy, line 1: There seems to be some duplication between the tests here and `statemachine_test.go`. It would be best to keep transition logic tests in `statemachine_test.go`.

common/log/tag/tags.go
bergundy: Yeah, I want to break this up as we are doing for configs, metrics, etc...

chasm/lib/worker/tasks.go
bergundy: Log is fine but you can also just ignore it and let proto getters remove the need to have nil checks.

chasm/lib/worker/statemachine.go
bergundy: That's fine, but keep in mind that you are adding redundant status checks in `worker.recordHeartbeat` due to this separation.

chasm/lib/worker/executors.go
bergundy: The cleanup task will need to notify all of the activity state machines that the worker went away. No need to create extra tasks if you already have an executor. You want to minimize persistence writes if possible. We will get to that in a follow up PR though. Just keep in mind you are going to have some refactoring here.

chasm/lib/worker/worker.go
```diff
@@ -0,0 +1,85 @@
...
+			LeaseDeadline: leaseDeadline,
+		})
+	case workerstatepb.WORKER_STATUS_INACTIVE:
+		// Handle worker resurrection (example network partition, overloaded worker, etc.)
+		return TransitionResurrected.Apply(ctx, w, EventHeartbeatReceived{
+			LeaseDeadline: leaseDeadline,
+		})
+	default:
+		// CLEANED_UP or other states - not allowed
+		return fmt.Errorf("cannot record heartbeat for worker in state %v", w.Status)
```
bergundy: This should be a failed precondition error and not an internal error that will end up as an SLA violation.

PR #8570, opened 2025-10-30
Async routing config propagation in deployment workflow

ShahabT:
## What changed?
- The following APIs now propagate Worker Deployment Routing Config to the Task Queues in an async manner. This means that the API will return if the update is valid but the routing behavior fully adjusts in an eventually consistent manner.
  - `SetCurrentWorkerDeploymentVersion`
  - `SetRampingWorkerDeploymentVersion`
  - `DeleteDeploymentVersion`
  
- `DescribeWorkerDeployment` now returns a `RoutingConfigUpdateState` enum to inform the user if the propagation is fully done. (usable for writing unflaky test) 
 
- RoutingConfig now has `revision_number` which tracks changes in the config. All task queues now receive the routing info and the revision number. (used for handling versioning race conditions)

- 

## Why?
The Routing Config info propagation to the task queues and their partitions can take mentioned APIs can take long when there are many task queues (and/or partitions) which could cause to client side timeouts that would be hard to trouble shoot for users. Also, the Worker Deployment would be locked during the update procedure in the old logic which would make all other APIs non-responsive.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
This is a breaking change to the API behavior in the sense that the affected APIs are now eventually consistent and may return before all the task queue receive the changes. This should be OK for practical use but if some one is writing tests for their deployment system they'd need to wait on the `RoutingConfigUpdateState` to become `COMPLETED` before proceeding with the test.

### DIFF
--- a/tests/worker_deployment_test.go
+++ b/tests/worker_deployment_test.go
@@ -3156,19 +3156,21 @@ func (s *WorkerDeploymentSuite) setAndVerifyRampingVersionUnversionedOption(
 	}
 	s.NoError(err)
 
-	if prevVersion := expectedResp.GetPreviousDeploymentVersion(); prevVersion != nil {
-		s.Equal(prevVersion.GetBuildId(), resp.GetPreviousDeploymentVersion().GetBuildId())
-		s.Equal(prevVersion.GetDeploymentName(), resp.GetPreviousDeploymentVersion().GetDeploymentName())
-	} else {
-		// nolint:staticcheck // SA1019: version v0.31
-		if expectedResp.GetPreviousVersion() == "" {
-			s.Nil(resp.GetPreviousDeploymentVersion())
+	if expectedResp != nil {
+		if prevVersion := expectedResp.GetPreviousDeploymentVersion(); prevVersion != nil {
+			s.Equal(prevVersion.GetBuildId(), resp.GetPreviousDeploymentVersion().GetBuildId())
+			s.Equal(prevVersion.GetDeploymentName(), resp.GetPreviousDeploymentVersion().GetDeploymentName())
 		} else {
 			// nolint:staticcheck // SA1019: version v0.31
-			s.Equal(expectedResp.GetPreviousVersion(), worker_versioning.ExternalWorkerDeploymentVersionToStringV31(resp.GetPreviousDeploymentVersion()))
+			if expectedResp.GetPreviousVersion() == "" {
+				s.Nil(resp.GetPreviousDeploymentVersion())
+			} else {
+				// nolint:staticcheck // SA1019: version v0.31
+				s.Equal(expectedResp.GetPreviousVersion(), worker_versioning.ExternalWorkerDeploymentVersionToStringV31(resp.GetPreviousDeploymentVersion()))
+			}
 		}
+		s.InDelta(expectedResp.GetPreviousPercentage(), resp.GetPreviousPercentage(), 0.01)
 	}
-	s.InDelta(expectedResp.GetPreviousPercentage(), resp.GetPreviousPercentage(), 0.01)
 }
 
 func (s *WorkerDeploymentSuite) setCurrentVersion(ctx context.Context, tv *testvars.TestVars, previousCurrent string, ignoreMissingTaskQueues bool, expectedError string) {


service/worker/workerdeployment/version_workflow.go
```diff
@@ -516,21 +519,33 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
-	// sync version information to all the task queues
-	err = d.syncVersionDataToTaskQueues(ctx, versionData)
-	if err != nil {
-		// TODO (Shivam): Compensation functions required to roll back the local state + activity changes.
-		return nil, err
+	if isAsyncMode {
+		// ASYNC MODE: propagate full routing config
+		err = d.syncRoutingConfigToTaskQueues(ctx, args.RoutingConfig, newStatus)
+		if err != nil {
+			// TODO (Shivam): Compensation functions required to roll back the local state + activity changes.
```
Shivs11: I think we can remove this comment now since we are not going down the path of compensating functions

service/worker/workerdeployment/fx.go
```diff
@@ -16,6 +16,18 @@ import (
 	"go.uber.org/fx"
 )
 
+type DeploymentWorkflowVersion int64
+
+const (
+	// Versions of workflow logic. When introducing a new version, consider generating a new
+	// history for TestReplays using generate_history.sh.
+
+	// Represents the state before DeploymentWorkflowVersion is introduced
```
Shivs11: nit: I think we should just say: Represents the state before the versioning API's received the option of becoming async in nature

service/worker/workerdeployment/workflow.go
```diff
@@ -47,6 +47,10 @@ type (
 		stateChanged  bool
 		signalHandler *SignalHandler
 		forceCAN      bool
+		// Tracks the version of the deployment manger when a particular run of a workflow starts base on the dynamic config of the
```
Shivs11: nit: typo in "manager"

service/worker/workerdeployment/workflow.go
```diff
@@ -559,12 +649,19 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
...
 		}
 		if !d.rampingVersionStringUnversioned(newRampingVersion) {
 			if _, err := d.syncVersion(ctx, newRampingVersion, setRampUpdateArgs, true); err != nil {
 				return nil, err
 			}
+		} else if asyncMode {
+			// Updates to unversioned ramp should go through the current version
+			if _, err := d.syncVersion(ctx, pendingRoutingConfig.CurrentVersion, setRampUpdateArgs, true); err != nil {
+				return nil, err
+			}
```
Shivs11: maybe we can make the comment better since this is not dealing with unversioned ramps only

service/worker/workerdeployment/version_workflow.go
```diff
@@ -808,3 +823,134 @@ func (d *VersionWorkflowRunner) syncVersionDataToTaskQueues(ctx workflow.Context
...
+			Version: state.GetVersion(),
+			Sync:    batch,
+		}).Get(ctx, &syncRes)
+
+		if err != nil {
+			d.logger.Error("async propagation batch failed", "error", err)
+			// Continue with other batches
+			continue
+		}
+
```
Shivs11: nit: Maybe we should move this to a different function altogether since this is not really tracking the propagation and is actually executing it

service/worker/workerdeployment/workflow.go
```diff
@@ -829,6 +930,33 @@ func (d *WorkflowRunner) handleSetCurrent(ctx workflow.Context, args *deployment
...
+	if newCurrentVersion == d.State.RoutingConfig.RampingVersion {
+		pendingRoutingConfig.RampingVersion = ""
+		pendingRoutingConfig.RampingDeploymentVersion = nil
+		pendingRoutingConfig.RampingVersionPercentage = 0
+		pendingRoutingConfig.RampingVersionChangedTime = updateTime
+		pendingRoutingConfig.RampingVersionPercentageChangedTime = updateTime
+	}
+
 	// TODO (Shivam): remove the empty string check once canary stops flaking out
 	if newCurrentVersion != worker_versioning.UnversionedVersionId && newCurrentVersion != "" {
```
Shivs11: should we not have `async` mode as an option when trying to set a new version to a current?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -808,3 +823,134 @@ func (d *VersionWorkflowRunner) syncVersionDataToTaskQueues(ctx workflow.Context
...
+// signalPropagationComplete sends a signal to the deployment workflow when async propagation completes
+func (d *VersionWorkflowRunner) signalPropagationComplete(ctx workflow.Context, revisionNumber int64) {
+	err := workflow.SignalExternalWorkflow(
+		ctx,
+		worker_versioning.GenerateDeploymentWorkflowID(d.VersionState.Version.DeploymentName),
+		"",
+		PropagationCompleteSignal,
+		&deploymentspb.PropagationCompletionInfo{
+			RevisionNumber: revisionNumber,
+			BuildId:        d.VersionState.Version.BuildId,
```
Shivs11: nit (maybe not nit): In theory, I know what I am about to say is conservative and super-safe programming we should have a .GetVersion().GetBuildID() call here rather than prevent any nil dereferences which could happen during nightlies/long-hauls

Shivs11:
I did a first-pass and the overall async logic looks nice to me - I shall do another pass when we add in some new tests so that we can see the whole workflow (no pun intended) happen from start to finish

service/matching/matching_engine.go
```diff
@@ -3097,3 +3211,122 @@ func (e *matchingEngineImpl) UpdateTaskQueueConfig(
...
+		rc.RampingVersionPercentageChangedTime = nil
+		rc.RampingVersionChangedTime = nil
+	}
+
+	// Check if current role changed. If it did, clear current from RC.
+	oldCurrent := oldVd.GetCurrentSinceTime() != nil
+	newCurrent := newVd.GetCurrentSinceTime() != nil
+	if oldCurrent != newCurrent || newCurrent {
+		//nolint:staticcheck // SA1019
+		rc.CurrentVersion = ""
```
Shivs11, line 3326: I am assuming this caused some breakages in some older test? If so, sorry I wrote this function.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -209,12 +211,13 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
...
-				d.drainageStatusSyncInProgress == false &&
-				d.asyncPropagationsInProgress == 0 && // Don't CaN while async propagations are in progress
-				(d.forceCAN || d.stateChanged))
+		return !d.propagatingDelete &&
+			(d.deleteVersion || // version is deleted -> it's ok to drop all signals and updates.
+				// There is no pending signal or update, but the state is dirty or forceCaN is requested:
+				(!d.signalHandler.signalSelector.HasPending() && d.signalHandler.processingSignals == 0 && workflow.AllHandlersFinished(ctx) &&
+					!d.drainageStatusSyncInProgress &&
+					d.asyncPropagationsInProgress == 0 && // Don't CaN while async propagations are in progress
+					(d.forceCAN || d.stateChanged)))
```
Shivs11: (not blaming you since I worked on this for the longest before handing this over to you) - this seems to be kinda convoluted - we should def come back to this to clean it up.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -209,12 +211,13 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
 	// Wait until we can continue as new or are cancelled. The workflow will continue-as-new iff
 	// there are no pending updates/signals and the state has changed.
 	err := workflow.Await(ctx, func() bool {
-		return d.deleteVersion || // version is deleted -> it's ok to drop all signals and updates.
-			// There is no pending signal or update, but the state is dirty or forceCaN is requested:
-			(!d.signalHandler.signalSelector.HasPending() && d.signalHandler.processingSignals == 0 && workflow.AllHandlersFinished(ctx) &&
-				d.drainageStatusSyncInProgress == false &&
-				d.asyncPropagationsInProgress == 0 && // Don't CaN while async propagations are in progress
-				(d.forceCAN || d.stateChanged))
+		return !d.propagatingDelete &&
```
Shivs11: When commencing an async delete operation, why not just increment the value of d.asyncPropagationInProgress? Why make this new variable?

Can't we just track all propagations using the same variable?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -430,6 +451,16 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
...
+	// In case this version just got deleted, we wait until it finished propagating delete to all task queue before reviving it
+	err = workflow.Await(ctx, func() bool {
+		return !d.propagatingDelete
+	})
+	if err != nil {
+		return err
+	}
+	// In case it was marked as deleted we make it undeleted
+	d.deleteVersion = false
+
```
Shivs11: So in theory, this could happen one world:

- User deletes this version. 
- Deletion is taking place in an async manner. It is taking sometime since the version had 1000 task-queues.
- Now, the same user registers the worker in this version.
- The user then queries the version state by doing a describe call.

IIUC, we should then be in a situation where these calls are backlogged and the query shall only go through after the register successfully sets delete version to be true, right?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -546,47 +588,77 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
 		}
 
 		// Clear drainage information, if present, when a version gets activated.
 		// This handles the rollback scenario where a previously draining/drained version
-		// is reactivated and should have it's drainage information cleared.
-
-		v := workflow.GetVersion(ctx, "clear-drainage-on-activation", workflow.DefaultVersion, 0)
-		if v != workflow.DefaultVersion {
-			d.VersionState.DrainageInfo = nil
-		}
```
Shivs11, line 563: I think we want this no? This may break backwards compatibility for folks started on pre-release?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -864,56 +956,52 @@ func (d *VersionWorkflowRunner) syncRoutingConfigToTaskQueues(
...
+	completedBatches := 0
+	totalBatches := len(batches)
+	resultChannel := workflow.NewChannel(ctx)
 
-		err := workflow.ExecuteActivity(activityCtx, d.a.SyncDeploymentVersionUserData, &deploymentspb.SyncDeploymentVersionUserDataRequest{
-			Version: state.GetVersion(),
-			Sync:    batch,
-		}).Get(ctx, &syncRes)
+	for _, batch := range batches {
+		batch := batch // capture loop variable
```
Shivs11: nit: I think we can remove this loop variable here - go versions above 1.22 should allow it

service/worker/workerdeployment/version_workflow.go
```diff
@@ -934,7 +1022,43 @@ func (d *VersionWorkflowRunner) trackAsyncPropagation(
...
+	// Only check propagation for task queues where routing config actually changed
+	// Task queues with max_version = -1 indicate no routing change, skip those
+	result := make(map[string]int64)
+	for _, tqName := range workflow.DeterministicKeys(syncRes.TaskQueueMaxVersions) {
+		maxVersion := syncRes.TaskQueueMaxVersions[tqName]
+		if maxVersion != -1 {
+			result[tqName] = maxVersion
+		}
+	}
+	return result
```
Shivs11: From matching, we are returning a boolean which denotes if the routingConfig changed or not. Why not use that in our internal protos? I think it could make things simpler and remove this -1 check too

wdyt?

service/matching/matching_engine.go
ShahabT: I think this one cause an actual issue but since still some part of the logic is based on the old strings I thought I should keep it consistent until we clean up old stuff completely.

service/worker/workerdeployment/version_workflow.go
ShahabT: hmm... I think we can lm change it.

service/worker/workerdeployment/version_workflow.go
ShahabT: correct. I have a test for the scenario Test_DeleteVersion_AsyncPropagation_BlocksWorkerRegistration

service/worker/workerdeployment/version_workflow.go
ShahabT: Break in what sense?

If you mean break replay it shouldn't happen because this info (d.VersionState.DrainageInfo) is read only and doesn't affect the workflow flow (history events).

If you mean DescribeVersion API will return different output then I think we want to fix it anyways.

BTW, I think it's safe to break V3 prerelease users because there probably is none! Not the same for V1-2

service/worker/workerdeployment/version_workflow.go
ShahabT: In `SyncDeploymentVersionUserDataResponse` we have a map from name to number it felt best to just use that rather than replace it with a new map from name to pair of number and bool.

service/worker/workerdeployment/version_workflow.go
Shivs11: Never mind - I just realized why.

ShahabT (commented):
comments

service/worker/workerdeployment/version_workflow.go
```diff
@@ -234,6 +233,11 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
 }
 
 func (d *VersionWorkflowRunner) handleUpdateVersionMetadata(ctx workflow.Context, args *deploymentspb.UpdateVersionMetadataArgs) (*deploymentspb.UpdateVersionMetadataResponse, error) {
+	if d.deleteVersion {
+		// Deployment workflow should not call this function if version is marked for deletion, but still checking for safety.
+		return nil, temporal.NewNonRetryableApplicationError(errVersionDeleted, errVersionDeleted, nil)
+	}
+
```
Shivs11, line 248: maybe this can go in the validator for this update handler since other updates are currently following that paradigm

Shivs11 (approved):

PR #8571, opened 2025-10-30
Add Sometimes assertions

stephanos:
## What changed?

WISOTT

## Why?

Sometimes assertions help our internal testing platform to identify interesting test cases.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks

Performance overhead in production; but it should be minimal as the assertions are not placed on the hot path but on error cases.

### DIFF
--- a/service/history/queues/executable.go
+++ b/service/history/queues/executable.go
@@ -555,7 +555,7 @@ func (e *executableImpl) HandleErr(err error) (retErr error) {
 		tag.NewStringTag("task-category", e.GetCategory().Name()),
 	)
 	if attempt > taskCriticalLogMetricAttempts {
-		softassert.Sometimes(logger).Error("Critical error processing task, retrying.")
+		softassert.Sometimes(logger).Error("Critical error processing task, retrying.", tag.OperationCritical)
 	} else {
 		softassert.Sometimes(logger).Warn("Fail to process task")
 	}


common/softassert/softassert.go
```diff
@@ -44,27 +48,66 @@ func Fail(logger log.Logger, staticMessage string, tags ...tag.Tag) {
...
 // Dynamic information should be passed via `tags`.
+func (s *SometimesLogger) Debug(staticMessage string, tags ...tag.Tag) {
+	s.logger.Debug(staticMessage, tags...)
+}
+
+// Info logs a message at info level.
 //
-// Example:
-// softassert.ThatSometimes(logger, object.state == "terminated", "termination event", tag.NewStringTag("state", object.state))
-func ThatSometimes(logger log.Logger, condition bool, staticMessage string, tags ...tag.Tag) bool {
```
stephanos, line 54: Remove for now until use case presents itself.

common/softassert/softassert.go
```diff
@@ -44,27 +48,66 @@ func Fail(logger log.Logger, staticMessage string, tags ...tag.Tag) {
 	logger.Error("failed assertion: "+staticMessage, append([]tag.Tag{tag.FailedAssertion}, tags...)...)
 }
 
-// ThatSometimes is used to conditionally log a debug message of a noteworthy but non-problematic event.
+// Sometimes is used to log a message of a noteworthy but non-problematic event.
+//
+// Example:
+// softassert.Sometimes(logger).Warn("termination event", tag.NewStringTag("state", object.state))
+func Sometimes(logger log.Logger) *sometimesLogger {
```
stephanos, line 55: Changed the API to allow replacing existing warn/error/info logs seemlessly.

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -129,6 +130,12 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 				metrics.StaleMutableStateCounter.With(handler.metricsHandler).Record(
 					1,
 					metrics.OperationTag(metrics.HistoryRespondWorkflowTaskCompletedScope))
+				softassert.Sometimes(handler.logger).Debug("stale mutable state detected",
+					tag.WorkflowID(token.GetWorkflowId()),
```
spkane31, line 134: I noticed in logs I added that when I added the `tag.WorkflowID` it ended up duplicated, do you know if we have a way to de-dupe those?

service/history/queues/executable.go
```diff
@@ -538,11 +552,12 @@ func (e *executableImpl) HandleErr(err error) (retErr error) {
 		tag.Attempt(int32(attempt)),
 		tag.UnexpectedErrorAttempts(int32(e.unexpectedErrorAttempts)),
 		tag.LifeCycleProcessingFailed,
+		tag.NewStringTag("task-category", e.GetCategory().Name()),
 	)
 	if attempt > taskCriticalLogMetricAttempts {
-		logger.Error("Critical error processing task, retrying.", tag.OperationCritical)
+		softassert.Sometimes(logger).Error("Critical error processing task, retrying.")
```
spkane31: should we keep the tag?

stephanos: Good catch; I'll add it back 👍

service/history/api/respondworkflowtaskcompleted/api.go
stephanos: I've never noticed/had that issue. I don't know; I guess under the hood it's an array then.

Here it should be fine since the other existing calls are also adding `tag.WorkflowRunID` one by one.

spkane31: More of an annoyance than anything, certainly not blocking this PR

PR #8573, opened 2025-10-30
Add timeout tasks for Standalone Activities

fretz12:
## What changed?
Added timeout tasks for standalone activities. Refactored existing backoff algorithm to common. Added outcome field. Fixed a few errors as the result of main rebase. Regen go.sum. Fixed broken redirection tests. Uncommented `LifecycleStateTimedout` and added supporting logic in `closeTransactionHandleRootLifecycleChange` 

## Why?
We need schedule-to-start, schedule-to-close, and start-to-close timeout support for standalone activities. Also reusing existing activity retry backoff code, so refactored it to a common package. We need to bring back LifecycleStateTimedout
because when a timeout lifecycle occurs the workflow status needs to update to `WORKFLOW_EXECUTION_STATUS_TIMED_OUT`, else we encounter an invalidate state change: `unable to change workflow state from Created to Completed, status Failed`

## How did you test it?
- [X] built
- [ ] added new unit test(s)

## Potential risks
E2E timeout tests not added yet as we need complete workflow API plumbed before we can test this.

### DIFF
--- a/chasm/lib/activity/statemachine_test.go
+++ b/chasm/lib/activity/statemachine_test.go
@@ -9,6 +9,7 @@ import (
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
@@ -68,8 +69,11 @@ func TestTransitionScheduled(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			ctx := &chasm.MockMutableContext{}
-			ctx.HandleNow = func(chasm.Component) time.Time { return defaultTime }
+			ctx := &chasm.MockMutableContext{
+				MockContext: chasm.MockContext{
+					HandleNow: func(chasm.Component) time.Time { return defaultTime },
+				},
+			}
 			attemptState := &activitypb.ActivityAttemptState{Count: tc.startingAttemptCount}
 			outcome := &activitypb.ActivityOutcome{}
 
@@ -95,81 +99,21 @@ func TestTransitionScheduled(t *testing.T) {
 			for i, expectedTask := range tc.expectedTasks {
 				actualTask := ctx.Tasks[i]
 
+				require.IsType(t, expectedTask.Payload, actualTask.Payload, "expected %T at index %d, got %T",
+					expectedTask.Payload, i, actualTask.Payload)
+
 				switch expectedTask.Payload.(type) {
 				case *activitypb.ActivityDispatchTask:
-					_, ok := actualTask.Payload.(*activitypb.ActivityDispatchTask)
-					require.True(t, ok, "expected ActivityDispatchTask at index %d", i)
 					require.Empty(t, actualTask.Attributes.ScheduledTime)
 				case *activitypb.ScheduleToStartTimeoutTask:
-					_, ok := actualTask.Payload.(*activitypb.ScheduleToStartTimeoutTask)
-					require.True(t, ok, "expected ScheduleToStartTimeoutTask at index %d", i)
 					require.Equal(t, defaultTime.Add(tc.scheduleToStartTimeout), actualTask.Attributes.ScheduledTime)
 				case *activitypb.ScheduleToCloseTimeoutTask:
-					_, ok := actualTask.Payload.(*activitypb.ScheduleToCloseTimeoutTask)
-					require.True(t, ok, "expected ScheduleToCloseTimeoutTask at index %d", i)
 					require.Equal(t, defaultTime.Add(tc.scheduleToCloseTimeout), actualTask.Attributes.ScheduledTime)
 				default:
 					t.Fatalf("unexpected task payload type at index %d: %T", i, actualTask.Payload)
 				}
 
 			}
-		})
-	}
-}
-
-func TestTransitionScheduledFromInvalidStatus(t *testing.T) {
-	testCases := []struct {
-		name           string
-		startingStatus activitypb.ActivityExecutionStatus
-	}{
-		{
-			name:           "from scheduled status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED,
-		},
-		{
-			name:           "from started status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_STARTED,
-		},
-		{
-			name:           "from cancel requested status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
-		},
-		{
-			name:           "from completed status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_COMPLETED,
-		},
-		{
-			name:           "from cancelled status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED,
-		},
-		{
-			name:           "from terminated status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TERMINATED,
-		},
-		{
-			name:           "from timed out status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT,
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			ctx := &chasm.MockMutableContext{}
-
-			activity := &Activity{
-				ActivityState: &activitypb.ActivityState{
-					RetryPolicy:            defaultRetryPolicy,
-					ScheduleToCloseTimeout: durationpb.New(defaultScheduleToCloseTimeout),
-					ScheduleToStartTimeout: durationpb.New(defaultScheduleToStartTimeout),
-					StartToCloseTimeout:    durationpb.New(defaultStartToCloseTimeout),
-					Status:                 tc.startingStatus,
-				},
-				Attempt: chasm.NewDataField(ctx, &activitypb.ActivityAttemptState{Count: 0}),
-				Outcome: chasm.NewDataField(ctx, &activitypb.ActivityOutcome{}),
-			}
-
-			err := TransitionScheduled.Apply(activity, ctx, nil)
-			require.Error(t, err)
 		})
 	}
 }
@@ -240,7 +184,10 @@ func TestTransitionRescheduled(t *testing.T) {
 			require.NoError(t, err)
 			require.Equal(t, activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED, activity.Status)
 			require.Equal(t, tc.startingAttemptCount+1, attemptState.Count)
-			require.Equal(t, durationpb.New(tc.expectedRetryInterval), attemptState.GetCurrentRetryInterval())
+			require.True(t, proto.Equal(
+				durationpb.New(tc.expectedRetryInterval),
+				attemptState.GetCurrentRetryInterval(),
+			), "retry intervals should be equal")
 
 			// Verify attempt state failure details updated correctly
 			lastFailureDetails := attemptState.GetLastFailureDetails()
@@ -268,63 +215,6 @@ func TestTransitionRescheduled(t *testing.T) {
 				}
 
 			}
-		})
-	}
-}
-
-func TestTransitionRescheduledFromInvalidStatus(t *testing.T) {
-	testCases := []struct {
-		name           string
-		startingStatus activitypb.ActivityExecutionStatus
-	}{
-		{
-			name:           "from unspecified status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_UNSPECIFIED,
-		},
-		{
-			name:           "from scheduled status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED,
-		},
-		{
-			name:           "from cancel requested status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
-		},
-		{
-			name:           "from completed status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_COMPLETED,
-		},
-		{
-			name:           "from cancelled status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED,
-		},
-		{
-			name:           "from terminated status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TERMINATED,
-		},
-		{
-			name:           "from timed out status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT,
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			ctx := &chasm.MockMutableContext{}
-
-			activity := &Activity{
-				ActivityState: &activitypb.ActivityState{
-					RetryPolicy:            defaultRetryPolicy,
-					ScheduleToCloseTimeout: durationpb.New(defaultScheduleToCloseTimeout),
-					ScheduleToStartTimeout: durationpb.New(defaultScheduleToStartTimeout),
-					StartToCloseTimeout:    durationpb.New(defaultStartToCloseTimeout),
-					Status:                 tc.startingStatus,
-				},
-				Attempt: chasm.NewDataField(ctx, &activitypb.ActivityAttemptState{Count: 1}),
-				Outcome: chasm.NewDataField(ctx, &activitypb.ActivityOutcome{}),
-			}
-
-			err := TransitionRescheduled.Apply(activity, ctx, nil)
-			require.Error(t, err)
 		})
 	}
 }
@@ -357,63 +247,6 @@ func TestTransitionStarted(t *testing.T) {
 	_, ok := ctx.Tasks[0].Payload.(*activitypb.StartToCloseTimeoutTask)
 	require.True(t, ok, "expected ScheduleToStartTimeoutTask")
 	require.Equal(t, defaultTime.Add(defaultStartToCloseTimeout), ctx.Tasks[0].Attributes.ScheduledTime)
-}
-
-func TestTransitionStartedFromInvalidStatus(t *testing.T) {
-	testCases := []struct {
-		name           string
-		startingStatus activitypb.ActivityExecutionStatus
-	}{
-		{
-			name:           "from unspecified status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_UNSPECIFIED,
-		},
-		{
-			name:           "from started status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_STARTED,
-		},
-		{
-			name:           "from cancel requested status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
-		},
-		{
-			name:           "from completed status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_COMPLETED,
-		},
-		{
-			name:           "from cancelled status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED,
-		},
-		{
-			name:           "from terminated status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TERMINATED,
-		},
-		{
-			name:           "from timed out status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT,
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			ctx := &chasm.MockMutableContext{}
-
-			activity := &Activity{
-				ActivityState: &activitypb.ActivityState{
-					RetryPolicy:            defaultRetryPolicy,
-					ScheduleToCloseTimeout: durationpb.New(defaultScheduleToCloseTimeout),
-					ScheduleToStartTimeout: durationpb.New(defaultScheduleToStartTimeout),
-					StartToCloseTimeout:    durationpb.New(defaultStartToCloseTimeout),
-					Status:                 tc.startingStatus,
-				},
-				Attempt: chasm.NewDataField(ctx, &activitypb.ActivityAttemptState{Count: 1}),
-				Outcome: chasm.NewDataField(ctx, &activitypb.ActivityOutcome{}),
-			}
-
-			err := TransitionStarted.Apply(activity, ctx, nil)
-			require.Error(t, err)
-		})
-	}
 }
 
 func TestTransitionTimedout(t *testing.T) {
@@ -497,59 +330,6 @@ func TestTransitionTimedout(t *testing.T) {
 			}
 
 			require.Empty(t, ctx.Tasks)
-		})
-	}
-}
-
-func TestTransitionTimeOutFromInvalidStatus(t *testing.T) {
-	testCases := []struct {
-		name           string
-		startingStatus activitypb.ActivityExecutionStatus
-	}{
-		{
-			name:           "from unspecified status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_UNSPECIFIED,
-		},
-		{
-			name:           "from cancel requested status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
-		},
-		{
-			name:           "from completed status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_COMPLETED,
-		},
-		{
-			name:           "from cancelled status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED,
-		},
-		{
-			name:           "from terminated status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TERMINATED,
-		},
-		{
-			name:           "from timed out status",
-			startingStatus: activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT,
-		},
-	}
-
-	for _, tc := range testCases {
-		t.Run(tc.name, func(t *testing.T) {
-			ctx := &chasm.MockMutableContext{}
-
-			activity := &Activity{
-				ActivityState: &activitypb.ActivityState{
-					RetryPolicy:            defaultRetryPolicy,
-					ScheduleToCloseTimeout: durationpb.New(defaultScheduleToCloseTimeout),
-					ScheduleToStartTimeout: durationpb.New(defaultScheduleToStartTimeout),
-					StartToCloseTimeout:    durationpb.New(defaultStartToCloseTimeout),
-					Status:                 tc.startingStatus,
-				},
-				Attempt: chasm.NewDataField(ctx, &activitypb.ActivityAttemptState{Count: 1}),
-				Outcome: chasm.NewDataField(ctx, &activitypb.ActivityOutcome{}),
-			}
-
-			err := TransitionTimedOut.Apply(activity, ctx, enumspb.TIMEOUT_TYPE_START_TO_CLOSE)
-			require.Error(t, err)
 		})
 	}
 }


chasm/lib/activity/statemachine.go
```diff
@@ -25,20 +36,48 @@ func (a *Activity) SetStateMachineState(state activitypb.ActivityExecutionStatus
...
 	},
 	activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED,
-	func(a *Activity, ctx chasm.MutableContext, _ any) error {
+	func(a *Activity, ctx chasm.MutableContext, scheduledEvent scheduledEvent) error {
 		attempt, err := a.Attempt.Get(ctx)
 		if err != nil {
 			return err
 		}
 
+		currentTime := time.Now()
```
bergundy: Use `ctx.Now(a)` so we can later support time skipping and advanced pause features.

chasm/lib/activity/statemachine.go
```diff
@@ -49,7 +88,21 @@ var TransitionStarted = chasm.NewTransition(
...
+	func(a *Activity, ctx chasm.MutableContext, _ any) error {
+		attempt, err := a.Attempt.Get(ctx)
+		if err != nil {
+			return err
+		}
+
+		ctx.AddTask(
+			a,
+			chasm.TaskAttributes{
+				ScheduledTime: time.Now().Add(a.GetStartToCloseTimeout().AsDuration()),
```
bergundy: Same here, don't get use wall clock in general in CHASM.

chasm/lib/activity/statemachine.go
```diff
@@ -95,7 +148,7 @@ var TransitionTimedOut = chasm.NewTransition(
 		activitypb.ACTIVITY_EXECUTION_STATUS_STARTED,
 	},
 	activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT,
-	func(_ *Activity, _ chasm.MutableContext, _ any) error {
-		return nil
+	func(a *Activity, ctx chasm.MutableContext, event timeoutEvent) error {
```
bergundy: Doesn't seem like you need this closure, use the method reference instead.

common/backoff/retry.go
```diff
@@ -169,3 +171,24 @@ func IgnoreErrors(errorsToExclude []error) func(error) bool {
...
+
+// ExponentialBackoffAlgorithm calculates the backoff duration using exponential algorithm.
+// The result is initInterval * (backoffCoefficient ^ (currentAttempt - 1)).
+func ExponentialBackoffAlgorithm(initInterval *durationpb.Duration, backoffCoefficient float64, currentAttempt int32) time.Duration {
+	return time.Duration(int64(float64(initInterval.AsDuration().Nanoseconds()) * math.Pow(backoffCoefficient, float64(currentAttempt-1))))
+}
+
+// MakeBackoffAlgorithm creates a BackoffCalculatorAlgorithmFunc that returns a fixed delay if requestedDelay is non-nil,
+// otherwise falls back to exponential backoff algorithm.
+func MakeBackoffAlgorithm(requestedDelay *time.Duration) BackoffCalculatorAlgorithmFunc {
```
bergundy, line 190: Not blocking this PR because I see that you copied the code over but can't we use `0` for checking if requestedDelay is provided?

chasm/lib/activity/statemachine.go
```diff
@@ -1,22 +1,33 @@
...
+	"time"
+
+	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/server/chasm"
-	activitypb "go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"
+	"go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"
 )
 
+type scheduledEvent struct {
+	TaskStartDelay time.Duration // Delay before the activity tasks should be run, e.g., for retries
```
bergundy: Why do you need to provide this as input? You should be able to derive this from the current activity state.
EDIT: After reading the code where it's used, I see why you did it this way. The only reason seems to be that we want to ignore a negative interval and I don't understand the reasoning for that.

chasm/lib/activity/library.go
```diff
@@ -41,8 +50,23 @@ func (l *library) Tasks() []*chasm.RegistrableTask {
 	return []*chasm.RegistrableTask{
 		chasm.NewRegistrableSideEffectTask[*Activity, *activitypb.ActivityDispatchTask](
 			"dispatchActivity",
-			l.ActivityDispatchTaskExecutor,
-			l.ActivityDispatchTaskExecutor,
+			l.activityDispatchTaskExecutor,
+			l.activityDispatchTaskExecutor,
+		),
+		chasm.NewRegistrablePureTask(
+			"scheduleToStartTimer",
```
bergundy, line 57: I see the other task qualifies itself as an activity task and these don't, CHASM already qualifies all of the tasks with the library name, we should rename `dispatchActivity` to `dispatch`.

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+		return err
+	}
+
+	interval := calculateRetryInterval(retryPolicy, attempt.Count)
+	isNegativeInterval := retryPolicy.GetMaximumInterval().AsDuration() == 0 && interval <= 0 // From existing workflow activity retry behavior
+
+	// Retry task if we have remaining attempts. A retry involves transitioning the activity back to scheduled state.
+	if (retryPolicy.GetMaximumAttempts() == 0 || task.GetAttempt() < retryPolicy.GetMaximumAttempts()) && (!isNegativeInterval) {
+		attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(activity))
+		attempt.Count += 1
```
bergundy: This should happen in the state machine transition function.

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	task *activitypb.StartToCloseTimeoutTask,
+) error {
+	retryPolicy := activity.RetryPolicy
+	attempt, err := activity.Attempt.Get(ctx)
+	if err != nil {
+		return err
+	}
+
+	interval := calculateRetryInterval(retryPolicy, attempt.Count)
+	isNegativeInterval := retryPolicy.GetMaximumInterval().AsDuration() == 0 && interval <= 0 // From existing workflow activity retry behavior
```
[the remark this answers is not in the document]
bergundy: Looks like by definition this can only happen when `retryPolicy.GetMaximumInterval().AsDuration() == 0` based on the implementation of `calculateRetryInterval`.

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+
+	// Retry task if we have remaining attempts. A retry involves transitioning the activity back to scheduled state.
+	if (retryPolicy.GetMaximumAttempts() == 0 || task.GetAttempt() < retryPolicy.GetMaximumAttempts()) && (!isNegativeInterval) {
+		attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(activity))
+		attempt.Count += 1
+
+		return TransitionScheduled.Apply(activity, ctx, scheduledEvent{TaskStartDelay: interval})
+	}
+
+	// Reached maximum attempts, timeout the activity
```
bergundy, line 169: You should call `a.Store.RecordCompletion` here.

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	if err != nil {
+		return err
+	}
+
+	interval := calculateRetryInterval(retryPolicy, attempt.Count)
+	isNegativeInterval := retryPolicy.GetMaximumInterval().AsDuration() == 0 && interval <= 0 // From existing workflow activity retry behavior
+
+	// Retry task if we have remaining attempts. A retry involves transitioning the activity back to scheduled state.
+	if (retryPolicy.GetMaximumAttempts() == 0 || task.GetAttempt() < retryPolicy.GetMaximumAttempts()) && (!isNegativeInterval) {
+		attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(activity))
```
bergundy: Also all of the attempt information should be reset at this point otherwise it would seem as if the previous attempt's failure is relevant.

chasm/lib/activity/activity_tasks.go
bergundy: In a workflow this would trigger `TransitionTimedOut.Apply`...

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	return true, nil
+}
+
+func (e *scheduleToStartTimeoutTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	activity *Activity,
+	_ chasm.TaskAttributes,
+	_ *activitypb.ScheduleToStartTimeoutTask,
+) error {
+	return TransitionTimedOut.Apply(activity, ctx, timeoutEvent{TimeoutType: enumspb.TIMEOUT_TYPE_SCHEDULE_TO_START})
```
bergundy: This should go through `store.RecordCompletion`

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	activity *Activity,
+	_ chasm.TaskAttributes,
+	task *activitypb.ScheduleToStartTimeoutTask,
+) (bool, error) {
+	attempt, err := activity.Attempt.Get(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	if activity.Status != activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED || task.Attempt != attempt.Count {
```
bergundy: nit: I feel like the negative logic here is harder to parse.

How about this?

```go
valid := activity.Status == activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED && task.Attempt == attempt.Count
return valid, nil
```

chasm/lib/activity/activity.go
```diff
@@ -150,31 +155,52 @@ func (a *Activity) RecordActivityTaskStarted(ctx chasm.MutableContext, params Re
...
+			Failure: failure,
+		},
+	}
+
+	attempt, err := a.Attempt.Get(ctx)
+	if err != nil {
+		return err
+	}
+
+	attempt.LastFailure = failure
```
bergundy: Don't overwrite the last failure unless the timeout is a timeout for a certain attempt.
We want to preserve this information if possible.

chasm/lib/activity/activity.go
```diff
@@ -150,31 +155,52 @@ func (a *Activity) RecordActivityTaskStarted(ctx chasm.MutableContext, params Re
...
+	attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(a))
```
bergundy: Same for this field.

chasm/lib/activity/activity_tasks.go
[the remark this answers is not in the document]
bergundy: Also agree with copilot on the comment being vague.

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +69,134 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	attempt, err := activity.Attempt.Get(ctx)
+	if err != nil {
+		return err
+	}
+
+	interval := calculateRetryInterval(retryPolicy, attempt.Count)
+	isNegativeInterval := retryPolicy.GetMaximumInterval().AsDuration() == 0 && interval <= 0 // From existing workflow activity retry behavior
+
+	// Retry task if we have remaining attempts. A retry involves transitioning the activity back to scheduled state.
+	if (retryPolicy.GetMaximumAttempts() == 0 || task.GetAttempt() < retryPolicy.GetMaximumAttempts()) && (!isNegativeInterval) {
```
bergundy: Looks like we can drop the negative interval check since this implementation doesn't return a negative value.

common/backoff/retry.go
```diff
@@ -169,3 +171,24 @@ func IgnoreErrors(errorsToExclude []error) func(error) bool {
...
 	}
 }
+
+// BackoffCalculatorAlgorithmFunc is a function type that calculates backoff duration based on
+// initial duration, coefficient, and current attempt number.
+type BackoffCalculatorAlgorithmFunc func(duration *durationpb.Duration, coefficient float64, currentAttempt int32) time.Duration
+
+// ExponentialBackoffAlgorithm calculates the backoff duration using exponential algorithm.
+// The result is initInterval * (backoffCoefficient ^ (currentAttempt - 1)).
+func ExponentialBackoffAlgorithm(initInterval *durationpb.Duration, backoffCoefficient float64, currentAttempt int32) time.Duration {
```
bergundy, line 183: Let's make sure that this function doesn't overflow.

chasm/lib/activity/statemachine.go
bergundy: Okay, I reviewed the code again and you should just calculate the next start time in the transition function.

chasm/lib/activity/activity_tasks.go
fretz12: much clearer, thx. Applied in all occurrences.

chasm/lib/activity/activity_tasks.go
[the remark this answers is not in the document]
fretz12: I reworked this a bit so interval is always guranteed to be >= 0. Also verified initialInterval is always > 0 and added test to confirm.

common/backoff/retry.go
fretz12: Added a clamp so it never overflows.

common/backoff/retry.go
fretz12: seems like a safe thing to do since requestedDelay as a value will cover nil pointers to 0. Changed.

chasm/lib/activity/activity_tasks.go
fretz12: Dropped, and added check if enough time for a retry

chasm/lib/activity/activity_tasks.go
fretz12: Moved

chasm/lib/activity/activity_tasks.go
fretz12: Done, and moved into transition

chasm/lib/activity/library.go
fretz12: renamed dispatchActivity to dispatch

chasm/lib/activity/statemachine.go
fretz12: I reworked this so we don't need this anymore

chasm/lib/activity/statemachine.go
```diff
@@ -1,22 +1,33 @@
...
+
+type timeoutEvent struct {
+	TimeoutType enumspb.TimeoutType
+}
+
 // Ensure that Activity implements chasm.StateMachine interface
 var _ chasm.StateMachine[activitypb.ActivityExecutionStatus] = (*Activity)(nil)
 
-// State returns the current status of the activity.
+// StateMachineState State returns the current status of the activity.
```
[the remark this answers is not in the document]

chasm/lib/activity/statemachine.go
```diff
@@ -1,22 +1,33 @@
...
 func (a *Activity) StateMachineState() activitypb.ActivityExecutionStatus {
 	if a.ActivityState == nil {
 		return activitypb.ACTIVITY_EXECUTION_STATUS_UNSPECIFIED
 	}
 	return a.Status
 }
 
-// SetState sets the status of the activity.
+// SetStateMachineState SetState sets the status of the activity.
```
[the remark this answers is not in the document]

chasm/lib/activity/statemachine.go
fretz12: Fixed all occurrences

chasm/lib/activity/statemachine.go
fretz12: Fixed

chasm/lib/activity/statemachine.go

chasm/lib/activity/activity_tasks.go
fretz12: Done, and RecordCompletion to take in an apply func as we discussed.

chasm/lib/activity/activity.go
fretz12: Reworked to handle different timeout cases and attempt/outcome failure population as we discussed.

chasm/lib/activity/activity.go
fretz12: Reworked.

chasm/lib/activity/activity_tasks.go
fretz12: TransitionTimedOut now handles store.RecordCompletion

chasm/component.go
```diff
@@ -53,7 +53,7 @@ const (
 	LifecycleStateCompleted
 	LifecycleStateFailed
 	// LifecycleStateTerminated
-	// LifecycleStateTimedout
+	LifecycleStateTimedout
```
fretz12: We need to bring back LifecycleStateTimedout
because when a timeout lifecycle occurs the workflow status needs to update to WORKFLOW_EXECUTION_STATUS_TIMED_OUT (see change in tree.go), else we encounter an invalid state change: `unable to change workflow state from Created to Completed, status Failed`. This was found when testing a schedule-to-X timeout and an activity was in scheduled state and allowed to timeout.

chasm/lib/activity/proto/v1/activity_state.proto
```diff
@@ -157,6 +167,8 @@ message ActivityOutcome {
     }
 
     message Failed {
+        // Only filled on schedule-to-start timeouts, schedule-to-close timeouts or terminations. All other attempt
+        // failures will be recorded in ActivityAttemptState.last_failure_details
```
bergundy:
```suggestion
        // failures will be recorded in ActivityAttemptState.last_failure_details.
```

chasm/lib/activity/activity.go
```diff
@@ -16,8 +18,11 @@ import (
 )
 
 type ActivityStore interface {
+	// PopulateRecordActivityTaskStartedResponse populates the response for RecordActivityTaskStarted
 	PopulateRecordActivityTaskStartedResponse(ctx chasm.Context, key chasm.EntityKey, response *historyservice.RecordActivityTaskStartedResponse) error
-	RecordCompletion(ctx chasm.MutableContext) error
+
+	// RecordCompletion applies the provided function to record activity completion
+	RecordCompletion(ctx chasm.MutableContext, applyFn func(ctx chasm.MutableContext) error) error
```
bergundy, line 29: This is likely to change when we have an embedded-within-workflow implementation but don't stress about that too much now.

chasm/lib/activity/activity.go
```diff
@@ -116,11 +124,10 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
 
 	// Note: No need to set the vector clock here, as the components track version conflicts for read/write
 	return &matchingservice.AddActivityTaskRequest{
-		NamespaceId:            namespaceID,
-		TaskQueue:              a.GetTaskQueue(),
-		ScheduleToStartTimeout: a.GetScheduleToStartTimeout(),
```
bergundy, line 121: Why did you remove this? Can you also confirm that we've populated all of the relevant fields?

chasm/lib/activity/activity.go
```diff
@@ -150,24 +157,13 @@ func (a *Activity) RecordActivityTaskStarted(ctx chasm.MutableContext, params Re
...
-		return nil, err
-	}
-
-	activityRef, err := chasm.DeserializeComponentRef(activityRefBytes)
-	if err != nil {
-		return nil, err
-	}
-
 	response := &historyservice.RecordActivityTaskStartedResponse{}
 	if store == nil {
```
bergundy, line 167: We will need to revisit whether this component should be optional or a pointer to self.

chasm/lib/activity/proto/v1/activity_state.proto
```diff
@@ -125,14 +125,24 @@ message ActivityAttemptState {
...
+      google.protobuf.Timestamp last_failure_time = 1;
 
-    string last_worker_identity = 6;
+      // Failure details from the last failed attempt.
+      temporal.api.failure.v1.Failure last_failure = 2;
+    }
+
+    // Details about the last failure. This will only be updated when an activity attempt fails,
+    // including start-to-close timeout. Activity success, termination, schedule-to-start and schedule-to-close timeouts
+    // will not reset it.
```
bergundy, line 138: This statement needs to be tested in unit tests for the state machine transition functions.

chasm/lib/activity/proto/v1/activity_state.proto
bergundy: Also needs to be tested.

chasm/lib/activity/activity.go
```diff
@@ -218,8 +214,58 @@ func (a *Activity) PopulateRecordActivityTaskStartedResponse(ctx chasm.Context,
...
+}
+
+func (a *Activity) recordActivityTimedOut(ctx chasm.MutableContext, timeoutType enumspb.TimeoutType) error {
+	outcome, err := a.Outcome.Get(ctx)
+	if err != nil {
+		return err
+	}
+
+	failure := &failurepb.Failure{
+		Message: fmt.Sprintf("activity %s timed out", timeoutType.String()),
```
bergundy: Is this error message consistent with the failure message for workflow activities? I didn't see this string template in the codebase.

chasm/lib/activity/activity.go
```diff
@@ -218,8 +214,58 @@ func (a *Activity) PopulateRecordActivityTaskStartedResponse(ctx chasm.Context,
...
+		}
+
+		currentTime := timestamppb.New(ctx.Now(a))
+
+		attempt.LastFailureDetails = &activitypb.ActivityAttemptState_LastFailureDetails{
+			LastFailure:     failure,
+			LastFailureTime: currentTime,
+		}
+		attempt.LastAttemptCompleteTime = currentTime
+		outcome.Variant = &activitypb.ActivityOutcome_Failed_{}
```
bergundy: The outcome should only be updated here if you know for sure that this is the last attempt.

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +66,120 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	// Retry task if we have remaining attempts and time. A retry involves transitioning the activity back to scheduled state.
+	if enoughAttempts && enoughTime {
+		return TransitionScheduled.Apply(activity, ctx, nil)
+	}
+
+	// Reached maximum attempts, timeout the activity
+	return TransitionTimedOut.Apply(activity, ctx, enumspb.TIMEOUT_TYPE_START_TO_CLOSE)
+}
+
+func hasEnoughTimeForRetry(ctx chasm.Context, activity *Activity) (bool, error) {
```
bergundy: nit: this could be a method on the activity struct.

chasm/lib/activity/statemachine.go
```diff
@@ -33,12 +41,48 @@ var TransitionScheduled = chasm.NewTransition(
...
 
+		currentTime := ctx.Now(a)
+		retryInterval := time.Duration(0)
+		isRetry := attempt.GetCount() >= 1
+		attempt.Count += 1
+
+		// If this is a retry, calculate the delay before scheduling tasks and update attempt fields
+		if isRetry {
+			retryInterval = backoff.CalculateExponentialRetryInterval(a.GetRetryPolicy(), attempt.GetCount())
+			attempt.CurrentRetryInterval = durationpb.New(retryInterval)
```
bergundy: You might want to put a TODO here to set this to the next interval from the activity failure on a failed attempt.

chasm/lib/activity/statemachine.go
```diff
@@ -33,12 +41,48 @@ var TransitionScheduled = chasm.NewTransition(
...
+				a,
+				chasm.TaskAttributes{
+					ScheduledTime: currentTime.Add(timeout).Add(retryInterval),
+				},
+				&activitypb.ScheduleToStartTimeoutTask{
+					Attempt: attempt.GetCount(),
+				})
+		}
+
+		if timeout := a.GetScheduleToCloseTimeout().AsDuration(); timeout > 0 {
```
bergundy, line 56: This is just relevant for the first attempt.

You may want to have two separate transitions: scheduled and rescheduled or just put this logic in a conditional.

chasm/lib/activity/statemachine.go
```diff
@@ -25,6 +32,7 @@ func (a *Activity) SetStateMachineState(state activitypb.ActivityExecutionStatus
 var TransitionScheduled = chasm.NewTransition(
```
bergundy, line 31: You also need to call `recordActivityTimedOut` in this transition for start to close.

chasm/lib/activity/activity.go
fretz12: Ah, added back. Mistakenly assumed that field is no longer relevant since activity is taking over the timer for it, but looks like matching sets its own expire time. 
Do any of these fields matter for our case:
```
VersionDirective *v18.TaskVersionDirective
ForwardInfo      *v18.TaskForwardInfo      
Stamp            int32              
```

chasm/lib/activity/activity.go
fretz12: Yes, I've added the TODO docstring on the store field in the component.

chasm/lib/activity/activity.go
fretz12: I refactored this into a shared error message that wf-activities are using today. I'd like to make the message grammar a bit better, but not sure if we have the freedom to do it.

chasm/lib/activity/activity.go
fretz12: Added a check now since I realized this will be called on every retry attempt.

chasm/lib/activity/proto/v1/activity_state.proto
fretz12: Fixed

chasm/lib/activity/statemachine.go
fretz12: Ah, I missed this bit that we need to record timeout on every retry attempt. Added.

chasm/lib/activity/statemachine.go
fretz12: Yup, added

chasm/lib/activity/statemachine.go
fretz12: Good catch... I've added the rescheduled state, seems cleaner easier to read.

chasm/lib/activity/activity_tasks.go

chasm/lib/activity/proto/v1/activity_state.proto
fretz12: Statemachine unit tests added

chasm/lib/activity/proto/v1/activity_state.proto
fretz12: Statemachine unit tests added

chasm/lib/activity/statemachine_test.go
```diff
@@ -0,0 +1,588 @@
...
+			// Verify added tasks
+			require.Len(t, ctx.Tasks, len(tc.expectedTasks))
+			for i, expectedTask := range tc.expectedTasks {
+				actualTask := ctx.Tasks[i]
+
+				switch expectedTask.Payload.(type) {
+				case *activitypb.ActivityDispatchTask:
+					_, ok := actualTask.Payload.(*activitypb.ActivityDispatchTask)
+					require.True(t, ok, "expected ActivityDispatchTask at index %d", i)
+					require.Empty(t, 0, actualTask.Attributes.ScheduledTime)
```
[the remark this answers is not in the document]
fretz12: fixed

chasm/lib/activity/proto/v1/activity_state.proto
```diff
@@ -125,14 +125,24 @@ message ActivityAttemptState {
...
+    LastFailureDetails last_failure_details = 5;
+
+    string last_worker_identity =7;
```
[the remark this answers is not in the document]
fretz12: fixed

chasm/lib/activity/proto/v1/activity_state.proto
```diff
@@ -125,14 +125,24 @@ message ActivityAttemptState {
...
 
-    // Failure details from the last failed attempt.
-    temporal.api.failure.v1.Failure last_failure = 5;
+    message LastFailureDetails {
+      // The last time the activity attempt failed.
+      google.protobuf.Timestamp last_failure_time = 1;
 
-    string last_worker_identity = 6;
+      // Failure details from the last failed attempt.
+      temporal.api.failure.v1.Failure last_failure = 2;
```
bergundy: I wouldn't stutter here.


```suggestion
      google.protobuf.Timestamp time = 1;

      // Failure details from the last failed attempt.
      temporal.api.failure.v1.Failure failure = 2;
```

chasm/lib/activity/activity.go
bergundy: ForwardInfo is set by matching when forwarding between partitions.
VersionDirective I think is relevant, but you'd have to ask the matching team, @dandavison is supposed to discuss standalone activities and versioning with them.
Stamp is something that we will need to drop tasks after reset and update-options. We don't support those today but I think we should use that concept (or something similar).

chasm/lib/activity/activity.go
```diff
@@ -54,9 +62,10 @@ func (a *Activity) LifecycleState(_ chasm.Context) chasm.LifecycleState {
 		return chasm.LifecycleStateCompleted
 	case activitypb.ACTIVITY_EXECUTION_STATUS_FAILED,
 		activitypb.ACTIVITY_EXECUTION_STATUS_TERMINATED,
-		activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED,
-		activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT:
+		activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED:
 		return chasm.LifecycleStateFailed
+	case activitypb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT:
+		return chasm.LifecycleStateTimedout
```
bergundy: Discussed internally, we do not want to add this new lifecycle state.

common/util.go
```diff
@@ -89,6 +89,8 @@ const (
 
 	contextExpireThreshold = 10 * time.Millisecond
 
+	// FailureReasonActivityTimeout is failureReason for when an activity times out, with %v as the timeout type.
+	FailureReasonActivityTimeout = "activity %v timeout"
```
bergundy, line 93: This is overkill IMHO, let's not make a constant for this.

chasm/lib/activity/statemachine.go
```diff
@@ -33,28 +40,120 @@ var TransitionScheduled = chasm.NewTransition(
...
+		attempt, err := a.Attempt.Get(ctx)
+		if err != nil {
+			return err
+		}
+
+		currentTime := ctx.Now(a)
+		attempt.Count += 1
+
+		// If this is a retry, calculate the delay before scheduling tasks and update attempt fields
+		// TODO: for activity failures it'll go through this retry path as well; we'll need to refactor the record timeout, probably passed as an event func
```
bergundy: You'll also want to set the `CurrentRetryInterval` based on the error the activity attempt returned.

chasm/lib/activity/statemachine.go
```diff
@@ -33,28 +40,120 @@ var TransitionScheduled = chasm.NewTransition(
 			return err
 		}
 
+		if attempt.Count != 0 {
```
bergundy: If you don't expect this to happen, don't check for it. This can only be due to a bug in this library.

chasm/lib/activity/statemachine_test.go
```diff
@@ -0,0 +1,588 @@
...
+			},
+			scheduleToStartTimeout: defaultScheduleToStartTimeout,
+			scheduleToCloseTimeout: 0,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := &chasm.MockMutableContext{}
+			ctx.HandleNow = func(chasm.Component) time.Time { return defaultTime }
```
bergundy: nit:

```suggestion
			ctx := &chasm.MockMutableContext{
				HandleNow: func(chasm.Component) time.Time { return defaultTime },
			}
```

chasm/lib/activity/statemachine_test.go
```diff
@@ -0,0 +1,588 @@
...
+
+			// Verify added tasks
+			require.Len(t, ctx.Tasks, len(tc.expectedTasks))
+			for i, expectedTask := range tc.expectedTasks {
+				actualTask := ctx.Tasks[i]
+
+				switch expectedTask.Payload.(type) {
+				case *activitypb.ActivityDispatchTask:
+					_, ok := actualTask.Payload.(*activitypb.ActivityDispatchTask)
+					require.True(t, ok, "expected ActivityDispatchTask at index %d", i)
```
bergundy: nit: You should be able to use `require.IsType` here.

chasm/lib/activity/statemachine_test.go
```diff
@@ -0,0 +1,588 @@
...
+		},
+		Attempt: chasm.NewDataField(ctx, &activitypb.ActivityAttemptState{Count: 1}),
+		Outcome: chasm.NewDataField(ctx, &activitypb.ActivityOutcome{}),
+	}
+
+	err := TransitionScheduled.Apply(activity, ctx, nil)
+	require.Error(t, err)
+}
+
+func TestTransitionScheduledFromInvalidStatus(t *testing.T) {
```
bergundy: Why are you testing this? It's guaranteed by the way that you've defined the transition.

chasm/lib/activity/statemachine_test.go
```diff
@@ -0,0 +1,588 @@
...
+				default:
+					t.Fatalf("unexpected task payload type at index %d: %T", i, actualTask.Payload)
+				}
+
+			}
+		})
+	}
+}
+
+func TestTransitionRescheduledFromInvalidStatus(t *testing.T) {
```
bergundy: Again here. Don't test this.

chasm/lib/activity/statemachine_test.go
```diff
@@ -0,0 +1,588 @@
...
+				},
+				Attempt: chasm.NewDataField(ctx, attemptState),
+				Outcome: chasm.NewDataField(ctx, outcome),
+			}
+
+			err := TransitionRescheduled.Apply(activity, ctx, nil)
+			require.NoError(t, err)
+			require.Equal(t, activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED, activity.Status)
+			require.Equal(t, tc.startingAttemptCount+1, attemptState.Count)
+			require.Equal(t, durationpb.New(tc.expectedRetryInterval), attemptState.GetCurrentRetryInterval())
```
bergundy: Typically better to use `protorequire` for testing proto equality, but if this passes you're fine.

bergundy (approved):
I left some comments, mostly nits.

chasm/lib/activity/activity.go
fretz12: Hmm.. what's the short/long term plan?
Right now, if I don't introduce a new status mapping (see tree.go), I will hit this [error](https://github.com/temporalio/temporal/blob/d3e4c8125d495a8c165401ceb93b1591c582e34b/service/history/workflow/mutable_state_state_status.go#L40) because schedule-to-X timeouts go from created to failed. I could possibly change the validation in `mutable_state_state_status.go` but not sure if that's advised.

chasm/lib/activity/activity.go
fretz12: Added a TODO for VersionDirective, and will connect with Dan on it.

chasm/lib/activity/proto/v1/activity_state.proto
fretz12: Renamed

chasm/lib/activity/statemachine.go
fretz12: Yea shouldn't happen. Removed.

chasm/lib/activity/statemachine.go
fretz12: Yup. Added reminder in TODO. Also refactored timeout method a bit to make the CurrentRetryInterval recorded within.

common/util.go
fretz12: yea it's kinda silly, but if want consistent failure messages I feel it should stay in one place.

chasm/lib/activity/activity.go
bergundy: Keep it as failed and implement the SearchAttributesProvider interface to set the status.

fretz12: I will revert to failed and implement SearchAttributesProvider as a separate PR after merging this if there are no objections.

chasm/lib/activity/statemachine_test.go
fretz12: This would be the one-liner. Replaced.
```
			ctx := &chasm.MockMutableContext{
				MockContext: chasm.MockContext{
					HandleNow: func(chasm.Component) time.Time { return defaultTime },
				},
			}
```

chasm/lib/activity/statemachine_test.go
fretz12: Great, used it and refactored.

chasm/lib/activity/statemachine_test.go
fretz12: It's to guard against someone inadvertently adding/removing a state. I suppose that may be overkill as they really have to do that explicitly and should be caught in a code review if not intended. I will remove it then.

chasm/lib/activity/statemachine_test.go
fretz12: removed

chasm/lib/activity/statemachine_test.go
fretz12: Changed to use proto.equal

chasm/lib/activity/activity_tasks.go
```diff
@@ -64,3 +65,107 @@ func (e *activityDispatchTaskExecutor) Execute(
...
+	activity *Activity,
+	_ chasm.TaskAttributes,
+	task *activitypb.ScheduleToCloseTimeoutTask,
+) (bool, error) {
+	attempt, err := activity.Attempt.Get(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	valid := TransitionTimedOut.Possible(activity) && task.Attempt == attempt.Count
```
dandavison, line 116: This looks like a bug. The schedule-to-close timer task should still be considered valid if the attempt count has advanced beyond the attempt count when the task was created.

dandavison: I've repro'd this in a test; will create a PR.

dandavison: https://github.com/temporalio/temporal/pull/8720

PR #8576, opened 2025-10-30
Add workflow update-options identity to history event

carlydf:
## What changed?
Add workflow update-options identity to history event

## Why?
So that users can use the event as an audit log to find out who set an override

## How did you test it?
- [x] built
- [x] added new unit test(s)

---

> [!NOTE]
> Add identity to `WORKFLOW_EXECUTION_OPTIONS_UPDATED` history event and propagate through APIs, batcher, NDC, and tests.
> 
> - **History events**:
>   - Add `identity` to `WorkflowExecutionOptionsUpdated` attributes and wire through `EventFactory`, `HistoryBuilder`, and `MutableState` (`AddWorkflowExecutionOptionsUpdatedEvent`).
> - **APIs**:
>   - `UpdateWorkflowOptions.MergeAndApply` now accepts `identity` and passes it to mutable state.
>   - `updateworkflowoptions` handler uses `req.GetIdentity()`; `startworkflow` passes empty identity for attach-only updates.
> - **NDC/Reset**:
>   - Reapplier and workflow resetter pass through/propagate event `identity`.
> - **Worker batcher**:
>   - Batch update-workflow-options forwards `Identity` in requests.
> - **Tests**:
>   - Update mocks and unit tests to include `identity` in expectations.
>   - Add functional checks to assert event `identity` in history for direct and batch updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8958b2f2b1e71bf6bf497993a7f299d25188cd3c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/ndc/workflow_resetter.go
+++ b/service/history/ndc/workflow_resetter.go
@@ -1018,7 +1018,7 @@ func reapplyEvents(
 // This function is intended to pick up all the events for a child that was already initialized before the reset point.
 // Re-applying these events is needed to support reconnecting of the child with parent.
 func reapplyChildEvents(mutableState historyi.MutableState, event *historypb.HistoryEvent) error { // nolint:revive
-	switch event.GetEventType() {                                                                  // nolint:exhaustive
+	switch event.GetEventType() { // nolint:exhaustive
 	case enumspb.EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_FAILED:
 		childEventAttributes := event.GetStartChildWorkflowExecutionFailedEventAttributes()
 		_, childExists := mutableState.GetChildExecutionInfo(childEventAttributes.GetInitiatedEventId())


service/history/api/startworkflow/api.go
```diff
@@ -663,6 +663,7 @@ func (s *Starter) handleUseExistingWorkflowOnConflictOptions(
 					requestID,
 					completionCallbacks,
 					links,
+					"",
```
carlydf, line 671: start workflow request does not provide identity, so we don't have anything to put here

service/history/ndc/workflow_resetter.go
```diff
@@ -949,6 +949,7 @@ func reapplyEvents(
 				requestID,
 				callbacks,
 				event.Links,
+				attr.GetIdentity(),
```
carlydf, line 952: reapply the identity that was already on the update-options event

service/history/ndc/workflow_resetter.go
```diff
@@ -1162,7 +1163,7 @@ func (r *workflowResetterImpl) performPostResetOperations(ctx context.Context, r
 	for _, operation := range postResetOperations {
 		switch op := operation.GetVariant().(type) {
 		case *workflowpb.PostResetOperation_UpdateWorkflowOptions_:
-			_, _, err := updateworkflowoptions.MergeAndApply(resetMS, op.UpdateWorkflowOptions.GetWorkflowExecutionOptions(), op.UpdateWorkflowOptions.GetUpdateMask())
+			_, _, err := updateworkflowoptions.MergeAndApply(resetMS, op.UpdateWorkflowOptions.GetWorkflowExecutionOptions(), op.UpdateWorkflowOptions.GetUpdateMask(), "")
```
carlydf, line 1167: don't apply an identity for post-reset-operations update-options, because even if we applied the identity of the reset request, the reset code paths are used in a lot of places and I don't want to accidentally have inconsistencies on the passive cluster or something. Better to just be consistent and not have an identity if the update-options was part of a reset operation (for now at least)

service/history/api/startworkflow/api.go
carlydf: this is for nexus callbacks and not versioning override anyway

service/history/ndc/workflow_resetter.go
ShahabT: with state-based replication this code might not run on the passive side.

Ideally we should put the reset requester in the event but I'm fine doing it later. But please add this explanation in the code as a comment so the future contributor can understand the reasoning.

PR #8578, opened 2025-10-31
Change chasm.ReadComponent, PollComponent, and UpdateComponent to take an any

bergundy:
## What changed?

In title.

## Why?

Components are accessed via various interfaces and we do not expect all of those interfaces to embed `Component`.

### DIFF
--- a/chasm/engine.go
+++ b/chasm/engine.go
@@ -176,7 +176,7 @@ func UpdateWithNewEntity[C Component, I any, O1 any, O2 any](
 //   - consider remove ComponentRef from the return value and allow components to get
 //     the ref in the transition function. There are some caveats there, check the
 //     comment of the NewRef method in MutableContext.
-func UpdateComponent[C Component, R []byte | ComponentRef, I any, O any](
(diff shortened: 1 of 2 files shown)


chasm/engine.go
```diff
@@ -234,14 +234,14 @@ func ReadComponent[C Component, R []byte | ComponentRef, I any, O any](
 	return output, err
 }
 
-type PollComponentRequest[C Component, I any, O any] struct {
+type PollComponentRequest[C any, I any, O any] struct {
```
yycptt: hmm I think this struct is not actually used anywhere, I probably missed this in previous refactoring. Can just delete.

chasm/engine.go
```diff
@@ -234,14 +234,14 @@ func ReadComponent[C Component, R []byte | ComponentRef, I any, O any](
...
 	Ref         ComponentRef
 	PredicateFn func(C, Context, I) bool
 	OperationFn func(C, MutableContext, I) (O, error)
 	Input       I
 }
```
bergundy:
```suggestion
```

PR #8582, opened 2025-10-31
Add CHASM callback into mutable state

spkane31:
## What changed?
Integrated chasm/lib/callback into MutableState and updated callback-related APIs to use CHASM implementation. Added integration tests following the scheduler pattern.

## Why?
Porting callback functionality from HSM to CHASM as part of the ongoing HSM-to-CHASM migration effort.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

## Potential risks
This will need to be tested more, but is not being turned on with this PR

---

> [!NOTE]
> Integrates CHASM-based callbacks into mutable state and history APIs, adds configs and error types, updates routing/state transitions, and includes migration tests.
> 
> - **CHASM Callback Integration**:
>   - Add CHASM callback module (FX wiring, HTTP caller provider, request routing) and metrics shared with HSM implementation.
>   - Implement `ToAPICallback`, invocation flow changes, and refined error wrapping.
> - **History/MutableState**:
>   - Expose `ChasmEnabled`, `ChasmWorkflowComponent{,ReadOnly}`; use CHASM for adding/processing completion callbacks when enabled.
>   - Describe API now builds callback info from both CHASM and HSM trees.
>   - Continue-as-new/retry path aggregates callbacks from both implementations.
> - **State Machine/Executors**:
>   - Switch to `queues/errors` types; move `NamespaceIDAndDestination` to `queues/common`.
>   - Adjust transitions: generate `BackoffTask` with scheduled time; include destination on reschedule; no tasks on success/fail.
>   - Simplify retry result (policy passed separately) and validation signatures.
> - **Config**:
>   - Add `EnableCHASMCallbacks` and `MaxCHASMCallbacksPerWorkflow`; wire through history configs.
> - **Tests**:
>   - Add migration and CHASM-enabled functional tests; update unit tests to new error/types and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e3b51a931d3ea2dd7f18fb2f9c55fc880ba828d8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chasm/lib/callback/statemachine.go
+++ b/chasm/lib/callback/statemachine.go
@@ -20,7 +20,7 @@ var TransitionScheduled = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_STANDBY},
 	callbackspb.CALLBACK_STATUS_SCHEDULED,
 	func(cb *Callback, ctx chasm.MutableContext, event EventScheduled) error {
-		u, err := url.Parse(cb.Callback.GetNexus().Url)
+		u, err := url.Parse(cb.Callback.GetNexus().GetUrl())
 		if err != nil {
 			return fmt.Errorf("failed to parse URL: %v: %w", cb.Callback, err)
 		}


chasm/lib/callback/fx.go
```diff
@@ -2,21 +2,34 @@ package callback
...
+func register(
 	registry *chasm.Registry,
 	library *Library,
 ) error {
 	return registry.Register(library)
 }
 
+// httpCallerProviderProvider adapts the HSM callbacks HTTPCallerProvider to the CHASM callback HTTPCallerProvider.
+// Since both types have the same signature, we just need to wrap the function.
+func httpCallerProviderProvider(hsmProvider callbacks.HTTPCallerProvider) HTTPCallerProvider {
```
bergundy: Let's require both implementations to be injected, we eventually want to deprecate the HSM implementation completely.

chasm/lib/callback/invocation_tasks_test.go
```diff
@@ -0,0 +1,298 @@
...
+func (s *invocationTasksSuite) SetupTest() {
+	s.callbackSuite.SetupTest()
+}
+
+// TestInvocationTask_SuccessfulNexusCallback tests a successful HTTP callback invocation.
+// Mirrors HSM test: TestProcessInvocationTaskNexus_Outcomes/success
+// TODO (seankane): This test is functionally a duplicate of TestExecuteInvocationTaskNexus_Outcomes/success
+// in executors_test.go until CHASM testing infrastructure matures to support verifying task generation
+// in integration tests. Once that's available this test will test the full CHASM stack integration.
+func (s *invocationTasksSuite) TestInvocationTask_SuccessfulNexusCallback() {
```
bergundy: What is this testing? We already have `TestExecuteInvocationTaskNexus_Outcomes` from the previous PR.

chasm/lib/callback/callback_suite_test.go
```diff
@@ -0,0 +1,345 @@
...
+	"go.temporal.io/server/common/testing/testlogger"
+	"go.temporal.io/server/common/testing/testvars"
+	"go.temporal.io/server/service/history/queues"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// callbackSuite sets up a suite that has a basic CHASM tree ready
+// for use with the Callback library, integrated within a Workflow component.
+type callbackSuite struct {
```
bergundy: We said we will not create more testify suite based unit tests. Please don't add more of this to the codebase.

chasm/lib/callback/invocation_tasks_test.go
```diff
@@ -0,0 +1,298 @@
...
+	// Verify callback transitioned to SUCCEEDED
+	s.Equal(callbackspb.CALLBACK_STATUS_SUCCEEDED, cb.Status)
+}
+
+// TestInvocationTask_FailureTriggersBackoff tests that a retryable HTTP error triggers backoff.
+// Mirrors HSM test: TestProcessInvocationTaskNexus_Outcomes/retryable-error
+// TODO (seankane): This test is functionally a duplicate of TestExecuteInvocationTaskNexus_Outcomes/retryable-http-error
+// in executors_test.go until CHASM testing infrastructure matures to support verifying task generation
+// in integration tests. Once that's available this test will test the full CHASM stack integration.
+func (s *invocationTasksSuite) TestInvocationTask_FailureTriggersBackoff() {
```
bergundy: Why do you need to test this if we have `executors_test.go` and tests for the statemachine transition that where this task is added?

chasm/lib/callback/invocation_tasks_test.go
```diff
@@ -0,0 +1,298 @@
...
+	// Try to validate task with old attempt number (3)
+	valid, err := s.invocationExecutor.Validate(ctx, cb, chasm.TaskAttributes{}, &callbackspb.InvocationTask{
+		Attempt: 3,
+	})
+	s.NoError(err)
+	s.False(valid, "stale task should be rejected")
+}
+
+// TestInvocationTask_ValidateWrongState tests that validation rejects tasks when callback is in wrong state.
+func (s *invocationTasksSuite) TestInvocationTask_ValidateWrongState() {
```
bergundy: These tests feel like a bit of overkill to me because the logic is so straightforward but you're welcome to keep them if you feel like it's adding value.

chasm/lib/callback/metrics.go
```diff
@@ -1,12 +1,11 @@
...
-)
-var RequestLatencyHistogram = metrics.NewTimerDef(
-	"callback_outbound_latency",
-	metrics.WithDescription("Latency histogram of CHASM outbound callback requests made by the history service."),
+// Re-export the callback metrics from components/callbacks to maintain a consistent interface.
+// Both the HSM-based callbacks (components/callbacks) and CHASM callbacks use the same metrics,
+// allowing for unified monitoring and drop-in compatibility.
+var (
+	RequestCounter          = callbacks.RequestCounter
+	RequestLatencyHistogram = callbacks.RequestLatencyHistogram
```
bergundy: Don't re-export, just copy. We want to just leave the HSM codebase alone and eventually remove it (it just might take years before we can do that).

chasm/lib/callback/statemachine.go
```diff
@@ -31,7 +31,11 @@ var TransitionRescheduled = chasm.NewTransition(
 	callbackspb.CALLBACK_STATUS_SCHEDULED,
 	func(cb *Callback, ctx chasm.MutableContext, event EventRescheduled) error {
 		cb.NextAttemptScheduleTime = nil
-		ctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{Attempt: cb.Attempt})
+		ctx.AddTask(
+			cb,
+			chasm.TaskAttributes{ScheduledTime: time.Time{}},
```
bergundy: just omit ScheduledTime, but you will need the destination. I missed that in the last review.

chasm/lib/callback/statemachine.go
```diff
@@ -60,7 +64,11 @@ var TransitionAttemptFailed = chasm.NewTransition(
 				},
 			},
 		}
-		ctx.AddTask(cb, chasm.TaskAttributes{ScheduledTime: time.Time{}}, &callbackspb.InvocationTask{})
+		ctx.AddTask(
+			cb,
+			chasm.TaskAttributes{ScheduledTime: nextAttemptScheduleTime},
```
bergundy, line 79: I'm surprised I missed this in the previous review. Thanks for fixing.

chasm/lib/callback/statemachine.go
bergundy, line 1: Now that I see that there were bugs added when translating the code from HSM, I want to ask you to do another review and verify nothing else was missed.

common/dynamicconfig/constants.go
```diff
@@ -2622,6 +2622,20 @@ instead of the existing (V1) implementation.`,
...
 	)
 
+	EnableCHASMCallbackCreation = NewNamespaceBoolSetting(
+		"history.enableCHASMCallbackCreation",
+		false,
+		`EnableCHASMCallbackCreation controls whether new callbacks are created using the CHASM (V2) implementation
+instead of the existing (V1) implementation.`,
+	)
+
+	EnableCHASMCallbackMigration = NewNamespaceBoolSetting(
```
bergundy: We don't support this functionality and don't plan on it, you can remove this setting.

common/dynamicconfig/constants.go
```diff
@@ -2622,6 +2622,20 @@ instead of the existing (V1) implementation.`,
 to the CHASM (V2) implementation on active scheduler workflows.`,
 	)
 
+	EnableCHASMCallbackCreation = NewNamespaceBoolSetting(
+		"history.enableCHASMCallbackCreation",
+		false,
+		`EnableCHASMCallbackCreation controls whether new callbacks are created using the CHASM (V2) implementation
+instead of the existing (V1) implementation.`,
```
bergundy:
```suggestion
		`Controls whether new callbacks are created using the CHASM implementation
instead of the previous HSM backed implementation.`,
```

common/dynamicconfig/constants.go
```diff
@@ -2622,6 +2622,20 @@ instead of the existing (V1) implementation.`,
 to the CHASM (V2) implementation on active scheduler workflows.`,
 	)
 
+	EnableCHASMCallbackCreation = NewNamespaceBoolSetting(
```
bergundy:
```suggestion
	EnableCHASMCallbacks = NewNamespaceBoolSetting(
```

common/dynamicconfig/constants.go
```diff
@@ -2622,6 +2622,20 @@ instead of the existing (V1) implementation.`,
...
+		"history.enableCHASMCallbackCreation",
```
bergundy:
```suggestion
		"history.enableCHASMCallbacks",
```

service/history/api/describeworkflow/api.go
```diff
@@ -235,38 +241,44 @@ func Invoke(
...
-			return nil, serviceerror.NewInternal("failed to construct describe response")
+	// Check if using CHASM callbacks
+	var callbackInfos []*workflowpb.CallbackInfo
+	useChasmCallbacks := false
+	chasmTree := mutableState.ChasmTree()
+	if chasmTree != nil {
+		// Check if CHASM tree is enabled (not a noop)
+		if chasmTree.Archetype() != "" {
+			// CHASM tree is enabled, check if callback creation flag is on
+			if shard.GetConfig().EnableCHASMCallbackCreation(mutableState.GetNamespaceEntry().Name().String()) {
```
bergundy: You're still going to want to list any CHASM callbacks attached even if the feature is turned off.

service/history/api/describeworkflow/api.go
```diff
@@ -235,38 +241,44 @@ func Invoke(
...
-				tag.WorkflowRunID(executionState.RunId),
-				tag.Error(err),
-			)
-			return nil, serviceerror.NewInternal("failed to construct describe response")
-		}
-		if callbackInfo == nil {
-			continue
-		}
-		result.Callbacks = append(result.Callbacks, callbackInfo)
+	if useChasmCallbacks {
```
bergundy: This is wrong, a single execution could have both CHASM and HSM callbacks.

service/history/api/describeworkflow/api.go
```diff
@@ -368,6 +382,201 @@ func buildCallbackInfo(
...
+				"failed to load callback data while building describe response",
+				tag.WorkflowNamespaceID(namespaceID.String()),
+				tag.WorkflowID(executionInfo.WorkflowId),
+				tag.WorkflowRunID(executionState.RunId),
+				tag.Error(err),
+			)
+			return nil, serviceerror.NewInternal("failed to construct describe response")
+		}
+
+		callbackInfo, err := buildCallbackInfo(namespaceID, callback, outboundQueueCBPool)
```
bergundy:
```suggestion
		callbackInfo, err := buildCallbackInfoFromHSM(namespaceID, callback, outboundQueueCBPool)
```

service/history/api/describeworkflow/api.go
```diff
@@ -368,6 +382,201 @@ func buildCallbackInfo(
...
+	return result, nil
+}
+
+// buildCallbackInfoFromChasm converts a single CHASM callback to API format.
+func buildCallbackInfoFromChasm(
+	namespaceID namespace.ID,
+	callback *chasmcallback.Callback,
+	outboundQueueCBPool *circuitbreakerpool.OutboundQueueCircuitBreakerPool,
+) (*workflowpb.CallbackInfo, error) {
+	if callback.Callback.GetNexus() == nil {
```
bergundy: nit: prefer using getters when working with protos


```suggestion
	if callback.GetCallback().GetNexus() == nil {
```

service/history/api/describeworkflow/api.go
```diff
@@ -368,6 +382,201 @@ func buildCallbackInfo(
...
+			},
+		}
+	default:
+		data, err := proto.Marshal(callback.Callback)
+		if err != nil {
+			return nil, err
+		}
+		cbSpec.Variant = &commonpb.Callback_Internal_{
+			Internal: &commonpb.Callback_Internal{Data: data},
+		}
```
bergundy: We don't actually support internal callbacks anymore, you can remove this code.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2821,6 +2823,10 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
 	requestID string,
 	completionCallbacks []*commonpb.Callback,
 ) error {
+	if ms.chasmEnabled() && ms.config.EnableCHASMCallbackCreation(ms.GetNamespaceEntry().Name().String()) {
+		return ms.addCompletionCallbacksChasm(event, requestID, completionCallbacks)
+	}
+
 	coll := callbacks.MachineCollection(ms.HSM())
```
bergundy, line 2990: Would you split this logic into a separate `addCompletionCallbacksHSM` function please?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2866,6 +2872,77 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+) error {
+	ctx := chasm.NewMutableContext(context.Background(), ms.chasmTree.(*chasm.Node))
+	rootComponent, err := ms.chasmTree.ComponentByPath(ctx, nil)
+	if err != nil {
+		return err
+	}
+	wf, ok := rootComponent.(*chasmworkflow.Workflow)
+	if !ok {
+		return serviceerror.NewInternalf("expected workflow component, but got %T", rootComponent)
+	}
```
bergundy: Might as well add a helper in ms for this, it's going to come up more.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2866,6 +2872,77 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+
+	// Check max callbacks limit
+	maxCallbacksPerWorkflow := ms.config.MaxCallbacksPerWorkflow(ms.GetNamespaceEntry().Name().String())
+	currentCallbackCount := len(wf.Callbacks)
+	if len(completionCallbacks)+currentCallbackCount > maxCallbacksPerWorkflow {
+		return serviceerror.NewInvalidArgumentf(
```
bergundy: I realize that a failed precondition status code is better here.

bergundy: Could you change here and in the HSM implementation please?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2866,6 +2872,77 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+	if err != nil {
+		return err
+	}
+	wf, ok := rootComponent.(*chasmworkflow.Workflow)
+	if !ok {
+		return serviceerror.NewInternalf("expected workflow component, but got %T", rootComponent)
+	}
+
+	// Check max callbacks limit
+	maxCallbacksPerWorkflow := ms.config.MaxCallbacksPerWorkflow(ms.GetNamespaceEntry().Name().String())
```
bergundy: We will need a separate config for max callbacks with CHASM, we want to allow 2000 with this implementation.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2866,6 +2872,77 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+		id := ""
+		// This is for backwards compatibility: callbacks were initially only attached when the workflow
+		// execution started, but now they can be attached while the workflow is running.
+		if event.GetEventType() == enumspb.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED {
+			// Use the start event version and ID as part of the callback ID to ensure that callbacks have unique
+			// IDs that are deterministically created across clusters.
+			id = fmt.Sprintf("%d-%d-%d", event.GetVersion(), event.GetEventId(), idx)
+		} else {
+			id = fmt.Sprintf("cb-%s-%d", requestID, idx)
+		}
```
bergundy: You no longer have the backwards compatibility concerns.
```suggestion
		id := fmt.Sprintf("%s-%d", requestID, idx)
```

service/history/workflow/mutable_state_impl.go
```diff
@@ -2866,6 +2872,77 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+		return serviceerror.NewInvalidArgumentf(
+			"cannot attach more than %d callbacks to a workflow (%d callbacks already attached)",
+			maxCallbacksPerWorkflow,
+			currentCallbackCount,
+		)
+	}
+
+	// Initialize map if needed
+	if wf.Callbacks == nil {
+		wf.Callbacks = make(chasm.Map[string, *chasmcallback.Callback])
```
bergundy: Might as well allocate enough space for adding the requested callbacks.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6114,6 +6191,10 @@ func (ms *MutableStateImpl) AddHistorySize(size int64) {
 // processCloseCallbacks triggers "WorkflowClosed" callbacks, applying the state machine transition that schedules
 // callback tasks.
 func (ms *MutableStateImpl) processCloseCallbacks() error {
+	if ms.chasmEnabled() && ms.config.EnableCHASMCallbackCreation(ms.GetNamespaceEntry().Name().String()) {
+		return ms.processCloseCallbacksChasm()
```
bergundy: Can't return here, you have to process both callback implementations.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6139,6 +6220,42 @@ func (ms *MutableStateImpl) processCloseCallbacks() error {
 	return nil
 }
 
+// processCloseCallbacksChasm triggers "WorkflowClosed" callbacks using the CHASM implementation.
+func (ms *MutableStateImpl) processCloseCallbacksChasm() error {
+	// Skip callbacks if workflow was reset
+	resetRunID := ms.GetExecutionInfo().GetResetRunId()
+	if ms.GetExecutionInfo().GetWorkflowWasReset() && resetRunID != "" && resetRunID != ms.executionState.RunId {
+		return nil
+	}
```
bergundy: Move this to the top of `processCloseCallbacks` instead of duplicating.

chasm/lib/callback/chasm_invocation.go
```diff
@@ -33,11 +33,22 @@ type chasmInvocation struct {
...
+	}
+
+	// For permanent failures, return unprocessable task error
 	if failure, ok := result.(invocationResultFail); ok {
 		return queues.NewUnprocessableTaskError(failure.err.Error())
 	}
 
-	return result.error()
+	return nil
 }
```
bergundy, line 41: I realized there's no need to wrap errors here.

```suggestion
func (c chasmInvocation) WrapError(result invocationResult, err error) error {
	return err
}
```

chasm/lib/callback/backoff_tasks_test.go
bergundy, line 1: This should be covered by the state machine transition tests.

chasm/lib/callback/fx.go
spkane31: Can you elaborate on this? both the HSM `callbacks.HTTPCallerProvider` and `HTTPCaller` (CHASM version) should be included in the `fx.Module`?

chasm/lib/callback/metrics.go
spkane31: In the [`buildRegistry`](https://github.com/temporalio/temporal/blob/main/common/metrics/registry.go#L53-L70) we return an error if we have duplicated names. So the duplication adds a some complicated code that I'm hesitant to add. Why can't we do this and then when we remove copy the metrics definition over?

chasm/lib/callback/chasm_invocation.go
spkane31: Removing this change to be consistent with HSM, we'll wrap the `invocationResultFail` in `queues.NewUnprocessableTaskError`

bergundy: No need to wrap.

chasm/lib/callback/callback_suite_test.go
spkane31: Removed, using the `MockMutableContext` for these

chasm/lib/callback/invocation_tasks_test.go
spkane31: I was using the scheduler tests as a guide which were written before the `MockMutableContext` so I'm removing this and extending the previously existing test to assert on tasks added

chasm/lib/callback/invocation_tasks_test.go
spkane31: same as other comment

chasm/lib/callback/invocation_tasks_test.go
spkane31: Removed

chasm/lib/callback/statemachine.go
spkane31: Found a difference with how the retry policy was passed on a result, but otherwise I couldn't find anything different.

tests/callbacks_test.go
```diff
@@ -44,13 +44,29 @@ func (h *completionHandler) CompleteOperation(ctx context.Context, request *nexu
 
 type CallbacksSuite struct {
 	testcore.FunctionalTestBase
+
+	chasmEnabled bool
 }
 
 func TestCallbacksSuite(t *testing.T) {
```
bergundy: nit:

```suggestion
func TestCallbacksSuiteHSM(t *testing.T) {
```

tests/callbacks_migration_test.go
```diff
@@ -0,0 +1,544 @@
...
+	suite.Run(t, new(CallbacksMigrationSuite))
+}
+
+func (s *CallbacksMigrationSuite) SetupSuite() {
+	// Start with CHASM disabled for migration tests
+	s.SetupSuiteWithCluster(
+		testcore.WithDynamicConfigOverrides(map[dynamicconfig.Key]any{
+			dynamicconfig.EnableChasm.Key():          false,
+			dynamicconfig.EnableCHASMCallbacks.Key(): false,
+			dynamicconfig.EnableNexus.Key():          true,
```
bergundy: This is the default now, no need to set it anymore.

tests/callbacks_migration_test.go
```diff
@@ -0,0 +1,544 @@
...
+		}),
+	)
+}
+
+func (s *CallbacksMigrationSuite) SetupTest() {
```
bergundy, line 38: Looks like this already happens in `SetupSuite`?

tests/callbacks_migration_test.go
```diff
@@ -0,0 +1,544 @@
...
+	s.FunctionalTestBase.SetupTest()
+	// Start with CHASM disabled by default for migration tests
+	s.OverrideDynamicConfig(dynamicconfig.EnableChasm, false)
+	s.OverrideDynamicConfig(dynamicconfig.EnableCHASMCallbacks, false)
+	// Enable Nexus for callbacks
+	s.OverrideDynamicConfig(dynamicconfig.EnableNexus, true)
+}
+
+func (s *CallbacksMigrationSuite) TestWorkflowCallbacks_CHASM_Enabled_Mid_WF() {
```
bergundy, line 48: I'm not strictly opposed to adding this test but keep in mind that the value is very low and the cost is fairly high. Our functional test suite is already pretty big and we should be careful not to add tests that don't provide a lot of value.

My concern here is that we already know that both CHASM and HSM callbacks work and there is no logic that should prevent both from working.

tests/callbacks_migration_test.go
```diff
@@ -0,0 +1,544 @@
...
+		1 WorkflowExecutionStarted
+		2 WorkflowTaskScheduled
+		3 WorkflowTaskStarted
+		4 WorkflowTaskCompleted`,
+		s.GetHistoryFunc(s.Namespace().String(), workflowExecution),
+		5*time.Second,
+		10*time.Millisecond)
+
+	// Disable CHASM mid-workflow
+	s.OverrideDynamicConfig(dynamicconfig.EnableChasm, false)
```
bergundy: I would expect callbacks to only hang if we disable chasm entirely, which would be a destructive operation.
Otherwise, disabling CHASM callbacks should only prevent creation of new ones IMHO instead of causing the callbacks to be stuck.

tests/callbacks_migration_test.go
```diff
@@ -0,0 +1,544 @@
...
+		s.NoError(completion.Result.Consume(&callbackResult))
+		s.Equal(1, callbackResult)
+		// Acknowledge the callback
+		ch.requestCompleteCh <- nil
+	case <-time.After(10 * time.Second):
+		s.Fail("timeout waiting for callback")
+	}
+}
+
+func (s *CallbacksMigrationSuite) TestWorkflowCallbacks_MixedCallbacks() {
```
bergundy, line 314: This is also something that we can just test once manually and not have automated coverage for but not strongly opinionated against this to prevent regression.

service/history/workflow/util.go
```diff
@@ -290,9 +294,59 @@ func getCompletionCallbacksAsProtoSlice(ms historyi.MutableState) ([]*commonpb.C
 		}
 		result = append(result, cbSpec)
 	}
+
+	// Collect CHASM callbacks
+	if msImpl, ok := ms.(*MutableStateImpl); ok {
+		if msImpl.chasmEnabled() {
+			wf, ctx, err := msImpl.getChasmWorkflowComponent(context.Background())
```
bergundy: nit:

```suggestion
			wf, ctx, err := msImpl.chasmWorkflowComponent(context.Background())
```

common/dynamicconfig/constants.go
```diff
@@ -959,6 +959,11 @@ so forwarding by endpoint ID will not work out of the box.`,
 		32,
 		`MaxCallbacksPerWorkflow is the maximum number of callbacks that can be attached to a workflow.`,
 	)
+	MaxCHASMCallbacksPerWorkflow = NewNamespaceIntSetting(
```
bergundy, line 964: Maybe just add a comment saying that this is config is temporary? We will want to remove it and use `MaxCallbacksPerWorkflow` when CHASM is fully enabled.

common/dynamicconfig/constants.go
```diff
@@ -2638,6 +2643,13 @@ instead of the existing (V1) implementation.`,
 to the CHASM (V2) implementation on active scheduler workflows.`,
 	)
 
+	EnableCHASMCallbacks = NewNamespaceBoolSetting(
```
bergundy, line 2663: Being nitpicky here, but maybe put this config in the callback library and name it accordingly?

service/history/api/describeworkflow/api.go
```diff
@@ -235,38 +240,44 @@ func Invoke(
...
-	for _, node := range cbs {
-		callback, err := cbColl.Data(node.Key.ID)
+	// Collect callbacks from both CHASM and HSM sources
+	// A single execution could have callbacks in both places, so we need to check both
+	var callbackInfos []*workflowpb.CallbackInfo
+
+	// Check for CHASM callbacks (regardless of feature flag setting)
+	chasmTree := mutableState.ChasmTree()
+	// Only process CHASM callbacks if we have an actual chasm.Node (not a noopChasmTree)
+	if chasmNode, ok := chasmTree.(*chasm.Node); ok && chasmNode != nil {
```
bergundy: nit: We can export `chasmEnabled()` from mutable state instead and encapsulate this logic.

service/history/api/describeworkflow/api.go
```diff
@@ -368,6 +381,203 @@ func buildCallbackInfo(
...
+func buildCallbackInfosFromChasm(
+	ctx context.Context,
+	namespaceID namespace.ID,
+	mutableState historyi.MutableState,
+	executionInfo *persistencespb.WorkflowExecutionInfo,
+	executionState *persistencespb.WorkflowExecutionState,
+	outboundQueueCBPool *circuitbreakerpool.OutboundQueueCircuitBreakerPool,
+	logger log.Logger,
+) ([]*workflowpb.CallbackInfo, error) {
+	chasmTree := mutableState.ChasmTree()
```
bergundy: You could consider exporting `getChasmWorkflowComponent` and using that here.

service/history/api/describeworkflow/api.go
```diff
@@ -368,6 +381,203 @@ func buildCallbackInfo(
...
+			"expected workflow component but got different type",
+			tag.WorkflowNamespaceID(namespaceID.String()),
+			tag.WorkflowID(executionInfo.WorkflowId),
+			tag.WorkflowRunID(executionState.RunId),
+		)
+		return nil, serviceerror.NewInternal("failed to construct describe response")
+	}
+
+	result := make([]*workflowpb.CallbackInfo, 0, len(wf.Callbacks))
+	for _, field := range wf.Callbacks {
```
bergundy, line 453: nit: This logic could live in the `workflow` library.

service/history/workflow/mutable_state_impl.go
```diff
@@ -640,6 +642,21 @@ func (ms *MutableStateImpl) chasmEnabled() bool {
 	return !isNoop
 }
 
+// getChasmWorkflowComponent gets the root workflow component from the CHASM tree.
+// Returns both the workflow component and the CHASM mutable context.
+func (ms *MutableStateImpl) getChasmWorkflowComponent(ctx context.Context) (*chasmworkflow.Workflow, chasm.MutableContext, error) {
```
bergundy: Please don't use `get` for getters in Go.

```suggestion
func (ms *MutableStateImpl) chasmWorkflowComponent(ctx context.Context) (*chasmworkflow.Workflow, chasm.MutableContext, error) {
```

service/history/workflow/mutable_state_impl.go
```diff
@@ -2864,6 +2894,64 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
 	return nil
 }
 
+// addCompletionCallbacksChasm creates completion callbacks using the CHASM implementation.
+func (ms *MutableStateImpl) addCompletionCallbacksChasm(
```
bergundy, line 3033: nit: this could live in the `workflow` CHASM library.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2864,6 +2894,64 @@ func (ms *MutableStateImpl) addCompletionCallbacks(
...
+				return err
+			}
+		}
+
+		id := fmt.Sprintf("cb-%s-%d", requestID, idx)
+
+		// Create and add callback
+		callback := chasmcallback.NewCallback(requestID, event.EventTime, &callbackspb.CallbackState{}, chasmCB)
+		// Initialize the MSPointer field for accessing the underlying mutable state
+		callback.MSPointer = chasm.NewMSPointer(ms)
```
bergundy: The callback doesn't need a mutable state pointer.

chasm/lib/callback/component.go
```diff
@@ -29,8 +29,9 @@ type Callback struct {
 	// Persisted internal state
 	*callbackspb.CallbackState
 
-	// Interface to retrieve Nexus operation completion data
-	CompletionSource chasm.Field[CompletionSource]
+	// MSPointer is a runtime-only field for accessing the underlying mutable state to get completion data.
+	// Similar to MSPointer in the Workflow component, this is set at deserialization time.
+	chasm.MSPointer
```
bergundy: Why did you change this? We are supposed to use a pointer to the parent here, `MSPointer` is not supposed to be used outside of the `Workflow` component.

chasm/lib/callback/component.go
```diff
@@ -106,19 +102,24 @@ func (c *Callback) loadInvocationArgs(
...
+	switch r := input.result.(type) {
 	case invocationResultOK:
 		err := TransitionSucceeded.Apply(ctx, c, EventSucceeded{Time: ctx.Now(c)})
 		return nil, err
 	case invocationResultRetry:
 		err := TransitionAttemptFailed.Apply(ctx, c, EventAttemptFailed{
 			Time:        ctx.Now(c),
 			Err:         r.err,
-			RetryPolicy: r.retryPolicy,
+			RetryPolicy: input.config.RetryPolicy(),
```
bergundy: Would you pass in just the retry policy here instead of the whole config just so we minimize the dependencies and make it easier to unit test?

chasm/lib/callback/executors_test.go
```diff
@@ -91,28 +71,41 @@ func TestExecuteInvocationTaskNexus_Outcomes(t *testing.T) {
 				return nil, errors.New("fake failure")
 			},
 			expectedMetricOutcome: "unknown-error",
-			assertOutcome: func(t *testing.T, cb *Callback) {
+			assertOutcome: func(t *testing.T, cb *Callback, mctx *chasm.MockMutableContext) {
 				require.Equal(t, callbackspb.CALLBACK_STATUS_BACKING_OFF, cb.Status)
+				// Should generate a BackoffTask
+				require.Len(t, mctx.Tasks, 1)
+				require.IsType(t, &callbackspb.BackoffTask{}, mctx.Tasks[0].Payload)
+				backoffTask := mctx.Tasks[0].Payload.(*callbackspb.BackoffTask)
```
bergundy: Isn't this already tested in the state machine tests? Why is it important to cover here too?

chasm/lib/callback/executors_test.go
```diff
@@ -91,28 +71,41 @@ func TestExecuteInvocationTaskNexus_Outcomes(t *testing.T) {
...
+				require.Equal(t, cb.Attempt, backoffTask.Attempt)
 			},
+			expectedError: true,
```
bergundy: It would be better to have a function that would assert the error is of an expected type. You could add that to `assertOutcome`.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6126,6 +6227,31 @@ func (ms *MutableStateImpl) processCloseCallbacks() error {
 	return nil
 }
 
+// processCloseCallbacksChasm triggers "WorkflowClosed" callbacks using the CHASM implementation.
+func (ms *MutableStateImpl) processCloseCallbacksChasm() error {
```
bergundy, line 6339: I would move all of the code that uses the `Workflow` component into the CHASM `workflow` library.

common/dynamicconfig/constants.go
bergundy: I kept going back and forth on whether we need a separate config or not but I figure this would be easier to rollout.

tests/callbacks_migration_test.go
bergundy: Maybe just add a TODO to remove this test once CHASM callbacks are the default?

tests/callbacks_migration_test.go
spkane31: removed `SetupSuite` and kept `SetupTest` to make sure things are reset at test creation

service/history/api/describeworkflow/api.go
spkane31: adding `ChasmWorkflowComponent` for write ops and `ChasmWorkflowComponentReadOnly` for read only ops

service/history/api/describeworkflow/api.go
spkane31: I disagree, this is transforming the `Callback` into a protobuf struct which isn't relevant to the `workflow` library.

common/dynamicconfig/constants.go
spkane31: I think it makes the most sense to have this in the same place as scheduler dynamic configs. additionally this will be removed once this is fully rolled out right?

chasm/lib/callback/component.go
spkane31: Reverted.

chasm/lib/workflow/workflow.go
```diff
@@ -43,3 +52,92 @@ func (w *Workflow) LifecycleState(
...
+	return nil
+}
+
+func (w *Workflow) GetNexusCompletion(
+	ctx chasm.Context,
+	requestID string,
+) (nexusrpc.OperationCompletion, error) {
+	// Retrieve the completion data from the underlying mutable state via MSPointer
+	return w.MSPointer.GetNexusCompletion(ctx, requestID)
+}
```
spkane31, line 128: Is this how we want to embed a field, with `chasm.Field(chasm.ComponentPointerTo(ctx, C))` ? This was the best way I could find out to do this and included adding the `GetNexusCompletion` function to Workflow because there are errors when trying to use `chasm.ComponentPointerTo(ctx, w.MSPointer)`? @bergundy

tests/callbacks_migration_test.go
spkane31: In favor of keeping this, but adding a TODO to remove it in the future

chasm/lib/workflow/workflow.go
bergundy: Yes, the pointer should be to the `Workflow` component.

yycptt: I will be working a ParentPointer and will update this later to use that. But for now yeah, create a normal pointer to the Workflow component.

service/history/workflow/util.go
```diff
@@ -290,6 +295,38 @@ func getCompletionCallbacksAsProtoSlice(ms historyi.MutableState) ([]*commonpb.C
 		}
 		result = append(result, cbSpec)
 	}
+
+	// Collect CHASM callbacks
+	if msImpl, ok := ms.(*MutableStateImpl); ok {
```
bergundy: There is no need to cast here.

service/history/workflow/util.go
```diff
@@ -290,6 +295,38 @@ func getCompletionCallbacksAsProtoSlice(ms historyi.MutableState) ([]*commonpb.C
...
+		if msImpl.ChasmEnabled() {
+			component, ctx, err := msImpl.ChasmWorkflowComponent(context.Background())
```
bergundy: You should have a context to pass into this function.

service/history/workflow/util.go
```diff
@@ -290,6 +295,38 @@ func getCompletionCallbacksAsProtoSlice(ms historyi.MutableState) ([]*commonpb.C
...
+			if err != nil {
+				return nil, err
+			}
+			wf, ok := component.(*chasmworkflow.Workflow)
```
bergundy: I was expecting that `ChasmWorkflowComponent` would already cast for you.

service/history/workflow/util.go
```diff
@@ -290,6 +295,38 @@ func getCompletionCallbacksAsProtoSlice(ms historyi.MutableState) ([]*commonpb.C
...
+			if !ok {
+				return nil, serviceerror.NewInternalf("expected workflow component, but got %T", component)
+			}
+
+			for _, field := range wf.Callbacks {
```
bergundy: I would make this a method of `chasmworkflow.Workflow`, not blocking.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6106,6 +6172,22 @@ func (ms *MutableStateImpl) processCloseCallbacks() error {
 		return nil
 	}
 
+	// Process CHASM callbacks if CHASM is enabled. Note that we check ChasmEnabled() rather than
+	// chasmCallbacksEnabled() to ensure that callbacks created when both CHASM and CHASM callbacks
```
```suggestion
	// chasmCallbacksEnabled() to ensure that callbacks created when both HSM and CHASM callbacks
```

service/history/workflow/mutable_state_impl.go
```diff
@@ -631,15 +631,49 @@ func (ms *MutableStateImpl) ChasmTree() historyi.ChasmTree {
...
+// Returns both the workflow component and a read-only CHASM context.
+// This method is for read-only operations.
+func (ms *MutableStateImpl) ChasmWorkflowComponentReadOnly(ctx context.Context) (chasm.Component, chasm.Context, error) {
+	chasmCtx := chasm.NewContext(ctx, ms.chasmTree.(*chasm.Node))
+	rootComponent, err := ms.chasmTree.ComponentByPath(chasmCtx, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	return rootComponent, chasmCtx, nil
+}
```
bergundy: Cast to `*chasmworkflow.Workflow` in both of these here instead of doing that in all of the callsites.

service/history/workflow/util.go
bergundy:
```suggestion
			component, ctx, err := msImpl.ChasmWorkflowComponentReadOnly(context.Background())
```

service/history/api/describeworkflow/api.go
bergundy: Eventually we want all API handler to live in their corresponding libraries. Not blocking though, we will get there eventually.

service/history/api/describeworkflow/api.go
```diff
@@ -235,38 +239,43 @@ func Invoke(
...
 	}
 
-	cbColl := callbacks.MachineCollection(mutableState.HSM())
-	cbs := cbColl.List()
-	result.Callbacks = make([]*workflowpb.CallbackInfo, 0, len(cbs))
-	for _, node := range cbs {
-		callback, err := cbColl.Data(node.Key.ID)
+	// Collect callbacks from both CHASM and HSM sources
+	// A single execution could have callbacks in both places, so we need to check both
+	var callbackInfos []*workflowpb.CallbackInfo
```
bergundy: Technically you could append directly into `result.Callbacks` but 🤷

common/dynamicconfig/constants.go
bergundy: Not critical.

chasm/lib/workflow/workflow.go
```diff
@@ -43,3 +52,92 @@ func (w *Workflow) LifecycleState(
...
+	}
+	return nil
+}
+
+// AddCompletionCallbacks creates completion callbacks using the CHASM implementation.
+// maxCallbacksPerWorkflow is the configured maximum number of callbacks allowed per workflow.
+// msPointer is used to initialize the MSPointer field for accessing the underlying mutable state.
+func (w *Workflow) AddCompletionCallbacks(
+	ctx chasm.MutableContext,
+	event *historypb.HistoryEvent,
```
bergundy: This is fine but you could just pass in the time since the rest of the event isn't used and less dependencies will make it easier to test.

chasm/lib/workflow/workflow.go
```diff
@@ -43,3 +52,92 @@ func (w *Workflow) LifecycleState(
...
+	requestID string,
+	completionCallbacks []*commonpb.Callback,
+	maxCallbacksPerWorkflow int,
+	msPointer chasm.MSPointer,
```
bergundy: Unused and shouldn't be part of the contract.

chasm/lib/workflow/workflow.go
```diff
@@ -43,3 +52,92 @@ func (w *Workflow) LifecycleState(
...
+					Url:    variant.Nexus.GetUrl(),
+					Header: variant.Nexus.GetHeader(),
+				},
+			}
+		case *commonpb.Callback_Internal_:
+			err := proto.Unmarshal(cb.GetInternal().GetData(), chasmCB)
+			if err != nil {
+				return err
+			}
+		}
```
bergundy: This isn't supported anymore, remove this code and fail if the variant is not supported.

chasm/lib/workflow/workflow.go
bergundy, line 1: Makes me really happy to start seeing this library populated.

chasm/lib/callback/library.go
```diff
@@ -9,16 +9,19 @@ type (
 	Library struct {
 		chasm.UnimplementedLibrary
 
+		Config                 *Config
```
bergundy: Unused.

chasm/lib/callback/fx.go
bergundy: I missed this comment in the last review, had this file collapsed.

The idea is that CHASM shouldn't depend on HSM. The implementation should be completely separate even if it means that the injection needs to happen twice, once for each interface.

chasm/lib/callback/executors_test.go
```diff
@@ -461,9 +456,10 @@ func TestExecuteInvocationTaskChasm_Outcomes(t *testing.T) {
 				require.NoError(t, err)
 				return comp
 			}(),
-			headerValue:          encodedRef,
-			expectsInternalError: true,
-			assertOutcome: func(t *testing.T, cb *Callback) {
+			headerValue: encodedRef,
+			assertOutcome: func(t *testing.T, cb *Callback, err error) {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "internal error, reference-id:")
```
bergundy: This can be shortened to `require.ErrorContains` but ideally you would also check the error type if a concrete type is relevant and available (didn't check).

chasm/lib/callback/executors_test.go
```diff
@@ -91,7 +92,10 @@ func TestExecuteInvocationTaskNexus_Outcomes(t *testing.T) {
 				return nil, errors.New("fake failure")
 			},
 			expectedMetricOutcome: "unknown-error",
-			assertOutcome: func(t *testing.T, cb *Callback) {
+			assertOutcome: func(t *testing.T, cb *Callback, err error) {
+				require.Error(t, err)
```
bergundy: Redundant if you're already checking ErrorAs below.

bergundy (commented):
Most of this last round was very minor comments. I think we're very close.

components/callbacks/metrics.go
```diff
@@ -1,12 +1,10 @@
...
+import (
+	chasmcallbacks "go.temporal.io/server/chasm/lib/callback"
 )
-var RequestLatencyHistogram = metrics.NewTimerDef(
-	"callback_outbound_latency",
-	metrics.WithDescription("Latency histogram of outbound callback requests made by the history service."),
+
+var (
+	RequestCounter          = chasmcallbacks.RequestCounter
+	RequestLatencyHistogram = chasmcallbacks.RequestLatencyHistogram
```
spkane31, line 9: This removes any import cycles and allows for the same metrics so when we do remove the HSM impl we should not need to make changes to the chasm impl

chasm/lib/callback/executors.go
```diff
@@ -7,19 +7,23 @@ import (
...
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/namespace"
 	commonnexus "go.temporal.io/server/common/nexus"
 	"go.temporal.io/server/common/resource"
-	"go.temporal.io/server/service/history/queues"
 	"go.uber.org/fx"
 )
 
+// NamespaceIDAndDestination is a key for grouping callbacks by namespace and destination.
+type NamespaceIDAndDestination struct {
```
bergundy: Please don't redefine this here. Import from queues instead.

service/history/queues/errors/errors.go
bergundy, line 1: This should not be redefined.

chasm/lib/workflow/workflow.go
```diff
@@ -37,3 +45,84 @@ func (w *Workflow) LifecycleState(
...
+				Nexus: &callbackspb.Callback_Nexus{
+					Url:    variant.Nexus.GetUrl(),
+					Header: variant.Nexus.GetHeader(),
+				},
+			}
+		default:
+			return fmt.Errorf("unsupported callback variant: %T", variant)
+		}
+
+		id := fmt.Sprintf("cb-%s-%d", requestID, idx)
```
bergundy: You can save a couple of bytes on the prefix.

```suggestion
		id := fmt.Sprintf("%s-%d", requestID, idx)
```

service/history/workflow/util.go
```diff
@@ -318,3 +346,26 @@ func PersistenceCallbackToAPICallback(cb *persistencespb.Callback) (*commonpb.Ca
 	}
 	return res, nil
 }
+
+// ChasmCallbackToAPICallback converts a CHASM callback proto to API callback proto.
+func ChasmCallbackToAPICallback(cb *chasmcallback.Callback) (*commonpb.Callback, error) {
```
bergundy: nit: this can 100% be a method on `chasmcallback.Callback`.

chasm/lib/callback/executors.go
spkane31: Moved to it's own package

spkane31:
cursor review

chasm/lib/callback/statemachine.go
```diff
@@ -18,7 +20,11 @@ var TransitionScheduled = chasm.NewTransition(
 	[]callbackspb.CallbackStatus{callbackspb.CALLBACK_STATUS_STANDBY},
 	callbackspb.CALLBACK_STATUS_SCHEDULED,
 	func(cb *Callback, ctx chasm.MutableContext, event EventScheduled) error {
-		ctx.AddTask(cb, chasm.TaskAttributes{}, &callbackspb.InvocationTask{})
+		u, err := url.Parse(cb.Callback.GetNexus().Url)
```
bergundy: It would be safer to use which does nil checks:


```suggestion
		u, err := url.Parse(cb.Callback.GetNexus().GetUrl())
```

PR #8585, opened 2025-11-03
Fix leaking and modifying history event cache objects

rodrigozhou:
## What changed?
Alias search attributes of WorkflowStartedEvent in the eager workflow task in the StartWorkflowResponse in frontend instead of history.

## Why?
Modifying in history modifies the object in history cache, which breaks the retry logic.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -408,6 +408,25 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 	if err != nil {
 		return nil, err
 	}
+	return wh.convertToStartWorkflowExecutionResponse(resp, namespaceName)
+}
+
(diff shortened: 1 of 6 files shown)


bergundy (approved):
Thanks! Can you also add a functional test that would fail without this fix to ensure this works properly?

rodrigozhou:
@bergundy A functional to verify this is not simple, need to do some refactoring to untangle some search attributes registration between ES and SQL databases.

The existing functional tests already verifies that the frontend returns the aliased search attributes, and I've added a unit test in history that verifies that it returns unaliased search attributes.

I'll follow up with another PR to refactor and add more tests to verify this.

PR #8587, opened 2025-11-04
Drop repeated workflow task failures

yiminc:
## What changed?
Drop repeated workflow task failure reports

## Why?
To avoid busy loop of repeated workflow task failures

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
No


common/dynamicconfig/constants.go
```diff
@@ -2250,6 +2250,11 @@ the number of children greater than or equal to this threshold`,
 		10,
 		`If speculative workflow task shipped more than DiscardSpeculativeWorkflowTaskMaximumEventsCount events, it can't be discarded`,
 	)
+	EnableDropRepeatedWorkflowTaskFailures = NewNamespaceBoolSetting(
```
yycptt, line 2253: I was thinking an attempt count threshold, but yeah this works as well.

PR #8588, opened 2025-11-04
CHASM: Store component and task type ID instead of name

yycptt:
## What changed?
- Introduce typeID for registered CHASM components and tasks
- In longer term the relationship between component/task name and ID should be persisted (similar to namespace)

## Why?
- Store less data
- Decouple user visible name from internal implementation, making it possible in the future to rename components/tasks
- Required for separate ID space work, where archetype information needs to be stored in all history tasks as well.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/visibility/store/sql/visibility_store_test.go
+++ b/common/persistence/visibility/store/sql/visibility_store_test.go
@@ -72,7 +72,7 @@ func TestBuildQueryParams(t *testing.T) {
 					testNamespaceID,
 					tc.query,
 					sqlQC,
-					searchattribute.TestNameTypeMap,
+					searchattribute.TestNameTypeMap(),
 					&searchattribute.TestMapper{},
 				)
 				if tc.err != "" {


chasm/lib/workflow/library.go
```diff
@@ -13,11 +13,11 @@ func NewLibrary() *Library {
...
 
 func (l *Library) Name() string {
-	return "workflow"
+	return chasm.WorkflowLibraryName
 }
 
 func (l *Library) Components() []*chasm.RegistrableComponent {
 	return []*chasm.RegistrableComponent{
-		chasm.NewRegistrableComponent[*Workflow]("Workflow"),
+		chasm.NewRegistrableComponent[*Workflow](chasm.WorkflowComponentName),
```
bergundy, line 21: Side note: We should standardize on the naming convention for component and library names.

chasm/archetype.go
```diff
@@ -1,12 +1,13 @@
 package chasm
 
-type Archetype string
+// Archetype is the fully qualified name of the root component of a CHASM execution.
+type Archetype = string
```
bergundy, line 4: Should we rename this to `ArchetypeName`? I'm totally okay not doing that FTR.

chasm/export_test.go
bergundy, line 1: I don't understand the purpose of this file and why it's named `export_test` where it actually just adds exported variants of existing registry methods.

chasm/registrable_component.go
```diff
@@ -4,20 +4,21 @@ import (
 	"fmt"
 	"reflect"
 
-	enumspb "go.temporal.io/api/enums/v1"
+	"github.com/dgryski/go-farm"
```
bergundy, line 7: Wondering if it'd be worth using https://pkg.go.dev/hash/fnv from the standard library vs. using an external dependency. I'm not concerned with speed here and the distribution should be okay for us. Even crc should be okay.

chasm/tree.go
```diff
@@ -224,7 +226,9 @@ func NewTree(
 	logger log.Logger,
 ) (*Node, error) {
 	if len(serializedNodes) == 0 {
-		return NewEmptyTree(registry, timeSource, backend, pathEncoder, logger), nil
+		root := NewEmptyTree(registry, timeSource, backend, pathEncoder, logger)
+		root.serializedNode.Metadata.GetComponentAttributes().TypeId = WorkflowArchetypeID
```
bergundy, line 232: Is it important to do a nil check here for ComponentAttributes? The getter isn't protecting you here.

chasm/tree.go
```diff
@@ -2347,15 +2330,23 @@ func (n *Node) Terminate(
 	return nil
 }
 
-func (n *Node) Archetype() Archetype {
-	root := n.root()
-	if root.serializedNode == nil {
-		// Empty tree
-		return ""
+func (n *Node) ArchetypeID() ArchetypeID {
```
bergundy, line 2360: Please document exported methods.

chasm/tree.go
```diff
@@ -2347,15 +2330,23 @@ func (n *Node) Terminate(
...
+	// Root must be a component.
+	return n.root().serializedNode.Metadata.GetComponentAttributes().GetTypeId()
+}
+
+func (n *Node) Archetype() (Archetype, error) {
```
bergundy, line 2367: It's worth noting in the docstring that this involves a map lookup and isn't a "cheap" operation.

chasm/export_test.go
yycptt: Basically to export a set of methods for unit tests in chasm_test package and also at the same time prevent chasm package users from seeing and using them.

Though I agree in this case we can just not use the chasm_test package and call those private methods directly, but don't really want to make that change part of this PR.

chasm/registrable_component.go
yycptt: Guess for now I still want to stick to go-farm which is used throughout the codebase for hashing purpose and avoid an exception.

chasm/tree.go
yycptt: This is fine. NewEmptyTree takes care of initializing the serializedNode field of the root to be a valid ComponentNode. I can add a comment for this.

chasm/tree.go
yycptt: Added a comment, and this method will be removed in next PR.

chasm/tree.go
yycptt: Done.

chasm/lib/workflow/library.go
yycptt: 👍 renamed to `workflow` (lower case) for now.

chasm/archetype.go
yycptt: yeah.. I was also debating on this with myself.. On one hand using xxxName and xxxID is similar to what we do for namespace. But on the other hand, we used the word Archetype a lot in our conversation and I want the same name in the code as well. 

hmmm guess I will keep it as is since I also don't see a strong reason to change it.

yycptt:
>Can you just make sure that we are tracking the open TODOs somewhere?

Most TODOs will actually be resolved in my next PR for separateID space. But yeah I have created tasks for other TODOs like deprecating ShardingKey.

service/history/visibility_queue_task_executor.go
```diff
@@ -446,7 +447,9 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
 		memo,
 		searchattributes,
 	)
-	requestBase.SearchAttributes.IndexedFields[searchattribute.TemporalNamespaceDivision] = payload.EncodeString(tree.Archetype().String())
+
+	// We reuses the TemporalNamespaceDivision column to store the string representation of ArchetypeID.
+	requestBase.SearchAttributes.IndexedFields[searchattribute.TemporalNamespaceDivision] = payload.EncodeString(strconv.FormatUint(uint64(tree.ArchetypeID()), 10))
```
rodrigozhou, line 452: LGTM!

chasm/archetype.go
bergundy: SGTM.

chasm/registrable_component.go
bergundy: I guess we're already married to this library. My concern is the low activity in this repo and us using it for IDs that would be annoying to change later.

chasm/tree.go
bergundy: This comment is really a nit, since the getter isn't protecting you but it doesn't hurt either.

service/history/visibility_queue_task_executor.go
```diff
@@ -446,7 +447,9 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
 		memo,
 		searchattributes,
 	)
-	requestBase.SearchAttributes.IndexedFields[searchattribute.TemporalNamespaceDivision] = payload.EncodeString(tree.Archetype().String())
+
+	// We reuses the TemporalNamespaceDivision column to store the string representation of ArchetypeID.
```
bergundy: typo:

```suggestion
	// We reuse the TemporalNamespaceDivision column to store the string representation of ArchetypeID.
```

bergundy (approved):

PR #8590, opened 2025-11-04
Refactor searchattribute.TestNameTypeMap

rodrigozhou:
## What changed?
Split `searchattribute.TestNameTypeMap` in two: one for ES setup, and one for SQL DB setup.

## Why?
`searchattribute.TestNameTypeMap` is used in tests as the predefined list of custom search attributes. However, it mixes how custom search attributes work differently in Elasticsearch vs SQL databases. This PR splits the variable in two (one for each) so the tests will be a better representation of the actual system.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/visibility/store/elasticsearch/query_interceptors_test.go
+++ b/common/persistence/visibility/store/elasticsearch/query_interceptors_test.go
@@ -35,7 +35,7 @@ func (s *QueryInterceptorSuite) TearDownTest() {
 func (s *QueryInterceptorSuite) TestTimeProcessFunc() {
 	vi := NewValuesInterceptor(
 		"test-namespace",
-		searchattribute.TestNameTypeMap,
+		searchattribute.TestEsNameTypeMap(),
 	)
 
 	cases := []struct {
@@ -72,7 +72,7 @@ func (s *QueryInterceptorSuite) TestTimeProcessFunc() {
 func (s *QueryInterceptorSuite) TestStatusProcessFunc() {
 	vi := NewValuesInterceptor(
 		"test-namespace",
-		searchattribute.TestNameTypeMap,
+		searchattribute.TestEsNameTypeMap(),
 	)
 
 	cases := []struct {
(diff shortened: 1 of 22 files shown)


common/persistence/visibility/store/elasticsearch/query_interceptors_test.go
```diff
@@ -35,7 +35,7 @@ func (s *QueryInterceptorSuite) TearDownTest() {
 func (s *QueryInterceptorSuite) TestTimeProcessFunc() {
 	vi := NewValuesInterceptor(
 		"test-namespace",
-		searchattribute.TestNameTypeMap,
+		searchattribute.TestEsNameTypeMap(),
```
awln-temporal, line 38: can we add this when setting up the suite instead of asking for the name type map on each test case?

common/searchattribute/test_provider.go
```diff
@@ -78,12 +46,41 @@ var (
...
+	return res
+}
+
 func NewTestProvider() *TestProvider {
 	return &TestProvider{}
 }
 
+func NewTestEsProvider() *TestProvider {
+	return &TestProvider{
+		es: true,
```
awln-temporal, line 79: nit: any reason we want to switch on a boolean? When we support ClickHouse Visibility Stores and more, how would the current test provider work? Imo it's fine to just split into different implementations.

awln-temporal (approved):
Would these changes be needed once we merge the OSS ElasticSearch search attribute registration path with SQL?

common/persistence/visibility/store/elasticsearch/query_interceptors_test.go
rodrigozhou: Maybe, but this code will go away soon after we move to the unified query converter.

common/searchattribute/test_provider.go
rodrigozhou: This will go away after we move ES to use mappers as well.

rodrigozhou:
@awln-temporal Somewhat, yes. The map was working for both ES and SQL, so this PR makes a separation between them.

PR #8592, opened 2025-11-04
VLN-444: Use first-party action for GitHub app tokens

picatz:
## Summary

- `.github/workflows/create-tag.yml`: Set both checkout steps to disable credential persistence while continuing to use the generated app token.
- `.github/workflows/flaky-tests-report.yml`: Checkout now uses the generated app token and turns off persistent credentials.
- `.github/workflows/trigger-publish.yml`: Added explicit Temporal organization owner and repository scope so the app token can dispatch to docker-builds.
- `.github/workflows/trigger-version-info-service.yml`: Formatted the repository scope as a newline-delimited list to match the updated token requirements.

---

Previous summary:

## Summary

- `.github/workflows/trigger-publish.yml`: Swapped tibdex/github-app-token for actions/create-github-app-token@v2 and translated inputs to new kebab-case fields to keep workflow functional and remove deprecated action usage.

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
         with:
-          persist-credentials: false
+          persist-credentials: true
           token: ${{ steps.generate_token.outputs.token }}
           ref: ${{ github.event.inputs.branch }}
           fetch-depth: 0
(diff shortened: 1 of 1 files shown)


picatz:
☝️ Should be resolved with https://github.com/temporalio/temporal/pull/8565

.github/workflows/create-tag.yml
simvlad, line 127: Where does this get the token if we now have `persist-credentials: false`?

picatz: Should be resolved with https://github.com/temporalio/temporal/pull/8592/commits/750e7312d1eb928625758b23a48a04fa7585127f

PR #8595, opened 2025-11-04
Check for an existing CHASM component before writing Workflow as component root

lina-temporal:
# What changed?
- Only add the Workflow root CHASM component if no root component already exists.
- I think the overall feature should probably also be behind a feature flag.

## Why?
- CHASM Scheduler functional tests are broken without this fix (scheduler mutable states 'become' chasmworkflows)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -373,7 +373,8 @@ func (n *Node) Component(
 		return nil, errComponentNotFound
 	}
 
-	if err := node.prepareComponentValue(chasmContext); err != nil {
+	validationContext := NewContext(chasmContext.getContext(), node)
+	if err := node.prepareComponentValue(validationContext); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -2597,8 +2597,14 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 		// Initialize chasm tree once for new workflows.
 		// Using context.Background() because this is done outside an actual request context and the
 		// chasmworkflow.NewWorkflow does not actually use it currently.
-		mutableContext := chasm.NewMutableContext(context.Background(), ms.chasmTree.(*chasm.Node))
-		ms.chasmTree.(*chasm.Node).SetRootComponent(chasmworkflow.NewWorkflow(mutableContext, ms))
+		root, ok := ms.chasmTree.(*chasm.Node)
+		softassert.That(ms.logger, ok, "chasmTree cast failed")
+
+		mutableContext := chasm.NewMutableContext(context.Background(), root)
+		_, err := root.Component(mutableContext, chasm.ComponentRef{})
```
yycptt: I think we should use immutable context first to do the check.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2597,8 +2597,15 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
 		// Using context.Background() because this is done outside an actual request context and the
 		// chasmworkflow.NewWorkflow does not actually use it currently.
-		mutableContext := chasm.NewMutableContext(context.Background(), ms.chasmTree.(*chasm.Node))
-		ms.chasmTree.(*chasm.Node).SetRootComponent(chasmworkflow.NewWorkflow(mutableContext, ms))
+		root, ok := ms.chasmTree.(*chasm.Node)
+		softassert.That(ms.logger, ok, "chasmTree cast failed")
+
+		validationContext := chasm.NewContext(context.Background(), root)
+		_, err := root.Component(validationContext, chasm.ComponentRef{})
+		if common.IsNotFoundError(err) {
```
yycptt, line 2605: it returns NotFound due to which check? Initial versioned transition check? 
I thought it kinda like a undefined behavior. But if it works I don't have a objection. we probably want a test to make sure this always work though.

yycptt (approved):
+1 for adding a separate feature flag for workflow chasm node.

PR #8596, opened 2025-11-04
[Nexus] Allow temporal://internal URLs to validate

lina-temporal:
# What changed?
- Allows `temporal://internal` URLs to validate.

## Why?
- CHASM uses this URL for Nexus callback completions.

## Potential risks
- Risk is mitigated in Cloud by denying non-Admin callers.


bergundy (approved):
Thanks!

PR #8603, opened 2025-11-06
Implement rpc handler to record worker heartbeat.

rkannan82:
## What changed?
Implement the rpc handler that creates/updates the worker chasm component with the new heartbeat.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None. Feature under developement.


chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			return w, &workerstatepb.RecordHeartbeatResponse{}, nil
+		},
+		func(w *Worker, ctx chasm.MutableContext, heartbeat *HeartbeatRequest) (*workerstatepb.RecordHeartbeatResponse, error) {
+			// Update existing worker with new heartbeat
+			err := w.RecordHeartbeat(ctx, heartbeat.WorkerHeartbeat, heartbeat.LeaseDuration)
+			if err != nil {
+				return nil, err
+			}
+			return &workerstatepb.RecordHeartbeatResponse{}, nil
```
bergundy: nit: if you make `w.RecordHeartbeat` take `*RecordHeartbeatRequest` and return `*RecordHeartbeatResponse, error`, you can remove this closure.

chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// Get lease duration from request
+	leaseDuration := req.LeaseDuration.AsDuration()
+	if leaseDuration <= 0 {
+		leaseDuration = defaultLeaseDuration
+	}
+
+	// Try to update existing worker, or create new one if it doesn't exist
+	_, _, _, _, err := chasm.UpdateWithNewEntity(
```
bergundy: Assuming that you intend to always allow ID reuse and fail on conflict if already running.

chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,77 @@
...
+			WorkerHeartbeat: req.WorkerHeartbeat,
+			LeaseDuration:   leaseDuration,
+		},
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return &workerstatepb.RecordHeartbeatResponse{}, nil
```
bergundy: You're going to want to return a ref here to be used in the following heartbeat request to be able to detect stale state.

chasm/lib/worker/handler_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"go.temporal.io/server/common/testing/testvars"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+const (
+	testWorkerInstanceKey = "test-worker-1"
+)
+
+func TestRecordHeartbeatHandler(t *testing.T) {
```
bergundy: Make this a functional test, all of these mocks don't prove that this API actually works.

chasm/lib/worker/worker.go
```diff
@@ -15,6 +15,13 @@ const (
 	Archetype chasm.Archetype = "Worker"
 )
 
+// HeartbeatRequest is the internal request for CHASM component updates
+type HeartbeatRequest struct {
```
bergundy: Redundant, just use the proto.

chasm/lib/worker/handler.go
rkannan82: Ack. Is it ok if I do this in separate PR where I start working on activities? I believe the ordering semantics are relevant then.

chasm/lib/worker/handler.go
rkannan82: The behavior we want is:
- If worker (process) restarts, it should get a new ID.
- As long as the id is same, we should just keep updating it.
What is this **conflict-if-already-running** semantics?

chasm/lib/worker/handler.go
rkannan82: Done. Interesting go shorthands :)

rkannan82 (commented):
Please take another look.

chasm/lib/worker/handler_test.go
rkannan82: Right this test is useless. I have deleted it and added a tests/worker_heartbeat_test.go.

chasm/lib/worker/handler.go
rkannan82: Oh by the way, UpdateWithInsert is not implemented yet. So I ended up using the individual methods for Update and Insert for now.

chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,68 @@
...
+type handler struct {
+	workerstatepb.UnimplementedWorkerServiceServer
+}
+
+func newHandler() *handler {
+	return &handler{}
+}
+
+func (h *handler) RecordHeartbeat(ctx context.Context, req *workerstatepb.RecordHeartbeatRequest) (*workerstatepb.RecordHeartbeatResponse, error) {
+	// Validate that exactly one worker heartbeat is present
```
bergundy, line 26: Why only 1? And why do we allow more than one in the API?

chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,68 @@
...
+
+	workerHeartbeat := frontendReq.GetWorkerHeartbeat()[0]
+
+	// Try to update existing worker first
+	resp, _, err := chasm.UpdateComponent(
+		ctx,
+		chasm.NewComponentRef[*Worker](
+			chasm.EntityKey{
+				NamespaceID: req.NamespaceId,
+				BusinessID:  workerHeartbeat.WorkerInstanceKey,
```
bergundy: I think this is fine for now but you may need to consider situations where there are multiple runs for the same workflow instance key. This could happen if a worker is resurrected after a long time. Not really expected but you never know. You could consider returning a chasm ref (`[]byte`) back to the worker so it can be used for staleness checks.

chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,68 @@
...
+	// Validate that exactly one worker heartbeat is present
+	frontendReq := req.GetFrontendRequest()
+	if frontendReq == nil || len(frontendReq.GetWorkerHeartbeat()) != 1 {
+		return nil, fmt.Errorf("exactly one worker heartbeat must be present in the request")
+	}
+
+	workerHeartbeat := frontendReq.GetWorkerHeartbeat()[0]
+
+	// Try to update existing worker first
+	resp, _, err := chasm.UpdateComponent(
```
bergundy: You will want to do this transactionally with `UpdateWithNewExecution` (note we renamed Entity to Execution in main).

chasm/lib/worker/handler.go
```diff
@@ -0,0 +1,68 @@
...
+					w := NewWorker()
+					resp, err := w.recordHeartbeat(ctx, req)
+					return w, resp, err
+				},
+				req,
+			)
+			if err != nil {
+				return nil, err
+			}
+			return resp, nil
```
bergundy: nit:

```suggestion
			return resp, err
```

tests/worker_heartbeat_test.go
```diff
@@ -0,0 +1,162 @@
...
+				WorkerInstanceKey: workerInstanceKey,
+			},
+		},
+	})
+
+	require.NoError(t, err)
+	require.NotNil(t, resp)
+
+	// Verify the worker entity was created in CHASM
+	engineCtx := chasm.NewEngineContext(ctx, s.chasmEngine)
```
bergundy, line 67: Isn't there some describe worker API that you can implement on the frontend? Ideally you wouldn't read internal state in functional tests which are supposed to be more of a blackbox.

We use the chasm API for chasm functional tests for sanity but that's not a pattern we encourage in general.

tests/worker_heartbeat_test.go
```diff
@@ -0,0 +1,162 @@
...
+	worker2, err := chasm.ReadComponent(
+		engineCtx,
+		chasm.NewComponentRef[*chasmworker.Worker](
+			chasm.EntityKey{
+				NamespaceID: s.NamespaceID().String(),
+				BusinessID:  workerInstanceKey,
+			},
+		),
+		func(w *chasmworker.Worker, _ chasm.Context, _ any) (*chasmworker.Worker, error) {
+			return w, nil
```
bergundy, line 149: Avoid the data race:

```suggestion
			return w.Clone(), nil
```

PR #8604, opened 2025-11-06
Grouping replication task by workflow Id

xwduan:
## What changed?
Grouping replication task by workflow Id for high priority replication task
## Why?
Reduce the lock contention issue when there is a hot workflow

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
n/a

### DIFF
--- a/service/history/replication/fx.go
+++ b/service/history/replication/fx.go
@@ -279,7 +279,12 @@ func sequentialTaskQueueFactoryProvider(
 		if config.EnableReplicationTaskBatching() {
 			return NewSequentialBatchableTaskQueue(task, nil, logger, metricsHandler)
 		}
-		return NewSequentialTaskQueue(task)
+		item := task.QueueID()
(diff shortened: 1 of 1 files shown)


service/history/replication/fx.go
```diff
@@ -279,7 +279,12 @@ func sequentialTaskQueueFactoryProvider(
 		if config.EnableReplicationTaskBatching() {
 			return NewSequentialBatchableTaskQueue(task, nil, logger, metricsHandler)
```
yux0, line 280: Also update this with customized ID? I am not sure if batching support cross execution

xwduan: Batching does not support cross execution. Current temporal implementation only support two update mode:
1. Update a workflow
2. Update a workflow and create a new workflow (continueAsNew or Reset).

yux0 (approved):

PR #8607, opened 2025-11-07
Gate activity retry stamp behind feature flag

michaely520:
## What changed?
- Gating stamp increment on activity retries from https://github.com/temporalio/temporal/pull/8536/files behind a feature flag
- Returning err when dropping stale activities on passive so we have metrics emitted
 
## Why?
- We need stamp increments to only occur after the change from the PR has fully rolled out to all clusters, else we risk compatibility issues and dropped tasks. 

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -188,10 +188,10 @@ config as the other services.`,
 		false,
 		`EnableActivityEagerExecution indicates if activity eager execution is enabled per namespace`,
 	)
-	EnableActivityRetryStampIncrement = NewNamespaceBoolSetting(
+	EnableActivityRetryStampIncrement = NewGlobalBoolSetting(
 		"system.enableActivityRetryStampIncrement",
 		false,
-		`EnableActivityRetryStampIncrement indicates if activity retry stamp increment is enabled per namespace`,
+		`EnableActivityRetryStampIncrement indicates if activity retry stamp increment is enabled`,
 	)
 	EnableEagerWorkflowStart = NewNamespaceBoolSetting(
 		"system.enableEagerWorkflowStart",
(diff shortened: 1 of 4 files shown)


common/dynamicconfig/constants.go
```diff
@@ -188,6 +188,11 @@ config as the other services.`,
 		false,
 		`EnableActivityEagerExecution indicates if activity eager execution is enabled per namespace`,
 	)
+	EnableActivityRetryStampIncrement = NewNamespaceBoolSetting(
```
xwduan: Are we enabling this feature by namespace? Seemed too complicated. Cluster level should be fine?

service/history/transfer_queue_standby_task_executor.go
```diff
@@ -162,8 +162,12 @@ func (t *transferQueueStandbyTaskExecutor) processActivityTask(
...
 		}
 
-		if activityInfo.Stamp != transferTask.Stamp || activityInfo.Paused {
-			return nil, nil // drop the task
+		if activityInfo.Paused {
+			return nil, nil
+		}
+
+		if activityInfo.Stamp != transferTask.Stamp {
+			return nil, consts.ErrStaleReference
```
xwduan, line 170: The task will not be retried when returning `ErrStaleReference`?

service/history/workflow/mutable_state_impl.go
```diff
@@ -5860,12 +5862,14 @@ func (ms *MutableStateImpl) updateActivityInfoForRetries(
 	_ = ms.UpdateActivity(ai.ScheduledEventId, func(activityInfo *persistencespb.ActivityInfo, mutableState historyi.MutableState) error {
 		mutableStateImpl, ok := mutableState.(*MutableStateImpl)
 		if ok {
+			isActivityRetryStampIncrementEnabled := mutableStateImpl.shard.GetConfig().EnableActivityRetryStampIncrement(mutableStateImpl.GetNamespaceEntry().Name().String())
```
xwduan: nit: `ms.config`?

service/history/transfer_queue_standby_task_executor.go
michaely520: yes, see here https://github.com/temporalio/temporal/blob/main/service/history/queues/executable.go#L399

PR #8610, opened 2025-11-08
Terminate workflow if respond message too large

yiminc:
## What changed?
Terminate workflow if workflow task failed on GRPC message too large issue

## Why?
gRPC error message too large is not a recoverable error, there is no point of keep retrying. 

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
No


yycptt (approved):

yycptt:
cc @simvlad something needs to be changed for payload store work (to not fail the workflow).

PR #8614, opened 2025-11-10
Versioning bug: don't let the server set the last modifier identity to itself during garbage collection

Shivs11:
## What changed?
- WISOTT

## Why?
- The main reason why this should be avoided is because the worker-controller actually stops assuming control when it sees a different identity being present on the LastModifierIdentity. This should not be set when the server attempts to garbage collect a version which right now happens when someone crosses the 100 versions per deployment limit set per namespace.
 
## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None pertaining to the server imho

---

> [!NOTE]
> Adds a `server_delete` flag to version deletion, updates workflow to bypass manager identity and not set `last_modifier_identity` for server-initiated GC, and adds tests.
> 
> - **API/Proto**:
>   - Add `bool server_delete` to `deployment.v1.DeleteVersionArgs` and regenerate `message.pb.go` (getter and wire defs).
> - **Workflow**:
>   - Introduce `serverDeleteVersionIdentity` and set it in `tryDeleteVersion`.
>   - In `validateDeleteVersion`, skip manager identity check when `args.server_delete` is true.
>   - In `deleteVersion`, avoid setting `state.last_modifier_identity` when `args.server_delete` is true.
> - **Tests**:
>   - Add `TestDeleteVersion_ServerDeleteMaxVersionsReached` to verify oldest version GC at max limit and that `last_modifier_identity` is not set to workflow identity.
>   - Minor tweak in `setAndValidateManagerIdentity` to set `req.Identity` based on new manager.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6b6b98b41e59eef1e5483836fec048bf68183d21. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/worker_deployment_test.go
+++ b/tests/worker_deployment_test.go
@@ -1707,7 +1707,7 @@ func (s *WorkerDeploymentSuite) TestDeleteVersion_ServerDeleteMaxVersionsReached
 		})
 		a.NoError(err)
 		a.Len(resp.GetWorkerDeploymentInfo().GetVersionSummaries(), 1)
-		a.Equal(tv2.DeploymentVersionString(), resp.GetWorkerDeploymentInfo().GetVersionSummaries()[0].GetVersion())
+		a.Equal(tv2.ExternalDeploymentVersion().GetBuildId(), resp.GetWorkerDeploymentInfo().GetVersionSummaries()[0].GetDeploymentVersion().GetBuildId())
 
 		// Also verify that the last modifier identity is not set to the identity of the worker-deployment workflow.
 		a.NotEqual(tv.ClientIdentity(), resp.GetWorkerDeploymentInfo().GetLastModifierIdentity())


service/worker/workerdeployment/workflow.go
```diff
@@ -660,7 +660,6 @@ func (d *WorkflowRunner) deleteVersion(ctx workflow.Context, args *deploymentspb
 	}
 	// update local state
 	delete(d.State.Versions, args.Version)
-	d.State.LastModifierIdentity = args.Identity
```
ShahabT, line 663: so this will stop setting LastModifierIdentity for all deleteVersion calls, not only the automatic ones. Is that what we want?

carlydf: I think Cursor and Shahab are right, we only want to not set the identity if the server is doing it

Shivs11: ah, I should not have missed this. Apologies. 
Resolved.

service/worker/workerdeployment/workflow.go
```diff
@@ -660,7 +660,9 @@ func (d *WorkflowRunner) deleteVersion(ctx workflow.Context, args *deploymentspb
...
+	if args.Identity != tryDeleteForAddingVersionIdentity {
+		d.State.LastModifierIdentity = args.Identity
+	}
```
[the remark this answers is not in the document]
Shivs11: oh shit, yeah - I think I am going to embed a flag in the proto 
Thanks

carlydf (commented):
Could you also make it so that if the delete request is a serverDelete, the manager identity==identity requirement is ignored?

To do that, you'd change the validator function here. Then, the server will still be able to correctly garbage collect old versions even after we implement the new ManagerIdentity logic in the controller

```
func (d *WorkflowRunner) validateDeleteVersion(args *deploymentspb.DeleteVersionArgs) error {
	if _, ok := d.State.Versions[args.Version]; !ok {
		return temporal.NewApplicationError("version not found in deployment", errVersionNotFound)
	}
	// Check if the version is not current or ramping. This condition is better to be checked in the
	// deployment workflow because that's the source of truth for routing config.
	if d.State.RoutingConfig.CurrentVersion == args.Version || d.State.RoutingConfig.RampingVersion == args.Version {
		// activity won't retry on this error since version not eligible for deletion
		return serviceerror.NewFailedPrecondition(ErrVersionIsCurrentOrRamping)
	}
	if d.State.ManagerIdentity != "" && d.State.ManagerIdentity != args.Identity {
		return serviceerror.NewFailedPrecondition(fmt.Sprintf(ErrManagerIdentityMismatch, d.State.ManagerIdentity, args.Identity))
	}
	return nil
}
```

tests/worker_deployment_test.go
```diff
@@ -3184,6 +3228,11 @@ func (s *WorkerDeploymentSuite) setAndValidateManagerIdentity(ctx context.Contex
 			ManagerIdentity: newManager,
 		}
 		expectedManagerIdentity = newManager
+		if newManager == "" {
+			req.Identity = tv.ClientIdentity()
+		} else {
+			req.Identity = newManager
+		}
```
Shivs11, line 3235: i think this was a small bug in this helper that I found - basically, I think if we are setting the identity of the manager to a string X, we should also set that value to the Identity. This is because internally, the LastModifierIdentity is being set to Identity and we want both of them to be the same when someone sets X as the new manager of a worker-deployment

carlydf: The Modifier and the Manager could be different. Like, me, with ClientIdentity="carly" might want to set ManagerIdentity="controller" and that would be valid. I wanted to test that scenario

carlydf: but it's fine for now -- totally possible I didn't write setAndValidateManagerIdentity correctly. More important to just get this test in with it working :)

PR #8615, opened 2025-11-10
Bump stamp on workflow task failure

michaely520:
## What changed?
- Bump stamp on workflow task retry
- Propagate stamp on matching workflow task forwarding.
- Add functional test
- Refactor test interceptors

## Why?
Workflow tasks can pile up on passive side due to stamps not incrementing, meaning they never become stale and never become eligible for invalidation. This ensures tasks can be invalidated due to staleness.

We need to ensure we forward the stamp in matching, otherwise prior to dispatch the task will be deemed invalid as the stamp does not match what is stored in mutable state.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/testcore/task_queue_recorder.go
+++ b/tests/testcore/task_queue_recorder.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/persistence"
 	"go.temporal.io/server/service/history/tasks"
 )
@@ -22,6 +23,7 @@ type TaskQueueRecorder struct {
 	mu       sync.RWMutex
 	tasks    map[tasks.Category][]RecordedTask // All tasks by category, in order
 	delegate persistence.ExecutionManager
+	logger   log.Logger
 }
 
 // RecordedTask wraps a task with metadata about when and where it was written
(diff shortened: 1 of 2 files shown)


tests/testcore/task_queue_recorder.go
```diff
@@ -268,6 +207,52 @@ func (r *TaskQueueRecorder) CountMatchingTasks(category tasks.Category, matcher
...
+
+// GetRecordedTasksByCategoryFiltered returns recorded tasks WITH metadata for a specific category,
+// filtered by namespace (required) and optionally by workflow ID and run ID.
+// This is the preferred API for tests to ensure tasks are properly scoped.
+func (r *TaskQueueRecorder) GetRecordedTasksByCategoryFiltered(category tasks.Category, filter TaskFilter) []RecordedTask {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+
+	if filter.NamespaceID == "" {
+		panic("TaskFilter.NamespaceID is required - use GetRecordedTasksByCategoryFiltered to prevent accidentally checking all tasks")
```
yux0: can we avoid panic?

michaely520: I want to panic to force the test to fail, as otherwise people will erroneously not pass namespace and will be incorrectly asserting outside the scope of their specific test.

michaely520: the other option is I expand to accept testing.T and fail

yux0: Can we use logger.DPanic? what is the worst case that we don't filter on namespace ID?

michaely520: updated

worst case is someone doesnt filter accidentally and introduces a test that is flaky as it may or may not be asserting on data introduced by other tests depending on test ordering/parallelism

yux0 (approved):

PR #8618, opened 2025-11-11
Encapsulate replication related ns config into replication resolver

xwduan:
## What changed?
Encapsulate replication related ns config into replication resolver

## Why?
To support different replication config.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
No risk.

### DIFF
--- a/service/history/replication/eager_namespace_refresher_test.go
+++ b/service/history/replication/eager_namespace_refresher_test.go
@@ -113,7 +113,7 @@ func (s *EagerNamespaceRefresherSuite) TestSyncNamespaceFromSourceCluster_Create
 	}).Return(nsResponse, nil)
 	s.mockReplicationTaskExecutor.EXPECT().Execute(gomock.Any(), task).Return(nil).Times(1)
 	s.mockNamespaceRegistry.EXPECT().GetNamespaceByID(namespaceId).Return(nil, serviceerror.NewNamespaceNotFound("namespace not found")).Times(1)
-	nsFromResponse, err := fromAdminClientApiResponse(nsResponse)
+	nsFromResponse, err := fromAdminClientAPIResponse(nsResponse)
 	s.NoError(err)
 	s.mockNamespaceRegistry.EXPECT().RefreshNamespaceById(namespaceId).Return(nsFromResponse, nil).Times(1)
 	ns, err := s.eagerNamespaceRefresher.SyncNamespaceFromSourceCluster(context.Background(), namespaceId, "currentCluster")
@@ -156,7 +156,7 @@ func (s *EagerNamespaceRefresherSuite) TestSyncNamespaceFromSourceCluster_Update
 	}
 	s.mockReplicationTaskExecutor.EXPECT().Execute(gomock.Any(), task).Return(nil).Times(1)
 	s.mockNamespaceRegistry.EXPECT().GetNamespaceByID(namespaceId).Return(nil, nil).Times(1)
-	nsFromResponse, err := fromAdminClientApiResponse(nsResponse)
+	nsFromResponse, err := fromAdminClientAPIResponse(nsResponse)
 	s.NoError(err)
 	s.mockNamespaceRegistry.EXPECT().RefreshNamespaceById(namespaceId).Return(nsFromResponse, nil).Times(1)
 	ns, err := s.eagerNamespaceRefresher.SyncNamespaceFromSourceCluster(context.Background(), namespaceId, "currentCluster")
@@ -247,7 +247,7 @@ func (s *EagerNamespaceRefresherSuite) TestSyncNamespaceFromSourceCluster_Namesp
 	s.IsType(&serviceerror.FailedPrecondition{}, err)
 }
 
-func fromAdminClientApiResponse(response *adminservice.GetNamespaceResponse) (*namespace.Namespace, error) {
+func fromAdminClientAPIResponse(response *adminservice.GetNamespaceResponse) (*namespace.Namespace, error) {
 	info := &persistencespb.NamespaceInfo{
 		Id:          response.GetInfo().GetId(),
 		Name:        response.GetInfo().GetName(),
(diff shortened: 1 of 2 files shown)


common/namespace/replication_resolver.go
```diff
@@ -0,0 +1,129 @@
+package namespace
+
+import (
+	enumspb "go.temporal.io/api/enums/v1"
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+)
+
+type ReplicationResolver interface {
```
xwduan, line 8: APIs in this interface is not final version. The next task is to abstract the usage of namespace replication info into api. Based on the usage of ns, the api in this interface might need to change.

common/namespace/namespace.go
```diff
@@ -81,17 +79,32 @@ func FromPersistentState(
...
+	return FromPersistentStateWithResolver(detail, nil, mutations...)
+}
+
+func FromPersistentStateWithResolver(
+	detail *persistencespb.NamespaceDetail,
+	resolver ReplicationResolver,
+	mutations ...Mutation,
+) *Namespace {
+	// If no resolver provided, create a default one
+	if resolver == nil {
```
michaely520, line 82: should we be defaulting like this versus deferring to factory always injecting the resolver and throwing an error here if resolver is nil? feels like we can be masking issues here

common/namespace/replication_resolver.go
```diff
@@ -0,0 +1,129 @@
...
+	r.isGlobalNamespace = isGlobal
+}
+
+func (r *defaultReplicationResolver) SetActiveCluster(clusterName string) {
+	if r.replicationConfig != nil {
+		r.replicationConfig.ActiveClusterName = clusterName
+	}
+}
+
+func (r *defaultReplicationResolver) PretendLocalNamespace(localClusterName string) {
```
michaely520, line 93: can you explain/maybe comment what you believe the usage of this is, as its non-intuitive how it may be used

common/namespace/namespace.go
```diff
@@ -173,34 +193,25 @@ func (ns *Namespace) State() enumspb.NamespaceState {
...
 }
 
 // ClusterNames observes the names of the clusters to which this namespace is
 // replicated.
 func (ns *Namespace) ClusterNames() []string {
-	// copy slice to preserve immutability
-	out := make([]string, len(ns.replicationConfig.Clusters))
-	copy(out, ns.replicationConfig.Clusters)
-	return out
+	return ns.replicationResolver.ClusterNames(EmptyEntityID)
```
michaely520: not fully following the usage of empty entity id?

common/namespace/replication_resolver.go
michaely520: i guess this is an existing functionality it seems

common/namespace/namespace.go
xwduan: Good point. And debugging what resolver is using becomes very tricky. Will make change in next revision.

common/namespace/namespace.go
xwduan: This is only used in this step and make it the same behavior as what we have now. In next step, I will deprecate ClusterNames() api in the namespace and add ClusterNames(entityId) api instead, so we don't need empty entity id anymore.

common/namespace/replication_resolver.go
xwduan: Yeah, it is used here: https://github.com/temporalio/temporal/blob/main/common/namespace/mutate.go#L82

common/namespace/namespace.go
```diff
@@ -63,8 +60,9 @@ type (
 )
 
 const (
-	EmptyName Name = ""
-	EmptyID   ID   = ""
+	EmptyName     Name = ""
+	EmptyID       ID   = ""
+	EmptyEntityID      = ""
```
yycptt: nit: we have decided to call everything an execution to be consistent with existing names in the code base.

common/namespace/namespace.go
```diff
@@ -214,19 +225,19 @@ func (ns *Namespace) ConfigVersion() int64 {
 
 // FailoverVersion return the namespace failover version
 func (ns *Namespace) FailoverVersion() int64 {
-	return ns.failoverVersion
+	return ns.replicationResolver.FailoverVersion()
```
yycptt, line 214: nit: probably just embed replicationResolver in Namespace.

common/namespace/namespace.go
```diff
@@ -214,19 +225,19 @@ func (ns *Namespace) ConfigVersion() int64 {
...
 // IsGlobalNamespace returns whether the namespace is a global namespace.
 // Being a global namespace doesn't necessarily mean that there are multiple registered clusters for it, only that it
 // has a failover version. To determine whether operations should be replicated for a namespace, see ReplicationPolicy.
 func (ns *Namespace) IsGlobalNamespace() bool {
-	return ns.isGlobalNamespace
+	return ns.replicationResolver.IsGlobalNamespace()
 }
 
 // FailoverNotificationVersion return the global notification version of when failover happened
 func (ns *Namespace) FailoverNotificationVersion() int64 {
```
yycptt, line 225: is this actually used anywhere? guess I never seen any application logic calling this method.

common/namespace/replication_resolver.go
```diff
@@ -0,0 +1,129 @@
...
+	persistencespb "go.temporal.io/server/api/persistence/v1"
+)
+
+type ReplicationResolver interface {
+	ActiveClusterName(entityID string) string
+	ClusterNames(entityID string) []string
+	WorkflowReplicationState(entityID string) enumspb.ReplicationState
+	NamespaceReplicationState() enumspb.ReplicationState
+	IsGlobalNamespace() bool
+	FailoverVersion() int64
```
yycptt, line 13: it should always take an executionID as argument? guess same question for other methods like NamespaceReplicationState(). I am not sure we want two versions of the method on the interface.

xwduan: Will do this refactor in next pr. This pr is to abstract the replication resolver.

common/namespace/namespace.go
xwduan: We can probably deprecate this. Seems not used.

common/namespace/namespace.go
xwduan: So an execution is identified by a WorkflowId and RunId?

common/namespace/namespace.go
xwduan: Will do this in next refactor step.

PR #8619, opened 2025-11-11
Misc callback related fixes

bergundy:
## What changed?

- Allow using `chasm.Context` when accessing `MSPointer`
- Embed `MSPointer` in the `Workflow` component to allow callbacks to access the exposed mutable state method
- Delete `hsm_invocation` to eliminate dead code
- Don't wrap CHASM callback failures in `UnprocessableTask` errors, failed callbacks are not supposed to bubble back to the executor anyways


chasm/lib/callback/component.go
```diff
@@ -65,43 +69,41 @@ func (c *Callback) recordAttempt(ts time.Time) {
 
 //nolint:revive // context.Context is an input parameter for chasm.ReadComponent, not a function parameter
 func (c *Callback) loadInvocationArgs(
-	chasmCtx chasm.Context,
-	ctx context.Context,
+	ctx chasm.Context,
+	_ chasm.NoValue,
```
spkane31, line 73: Can we remove the second parameter entirely? Or does the `chasm.ReadComponent` require an input of some kind

PR #8620, opened 2025-11-12
Change deployment limit error types to proper FailedPrecondition

ShahabT:
## What changed?
Ensure the Deployments client return proper FailedPrecondition error when we hit max deployments. versions, or task queues in a version limits.

Also cleaned up some unused code.

## Why?
If error type/code is not set properly user would see "internal" error without proper info in Cloud.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/service/worker/workerdeployment/client.go
+++ b/service/worker/workerdeployment/client.go
@@ -318,17 +318,23 @@ func (d *ClientImpl) RegisterTaskQueueWorker(
 }
 
 func (d *ClientImpl) handleRegisterVersionFailures(outcome *updatepb.Outcome) error {
-	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errMaxTaskQueuesInVersionType {
-		// translate to client-side error type
-		return &serviceerror.ResourceExhausted{Message: failure.Message, Scope: enumspb.RESOURCE_EXHAUSTED_SCOPE_NAMESPACE}
-	} else if failure.GetApplicationFailureInfo().GetType() == errTooManyVersions {
-		return &serviceerror.ResourceExhausted{Message: failure.Message, Scope: enumspb.RESOURCE_EXHAUSTED_SCOPE_NAMESPACE}
+	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errMaxTaskQueuesInVersionType ||
+		failure.GetApplicationFailureInfo().GetType() == errTooManyVersions {
+		return newResourceExhaustedError(failure.GetMessage())
 	} else if failure.GetApplicationFailureInfo().GetType() == errNoChangeType {
 		return nil
 	} else if failure != nil {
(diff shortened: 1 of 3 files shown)


service/worker/workerdeployment/client.go
```diff
@@ -334,9 +320,9 @@ func (d *ClientImpl) RegisterTaskQueueWorker(
 func (d *ClientImpl) handleRegisterVersionFailures(outcome *updatepb.Outcome) error {
 	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errMaxTaskQueuesInVersionType {
 		// translate to client-side error type
-		return ErrMaxTaskQueuesInVersion{error: errors.New(failure.Message)}
+		return &serviceerror.ResourceExhausted{Message: failure.Message, Scope: enumspb.RESOURCE_EXHAUSTED_SCOPE_NAMESPACE} // ErrMaxTaskQueuesInVersion{error: errors.New(failure.Message)}
```
carlydf: do you want this commented-out code still here?
Also, wdyt about adding a new `RESOURCE_EXHAUSTED_CAUSE` for this? I'm happy to leave it as unknown cause with namespace scope, but just want to think through what it means to not provide a cause

ShahabT: yeah good idea to add a cause I was lazy because it was in the api repo. but here is PR https://github.com/temporalio/api/pull/662

My guess is the cause is mostly for metrics but still useful.

service/worker/workerdeployment/client.go
```diff
@@ -334,9 +320,9 @@ func (d *ClientImpl) RegisterTaskQueueWorker(
 func (d *ClientImpl) handleRegisterVersionFailures(outcome *updatepb.Outcome) error {
 	if failure := outcome.GetFailure(); failure.GetApplicationFailureInfo().GetType() == errMaxTaskQueuesInVersionType {
 		// translate to client-side error type
-		return ErrMaxTaskQueuesInVersion{error: errors.New(failure.Message)}
+		return &serviceerror.ResourceExhausted{Message: failure.Message, Scope: enumspb.RESOURCE_EXHAUSTED_SCOPE_NAMESPACE}
```
carlydf: This should be precondition failed. See https://grpc.io/docs/guides/status-codes/ and slack thread. SDK retries resource exhausted, but a retry likely won't help here.

carlydf (changes requested):

service/worker/workerdeployment/client.go
ShahabT: done

carlydf: While we're here can you also populate the Cause = WORKER_DEPLOYMENT_LIMITS ? Chad and I both approved that API PR

carlydf (changes requested):

carlydf (approved):

PR #8623, opened 2025-11-12
Add request ID validation on record start activity

fretz12:
## What changed?
When recording the start standalone activity, validate the the incoming request ID.

## Why?
The requestID is set by the matching service to a UUID, allowing safe retries if the response is lost. If the existing request ID is empty, it sets it to the incoming one as this is a first attempt. If they match, it's a valid retry, and it's a successful no-op. If there's a mismatch, it returns a TaskAlreadyStarted error.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -147,7 +147,7 @@ func (a *Activity) RecordActivityTaskStarted(ctx chasm.MutableContext, params Re
 	}
 
 	// The requestID is set by the matching service to a UUID, allowing safe retries if the response is lost. If the
-	// existing request ID is empty, this is fresh start so update attempt attributes. If there's a mismatch, then
+	// existing request ID is empty, this is a fresh start so update attempt attributes. If there's a mismatch, then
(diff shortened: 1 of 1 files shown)


chasm/lib/activity/proto/v1/activity_state.proto
```diff
@@ -143,6 +143,12 @@ message ActivityAttemptState {
     // The Worker Deployment Version this activity was dispatched to most recently.
     // If nil, the activity has not yet been dispatched or was last dispatched to an unversioned worker.
     temporal.api.deployment.v1.WorkerDeploymentVersion last_deployment_version = 8;
+
+    // Request identifier for idempotent task start operations. Matching service sets this to a
+    // UUID, allowing safe retries if the response is lost. History service returns success (not
+    // AlreadyStarted error) when retrying with a matching request_id, ensuring tasks are delivered
+    // to workers even after network failures.
+    string request_id = 9;
```
bergundy, line 151: You don't need this, use `chasm.WithRequestID` when handling the start request.

fretz12:
Per comment, this just needs a one liner `WithRequestID`. I will open a new PR for it.

PR #8626, opened 2025-11-12
lock before reading subqueues in emitZeroBacklogGauges

majiru:
## What changed?
We changed the approximateBacklogCount metric to be tagged with priority, this also meant adding that tag in emitZeroBacklogGauges. However now that we're reading the priority off the subqueues, we need to take the lock here before reading.

## Why?
Fix a race found in the functional tests.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/db.go
+++ b/service/matching/db.go
@@ -816,8 +816,8 @@ func (db *taskQueueDB) emitZeroBacklogGauges() {
 		return
 	}
 
-	db.Lock()
 	priorities := make(map[int32]struct{})
+	db.Lock()
 	for _, s := range db.subqueues {
 		priorities[s.Key.Priority] = struct{}{}
 	}


service/matching/db.go
```diff
@@ -816,10 +816,13 @@ func (db *taskQueueDB) emitZeroBacklogGauges() {
 		return
 	}
 
+	db.Lock()
```
dnr: nit: could put this after the make(map(...)), not sure if the compiler is smart enough to do that

majiru: Yup, of course, I can put it after the make. Yeah, the race detector found this in one of the other functional tests runs for another PR after this was merged.

stephanos:
Yup; it was caught here: https://github.com/temporalio/temporal/actions/runs/19305999783/job/55213819748

PR #8627, opened 2025-11-13
Translate context deadline from forwarding nexus tasks to nexus request timeout

bergundy:
## Why?

The previous behavior where we return a context deadline is wrong.

## How did you test it?
- [x] covered by existing tests (with modification)

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -1034,6 +1034,10 @@ func (e *matchingEngineImpl) QueryWorkflow(
 	}
 
 	taskID := uuid.New()
+	queryResultCh := make(chan *queryResult, 1)
+	e.queryResults.Set(taskID, queryResultCh)
+	defer e.queryResults.Delete(taskID)
+
 	resp, err := pm.DispatchQueryTask(ctx, taskID, queryRequest)
 
 	// if we get a response or error it means that query task was handled by forwarding to another matching host
@@ -1044,10 +1048,6 @@ func (e *matchingEngineImpl) QueryWorkflow(
 
 	// if we get here it means that dispatch of query task has occurred locally
 	// must wait on result channel to get query result
-	queryResultCh := make(chan *queryResult, 1)
-	e.queryResults.Set(taskID, queryResultCh)
-	defer e.queryResults.Delete(taskID)
-
 	select {
 	case result := <-queryResultCh:
 		if result.internalError != nil {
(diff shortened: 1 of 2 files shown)


service/matching/matching_engine.go
```diff
@@ -2352,20 +2352,32 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
...
+	defer e.nexusResults.Delete(taskID)
+
 	resp, err := pm.DispatchNexusTask(ctx, taskID, request)
 
-	// if we get a response or error it means that the Nexus task was handled by forwarding to another matching host
-	// this remote host's result can be returned directly
-	if resp != nil || err != nil {
+	if err != nil {
+		if ctx.Err() != nil {
+			// The context deadline has expired if it reaches here; return an explicit timeout response to the caller.
```
pdoerner, line 2365: Do we want to treat `context canceled` the same as `context deadline exceeded`?

service/matching/matching_engine.go
```diff
@@ -2352,20 +2352,32 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
 	ctx, cancel := contextutil.WithDeadlineBuffer(ctx, matching.DefaultTimeout, e.config.MinDispatchTaskTimeout(ns.Name().String()))
 	defer cancel()
 
+	// First allocate a result channel and register it so that when the task is completed locally (without forwarding) the
```
pdoerner, line 2355: I would prefer someone from the matching team confirms this will not cause additional resource leakage / overhead. There is [a comment](https://github.com/temporalio/temporal/blob/4b035d52bae6a86848dfe2385e56edd9c6839354/service/matching/task_queue_partition_manager.go#L558-L564) in task_queue_partition_manager that indicates this task could still get redirected, so I don't want to assume that it will always be completed locally.

cc @ShahabT

service/matching/matching_engine.go
```diff
@@ -2352,20 +2352,32 @@ func (e *matchingEngineImpl) DispatchNexusTask(ctx context.Context, request *mat
...
-	e.nexusResults.Set(taskID, resultCh)
-	defer e.nexusResults.Delete(taskID)
+	// If we get a response it means that the Nexus task was handled by forwarding to another matching host this remote
+	// host's result can be returned directly.
+	if resp != nil {
+		return resp, nil
+	}
 
+	// If we get here it means that task dispatch has occurred locally.
+	// Must wait on result channel to get query result.
```
pdoerner, line 2380:
```suggestion
	// Must wait on result channel to get task result.
```

pdoerner: Or just remove this line, it's not that helpful.

service/matching/matching_engine.go
dnr: "Redirected" is still matched on this node, "forwarded" is matched on another node. In general forwarding should not be a large proportion of tasks, so most of the time this will be used locally and isn't a waste. I.o.w. it's fine to do it here.

pdoerner (approved):

PR #8632, opened 2025-11-13
Add revision number mechanics for child and CAN workflows

Shivs11:
## What changed?
- Add revision number mechanics for child workflows and CAN workflows.
- Additionally, this PR introduces the following change conceptually: Child workflows of AutoUpgrade parents and CAN workflow executions of AutoUpgrade workflows have their first task as AutoUpgrade by default. Previously, it was unversioned.

## Why?
- Resolving eventual consistency issues.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None since this feature is gated by a dc

---

> [!NOTE]
> Adds `inherited_auto_upgrade_info` propagation and revision-number-based dispatch so child and continue-as-new workflows start with AutoUpgrade behavior when appropriate; updates matching/history logic and tests.
> 
> - **API/Proto**:
>   - Add `StartWorkflowExecutionRequest.inherited_auto_upgrade_info` (and getter) to carry deployment version and revision from parent/previous run.
> - **History Service**:
>   - Populate/propagate `InheritedAutoUpgradeInfo` when starting child workflows and during ContinueAsNew; avoid when pinned inheritance applies or namespaces differ.
>   - On start, set execution `VersioningInfo` (deployment, behavior AutoUpgrade) and revision number from inherited info.
>   - Minor fixes: use request context for version checks; guard `SetVersioningRevisionNumber` against nil.
> - **Matching**:
>   - Add test hook `MatchingIgnoreRoutingConfigRevisionCheck` and allow bypassing routing-config revision gate in updates.
>   - Minor cleanup in partition manager; keep revision-based queue selection logic.
> - **Tests**:
>   - Add/rename functional tests covering child/CAN inheritance, cross-TQ behavior, retry no-bounce with revision lag, and propagation helpers; introduce rollback utility for task queue user data.
> - **Deps**:
>   - Bump `go.temporal.io/api` to latest prerelease.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8a342ab287a8ab7f7ac7a4b3009db3d02c1012d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/versioning_3_test.go
```diff
@@ -3585,7 +3585,7 @@ func (s *Versioning3Suite) verifyVersioningSAs(
 }
 
 func (s *Versioning3Suite) TestAutoUpgradeWorkflows_NoBouncingBetweenVersions() {
-	s.T().Skip("This test is flaky right now and shall be fixed in a future PR.") // TODO (Shivam)
+	s.T().Skip("Skipping this test for now as there seems to be a flake. Shall fix it in the upcoming PR.")
```
Shivs11, line 3655: @ShahabT - for some reason, this test keeps on flaking on me. I have spent sometime trying to understand why but have not been able to (just yet). I do wanna get this PR in though and don't want to block on this test so shall be shipping this out in a different PR.

The failure seems to be that the test "times out" for some reason.

common/util.go
Shivs11, line 532: Shahab: "can the caller just pass this as one struct rather than two fields?"

service/history/workflow/mutable_state_impl.go
```diff
@@ -2403,6 +2403,34 @@ func (ms *MutableStateImpl) addWorkflowExecutionStartedEventForContinueAsNew(
...
+
+		newTQ := command.GetTaskQueue().GetName()
+		if newTQ != previousExecutionInfo.GetTaskQueue() {
+			// Cross-TQ CAN: check if new TQ is in parent's deployment
+			TQInSourceDeployment, err := IsWFTaskQueueInVersionDetector(
+				context.Background(),
+				ms.GetNamespaceEntry().ID().String(),
+				newTQ,
+				sourceDeploymentVersion,
+			)
```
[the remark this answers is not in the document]
Shivs11: I think this may be a good catch - thinking about this hmmm

service/history/workflow/mutable_state_impl.go
```diff
@@ -2403,6 +2403,34 @@ func (ms *MutableStateImpl) addWorkflowExecutionStartedEventForContinueAsNew(
...
+	var sourceDeploymentRevisionNumber int64
+	if previousExecutionState.GetEffectiveVersioningBehavior() == enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE {
+		sourceDeploymentVersion = worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(previousExecutionState.GetEffectiveDeployment())
+		sourceDeploymentRevisionNumber = previousExecutionState.GetVersioningRevisionNumber()
+
+		newTQ := command.GetTaskQueue().GetName()
+		if newTQ != previousExecutionInfo.GetTaskQueue() {
+			// Cross-TQ CAN: check if new TQ is in parent's deployment
+			TQInSourceDeployment, err := IsWFTaskQueueInVersionDetector(
+				context.Background(),
```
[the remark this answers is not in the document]
Shivs11: nah

service/history/transfer_queue_active_task_executor.go
```diff
@@ -942,6 +943,29 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
...
+
+		newTQ := attributes.GetTaskQueue().GetName()
+		if attributes.GetNamespaceId() != mutableState.GetExecutionInfo().GetNamespaceId() { // don't inherit auto upgrade info if child is in a different namespace
+			inheritedAutoUpgradeInfo = nil
+		} else if newTQ != mutableState.GetExecutionInfo().GetTaskQueue() {
+			TQInSourceDeployment, err := worker_versioning.GetIsWFTaskQueueInVersionDetector(t.matchingRawClient)(ctx, attributes.GetNamespaceId(), newTQ, inheritedAutoUpgradeInfo.GetSourceDeploymentVersion())
+			if err != nil {
+				//nolint:staticcheck
+				return errors.New(fmt.Sprintf("error determining child task queue presence in auto upgrade deployment: %s", err.Error()))
+			}
```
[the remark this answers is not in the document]
Shivs11: good point - changes will be made in the next commit.

tests/versioning_3_test.go
```diff
@@ -4015,95 +4107,714 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
+		RevisionNumber:            0,
+	}
+	s.syncTaskQueueDeploymentDataWithRoutingConfig(tv0, rc, map[string]*deploymentspb.WorkerDeploymentVersionData{tv0.DeploymentVersion().GetBuildId(): &deploymentspb.WorkerDeploymentVersionData{
+		Status: enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT,
+	}, tv1.DeploymentVersion().GetBuildId(): &deploymentspb.WorkerDeploymentVersionData{
+		Status: enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING,
+	}}, nil, tqTypeWf)
+
+	// Verify that the rollback propagated to all partitions
+	s.Eventually(func() bool {
```
Shivs11, line 4620: this seems to flake out: I want to assert that the current revision number changed to 0 for this task-queue partition (previously it was 1)

I have also added a testhook condition over here (the only change in matching_engine.go). However, sometimes the API updates the partition correctly while in others it just....doesn't?

I think majority of these errors are also only in cassandra....

go.mod
```diff
@@ -171,3 +171,5 @@ require (
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.11.0 // indirect
 )
+
+replace go.temporal.io/api => github.com/temporalio/api-go v1.57.1-0.20251124143622-8cd477726006
```
Shivs11: reminder: remove this please.

service/matching/matching_engine.go
```diff
@@ -2027,7 +2027,8 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
 				rc := req.GetUpdateRoutingConfig()
 				tqWorkerDeploymentData := deploymentData.GetDeploymentsData()[req.GetDeploymentName()]
 
-				if rc.GetRevisionNumber() > tqWorkerDeploymentData.GetRoutingConfig().GetRevisionNumber() {
+				ignoreRevCheck, _ := testhooks.Get[bool](e.testHooks, testhooks.MatchingIgnoreRoutingConfigRevisionCheck)
```
Shivs11, line 2030: I added a testhook here so that one could test routingConfig rollbacks without using the lower level matching API

chasm/tree_test.go
```diff
@@ -562,14 +562,12 @@ func (s *nodeSuite) assertParentPointer(testComponentNode *Node) {
 	_, found := testComponent.ParentPtr.TryGet(chasmContext)
 	s.False(found)
 
-	subComponent1, err := testComponent.SubComponent1.Get(chasmContext)
-	s.NoError(err)
+	subComponent1 := testComponent.SubComponent1.Get(chasmContext)
```
Shivs11, line 565: [for reviewer]: ignore this as I have another PR open. 
https://github.com/temporalio/temporal/pull/8699

service/history/workflow/retry.go
```diff
@@ -255,11 +255,24 @@ func SetupNewWorkflowForRetryOrCron(
...
+			startAttr.GetInheritedPinnedVersion() != nil {
+			inheritedPinnedVersion = worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(GetEffectiveDeployment(previousExecutionInfo.GetVersioningInfo()))
+		} else if GetEffectiveVersioningBehavior(previousExecutionInfo.GetVersioningInfo()) == enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE {
+			sourceDeploymentVersion := worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(previousMutableState.GetEffectiveDeployment())
+			sourceDeploymentRevisionNumber := previousMutableState.GetVersioningRevisionNumber()
+			inheritedAutoUpgradeInfo = &deploymentpb.InheritedAutoUpgradeInfo{
+				SourceDeploymentVersion:        sourceDeploymentVersion,
+				SourceDeploymentRevisionNumber: sourceDeploymentRevisionNumber,
+			}
+		}
```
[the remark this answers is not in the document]
Shivs11: I don't think this would change the functionality of the core feature here.

tests/versioning_3_test.go
```diff
@@ -72,31 +76,38 @@ type Versioning3Suite struct {
 	useV32                    bool
 	deploymentWorkflowVersion workerdeployment.DeploymentWorkflowVersion
 	useRevisionNumbers        bool
-	// TODO: this is always false. clean it up.
-	useNewDeploymentData bool
+	useNewDeploymentData      bool
```
ShahabT, line 79: these two always have same values, we should just use one var. IMO `useRevisionNumbers` describes the flag better.

tests/versioning_3_test.go
```diff
@@ -72,31 +76,38 @@ type Versioning3Suite struct {
...
+		suite.Run(t, &Versioning3Suite{
+			deploymentWorkflowVersion: workerdeployment.InitialVersion,
+			useV32:                    true,
+			useNewDeploymentData:      false,
+			useRevisionNumbers:        false,
+		})
 	})
 
-	t.Run("async_deployment", func(t *testing.T) {
+	t.Run("async_new_deployment_data", func(t *testing.T) {
```
ShahabT: Nit: async comes with new deployment data anyway so not sure if we need to mention it I tried to avoid making test names so long.

tests/versioning_3_test.go
```diff
@@ -857,6 +868,7 @@ func (s *Versioning3Suite) TestWorkflowRetry_Pinned_ExpectNoInherit() {
 	s.testWorkflowRetry(workflow.VersioningBehaviorPinned, false, false, false)
 }
 
+// TODO (Shivam): I think one may need to slightly modify this now that we are actually "inheriting" the autoupgrade version.
 func (s *Versioning3Suite) TestWorkflowRetry_Unpinned_ExpectNoInherit() {
```
ShahabT: Yup, if the logic is changing in this PR, shouldn't we update the test at the same time? otherwise, it'd fail, right?

tests/versioning_3_test.go
Shivs11: Yes. Sorry, I should have kept a comment there but basically, after this PR goes in, I am sending another one which will just deal with versioning_3_test refactors. 

That PR shall remove this variable completely and replace wherever we are using this with useRevisionNumbers.

I didn't wanna clutter this PR since it's anyways kinda big and is making some fundamental changes.

tests/versioning_3_test.go
Shivs11: sorry, that comment is misleading imo - what i meant to add there was change the name of this test since we are inheriting the version from the parent!!

What this test is testing will still pass with my changes. The reason being as follows:

This test is testing that a retry of an autoUpgrade workflow bounces "ahead" to the next current deployment version of the task-queue. This, in theory, shall also work with my changes since the retried workflow shall inherit the revision number (and deployment version) prior to the retry, but will always be scheduled on the current deployment version if it is more recent (in this case it will be since we wait for tv2 to be current prior to kickstarting the retry)

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
 	}, 10*time.Second, 100*time.Millisecond)
 }
 
-// Fallback of revision number mechanics: Independent activities are eventually consistent.
-func (s *Versioning3Suite) TestActivityTQLags_IndependentActivityDispatchesToItsCurrentDeployment() {
```
Shivs11, line 4019: no point in having this imho

ShahabT (commented):
Can we get @carlydf approval for history side? she's more familiar with this part and also this is related to trampolining.

service/history/workflow/mutable_state_impl.go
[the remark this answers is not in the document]
carlydf: Actually yeah, although I also used context.Background(), I think both calls to this function should pass the ctx variable that we already have in the environment instead.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2467,6 +2494,14 @@ func (ms *MutableStateImpl) addWorkflowExecutionStartedEventForContinueAsNew(
 		req.WorkflowExecutionExpirationTime = timestamppb.New(workflowTimeoutTime)
 	}
 
+	// Add InheritedAutoUpgradeInfo if source deployment version and revision number are set
+	if sourceDeploymentVersion != nil && sourceDeploymentRevisionNumber != 0 {
+		req.InheritedAutoUpgradeInfo = &deploymentpb.InheritedAutoUpgradeInfo{
+			SourceDeploymentVersion:        sourceDeploymentVersion,
+			SourceDeploymentRevisionNumber: sourceDeploymentRevisionNumber,
+		}
+	}
```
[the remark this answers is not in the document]
carlydf: Did you resolve this because when a wf is continuing-as-new, the revision number would never actually be 0?

Regardless, I think `sourceDeploymentVersion != nil && inheritedPinnedVersion == nil` is a sufficient condition and you don't need to think about all the possible cases where revision number could legitimately be 0.

I guess what I'm saying is, is there a case where revision # is 0 but we would still want to pass down the auto-upgrade version?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2853,6 +2888,17 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 	ms.approximateSize += ms.executionInfo.Size()
 	ms.approximateSize += ms.executionState.Size()
 
+	// Populate the versioningInfo if the inheritedAutoUpgradeInfo is present.
+	if event.GetInheritedAutoUpgradeInfo() != nil {
+		ms.SetVersioningRevisionNumber(event.GetInheritedAutoUpgradeInfo().GetSourceDeploymentRevisionNumber())
```
carlydf, line 2907: this call will have NDE if versioning Info is nil, so you should do it after creating the empty versioning info

carlydf: oh I see you also changed the `Set` function to be nil safe, so nevermind!

Should we make symmetrical `SetDeploymentVersion` and `SetVersioningBehavior` functions too, so that there is no need to repeat 
```
		if ms.executionInfo.VersioningInfo == nil {
			ms.executionInfo.VersioningInfo = &workflowpb.WorkflowExecutionVersioningInfo{}
		}
``` 
every time we set those things? not necessary but could be nice

service/history/workflow/retry.go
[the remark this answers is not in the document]
carlydf: I think this comment supports my idea of omitting the != 0 check in the continue-as-new code path

service/matching/matching_engine.go
carlydf: This is cool but I'd love to understand more how it works. It seems like in a test environment, line 2031 just always updates the routing config regardless of revision number; is that what we want?

service/matching/task_queue_partition_manager.go
```diff
@@ -1207,14 +1207,14 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
 // chooseTargetQueueByFlag picks the target queue and dispatch revision number.
 // If UseRevisionNumberForWorkerVersioning is enabled, it uses the revision
 // comparison; otherwise it always chooses targetDeployment.
-// TODO (Shivam): This function can be simplified to literally one if check.
```
carlydf, line 1210: lol can it be simplified to one if check?

tests/worker_deployment_test.go
```diff
@@ -38,10 +38,10 @@ type (
 
 func TestWorkerDeploymentSuite(t *testing.T) {
 	t.Parallel()
-	t.Run("v0", func(t *testing.T) {
+	t.Run("sync_workflows", func(t *testing.T) {
 		suite.Run(t, &WorkerDeploymentSuite{workflowVersion: workerdeployment.InitialVersion})
 	})
-	t.Run("v1", func(t *testing.T) {
+	t.Run("async_workflows", func(t *testing.T) {
```
carlydf, line 44: nice :)

tests/versioning_3_test.go
```diff
@@ -2079,16 +2069,18 @@ func (s *Versioning3Suite) testChildWorkflowInheritance_ExpectNoInherit(crossTq
...
+	// Wait for propagation to complete since we have tests using async entity workflows to set the current version
+	if s.deploymentWorkflowVersion == workerdeployment.AsyncSetCurrentAndRamping {
+		s.Eventually(func() bool {
+			resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+				Namespace:      s.Namespace().String(),
+				DeploymentName: tv1.DeploymentSeries(),
+			})
+			s.NoError(err)
+			return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
+		}, 10*time.Second, 100*time.Millisecond)
```
carlydf: should we not just put this "await" into `s.setCurrentDeployment`?

tests/versioning_3_test.go
carlydf: is there a specific upcoming PR? lmk if you want any help understanding or investigating the flake!

tests/versioning_3_test.go
carlydf: why?

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
+	s.Eventually(func() bool {
+		resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      s.Namespace().String(),
+			DeploymentName: tv2.DeploymentSeries(),
+		})
+		s.NoError(err)
+		return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
 	}, 10*time.Second, 100*time.Millisecond)
 
-	// Update the userData for the workflow TQ *only* by setting the current version to v1
```
carlydf, line 4071: wait im confused, doesn't removing this completely change the test?

carlydf (commented):
Left some comments. I think a good chunk of repeated test code can be avoided if we just wait for propagation from inside the `setCurrent` helper function. 

Also, I wonder if `TestChildStartsWithParentRevision_SameTQ_TQLags`, `TestChildStartsWithNoInheritedAutoUpgradeInfo_CrossTQ`, and `TestChildStartsWithParentRevision_SameTQ_TQAhead` could share more code. If it ends up being flaky or we want to fix the test, it's nice to only have to fix code in one place

tests/versioning_3_test.go
Shivs11: yes, there is a follow-up PR coming very soon!! I fixed it in that one https://github.com/temporalio/temporal/pull/8635

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -93,6 +93,11 @@ message StartWorkflowExecutionRequest {
     bool child_workflow_only = 14;
     // If present, the new workflow should start on this version with pinned base behavior.
     temporal.api.deployment.v1.WorkerDeploymentVersion inherited_pinned_version = 15;
+    // Passes deployment version and revision number from a parent/previous workflow with AutoUpgrade behavior
+    // to its child/continued-as-new workflow. The first workflow task of the child/CAN workflow is dispatched to
+    // either this deployment version or the current version of the task queue. After the first workflow task,
```
carlydf: actually I'm confused as to why we pass the version here. Isn't it always going to start on the current version of the task queue?

Shivs11: There could be a situation where the task gets routed to a matching partition that does not have the most recent updated current version. If that happens, the task can go to an older current version

service/matching/matching_engine.go
Shivs11: Not really; In a test environment, the value of the testhook "MatchingIgnoreRoutingConfigRevisionCheck" will be false by default. In the cases where I do want this value to be true, I run something like:

```
	cleanup := s.InjectHook(testhooks.MatchingIgnoreRoutingConfigRevisionCheck, true)
	defer cleanup()
```

What this does is that it "injects" the value for this knob to be true during the runtime of the test, helping me conduct rollbacks (in this case)

tests/versioning_3_test.go
carlydf: can be removed

tests/versioning_3_test.go
```diff
@@ -2131,27 +2123,19 @@ func (s *Versioning3Suite) testChildWorkflowInheritance_ExpectNoInherit(crossTq
...
+	// Wait for propagation to complete since we have tests using async entity workflows to set the current version
+	if s.deploymentWorkflowVersion == workerdeployment.AsyncSetCurrentAndRamping {
+		s.Eventually(func() bool {
+			resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+				Namespace:      s.Namespace().String(),
+				DeploymentName: tv2.DeploymentSeries(),
+			})
+			s.NoError(err)
+			return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
+		}, 10*time.Second, 100*time.Millisecond)
```
carlydf: can be removed

tests/versioning_3_test.go
```diff
@@ -3719,7 +3769,7 @@ func (s *Versioning3Suite) TestAutoUpgradeWorkflows_NoBouncingBetweenVersions()
 	}}, []string{}, tqTypeWf, tqTypeAct)
 
 	// Wait until all task queue partitions know that v1 is current.
```
carlydf, line 3735: do you mean to comment this out?

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
+	s.setCurrentDeployment(tvParent)
+	// Wait for propagation to complete
+	s.Eventually(func() bool {
+		resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      s.Namespace().String(),
+			DeploymentName: tvParent.DeploymentSeries(),
 		})
+		s.NoError(err)
+		return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
+	}, 10*time.Second, 100*time.Millisecond)
```
carlydf: can be removed

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
+	// Set v2 as the current version of the deployment
+	s.setCurrentDeployment(tv2)
+	// Wait for propagation to complete
+	s.Eventually(func() bool {
+		resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      s.Namespace().String(),
+			DeploymentName: tv2.DeploymentSeries(),
+		})
+		s.NoError(err)
+		return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
```
carlydf: can be removed

tests/versioning_3_test.go
Shivs11: This is something that revision numbers cannot solve. It didn't make sense to have a test for something for revision numbers where it's a limitation and also this is not something which we wanna solve in the near future either. I think it was decided to let customers know that while running indepedant activities, one could expect some bouncing of tasks so it is important to keep activity defs compatible

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
+	s.setCurrentDeployment(tvParent)
+	// Wait for propagation to complete
+	s.Eventually(func() bool {
+		resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      s.Namespace().String(),
+			DeploymentName: tvParent.DeploymentSeries(),
 		})
+		s.NoError(err)
+		return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
+	}, 10*time.Second, 100*time.Millisecond)
```
carlydf: can be removed

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
```
carlydf: can be removed

tests/versioning_3_test.go
```diff
@@ -4015,95 +4073,682 @@ func (s *Versioning3Suite) TestActivityTQLags_DependentActivityCompletesOnTheNew
...
+	s.setCurrentDeployment(tvChild)
+	// Wait for propagation to complete
+	s.Eventually(func() bool {
+		resp, err := s.FrontendClient().DescribeWorkerDeployment(context.Background(), &workflowservice.DescribeWorkerDeploymentRequest{
+			Namespace:      s.Namespace().String(),
+			DeploymentName: tvChild.DeploymentSeries(),
+		})
+		s.NoError(err)
+		return resp.GetWorkerDeploymentInfo().GetRoutingConfigUpdateState() == enumspb.ROUTING_CONFIG_UPDATE_STATE_COMPLETED
+	}, 10*time.Second, 100*time.Millisecond)
```
carlydf: can be removed

service/matching/task_queue_partition_manager.go
Shivs11: I think it can; I shall bring back the TODO but don't want to do it in this PR. This shall come in my next PR where I refactor versioning_3_test.

I generally like to keep refactorings and feature additions separate so that debugging becomes much simpler in the near future

service/history/workflow/mutable_state_impl.go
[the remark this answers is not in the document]
Shivs11: synced and we both agree to keep things conservative and have both these checks be present.

service/history/workflow/mutable_state_impl.go
Shivs11: I think I can do that in the upcoming PR that's gonna be a refactor one. But yes, I agree I can do this.

tests/versioning_3_test.go
Shivs11: this test is gonna undergo some changes so these changes can be ignored.

tests/versioning_3_test.go
Shivs11: I think the diff is misleading!
This is comparing the diff line by line, and on these lines stood the test TestActivityTQLags_IndependentActivityDispatchesToItsCurrentDeployment which I am removing in this PR. 

The diff you are seeing here is just showing the outdated test's content. This test that I am adding is brand new.

service/history/transfer_queue_active_task_executor.go
```diff
@@ -942,6 +942,28 @@ func (t *transferQueueActiveTaskExecutor) processStartChildExecution(
 		}
 	}
 
+	// If the parent has AutoUpgrade behavior, we populate the inherited auto upgrade info based on whether the child TQ is in the same deployment version as the parent TQ.
+	var inheritedAutoUpgradeInfo *deploymentpb.InheritedAutoUpgradeInfo
+	if effectiveVersioningBehavior := mutableState.GetEffectiveVersioningBehavior(); effectiveVersioningBehavior == enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE {
+		inheritedAutoUpgradeInfo = &deploymentpb.InheritedAutoUpgradeInfo{
+			SourceDeploymentVersion:        worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(mutableState.GetEffectiveDeployment()),
+			SourceDeploymentRevisionNumber: mutableState.GetVersioningRevisionNumber(),
+		}
```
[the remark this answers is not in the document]
Shivs11: thanks!

PR #8634, opened 2025-11-13
Add search attribute alias to field mapping to Namespace Detail for ElasticSearch

awln-temporal:
Add search attribute alias to field mapping to Namespace Detail for ElasticSearch.

* TODO: Run test cluster and register search attributes via Operator APIs.

## Why?
Currently, when a Cluster is configured with ElasticSearch, search attribute registration via the Operator API `AddSearchAttributes` persist the user aliased search attributes in the ClusterMetadata without writing to the Namespace Detail.

ClusterMetadata is meant to list all pre-allocated custom search attribute column names and their types.

Search attributes are namespace scoped, registering at the ClusterMetadata level leads unexpected behavior.

## What changed?
To support backwards compatibility with existing Clusters, existing search attributes in the ClusterMetadata must be replicated to each NamespaceDetail at startup.

All pre-allocated fields must be added to the ClusterMetadata, similar to existing SQL implementations.

## How did you test it?
- [X] built
- [X] added new unit test(s)


common/namespace/nsregistry/registry.go
```diff
@@ -588,3 +591,114 @@ func namespaceStateChanged(old *namespace.Namespace, new *namespace.Namespace) b
...
+		needsUpdate = true
+	}
+
+	if !needsUpdate {
+		return nil
+	}
+
+	// Update the namespace with new mappings
+	nsDetail.Namespace.Config.CustomSearchAttributeAliases = upsertFieldToAliasMap
+	err := metadataManager.UpdateNamespace(ctx, &persistence.UpdateNamespaceRequest{
```
awln-temporal: note: needs to go through frontend for replication.

common/dynamicconfig/constants.go
```diff
@@ -779,6 +779,11 @@ This config is EXPERIMENTAL and may be changed or removed in a later release.`,
 		false,
 		`SendRawWorkflowHistory is whether to enable raw history retrieving`,
 	)
+	EnableSearchAttributeCustomMapperOverride = NewGlobalBoolSetting(
+		"frontend.enableCustomSearchAttributeProviderOverride",
+		true,
+		`EnableCustomSearchAttributeProviderOverride is whether to enable custom search attribute mapper override. If disabled, the namespace detail search attribute mapper will be used.`,
+	)
```
rodrigozhou: What's the goal of this dynamic config?

common/namespace/nsregistry/registry.go
```diff
@@ -882,3 +884,114 @@ func namespaceStateChanged(old *namespace.Namespace, new *namespace.Namespace) b
...
+			if err := updateSingleNamespaceSearchAttributeMappings(
+				ctx,
+				metadataManager,
+				nsDetail,
+				indexSearchAttrs.CustomSearchAttributes,
+				logger,
+			); err != nil {
+				logger.Warn("Failed to update namespace search attribute mappings",
+					tag.WorkflowNamespace(nsDetail.Namespace.Info.Name),
+					tag.Error(err))
```
rodrigozhou: Maybe save the last error, and return it in the end?

common/namespace/nsregistry/registry.go
```diff
@@ -882,3 +884,114 @@ func namespaceStateChanged(old *namespace.Namespace, new *namespace.Namespace) b
...
+) error {
+	fieldToAliasMap := nsDetail.Namespace.Config.CustomSearchAttributeAliases
+	if fieldToAliasMap == nil {
+		fieldToAliasMap = make(map[string]string)
+	}
+
+	needsUpdate := false
+	upsertFieldToAliasMap := make(map[string]string)
+	for fieldName, fieldType := range clusterCustomSearchAttributes {
+		if sadefs.IsPreallocatedCSAFieldName(fieldName, fieldType) {
```
rodrigozhou: What is the purpose of this check?

common/namespace/nsregistry/registry.go
```diff
@@ -882,3 +884,114 @@ func namespaceStateChanged(old *namespace.Namespace, new *namespace.Namespace) b
...
+		needsUpdate = true
+	}
+
+	if !needsUpdate {
+		return nil
+	}
+
+	// Update the namespace with new mappings
+	nsDetail.Namespace.Config.CustomSearchAttributeAliases = upsertFieldToAliasMap
+	err := metadataManager.UpdateNamespace(ctx, &persistence.UpdateNamespaceRequest{
```
rodrigozhou: Wondering if you need to call the frontend handler instead.
cc: @yycptt

common/resource/fx.go
```diff
@@ -155,12 +155,17 @@ func SearchAttributeMapperProviderProvider(
 	namespaceRegistry namespace.Registry,
 	searchAttributeProvider searchattribute.Provider,
 	persistenceConfig *config.Persistence,
+	dynamicCollection *dynamicconfig.Collection,
 ) searchattribute.MapperProvider {
+	enableMapperFromNamespace := persistenceConfig.IsSQLVisibilityStore() ||
+		persistenceConfig.IsCustomVisibilityStore() ||
+		persistenceConfig.IsElasticsearchVisibilityStore()
```
rodrigozhou: Isn't this always true?

service/frontend/operator_handler.go
rodrigozhou: Wondering if we could have a single function that handles both ES and SQL with your changes.

rodrigozhou (changes requested):

common/resource/fx.go
awln-temporal: yeah, i was wondering if it makes sense to keep this in case there are other visibility stores onboarded in the future, but it's better to keep default as true

common/namespace/nsregistry/registry.go
awln-temporal: yeah this should fail startup.

i'm fine with returning early with the error, don't see much utility in storing the error for later as of now.

common/dynamicconfig/constants.go
awln-temporal: so the rollout would be in the following steps:

1. Create pre-allocated columns
2. Release this PR to backfill mappings to Namespace Detail
3. Adding search attributes goes to both namespace detail and custom mapper override
4. Disable custom mapper override through dc to rollout the unified mapper.

The DC allows us to control the rollout of the mapper in case something goes wrong, does this sound right to you?

common/resource/fx.go
rodrigozhou: Any future visibility store will have to handle this natively. There's no need to keep this check.

common/dynamicconfig/constants.go
rodrigozhou: We are not removing the custom mapper support. This could break backwards compatibility. If the user sets a custom mapper, it should always take precedence.

common/namespace/nsregistry/registry.go
awln-temporal: cluster metadata can contain the list of preallocated custom search attribute field names, along with the user registered custom search attributes. we should only backfill registered custom search attributes.

common/namespace/nsregistry/registry.go
awln-temporal: Originally, we thought this needed to be moved to the frontend handler, but as long as registering search attributes go through frontend and generate replication tasks, I believe we can assume the cluster metadata is identical? I can move this to the frontend handler to be safe though.

hehaifengcn:
Is it going to affect how searchattribute mapper works in Temporal?

common/namespace/nsregistry/registry.go
awln-temporal: moved this to frontend to be safe and reduce code duplication.

rodrigozhou:
@hehaifengcn It only affects for Elasticsearch to match the experience with SQL databases.

There won't be any impact in the user experience using custom search attributes. The main difference will be the introduction of maximum number of custom search attributes the user can create (currently unlimited).

.github/workflows/run-tests.yml
```diff
@@ -495,6 +502,21 @@ jobs:
...
+        if: ${{ toJson(matrix.containers) != '[]' && (contains(matrix.containers, 'elasticsearch') || contains(matrix.containers, 'elasticsearch8') || contains(matrix.containers, 'opensearch2') || contains(matrix.containers, 'opensearch3')) }}
+        run: |
+          echo "Waiting for visibility store on 127.0.0.1:9200..."
+          for i in $(seq 1 60); do
+            if curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:9200 | grep -q 200; then
+              echo "Ready after ${i}s"
+              exit 0
+            fi
+            echo "Attempt $i/60: not ready"
+            sleep 2
```
rodrigozhou: If you're sleeping 2 seconds between attempts, then it should be `"Ready after $((i*2))s"`, right?

common/resource/fx.go
```diff
@@ -154,13 +154,13 @@ func SearchAttributeMapperProviderProvider(
...
 	searchAttributeProvider searchattribute.Provider,
-	persistenceConfig *config.Persistence,
+	dynamicCollection *dynamicconfig.Collection,
 ) searchattribute.MapperProvider {
 	return searchattribute.NewMapperProvider(
 		saMapper,
 		namespaceRegistry,
 		searchAttributeProvider,
-		persistenceConfig.IsSQLVisibilityStore() || persistenceConfig.IsCustomVisibilityStore(),
+		true,
```
rodrigozhou: I wonder if we could just remove this argument since now it's always mapping per namespace.

schema/elasticsearch/visibility/versioned/v15/index_template_v7.json
```diff
@@ -0,0 +1,581 @@
...
+      },
+      "TemporalKeywordList01": {
+        "type": "keyword"
+      },
+      "TemporalKeywordList02": {
+        "type": "keyword"
+      },
+      "Bool01": {
+        "type": "boolean"
+      },
```
rodrigozhou, line 169: Only need to create for the default amount of custom search attributes.

service/frontend/fx.go
```diff
@@ -112,6 +113,7 @@ var Module = fx.Options(
 	fx.Provide(NexusEndpointRegistryProvider),
 	fx.Invoke(ServiceLifetimeHooks),
 	fx.Invoke(EndpointRegistryLifetimeHooks),
+	fx.Invoke(SearchAttributeMappingsBackfillLifetimeHooks),
```
rodrigozhou: This should be at the top level fx, not in the frontend.

temporal/fx.go
```diff
@@ -651,7 +651,7 @@ func ApplyClusterMetadataConfigProvider(
 	}
 	indexSearchAttributes := make(map[string]*persistencespb.IndexSearchAttributes)
 	for _, ds := range visDataStores {
-		if ds.SQL != nil || ds.CustomDataStoreConfig != nil {
+		if ds.Elasticsearch != nil || ds.SQL != nil || ds.CustomDataStoreConfig != nil {
 			indexSearchAttributes[ds.GetIndexName()] = sadefs.GetDBIndexSearchAttributes(visCSAOverride)
```
rodrigozhou: This should always be true, right?

rodrigozhou (changes requested):

.github/workflows/run-tests.yml
awln-temporal: Removing this, was trying to see if cass_es8 would succeed, but it seems unrelated to my changes.

temporal/fx.go
awln-temporal: yeah, meant to actually remove this since this would be done in a separate commit, but will change to always true for now.

service/frontend/fx.go
awln-temporal: this is a backfill call to the frontend handler, which I am not sure we want to move to the top level fx. If moved to top level fx, every pod's start would depend on frontend starting up successfully, otherwise the call would throw an error.

.cursor/commands/fork.md
rodrigozhou, line 1: What is this change?

.github/workflows/run-tests.yml
rodrigozhou, line 1: PR https://github.com/temporalio/temporal/pull/9312/ already does these changes. Please rebase.

service/frontend/fx.go
```diff
@@ -975,3 +977,35 @@ func EndpointRegistryLifetimeHooks(lc fx.Lifecycle, registry nexus.EndpointRegis
...
+				ctx,
+				metadataManager,
+				clusterMetadataManager,
+				currentClusterName,
+				visibilityIndexName,
+				handler,
+				logger,
+			)
+		},
+	})
```
rodrigozhou: Do we need to call this as a hook?

service/frontend/namespace_search_attributes_backfill.go
rodrigozhou, line 1: rename file to `namespace_search_attributes_backfill.go`

service/frontend/namespace_backfill.go
```diff
@@ -0,0 +1,116 @@
...
+	"go.temporal.io/server/common/persistence"
+	"go.temporal.io/server/common/searchattribute/sadefs"
+)
+
+const backfillListNamespacesPageSize = 1000
+
+// InitializeSearchAttributeMappings backfills namespace custom search attribute
+// aliases from cluster metadata with identity mappings. Runs on frontend startup
+// when Elasticsearch visibility is configured.
+func InitializeSearchAttributeMappings(
```
rodrigozhou: This function doesn't need to be exported, does it?

common/resource/fx.go
rodrigozhou: Need to remove this line, right?

rodrigozhou (changes requested):

.cursor/commands/fork.md
awln-temporal: not sure how this was included, removing

service/frontend/fx.go
awln-temporal: no, you're right, removed the startup hook.

rodrigozhou (approved):

service/frontend/namespace_search_attributes_backfill.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	nsName := nsDetail.Namespace.Info.Name
+	_, err := handler.UpdateNamespace(ctx, &workflowservice.UpdateNamespaceRequest{
+		Namespace: nsName,
+		Config: &namespacepb.NamespaceConfig{
+			CustomSearchAttributeAliases: upsertFieldToAliasMap,
+		},
+	})
+	if err != nil {
```
awln-temporal, line 98: need to gracefully handle error due to invalid sequence id

code4dc:
Is it a correct summary that once this ticket is merged we'll be able to have per-namespace search-attributes in a self-hosted cluster with an elasticsearch visibility store? 

Is there a timeline for this ticket?

PR #8635, opened 2025-11-13
Worker-Versioning GA: Tests for AutoUpgrade workflows not bouncing back and forth. 

Shivs11:
## What changed?
- WISOTT

## Why?
- To be sure of the revision number mechanics we have implemented.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
 - None until we flip the DC switch :) 

---

> [!NOTE]
> Enables and rewrites AutoUpgrade no-bounce tests using revision numbers, adds several new lag/child/CAN/retry tests, and updates idle workflow poller to accept context for cancellable polling.
> 
> - **Tests (worker versioning v3)**:
>   - **Enable/Rewrite**: Implement `TestAutoUpgradeWorkflows_NoBouncingBetweenVersions` using revision-number mechanics and cancellable idle pollers.
>   - **New/Expanded Scenarios (revision numbers)**:
>     - Workflow/Activity TQ lag transitions: `TestWorkflowTQLags_DependentActivityStartsTransition`, `TestActivityTQLags_DependentActivityCompletesOnTheNewVersion`.
>     - Child workflow behaviors across TQ/version lag: `TestChildStartsWithParentRevision_SameTQ_TQAhead`, `TestChildStartsWithParentRevision_SameTQ_TQLags`, `TestChildStartsWithNoInheritedAutoUpgradeInfo_CrossTQ`.
>     - Continue-as-new and retry stability: `TestContinueAsNewOfAutoUpgradeWorkflow_RevisionNumberMechanics`, `testRetryNoBounceBack` (and callers) to avoid bounce-back under rollback.
>   - **Skip gating**: Gate several tests on `useRevisionNumbers` instead of `useNewDeploymentData`.
> - **Infrastructure**:
>   - `idlePollWorkflow` now takes `context.Context` and passes it via `taskpoller.WithContext`, enabling time-bounded idle pollers; update all call sites accordingly.
>   - Remove ad-hoc `sync.WaitGroup` uses in favor of context-based cancellation/timeouts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3a6fab30587623f164791b5ac105fcc0b395a75f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -347,7 +347,7 @@ func (s *Versioning3Suite) testPinnedQuery_DrainedVersion(pollersPresent bool, r
 	// create version v1 and make it current
 	idlePollerDone := make(chan struct{})
 	go func() {
-		s.idlePollWorkflow(nil, tv, true, ver3MinPollTime, "should not have gotten any tasks since there are none")
+		s.idlePollWorkflow(context.Background(), tv, true, ver3MinPollTime, "should not have gotten any tasks since there are none")
 		close(idlePollerDone)
 	}()
 	s.setCurrentDeployment(tv)
@@ -3655,19 +3655,9 @@ func (s *Versioning3Suite) verifyVersioningSAs(
 }
 
 func (s *Versioning3Suite) TestAutoUpgradeWorkflows_NoBouncingBetweenVersions() {
-	// if !s.useRevisionNumbers {
-	// 	s.T().Skip("This test is only supported on revision number mechanics")
-	// }
-	/*
-
-		Test plan:
-		- Use only one read and write partition.
-		- Set v1 to be the current version.
-		- Start a workflow on v1 and then make it block on a signal.
-		- While blocked, rollback the userData to v0.
-		- Unblock the workflow and verify that it completes on v1.
-
-	*/
+	if !s.useRevisionNumbers {
+		s.T().Skip("This test is only supported on revision number mechanics")
+	}
 
 	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
 	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
@@ -3680,7 +3670,7 @@ func (s *Versioning3Suite) TestAutoUpgradeWorkflows_NoBouncingBetweenVersions() 
 		return "v1", nil
 	}
 
-	// Define the v1 workers
+	// v1 workers
 	w1 := worker.New(s.SdkClient(), tv1.TaskQueue().GetName(), worker.Options{
 		DeploymentOptions: worker.DeploymentOptions{Version: tv1.SDKDeploymentVersion(), UseVersioning: true, DefaultVersioningBehavior: workflow.VersioningBehaviorAutoUpgrade},
 	})
@@ -3691,7 +3681,7 @@ func (s *Versioning3Suite) TestAutoUpgradeWorkflows_NoBouncingBetweenVersions() 
 	// Set v1 to be the current version
 	s.setCurrentDeployment(tv1)
 
-	// Start a workflow on v1.
+	// Start a workflow on v1
 	run, err := s.SdkClient().ExecuteWorkflow(context.Background(), sdkclient.StartWorkflowOptions{
 		ID:        tv1.WorkflowID(),
 		TaskQueue: tv1.TaskQueue().GetName(),


tests/versioning_3_test.go
```diff
@@ -3584,8 +3584,176 @@ func (s *Versioning3Suite) verifyVersioningSAs(
 	}, 5*time.Second, 50*time.Millisecond)
 }
 
+func (s *Versioning3Suite) TestAutoUpgradeWorkflow_NoBouncingBetweenVersions() {
+	if !s.useNewDeploymentData {
```
ShahabT: this bool is always false, right? you should use `useRevisionNumbers` because that's the flag that controls this logic. in that case you don't need to set UseRevisionNumberForWorkerVersioning again.

tests/versioning_3_test.go
```diff
@@ -3584,8 +3584,176 @@ func (s *Versioning3Suite) verifyVersioningSAs(
...
+	s.updateTaskQueueDeploymentDataWithRoutingConfig(tv1, &deploymentpb.RoutingConfig{
+		CurrentDeploymentVersion:  worker_versioning.ExternalWorkerDeploymentVersionFromStringV31(tv1.DeploymentVersionString()),
+		CurrentVersionChangedTime: timestamp.TimePtr(time.Now()),
+		RevisionNumber:            1,
+	}, map[string]*deploymentspb.WorkerDeploymentVersionData{tv1.DeploymentVersion().GetBuildId(): &deploymentspb.WorkerDeploymentVersionData{
+		Status: enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT,
+	}}, []string{}, tqTypeWf, tqTypeAct)
+
+	// Wait until all task queue partitions know that v1 is current.
+	s.waitForDeploymentDataPropagation(tv1, versionStatusCurrent, false, tqTypeWf, tqTypeAct)
```
ShahabT: btw, later when we clean up all old tests we should maybe change this function to wait for the particular revision number propagation rather than status of the version.

tests/versioning_3_test.go
```diff
@@ -3584,8 +3584,176 @@ func (s *Versioning3Suite) verifyVersioningSAs(
...
+
+	// Rollback the userData to v1. Using the lower API here.
+	currentData, err := s.GetTestCluster().MatchingClient().GetTaskQueueUserData(context.Background(), &matchingservice.GetTaskQueueUserDataRequest{
+		NamespaceId:   s.NamespaceID().String(),
+		TaskQueue:     tv1.TaskQueue().GetName(),
+		TaskQueueType: tqTypeWf,
+	})
+	s.NoError(err)
+
+	_, err = s.GetTestCluster().MatchingClient().UpdateTaskQueueUserData(context.Background(), &matchingservice.UpdateTaskQueueUserDataRequest{
```
ShahabT: I imagine you use this in other similar tests, no? seems like a good candidate to extract as a helper

tests/versioning_3_test.go
```diff
@@ -3584,8 +3584,176 @@ func (s *Versioning3Suite) verifyVersioningSAs(
...
+			return respondCompleteWorkflow(tv2, vbUnpinned), nil
+		})
+
+	// Verify that the workflow is running on v2.
+	s.EventuallyWithT(func(t *assert.CollectT) {
+		s.verifyWorkflowVersioning(tv2, vbUnpinned, tv2.Deployment(), nil, nil)
+	}, 10*time.Second, 100*time.Millisecond)
+}
+
 func (s *Versioning3Suite) TestAutoUpgradeWorkflows_NoBouncingBetweenVersions() {
```
ShahabT: this test is names same as above, is it testing the same scenario?

tests/versioning_3_test.go
```diff
@@ -3584,8 +3584,176 @@ func (s *Versioning3Suite) verifyVersioningSAs(
...
-	s.T().Skip("This test is flaky right now and shall be fixed in a future PR.") // TODO (Shivam)
+	if !s.useNewDeploymentData {
```
ShahabT: same

tests/versioning_3_test.go
Shivs11: Yes - actually, now that I think about this and read all your latter comments, it's best if we get this PR in after my other open PR goes in.

The reason why I say this is because I have done some refactoring there anyways, which shall reduce the clutter here.

tests/versioning_3_test.go
Shivs11: IMHO I think this test should be thinking of "propagation complete" == "the right revision number has been synced to matching partitions. We should have tests in the worker_deployment_suite that should test out if the right revision number is synced to these partitions alongside the statuses of each version.

I think having this differences, given we have two suites, would make test writing simpler and easier to read for a new user.

tests/versioning_3_test.go
Shivs11: this is done

tests/versioning_3_test.go
Shivs11: duplicate imo - removed.

tests/versioning_3_test.go
```diff
@@ -3355,6 +3357,7 @@ func (s *Versioning3Suite) doPollActivityAndHandle(
 }
 
 func (s *Versioning3Suite) idlePollWorkflow(
+	ctx context.Context,
```
Shivs11, line 3359: this function shall now take in a context. The reason this was done is as follows:

There are some tests that kickstart a v0 poller (v0 is just an example) for the sole purpose of checking if a task were to ever go this worker. These tests initialize them in a waitGroup and then towards the end of the test, just before it gets cleaned up, there is a wg.Wait() present which ensures that this poller completes running before the test can be terminated safely.

However, this does mean that the minimum wait time for the test will never be lesser than the minimum poll time of this v0 poller. This makes our tests long to run. Thus, passing this context and cancelling it is a safe and efficient way to terminate the test.

ShahabT: nice!

PR #8643, opened 2025-11-14
Fix CalculateTaskQueueVersioningInfo non-deterministic bug

ShahabT:
## What changed?
Fixing CalculateTaskQueueVersioningInfo so it does not randomly choose the wrong current or ramping version based on the order of map iteration.

## Why?
Flaky wrong behavior.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None


ShahabT (commented):
LGTM

PR #8647, opened 2025-11-14
Added request ID transition option when creating activity entity.

fretz12:
## What changed?
Added request ID transition option when creating activity entity.

## Why?
The requestID is set by the matching service to a UUID, allowing safe retries if the response is lost. Here we add the request ID transition option so chasm engine handles it for us under the hood. 

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/activity/handler.go
+++ b/chasm/lib/activity/handler.go
@@ -39,7 +39,9 @@ func (h *handler) StartActivityExecution(ctx context.Context, req *activitypb.St
 				// EagerTask: TODO when supported, need to call the same code that would handle the RecordActivityTaskStarted API
 			}, nil
 		},
-		req.GetFrontendRequest())
+		req.GetFrontendRequest(),
+		chasm.WithRequestID(req.GetFrontendRequest().GetRequestId()),
+	)
 
 	if err != nil {
 		return nil, err


chasm/lib/activity/handler.go
```diff
@@ -39,7 +39,9 @@ func (h *handler) StartActivityExecution(ctx context.Context, req *activitypb.St
 				// EagerTask: TODO when supported, need to call the same code that would handle the RecordActivityTaskStarted API
 			}, nil
 		},
-		req.GetFrontendRequest())
+		req.GetFrontendRequest(),
+		chasm.WithRequestID(req.GetFrontendRequest().GetRequestId()),
+	)
```
dandavison, line 44: Should there have been some failing tests when this was missing?

bergundy: Not sure, but that's a good point, I wonder if we want to require a request ID for any entity creation.

bergundy (approved):

chasm/lib/activity/handler.go
fretz12: Would there be any future entities where we don't want to honor idempotency?

PR #8650, opened 2025-11-14
auto enabling priority and fairness

majiru:
## What changed?
Add a new dynamic config for Auto Enabling fairness and priority if we see the relevant tasks coming in.

## Why?
Seamlessly start to transition users who start using the fields over to the new code path.

## How did you test it?
- [X] built
- [x] added new unit test(s)

## Potential risks
Due to storing this in the userdata we are using that interface a bit more, we also need to change the initialization such that we start it before being able to substantiate the defaultQ, this change in initialization might have unintended side effects that I'm not currently seeing.

## Notes

I've left comments for pieces of code that I am less sure or unsure about, or where things different slightly from our original plan.  Leaving as a draft until I can get some new tests for testing this more thoroughly.

### DIFF
--- a/service/matching/task_queue_partition_manager.go
+++ b/service/matching/task_queue_partition_manager.go
@@ -185,8 +185,8 @@ func (pm *taskQueuePartitionManagerImpl) initialize() (retErr error) {
 	if err != nil {
 		return err
 	}
-	defaultQ.Start()
 	pm.defaultQueueFuture.Set(defaultQ, nil)
+	defaultQ.Start()
 	return nil
 }
 


service/matching/task_queue_partition_manager.go
```diff
@@ -197,6 +248,24 @@ func (pm *taskQueuePartitionManagerImpl) AddTask(
 ) (buildId string, syncMatched bool, err error) {
 	var spoolQueue, syncMatchQueue physicalTaskQueueManager
 	directive := params.taskInfo.GetVersionDirective()
+
+	if pm.Partition().IsRoot() && pm.config.AutoEnable && pm.fairnessState == persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_UNSPECIFIED {
+		// TODO(moody): unsure about this check, what is the correct way to check to see if PriorityKey is set?
+		// TODO(moody): This was originally discussed to be a separate API, but is just exposing this through the generic UpdateUserData sufficient?
+		// what is the pro/cons of adding the new API and perhaps invoking that instead? We're going to unload either way...
+		if params.taskInfo.Priority != nil && (params.taskInfo.Priority.FairnessKey != "" || params.taskInfo.Priority.PriorityKey != int32(pm.config.DefaultPriorityKey)) {
```
majiru: Check for a zero value for priority instead of DefaultPriorityKey

common/dynamicconfig/constants.go
```diff
@@ -1382,6 +1382,11 @@ second per poller by one physical queue manager`,
 The metric has 2 dimensions: namespace_id and plugin_name. Disabled by default as this is 
 an optional feature and also requires a metrics collection system that can handle higher cardinalities.`,
 	)
+	MatchingAutoEnable = NewTaskQueueBoolSetting(
+		"matching.autoEnable",
```
dnr: auto-enable what? we probably want `matching.autoEnableFairness` or maybe `matching.autoEnableV2`

common/dynamicconfig/constants.go
```diff
@@ -1382,6 +1382,11 @@ second per poller by one physical queue manager`,
...
+		false,
+		`MatchingAutoEnable automatically enables fairness when a fairness key is seen`,
```
dnr: also when a priority key is seen?

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -134,6 +134,13 @@ message TaskQueueTypeUserData {
     DeploymentData deployment_data = 1;
 
     temporal.api.taskqueue.v1.TaskQueueConfig config = 2;
+
+    enum FairnessState {
+        FAIRNESS_STATE_UNSPECIFIED = 0;
+        FAIRNESS_STATE_V1 = 1;
+        FAIRNESS_STATE_V2 = 2;
+    };
```
dnr: hmm, I think we usually put enums at the top-level. but I guess this is fine, we don't expect to use it anywhere else.

service/matching/config.go
```diff
@@ -82,6 +82,7 @@ type (
 		MembershipUnloadDelay                    dynamicconfig.DurationPropertyFn
 		TaskQueueInfoByBuildIdTTL                dynamicconfig.DurationPropertyFnWithTaskQueueFilter
 		PriorityLevels                           dynamicconfig.IntPropertyFnWithTaskQueueFilter
+		AutoEnableSub                            dynamicconfig.TypedSubscribableWithTaskQueueFilter[bool]
```
dnr: as we discussed: I don't think this needs to be a subscription, just check it whenever you need to.

service/matching/physical_task_queue_manager_test.go
```diff
@@ -91,9 +91,16 @@ func (s *PhysicalTaskQueueManagerTestSuite) SetupTest() {
 	s.NoError(err)
 	engine.partitions[prtn.Key()] = prtnMgr
 
+	if s.fairness {
+		prtnMgr.config.NewMatcher = true
+		prtnMgr.config.EnableFairness = true
+	} else if s.newMatcher {
+		prtnMgr.config.NewMatcher = true
+	}
```
dnr, line 94: hmm, the stuff on line 70-74 doesn't take care of this? if we do need this, maybe we don't need that?

reading this whole function again... it duplicates a lot of real code to set things up. I wish we could consolidate. but probably not now.

service/matching/task_queue_partition_manager.go
```diff
@@ -63,8 +63,12 @@ type (
 		// TODO(stephanos): move cache out of partition manager
 		cache cache.Cache // non-nil for root-partition
 
+		fairnessState persistencespb.TaskQueueTypeUserData_FairnessState // Set once on initialization and read only after
+		managerReady  *future.FutureImpl[physicalTaskQueueManager]
```
dnr: since this actually contains the default queue, maybe it should be named `defaultQueueFuture`?

service/matching/task_queue_partition_manager.go
```diff
@@ -115,32 +120,74 @@ func newTaskQueuePartitionManager(
...
-	pm.defaultQueue = defaultQ
-	return pm, nil
+	defaultQ.Start()
+	pm.managerReady.Set(defaultQ, nil)
+}
+
+func (pm *taskQueuePartitionManagerImpl) MustGetDefaultQueue(ctx context.Context) physicalTaskQueueManager {
+	queue, err := pm.managerReady.Get(ctx)
+	if err != nil {
+		softassert.Fail(pm.logger, "mustGetDefaultQueue: "+err.Error())
```
dnr: use tags, not dynamic strings
```suggestion
		softassert.Fail(pm.logger, "mustGetDefaultQueue", tag.Error(err))
```

service/matching/task_queue_partition_manager.go
```diff
@@ -115,32 +120,74 @@ func newTaskQueuePartitionManager(
...
+		pm.managerReady.Set(nil, err)
+		return
 	}
-	pm.defaultQueue = defaultQ
-	return pm, nil
+	defaultQ.Start()
+	pm.managerReady.Set(defaultQ, nil)
+}
+
+func (pm *taskQueuePartitionManagerImpl) MustGetDefaultQueue(ctx context.Context) physicalTaskQueueManager {
```
dnr: it's fine to have a helper here, but:

- I think actually this should not take a context or wait at all.. in proper usage, it should only be called after WaitUntilInitialized, right? so we always expect the future to be ready otherwise it's an error. (so it can either Get using a canceled context, or split it into a atomic.Value plus a future for signaling.)
- unfortunately I think we need all the callers to do a nil check and return an internal error (or do nothing). it shouldn't happen if they WaitUntilInitialized as intended, but we shouldn't panic if a caller misuses this.
- name it `mustGetDefaultQueue`? it's only used internally within tqpm, right? actually on second thought name it `defaultQueue()`

service/matching/task_queue_partition_manager.go
```diff
@@ -115,32 +120,74 @@ func newTaskQueuePartitionManager(
...
+		// Fairness is disabled for sticky queues for now so that we can still use TTLs.
+		pm.config.EnableFairness = fairness && pm.partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY
+		if fairness {
+			pm.config.NewMatcher = true
+		} else {
+			pm.config.NewMatcher, pm.cancelNewMatcherSub = pm.config.NewMatcherSub(unload)
+		}
+	case pm.fairnessState == persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_V1:
+		pm.config.NewMatcher = true
+		pm.config.EnableFairness = false
```
dnr: hmm.. this makes me wonder if we should have a fairnessState value that forces NewMatcher = false. of course we can do it with dynamic config too (turn off all three configs), but it could be more consistent to have that option in userdata?

service/matching/task_queue_partition_manager.go
```diff
@@ -115,32 +120,74 @@ func newTaskQueuePartitionManager(
...
+		softassert.Fail(pm.logger, "mustGetDefaultQueue: "+err.Error())
+	}
+	return queue
 }
 
 func (pm *taskQueuePartitionManagerImpl) Start() {
 	pm.engine.updateTaskQueuePartitionGauge(pm.Namespace(), pm.partition, 1)
 	pm.userDataManager.Start()
-	pm.defaultQueue.Start()
+	go pm.initialize(context.Background())
```
dnr: I'm starting to worry about the Background()s 🙁  maybe we really do need a lifecycle context for the tqpm. specifically: what if Stop is called after Start returns but before initialization is complete? ideally it would interrupt initialization and return immediately. it's not great that now Stop would wait for init to succeed all the way or fail.

service/matching/task_queue_partition_manager.go
```diff
@@ -197,6 +250,21 @@ func (pm *taskQueuePartitionManagerImpl) AddTask(
 ) (buildId string, syncMatched bool, err error) {
 	var spoolQueue, syncMatchQueue physicalTaskQueueManager
 	directive := params.taskInfo.GetVersionDirective()
+
+	if pm.Partition().IsRoot() && pm.config.AutoEnable && pm.fairnessState == persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_UNSPECIFIED {
+		if params.taskInfo.Priority != nil && (params.taskInfo.Priority.FairnessKey != "" || params.taskInfo.Priority.PriorityKey != int32(0)) {
```
dnr: simplify?
```suggestion
		if params.taskInfo.Priority.GetFairnessKey() != "" || params.taskInfo.Priority.GetPriorityKey() != int32(0) {
```

service/matching/task_queue_partition_manager.go
```diff
@@ -197,6 +250,21 @@ func (pm *taskQueuePartitionManagerImpl) AddTask(
...
+			updateFn := func(old *persistencespb.TaskQueueUserData) (*persistencespb.TaskQueueUserData, bool, error) {
+				data := common.CloneProto(old)
+				perType := data.GetPerType()[int32(pm.Partition().TaskType())]
+				perType.FairnessState = persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_V2
+				return data, true, nil
+			}
+			_, err := pm.userDataManager.UpdateUserData(ctx, UserDataUpdateOptions{Source: "Matching auto enable"}, updateFn)
+			if err != nil {
+				pm.logger.Error("could not update userdata for autoenable: " + err.Error())
```
dnr: use tags

service/matching/task_queue_partition_manager.go
```diff
@@ -197,6 +250,21 @@ func (pm *taskQueuePartitionManagerImpl) AddTask(
...
+
+	if pm.Partition().IsRoot() && pm.config.AutoEnable && pm.fairnessState == persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_UNSPECIFIED {
+		if params.taskInfo.Priority != nil && (params.taskInfo.Priority.FairnessKey != "" || params.taskInfo.Priority.PriorityKey != int32(0)) {
+			updateFn := func(old *persistencespb.TaskQueueUserData) (*persistencespb.TaskQueueUserData, bool, error) {
+				data := common.CloneProto(old)
+				perType := data.GetPerType()[int32(pm.Partition().TaskType())]
+				perType.FairnessState = persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_V2
+				return data, true, nil
+			}
+			_, err := pm.userDataManager.UpdateUserData(ctx, UserDataUpdateOptions{Source: "Matching auto enable"}, updateFn)
```
dnr: unfortunately, we decided to store user data for all task queue types on the root of the _workflow_ task queue. so this local call can only be done on the workflow tq. for activity, we need to do an rpc to the worker tq.

service/matching/task_queue_partition_manager.go
```diff
@@ -197,6 +250,21 @@ func (pm *taskQueuePartitionManagerImpl) AddTask(
 ) (buildId string, syncMatched bool, err error) {
 	var spoolQueue, syncMatchQueue physicalTaskQueueManager
 	directive := params.taskInfo.GetVersionDirective()
+
+	if pm.Partition().IsRoot() && pm.config.AutoEnable && pm.fairnessState == persistencespb.TaskQueueTypeUserData_FAIRNESS_STATE_UNSPECIFIED {
+		if params.taskInfo.Priority != nil && (params.taskInfo.Priority.FairnessKey != "" || params.taskInfo.Priority.PriorityKey != int32(0)) {
+			updateFn := func(old *persistencespb.TaskQueueUserData) (*persistencespb.TaskQueueUserData, bool, error) {
+				data := common.CloneProto(old)
+				perType := data.GetPerType()[int32(pm.Partition().TaskType())]
```
dnr: you might need to set up the per type map if it's nil here

service/matching/task_queue_partition_manager.go
```diff
@@ -1303,10 +1377,18 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+func (pm *taskQueuePartitionManagerImpl) userDataChanged(to *persistencespb.VersionedTaskQueueUserData) {
 	// Update rateLimits if any change is userData.
 	pm.rateLimitManager.UserDataChanged()
 
+	if !pm.managerReady.Ready() {
+		return
+	}
+	taskType := int32(pm.Partition().TaskType())
+	if to.GetData().GetPerType()[taskType].GetFairnessState() != pm.fairnessState {
+		pm.unloadFromEngine(unloadCauseConfigChange)
```
dnr, line 1759: return after this?

service/matching/task_queue_partition_manager.go
dnr: for the rpc, and even for local UpdateUserData, we need some kind of rate limit or something so we don't do it again immediately on the next task, because updating user data is async.

dnr: also, I'd move this whole thing into another function so there's just one line here, to not break the flow of reading the main path of AddTask

dnr (commented):
for the PR title: it should just talk about auto-enabling priority/fairness. the fact that there's a dynamic config to enable the auto-enabling is the least interesting part.

service/matching/config.go
```diff
@@ -82,7 +82,7 @@ type (
 		MembershipUnloadDelay                    dynamicconfig.DurationPropertyFn
 		TaskQueueInfoByBuildIdTTL                dynamicconfig.DurationPropertyFnWithTaskQueueFilter
 		PriorityLevels                           dynamicconfig.IntPropertyFnWithTaskQueueFilter
-		AutoEnableSub                            dynamicconfig.TypedSubscribableWithTaskQueueFilter[bool]
+		AutoEnable                               dynamicconfig.BoolPropertyFnWithTaskQueueFilter
```
dnr: nit: name AutoEnableV2 also?

service/matching/task_queue_partition_manager.go
```diff
@@ -61,14 +61,16 @@ type (
 		matchingClient      matchingservice.MatchingServiceClient
 		metricsHandler      metrics.Handler // namespace/taskqueue tagged metric scope
 		// TODO(stephanos): move cache out of partition manager
-		cache cache.Cache // non-nil for root-partition
+		cache      cache.Cache // non-nil for root-partition
+		autoEnable bool
```
dnr: maybe we shouldn't copy this into the struct but just read it when we need?

I'm thinking along the lines of: we find a bug with v2 that makes us want to disable auto-enable to prevent further problems. if we've copied it into the struct, that disabling won't take effect on loaded task queues unless we reload everything.

service/matching/task_queue_partition_manager.go
```diff
@@ -123,26 +126,28 @@ func newTaskQueuePartitionManager(
 	return pm, nil
 }
 
-func (pm *taskQueuePartitionManagerImpl) initialize(ctx context.Context) {
+var errDefaultQueueNotInit = serviceerror.NewInternal("defaultQueue is not initializaed")
```
dnr: put this at the top of the file

service/matching/task_queue_partition_manager.go
```diff
@@ -61,14 +61,16 @@ type (
...
 		// TODO(stephanos): move cache out of partition manager
-		cache cache.Cache // non-nil for root-partition
+		cache      cache.Cache // non-nil for root-partition
+		autoEnable bool
 
-		fairnessState persistencespb.TaskQueueTypeUserData_FairnessState // Set once on initialization and read only after
-		managerReady  *future.FutureImpl[physicalTaskQueueManager]
+		fairnessState      persistencespb.TaskQueueTypeUserData_FairnessState // Set once on initialization and read only after
+		defaultQueueFuture *future.FutureImpl[physicalTaskQueueManager]
+		initCtx            context.Context
```
dnr: maybe just a few-word comment on what this is for?

service/matching/task_queue_partition_manager.go
```diff
@@ -163,31 +168,33 @@ func (pm *taskQueuePartitionManagerImpl) initialize(ctx context.Context) {
...
 	defaultQ.Start()
-	pm.managerReady.Set(defaultQ, nil)
+	pm.defaultQueueFuture.Set(defaultQ, nil)
 }
 
-func (pm *taskQueuePartitionManagerImpl) MustGetDefaultQueue(ctx context.Context) physicalTaskQueueManager {
-	queue, err := pm.managerReady.Get(ctx)
+func (pm *taskQueuePartitionManagerImpl) defaultQueue() physicalTaskQueueManager {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
```
dnr: I kinda hate that we have to this (I did it somewhere else too). this is our future implementation, I think we should just add a GetIfReady to it.

client/matching/client.go
```diff
@@ -153,6 +153,24 @@ func (c *clientImpl) QueryWorkflow(ctx context.Context, request *matchingservice
 	return client.QueryWorkflow(ctx, request, opts...)
 }
 
+func (c *clientImpl) EnablePriorityAndFairness(
```
dnr: This shouldn't get a manual implementation, it should use the generated one. You may need to tweak the generation code to pick the right partition. It should work like all the ones that mess with userdata, e.g. UpdateTaskQueueConfigRequest.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -641,3 +641,11 @@ message DescribeWorkerRequest {
 message DescribeWorkerResponse {
     temporal.api.worker.v1.WorkerInfo worker_info = 1;
 }
+
+message EnablePriorityAndFairnessRequest {
```
dnr: We have so many rpcs.. I'm wondering if it would easier to make this another field of UpdateTaskQueueConfigRequest, like there could either be an updated_taskqueue_config, or a internal_set_fairness_state. I guess overall it's not that much simpler?

Anyway, can we name this SetFairnessState since that's a more direct description of what it does. And it should take a FairnessState to set, so it can be used for setting to V1 or V0.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
```diff
@@ -160,3 +162,10 @@ message VersionedTaskQueueUserData {
     TaskQueueUserData data = 1;
     int64 version = 2;
 }
+
+enum FairnessState {
```
dnr: Put this at the top of the file between imports and the first message

service/matching/config.go
```diff
@@ -82,6 +82,7 @@ type (
 		MembershipUnloadDelay                    dynamicconfig.DurationPropertyFn
 		TaskQueueInfoByBuildIdTTL                dynamicconfig.DurationPropertyFnWithTaskQueueFilter
 		PriorityLevels                           dynamicconfig.IntPropertyFnWithTaskQueueFilter
+		AutoEnable                               dynamicconfig.BoolPropertyFnWithTaskQueueFilter
```
dnr: nit: name AutoEnableV2 also?

service/matching/config.go
```diff
@@ -221,6 +223,7 @@ const (
 	loadCauseOtherRead  // any other read-only rpc
 	loadCauseOtherWrite // any other mutating rpc
 	loadCauseForce      // root partition loaded, force load to ensure matching with back logged partitions
+	loadCauseAutoEnable
```
dnr: this should basically never be an actual load cause since the root should be loaded while any child is loaded. just use loadCauseOtherWrite to keep it simpler?

service/matching/matching_engine.go
```diff
@@ -3342,3 +3342,36 @@ func unsetRampingFromRoutingConfig(
 	rc.RampingVersionPercentageChangedTime = nil
 	rc.RampingVersionChangedTime = nil
 }
+
+func (e *matchingEngineImpl) EnablePriorityAndFairness(
```
dnr: can you move this after the last rpc handler in this file, above these little helpers?

service/matching/matching_engine.go
```diff
@@ -3342,3 +3342,36 @@ func unsetRampingFromRoutingConfig(
...
+) (*matchingservice.EnablePriorityAndFairnessResponse, error) {
+	partition, err := tqid.PartitionFromProto(req.TaskQueue, req.NamespaceId, enumspb.TASK_QUEUE_TYPE_WORKFLOW)
+	if err != nil {
+		return nil, err
+	}
+
+	if partition.Kind() != enumspb.TASK_QUEUE_KIND_NORMAL {
+		return nil, serviceerror.NewUnimplemented("fairness and priority not implemented for non-normal queues")
+	}
+	pm, _, err := e.getTaskQueuePartitionManager(ctx, partition, false, loadCauseAutoEnable)
```
dnr: should use true to load on demand. I mean, it should already be loaded, but if it's somehow not, just load it.

service/matching/matching_engine.go
```diff
@@ -3342,3 +3342,36 @@ func unsetRampingFromRoutingConfig(
...
+	if err != nil {
+		return nil, err
+	}
+	if pm == nil {
+		return nil, serviceerror.NewInvalidArgument("task queue provided does not exist")
+	}
```
dnr: this shouldn't happen if you pass true above

service/matching/matching_engine.go
```diff
@@ -3342,3 +3342,36 @@ func unsetRampingFromRoutingConfig(
...
+
+	updateFn := func(old *persistencespb.TaskQueueUserData) (*persistencespb.TaskQueueUserData, bool, error) {
+		data := common.CloneProto(old)
+		perType := data.GetPerType()[int32(pm.Partition().TaskType())]
+		perType.FairnessState = persistencespb.FAIRNESS_STATE_V2
```
dnr: perType might be nil here

service/matching/physical_task_queue_manager_test.go
```diff
@@ -334,13 +341,17 @@ func (s *PhysicalTaskQueueManagerTestSuite) TestAddTaskStandby() {
 	s.tqMgr.namespaceRegistry = mockNamespaceCache
 
 	s.tqMgr.Start()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Millisecond)
```
dnr: maybe give 100ms even for a unit test

service/matching/task_queue_partition_manager.go
```diff
@@ -96,51 +103,103 @@ func newTaskQueuePartitionManager(
...
 	if pm.partition.IsRoot() {
 		pm.cache = cache.New(10000, &cache.Options{
 			TTL: max(1, tqConfig.TaskQueueInfoByBuildIdTTL())}, // ensure TTL is never zero (which would disable TTL)
 		)
 	}
 
+	return pm, nil
+}
+
+var errDefaultQueueNotInit = serviceerror.NewInternal("defaultQueue is not initializaed")
```
dnr: put this at the top of the file

service/matching/task_queue_partition_manager.go
```diff
@@ -96,51 +103,103 @@ func newTaskQueuePartitionManager(
...
 	}
-	pm.defaultQueue = defaultQ
-	return pm, nil
+	defaultQ.Start()
+	pm.defaultQueueFuture.Set(defaultQ, nil)
+}
+
+func (pm *taskQueuePartitionManagerImpl) defaultQueue() physicalTaskQueueManager {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
```
dnr: I kinda hate that we have to this (I did it somewhere else too). this is our own future implementation, I think we should just add a GetIfReady to it.

service/matching/task_queue_partition_manager.go
```diff
@@ -150,8 +209,11 @@ func (pm *taskQueuePartitionManagerImpl) GetRateLimitManager() *rateLimitManager
 // Stop does not unload the partition from matching engine. It is intended to be called by matching engine when
 // unloading the partition. For stopping and unloading a partition call unloadFromEngine instead.
 func (pm *taskQueuePartitionManagerImpl) Stop(unloadCause unloadCause) {
-	pm.versionedQueuesLock.Lock()
-	defer pm.versionedQueuesLock.Unlock()
+	pm.initCancel()
+	queue, err := pm.defaultQueueFuture.Get(context.Background())
```
dnr, line 216: this could block? or is initialize() guaranteed to resolve the future quickly after initCtx is canceled?

service/matching/task_queue_partition_manager.go
```diff
@@ -180,15 +243,44 @@ func (pm *taskQueuePartitionManagerImpl) Namespace() *namespace.Namespace {
...
+	queue, err := pm.defaultQueueFuture.Get(ctx)
 	if err != nil {
 		return err
 	}
-	return pm.defaultQueue.WaitUntilInitialized(ctx)
+	return queue.WaitUntilInitialized(ctx)
+}
+
+func (pm *taskQueuePartitionManagerImpl) autoEnableIfNeeded(ctx context.Context, params addTaskParams) {
+	if !pm.autoEnable || !pm.Partition().IsRoot() || pm.Partition().Kind() == enumspb.TASK_QUEUE_KIND_STICKY {
```
dnr: I think we shouldn't resolve+copy the autoEnable bool into the tqpm struct, just get it on-demand. (and evaluate it after the root and sticky check since those are cheaper)

service/matching/task_queue_partition_manager.go
```diff
@@ -1303,10 +1443,24 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
 }
 
-func (pm *taskQueuePartitionManagerImpl) userDataChanged() {
+func (pm *taskQueuePartitionManagerImpl) userDataChanged(to *persistencespb.VersionedTaskQueueUserData) {
 	// Update rateLimits if any change is userData.
 	pm.rateLimitManager.UserDataChanged()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	defaultQ, err := pm.defaultQueueFuture.Get(ctx)
```
dnr: just use defaultQueue()?

dnr (commented):
(trying out new code review tool, apologies if any of the comments end up on the wrong lines)

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
majiru: I had it this way originally as well, but the proto linter disagrees with this (a bit silly I know):

```
- file_path: temporal/server/api/persistence/v1/task_queues.proto
  problems:
  - message: Messages should precede all top-level enums.
    location:
      start_position:
        line_number: 11
        column_number: 1
      end_position:
        line_number: 16
        column_number: 1
    rule_id: core::0191::file-layout
    rule_doc_uri: https://linter.aip.dev/191/file-layout
```

service/matching/task_queue_partition_manager.go
majiru: I avoided it because if the queue is not ready defaultQueue() has that softassert in it, in this case it not being ready yet is more expected and not really an error, so I didn't want to spam the output for it. Perhaps that should be in a comment there? Or maybe defaultQueue takes another flag? If we add a GetIfReady() to our future implementation I think my preference would be to use that and a comment why we dont use defaultQueue()

service/matching/task_queue_partition_manager.go
majiru: It's the later, its a bit tricky, at some point in writing this I had a comment block explaining my thinking about but I must have removed it before pushing. You asking this is probably more evidence that I should put something  back there.

My thinking on this is initialization can be in one of three states, it could be before we make the call to wait on the userdatainitialization, currently blocked on initialization, or finishing up the initialization after the wait call returns. In the first two, cancelling the context will short circuit the rest of initialization and not block Stop. However the last one is a bit more complex, there isn't a great spot to short circuit the initialization after the wait call, and we don't want to race with the subscriptions. So my thought was that as long as there is nothing that can block us for long after that first Wait call, then we should just let it finish first before tearing down. The idea being that the rest of initialization should not have any more blocking calls.

I'm not in love with this solution, and I'm open to discussing alternatives, but I think this works "ok" as-is.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -641,3 +641,18 @@ message DescribeWorkerRequest {
 message DescribeWorkerResponse {
     temporal.api.worker.v1.WorkerInfo worker_info = 1;
 }
+
+// (-- api-linter: core::0134::request-resource-required=disabled
+//     aip.dev/not-precedent: UpdateFairnessStateRequest RPC doesn't follow Google API format. --)
+// (-- api-linter: core::0134::request-mask-required=disabled
+//     aip.dev/not-precedent: UpdateFairnessStateRequest RPC doesn't follow Google API format. --)
+// (-- api-linter: core::0134::method-signature=disabled
+//     aip.dev/not-precedent: UpdateFairnessStateRequest RPC doesn't follow Google API format. --)
```
dnr: test comment, please ignore

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -641,3 +641,18 @@ message DescribeWorkerRequest {
...
+message UpdateFairnessStateRequest {
```
dnr: second test comment, please ignore

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
dnr, line 651: should we just call it SetFairnessState to avoid this silly lint stuff? :)

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -641,3 +641,18 @@ message DescribeWorkerRequest {
...
+    string namespace_id = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
```
dnr: so, all the rpcs that mess with user data use just a `string` for `task_queue`, which seems weird when we have this TaskQueue message, but overall I think it makes things simpler. (and probably that message shouldn't have existed, but in any case can be left for workflow progress apis only.)

that way you don't have to worry about sticky queues in the handler.

service/matching/config.go
```diff
@@ -308,6 +310,7 @@ func NewConfig(
 		FairnessKeyRateLimitCacheSize:            dynamicconfig.MatchingFairnessKeyRateLimitCacheSize.Get(dc),
 		MaxFairnessKeyWeightOverrides:            dynamicconfig.MatchingMaxFairnessKeyWeightOverrides.Get(dc),
 		MaxIDLengthLimit:                         dynamicconfig.MaxIDLengthLimit.Get(dc),
+		AutoEnableV2:                             dynamicconfig.MatchingAutoEnableV2.Get(dc),
```
dnr: I know it's a pain but I try to keep these in similar order so it's easier to see if we missed anything. put AutoEnableV2 immediately following EnableMigration in all structs/initalizers?

service/matching/matching_engine.go
```diff
@@ -3225,6 +3225,44 @@ func (e *matchingEngineImpl) UpdateTaskQueueConfig(
...
+	ctx context.Context,
+	req *matchingservice.UpdateFairnessStateRequest,
+) (*matchingservice.UpdateFairnessStateResponse, error) {
+	partition, err := tqid.PartitionFromProto(req.TaskQueue, req.NamespaceId, enumspb.TASK_QUEUE_TYPE_WORKFLOW)
+	if err != nil {
+		return nil, err
+	}
+
+	if partition.Kind() != enumspb.TASK_QUEUE_KIND_NORMAL {
+		return nil, serviceerror.NewUnimplemented("fairness and priority not implemented for non-normal queues")
```
dnr: I would say InvalidArgument instead, but actually I think this will go away if you just use a string for the task queue name

service/matching/matching_engine.go
```diff
@@ -3225,6 +3225,44 @@ func (e *matchingEngineImpl) UpdateTaskQueueConfig(
...
+	if err != nil {
+		return nil, err
+	}
+
+	updateFn := func(old *persistencespb.TaskQueueUserData) (*persistencespb.TaskQueueUserData, bool, error) {
+		data := common.CloneProto(old)
+		if data.PerType == nil {
+			data.PerType = make(map[int32]*persistencespb.TaskQueueTypeUserData)
+		}
+		typ := int32(pm.Partition().TaskType())
```
dnr: this will always be workflow, right? because we're "in" the workflow queue here. but we might be changing state on the activity queue. so we need to pass the type as part of the request.

service/matching/task_queue_partition_manager.go
```diff
@@ -96,51 +104,98 @@ func newTaskQueuePartitionManager(
...
 	if pm.partition.IsRoot() {
 		pm.cache = cache.New(10000, &cache.Options{
 			TTL: max(1, tqConfig.TaskQueueInfoByBuildIdTTL())}, // ensure TTL is never zero (which would disable TTL)
 		)
 	}
 
+	return pm, nil
+}
+
+func (pm *taskQueuePartitionManagerImpl) initialize() {
```
dnr: I just thought of this, but what do you think about this structure:

```suggestion
func (pm *taskQueuePartitionManagerImpl) initialize() (retErr error) {
	defer func() { pm.defaultQueueFuture.SetIfNotReady(nil, retErr) }
```

then there's no way to accidentally return without the future being resolved one way or the other.

service/matching/task_queue_partition_manager.go
```diff
@@ -180,15 +239,45 @@ func (pm *taskQueuePartitionManagerImpl) Namespace() *namespace.Namespace {
...
+	queue, err := pm.defaultQueueFuture.Get(ctx)
 	if err != nil {
 		return err
 	}
-	return pm.defaultQueue.WaitUntilInitialized(ctx)
+	return queue.WaitUntilInitialized(ctx)
+}
+
+func (pm *taskQueuePartitionManagerImpl) autoEnableIfNeeded(ctx context.Context, params addTaskParams) {
+	if !pm.config.AutoEnableV2() || !pm.Partition().IsRoot() || pm.Partition().Kind() == enumspb.TASK_QUEUE_KIND_STICKY {
```
dnr: as a micro-optimization, I would put this condition after checking fairnessState and the priority on the task since getting the dynamic config is more expensive.

proto/internal/temporal/server/api/persistence/v1/task_queues.proto
dnr: weird, I really thought I remembered seeing enums and files in the same file. we have most enums in a separate enums package, but I don't see any problem with mixing it like this. so up to you if you want to move it to temporal.server.api.enums

service/matching/physical_task_queue_manager_test.go
dnr: I still don't understand why this change was needed. maybe we can huddle on it.

service/matching/task_queue_partition_manager.go
dnr: I read it over and I'm satisfied. nothing in initialize should block except wait on user data, and that will return immediately on initCancel. it may be worth a one-line comment to say that initCancel() will cause initialize to return immediately if blocked, which will resolve this future, so this shouldn't block.

the only thing to worry about would be a bug in initialize, but I think the defer will fix that.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
majiru: SetFairnessState fails a lint saying Set is a synonym of Update, maybe that results in less comment mess though.

service/matching/matching_engine_test.go
```diff
@@ -211,6 +212,7 @@ func (s *matchingEngineSuite) newConfig() *Config {
 	} else if s.newMatcher {
 		useNewMatcher(res)
 	}
+	res.AutoEnableV2 = func(_, _ string, _ enumspb.TaskQueueType) bool { return true }
```
dnr: I'd just put this in defaultTestConfig, no? it won't affect any of the other tests that use that.

also we have helpers for those:

```suggestion
	res.AutoEnableV2 = dynamicconfig.GetBoolPropertyFnFilteredByTaskQueue(true)
```

service/matching/matching_engine_test.go
```diff
@@ -767,21 +769,60 @@ func (s *matchingEngineSuite) TestPollActivityTaskQueues_NamespaceHandover() {
...
+	mgr.GetUserDataManager().(*mockUserDataManager).onChange = cMgr.userDataChanged
+	s.matchingEngine.updateTaskQueue(dbq.partition, mgr)
+	mgr.Start()
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	err := mgr.WaitUntilInitialized(ctx)
+	s.Require().NoError(err)
+	cancel()
+
+	s.logger.Expect(testlogger.Error, "unexpected error dispatching task", tag.Error(errTaskQueueClosed))
+	s.AddTasksTest(enumspb.TASK_QUEUE_TYPE_WORKFLOW, false, true)
```
dnr: tbh I'd almost rather just see an explicit s.matchingEngine.AddWorkflowTask call here instead of reusing the test. reusing AddTasksTest and having to match the names, etc. overcomplicates that function when we really just need one call here.

service/matching/matching_engine_test.go
```diff
@@ -767,21 +769,60 @@ func (s *matchingEngineSuite) TestPollActivityTaskQueues_NamespaceHandover() {
...
+	data, _, _ := mgr.GetUserDataManager().GetUserData()
+	s.Require().Equal(persistencespb.FAIRNESS_STATE_V2, data.GetData().GetPerType()[int32(enumspb.TASK_QUEUE_TYPE_WORKFLOW)].FairnessState)
+	// At this point the partition manager should be unloaded
+	select {
+	case <-cMgr.initCtx.Done():
+	case <-time.Tick(100 * time.Millisecond):
```
dnr: I think you want time.After? and give it 1s for CI slowness

service/matching/matching_engine_test.go
```diff
@@ -2988,7 +3040,9 @@ func (s *matchingEngineSuite) TestUpdatePhysicalTaskQueueGauge_UnVersioned() {
 	// the size of the map to 1 and it's counter to 1.
 	s.PhysicalQueueMetricValidator(capture, 1, 1)
 
-	tlmImpl, ok := tqm.(*taskQueuePartitionManagerImpl).defaultQueue.(*physicalTaskQueueManagerImpl)
+	defaultQ, err := tqm.(*taskQueuePartitionManagerImpl).defaultQueueFuture.Get(context.Background())
```
dnr: could split up getPhysicalTaskQueueManagerImpl into two functions so you can reuse the inner one here?

service/matching/matching_engine_test.go
```diff
@@ -3223,8 +3277,9 @@ func (s *matchingEngineSuite) pollWorkflowTasksConcurrently(
 
 // getPhysicalTaskQueueManagerImpl extracts the physicalTaskQueueManagerImpl for the given PhysicalTaskQueueKey
 func (s *matchingEngineSuite) getPhysicalTaskQueueManagerImpl(ptq *PhysicalTaskQueueKey) *physicalTaskQueueManagerImpl {
-	pgMgr, ok := s.matchingEngine.partitions[ptq.Partition().Key()].(*taskQueuePartitionManagerImpl).defaultQueue.(*physicalTaskQueueManagerImpl)
-	s.True(ok, "taskQueueManger doesn't implement taskQueuePartitionManager interface")
+	defaultQ, err := s.matchingEngine.partitions[ptq.Partition().Key()].(*taskQueuePartitionManagerImpl).defaultQueueFuture.Get(context.Background())
```
dnr: should this use GetIfReady? or do you need to wait sometimes?

service/matching/task_queue_partition_manager_test.go
```diff
@@ -76,8 +77,10 @@ func (s *PartitionManagerTestSuite) SetupTest() {
 		useNewMatcher(config)
 	}
 
-	matchingClientMock := matchingservicemock.NewMockMatchingServiceClient(s.controller)
-	engine := createTestMatchingEngine(logger, s.controller, config, matchingClientMock, registry)
+	config.AutoEnableV2 = func(_, _ string, _ enumspb.TaskQueueType) bool { return true }
```
dnr: should we use defaultTestConfig instead of NewConfig above? and then we don't need this here if you put AutoEnableV2 in defaultTestConfig.

service/matching/task_queue_partition_manager_test.go
```diff
@@ -625,13 +652,18 @@ func (m *mockUserDataManager) GetUserData() (*persistencespb.VersionedTaskQueueU
 
 func (m *mockUserDataManager) UpdateUserData(_ context.Context, _ UserDataUpdateOptions, updateFn UserDataUpdateFunc) (int64, error) {
 	m.Lock()
```
dnr, line 1041: we should still defer unlock here? you can't read m.data outside of the lock, no?

cmd/tools/genrpcwrappers/main.go
```diff
@@ -340,6 +340,10 @@ func makeGetMatchingClient(reqType reflect.Type) string {
 		tq = fieldWithPath{path: `"not-applicable"`}
 		tqt = fieldWithPath{path: "enumspb.TASK_QUEUE_TYPE_UNSPECIFIED"}
 		nsID = fieldWithPath{path: `"not-applicable"`}
+	case "UpdateFairnessStateRequest":
+		tq = tryFindOneNestedField(t, "TaskQueue", "request", 1)
+		tqt = tryFindOneNestedField(t, "TaskQueueType", "request", 1)
```
dnr: this was correct before and is wrong now.. this request has to be routed to the root workflow partition like all the others that affect userdata.

service/matching/matching_engine_test.go
```diff
@@ -778,10 +780,78 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
+		TaskQueue:     "makeToast",
+		TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+		FairnessState: enumsspb.FAIRNESS_STATE_V2,
+	}
+	const N = 64
+	ch := make(chan struct{}, N)
+	s.mockMatchingClient.EXPECT().UpdateFairnessState(context.Background(), req).DoAndReturn(
+		func(ctx context.Context, req *matchingservice.UpdateFairnessStateRequest, opts ...grpc.CallOption) (resp *matchingservice.UpdateFairnessStateResponse, err error) {
+			resp, err = s.matchingEngine.UpdateFairnessState(ctx, req)
+			ch <- struct{}{}
```
dnr: this is fine, just another suggestion for how to write this: consider having a `var didUpdate atomic.Bool` and then on this line `didUpdate.Store(true)`. then at the bottom, do `s.Eventually(didUpdate.Load, time.Milliscond, time.Second)`

service/matching/matching_engine_test.go
```diff
@@ -778,10 +780,78 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
+	s.Require().NoError(err)
+	select {
+	case <-ch:
+	case <-time.After(time.Second):
+		s.Require().Fail("AddWorkflowTask did not trigger AutoEnable")
+	}
+
+	data, _, _ := mgr.GetUserDataManager().GetUserData()
+	s.Require().Equal(enumsspb.FAIRNESS_STATE_V2, data.GetData().GetPerType()[int32(enumspb.TASK_QUEUE_TYPE_WORKFLOW)].FairnessState)
+	// At this point the partition manager should be unloaded
```
dnr, line 831: something feels weird here.. we expect mgr/cMgr to be unloaded, but we're also asking its userdatamanager for the current state? shouldn't we ask the current instance of the partition manager for this tq?

service/matching/matching_engine_test.go
```diff
@@ -3221,11 +3288,22 @@ func (s *matchingEngineSuite) pollWorkflowTasksConcurrently(
...
+func (s *matchingEngineSuite) getTaskQueuePartitionManagerImpl(ptq *PhysicalTaskQueueKey) *taskQueuePartitionManagerImpl {
+	return s.matchingEngine.partitions[ptq.Partition().Key()].(*taskQueuePartitionManagerImpl)
+}
+
+// getPhysicalTaskQueueManagerImpl extracts the physicalTaskQueueManagerImpl for the given taskQueuePartitionManager
+func (s *matchingEngineSuite) getPhysicalTaskQueueManagerImpl(mgr taskQueuePartitionManager) *physicalTaskQueueManagerImpl {
+	defaultQ, err := mgr.(*taskQueuePartitionManagerImpl).defaultQueueFuture.GetIfReady()
+	if err != nil {
+		panic("BAD")
+	}
```
dnr: I think the Require handles this?

service/matching/task_queue_partition_manager.go
```diff
@@ -96,51 +105,98 @@ func newTaskQueuePartitionManager(
...
 
-	var fairness bool
-	fairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(unload)
-	// Fairness is disabled for sticky queues for now so that we can still use TTLs.
-	tqConfig.EnableFairness = fairness && partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY
-	if fairness {
-		tqConfig.NewMatcher = true
-	} else {
-		tqConfig.NewMatcher, pm.cancelNewMatcherSub = tqConfig.NewMatcherSub(unload)
+	retErr = pm.userDataManager.WaitUntilInitialized(pm.initCtx)
```
dnr: nitpick, but what I meant was to name retErr just so that the defer could see it, but otherwise don't change the code, i.e. still `return err` or `return serviceerror.NewInternal...`

then you don't have any confusion about whether it's being shadowed or not. I think it's correct as-is, just not the usual style.

service/matching/task_queue_partition_manager.go
```diff
@@ -1303,10 +1439,25 @@ func (pm *taskQueuePartitionManagerImpl) getPerTypeUserData() (*persistencespb.T
...
+	// not being ready as an error, which is expected during bringup here.
+	defaultQ, err := pm.defaultQueueFuture.GetIfReady()
+	// Initialization error or not ready yet
+	if err != nil {
+		return
+	}
+
+	taskType := int32(pm.Partition().TaskType())
+	if to.GetData().GetPerType()[taskType].GetFairnessState() != pm.fairnessState {
+		pm.logger.Debug("unloading partitionManager due to change in FairnessState")
```
dnr, line 1758: do you want to leave this in? we already log one line for unload with cause:ConfigChange

service/matching/matching_engine_test.go
majiru: Ah my bad, left that in on accident when I was debugging

service/matching/matching_engine_test.go
majiru: (replicating the reasoning here that was discussed)
This would be ideal but due to how we pass around the mockUserData manager would make it such that a newly created partition manager would not see the modifications of the previous one. This would be nice to fix and make the tests better match reality, but I think outside the scope of this PR. I've added a comment explaining this to the PR.

cmd/tools/genrpcwrappers/main.go
```diff
@@ -340,6 +340,10 @@ func makeGetMatchingClient(reqType reflect.Type) string {
 		tq = fieldWithPath{path: `"not-applicable"`}
 		tqt = fieldWithPath{path: "enumspb.TASK_QUEUE_TYPE_UNSPECIFIED"}
 		nsID = fieldWithPath{path: `"not-applicable"`}
+	case "UpdateFairnessStateRequest":
```
dnr: this can be merged with the case for UpdateTaskQueueConfigRequest?

service/matching/matching_engine.go
```diff
@@ -3280,6 +3280,46 @@ func (e *matchingEngineImpl) UpdateTaskQueueConfig(
...
+		if data.PerType == nil {
+			data.PerType = make(map[int32]*persistencespb.TaskQueueTypeUserData)
+		}
+		typ := int32(req.GetTaskQueueType())
+		perType := data.PerType[typ]
+		if perType == nil {
+			data.PerType[typ] = &persistencespb.TaskQueueTypeUserData{}
+			perType = data.PerType[typ]
+		}
+		perType.FairnessState = req.FairnessState
```
dnr, line 3313: nice, this can use those new helpers:

```suggestion
util.GetOrSetNew(data.PerType, typ).FairnessState = req.FairnessState
```

(don't have to do in this PR, I'm going to search for spots to simplify later)

service/matching/matching_engine_test.go
```diff
@@ -778,10 +781,70 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
 }
 
+func (s *matchingEngineSuite) TestAddWorkflowAutoEnable() {
+	req := &matchingservice.UpdateFairnessStateRequest{
+		NamespaceId:   s.ns.ID().String(),
+		TaskQueue:     "makeToast",
+		TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+		FairnessState: enumsspb.FAIRNESS_STATE_V2,
+	}
+	didUpdate := &atomic.Bool{}
```
dnr: nit: more idiomatic:

```suggestion
var didUpdate atomic.Bool
```

service/matching/matching_engine_test.go
```diff
@@ -778,10 +781,70 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
+	s.mockMatchingClient.EXPECT().UpdateFairnessState(context.Background(), req).DoAndReturn(
+		func(ctx context.Context, req *matchingservice.UpdateFairnessStateRequest, opts ...grpc.CallOption) (resp *matchingservice.UpdateFairnessStateResponse, err error) {
+			resp, err = s.matchingEngine.UpdateFairnessState(ctx, req)
+			didUpdate.Store(true)
+			return
```
dnr: nit: could be

```suggestion
			didUpdate.Store(true)
			return s.matchingEngine.UpdateFairnessState(ctx, req)
```

without named returns?

service/matching/matching_engine_test.go
```diff
@@ -778,10 +781,70 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
+	dbq := newUnversionedRootQueueKey(s.ns.ID().String(), "makeToast", enumspb.TASK_QUEUE_TYPE_WORKFLOW)
+	mgr := s.newPartitionManager(dbq.partition, s.matchingEngine.config)
+	cMgr := mgr.(*taskQueuePartitionManagerImpl)
+	mgr.GetUserDataManager().(*mockUserDataManager).onChange = cMgr.userDataChanged
+	s.matchingEngine.updateTaskQueue(dbq.partition, mgr)
+	mgr.Start()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	err := mgr.WaitUntilInitialized(ctx)
+	s.Require().NoError(err)
+	defer cancel()
```
dnr: might as well just `cancel()` here since we're initialized, right? no need to defer

service/matching/matching_engine_test.go
```diff
@@ -778,10 +781,70 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
+		&matchingservice.AddWorkflowTaskRequest{
+			NamespaceId: s.ns.ID().String(),
+			Execution: &commonpb.WorkflowExecution{
+				WorkflowId: "workflow1",
+				RunId:      uuid.NewString(),
+			},
+			TaskQueue: &taskqueuepb.TaskQueue{
+				Name: "makeToast",
+				Kind: enumspb.TASK_QUEUE_KIND_NORMAL,
+			},
```
dnr: nit: in general using testvars can simplify some of this fake data

service/matching/task_queue_partition_manager.go
```diff
@@ -99,51 +108,99 @@ func newTaskQueuePartitionManager(
...
 }
 
 func (pm *taskQueuePartitionManagerImpl) Start() {
 	pm.engine.updateTaskQueuePartitionGauge(pm.Namespace(), pm.partition, 1)
 	pm.userDataManager.Start()
-	pm.defaultQueue.Start()
+	go func() {
+		// Make the linter happy...
+		_ = pm.initialize()
+	}()
```
dnr: grumble... I would rather use a nolint comment, I don't like complicating the actual code for stupid reasons (nit)

service/matching/task_queue_partition_manager.go
```diff
@@ -183,15 +244,43 @@ func (pm *taskQueuePartitionManagerImpl) Namespace() *namespace.Namespace {
...
 	}
-	return pm.defaultQueue.WaitUntilInitialized(ctx)
+	return queue.WaitUntilInitialized(ctx)
+}
+
+func (pm *taskQueuePartitionManagerImpl) autoEnableIfNeeded(ctx context.Context, params addTaskParams) {
+	if pm.fairnessState != enumsspb.FAIRNESS_STATE_UNSPECIFIED {
+		return
+	}
+	if !pm.config.AutoEnableV2() || !pm.Partition().IsRoot() || pm.Partition().Kind() == enumspb.TASK_QUEUE_KIND_STICKY {
```
dnr: check pm.config.AutoEnableV2() after the other two clauses since they're cheaper

service/matching/task_queue_partition_manager.go
```diff
@@ -183,15 +244,43 @@ func (pm *taskQueuePartitionManagerImpl) Namespace() *namespace.Namespace {
...
+		return
+	}
+	if params.taskInfo.Priority.GetFairnessKey() == "" && params.taskInfo.Priority.GetPriorityKey() == int32(0) {
```
dnr: move this one up above pm.config.AutoEnableV2()

tests/priority_fairness_test.go
```diff
@@ -240,35 +240,97 @@ func wrongorderness(vs []int) float64 {
...
+	} else {
+		dynamicConfigOverrides[dynamicconfig.MatchingUseNewMatcher.Key()] = true
+		dynamicConfigOverrides[dynamicconfig.MatchingEnableFairness.Key()] = true
+	}
 	s.FunctionalTestBase.SetupSuiteWithCluster(testcore.WithDynamicConfigOverrides(dynamicConfigOverrides))
 }
 
+func (s *FairnessSuite) TriggerAutoEnable(tv *testvars.TestVars) {
+	// We need to trigger both a worklow and activity reload.
+	// We gurantee that a successfull call triggers the reload
```
dnr:
```suggestion
	// We need to trigger both a workflow and activity reload.
	// We guarantee that a successful call triggers the reload
```

tests/priority_fairness_test.go
```diff
@@ -240,35 +240,97 @@ func wrongorderness(vs []int) float64 {
...
+		_, err := s.FrontendClient().StartWorkflowExecution(context.Background(), &workflowservice.StartWorkflowExecutionRequest{
+			Namespace:    s.Namespace().String(),
+			WorkflowId:   "trigger",
+			WorkflowType: tv.WorkflowType(),
+			TaskQueue:    tv.TaskQueue(),
+			Priority: &commonpb.Priority{
+				PriorityKey: 3,
+			},
+		})
+		s.T().Log("AutoEnable StartWorkflowExecution:", err)
```
dnr: can this fail? should the StartWorkflowExecution be outside the loop?

tests/priority_fairness_test.go
```diff
@@ -240,35 +240,97 @@ func wrongorderness(vs []int) float64 {
...
+								ScheduleToCloseTimeout: durationpb.New(time.Minute),
+							},
+						},
+					},
+				)
+				return &workflowservice.RespondWorkflowTaskCompletedRequest{Commands: commands}, nil
+			},
+			taskpoller.WithContext(ctx),
+		)
+		if err == nil {
```
dnr: what error might we see here?

service/matching/matching_engine_test.go
dnr: oof, yeah, that is kind of a mess. also manually hooking up userDataChanged... let's think about how to improve it but fine for now.

tests/priority_fairness_test.go
majiru: It can fail, I tried to explain in the top comment what this is doing here. After triggering a reload the original workflow may or may not be enqueued. We need it to be enqueued for us to create an activity from it and trigger autoenable for activities as well. The version there is what I've found to flake the least, always matching the workflow creation with a poll and just looping until we see the activity get created. I added some more comments to hopefully explain this more, open to changing this to something else if it makes more sense but this was the result of a lot of trial and error.

service/matching/matching_engine_test.go
```diff
@@ -778,10 +781,92 @@ func (s *matchingEngineSuite) TestAddWorkflowTasksForwarded() {
...
+		&matchingservice.AddWorkflowTaskRequest{
+			NamespaceId: tv.NamespaceID().String(),
+			Execution:   tv.WorkflowExecution(),
+			TaskQueue:   tv.TaskQueue(),
+			Priority: &commonpb.Priority{
+				PriorityKey: 3,
+			},
+		},
+	)
+	s.Require().NoError(err)
```
dnr, line 863: My only concern here is that the UpdateFairnessState is async, so maybe we should wait a sec to be sure it doesn't arrive? Otherwise I think this could pass by accident. I'm not sure of a better way to test it at this level.

tests/priority_fairness_test.go
```diff
@@ -240,35 +240,116 @@ func wrongorderness(vs []int) float64 {
...
+		WorkflowType: tv.WorkflowType(),
+		TaskQueue:    tv.TaskQueue(),
+		Priority: &commonpb.Priority{
+			PriorityKey: 3,
+		},
+	})
+	s.Require().NoError(err)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	s.Eventually(func() bool {
+		resp, err := s.AdminClient().GetTaskQueueTasks(ctx, &adminservice.GetTaskQueueTasksRequest{
```
dnr, line 292: in other tests I've been using DescribeTQ/Partition and ApproximateBacklogCount, but I forgot this existed.. it's more straightforward. maybe we should switch other tests to just reading the tasks directly.

service/matching/task_queue_partition_manager.go
```diff
@@ -100,55 +109,100 @@ func newTaskQueuePartitionManager(
...
-	defaultQ, err := newPhysicalTaskQueueManager(pm, UnversionedQueueKey(partition))
+	defaultQ, err := newPhysicalTaskQueueManager(pm, UnversionedQueueKey(pm.partition))
 	if err != nil {
-		return nil, err
+		return err
 	}
-	pm.defaultQueue = defaultQ
-	return pm, nil
+	defaultQ.Start()
+	pm.defaultQueueFuture.Set(defaultQ, nil)
```
spkane31: Flip the order of these two per @prathyushpv comments offline

PR #8653, opened 2025-11-16
Add standalone activity completion and failure support

fretz12:
## What changed?
Added standalone activity completion and failure handling. Refactored existing timeout failure handling. Refactored existing check for retry method.

## Why?
Needed to support standalone activities full operation. 

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/tests/standalone_activity_test.go
+++ b/tests/standalone_activity_test.go
@@ -899,33 +899,6 @@ func (s *standaloneActivityTestSuite) assertActivityExecutionInfo(
 	// info.LastDeploymentVersion
 }
 
-//func (s *standaloneActivityTestSuite) startActivity(ctx context.Context, activityID string, taskQueue string) (*workflowservice.StartActivityExecutionResponse, error) {
-//	return s.FrontendClient().StartActivityExecution(ctx, &workflowservice.StartActivityExecutionRequest{
-//		Namespace:  s.Namespace().String(),
-//		ActivityId: activityID,
-//		ActivityType: &commonpb.ActivityType{
-//			Name: "test-activity-type",
-//		},
-//		Input: &commonpb.Payloads{
-//			Payloads: []*commonpb.Payload{
-//				{
-//					Metadata: map[string][]byte{
-//						"encoding": []byte("json/plain"),
-//					},
-//					Data: []byte("test-activity-input"),
-//				},
-//			},
-//		},
-//		Options: &activitypb.ActivityOptions{
-//			TaskQueue: &taskqueuepb.TaskQueue{
-//				Name: taskQueue,
-//			},
-//			StartToCloseTimeout: durationpb.New(1 * time.Minute),
-//		},
-//		RequestId: "test-request-id",
-//	})
-//}
-
 func (s *standaloneActivityTestSuite) pollActivityTaskQueue(ctx context.Context, taskQueue string) (*workflowservice.PollActivityTaskQueueResponse, error) {
 	return s.FrontendClient().PollActivityTaskQueue(ctx, &workflowservice.PollActivityTaskQueueRequest{
 		Namespace: s.Namespace().String(),


chasm/lib/activity/activity.go
```diff
@@ -138,6 +134,12 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
 	}, nil
 }
 
+// RecordActivityTaskStartedParams holds parameters for RecordActivityTaskStarted
+type RecordActivityTaskStartedParams struct {
+	VersionDirective *taskqueuespb.TaskVersionDirective
+	WorkerIdentity   string
+}
+
```
dandavison: Let's think again about naming of methods in this file. Do we definitely want the stutter (`activity.RecordActivityTaskStarted` vs `activity.RecordStarted`)? Is "Task" appropriate/desired? I don't think we want the inconsistency between `ActivityTaskStarted` and `ActivityCompleted`.

chasm/lib/activity/activity.go
```diff
@@ -225,9 +227,75 @@ func (a *Activity) RecordCompletion(ctx chasm.MutableContext, applyFn func(ctx c
...
+// RecordActivityCompleted updates the activity on activity completion.
+func (a *Activity) RecordActivityCompleted(ctx chasm.MutableContext, params RecordActivityCompletedParams) (*historyservice.RespondActivityTaskCompletedResponse, error) {
+	if err := TransitionCompleted.Apply(a, ctx, params); err != nil {
+		return nil, err
+	}
+
+	return &historyservice.RespondActivityTaskCompletedResponse{}, nil
+}
```
dandavison, line 229: Not urgent to address in this PR but I'm uncomfortable with APIs in `chasm/lib` returning `historyservice` response structs, especially when they take in their own "params" struct rather than the `historyservice` request.

chasm/lib/activity/activity.go
```diff
@@ -225,9 +227,75 @@ func (a *Activity) RecordCompletion(ctx chasm.MutableContext, applyFn func(ctx c
...
+
+// RecordActivityFailedParams holds parameters for HandleActivityFailed
+type RecordActivityFailedParams struct {
+	Failure              *failurepb.Failure
+	LastHeartbeatDetails *commonpb.Payloads
+	WorkerIdentity       string
+}
+
+// HandleActivityFailed updates the activity on activity failure. if the activity is retriable, it will be rescheduled for retry instead.
```
```suggestion
// HandleActivityFailed updates the activity on activity failure. if the activity is retryable, it will be rescheduled for retry instead.
```

chasm/lib/activity/activity.go
```diff
@@ -225,9 +227,75 @@ func (a *Activity) RecordCompletion(ctx chasm.MutableContext, applyFn func(ctx c
...
+		return false, err
+	}
+
+	return enoughAttempts && enoughTime, nil
+}
+
+// recordTimeoutFromScheduledStatus records schedule-to-start or schedule-to-close timeouts. Such timeouts are not retried so we
 // set the outcome failure directly and leave the attempt failure as is.
-func (a *Activity) recordFromScheduledTimeOut(ctx chasm.MutableContext, timeoutType enumspb.TimeoutType) error {
+func (a *Activity) recordTimeoutFromScheduledStatus(ctx chasm.MutableContext, timeoutType enumspb.TimeoutType) error {
```
dandavison: "FromScheduled" isn't an established concept, but the names "schedule-to-start" and "schedule-to-close" are very established. Suggest `recordScheduleToStartOrCloseTimeoutFailure` for consistency with `createStartToCloseTimeoutFailure`.

chasm/lib/activity/statemachine.go
```diff
@@ -153,8 +158,45 @@ var TransitionCompleted = chasm.NewTransition(
...
-	func(_ *Activity, _ chasm.MutableContext, _ any) error {
-		return nil
+	func(a *Activity, ctx chasm.MutableContext, params RecordActivityCompletedParams) error {
+		store, err := a.Store.Get(ctx)
+		if err != nil {
+			return err
+		}
+
+		if store == nil {
+			err = a.RecordCompletion(ctx, func(ctx chasm.MutableContext) error {
```
dandavison: Why do method names use "Completion" in some cases and "Completed" in others?

chasm/lib/activity/activity.go
```diff
@@ -225,9 +227,75 @@ func (a *Activity) RecordCompletion(ctx chasm.MutableContext, applyFn func(ctx c
 	return applyFn(ctx)
```
dandavison, line 218: What is the point of `RecordCompletion`? Why not just use the `applyFn` directly at the call sites?

chasm/lib/activity/activity.go
```diff
@@ -251,25 +319,20 @@ func (a *Activity) recordFromScheduledTimeOut(ctx chasm.MutableContext, timeoutT
 	return nil
 }
 
-// recordStartToCloseTimedOut records start-to-close timeouts. These come from retried attempts so we update the attempt
-// failure info but leave the outcome failure empty to avoid duplication
-func (a *Activity) recordStartToCloseTimedOut(ctx chasm.MutableContext, retryInterval time.Duration, noRetriesLeft bool) error {
+// recordFailedAttempt records any failures resulting from a tried attempt, including worker application failures and
+// start-to-close timeouts. Since the calls come from retried attempts so we update the attempt failure info but leave
+// the outcome failure empty to avoid duplication
```
```suggestion
// start-to-close timeouts. Since the calls come from retried attempts we update the attempt failure info but leave
// the outcome failure empty to avoid duplication.
```

chasm/lib/activity/activity.go
fretz12: We can just drop the "task" and "activity" in the name. Renamed.

chasm/lib/activity/activity.go
fretz12: This will involve more for workflow activities, i.e. writing to history. As discussed with Roey, we want to put this interface for now. It might evolve as we start implementing workflow activities.

chasm/lib/activity/activity.go
fretz12: Fixed

chasm/lib/activity/activity.go
fretz12: Fixed

chasm/lib/activity/activity.go
fretz12: Renamed.

chasm/lib/activity/activity.go
fretz12: This pattern was suggested by @bergundy when we did the start API as a way to avoid using closures from the caller and to transactionally return the filled in response as opposed to marshaling to an intermediate struct. I do understand your point of model leakage. We can bring this up for discussion next time we meet.

chasm/lib/activity/statemachine.go
fretz12: I like completed. Changed.

chasm/lib/activity/activity.go
dandavison: OK. What's the difference between `RecordFoo` and `HandleFoo`? Currently we have
```
// RecordCompletedParams holds parameters for HandleCompleted
```
and
```
// RecordFailedParams holds parameters for HandleFailed
```

bergundy: JFYI this interface won't work for workflow activities as I mentioned in the last PR but I don't want to dive into this now. It's easy to refactor later.

chasm/lib/activity/activity.go
bergundy: I would be fine with taking historyservice request as params too FTR. I just probably would keep the state machines a bit cleaner.

Not super opinionated here. Originally I just saw that we were translating and converting without good enough justification.

chasm/lib/activity/activity.go
```diff
@@ -308,6 +372,17 @@ func (a *Activity) hasEnoughTimeForRetry(ctx chasm.Context) (bool, error) {
 	return ctx.Now(a).Add(retryInterval).Before(deadline), nil
 }
 
+func (a *Activity) createStartToCloseTimeoutFailure() *failurepb.Failure {
```
bergundy: nit: this doesn't need to be a method if it doesn't reference the activity.

chasm/lib/activity/activity.go
```diff
@@ -221,13 +223,80 @@ func (a *Activity) PopulateRecordActivityTaskStartedResponse(ctx chasm.Context,
...
+	}
+	retryPolicy := a.RetryPolicy
+
+	enoughAttempts := retryPolicy.GetMaximumAttempts() == 0 || attempt.GetCount() < retryPolicy.GetMaximumAttempts()
+	enoughTime, err := a.hasEnoughTimeForRetry(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	return enoughAttempts && enoughTime, nil
```
bergundy: This can be extracted into a method, this logic is duplicated in one of the executors.

chasm/lib/activity/activity.go
bergundy: Eventually, these methods should be defined as CHASM activity APIs. But it'll take a while for us to support workflow activities in CHASM.

chasm/lib/activity/statemachine.go
```diff
@@ -90,10 +93,12 @@ var TransitionRescheduled = chasm.NewTransition(
 		currentTime := ctx.Now(a)
 		attempt.Count += 1
 
-		// If this is a retry, calculate the delay before scheduling tasks and update attempt fields
-		// TODO: for activity failures it'll go through this retry path as well; we'll need to refactor the record timeout and retryInterval recording, probably passed as an event func
-		retryInterval := backoff.CalculateExponentialRetryInterval(a.GetRetryPolicy(), attempt.GetCount())
-		err = a.recordStartToCloseTimedOut(ctx, retryInterval, false)
+		retryInterval := failure.GetApplicationFailureInfo().GetNextRetryDelay().AsDuration()
+		if retryInterval <= 0 {
+			retryInterval = backoff.CalculateExponentialRetryInterval(a.GetRetryPolicy(), attempt.GetCount())
```
bergundy: This interval should be taken into account when calculating `hasEnoughTimeForRetry`.

chasm/lib/activity/statemachine.go
```diff
@@ -153,8 +158,45 @@ var TransitionCompleted = chasm.NewTransition(
...
+				}
+
+				return nil
+			})
+		} else {
+			err = store.RecordCompleted(ctx, func(ctx chasm.MutableContext) error {
+				// Implement workflow activity completion handling here, including logic to rebuild the workflow state from history if needed.
+				return status.Errorf(codes.Unimplemented, "workflow activity completed handling is not implemented")
+			})
+		}
```
bergundy, line 169:
```suggestion
		if store == nil {
			store = a
		}
		return store.RecordCompleted(ctx, func(ctx chasm.MutableContext) error {
			attempt, err := a.Attempt.Get(ctx)
			if err != nil {
				return err
			}

			attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(a))
			if params.WorkerIdentity != "" {
				attempt.LastWorkerIdentity = params.WorkerIdentity
			}

			outcome, err := a.Outcome.Get(ctx)
			if err != nil {
				return err
			}

			outcome.Variant = &activitypb.ActivityOutcome_Successful_{
				Successful: &activitypb.ActivityOutcome_Successful{
					Output: params.Payload,
				},
			}

			return nil
		})
```

chasm/lib/activity/statemachine.go
```diff
@@ -164,8 +206,41 @@ var TransitionFailed = chasm.NewTransition(
...
 	activitypb.ACTIVITY_EXECUTION_STATUS_FAILED,
-	func(_ *Activity, _ chasm.MutableContext, _ any) error {
-		return nil
+	func(a *Activity, ctx chasm.MutableContext, params RecordFailedParams) error {
+		store, err := a.Store.Get(ctx)
+		if err != nil {
+			return err
+		}
+
+		if store == nil {
```
bergundy, line 208: Same here, the idea is that if the store is nil, you default to "self".

chasm/lib/activity/statemachine.go
```diff
@@ -153,8 +158,45 @@ var TransitionCompleted = chasm.NewTransition(
...
+			err = a.RecordCompleted(ctx, func(ctx chasm.MutableContext) error {
+				attempt, err := a.Attempt.Get(ctx)
+				if err != nil {
+					return err
+				}
+
+				attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(a))
+				if params.WorkerIdentity != "" {
```
bergundy: When would this be empty? Can you document this please?

common/failure/failure.go
```diff
@@ -89,3 +92,43 @@ func TruncateWithDepth(f *failurepb.Failure, maxSize, maxDepth int) *failurepb.F
 
 	return newFailure
 }
+
+// IsRetryable determines if a failure is retryable based on its type and non-retryable types list.
+func IsRetryable(failure *failurepb.Failure, nonRetryableTypes []string) bool {
```
bergundy: FWIW the failure you get from the API must always be an ApplicationFailure instance.

common/failure/failure.go
```diff
@@ -89,3 +92,43 @@ func TruncateWithDepth(f *failurepb.Failure, maxSize, maxDepth int) *failurepb.F
...
+	if failure == nil {
```
bergundy: Why is the default true? This is undefined behavior AFAIC.

chasm/lib/activity/activity.go
fretz12: @dandavison - Record was is more or less updating the model only. Handle does more than just updating, including retries for example. That said, I like consistency, and feel Handle is the better all-encompassing term

chasm/lib/activity/activity.go
fretz12: Sounds like we're leaning towards using historyservice request and response to make it consistent. I'm ok with that.

common/failure/failure.go
fretz12: This func is refactored as is from history service for reuse. Seems like that's overkill and I will just grab the part for ApplicationFailure an reuse locally.

chasm/lib/activity/statemachine.go
fretz12: This originates from the worker called RespondActivityTaskCompletedRequest and the Identity field is optional, so it could be blank at this point. I will doc it.

common/failure/failure.go
fretz12: From refactored func as is, but irrelevant now since I'm going to unrefactor this as stated above.

chasm/lib/activity/activity.go
fretz12: The RecordCompleted interface method is kept as is, but for other occurrences of public methods with the prefix `Record` has been changed the `Handle`

chasm/lib/activity/activity.go
fretz12: Changed.

chasm/lib/activity/activity.go
fretz12: Made into method.

chasm/lib/activity/activity.go
fretz12: Addressed.

chasm/lib/activity/statemachine.go
fretz12: The callers of TransitionRescheduled already do the check if it should retry, but I did miss taking into account NextRetryDelay (if set) for the retry. I've refactored this a bit so that all calcs including retryInterval are done once and passed into the transition func.

chasm/lib/activity/statemachine.go
fretz12: done

chasm/lib/activity/statemachine.go
fretz12: done

chasm/lib/activity/activity.go
```diff
@@ -53,12 +53,6 @@ type Activity struct {
 	Store chasm.Field[ActivityStore]
 }
 
-// RecordActivityTaskStartedParams holds parameters for RecordActivityTaskStarted
-type RecordActivityTaskStartedParams struct {
-	VersionDirective *taskqueuespb.TaskVersionDirective
-	WorkerIdentity   string
-}
-
 // LifecycleState TODO: we need to add more lifecycle states to better categorize some activity states, particulary for terminated/canceled.
```
bergundy: You can remove this TODO

chasm/lib/activity/activity.go
```diff
@@ -104,8 +98,9 @@ func NewStandaloneActivity(
 			Header:       request.Header,
 			UserMetadata: request.UserMetadata,
 		}),
-		Outcome:    chasm.NewDataField(ctx, &activitypb.ActivityOutcome{}),
-		Visibility: chasm.NewComponentField(ctx, visibility),
+		LastHeartbeat: chasm.NewDataField(ctx, &activitypb.ActivityHeartbeatState{}),
```
bergundy: IIUC, if you don't create this component by default, it will reduce load on the DB by not creating the cell for this node. @yycptt correct me if I'm wrong.

chasm/lib/activity/statemachine.go
```diff
@@ -75,13 +76,19 @@ var TransitionScheduled = chasm.NewTransition(
 	},
 )
 
-// TransitionRescheduled affects a transition to Scheduled from Started, which happens on retries
+type rescheduleInfo struct {
```
bergundy: nit: you could consider calling this `rescheduleInput` or `rescheduleEvent` or `transitionRescheduledInput`.

`info` is a very generic term that is overused in our codebase IMHO.

chasm/lib/activity/statemachine.go
```diff
@@ -153,8 +157,43 @@ var TransitionCompleted = chasm.NewTransition(
...
+			if err != nil {
+				return err
+			}
+
+			attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(a))
+
+			// The Identity field is optionally set by the worker, so it could be blank at this point. Set identity
+			// only if has been provided to avoid overwriting previous value with blank.
+			if identity := request.GetCompleteRequest().GetIdentity(); identity != "" {
+				attempt.LastWorkerIdentity = identity
```
bergundy: My concern here is that we will leave the identity of the previous attempt, which would be incorrect. I think we always should set this when an attempt concludes.

chasm/lib/activity/statemachine.go
```diff
@@ -153,8 +157,43 @@ var TransitionCompleted = chasm.NewTransition(
...
+			store = a
+		}
+
+		return store.RecordCompleted(ctx, func(ctx chasm.MutableContext) error {
+			attempt, err := a.Attempt.Get(ctx)
+			if err != nil {
+				return err
+			}
+
+			attempt.LastAttemptCompleteTime = timestamppb.New(ctx.Now(a))
```
bergundy: Not blocking this PR but we should call this field `LastCompleteTime` because it's already on an `Attempt` message.

chasm/lib/activity/statemachine.go
```diff
@@ -164,8 +203,40 @@ var TransitionFailed = chasm.NewTransition(
...
+					return err
+				}
+
+				heartbeatDetails.Details = details
+				heartbeatDetails.RecordedTime = timestamppb.New(ctx.Now(a))
+			}
+
+			// The Identity field is optionally set by the worker, so it could be blank at this point. Set identity
+			// only if has been provided to avoid overwriting previous value with blank.
+			if identity := req.GetFailedRequest().GetIdentity(); identity != "" {
```
bergundy: Same here, the information on the `Attempt` struct must always be relevant to the last attempt and never the previous one.

chasm/lib/activity/statemachine_test.go
```diff
@@ -333,3 +340,85 @@ func TestTransitionTimedout(t *testing.T) {
...
+			Result:   payload,
+			Identity: "worker",
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, activitypb.ACTIVITY_EXECUTION_STATUS_COMPLETED, activity.Status)
+	require.EqualValues(t, 1, attemptState.Count)
+	require.Equal(t, "worker", attemptState.GetLastWorkerIdentity())
+	require.NotNil(t, attemptState.GetLastAttemptCompleteTime())
+	require.True(t, proto.Equal(payload, outcome.GetSuccessful().GetOutput()))
```
bergundy: Use `protorequire.ProtoEqual`

service/frontend/workflow_handler.go
```diff
@@ -1445,6 +1444,21 @@ func (wh *WorkflowHandler) RespondActivityTaskCompletedById(ctx context.Context,
 		return nil, errIdentityTooLong
 	}
 
+	// If workflowID is empty, it means the activity is a standalone activity and we need to set the component ref
+	var componentRef []byte
+	if workflowID == "" {
```
bergundy, line 1453: This should be possible only if standalone activities are enabled.
We still need to add that dynamic config, add a TODO here.

service/frontend/workflow_handler.go
```diff
@@ -1445,6 +1444,21 @@ func (wh *WorkflowHandler) RespondActivityTaskCompletedById(ctx context.Context,
...
+		componentRef, err = (&persistencespb.ChasmComponentRef{
+			NamespaceId: namespaceID.String(),
+			BusinessId:  activityID,
+			EntityId:    runID,
+			Archetype:   "activity.activity",
```
bergundy: Consider making this token creation a method in `chasm/lib/activity` to keep the logic contained.

service/frontend/workflow_handler.go
```diff
@@ -1445,6 +1444,21 @@ func (wh *WorkflowHandler) RespondActivityTaskCompletedById(ctx context.Context,
...
+		}).Marshal()
+
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
bergundy: Ah even copilot is requesting making a method for this.

service/history/handler.go
```diff
@@ -148,6 +148,7 @@ var (
 	errWorkflowExecutionNotSet = serviceerror.NewInvalidArgument("WorkflowExecution not set on request.")
 	errTaskQueueNotSet         = serviceerror.NewInvalidArgument("Task queue not set.")
 	errWorkflowIDNotSet        = serviceerror.NewInvalidArgument("WorkflowId is not set on request.")
+	errBusinessIDNotSet        = serviceerror.NewInvalidArgument("Business ID is not set on request.")
```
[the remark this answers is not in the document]
bergundy: This should be an internal error IMHO, if it happens it's because the frontend has a bug. But let's not change the semantics here without a broader discussion.

tests/standalone_activity_test.go
```diff
@@ -11,20 +11,60 @@ import (
...
+	"go.temporal.io/server/common/testing/testvars"
 	"go.temporal.io/server/tests/testcore"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
+// TODO(fred) add tests for retries when we implement search attributes
+
+var (
+	defaultInput = &commonpb.Payloads{
```
bergundy: Why not use `payloads.EncodeString` here too?

tests/standalone_activity_test.go
```diff
@@ -11,20 +11,60 @@ import (
...
+
+	defaultFailure = &failurepb.Failure{
+		Message: "Failed Activity",
+		FailureInfo: &failurepb.Failure_ApplicationFailureInfo{ApplicationFailureInfo: &failurepb.ApplicationFailureInfo{
+			Type:         "Test",
+			NonRetryable: true,
+		}},
+	}
+
+	defaultHeartbeatDetails = &commonpb.Payloads{
```
bergundy: Same here, use `EncodeString`

tests/standalone_activity_test.go
```diff
@@ -35,6 +75,7 @@ func TestStandaloneActivityTestSuite(t *testing.T) {
 
 func (s *standaloneActivityTestSuite) SetupSuite() {
 	s.FunctionalTestBase.SetupSuite()
+	s.tv = testvars.New(s.T())
```
bergundy: Shouldn't this be set in `SetupTest()`?

tests/standalone_activity_test.go
```diff
@@ -45,49 +86,124 @@ func (s *standaloneActivityTestSuite) SetupSuite() {
...
+
+	_, err := s.FrontendClient().RespondActivityTaskCompleted(ctx, &workflowservice.RespondActivityTaskCompletedRequest{
+		Namespace: s.Namespace().String(),
+		TaskToken: pollTaskResp.TaskToken,
+		Result:    defaultResult,
+		Identity:  "new-worker",
+	})
+	require.NoError(t, err)
+
+	s.validateCompletion(ctx, t, activityID, runID, "new-worker")
```
bergundy, line 85: I am not a fan of these test helpers that hide a lot of the logic and make it hard to follow exactly what is happening. Personally, I am on team repeat yourself in tests.
I developed this notion independently and finally took a stronger stance after watching https://www.youtube.com/watch?v=8hQG7QlcLBk.

That being said, I don't think everyone shares this stance, and I don't think it would be fair to block the PR because tests aren't written in this specific style.

tests/standalone_activity_test.go
```diff
@@ -263,8 +369,111 @@ func (s *standaloneActivityTestSuite) Test_PollActivityExecution_WaitAnyStateCha
...
+		// Worker picks up activity task and completes it
+		go func() {
+			pollTaskResp, err := s.pollActivityTaskQueue(ctx, taskQueue)
+			if err != nil {
+				taskError = err
+				return
+			}
+
+			taskError = tc.taskCompletionFn(ctx, pollTaskResp.TaskToken)
+		}()
```
bergundy: Why spin up these goroutines? It seems like you can write the test to first poll, then complete, then poll on completion.

chasm/lib/activity/activity.go
fretz12: Removed

chasm/lib/activity/activity.go
fretz12: ah nice optimization. I've made a helper method that will get and create it lazily.
cc: @dandavison

chasm/lib/activity/statemachine.go
fretz12: event seems suitable given we're using that term in the chasm SM

chasm/lib/activity/statemachine.go
fretz12: Renamed

chasm/lib/activity/statemachine.go
fretz12: I'm ok with that. Either choice will be tricky, but at least this will put the onus on the user to set the identity.

chasm/lib/activity/statemachine.go

chasm/lib/activity/statemachine_test.go
fretz12: changed all occurrences

service/frontend/workflow_handler.go
fretz12: I see. Added TODO.

service/frontend/workflow_handler.go
fretz12: done

tests/standalone_activity_test.go
fretz12: sure why not

tests/standalone_activity_test.go
fretz12: Done

tests/standalone_activity_test.go
fretz12: Moved

tests/standalone_activity_test.go
fretz12: I see the point and thanks for sharing the vid. I think there's a balance between the two.
"Limit helpers to very reused logic that doesn't fail often or fails all at once" - I like that philosophy.
Let me run by the team in the next standup than we can lock it down.

tests/standalone_activity_test.go
fretz12: Seems cleaner if we do that. However, there's value in testing the poll await completion if no completion ever happens. I will add a test for that too.

tests/standalone_activity_test.go
```diff
@@ -263,8 +369,111 @@ func (s *standaloneActivityTestSuite) Test_PollActivityExecution_WaitAnyStateCha
...
+			)
+
+			require.True(t, proto.Equal(defaultInput, activityPollResp.GetInput()))
+			tc.completionValidationFn(t, activityPollResp)
+		case <-ctx.Done():
+			t.Fatal("PollActivityExecution timed out")
+		}
+
+		require.NoError(t, taskError)
+	}
```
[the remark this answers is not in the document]
fretz12: fixed

chasm/lib/activity/proto/v1/activity_state.proto
```diff
@@ -123,7 +123,7 @@ message ActivityAttemptState {
     google.protobuf.Timestamp last_started_time = 3;
 
     // The time when the last activity attempt completed. If activity has not been completed yet, it will be null.
-    google.protobuf.Timestamp last_attempt_complete_time = 4;
+    google.protobuf.Timestamp last_complete_time = 4;
```
bergundy: Just `started_time` and `completed_time` IMHO because this is information of a single attempt. And we would want to be consistent with the tense.

chasm/lib/activity/activity.go
```diff
@@ -68,6 +68,16 @@ func (a *Activity) LifecycleState(_ chasm.Context) chasm.LifecycleState {
 	}
 }
 
+// ToRef creates a serialized component reference for the given activity identifiers.
+func ToRef(namespaceID string, activityID string, runID string) ([]byte, error) {
```
bergundy: You should be able to use `chasm.NewComponentRef(...).Serialize()` instead of defining this method.
Just wondering if the chasm registry will be aware of the activity component outside of history service.  We should probably make sure we register all of the libraries everywhere regardless.

service/history/workflow/transaction_impl.go
```diff
@@ -190,6 +191,11 @@ func (t *TransactionImpl) UpdateWorkflowExecution(
 	if persistence.OperationPossiblySucceeded(err) {
 		NotifyWorkflowMutationTasks(engine, currentWorkflowMutation)
 		NotifyWorkflowSnapshotTasks(engine, newWorkflowSnapshot)
+		engine.NotifyChasmExecution(chasm.EntityKey{
```
bergundy: Wondering why you are checking in @dandavison's changes in your PR...

chasm/lib/activity/proto/v1/activity_state.proto
fretz12: Renamed

service/history/workflow/transaction_impl.go
fretz12: I've merged his changes, should be good now

chasm/lib/activity/activity.go
fretz12: Changed to use Serialize and added injected deps.

chasm/lib/activity/activity.go
```diff
@@ -25,11 +25,11 @@ import (
 )
 
 type ActivityStore interface {
-	// PopulateRecordActivityTaskStartedResponse populates the response for RecordActivityTaskStarted
-	PopulateRecordActivityTaskStartedResponse(ctx chasm.Context, key chasm.EntityKey, response *historyservice.RecordActivityTaskStartedResponse) error
```
bergundy, line 28: Not blocking but you will want `Activity` in both of these method names because when the workflow component implements these APIs it will be important to qualify.
Don't worry about this though, we are going to want to rewrite all of this eventually anyways.

chasm/lib/activity/handler.go
```diff
@@ -112,17 +112,22 @@ func (h *handler) PollActivityExecution(
 			ctx chasm.Context,
 			req *activitypb.PollActivityExecutionRequest,
 		) (*activitypb.PollActivityExecutionResponse, bool, error) {
-			// TODO(dan): check for terminal activity states
-			panic("pollActivityExecutionWaitCompletion is not implemented")
-			completed := false
-			if completed {
+
+			switch a.Status {
```
bergundy: You could also check if `a.LifecycleState(ctx)` is not running and save this duplicated logic.

chasm/lib/activity/statemachine.go
```diff
@@ -75,13 +76,19 @@ var TransitionScheduled = chasm.NewTransition(
...
+// TransitionRescheduled affects a transition to Scheduled from Started, which happens on retries. The event to pass in
+// is the failure to be recorded from the previously failed attempt.
 var TransitionRescheduled = chasm.NewTransition(
 	[]activitypb.ActivityExecutionStatus{
-		activitypb.ACTIVITY_EXECUTION_STATUS_STARTED, // On retries for start-to-close timeouts
+		activitypb.ACTIVITY_EXECUTION_STATUS_STARTED, // For retries the activity will be in started status
 	},
 	activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED,
-	func(a *Activity, ctx chasm.MutableContext, _ any) error {
+	func(a *Activity, ctx chasm.MutableContext, info rescheduleEvent) error {
```
bergundy:
```suggestion
	func(a *Activity, ctx chasm.MutableContext, event rescheduleEvent) error {
```

chasm/lib/activity/statemachine.go
```diff
@@ -153,8 +157,38 @@ var TransitionCompleted = chasm.NewTransition(
...
-		return nil
+	func(a *Activity, ctx chasm.MutableContext, request *historyservice.RespondActivityTaskCompletedRequest) error {
+		store, err := a.Store.Get(ctx)
+		if err != nil {
+			return err
+		}
+
+		if store == nil {
+			store = a
+		}
```
bergundy, line 169: When you rebase you won't have to handle errors anymore and I would recommend adding the following helper:

```suggestion
		store := a.LoadStore(ctx)
```

chasm/lib/activity/statemachine_test.go
```diff
@@ -334,3 +337,85 @@ func TestTransitionTimedout(t *testing.T) {
...
+
+	activity := &Activity{
+		ActivityState: &activitypb.ActivityState{
+			RetryPolicy:            defaultRetryPolicy,
+			ScheduleToCloseTimeout: durationpb.New(defaultScheduleToCloseTimeout),
+			ScheduleToStartTimeout: durationpb.New(defaultScheduleToStartTimeout),
+			StartToCloseTimeout:    durationpb.New(defaultStartToCloseTimeout),
+			Status:                 activitypb.ACTIVITY_EXECUTION_STATUS_STARTED,
+		},
+		Attempt: chasm.NewDataField(ctx, attemptState),
```
bergundy: This field should be called `LastAttempt` IMHO.

chasm/transition_history.go
bergundy: This doesn't seem like the right error. If a user provides an invalid ref, we should return an InvalidArgument error not internal.

AGENTS.md
```diff
@@ -99,3 +99,83 @@ When planning (under 'Software Engineering Tasks'):
   c. Complexity trade-offs
   d. Security trade-offs
 7. Reason about the failure modes of your design. How does it handle crashes? A 10x increase in load?
+The following instructions take precedence over any previous ones derived from different agent instruction files.
```
bergundy: This doesn't belong here... looks like it was taken from an SDK repo?

Ideally if you want to make this type of change it isn't bundled with other features.

fretz12: Seems like a GH bug. If you compare the files, they are identical

chasm/lib/activity/activity.go
fretz12: Noted.

chasm/lib/activity/handler.go
fretz12: Love it, changed.

chasm/lib/activity/statemachine.go
fretz12: refactored.

chasm/lib/activity/statemachine.go
fretz12: Excellent. Added TODO for now.

chasm/lib/activity/statemachine_test.go
fretz12: I feel it is implicitly understood in the way it's used here as there's only a single attemptstate here and attemptstate is not used in any other way elsewhere. So, unless you feel strongly about it I'd prefer to keep it simple as Attempt.

bergundy: I think last is clearer because all of the fields in attempt state are not supposed to be specific to the last attempt, they are just fields of an attempt.

common/persistence/visibility/store/elasticsearch/visibility_store_read_test.go
bergundy, line 1: You shouldn't need to touch these files in this PR.

service/history/handler.go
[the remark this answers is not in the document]
bergundy: Just make sure this is tracked please @fretz12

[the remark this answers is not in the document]
fretz12: ok, let me revert this for now then. Really shouldn't be happening and at worst it's just a slightly confusing message

common/persistence/visibility/store/elasticsearch/visibility_store_read_test.go
fretz12: These are from running lint-code. I think it's been cleaned up in main already. Let me revert them here to avoid confusion.

chasm/lib/activity/statemachine_test.go
fretz12: Ok, renamed.

PR #8658, opened 2025-11-17
Enable replication separately from namespace replication

michaely520:
## What changed?
- Add a new flag that controls whether clusters setup replication streams to each other
- Feature flag to gate the new flag for compatibility
- Functional test to vet the changes

## Why?
Optimization to avoid excessive network activity when we only want namespace replication.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/xdc/replication_enable_test.go
+++ b/tests/xdc/replication_enable_test.go
@@ -223,7 +223,7 @@ func (s *ReplicationEnableTestSuite) TestReplicationEnableFlow() {
 	s.Require().NoError(err)
 	s.Require().Equal("workflow completed successfully", result)
 
-	// Verify workflow is in completed state
+	// Verify workflow is in completed state on active cluster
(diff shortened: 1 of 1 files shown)


proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -284,6 +284,8 @@ message AddOrUpdateRemoteClusterRequest {
   string frontend_address = 1;
   bool enable_remote_cluster_connection = 2;
   string frontend_http_address = 3 [deprecated = true];
+  // is_replication_enabled controls whether replication streams are active.
+  bool is_replication_enabled = 4;
```
xwduan, line 288: Nit: maybe consistent with the name above i.e. `enable_remote_cluster_connection`

xwduan (approved):

PR #8605, opened 2025-11-06
Single workflow pause API

gow:
## What changed?
Note: This API is part of an experimental feature - "pause single workflow"

- Added a new API (in Frontend & History) to pause workflow execution.
- Updated DescribeWorkflowExecution API to show the pause info.
- Added a dynamic config to control the access to this feature. The PauseWorkflowExecution API fails if the config is not set.
- Added simple acceptance functional tests. These tests will evolve and new ones added as the feature implementation progresses.

Followup PRs:
- Idempotency check in pause api.
- Validate input sizes.
- Intercept workflow task creation.
- Intercept and pause activities.
- Unpause API

## Why?
This API is needed to implement the "pause single workflow" feature.

## How did you test it?
- [x] built
- [x] run locally and tested manually
- [ ] covered by existing tests
- [ ] added new unit test(s)
- [x] added new functional test(s)

---

> [!NOTE]
> Adds a PauseWorkflowExecution API across frontend/history, wires clients/GRPC/mocks, surfaces pause info in DescribeWorkflowExecution, gates via new WorkflowPauseEnabled config, and adds functional tests.
> 
> - **APIs/Protos**:
>   - Add `PauseWorkflowExecutionRequest/Response` messages in `api/historyservice/v1/request_response.proto` and generated Go files.
>   - Extend `HistoryService` with `PauseWorkflowExecution` RPC (server, client, grpc, mocks).
> - **Frontend**:
>   - Implement `WorkflowHandler.PauseWorkflowExecution`, resolving namespace, invoking history, and gating by `WorkflowPauseEnabled`.
>   - Wire new config in `service.go`.
> - **History**:
>   - Implement handler `PauseWorkflowExecution` in engine and new API `service/history/api/pauseworkflow` to add paused event and handle running-state checks.
>   - Enhance `DescribeWorkflowExecution` to include `PauseInfo` in `WorkflowExtendedInfo`.
> - **Clients**:
>   - Add client methods (`client_gen`, `metric_client_gen`, `retryable_client_gen`) and historyservice mocks.
> - **Config**:
>   - Introduce `dynamicconfig.WorkflowPauseEnabled` (namespace-scoped) and thread through frontend config.
> - **Telemetry/Logging**:
>   - Add log tags for `PauseWorkflowExecution` and include RPC in telemetry interceptor.
> - **Tests**:
>   - Add functional tests `tests/pause_workflow_execution_test.go` covering enabled/disabled scenarios.
> - **Dependencies**:
[...]

PR #8660, opened 2025-11-18
[WorkflowPause] Input validation and idempotency check in pause API

gow:
## What changed?
 - Pause API validates lengths of identity, reason & request ID before processing the request.
 - Idempotency check is made against the saved request ID in mutable state.
 - Added functional tests.

Depends on https://github.com/temporalio/temporal/pull/8605

## Why?
Make the API guard against bad inputs.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

---

> [!NOTE]
> Adds input length validation and idempotent handling to PauseWorkflowExecution, with new functional tests.
> 
> - **Backend**:
>   - **Pause API** (`service/history/api/pauseworkflow/api.go`):
>     - Validate `PauseWorkflowExecutionRequest` fields (`identity`, `reason`, `requestId`) against `MaxIDLengthLimit` via `validatePauseRequest`.
>     - Idempotency: if workflow is already paused and `PauseInfo.RequestId` matches incoming `requestId`, return no-op; otherwise return `FailedPrecondition`.
>     - Minor import addition for `workflowservice/v1`.
> - **Tests** (`tests/pause_workflow_execution_test.go`):
>   - Add `TestPauseWorkflowExecutionRequestValidation` covering too-long `identity`, `reason`, `requestId`, and disabled dynamic config (expects `Unimplemented`).
>   - Add `TestPauseWorkflowExecutionAlreadyPaused` verifying first pause succeeds and second with different `requestId` fails with `FailedPrecondition`.
>   - Keep baseline `TestPauseWorkflowExecution` asserting pause status and stored `PauseInfo`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ffbc41bc3b39d07aa737c22ecf7ac8f0637bd538. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/pause_workflow_execution_test.go
+++ b/tests/pause_workflow_execution_test.go
@@ -235,7 +235,7 @@ func (s *PauseWorkflowExecutionSuite) TestPauseWorkflowExecutionAlreadyPaused() 
 		info := desc.GetWorkflowExecutionInfo()
 		require.NotNil(t, info)
 		require.Equal(t, enumspb.WORKFLOW_EXECUTION_STATUS_PAUSED, info.GetStatus())
-		if pauseInfo := info.GetPauseInfo(); pauseInfo != nil {
+		if pauseInfo := desc.GetWorkflowExtendedInfo().GetPauseInfo(); pauseInfo != nil {
 			require.Equal(t, s.pauseIdentity, pauseInfo.GetIdentity())
 			require.Equal(t, s.pauseReason, pauseInfo.GetReason())
 		}


service/history/api/pauseworkflow/api.go
```diff
@@ -89,3 +95,22 @@ func Invoke(
...
+	// verify size limits of reason, request id and identity.
+	if len(req.GetReason()) > shard.GetConfig().MaxIDLengthLimit() {
+		return serviceerror.NewInvalidArgument("reason is too long.")
+	}
+	if len(req.GetRequestId()) > shard.GetConfig().MaxIDLengthLimit() {
+		return serviceerror.NewInvalidArgument("request id is too long.")
+	}
+	if len(req.GetIdentity()) > shard.GetConfig().MaxIDLengthLimit() {
+		return serviceerror.NewInvalidArgument("identity is too long.")
+	}
```
spkane31, line 109: Does the same config limit the reason, identity, and request id?

tests/pause_workflow_execution_test.go
```diff
@@ -112,17 +113,93 @@ func (s *PauseWorkflowExecutionSuite) TestPauseWorkflowExecution() {
...
+		RunId:      uuid.New(),
+		Identity:   s.pauseIdentity,
+		Reason:     s.pauseReason,
+		RequestId:  uuid.New(),
+	}
+	resp, err = s.FrontendClient().PauseWorkflowExecution(ctx, pauseRequest)
+	s.Error(err)
+	s.Nil(resp)
+	var unimplementedErr *serviceerror.Unimplemented
+	s.ErrorAs(err, &unimplementedErr)
```
spkane31, line 190: Is `Unimplemented` always the error we return for a feature that exists but isn't turned on for a namespace? That seems unintuitive but a reasonable response.

service/history/api/pauseworkflow/api.go
gow: Yes. Based on the [comment on this config](https://github.com/temporalio/temporal/blob/main/common/dynamicconfig/constants.go#L415-L416) - `MaxIDLengthLimit is the length limit for various IDs, including: Namespace, TaskQueue, WorkflowID, ActivityID, TimerID,
WorkflowType, ActivityType, SignalName, MarkerName, ErrorReason/FailureReason/CancelCause, Identity, RequestID`

tests/pause_workflow_execution_test.go
gow: Good point. There is actually `Unavailable` error as well. That seems like the best error to return in this case.

service/history/api/pauseworkflow/api.go
```diff
@@ -89,3 +95,22 @@ func Invoke(
 	// Return response
 	return &historyservice.PauseWorkflowExecutionResponse{}, nil
 }
+
+func validatePauseRequest(req *workflowservice.PauseWorkflowExecutionRequest, shard historyi.ShardContext, namespaceEntry *namespace.Namespace) error {
+	if !shard.GetConfig().WorkflowPauseEnabled(namespaceEntry.ID().String()) {
+		return serviceerror.NewUnavailablef("workflow pause is not enabled for namespace: %s", namespaceEntry.Name().String())
```
yycptt: Need to return a non retryable error BadRequest/FailedPrecondition/Unimplemented and make sure it doesn't count against our SLA I think. 

Also I think this check should be done on frontend. Same Q for other checks below.

service/history/api/pauseworkflow/api.go
```diff
@@ -56,11 +57,16 @@ func Invoke(
...
+				pauseInfo := mutableState.GetExecutionInfo().PauseInfo
+				if pauseInfo != nil && pauseInfo.RequestId != "" && pauseInfo.RequestId == pauseRequest.GetRequestId() {
+					// Already paused with the same request id, nothing to do
+					return &api.UpdateWorkflowAction{
+						Noop:               true,
+						CreateWorkflowTask: false,
+					}, nil
+				}
+				releaseFn(nil)
+				return nil, serviceerror.NewFailedPrecondition("workflow is already paused.")
```
yycptt, line 69: could also just return a success response I guess (similar to workflow cancel). not sure what's the semantic we want here. 

I will defer that to your team, just want to make sure that's a conscious decision.

gow: Yes. It was discussed in the internal blueprint doc. We want to explicitly fail any (non-duplicate) pause requests while the workflow is paused. The idea is to provide a different API to configure pause behavior (ex: include timers, child workflows etc). When such API exists, this error message would point users to use that API instead.

service/history/api/pauseworkflow/api.go
gow: Done. Moved it to frontend (in the [base PR](https://github.com/temporalio/temporal/pull/8605/files#diff-c9a59445e489bd70346630f2773ad69e1723b2298fadbae50ff5c1a424917e1c)) and returning `Unimplemented` now.

service/history/api/pauseworkflow/api.go
```diff
@@ -52,11 +57,16 @@ func Invoke(
...
```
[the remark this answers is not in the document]
gow: `RequestId` is optional. So it could be missing (either in request or in the stored pauseInfo). When it's empty in either of them, it's expected that we skip the idempotency check.

PR #8669, opened 2025-11-20
Change chasm.Field.Get signature to not return an error

bergundy:
## What changed?

- Also changed other methods that no longer need to return an error
- Also added a `TryGet` method that returns `(T, bool)` in case a nil check isn't enough
- Fixed various lint warnings that my editor warned about

## Why?

We decided to handle these errors on the framework level instead of the application level.

### DIFF
--- a/chasm/field.go
+++ b/chasm/field.go
@@ -70,7 +70,7 @@ func DataPointerTo[D proto.Message](
 
 // TryGet returns the value of the field and a boolean indicating if the value was found, deserializing if necessary.
 // Panics rather than returning an error, as errors are supposed to be handled by the framework as opposed to the
-// application.
+// application, even if the error is an application bug.
 func (f Field[T]) TryGet(chasmContext Context) (T, bool) {
 	var nilT T
 
@@ -146,7 +146,7 @@ func (f Field[T]) TryGet(chasmContext Context) (T, bool) {
 
 // Get returns the value of the field, deserializing it if necessary.
 // Panics rather than returning an error, as errors are supposed to be handled by the framework as opposed to the
-// application.
+// application, even if the error is an application bug.
 func (f Field[T]) Get(chasmContext Context) T {
 	v, ok := f.TryGet(chasmContext)
 	if !ok {


chasm/field.go
```diff
@@ -113,25 +120,40 @@ func (f Field[T]) Get(chasmContext Context) (T, error) {
...
-		return nilT, serviceerror.NewInternalf("node value doesn't implement %s", reflect.TypeFor[T]().Name())
+		// nolint:forbidigo // Panic is intended here for framework error handling.
+		panic(serviceerror.NewInternalf("node value doesn't implement %s", reflect.TypeFor[T]().Name()))
 	}
-	return vT, nil
+	return vT, true
+}
+
+// Get returns the value of the field, deserializing it if necessary.
+// Panics rather than returning an error, as errors are supposed to be handled by the framework as opposed to the
```
yycptt, line 148: nit: fwiw some errors could be due to errors in component code (like wrong T is specified, or expecting a component that doesn't exist), though yeah framework can handle them as well.

chasm/field.go
```diff
@@ -113,25 +120,40 @@ func (f Field[T]) Get(chasmContext Context) (T, error) {
...
+// application.
+func (f Field[T]) Get(chasmContext Context) T {
+	v, ok := f.TryGet(chasmContext)
+	if !ok {
+		// nolint:forbidigo // Panic is intended here for framework error handling.
+		panic(serviceerror.NewInternalf("field value of type %s not found", reflect.TypeFor[T]().Name()))
```
yycptt, line 154: nit: not sure if we want to mention the "of type %s" part, which should already panic in the type assertion in TryGet(). I think here it just means this method is called on an empty field or due to some bugs in the framework, value can't find/ is not populated.

yycptt (approved):

lina-temporal (approved):

chasm/field.go
bergundy: I thought it might be helpful for debugging...

PR #8671, opened 2025-11-20
Add Docker Images

chaptersix:
## What changed?

This PR add the new docker images with minimized dependencies. They are based on images from the [`temporalio/docker-builds`](https://github.com/temporalio/docker-builds/tree/d6f9c20abb3c9d21e22be26c255e4884af56a525) repository. These images will be used to build from version 1.30.0 and later. 

**This PR focuses on the Dockerfiles and local build tooling. Future PRs will add GitHub Actions for automated builds.**

## Changes Made

- Added Dockerfiles and supporting scripts for server and admin-tools images 
- Added docker-bake.hcl configuration for building images

## Why?

The existing build process is more complex than it has to be and existing images have dependencies that are not necessary.

## How did you test it?

- [x] run new images with the helm-charts
- [x] run new images with docker compose (single container)

## Potential risks

Some users might depend on some of the dependencies we've included in our images until now. Or there some special case that we haven't taken into account.

### DIFF
--- a/.github/docker/scripts/sh/entrypoint.sh
+++ b/.github/docker/scripts/sh/entrypoint.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
-
+# pipefail is included in alpine
+# shellcheck disable=SC3040
 set -eu -o pipefail
 
 # Resolve hostname to IP address for binding if not already set (supports both IPv4 and IPv6)


.github/docker/targets/admin-tools.Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY --chmod=755 ./build/${TARGETARCH}/tdbg /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-cassandra-tool /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-sql-tool /usr/local/bin/
+COPY ./build/${TARGETARCH}/ /tmp/binaries/
+RUN if [ -f /tmp/binaries/temporal-elasticsearch-tool ]; then \
+    cp /tmp/binaries/temporal-elasticsearch-tool /usr/local/bin/ && \
+    chmod 755 /usr/local/bin/temporal-elasticsearch-tool; \
+    fi && \
+    rm -rf /tmp/binaries
+COPY ./build/temporal/schema /etc/temporal/schema
```
chaptersix: This is here because the elastic search tool is not in an existing release. I'll probably be able to update in to a normal copy command when I add the GHA in a future PR.

.github/docker/targets/legacy-admin-tools.Dockerfile
```diff
@@ -0,0 +1,67 @@
+ARG BASE_IMAGE=alpine:3.22@sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d
```
picatz:
```suggestion
ARG ALPINE_IMAGE=alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
```

```console
$ docker buildx imagetools inspect alpine:3.22
Name:      docker.io/library/alpine:3.22
MediaType: application/vnd.oci.image.index.v1+json
Digest:    sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
...
```

.github/docker/targets/admin-tools.Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG BASE_IMAGE=alpine:3.22@sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d
```
picatz:
```suggestion
ARG ALPINE_IMAGE=alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
```

```console
$ docker buildx imagetools inspect alpine:3.22
Name:      docker.io/library/alpine:3.22
MediaType: application/vnd.oci.image.index.v1+json
Digest:    sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
...
```

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,33 @@
+ARG ALPINE_IMAGE=alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
```
bergundy: Why does this need to be an arg?

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN addgroup -g 1000 temporal && \
+    adduser -u 1000 -G temporal -D temporal && \
+    mkdir -p /etc/temporal/config/dynamicconfig && \
+    touch /etc/temporal/config/dynamicconfig/docker.yaml && \
+    chown -R temporal:temporal /etc/temporal/config
+
+WORKDIR /etc/temporal
+
+ENV TEMPORAL_HOME=/etc/temporal
+ENV TEMPORAL_SHA=${TEMPORAL_SHA}
```
bergundy: Why do we need these env vars?

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN apk add --no-cache \
+    ca-certificates \
+    tzdata
+
+
+RUN addgroup -g 1000 temporal && \
+    adduser -u 1000 -G temporal -D temporal && \
+    mkdir -p /etc/temporal/config/dynamicconfig && \
+    touch /etc/temporal/config/dynamicconfig/docker.yaml && \
```
bergundy: Does this need to be called `docker.yaml`?

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ARG TEMPORAL_VERSION=dev
+
+RUN apk add --no-cache \
+    ca-certificates \
+    tzdata
+
+
+RUN addgroup -g 1000 temporal && \
+    adduser -u 1000 -G temporal -D temporal && \
+    mkdir -p /etc/temporal/config/dynamicconfig && \
```
bergundy: `/etc` is already for config files, why do we need the extra `/config` and `/dynamicconfig` dirs?

.github/docker/scripts/sh/start-temporal.sh
bergundy, line 1: Can we get rid of this wrapper script if we're cleaning things up?

.github/docker/scripts/sh/entrypoint.sh
```diff
@@ -0,0 +1,40 @@
...
+            export TEMPORAL_ADDRESS="[${BIND_ON_IP}]:7233"
+            ;;
+        *)
+            # ipv4
+            export TEMPORAL_ADDRESS="${BIND_ON_IP}:7233"
+            ;;
+    esac
+fi
+
+# Support "bash" argument for debugging (drops into shell instead of starting server)
```
bergundy:
```suggestion
# Support "sh" argument for debugging (drops into shell instead of starting server)
```

.github/docker/scripts/sh/entrypoint.sh
bergundy, line 1: I wonder how much of this can be cleaned up.

.github/docker/targets/legacy-admin-tools.Dockerfile
bergundy, line 1: Why did you add the legacy image?

bergundy (commented):
I skimmed this and added some comments.

.github/docker/scripts/sh/entrypoint.sh
chaptersix: I looked into it briefly
1. `TEMPORAL_BROADCAST_ADDRESS` is needed so the server knows what addresses to broadcast for ringpop. I don't believe `0.0.0.0` is sufficient
2. You have to specify the port for IPv4 and IPv6 differently. 
3. It's possible to have multiple network interfaces for so `BIND_ON_IP` is needed. The interface the server is binding should be the same as what it's board casting

I'll look and see if the server itself is able to handle this in some way. I found some config that controls whether or not the server binds on localhost in the dynamic or normal config.

.github/docker/scripts/sh/start-temporal.sh
chaptersix: I'll look into it a bit more. There's probably some additional changes that are required in the server.

.github/docker/targets/legacy-admin-tools.Dockerfile
chaptersix: to support 3 minor release security update requirement.

.github/docker/targets/server.Dockerfile
chaptersix: I was keeping the existing pattern since the helm-charts writes to this config directory. I assume it has to exist in the image but will double check.

.github/docker/targets/server.Dockerfile
chaptersix: no, the normal config defines the path so we can change it.

.github/docker/targets/server.Dockerfile
chaptersix: I'll remove them. we don't need the sha since it's in the label info and home is not referenced anywhere.

.github/docker/targets/server.Dockerfile
chaptersix: So we can update it when doing security updates. I was planning on allowing the a workflow run to define it so we don't have to edit the github tag with the latest alpine tag.

.github/docker/scripts/sh/entrypoint.sh
chaptersix: We can just set good defaults for the embedded config and make sure helm has a valued. I'll make sure to include something about this in the release notes. 
Removed all scripts from the server image

.github/docker/scripts/sh/start-temporal.sh
chaptersix: yes, done

.github/docker/Makefile
```diff
@@ -0,0 +1,56 @@
+SERVER_VERSION ?= 1.29.1
+CLI_VERSION ?= 1.5.0
+TCTL_VERSION ?= 1.18.4
+IMAGE_REPO ?= temporaliotest
+TAG_LATEST ?= false
+
+TEMPORAL_SHA := $(shell git rev-parse HEAD)
+IMAGE_SHA_TAG := $(shell git rev-parse --short HEAD)
+IMAGE_BRANCH_TAG := $(shell git rev-parse --abbrev-ref HEAD)
+ALPINE_IMAGE ?= alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
```
chaptersix:
```suggestion
ALPINE_IMAGE ?= alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
```

chaptersix:
The makefile and build scripts will be removed In a future PR. Just here for demo/testing. 

The Alpine Image is an argument so we can trigger builds via workflows without committing to given branch. It should make re-building images easier

.github/docker/targets/legacy-server.Dockerfile
```diff
@@ -0,0 +1,55 @@
+ARG ALPINE_IMAGE
+
+FROM ${ALPINE_IMAGE}
+ARG TARGETARCH
+ARG TEMPORAL_SHA=unknown
+ARG DOCKERIZE_VERSION=v0.9.2
```
picatz: If we're defaulting to a particular version for `dockerize` (and assume we'll almost always use the `alpine` flavored release), might also be worth **considering** a `sha256:...` checksum match for the expected binary we download: https://github.com/jwilder/dockerize/releases/tag/v0.9.7

Something like:
```Dockerfile
ARG DOCKERIZE_CHECKSUM=sha256:152972b7c317d87ab24d0a173e18c2bc971f688652440fe3c931218231cff5e2
```

Then checking that matches the downloaded file before extracting the contents.

chaptersix: swapped to using go to install it.

.github/docker/scripts/bash/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
...
+    if [[ "${BIND_ON_IP}" =~ ":" ]]; then
+        # ipv6
+        export TEMPORAL_ADDRESS="[${BIND_ON_IP}]:7233"
+    else
+        # ipv4
+        export TEMPORAL_ADDRESS="${BIND_ON_IP}:7233"
+    fi
+fi
+
+# Support TEMPORAL_CLI_ADDRESS for backwards compatibility.
```
bergundy: I don't think we have any backwards compatibility concerns here.

.github/docker/scripts/bash/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
bergundy: Can we do with just `sh` in the image?

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,17 @@
+ARG ALPINE_IMAGE
+
+FROM ${ALPINE_IMAGE}
```
bergundy: Hardcode this instead of taking an arg.

.github/docker/targets/legacy-server.Dockerfile
bergundy, line 1: I still think you should just keep the legacy stuff in docker-builds and move forward here. @lilydoar just refactored all of the server image builds and I wouldn't want to regress here.

https://github.com/temporalio/docker-builds/pull/295

.github/docker/targets/server.Dockerfile
bergundy, line 1: Talk to @lilydoar and see the changes they made to the docker images recently to optimize the docker images. Those optimizations should be made here. Specifically, we can reduce the number of layers in the final container.

.github/docker/targets/admin-tools.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+COPY ./build/temporal/schema /etc/temporal/schema
+
+RUN addgroup -g 1000 temporal && \
+    adduser -u 1000 -G temporal -D temporal
+
+USER temporal
+WORKDIR /etc/temporal
+
+ENTRYPOINT ["tini", "--", "sleep", "infinity"]
```
bergundy: Do we need to be in `/etc/temporal` still?

Do we need `tini` here to call `sleep infinity`? Does it need to be the entrypoint? Can't we use a `CMD` for that?

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ARG TARGETARCH
+
+RUN apk add --no-cache \
+    ca-certificates \
+    tzdata && addgroup -g 1000 temporal && \
+    adduser -u 1000 -G temporal -D temporal
+
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-server /usr/local/bin/
+
+WORKDIR /etc/temporal
```
bergundy, line 13: Why do we need to be in `/etc/temporal`?

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+USER temporal
+
+ENTRYPOINT [ "temporal-server", "start" ]
```
bergundy: Shouldn't this be a `CMD`?

bergundy: I am concerned with the server listening on a public interface, can we make sure the server is accessible from outside?

.github/docker/build-from-source.sh
bergundy, line 1: Please don't add more shell scripts to the repo. Prefer writing this in Go.

.github/docker/build-from-source.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Build binaries from source by cloning temporal repo at specific SHA using goreleaser
```
bergundy: Why is this script required? Can't we just trigger the GH job for building the images from a specific sha?

.github/docker/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	TCTL_VERSION=$(TCTL_VERSION) \
+	IMAGE_REPO=$(IMAGE_REPO) \
+	IMAGE_SHA_TAG=$(IMAGE_SHA_TAG) \
+	IMAGE_BRANCH_TAG=$(IMAGE_BRANCH_TAG) \
+	TEMPORAL_SHA=$(TEMPORAL_SHA) \
+	TAG_LATEST=$(TAG_LATEST) \
+	ALPINE_IMAGE=$(ALPINE_IMAGE)
+
+.PHONY: build-tools
+build-admin-tools:
```
bergundy: Don't we need the schemas here?

.github/docker/targets/server.Dockerfile
chaptersix: Part of the script that I removed was for looking up the broadcast address if it was not defined or `BIND_ON_IP` was set to `0.0.0.0`. In the helm-charts that we publish both the `BIND_ON_IP` and the `TEMPORAL_ADDRESS` are set to the `POD_IP`. So that logic would never run.

chaptersix: Ports that are exposed are controlled outside of the Dockerfile. ex: helm chart or docker compose file. If we declare a set or port in this file it has no impact on what's actually exposed.

.github/docker/targets/server.Dockerfile
chaptersix: Kept this mostly because it's already there and it makes it easier to understand where a shell will launch when you exec into a pod

chaptersix: It also makes it a bit more clear where the configs should be mounted for the app. alpine has other configs in the `/etc` dir.

.github/docker/targets/server.Dockerfile
chaptersix:
```
# Resolve hostname to IP address for binding if not already set (supports both IPv4 and IPv6)
: "${BIND_ON_IP:=$(getent hosts "$(hostname)" | awk '{print $1;}')}"
export BIND_ON_IP

# If binding to wildcard address (0.0.0.0 or ::0), set broadcast address if not already set
if [[ "${BIND_ON_IP}" == "0.0.0.0" || "${BIND_ON_IP}" == "::0" ]]; then
    : "${TEMPORAL_BROADCAST_ADDRESS:=$(getent hosts "$(hostname)" | awk '{print $1;}')}"
    export TEMPORAL_BROADCAST_ADDRESS
fi

# Set default Temporal server address if not already configured
if [[ -z "${TEMPORAL_ADDRESS:-}" ]]; then
    echo "TEMPORAL_ADDRESS is not set, setting it to ${BIND_ON_IP}:7233"

    # IPv6 addresses contain colons and must be wrapped in brackets to avoid ambiguity with port separator
    if [[ "${BIND_ON_IP}" =~ ":" ]]; then
        # ipv6
        export TEMPORAL_ADDRESS="[${BIND_ON_IP}]:7233"
    else
        # ipv4
        export TEMPORAL_ADDRESS="${BIND_ON_IP}:7233"
    fi
fi
```

chaptersix: I don't think so. We expect this command to be run when using these images. I don't see a use case where a user will need to change this. 

https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/

```
The ENTRYPOINT instruction sets the default executable for the container. Any arguments supplied to the docker run command are appended to the ENTRYPOINT command.
```
If you'd like, I could move the start command to `CMD`.

.github/docker/targets/admin-tools.Dockerfile
chaptersix:
```suggestion
ENTRYPOINT ["tini", "--"]
CMD ["sleep", "infinity"]
```

chaptersix: https://github.com/krallin/tini?tab=readme-ov-file#why-tini

chaptersix: https://github.com/krallin/tini?tab=readme-ov-file#why-tini

.github/docker/Makefile
chaptersix:
```
# Schema is already available in the cloned repo at build/temporal/schema
# The Dockerfile expects it at ./build/temporal/schema which is where we cloned it
echo "Schema directory available at ${TEMPORAL_CLONE_DIR}/schema"
```

build from source handles this.

.github/docker/build-from-source.sh
chaptersix: I intend to remove it when I add the GHA that will actually execute the builds.

.github/docker/build-from-source.sh
chaptersix: yes, like I said above this be only adds the Dockerfiles and a few scripts that demonstrate they can be built. Trying to split up the work a bit so it's not one giant PR.

.github/docker/targets/admin-tools.Dockerfile
bergundy: How about this?


```
CMD ["sh" "-c", "trap exit INT HUP TERM; sleep infinity"]
```

.github/docker/targets/server.Dockerfile
bergundy: I would want `docker run ... temporalio/server sh` to execute a shell. That won't be possible with the entrypoint approach, you would have to specify `--entrypoint ''` for that.

bergundy: This should work on helm and just arbitrary docker run.

.github/docker/build-from-source.sh
bergundy: Can't we trigger the goreleaser action for building?

bergundy: I want to avoid introducing multiple ways of building the server.

.github/docker/targets/admin-tools.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-sql-tool /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-elasticsearch-tool /usr/local/bin/
+
+COPY ./build/temporal/schema /etc/temporal/schema
+
+USER temporal
+WORKDIR /etc/temporal
+
+ENTRYPOINT ["tini", "--"]
+CMD ["sleep", "infinity"]
```
bergundy:
```suggestion
CMD ["sh", "-c", "trap exit INT HUP TERM; sleep infinity"]
```

.github/docker/targets/admin-tools.Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM alpine:3.22@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412
+
+ARG TARGETARCH
+
+RUN apk add --no-cache \
+    ca-certificates \
+    tzdata \
+    tini && addgroup -g 1000 temporal && \
```
bergundy:
```suggestion
    tzdata && addgroup -g 1000 temporal && \
```

.github/docker/targets/admin-tools.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY --chmod=755 ./build/${TARGETARCH}/temporal /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/tdbg /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-cassandra-tool /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-sql-tool /usr/local/bin/
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-elasticsearch-tool /usr/local/bin/
+
+COPY ./build/temporal/schema /etc/temporal/schema
+
+USER temporal
+WORKDIR /etc/temporal
```
bergundy:
```suggestion
```

.github/docker/targets/server.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    tzdata && addgroup -g 1000 temporal && \
+    adduser -u 1000 -G temporal -D temporal
+
+COPY --chmod=755 ./build/${TARGETARCH}/temporal-server /usr/local/bin/
+COPY --chmod=755 ./scripts/sh/entrypoint.sh /etc/temporal/entrypoint.sh
+
+WORKDIR /etc/temporal
+USER temporal
+
+ENTRYPOINT [ "/etc/temporal/entrypoint.sh" ]
```
bergundy:
```suggestion
CMD [ "/etc/temporal/entrypoint.sh" ]
```

.github/docker/build-from-source.go
```diff
@@ -0,0 +1,305 @@
+package main
```
bergundy: Let's remove this script in favor of triggering the GH action for goreleaser.

bergundy (approved):
Approved with comments.

.github/docker/targets/admin-tools.Dockerfile
chaptersix: should be usr dir
will need to reference script folder in helm charts and docker compose.

.github/docker/targets/server.Dockerfile
chaptersix: `TEMPORAL_ADDRESS` was included at part of auto setup but we don't need it. removed

chaptersix: updated

.github/docker/scripts/sh/entrypoint.sh
chaptersix: ended up including the shell script. bit to dangerous to remove right now.

chaptersix:
> The makefile and build scripts will be removed In a future PR. Just here for demo/testing.
> 
> The Alpine Image is an argument so we can trigger builds via workflows without committing to given branch. It should make re-building images easier

removed in this PR

PR #8672, opened 2025-11-20
CHASM: Add ParentPtr field

yycptt:
## What changed?
- Add CHASM ParentPtr 

## Why?
- Current implementation of CHASM pointer is a separate physical node. But in most of the use case we have, we only need to point to the parent component which can be done entirely in memory without persisting anything.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/parent_pointer.go
+++ b/chasm/parent_pointer.go
@@ -31,7 +31,7 @@ type parentPtrInternal struct {
 	currentNode *Node
 }
 
-// Get returns the parent component this ParentPtr points to, deserializing it if necessary.
+// Get returns the parent component, deserializing it if necessary.
 // Panics rather than returning an error, as errors are supposed to be handled by the framework as opposed to the
 // application.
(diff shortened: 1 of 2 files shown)


chasm/parent_pointer.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Get returns the parent component, deserializing it if necessary.
+// Panics rather than returning an error, as errors are supposed to be handled by the framework as opposed to the
+// application.
+func (p ParentPtr[T]) Get(chasmContext Context) T {
+	vT, ok := p.TryGet(chasmContext)
+	if !ok {
+		// nolint:forbidigo // Panic is intended here for framework error handling.
+		panic(serviceerror.NewInternal("expect parent component value but got nil"))
```
bergundy, line 41:
```suggestion
		panic(serviceerror.NewInternal("expected parent component value but got nil"))
```

PR #8673, opened 2025-11-20
Remove transient dependencies on persistence in the chasm package

awln-temporal:
## What changed?
Remove transient dependencies on persistence in the chasm package. 

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/search_attribute.go
+++ b/chasm/search_attribute.go
@@ -5,7 +5,7 @@ import (
 	"time"
 
 	enumspb "go.temporal.io/api/enums/v1"
-	sadefs "go.temporal.io/server/common/searchattribute/defs"
+	"go.temporal.io/server/common/searchattribute/sadefs"
 )
 
 // CHASM Search Attribute User Guide:
--- a/chasm/visibility_value.go
+++ b/chasm/visibility_value.go
@@ -7,7 +7,7 @@ import (
 	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/server/common/payload"
-	sadefs "go.temporal.io/server/common/searchattribute/defs"
+	"go.temporal.io/server/common/searchattribute/sadefs"
 )
 
 type VisibilityValue interface {
--- a/common/persistence/sql/sqlplugin/mysql/query_converter.go
+++ b/common/persistence/sql/sqlplugin/mysql/query_converter.go
@@ -9,7 +9,7 @@ import (
 	"github.com/temporalio/sqlparser"
 	"go.temporal.io/server/common/persistence/sql/sqlplugin"
 	"go.temporal.io/server/common/persistence/visibility/store/query"
-	sadefs "go.temporal.io/server/common/searchattribute/defs"
+	"go.temporal.io/server/common/searchattribute/sadefs"
 )
 
 var maxDatetime = time.Date(9999, 12, 31, 23, 59, 59, 0, time.UTC)
--- a/common/persistence/sql/sqlplugin/mysql/query_converter_test.go
+++ b/common/persistence/sql/sqlplugin/mysql/query_converter_test.go
@@ -11,7 +11,7 @@ import (
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/server/common/persistence/sql/sqlplugin"
 	"go.temporal.io/server/common/persistence/visibility/store/query"
-	sadefs "go.temporal.io/server/common/searchattribute/defs"
+	"go.temporal.io/server/common/searchattribute/sadefs"
 )
 
 func TestQueryConverter_GetCoalesceCloseTimeExpr(t *testing.T) {
--- a/common/persistence/sql/sqlplugin/postgresql/query_converter.go
+++ b/common/persistence/sql/sqlplugin/postgresql/query_converter.go
@@ -8,7 +8,7 @@ import (
 	"github.com/temporalio/sqlparser"
 	"go.temporal.io/server/common/persistence/sql/sqlplugin"
 	"go.temporal.io/server/common/persistence/visibility/store/query"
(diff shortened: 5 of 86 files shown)


chasm/test_helpers_test.go
```diff
@@ -0,0 +1,155 @@
...
+// newTestHelper creates a new test helper from a test name.
+func newTestHelper(testName string) *testHelper {
+	return &testHelper{
+		testName: testName,
+		testHash: hashTestName(testName),
+	}
+}
+
+// NewTestHelper creates a test helper from a testing.T.
+func NewTestHelper(t testing.TB) *testHelper {
```
yycptt: looks like this doesn't need to be exported.

chasm/test_helpers_test.go
yycptt, line 1: not sure if it worth to build the same functionality as today's testvar package. Use some random/hardcoded value in tests works as well. But I don't have strong opinion.

common/nexus/callback_token.go
```diff
@@ -13,10 +14,12 @@ import (
 const (
 	// Currently supported token version.
 	TokenVersion = 1
-	// Header key for the callback token in StartOperation requests.
-	CallbackTokenHeader = "Temporal-Callback-Token"
 )
 
+// CallbackTokenHeader is the HTTP header name for the Nexus callback token.
+// Deprecated: Use types.CallbackTokenHeader instead.
+const CallbackTokenHeader = types.CallbackTokenHeader
```
awln-temporal: is this pattern acceptable 👀

common/nexus/types/constants.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+// CallbackTokenHeader is the HTTP header name for the Nexus callback token.
+// This is a shared constant used across packages that need to reference
+// the callback token header without importing the full nexus package.
+const CallbackTokenHeader = "Temporal-Callback-Token"
```
yycptt: cc @bergundy @pdoerner on the refactoring. 

I personally think it's probably better to move `chasm/nexus_completion.go` to a separate `nexus` package under `chasm`

common/persistence/task/constants.go
```diff
@@ -0,0 +1,7 @@
+package task
+
+import "time"
+
+// ScheduledTaskMinPrecision is the minimum precision for scheduled task times.
+// This matches the precision used by the persistence layer when storing and retrieving tasks.
+const ScheduledTaskMinPrecision = time.Millisecond
```
yycptt: let's define this in common/constants.go

common/persistence/data_interfaces.go
yycptt, line 1: can you add a type definition say `type Archetype = chasm.Archetype` to introduce a dependency on chasm package? just to make sure people won't accidentally land code that breaks it again.

chasm/test_helpers_test.go
```diff
@@ -0,0 +1,155 @@
+package chasm
+
+import (
+	"fmt"
+	"hash/fnv"
+	"math"
+	"math/rand"
+	"testing"
+
+	"github.com/pborman/uuid"
```
bergundy: Please use google/uuid

common/nexus/types/constants.go
awln-temporal: yea, that seems simpler

common/nexus/callback_token.go
bergundy: There's no need to deprecate or create this alias, but we can handle this later. Not blocking the PR.

chasm/test_helpers_test.go
awln-temporal: Yea, just wasn't sure if we wanted the test inputs to be relatively the same format as other unit tests, I can defer to a simpler approach as well.

chasm/field_test.go
```diff
@@ -12,7 +12,6 @@ import (
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/testing/protorequire"
 	"go.temporal.io/server/common/testing/testlogger"
-	"go.temporal.io/server/common/testing/testvars"
```
rodrigozhou, line 15: Why you cannot use `testvars` here?

yycptt: it depends on tons of packages to return typed values. one of them is "go.temporal.io/server/common/worker_versioning" which also some how depends on persistence...

```
go: can't load test package: package go.temporal.io/server/chasm
        imports go.temporal.io/server/common/testing/testvars from tree_test.go
        imports go.temporal.io/server/common/worker_versioning from test_vars.go
        imports go.temporal.io/server/common/persistence/visibility/manager from worker_versioning.go
        imports go.temporal.io/server/common/persistence from visibility_manager.go
        imports go.temporal.io/server/chasm from data_interfaces.go: import cycle not allowed in test
```

common/searchattribute/defs/constants.go
```diff
@@ -1,7 +1,8 @@
-package searchattribute
+package defs
```
bergundy: Can we rename this package `searchattributedefs` or `sadefs`? The name `defs` is very generic and makes it hard to understand from usage what it's referring to.

chasm/nexus_completion.go
yycptt, line 1: we can remove this file now.

chasm/test_helpers_test.go
yycptt: > if we wanted the test inputs to be relatively the same format as other unit tests

hmm, don't think there's such requirement.

common/searchattribute/defs/constants.go
awln-temporal: is it fine for nested packages to have simple names, and package importers to decide when to alias the package? Maybe the full name `definitions` or `constants` would be better as well.

I'm leaning towards separating search attribute definitions into a separate package, since it leaves constants in an isolated place for access if that's all we need. Which, most places of the code only need these constants.

awln-temporal: renamed to sadefs

PR #8676, opened 2025-11-21
Add command to dump dynamic config values (#421)

### LINKED ISSUE #421: Add admin command to dump dynamic config value[s]

vaibhavyadav-dev:
## What changed?
Added an admin command that dump all frontend dynamic config values,

## Why?
Operators currently cannot easily inspect frontend runtime dynamic configuration, dumping values aids troubleshooting, incident response, and verification after config changes.

## How did you test it?
- [x] built
- [x] added new unit test(s)

Closes #421

### DIFF
--- a/proto/internal/temporal/server/api/adminservice/v1/service.proto
+++ b/proto/internal/temporal/server/api/adminservice/v1/service.proto
@@ -218,4 +218,5 @@ service AdminService {
   }
 
     rpc GetDynamicConfigurations(GetDynamicConfigurationsRequest) returns (GetDynamicConfigurationsResponse) {}
+    rpc GetDynamicConfigurations(GetDynamicConfigurationsRequest) returns (GetDynamicConfigurationsResponse) {}
 }


vaibhavyadav-dev:
@bergundy I think It's better to add DefaultKey in this [handler](https://github.com/temporalio/temporal/blob/main/common/dynamicconfig/file_based_client.go#L106) .

Do let me know what you think,
Thanks

common/dynamicconfig/collection.go
```diff
@@ -161,6 +161,10 @@ func (c *Collection) GetPingChecks() []pingable.Check {
 	}
 }
 
+func (c *Collection) GetClient() Client {
```
dnr: is this used anywhere? I'd prefer not exposing the underlying Client if we don't really need to

cmd/tools/gendynamicconfig/dynamic_config.tmpl
```diff
@@ -75,13 +75,19 @@ func (s {{$P.Name}}TypedSetting[T]) Validate(v any) error {
...
+}
 
 func (s {{$P.Name}}TypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
 func (s {{$P.Name}}TypedConstrainedDefaultSetting[T]) Precedence() Precedence { return Precedence{{$P.Name}} }
 func (s {{$P.Name}}TypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
 }
+func (s {{$P.Name}}TypedConstrainedDefaultSetting[T]) DefaultValue() any {
+	return s.cdef
```
dnr: this feels weird, `s.cdef` is a `[]TypedConstrainedValue[T]`, which is a type that's going to be pretty hard/awkward for any client code to use. I suppose you're only using it serialize to json or something, so it doesn't really matter what the type is?

common/dynamicconfig/file_based_client.go
```diff
@@ -248,6 +249,16 @@ func loadFile(contents []byte) (configValueMap, *LoadResult) {
 
 			cvs[i].Value = val
 			cvs[i].Constraints = convertYamlConstraints(key, cv.Constraints, precedence, lr)
+
+			// check for default value
+			if cvs[i].Constraints == (Constraints{}) {
+				hasDefaultConstraint = true
+			}
+		}
+		// add default value if not present
```
dnr: this feels kind of weird. why do we need to insert the default values here? first of all, it doesn't work for settings with multiple constrained defaults. second, it's using more memory for something we don't actually need for looking up settings. why can't the code to retrieve settings just have logic to get the default from the registry instead of doing it here? it needs that anyway to get the default for a setting that's not preset in the file at all, right?

common/dynamicconfig/registry.go
```diff
@@ -17,6 +17,14 @@ var (
 	globalRegistry registry
 )
 
+func GetDefaultValueForKey(k Key) ConstrainedValue {
```
dnr: as above, this just doesn't make any sense for settings with constrained default values, there is no single default.

config/dynamicconfig/development-sql.yaml
```diff
@@ -61,6 +61,6 @@ matching.queryWorkflowTaskTimeoutLogRate:
 history.ReplicationEnableUpdateWithNewTaskMerge:
   - value: true
 history.hostLevelCacheMaxSize:
-  - value: 8192
+  - value: 8000
```
dnr: why is this changed?

proto/internal/temporal/server/api/dynamicconfig/v1/message.proto
```diff
@@ -0,0 +1,25 @@
...
+	string task_queue_name = 3;
+	int32 task_queue_type = 4;
+	int32 shard_id = 5;
+	int32 task_type = 6;
+	string destination = 7;
+}
+
+message ConstrainedValue {
+  Constraints constraints = 1;
+  google.protobuf.Struct value = 2;
```
dnr: hmm, shouldn't it be a `google.protobuf.Value` instead? most configs are scalars, there are structs and lists but they're not the majority.

service/frontend/admin_handler.go
```diff
@@ -24,6 +24,7 @@ import (
 	"go.temporal.io/server/api/adminservice/v1"
 	clusterspb "go.temporal.io/server/api/cluster/v1"
 	commonspb "go.temporal.io/server/api/common/v1"
+	dc "go.temporal.io/server/api/dynamicconfig/v1"
```
dnr: import should be named `dynamicconfigspb` ("spb" for "server proto buf")

service/frontend/admin_handler.go
```diff
@@ -2177,6 +2183,66 @@ func (adh *AdminHandler) getDLQWorkflowID(
...
+		dcEntry = adh.dynamicClient.GetValue(k)
+
+		// get global default value
+		if dcEntry == nil {
+			defaultValue := dynamicconfig.GetDefaultValueForKey(k)
+			dcEntry = append(dcEntry, defaultValue)
+		}
+
+		protoValues := make([]*dc.ConstrainedValue, 0)
+		for _, ConstrainedValue := range dcEntry {
```
dnr: use lowercase for variable names
```suggestion
		for _, cv := range dcEntry {
```

service/frontend/admin_handler.go
```diff
@@ -2177,6 +2183,66 @@ func (adh *AdminHandler) getDLQWorkflowID(
...
+		k := dynamicconfig.Key(key)
+		dcEntry = adh.dynamicClient.GetValue(k)
+
+		// get global default value
+		if dcEntry == nil {
+			defaultValue := dynamicconfig.GetDefaultValueForKey(k)
+			dcEntry = append(dcEntry, defaultValue)
+		}
+
+		protoValues := make([]*dc.ConstrainedValue, 0)
```
dnr:
```suggestion
		protoValues := make([]*dc.ConstrainedValue, 0, len(dcEntry))
```
or assign to specific indexes. or use util.MapSlice.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -598,3 +599,18 @@ message ForceUnloadTaskQueuePartitionRequest {
...
+  // The raw dynamic config loaded with all constraints.
+  map<string, temporal.server.api.dynamicconfig.v1.ConstrainedValues> dynamic_config = 2;
+}
+
+message GetDynamicConfigurationsRequest {
+  repeated string dynamic_config_keys = 1;
+}
+
+message GetDynamicConfigurationsResponse {
+  repeated HostConfig host_config = 1;
```
dnr, line 615: why repeated? do we allow querying multiple hosts at once?

config/dynamicconfig/development-sql.yaml
vaibhavyadav-dev: My bad, I did this to test my changes and forgot revert it.

proto/internal/temporal/server/api/dynamicconfig/v1/message.proto
vaibhavyadav-dev: ok

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
vaibhavyadav-dev: Not now, but do have plan to make so.

vaibhavyadav-dev:
Thanks for the review @dnr 
I have a question about the approach since It doesn't work for settings with multiple constrained defaults, and inserting defaults at load time is bad.

I'm considering two alternatives.
-  Only display values that are explicitly configured in the dynamic config file. If a setting isn't present, don't show it.
-  Retrieve defaults at display time from the registry based on the actual constraints being queried.

Which approach would you prefer? I'm leaning toward first one as it's simpler and provides value for operators troubleshooting config issues.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
bergundy: This was my idea, so you can compare dynamic configs across hosts. This will be a future capability though.

proto/internal/temporal/server/api/dynamicconfig/v1/message.proto
bergundy: I'm not sure it should be `Value`, should we just dump the struct as JSON? It could be simpler. But `Value` would be better than `Struct`, agree.

common/dynamicconfig/registry.go
```diff
@@ -38,6 +38,11 @@ func queryRegistry(k Key) GenericSetting {
 	return globalRegistry.settings[strings.ToLower(k.String())]
 }
 
+func GetDefaultValues(k Key) []ConstrainedValue {
```
bergundy, line 40:
```suggestion
// DefaultValue returns the default value defined for the setting of the given key.
func DefaultValue(k Key) []ConstrainedValue {
```

proto/internal/temporal/server/api/dynamicconfig/v1/message.proto
```diff
@@ -0,0 +1,25 @@
...
+	string task_queue_name = 3;
+	int32 task_queue_type = 4;
+	int32 shard_id = 5;
+	int32 task_type = 6;
+	string destination = 7;
+}
+
+message ConstrainedValue {
+  Constraints constraints = 1;
+  google.protobuf.Value value = 2;
```
bergundy, line 20: I wonder if it would be simpler to give back a serialized JSON object here.

service/frontend/admin_handler.go
```diff
@@ -2230,6 +2236,65 @@ func (adh *AdminHandler) archetypeNameToID(archetype chasm.Archetype) (chasm.Arc
...
+	dynamicConfig := make(map[string]*dynamicconfigspb.ConstrainedValues, len(requestedKeys))
+
+	for _, key := range requestedKeys {
+		var dcEntry []dynamicconfig.ConstrainedValue
+		k := dynamicconfig.Key(key)
+		dcEntry = adh.dynamicClient.GetValue(k)
+
+		// get global default value
+		if dcEntry == nil {
+			dcEntry = dynamicconfig.GetDefaultValues(k)
```
bergundy, line 2275: Do we even want to give back the default value? I was thinking we would return the "raw" values without deserializing them. I'm not sure if the default values are always JSON serializable.

I can look into this a bit more but @dnr can speak to it better than I can.

service/frontend/admin_handler.go
```diff
@@ -2230,6 +2236,65 @@ func (adh *AdminHandler) archetypeNameToID(archetype chasm.Archetype) (chasm.Arc
...
+		}
+
+		protoValues := make([]*dynamicconfigspb.ConstrainedValue, 0, len(dcEntry))
+		for _, cv := range dcEntry {
+			s, err := structpb.NewValue(normalizeValue(cv.Value))
```
bergundy, line 2280: This isn't ideal, we should be able to return all of the values back. Users of this API will be surprised to not get responses back.

Reading the implementation of `NewValue` I'm not convinced that it will be able to convert everything that we support, which is why I think returning a JSON serialized representation is probably a better option.
We should also test this API with various complex configs to see that we return what is expected.

proto/internal/temporal/server/api/dynamicconfig/v1/message.proto
```diff
@@ -0,0 +1,25 @@
...
+package temporal.server.api.dynamicconfig.v1;
+option go_package = "go.temporal.io/server/api/dynamicconfig/v1;dynamicconfig";
+
+import "google/protobuf/struct.proto";
+
+message Constraints {
+	string namespace = 1;
+	string namespace_id = 2;
+	string task_queue_name = 3;
+	int32 task_queue_type = 4;
```
bergundy, line 12: We should be able to use 
`TaskQueueType` from `temporal/api/enums/v1/task_queue.proto` and `TaskType` from `temporal/server/api/enums/v1/task.proto`.

common/dynamicconfig/registry.go
bergundy: But I'm not even convinced we want this.

service/frontend/admin_handler_test.go
```diff
@@ -85,6 +85,7 @@ type (
 		mockProducer               *persistence.MockNamespaceReplicationQueue
 		mockMatchingClient         *matchingservicemock.MockMatchingServiceClient
 		mockSaMapper               *searchattribute.MockMapper
+		dynamicClient              dynamicconfig.Client
```
bergundy, line 90: nit: call this `dynamicConfigClient`

service/frontend/admin_handler_test.go
```diff
@@ -138,6 +139,14 @@ func (s *adminHandlerSuite) SetupTest() {
 		NumHistoryShards: 1,
 	}
 
+	dynamicConfigClient := dynamicconfig.NewMemoryClient()
+	dynamicConfigClient.OverrideValue("frontend.rps", 10)
+	dynamicConfigClient.OverrideValue("matching.rps", 20)
+	dynamicConfigClient.OverrideValue("history.rps", 30)
+	dynamicConfigClient.OverrideValue("limit.maxIDLength", 255)
+	dynamicConfigClient.OverrideValue("history.enableTransitionHistory", true)
```
bergundy, line 149: Test more complex configs here. Anything that has slices or maps or is a more complex struct. There are a few of those in `dynamicconfig/constants.go`.

tools/tdbg/tdbg_commands.go
```diff
@@ -69,6 +69,12 @@ func getCommands(
 			Usage:       "Decode payload",
 			Subcommands: newDecodeCommands(taskBlobEncoder),
 		},
+		{
+			Name: "config",
+			Usage: "Command to get dynamic config values",
+			Aliases: []string{"cfg"},
```
bergundy, line 81:
```suggestion
```

tools/tdbg/tdbg_commands.go
```diff
@@ -897,3 +903,23 @@ func newDecodeCommands(
 		},
 	}
 }
+
+func newGetConfigValuesCommands(ClientFactory ClientFactory) []*cli.Command {
+	return []*cli.Command{
+		{
+			Name:    "get",
+			Aliases: []string{"g"},
```
bergundy, line 975:
```suggestion
```

tools/tdbg/tdbg_commands.go
```diff
@@ -897,3 +903,23 @@ func newDecodeCommands(
...
+			Usage:   "Get frontend dynamic configuration",
+			Flags: []cli.Flag{
+				&cli.StringSliceFlag{
+					Name:  FlagKey,
+					Usage: "Enter keys",
```
bergundy, line 980:
```suggestion
					Usage: "Dynamic config keys to get",
```

tools/tdbg/commands.go
```diff
@@ -745,3 +745,29 @@ func AdminReplicateWorkflow(
...
 }
+
+func AdminGetClusterConfig(c *cli.Context, clientFactory ClientFactory) error {
+	adminClient := clientFactory.AdminClient(c)
+	ctx, cancel := newContext(c)
+	defer cancel()
+
+    keys := c.StringSlice(FlagKey)
+    if len(keys) == 0 {
+        return fmt.Errorf("flag %q is required", FlagKey)
```
bergundy, line 882:
```suggestion
        return fmt.Errorf("at least one %q flag is required", FlagKey)
```

vaibhavyadav-dev:
Closing this PR due to it being outdated and hard to rebase cleanly.
I'll open a fresh PR with updated changes.

PR #8680, opened 2025-11-21
Ensure versioning transition properly reschedules tasks with statebased replication

ShahabT:
## What changed?
Ensuring the activity and workflow task re-generation works fine on the passive side when statebased replication is turned on.

## Why?
Otherwise, the workflows can get stuck after failover in certain corner case.

## How did you test it?
Existing tests should cover regressions.
Will add new tests to ensure the fix works as intended.

- [ ] built
- [ ] added new unit test(s)

## Potential risks
Patch candidate.

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -8577,13 +8577,13 @@ func (ms *MutableStateImpl) reschedulePendingWorkflowTask() error {
 		ms.logInfo("start transition did not reschedule pending speculative task")
 		return nil
 	}
+	// Reset the attempt; forcing a non-transient workflow task to be scheduled.
+	ms.executionInfo.WorkflowTaskAttempt = 1
 
 	// Increase the stamp ("version") to invalidate the pending non-speculative WFT.
 	// We don't invalidate speculative WFTs because they are very latency sensitive.
 	ms.executionInfo.WorkflowTaskStamp += 1
-
-	// Reset the attempt; forcing a non-transient workflow task to be scheduled.
-	ms.executionInfo.WorkflowTaskAttempt = 1
+	ms.workflowTaskUpdated = true
 
 	return ms.taskGenerator.GenerateScheduleWorkflowTaskTasks(pendingTask.ScheduledEventID)
 }


service/history/workflow/mutable_state_impl.go
```diff
@@ -3143,7 +3143,7 @@ func (ms *MutableStateImpl) ApplyBuildIdRedirect(
 	ms.GetExecutionInfo().BuildIdRedirectCounter = redirectCounter
 
 	// Re-scheduling pending workflow and activity tasks.
-	err = ms.reschedulePendingWorkflowTask(false)
+	err = ms.reschedulePendingWorkflowTask()
```
ShahabT, line 3146: these should be true but was scoped out from @stephanos PR based on https://github.com/temporalio/temporal/pull/8103/files#r2403627911

service/history/workflow/mutable_state_impl.go
```diff
@@ -8559,14 +8578,12 @@ func (ms *MutableStateImpl) reschedulePendingWorkflowTask(invalidatePendingTasks
 		return nil
 	}
 
-	if invalidatePendingTasks {
-		// Increase the stamp ("version") to invalidate the pending non-speculative WFT.
-		// We don't invalidate speculative WFTs because they are very latency sensitive.
-		ms.executionInfo.WorkflowTaskStamp += 1
+	// Increase the stamp ("version") to invalidate the pending non-speculative WFT.
+	// We don't invalidate speculative WFTs because they are very latency sensitive.
+	ms.executionInfo.WorkflowTaskStamp += 1
```
yycptt, line 8585: let's set ms.workflowTaskUpdated = true since we are not using UpdateWorkflowTask() method here.

service/history/workflow/mutable_state_impl.go
yycptt, line 3151: nit for later: should this also somehow reuse reschedulePendingActivities as well?

yycptt (approved):

service/history/workflow/mutable_state_impl.go
stephanos: Right; I didn't want to make that change since I was worried it would require more reviews and introduce more risk.

service/history/workflow/mutable_state_impl.go
ShahabT: old versioning stuff will be cleaned up

PR #8685, opened 2025-11-23
Simplify Config Templating

chaptersix:
## What changed?
* Remove previously support for `dockerize` syntax. Now we just support `sprig`
* updated embedded config 

## Why?
Simplification 

## How did you test it?
- [x] added new unit test(s)

## Potential risks
Existing config templates will need to be updated to use `sprig` only syntax. Otherwise the server will not be able to read it.


common/config/config_template_embedded.yaml
```diff
@@ -1,163 +1,162 @@
 # enable-template
-# TODO: Remove this file after temporalio/docker-builds repository is archived
 log:
     stdout: true
-    level: {{ default .Env.LOG_LEVEL "info" }}
+    level: {{ default "info" (env "LOG_LEVEL") }}
```
chaptersix, line 4: This is worth looking at. (Hidden by Github do to the large number of changes)

common/config/config_test.go
```diff
@@ -13,17 +14,31 @@ func TestToString(t *testing.T) {
 	require.NotEmpty(t, cfg.String())
 }
 
-func TestEmbeddedTemplateOnlyDiffersFromDockerByComment(t *testing.T) {
-	embeddedContent, err := os.ReadFile("config_template_embedded.yaml")
+func TestEmbeddedTemplateConsistentWithDockerTemplate(t *testing.T) {
```
chaptersix:
```suggestion
func TestEmbeddedTemplateConsistentWithDockerTemplate(t *testing.T) {
```
this test in no longer relevant since there's a syntax different between the 2 files. docker-builds still expects the `dockerize` syntax.

common/config/loader.go
```diff
@@ -253,83 +237,22 @@ func processConfigFile(data []byte, filename string, envMap map[string]string) (
...
-
-	tpl, err := template.New(filename).Funcs(templateFuncs).Parse(string(data))
+	tpl, err := template.New(filename).Funcs(sprig.FuncMap()).Parse(string(data))
 	if err != nil {
 		return nil, err
 	}
 
 	var rendered bytes.Buffer
-	err = tpl.Execute(&rendered, ctx)
+	err = tpl.Execute(&rendered, nil)
```
chaptersix, line 246: no need to pass in the environment as context

CONTRIBUTING.md
```diff
@@ -265,8 +265,6 @@ All PR titles should start with Upper case and have no dot at the end.
 and [base-builder](https://github.com/temporalio/docker-builds/blob/main/docker/base-images/base-builder.Dockerfile)~~ 
 **Note:** The docker-builds repository is now deprecated and will be archived.
 
-<!-- TODO: Remove docker/config_template.yaml after temporalio/docker-builds repository is archived -->
```
chaptersix, line 268: note is already in the file and this does not render.

PR #8665, opened 2025-11-19
Add support for standalone activity termination

fretz12:
## What changed?
Added standalone activity termination handling.

## Why?
Needed to support standalone activities full operation.

## How did you test it?
- [X] built
- [X] run locally and tested manually
- [X] covered by existing tests
- [X] added new unit test(s)
- [X] added new functional test(s)

---

> [!NOTE]
> Adds termination support for standalone activities across API, frontend/handler, state machine, and tests.
> 
> - **API/Proto**:
>   - Add `TerminateActivityExecution` RPC: request/response messages in `proto/v1/request_response.proto` and service in `proto/v1/service.proto`.
>   - Regenerate clients/servers: `activitypb` request/response, service, layered client, and gRPC stubs.
> - **Frontend**:
>   - Implement `TerminateActivityExecution` in `frontend.go` to resolve namespace and forward to activity service.
> - **Backend/Handler**:
>   - Add handler `TerminateActivityExecution` to update component via `(*Activity).handleTerminated`.
>   - Add `Activity.handleTerminated` to apply termination transition.
> - **State Machine**:
>   - Enhance `TransitionTerminated` to finalize execution, set `LastWorkerIdentity`, and record `TerminatedFailureInfo` in outcome.
> - **Tests**:
>   - Add unit test `TestTransitionTerminated`.
>   - Add functional tests for terminating an activity and preventing termination after completion.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de5e4f2a7f95ffbb90d64479f035b13ad2d14e0b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #8688, opened 2025-11-24
Add support for standalone activity cancellation

fretz12:
## What changed?
Added standalone activity cancellation request and respond handling.

## Why?
Needed to support standalone activities full operation.

## How did you test it?
- [X] built
- [X] added new unit test(s)

---

> [!NOTE]
> Implements end-to-end standalone activity cancellation with new RPCs, state transitions, token routing, and tests.
> 
> - **API/Proto**:
>   - Add `CancelActivityExecution` RPC and messages (`CancelActivityExecutionRequest/Response`) in `proto/v1` and generated clients/servers.
> - **Frontend**:
>   - New handler `RequestCancelActivityExecution` validating/normalizing `request_id` and calling `ActivityService.CancelActivityExecution`.
>   - For ById activity APIs, when `workflow_id` is empty, embed `componentRef` in task tokens (completed/failed/canceled paths).
> - **Activity State Machine**:
>   - Add transitions `TransitionCancelRequested` and `TransitionCanceled`; allow `CANCEL_REQUESTED` in `Completed/Failed/Terminated/TimedOut` sources.
>   - Implement `handleCancellationRequested` and `HandleCanceled`; surface `CanceledReason` in `buildActivityExecutionInfo`.
>   - Guard retries when reschedule is not possible.
> - **History**:
>   - `RespondActivityTaskCanceled` handles standalone activities via `componentRef` by invoking `Activity.HandleCanceled`.
> - **Validation**:
>   - Rename and use `validateAndNormalizeRequestID`.
> - **Tests**:
>   - Add unit and functional tests covering cancel request idempotency/conflicts, respond-canceled flows, and post-cancel completions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a42eeafdde6046dc6f82f57ecbb1a1df9b9ed929. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/matching/matching_engine.go
+++ b/service/matching/matching_engine.go
@@ -1332,8 +1332,8 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 					if v.GetVersion() == nil || v.GetVersion().GetDeploymentName() == "" || v.GetVersion().GetBuildId() == "" {
 						continue
 					}
-					buildID := worker_versioning.WorkerDeploymentVersionToStringV32(v.GetVersion())
-					buildIds = append(buildIds, buildID)
+					buildId := worker_versioning.WorkerDeploymentVersionToStringV32(v.GetVersion())
+					buildIds = append(buildIds, buildId)
 				}
 			}
 


fretz12:
cursor review

chasm/lib/activity/proto/v1/service.proto
```diff
@@ -19,4 +19,8 @@ service ActivityService {
     rpc TerminateActivityExecution(TerminateActivityExecutionRequest) returns (TerminateActivityExecutionResponse) {
       option (temporal.server.api.routing.v1.routing).execution_id = "frontend_request.activity_id";
     }
+
+    rpc CancelActivityExecution(CancelActivityExecutionRequest) returns (CancelActivityExecutionResponse) {
```
bergundy: Use `Request` here, it's not actually canceling the activity. It's requesting it to be canceled.

chasm/lib/activity/activity.go
```diff
@@ -259,6 +259,25 @@ func (a *Activity) HandleFailed(ctx chasm.MutableContext, req *historyservice.Re
...
+func (a *Activity) HandleCanceled(ctx chasm.MutableContext, request *historyservice.RespondActivityTaskCanceledRequest) (
+	*historyservice.RespondActivityTaskCanceledResponse, error) {
+	if err := TransitionCanceled.Apply(a, ctx, request); err != nil {
+		return nil, err
+	}
+
+	return &historyservice.RespondActivityTaskCanceledResponse{}, nil
+}
+
+func (a *Activity) handleTerminated(ctx chasm.MutableContext, req *activitypb.TerminateActivityExecutionRequest) (
```
bergundy, line 276: I thought you only implemented cancellation in this PR based on the title?

chasm/lib/activity/statemachine.go
```diff
@@ -280,11 +283,75 @@ var TransitionTerminated = chasm.NewTransition(
...
 
+// TransitionCancelRequested affects a transition to CancelRequested status
+var TransitionCancelRequested = chasm.NewTransition(
+	[]activitypb.ActivityExecutionStatus{
+		activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED,
+		activitypb.ACTIVITY_EXECUTION_STATUS_STARTED,
+		activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED, // Allow idempotent transition
+	},
+	activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
+	func(a *Activity, ctx chasm.MutableContext, req *activitypb.CancelActivityExecutionRequest) error {
```
bergundy: nit: take the frontend request as argument here so you don't have to convert it.

tests/standalone_activity_test.go
```diff
@@ -190,6 +191,334 @@ func (s *standaloneActivityTestSuite) TestActivityFailedByID() {
...
+		},
+	}
+
+	_, err := s.FrontendClient().RespondActivityTaskCanceled(ctx, &workflowservice.RespondActivityTaskCanceledRequest{
+		Namespace: s.Namespace().String(),
+		TaskToken: pollTaskResp.TaskToken,
+		Details:   details,
+		Identity:  "new-worker",
+	})
+	require.Error(t, err)
```
bergundy: This should fail with an invalid argument error.

tests/standalone_activity_test.go
```diff
@@ -190,6 +191,334 @@ func (s *standaloneActivityTestSuite) TestActivityFailedByID() {
...
+				_, err := s.FrontendClient().RespondActivityTaskCompleted(ctx, &workflowservice.RespondActivityTaskCompletedRequest{
+					Namespace: s.Namespace().String(),
+					TaskToken: taskToken,
+					Result:    defaultResult,
+				})
+
+				return err
+			},
+			expectedStatus: enumspb.ACTIVITY_EXECUTION_STATUS_COMPLETED,
+			expectedState:  enumspb.PENDING_ACTIVITY_STATE_UNSPECIFIED,
```
bergundy: Can remove this if it's always expected to be unspecified.

tests/standalone_activity_test.go
```diff
@@ -190,6 +191,334 @@ func (s *standaloneActivityTestSuite) TestActivityFailedByID() {
...
+		Namespace:  s.Namespace().String(),
+		ActivityId: s.tv.ActivityID(),
+		RunId:      runID,
+		Identity:   "cancelling-worker",
+		RequestId:  s.tv.RequestID(),
+		Reason:     "Test Cancellation",
+	})
+	require.NoError(t, err)
+
+	// TODO: we should get the cancel request from heart beat once we implement it
```
bergundy, line 211: Yes!

chasm/lib/activity/validator.go
```diff
@@ -208,7 +208,7 @@ func ValidateStandaloneActivity(
 	return nil
 }
 
-func validateRequestID(requestID *string, maxIDLengthLimit int) error {
+func validateAndNormalizeRequestID(requestID *string, maxIDLengthLimit int) error {
 	if *requestID == "" {
```
bergundy: Just wondering if this could be nil and panic...

chasm/lib/activity/statemachine.go
```diff
@@ -280,11 +283,75 @@ var TransitionTerminated = chasm.NewTransition(
...
+		}
+
+		return nil
+	},
+)
+
+// TransitionCanceled affects a transition to Canceled status
+var TransitionCanceled = chasm.NewTransition(
+	[]activitypb.ActivityExecutionStatus{
+		activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
```
bergundy, line 303: It can transition from scheduled too, just not from started.

chasm/lib/activity/activity.go
```diff
@@ -275,13 +294,26 @@ func (a *Activity) getLastHeartbeat(ctx chasm.MutableContext) (*activitypb.Activ
...
+	if a.GetStatus() == activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED {
+		if existingReqID != newReqID {
+			return nil, serviceerror.NewFailedPrecondition(
+				fmt.Sprintf("cancellation already requested with request ID %s", existingReqID))
+		}
+
+		return &activitypb.CancelActivityExecutionResponse{}, nil
+	}
+
+	if err := TransitionCancelRequested.Apply(a, ctx, req); err != nil {
```
bergundy: You need to check if you're in scheduled state and immediate transition to canceled.

bergundy (commented):
Overall this looks really good, it's almost completely correct.

chasm/lib/activity/activity.go
fretz12: The [terminate PR](https://github.com/temporalio/temporal/pull/8665) is the destination for this PR. Here i've moved the "handle" methods together so it's easier to find

chasm/lib/activity/statemachine.go
fretz12: Added

chasm/lib/activity/activity.go
fretz12: Gotcha, added the check and modified the transition-from states. Also added a test to cover such a case (disabled for now, see below).

However, I'm running into `unable to change workflow state from Created to Completed, status Failed` from the chasm engine. This was the [same issue with the schedule-to-X timeouts](https://github.com/temporalio/temporal/pull/8573#discussion_r2511636330). I believe once we implement the SearchAttributes interface (and @yycptt and team address anything outstanding?), this shouldn't fail then?

chasm/lib/activity/proto/v1/service.proto
fretz12: Renamed rpc + the req/resp messages

chasm/lib/activity/statemachine.go

chasm/lib/activity/validator.go
fretz12: The callers of this func all pass the reqID from a proto string so shouldn't be nil. Nonetheless, I've added a nil pointer check if this ever gets abused. 
Not thrilled about passing in a string ptr but it also encapsulates validation + normalization in one go. If you'd prefer we take in a string and return a string the caller should set, we can consider that.

tests/standalone_activity_test.go
fretz12: Invalid arg error doesn't seem to fit well here cuz it's not really passing a wrong arg but making a call with the wrong conditions... maybe FailedPrecondition is more fitting?

tests/standalone_activity_test.go
fretz12: Removed all occurences

chasm/lib/activity/validator.go
```diff
@@ -208,7 +208,11 @@ func ValidateStandaloneActivity(
 	return nil
 }
 
-func validateRequestID(requestID *string, maxIDLengthLimit int) error {
+func validateAndNormalizeRequestID(requestID *string, maxIDLengthLimit int) error {
+	if requestID == nil {
+		return serviceerror.NewInvalidArgument("RequestID is required.")
```
bergundy: This error is misleading, it's not that request ID is required, it's that within the server codebase you expect this to be a non-nil pointer.

chasm/lib/activity/validator.go
```diff
@@ -208,7 +208,11 @@ func ValidateStandaloneActivity(
 	return nil
 }
 
-func validateRequestID(requestID *string, maxIDLengthLimit int) error {
+func validateAndNormalizeRequestID(requestID *string, maxIDLengthLimit int) error {
```
bergundy: Given the simplicity of the body of this function and how complex it gets with passing string pointers around, I would recommend that you inline the implementation.

tests/standalone_activity_test.go
bergundy: Yeah, that makes sense. Please always check with `require.ErrorAs` (or `ErrorIs` if applicable) in tests to verify the contract. Especially important for functional tests.

tests/standalone_activity_test.go
```diff
@@ -190,6 +191,365 @@ func (s *standaloneActivityTestSuite) TestActivityFailedByID() {
...
+
+	details := &commonpb.Payloads{
+		Payloads: []*commonpb.Payload{
+			payload.EncodeString("Canceled Details"),
+		},
+	}
+
+	_, err := s.FrontendClient().RespondActivityTaskCanceled(ctx, &workflowservice.RespondActivityTaskCanceledRequest{
```
bergundy: No need to test after the second request is issued. This should already be covered by previous tests.

tests/standalone_activity_test.go
```diff
@@ -190,6 +191,365 @@ func (s *standaloneActivityTestSuite) TestActivityFailedByID() {
...
+
+	_, err = s.FrontendClient().RequestCancelActivityExecution(ctx, &workflowservice.RequestCancelActivityExecutionRequest{
+		Namespace:  s.Namespace().String(),
+		ActivityId: s.tv.ActivityID(),
+		RunId:      runID,
+		Identity:   "cancelling-worker",
+		RequestId:  "different-cancel-request-id",
+		Reason:     "Test Cancellation",
+	})
+	require.Error(t, err)
```
bergundy: Check that the right error is returned here (failed precondition).

tests/standalone_activity_test.go
```diff
@@ -226,7 +586,6 @@ func (s *standaloneActivityTestSuite) TestActivityTerminated() {
 	require.NoError(t, err)
 	s.validateBaseActivityResponse(t, activityID, runID, activityResp)
 	require.Equal(t, enumspb.ACTIVITY_EXECUTION_STATUS_TERMINATED, info.GetStatus())
-	require.Equal(t, enumspb.PENDING_ACTIVITY_STATE_UNSPECIFIED, info.GetRunState())
```
bergundy, line 223: Why did you remove these assertions? Seems unrelated.

chasm/lib/activity/activity.go
```diff
@@ -275,13 +294,36 @@ func (a *Activity) getLastHeartbeat(ctx chasm.MutableContext) (*activitypb.Activ
...
+
+	// If in scheduled state, cancel immediately
+	if a.GetStatus() == activitypb.ACTIVITY_EXECUTION_STATUS_SCHEDULED {
+		err := TransitionCanceled.Apply(a, ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		return &activitypb.RequestCancelActivityExecutionResponse{}, nil
+	}
```
bergundy: First transition to CancelRequested IMHO and then cancel. You will want the request ID deduplication and to record the cancel request reason.

fretz12: That seems more consistent. Changed.

chasm/lib/activity/validator.go
fretz12: Seems to be the right thing to do at this point. Changed.

chasm/lib/activity/validator.go
fretz12: No longer relevant given above changes.

tests/standalone_activity_test.go
fretz12: Done. I've made the change in statemachine `Apply` during the the Possible check to reutrn FailedPrecondition, so any invalid transition will return this err type now.

tests/standalone_activity_test.go
fretz12: Removed this, and changed the check of state/status

tests/standalone_activity_test.go
fretz12: done

tests/standalone_activity_test.go
fretz12: I made this change in response to this comment for any terminal states, including terminate. LMK if there's a misunderstanding. https://github.com/temporalio/temporal/pull/8688#discussion_r2566944132

chasm/lib/activity/activity.go
```diff
@@ -259,6 +260,25 @@ func (a *Activity) HandleFailed(ctx chasm.MutableContext, req *historyservice.Re
 	return &historyservice.RespondActivityTaskFailedResponse{}, nil
 }
 
+// HandleCanceled updates the activity on activity canceled.
+func (a *Activity) HandleCanceled(ctx chasm.MutableContext, request *historyservice.RespondActivityTaskCanceledRequest) (
+	*historyservice.RespondActivityTaskCanceledResponse, error) {
```
dandavison: nit: this isn't the formatting style that the rest of the codebase uses. They do
```go
func (a *Activity) HandleCanceled(
	ctx chasm.MutableContext,
	request *historyservice.RespondActivityTaskCanceledRequest,
) (*historyservice.RespondActivityTaskCanceledResponse, error) {
```

fretz12: reformatted all occurences.

tests/standalone_activity_test.go
bergundy: Ah makes sense, thanks.

chasm/lib/activity/activity.go
```diff
@@ -232,7 +235,8 @@ func (a *Activity) HandleCompleted(ctx chasm.MutableContext, request *historyser
 // HandleFailed updates the activity on activity failure. if the activity is retryable, it will be rescheduled
 // for retry instead.
 func (a *Activity) HandleFailed(ctx chasm.MutableContext, req *historyservice.RespondActivityTaskFailedRequest) (
-	*historyservice.RespondActivityTaskFailedResponse, error) {
+	*historyservice.RespondActivityTaskFailedResponse, error,
+) {
```
bergundy, line 238: Don't have a strong opinion over the style here. Feel free to do whatever you want as long as its consistent with other places in the codebase.

I personally typically prefer this style (but don't think it's necessarily others opinion and I don't wish to block the PR on this):


```suggestion
func (a *Activity) HandleFailed(
	ctx chasm.MutableContext,
	req *historyservice.RespondActivityTaskFailedRequest
) (*historyservice.RespondActivityTaskFailedResponse, error) {
```

chasm/lib/activity/frontend.go
```diff
@@ -132,6 +134,37 @@ func (h *frontendHandler) TerminateActivityExecution(
...
+	// Since validation potentially mutates the request, we clone it first so that any retries use the original request.
+	req = common.CloneProto(req)
+
+	if req.GetRequestId() == "" {
+		req.RequestId = uuid.NewString()
+	}
+
+	if len(req.GetRequestId()) > dynamicconfig.MaxIDLengthLimit.Get(h.dc)() {
+		return nil, serviceerror.NewInvalidArgument("RequestID length exceeds limit.")
+	}
```
bergundy: nit: you can validate only if this is set by the client.

chasm/lib/activity/frontend.go
```diff
@@ -132,6 +134,37 @@ func (h *frontendHandler) TerminateActivityExecution(
 	return &workflowservice.TerminateActivityExecutionResponse{}, nil
 }
 
+func (h *frontendHandler) RequestCancelActivityExecution(
```
bergundy, line 151: I notice that you didn't validate the `reason` field here. We need to cap the length of any string we store.

chasm/lib/activity/statemachine.go
```diff
@@ -280,11 +285,73 @@ var TransitionTerminated = chasm.NewTransition(
...
+	},
+	activitypb.ACTIVITY_EXECUTION_STATUS_CANCELED,
+	func(a *Activity, ctx chasm.MutableContext, details *commonpb.Payloads) error {
```
bergundy, line 306: You should accept the entire failure here and use the cancel reason as the message when we immediately cancel IMHO. I don't think encoding the details in a string is necessarily the right behavior but you're welcome to start a discussion how to best model this.

bergundy: Thinking about this a bit more, it might be best to have default message here and avoid exposing the cancel reason in the failure. That would be consistent with workflow activities AFAICT.

chasm/lib/activity/activity.go
fretz12: yea i prefer that style as well... let me get through all these PRs and a one time reformat on anything remaining as I believe Dan has applied it on PRs after this.

chasm/lib/activity/frontend.go
fretz12: I'm not seeing that validation for RequestCancelWorkflowExecutionRequest, but seems like a valid guard. Added validation + test.

chasm/lib/activity/frontend.go
fretz12: I moved setting the default req ID to below this check.

chasm/lib/activity/statemachine.go
fretz12: Help me understand... are you thinking something like this?
```
func(a *Activity, ctx chasm.MutableContext, failure *failurepb.Failure) error {
		store, err := a.Store.Get(ctx)
		if err != nil {
			return err
		}

		if store == nil {
			store = a
		}

		return store.RecordCompleted(ctx, func(ctx chasm.MutableContext) error {
			outcome, err := a.Outcome.Get(ctx)
			if err != nil {
				return err
			}

			if failure == nil {
				failure = &failurepb.Failure{
					Message: "Activity canceled by server",
					FailureInfo: &failurepb.Failure_CanceledFailureInfo{
						CanceledFailureInfo: &failurepb.CanceledFailureInfo{},
					},
				}
			}

			outcome.Variant = &activitypb.ActivityOutcome_Failed_{
				Failed: &activitypb.ActivityOutcome_Failed{
					Failure: failure,
				},
			}

			return nil
		})
	}
```

chasm/lib/activity/statemachine.go
```diff
@@ -280,11 +285,73 @@ var TransitionTerminated = chasm.NewTransition(
...
+			store = a
+		}
+
+		return store.RecordCompleted(ctx, func(ctx chasm.MutableContext) error {
+			outcome, err := a.Outcome.Get(ctx)
+			if err != nil {
+				return err
+			}
+
+			failure := &failurepb.Failure{
```
bergundy, line 322:
```suggestion
			failure := &failurepb.Failure{
				Message: "Activity canceled",
```

chasm/lib/activity/statemachine.go
bergundy: See my suggestion below.

fretz12: Accepted

PR #8689, opened 2025-11-24
Add alert for slow async versioning propagations

ShahabT:
## What changed?
WTTS

## Why?
Be able to alert on slow propagation to find eventual consistency bugs.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
None


service/worker/workerdeployment/activities.go
```diff
@@ -48,7 +46,7 @@ func (a *Activities) SyncUnversionedRamp(
 ) (*deploymentspb.SyncDeploymentVersionUserDataResponse, error) {
 	logger := activity.GetLogger(ctx)
 	// Get all the task queues in the current version and put them into SyncUserData format
-	currVersionInfo, _, err := a.deploymentClient.DescribeVersion(ctx, a.namespace, input.CurrentVersion, false)
+	currVersionInfo, _, err := a.WorkerDeploymentClient.DescribeVersion(ctx, a.namespace, input.CurrentVersion, false)
```
Shivs11, line 49: nit: is changing from deploymentClient to WorkerDeploymentClient a design change only? I looked around and I think we were repeating the declaration of this earlier on but I just wanna be sure.

service/worker/workerdeployment/version_activities.go
```diff
@@ -50,6 +53,16 @@ func (a *VersionActivities) SyncDeploymentVersionUserData(
 	input *deploymentspb.SyncDeploymentVersionUserDataRequest,
 ) (*deploymentspb.SyncDeploymentVersionUserDataResponse, error) {
 	logger := activity.GetLogger(ctx)
+	scheduledTime := activity.GetInfo(ctx).ScheduledTime
+
+	if scheduledTime.Add(SlowPropagationDelay).Before(time.Now()) {
+		a.Logger.Warn("Slow propagation detected, attempting to sync user data",
```
Shivs11: Should we be using logger, which is defined above, here? Wonder if these two are different fundamentally

common/metrics/metric_defs.go
```diff
@@ -1281,6 +1281,8 @@ var (
 	WorkerDeploymentVersionVisibilityQueryCount       = NewCounterDef("worker_deployment_version_visibility_query_count")
 	WorkerDeploymentVersioningOverrideCounter         = NewCounterDef("worker_deployment_versioning_override_count")
 	StartDeploymentTransitionCounter                  = NewCounterDef("start_deployment_transition_count")
+	VersioningDataPropagationLatency                  = NewTimerDef("versioning_data_propagation_latency")
+	SlowVersioningDataPropagationCounter              = NewCounterDef("slow_versioning_data_propagation")
```
Shivs11, line 1294: curious: do we wanna also have a counter to track the data propagation for those users using sync workflows?

The reason why I say this is because they could make things easier to debug/understand when they have an operator's hat on

Shivs11: If you agree to the above point I made, we need to rename the variables just fyi

service/worker/workerdeployment/version_activities.go
Shivs11: I think I see what you were trying to find here, but I took a minute/two to understand that this was actually finding out how long it took for the already scheduled activity to start.

Not sure if the name of SlowVersioningDataPropagationCounter makes sense if my above understanding is right. Something like ActivitySchedulingDelayCounter could make sense but shall leave the naming up to ya!

service/worker/workerdeployment/version_workflow.go
```diff
@@ -374,11 +374,17 @@ func (d *VersionWorkflowRunner) deleteVersionFromTaskQueuesAsync(ctx workflow.Co
 	workflow.Await(ctx, func() bool { return d.asyncPropagationsInProgress == 1 }) // delete itself is counted as one
 	d.cancelPropagations = false                                                   // need to unset this in case the version is revived
 
+	// Not counting the possible wait for previous propagations in this propagation latency.
+	startTime := workflow.Now(ctx)
+	defer func() {
+		d.metrics.Timer(metrics.VersioningDataPropagationLatency.Name()).Record(time.Since(startTime))
```
Shivs11: is using something like time.Since NDE prone?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -1044,7 +1055,11 @@ func (d *VersionWorkflowRunner) executeAndTrackAsyncPropagation(
 	}
 
 	if routingConfig != nil {
-		d.syncSummary(ctx)
+		if workflow.GetVersion(ctx, "no-propagation-sync-summary", workflow.DefaultVersion, 0) == workflow.DefaultVersion {
+			// TODO: clean this unnecessary sync up.
+			// No summary changes happen in async propagation that the deployment workflow
```
Shivs11: nit: unfinished comment i believe

common/metrics/metric_defs.go
ShahabT: but we're not gonna support sync wfs for long. once we verify async then we'd make that the default and eventually cleanup async path.

service/worker/workerdeployment/activities.go
ShahabT: It's just the name that was there in `activityDeps`. now that we embed that we don't need to have `deploymentClient`.

service/worker/workerdeployment/version_activities.go
ShahabT: So this scheduledTime is the first attempt schedule time, not the current attempt scheduling time. so basically if the activity has to retry for long it should fire.

service/worker/workerdeployment/version_workflow.go
ShahabT: For metrics purposes yes, but the linter is not happy about it so I need to change it.

service/worker/workerdeployment/version_activities.go
ShahabT: I think the logger in SDK adds activity context stuff like wf and activity names which is a good thing, changing to that.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -1003,6 +1004,11 @@ func (d *VersionWorkflowRunner) syncVersionDataToTaskQueues(ctx workflow.Context
 // syncTaskQueuesAsync must be called within the lock. It first increments the version revision number and then
 // starts async propagation of version data (and routing info if given) to all task queues.
 func (d *VersionWorkflowRunner) syncTaskQueuesAsync(ctx workflow.Context, routingConfig *deploymentpb.RoutingConfig, versionDataChanged bool) {
+	startTime := workflow.Now(ctx)
+	defer func() {
+		d.metrics.Timer(metrics.VersioningDataPropagationLatency.Name()).Record(workflow.Now(ctx).Sub(startTime))
+	}()
+
```
Shivs11, line 1008: This defer function is executed the second function execution completes. In the case of async mode, we would kickstart the go-routine that starts the async propagation and then immediately return from the function no????

so in theory, this metric is not calculating what you want it to?  or did I miss something?

Shivs11 (approved):
just one comment which may be valid - approving since rest all looks fine

service/worker/workerdeployment/version_workflow.go
ShahabT: good catch. fixed

PR #8693, opened 2025-11-24
CHASM: Propagate ArchetypeID

yycptt:
## What changed?
- Use archetypeID everywhere in history service and pass them to persistence 

## Why?
- Required to separate ID space work.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/proto/internal/temporal/server/api/replication/v1/message.proto
+++ b/proto/internal/temporal/server/api/replication/v1/message.proto
@@ -258,8 +258,10 @@ message VersionedTransitionArtifact {
 }
 
 message MigrationExecutionInfo {
-    // Should be called business_id, but to maintain backward compatibility
-    // with old type definition used in migration workflows.
+    // The proto json name of this field needs to be "workflowId",
+    // to be backward compatibility with commonpb.WorkflowExecution,
+    // which is what used to be used in migration workflow's activity
+    // input/output.
     string business_id = 1 [json_name = "workflowId"];
     string run_id = 2;
     // (-- api-linter: core::0141::forbidden-types=disabled --)
--- a/api/replication/v1/message.pb.go
+++ b/api/replication/v1/message.pb.go
@@ -2048,8 +2048,10 @@ func (*VersionedTransitionArtifact_SyncWorkflowStateSnapshotAttributes) isVersio
 
 type MigrationExecutionInfo struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
-	// Should be called business_id, but to maintain backward compatibility
-	// with old type definition used in migration workflows.
+	// The proto json name of this field needs to be "workflowId",
+	// to be backward compatibility with commonpb.WorkflowExecution,
+	// which is what used to be used in migration workflow's activity
+	// input/output.
 	BusinessId string `protobuf:"bytes,1,opt,name=business_id,json=workflowId,proto3" json:"business_id,omitempty"`
 	RunId      string `protobuf:"bytes,2,opt,name=run_id,json=runId,proto3" json:"run_id,omitempty"`
 	// (-- api-linter: core::0141::forbidden-types=disabled --)


service/history/ndc/workflow_state_replicator.go
```diff
@@ -1330,10 +1348,25 @@ func (r *WorkflowStateReplicatorImpl) backfillHistory(
 	if err != nil {
 		return err
 	}
+
+	archetypeID := mutableState.ChasmTree().ArchetypeID()
+	if archetypeID != chasm.WorkflowArchetypeID {
```
xwduan, line 1400: Will it have backward compatibility issue? i.e. for some old workflow which dose not have archetype? And do we need any logic to prevent other Archetype not call this func or return early?

common/persistence/execution_manager.go
```diff
@@ -1119,6 +1178,27 @@ func (m *executionManagerImpl) toWorkflowMutableState(internState *InternalWorkf
...
+		m.logger,
+		archetypeID != chasm.UnspecifiedArchetypeID,
+		"ArchetypeID not specified, defaulting to Workflow.",
+		tag.Operation(methodName),
+		tag.SysStackTrace(string(debug.Stack())),
+	) {
+		return chasm.WorkflowArchetypeID, true
+	}
+
+	if archetypeID == chasm.SchedulerArchetypeID {
```
bergundy, line 1196: Put a comment here that this is because schedulers have a temporary exception to allow them run in the same ID space as workflows.

proto/internal/temporal/server/api/replication/v1/message.proto
```diff
@@ -252,3 +256,12 @@ message VersionedTransitionArtifact {
     bool is_close_transfer_task_acked = 6;
     bool is_force_replication = 7;
 }
+
+message MigrationExecutionInfo {
+    // Should be called business_id, but to maintain backward compatibility
+    // with old type definition used in migration workflows.
+    string workflow_id = 1;
```
bergundy: Not sure sure why we can't call this business ID. Can you explain?

service/history/workflow/context.go
```diff
@@ -166,25 +174,21 @@ func (c *ContextImpl) LoadMutableState(ctx context.Context, shardContext history
...
-
-	if actualArchetype != "" && c.archetype != chasm.ArchetypeAny && c.archetype != actualArchetype {
+	actualArchetypeID := c.MutableState.ChasmTree().ArchetypeID()
+	if c.archetypeID != chasm.UnspecifiedArchetypeID && c.archetypeID != actualArchetypeID {
+		// TODO: convert to archetype name for better observability
 		c.logger.Warn("Potential ID conflict across different archetypes",
-			tag.Archetype(c.archetype),
-			tag.NewStringTag("actual-archetype", actualArchetype),
+			tag.ArchetypeID(c.archetypeID),
+			tag.NewUInt32("actual-archetype-id", actualArchetypeID),
```
bergundy: My concern with these logs is that they're going to be very hard to reason about. Ideally we would resolve the archetype names before logging and emitting other operational information such as metrics and internal errors.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -1330,10 +1348,25 @@ func (r *WorkflowStateReplicatorImpl) backfillHistory(
...
+		return softassert.UnexpectedInternalErr(
+			r.logger,
+			"Backfilling history for non-workflow archetype",
```
bergundy:
```suggestion
			"Backfilling history not supported for non-workflow archetype",
```

service/history/workflow/mutable_state_impl.go
```diff
@@ -890,8 +848,21 @@ func (ms *MutableStateImpl) SetHistoryTree(
...
 	// NOTE: Unfortunately execution timeout and run timeout are not yet initialized into ms.executionInfo at this point.
 	// TODO: Consider explicitly initializing mutable state with these timeout parameters instead of passing them in.
-
 	workflowKey := ms.GetWorkflowKey()
+
+	archetypeID := ms.ChasmTree().ArchetypeID()
+	if archetypeID != chasm.WorkflowArchetypeID {
+		return softassert.UnexpectedInternalErr(
+			ms.logger,
+			"Backfilling history for non-workflow archetype",
```
```suggestion
			"Backfilling history not supported for non-workflow archetype",
```

service/history/handler.go
```diff
@@ -1882,10 +1884,8 @@ func (h *Handler) RefreshWorkflowTasks(ctx context.Context, request *historyserv
 	err = engine.RefreshWorkflowTasks(
```
bergundy, line 1883: Should we rename this to refresh execution tasks if it is expected to support non-workflows?

service/worker/migration/activities.go
```diff
@@ -372,9 +373,20 @@ func (a *activities) generateWorkflowReplicationTask(
 
 	var stateTransitionCount, historyLength int64
 	if generateViaFrontend {
+		archetype, ok := a.chasmRegistry.ComponentFqnByID(execution.ArchetypeId)
+		if !ok {
+			err := fmt.Errorf("unknown archetypeID: %v", execution.ArchetypeId)
+			a.logger.Error("force-replication failed to translate archetypeID to name", tag.Error(err), tag.ArchetypeID(execution.ArchetypeId))
```
bergundy: Put the execution key in the log message to help debug this?

service/worker/migration/activities.go
```diff
@@ -474,9 +490,38 @@ func (a *activities) ListWorkflows(ctx context.Context, request *workflowservice
...
+			// but for backward compatibility reason we need this field to be 0
+			// to avoid unmarshaling errors since the previous type returned in
+			// ListWorkflowExecutions.Executions is commonpb.WorkflowExecution
+			// which does not have ArchetypeId field.
+			//
+			// TODO: switch to chasm.WorkflowArchetypeID in release 1.31.0
+			ArchetypeId: chasm.UnspecifiedArchetypeID,
+		}
+
+		nsDivisionPayload, ok := e.SearchAttributes.IndexedFields[sadefs.TemporalNamespaceDivision]
```
bergundy: Is this ensured to not be nil?

tests/testcore/flag.go
```diff
@@ -17,8 +17,8 @@ var cliFlags struct {
 }
 
 func init() {
-	flag.StringVar(&cliFlags.persistenceType, "persistenceType", "sql", "type of persistence - [nosql or sql]")
-	flag.StringVar(&cliFlags.persistenceDriver, "persistenceDriver", "sqlite", "driver of nosql/sql - [cassandra, mysql8, postgres12, sqlite]")
+	flag.StringVar(&cliFlags.persistenceType, "persistenceType", "nosql", "type of persistence - [nosql or sql]")
+	flag.StringVar(&cliFlags.persistenceDriver, "persistenceDriver", "cassandra", "driver of nosql/sql - [cassandra, mysql8, postgres12, sqlite]")
```
bergundy: This should be reverted.

service/history/ndc_standby_task_util.go
```diff
@@ -78,18 +79,32 @@ func standbyTimerTaskPostActionTaskDiscarded(
 	return consts.ErrTaskDiscarded
 }
 
-func isWorkflowExistOnSource(
+func isExecutionExistOnSource(
```
bergundy: Grammar fix while you're here:

```suggestion
func executionExistsOnSource(
```

service/history/transfer_queue_standby_task_executor.go
```diff
@@ -376,7 +375,7 @@ func (t *transferQueueStandbyTaskExecutor) processCancelExecution(
 			transferTask,
 			t.getCurrentTime,
 			t.config.StandbyTaskMissingEventsDiscardDelay(transferTask.GetType()),
-			t.checkWorkflowStillExistOnSourceBeforeDiscard,
+			t.checkExecutionStillExistOnSourceBeforeDiscard,
```
bergundy:
```suggestion
			t.checkExecutionStillExistsOnSourceBeforeDiscard,
```

tests/chasm_test.go
```diff
@@ -331,3 +331,5 @@ func (s *ChasmTestSuite) TestPayloadStoreVisibility() {
 }
 
 // TODO: More tests here...
+
+// TODO: add a test to make sure workflow can run successfully when chasm is enabled.
```
bergundy: That's technically already covered by the callback tests.

proto/internal/temporal/server/api/replication/v1/message.proto
yycptt: We used to return `commonpb.WorkflowExecution` in activity result, which is encoded by the default dataconverter via protojson, so can't directly change the field name...

I can do `string business_id = 1 [json_name = "workflowId"];` though, I believe that's still compatible. WDYT?

service/worker/migration/activities.go
yycptt: Updated to use GetIndexedFields() and performs nil check on the map.

service/worker/migration/activities.go
yycptt: Done.

service/history/handler.go
yycptt: Yeah we should, there are tons of renaming to be done. Don't want to touch that in this PR though.

service/history/workflow/context.go
yycptt: Updated the logic to do a best effort convert to archetype names.

proto/internal/temporal/server/api/replication/v1/message.proto
yycptt: I changed it to `string business_id = 1 [json_name = "workflowId"]`

common/persistence/execution_manager.go
yycptt: Done.

service/history/ndc/workflow_state_replicator.go
yycptt: > Will it have backward compatibility issue?

No there shouldn't be compatibility issue. An empty chasm tree will also return workflow archetypeID.

> do we need any logic to prevent other Archetype not call this func or return early?

There's a check early in the function to return error when version history is empty, which applies to all non workflow archetypes.

yycptt:
Also requires: https://github.com/temporalio/temporal/pull/8704

prathyushpv (approved):

PR #8692, opened 2025-11-24
Clean up Nexus CHASM completion code

bergundy:
## What changed?

- Move the CHASM / Nexus code back into the `chasm` package.
- Remove wrong and unused callback `Archetype` constant.
- Read the callback header using the case insensitive header util from the Nexus SDK.

## How did you test it?
- [ ] covered by existing tests

PR #8694, opened 2025-11-24
Hook up CHASM Scheduler to Frontend handler

lina-temporal:
## What changed?
- Frontend workflow_handler can now route to CHASM scheduler, depending on experiment flag/dynamic config
- Scheduler functional test has been updated to exercise both V1 and V2 scheduler
- Small fixes throughout scheduler codebase to address differences noticed during functional testing
- Nexus callback mechanisms updated to allow the CHASM internal callback URL

I plan to address the TODO around LastCompletionResult's assertion in a separate follow-up PR.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- Existing V1 tests are preserved, to show that the V1 codepath still works as-is. All new CHASM handling logic falls back to V1 in absence of a positive response. CHASM can only be opted in for a schedule during creation, controlled by dynamic config/experimental header.


chasm/lib/scheduler/scheduler.go
```diff
@@ -523,16 +535,72 @@ func (s *Scheduler) Describe(
...
+		return nil, ErrClosed
+	}
+
+	visibility := s.Visibility.Get(ctx)
+	memo := visibility.GetMemo(ctx)
+	delete(memo, visibilityMemoFieldInfo) // We don't need to return a redundant info block.
+
+	attributes := visibility.GetSearchAttributes(ctx)
+	delete(attributes, sadefs.TemporalNamespaceDivision)
+	delete(attributes, sadefs.TemporalSchedulePaused)
```
bergundy: These shouldn't come back from visibility... they are system fields, not user fields.
Are you seeing those come back?

chasm/lib/scheduler/scheduler.go
```diff
@@ -523,16 +535,72 @@ func (s *Scheduler) Describe(
...
+
+	if s.Schedule.Policies == nil {
```
bergundy: Why not call `setNullableFields` again here?

chasm/lib/scheduler/scheduler.go
```diff
@@ -523,16 +535,72 @@ func (s *Scheduler) Describe(
...
+		s.Schedule.Policies = &schedulepb.SchedulePolicies{}
+	}
+	if s.Schedule.GetPolicies().GetOverlapPolicy() == enumspb.SCHEDULE_OVERLAP_POLICY_UNSPECIFIED {
+		s.Schedule.Policies.OverlapPolicy = s.overlapPolicy()
+	}
+	if !s.Schedule.GetPolicies().GetCatchupWindow().IsValid() {
+		// TODO - this should be set from Tweakables.DefaultCatchupWindow.
```
bergundy, line 551: Why not pass this into the function?

chasm/lib/scheduler/scheduler.go
```diff
@@ -557,13 +625,32 @@ func (s *Scheduler) Update(
...
+		// attributes.
+
+		// Preserve the old custom memo in the new Visibility component.
+		oldVisibility := s.Visibility.Get(ctx)
+		oldMemo := oldVisibility.GetMemo(ctx)
+
+		visibility := chasm.NewVisibility(ctx)
+		s.Visibility = chasm.NewComponentField(ctx, visibility)
+		visibility.SetSearchAttributes(ctx, req.FrontendRequest.GetSearchAttributes().GetIndexedFields())
+		visibility.SetMemo(ctx, oldMemo)
```
bergundy: nit:


```suggestion
		visibility := chasm.NewVisibilityWithData(ctx, req.FrontendRequest.GetSearchAttributes().GetIndexedFields(), oldMemo)
		s.Visibility = chasm.NewComponentField(ctx, visibility)
```

chasm/scheduler.go
```diff
@@ -0,0 +1,15 @@
+package chasm
```
bergundy, line 1: Kinda hate that we have to put this here but this in not blocking the PR.

components/callbacks/chasm_invocation.go
```diff
@@ -44,7 +45,7 @@ func logInternalError(logger log.Logger, internalMsg string, internalErr error)
 
 func (c chasmInvocation) Invoke(ctx context.Context, ns *namespace.Namespace, e taskExecutor, task InvocationTask) invocationResult {
 	// Get back the base64-encoded ComponentRef from the header.
-	encodedRef, ok := c.nexus.GetHeader()[commonnexus.CallbackTokenHeader]
+	encodedRef, ok := c.nexus.GetHeader()[strings.ToLower(commonnexus.CallbackTokenHeader)]
```
bergundy: I am fixing this in https://github.com/temporalio/temporal/pull/8692 FYI.

service/worker/scheduler/fx.go
```diff
@@ -26,16 +28,15 @@ const (
...
-	)
+// VisibilityBaseListQuery will select schedules handled by both V1 scheduler and
+// V2 CHASM scheduler.
+var VisibilityBaseListQuery = fmt.Sprintf(
+	"%s IN ('%s', '%s') AND %s = '%s'",
+	sadefs.TemporalNamespaceDivision,
+	NamespaceDivision,
+	strconv.FormatUint(uint64(chasm.SchedulerArchetypeID), 10),
+	sadefs.ExecutionStatus,
+	enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING.String(),
```
bergundy: nit:

```suggestion
	"%s IN ('%s', '%d') AND %s = '%s'",
	sadefs.TemporalNamespaceDivision,
	NamespaceDivision,
	chasm.SchedulerArchetypeID,
	sadefs.ExecutionStatus,
	enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING.String(),
```

tests/schedule_test.go
```diff
@@ -364,8 +395,8 @@ func (s *ScheduleFunctionalSuite) TestBasics() {
 
 	// wait for one new run
 	s.Eventually(
-		func() bool { return atomic.LoadInt32(&runs2) == 1 },
-		7*time.Second,
+		func() bool { return atomic.LoadInt32(&runs2) >= 1 },
+		10*time.Second,
```
bergundy: Wondering why this had to change. Did the semantics change?

tests/schedule_test.go
```diff
@@ -870,7 +922,15 @@ func (s *ScheduleFunctionalSuite) TestRateLimit() {
...
 }
 
-func (s *ScheduleFunctionalSuite) TestListSchedulesReturnsWorkflowStatus() {
+func (s *ScheduleV1FunctionalSuite) TestListSchedulesReturnsWorkflowStatus_V1() {
+	s.testListSchedulesReturnsWorkflowStatus()
+}
+
+func (s *ScheduleCHASMFunctionalSuite) TestListSchedulesReturnsWorkflowStatus_CHASM() {
+	s.testListSchedulesReturnsWorkflowStatus()
+}
```
bergundy: Why do you need the duplicate test definition here?

tests/schedule_test.go
```diff
@@ -91,12 +111,19 @@ func (s *ScheduleFunctionalSuite) TearDownTest() {
...
 }
 
-func (s *ScheduleFunctionalSuite) TestBasics() {
+func (s *ScheduleCHASMFunctionalSuite) TestBasics_V1() {
+	s.testBasics()
+}
+
+func (s *ScheduleV1FunctionalSuite) TestBasics_CHASM() {
+	s.testBasics()
+}
```
bergundy: Not sure why you split this up. Why can't there be just a single `TestBasics()` method? I see this duplication in other test functions.

tests/schedule_test.go
```diff
@@ -659,14 +710,15 @@ func (s *ScheduleFunctionalSuite) TestLastCompletionAndError() {
...
+	ctx := s.newContext()
+	_, err := s.FrontendClient().CreateSchedule(ctx, req)
 	s.NoError(err)
 	s.cleanup(sid)
 
 	s.Eventually(func() bool { return atomic.LoadInt32(&testComplete) == 1 }, 20*time.Second, 200*time.Millisecond)
 }
 
-func (s *ScheduleFunctionalSuite) TestRefresh() {
+func (s *ScheduleV1FunctionalSuite) TestRefresh() {
```
bergundy, line 716: Is this intentionally only run on "V1"? If so please add a comment in the code.
Same for the other methods here.

tests/schedule_test.go
```diff
@@ -1169,3 +1234,14 @@ func (s *ScheduleFunctionalSuite) cleanup(sid string) {
...
+
+func (s *scheduleFunctionalSuiteBase) newContext() context.Context {
+	return testcore.NewContext()
+}
+
+func (s *ScheduleCHASMFunctionalSuite) newContext() context.Context {
+	baseCtx := testcore.NewContext()
+	return metadata.NewOutgoingContext(baseCtx, metadata.Pairs(
+		headers.ExperimentHeaderName, "chasm-scheduler",
+	))
```
bergundy: There should be a way to run the scheduler just with overriding the dynamic config without these specific headers, no?

service/frontend/workflow_handler.go
```diff
@@ -3045,57 +3049,67 @@ func (wh *WorkflowHandler) ListTaskQueuePartitions(ctx context.Context, request
...
 
-	workflowID := scheduler.WorkflowIDPrefix + request.ScheduleId
-
-	if err := wh.validateWorkflowID(workflowID); err != nil {
-		return nil, err
+	// Validate blob size limit here. In the V1 codepath, this is done automatically
+	// as part of StartWorkflowExecution, but here it must be done separately (to
+	// maintain equal payload size limits between versions).
+	switch action := request.GetSchedule().GetAction().GetAction().(type) {
+	case *schedulepb.ScheduleAction_StartWorkflow:
```
bergundy, line 3073: I think we should also verify the memo and header size. Those are also payloads.

service/frontend/workflow_handler.go
```diff
@@ -3593,6 +3665,32 @@ func (wh *WorkflowHandler) UpdateWorkerDeploymentVersionMetadata(ctx context.Con
 func (wh *WorkflowHandler) DescribeSchedule(ctx context.Context, request *workflowservice.DescribeScheduleRequest) (_ *workflowservice.DescribeScheduleResponse, retError error) {
 	defer log.CapturePanic(wh.logger, &retError)
 
+	// Prefer CHASM scheduler if enabled.
```
bergundy: Seems like we should first check if CHASM scheduler is enabled to avoid every call to the schedule going through this new code path, especially before this code has been thoroughly tested.

bergundy: We should also move the check if schedules are enabled to the top of this function (or any other function that handles schedules). Users may have disabled the feature completely for some arbitrary reason.

service/frontend/workflow_handler.go
```diff
@@ -3789,14 +3887,45 @@ func (wh *WorkflowHandler) UpdateSchedule(
...
+		return nil, errSchedulesNotAllowed
+	}
+
 	if request == nil {
 		return nil, errRequestNotSet
 	}
 
-	if !wh.config.EnableSchedules(request.Namespace) {
-		return nil, errSchedulesNotAllowed
+	res, err := wh.updateScheduleCHASM(ctx, request)
```
bergundy: What about handling not found errors here?

bergundy: What about request validation? (and make sure that you are not unaliasing search attributes on the CHASM path)

service/frontend/workflow_handler.go
```diff
@@ -3910,6 +4035,24 @@ func (wh *WorkflowHandler) PatchSchedule(ctx context.Context, request *workflows
 		trigger.ScheduledTime = timestamppb.Now()
 	}
 
+	res, err := wh.patchScheduleCHASM(ctx, request)
```
bergundy: Missing validation and not found error checks.
Also don't issue these RPCs unless CHASM scheduler is enabled.

chasm/lib/scheduler/scheduler.go
lina-temporal: No, you're right, I'd just assumed they would. Removed that.

chasm/lib/scheduler/scheduler.go
lina-temporal: I can actually just remove this from Describe, since I call `setNullableFields` on both the create and update paths, fixed.

components/callbacks/chasm_invocation.go
lina-temporal: Thanks, I'll merge main and get rid of this.

tests/schedule_test.go
lina-temporal: I think I probably inadvertently changed it when I was debugging. Reverted to how it was.

tests/schedule_test.go
lina-temporal: I was doing this so that it showed up a little bit more cleanly in test input (`TestListSchedulesReturnsWorkflowStatus_<version>` vs `TestBasics` and `TestBasics#01`), but I suppose it's unnecessary. Dropped em.

tests/schedule_test.go
lina-temporal: Correct, yeah. Added.

tests/schedule_test.go
lina-temporal: Fixed.

service/frontend/workflow_handler.go
lina-temporal: Sure, will fix both.

service/frontend/workflow_handler.go
lina-temporal: Fixed on both. Added handling for not found in the other handlers.

service/frontend/workflow_handler.go
lina-temporal: Fixed.

tests/schedule_test.go
lina-temporal: Yeah, you can just flip it on for the namespace. The tests could really take either approach.

service/frontend/workflow_handler.go
lina-temporal: Added validation for memo. I'm not sure which header size you're referring to though? Nexus callback headers? I think if users had a way to pass those, that'd be handled in the preexisting WF start args validation function (and they aren't).

bergundy: StartWorkflowExecutionRequest has a header too but I think we never validate header sizes. That should be fixed across the codebase.

chasm/lib/scheduler/scheduler.go
bergundy: @lina-temporal WDYT?

lina-temporal: Pass `Tweakables` into `Describe`? I guess it could be done that way, it'd be asymmetric from all of the other handlers, however.

chasm/lib/scheduler/scheduler.go
```diff
@@ -525,16 +536,65 @@ func (s *Scheduler) Describe(
...
+			ConflictToken:    s.generateConflictToken(),
+			Memo:             &commonpb.Memo{Fields: memo},
+			SearchAttributes: &commonpb.SearchAttributes{IndexedFields: visibility.GetSearchAttributes(ctx)},
 		},
 	}, nil
 }
 
+// cleanSpec sets default values in ranges for the DescribeSchedule response.
+func cleanSpec(spec *schedulepb.ScheduleSpec) {
+	cleanRanges := func(ranges []*schedulepb.Range) {
```
bergundy, line 571: nit: this could be a top level function.

tests/schedule_test.go
```diff
@@ -649,7 +673,10 @@ func (s *ScheduleFunctionalSuite) TestLastCompletionAndError() {
 			s.Equal("this one succeeds", lcr)
 			return "", errors.New("this one fails")
 		case 3:
-			s.Equal("this one succeeds", lcr)
+			// TODO - CHASM scheduler only keeps a single one of these set at a time, not both. IMO, that's more correct than
```
bergundy, line 694: This is fixed in the next PR right?

tests/schedule_test.go
```diff
@@ -174,8 +193,9 @@ func (s *ScheduleFunctionalSuite) TestBasics() {
 
 	// create
 
+	ctx := s.newContext()
```
bergundy, line 214: This won't work... Go won't invoke the method of the subclass on a method of the base class.

You'll need to pass in a function pointer on instantiation or set a different dynamic configs per class on setup.

lina-temporal: Fixed, this happened in the last requested refactor for the suite.

PR #8702, opened 2025-11-25
Fix handling of context errors

awln-temporal:
## What changed?
Context cancelations are getting incorrectly translated to `Unavailable` service errors, which lead to Visibility Availability Alert false positives. 

This code change adds branches to check context error types before casting to the resulting `serviceerror` for both ElasticSearch and SQL visibility stores.

## Why?
Separate client and service side errors.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/visibility/store/sql/visibility_store.go
+++ b/common/persistence/visibility/store/sql/visibility_store.go
@@ -75,9 +75,9 @@ func (s *VisibilityStore) GetName() string {
 
 func convertSQLError(message string, err error) error {
 	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-		return fmt.Errorf("%s: %w", operation, err)
+		return fmt.Errorf("%s: %w", message, err)
 	}
-	return serviceerror.NewUnavailable(fmt.Sprintf("%s: %v", operation, err))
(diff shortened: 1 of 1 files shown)


common/persistence/visibility/store/sql/visibility_store.go
```diff
@@ -73,6 +73,13 @@ func (s *VisibilityStore) GetName() string {
 	return s.sqlStore.GetName()
 }
 
+func convertSQLError(operation string, err error) error {
```
rodrigozhou: nit
```suggestion
func convertSQLError(message string, err error) error {
```

common/persistence/visibility/store/sql/visibility_store.go
```diff
@@ -160,7 +167,7 @@ func (s *VisibilityStore) DeleteWorkflowExecution(
 		RunID:       request.RunID,
 	})
 	if err != nil {
-		return serviceerror.NewUnavailable(err.Error())
+		return convertSQLError("DeleteWorkflowExecution operation failed. Delete failed", err)
```
rodrigozhou: This looks redundant. ditto other mesages.
```suggestion
		return convertSQLError("DeleteWorkflowExecution operation failed.", err)
```

rodrigozhou (approved):
Please address the comments before merging.

PR #8705, opened 2025-11-26
CHASM: Use ParentPtr in scheduler and callback

yycptt:
## What changed?
- Use ParentPtr in scheduler and callback

## Why?
- Avoid the overhead of a persisted pointer

## How did you test it?
- [x] built
- [ ] added new unit test(s)


bergundy:
We really need a better way to test using the full engine but that's already known.

PR #8706, opened 2025-11-26
Reject TaskQueue's with empty normalName + Refactorings

Shivs11:
## What changed?
- WISOTT
- [Context on the re-factorings]: The main motivation for the refactoring was to have the worker_deployment package as a subset to the worker_versioning package. This would also mean that we would keep things that are specific to the worker-deployment package in it's own package (ofc, as much as possible)

## Why?
- Versioning correctness. Also, an empty task-queue should never be part of a deployment version.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None

---

> [!NOTE]
> Validates deployment options and enforces NormalName on versioned sticky polls, refactors worker-deployment ID utilities and validations, updates constants, and adds tests; minor ES copy and naming cleanups.
> 
> - **Frontend**:
>   - Validate `DeploymentOptions` on polls/WT completions/Nexus; require non-empty `TaskQueue.NormalName` for sticky polls when versioned or `DeploymentOptions` provided.
>   - Add `errDeploymentOptionsNotSet`; wire `DeploymentOptions` through handlers; new `validateDeploymentOptions` helper.
> - **Worker Versioning**:
>   - Rename/standardize delimiters and constants; add `ValidateDeploymentVersion(version, maxIDLengthLimit)` and `ValidateDeploymentVersionFields`.
>   - Remove ID generator helpers from this package; keep string conversion helpers; minor copy simplifications.
> - **Worker Deployment (service)**:
>   - Move ID helpers (`GenerateDeploymentWorkflowID`, `GenerateVersionWorkflowID`, `GetDeploymentNameFromWorkflowID`) to `workerdeployment/util` and update all call sites.
>   - Use new validation helpers/field name constants; error handling tweaks; small refactors.
> - **Elasticsearch Visibility**:
>   - Replace manual slice loops with `copy` for sorters and group value assembly.
> - **Tests**:
>   - Add functional test verifying versioned sticky pollers fail without `NormalName`.
>   - Update tests and utilities to new helpers/constant names and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c61e6c4458ad4c515c46086bf8992b09a7077437. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -5288,10 +5288,7 @@ func (wh *WorkflowHandler) validateDeploymentOptions(deploymentOptions *deployme
 	if deploymentOptions == nil {
 		return nil
 	}
-	if deploymentOptions.GetDeploymentName() != "" && deploymentOptions.GetBuildId() == "" {
-		return errDeploymentOptionsNotSet
-	}
-	if deploymentOptions.GetDeploymentName() == "" && deploymentOptions.GetBuildId() != "" {
+	if deploymentOptions.GetDeploymentName() == "" || deploymentOptions.GetBuildId() == "" {
 		return errDeploymentOptionsNotSet
 	}
 	deploymentVersion := &deploymentspb.WorkerDeploymentVersion{


service/frontend/workflow_handler.go
```diff
@@ -5258,11 +5259,17 @@ type buildIdAndFlag interface {
...
+	GetBuildId() string
+	GetWorkerVersioningMode() enumspb.WorkerVersioningMode
+}
+
+func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFlag, buildAndVersioningMode buildAndVersioningMode, tq *taskqueuepb.TaskQueue) error {
 	if id.GetUseVersioning() && !wh.config.EnableWorkerVersioningWorkflow(nsName) {
 		return errWorkerVersioningWorkflowAPIsNotAllowed
 	}
-	if id.GetUseVersioning() && tq.GetKind() == enumspb.TASK_QUEUE_KIND_STICKY && len(tq.GetNormalName()) == 0 {
+	if (id.GetUseVersioning() || buildAndVersioningMode.GetWorkerVersioningMode() == enumspb.WORKER_VERSIONING_MODE_VERSIONED) && tq.GetKind() == enumspb.TASK_QUEUE_KIND_STICKY && len(tq.GetNormalName()) == 0 {
```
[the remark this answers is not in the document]
Shivs11: The go-generated code, by protobuf, handles this!

service/frontend/workflow_handler.go
```diff
@@ -4908,15 +4909,15 @@ func (wh *WorkflowHandler) PollNexusTaskQueue(ctx context.Context, request *work
 	}
 
 	//nolint:staticcheck // SA1019: worker versioning v0.31
-	if err := wh.validateVersioningInfo(request.Namespace, request.WorkerVersionCapabilities, request.TaskQueue); err != nil {
+	if err := wh.validateVersioningInfo(request.Namespace, request.WorkerVersionCapabilities, request.DeploymentOptions, request.TaskQueue); err != nil {
```
Shivs11, line 4916: this made me wonder: did we ever talk about how nexus pollers would interact with versioning? I haven't included a nexus poller in my test case at the bottom only because I don't recall us talking about nexus + versioning. However, included the code here since a nexus poller uses the same helper function.

cc - @ShahabT

ShahabT: yeah we have functional tests for nexus in versioning_3_test

service/frontend/workflow_handler.go
```diff
@@ -5258,11 +5259,16 @@ type buildIdAndFlag interface {
 	GetUseVersioning() bool
 }
 
-func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFlag, tq *taskqueuepb.TaskQueue) error {
+type buildAndVersioningMode interface {
```
ShahabT: no need for this because unlike the buildIdAndFlag there is no more than one type (DeploymentOptions) that we want to cover here.

service/frontend/workflow_handler.go
```diff
@@ -5258,11 +5259,16 @@ type buildIdAndFlag interface {
...
+	GetBuildId() string
+	GetWorkerVersioningMode() enumspb.WorkerVersioningMode
+}
+
+func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFlag, buildAndVersioningMode buildAndVersioningMode, tq *taskqueuepb.TaskQueue) error {
```
ShahabT:
```suggestion
func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFlag, deploymentOptions *DeploymentOptions, tq *taskqueuepb.TaskQueue) error {
```

service/frontend/workflow_handler.go
ShahabT, line 5282: let's improve these checks to cover versioning 3 two. In case of versioning 3, we need to also ensure deployment name existence and length

common/worker_versioning/worker_versioning.go
```diff
@@ -969,24 +1010,3 @@ func WorkerDeploymentVersionFromStringV32(s string) (*deploymentspb.WorkerDeploy
...
-
-// GenerateVersionWorkflowID is a helper that generates a system accepted
-// workflowID which are used in our Worker Deployment Version workflows
-func GenerateVersionWorkflowID(deploymentName string, buildID string) string {
-	versionString := ExternalWorkerDeploymentVersionToString(&deploymentpb.WorkerDeploymentVersion{
-		DeploymentName: deploymentName,
-		BuildId:        buildID,
-	})
-	return WorkerDeploymentVersionWorkflowIDPrefix + WorkerDeploymentVersionWorkflowIDDelimeter + versionString
-}
```
Shivs11, line 992: I think these functions are quite specific to the worker-deployment(version) entities. IMHO, they should be moved from the worker-versioning package and over to the worker_deployment one.

service/frontend/workflow_handler.go
```diff
@@ -5271,7 +5274,44 @@ func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFl
...
+func validateDeploymentVersionMode(
+	versioningMode enumspb.WorkerVersioningMode,
+	deploymentVersion *deploymentspb.WorkerDeploymentVersion,
+	tq *taskqueuepb.TaskQueue,
+) error {
+	switch versioningMode {
+	case enumspb.WORKER_VERSIONING_MODE_UNVERSIONED:
+		if deploymentVersion != nil {
+			return serviceerror.NewInvalidArgument("deployment version cannot be set for unversioned worker")
+		}
```
Shivs11: @ShahabT - this is something that I was not totally sure about. Should we be rejecting unversioned pollers that have a deployment version set?

I think right now, it just no-ops but I think being conservative could help us enforce our requirements on the user even more?

common/worker_versioning/worker_versioning.go
```diff
@@ -429,27 +433,64 @@ func ValidateDeployment(deployment *deploymentpb.Deployment) error {
...
-		return serviceerror.NewInvalidArgument("deployment name cannot be empty")
+
+	// Validate deployment name
+	err := ValidateDeploymentVersionFields(WorkerDeploymentNameFieldName, version.GetDeploymentName(), WorkerDeploymentVersionWorkflowIDInitialSize)
+	if err != nil {
+		return err
+	}
+
+	// Validate build ID
+	err = ValidateDeploymentVersionFields(WorkerDeploymentBuildIDFieldName, version.GetBuildId(), WorkerDeploymentVersionWorkflowIDInitialSize)
```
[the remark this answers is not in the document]
Shivs11: ah, good catch - fixed it.

service/frontend/workflow_handler.go
```diff
@@ -5271,7 +5274,44 @@ func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFl
...
+	deploymentVersion := worker_versioning.DeploymentVersionFromOptions(deploymentOptions)
+	versioningMode := deploymentOptions.GetWorkerVersioningMode()
+
+	// Validate deployment version based on the versioning mode
+	err := validateDeploymentVersionMode(versioningMode, deploymentVersion, tq)
+	if err != nil {
+		return err
+	}
+
+	return worker_versioning.ValidateDeploymentVersion(deploymentVersion)
```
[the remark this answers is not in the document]
Shivs11: No; Un-versioned pollers polling any type of task-queue (with or without using versioning mode) should not have a deploymentVersion set and with my changes, it would simply return nil when the pre-requisite that I just described is true. Sure, if they were to set it to a deployment version and have the worker to be unversioned, it throws an error but I wonder if we want this behavior.

Now, in the case of older SDK's, the versioning mode would be set to Unspecified which would make the control loop fall in the default case and return a nil for the error.

common/worker_versioning/worker_versioning.go
```diff
@@ -429,27 +426,70 @@ func ValidateDeployment(deployment *deploymentpb.Deployment) error {
...
+
+	// Length of each field should be: (MaxIDLengthLimit - (prefix + delimeter length)) / 2
+	if fieldName == WorkerDeploymentNameFieldName {
+		if len(field) > (maxIDLengthLimit-WorkerDeploymentWorkerDeploymentIDInitialSize)/2 {
+			return serviceerror.NewInvalidArgumentf("size of %v larger than the maximum allowed", fieldName)
+		}
+	} else if fieldName == WorkerDeploymentBuildIDFieldName {
+		if len(field) > (maxIDLengthLimit-WorkerDeploymentVersionWorkflowIDInitialSize)/2 {
+			return serviceerror.NewInvalidArgumentf("size of %v larger than the maximum allowed", fieldName)
+		}
```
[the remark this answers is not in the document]
Shivs11: oh, thanks!

service/frontend/workflow_handler.go
```diff
@@ -5271,9 +5277,49 @@ func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFl
...
+			DeploymentName: deploymentOptions.GetDeploymentName(),
+			BuildId:        deploymentOptions.GetBuildId(),
+		}
+	}
+
+	// Validation checks
+	switch versioningMode {
+	case enumspb.WORKER_VERSIONING_MODE_UNVERSIONED:
+		// if deploymentVersion != nil {
+		// 	return serviceerror.NewInvalidArgument("deployment version cannot be set for unversioned worker")
```
Shivs11: @ShahabT - question for ya; do we wanna reject all those polls that have unversioned as the mode but have also passed in a deployment version and name?

I think we should, we don't right now though

ShahabT: IMO, if the deployment options is passed, regardless of mode the deployment name and build ID should be present. right now we don't show those deployments in the UI but maybe later we do. still user can benefit from the values in wf history events even if versioning is disabled.

Also, I'm thinking switching the useVersioning flag from worker side should work smoothly and just work. i.e. we don't want to force users to comment out deployment name and build id if they sent useVersioning from true to false.

Shivs11: > IMO, if the deployment options is passed, regardless of mode the deployment name and build ID should be present. right now we don't show those deployments in the UI but maybe later we do. still user can benefit from the values in wf history events even if versioning is disabled.

Slightly confused though - why would someone pass in deploymentOptions for an unversioned worker? Didn't we state that unversioned workers don't really belong to one specific worker-deployment?

ShahabT: there is still benefits:
- be able to reset by build id
- seeing build id in the event history to know which build the wf/activity ran on.
- in the future, user can get scaling decisions at deployment level rather than TQ. for versioned deployments it'll be at deployment _version_ level. for unversioned, deployment level.
- in the future we might add those deployments in the deployment page as Rolling deploys (as opposed to Rainbow)

Shivs11: thanks, convinced that users could pass in unversioned mode with deploymentOptions (although it does seem kinda non-intuitive to me)

common/persistence/visibility/store/elasticsearch/visibility_store.go
```diff
@@ -1052,9 +1050,7 @@ func (s *VisibilityStore) parseCountGroupByResponse(
 				return fmt.Errorf("unable to parse 'doc_count' field: %w", err)
 			}
 			groupValues := make([]*commonpb.Payload, len(groupByFields))
-			for i := range bucketValues {
-				groupValues[i] = bucketValues[i]
-			}
+			copy(groupValues, bucketValues)
```
Shivs11, line 1053: this was suggested to me by the lint; Not having this was failing the lint check for some weird reason (weird, ik)

however, I do think this is right so keeping this here

service/frontend/workflow_handler.go
```diff
@@ -5271,9 +5277,44 @@ func (wh *WorkflowHandler) validateVersioningInfo(nsName string, id buildIdAndFl
...
+	versioningMode enumspb.WorkerVersioningMode,
+	deploymentOptions *deploymentpb.WorkerDeploymentOptions,
+	tq *taskqueuepb.TaskQueue,
+) error {
+	var deploymentVersion *deploymentspb.WorkerDeploymentVersion
+	if deploymentOptions.GetDeploymentName() != "" || deploymentOptions.GetBuildId() != "" {
+		deploymentVersion = &deploymentspb.WorkerDeploymentVersion{
+			DeploymentName: deploymentOptions.GetDeploymentName(),
+			BuildId:        deploymentOptions.GetBuildId(),
+		}
```
Shivs11: have an extra validation which should not allow options to be set when only one of buildID and deployment set.

Shivs11: might cause things to break in SDK.

service/frontend/workflow_handler.go
```diff
@@ -5258,20 +5263,42 @@ type buildIdAndFlag interface {
...
+func (wh *WorkflowHandler) validateDeploymentOptions(deploymentOptions *deploymentpb.WorkerDeploymentOptions) error {
+	if deploymentOptions == nil {
+		return nil
+	}
+	if deploymentOptions.GetDeploymentName() != "" && deploymentOptions.GetBuildId() == "" {
+		return errDeploymentOptionsNotSet
+	}
+	if deploymentOptions.GetDeploymentName() == "" && deploymentOptions.GetBuildId() != "" {
+		return errDeploymentOptionsNotSet
+	}
```
ShahabT:
```suggestion
	if deploymentOptions.GetBuildId() == "" || deploymentOptions.GetDeploymentName() == "" {
			return errDeploymentOptionsNotSet
	}
```

PR #8707, opened 2025-11-26
Change DeserializeComponentRef to return error on empty input

dandavison:
Currently, `DeserializeComponentRef` given empty input yields a zero-value `ComponentRef`. Although this matches the behavior of proto `Unmarshal`, we don't want to support empty serialized refs. A zero-value `ComponentRef` compares less than other values, but it's otherwise mostly invalid (and can't be serialized).

---

> [!NOTE]
> Make `DeserializeComponentRef` error on empty input and add tests for nil/empty cases.
> 
> - **chasm**:
>   - **`DeserializeComponentRef`**: Return error on empty input (`len(data) == 0`); add `errors` import.
>   - **Tests**: Update `TestSerializeDeserialize` to assert errors for `nil` and empty byte slices; retain serialization roundtrip checks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f968d826281c1a67845236498b40d946f6b6e5fd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chasm/ref.go
+++ b/chasm/ref.go
@@ -1,6 +1,7 @@
 package chasm
 
 import (
+	"errors"
 	"reflect"
 
 	"go.temporal.io/api/serviceerror"
@@ -129,7 +130,7 @@ func (r *ComponentRef) Serialize(
 // Provides caller the access to information including ExecutionKey, Archetype, and ShardingKey.
 func DeserializeComponentRef(data []byte) (ComponentRef, error) {
 	if len(data) == 0 {
-		return ComponentRef{}, serviceerror.NewInternal("empty chasm component ref")
+		return ComponentRef{}, errors.New("empty chasm component ref")
 	}
 	var pRef persistencespb.ChasmComponentRef
 	if err := pRef.Unmarshal(data); err != nil {


chasm/ref.go
```diff
@@ -128,6 +128,9 @@ func (r *ComponentRef) Serialize(
 // DeserializeComponentRef deserializes a byte slice into a ComponentRef.
 // Provides caller the access to information including ExecutionKey, Archetype, and ShardingKey.
 func DeserializeComponentRef(data []byte) (ComponentRef, error) {
+	if len(data) == 0 {
+		return ComponentRef{}, serviceerror.NewInternal("empty chasm component ref")
```
yycptt: if we are using service error, let's switch to InvalidArgument.

chasm/ref.go
```diff
@@ -128,6 +129,9 @@ func (r *ComponentRef) Serialize(
 // DeserializeComponentRef deserializes a byte slice into a ComponentRef.
 // Provides caller the access to information including ExecutionKey, Archetype, and ShardingKey.
 func DeserializeComponentRef(data []byte) (ComponentRef, error) {
+	if len(data) == 0 {
+		return ComponentRef{}, errors.New("empty chasm component ref")
```
yycptt, line 133: This will cause unnecessary retries so I can't approve it.

PR #8662, opened 2025-11-18
Add implementation of CHASM List/Count Runs

awln-temporal:
## What changed?
Add implementation of CHASM List/Count Runs. Given a CHASM archetype and supported MemoType, a Visibility Query can be ran against both custom and CHASM search attributes. Both operations wrap around the Visibility Manager methods to List/Count Workflows. 

In the initial handler of the query in visibility_store.go, the query converter is passed the chasm VisibilitySearchAttributesMapper and the archetypeID. When resolving query parameters, the order of resolving aliases to field names is Custom -> CHASM -> System/Predefined attribute.

## Why?
Required to support CHASM Visibility and Component authors.

## How did you test it?
- [X] built
- [X] run locally and tested manually
- [X] covered by existing tests
- [X] added new unit test(s)
- [x] added new functional test(s)

PR #8715, opened 2025-11-26
Emit error log if deregistered search attribute in chasm vis task

awln-temporal:
## What changed?
Emit warn log if deregistered search attribute in chasm vis task. 

Original thread: https://github.com/temporalio/temporal/pull/8662#discussion_r2561591969

## Current State
If user deregisters search attributes while the task is getting processed, current code leads to task failure.

In ES, if the search attribute has been deregistered, upsert SA task execution would still fail when generating the ES Doc and attempting to decode the value. 

In the SQL implementation, since these fields are constant and always attached to the NameTypeMap/ClusterMetadata, Upserting these fields would succeed, so there is a discrepancy between these stores.

## Why?

Once we have parity between ES and SQL for preallocated columns, we should not throw an error if a search attribute gets deregistered when executing the CHASM vis task and just do a best effort upsert call. User deregistering search attributes after archetype execution is a valid use case.

A valid concern is if the visibility component somehow has an incorrect search attribute (eg. field name). There was an issue the past caught because errors were thrown during translation, which showed search attribute already translated to field name before hitting the translation layer (due to application code bug). However, Error logs should be sufficient to debug.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/visibility_queue_task_executor.go
+++ b/service/history/visibility_queue_task_executor.go
@@ -410,9 +410,9 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
 	for alias, value := range aliasedSearchAttributes {
 		fieldName, err := searchAttributesMapper.GetFieldName(alias, namespaceEntry.Name().String())
 		if err != nil {
-			// INFO: To reach here, either the search attribute has been deregistered before task execution, which is valid behavior,
-			// or the upserted search attribute is incorrectly mapped, indicating possible bug in application code.
-			t.logger.Error("Failed to get field name for alias, ignoring search attribute", tag.NewStringTag("alias", alias), tag.Error(err))
+			// To reach here, either the search attribute has been deregistered before task execution, which is valid behavior,
+			// or there are delays in propagating search attribute mappings to History.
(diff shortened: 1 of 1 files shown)


rodrigozhou (approved):

service/history/visibility_queue_task_executor.go
```diff
@@ -409,7 +410,10 @@ func (t *visibilityQueueTaskExecutor) processChasmTask(
 	for alias, value := range aliasedSearchAttributes {
 		fieldName, err := searchAttributesMapper.GetFieldName(alias, namespaceEntry.Name().String())
 		if err != nil {
-			return err
+			// INFO: To reach here, either the search attribute has been deregistered before task execution, which is valid behavior,
+			// or the upserted search attribute is incorrectly mapped, indicating possible bug in application code.
```
bergundy: WDYM incorrectly mapped? Do you mean that there's a bug in the frontend code?

There's another cause for this that we discussed earlier. The search attribute mapping may have propagation delays to the history node executing this task.
We already ensure that all search attributes are mappable on the frontend.

awln-temporal: in the case of eager workflows, following executions bypassed frontend validators. error message is more specific now.

PR #8716, opened 2025-11-26
Forbid use of time.Now() in chasm libraries

bergundy:
## What changed?

Use forbidigo to enforce `time.Now()` is not used in CHASM code.

## Why?

We want to support pause and time skipping as part of the CHASM framework, which requires that the framework provides the time.

### DIFF
--- a/.github/.golangci.yml
+++ b/.github/.golangci.yml
@@ -36,6 +36,8 @@ linters:
           msg: "Please use require.Eventually or assert.Eventually instead unless you've no other option"
         - pattern: panic
           msg: "Please avoid using panic in application code"
+        - pattern: time\.Now
+          msg: "Using time.Now is not allowed in chasm/lib package (non-test files), use ctx.Now(component) instead"
     depguard:
       rules:
         main:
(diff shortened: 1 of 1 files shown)


common/rpc/test/http_test.go
```diff
@@ -16,6 +17,7 @@ import (
 )
 
 func TestCreateLocalFrontendHTTPClient_UsingMembership(t *testing.T) {
+	_ = time.Now()
```
stephanos: Was this just a test to see it works?

stephanos (approved):

common/rpc/test/http_test.go
bergundy: Lesson: don't use `git commit -a`


```suggestion
```

common/rpc/test/http_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http/httptest"
 	"strconv"
 	"testing"
+	"time"
```
bergundy:
```suggestion
```

PR #8717, opened 2025-11-27
Metrics for the size of the header field

simvlad:
## What changed?
Start to track the size of the header field in the requests

## Why?
There are few reasons to start tracking the header size. First, it helps to picture the size of the request used by header field. Second, unlike payload and memo fields, we currently do not enforce the size of the header size. Better visibility into its size will determine at what level we would need to set up the enforcement.

## How did you test it?
- [X] built
- [X] run locally and tested manually - ran `samples-go/ctxpropagation` and checked that the header size metric is populated in Prometheus
- [ ] added new unit test(s)

## Potential risks
No

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -3122,6 +3122,11 @@ func (wh *WorkflowHandler) createScheduleWorkflow(
 		return nil, err
 	}
 
+	if startWorkflow := request.GetSchedule().GetAction().GetStartWorkflow(); startWorkflow != nil {
+		metricsHandler := wh.metricsScope(ctx).WithTags(metrics.HeaderCallsiteTag("CreateSchedule"))
+		metrics.HeaderSize.With(metricsHandler).Record(int64(startWorkflow.GetHeader().Size()))
+	}
+
 	// size limits will be validated on history. note that the start workflow request is
 	// embedded in the schedule, which is in the scheduler input. so if the scheduler itself
 	// doesn't exceed the limit, the started workflows should be safe as well.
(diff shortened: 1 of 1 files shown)


service/frontend/workflow_handler.go
```diff
@@ -1955,6 +1939,9 @@ func (wh *WorkflowHandler) SignalWorkflowExecution(ctx context.Context, request
 		return nil, err
 	}
 
+	metricsHandler := wh.metricsScope(ctx).WithTags(metrics.CommandTypeTag(enumspb.COMMAND_TYPE_UNSPECIFIED.String()))
```
bergundy: Ideally we would know what API this metric was emitted from. I would use a single label that can be applied to both a command and an API.

common/metrics/metric_defs.go
```diff
@@ -636,6 +634,7 @@ var (
 	TlsCertsExpiring                         = NewGaugeDef("certificates_expiring")
 	ServiceAuthorizationLatency              = NewTimerDef("service_authorization_latency")
 	EventBlobSize                            = NewBytesHistogramDef("event_blob_size")
+	HeaderSize                               = NewBytesHistogramDef("header_size", WithDescription("The size of the header in bytes passed to the server by the client."))
```
bergundy: I would call this out that this metric is experimental and can be removed without notice.

service/history/api/create_workflow_util.go
```diff
@@ -217,37 +215,39 @@ func ValidateStart(
...
+	workflowHeaderSize int,
 	operation string,
 ) error {
 	config := shard.GetConfig()
 	logger := shard.GetLogger()
 	throttledLogger := shard.GetThrottledLogger()
 	namespaceName := namespaceEntry.Name().String()
 
+	metricsHandler := interceptor.GetMetricsHandlerFromContext(ctx, logger).WithTags(metrics.CommandTypeTag(operation))
+	metrics.HeaderSize.With(metricsHandler).Record(int64(workflowHeaderSize))
```
bergundy: Why emit this both on the history and frontend?

bergundy (commented):
Looks like you missed workflow update headers.
I think all of the commands are covered by I would double check.
Look [here](https://github.com/search?q=repo%3Atemporalio%2Fapi%20v1.header&type=code) for all of the places where header is referenced.
Maybe also inspect the header in the `CreateSchedule` and UpdateSchedule` APIs in the action to start a workflow.

service/history/api/create_workflow_util.go
simvlad: I tried to emit this one close to where the payload/memo sizes are being enforced (right after this). Also I don't think this is emitted for start workflow execution requests specifically in the frontend.

bergundy (approved):

PR #8720, opened 2025-11-29
Fix bug in CHASM activity schedule-to-close timer task validation

dandavison:
## What changed?
- Fix bug: schedule-to-close timer task validator was incorrectly requiring activity attempt at task execution time to be equal to activity attempt at task creation
- Add test of schedule-to-close timeout that fails with the bug fix reverted
- Do not set empty struct as outcome failure on attempt failure when retries are exhausted.
- Improve doc comments

## Why?
- Standalone activity schedule-to-close was incorrect: would not have fired after attempt 1 without this fix
- Setting empty struct on attempt failure when retries are exhausted should not be necessary and it is fragile to introduce special values that code might start to rely on.

## How did you test it?
- [x] built

---

> [!NOTE]
> Decouples schedule-to-close timeout from attempt matching, updates proto to empty task payload, and adds a retry-based timeout test.
> 
> - **Activity execution**:
>   - Relax `schedule-to-close` timeout validation to only require `TransitionTimedOut` to be possible (remove attempt check) in `activity_tasks.go`.
>   - When scheduling `ScheduleToCloseTimeoutTask`, stop setting `Attempt` in `statemachine.go`.
> - **Proto/Generated code**:
>   - Make `activity.proto.v1.ScheduleToCloseTimeoutTask` an empty message; regenerate Go (`tasks.pb.go`).
> - **Tests**:
>   - Add `TestScheduleToCloseTimeout_WithRetry` to verify schedule-to-close timeout across a retry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 58f13fc34b57f2a383d3d0b0a978438984ef7700. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/standalone_activity_test.go
+++ b/tests/standalone_activity_test.go
@@ -802,7 +802,7 @@ func (s *standaloneActivityTestSuite) TestScheduleToCloseTimeout_WithRetry() {
 		},
 	})
 	require.NoError(t, err)
-	pollTaskResp, err = s.pollActivityTaskQueue(ctx, taskQueue)
+	_, err = s.pollActivityTaskQueue(ctx, taskQueue)
 	require.NoError(t, err)
 
 	// Wait for schedule-to-close timeout.


chasm/lib/activity/library.go
```diff
@@ -48,7 +48,7 @@ func (l *library) Components() []*chasm.RegistrableComponent {
 
 func (l *library) Tasks() []*chasm.RegistrableTask {
 	return []*chasm.RegistrableTask{
-		chasm.NewRegistrableSideEffectTask[*Activity, *activitypb.ActivityDispatchTask](
+		chasm.NewRegistrableSideEffectTask(
```
dandavison: I removed this because these types are inferred.

chasm/lib/activity/activity_tasks.go
```diff
@@ -96,25 +101,21 @@ func (e *scheduleToStartTimeoutTaskExecutor) Execute(
...
+	_ *activitypb.ScheduleToCloseTimeoutTask,
 ) (bool, error) {
-	attempt, err := activity.Attempt.Get(ctx)
-	if err != nil {
-		return false, err
-	}
-
-	valid := TransitionTimedOut.Possible(activity) && task.Attempt == attempt.Count
-	return valid, nil
+	return TransitionTimedOut.Possible(activity), nil
```
dandavison, line 111: This is the bug fix

fretz12: I believe you're right since there could be multiple retry attempts within a scheduleToClose tasks. Good catch.

chasm/lib/activity/activity_tasks.go
```diff
@@ -45,6 +45,9 @@ func (e *activityDispatchTaskExecutor) Validate(
 	return true, nil
 }
 
+// ActivityDispatchTask is a side-effect task that calls AddActivityTask in the matching service.
```
fretz12: Maybe move this doc to where the struct resides?

chasm/lib/activity/library.go
fretz12: I had added it cuz my IDE had trouble inferring it, though everything compiles.

chasm/lib/activity/proto/v1/tasks.proto
```diff
@@ -4,26 +4,26 @@ package temporal.server.chasm.lib.activity.proto.v1;
...
 message ScheduleToStartTimeoutTask {
-  // The current attempt number for this activity execution. Since task validation/exec happen outside of a lock, we
-  // need to guard against any concurrent operations where the originally intended task may be outdated.
+  // The current attempt number for this activity execution. Used for task validation.
   int32 attempt = 1;
 }
 
+// ScheduleToCloseTimeoutTask is a pure task that enforces a timeout across the sequence of activity
+// attempts.
 message ScheduleToCloseTimeoutTask {
```
fretz12, line 19: I believe you can remove this altogether now. In place of the interface arg you can use `_ any`

fretz12:
Thanks for catching this.

chasm/lib/activity/activity.go
```diff
@@ -389,10 +384,7 @@ func (a *Activity) recordFailedAttempt(
 	}
 	attempt.CompleteTime = currentTime
 
-	// If the activity has exhausted retries, mark the outcome failure as well but don't store duplicate failure info.
-	// Also reset the retry interval as there won't be any more retries.
 	if noRetriesLeft {
-		outcome.Variant = &activitypb.ActivityOutcome_Failed_{}
```
dandavison, line 417: @fretz12 thanks for reviewing. After you reviewed, I added this commit which gets rid of this set-to-empty-struct on this line. Would you mind looking and seeing if you agree that it's unnecessary? I'd prefer not to do it because I don't want code relying on a special value set here -- I feel that the code should just be able to take the failure from the right place without any special empty value here. https://github.com/temporalio/temporal/pull/8720/commits/0cbe6a96ddfbb38cf458d18513d6e10a079efc57

fretz12: I'm ok with that as long as the outcome is filled out on any GET API responses if an activity has reached terminal state, whether from the Attempt or Outcome field stored internally.

chasm/lib/activity/activity_tasks.go
dandavison: I have documented it with the struct also. I felt that it was helpful to have a sentence saying what these tasks actually are in the file where they are implemented. WDYT?

chasm/lib/activity/library.go
dandavison: Ah OK. Well shout out if you think we should keep it. I'd prefer to have the code be in line with Go rather than tracking IDE deficiencies, but then my IDE doesn't have a problem with it :)

chasm/lib/activity/activity_tasks.go
fretz12: ah ok, yea works for me

chasm/lib/activity/library.go
fretz12: yea we can remove it. Just a slight annoyance for me, hopefully the IDE will address the issue soon.

chasm/lib/activity/activity.go
dandavison: I've removed this change from the PR so that we can discuss it separately.

chasm/lib/activity/proto/v1/tasks.proto
bergundy: Keep it because we will have validation later when we support activity resets.

tests/standalone_activity_test.go
```diff
@@ -709,8 +709,69 @@ func (s *standaloneActivityTestSuite) TestCompletedActivity_CannotTerminate() {
 	require.Error(t, err)
 }
 
+func (s *standaloneActivityTestSuite) Test_ScheduleToCloseTimeout_WithRetry() {
```
bergundy: This test could be flaky when CI is under load.
2 seconds is fairly short to ensure at least one attempt is issued.

tests/standalone_activity_test.go
```diff
@@ -709,8 +709,69 @@ func (s *standaloneActivityTestSuite) TestCompletedActivity_CannotTerminate() {
...
+	pollTaskResp, err := s.pollActivityTaskQueue(ctx, taskQueue)
+	require.NoError(t, err)
+	_, err = s.FrontendClient().RespondActivityTaskFailed(ctx, &workflowservice.RespondActivityTaskFailedRequest{
+		Namespace: s.Namespace().String(),
+		TaskToken: pollTaskResp.TaskToken,
+		Failure: &failurepb.Failure{
+			Message: "Retryable failure",
+			FailureInfo: &failurepb.Failure_ApplicationFailureInfo{ApplicationFailureInfo: &failurepb.ApplicationFailureInfo{
+				NonRetryable:   false,
+				NextRetryDelay: durationpb.New(1 * time.Second),
```
bergundy, line 800: Depending on timing, this may prevent the schedule to close timeout from firing because we would know there's not enough time for the next attempt and avoid scheduling it.

bergundy: We should have a test for this behavior if we don't yet.

tests/standalone_activity_test.go
```diff
@@ -709,8 +709,69 @@ func (s *standaloneActivityTestSuite) TestCompletedActivity_CannotTerminate() {
...
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, enumspb.ACTIVITY_EXECUTION_STATUS_TIMED_OUT, pollResp.GetInfo().GetStatus())
+	require.Equal(t, enumspb.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE, pollResp.GetFailure().GetTimeoutFailureInfo().GetTimeoutType())
+}
+
 // TestStartToCloseTimeout tests that a start-to-close timeout is recorded after the activity is started.
-func (s *standaloneActivityTestSuite) TestStartToCloseTimeout() {
+func (s *standaloneActivityTestSuite) Test_StartToCloseTimeout() {
```
bergundy: We typically don't put an underscore after the word `Test` in the codebase.


```suggestion
func (s *standaloneActivityTestSuite) TestStartToCloseTimeout() {
```

dandavison:
> The test you added is okay but I am slightly worried that it will be flaky. I would recommend unit testing instead where you have more control over timing.

Agreed, I'm aware that some of the functional tests I've been writing involving timer tasks may be flaky. Let's address this in follow-on PRs. (They are doing their immediate job of verifying that the intended algorithm has been implemented.)

tests/standalone_activity_test.go
dandavison: Fixed

PR #8722, opened 2025-11-30
Add revision number to Version Data

ShahabT:
## What changed?
Add revision number to Worker Deployment Version Data that is synced to Task Queues.

## Why?
This allows making task queue registration also async, but more importantly, it prevent's race conditions between concurrent registrations, setCurrent/Ramping, drainage, and deletion.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None.


service/worker/workerdeployment/fx.go
```diff
@@ -23,9 +25,11 @@ const (
 	// history for TestReplays using generate_history.sh.
 
 	// Represents the state before the versioning API's received the option of becoming async in nature
-	InitialVersion DeploymentWorkflowVersion = 0
-	// SetCurrent and SetRamping APIs are async
-	AsyncSetCurrentAndRamping = 1
+	InitialVersion DeploymentWorkflowVersion = iota
+	// SetCurrent and SetRamping and DeleteVersion APIs are async
+	AsyncSetCurrentAndRamping
+	// Version Data has its own revision number and Task Queue registration is async as well
```
Shivs11: nit: "with TaskQueue registration being async as well"

service/matching/matching_engine.go
```diff
@@ -2039,8 +2039,28 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
...
+						for idx, oldVersions := range deploymentData.GetVersions() {
+							if oldVersions.GetVersion().GetDeploymentName() == req.GetDeploymentName() &&
+								oldVersions.GetVersion().GetBuildId() == buildID {
+								//nolint:staticcheck // SA1019 deprecated versions will clean up later
+								deploymentData.Versions = append(deploymentData.Versions[:idx], deploymentData.Versions[idx+1:]...)
+								changed = true
+								break
+							}
+						}
+					}
```
Shivs11, line 2043: Just underneath this function is the helper function: `removeDeploymentVersions`. This function was created by me to serve the purpose of removing versions from the new and old formats.

I think we should either have your logic go into that function so that one can read the code easily, or mark that as deprecated. I prefer the former method but shall leave it up to you.

service/matching/matching_engine.go
```diff
@@ -2085,6 +2105,10 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
 						tqWorkerDeploymentData,
 					)
 				}
+
+				if cleanupOldDeletedVersions(tqWorkerDeploymentData) {
+					changed = true
```
Shivs11, line 2090: maybe we could just have this function placed inside of `removeDeploymentVersions` and call it something like `removeAndCleanUpDeploymentVersions`. It shall then have the contained logic of deleting old versions, marking the version in the new format as deleted and also Garbage Collecting any versions if present.

service/matching/matching_engine.go
```diff
@@ -2100,6 +2124,20 @@ func (e *matchingEngineImpl) SyncDeploymentUserData(
 	return &matchingservice.SyncDeploymentUserDataResponse{Version: version, RoutingConfigChanged: applyUpdatesToRoutingConfig}, nil
 }
 
+func cleanupOldDeletedVersions(deploymentData *persistencespb.WorkerDeploymentData) bool {
+	cleaned := false
+	for buildID, versionData := range deploymentData.Versions {
+		// TODO: improve this logic to remove even more old versions if we're reaching the limit of max versions per task queue.
+		// max versions per task queue is not implemented yet but we should have it because as of now a task queue can be polled from
+		// numerous deployment versions (accross different deployments) and that leads to a very large user data size.
```
Shivs11, line 2112: thinking about this a bit more - I think we should have a safeguard (dynamic config) for this prior to launching this as a GA. If folks use versioning with their CI/CD systems (as is done by our internal saas team) on the same task-queue, what you have placed in the comment can 100% come to life.

wdyt?

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -370,20 +370,20 @@ message SyncDeploymentUserDataRequest {
...
     repeated string forget_versions = 12;
 }
 
 message SyncDeploymentUserDataResponse {
     // New task queue user data version. Can be used to wait for propagation.
     int64 version = 1;
     // If the routing config changed after applying this operation. Compared base on revision number.
-    bool routing_config_changed = 2;
+    // Deprecated. using this is not totaly safe in case of retries.
+    bool routing_config_changed = 2 [deprecated = true];
```
Shivs11, line 376: this has confused me a little....might be worth discussing on call.

service/worker/workerdeployment/version_activities.go
```diff
@@ -93,12 +93,9 @@ func (a *VersionActivities) SyncDeploymentVersionUserData(
...
 			} else {
-				if syncData.GetData() != nil || res.GetRoutingConfigChanged() {
-					// No need to wait for partition propagation if it's async mode (no old-format data is provided) and routing config did not change
-					lock.Lock()
-					maxVersionByName[syncData.Name] = max(maxVersionByName[syncData.Name], res.Version)
-					lock.Unlock()
-				}
+				lock.Lock()
+				maxVersionByName[syncData.Name] = max(maxVersionByName[syncData.Name], res.Version)
+				lock.Unlock()
```
Shivs11, line 98: i'm wondering if we truly do need this max(maxVersionByName,....) behaviour now. We ensure that in every sync request going to matching, we have one task-queue name with it's different types (activity, workflow, nexus). 

Plus, making things async doesn't really matter here since the map we are conversing about is not global here.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -49,8 +50,12 @@ type (
...
 		asyncPropagationsInProgress int
 		// When true, all the ongoing propagations should cancel themselves
-		cancelPropagations          bool
-		unsafeWorkflowVersionGetter func() DeploymentWorkflowVersion
+		// Deprecated. With version data revision number, we don't need to cancel propagations anymore.
+		cancelPropagations bool
+		// Tracks the version of the deployment version workflow when a particular run of a workflow starts base on the dynamic config of the
+		// worker who completes the first task of the workflow. `workflowVersion` remains the same until the workflow CaNs when it
+		// will get another chance to pick the latest manager version.
+		workflowVersion DeploymentWorkflowVersion
```
Shivs11: nit: 

> // workflowVersion is set at workflow start based on the dynamic config of the worker
// that completes the first task. It remains constant for the lifetime of the run and
// only updates when the workflow performs continue-as-new.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -496,6 +516,17 @@ func (d *VersionWorkflowRunner) handleRegisterWorker(ctx workflow.Context, args
 	}
 	d.VersionState.TaskQueueFamilies[args.TaskQueueName].TaskQueues[int32(args.TaskQueueType)] = &deploymentspb.TaskQueueVersionData{}
 
+	if withRevisionNumbers && args.GetRoutingConfig() != nil {
+		// Still need to check RoutingConfig not being nil because of edge cases during enabling dynamic config.
+		// i.e. the deployment workflow might run old version and not send the routing config.
```
Shivs11, line 521: As far as my understanding goes, you have placed this check since you are worried that the deployment workflow and the version workflow may not be in full sync when it comes to the version of the code they are running.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -1095,6 +1168,45 @@ func (d *VersionWorkflowRunner) signalPropagationComplete(ctx workflow.Context,
...
+}
+
+// syncRegisteredTaskQueueAsync syncs the routing config and version data to the new task queue.
+// This method does not increment version data revision number.
+// Note: task queue registration does not affect WorkerDeploymentInfo.RoutingConfigUpdateState hence we do not signal
+// the deployment workflow about propagation completion.
+// TODO: should RoutingConfigUpdateState become IN_PROGRESS while registration is propagating? Current thoughts: It
+// doesn't seem useful because the following setCurrent/Ramping will set the status to IN_PROGRESS. The only case is if
+// user is calling setCurrent/Ramping before registration passing IgnoreMissingTaskQueues, but in that case still the
+// registration is likely not in control of the operator but is async all the way from the worker side.
```
Shivs11: lets discuss this on a call; have some concerns/thoughts.

Shivs11: FWIW - I think we should be a bit conservative here and have the routing config state to be in progress. I think convoluting different API calls further with an in-progress propagation is more error prone

service/worker/workerdeployment/workflow.go
```diff
@@ -596,8 +596,13 @@ func (d *WorkflowRunner) handleSetRampingVersion(ctx workflow.Context, args *dep
 
 		// Build pending routing config with the updated ramping version
 		// Initialize for both sync and async modes to simplify state update logic
+		// Ensure CurrentDeploymentVersion is populated from deprecated field if needed for backward compatibility
+		currentDeploymentVersion := d.State.RoutingConfig.CurrentDeploymentVersion
+		if currentDeploymentVersion == nil {
+			currentDeploymentVersion = worker_versioning.ExternalWorkerDeploymentVersionFromStringV31(d.State.RoutingConfig.CurrentVersion)
+		}
```
Shivs11, line 617: I think we should just make our life simpler by always populating both values while the workflow CAN's. It's simpler since then, in all our API handlers, we can just focus on using the new value and not worry about workflow versions

service/matching/matching_engine.go
ShahabT: will follow up with a PR to clean up more TQs when reaching a limit

service/worker/workerdeployment/version_workflow.go
ShahabT: Yeah I thought more about this and I think we should do in the case that version is already ramping or current (but not generally). Updated my comment. I think it's fine to do it later since it's a very small edge case.

service/worker/workerdeployment/workflow.go
ShahabT: Will follow up in another PR

PR #8726, opened 2025-12-01
panic test hooks if not enabled

stephanos:
## What changed?

Changed use of `testhooks`'s `Set` to be runtime error instead of compile time error.

It still ensures that the default impl (used in `go build` artifacts without `test_dep` tag) is the noop impl.

## Why?

When running a functional test via plain `go test` it will fail, as the `test_dep` build flag is required. It is required to prevent accidentally using the default/noop implementation of the `testhooks` package which allows injecting custom behavior into production code from tests.

But requiring `test_dep` to run any test adds friction: CLI invocations need to specify it manually, editor setups need to be adjusted, AI agents need to be instructed. That's a high price to pay for a test helper that is used fairly sparsely. 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

Functional test without testhook:
```
% go test ./tests -run "OwnershipLostErrorSuite" -count=1 .
ok  	go.temporal.io/server/tests	7.062s
```

Functional test with testhook and build tag:
```
% go test ./tests -run "TestUpdateWorkflowSuite" -tags test_dep -count=1 .
ok  	go.temporal.io/server/tests	36.975s
```

Functional test with testhook but no build tag:
```
% go test ./tests -run "TestUpdateWorkflowSuite" -count=1 .
...
noop_impl.go:39: test panicked: testhooks.Set called but TestHooks are not enabled: use -tags=test_dep when running `go test`
                    goroutine 33293 [running]:
                    runtime/debug.Stack()
...
FAIL
FAIL	go.temporal.io/server/tests	57.128s
FAIL
```

### DIFF
--- a/common/testing/testhooks/test_impl.go
+++ b/common/testing/testhooks/test_impl.go
@@ -12,10 +12,6 @@ var Module = fx.Options(
 	fx.Provide(NewTestHooks),
 )
 
-type testHooksData struct {
-	m sync.Map
-}
-
 // TestHooks holds a registry of active test hooks. It should be obtained through fx and
 // used with Get and Set.
 //
@@ -23,11 +19,11 @@ type testHooksData struct {
 // concerns into production code. In general you should prefer other ways of writing tests
 // wherever possible, and only use TestHooks sparingly, as a last resort.
 type TestHooks struct {
-	data *testHooksData
+	data *sync.Map
 }
 
 func NewTestHooks() TestHooks {
-	return TestHooks{data: &testHooksData{}}
+	return TestHooks{data: &sync.Map{}}
 }
 
 // Get gets the value of a test hook from the registry.
@@ -38,7 +34,7 @@ func Get[T any](th TestHooks, key Key) (T, bool) {
 	if th.data == nil {
 		return zero, false
 	}
-	if val, ok := th.data.m.Load(key); ok {
+	if val, ok := th.data.Load(key); ok {
 		// this is only used in test so we want to panic on type mismatch:
 		return val.(T), ok //nolint:revive
 	}
@@ -57,6 +53,6 @@ func Call(th TestHooks, key Key) {
 // Set sets a test hook to a value and returns a cleanup function to unset it.
 // Calls to Set and the cleanup function should form a stack.
 func Set[T any](th TestHooks, key Key, val T) func() {
-	th.data.m.Store(key, val)
-	return func() { th.data.m.Delete(key) }
+	th.data.Store(key, val)
+	return func() { th.data.Delete(key) }
 }


common/testing/testhooks/noop_impl.go
```diff
@@ -31,23 +38,14 @@ func Get[T any](_ TestHooks, key Key) (T, bool) {
...
 }
 
-// CallCtx calls a func(context.Context) hook if present.
-//
-// TestHooks should be used very sparingly, see comment on TestHooks.
-func CallCtx(_ context.Context, key Key) {
+// Set is only to be used by test code together with the test_dep build tag.
+func Set[T any](th TestHooks, key Key, val T) func() {
+	softassert.Fail(th.logger, "testhooks.Set called without test_dep build tag")
+	return func() {}
```
stephanos: The main change: instead of a compile time error, a runtime error for the very few tests that rely on testhooks is emitted.

Risk and impact of accidentally using it in production code are very low (ie extra logging, which should be caught in testing). But could be mitigated via linter, if needed.

common/testing/testhooks/test_impl.go
```diff
@@ -50,19 +47,6 @@ func Get[T any](th TestHooks, key Key) (T, bool) {
...
-// TestHooks should be used sparingly, see comment on TestHooks.
-func GetCtx[T any](ctx context.Context, key Key) (T, bool) {
-	hooks := ctx.Value(contextKey{})
-	if hooks, ok := hooks.(TestHooks); ok {
-		return Get[T](hooks, key)
-	}
-	var zero T
-	return zero, false
-}
-
```
stephanos, line 65: It's been a while and these haven't seen any use. Safe to remove IMO.

common/testing/testhooks/noop_impl.go
stephanos: Thoughts from a few days later; could also consider using `panic` actually to make the error clearer when a user forgets to set `test_dep` when testing.

I wouldn't expect this function to be used in production code. Again, linter could be leveraged to make sure.

common/testing/testhooks/noop_impl.go
```diff
@@ -3,51 +3,38 @@
...
 
 import (
-	"context"
-
 	"go.uber.org/fx"
 )
 
 var Module = fx.Options(
-	fx.Provide(func() (_ TestHooks) { return }),
+	fx.Provide(NewTestHooks),
```
dnr, line 10: what was wrong with the interface/concrete switch? an interface takes 16 bytes, a concrete struct{} takes 0. yeah, it may not be noticeable but why change it?

stephanos: I was playing around with some options; the size wasn't top of mind for me here. I'll revert that, no problem.

common/testing/testhooks/test_impl.go
```diff
@@ -3,63 +3,45 @@
...
-		m sync.Map
-	}
+// TestHooks holds a registry of active test hooks. It should be obtained through fx and
+// used with Get and Set.
+//
+// TestHooks are an inherently unclean way of writing tests. They require mixing test-only
+// concerns into production code. In general you should prefer other ways of writing tests
+// wherever possible, and only use TestHooks sparingly, as a last resort.
+type TestHooks struct {
+	data *testHooksData
```
dnr: could probably just be `data *sync.Map`?

PR #8727, opened 2025-12-01
[Refactoring]: Move a subset of validators to the frontend

Shivs11:
## What changed?
- Move some of the validator functions to the frontend, where some pre-existing validations are already present.
- Quite specifically, move the function ValidateVersioningOverride to the frontend. The detailed reasoning is present below.
- Additionally, move some of the validator functions that internally call the function ValidateVersioningOverride to the frontend.

## Why?
- The reasons I have moved the validators to the frontend are as follows:
1. ValidateVersioningOverride will soon be making an RPC call. To make this RPC call, the function would require _workerdeployment.Client_. The frontend handler, currently, has this already present as a member. Note - I thought of embedding this directly in the history handler itself but chose against it since the  _workerdeployment.Client_ internally has a dependency on history, which would have then caused circular dependencies. 
2. Additionally, failed validations would unnecessary make a RPC roundtrip to history (in some cases). Instead, we should be validating user inputs in the frontend only and quickly returning an error whenever the inputs don't match what we expect.

**NOTE**: This is part 1 of a change that I am working on. Part 2 involves actually adding that RPC call that I mentioned above. The reason why I have a separate PR open for this is because this touches parts of the codebase worked on by many different folks. Thus, for the sole purpose of iterating faster, I decided to make refactoring into it's own PR.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
This should not be a risky change.

---

> [!NOTE]
> Shift versioning-override and related input validations from history to frontend handlers, add reset post-op validation, and simplify batch workflow by validating at API layer.
> 
> - **Frontend**:
>   - Validate `VersioningOverride` in `StartWorkflowExecution`, `SignalWithStartWorkflowExecution`, `ResetWorkflowExecution` (via new `validatePostResetOperationInputs`), and `UpdateWorkflowExecutionOptions`.
>   - Adjust `StartBatchOperation` flow: ensure `EnableBatcher` check precedes `ValidateBatchOperation`.
>   - Minor: fetch `namespaceID` before validating options in `UpdateWorkflowExecutionOptions`.
> - **History**:
>   - Remove `VersioningOverride` validations from `create_workflow_util.go`, `resetworkflow/api.go` (and delete its local post-reset validator), and `updateworkflowoptions/api.go` (also pass `req.GetWorkflowExecutionOptions()` directly to `MergeAndApply`).
> - **Worker/Batcher**:
>   - Drop in-workflow `ValidateBatchOperation` call; rely on frontend validation. Update tests accordingly (remove missing-params test).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d1e90031ff0a6dae30c8bdf972ade522a3a5b823. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -4515,6 +4515,10 @@ func (wh *WorkflowHandler) StartBatchOperation(
 		return nil, errBatchAPINotAllowed
 	}
 
+	if err := batcher.ValidateBatchOperation(request); err != nil {
+		return nil, err
+	}
+
 	// Validate concurrent batch operation
 	maxConcurrentBatchOperation := wh.config.MaxConcurrentBatchOperation(request.GetNamespace())
 	countResp, err := wh.CountWorkflowExecutions(ctx, &workflowservice.CountWorkflowExecutionsRequest{
@@ -4587,10 +4591,6 @@ func (wh *WorkflowHandler) StartBatchOperation(
 		identity = op.UpdateActivityOptionsOperation.GetIdentity()
 	default:
 		return nil, serviceerror.NewInvalidArgumentf("The operation type %T is not supported", op)
-	}
-
-	if err := batcher.ValidateBatchOperation(request); err != nil {
-		return nil, err
 	}
 
 	inputPayload, err := payloads.Encode(input)
--- a/service/worker/batcher/workflow_test.go
+++ b/service/worker/batcher/workflow_test.go
@@ -37,13 +37,6 @@ func (s *batcherSuite) TearDownTest() {
 	s.env.AssertExpectations(s.T())
 }
 
-func (s *batcherSuite) TestBatchWorkflow_MissingParams_Protobuf() {
-	s.env.ExecuteWorkflow(BatchWorkflowProtobuf, &batchspb.BatchOperationInput{})
-	err := s.env.GetWorkflowError()
-	s.Require().Error(err)
-	s.Contains(err.Error(), "must provide required parameters")
-}
-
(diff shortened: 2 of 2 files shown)


service/frontend/workflow_handler.go
```diff
@@ -2117,6 +2122,10 @@ func (wh *WorkflowHandler) ResetWorkflowExecution(ctx context.Context, request *
 		return nil, err
 	}
 
+	if err := validatePostResetOperationInputs(request.GetPostResetOperations()); err != nil {
```
Shivs11, line 2125: cc - @gow moved this guy outside from history to frontend.

service/frontend/workflow_handler.go
Shivs11, line 4488: seems to be duplicated call - see L4592 
@bergundy / @spkane31

service/history/api/updateworkflowoptions/api.go
```diff
@@ -30,11 +29,6 @@ func Invoke(
 	req := request.GetUpdateRequest()
 	ret := &historyservice.UpdateWorkflowExecutionOptionsResponse{}
 
-	opts := req.GetWorkflowExecutionOptions()
-	if err := worker_versioning.ValidateVersioningOverride(opts.GetVersioningOverride()); err != nil {
-		return nil, err
-	}
-
```
Shivs11, line 37: removing this validation from history and putting this in the frontend for updateWorkflowExecutionOptions. If this fails, we would simply waste one RPC trip to history so we better catch this in the frontend.

cc - @carlydf

PR #8730, opened 2025-12-01
Standalone activity heartbeating

dandavison:
## What changed?
Implement heartbeating for CHASM activities (standalone activity)

## Why?
Required feature

## How did you test it?
- [x] built

---

> [!NOTE]
> Implements heartbeating for standalone activities, including heartbeat timers, timeout-driven retry/failure, task-token validation, frontend routing, and supporting proto/executors with comprehensive tests.
> 
> - **Activity/State Machine**:
>   - Add `RecordHeartbeat` with task-token validation; store heartbeat details/time and schedule heartbeat timer; response signals `cancelRequested`.
>   - Start schedules heartbeat and start-to-close timers; `TransitionTimedOut` handles `HEARTBEAT`; `getOrCreateLastHeartbeat` used when recording failures.
>   - Introduce `WithToken` wrapper and enforce token validation in `HandleCompleted/Failed/Canceled` and heartbeats.
>   - Extract retry flow via `tryReschedule/shouldRetry/hasEnoughTimeForRetry`; start-to-close timeout uses retry-first logic.
>   - `PopulateRecordStartedResponse` includes prior `HeartbeatDetails`; `TransitionStarted` records worker/deployment and uses started-time-based scheduling.
> - **Tasks/Proto/FX**:
>   - Define `HeartbeatTimeoutTask` in proto (+ helpers); implement validate/execute executors; register in `fx` and task registry.
> - **History APIs**:
>   - Route standalone `RecordActivityTaskHeartbeat` and `RespondActivityTask{Completed,Failed,Canceled}` via `chasm.UpdateComponent` using `WithToken`.
> - **Tests**:
>   - Add/extend functional tests for heartbeats: invalid/stale tokens, cancelRequested flag, retry on heartbeat timeout, timeout without heartbeats, and heartbeats preventing timeout; adjust started transition test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit af7ca83a7e181bd5c0a1b36829df73807eaa5a9c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/standalone_activity_test.go
+++ b/tests/standalone_activity_test.go
@@ -1577,10 +1577,6 @@ func (s *standaloneActivityTestSuite) TestDescribeActivityExecution_InvalidArgum
 }
 
 func (s *standaloneActivityTestSuite) TestHeartbeat() {
-	// TODO(dan)
-	// - a heartbeat timer set on one attempt should not timeout a subsequent attempt (not unless
-	//   there have been some appropriate adjustments to the schedule time)
-
 	t := s.T()
 	ctx := testcore.NewContext()
 	heartbeatDetails := payloads.EncodeString("Heartbeat Details")


chasm/lib/activity/activity.go
```diff
@@ -132,8 +139,10 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
 }
 
 // HandleStarted updates the activity on recording activity task started and populates the response.
-func (a *Activity) HandleStarted(ctx chasm.MutableContext, request *historyservice.RecordActivityTaskStartedRequest) (
-	*historyservice.RecordActivityTaskStartedResponse, error) {
+func (a *Activity) HandleStarted(
+	ctx chasm.MutableContext,
+	request *historyservice.RecordActivityTaskStartedRequest,
+) (*historyservice.RecordActivityTaskStartedResponse, error) {
```
dandavison: ignore: formatting only

chasm/lib/activity/activity.go
```diff
@@ -172,7 +181,11 @@ func (a *Activity) HandleStarted(ctx chasm.MutableContext, request *historyservi
 	return response, nil
 }
 
-func (a *Activity) PopulateRecordStartedResponse(ctx chasm.Context, key chasm.EntityKey, response *historyservice.RecordActivityTaskStartedResponse) error {
+func (a *Activity) PopulateRecordStartedResponse(
+	ctx chasm.Context,
+	key chasm.EntityKey,
+	response *historyservice.RecordActivityTaskStartedResponse,
+) error {
```
dandavison: ignore: formatting only

chasm/lib/activity/activity.go
```diff
@@ -294,8 +330,10 @@ func (a *Activity) getLastHeartbeat(ctx chasm.MutableContext) (*activitypb.Activ
 	return heartbeat, nil
 }
 
-func (a *Activity) handleCancellationRequested(ctx chasm.MutableContext, req *activitypb.CancelActivityExecutionRequest) (
-	*activitypb.CancelActivityExecutionResponse, error) {
+func (a *Activity) handleCancellationRequested(
+	ctx chasm.MutableContext,
+	req *activitypb.CancelActivityExecutionRequest,
+) (*activitypb.CancelActivityExecutionResponse, error) {
```
dandavison: ignore: formatting only

chasm/lib/activity/validator.go
```diff
@@ -309,3 +312,23 @@ func ValidatePollActivityExecutionRequest(
...
+
+// ValidateActivityTaskToken validates a task token against the current activity state.
+func ValidateActivityTaskToken(
+	ctx chasm.Context,
+	a *Activity,
+	token *tokenspb.Task,
+) error {
+	if a.Status != activitystatepb.ACTIVITY_EXECUTION_STATUS_STARTED &&
+		a.Status != activitystatepb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED {
+		return serviceerror.NewNotFound("activity task not found")
```
fretz12: Wrong error message? Also NotFound doesn't feel like the right error type, even though I get it's checking if a token "matches". Perhaps FailPrecondition?

chasm/lib/activity/validator.go
```diff
@@ -309,3 +312,23 @@ func ValidatePollActivityExecutionRequest(
...
+	}
+	attempt, err := a.Attempt.Get(ctx)
+	if err != nil {
+		return err
+	}
+	if token.Attempt != attempt.GetCount() {
+		return serviceerror.NewNotFound("activity task not found")
```
fretz12: Same comment as above

chasm/lib/activity/activity.go
```diff
@@ -455,12 +495,33 @@ func createStartToCloseTimeoutFailure() *failurepb.Failure {
...
+
+// RecordHeartbeat records a heartbeat for the activity.
+func (a *Activity) RecordHeartbeat(
+	ctx chasm.MutableContext,
+	input WithToken[*historyservice.RecordActivityTaskHeartbeatRequest],
+) (*historyservice.RecordActivityTaskHeartbeatResponse, error) {
+	if err := ValidateActivityTaskToken(ctx, a, input.Token); err != nil {
+		return nil, err
+	}
 	a.LastHeartbeat = chasm.NewDataField(ctx, &activitypb.ActivityHeartbeatState{
```
fretz12, line 445: @bergundy  is there any performance pentalty if we create a new field on every hearbeat?

chasm/lib/activity/activity.go
```diff
@@ -455,12 +495,33 @@ func createStartToCloseTimeoutFailure() *failurepb.Failure {
...
 		RecordedTime: timestamppb.New(ctx.Now(a)),
-		Details:      details,
+		Details:      input.Request.HeartbeatRequest.GetDetails(),
```
fretz12:
```suggestion
		Details:      input.Request.GetHeartbeatRequest().GetDetails(),
```

chasm/lib/activity/activity.go
bergundy: Looks like @fretz12 and you have different ideas on how to format these signatures. I personally like the style in this PR better.

chasm/lib/activity/activity_tasks.go
```diff
@@ -172,3 +173,91 @@ func (e *startToCloseTimeoutTaskExecutor) Execute(
...
+		activity.Status == activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED) &&
+		attempt.GetCount() == task.Attempt
+	return valid, nil
+}
+
+// Execute executes a HeartbeatTimeoutTask.
+func (e *heartbeatTimeoutTaskExecutor) Execute(
+	ctx chasm.MutableContext,
+	activity *Activity,
+	taskAttrs chasm.TaskAttributes,
```
fretz12: nit: taskAttrs, task are unused

chasm/lib/activity/activity_tasks.go
```diff
@@ -172,3 +173,91 @@ func (e *startToCloseTimeoutTaskExecutor) Execute(
...
+
+// HeartbeatTimeoutTask is a pure task that enforces heartbeat timeouts.
+type heartbeatTimeoutTaskExecutor struct{}
+
+func newHeartbeatTimeoutTaskExecutor() *heartbeatTimeoutTaskExecutor {
+	return &heartbeatTimeoutTaskExecutor{}
+}
+
+// Validate validates a HeartbeatTimeoutTask.
+func (e *heartbeatTimeoutTaskExecutor) Validate(
```
bergundy, line 154: You will want to validate that the schedule time in the attributes is still relevant here not in the execute function. It should deterministic function of the last heartbeat time and the attempt start time (`hbDeadline` below should be equal to the schedule time).

chasm/lib/activity/activity_tasks.go
```diff
@@ -172,3 +173,91 @@ func (e *startToCloseTimeoutTaskExecutor) Execute(
...
+			activity,
+			chasm.TaskAttributes{
+				ScheduledTime: hbDeadline,
+			},
+			&activitypb.HeartbeatTimeoutTask{
+				Attempt: attempt.GetCount(),
+			},
+		)
+		return nil
+	}
```
bergundy: This shouldn't happen here. Validate in the `Validate` function. Schedule a new task any time a heartbeat is received.

chasm/lib/activity/activity_tasks.go
```diff
@@ -172,3 +173,91 @@ func (e *startToCloseTimeoutTaskExecutor) Execute(
...
+	// No heartbeats in the attempt so far is equivalent to a heartbeat having been sent at attempt
+	// start time.
+	hbDeadline := util.MaxTime(lastHbTime, attemptStartTime).Add(hbTimeout)
+
+	if ctx.Now(activity).Before(hbDeadline) {
+		// Deadline has not expired; schedule a new task.
+		ctx.AddTask(
+			activity,
+			chasm.TaskAttributes{
+				ScheduledTime: hbDeadline,
```
fretz12: If `hbDeadline` happens to be `attemptStartTime + hbTimeout` (i.e., no hearbeat recorded yet), won't the next timer basically pop the same time as the current one being executed?

chasm/lib/activity/library.go
```diff
@@ -53,6 +56,7 @@ func (l *library) Tasks() []*chasm.RegistrableTask {
 			l.activityDispatchTaskExecutor,
 			l.activityDispatchTaskExecutor,
 		),
+		// TODO(dan): why are the task names "FooTimer" but "FooTimeoutTask" in the struct names?
```
bergundy: We agreed not to add `Task` to the task names. Not sure if we want to use `TaskExecutor` vs. just `Executor` for the struct names but that is not critical and can change easily. The string names cannot be changed easily OTOH because they affect how tasks are represented in persistence.

chasm/lib/activity/activity.go
```diff
@@ -455,12 +495,33 @@ func createStartToCloseTimeoutFailure() *failurepb.Failure {
...
 	})
-	return nil, nil
+	return &historyservice.RecordActivityTaskHeartbeatResponse{
```
bergundy, line 458: Just before returning here you want to generate a new heartbeat task if the heartbeat timeout is set.

tests/standalone_activity_test.go
```diff
@@ -1407,6 +1415,433 @@ func (s *standaloneActivityTestSuite) Test_PollActivityExecution_InvalidArgument
...
+			t.Run(tc.name, func(t *testing.T) {
+				_, err := s.FrontendClient().RecordActivityTaskHeartbeat(ctx, &workflowservice.RecordActivityTaskHeartbeatRequest{
+					Namespace: s.Namespace().String(),
+					TaskToken: tc.taskToken,
+					Details:   heartbeatDetails,
+				})
+				require.Error(t, err)
+				statusErr := serviceerror.ToStatus(err)
+				require.NotNil(t, statusErr)
+				require.Equal(t, codes.InvalidArgument, statusErr.Code())
```
bergundy, line 1612: Alternatively, you can use `require.ErrorAs(err, &invalidArgumentErr)`

chasm/lib/activity/activity.go
fretz12: I like this formatting better too, but ideally we should have the linter take care of it. 
I'm running [gofumpt](https://github.com/mvdan/gofumpt) locally, it seems to do a good job... maybe introduce it as part of the lint-code target?

chasm/lib/activity/activity.go
bergundy: No, no penalty.

chasm/lib/activity/activity_tasks.go
dandavison: Thanks, marked as `_` unused parameters

chasm/lib/activity/activity_tasks.go
dandavison: Thanks, switched to new design.

chasm/lib/activity/activity.go
dandavison: Thanks, taken

chasm/lib/activity/validator.go
dandavison: I used `NotFound` because it'd what the workflow implementation uses (see `IsActivityTaskNotFoundForToken`)

chasm/lib/activity/validator.go
dandavison: I used `NotFound` because it'd what the workflow implementation uses (see `IsActivityTaskNotFoundForToken`)

tests/standalone_activity_test.go
dandavison: Leaving as is for now

chasm/lib/activity/activity_tasks.go
dandavison: I currently still think that the design I've proposed is preferable:

image

I've added a high water mark mechanism to make the tasks become invalid on execution.

chasm/lib/activity/activity_tasks.go
dandavison: See reply below

chasm/lib/activity/activity.go
dandavison: See reply above; I currently still think the proposed design is preferable.

chasm/lib/activity/activity_tasks.go
dandavison: That should not happen because of this condition

```go
if ctx.Now(activity).Before(hbDeadline)
```

the heartbeat task will not execute before time `hbDeadline` since the first task is scheduled for `hbTimeout`.

chasm/lib/activity/activity.go
```diff
@@ -419,12 +446,34 @@ func createStartToCloseTimeoutFailure() *failurepb.Failure {
...
 	a.LastHeartbeat = chasm.NewDataField(ctx, &activitypb.ActivityHeartbeatState{
 		RecordedTime: timestamppb.New(ctx.Now(a)),
-		Details:      details,
+		Details:      input.Request.GetHeartbeatRequest().GetDetails(),
 	})
-	return nil, nil
+	return &historyservice.RecordActivityTaskHeartbeatResponse{
+		CancelRequested: a.Status == activitypb.ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED,
+		// TODO(dan): ActivityPaused, ActivityReset
+	}, nil
```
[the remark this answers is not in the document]
dandavison: The PR implements a different design; see comment above.

chasm/lib/activity/activity.go
```diff
@@ -134,25 +141,26 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
...
 		attempt.LastDeploymentVersion = &deploymentpb.WorkerDeploymentVersion{
 			BuildId:        versionDirective.GetBuildId(),
 			DeploymentName: versionDirective.GetDeploymentName(),
 		}
 	}
+
+	if err := TransitionStarted.Apply(a, ctx, nil); err != nil {
+		return nil, err
+	}
+
```
dandavison: The movement here was required to fix a bug that I encountered when implementing the other proposed design. We needed `attempt.StartedTime` to be available in `TransitionStarted` so that we could use it in scheduling the first heartbeat task, as well as the base for the start-to-close timeout. Before this PR we were using an ad-hoc `ctx.Now()` that was close to but not equal to `StartedTime`.

Reverting it doesn't cause a test failure with the hreartbeating design in this PR, but it's better to set the field state before calling the transition function.

chasm/lib/activity/statemachine.go
```diff
@@ -125,14 +125,28 @@ var TransitionStarted = chasm.NewTransition(
 	},
 	activitypb.ACTIVITY_EXECUTION_STATUS_STARTED,
 	func(a *Activity, ctx chasm.MutableContext, _ any) error {
+		attempt := a.LastAttempt.Get(ctx)
+		startTime := attempt.GetStartedTime().AsTime()
```
dandavison, line 138: This is where we use `StartedTime` that I referred to above; it wasn't possible before because we were calling the transition function before setting the attempt field state.

Reverting this doesn't cause a test failure in the current PR. But it did with the alternative design, and in any case it's better to use `StartedTime` rather than an arbitrary timestamp that's close to it in time.

chasm/lib/activity/activity_tasks.go
dandavison: Discussed offline -- design B incurs extra persistence writes that aren't balanced out by the changes to numbers of logical tasks created. I've switched the PR over to the design that @bergundy proposes.

chasm/lib/activity/activity.go
dandavison: This is done now -- switched back to your design.

chasm/lib/activity/statemachine.go
dandavison: Reverting this would now cause a test failure since the PR is back to design A.

chasm/lib/activity/activity.go
dandavison: Reverting this would now cause a test failure since the PR is back to design A.

chasm/lib/activity/activity.go
```diff
@@ -134,25 +141,26 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
 func (a *Activity) HandleStarted(ctx chasm.MutableContext, request *historyservice.RecordActivityTaskStartedRequest) (
 	*historyservice.RecordActivityTaskStartedResponse, error,
 ) {
-	if err := TransitionStarted.Apply(a, ctx, nil); err != nil {
-		return nil, err
-	}
-
 	attempt := a.LastAttempt.Get(ctx)
 	attempt.StartedTime = timestamppb.New(ctx.Now(a))
 	attempt.LastWorkerIdentity = request.GetPollRequest().GetIdentity()
```
bergundy: Why is this logic outside of the transition function? Ideally no mutation would be done outside of transitions.

chasm/lib/activity/activity_tasks.go
```diff
@@ -146,3 +147,72 @@ func (e *startToCloseTimeoutTaskExecutor) Execute(
...
+	if err != nil {
+		return err
+	}
+	if shouldRetry {
+		return TransitionRescheduled.Apply(activity, ctx, rescheduleEvent{
+			retryInterval: retryInterval,
+			failure:       createHeartbeatTimeoutFailure(),
+		})
+	}
+	return TransitionTimedOut.Apply(activity, ctx, enumspb.TIMEOUT_TYPE_HEARTBEAT)
```
bergundy: nit: this logic seems duplicated any time an attempt is failed, can we add a function that encapsulates it?

chasm/lib/activity/validator.go
```diff
@@ -306,3 +309,19 @@ func ValidateGetActivityExecutionOutcomeRequest(
 	}
 	return nil
 }
+
+// ValidateActivityTaskToken validates a task token against the current activity state.
+func ValidateActivityTaskToken(
+	ctx chasm.Context,
+	a *Activity,
+	token *tokenspb.Task,
+) error {
```
bergundy: Make it a method:

```suggestion
func (a *Activity) ValidateTaskToken(
	ctx chasm.Context,
	token *tokenspb.Task,
) error {
```

bergundy: And move it to `activity.go`?

tests/standalone_activity_test.go
```diff
@@ -1575,6 +1576,413 @@ func (s *standaloneActivityTestSuite) TestDescribeActivityExecution_InvalidArgum
...
+		// Worker accepts task
+		pollTaskResp, err := s.FrontendClient().PollActivityTaskQueue(ctx, &workflowservice.PollActivityTaskQueueRequest{
+			Namespace: s.Namespace().String(),
+			TaskQueue: &taskqueuepb.TaskQueue{Name: taskQueue, Kind: enumspb.TASK_QUEUE_KIND_NORMAL},
+		})
+		require.NoError(t, err)
+		require.NotEmpty(t, pollTaskResp.TaskToken)
+
+		// Heartbeat before timeout
+		time.Sleep(600 * time.Millisecond) //nolint:forbidigo
```
bergundy, line 1943: What value are you getting from adding those sleeps? They are just making the test flaky.

bergundy (approved):
Approved with comments.

tests/standalone_activity_test.go
dandavison: The sleeps allow us to prove that an attempt with duration longer than the heartbeat interval can succeed, due to the heartbeats. Without the sleep, the attempt would just immediately succeed and we would have demonstrated nothing about the heartbeat timeout.

I'd like to know of better ways to demonstrate that, but tests like that are a necessity while developing the feature. We could delete the test if there is no good way of testing this in a functional test.

chasm/lib/activity/validator.go
dandavison: Thanks, yes, agree with both, done.

chasm/lib/activity/activity_tasks.go
dandavison: Good call, refactored along those lines in ffcb33892.

chasm/lib/activity/activity.go
dandavison: Good point, moved and updated tests.

PR #8740, opened 2025-12-02
Add workflowId parameter to namespace.ActiveClusterName api.

xwduan:
- **init**
- **fix tests**

## What changed?
Add workflowId parameter to namespace.ActiveClusterName api

## Why?
So we can have different replication config for different workflowId.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
No risk.

### DIFF
--- a/chasm/interceptor_test.go
+++ b/chasm/interceptor_test.go
@@ -43,20 +43,35 @@ func (l *ServiceLibrary) RegisterServices(server *grpc.Server) {
 	testspb.RegisterTestServiceServer(server, ServiceHandler{})
 }
 
-func TestChasmRequestInterceptor_ShouldRespond(t *testing.T) {
+func TestChasmEngineInterceptor_ShouldRespond(t *testing.T) {
 	ctrl := gomock.NewController(t)
+
 	mockEngine := chasm.NewMockEngine(ctrl)
-	mockVisibilityManager := chasm.NewMockVisibilityManager(ctrl)
-	requestInterceptor := chasm.ChasmRequestInterceptorProvider(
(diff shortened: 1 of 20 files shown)


service/history/history_engine.go
```diff
@@ -370,7 +370,7 @@ func (e *historyEngineImpl) registerNamespaceStateChangeCallback() {
 
 		if ns.IsGlobalNamespace() &&
 			ns.ReplicationPolicy() == namespace.ReplicationPolicyMultiCluster &&
-			ns.ActiveClusterName() == e.currentClusterName {
+			ns.ActiveInCluster(e.currentClusterName) {
```
yux0, line 376: We need to think more about this. I am not sure if this is correct

common/rpc/interceptor/dc_redirection_policy.go
```diff
@@ -178,5 +179,5 @@ func (policy *SelectedAPIsForwardingRedirectionPolicy) getTargetClusterAndIsName
 		return policy.currentClusterName, false
 	}
 
-	return namespaceEntry.ActiveClusterName(), true
+	return namespaceEntry.ActiveClusterName(""), true
```
michaely520: emptyWorkflowId here?

service/history/history_engine.go
xwduan: From code semantic, it is correct, because ActiveInCluster underlying calling ns.ActiveClusterName api. From logic perspective, you are right to question this. This is to detect the when the current cluster become active. We will introduce ns.StateChanged API to handle this case instead of using ns.ActiveClusterName.

michaely520 (approved):

PR #8746, opened 2025-12-03
Use a read Context when checking for callbacks to close

spkane31:
When processing callbacks in mutable state use a read only context to validate there are callbacks to process, if there are get the writeable context to process the callbacks.

## What changed?
Use a read context to check for callbacks

## Why?
Removes a write of the root workflow component persisted / replicated upon close.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
NA, not rolled out for users yet

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -6421,6 +6421,17 @@ func (ms *MutableStateImpl) processCloseCallbacksHsm() error {
 
 // processCloseCallbacksChasm triggers "WorkflowClosed" callbacks using the CHASM implementation.
 func (ms *MutableStateImpl) processCloseCallbacksChasm() error {
+	wf, _, err := ms.ChasmWorkflowComponentReadOnly(context.Background())
+	if err != nil {
+		return err
+	}
+
+	// Return early if there are no chasm callbacks to process.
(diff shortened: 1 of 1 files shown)


yycptt (approved):

service/history/workflow/mutable_state_impl.go
```diff
@@ -6421,6 +6421,17 @@ func (ms *MutableStateImpl) processCloseCallbacksHsm() error {
 
 // processCloseCallbacksChasm triggers "WorkflowClosed" callbacks using the CHASM implementation.
 func (ms *MutableStateImpl) processCloseCallbacksChasm() error {
+	wf, _, err := ms.ChasmWorkflowComponentReadOnly(context.Background())
```
bergundy, line 6424: Wouldn't there be an `err` if the workflow component wasn't initialized?

yycptt: no... it's really a weird/undefined case TBH. but for now, an empty tree is assumed to be workflow and this will return an empty Workflow struct I believe.

bergundy: That seems wrong but as long as it works that unblocks us. Tests seems to pass so I guess this is okay.

PR #8747, opened 2025-12-03
Make the enableChasm flag per-NS instead of global

lina-temporal:
## What changed?
- The `enableChasm` dynamic config flag is now applied per-namespace.

## Why?
- For this release, we only want CHASM enabled in canary codepaths, for both callbacks and scheduler.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- This reduces the risk by reducing `enableChasm`'s blast radius. However, we'll need to be wary of writing tests that rely on `enableChasm` being set to `false` (current default), since in the future it'll be enabled everywhere.

### DIFF
--- a/service/history/tests/vars.go
+++ b/service/history/tests/vars.go
@@ -165,5 +165,6 @@ func NewDynamicConfig() *configs.Config {
 	config.ReplicationEnableUpdateWithNewTaskMerge = dynamicconfig.GetBoolPropertyFn(true)
 	config.EnableWorkflowIdReuseStartTimeValidation = dynamicconfig.GetBoolPropertyFnFilteredByNamespace(true)
 	config.EnableTransitionHistory = dynamicconfig.GetBoolPropertyFn(true)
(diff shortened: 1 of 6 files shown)


service/history/tests/vars.go
```diff
@@ -165,5 +165,6 @@ func NewDynamicConfig() *configs.Config {
 	config.ReplicationEnableUpdateWithNewTaskMerge = dynamicconfig.GetBoolPropertyFn(true)
 	config.EnableWorkflowIdReuseStartTimeValidation = dynamicconfig.GetBoolPropertyFnFilteredByNamespace(true)
 	config.EnableTransitionHistory = dynamicconfig.GetBoolPropertyFn(true)
+	config.EnableChasm = dynamicconfig.GetBoolPropertyFnFilteredByNamespace(false)
```
yycptt, line 168: hmm I don't think we need this? not sure if it will cause other tests to fail.

yycptt: I am blind, it's setting it to false.

PR #8748, opened 2025-12-03
Handle Duplicated Error for poll based replication

xwduan:
## What changed?
Handle Duplicated Error for poll based replication

## Why?
To ignore the task if the task is already applied.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
no risk.


service/history/replication/task_executor.go
```diff
@@ -239,6 +244,9 @@ func (e *taskExecutorImpl) handleHistoryReplicationTask(
 	// This might be extra cost if the workflow belongs to local shard.
 	// Add a wrapper of the history client to call history engine directly if it becomes an issue.
 	_, err = e.shardContext.GetHistoryClient().ReplicateEventsV2(ctx, request)
+	if errors.Is(err, consts.ErrDuplicate) {
```
hehaifengcn: should we move this to default below? (line 297)

PR #8750, opened 2025-12-03
fix: make release dep check job always run on PRs

stpierre:
## What changed?

Always run the release dependency check job on all PRs. When the PR is not against a release branch, it gracefully degrades to a no-op.

## Why?

This will let us make the status required, so that it can actually block releases that fail the job.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/.github/workflows/check-release-dependencies.yml
+++ b/.github/workflows/check-release-dependencies.yml
@@ -1,16 +1,24 @@
 name: Check Release Dependencies
 on:
-  pull_request:
-    branches:
-      - release/*
-      - cloud/*
+  pull_request: {}
+
+permissions:
+  contents: read
+
 jobs:
   check-dependencies:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/check-release-dependencies.yml
```diff
@@ -1,16 +1,24 @@
...
+  contents: read
+
 jobs:
   check-dependencies:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
+        if: >-
```
tdyas, line 14: Does the job still work as you intend if the `if` keys at the step level are moved to a single `if` key directly under `check-dependencies`? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idif

stpierre: It would not; in that case, no job would run, and the status would be `skipped`. Github Actions requires a `success` status; `skipped` counts as failed. We need a job to run, it just doesn't have to actually do anything 🙄

PR #8752, opened 2025-12-04
CHASM Visibility Manager

rodrigozhou:
## What changed?
Create CHASM Visibility Manager as a thin wrapper of VisibilityManager.
This is a refactor to separate the Visibility related API from the rest of the CHASM engine.

## Why?
Make it possible to be inject anywhere (history, frontend, etc), without needing to inject the whole CHASM engine.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/interceptor_test.go
+++ b/chasm/interceptor_test.go
@@ -44,8 +44,15 @@ func (l *ServiceLibrary) RegisterServices(server *grpc.Server) {
 }
 
 func TestChasmRequestInterceptor_ShouldRespond(t *testing.T) {
-	mockEngine := chasm.NewMockEngine(gomock.NewController(t))
-	requestInterceptor := chasm.ChasmRequestInterceptorProvider(mockEngine, log.NewNoopLogger(), metrics.NoopMetricsHandler)
+	ctrl := gomock.NewController(t)
+	mockEngine := chasm.NewMockEngine(ctrl)
+	mockVisibilityManager := chasm.NewMockVisibilityManager(ctrl)
+	requestInterceptor := chasm.ChasmRequestInterceptorProvider(
+		mockEngine,
+		mockVisibilityManager,
+		log.NewNoopLogger(),
+		metrics.NoopMetricsHandler,
+	)
 
 	server, address := startTestServer(t, grpc.UnaryInterceptor(requestInterceptor.Intercept))
 	defer server.Stop()
--- a/service/history/chasm_engine.go
+++ b/service/history/chasm_engine.go
@@ -4,9 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
 
-	commonpb "go.temporal.io/api/common/v1"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
 	enumsspb "go.temporal.io/server/api/enums/v1"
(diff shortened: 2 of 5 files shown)


chasm/visibility_manager.go
```diff
@@ -0,0 +1,156 @@
...
+// PageSize is required, must be greater than 0.
+// NextPageToken is optional, set on subsequent requests to continue listing the next page of executions.
+// Note: For CHASM executions, TemporalNamespaceDivision is the predefined search attribute
+// that is used to identify the archetype of the execution.
+// If the query string does not specify TemporalNamespaceDivision, the archetype C of the request will be used to filter the executions.
+// If the initial query already specifies TemporalNamespaceDivision, the archetype C of the request will
+// only be used to get the registered SearchAttributes.
+func ListExecutions[C Component, M proto.Message](
+	ctx context.Context,
+	request *ListExecutionsRequest,
```
bergundy, line 80: I missed this in the initial review, can we make this not a pointer? It's slightly nicer/safer to know that the arguments are copied when the function is called. Same for the response and count BTW.

chasm/visibility_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	"reflect"
+	"time"
+
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/common/payload"
+	"google.golang.org/protobuf/proto"
+)
+
+type VisibilityManager interface {
```
yycptt, line 16: nit: is it possible to call it like VisibilityEngine? `Manager` makes me think it's a persistence layer thing.

common/persistence/visibility/chasm_visibility_manager.go
```diff
@@ -0,0 +1,73 @@
...
+	"reflect"
+
+	commonpb "go.temporal.io/api/common/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/persistence/visibility/manager"
+)
+
+type ChasmVisibilityManager struct {
```
yycptt, line 14: is this the corresponding part of the methods implemetation in chasm_engine.go today?

yycptt (commented):
Do we also need to remove the List/CountExecution method from chasm_engine.go?

chasm/visibility_manager.go
yycptt: oh I see, this is a persistence layer object.

rodrigozhou: Yes, since we want to be able to inject in different places, I thought it would be better the implementation to be in the `common` folder. And this is basically a wrapper of `VisibilityManager`, I put in `common/persistence/visibility`.

common/persistence/visibility/chasm_visibility_manager.go
rodrigozhou: Yes! I forgot to delete from `chasm_engine.go`. I'll update the PR

awln-temporal (approved):

chasm/visibility_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	return visibilityManagerFromContext(ctx).CountExecutions(ctx, archetypeType, request)
+}
+
+type visibilityManagerCtxKeyType string
+
+const visibilityManagerCtxKey visibilityManagerCtxKeyType = "chasmVisibilityManager"
+
+func NewVisibilityManagerContext(
+	ctx context.Context,
+	engine VisibilityManager,
```
awln-temporal: nit: manager

chasm/visibility_manager.go
rodrigozhou: No strong opinions on that. Maybe a separate PR.
cc: @awln-temporal

PR #8754, opened 2025-12-04
Support CHASM Count Group By Filtering

awln-temporal:
Support CHASM Count "Group By" Filtering.

## What changed?
* Adds implementation of CHASM Count Executions `Group By` Filter support.
* Unifies `visibility_store.go` implementation to get count execution info.
* Adds functional tests for validation.

## Why?
Needed to unblock standalone activities, and fully support CHASM Count Execution requirements.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/common/persistence/visibility/store/elasticsearch/query_interceptors.go
+++ b/common/persistence/visibility/store/elasticsearch/query_interceptors.go
@@ -2,13 +2,13 @@ package elasticsearch
 
 import (
 	"fmt"
-	"maps"
 	"strconv"
 	"time"
 
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/persistence/visibility/store"
 	"go.temporal.io/server/common/persistence/visibility/store/query"
 	"go.temporal.io/server/common/primitives"
 	"go.temporal.io/server/common/searchattribute"
@@ -51,10 +51,7 @@ func NewValuesInterceptor(
 	csaTypeMap searchattribute.NameTypeMap,
 	chasmMapper *chasm.VisibilitySearchAttributesMapper,
 ) *valuesInterceptor {
-	combinedTypeMap := make(map[string]enumspb.IndexedValueType)
-	maps.Copy(combinedTypeMap, csaTypeMap.Custom())
-	maps.Copy(combinedTypeMap, chasmMapper.SATypeMap())
-	saTypeMap := searchattribute.NewNameTypeMap(combinedTypeMap)
+	saTypeMap := store.CombineTypeMaps(csaTypeMap, chasmMapper)
 	return &valuesInterceptor{
 		namespace:   namespaceName,
 		saTypeMap:   saTypeMap,
--- a/common/persistence/visibility/store/elasticsearch/visibility_store.go
+++ b/common/persistence/visibility/store/elasticsearch/visibility_store.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"maps"
 	"math"
 	"strconv"
 	"strings"
@@ -999,10 +998,7 @@ func (s *VisibilityStore) ParseESDoc(
 		return nil, serviceerror.NewInternalf("unable to unmarshal JSON from Elasticsearch document(%s): %v", docID, err)
 	}
 
-	combinedTypeMap := make(map[string]enumspb.IndexedValueType)
-	maps.Copy(combinedTypeMap, saTypeMap.Custom())
-	maps.Copy(combinedTypeMap, chasmMapper.SATypeMap())
-	finalTypeMap := searchattribute.NewNameTypeMap(combinedTypeMap)
+	combinedTypeMap := store.CombineTypeMaps(saTypeMap, chasmMapper)
 
 	var (
 		isValidType         bool
@@ -1034,7 +1030,7 @@ func (s *VisibilityStore) ParseESDoc(
 			continue
 		}
 
-		fieldType, err := finalTypeMap.GetType(fieldName)
+		fieldType, err := combinedTypeMap.GetType(fieldName)
 		if err != nil {
 			// Silently ignore ErrInvalidName because it indicates an unknown field in an Elasticsearch document.
 			if errors.Is(err, searchattribute.ErrInvalidName) {
@@ -1095,7 +1091,7 @@ func (s *VisibilityStore) ParseESDoc(
 	}
 
 	var err error
-	record.SearchAttributes, err = searchattribute.Encode(allSearchAttributes, &finalTypeMap)
+	record.SearchAttributes, err = searchattribute.Encode(allSearchAttributes, &combinedTypeMap)
 	if err != nil {
 		metrics.ElasticsearchDocumentParseFailuresCount.With(s.metricsHandler).Record(1)
 		return nil, serviceerror.NewInternalf(
@@ -1136,14 +1132,11 @@ func (s *VisibilityStore) parseCountGroupByResponse(
 		)
 	}
 
-	combinedTypeMap := make(map[string]enumspb.IndexedValueType)
-	maps.Copy(combinedTypeMap, saTypeMap.Custom())
-	maps.Copy(combinedTypeMap, chasmMapper.SATypeMap())
-	typeMap := searchattribute.NewNameTypeMap(combinedTypeMap)
+	combinedTypeMap := store.CombineTypeMaps(saTypeMap, chasmMapper)
 
 	groupByTypes := make([]enumspb.IndexedValueType, len(groupByFields))
 	for i, saName := range groupByFields {
-		tp, err := typeMap.GetType(saName)
+		tp, err := combinedTypeMap.GetType(saName)
 		if err != nil {
 			return nil, err
 		}
--- a/common/persistence/visibility/store/query/converter.go
+++ b/common/persistence/visibility/store/query/converter.go
@@ -129,22 +129,6 @@ var (
 	}
 )
 
-func IsGroupByFieldAllowed(fieldName string) bool {
-	for _, allowedField := range groupByFieldAllowlist {
-		if fieldName == allowedField {
-			return true
-		}
-	}
-
-	for _, allowedPrefix := range groupByFieldPrefixAllowlist {
-		if strings.HasPrefix(fieldName, allowedPrefix) {
-			return true
-		}
-	}
-
-	return false
-}
-
 func NewQueryConverter[ExprT any](
 	storeQC StoreQueryConverter[ExprT],
 	namespaceName namespace.Name,
@@ -726,6 +710,20 @@ func (c *QueryConverter[ExprT]) validateValueType(
 			saName,
 		)
 	}
+}
+
+func IsGroupByFieldAllowed(fieldName string) bool {
+	for _, allowedField := range groupByFieldAllowlist {
+		if fieldName == allowedField {
+			return true
+		}
+	}
+	for _, allowedPrefix := range groupByFieldPrefixAllowlist {
+		if strings.HasPrefix(fieldName, allowedPrefix) {
+			return true
+		}
+	}
+	return false
 }
 
 func parseExecutionStatusValue(value any) (string, error) {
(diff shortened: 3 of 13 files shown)


rodrigozhou (changes requested):
I think you messed the commit. I see schema changed unrelated to this PR.

tests/chasm_test.go
```diff
@@ -335,4 +335,100 @@ func (s *ChasmTestSuite) TestPayloadStoreVisibility() {
...
+	_, err = chasm.CountExecutions[*tests.PayloadStore](
+		ctx,
+		&chasm.CountExecutionsRequest{
+			NamespaceID:   string(s.NamespaceID()),
+			NamespaceName: s.Namespace().String(),
+			Query:         "GROUP BY `PayloadTotalCount`",
+		},
+	)
+	s.Error(err)
+	s.Contains(err.Error(), "GROUP BY")
```
yycptt, line 433: probably assert this is an InvalidArgument error

chasm/visibility_manager.go
```diff
@@ -61,7 +61,13 @@ type CountExecutionsRequest struct {
 }
 
 type CountExecutionsResponse struct {
-	Count int64
+	Count  int64
+	Groups []Group
+}
+
+type Group struct {
+	Values []any
```
yycptt: is this always string? If so we should just make it a string, or we can future proof and make the struct generic.

common/persistence/visibility/store/visibility_store.go
```diff
@@ -34,8 +34,8 @@ type (
 		// Read APIs.
 		ListWorkflowExecutions(ctx context.Context, request *manager.ListWorkflowExecutionsRequestV2) (*InternalListExecutionsResponse, error)
 		ListChasmExecutions(ctx context.Context, request *manager.ListChasmExecutionsRequest) (*InternalListExecutionsResponse, error)
-		CountWorkflowExecutions(ctx context.Context, request *manager.CountWorkflowExecutionsRequest) (*manager.CountWorkflowExecutionsResponse, error)
-		CountChasmExecutions(ctx context.Context, request *manager.CountChasmExecutionsRequest) (*manager.CountChasmExecutionsResponse, error)
+		CountWorkflowExecutions(ctx context.Context, request *manager.CountWorkflowExecutionsRequest) (*InternalCountExecutionsResponse, error)
```
yycptt, line 37: why do we need to change workflow? it looks like it introduces an extra conversion.

yycptt: ah I see why now, we are reusing the underlying implementation for count group by.

chasm/visibility_manager.go
awln-temporal: yea, i'll change it to string, was mainly concerned user wants to group by WorkflowExecutionStatus and the underlying storage is int. But, should be fine to just return a string version of the value

common/persistence/visibility/visibility_manager_impl.go
```diff
@@ -310,6 +311,75 @@ func (p *visibilityManagerImpl) AddSearchAttributes(
...
+	}
+
+	for _, group := range internal.Groups {
+		groupValues := make([]string, len(group.GroupValues))
+		for i, val := range group.GroupValues {
+			switch v := val.(type) {
+			case string:
+				groupValues[i] = v
+			case int: // needed to support legacy WorkflowExecutionStatus int type
+				groupValues[i] = strconv.Itoa(v)
```
yycptt: hmm we should probably convert it to the string representation of the workflowExecutionStatus, not the number as is.

cc @lina-temporal @bergundy I think this only applies to scheduler basically, want to understand how (if) scheduler will use the chasm count groupby.

chasm/visibility_manager.go
yycptt: do we have any type defined here so that we can do something like GetString(), []string is probably fine for now, am just thinking if we want to groupby SA with different types in the future.

common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -88,10 +88,9 @@ func (ni *nameInterceptor) Name(name string, usage query.FieldNameUsage) (string
 			)
 		}
 	case query.FieldNameGroupBy:
-		if fieldName != sadefs.ExecutionStatus {
+		if !query.IsGroupByFieldAllowed(fieldName) {
 			return "", query.NewConverterError(
-				"'group by' clause is only supported for %s search attribute",
-				sadefs.ExecutionStatus,
+				"'GROUP BY' clause is only supported for ExecutionStatus and LowCardinalityKeyword fields",
```
bergundy: Is this a user facing error? `LowCardinalityKeyword` is not a user facing concept.

chasm/visibility_manager.go
rodrigozhou: Currently, it's only a string, but I had set as `any` to future-proof when we start allowing to group by other search attributes types.

common/persistence/visibility/store/elasticsearch/visibility_store_read_test.go
```diff
@@ -1085,38 +1076,22 @@ func (s *ESVisibilitySuite) TestCountWorkflowExecutions_GroupBy() {
 	s.Contains(err.Error(), "'GROUP BY' clause supports only a single field")
 	s.Nil(resp)
 
-	// test only allowed to group by ExecutionStatus
+	// test only allowed to group by ExecutionStatus and LowCardinalityKeyword fields
 	request.Query = "GROUP BY WorkflowType"
 	resp, err = s.visibilityStore.CountWorkflowExecutions(context.Background(), request)
 	s.Error(err)
-	s.Contains(err.Error(), "'GROUP BY' clause is only supported for search attributes")
+	s.Contains(err.Error(), "'GROUP BY' clause is only supported for ExecutionStatus and LowCardinalityKeyword fields")
```
bergundy: nit:


```suggestion
	s.ErrorContains(err, "'GROUP BY' clause is only supported for ExecutionStatus and LowCardinalityKeyword fields")
```

chasm/visibility_manager.go
yycptt: or we can follow what we have for `GetValue[T any](SearchAttributesMap, typedSearchAttribute[T])`, and similarly have a `GetValue[T](VisibilityValue, typeSA[T])` to be future proof. 

cc @bergundy if you have any opinions here.

btw I don't think this discussion need to block the PR, using []string is good enough for now. We can change all the call sites if we change the interface.

common/persistence/visibility/visibility_manager_impl.go
bergundy: Do we even need to support workflow execution status? Ideally the scheduler itself uses the new low cardinality status field to decouple from workflow execution status. For listing both workflow and chasm schedulers, I'm still not sure how we that will work. Maybe @lina-temporal has given it some thought...

chasm/visibility_manager.go
rodrigozhou: Not sure I get your suggestion. To be clear, this `Values []any` is the group by key values. Eg: if you do `GROUP BY ExecutionStatus`, then `Values = []any{"Running"}` or `Values = []any{"Completed"}`, etc. If you could do `GROUP BY ExecutionStatus, CustomIntField`, then you would see things like `Values = []any{"Running", 2}` or `Values = []any{"Failed", 1}`.

yycptt: yeah I mean it can be `Value []VisibilityValue` and caller use GetValue[string](VisibilityValue, typeSA) to get the strongly typed value.

rodrigozhou: Oh, that works for me as well.

chasm/visibility_manager.go
```diff
@@ -61,7 +61,13 @@ type CountExecutionsRequest struct {
 }
 
 type CountExecutionsResponse struct {
-	Count int64
+	Count  int64
+	Groups []Group
+}
+
+type Group struct {
+	Values []string
```
bergundy: In the public API we are using:

```
    message AggregationGroup {
        repeated temporal.api.common.v1.Payload group_values = 1;
        int64 count = 2;
    }
```

It would be more convenient for the CHASM API to also return payloads.

common/persistence/visibility/store/elasticsearch/query_interceptors.go
awln-temporal: hm, how important is it that the error message is accurate? from the store and below, they are unified. If we're ok with the error message only specifying Workflow ExecutionStatus, I can revert this.

common/persistence/visibility/visibility_manager_impl.go
awln-temporal: if we're ok with encoded payloads as Group values, we can support either workflow execution status or low cardinality status from chasm visibility.

chasm/visibility_manager.go
awln-temporal: summarizing from offline discussion, we agreed to return payloads from the visibility manager

common/persistence/visibility/visibility_manager_impl.go
lina-temporal: > For listing both workflow and chasm schedulers, I'm still not sure how we that will work. Maybe @lina-temporal has given it some thought...

`ListSchedules` will only return running schedules to the caller, without execution status info on the schedule itself, so I think the only concern is around the ES query returning both simultaneously. Prior to this change, we're doing this:

```go
var VisibilityBaseListQuery = fmt.Sprintf(
	"%s IN ('%s', '%d') AND %s = '%s'",
	sadefs.TemporalNamespaceDivision,
	NamespaceDivision,
	chasm.SchedulerArchetypeID,
	sadefs.ExecutionStatus,
	enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING.String(),
)
```

Is there a reason `.. AND (HighCardinalityStatus = 'RUNNING' OR LowCardinalityStatus = 'RUNNING')` wouldn't work? 

> cc @lina-temporal @bergundy I think this only applies to scheduler basically, want to understand how (if) scheduler will use the chasm count groupby.

We don't do any count queries, just `ListWorkflowExecutions` (as part of `ListSchedules`).

bergundy: Yeah, I later realized that there's no count schedules API. We will probably want to add that with the group by feature for the UI when we take on some of the scheduler enhancements.

common/persistence/visibility/store/elasticsearch/visibility_store.go
```diff
@@ -1123,14 +1126,21 @@ func (s *VisibilityStore) ParseESDoc(
...
 	if err != nil {
 		return nil, serviceerror.NewUnavailablef(
 			"unable to read search attribute types: %v", err,
 		)
 	}
+
+	combinedTypeMap := make(map[string]enumspb.IndexedValueType)
+	maps.Copy(combinedTypeMap, saTypeMap.Custom())
+	maps.Copy(combinedTypeMap, chasmMapper.SATypeMap())
+	typeMap := searchattribute.NewNameTypeMap(combinedTypeMap)
```
rodrigozhou: This pattern is showing up quite a lot, might be good to have util function to combine type maps.

common/persistence/visibility/store/query/converter.go
```diff
@@ -125,6 +129,22 @@ var (
...
+	}
+
+	for _, allowedPrefix := range groupByFieldPrefixAllowlist {
+		if strings.HasPrefix(fieldName, allowedPrefix) {
+			return true
+		}
+	}
+
+	return false
+}
```
rodrigozhou: Util functions are usually added at the end of file.

common/persistence/visibility/visibility_manager_impl.go
```diff
@@ -310,6 +311,45 @@ func (p *visibilityManagerImpl) AddSearchAttributes(
...
+	if len(internal.Groups) > 0 {
+		response.Groups = make([]*workflowservice.CountWorkflowExecutionsResponse_AggregationGroup, 0, len(internal.Groups))
+	}
+
+	for _, group := range internal.Groups {
+		response.Groups = append(response.Groups, &workflowservice.CountWorkflowExecutionsResponse_AggregationGroup{
+			GroupValues: group.GroupValues,
+			Count:       group.Count,
+		})
+	}
```
rodrigozhou: This could be inside the `if` block

PR #8755, opened 2025-12-04
Fix stale namespace name cache entries after rename

jprieto-temporal:
## What changed?
Renaming is a possible operation on namespaces (see [MetadataManager](https://github.com/temporalio/temporal/blob/c0aaf5d1cb932d6e1295e916d6ecba7896f3711e/common/persistence/metadata_manager.go#L95)). When a namespace is renamed, the namespace registry now removes the old name from the `nameToID` map. This fix applies to both `refreshNamespaces` (background refresh) and `updateSingleNamespace` (readthrough/explicit refresh) code paths.

## Why?
Previously, renaming a namespace would add the new name to the cache but leave the old name pointing to the namespace ID. This caused stale name->ID mappings to persist, allowing lookups by the old name to incorrectly succeed.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
Low risk. The change adds a delete operation to the nameToID map only when a rename is detected (old and new names differ). The delete occurs in the same critical section where the map is already being modified.


common/namespace/nsregistry/registry.go
```diff
@@ -378,6 +378,10 @@ func (r *registry) refreshNamespaces(ctx context.Context) error {
 	var stateChanged []*namespace.Namespace
 	for _, aNamespace := range namespacesDb {
 		oldNS := r.updateIDToNamespace(newIDToNamespace, aNamespace.ID(), aNamespace)
+		// If namespace was renamed, remove entry for the old name
+		if oldNS != nil && oldNS.Name() != aNamespace.Name() {
+			delete(newNameToID, oldNS.Name())
```
dnr, line 383: I assume we don't have to worry about a situation where B was renamed to A in the same update that C was renamed to B. I think you have to get fancier to handle that, but I'm not sure it's necessary.

Also, should we count Name changes for namespaceStateChanged? It seems safer to do that. Looks like it marginally simplifies the test too.

PR #8759, opened 2025-12-04
Add standalone activity metrics

fretz12:
## What changed?
Add standalone activity metrics

## Why?
Need standalone activity metrics for observability purposes

## How did you test it?
- [X] built
- [X] added new unit test(s)

---

> [!NOTE]
> Adds comprehensive metrics and payload-size recording for standalone activities, introducing contextual request wrappers and wiring metrics through state transitions, timeout executors, and history APIs.
> 
> - **Activity component/state machine**:
>   - Introduces `RequestWithContext` carrying `Token`, `MetricsHandler`, `NamespaceName`, and `BreakdownMetricsByTaskQueue` and replaces prior request wrappers.
>   - Records metrics on schedule, attempts, and terminal transitions (success, fail, cancel, timeouts), including `ActivityStartToCloseLatency`, `ActivityScheduleToCloseLatency`, success/fail/cancel/timeout counters, and per-timeout tags.
>   - Adds `recordPayloadSize(...)` to emit payload sizes for input, heartbeat details, results, and failures.
> - **Timeout and dispatch executors (`chasm/lib/activity/activity_tasks.go`)**:
>   - Add `timeoutTaskExecutorOptions` (dynamic config, metrics, namespace registry); resolve namespace and emit timeout metrics during schedule/start/close/heartbeat timeouts and retries.
> - **Activity handler**:
>   - Injects `metrics.Handler` and `namespace.Registry`; `StartActivityExecution` emits input payload size on scheduling.
> - **History APIs (record/respond activity ops)**:
>   - Pass `RequestWithContext` (token, metrics handler, namespace, breakdown setting) into chasm component calls.
> - **Tests**:
>   - Update/add unit tests to validate metric emissions and payload-size recording across transitions and timeouts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 02c6abafb29e0438c4cbad319919ed3dc4973f98. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/common/dynamicconfig/setting_gen.go
+++ b/common/dynamicconfig/setting_gen.go
@@ -924,10 +924,8 @@ func (s NamespaceTypedSetting[T]) Validate(v any) error {
 	return err
 }
 
-func (s NamespaceTypedConstrainedDefaultSetting[T]) Key() Key { return s.key }
-func (s NamespaceTypedConstrainedDefaultSetting[T]) Precedence() Precedence {
-	return PrecedenceNamespace
-}
+func (s NamespaceTypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
+func (s NamespaceTypedConstrainedDefaultSetting[T]) Precedence() Precedence { return PrecedenceNamespace }
 func (s NamespaceTypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
@@ -1062,10 +1060,8 @@ func (s NamespaceIDTypedSetting[T]) Validate(v any) error {
 	return err
 }
 
-func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Key() Key { return s.key }
-func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Precedence() Precedence {
-	return PrecedenceNamespaceID
-}
+func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
+func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Precedence() Precedence { return PrecedenceNamespaceID }
 func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
@@ -1200,10 +1196,8 @@ func (s TaskQueueTypedSetting[T]) Validate(v any) error {
 	return err
 }
 
-func (s TaskQueueTypedConstrainedDefaultSetting[T]) Key() Key { return s.key }
-func (s TaskQueueTypedConstrainedDefaultSetting[T]) Precedence() Precedence {
-	return PrecedenceTaskQueue
-}
+func (s TaskQueueTypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
+func (s TaskQueueTypedConstrainedDefaultSetting[T]) Precedence() Precedence { return PrecedenceTaskQueue }
 func (s TaskQueueTypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
@@ -1634,10 +1628,8 @@ func (s DestinationTypedSetting[T]) Validate(v any) error {
 	return err
 }
 
-func (s DestinationTypedConstrainedDefaultSetting[T]) Key() Key { return s.key }
-func (s DestinationTypedConstrainedDefaultSetting[T]) Precedence() Precedence {
-	return PrecedenceDestination
-}
+func (s DestinationTypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
+func (s DestinationTypedConstrainedDefaultSetting[T]) Precedence() Precedence { return PrecedenceDestination }
 func (s DestinationTypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
@@ -1738,3 +1730,4 @@ func GetTypedPropertyFnFilteredByDestination[T any](value T) TypedPropertyFnWith
 		return value
 	}
 }
+


chasm/lib/activity/activity.go
```diff
@@ -54,10 +58,13 @@ type Activity struct {
 	Store chasm.Field[ActivityStore]
 }
 
-// WithToken wraps a request with its deserialized task token.
-type WithToken[R any] struct {
-	Token   *tokenspb.Task
-	Request R
+// RequestWithContext wraps a request context specific metadata.
+type RequestWithContext[R any] struct {
```
fretz12: Using @dandavison naming. Open to better ideas.

fretz12:
cursor review

chasm/lib/activity/activity.go
```diff
@@ -54,10 +58,13 @@ type Activity struct {
...
+	Request                     R
+	Token                       *tokenspb.Task
```
dandavison, line 93: This is specific to certain worker APIs isn't it. So I think it's starting to feel a bit hacky to share the same wrapper across multiple APIs that only need some things.

chasm/lib/activity/activity.go
```diff
@@ -54,10 +58,13 @@ type Activity struct {
...
+	MetricsHandler              metrics.Handler
+	NamespaceName               namespace.Name
+	BreakdownMetricsByTaskQueue dynamicconfig.BoolPropertyFnWithTaskQueueFilter
```
dandavison: Is it definitely appropriate to pass this; DC can be queried where needed, right?

chasm/lib/activity/activity.go
```diff
@@ -54,10 +58,13 @@ type Activity struct {
 	Store chasm.Field[ActivityStore]
 }
 
-// WithToken wraps a request with its deserialized task token.
-type WithToken[R any] struct {
-	Token   *tokenspb.Task
-	Request R
+// RequestWithContext wraps a request context specific metadata.
```
```suggestion
// RequestWithContext wraps a request with context-specific metadata.
```

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
 		FrontendResponse: response,
 	}, nil
 }
+
+// recordOnAttemptedMetrics records metrics for attempted activities, including retries and originating from any
+// terminal state transitions.
```
dandavison: I don't quite follow the "and originating from any terminal state transitions" bit.

For the first bit, would "records metrics for an activity attempt" be clearer?

chasm/lib/activity/statemachine.go
```diff
@@ -120,6 +136,14 @@ var TransitionRescheduled = chasm.NewTransition(
 				Attempt: attempt.GetCount(),
 			})
 
+		a.recordOnAttemptedMetrics(
+			attempt.GetStartedTime().AsTime(),
+			event.namespace.String(),
+			event.handler,
+			event.breakdownMetricsByTaskQueue,
+			event.operationTag,
+			event.timeoutType)
```
dandavison: I'm finding it odd that we're referencing `GetStartedTime` time in a transition to `Scheduled`.

I don't think `attempt.GetStartedTime().AsTime()` is guaranteed to be zero here, right? So then we'll be emitting `ActivityStartToCloseLatency` which would be wrong?

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
...
+	operationTag string,
+	timeoutType enumspb.TimeoutType,
+) {
+	a.recordOnAttemptedMetrics(
+		startedTime,
+		namespaceName,
+		metricsHandler,
+		breakdownMetricsByTaskQueue,
+		operationTag,
+		timeoutType)
```
dandavison: Does this not mean that `recordOnAttemptedMetrics()` gets called twice, since it is called when a retry transitions to `SCHEDULED`?

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
...
+		namespaceName,
+		tqid.UnsafeTaskQueueFamily(namespaceName, taskQueueFamily),
+		breakdownMetricsByTaskQueue(namespaceName, taskQueueFamily, enumspb.TASK_QUEUE_TYPE_ACTIVITY),
+		metrics.OperationTag(operationTag),
+		metrics.ActivityTypeTag(a.GetActivityType().GetName()),
+		// metrics.VersioningBehaviorTag(versioningBehavior), TODO add when we have versioning
+	)
+
+	scheduleToCloseLatency := time.Since(a.GetScheduledTime().AsTime())
+	metrics.ActivityScheduleToCloseLatency.With(metricsHandler).Record(scheduleToCloseLatency)
```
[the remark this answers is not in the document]
fretz12: fixed

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
 		FrontendResponse: response,
 	}, nil
 }
+
+// recordOnAttemptedMetrics records metrics for attempted activities, including retries and originating from any
+// terminal state transitions.
+func (a *Activity) recordOnAttemptedMetrics(
```
dandavison: It's not urgent but when we get time I'd like us to think about a consistent and sensible mapping of events to method names. I am already finding methods named "handleX" and "recordX" confusingly similar, and now we're adding "recordX" with a second meaning.

chasm/lib/activity/activity.go
bergundy: You need to pass this in.

chasm/lib/activity/activity.go
bergundy: I would create a struct per request and avoid this generic wrapper.

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
...
+	startedTime time.Time,
+	namespaceName string,
+	metricsHandler metrics.Handler,
+	breakdownMetricsByTaskQueue dynamicconfig.BoolPropertyFnWithTaskQueueFilter,
+	operationTag string,
+	timeoutType enumspb.TimeoutType,
+) {
+	taskQueueFamily := a.GetTaskQueue().GetName()
+
+	handler := metrics.GetPerTaskQueueFamilyScope(
```
bergundy: You can build this handler outside of the component logic and simplify testing.

chasm/lib/activity/activity.go
bergundy: Suggesting the verb `emit` here.

chasm/lib/activity/activity.go
bergundy: I would suggest passing in a fully baked metrics handler into the application logic.

You can build the metrics handler in the API handlers or task executors if needed, it will save you from carrying all of these parameters around.

chasm/lib/activity/activity.go
```diff
@@ -509,15 +520,21 @@ func createHeartbeatTimeoutFailure() *failurepb.Failure {
...
+
+	details := req.Request.HeartbeatRequest.GetDetails()
+
 	a.LastHeartbeat = chasm.NewDataField(ctx, &activitypb.ActivityHeartbeatState{
 		RecordedTime: timestamppb.New(ctx.Now(a)),
-		Details:      input.Request.HeartbeatRequest.GetDetails(),
+		Details:      details,
 	})
+
+	recordPayloadSize(details.Size(), req.MetricsHandler, req.NamespaceName.String(), metrics.HistoryRecordActivityTaskHeartbeatScope)
```
bergundy: We don't need these payload size metrics IMHO, we added them to understand the implications of putting payloads in mutable state for standalone activities.

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
...
+	taskQueueFamily := a.GetTaskQueue().GetName()
+
+	handler := metrics.GetPerTaskQueueFamilyScope(
+		metricsHandler,
+		namespaceName,
+		tqid.UnsafeTaskQueueFamily(namespaceName, taskQueueFamily),
+		breakdownMetricsByTaskQueue(namespaceName, taskQueueFamily, enumspb.TASK_QUEUE_TYPE_ACTIVITY),
+		metrics.OperationTag(operationTag),
+		metrics.ActivityTypeTag(a.GetActivityType().GetName()),
+		// metrics.VersioningBehaviorTag(versioningBehavior), TODO add when we have versioning
```
bergundy: You can't not emit this tag here, all metrics have to have the same tags no matter where they're omitted from.

chasm/lib/activity/activity.go
```diff
@@ -679,3 +696,103 @@ func (a *Activity) buildPollActivityExecutionResponse(
...
+	)
+
+	scheduleToCloseLatency := time.Since(a.GetScheduledTime().AsTime())
+	metrics.ActivityScheduleToCloseLatency.With(handler).Record(scheduleToCloseLatency)
+
+	switch operationTag {
```
bergundy: I think it would be much easier to follow this code if you emitted the appropriate metric where it is relevant.

chasm/lib/activity/activity_tasks.go
```diff
@@ -70,12 +73,24 @@ func (e *activityDispatchTaskExecutor) Execute(
 	return err
 }
 
+type timeoutTaskExecutorOptions struct {
+	fx.In
+
+	Dc                *dynamicconfig.Collection
```
bergundy: Use the config struct that Dan added in a separate PR and initialize it in fx instead of using the collection directly.

chasm/lib/activity/statemachine.go
```diff
@@ -70,13 +79,20 @@ var TransitionScheduled = chasm.NewTransition(
 				Attempt: attempt.GetCount(),
 			})
 
+		recordPayloadSize(event.inputSize, event.handler, event.namespace.String(), metrics.HistoryRecordActivityTaskStartedScope)
```
bergundy: Same here, not required to record this IMHO.

bergundy (commented):
Overall this looks really good. I would consider restructuring the code as I suggested to make it easier to follow.

chasm/lib/activity/activity_tasks.go
fretz12: Awaiting heartbeat branch rebase on standalone-activity, so added TODO. Will do before this PR merges.

chasm/lib/activity/activity.go
fretz12: defaulting to VERSIONING_BEHAVIOR_UNSPECIFIED unless you think that's wrong

chasm/lib/activity/activity.go
fretz12: emit sounds appropriate

chasm/lib/activity/activity.go
fretz12: Good, they've been removed.

chasm/lib/activity/statemachine.go
fretz12: removed

chasm/lib/activity/statemachine.go
fretz12: This has been refactored to individual instead of shared methods. 

> I don't think attempt.GetStartedTime().AsTime() is guaranteed to be zero here, right?

In my refactored code, this should be guranteed as it happens after a reschedule transition and that should guarantee the activity is started, so check is removed.

chasm/lib/activity/activity.go
fretz12: Refactored so no longer relevant

chasm/lib/activity/activity.go
fretz12: Done

chasm/lib/activity/activity.go
fretz12: Refactored to request specific wrappers

chasm/lib/activity/activity.go
fretz12: Refactored so that handlers are enriched at the service APIs, so no longer relevant.

chasm/lib/activity/activity.go
fretz12: Refactored so no longer relevant

chasm/lib/activity/activity.go
fretz12: done

chasm/lib/activity/activity.go
fretz12: the intention is that such metrics will get recorded on every attempt, including that induced by retries. This has been refactored FYI.

chasm/lib/activity/activity.go
fretz12: refactored

chasm/lib/activity/activity_tasks.go

chasm/lib/activity/activity.go
```diff
@@ -61,6 +62,39 @@ type WithToken[R any] struct {
...
+type requestCancelActivityReqWrapper struct {
+	request        *activitypb.RequestCancelActivityExecutionRequest
+	metricsHandler metrics.Handler
+}
+
+// terminateActivityReqWrapper wraps the TerminateActivityExecutionRequest with context-specific data.
+type terminateActivityReqWrapper struct {
+	request        *activitypb.TerminateActivityExecutionRequest
+	metricsHandler metrics.Handler
+}
```
bergundy: Is there a reason that some of these are exported and others aren't?

chasm/lib/activity/activity.go
```diff
@@ -61,6 +62,39 @@ type WithToken[R any] struct {
 	Request R
 }
 
+// RespondCompletedReqWrapper wraps the RespondActivityTaskCompletedRequest with context-specific data.
+type RespondCompletedReqWrapper struct {
```
bergundy: I think the name input or event is better here. Even if you think RequestWrapper works better, please don't abbreviate, we shouldn't worry about typing a few more characters and prefer readability.

```suggestion
type RespondCompletedInput struct {
```

chasm/lib/activity/activity_tasks.go
```diff
@@ -81,13 +96,41 @@ func (e *scheduleToStartTimeoutTaskExecutor) Execute(
...
+	breakdownMetricsByTaskQueue := e.opts.Config.BreakdownMetricsByTaskQueue
+	nsID := namespace.ID(ctx.ExecutionKey().NamespaceID)
+	namespaceName, err := e.opts.NamespaceRegistry.GetNamespaceName(nsID)
+	if err != nil {
+		return err
+	}
+
+	taskQueueFamily := activity.GetTaskQueue().GetName()
+
+	metricsHandler := metrics.GetPerTaskQueueFamilyScope(
```
bergundy: I think it's worth trying to abstract this and avoid repeating the logic everywhere we construct the handler.

chasm/lib/activity/activity_tasks.go
```diff
@@ -81,13 +96,41 @@ func (e *scheduleToStartTimeoutTaskExecutor) Execute(
 	_ chasm.TaskAttributes,
 	_ *activitypb.ScheduleToStartTimeoutTask,
 ) error {
-	return TransitionTimedOut.Apply(activity, ctx, enumspb.TIMEOUT_TYPE_SCHEDULE_TO_START)
+	breakdownMetricsByTaskQueue := e.opts.Config.BreakdownMetricsByTaskQueue
+	nsID := namespace.ID(ctx.ExecutionKey().NamespaceID)
+	namespaceName, err := e.opts.NamespaceRegistry.GetNamespaceName(nsID)
+	if err != nil {
+		return err
+	}
```
bergundy: See comment below, this can be part of the abstraction.

chasm/lib/activity/handler.go
```diff
@@ -226,11 +232,37 @@ func (h *handler) TerminateActivityExecution(
 		RunID:       frontendReq.GetRunId(),
 	})
 
+	breakdownMetricsByTaskQueue := h.config.BreakdownMetricsByTaskQueue
+	namespaceName, err := h.namespaceRegistry.GetNamespaceName(namespace.ID(req.GetNamespaceId()))
+	if err != nil {
+		return nil, err
+	}
+
+	metricsHandlerBuilderParams, err := chasm.ReadComponent(ctx, ref, (*Activity).GetMetricsHandlerParams, nil)
```
bergundy: It would be better to not have to read the component separately. You can wrap `(*Activity).handleTerminated` in a closure that will build everything that you need.

chasm/lib/activity/statemachine.go
```diff
@@ -203,19 +218,22 @@ var TransitionTerminated = chasm.NewTransition(
...
+				Message:     reqWrapper.request.GetFrontendRequest().GetReason(),
 				FailureInfo: &failurepb.Failure_TerminatedFailureInfo{},
 			}
 			outcome.Variant = &activitypb.ActivityOutcome_Failed_{
 				Failed: &activitypb.ActivityOutcome_Failed{
 					Failure: failure,
 				},
 			}
+
+			a.emitOnCanceledMetrics(ctx, reqWrapper.metricsHandler)
```
bergundy: This is the wrong metric.

chasm/lib/activity/activity_tasks.go
```diff
@@ -208,11 +313,43 @@ func (e *heartbeatTimeoutTaskExecutor) Execute(
 	if err != nil {
 		return err
 	}
+
+	breakdownMetricsByTaskQueue := e.opts.Config.BreakdownMetricsByTaskQueue
```
bergundy: Isn't this check required for API handlers too?

chasm/lib/activity/library.go
```diff
@@ -59,11 +59,12 @@ func (l *library) RegisterServices(server *grpc.Server) {
 
 func (l *library) Tasks() []*chasm.RegistrableTask {
 	return []*chasm.RegistrableTask{
-		chasm.NewRegistrableSideEffectTask[*Activity, *activitypb.ActivityDispatchTask](
+		chasm.NewRegistrableSideEffectTask(
 			"dispatch",
 			l.activityDispatchTaskExecutor,
 			l.activityDispatchTaskExecutor,
 		),
+		// TODO(dan): why are the task names "FooTimer" but "FooTimeoutTask" in the struct names?
```
bergundy: Can we delete this TODO?

chasm/lib/activity/activity.go
```diff
@@ -137,6 +171,20 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
...
 }
 
+// MetricsHandlerBuilderParams contains parameters for building a metrics handler for activity operations.
+type MetricsHandlerBuilderParams struct {
+	ActivityType  string
+	TaskQueueName string
+}
+
+// GetMetricsHandlerParams retrieves parameters for building a metrics handler for activity operations.
+func (a *Activity) GetMetricsHandlerParams(_ chasm.Context, _ any) (MetricsHandlerBuilderParams, error) {
```
bergundy: Mentioned elsewhere, replace this with a method that will construct a metrics handler.

chasm/lib/activity/activity.go
```diff
@@ -635,3 +693,98 @@ func (a *Activity) StoreOrSelf(ctx chasm.Context) ActivityStore {
...
+	scheduleToCloseLatency := time.Since(a.GetScheduleTime().AsTime())
+	metrics.ActivityScheduleToCloseLatency.With(handler).Record(scheduleToCloseLatency)
+
+	metrics.ActivityTaskFail.With(handler).Record(1)
+	metrics.ActivityFail.With(handler).Record(1)
+}
+
+func (a *Activity) emitOnCanceledMetrics(ctx chasm.Context, handler metrics.Handler) {
+	attempt := a.LastAttempt.Get(ctx)
+	startedTime := attempt.GetStartedTime().AsTime()
```
bergundy: Couldn't this be set from the previous attempt? I thought we said we would preserve previous attempt information until a new attempt is made?

chasm/lib/activity/activity.go
```diff
@@ -635,3 +693,98 @@ func (a *Activity) StoreOrSelf(ctx chasm.Context) ActivityStore {
...
+
+	// Terminate can happen before start, so guard against zero time
+	if !startedTime.IsZero() {
+		startToCloseLatency := time.Since(startedTime)
+		metrics.ActivityStartToCloseLatency.With(handler).Record(startToCloseLatency)
+	}
+
+	scheduleToCloseLatency := time.Since(a.GetScheduleTime().AsTime())
+	metrics.ActivityScheduleToCloseLatency.With(handler).Record(scheduleToCloseLatency)
```
bergundy, line 823: This is technically incorrect. An attempt didn't complete, we shouldn't emit start-to-close latency.

common/metrics/metric_defs.go
```diff
@@ -862,6 +862,7 @@ var (
 	ActivityFail                                         = NewCounterDef("activity_fail", WithDescription("Number of activities that failed and won't be retried anymore."))
 	ActivityTaskFail                                     = NewCounterDef("activity_task_fail", WithDescription("Number of activity task failures (includes retries)."))
 	ActivityCancel                                       = NewCounterDef("activity_cancel")
+	ActivityTerminate                                    = NewCounterDef("activity_terminate")
```
bergundy: Can you add descriptions to both of these metrics?
Mention that terminate is only available for standalone activities.

service/history/api/respondactivitytaskcanceled/api.go
```diff
@@ -40,13 +41,31 @@ func Invoke(
...
 	// Handle standalone activity if component ref is present in the token
 	if componentRef := token.GetComponentRef(); len(componentRef) > 0 {
+		metricsHandlerBuilderParams, err := chasm.ReadComponent(ctx, componentRef, (*activity.Activity).GetMetricsHandlerParams, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		taskQueueName := metricsHandlerBuilderParams.TaskQueueName
+
+		handler := metrics.GetPerTaskQueueFamilyScope(
```
bergundy: For any metric you emit here make sure it has the exact same labels as emitted for workflow activities.
I see you are missing `WorkflowTypeTag`, it should be set to `__temporal_standalone_activity__`. Note that for the terminate metric, we don't need to emit that tag but you can emit it for completeness.

chasm/lib/activity/activity.go
fretz12: Those that are accessed only within the activity package are private

chasm/lib/activity/library.go
fretz12: Removed

chasm/lib/activity/statemachine.go
fretz12: Oops, test was expecting cancel too... fixed in both places

chasm/lib/activity/activity.go
fretz12: ok that's true, I'll remove these latency metrics. Good catch.

chasm/lib/activity/activity.go
fretz12: hmm, I can think of mainly this scenario:
- Activity in `ACTIVITY_EXECUTION_STATUS_SCHEDULED` due to retry. Start time recorded from previous attempt.
- Cancel request comes in before next retry. Status is now in ACTIVITY_EXECUTION_STATUS_CANCEL_REQUESTED
- Respond cancelation comes in, and hits this line in the code. 
I think Ignore start to complete in such a case? 

FYI, I don't think existing WF activities is doing any special logic: 
https://github.com/temporalio/temporal/blob/f383ea4b02a0450d024fa9840489852bad4552a9/service/history/workflow/metrics.go#L176

fretz12: We need to tee off on attempt.LastCompleted, and make sure that gets cleared out on every start

common/metrics/metric_defs.go
fretz12: Done

chasm/lib/activity/activity.go
fretz12: Changed to *Event since we already started using that term

service/history/api/respondactivitytaskcanceled/api.go
fretz12: ok, didn't think we needed WorkflowTypeTag since this is SAA, but I'll set it as you mentioned

chasm/lib/activity/handler.go
fretz12: Refactored

chasm/lib/activity/activity.go
fretz12: done

chasm/lib/activity/activity_tasks.go

chasm/lib/activity/activity_tasks.go
fretz12: refactored into a common method

chasm/lib/activity/activity_tasks.go
fretz12: Yes, it's used in some API handlers. I'm not quite following...what's the ask and did I miss something?

chasm/lib/activity/activity.go
```diff
@@ -20,11 +20,16 @@ import (
...
 	"go.temporal.io/server/common/backoff"
+	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/payload"
+	"go.temporal.io/server/common/tqid"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+const WorkflowTypeTag = "__temporal_standalone_activity__"
```
bergundy, line 31: Add a docstring explaining that this is a sentinel value that is used to ensure workflow and standalone activity metrics have the same label.

chasm/lib/activity/activity.go
```diff
@@ -137,6 +175,12 @@ func (a *Activity) createAddActivityTaskRequest(ctx chasm.Context, namespaceID s
 	}, nil
 }
 
+// MetricsHandlerBuilderParams contains parameters for building a metrics handler for activity operations.
+type MetricsHandlerBuilderParams struct {
```
bergundy: Unused?

chasm/lib/activity/handler.go
```diff
@@ -253,11 +278,30 @@ func (h *handler) RequestCancelActivityExecution(
...
+	}
+
 	response, _, err = chasm.UpdateComponent(
 		ctx,
 		ref,
 		(*Activity).handleCancellationRequested,
-		req,
+		requestCancelEvent{
+			request: req,
+			handlerBuilder: func(activityType string, taskQueueName string) metrics.Handler {
```
bergundy: This can take `*Activity` and you can share the code for generating this.

chasm/lib/activity/activity.go
bergundy: But discussing this again, we said we would just pass in this in all of the "events" and maybe also include the metrics handler.

chasm/lib/activity/activity.go
fretz12: ok, based on our discussion of using a `fromStatus` and thinking through the cases, I think the cases to emit start-to-close latency covered now

chasm/lib/activity/activity.go
fretz12: Added

chasm/lib/activity/activity.go
fretz12: Refactored as we discussed.

chasm/lib/activity/handler.go
fretz12: refactored and done

PR #8768, opened 2025-12-05
CHASM: Use force deletion for chasm execution in namespace deletion

yycptt:
## What changed?
- Use force deletion for chasm execution in namespace deletion
- Add functional test for forcing deleting chasm executions
- We have log for force deletion as well

## Why?
- DeleteWorkflowExecution API is an async graceful deletion api that only works for Workflow, but not other CHASM archetypes. 
- Upon namespace deletion we need to delete chasm executions as well and we can use force deletion to do it.

## How did you test it?
- [x] built
- [x] added new unit test(s)


service/worker/deletenamespace/deleteexecutions/activities.go
```diff
@@ -179,10 +182,35 @@ func (a *Activities) DeleteExecutionsActivity(ctx context.Context, params Delete
...
-		})
+
+		archetypeID, err := workercommon.ArchetypeIDFromExecutionInfo(execution)
+		if err != nil {
+			logger.Error("Failed to extract archetype ID from execution info.", tag.Error(err))
+			return result, fmt.Errorf("archetypeID extraction error: %w", err)
+		}
+
+		if archetypeID == chasm.WorkflowArchetypeID {
+			// TODO: consider using ForceDeleteWorkflowExecution for workflow as well.
```
yiminc, line 193: this is only used for namespace deletion? if true, then we should just use force delete here.

yiminc (approved):

service/worker/deletenamespace/deleteexecutions/activities.go
yycptt: We will need to refactor force deletion api handle a bit though. Today it only does best effort deletion on history events, we will need a non best effort mode first and then switch to it.

PR #8769, opened 2025-12-06
Fix TemporalReportedProblems SA application for buffered events

spkane31:
## What changed?
When buffered events are applied, like when applying signals after workflow task failures occur, preserve the workflow task attempt so the `TemporalReportedProblems` search attribute is still properly added.

## Why?
Without this change the `TemporalReportedProblems` search attribute will never be added if a workflow is consistently queried.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Minimal, this adds a new int to the workflow task execution, but it's only used in one place.


service/history/workflow/workflow_task_state_machine.go
```diff
@@ -319,6 +319,14 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskScheduledEventAsHeartbeat(
 	// Flush any buffered events before creating the workflow task, otherwise it will result in invalid IDs for
 	// transient/speculative workflow task and will cause in timeout processing to not work for transient workflow tasks.
 	if m.ms.HasBufferedEvents() {
+		// When buffered events are applied, preserve the total failure count from the previous workflow task
+		// before resetting Attempt to 1. This ensures we track continuous failures across workflow task resets.
+		if m.ms.IsTransientWorkflowTask() {
+			// Calculate total failures from previous WFT: AttemptsSinceLastSuccess + Attempt
+			// This becomes the new AttemptsSinceLastSuccess for the next WFT
+			previousTotalFailures := m.ms.executionInfo.WorkflowTaskAttemptsSinceLastSuccess + m.ms.executionInfo.WorkflowTaskAttempt
+			m.ms.executionInfo.WorkflowTaskAttemptsSinceLastSuccess = previousTotalFailures
```
bergundy: Can you update WorkflowTaskAttemptsSinceLastSuccess every time there is a WFT failure instead of only doing it when flushing buffered events? That seems more intuitive behavior to keep this attribute up-to-date.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -301,3 +301,259 @@ func (s *WorkflowTaskReportedProblemsReplicationSuite) TestWFTFailureReportedPro
...
+		TaskQueue: taskQueue,
+	}
+
+	workflowRun, err := activeSDKClient.ExecuteWorkflow(ctx, workflowOptions, s.workflowWithSignalsThatFails)
+	s.NoError(err)
+
+	// Start sending signals every second in a goroutine
+	signalDone := make(chan struct{})
+	go func() {
+		defer close(signalDone)
```
bergundy: You could instead signal for a side effect in the workflow and immediately panic. No need to add concurrency to this test.

tests/xdc/workflow_task_reported_problems_test.go
```diff
@@ -301,3 +301,259 @@ func (s *WorkflowTaskReportedProblemsReplicationSuite) TestWFTFailureReportedPro
...
+		workflowRun.GetID(),
+		workflowRun.GetRunID(),
+		ns,
+		"",
+		"",
+		false,
+	)
+}
+
+func (s *WorkflowTaskReportedProblemsReplicationSuite) TestWFTFailureReportedProblems_SetAndClear_FailAfterActivity() {
```
bergundy: Shouldn't this be also testing the buffered signal case?

tests/workflow_task_reported_problems_test.go
```diff
@@ -45,6 +46,24 @@ func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
...
 }
 
+// workflowWithSignalsThatFails creates a workflow that listens for signals and fails on each workflow task.
+// This is used to test that the TemporalReportedProblems search attribute is not incorrectly removed
+// when signals keep coming in despite continuous workflow task failures.
+func (s *WFTFailureReportedProblemsTestSuite) workflowWithSignalsThatFails(ctx workflow.Context) (string, error) {
+	signalChan := workflow.GetSignalChannel(ctx, "test-signal")
+
+	var signalValue string
+	signalChan.Receive(ctx, &signalValue)
```
bergundy: You don't really need to handle the signals in the workflow, it doesn't add value in the test and prevents your workflow task from failing because you never reach the line where it panics.

bergundy (commented):
@yycptt would you please also take a look? I'm less familiar with workflow tasks than you are. Also wondering if this covers replication properly, we are assuming state based replication is enabled for this.

tests/xdc/workflow_task_reported_problems_test.go
spkane31: Duplicated the wrong test here, removing this.

tests/workflow_task_reported_problems_test.go
```diff
@@ -45,6 +43,22 @@ func (s *WFTFailureReportedProblemsTestSuite) simpleActivity() (string, error) {
...
 
+// workflowWithSignalsThatFails creates a workflow that listens for signals and fails on each workflow task.
+// This is used to test that the TemporalReportedProblems search attribute is not incorrectly removed
+// when signals keep coming in despite continuous workflow task failures.
+func (s *WFTFailureReportedProblemsTestSuite) workflowWithSignalsThatFails(ctx workflow.Context) (string, error) {
+	// If we should fail, signal ourselves (creating a side effect) and immediately panic.
+	// This will create buffered.
+	if s.shouldFail.Load() {
+		// Signal ourselves to create buffered events
+		_ = workflow.SignalExternalWorkflow(ctx, workflow.GetInfo(ctx).WorkflowExecution.ID, "", "test-signal", "self-signal")
```
bergundy: This doesn't create buffered events, you need to do it from a `workflow.SideEffect`: https://pkg.go.dev/go.temporal.io/sdk/workflow#hdr-SideEffect_API

bergundy: You will need to use a temporal client to issue the signal, and make sure you don't ignore errors, we should be able to tell that the signal was successfully submitted.

tests/workflow_task_reported_problems_test.go
```diff
@@ -110,6 +127,49 @@ func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_Set
...
+		saVal, ok := description.TypedSearchAttributes.GetKeywordList(temporal.NewSearchAttributeKeyKeywordList(sadefs.TemporalReportedProblems))
+		require.True(t, ok)
+		require.NotEmpty(t, saVal)
+		require.Contains(t, saVal, "category=WorkflowTaskFailed")
+		require.Contains(t, saVal, "cause=WorkflowTaskFailedCauseWorkflowWorkerUnhandledFailure")
+	}, 20*time.Second, 500*time.Millisecond)
+
+	// Verify the search attribute persists even as the workflow continues to fail and create buffered events
+	// This is the key part of the test - buffered events should not cause the search attribute to be cleared
+	s.EventuallyWithT(func(t *assert.CollectT) {
```
bergundy, line 160: Before this, wait for the history to have a second failing task. You should see a repeated pattern where the task gets started, failed and a signal is sandwiched in between.

tests/workflow_task_reported_problems_test.go
```diff
@@ -110,6 +127,49 @@ func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_Set
 	s.False(ok)
 }
 
+func (s *WFTFailureReportedProblemsTestSuite) TestWFTFailureReportedProblems_NotClearedBySignals() {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	s.shouldFail.Store(true)
```
bergundy, line 134: If you don't wait for the workflow to recover (I think that's fine because we test recovery elsewhere) you  don't need this atomic bool.

PR #8771, opened 2025-12-07
Standalone Activity: DescribeActivityExecution and GetActivityExecutionOutcome

dandavison:
## What changed?
- Add new public gRPC methods `DescribeActivityExecution` and `GetActivityExecutionOutcome`. See https://github.com/temporalio/api/pull/673
- These replace the previous `PollActivityExecution`
- Respond to additional API changes: inlining of `ActivityOptions`
- Configure quota for the new methods in their blocking and non-blocking forms
- Update test suite

## Why?
- Implements agreed Standalone Activity design

## How did you test it?
- [x] built

---

> [!NOTE]
> Introduce DescribeActivityExecution and GetActivityExecutionOutcome APIs (replacing PollActivityExecution), inline ActivityOptions in StartActivityExecution, and update plumbing, quotas, and tests accordingly.
> 
> - **Public API changes**:
>   - Replace `workflowservice.PollActivityExecution` with `DescribeActivityExecution` and `GetActivityExecutionOutcome` (long-poll support via `LongPollToken`).
>   - Inline `ActivityOptions` fields in `StartActivityExecutionRequest` and propagate via helpers.
> - **Server/Frontend**:
>   - Add handlers, validation (`ValidateDescribeActivityExecutionRequest`, `ValidateGetActivityExecutionOutcomeRequest`), and long-poll behavior.
>   - Update rate limiting, redirection, metadata, and quotas (new poll aliases; priority classifications).
> - **Activity component**:
>   - Rename `ScheduledTime` → `ScheduleTime` throughout; adjust retry deadline calc.
>   - Add `outcome` helper and `buildGetActivityExecutionOutcomeResponse`; refactor describe response builder.
> - **Protos/clients (generated)**:
>   - Update proto messages/services, gRPC stubs, layered/metric/retryable clients, and mocks to new methods and fields.
> - **Tests**:
>   - Migrate tests from `PollActivityExecution` to `DescribeActivityExecution`/`GetActivityExecutionOutcome`; adapt assertions to new fields.
> - **Deps**:
>   - Bump `go.temporal.io/api` version.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 32cd5cff685184af344f2fdafcf527c504c6cdab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -352,7 +352,6 @@ func (a *Activity) recordFailedAttempt(
 	failure *failurepb.Failure,
 	noRetriesLeft bool,
 ) error {
-	outcome := a.Outcome.Get(ctx)
 	attempt := a.LastAttempt.Get(ctx)
 	currentTime := timestamppb.New(ctx.Now(a))
 
@@ -365,7 +364,6 @@ func (a *Activity) recordFailedAttempt(
 	// If the activity has exhausted retries, mark the outcome failure as well but don't store duplicate failure info.
 	// Also reset the retry interval as there won't be any more retries.
 	if noRetriesLeft {
-		outcome.Variant = &activitypb.ActivityOutcome_Failed_{}
 		attempt.CurrentRetryInterval = nil
 	} else {
 		attempt.CurrentRetryInterval = durationpb.New(retryInterval)


chasm/lib/activity/proto/v1/tasks.proto
```diff
@@ -11,16 +11,16 @@ message ActivityDispatchTask {
 }
 
 message ScheduleToStartTimeoutTask {
-  // The current attempt number for this activity execution. Since task validation/exec happen outside of a lock, we
-  // need to guard against any concurrent operations where the originally intended task may be outdated.
-  int32 attempt = 1;
+    // The current attempt number for this activity execution. Since task validation/exec happen outside of a lock, we
```
bergundy, line 14: We need a linter/formatter so this doesn't happen again. Maybe open an issue and we'll tackle it at some later point?

chasm/lib/activity/activity.go
```diff
@@ -486,80 +483,104 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+		response.Outcome = &workflowservice.GetActivityExecutionOutcomeResponse_Failure{Failure: failure}
+	}
+	return &activitypb.GetActivityExecutionOutcomeResponse{
+		FrontendResponse: response,
+	}, nil
+}
+
+// getOutcome retrieves the activity outcome (result or failure) if the activity has completed.
+// Returns (result, failure, error) where at most one of result/failure is non-nil.
+func (a *Activity) getOutcome(ctx chasm.Context) (*commonpb.Payloads, *failurepb.Failure, error) {
```
bergundy: Please don't use `get` for getters in Go: https://go.dev/doc/effective_go#Getters

chasm/lib/activity/activity.go
```diff
@@ -486,80 +483,104 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
 		Priority:                a.GetPriority(),
 		RunId:                   key.RunID,
 		RunState:                runState,
-		ScheduledTime:           a.GetScheduledTime(),
+		ScheduleTime:            a.GetScheduledTime(),
```
bergundy: Might as well rename to `schedule_time` in the internal protos too.

chasm/lib/activity/activity.go
```diff
@@ -486,80 +483,104 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
-						Failure: details.GetFailure(),
-					}
-				}
-			}
+		result, failure, err := a.getOutcome(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if result != nil {
+			response.Outcome = &workflowservice.DescribeActivityExecutionResponse_Result{Result: result}
```
bergundy: We can add an `Outcome` message in the protos and use it in both APIs instead of having a separate oneof in each response.

chasm/lib/activity/activity.go
```diff
@@ -486,80 +483,104 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+// getOutcome retrieves the activity outcome (result or failure) if the activity has completed.
+// Returns (result, failure, error) where at most one of result/failure is non-nil.
+func (a *Activity) getOutcome(ctx chasm.Context) (*commonpb.Payloads, *failurepb.Failure, error) {
+	activityOutcome := a.Outcome.Get(ctx)
+	// Check for successful outcome
+	if successful := activityOutcome.GetSuccessful(); successful != nil {
+		return successful.GetOutput(), nil, nil
+	}
+	// Check for failure in outcome
+	if failure := activityOutcome.GetFailed().GetFailure(); failure != nil {
```
bergundy, line 551: Please verify that we don't set failure to an empty struct when we fail from an attempt.

chasm/lib/activity/activity.go
```diff
@@ -486,80 +483,104 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+		return nil, failure, nil
+	}
+	// Check for failure in last attempt details
+	shouldHaveFailure := (a.GetStatus() == activitypb.ACTIVITY_EXECUTION_STATUS_FAILED ||
```
bergundy, line 557: Just noticed this: terminated and timed out should never return outcome from the last attempt, canceled would sometimes have to read the last attempt failure.

chasm/lib/activity/frontend.go
```diff
@@ -224,3 +251,27 @@ func (h *frontendHandler) validateAndPopulateStartRequest(
...
+		ScheduleToStartTimeout: req.ScheduleToStartTimeout,
+		StartToCloseTimeout:    req.StartToCloseTimeout,
+		HeartbeatTimeout:       req.HeartbeatTimeout,
+		RetryPolicy:            req.RetryPolicy,
+	}
+}
+
+// applyActivityOptionsToStartRequest copies normalized values from ActivityOptions
+// back to the StartActivityExecutionRequest.
+func applyActivityOptionsToStartRequest(opts *apiactivitypb.ActivityOptions, req *workflowservice.StartActivityExecutionRequest) {
```
bergundy, line 270: Why not pass in the request into the `ValidateAndNormalizeActivityAttributes` function instead of using this options struct which isn't part of the API? Is that function used for the workflow command and the activity operator APIs?

chasm/lib/activity/handler.go
```diff
@@ -132,67 +123,91 @@ func (h *handler) PollActivityExecution(
...
+			return response, true, err
+		}
+		return nil, false, nil
+	}, req)
 
-			return nil, false, nil
-		}, req)
-	default:
-		return nil, serviceerror.NewInvalidArgumentf("unexpected wait policy type: %T", waitPolicy)
+	if ctx.Err() != nil {
```
bergundy: I thought about it during the weekend and I think we should only do this check if `err != nil` just in case there's a race and we could potentially fulfill the response.

chasm/lib/activity/activity.go
bergundy: Might be best to first check if the lifecycle state is not running at the top of this function and then go through, outcome success, outcome failure, and last result without checking statuses here.

chasm/lib/activity/validator.go
```diff
@@ -285,3 +285,24 @@ func ValidatePollActivityExecutionRequest(
...
+) error {
+	if req.GetActivityId() == "" {
+		return serviceerror.NewInvalidArgument("activity ID is required")
+	}
+	if len(req.GetActivityId()) > maxIDLengthLimit {
+		return serviceerror.NewInvalidArgumentf("activity ID exceeds length limit. Length=%d Limit=%d",
+			len(req.GetActivityId()), maxIDLengthLimit)
+	}
+	if runID := req.GetRunId(); runID != "" {
+		_, err := uuid.Parse(runID)
```
bergundy, line 302: Do we typically validate that run IDs are uuids in our API handlers?

service/frontend/configs/quotas.go
```diff
@@ -24,32 +24,34 @@ const (
 	CompleteNexusOperation                          = "/temporal.api.nexusservice.v1.NexusService/CompleteNexusOperation"
 	// PollWorkflowHistoryAPIName is used instead of GetWorkflowExecutionHistory if WaitNewEvent is true in request.
 	PollWorkflowHistoryAPIName = "/temporal.api.workflowservice.v1.WorkflowService/PollWorkflowExecutionHistory"
+	// PollActivityExecutionAPIName is used instead of GetActivityExecutionOutcome if LongPollToken is set in request.
```
bergundy:
```suggestion
	// PollActivityExecutionAPIName is used instead of DescribeActivityExecution if LongPollToken is set in request.
```

service/frontend/configs/quotas.go
```diff
@@ -24,32 +24,34 @@ const (
...
+	PollActivityExecutionAPIName = "/temporal.api.workflowservice.v1.WorkflowService/PollActivityExecution"
```
bergundy: I think being more specific is preferable here.

```suggestion
	PollActivityExecutionAPIName = "/temporal.api.workflowservice.v1.WorkflowService/PollActivityExecutionDescription"
```

chasm/lib/activity/activity.go
dandavison: Done

chasm/lib/activity/activity.go
dandavison: Thanks, done, pushed to proto repos. Named it `ActivityExecutionOutcome` but lmk if you want it to be just `Outcome`.

chasm/lib/activity/activity.go
dandavison: Thanks! changed to `outcome`

chasm/lib/activity/activity.go
dandavison: I've moved this to an issue to track because I want to write a test to repro it and we can unblock others by merging this now.

chasm/lib/activity/frontend.go
dandavison: Yes, it's because it's shared with workflow code. I did try to say that in the comment above `for use with shared validation logic`. Anyone got better ideas here?

chasm/lib/activity/handler.go
dandavison: Done! Looks like Cursorbot agrees (it had taken that position it out in a previous PR)

chasm/lib/activity/validator.go
dandavison: Yes, e.g. https://github.com/temporalio/temporal/blob/main/service/frontend/validators.go#L23

chasm/lib/activity/proto/v1/tasks.proto
dandavison: Done

chasm/lib/activity/activity.go
dandavison: We were still doing that (I'd brought it up before but we hadn't removed it). I removed it in 32cd5cff6 (PTAL).

PR #8772, opened 2025-12-08
New matching RPC: Check if version exists in TaskQueue 

Shivs11:
## What changed?
- Added a new RPC method, scoped to matching only, which shall check if a specific version exists in the userData of a task queue.

## Why?
- Currently, versioning has several places in history which call the method GetTaskQueueUserData to validate if a version exists in a task queue or not.
- However, with [this](https://github.com/temporalio/temporal/pull/8722) going in very soon, one of the biggest changes put forward by it is the fact that a task queue's userData could go on to become big (if a task queue was part of a large number of versions). The main reason for this is because we keep an entry for versions that are deleted inside the task-queue's user data so that we can ensure read-after-write consistencies.
- Thus, to prevent large number of bytes flowing across the wire, it's best to just make this new RPC that does the check in matching and only sends a boolean across. 

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- No risk honestly I think this is an optimization

---

> [!NOTE]
> Introduces a matching RPC to check if a task queue belongs to a deployment version, wires it through proto, gRPC, clients, engine, quotas, and tests, and updates worker versioning to use it.
> 
> - **API/Protobuf**:
>   - Add `CheckTaskQueueVersionMembershipRequest`/`Response` messages in `api/matchingservice/v1/request_response.proto` and generated Go files.
>   - Add `CheckTaskQueueVersionMembership` method to `MatchingService` in `service.proto` and generated gRPC stubs.
> - **Server**:
>   - Implement handler `Handler.CheckTaskQueueVersionMembership` forwarding to engine.
>   - Implement engine method `matchingEngineImpl.CheckTaskQueueVersionMembership` to read user data and return membership via `worker_versioning.HasDeploymentVersion`.
>   - Include new RPC in rate-limiting quotas and logtags switch.
> - **Clients/Mocks**:
>   - Add client methods in generated `client_gen.go`, `metric_client_gen.go`, and `retryable_client_gen.go`.
>   - Update gRPC client/server interfaces and register handler; add mocks in `matchingservicemock`.
> - **Worker Versioning**:
>   - Replace `GetTaskQueueUserData` check with `CheckTaskQueueVersionMembership` in detector.
> - **Tests**:
>   - Add functional test `TestCheckTaskQueueVersionMembership` validating false-before/true-after worker registration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ff2ea79afbc0436fb329bfd7bc0c461a7548a3c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/matching/user_data_manager.go
+++ b/service/matching/user_data_manager.go
@@ -8,7 +8,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/dgryski/go-farm"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/api/matchingservice/v1"
@@ -250,11 +249,7 @@ func (m *userDataManagerImpl) userDataFetchSource() (*tqid.NormalPartition, erro
 		}
 		// sticky queue can only be of workflow type as of now. but to be future-proof, we make sure
 		// change to workflow task queue here
-		wfTQ := normalQ.Family().TaskQueue(enumspb.TASK_QUEUE_TYPE_WORKFLOW)
-		// use hash of the sticky queue name to pick a consistent "parent"
-		partitions := m.config.NumReadPartitions()
-		partition := int(farm.Fingerprint32([]byte(m.partition.RpcName()))) % partitions
-		return wfTQ.NormalPartition(partition), nil
+		return normalQ.Family().TaskQueue(enumspb.TASK_QUEUE_TYPE_WORKFLOW).RootPartition(), nil
 	}
 
 }


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -641,3 +641,14 @@ message DescribeWorkerRequest {
 message DescribeWorkerResponse {
     temporal.api.worker.v1.WorkerInfo worker_info = 1;
 }
+
+message CheckTaskQueueVersionMembershipRequest {
+    string namespace_id = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+    temporal.api.enums.v1.TaskQueueType task_queue_type = 3;
```
Shivs11, line 648: Technically, we don't need this (right now) since all the version presence checks for a task queue go through the workflow task queue type. However, I just wanted to be consistent with other matching API's since those that accept task-queue also accept the type. Moreover, we could have a future use-case of involving the activity task queue type of a task queue.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -641,3 +641,14 @@ message DescribeWorkerRequest {
 message DescribeWorkerResponse {
     temporal.api.worker.v1.WorkerInfo worker_info = 1;
 }
+
+message CheckTaskQueueVersionMembershipRequest {
+    string namespace_id = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
```
ShahabT: we're trying to move away from this proto as much as possible because it mixes multiple levels of TQs. In this case the API is only relevant to normal TQs, so just a string field is enough.

common/worker_versioning/worker_versioning.go
```diff
@@ -265,21 +265,19 @@ func GetIsWFTaskQueueInVersionDetector(matchingClient resource.MatchingClient) I
...
-				TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW,
-			})
+		resp, err := matchingClient.CheckTaskQueueVersionMembership(ctx, &matchingservice.CheckTaskQueueVersionMembershipRequest{
+			NamespaceId:   namespaceID,
+			TaskQueue:     tq,
+			TaskQueueType: enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+			Version: &deploymentspb.WorkerDeploymentVersion{
+				DeploymentName: version.GetDeploymentName(),
+				BuildId:        version.GetBuildId(),
+			},
```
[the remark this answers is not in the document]
Shivs11: I think this should not happen and is the caller's responsibility for passing non-empty structs. Even if it does happen, worst case would be a slice check.

PR #8773, opened 2025-12-08
CHASM Visibility ListWorkflowExecutions ignore unregistered CHASM search attributes

awln-temporal:
## What changed?
Silently ignore unregistered CHASM search attributes for SQL visibility store.
ES already ignores unregistered search attributes (custom or chasm).

## Why?
ListWorkflowExecutions needs to succeed when querying on all TemporalNamespaceDivisions.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/chasm/lib/callback/executors_test.go
+++ b/chasm/lib/callback/executors_test.go
@@ -184,7 +184,7 @@ func TestExecuteInvocationTaskNexus_Outcomes(t *testing.T) {
 			}
 
 			chasmRegistry := chasm.NewRegistry(logger)
-			err := chasmRegistry.Register(&Library{
+			err = chasmRegistry.Register(&Library{
 				InvocationTaskExecutor: executor,
 			})
 			require.NoError(t, err)
@@ -587,7 +587,7 @@ func TestExecuteInvocationTaskChasm_Outcomes(t *testing.T) {
 			}
 
 			chasmRegistry := chasm.NewRegistry(logger)
-			err := chasmRegistry.Register(&Library{
+			err = chasmRegistry.Register(&Library{
 				InvocationTaskExecutor: executor,
 			})
 			require.NoError(t, err)


common/persistence/visibility/store/sql/visibility_store.go
```diff
@@ -859,13 +860,19 @@ func (s *VisibilityStore) encodeRowSearchAttributes(
...
 
 	combinedTypeMap := store.CombineTypeMaps(saTypeMap, chasmMapper)
+	registeredSearchAttributes := sqlplugin.VisibilitySearchAttributes{}
 
 	// Fix SQLite keyword list handling (convert string to []string for keyword lists)
 	for name, value := range rowSearchAttributes {
 		tp, err := combinedTypeMap.GetType(name)
 		if err != nil {
+			// Silently ignore ErrInvalidName for unregistered chasm search attributes
+			if sadefs.IsChasmSearchAttribute(name) && errors.Is(err, searchattribute.ErrInvalidName) {
```
rodrigozhou, line 870: The `combinedTypeMap` already contains the CHASM search attributes. Do you need to check `IsChasmSearchAttribute` here?

common/persistence/visibility/store/sql/visibility_store.go
rodrigozhou: Need to update

common/persistence/visibility/store/sql/visibility_store.go
awln-temporal: yeah, since the chasm search search attributes are not registered when calling ListWorkflowExecutions.

rodrigozhou: Hm, in this case, isn't it better to check before calling `combinedTypeMap.GetType`?

awln-temporal: no, since the chasm search attribute could be registered, in which case we do want to try returning it. visibility store implementations are unified for `ListWorkflowExecutions` and `ListChasmExecutions` as of now.

rodrigozhou (approved):

PR #8775, opened 2025-12-08
Track external payloads stats for workflow execution

simvlad:
## What changed?
Keep the total number and the size of the external payloads per the workflow execution

## Why?
We are working on building the support for external payloads in SDK, which are stored outside of Temporal. We'd like to be able to show the total size and the number of external payloads in the given workflow execution. 

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
N/A

### DIFF
--- a/api/persistence/v1/executions.pb.go
+++ b/api/persistence/v1/executions.pb.go
@@ -1111,9 +1111,11 @@ func (*WorkflowExecutionInfo_LastWorkflowTaskTimedOutType) isWorkflowExecutionIn
 type ExecutionStats struct {
 	state       protoimpl.MessageState `protogen:"open.v1"`
 	HistorySize int64                  `protobuf:"varint,1,opt,name=history_size,json=historySize,proto3" json:"history_size,omitempty"`
-	// Total size in bytes of all external payloads referenced in the workflow history.
+	// Total size in bytes of all external payloads referenced in the entire history tree of the execution, not just the current branch.
+	// This number doesn't include payloads in buffered events.
 	ExternalPayloadSize int64 `protobuf:"varint,2,opt,name=external_payload_size,json=externalPayloadSize,proto3" json:"external_payload_size,omitempty"`
-	// Total count of external payloads referenced in the workflow history.
+	// Total count of external payloads referenced in the entire history tree of the execution, not just the current branch.
+	// This number doesn't include payloads in buffered events.
 	ExternalPayloadCount int64 `protobuf:"varint,3,opt,name=external_payload_count,json=externalPayloadCount,proto3" json:"external_payload_count,omitempty"`
 	unknownFields        protoimpl.UnknownFields
 	sizeCache            protoimpl.SizeCache


yycptt:
I think state-based replication logic in workflow_state_replicator.go also changes HistorySize and needs to be updated as well. Haven't finished the review yet, will continue tomorrow.

cc @xwduan @yux0 Please help review this PR as most of the changes are actually on the replication side.

simvlad:
@xwduan, @yux0 - I fixed the replication logic in service/history/ndc/workflow_state_replicator.go and added the test to service/history/ndc/workflow_state_replicator_test.go. Please take a look when you have time.

service/history/ndc/state_rebuilder.go
```diff
@@ -360,6 +380,16 @@ func (r *StateRebuilderImpl) getPaginationFn(
 				TransactionID: resp.TransactionIDs[i],
 			}
 			paginateItems = append(paginateItems, nextBatch)
+
+			// Calculate and accumulate external payload size and count for this batch of history events
+			if r.shard.GetConfig().ExternalPayloadsEnabled(namespaceName) {
```
yycptt, line 385: nit: feel like checking the flag once per page/request is also good enough. no strong opinion.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6399,6 +6399,28 @@ func (ms *MutableStateImpl) AddHistorySize(size int64) {
 	ms.executionInfo.ExecutionStats.HistorySize += size
 }
 
+func (ms *MutableStateImpl) GetExternalPayloadSize() int64 {
+	return ms.executionInfo.ExecutionStats.ExternalPayloadSize
```
yycptt: do we need nil check here as well? or use Get methods?

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -292,6 +292,10 @@ message WorkflowExecutionInfo {
 
 message ExecutionStats {
     int64 history_size = 1;
+    // Total size in bytes of all external payloads referenced in the workflow history.
+    int64 external_payload_size = 2;
+    // Total count of external payloads referenced in the workflow history.
+    int64 external_payload_count = 3;
```
yycptt, line 300: - plz remove `workflow` from the comments, we will support history events for other archetypes as well later.
- we need to clarify that this won't include payloads in buffered events (is that ok?)
- we also need to clarify that this is the size & count for the entire history tree of the execution, not just the current history branch.

yycptt:
@yux0 @xwduan plz take a look as well. Thanks!

service/history/workflow/mutable_state_impl.go
```diff
@@ -7535,6 +7557,16 @@ func (ms *MutableStateImpl) closeTransactionPrepareEvents(
 		}
 		ms.executionInfo.LastFirstEventId = eventBatch[0].GetEventId()
 		ms.executionInfo.LastFirstEventTxnId = historyNodeTxnIDs[index]
+
+		// Calculate and add the external payload size and count for this batch
+		if ms.config.ExternalPayloadsEnabled(ms.GetNamespaceEntry().Name().String()) {
```
yycptt, line 7562: I wonder if we should reset the size & count if the feature flag is disabled.

simvlad: We can exclude it from the results of DescribeWorkflowExecution, but not sure if resetting would make this better. We still need to explicitly retrigger the recomputation of it, if it somehow gets computed incorrectly.

simvlad:
This feature is gated/in pre-release, so I'll merge this change. But @xwduan, @yux0, let me know if there are any changes needed to the replication part.

service/history/workflow/mutable_state_impl.go
yycptt: > We still need to explicitly retrigger the recomputation of it

How do we know if recomputation is needed for an execution if the feature flag get disabled and then enabled?

simvlad: But having those fields set tot 0, is also not a good indicator that the recomputation is necessary.

PR #8780, opened 2025-12-09
Add metric to track inflight visibility tasks

rodrigozhou:
## What changed?
Added metric to track number of inflight visibility tasks.

## Why?
Help debug bottlenecks in the visibility task flow: task queue or bulk processor.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -846,7 +846,10 @@ var (
 	QueueReaderCountHistogram = NewDimensionlessHistogramDef("queue_reader_count")
 	QueueSliceCountHistogram  = NewDimensionlessHistogramDef("queue_slice_count")
 	QueueActionCounter        = NewCounterDef("queue_actions")
-	ActivityE2ELatency        = NewTimerDef(
+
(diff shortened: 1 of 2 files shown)


yycptt:
Can you provide some context for this metric? I mean it will limited by the visibility task gorouting pool size.

FWIW, we already have `pending_tasks` metric and can be scoped to visibility queue, though it's a histogram for pending task count across shards.

rodrigozhou:
IIUC, `pending_tasks` included all tasks that were load into memory, ie., not only the tasks dispatched to the executor. I wanted to track tasks that are in the executor to check if we are exhausting the `history.visibilityProcessorSchedulerWorkerCount` value. Basically, help diagnose where the bottleneck is in visibility: task queue or bulk processor.

yycptt:
> Basically, help diagnose where the bottleneck is in visibility: task queue or bulk processor.

The `bulk_processor_bulk_size` and `bulk_processor_queued_requests` can provide that insight? Or I am still missing something here?

rodrigozhou:
@yycptt Yes, I think it could.
Closing the PR.

PR #8783, opened 2025-12-09
Revert "fix: make release dep check job always run on PRs"

stpierre:
Reverts temporalio/temporal#8750

This turned out to be unnecessary; we already have a separate ruleset for release branches, so we can just make this status required in that ruleset. Usually Github makes it as hard as possible to require different statuses for different PRs, but the exception turns out to be when those PRs can be clearly separated by target branch name.


lilydoar:
@stpierre I assume this PR is no longer relevant? If it's still needed then I can take a look at it

lilydoar (approved):

PR #8784, opened 2025-12-09
AutoUpgrade-on-Continue-as-New option (Trampolining)

carlydf:
## What changed?
Give the option to upgrade-on-Continue-as-New with a continue-as-new option.
Ensure that Pinned Overrides are inherited across the entire continue-as-new chain, regardless of the initial versioning behavior of the continue-as-new.

## Why?
To unlock "Trampolining"

## How did you test it?
- [x] built
- [ ] added new unit test(s)

---

> [!NOTE]
> Introduces explicit Continue-as-New tracking and version-targeting across APIs and server internals.
> 
> - Protos: add `target_deployment_version` to `RecordWorkflowTaskStartedRequest`; add repeated `workflow_task_suggest_continue_as_new_reasons` to persistence `WorkflowExecutionInfo`; regenerate bindings and enum imports
> - History: plumb `target_deployment_version` through `AddWorkflowTaskStartedEvent`; include `suggest_continue_as_new_reasons` in `WorkflowTaskStarted` events; choose non-nil `VersioningOverride` when starting executions (user or inherited)
> - Metrics: remove `workflow_update_continue_as_new_suggestions`; add `workflow_continue_as_new_count` and `workflow_suggest_continue_as_new_count`; introduce tags `continue_as_new_versioning_behavior` and per-reason suggest flags; tag helpers added
> - Interfaces/tests: extend `MutableState.Add/ApplyWorkflowTaskStartedEvent` and `WorkflowTaskInfo` with suggest reasons; update call sites and tests; update `common/util_test.go` scan; bump `go.temporal.io/api`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 849687bd0cee86dfa8a7949e76161a5c5a090f25. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/api/matchingservice/v1/request_response.go-helpers.pb.go
+++ b/api/matchingservice/v1/request_response.go-helpers.pb.go
@@ -2816,3 +2816,40 @@ func (this *CheckTaskQueueVersionMembershipResponse) Equal(that interface{}) boo
 
 	return proto.Equal(this, that1)
 }
+
+// Marshal an object of type PollConditions to the protobuf v3 wire format
+func (val *PollConditions) Marshal() ([]byte, error) {
+	return proto.Marshal(val)
+}
+
+// Unmarshal an object of type PollConditions from the protobuf v3 wire format
+func (val *PollConditions) Unmarshal(buf []byte) error {
+	return proto.Unmarshal(buf, val)
+}
+
+// Size returns the size of the object, in bytes, once serialized
+func (val *PollConditions) Size() int {
+	return proto.Size(val)
+}
+
+// Equal returns whether two PollConditions values are equivalent by recursively
+// comparing the message's fields.
+// For more information see the documentation for
+// https://pkg.go.dev/google.golang.org/protobuf/proto#Equal
+func (this *PollConditions) Equal(that interface{}) bool {
+	if that == nil {
+		return this == nil
+	}
+
+	var that1 *PollConditions
+	switch t := that.(type) {
+	case *PollConditions:
+		that1 = t
+	case PollConditions:
+		that1 = &t
+	default:
+		return false
+	}
+
+	return proto.Equal(this, that1)
+}
--- a/api/matchingservice/v1/request_response.pb.go
+++ b/api/matchingservice/v1/request_response.pb.go
@@ -45,8 +45,10 @@ type PollWorkflowTaskQueueRequest struct {
 	PollerId        string                           `protobuf:"bytes,2,opt,name=poller_id,json=pollerId,proto3" json:"poller_id,omitempty"`
 	PollRequest     *v1.PollWorkflowTaskQueueRequest `protobuf:"bytes,3,opt,name=poll_request,json=pollRequest,proto3" json:"poll_request,omitempty"`
 	ForwardedSource string                           `protobuf:"bytes,4,opt,name=forwarded_source,json=forwardedSource,proto3" json:"forwarded_source,omitempty"`
-	unknownFields   protoimpl.UnknownFields
-	sizeCache       protoimpl.SizeCache
+	// Extra conditions on this poll request. Only supported with new matcher.
+	Conditions    *PollConditions `protobuf:"bytes,5,opt,name=conditions,proto3" json:"conditions,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *PollWorkflowTaskQueueRequest) Reset() {
@@ -105,6 +107,13 @@ func (x *PollWorkflowTaskQueueRequest) GetForwardedSource() string {
 		return x.ForwardedSource
 	}
 	return ""
+}
+
+func (x *PollWorkflowTaskQueueRequest) GetConditions() *PollConditions {
+	if x != nil {
+		return x.Conditions
+	}
+	return nil
 }
 
 type PollWorkflowTaskQueueResponse struct {
@@ -312,8 +321,10 @@ type PollActivityTaskQueueRequest struct {
 	PollerId        string                           `protobuf:"bytes,2,opt,name=poller_id,json=pollerId,proto3" json:"poller_id,omitempty"`
 	PollRequest     *v1.PollActivityTaskQueueRequest `protobuf:"bytes,3,opt,name=poll_request,json=pollRequest,proto3" json:"poll_request,omitempty"`
 	ForwardedSource string                           `protobuf:"bytes,4,opt,name=forwarded_source,json=forwardedSource,proto3" json:"forwarded_source,omitempty"`
-	unknownFields   protoimpl.UnknownFields
-	sizeCache       protoimpl.SizeCache
+	// Extra conditions on this poll request. Only supported with new matcher.
+	Conditions    *PollConditions `protobuf:"bytes,5,opt,name=conditions,proto3" json:"conditions,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *PollActivityTaskQueueRequest) Reset() {
@@ -372,6 +383,13 @@ func (x *PollActivityTaskQueueRequest) GetForwardedSource() string {
 		return x.ForwardedSource
 	}
 	return ""
+}
+
+func (x *PollActivityTaskQueueRequest) GetConditions() *PollConditions {
+	if x != nil {
+		return x.Conditions
+	}
+	return nil
 }
 
 type PollActivityTaskQueueResponse struct {
@@ -3580,8 +3598,10 @@ type PollNexusTaskQueueRequest struct {
 	Request *v1.PollNexusTaskQueueRequest `protobuf:"bytes,3,opt,name=request,proto3" json:"request,omitempty"`
 	// Non-empty if this poll was forwarded from a child partition.
 	ForwardedSource string `protobuf:"bytes,4,opt,name=forwarded_source,json=forwardedSource,proto3" json:"forwarded_source,omitempty"`
-	unknownFields   protoimpl.UnknownFields
-	sizeCache       protoimpl.SizeCache
+	// Extra conditions on this poll request. Only supported with new matcher.
+	Conditions    *PollConditions `protobuf:"bytes,5,opt,name=conditions,proto3" json:"conditions,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *PollNexusTaskQueueRequest) Reset() {
@@ -3640,6 +3660,13 @@ func (x *PollNexusTaskQueueRequest) GetForwardedSource() string {
 		return x.ForwardedSource
 	}
 	return ""
+}
+
+func (x *PollNexusTaskQueueRequest) GetConditions() *PollConditions {
+	if x != nil {
+		return x.Conditions
+	}
+	return nil
 }
 
 type PollNexusTaskQueueResponse struct {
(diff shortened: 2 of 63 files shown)


service/history/workflow/workflow_task_state_machine.go
```diff
@@ -472,9 +474,19 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskStartedEvent(
...
+	historySizeBytes, suggestContinueAsNewReasons := m.getHistorySizeInfo()
+	suggestContinueAsNew := len(suggestContinueAsNewReasons) > 0
+	if updateReg != nil && updateReg.SuggestContinueAsNew() {
+		suggestContinueAsNew = cmp.Or(suggestContinueAsNew, true)
+		suggestContinueAsNewReasons = append(suggestContinueAsNewReasons, enumspb.SUGGEST_CONTINUE_AS_NEW_REASON_TOO_MANY_UPDATES)
+	}
+
+	if m.ms.GetEffectiveVersioningBehavior() != enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED && targetDeploymentVersion != nil {
+		if currentDeploymentVersion := m.ms.GetEffectiveDeployment(); currentDeploymentVersion.BuildId != targetDeploymentVersion.BuildId ||
+			currentDeploymentVersion.SeriesName != targetDeploymentVersion.DeploymentName {
```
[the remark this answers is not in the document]
carlydf: fixed

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -708,6 +722,7 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskCompletedEvent(
 			workflowTask.HistorySizeBytes,
 			request.WorkerVersionStamp,
 			workflowTask.BuildIdRedirectCounter,
+			nil,
 		)
```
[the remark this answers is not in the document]
ShahabT: this seems like a valid concern

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -247,6 +247,9 @@ message RecordWorkflowTaskStartedRequest {
     // Revision number that was sent by matching when the task was dispatched. Used to resolve eventual consistency issues
     // that may arise due to stale routing configs in task queue partitions.
     int64 task_dispatch_revision_number = 12;
+    // Target worker deployment version according to matching when starting the task.
+    // Computed after matching with a poller, right before calling RecordWorkflowTaskStarted.
+    temporal.api.deployment.v1.WorkerDeploymentVersion target_deployment_version = 13;
```
ShahabT, line 253: is it passed all the time or only when different from poller version?

service/history/workflow/workflow_task_state_machine.go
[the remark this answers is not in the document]
carlydf: Hm yeah, when I wrote this I was at first thinking that it's fine if the speculative tasks don't send the reasons, because the next task is probably not speculative, and the reason would eventually get to the workflow.

But, there could be a workflow that almost exclusively handles conditional updates, and then a user would definitely notice if they never get a CaN-reason..

And the difference between speculative and normal WFT should not be a concern for the user. So I should store it in the `WorkflowTaskInfo` probably

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -472,9 +478,20 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskStartedEvent(
...
-		suggestContinueAsNew = cmp.Or(suggestContinueAsNew, updateReg.SuggestContinueAsNew())
+	historySizeBytes, suggestContinueAsNewReasons := m.getHistorySizeInfo()
+	suggestContinueAsNew := len(suggestContinueAsNewReasons) > 0
+	if updateReg != nil && updateReg.SuggestContinueAsNew() {
+		suggestContinueAsNew = cmp.Or(suggestContinueAsNew, true)
+		suggestContinueAsNewReasons = append(suggestContinueAsNewReasons, enumspb.SUGGEST_CONTINUE_AS_NEW_REASON_TOO_MANY_UPDATES)
+	}
+
+	if m.ms.GetEffectiveVersioningBehavior() != enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED && targetDeploymentVersion != nil {
+		if currentDeploymentVersion := m.ms.GetEffectiveDeployment(); currentDeploymentVersion != nil &&
```
ShahabT, line 493: Q: would this work fine if a transition started just at this task for AU wf? expected behavior is no CaN suggestion.

just to double check... I assume `AddWorkflowTaskStartedEvent` is called after transition is started and then `GetEffectiveDeployment` would return the new version which equals target so the condition does not satisfy.

carlydf: Actually the AU workflow _is_ receiving the CaN suggestion on the transition-task, and then once the transition completes, the CaN suggestion is no longer sent to the AU workflow (confirmed by my functional test) -- do we not want that to happen?

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
carlydf: all the time, but I see your comment on how that's wasteful

service/history/workflow/mutable_state_impl.go
```diff
@@ -2483,9 +2487,15 @@ func (ms *MutableStateImpl) addWorkflowExecutionStartedEventForContinueAsNew(
 
 	// New run initiated by ContinueAsNew of an AUTO_UPGRADE workflow execution will inherit the previous run's
 	// deployment version and revision number iff the new run's Task Queue belongs to source deployment version.
+	//
+	// If the initiating workflow is PINNED and the continue-as-new command says to use InitialVersioningBehavior
+	// AutoUpgrade, the new run will start as AutoUpgrade in the first task and then assume the SDK-sent behavior
+	// after first workflow task completion.
 	var sourceDeploymentVersion *deploymentpb.WorkerDeploymentVersion
 	var sourceDeploymentRevisionNumber int64
-	if previousExecutionState.GetEffectiveVersioningBehavior() == enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE {
```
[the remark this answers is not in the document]
carlydf: No, I want the Pinned override to be inherited regardless of the initial base behavior

service/history/workflow/workflow_task_state_machine.go
carlydf: `StartDeploymentTransition` is called after `AddWorkflowTaskStartedEvent`, which is why the effective behavior and version is still the pre-transition one

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
carlydf: resolved!

service/history/workflow/workflow_task_state_machine.go
carlydf: done! AU wfs won't get the suggestion while transitioning

tests/versioning_3_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	commonpb "go.temporal.io/api/common/v1"
```
Shivs11, line 18: apologies for the question if I just have missed this in the file:

Do you think we should test VersioningOverride with the initial CAN behaviour? Maybe the test could then test that even though the SDK suggests a new deployment version, the pinned override would take precedence and keep the workflow running on that version

service/matching/task_queue_partition_manager.go
```diff
@@ -1056,34 +1072,30 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
...
 			if !deployment.Equal(targetDeployment) {
 				// Current deployment has changed, so the workflow should move to a normal queue to
 				// get redirected to the new deployment.
-				return nil, nil, nil, 0, serviceerrors.NewStickyWorkerUnavailable()
+				return nil, nil, nil, 0, nil, serviceerrors.NewStickyWorkerUnavailable()
 			}
 
 			// TODO (shahab): we can verify the passed deployment matches the last poller's deployment
-			return pm.defaultQueue, pm.defaultQueue, userDataChanged, 0, nil
+			return pm.defaultQueue, pm.defaultQueue, userDataChanged, 0, targetDeploymentVersion, nil
```
Shivs11, line 1098: question: is it important to pass this information when using a sticky worker as well?
moreover, should we test this behaviour out (as in, when using a sticky poller, is the behaviour that we see as expected? )

service/matching/task_queue_partition_manager.go
```diff
@@ -1018,26 +1030,30 @@ func (pm *taskQueuePartitionManagerImpl) getPhysicalQueuesForAdd(
...
-		return nil, nil, nil, 0, err
+		return nil, nil, nil, 0, nil, err
 	}
 	deploymentData := perTypeUserData.GetDeploymentData()
 	taskDirectiveRevisionNumber := directive.GetRevisionNumber()
 
+	current, currentRevisionNumber, _, ramping, _, rampingPercentage, rampingRevisionNumber, _ := worker_versioning.CalculateTaskQueueVersioningInfo(deploymentData)
+	targetDeploymentVersion, targetDeploymentRevisionNumber := worker_versioning.FindTargetDeploymentVersionAndRevisionNumberForWorkflowID(current, currentRevisionNumber, ramping, rampingPercentage, rampingRevisionNumber, workflowId)
+	targetDeployment := worker_versioning.DeploymentFromDeploymentVersion(targetDeploymentVersion)
+
```
Shivs11, line 1041: maybe I missed something here but pasting this if it actually is a valid concern:

The targetDeploymentVersion that we are extracting here is the potential target version the workflow may be sent to based on the revision number that is found in the task queue's user data at that point in time. What this would mean is that there could be a time where the task scheduled from history has a more up-to-date scheduled deployment version for this task than matching does. This is the reason why we have the function: chooseTargetQueueByFlag.

The changes you have added may always be returning the targetDeploymentVersion (which could not be up-to-date) to history. 

Say history has scheduled workflow task with version "foo.A" which is the current version of the deployment. Matching, which has not gotten the current version propagation yet, thinks "foo.B" is still the current version. The matching code correctly schedules the task to be sent to "foo.A". However, we still send back "foo.B" to history. 

Should we not be taking into consideration the revision number semantics here since they are present to resolve eventual consistencies between history and matching?

Shivs11 (commented):
one comment that could be important

service/matching/task_queue_partition_manager.go
carlydf: I think this commit resolves this issue! thanks for highlighting it! https://github.com/temporalio/temporal/pull/8784/commits/a75ec00f477cf4153f7369dc1f248da73f03ecd3

tests/versioning_3_test.go
carlydf: done!

service/matching/task_queue_partition_manager.go
carlydf: I couldn't get a sticky worker test working for this PR, but I will add a follow-up.

tests/versioning_3_test.go
```diff
@@ -2180,7 +2608,12 @@ func (s *Versioning3Suite) testCan(crossTq bool, behavior enumspb.VersioningBeha
 	}
 	canxTq := tv1.TaskQueue().GetName() + "_XTQ_CaN"
 
-	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	var override *workflowpb.VersioningOverride
+	if pinnedOverride {
+		override = s.makePinnedOverride(tv1)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 600*time.Second)
```
carlydf, line 2616: TODO: revert this to 60s

PR #8788, opened 2025-12-10
Enforce version limit in task queues

ShahabT:
## What changed?
Enforce a max limit on number of versions registered in a task queue.

## Why?
Safety agains user data becoming too big.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
User's polls will be rejected if server cannot allow new version registration due to the limit.

### DIFF
--- a/service/matching/matching_engine_test.go
+++ b/service/matching/matching_engine_test.go
@@ -4479,7 +4479,7 @@ func (s *matchingEngineSuite) TestSyncDeploymentUserData_CleanupOldDeletedVersio
 	deploymentName := "foo"
 
 	// First, insert versions with recent update times
-	recentTime := timestamppb.New(time.Now().Add(-29 * 24 * time.Hour)) // 29 days old (should remain)
+	recentTime := timestamppb.New(time.Now().Add(-6 * 24 * time.Hour)) // 6 days old (should remain)
 
 	versionsData := map[string]*deploymentspb.WorkerDeploymentVersionData{
 		"to-be-old-build": {
@@ -4523,8 +4523,8 @@ func (s *matchingEngineSuite) TestSyncDeploymentUserData_CleanupOldDeletedVersio
 	s.Contains(versions, "recent-build")
 	s.Contains(versions, "active-build")
 
-	// Now update "to-be-old-build" to have an old timestamp (31 days)
-	oldTime := timestamppb.New(time.Now().Add(-31 * 24 * time.Hour))
+	// Now update "to-be-old-build" to have an old timestamp (8 days)
+	oldTime := timestamppb.New(time.Now().Add(-8 * 24 * time.Hour))
 	_, err = s.matchingEngine.SyncDeploymentUserData(context.Background(), &matchingservice.SyncDeploymentUserDataRequest{
 		NamespaceId:    namespaceID,
 		TaskQueue:      tq,


service/matching/config.go
```diff
@@ -203,7 +204,8 @@ type (
 
 		FairnessCounter func() counter.CounterParams
 
-		loadCause loadCause
+		loadCause              loadCause
+		MaxVersionsInTaskQueue func() int
```
dnr: can you keep these settings consistent in the four places they have to appear? like put this right underneath MaxTaskQueuesInDeployment for all of them

common/worker_versioning/worker_versioning.go
```diff
@@ -1004,3 +1020,53 @@ func WorkerDeploymentVersionFromStringV32(s string) (*deploymentspb.WorkerDeploy
...
+		// Stop if:
+		// 1. This version is not older than 7 days AND
+		// 2. We're not exceeding the limit because of deleted versions
+		if !dv.updateTime.Before(aWeekAgo) && totalCount <= maxVersions {
+			break
+		}
+
+		// Remove this deleted version
+		delete(deploymentData.Versions, dv.buildID)
+		deletedVersions = deletedVersions[1:] // Remove from slice to keep count accurate
```
marius-swfy: A more idiomatic way will be to use a dedicated variable to know the total number of elements that were deleted.
Example:
```go
cleaned := false
deletedRemaining := len(deletedVersions)
for _, dv := range deletedVersions {
    totalCount := undeletedCount + deletedRemaining

    if !dv.updateTime.Before(aWeekAgo) && totalCount <= maxVersions {
        break
    }

    delete(deploymentData.Versions, dv.buildID)
    deletedRemaining--
    cleaned = true
}
```

service/matching/config.go
ShahabT: not all the versioning settings appear in per-tq config. but I moved this one around to make things more consistent.

common/dynamicconfig/constants.go
```diff
@@ -1319,6 +1319,12 @@ these log lines can be noisy, we want to be able to turn on and sample selective
 		100,
 		`MatchingMaxVersionsInDeployment represents the maximum number of versions that can be registered in a single worker deployment`,
 	)
+	MatchingMaxVersionsInTaskQueue = NewNamespaceIntSetting(
+		"matching.maxVersionsInTaskQueue",
+		200,
+		`MatchingMaxVersionsInTaskQueue represents the maximum number of versions that can be registered in a single task queue. 
+ Should be larger than MatchingMaxVersionsInDeployment because a task queue can be in versions spanning across more than one deployments.`,
+	)
```
Shivs11, line 1327: curious: was 200 just chosen arbitrarily?

common/worker_versioning/worker_versioning.go
```diff
@@ -329,6 +330,22 @@ func HasDeploymentVersion(deployments *persistencespb.DeploymentData, v *deploym
 	return false
 }
 
+func CountDeploymentVersions(deployments *persistencespb.DeploymentData) int {
```
Shivs11, line 313: nit: CountDeploymentVersions could be renamed to something like CountDeploymentVersionsInTaskQueue or something to signify that we are interested in finding out the number of versions present in the task queue 

Right now, from a very initial glance, one could just think it's counting the versions from some other entity maybe

common/worker_versioning/worker_versioning.go
```diff
@@ -1004,3 +1021,52 @@ func WorkerDeploymentVersionFromStringV32(s string) (*deploymentspb.WorkerDeploy
 		BuildId:        after,
 	}, nil
 }
+
+// CleanupOldDeletedVersions removes versions deleted more than 7 days ago. Also removes more deleted versions if
```
Shivs11, line 992: note: we should also mention the fact that older, stale versions are only deleted from this task queue iff there is some caller calling SyncDeploymentUserData for this specific task queue and is not a recurring/background GC process. 

This makes me wonder that in the worst case, if no one were to call this Sync method, there could be a world where a task queue just has a bunch of versions and would still keep on serving tasks to pollers. This could happen if someone were to not make a new deployment and someone had made a bunch of deployments in the past 7 days (I could see this coming up in our internal dogfooding environments)

Can we/Should we make this an automatic process? How much additional work would that be? If not, maybe we should document this fact here on this function since if there is a bottleneck tomorrow, we know where to look at

tests/versioning_3_test.go
```diff
@@ -4570,3 +4570,78 @@ func (s *Versioning3Suite) TestWorkflowRetry_AutoUpgrade_AfterCAN_NoBounceBack()
...
+		"Error message should include the limit value")
+}
+
+func (s *Versioning3Suite) skipBeforeVersion(version workerdeployment.DeploymentWorkflowVersion) {
+	if s.deploymentWorkflowVersion < version {
+		s.T().Skipf("test supports workflow version %v and newer", version)
+	}
+}
+
+func (s *Versioning3Suite) skipFromVersion(version workerdeployment.DeploymentWorkflowVersion) {
```
Shivs11, line 4676: this guy seems to not be used anywhere

tests/versioning_3_test.go
```diff
@@ -4570,3 +4570,78 @@ func (s *Versioning3Suite) TestWorkflowRetry_AutoUpgrade_AfterCAN_NoBounceBack()
...
+		deploymentName := tvVersion.DeploymentVersion().GetDeploymentName()
+		_, err := s.GetTestCluster().MatchingClient().SyncDeploymentUserData(
+			ctx, &matchingservice.SyncDeploymentUserDataRequest{
+				NamespaceId:        s.NamespaceID().String(),
+				TaskQueue:          tv.TaskQueue().GetName(),
+				TaskQueueTypes:     []enumspb.TaskQueueType{tqTypeWf},
+				DeploymentName:     deploymentName,
+				UpsertVersionsData: upsertVersions,
+			},
+		)
```
Shivs11, line 4640: nit: rather than doing this, you could also just have 5 different pollers, each coming from different deployments, poll on the task queue right?

common/worker_versioning/worker_versioning.go
ShahabT: > if no one were to call this Sync method, there could be a world where a task queue just has a bunch of versions and would still keep on serving tasks to pollers.

the tasks will be served if the version is not deleted only. and if it's not deleted, then event automatic CG would not touch it.

tests/versioning_3_test.go
ShahabT: yeah I felt this way is more deterministic and less chance of flakes.

common/worker_versioning/worker_versioning.go
ShahabT: well the input is something that lives in per-tq user data so not one can call it for counting something else.

common/dynamicconfig/constants.go
ShahabT: it's twice MatchingMaxVersionsInDeployment so in case someone moves to another deployment, they'd get "maximum versions in deployment exceeded" before getting "maximum versions in task queue exceeded" which is harder for user to deal with because there is no direct way to see all the versions in a TQ.

common/worker_versioning/worker_versioning.go
Shivs11: yep, I get that - my concern was about to this being a bottleneck when it comes to delivery of tasks if noone has called the Sync call in like a week or something and the task queue has like 199 versions (Theoretically possible, but low probability)

tests/versioning_3_test.go
Shivs11: not sure i understand why using pollers would not make it deterministic though - you just start 5 pollers, wait for those to appear in the task queue and you then proceed with the test.

PR #8791, opened 2025-12-10
Reject versioning override when version does not exist

Shivs11:
## What changed?
- WISOTT
- Also added a cache per history host so that we don't overburden matching with these calls.
- Also added a whole new unit test testing the function `ValidateVersioningOverride`
- TODO in a follow-up PR: add metrics for this cache. Doing this as a follow-up in the interest of time but have it tracked in JIRA.

## Why?
- Versioning correctness.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
- I don't think this is risky

---

> [!NOTE]
> Validate versioning overrides by ensuring pinned versions exist in the task queue via matching RPC with a per-host cache, refactor validation into history APIs, and add configs and tests.
> 
> - **Worker Versioning / Validation**:
>   - Add `ValidateVersioningOverride(ctx, ...)` to verify pinned versions exist in a task queue via `matching.CheckTaskQueueVersionMembership`.
>   - Introduce per-host `versionMembershipCache` to cache membership results; reject with `FailedPrecondition` if not present.
>   - Remove frontend/batcher inline override validation; perform it in history layer (start, signal-with-start, update options, reset post-ops).
> - **History Service Wiring**:
>   - Thread `matchingClient` and `versionMembershipCache` through history engine, starter, multi-op, signal-with-start, reset, and update-workflow-options APIs.
>   - Mark batch UpdateWorkflowOptions non-retryable for "Pinned version is not present in the task queue".
> - **Config / Dynamic Config**:
>   - Add `history.versionMembershipCacheTTL` and `history.versionMembershipCacheMaxSize`; provide cache in `fx` with lifecycle management.
> - **Testing**:
>   - Add unit tests for `ValidateVersioningOverride` covering cache hits/misses and v0.31/v0.32 paths.
>   - Extend functional tests to assert membership checks, cache behavior, batch update failures, and reset with post-reset options.
>   - Test helpers: `TestVars.WithDeploymentSeries`, `WithBuildID`, and utilities to ensure versions are present via matching RPC.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d8b755273fe41c823a30c4cc2ecbbeb2c2be124d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/frontend/workflow_handler.go
```diff
@@ -4822,6 +4793,8 @@ func (wh *WorkflowHandler) StartBatchOperation(
 	case *workflowservice.StartBatchOperationRequest_UpdateWorkflowOptionsOperation:
 		input.BatchType = enumspb.BATCH_OPERATION_TYPE_UPDATE_EXECUTION_OPTIONS
 		identity = op.UpdateWorkflowOptionsOperation.GetIdentity()
+		// TODO (Shivam): Verify if this is the correct way to handle these errors.
+		input.NonRetryableErrors = append(input.NonRetryableErrors, "Pinned version is not present in the task queue")
```
Shivs11, line 4800: would love to get some clarity over here - theoretically, this works as I don't want the internal batch workflow's activity to retry on these errors. But should something like this be done in this manner is a different question that I am trying to answer.

Would love to get someone's opinion here - have also mentioned this in our slack channel

tests/workflow_alias_search_attribute_test.go
```diff
@@ -114,7 +167,7 @@ func (s *WorkflowAliasSearchAttributeTestSuite) TestWorkflowAliasSearchAttribute
 	s.Require().NoError(err)
 }
 
-func (s *WorkflowAliasSearchAttributeTestSuite) TestWorkflowAliasSearchAttribute_CustomSearchAttributeOverride() {
+func (s *WorkflowAliasSearchAttributeTestSuite) TestWorkflowAliasSearchAttribute_CustomSAOverride() {
```
Shivs11: had to make the test name shorter here otherwise there were registration errors in our worker-versioning entity workflows since we have some length checks

service/frontend/workflow_handler.go
```diff
@@ -494,11 +494,6 @@ func (wh *WorkflowHandler) prepareStartWorkflowRequest(
 		return nil, err
 	}
 
-	// Validate versioning override, if present.
-	if err := worker_versioning.ValidateVersioningOverride(request.GetVersioningOverride()); err != nil {
-		return nil, err
-	}
-
```
Shivs11, line 497: disclaimer for reviewer: all validation checks that you see removed from the frontend, in this PR, are simply moved from frontend to history.

Coincidentally, I had worked on [this](https://github.com/temporalio/temporal/pull/8727) which had initially moved them from history to frontend. Little did I know that I was wrong and that the authors of these changes had some incredible foresight (or maybe they did not and were just lucky)

common/worker_versioning/worker_versioning.go
```diff
@@ -552,7 +592,12 @@ func ValidateVersioningOverride(override *workflowpb.VersioningOverride) error {
 			return ValidateDeployment(override.GetDeployment())
```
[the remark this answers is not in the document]
Shivs11: this is gonna be removed soon anyways so ignoring.

service/history/api/resetworkflow/api.go
```diff
@@ -56,6 +63,16 @@ func Invoke(
 		return nil, serviceerror.NewInvalidArgument("Workflow task finish ID must be > 1 && <= workflow last event ID.")
 	}
 
+	// Validate versioning override, if any.
+	err = validatePostResetOperationInputs(ctx, request.GetPostResetOperations(), matchingClient, versionMembershipCache,
+		&taskqueuepb.TaskQueue{
```
ShahabT: just pass task queue name.

Also, I'm assuming you double checked that `GetTaskQueue` does not return sticky and always normal queue name.

service/history/history_engine.go
```diff
@@ -313,6 +315,10 @@ func NewEngineWithShardContext(
 		replicationTaskExecutorProvider,
 		dlqWriter,
 	)
+
+	historyEngImpl.versionMembershipCache = cache.New(10000, &cache.Options{
```
ShahabT: I think the size should come from config too. In the past, non configurable cache sizes caused issues in produciton.

Also, can we have cache hit and mix metrics if the generic cache implementation does not provide it already? (so we can later reason about the size and ttl and adjust them)

ShahabT: More importantly, the cache should be per node, not per shard, it seems we need to move it out of history_engine to somewhere else. I'm not sure where to but @prathyushpv @yycptt should know.

common/worker_versioning/worker_versioning.go
```diff
@@ -528,7 +524,53 @@ func ExtractVersioningBehaviorFromOverride(override *workflowpb.VersioningOverri
...
+	pinnedVersion *deploymentpb.WorkerDeploymentVersion,
+	matchingClient resource.MatchingClient,
+	versionMembershipCache cache.Cache,
+	tq string,
+	tqType enumspb.TaskQueueType,
+	namespaceID string) error {
+
+	// Check if we have recently queried matching recently to validate if this version exists in the task queue.
+	// Key format: namespaceID:taskQueue:DeploymentName:BuildID
+	key := fmt.Sprintf("%s:%s:%s:%s", namespaceID, tq, pinnedVersion.DeploymentName, pinnedVersion.BuildId)
```
ShahabT: does the key have to be string? if not we should pass a struct here with the 4 fields. because this concatenation is not safe without escaping `:` from `tq`.

ShahabT: Also struct might be more efficient than string concat

common/worker_versioning/worker_versioning.go
```diff
@@ -528,7 +524,51 @@ func ExtractVersioningBehaviorFromOverride(override *workflowpb.VersioningOverri
...
+			"Pinned version is not present in the task queue",
+		)
+	}
+
+	resp, err := matchingClient.CheckTaskQueueVersionMembership(context.Background(), &matchingservice.CheckTaskQueueVersionMembershipRequest{
+		NamespaceId:   namespaceID,
+		TaskQueue:     tq,
+		TaskQueueType: tqType,
+		Version:       DeploymentVersionFromDeployment(DeploymentFromExternalDeploymentVersion(pinnedVersion)),
+	})
```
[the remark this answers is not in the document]
ShahabT: this seems a valid concern

common/worker_versioning/worker_versioning.go
```diff
@@ -528,7 +524,53 @@ func ExtractVersioningBehaviorFromOverride(override *workflowpb.VersioningOverri
...
+	// Check if we have recently queried matching recently to validate if this version exists in the task queue.
+	// Key format: namespaceID:taskQueue:DeploymentName:BuildID
+	key := fmt.Sprintf("%s:%s:%s:%s", namespaceID, tq, pinnedVersion.DeploymentName, pinnedVersion.BuildId)
+	if cached := versionMembershipCache.Get(key); cached != nil {
+		if isMember, ok := cached.(bool); ok && isMember {
+			return nil
+		}
+		return serviceerror.NewFailedPrecondition(
+			"Pinned version is not present in the task queue",
+		)
```
ShahabT: the false case should also be within `ok`

```suggestion
		if isMember, ok := cached.(bool); ok {
		    if isMember {
   			  return nil
   			}
   			return serviceerror.NewFailedPrecondition(
			   "Pinned version is not present in the task queue",
		    )
		}
```

common/worker_versioning/worker_versioning.go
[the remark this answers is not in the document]
Shivs11: lol - I made the changes to pass in the context to this function but then did not end up using it. thanks.

service/history/api/resetworkflow/api.go
Shivs11: I looked again, and no the sticky name of a task queue should not be passed here by this method. 

Inside executions.pb, I see that the WorkflowExecutionInfo struct contains both fields for TaskQueue and StickyTaskQueue name. The GetTaskQueue method is the accessor only for the TaskQueue name, so we should be ok.

service/history/history_engine.go
Shivs11: yeah thought about this a bit more and here's what I think: 

Since the keys of the cache (namespace, task queue, deploymentName, buildID) are and should be unique across each history shard, this cache can be moved one level higher. The only thing that I am worried about is if this cache just serves a bunch of concurrent requests and becomes a bottleneck. Would love to have an answer to this question.

Shivs11: As for the location of the cache, we can just move this up one level and place it inside the fx constructor i believe

common/worker_versioning/worker_versioning.go
Shivs11: not sure how we could ever get a corrupted cache where we would not have a boolean as the cache value since we are controlling the writes that go to a cache - maybe the cache could be corrupted?

not sure, anyways, i am okay to make this change if things are safer.

service/history/history_engine.go
Shivs11: In my latest commit, I have moved the cache so that it is present in each history node rather than shard. LMK what y'all think.

service/history/fx.go
```diff
@@ -346,3 +358,19 @@ func ReplicationProgressCacheProvider(
 ) replication.ProgressCache {
 	return replication.NewProgressCache(serviceConfig, logger, handler)
 }
+
+func VersionMembershipCacheProvider(
+	lc fx.Lifecycle,
+	serviceConfig *configs.Config,
+) commoncache.Cache {
+	c := commoncache.New(10000, &commoncache.Options{
```
ShahabT: let's have a config for cache capacity

common/worker_versioning/worker_versioning.go
```diff
@@ -526,7 +527,66 @@ func ExtractVersioningBehaviorFromOverride(override *workflowpb.VersioningOverri
...
+	versionMembershipCache.Put(key, resp.GetIsMember())
+	if !resp.GetIsMember() {
+		return serviceerror.NewFailedPrecondition(
+			"Pinned version is not present in the task queue",
+		)
+	}
+	return nil
+}
+
+type versionMembershipCacheKey struct {
```
prathyushpv, line 577: Don't we need taskQueueType in cache key? Can version membership differ in two task queues with different types?

common/dynamicconfig/constants.go
```diff
@@ -2707,6 +2707,18 @@ to the CHASM (V2) implementation on active scheduler workflows.`,
 instead of the previous HSM backed implementation.`,
 	)
 
+	VersionMembershipCacheTTL = NewGlobalDurationSetting(
+		"history.versionMembershipCacheTTL",
+		1*time.Second,
```
prathyushpv, line 2712: We may not get enough cache reuse if it is too short. Is there any reason to keep it this short?

Shivs11: I chose one second because in theory, you could have a version removed from a task-queue. If we were to keep this value to a large number, I worry that someone could incorrectly set a versioning override on a workflow for a version that was deleted. 

I think we should be conservative here with this value since I think we should value correctness. What I could do is observe the number of times this cache was hit (i should add metrics but maybe shall ship out a new PR for that), and then if we realize that the number of cache hits have been too low, we could change this value using the DC. 

WDYT

common/worker_versioning/worker_versioning.go
Shivs11: oh! good catch and a big thank you.
making the change.

Shivs11: actually, investigating this a bit more i realize the following:

all the call-sites to the function ValidateVersioningOverride are workflow start/reset/update-options/signal-with-start. These, I think, should be passing in the workflow task queue type all the time to this function. Maybe it's ok right now to not have the task queue type field in the cache key but for the future, if someone were to query this using the activity task queue type, this could come handy.

ShahabT: > for the future, if someone were to query this using the activity task queue type, this could come handy.

this, since validatePinnedVersionInTaskQueue is receiving type tomorrow someone may call it for activity without paying attention to caching logic. to be safe and consistent we should add the key now, or remove the type from function inputs.

common/dynamicconfig/constants.go
prathyushpv: Ok. Sounds good!

PR #8793, opened 2025-12-10
Add batch workflow refresh tasks

awln-temporal:
## What changed?
Add batch workflow refresh tasks Admin API.

Add tdbg commands to invoke `StartAdminBatchOperation`.

## Details

AdminBatchOperations are executed in the Batched System Worker, similar to existing Frontend BatchOperations. The worker activity will fetch pages of workflow executions to perform operations on, and periodically heartbeat its results. For RefreshWorkflowTasks, each individual Refresh call will go through the admin handler.

Refreshing Workflow Tasks regenerates all pending tasks of an execution given its mutable state.

## CLI usage

```
tdbg workflow batch-refresh-tasks \
  --namespace <ns> \
  --query "WorkflowType='MyType'" \
  --reason "fixing stuck workflows" \
  [--job-id <optional-job-id>] \
  [--archetype <optional-archetype>]
```

## Why?
Unblock new Matcher migration and allow for general use case batch Admin calls. Currently, only supports `BatchOperationRefreshWorkflowTasks`.

## How did you test it?
- [X] built
- [X] added new unit test(s)


yycptt:
Haven't review the actual changes yet but for the tdbg experience, let’s match what we have in the temporal CLI. so something like `tdbg workflow refresh-tasks -q '<query>' means to start a batch operation. and if -q is not specified and --wid is specified, then it's not a batch op, but only operating on that single workflow.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -604,3 +604,41 @@ message ForceUnloadTaskQueuePartitionRequest {
...
   bool was_loaded = 1;
 }
+
+// StartAdminBatchOperationRequest starts an admin batch operation.
+message StartAdminBatchOperationRequest {
+  // Namespace that contains the batch operation.
+  string namespace = 1;
+
+  // Visibility query defines the group of workflows to apply the batch operation.
+  // This field and `executions` are mutually exclusive.
```
yycptt: I think we will want to support a list of executions as well, that can be quite useful during incident mitigation, but can be done as a follow up item.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -604,3 +604,41 @@ message ForceUnloadTaskQueuePartitionRequest {
...
+    BatchOperationRefreshWorkflowTasks refresh_workflow_tasks_operation = 10;
+  }
+}
+
+message StartAdminBatchOperationResponse {
+}
+
+// BatchOperationRefreshWorkflowTasks refreshes workflow tasks for batch workflows.
+// This regenerates all pending tasks for each workflow.
+message BatchOperationRefreshWorkflowTasks {
```
yycptt: plz remove `Workflow` from the names

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -604,3 +604,41 @@ message ForceUnloadTaskQueuePartitionRequest {
...
+  // The identity of the worker/client.
+  string identity = 1;
+
+  // Archetype to use for the refresh operation. If not specified, defaults to workflow archetype.
+  string archetype = 2;
```
yycptt: hmm why do we need this? I mean the visibility records contain the archetype information. refresh tasks may be applied to executions of different archetypes as well if we have some general task lost bug in queue processing logic.

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -22,4 +23,7 @@ message BatchOperationInput {
 
   // The request to start the batch operation.
   temporal.api.workflowservice.v1.StartBatchOperationRequest request = 7;
+
+  // The request to start an admin batch operation.
```
yycptt, line 29: plz comment that this field and `request` are mutually exclusive.
plz also add a comment for batch_type and say that it only applies to `request`

common/resource/fx.go
```diff
@@ -246,6 +253,18 @@ func FrontendClientProvider(clientBean client.Bean) workflowservice.WorkflowServ
 	)
 }
 
+func AdminRawClientProvider(clientBean client.Bean, clusterMetadata cluster.Metadata) (AdminRawClient, error) {
+	return clientBean.GetRemoteAdminClient(clusterMetadata.GetCurrentClusterName())
+}
```
yycptt: nit: this can be done inside `AdminClientProvider`?

service/frontend/admin_handler.go
```diff
@@ -1568,6 +1573,132 @@ func (adh *AdminHandler) RefreshWorkflowTasks(
...
+	// Validate concurrent batch operation
+	maxConcurrentBatchOperation := adh.config.MaxConcurrentBatchOperation(request.GetNamespace())
+	countResp, err := adh.visibilityMgr.CountWorkflowExecutions(ctx, &manager.CountWorkflowExecutionsRequest{
+		NamespaceID: namespaceID,
+		Namespace:   namespace.Name(request.GetNamespace()),
+		Query:       batcher.OpenBatchOperationQuery,
+	})
+	if err != nil {
+		return nil, err
+	}
```
yycptt, line 1605: I don't think this is necessary for admin batch op, but fine having that as well.

But I think more importantly we cannot let admin batch op taking user batch op slots, so we will need another TemporalNamespaceDivision value on L1645 to exclude admin batch op from user `OpenBatchOperationQuery`.
Having a different TemporalNamespaceDivision also prevents user ListBatchOperations request from seeing admin batch op. 

Using another search attributes works as well as long as it can separate user and admin batch op, though  a new, say, `TemporalAdminBatcher`  namespace division value seems the easiest approach to me. Please bring this up in the standup and see if team has concerns with it. The migration and ns deletion operation works for any namespace division value, so there shouldn't be any concern there.

service/worker/batcher/activities.go
```diff
@@ -353,6 +368,23 @@ func (a *activities) adjustQueryBatchTypeEnum(query string, batchType enumspb.Ba
...
+	if len(query) == 0 {
+		return query
+	}
+
+	switch adminReq.Operation.(type) {
+	case *adminservice.StartAdminBatchOperationRequest_RefreshWorkflowTasksOperation:
+		// RefreshWorkflowTasks only for running workflows.
+		// Non-running workflows will be a no-op at the history level but still
+		// incur gRPC and DB costs, so filter them out.
+		return fmt.Sprintf("(%s) AND (%s)", query, statusRunningQueryFilter)
```
yycptt: refresh task may be needed for closed workflow as well. Some tasks are only generated on execution close.

service/worker/batcher/activities.go
```diff
@@ -551,22 +590,57 @@ func startTaskProcessorProtobuf(
...
+) error {
+	adminReq := batchOperation.AdminRequest
+	switch op := adminReq.Operation.(type) {
+	case *adminservice.StartAdminBatchOperationRequest_RefreshWorkflowTasksOperation:
+		return processTask(ctx, limiter, task,
+			func(execution *commonpb.WorkflowExecution) error {
+				_, err := a.AdminClient.RefreshWorkflowTasks(ctx, &adminservice.RefreshWorkflowTasksRequest{
+					NamespaceId: batchOperation.NamespaceId,
+					Execution:   execution,
+					Archetype:   op.RefreshWorkflowTasksOperation.GetArchetype(),
```
yycptt: let's retrieve this from visibility records itself. You can follow what I did in this PR: https://github.com/temporalio/temporal/pull/8768/changes#diff-dc576c64d23020364cc358402a580e400ec6f3c18015af377e2540db66fbaafaR186.

I am also ok with calling history client refresh tasks api directly (so that you don't have to convert archetypeID to archetype name) as well as long as the outbound call is captured by OPS rate limiter.

Changing the data structure used in the activity's pagination logic shouldn't cause any compatibility issue as everything is in memory.

yycptt: Speaking of OPS rate limiter, let's use CallerTypePreemptable when making the call. Probably set it at the beginning of the activity when this is a admin batch op.

yycptt:
After this PR is landed, CGS can considering adding a batch force migrate command as well.
cc @yux0 @xwduan @hai719

common/resource/fx.go
awln-temporal: yea can merge them, was just following existing pattern.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
awln-temporal: yeah fair enough, will remove and have batcher activity use `TemporalNamespaceDivision` from WorkflowExecutions as archetypeID.

awln-temporal:
> Haven't review the actual changes yet but for the tdbg experience, let’s match what we have in the temporal CLI. so something like `tdbg workflow refresh-tasks -q '' means to start a batch operation. and if -q is not specified and --wid is specified, then it's not a batch op, but only operating on that single workflow.

Do we care about setting Memos for the Admin Batch Operation, eg. `Reason`, `JobID`, that are common with User Batch Operations, but not usable for single RefresWorkflowTasks ops?

service/frontend/admin_handler.go
awln-temporal: will discuss with Prathyush and team as a follow up.

service/worker/batcher/activities.go
```diff
@@ -551,22 +590,57 @@ func startTaskProcessorProtobuf(
...
+	batchOperation *batchspb.BatchOperationInput,
+	task task,
+	limiter *rate.Limiter,
+) error {
+	adminReq := batchOperation.AdminRequest
+	switch op := adminReq.Operation.(type) {
+	case *adminservice.StartAdminBatchOperationRequest_RefreshWorkflowTasksOperation:
+		return processTask(ctx, limiter, task,
+			func(execution *commonpb.WorkflowExecution) error {
+				_, err := a.AdminClient.RefreshWorkflowTasks(ctx, &adminservice.RefreshWorkflowTasksRequest{
```
dnr: just to be clear, we're okay with any user being able to do this at any time?

refresh tasks is safe, but my concern is that someone is going to add a more dangerous admin operation here and not realize that it's exposed to all users in the namespace.

I would say we need a big warning comment somewhere but I'm not sure where it should go.

service/frontend/admin_handler.go
```diff
@@ -1568,6 +1573,132 @@ func (adh *AdminHandler) RefreshWorkflowTasks(
...
+		WorkflowType:             &commonpb.WorkflowType{Name: batcher.BatchWFTypeProtobufName},
+		TaskQueue:                &taskqueuepb.TaskQueue{Name: primitives.PerNSWorkerTaskQueue},
+		Input:                    inputPayload,
+		Identity:                 identity,
+		RequestId:                uuid.NewString(),
+		WorkflowIdConflictPolicy: enumspb.WORKFLOW_ID_CONFLICT_POLICY_FAIL,
+		WorkflowIdReusePolicy:    enumspb.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE,
+		Memo:                     memo,
+		SearchAttributes:         searchAttributes,
+		Priority:                 &commonpb.Priority{}, // ie default priority
```
dnr: just leave this out? why include it if it's the default

service/worker/batcher/activities.go
awln-temporal: I can add a big warning to the proto definition

yycptt: I think we will need a general way for preventing user from starting this workflow directly, even for normal operations like signal/terminate etc, which allows them to bypass, for example, the concurrent limit.

service/worker/batcher/workflow.go
```diff
@@ -59,6 +59,11 @@ var (
 		sadefs.ExecutionStatus,
 		int(enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING),
 	)
+
+	OpenAdminBatchOperationQuery = fmt.Sprintf("%s = '%s'",
+		sadefs.TemporalNamespaceDivision,
+		AdminNamespaceDivision,
+	)
```
yycptt: hmm we want open workflow only right?

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -604,3 +604,39 @@ message ForceUnloadTaskQueuePartitionRequest {
...
+}
+
+message StartAdminBatchOperationResponse {
+}
+
+// BatchOperationRefreshTasks refreshes tasks for batch executions.
+// This regenerates all pending tasks for each execution.
+message BatchOperationRefreshTasks {
+  // The identity of the worker/client.
+  string identity = 1;
```
yycptt: I actually don't quite follow why today identity is a per operation field in user batch op request. All operations have it. I think we can move it into the start batch request.

proto/internal/temporal/server/api/batch/v1/request_response.proto
```diff
@@ -18,8 +19,12 @@ message BatchOperationInput {
...
+  // Only needed if StartBatchOperationRequest request is set.
   temporal.api.enums.v1.BatchOperationType batch_type = 6;
 
-  // The request to start the batch operation.
+  // The request to start the batch operation. Mutually exclusive with StartAdminBatchOperationRequest admin_request.
   temporal.api.workflowservice.v1.StartBatchOperationRequest request = 7;
+
+  // The request to start an admin batch operation. Mutually exclusive with StartBatchOperationRequest request.
+  temporal.server.api.adminservice.v1.StartAdminBatchOperationRequest admin_request = 8;
 }
```
yycptt, line 32: nit: plz help add a newline :)

service/frontend/admin_handler.go
```diff
@@ -1568,6 +1573,131 @@ func (adh *AdminHandler) RefreshWorkflowTasks(
...
+	input := &batchspb.BatchOperationInput{
+		AdminRequest: request,
+		NamespaceId:  namespaceID.String(),
+	}
+
+	var identity string
+	var batchTypeMemo string
+	switch op := request.Operation.(type) {
+	case *adminservice.StartAdminBatchOperationRequest_RefreshTasksOperation:
+		batchTypeMemo = "refresh_workflow_tasks"
```
yycptt: nit: remove `workflow`

service/worker/common/chasm_util.go
yycptt, line 1: oh I just got my PR landed, please rebase.

service/worker/batcher/activities.go
```diff
@@ -551,37 +590,80 @@ func startTaskProcessorProtobuf(
...
+	adminReq := batchOperation.AdminRequest
+	switch adminReq.Operation.(type) {
+	case *adminservice.StartAdminBatchOperationRequest_RefreshTasksOperation:
+		return processTask(ctx, limiter, task,
+			func(executionInfo *workflowpb.WorkflowExecutionInfo) error {
+				archetypeID, err := workercommon.ArchetypeIDFromExecutionInfo(executionInfo)
+				if err != nil {
+					return fmt.Errorf("archetypeID extraction error: %w", err)
+				}
+				ctx = headers.SetCallerType(ctx, headers.CallerTypePreemptable)
```
yycptt: nit: I'd see if this can be done only once per activity instead of per execution. maybe on L304?

yycptt:
> > Haven't review the actual changes yet but for the tdbg experience, let’s match what we have in the temporal CLI. so something like `tdbg workflow refresh-tasks -q '' means to start a batch operation. and if -q is not specified and --wid is specified, then it's not a batch op, but only operating on that single workflow.
> 
> Do we care about setting Memos for the Admin Batch Operation, eg. `Reason`, `JobID`, that are common with User Batch Operations, but not usable for single RefresWorkflowTasks ops?

I think it's nice to have them (reason and batch type) in the memo. `JobID` is not part of the memo I believe.

common/dynamicconfig/constants.go
```diff
@@ -991,6 +991,11 @@ so forwarding by endpoint ID will not work out of the box.`,
 		true,
 		`FrontendEnableBatcher enables batcher-related RPCs in the frontend`,
 	)
+	FrontendMaxConcurrentAdminBatchOperationPerNamespace = NewNamespaceIntSetting(
```
prathyushpv, line 984: Where is this config used?

service/worker/batcher/workflow.go
awln-temporal: yea...

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -604,3 +604,39 @@ message ForceUnloadTaskQueuePartitionRequest {
 message ForceUnloadTaskQueuePartitionResponse {
   bool was_loaded = 1;
 }
+
+// StartAdminBatchOperationRequest starts an admin batch operation.
+// WARNING: Batch Operations are exposed to all users of the namespace. Admin Batch Operations should be exercised with caution.
+message StartAdminBatchOperationRequest {
```
prathyushpv, line 610: I was wondering if a field to specify batch operation RPS would be helpful in input. We have `max_operations_per_second` in `StartBatchOperationRequest`. We can use this field to control the rate. It may not be strictly necessary since we are marking this batch operation with the lowest priority, and it should not block other operations.
cc: @yycptt

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
awln-temporal: some operations do not use the field for some reason, but maybe we can move this above for admin batch ops.

service/worker/batcher/activities.go
awln-temporal: yea fair enough, i just thought setting the header near where the call is made would make it easier for others to debug

common/dynamicconfig/constants.go
awln-temporal: updated ref

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
awln-temporal: looks like this `max_operations_per_second` field isn't properly handled by the activity? It seems to always default to the Dynamic Config RPS value https://github.com/temporalio/temporal/blob/admin-batch-refresh-workflows/service/worker/batcher/fx.go#L89.

yycptt: yeah looks like it's not used at all right now and we just rely on the dc value to protect us.

service/worker/batcher/activities.go
yycptt: yeah we can add a comment here.

tools/tdbg/tdbg_commands.go
```diff
@@ -185,6 +185,36 @@ func newAdminWorkflowCommands(clientFactory ClientFactory, prompterFactory Promp
 				return AdminRefreshWorkflowTasks(c, clientFactory)
 			},
 		},
+		{
+			Name:    "batch-refresh-tasks",
```
yycptt: let's make it a flag `-q <visibility_query>` in the existing `workflow refresh-task` command.

awln-temporal: done

tools/tdbg/tdbg_commands.go
```diff
@@ -180,9 +180,23 @@ func newAdminWorkflowCommands(clientFactory ClientFactory, prompterFactory Promp
 					Usage:       "Fully qualified archetype name of the execution",
 					DefaultText: chasm.WorkflowArchetype,
 				},
+				&cli.StringFlag{
+					Name:     FlagVisibilityQuery,
+					Usage:    "Visibility query to select workflows",
+					Required: true,
```
yycptt: hmm why it's required? Operator can either do batch or just refresh task for one execution.

tools/tdbg/commands.go
```diff
@@ -635,6 +635,13 @@ func AdminDescribeHistoryHost(c *cli.Context, clientFactory ClientFactory) error
 	return nil
 }
 
+func adminRefreshWorkflowTasks(c *cli.Context, clientFactory ClientFactory) error {
+	if c.IsSet(FlagVisibilityQuery) {
```
yycptt: nit: validate workflowID and runID are not specified.

service/worker/deployment/deployment_reachability_test.go
yycptt, line 1: looks like git history got messed up a bit.

tools/tdbg/commands.go
```diff
@@ -675,6 +682,52 @@ func AdminRefreshWorkflowTasks(c *cli.Context, clientFactory ClientFactory) erro
...
+	if jobID == "" {
+		jobID = fmt.Sprintf("batch-refresh-%d", time.Now().UnixNano())
+	}
+
+	ctx, cancel := newContext(c)
+	defer cancel()
+
+	_, err = adminClient.StartAdminBatchOperation(ctx, &adminservice.StartAdminBatchOperationRequest{
+		Namespace:       nsName,
+		VisibilityQuery: query,
```
yycptt, line 732: not related to this PR. 

Does visibility query parser support `BusinessID` today? I mean we added support for each component to define it's own alias for the underlying WorkflowID column, but we will also need a default alias I think. Very low priority though.

tools/tdbg/commands.go
awln-temporal: behavior of tctl commands is to check visibility query and ignore validation of `workflowID/runID`, but I can make this change.

tools/tdbg/commands.go
awln-temporal: At least right now, `WorkflowID` should be able to be used as a default system search attribute to retrieve the BusinessID, we never made the change to disallow querying against `WorkflowID` yet, since that might impact scheduler query backwards compatibility.

tools/tdbg/commands.go
```diff
@@ -635,6 +636,13 @@ func AdminDescribeHistoryHost(c *cli.Context, clientFactory ClientFactory) error
 	return nil
 }
 
+func adminRefreshWorkflowTasks(c *cli.Context, clientFactory ClientFactory, prompter *Prompter) error {
+	if c.IsSet(FlagVisibilityQuery) && !c.IsSet(FlagWorkflowID) && !c.IsSet(FlagRunID) {
```
yycptt, line 643: nit: error out if both are provided.

tools/tdbg/commands.go
```diff
@@ -675,6 +683,66 @@ func AdminRefreshWorkflowTasks(c *cli.Context, clientFactory ClientFactory) erro
...
+	// Count workflows matching the query to confirm with user
+	countResp, err := workflowClient.CountWorkflowExecutions(ctx, &workflowservice.CountWorkflowExecutionsRequest{
+		Namespace: nsName,
+		Query:     query,
+	})
+	if err != nil {
+		return fmt.Errorf("unable to count workflow executions: %w", err)
+	}
+
+	msg := fmt.Sprintf("Will refresh workflow tasks for %d workflow(s) matching query %q in namespace %q. Continue Y/N?",
```
yycptt:
```suggestion
	msg := fmt.Sprintf("Will refresh tasks for %d execution(s) matching query %q in namespace %q. Continue Y/N?",
```

tools/tdbg/commands.go
yycptt: oh I didn't mean to disallow WorkflowID. just we need to allow BusinessID.

tools/tdbg/commands.go
```diff
@@ -675,6 +683,66 @@ func AdminRefreshWorkflowTasks(c *cli.Context, clientFactory ClientFactory) erro
...
+		Operation: &adminservice.StartAdminBatchOperationRequest_RefreshTasksOperation{
+			RefreshTasksOperation: &adminservice.BatchOperationRefreshTasks{},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("unable to start batch refresh workflow tasks: %w", err)
+	}
+
+	// nolint:errcheck // assuming that write will succeed.
+	fmt.Fprintln(c.App.Writer, "Batch Refresh Workflow Tasks started successfully.")
```
yycptt: plz also print the jobID here so operator can use it to describe the started batch op.

tests/admin_batch_refresh_workflow_tasks_test.go
yycptt, line 1: nit: consider adding tests for 1. admin batch and normal batch don't share the same limit and 2. ListBatchOperations doesn't show admin batch op.

yycptt (approved):

tools/tdbg/commands.go
awln-temporal: gotcha, I just recall a discussion with Roey/Rodrigo about disallowing before, but maybe we just ignore that.

tests/admin_batch_refresh_workflow_tasks_test.go
awln-temporal: added tests to validate separate namespace divisions.

PR #8796, opened 2025-12-11
Add dynamic config to toggle standalone activity functionality

fretz12:
## What changed?
Add dynamic config to toggle standalone activity functionality. Refactored frontend dynamic config into activity config. Removed standalone activity dc prefix `chasm`

## Why?
We need the ability to toggle standalone activity functionality as we go to prod. We also agreed to remove the `chasm` prefix from any chasm originated dynamic config keys

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/common/dynamicconfig/setting_gen.go
+++ b/common/dynamicconfig/setting_gen.go
@@ -924,10 +924,8 @@ func (s NamespaceTypedSetting[T]) Validate(v any) error {
 	return err
 }
 
-func (s NamespaceTypedConstrainedDefaultSetting[T]) Key() Key { return s.key }
-func (s NamespaceTypedConstrainedDefaultSetting[T]) Precedence() Precedence {
-	return PrecedenceNamespace
-}
+func (s NamespaceTypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
+func (s NamespaceTypedConstrainedDefaultSetting[T]) Precedence() Precedence { return PrecedenceNamespace }
 func (s NamespaceTypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
@@ -1062,10 +1060,8 @@ func (s NamespaceIDTypedSetting[T]) Validate(v any) error {
 	return err
 }
 
-func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Key() Key { return s.key }
-func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Precedence() Precedence {
-	return PrecedenceNamespaceID
-}
+func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Key() Key               { return s.key }
+func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Precedence() Precedence { return PrecedenceNamespaceID }
 func (s NamespaceIDTypedConstrainedDefaultSetting[T]) Validate(v any) error {
 	_, err := s.convert(v)
 	return err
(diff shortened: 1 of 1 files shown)


chasm/lib/activity/frontend.go
```diff
@@ -71,6 +70,10 @@ func NewFrontendHandler(
 // before mutation to preserve the original for retries.
 // 3. Sends the request to the history activity service.
 func (h *frontendHandler) StartActivityExecution(ctx context.Context, req *workflowservice.StartActivityExecutionRequest) (*workflowservice.StartActivityExecutionResponse, error) {
+	if !h.config.Enabled(req.GetNamespace()) {
```
fretz12, line 81: Alternatively we can share this through an rpc interceptor, but not sure if overkill for this purpose

chasm/lib/activity/config.go
```diff
@@ -4,17 +4,24 @@ import (
...
 var (
+	Enabled = dynamicconfig.NewNamespaceBoolSetting(
+		"activity.enabled",
+		false,
+		`Toggles standalone activity functionality on the server.`,
+	)
+
 	LongPollTimeout = dynamicconfig.NewNamespaceDurationSetting(
-		"chasm.activity.longPollTimeout",
+		"activity.longPollTimeout",
```
fretz12, line 18: IIRC we agreed to remove the chasm prefix

chasm/lib/activity/frontend.go
bergundy: Yeah, no need for that IMHO.

chasm/lib/activity/frontend.go
```diff
@@ -71,6 +70,10 @@ func NewFrontendHandler(
...
+		return nil, serviceerror.NewUnavailable("Standalone activity is disabled.")
```
bergundy: We are very inconsistent on use of punctuation in error messages across the codebase. I tend to omit punctuation for single sentence messages. Not blocking. @dandavison, thoughts?

chasm/lib/activity/config.go
```diff
@@ -4,17 +4,24 @@ import (
 	"time"
 
 	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/retrypolicy"
 )
 
 var (
+	Enabled = dynamicconfig.NewNamespaceBoolSetting(
+		"activity.enabled",
```
bergundy: This will be more meaningful to operators and the tense matches the rest of our configs (enable vs. enabled).

```suggestion
		"activity.enableStandalone",
```

bergundy (commented):
You need to integrate with the frontend's `workflow_handler.go` too for all of the poll and complete APIs.

chasm/lib/activity/frontend.go
fretz12: I'm ok with no period at the end... but any sentence that needs proper punctuation within (i.e. commas), we should use.

chasm/lib/activity/config.go
fretz12: Changed

fretz12:
> You need to integrate with the frontend's `workflow_handler.go` too for all of the poll and complete APIs.

@bergundy good call, I've added checks there too. I've not added one for `PollActivityTaskQueue` as we won't be able to distinguish a standalone activity until we get a response back from the matching service. Currently the enabled DC is kept by chasm activity and we don't have any dependencies in matching yet. Hopefully this is not a big deal since we have the check in all other APIs and should be impossible for a user to start any SAA when the other APIs are guarded.

chasm/lib/activity/frontend.go
```diff
@@ -10,14 +10,15 @@ import (
...
 	"go.temporal.io/server/common"
-	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/searchattribute"
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
+const StandaloneActivityDisabledError = "Standalone activity is disabled"
```
bergundy, line 20: nit: I would just repeat the string instead of adding a layer of indirection. You're not getting much from this constant.

fretz12: I think we should just keep the message in one place to keep it consistent.

PR #8797, opened 2025-12-11
Fix visibility schema example to use temporal_visibility DB

mkthefreeman:
## What changed?
Corrected the temporal-sql-tool visibility schema example to use --db temporal_visibility instead of --db temporal.

## Why?
The previous example used the wrong database name, which could confuse users and lead to applying the visibility schema to the wrong DB.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


yycptt:
Thanks for the fix!

PR #8798, opened 2025-12-11
Add concurrency test for async Versioning operations

ShahabT:
## What changed?
Add concurrency test and refine some rough edges.

## Why?
testing is good!

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/tests/worker_deployment_test.go
+++ b/tests/worker_deployment_test.go
@@ -2122,6 +2122,8 @@ func (s *WorkerDeploymentSuite) TestSetRampingVersion_Concurrent_SameVersion_NoU
 // ex: "history_events: premature end of stream, expectedLastEventID=13 but no more events after eventID=11"
 // (error does not seems to be related to versioning)
 func (s *WorkerDeploymentSuite) TestConcurrentPollers_ManyTaskQueues_RapidRoutingUpdates_RevisionConsistency() {
+	s.T().Skip("Skipping until we can figure out why this test is flaky in sqlite")
+
 	// This test should work for InitialVersion, but it takes much longer (4m vs 1m15s vs 55s, in order, for 50 TQs).
 	// Also skipping for AsyncSetCurrentAndRampingVersion, to reduce flake chance.
 	s.skipBeforeVersion(workerdeployment.VersionDataRevisionNumber)


service/worker/workerdeployment/version_workflow.go
```diff
@@ -204,17 +204,21 @@ func (d *VersionWorkflowRunner) run(ctx workflow.Context) error {
...
 
-	// First ensure deployment workflow is running
-	if !d.VersionState.StartedDeploymentWorkflow {
-		activityCtx := workflow.WithActivityOptions(ctx, defaultActivityOptions)
-		err := workflow.ExecuteActivity(activityCtx, d.a.StartWorkerDeploymentWorkflow, &deploymentspb.StartWorkerDeploymentRequest{
-			DeploymentName: d.VersionState.Version.DeploymentName,
-			RequestId:      d.newUUID(ctx),
-		}).Get(ctx, nil)
-		if err != nil {
-			return err
```
Shivs11, line 215: I wonder: Could there ever be a case where a deployment workflow CAN's while the version workflow is processing an update?

In theory, this could happen. Imagine this scenario:

1. User sends in a setCurrent update.
2. Deployment workflow commences an async operation of the sync propagation to the task queues by calling the version workflow.
3. Version workflow, when it comes here, sends a signal.
4. The deployment workflow CAN's since according to itself, the state has changed.


I am not fully familiar about the signal "sending" semantics when a workflow is CAN'ing, but this could be important to consider so pasting this here.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -324,11 +340,18 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context, args *
...
+		// already deleted, returning success so it is idempotent
+		return nil
+	}
+
+	if !d.hasMinVersion(VersionDataRevisionNumber) {
+		// wait until deployment workflow started
+		err = workflow.Await(ctx, func() bool { return d.VersionState.StartedDeploymentWorkflow })
+		if err != nil {
+			d.logger.Error("Update canceled before worker deployment workflow started")
+			return serviceerror.NewDeadlineExceeded("Update canceled before worker deployment workflow started")
```
Shivs11, line 360: nit: Can improve error message by saying something like:

> Update canceled since the corresponding worker deployment workflow for this version did not start.

service/worker/workerdeployment/workflow.go
```diff
@@ -47,9 +47,9 @@ type (
 		stateChanged  bool
 		signalHandler *SignalHandler
 		forceCAN      bool
-		// workflowVersion is set at workflow start based on the dynamic config of the worker
-		// that completes the first task. It remains constant for the lifetime of the run and
-		// only updates when the workflow performs continue-as-new.
+		// Tracks the version of the deployment workflow when a particular run of a workflow starts base on the dynamic config of the
```
Shivs11: >  run of a workflow starts based on the dynamic config of the
worker that completes the first task of the workflow. `workflowVersion` remains the same until the workflow CaNs, after which it will get another chance to pick the latest manager version.

service/worker/workerdeployment/workflow.go
```diff
@@ -746,9 +776,10 @@ func (d *WorkflowRunner) setRamp(
 				)
 			}
 		}
-
-		// Erase summary drainage status immediately, so it is not draining/drained.
-		d.setDrainageStatus(newRampingVersion, enumspb.VERSION_DRAINAGE_STATUS_UNSPECIFIED, routingUpdateTime)
+		if !asyncMode {
+			// Erase summary drainage status immediately, so it is not draining/drained.
+			d.setDrainageStatus(newRampingVersion, enumspb.VERSION_DRAINAGE_STATUS_UNSPECIFIED, routingUpdateTime)
+		}
```
Shivs11, line 787: slightly confused;  I think we did have this functionality of changing the version's drainage status almost instantly if we realize it's being promoted. Did we delete it by mistake?

tests/worker_deployment_test.go
```diff
@@ -38,12 +40,12 @@ type (
...
-	})
-	t.Run("async", func(t *testing.T) {
-		suite.Run(t, &WorkerDeploymentSuite{workflowVersion: workerdeployment.AsyncSetCurrentAndRamping})
-	})
+	//t.Run("sync", func(t *testing.T) {
+	//	suite.Run(t, &WorkerDeploymentSuite{workflowVersion: workerdeployment.InitialVersion})
+	//})
+	//t.Run("async", func(t *testing.T) {
+	//	suite.Run(t, &WorkerDeploymentSuite{workflowVersion: workerdeployment.AsyncSetCurrentAndRamping})
+	//})
```
Shivs11: reminder: remove the comments here

service/worker/workerdeployment/version_workflow.go
ShahabT: yes it's possible, but CaN is atomic which means there is no gap between closing the old execution and creating the new one. both pieces happen at the same time.

so the signal will go to the old or the new execution but it can't see "not found".

Even if that was not the case, this check does not protect against the mentioned scenario because the CaN can happen just after this check and before the signal.

service/worker/workerdeployment/workflow.go
ShahabT: this part was undone unintentionally, will revert to the old comment which was your suggestion.

service/worker/workerdeployment/workflow.go
ShahabT: No need for this because syncVersion takes care of updating summary already.

config/dynamicconfig/development-sql.yaml
```diff
@@ -31,14 +31,14 @@
...
-#matching.wv.VersionDrainageStatusRefreshInterval:
-# - value: 5s
+matching.deploymentWorkflowVersion:
+ - value: 2 # put the integer value of the workflow version you want to target (enum DeploymentWorkflowVersion)
+matching.PollerHistoryTTL:
+ - value: 1s
+matching.wv.VersionDrainageStatusVisibilityGracePeriod:
+ - value: 5s
+matching.wv.VersionDrainageStatusRefreshInterval:
+ - value: 5s
```
carlydf: do you want to leave these in by default?

carlydf: I guess maybe it's not worth re-running CI to remove -- I can revert it in a subsequent PR like my remove-verbose-logging PR...

service/worker/workerdeployment/activities.go
```diff
@@ -141,20 +147,42 @@ func (a *Activities) IsVersionMissingTaskQueues(ctx context.Context, args *deplo
 }
 
 func (a *Activities) DeleteWorkerDeploymentVersion(ctx context.Context, args *deploymentspb.DeleteVersionActivityArgs) error {
+	fmt.Printf("act Deleting version %s from deployment %s\n", args.Version, args.DeploymentName)
```
carlydf: ah this probably should not stay

service/worker/workerdeployment/client.go
```diff
@@ -1335,138 +1207,58 @@ func (d *ClientImpl) updateWithStartWorkerDeploymentVersion(
...
 				)
 			} else {
-				d.logger.Error("deployment client error",
-					tag.Error(*retErr),
-					tag.Operation(operation),
-					tag.NewDurationTag("elapsed", elapsed),
-					tag.NewAnyTag("args", args),
-				)
+				if isRetryableUpdateError(*retErr) || isRetryableQueryError(*retErr) {
+					d.logger.Debug("deployment client throttling due to retriable error",
```
carlydf: Nit: I just googled this, turns out "retriable" and "retryable" are both valid spellings, but apparently in tech contexts "retryable" is more common (see for example `isRetryableQueryError` and `isRetryableUpdateError` names).

Anyways, doesn't really matter except "retryable" might be what I search for in the logs first. Now I know to also look for "retriable" though :)

service/worker/workerdeployment/replaytester/testdata/v2/run_1765589504/expected_counts.txt
```diff
@@ -1,5 +1,5 @@
 # Expected workflow counts for replay testing
-# Generated by generate_history.sh on Tue Dec  9 15:28:20 PST 2025
+# Generated by generate_history.sh on Fri Dec 12 17:31:46 PST 2025
```
carlydf, line 2: is the idea that, because this was never released anywhere, breaking the wf changes that you made a few days ago is ok?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -324,11 +342,19 @@ func (d *VersionWorkflowRunner) handleDeleteVersion(ctx workflow.Context, args *
...
+		return nil
+	}
+
+	if !d.hasMinVersion(VersionDataRevisionNumber) {
+		// wait until deployment workflow started
+		//nolint:staticcheck // SA1019
+		err = workflow.Await(ctx, func() bool { return d.VersionState.StartedDeploymentWorkflow })
+		if err != nil {
+			d.logger.Error("Update canceled before worker deployment workflow started")
+			return serviceerror.NewDeadlineExceeded("Update canceled before worker deployment workflow started")
```
carlydf, line 360: Ideally we should include the worker deployment name in this error, since we have access to it

service/worker/workerdeployment/version_workflow.go
```diff
@@ -458,6 +484,8 @@ func (d *VersionWorkflowRunner) doesVersionHaveActivePollers(ctx workflow.Contex
 }
 
 func (d *VersionWorkflowRunner) validateRegisterWorker(args *deploymentspb.RegisterWorkerInVersionArgs) error {
+	// Should not ensure not deleted, instead the version would revive if deleted.
+
 	if _, ok := d.VersionState.TaskQueueFamilies[args.TaskQueueName].GetTaskQueues()[int32(args.TaskQueueType)]; ok {
 		return temporal.NewApplicationError("task queue already exists in deployment version", errNoChangeType)
```
carlydf, line 495: should we include task queue name and version name here?

service/worker/workerdeployment/version_workflow.go
```diff
@@ -613,11 +658,13 @@ func (d *VersionWorkflowRunner) handleSyncState(ctx workflow.Context, args *depl
...
-	err = workflow.Await(ctx, func() bool { return d.VersionState.StartedDeploymentWorkflow })
-	if err != nil {
-		d.logger.Error("Update canceled before worker deployment workflow started")
-		return nil, serviceerror.NewDeadlineExceeded("Update canceled before worker deployment workflow started")
+	if !d.hasMinVersion(VersionDataRevisionNumber) {
+		// wait until deployment workflow started
+		err = workflow.Await(ctx, func() bool { return d.VersionState.StartedDeploymentWorkflow })
+		if err != nil {
+			d.logger.Error("Update canceled before worker deployment workflow started")
+			return nil, serviceerror.NewDeadlineExceeded("Update canceled before worker deployment workflow started")
```
carlydf, line 671: same here: recommend to include deployment name.

I wouldnt suggest it if there weren't other things to push like removing the print statement, but if another commit is needed anyway.. consider this

service/worker/workerdeployment/util.go
```diff
@@ -63,20 +76,22 @@ const (
...
+	ErrVersionHasPollers         = "version '%s' cannot be deleted since it has active pollers"
 	ErrVersionIsCurrentOrRamping = "version '%s' cannot be deleted since it is current or ramping"
 
 	ErrRampingVersionDoesNotHaveAllTaskQueues = "proposed ramping version '%s' is missing active task queues from the current version; these would become unversioned if it is set as the ramping version"
 	ErrCurrentVersionDoesNotHaveAllTaskQueues = "proposed current version '%s' is missing active task queues from the current version; these would become unversioned if it is set as the current version"
 	ErrManagerIdentityMismatch                = "ManagerIdentity '%s' is set and does not match user identity '%s'; to proceed, set your own identity as the ManagerIdentity, remove the ManagerIdentity, or wait for the other client to do so"
 	ErrWorkerDeploymentNotFound               = "no Worker Deployment found with name '%s'; does your Worker Deployment have pollers?"
 	ErrWorkerDeploymentVersionNotFound        = "build ID '%s' not found in Worker Deployment '%s'"
+	ErrTooManyRequests                        = "too many requests issued to the same Worker Deployment. Please try again later"
```
carlydf: if this is user facing, consider including Worker Dpeloyment name

service/worker/workerdeployment/client.go
Shivs11: @carlydf TIL, ty!!

> "retriable" and "retryable" are both valid spellings

service/worker/workerdeployment/replaytester/testdata/v2/run_1765589504/expected_counts.txt
ShahabT: yup I'm making incompatible changes to v2 and it's fine.

service/worker/workerdeployment/client.go
ShahabT: good to know. fixed

config/dynamicconfig/development-sql.yaml
ShahabT: commented them out

service/worker/workerdeployment/version_workflow.go
ShahabT: this path will not execute in v2

service/worker/workerdeployment/version_workflow.go
ShahabT: no, this is only used internally

PR #8799, opened 2025-12-11
Update Nexus error model to use Temporal failures

pdoerner:
## What changed?
Many changes to support the Temporal SDK sending Temporal `Failure`s instead of errors.

Depends on https://github.com/temporalio/api/pull/682 and https://github.com/nexus-rpc/sdk-go/pull/69

## Why?
Consistency with other APIs and more rich information.

## How did you test it?
- [x] built
- [x] added new unit test(s)
- [x] https://github.com/bergundy/nexus-error-compat-tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/lib/pq v1.10.9
 	github.com/maruel/panicparse/v2 v2.4.0
 	github.com/mitchellh/mapstructure v1.5.0
-	github.com/nexus-rpc/sdk-go v0.5.2-0.20260210000428-3d8ad6dc9742
+	github.com/nexus-rpc/sdk-go v0.5.2-0.20260211051645-26b0b4c584e5
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/olivere/elastic/v7 v7.0.32
 	github.com/pkg/errors v0.9.1
--- a/go.sum
+++ b/go.sum
@@ -236,8 +236,8 @@ github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 h1:C3w9PqII01/Oq
 github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822/go.mod h1:+n7T8mK8HuQTcFwEeznm/DIxMOiR9yIdICNftLE1DvQ=
 github.com/ncruces/go-strftime v1.0.0 h1:HMFp8mLCTPp341M/ZnA4qaf7ZlsbTc+miZjCLOFAw7w=
 github.com/ncruces/go-strftime v1.0.0/go.mod h1:Fwc5htZGVVkseilnfgOVb9mKy6w1naJmn9CehxcKcls=
-github.com/nexus-rpc/sdk-go v0.5.2-0.20260210000428-3d8ad6dc9742 h1:nE/NqmspHqHkLh8rsSsQ/dLXOrbx4N7dU7GoGsO0DdI=
-github.com/nexus-rpc/sdk-go v0.5.2-0.20260210000428-3d8ad6dc9742/go.mod h1:FHdPfVQwRuJFZFTF0Y2GOAxCrbIBNrcPna9slkGKPYk=
+github.com/nexus-rpc/sdk-go v0.5.2-0.20260211051645-26b0b4c584e5 h1:Van9KGGs8lcDgxzSNFbDhEMNeJ80TbBxwZ45f9iBk9U=
+github.com/nexus-rpc/sdk-go v0.5.2-0.20260211051645-26b0b4c584e5/go.mod h1:FHdPfVQwRuJFZFTF0Y2GOAxCrbIBNrcPna9slkGKPYk=
 github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e/go.mod h1:zD1mROLANZcx1PVRCS0qkT7pwLkGfwJo4zjcN/Tysno=
 github.com/olekukonko/tablewriter v0.0.5 h1:P2Ga83D34wi1o9J6Wh1mRuqd4mF/x/lgBS7N7AbDhec=
 github.com/olekukonko/tablewriter v0.0.5/go.mod h1:hPp6KlRPjbx+hW8ykQs1w3UBbZlj6HuIJcUGPhkA7kY=


common/nexus/failure.go
```diff
@@ -74,56 +76,140 @@ func NexusFailureToProtoFailure(failure nexus.Failure) *nexuspb.Failure {
...
-	// Ensure this always gets written.
-	apiFailure.Message = failure.Message
+
+	payloads, err := nexusFailureMetadataToPayloads(f)
+	if err != nil {
+		return nil, err
+	}
+	apiFailure.FailureInfo = &failurepb.Failure_ApplicationFailureInfo{
+		ApplicationFailureInfo: &failurepb.ApplicationFailureInfo{
+			// Make up a type here, it's not part of the Nexus Failure spec.
```
bergundy: We have `NexusSDKFailureErrorFailureInfo` in the API PR.

common/nexus/failure.go
```diff
@@ -74,56 +76,140 @@ func NexusFailureToProtoFailure(failure nexus.Failure) *nexuspb.Failure {
...
+			Type:    "NexusFailure",
+			Details: payloads,
+		},
+	}
 	return apiFailure, nil
 }
 
-func OperationErrorToTemporalFailure(opErr *nexus.OperationError) (*failurepb.Failure, error) {
+func OperationErrorToTemporalFailure(opErr *nexus.OperationError, retryable bool) (*failurepb.Failure, error) {
```
bergundy: You shouldn't need the retryable flag here, operation errors are non-retryable by definition and the resulting failure object should be a NexusOperationFailureInfo. This function doesn't seem necessary anymore, you should already have the original nexus failure on the operation error so all you need to do is convert to a temporal failure.

common/nexus/nexusrpc/client.go
```diff
@@ -355,90 +363,63 @@ func (c *HTTPClient) failureFromResponse(response *http.Response, body []byte) (
...
-	return failureErr
+func (c *HTTPClient) bestEffortHandlerErrorFromResponse(response *http.Response, body []byte) error {
+	// TODO: support old servers
+	failure, err := c.failureFromResponse(response, body)
+	if err != nil {
+		return c.defaultErrorFromResponse(response, body, nil)
+	}
+	convErr, err := c.options.FailureConverter.FailureToError(failure)
+	if err != nil {
+		return fmt.Errorf("failed to convert Failure to error: %w", err)
```
bergundy: Let's make this an unexpected error?

common/nexus/nexusrpc/client.go
```diff
@@ -355,90 +363,63 @@ func (c *HTTPClient) failureFromResponse(response *http.Response, body []byte) (
...
+	}
+	if _, ok := convErr.(*nexus.HandlerError); !ok {
+		convErr = c.defaultErrorFromResponse(response, body, convErr)
+	}
+	return convErr
 }
 
-func (c *HTTPClient) bestEffortHandlerErrorFromResponse(response *http.Response, body []byte) error {
+func httpStatusCodeToHandlerErrorType(response *http.Response) (nexus.HandlerErrorType, error) {
```
bergundy, line 429: Can you please double check that we cover all of the error types in the nexus SDK?

common/nexus/nexusrpc/client.go
```diff
@@ -355,90 +363,63 @@ func (c *HTTPClient) failureFromResponse(response *http.Response, body []byte) (
...
+func (c *HTTPClient) defaultErrorFromResponse(response *http.Response, body []byte, cause error) error {
+	errorType, err := httpStatusCodeToHandlerErrorType(response)
 	if err != nil {
-		failure.Message = defaultMessage
+		// TODO: use the provided cause, it's already a deserialized failure.
+		return newUnexpectedResponseError(err.Error(), response, body)
+	}
+	return &nexus.HandlerError{
+		Type:    errorType,
+		Message: response.Status,
```
bergundy: This includes the HTTP status code, we need to trim that.

common/nexus/nexusrpc/client.go
```diff
@@ -355,90 +363,63 @@ func (c *HTTPClient) failureFromResponse(response *http.Response, body []byte) (
...
 	}
-	return failure
 }
 
-func (c *HTTPClient) failureErrorFromResponseOrDefault(response *http.Response, body []byte, defaultMessage string) error {
-	failure := c.failureFromResponseOrDefault(response, body, defaultMessage)
-	failureErr := c.options.FailureConverter.FailureToError(failure)
-	return failureErr
+func (c *HTTPClient) bestEffortHandlerErrorFromResponse(response *http.Response, body []byte) error {
+	// TODO: support old servers
```
bergundy: I think this TODO is addressed?

common/nexus/nexusrpc/completion.go
```diff
@@ -165,6 +165,9 @@ type OperationCompletionUnsuccessful struct {
 type OperationCompletionUnsuccessfulOptions struct {
 	// A [FailureConverter] to convert a [Failure] instance to and from an [error]. Defaults to
 	// [DefaultFailureConverter].
+	//
+	// NOTE: To call server versions <= 0.4.0, use a FailureConverter that unwraps the error cause if message is not
```
bergundy: I think this is the default behavior for the failure converter today? Can you verify that and add a comment?

common/nexus/nexusrpc/server.go
```diff
@@ -98,7 +98,12 @@ func (h *baseHTTPHandler) writeFailure(writer http.ResponseWriter, err error) {
 
 	if errors.As(err, &opError) {
 		operationState = opError.State
-		failure = h.failureConverter.ErrorToFailure(opError.Cause)
+		var convErr error
+		failure, convErr = h.failureConverter.ErrorToFailure(opError)
+		if convErr != nil {
+			h.logger.Error("failed to convert operation error to failure", "error", convErr)
+			writer.WriteHeader(http.StatusInternalServerError)
```
bergundy, line 109: Missing return.

common/nexus/nexusrpc/server.go
```diff
@@ -108,7 +113,12 @@ func (h *baseHTTPHandler) writeFailure(writer http.ResponseWriter, err error) {
...
 		writer.Header().Set(headerOperationState, string(operationState))
 	} else if errors.As(err, &handlerError) {
-		failure = h.failureConverter.ErrorToFailure(handlerError.Cause)
+		var convErr error
+		failure, convErr = h.failureConverter.ErrorToFailure(handlerError)
+		if convErr != nil {
+			h.logger.Error("failed to convert handler error to failure", "error", convErr)
+			writer.WriteHeader(http.StatusInternalServerError)
+		}
 		switch handlerError.Type {
```
bergundy, line 137: Please verify that this covers all of the handler error types in the latest Nexus SDK.

common/nexus/nexusrpc/server.go
```diff
@@ -108,7 +113,12 @@ func (h *baseHTTPHandler) writeFailure(writer http.ResponseWriter, err error) {
 		}
 		writer.Header().Set(headerOperationState, string(operationState))
 	} else if errors.As(err, &handlerError) {
-		failure = h.failureConverter.ErrorToFailure(handlerError.Cause)
+		var convErr error
+		failure, convErr = h.failureConverter.ErrorToFailure(handlerError)
+		if convErr != nil {
+			h.logger.Error("failed to convert handler error to failure", "error", convErr)
+			writer.WriteHeader(http.StatusInternalServerError)
```
bergundy, line 130: Missing return.

components/nexusoperations/executors.go
```diff
@@ -862,44 +862,18 @@ func isDestinationDown(err error) bool {
...
-		}
-		failure := &failurepb.Failure{
-			Message: handlerErr.Error(),
-			FailureInfo: &failurepb.Failure_NexusHandlerFailureInfo{
-				NexusHandlerFailureInfo: &failurepb.NexusHandlerFailureInfo{
-					Type:          string(handlerErr.Type),
-					RetryBehavior: retryBehavior,
-				},
-			},
+		nf, err := nexus.DefaultFailureConverter().ErrorToFailure(handlerErr)
```
bergundy: You should use the original failure here and then convert it to a temporal failure.

components/nexusoperations/executors_test.go
```diff
@@ -287,7 +287,7 @@ func TestProcessInvocationTask(t *testing.T) {
 			checkOutcome: func(t *testing.T, op nexusoperations.Operation, events []*historypb.HistoryEvent) {
 				require.Equal(t, enumsspb.NEXUS_OPERATION_STATE_BACKING_OFF, op.State())
 				require.NotNil(t, op.LastAttemptFailure.GetNexusHandlerFailureInfo())
-				require.Equal(t, "handler error (INTERNAL): internal server error", op.LastAttemptFailure.Message)
+				require.Equal(t, "internal server error", op.LastAttemptFailure.Message)
```
bergundy, line 318: Please also check the handler error type in all of these tests.

components/nexusoperations/frontend/handler.go
```diff
@@ -198,16 +198,23 @@ func (h *completionHandler) CompleteOperation(ctx context.Context, r *nexusrpc.C
 	}
 	switch r.State { // nolint:exhaustive
 	case nexus.OperationStateFailed, nexus.OperationStateCanceled:
-		failureErr, ok := r.Error.(*nexus.FailureError)
-		if !ok {
+		var failureErr *nexus.FailureError
+		var operationErr *nexus.OperationError
+		switch {
+		case errors.As(r.Error, &failureErr):
```
bergundy: You should always get an operation error here.

service/frontend/nexus_handler.go
```diff
@@ -444,19 +447,32 @@ func (h *nexusHandler) StartOperation(
 	}
 	// Convert to standard Nexus SDK response.
 	switch t := response.GetOutcome().(type) {
+	case *matchingservice.DispatchNexusTaskResponse_Failure:
+		oc.metricsHandler = oc.metricsHandler.WithTags(metrics.OutcomeTag("handler_error:" + t.Failure.GetNexusHandlerFailureInfo().GetType()))
+		nf, err := commonnexus.APIFailureToNexusFailure(t.Failure)
```
bergundy: You want to create a properly structured nexus failure here with metadata `type` set to `nexus.HandlerError` or `nexus.OperationError` and make sure the cause chain is populated correctly.

common/nexus/failure.go
pdoerner: `NexusSDKFailureErrorFailureInfo` doesn't have a `Details` field. We only get to this point if we get an unexpected error type, so I thought it was better to capture the full information.

common/nexus/failure.go
pdoerner: I also thought it would be unnecessary but I had some trouble removing it as this is the only place where we have the specific handling for `CanceledFailureInfo`.

common/nexus/nexusrpc/client.go
bergundy: Done.

common/nexus/nexusrpc/server.go
bergundy: Done.

bergundy:
Closed in favor of https://github.com/temporalio/temporal/pull/9290

PR #8800, opened 2025-12-11
Add external payload stats to DescribeWorkflowExecution

simvlad:
## What changed?
Added external payload stats (size and count) to DescribeWorkflowExecution. Also change and add functional tests. **This PR depends on Server PR https://github.com/temporalio/temporal/pull/8775 and API PR https://github.com/temporalio/api/pull/689.**

## Why?
We'd like to surface the external payload stats when DescribeWorkflowExecution is called to see how many and what size of external payloads the workflow is referencing. Numbers are coming from the workflow mutable state. This is similar to how the HistorySize is kept.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/service/history/api/describeworkflow/api.go
+++ b/service/history/api/describeworkflow/api.go
@@ -89,6 +89,7 @@ func Invoke(
 	defer func() { workflowLease.GetReleaseFn()(retError) }()
 
 	mutableState := workflowLease.GetMutableState()
+	namespaceName := mutableState.GetNamespaceEntry().Name().String()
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -170,3 +171,8 @@ require (
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.11.0 // indirect
 )
+
```
simvlad: The API change is not released yet. Will remove this before merging this PR.

PR #8803, opened 2025-12-11
Add BusinessID search attribute option

awln-temporal:
## What changed?
Add BusinessID search attribute option. CHASM archetype developers that require Visibility will need to call `WithBusinessIDAlias` as a required RegistrableComponent option.

Adds validation if BusinessIDAlias option is not set, but contains Visibility as a child component, registration to the CHASM Registry will fail.

## Why?
CHASM archetypes define separate aliases for BusinessID, eg. `ActivityID`, `ScheduleID`, etc. End users of these archetypes will query Visibility using a filter such as `ActivityID = my-activity-id`, and should be able to map to the underlying shared `system` visibility storage column.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/activity/library.go
+++ b/chasm/lib/activity/library.go
@@ -26,6 +26,7 @@ func (l *componentOnlyLibrary) Components() []*chasm.RegistrableComponent {
 				StatusSearchAttribute,
 				TaskQueueSearchAttribute,
 			),
+			chasm.WithBusinessIDAlias("ActivityId"),
 		),
 	}
 }
--- a/common/persistence/visibility/store/query/resolve.go
+++ b/common/persistence/visibility/store/query/resolve.go
@@ -38,13 +38,6 @@ func ResolveSearchAttributeAlias(
 			return sadefs.WorkflowID, saType, nil
 		}
 
-		// Handle ActivityId → WorkflowID transformation for standalone activities.
-		// TODO: Remove this hardcoded transformation.
-		if name == sadefs.ActivityID {
-			saType, _ := saTypeMap.GetType(sadefs.WorkflowID)
-			return sadefs.WorkflowID, saType, nil
-		}
-
 		fieldName, fieldType = tryChasmMapper(name, chasmMapper)
 		if fieldName != "" {
 			return fieldName, fieldType, nil
--- a/tests/chasm_test.go
+++ b/tests/chasm_test.go
@@ -228,7 +228,7 @@ func (s *ChasmTestSuite) TestListExecutions() {
 	archetypeID, ok := s.FunctionalTestBase.GetTestCluster().Host().GetCHASMRegistry().ComponentIDFor(&tests.PayloadStore{})
 	s.True(ok)
 
-	visQuery := fmt.Sprintf("TemporalNamespaceDivision = '%d' AND WorkflowId = '%s'", archetypeID, storeID)
+	visQuery := fmt.Sprintf("TemporalNamespaceDivision = '%d' AND PayloadStoreId = '%s'", archetypeID, storeID)
 
 	var visRecord *chasm.ExecutionInfo[*testspb.TestPayloadStore]
 	s.Eventually(


chasm/registrable_component.go
```diff
@@ -67,24 +68,68 @@ func WithShardingFn(
...
+	alias string,
+) RegistrableComponentOption {
+	return func(rc *RegistrableComponent) {
+		if rc.searchAttributesMapper == nil {
+			rc.searchAttributesMapper = &VisibilitySearchAttributesMapper{
+				aliasToField:       make(map[string]string),
+				fieldToAlias:       make(map[string]string),
+				saTypeMap:          make(map[string]enumspb.IndexedValueType),
+				systemAliasToField: make(map[string]string),
+			}
```
rodrigozhou, line 83: Create a function newVisibilitySearchAttributesMapper.

chasm/registrable_component.go
```diff
@@ -67,24 +68,68 @@ func WithShardingFn(
...
+		}
+		if rc.searchAttributesMapper.systemAliasToField == nil {
+			rc.searchAttributesMapper.systemAliasToField = make(map[string]string)
+		}
```
rodrigozhou, line 87: This doesn't seem necessary?

chasm/registrable_component.go
```diff
@@ -67,24 +68,68 @@ func WithShardingFn(
...
+		if _, ok := rc.searchAttributesMapper.aliasToField[alias]; ok {
+			//nolint:forbidigo
+			panic(fmt.Sprintf("registrable component validation error: business ID alias %q is already defined as a search attribute", alias))
+		}
+		if _, ok := rc.searchAttributesMapper.systemAliasToField[alias]; ok {
+			//nolint:forbidigo
+			panic(fmt.Sprintf("registrable component validation error: business ID alias %q is already defined as a system search attribute", alias))
```
rodrigozhou, line 94:
```suggestion
			panic(fmt.Sprintf("registrable component validation error: business ID alias %q is already defined as a search attribute", alias))
```

chasm/registrable_component.go
```diff
@@ -67,24 +68,68 @@ func WithShardingFn(
...
+		}
+		rc.searchAttributesMapper.systemAliasToField[alias] = sadefs.WorkflowID
+		rc.searchAttributesMapper.fieldToAlias[sadefs.WorkflowID] = alias
+		rc.searchAttributesMapper.saTypeMap[sadefs.WorkflowID] = enumspb.INDEXED_VALUE_TYPE_KEYWORD
```
rodrigozhou, line 98: Is this necessary? System search attributes are already mapped to types in the `sadefs` package. This should not be necessary.

chasm/registrable_component.go
```diff
@@ -67,24 +68,68 @@ func WithShardingFn(
...
+				systemAliasToField: make(map[string]string),
+			}
 		}
 
 		for _, sa := range searchAttributes {
 			alias := sa.definition().alias
 			field := sa.definition().field
 			valueType := sa.definition().valueType
 
+			if sadefs.IsSystem(alias) || sadefs.IsReserved(alias) {
```
rodrigozhou, line 124: I think you could call `IsMappable` instead. You also need to check this with the business ID alias above.

chasm/registrable_component.go
```diff
@@ -67,24 +68,68 @@ func WithShardingFn(
...
+				//nolint:forbidigo
+				panic(fmt.Sprintf("registrable component validation error: CHASM search attribute alias %q is a system or reserved search attribute", alias))
+			}
+
+			if _, ok := rc.searchAttributesMapper.systemAliasToField[alias]; ok {
+				//nolint:forbidigo
+				panic(fmt.Sprintf("registrable component validation error: CHASM search attribute alias %q is already defined as a system search attribute alias", alias))
```
rodrigozhou, line 131: The error messages below says "search attribute alias", not "CHASM search attribute alias". Be consistent.

chasm/search_attribute.go
```diff
@@ -70,6 +70,10 @@ var (
 	SearchAttributeTemporalWorkerDeploymentVersion    = newSearchAttributeKeywordByField(sadefs.TemporalWorkerDeploymentVersion)
 	SearchAttributeTemporalWorkflowVersioningBehavior = newSearchAttributeKeywordByField(sadefs.TemporalWorkflowVersioningBehavior)
 	SearchAttributeTemporalWorkerDeployment           = newSearchAttributeKeywordByField(sadefs.TemporalWorkerDeployment)
+
+	// SearchAttributeBusinessIDField is a search attribute field mapping to underlying system search attribute WorkflowID.
+	// Used only with WithBusinessIDAlias option.
+	searchAttributeBusinessIDField = newSearchAttributeFieldKeywordByField(sadefs.WorkflowID)
```
rodrigozhou: This is not used.

chasm/registrable_component.go
awln-temporal: Is this needed if there's no args in the constructor?

chasm/registrable_component.go
awln-temporal: removed

chasm/registrable_component.go
awln-temporal: We need to map from BusinessID alias to system search attribute WorkflowID when resolving search attribute aliases, using the CHASM mapper. In `resolveSearchAttributeAlias`,

```
// resolveChasmSA only returns true if `alias` is a CHASM search attribute.
	resolveChasmSA := func(alias string) bool {
		if c.chasmMapper == nil {
			return false
		}
		fn, err := c.chasmMapper.Field(alias)
		if err != nil {
			return false
		}
		ft, err := c.chasmMapper.ValueType(fn)
		if err != nil {
			return false
		}
		fieldName, fieldType = fn, ft
		return true
	}
```

we return if we find a matching chasm search attribute alias. It's an extra step to then check the regular search attribute type map than checking the chasm mapper for the value type.

chasm/registrable_component.go
awln-temporal: will also add a check for businessID alias above

chasm/visibility.go
```diff
@@ -46,6 +48,17 @@ type VisibilitySearchAttributesMapper struct {
 	aliasToField map[string]string
 	fieldToAlias map[string]string
 	saTypeMap    map[string]enumspb.IndexedValueType
+
+	systemAliasToField map[string]string
```
rodrigozhou, line 54: Add a comment explaining this field

chasm/visibility.go
```diff
@@ -65,11 +78,37 @@ func (v *VisibilitySearchAttributesMapper) Field(alias string) (string, error) {
...
+		withoutPrefix := alias[len(sadefs.ReservedPrefix):]
+		if field, ok := v.systemAliasToField[withoutPrefix]; ok {
+			return field, true
+		}
+	}
+	// Try with the `Temporal` prefix.
+	withPrefix := sadefs.ReservedPrefix + alias
+	if field, ok := v.systemAliasToField[withPrefix]; ok {
+		return field, true
 	}
```
rodrigozhou: This should be inside the `else` block

chasm/registrable_component.go
rodrigozhou: No needed, but it makes easier since you build this object in multiple places.

chasm/search_attribute.go
```diff
@@ -172,6 +172,12 @@ func newSearchAttributeFieldKeyword(index int) SearchAttributeFieldKeyword {
 	}
 }
 
+func newSearchAttributeFieldKeywordByField(field string) SearchAttributeFieldKeyword {
+	return SearchAttributeFieldKeyword{
+		field: field,
+	}
+}
+
```
rodrigozhou: Remove, unused

rodrigozhou (approved):

chasm/visibility.go
awln-temporal: moved

PR #8804, opened 2025-12-11
[CHASM] Split and wire up engine and visibilityMgr request interceptors

lina-temporal:
## What changed?
- The `ChasmRequestInterceptor` has been split in two! `ChasmRequestEngineInterceptor` does everything except inject `chasm.VisibilityManager`, which is done by `ChasmRequestVisibilityInterceptor`.
- Note that the test doesn't include a matching `hasVisibilityCtx` assert to `hasEngineCtx` because I think making that getter public would make the public API earnestly more confusing for consumers

## Why?
- Frontend needs visibilityMgr injected so that it can handle FE->Vis queries without hopping through history. The previous interceptor assumed `chasm.Engine` was always available, which it isn't, for history. I'd also workshopped making engine an optional fx parameter, but eh. 

## How did you test it?
- Validated with scheduler functional tests, which use `chasm.ListExecutions` directly in frontend's `WorkflowHandler`.

### DIFF
--- a/chasm/interceptor_test.go
+++ b/chasm/interceptor_test.go
@@ -43,11 +43,11 @@ func (l *ServiceLibrary) RegisterServices(server *grpc.Server) {
 	testspb.RegisterTestServiceServer(server, ServiceHandler{})
 }
 
-func TestChasmRequestEngineInterceptor_ShouldRespond(t *testing.T) {
+func TestChasmEngineInterceptor_ShouldRespond(t *testing.T) {
 	ctrl := gomock.NewController(t)
 
 	mockEngine := chasm.NewMockEngine(ctrl)
-	engineInterceptor := chasm.ChasmRequestEngineInterceptorProvider(
+	engineInterceptor := chasm.ChasmEngineInterceptorProvider(
 		mockEngine,
 		log.NewNoopLogger(),
 		metrics.NoopMetricsHandler,
@@ -60,11 +60,11 @@ func TestChasmRequestEngineInterceptor_ShouldRespond(t *testing.T) {
 	require.True(t, response.HasEngineCtx)
 }
 
-func TestChasmRequestVisibilityInterceptor_ShouldRespond(t *testing.T) {
+func TestChasmVisibilityInterceptor_ShouldRespond(t *testing.T) {
 	ctrl := gomock.NewController(t)
 
 	mockVisibilityManager := chasm.NewMockVisibilityManager(ctrl)
-	visibilityInterceptor := chasm.ChasmRequestVisibilityInterceptorProvider(mockVisibilityManager)
+	visibilityInterceptor := chasm.ChasmVisibilityInterceptorProvider(mockVisibilityManager)
 
 	server, address := startTestServer(t, grpc.UnaryInterceptor(visibilityInterceptor.Intercept))
 	defer server.Stop()
--- a/chasm/interceptors.go
+++ b/chasm/interceptors.go
@@ -11,19 +11,17 @@ import (
 
 const chasmRequestPrefix = "/temporal.server.chasm"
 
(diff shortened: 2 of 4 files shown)


chasm/interceptor_test.go
```diff
@@ -43,20 +43,35 @@ func (l *ServiceLibrary) RegisterServices(server *grpc.Server) {
...
+func TestChasmRequestVisibilityInterceptor_ShouldRespond(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	mockVisibilityManager := chasm.NewMockVisibilityManager(ctrl)
+	visibilityInterceptor := chasm.ChasmRequestVisibilityInterceptorProvider(mockVisibilityManager)
+
+	server, address := startTestServer(t, grpc.UnaryInterceptor(visibilityInterceptor.Intercept))
+	defer server.Stop()
+	testRoundTrip(t, address)
+}
```
awln-temporal, line 72: any verifications we want that the context contains the chasm vis manager like we do for engine context?

awln-temporal (approved):

chasm/interceptors.go
```diff
@@ -33,21 +34,41 @@ func (i *ChasmRequestInterceptor) Intercept(
...
 	}
 }
+
+// ChasmRequestVisibilityInterceptor intercepts RPC requests and adds the CHASM
+// VisibilityManager to their context.
+type ChasmRequestVisibilityInterceptor struct {
+	visibilityMgr VisibilityManager
+}
+
+var _ grpc.UnaryServerInterceptor = (*ChasmRequestVisibilityInterceptor)(nil).Intercept
```
pdoerner: This is just for the type assertion? Don't know that it's really necessary

lina-temporal: Yeah, just for the type assertion.

chasm/interceptor_test.go
lina-temporal: Yeah, I hadn't done any here because we don't have any public interface way to get visibilityMgr context, as we do with the engine. Mocks wouldn't really work either, since I'd still have to make a call to the public `ListExecutions` which'll try to get the visibility context first (and then immediately nil-panic). I suppose I could panic-recover in the test handler but that seemed a little ugly to do in a test.

chasm/interceptors.go
```diff
@@ -11,18 +11,19 @@ import (
...
 const chasmRequestPrefix = "/temporal.server.chasm"
 
-// ChasmRequestInterceptor Interceptor that intercepts RPC requests, detects Chasm-specific calls and does additional
-// boilerplate processing before handing off.
-type ChasmRequestInterceptor struct {
+// ChasmRequestEngineInterceptor Interceptor that intercepts RPC requests,
+// detects CHASM-specific calls and does additional boilerplate processing before
+// handing off. Visibility is injected separately with
+// ChasmRequestVisibilityInterceptor.
+type ChasmRequestEngineInterceptor struct {
```
pdoerner: nit: I don't think `Request` is necessary in the names of either interceptor

lina-temporal: It does seem a bit wordy doesn't it :x I'll fix.

chasm/interceptors.go
pdoerner: Realistically it's never not going to implement `Intercept` and I think `HistoryAdditionalInterceptorsProvider` would throw a compile error if it didn't, so it's a bit unnecessary IMHO, but doesn't really hurt.

lina-temporal: You've convinced me, removed 'em here.

chasm/interceptor_test.go
awln-temporal: yup didn't read the description

PR #8809, opened 2025-12-11
Fix handle first replication task logic for state based replication

xwduan:
## What changed?
Fix handle first replication task logic for state based replication

## Why?
To prevent a potential lock scenario: run2 is run1's successor.
run1 is current run at passive, while run1's first replication task triggered the handleFirstReplicationTask code path and has  run2 as newRun. When create run2, it will try to get current lock again which caused the deadlock and task will go to DLQ.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
n/a

### DIFF
--- a/service/history/ndc/workflow_state_replicator.go
+++ b/service/history/ndc/workflow_state_replicator.go
@@ -494,7 +494,7 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTaskWithNewRun(
 	sourceClusterName string,
 ) (continueProcess bool, retErr error) {
 	if archetypeID != chasm.WorkflowArchetypeID {
-		return false, serviceerror.NewInvalidArgument("Non worklfow should not have new run")
+		return false, serviceerror.NewInvalidArgument("Non workflow should not have new run")
 	}
 	nsEntry, err := r.namespaceRegistry.GetNamespaceByID(namespace.ID(executionInfo.NamespaceId))
 	if err != nil {


service/history/ndc/workflow_state_replicator.go
yycptt: nit: can we call it xxxArtifact?

service/history/ndc/workflow_state_replicator.go
```diff
@@ -352,55 +338,84 @@ func (r *WorkflowStateReplicatorImpl) ReplicateVersionedTransition(
...
 
-	wfCtx, releaseFn, err := r.workflowCache.GetOrCreateChasmExecution(
+func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTask(
+	ctx context.Context,
+	archetypeID chasm.ArchetypeID,
+	wfCtx historyi.WorkflowContext,
+	versionedTransition *replicationspb.VersionedTransitionArtifact,
+	sourceClusterName string,
+) (continueProcess bool, retErr error) {
+	mutation, snapshot, executionState, executionInfo, err := parseVersionedTransitionAttributes(versionedTransition)
```
yycptt, line 373: already parsed it before?

service/history/ndc/workflow_state_replicator.go
```diff
@@ -429,52 +444,193 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTask(
...
+
 	defer func() {
 		r.deleteNewBranchWhenError(ctx, newBranchToken, retErr)
 	}()
-	if err != nil {
-		return err
-	}
 
 	if mutation != nil {
 		err = localMutableState.ApplyMutation(mutation.StateMutation)
```
yycptt, line 457: oh didn't know we have apply mutation when not ms not exist case.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -429,52 +444,193 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTask(
...
 	} else {
 		err = localMutableState.ApplySnapshot(snapshot.State)
```
yycptt, line 459: shall we reuse applySnapshotWhenWorkflowNotExist in this case?

service/history/ndc/workflow_state_replicator.go
```diff
@@ -429,52 +444,193 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTask(
...
+	localMutableState := workflow.NewMutableState(
+		r.shardContext,
+		r.shardContext.GetEventsCache(),
+		r.logger,
+		nsEntry,
+		executionInfo.WorkflowId,
+		executionState.RunId,
+		timestamp.TimeValue(executionState.StartTime),
+	)
+	if archetypeID == chasm.WorkflowArchetypeID {
```
yycptt: let's return an error (softassert) if archetypeID is not workflow here.

service/history/ndc/workflow_state_replicator.go
xwduan: yeah, but passing those parameters around make the func parameter list too long. Those parse is very light weight, not really parsing, just extracting.

service/history/ndc/workflow_state_replicator.go
xwduan: applySnapshotWhenWorkflowNotExist has some complicated logic to handle event based replication. I think the logic is cleaner within this dedicated func.

service/history/ndc/workflow_state_replicator.go
yycptt: ah I see, cool.

service/history/ndc/workflow_state_replicator.go
```diff
@@ -430,52 +445,196 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTask(
...
+	ctx context.Context,
+	archetypeID chasm.ArchetypeID,
+	wfCtx historyi.WorkflowContext,
+	executionInfo *persistencespb.WorkflowExecutionInfo,
+	executionState *persistencespb.WorkflowExecutionState,
+	versionedTransition *replicationspb.VersionedTransitionArtifact,
+	sourceClusterName string,
+) (continueProcess bool, retErr error) {
+	if archetypeID != chasm.WorkflowArchetypeID {
+		return false, serviceerror.NewInvalidArgument("Non worklfow should not have new run")
```
yycptt:
```suggestion
		return false, serviceerror.NewInvalidArgument("Non workflow should not have new run")
```

PR #8811, opened 2025-12-12
Parallel Workflow Update Tests

stephanos:
## What changed?

Migrated `TestWorkflowUpdateSuite` away from testify's `Suite`; enabling parallel test execution.

**How it works**

- a test invokes `testcore.NewEnv(t)` to obtain a new `TestEnv`
- `TestEnv` sets `t.Parallel()` (_intentionally not giving a way to opt out!_)
- `TestEnv` obtains a test cluster from `clusterPool` (_or blocks if all are in-use right now_)
- env var `TEMPORAL_TEST_SHARED_CLUSTERS` controls size of the pool
- if a test relies on APIs like InjectHook, a dedicated cluster is used to prevent overlap
- env var `TEMPORAL_TEST_DEDICATED_CLUSTERS` controls number of dedicated clusters

**testify suites**

Existing test suites are limited by the same dedicated cluster pool to prevent creating too many clusters.

**Database connections**

SQLite setup for TestEnv-based func tests (ie only TestWorkflowUpdateSuite so far) has been changed to a file-based approach since that supports much better concurrency due to its WAL that an in-memory SQLite database does not support.

Connection limits for other databases were also raised due to connection errors.

**Planned follow-ups**

- Migrating the other testify suites should be fairly straight-forward with the use of AI agents.
- Reduce need for dedicated clusters by leveraging isolated namespace-per-test more.
- Eliminate all `time.Sleep`s.
- Tweak test cluster pool behavior.

## Why?

1. **Local speedup**: benchmarks show a ~50% speed increase (36.1s → 16.6s) for `TestWorkflowUpdateSuite`.

5. **Namespace isolation**: every test runs in its own namespace. This greatly reduces the risk of (accidental) collisions and also reduces the need to craft unique identifiers such as for task queues and workflow IDs.

6. **Deprecate testify suites**: Long-term strategy to remove use of testify suites in functional tests (one reason being their inability to run tests within a suite in parallel).

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential Issues

1. Logs become less useful since there is more interleaving of tests. 
2. Higher resource consumption: it requires more concurrent connections to databases and shows higher memory consumption (see 3 and 4). This could cause some short-term instability on CI. Note that some other PRs were merged to add mechanics for monitoring memory usage much better; which will help here.
4. Until all functional tests are converted, there is an imbalance in test cluster creation: migrated tests use the shared pool while current tests create one cluster each. Especially given the fact that some tests don't allow for test cluster sharing as they use non-parallelizable actions such as `InjectHook` or dynamic config overrides. With some more effort the number of these can be reduced.
7. Setup of test clusters was designed around the idea of short-lived clusters, one per suite. But when re-using them for longer, some of the assumptions don't hold anymore and increase memory usage. There's a band aid in place to limit how often a test cluster can be used before it's torn down. A long-term solution requires some design changes to how test clusters are started/used/torn down.
8. If there are certain cross-namespace issues or bugs that affect multiple tests, it might be harder to identify the root cause now. However; the existing test re-runs should at least mitigate these short-term.

### DIFF
--- a/tests/update_workflow_test.go
+++ b/tests/update_workflow_test.go
@@ -90,6 +90,11 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				s := testcore.NewEnv(t, testcore.WithDedicatedCluster())
 				runID := mustStartWorkflow(s, s.Tv())
 
+				tv := s.Tv()
+				if tc.useRunID {
+					tv = tv.WithRunID(runID)
+				}
+
 				capture := s.GetTestCluster().Host().CaptureMetricsHandler().StartCapture()
 				defer s.GetTestCluster().Host().CaptureMetricsHandler().StopCapture(capture)
 
@@ -152,7 +157,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				_, err := poller.PollAndProcessWorkflowTask()
 				s.NoError(err)
 
-				updateResultCh := sendUpdateNoError(s, useRunID(s.Tv(), tc.useRunID, runID))
+				updateResultCh := sendUpdateNoError(s, tv)
 
 				// Process update in workflow.
 				res, err := poller.PollAndProcessWorkflowTask(testcore.WithoutRetries)
@@ -179,7 +184,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				s.Equal(1, commits)
 				s.Equal(0, rollbacks)
 
-				events := s.GetHistory(s.Namespace().String(), useRunID(s.Tv(), tc.useRunID, runID).WorkflowExecution())
+				events := s.GetHistory(s.Namespace().String(), tv.WorkflowExecution())
 
 				s.EqualHistoryEvents(`
   1 WorkflowExecutionStarted
@@ -215,6 +220,10 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				s := testcore.NewEnv(t)
 				runID := mustStartWorkflow(s, s.Tv())
+				tv := s.Tv()
+				if tc.useRunID {
+					tv = tv.WithRunID(runID)
+				}
 
 				wtHandlerCalls := 0
 				wtHandler := func(task *workflowservice.PollWorkflowTaskQueueResponse) ([]*commandpb.Command, error) {
@@ -285,7 +294,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				_, err := poller.PollAndProcessWorkflowTask()
 				s.NoError(err)
 
-				updateResultCh := sendUpdateNoError(s, useRunID(s.Tv(), tc.useRunID, runID))
+				updateResultCh := sendUpdateNoError(s, tv)
 
 				// Process update in workflow.
 				res, err := poller.PollAndProcessWorkflowTask(testcore.WithoutRetries)
@@ -298,7 +307,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				s.Equal(2, wtHandlerCalls)
 				s.Equal(2, msgHandlerCalls)
 
-				events := s.GetHistory(s.Namespace().String(), useRunID(s.Tv(), tc.useRunID, runID).WorkflowExecution())
+				events := s.GetHistory(s.Namespace().String(), tv.WorkflowExecution())
 
 				s.EqualHistoryEvents(`
   1 WorkflowExecutionStarted
@@ -335,6 +344,10 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				s := testcore.NewEnv(t)
 				runID := mustStartWorkflow(s, s.Tv())
+				tv := s.Tv()
+				if tc.useRunID {
+					tv = tv.WithRunID(runID)
+				}
 
 				wtHandlerCalls := 0
 				wtHandler := func(task *workflowservice.PollWorkflowTaskQueueResponse) ([]*commandpb.Command, error) {
@@ -384,7 +397,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 					T:                   s.T(),
 				}
 
-				updateResultCh := sendUpdateNoError(s, useRunID(s.Tv(), tc.useRunID, runID))
+				updateResultCh := sendUpdateNoError(s, tv)
 
 				// Process update in workflow.
 				res, err := poller.PollAndProcessWorkflowTask(testcore.WithoutRetries)
@@ -398,7 +411,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				s.Equal(1, wtHandlerCalls)
 				s.Equal(1, msgHandlerCalls)
 
-				events := s.GetHistory(s.Namespace().String(), useRunID(s.Tv(), tc.useRunID, runID).WorkflowExecution())
+				events := s.GetHistory(s.Namespace().String(), tv.WorkflowExecution())
 
 				s.EqualHistoryEvents(`
   1 WorkflowExecutionStarted
@@ -431,6 +444,10 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				s := testcore.NewEnv(t)
 				runID := mustStartWorkflow(s, s.Tv())
+				tv := s.Tv()
+				if tc.useRunID {
+					tv = tv.WithRunID(runID)
+				}
 
 				wtHandlerCalls := 0
 				wtHandler := func(task *workflowservice.PollWorkflowTaskQueueResponse) ([]*commandpb.Command, error) {
@@ -497,7 +514,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				err = s.SendSignal(s.Namespace().String(), s.Tv().WorkflowExecution(), s.Tv().Any().String(), s.Tv().Any().Payloads(), s.Tv().Any().String())
 				s.NoError(err)
 
-				updateResultCh := sendUpdateNoError(s, useRunID(s.Tv(), tc.useRunID, runID))
+				updateResultCh := sendUpdateNoError(s, tv)
 
 				// Process update in workflow. It will be attached to existing WT.
 				res, err := poller.PollAndProcessWorkflowTask(testcore.WithoutRetries)
@@ -511,7 +528,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				s.Equal(2, wtHandlerCalls)
 				s.Equal(2, msgHandlerCalls)
 
-				events := s.GetHistory(s.Namespace().String(), useRunID(s.Tv(), tc.useRunID, runID).WorkflowExecution())
+				events := s.GetHistory(s.Namespace().String(), tv.WorkflowExecution())
 
 				s.EqualHistoryEvents(`
   1 WorkflowExecutionStarted
@@ -1307,6 +1324,10 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 			t.Run(tc.name, func(t *testing.T) {
 				s := testcore.NewEnv(t)
 				runID := mustStartWorkflow(s, s.Tv())
+				tv := s.Tv()
+				if tc.useRunID {
+					tv = tv.WithRunID(runID)
+				}
 
 				// Drain existing first WT from regular task queue, but respond with sticky queue enabled response, next WT will go to sticky queue.
 				_, err := s.TaskPoller().PollAndHandleWorkflowTask(s.Tv(),
@@ -1350,7 +1371,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				// This is to make sure that sticky poller above reached server first.
 				// And when update comes, stick poller is already available.
 				time.Sleep(500 * time.Millisecond) //nolint:forbidigo
-				updateResult := <-sendUpdateNoError(s, useRunID(s.Tv(), tc.useRunID, runID))
+				updateResult := <-sendUpdateNoError(s, tv)
 
 				s.Equal("success-result-of-"+s.Tv().UpdateID(), testcore.DecodeString(s.T(), updateResult.GetOutcome().GetSuccess()))
 
@@ -1364,7 +1385,7 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 			  7 WorkflowTaskCompleted
 			  8 WorkflowExecutionUpdateAccepted {"AcceptedRequestSequencingEventId": 5} // WTScheduled event which delivered update to the worker.
 			  9 WorkflowExecutionUpdateCompleted {"AcceptedEventId": 8}
-			`, s.GetHistory(s.Namespace().String(), useRunID(s.Tv(), tc.useRunID, runID).WorkflowExecution()))
+			`, s.GetHistory(s.Namespace().String(), tv.WorkflowExecution()))
 			})
 		}
 	})
@@ -3143,6 +3164,10 @@ func TestWorkflowUpdateSuite(t *testing.T) {
 				t.Run(tc.name+" "+wfCC.name, func(t *testing.T) {
 					s := testcore.NewEnv(t)
 					runID := mustStartWorkflow(s, s.Tv())
+					tv := s.Tv()
+					if wfCC.useRunID {
+						tv = tv.WithRunID(runID)
+					}
 
 					wtHandlerCalls := 0
 					wtHandler := func(task *workflowservice.PollWorkflowTaskQueueResponse) ([]*commandpb.Command, error) {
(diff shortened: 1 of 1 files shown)


common/cache/lru.go
```diff
@@ -144,8 +144,11 @@ func NewWithMetrics(maxSize int, opts *Options, handler metrics.Handler) Stoppab
 	if opts == nil {
 		opts = &Options{}
 	}
-	if opts.BackgroundEvict == nil {
-		opts.BackgroundEvict = func() dynamicconfig.CacheBackgroundEvictSettings {
+
+	// Use local variables to avoid modifying the input opts (which could be shared).
+	backgroundEvict := opts.BackgroundEvict
```
stephanos, line 148: Fixes race condition.

.github/workflows/run-tests.yml
```diff
@@ -532,8 +532,6 @@ jobs:
         run: ./develop/github/monitor_test.sh make functional-test-coverage
         env:
           TEST_ARGS: ${{ needs.set-up-single-test.outputs.single_test_args }}
-          TEMPORAL_TEST_OTEL_OUTPUT: ${{ github.workspace }}/.testoutput
-          TEMPORAL_OTEL_DEBUG: true
```
stephanos, line 538: Removed OTEL tracing for now as its design caused increased memory usage (details: the OTEL exporter is scoped to the lifetime of the cluster instead of a namespace which would be more efficient here).

common/persistence/persistence-tests/setup.go
```diff
@@ -66,21 +110,28 @@ func GetPostgreSQLPGXTestClusterOption() *TestBaseOptions {
...
+		DBPassword:      testSQLitePassword,
+		DBHost:          environment.GetLocalhostIP(),
+		DBPort:          0,
+		SchemaDir:       testSQLiteSchemaDir,
+		StoreType:       config.StoreTypeSQL,
+		ConnectAttributes: map[string]string{
+			"cache":        "shared",
+			"busy_timeout": "30000",
+			"journal_mode": "wal",
+			"synchronous":  "normal",
```
stephanos, line 124: Recommended settings I found online.

common/persistence/persistence-tests/setup.go
```diff
@@ -24,10 +30,45 @@ const (
 	testSQLiteSchemaDir = "schema/sqlite/v3" // specify if mode is not "memory"
 )
 
+// GetTestClusterOption returns test options for the given store type and driver.
```
stephanos, line 33: A little refactoring to make the persistence setup slightly more ergonomic.

tests/workflow_utils.go
```diff
@@ -1,21 +1,12 @@
...
 import (
-	"testing"
-
 	"go.temporal.io/api/workflowservice/v1"
-	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/testing/testvars"
 	"go.temporal.io/server/tests/testcore"
 )
 
-type testEnv interface {
```
stephanos, line 12: Superseded by `testcore.Env`.

tests/update_workflow_test.go
```diff
@@ -69,170 +45,198 @@ func (s *UpdateWorkflowSuite) speculativeWorkflowTaskOutcomes(
...
 	}
 
-	for _, tc := range testCases {
-		s.Run(tc.name, func() {
-			tv := testvars.New(s.T())
+	s.CloseShard(resp.NamespaceInfo.Id, wid)
+}
 
-			runID := mustStartWorkflow(s, tv)
+func TestWorkflowUpdateSuite(t *testing.T) {
```
stephanos, line 71: Changes here were kept to a minimum; helpers and tests were migrated to regular functions. The nested nature of testify suites was preserved, too.

tests/testcore/test_cluster_pool.go
```diff
@@ -0,0 +1,198 @@
...
+		return cluster
+	}
+
+	// If no custom dynamic config is provided, reuse an existing cluster.
+	return p.dedicated.get(t, func() *FunctionalTestBase {
+		return p.createCluster(t, nil, false)
+	})
+}
+
+func (p *clusterPool) createCluster(t *testing.T, dynamicConfig map[dynamicconfig.Key]any, shared bool) *FunctionalTestBase {
```
chaptersix, line 174: why does this return a `FunctionalTestBase` and not `TestCluster`? Does this keep use married to using testify suites?

stephanos: Good question; the short answer is: simplicity. I didn't want to tease out what we still need from  `FunctionalTestBase`. There's a lot in there. The right thing to do what be to consolidate `FunctionalTestBase` and `TestCluster`. That is probably much easier once all suites are migrated; and isn't a blocker IMO.

tests/testcore/test_env.go
```diff
@@ -0,0 +1,116 @@
...
+// This implies WithDedicatedCluster.
+func WithDynamicConfig(dc map[dynamicconfig.Key]any) TestOption {
+	return func(o *testOptions) {
+		o.dedicatedCluster = true
+		if o.dynamicConfig == nil {
+			o.dynamicConfig = make(map[dynamicconfig.Key]any)
+		}
+		maps.Copy(o.dynamicConfig, dc)
+	}
+}
```
chaptersix: I'd prefer that it doesn't imply a dedicated cluster and a change to the interface that allows checking the level as witch the namespace applies. Something like this:
```
  func WithDynamicConfig(dc map[dynamicconfig.GenericSetting]any) TestOption {
      return func(o *testOptions) {
          o.dedicatedCluster = true
          if o.dynamicConfig == nil {
              o.dynamicConfig = make(map[dynamicconfig.Key]any)
          }
          for setting, value := range dc {
              if setting.Precedence() == dynamicconfig.PrecedenceGlobal {
                  panic(fmt.Sprintf("global setting %s not allowed", setting.Key()))
              }
              if err := setting.Validate(value); err != nil {
                  panic(fmt.Sprintf("invalid value for %s: %v", setting.Key(), err))
              }
              o.dynamicConfig[setting.Key()] = value
          }
      }
  }
```

chaptersix: Maybe we also ass a required field that indicates when/if the override should be removed.

tests/testcore/functional_test_base.go
```diff
@@ -572,13 +598,34 @@ func (s *FunctionalTestBase) DurationNear(value, target, tolerance time.Duration
 // will automatically be reverted at the end of the test (using t.Cleanup). The cleanup
 // function is also returned if you want to revert the change before the end of the test.
 func (s *FunctionalTestBase) OverrideDynamicConfig(setting dynamicconfig.GenericSetting, value any) (cleanup func()) {
+	if s.isShared {
+		// Technically this doesn't always require a dedicated cluster since some dynamic configs can be applied to only
```
dnr: Can't this construct a ConstrainedValue with the namespace constraint from the value and use that? What specific flakiness are you worried about in that case?

And similar comment for WithDynamicConfig as Alex pointed out.

(Check precedence first, of course, and handle the case where value already is a []ConstrainedValue)

tests/testcore/test_env.go
stephanos: I was thinking that could be a follow-up but since both of you asked about it, I can get it in here already. I hadn;t thought about it deeper, so if it's that simple that's great.

tests/testcore/functional_test_base.go
stephanos: Alright; I've pushed https://github.com/temporalio/temporal/pull/8811/changes/ea9c1be081f0263a5c55723b60862dc11f3d7e66 which should address this. 

Both `OverrideDynamicConfig` and `WithDynamicConfig` now apply the namespace constraint automatically.

tests/testcore/test_env.go
stephanos: Alright; I've pushed https://github.com/temporalio/temporal/pull/8811/changes/ea9c1be081f0263a5c55723b60862dc11f3d7e66 which should address this. 

Both `OverrideDynamicConfig` and `WithDynamicConfig` now apply the namespace constraint automatically.

common/persistence/sql/test_sql_persistence.go
```diff
@@ -143,10 +143,11 @@ func (s *TestCluster) DropDatabase() {
...
 		if len(cfg2.DatabaseName) > 3 { // 3 should mean not ., .., empty, or /
-			err := os.Remove(cfg2.DatabaseName)
-			if err != nil {
-				panic(err)
-			}
+			// Remove main database file
+			_ = os.Remove(cfg2.DatabaseName)
+			// Remove WAL mode files (may not exist if WAL wasn't used)
+			_ = os.Remove(cfg2.DatabaseName + "-wal")
+			_ = os.Remove(cfg2.DatabaseName + "-shm")
```
dandavison, line 150: Is it attractive to place these in a directory and thus just remove the directory on cleanup?

docs/development/testing.md
```diff
@@ -14,6 +14,8 @@ This document describes the project's testing setup, utilities and best practice
 - `TEMPORAL_TEST_LOG_FORMAT`: Controls the output format for test logs. Available options: `json` or `console`
 - `TEMPORAL_TEST_LOG_LEVEL`:  Sets the verbosity level for test logging. Available levels: `debug`, `info`, `warn`, `error`, `fatal`
 - `TEMPORAL_TEST_OTEL_OUTPUT`: Enables OpenTelemetry (OTEL) trace output for failed tests to the provided file path.
+- `TEMPORAL_TEST_SHARED_CLUSTERS`: Number of shared clusters in the pool. Tests can use each simultaneously. 
+- `TEMPORAL_TEST_DEDICATED_CLUSTERS`: Number of dedicated clusters in the pool. Only one test can use each at a time.
```
dandavison: nit: I find this slightly easier to understand
 
```suggestion
- `TEMPORAL_TEST_SHARED_CLUSTERS`: Number of shared clusters in the pool. Each can be used by multiple tests simultaneously.
- `TEMPORAL_TEST_DEDICATED_CLUSTERS`: Number of dedicated clusters in the pool. Each can be used by one test only at a time.
```

tests/testcore/test_cluster_pool.go
```diff
@@ -0,0 +1,198 @@
...
+	"sync/atomic"
+	"testing"
+
+	"go.temporal.io/server/common/dynamicconfig"
+)
+
+var testClusterPool *clusterPool
+
+func init() {
+	sharedSize := runtime.GOMAXPROCS(0) / 2
```
dandavison: clamp at lower limit of 1?

tests/testcore/test_cluster_pool.go
```diff
@@ -0,0 +1,198 @@
...
+		usage := p.usageCounts[idx].Add(1)
+		if usage > int64(p.maxUsage) {
+			p.clusterMu[idx].Lock()
+			// Double-check after acquiring lock
+			if p.usageCounts[idx].Load() > int64(p.maxUsage) && p.clusters[idx] != nil {
+				poolType := "shared"
+				if p.slots != nil {
+					poolType = "dedicated"
+				}
+				fmt.Printf("[cluster] recreating %s cluster %d after %d tests\n", poolType, idx, p.maxUsage)
```
dandavison: should these print statements to stdout be logger calls? (here and below)

chaptersix (approved):

tests/update_workflow_test.go
```diff
@@ -4488,1156 +4566,1233 @@ func (s *UpdateWorkflowSuite) TestSpeculativeWorkflowTask_WorkerSkippedProcessin
...
-			s.Error(uwsRes.err)
-			s.Equal("Update-with-Start could not be executed.", uwsRes.err.Error())
-			errs := uwsRes.err.(*serviceerror.MultiOperationExecution).OperationErrors()
-			s.Len(errs, 2)
-			var alreadyStartedErr *serviceerror.WorkflowExecutionAlreadyStarted
-			s.ErrorAs(errs[0], &alreadyStartedErr)
-			s.Equal("Operation was aborted.", errs[1].Error())
-		})
+		t.Run("workflow is not running", func(t *testing.T) {
+			_ = testcore.NewEnv(t) // unused s
```
dandavison: Am I right in thinking that this holds on to a cluster slot for the duration of all the nested subtests, but doesn't actually do anything with it? And also creates a pointless namespace? Should we just call `t.Parallel()` here?

tests/testcore/test_env.go
```diff
@@ -0,0 +1,179 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/testing/historyrequire"
+	"go.temporal.io/server/common/testing/taskpoller"
+)
+
+type Env interface {
```
dnr, line 20: just curious: what other names did you consider besides Env? it feels slightly off, like, this is not an "environment" itself, it's more like a claim or reservation on a test environment. but I don't have any better suggestions

tests/testcore/test_env.go
```diff
@@ -0,0 +1,179 @@
...
+	nsName     namespace.Name
+	taskPoller *taskpoller.TaskPoller
+	t          *testing.T
+}
+
+type TestOption func(*testOptions)
+
+type testOptions struct {
+	dedicatedCluster      bool
+	dynamicConfigSettings []dynamicConfigOverride
```
dnr, line 47: curious why not map? don't most of the other dc-related things use maps?

tests/testcore/test_env.go
```diff
@@ -0,0 +1,179 @@
...
+	CloseShard(namespaceID string, workflowID string)
+	OverrideDynamicConfig(setting dynamicconfig.GenericSetting, value any) (cleanup func())
+}
+
+type testEnv struct {
+	*FunctionalTestBase
+	*require.Assertions
+	historyrequire.HistoryRequire
+
+	Logger log.Logger
```
dnr, line 34: should we throw in a TestVars or is that too opinionated?

tests/testcore/test_env.go
```diff
@@ -0,0 +1,179 @@
...
+type Env interface {
+	T() *testing.T
+	Namespace() namespace.Name
+	FrontendClient() workflowservice.WorkflowServiceClient
+	GetTestCluster() *TestCluster
+	CloseShard(namespaceID string, workflowID string)
+	OverrideDynamicConfig(setting dynamicconfig.GenericSetting, value any) (cleanup func())
+}
+
+type testEnv struct {
```
dnr, line 29: I assume this should implement Env? add assertion?

tests/testcore/test_env.go
```diff
@@ -0,0 +1,179 @@
...
+// OverrideDynamicConfig overrides a dynamic config setting for the duration of this test.
+// For settings that can be namespace-scoped, a namespace constraint is applied.
+// All others cannot be applied to a shared cluster and require `WithDedicatedCluster`.
+func (e *testEnv) OverrideDynamicConfig(setting dynamicconfig.GenericSetting, value any) (cleanup func()) {
+	if e.isShared {
+		if !canBeNamespaceScoped(setting.Precedence()) {
+			e.t.Fatalf("OverrideDynamicConfig for setting %s (precedence %v) cannot be called on a shared cluster; use testcore.WithDedicatedCluster()", setting.Key(), setting.Precedence())
+		}
+
+		// Wrap value with namespace constraint for test isolation on shared clusters.
```
dnr, line 160: nice!

tests/testcore/test_env.go
```diff
@@ -0,0 +1,179 @@
...
+			value = []dynamicconfig.ConstrainedValue{{
+				Constraints: dynamicconfig.Constraints{Namespace: ns},
+				Value:       value,
+			}}
+		}
+	}
+	return e.cluster.host.overrideDynamicConfig(e.t, setting.Key(), value)
+}
+
+func canBeNamespaceScoped(p dynamicconfig.Precedence) bool {
```
dnr, line 179: ah.. ideally this logic should be owned by dynamicconfig. but fine for now.

tests/testcore/test_cluster_pool.go
stephanos: oh sorry, these were part of my debugging. Will remove!

tests/update_workflow_test.go
stephanos: Good catch; these were still there from an earlier version where I had hierarchical test envs. But that didn't prove too useful. Will remove!

tests/testcore/test_env.go
stephanos: I had a map at first (see original commit), but then we changed it from `dynamicconfig.Key` to `dynamicconfig.GenericSetting` for the namespace constraint and then it wasn't indexable anymore.

tests/testcore/test_env.go
stephanos: I had that at first; but removed it again as I thought that would be too controversial and didn't want to get the review blocked on that discussion. It would save an extra line from every test that uses it. I'm in favor.

tests/testcore/test_env.go
stephanos: `Ctx` was one. `Suite` was another. Both seemed to close to their original inspiration. AI was suggesting `Harness`, too. Eh. Easy to rename if we find a better name later.

tests/testcore/test_env.go
dnr: yeah, I don't love the long names that TestVars generates and I've been meaning to fix it there, but good to consolidate on it

tests/testcore/test_env.go
dnr: oh right, I think I've run into that myself

PR #8814, opened 2025-12-12
Standalone activities to support for ID conflict use existing

fretz12:
## What changed?
Added support for ID conflict use existing. Added error details on conflict policy fail. 

## Why?
We now have the use existing policy support on chasm engine, so we can support it. Also need to pass back error details on conflict policy fail in struct ActivityExecutionAlreadyStartedError

## How did you test it?
- [X] built
- [ ] added new unit test(s)

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit f264bba049abc7357237f7e410fdc82be04a6f45. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/standalone_activity_test.go
+++ b/tests/standalone_activity_test.go
@@ -197,6 +197,7 @@ func (s *standaloneActivityTestSuite) TestIDConflictPolicy() {
 			},
 			StartToCloseTimeout: durationpb.New(1 * time.Minute),
 			IdConflictPolicy:    enumspb.ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING,
+			RequestId:           s.tv.RequestID(),
 		})
 		require.NoError(t, err)
 		require.Equal(t, firstStartResp.RunId, secondStartResp.RunId)


fretz12:
cursor review

fretz12:
> ### Bug: UseExisting policy returns incorrect Started=true for existing activity
> When `ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING` is used and an activity with the same ID already exists, the response incorrectly returns `Started: true`. The callback that creates the response with `Started: true` is always executed, even when the engine returns an existing activity via the `UseExisting` conflict policy. Per the established convention (seen in workflow's similar `USE_EXISTING` behavior in telemetry tests), `Started` should be `false` when returning an existing execution rather than creating a new one. This could mislead callers into thinking a new activity was started when it wasn't.
> 
> `chasm/lib/activity/handler.go#L68-L84`
> https://github.com/temporalio/temporal/blob/76033ba2d34992393cf49cf15f11b41dae85e553/chasm/lib/activity/handler.go#L68-L84
> 
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjI5YWM5OTBjLWQ2NzYtNDQ0Ny1hZWFlLWI0YTY3YzJmZjc1OCIsImVuY3J5cHRpb25LZXkiOiJOS3U2OG8tdU93Y3BybXJnNm5NOHFHcTY4NnQ1SERsQ3hCeTZuN01JUHRnIiwiYnJhbmNoIjoic2FhLWlkLXBvbGljeS1lbmhhbmNlIn0sImlhdCI6MTc2NTU4MjE1OCwiZXhwIjoxNzY4MTc0MTU4fQ.ETwYFd4kt9bKGGLml0uHsZTm-E7nq98NKn7mjuGHujp3Qb5RAFMAR0gbLRfzK8G0FBiq31Safcg4IKh7_pZufNdSS_ZLxKf1-rtao3cmTfA9a-yeO_FqPeS6ouRVsZfvbCVG-R74K14foJyCPBIx5vEgBlDOvprKkhBu9P82eJRx7kgInvpScL-kutq0aXmxj8-xV1PDp9b9Sdkuwrw9AR8SY1BzTnNQcHzCvSJKyHv4AR32JvOHgCT3OcBdsatLITKhOkcI4fgrWYfmlQtmwk25mepAjjJQXfZn1JvozUkwrFmu6a9gao3MVfJ1FmKoI5glmk6I1oemgYVuLLqhVQ) [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjI5YWM5OTBjLWQ2NzYtNDQ0Ny1hZWFlLWI0YTY3YzJmZjc1OCIsImVuY3J5cHRpb25LZXkiOiJOS3U2OG8tdU93Y3BybXJnNm5NOHFHcTY4NnQ1SERsQ3hCeTZuN01JUHRnIiwiYnJhbmNoIjoic2FhLWlkLXBvbGljeS1lbmhhbmNlIiwicmVwb093bmVyIjoidGVtcG9yYWxpbyIsInJlcG9OYW1lIjoidGVtcG9yYWwiLCJwck51bWJlciI6ODgxNCwiY29tbWl0U2hhIjoiNzYwMzNiYTJkMzQ5OTIzOTNjZjQ5Y2YxNWYxMWI0MWRhZTg1ZTU1MyIsInByb3ZpZGVyIjoiZ2l0aHViIn0sImlhdCI6MTc2NTU4MjE1OCwiZXhwIjoxNzY4MTc0MTU4fQ.OAFxEgHFOCL9jpbEnNYyu2TG_TVuj1-ZimhM_JYpZ2XzSIzT8ENd_wGlItDimcr1vbeAanjmmlAuS0aIGiQoD2vrsLFwkdPULEQdduNo-nghdqtIylLoJEpvcPzaBK3pGreHwJpmBRbxIdSRQzT8Fs0QZE3JUkThg33j0uqskO0qAnrWnxdOY_ceO8gVTSWkWh9CZ-ZbwXxBOnSXbWYFa1w18kWT57B_bltj89GZqHSFOj7MlGjv_H24Cw_FtFVjxVKEn9lxZ8_-FtQNPBzNFfJ7EZWfarWKdqjS-aOjd6ottTjX9-m2oTrJ4ZEu4GglujzIoryDp_egkEs82NMQwg)

That behavior would seem confusing as the activity is in Started state. WHere are you deducing this for workflows? Give me evidence

tests/standalone_activity_test.go
```diff
@@ -164,10 +165,48 @@ func (s *standaloneActivityTestSuite) TestIDConflictPolicy_FailsIfExists() {
...
+	t := s.T()
+	ctx, cancel := context.WithTimeout(t.Context(), 10*time.Second)
+	defer cancel()
+
+	activityID := s.tv.ActivityID()
+	taskQueue := s.tv.TaskQueue().String()
+
+	s.startAndValidateActivity(ctx, t, activityID, taskQueue)
+
+	// By default, unspecified conflict policy should be set to ACTIVITY_ID_CONFLICT_POLICY_FAIL, so no need to set explicitly
```
dandavison: I think you meant to delete this comment.

tests/standalone_activity_test.go
```diff
@@ -164,10 +165,48 @@ func (s *standaloneActivityTestSuite) TestIDConflictPolicy_FailsIfExists() {
...
+		ActivityType: s.tv.ActivityType(),
+		Identity:     s.tv.WorkerIdentity(),
+		Input:        defaultInput,
+		TaskQueue: &taskqueuepb.TaskQueue{
+			Name: taskQueue,
+		},
+		StartToCloseTimeout: durationpb.New(1 * time.Minute),
+		IdConflictPolicy:    enumspb.ACTIVITY_ID_CONFLICT_POLICY_USE_EXISTING,
+	})
+	require.NoError(t, err)
```
dandavison: I think we should check that the returned `runID` is the same as the first one, and that `started` is false.

tests/standalone_activity_test.go
```diff
@@ -164,10 +165,48 @@ func (s *standaloneActivityTestSuite) TestIDConflictPolicy_FailsIfExists() {
...
+			break
+		}
+	}
+	require.NotNil(t, details, "expected ActivityExecutionAlreadyStartedFailure in error details")
+	require.Equal(t, s.tv.RequestID(), details.StartRequestId)
+	require.Equal(t, startResponse.GetRunId(), details.RunId)
 }
 
-// TODO(fred): add test for BusinessIDConflictPolicyUseExisting after rebasing on main
+func (s *standaloneActivityTestSuite) TestIDConflictPolicy_UseExistingNoError() {
```
dandavison: nit: How about we group together all the conflict policy tests as subtests of `TestIDConflictPolicy`? I've been doing that for e.g. heartbeat tests.

chasm/lib/activity/handler.go
```diff
@@ -82,7 +86,22 @@ func (h *handler) StartActivityExecution(ctx context.Context, req *activitypb.St
...
+		var alreadyStartedErr *chasm.ExecutionAlreadyStartedError
+		if errors.As(err, &alreadyStartedErr) {
+			details := &errordetailspb.ActivityExecutionAlreadyStartedFailure{
+				StartRequestId: alreadyStartedErr.CurrentRequestID,
+				RunId:          alreadyStartedErr.CurrentRunID,
+			}
+
+			errStatus, errDetail := status.New(codes.AlreadyExists, "activity execution already started").WithDetails(details)
+			if errDetail != nil {
+				return nil, errDetail
```
dandavison: Should we actually be returning `err` here as opposed to `errDetail`?

My worry is that `errDetail` is going to contain an obscure message about something that went wrong while we were trying to construct details, whereas it is more helpful and appropriate to give the user `chasm.ExecutionAlreadyStartedError`.

chasm/lib/activity/handler.go
```diff
@@ -82,7 +86,22 @@ func (h *handler) StartActivityExecution(ctx context.Context, req *activitypb.St
...
+			}
+			return nil, errStatus.Err()
```
dandavison: I think that if there's a problem constructing details we should make sure we still return an `AlreadyExists`. So like this, but I think we should log the error also.

```suggestion
			s := status.New(codes.AlreadyExists, "activity execution already started")
			sd, err := s.WithDetails(details)
			if err != nil {
			    // log err
				return nil, s.Err()
			}
			return nil, sd.Err()
```

dandavison (approved):
Approved with that suggestion to change the error and some minor comments.

chasm/lib/activity/handler.go
fretz12: that makes sense, refactored and added logging

tests/standalone_activity_test.go
fretz12: removed

tests/standalone_activity_test.go
fretz12: Added runID assertion. 
As discussed, added a TODO. We need a way for chasm engine to tell us execution is already created. I'll dig through the code and ping chasm crew

tests/standalone_activity_test.go
fretz12: yup, saw it and love it. Refactored. 
Since the tests don't run in parallel, I would limit those to quick tests and not group too many together

PR #8817, opened 2025-12-13
Add standalone heartbeat by ID

fretz12:
## What changed?
Add standalone heartbeat by ID

## Why?
We want to support existing RPCs to activity heartbeat

## How did you test it?
- [X] built
- [ ] added new unit test(s)

---

> [!NOTE]
> Enable `RecordActivityTaskHeartbeatById` to heartbeat standalone activities (no `workflowId`) by embedding a `componentRef` in the activity task token; add a functional test verifying heartbeats keep the activity alive.
> 
> - **Frontend**
>   - `service/frontend/workflow_handler.go`
>     - `RecordActivityTaskHeartbeatById`: accept empty `workflowId` for standalone activities; construct and serialize `componentRef` via `chasm.NewComponentRef[*activity.Activity]` and include it in the activity `taskToken`.
> - **Tests**
>   - `tests/standalone_activity_test.go`
>     - Add `RecordHeartbeatByIDStaysAlive` to verify heartbeating-by-ID prevents timeout and activity completes successfully.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c2a431048a68a9f50ca6be1e5a304fa7bb5c51e9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/standalone_activity_test.go
+++ b/tests/standalone_activity_test.go
@@ -2011,17 +2011,6 @@ func (s *standaloneActivityTestSuite) TestHeartbeat() {
 		require.NotEmpty(t, pollTaskResp.TaskToken)
 
 		// Heartbeat before timeout
-		time.Sleep(600 * time.Millisecond) //nolint:forbidigo
-		_, err = s.FrontendClient().RecordActivityTaskHeartbeatById(ctx, &workflowservice.RecordActivityTaskHeartbeatByIdRequest{
-			Namespace:  s.Namespace().String(),
-			ActivityId: activityID,
-			RunId:      startResp.RunId,
-			Details:    heartbeatDetails,
-		})
-		require.NoError(t, err)
-
-		// Wait again, then heartbeat again
-		time.Sleep(600 * time.Millisecond) //nolint:forbidigo
 		_, err = s.FrontendClient().RecordActivityTaskHeartbeatById(ctx, &workflowservice.RecordActivityTaskHeartbeatByIdRequest{
 			Namespace:  s.Namespace().String(),
 			ActivityId: activityID,


fretz12:
cursor review

service/frontend/workflow_handler.go
```diff
@@ -1305,13 +1305,30 @@ func (wh *WorkflowHandler) RecordActivityTaskHeartbeatById(ctx context.Context,
...
 	if activityID == "" {
 		return nil, errActivityIDNotSet
 	}
 
+	// If workflowID is empty, it means the activity is a standalone activity and we need to set the component ref.
+	// Else this should be a validation error.
+	var componentRef []byte
+	if workflowID == "" {
+		if !wh.IsStandaloneActivityEnabled(request.GetNamespace()) {
+			return nil, errWorkflowIDNotSet
```
dandavison, line 1317: I think this should be `serviceerror.NewUnavailable(activity.StandaloneActivityDisabledError)`
for consistency with `RecordActivityTaskHeartbeat`.

tests/standalone_activity_test.go
```diff
@@ -1981,6 +1981,75 @@ func (s *standaloneActivityTestSuite) TestHeartbeat() {
 			"expected status=Completed but is %s", pollResp.GetInfo().GetStatus())
 		protorequire.ProtoEqual(t, defaultResult, pollResp.GetOutcome().GetResult())
 	})
+
+	t.Run("RecordHeartbeatByIDStaysAlive", func(t *testing.T) {
```
dandavison, line 1985: nit: maybe rather than this slow test we could just confirm that the API call works and returns cancel requested, or something like that?

dandavison (approved):
Approved with a comment.

service/frontend/workflow_handler.go
fretz12: If it returns here, this is the original error embedded activities had when no workflow ID is provided. I don't want to change existing behavior for that reason.

tests/standalone_activity_test.go
fretz12: I didn't understand the part on cancel requested, but you gave the idea to just heartbeat once and not to sleep since the other heartbeat test already does the long test.

service/frontend/workflow_handler.go
dandavison: Ah, I see, thanks. I agree with your version.

tests/standalone_activity_test.go
dandavison: In case it's helpful I was suggesting:

1. Start
2. RequestCancel
3. HeartbeatWithId
4. Confirm response contains `cancel_requested=true` (as a sanity check that the API call did pick up state from the desired activity)

PR #8821, opened 2025-12-14
Fix ContinuedFailure not cleared on null success payloads

### LINKED ISSUE #8490: Scheduled Actions doesn't clear ContinuedFailure on null success payloads
## Expected Behavior
- When a scheduled action succeeds, subsequent scheduled actions should not receive a `ContinuedFailure` failure payload.

## Actual Behavior
- If a scheduled action fails, setting `ContinuedFailure`'s payload, and then a subsequent scheduled action succeeds without returning a payload (a `null` payload), `ContinuedFailure` will continue to propagate to scheduled actions.

anmol242:
## What changed?
- Clear ContinuedFailure when workflow completes successfully regardless of whether the result payload is null
- Previously, ContinuedFailure was only cleared when res.GetResult() != nil, causing it to persist when workflows completed successfully with null payloads
- Add test TestContinuedFailureCleared to verify the fix
- Edge cases handled:
   - COMPLETED with result payload: Works as before (clears failure + sets result)
   - COMPLETED with null payload: Now correctly clears failure (the bug fix)
   - Other statuses (FAILED, CANCELED, TERMINATED): Unchanged behavior

## Why?
Fixed a bug where ContinuedFailure persisted when scheduled workflows completed successfully with null/empty result payloads. Previously, ContinuedFailure was only cleared when res.GetResult() != nil, causing it to incorrectly propagate to subsequent scheduled actions even after a successful completion. This ensures that any successful workflow completion clears the failure state, regardless of whether a result payload is present.

Fixes #8490 

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/worker/scheduler/workflow.go
+++ b/service/worker/scheduler/workflow.go
@@ -855,12 +855,13 @@ func (s *scheduler) processWatcherResult(id string, f workflow.Future, long bool
 	}
 
 	// handle last completion/failure and record resulting payload sizes
-	if res.GetResult() != nil {
-		s.State.LastCompletionResult = res.GetResult()
+	if res.Status == enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED {
 		s.State.ContinuedFailure = nil
-
-		if resultPayload, err := res.GetResult().Marshal(); err == nil {
-			s.metrics.Counter(metrics.SchedulePayloadSize.Name()).Inc(int64(len(resultPayload)))
+		if res.GetResult() != nil {
+			s.State.LastCompletionResult = res.GetResult()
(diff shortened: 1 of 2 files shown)


lina-temporal:
Hi Anmol,
Thanks for the fix! However, I believe this would break workflow determinism without introducing a new version to the scheduler's workflow, which prevents me from merging as-is. We're presently moving to [a new scheduler architecture](https://github.com/temporalio/temporal/blob/main/docs/architecture/schedules.md) where we'll be able to quicker address these sorts of issues. If you do need the fix in, please also add a replay compatibility test and gate the new behavior behind a workflow version bump.

lina-temporal (changes requested):
Requesting changes as above.

PR #8477, opened 2025-10-13
Update Configuration Loading

chaptersix:
## What Changed

This PR introduces a new `--config-file` flag (and the `TEMPORAL_SERVER_CONFIG_FILE_PATH ` environment variable) to remove the dependency on `dockerize` in the Temporal server Docker image.

When a configuration file is specified using either the CLI flag or the environment variable, the server will load configuration **only** from that file.
Users who want templating behavior similar to `dockerize` can enable it by adding the comment `# enable-template` at the top of the configuration file.

---

### Key Changes

1. **New `--config-file` flag:**

   * Adds a global `--config-file` flag that accepts a path to a single configuration file (absolute or relative to the project root).
   * Can also be set via the `TEMPORAL_SERVER_CONFIG_FILE_PATH ` environment variable.

2. **Deprecated legacy flags:**

   * The `--config`, `--env`, and `--zone` flags are now marked as **deprecated** in CLI help text.
   * These flags still work for backward compatibility.

3. **Embedded config template:**

   * The `config_template.yaml` file is now embedded in the binary to support loading configuration from environment variables.
   * Templating is supported if the file includes the `# enable-template` comment at the top.

4. **Templating support:**

   * Configuration files can use templating by including `# enable-template` at the beginning of the YAML file.

---

### Configuration Loading Priority (Highest to Lowest)

1. **`--config-file` specified** → Load that specific file
2. **`--config`, `--env`, or `--zone` specified** → Load from configuration directory (**deprecated**)
3. **No configuration specified** → Load from embedded template using environment variables (default)

---

### Expected Behavior

The following examples illustrate how the new configuration loading logic behaves:

* **Default behavior:**
  Running `temporal start` without flags loads configuration from environment variables only using the embedded template.

* **Using `--config-file`:**
  `temporal --config-file=/path/to/config.yaml start` loads configuration from the specified file path.

* **Using `TEMPORAL_SERVER_CONFIG_FILE_PATH`:**
  Setting `TEMPORAL_SERVER_CONFIG_FILE_PATH=/path/to/config.yaml temporal start` has the same effect as using the flag.

* **Validation and error handling:**
  The CLI returns clear error messages when conflicting flags or environment variables are used, or when a specified file does not exist.
---

## Breaking Change

The default behavior of `temporal start` has changed.
It now loads configuration **from environment variables** instead of using a default template path.

PR #8825, opened 2025-12-15
New Docker Build GH Actions

chaptersix:
## What changed?
* updates goreleaser to v2
* add GHA for build admin-tools and server images within this repo. `docker-builds` will only be used for building pre 1.30 images
* added GHA to promote docker builds from temporaliotest to temproalio docker image repos
* trivy security scanning gates for image promotion (pulled rom docker-builds repo). The gate can be overridden 

## Why?
* we decided to move away from building images in `docker-builds`. The complexity is not needed
* updates goreleaser to v2 because the v1 definitions might be no longer supported at some point and this is a good time to do it
* I used go scripts for the more complex flows instead of js or bash so we don't introduce another language contributors need to be familiar with. IMO the js or bash I did use it simple enough to understand. 

## How did you test it?
It passes in CI. 

## Potential risks
* these build pipelines are only compatible with the new docker images. 
* merging this PR may break our nightly tests. Will double check before merging
* flows that are not triggered by opening a PR are untested and therefore not completely validated

### DIFF
--- a/.github/actions/build-docker-images/scripts/main.go
+++ b/.github/actions/build-docker-images/scripts/main.go
@@ -322,7 +322,7 @@ func downloadCLI() error {
 func downloadCLIForArch(arch string) error {
 	cliVersion := os.Getenv("CLI_VERSION")
 	if cliVersion == "" {
-		defaultCliVersion
+		cliVersion = defaultCliVersion
 	}
 
 	tarballName := fmt.Sprintf("temporal_cli_%s_linux_%s.tar.gz", cliVersion, arch)


.github/actions/build-docker-images/scripts/main.go
```diff
@@ -0,0 +1,500 @@
+package main
```
chaptersix, line 1: go script used in the `build-docker-image` action

chaptersix:
Related Changes:
https://github.com/temporalio/temporal/pull/8671
https://github.com/temporalio/features/pull/709
https://github.com/temporalio/helm-charts/pull/797

chaptersix:
> Related Changes: #8671 [temporalio/features#709](https://github.com/temporalio/features/pull/709) [temporalio/helm-charts#797](https://github.com/temporalio/helm-charts/pull/797)

https://github.com/temporalio/temporal/pull/8477

.github/workflows/build-and-publish.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Build binaries
+        uses: ./.github/actions/build-binaries
+        with:
+          snapshot: true
+
+      - name: Build Docker images
+        uses: ./.github/actions/build-docker-images
+        with:
+          push: false
+          tag-latest: false
```
chaptersix, line 62: double check builds for main get the latest tag

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: ./.github/actions/build-binaries
+        with:
+          snapshot: false
+          release: true
+
+      - name: Build and push Docker images
+        uses: ./.github/actions/build-docker-images
+        with:
+          push: true
+          tag-latest: true
```
chaptersix: lates tag should false.

chaptersix: actually we may not need to build and image here.  build-and-publish Will publish and image.

.github/workflows/build-and-publish.yml
chaptersix: this is only for the feature builds. line 34 is correct.

.github/workflows/promote-server-image.yml
```diff
@@ -0,0 +1,199 @@
...
+              core.info('');
+              core.info('✓ Security scan passed. Proceeding with promotion.');
+              core.setOutput('can-promote', 'true');
+            }
+
+  promote-server:
+    needs: [validate-inputs, check-security-gate]
+    if: needs.check-security-gate.outputs.can-promote == 'true'
+    runs-on: ubuntu-latest
+    steps:
```
gow: There is a step `Validate Docker Hub credentials` in admin-tools promotion which seems to be missing here. Is that intentional?

.github/workflows/promote-server-image.yml
```diff
@@ -0,0 +1,199 @@
+name: Promote Server Image
```
gow, line 1: This is very similar to the previous file (`promote-admin-tools-image.yml`). Can these 2 workflows be composed from the same base file? (I'm not super familiar with github workflows and yaml composition)

.github/actions/build-docker-images/scripts/main.go
```diff
@@ -0,0 +1,558 @@
...
+		}
+	}
+
+	return nil
+}
+
+func downloadCLIForArch(arch string) error {
+	cliVersion := os.Getenv("CLI_VERSION")
+	if cliVersion == "" {
+		cliVersion = "1.5.1" // default fallback
```
gow: nit: the default is hard coded deep in the file. Maybe add const?

.github/actions/build-docker-images/scripts/main.go
```diff
@@ -0,0 +1,558 @@
...
+		// Copy file
+		if err := copyFile(src, dst); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func downloadFile(url, filepath string) error {
```
gow: Did the compiler not complain about this variable (`filepath`) being the same name as an imported package in this file?

gow (approved):
Looks good. See some comments below.

.github/workflows/promote-server-image.yml
chaptersix: thanks, I meant to remove it from both. It's not needed. If they are invalid creds the workflow will fail shortly after

.github/workflows/promote-server-image.yml
chaptersix: I can probably use a composite action. I'll give it a try

.github/actions/build-docker-images/scripts/main.go
chaptersix: will do

.github/actions/build-docker-images/scripts/main.go
chaptersix: I think that ends up being a lint warning. I'll modify it to be different

PR #8827, opened 2025-12-15
Define worker registry settings as dynamic config with aggressive eviction policy

rkannan82:
## What changed?
Define eviction related thresholds as dynamic config and set default values to evict aggressively. 

There are 2 types of eviction. Time based and capacity based. Capacity based kicks in once the max number of entries limit is reached.

| Setting | Before | After |
| ------- | -------- |---- |
|Eviction interval|1 hour|1 min|
|TTL for dead entry|24 hours|5 min|
|Min age to evict entry by capacity|10 min|1 min|

## Why?
To handle memory bloat when workers are constantly restarting. The new settings ensure we detect dead workers earlier and GC them.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Increased cpu due to aggressive gc schedule. If we see a problem, we can adjust values via dynamic config.

### DIFF
--- a/service/matching/workers/registry_impl.go
+++ b/service/matching/workers/registry_impl.go
@@ -290,22 +290,13 @@ func (m *registryImpl) filterWorkers(
 
 // evictLoop periodically triggers TTL and capacity-based eviction.
 func (m *registryImpl) evictLoop() {
-	evictionInterval := m.evictionIntervalFn()
-	ticker := time.NewTicker(evictionInterval)
 	for {
 		select {
-		case <-ticker.C:
+		case <-time.After(m.evictionIntervalFn()):
 			m.evictByTTL()
 			m.evictByCapacity()
 			m.recordUtilizationMetric()
-			// Check if eviction interval has changed and reset ticker if needed
-			newInterval := m.evictionIntervalFn()
-			if newInterval != evictionInterval {
-				ticker.Reset(newInterval)
-				evictionInterval = newInterval
-			}
 		case <-m.quit:
-			ticker.Stop()
 			return
 		}
 	}


common/dynamicconfig/constants.go
```diff
@@ -1390,6 +1390,34 @@ The metric has 2 dimensions: namespace_id and plugin_name. Disabled by default a
 an optional feature and also requires a metrics collection system that can handle higher cardinalities.`,
 	)
 
+	// Worker registry settings
+	MatchingWorkerRegistryEntryTTL = NewGlobalDurationSetting(
```
dnr, line 1400: what do you think about making these a single struct setting instead of four (actually five, include the one above?) individual settings?

service/matching/workers/registry_impl.go
```diff
@@ -43,16 +39,16 @@ type (
...
-		total                     atomic.Int64    // atomic counter of total entries
-		quit                      chan struct{}   // channel to signal shutdown of the eviction loop
-		seed                      maphash.Seed    // seed for the hasher, used to ensure consistent hashing
-		metricsHandler            metrics.Handler // metrics handler for recording registry metrics
-		enableWorkerPluginMetrics func() bool     // dynamic config function to control plugin metrics export
+		buckets                   []*bucket                        // buckets for partitioning the keyspace
+		maxItemsFn                dynamicconfig.IntPropertyFn      // dynamic config for maximum entries
+		ttlFn                     dynamicconfig.DurationPropertyFn // dynamic config for entry TTL
+		minEvictAgeFn             dynamicconfig.DurationPropertyFn // dynamic config for minimum evict age
+		evictionIntervalFn        dynamicconfig.DurationPropertyFn // dynamic config for eviction interval
```
dnr, line 42: try to ensure that the order of the fields in the struct def, the args in the constructor, and the fields in the initializer all match up, unless there's a good reason to differ. it makes things more readable and avoids confusion about missing fields/args.

service/matching/workers/registry_impl.go
```diff
@@ -302,13 +309,20 @@ func (m *registryImpl) filterWorkers(
...
 		case <-ticker.C:
 			m.evictByTTL()
 			m.evictByCapacity()
 			m.recordUtilizationMetric()
+			// Check if eviction interval has changed and reset ticker if needed
+			newInterval := m.evictionIntervalFn()
+			if newInterval != evictionInterval {
+				ticker.Reset(newInterval)
+				evictionInterval = newInterval
+			}
```
dnr: the simpler way to do this just `case <-time.After(m.evictionIntervalFn()):` instead of a ticker

service/matching/workers/registry_impl_test.go
```diff
@@ -29,20 +29,37 @@ func disablePluginMetrics() func() bool {
 	return func() bool { return false }
 }
 
+// Test helper config functions with configurable values
```
dnr: replace all of these (including above) with `dynamiconfig.Get{Int,Bool,Duration}ProperyFn`

rkannan82: Done.

service/matching/workers/registry_impl.go
rkannan82: Done.

common/dynamicconfig/constants.go
rkannan82: IIUC, then to change a single setting we always have to specify the entire set. That's not convenient. Right?

But I do see a readability problem in the constructor (newRegistryImpl). It is better to encapsulate these into a struct and pass a single argument. What do you think?

service/matching/workers/registry_impl.go
rkannan82: Sounds good. Especially since our timer is at minute level. So ok to create a new timer every time.

service/matching/workers/registry_impl.go
```diff
@@ -179,40 +182,23 @@ func (b *bucket) evictByCapacity(threshold time.Time) bool {
...
-		defaultMinEvictAge,
-		defaultMaxEntries,
-		defaultEvictionInterval,
-		metricsHandler,
-		enableWorkerPluginMetrics,
-	)
-
+func NewRegistry(lc fx.Lifecycle, params RegistryParams) Registry {
+	m := newRegistryImpl(params)
 	lc.Append(fx.StartStopHook(m.Start, m.Stop))
```
dnr, line 187: I would arrange this differently so that we keep the fx stuff out of this file and just in fx.go, and either do it without a struct or use fx.In, but I don't think we should hold this up over that, we can clean it up later.

common/dynamicconfig/constants.go
dnr: No, you can have a struct with default values and override just one. See e.g. DefaultDynamicRateLimitingParams in this package. And see ConvertStructure for the implementation.

rkannan82: Ok. will follow up in separate PR.

PR #8829, opened 2025-12-16
Send raw history events from matching to frontend service

prathyushpv:
## What changed?
This PR extends the raw history optimization to pass raw history bytes from History Service → Matching Service → Frontend without deserialization in Matching Service.

**Key changes:**
1. History Service:
    - When `SendRawHistoryBetweenInternalServices` is enabled, sets `RawHistoryBytes` (field 21) with raw proto-encoded history batches
2. Matching Service:
   - Passes raw history bytes through to frontend via `PollWorkflowTaskQueueResponseWithRawHistory`
   - Uses wire-compatible proto messages so gRPC auto-deserializes `[][]byte` → `History` on the client side
3. Frontend:
   - Receives raw history in `RawHistory` field (auto-deserialized by gRPC)
   - Processes search attributes for raw history since it bypasses history service's normal processing
4. Proto definitions:
   - Added `raw_history_bytes` (field 21) to `RecordWorkflowTaskStartedResponse`
   - Added `PollWorkflowTaskQueueResponseWithRawHistory` message with wire-compatible layout
   - Added `raw_history` (field 22) to `PollWorkflowTaskQueueResponse`

## Why?
When `history.sendRawHistoryBetweenInternalServices` is enabled, the previous implementation only avoided deserialization from persistence → History Service. However, Matching Service was still deserializing history events (via gRPC auto-deserialization) and re-serializing them when forwarding to Frontend.

This change eliminates that unnecessary serialization/deserialization cycle in Matching Service by:
1. Having History Service send raw bytes directly
2. Having Matching Service forward these raw bytes without parsing
3. Having Frontend receive the bytes which gRPC auto-deserializes

  This reduces CPU usage in Matching Service for workflows with large histories.

  ## How did you test it?
  - [x] built
  - [x] added new unit test(s)
  - [x] added new functional test(s) (`tests/workflow_task_test.go`)

## Potential risks
SendRawHistoryBetweenInternalServices must be disabled when rolling back from this version to an older version.

### DIFF
--- a/service/history/history_engine2_test.go
+++ b/service/history/history_engine2_test.go
@@ -40,6 +40,7 @@ import (
 	"go.temporal.io/server/common/payload"
 	"go.temporal.io/server/common/payloads"
 	"go.temporal.io/server/common/persistence"
+	"go.temporal.io/server/common/persistence/serialization"
 	"go.temporal.io/server/common/persistence/versionhistory"
 	"go.temporal.io/server/common/persistence/visibility/manager"
 	"go.temporal.io/server/common/primitives/timestamp"
@@ -234,9 +235,6 @@ func (s *engine2Suite) TearDownSubTest() {
 }
 
 func (s *engine2Suite) TestRecordWorkflowTaskStartedSuccessStickyEnabled() {
-	// Disable raw history to test the non-raw history path
-	s.config.SendRawHistoryBetweenInternalServices = func() bool { return false }
-
 	fakeHistory := []*historypb.HistoryEvent{
 		{
 			EventId:   int64(1),
@@ -341,9 +339,8 @@ func (s *engine2Suite) TestRecordWorkflowTaskStartedSuccessStickyEnabled() {
 }
 
 func (s *engine2Suite) TestRecordWorkflowTaskStartedSuccessStickyEnabled_WithInternalRawHistory() {
-	// Disable raw history to test the non-raw history path
-	s.config.SendRawHistoryBetweenInternalServices = func() bool { return false }
-
+	s.config.SendRawHistoryBetweenInternalServices = func() bool { return true }
+	serializer := serialization.NewSerializer()
 	fakeHistory := []*historypb.HistoryEvent{
 		{
 			EventId:   int64(1),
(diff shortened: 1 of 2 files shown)


proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -269,13 +269,30 @@ message RecordWorkflowTaskStartedResponse {
...
-    temporal.api.history.v1.History raw_history = 20;
-}
-
-// RecordWorkflowTaskStartedResponseWithRawHistory should be wire compatible with RecordWorkflowTaskStartedResponse.
-// The only difference is that RecordWorkflowTaskStartedResponseWithRawHistory has a `history` field that contains the
-// raw history batches. RecordWorkflowTaskStartedResponseWithRawHistory will be returned by history service.
-// History client will deserialize this message to RecordWorkflowTaskStartedResponse.
+    // Deprecated: This field is being replaced by raw_history_bytes which sends raw bytes
+    // instead of a proto-decoded History. This avoids matching service having to decode history.
+    temporal.api.history.v1.History raw_history = 20 [deprecated = true];
```
yycptt, line 289: plz add comment for the plan of deprecating related fields/messages.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -269,13 +269,30 @@ message RecordWorkflowTaskStartedResponse {
...
+    repeated bytes raw_history_bytes = 21;
+}
+
+// RecordWorkflowTaskStartedResponseWithRawHistory is wire-compatible with RecordWorkflowTaskStartedResponse.
```
yycptt, line 293: looks like this is no longer used and can be removed?

prathyushpv: We will have to keep this to add a flag to disable/enable this change. This message will be needed for the disabled case.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
prathyushpv: Done

yycptt: Thanks!

PR #8837, opened 2025-12-16
Emit poll_success/sync only on parent partition in new matcher

dnr:
## What changed?
Emit `poll_success` and `poll_success_sync` metrics only on parent partition in new matcher, to match the behavior of classic matcher.

## Why?
These were inadvertently being emitted on both child and parent when forwarding, which double-counts them. Also in the case of a poll timeout, the child was counting timeouts in poll_success, since a timeout looks like an empty "started" task.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/service/matching/pri_matcher.go
+++ b/service/matching/pri_matcher.go
@@ -478,7 +478,7 @@ func (tm *priTaskMatcher) poll(
 	defer func() {
 		// TODO(pri): can we consolidate all the metrics code below?
 		if pollMetadata.forwardedFrom == "" {
-			// Only recording for original polls
+			// Only recording for original polls (i.e. on child if forwarded)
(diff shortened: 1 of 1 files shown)


rkannan82 (approved):

service/matching/pri_matcher.go
```diff
@@ -510,16 +510,18 @@ func (tm *priTaskMatcher) poll(
 	pollWasForwarded = task.isStarted()
 	priority = task.getPriority().GetPriorityKey()
 
-	if !task.isQuery() {
-		if task.isSyncMatchTask() {
+	if !pollWasForwarded {
```
rkannan82, line 513: Suggest adding a comment above stating the semantics of this check in the case when the poll is forwarded.

PR #8840, opened 2025-12-16
Use commonnexus.PayloadSerializer to deserialize chasm invocations

spkane31:
## What changed?
Use the `commonnexus.PayloadSerializer` for deserializing the nexus payloads. This is applied to both CHASM and HSM

## Why?
Fixes a bug with payload deserialization in both CHASM and HSM implementations.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Minimal, this is not enabled out for CHASM and was an existing bug for HSM impl.

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -336,6 +336,9 @@ func (n *Node) SetRootComponent(
 	root := n.root()
 	root.value = rootComponent
 	root.setValueState(valueStateNeedSyncStructure)
+	if componentID, ok := n.registry.ComponentIDFor(rootComponent); ok {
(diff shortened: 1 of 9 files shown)


pdoerner (approved):
Do we need the same for HSM invocations?

lina-temporal:
> Do we need the same for HSM invocations?

Yeah, this should be done for HSM invocations as well.

spkane31:
> > Do we need the same for HSM invocations?
> 
> Yeah, this should be done for HSM invocations as well.

@pdoerner @lina-temporal this is done already in components/callbacks

PR #8841, opened 2025-12-16
Add a slack notification when main runs fail

spkane31:
## What changed?
Alert company slack when the main build fails after a PR merge

## Why?
Help us combat our issues with flaky tests.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
No,  not running in production

### DIFF
--- a/.github/workflows/features-integration.yml
+++ b/.github/workflows/features-integration.yml
@@ -82,33 +82,3 @@ jobs:
           if [[ -n $(echo "$RESULTS" | jq '.[] | select (. != "success")') ]]; then
             exit 1
           fi
-
-  notify-failure:
-    name: Notify Slack on Failure
-    if: |
-      always() &&
-      github.ref == 'refs/heads/main' &&
-      needs.feature-tests-status.result == 'failure'
-    needs: feature-tests-status
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      actions: read
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version-file: 'go.mod'
-          cache: true
-
-      - name: Send Slack notification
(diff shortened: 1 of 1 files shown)


.github/workflows/features-integration.yml
```diff
@@ -82,3 +82,38 @@ jobs:
           if [[ -n $(echo "$RESULTS" | jq '.[] | select (. != "success")') ]]; then
             exit 1
           fi
+
+  notify-failure:
```
stephanos: I believe the feature tests are not blocking merge?

.github/workflows/run-tests.yml
```diff
@@ -930,3 +930,38 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+
+      - name: Build ci-notify tool
+        run: |
+          cd cmd/tools/ci-notify
+          go build -o /tmp/ci-notify .
```
stephanos: Would it be simpler to use `go run` directly, skipping the `go build`?

tools/ci-notify/app_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Error("Debug output should contain short SHA")
+	}
+
+	if !strings.Contains(output, "Test Author") {
+		t.Error("Debug output should contain author")
+	}
+
+	if !strings.Contains(output, "test-job-1") {
+		t.Error("Debug output should contain failed job name")
+	}
```
stephanos: Should/could these be testify assertions instead? The entire test file actually.

spkane31: good call, used claude to generate most of this and didn't catch this

.github/workflows/features-integration.yml
spkane31: Good call, removing

PR #8844, opened 2025-12-17
Banish ghost tasks

dnr:
## What changed and why?
When a forwarded poll times out and returns a successful "no tasks", we created a "started" internal task wrapping the "no task" return value. This is a non-nil `*internalTask` but it has no task, it's a ghost. This is very confusing. It's also unnecessary: we can just use the error value `errNoTasks` which we already use to represent exactly this condition, and then construct a new empty task just before returning to the client.

There is one harmless change in behavior: previously when a poll timed out on the parent, the child just returned the "no tasks" value directly without waiting longer. Now the child will wait for a local match for another second before returning (both old and new matcher).

## How did you test it?
- [x] built
- [ ] added new unit test(s)


ShahabT (approved):
sweet

PR #8847, opened 2025-12-17
Do not throw unimplemented error if worker heartbeat feature is disabled

rkannan82:
## What changed?
Changed RecordWorkerHeartbeat API to return success instead of unimplemented error if worker heartbeat feature is disabled.

In addition, added the dynamic config check to other 2 APIs: Poll NexusTaskQueue and ShutdownWorker so that the behavior is consistent.

## Why?
Heartbeat API calls are internally made by SDK. If we return an error, then it can confuse users to think there is something problematic with their workflows.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -2744,7 +2744,7 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
 	}
 
 	// route heartbeat to the matching service
-	if request.WorkerHeartbeat != nil {
+	if request.WorkerHeartbeat != nil && wh.config.WorkerHeartbeatsEnabled(request.GetNamespace()) {
 		heartbeats := []*workerpb.WorkerHeartbeat{request.WorkerHeartbeat}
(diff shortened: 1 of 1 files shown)


service/frontend/workflow_handler.go
```diff
@@ -2744,7 +2744,7 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
 	}
 
 	// route heartbeat to the matching service
-	if request.WorkerHeartbeat != nil {
+	if request.WorkerHeartbeat != nil && wh.config.WorkerHeartbeatsEnabled(request.GetNamespace()) {
 		heartbeats := []*workerpb.WorkerHeartbeat{request.WorkerHeartbeat}
 		_, err = wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
```
dnr, line 2749: Do you think we should do this in a goroutine like in PollNexusTaskQueue? No need to change now though.

dnr (approved):

service/frontend/workflow_handler.go
rkannan82: Ack.

PR #8849, opened 2025-12-17
Implement pagination for ListWorkers.

rkannan82:
## What changed?
Implements cursor-based pagination for the ListWorkers API in the worker registry. This allows clients to efficiently retrieve large sets of workers in manageable chunks.

Changes
API Changes:
- Added ListWorkersParams struct to encapsulate query parameters (Query, PageSize, NextPageToken)
- Added ListWorkersResponse struct with Workers slice and NextPageToken
- Updated Registry.ListWorkers interface signature to use the new structs

Implementation (registry_impl.go):
- Added paginateWorkers function that handles cursor-based pagination
- Workers are sorted by WorkerInstanceKey for deterministic ordering when pagination is requested Sorting overhead is acceptable as we expect the number of workers to be in O(1000s). Even for 100K workers, the sort time < 100 milliseconds.

Handler (handler.go):
- Updated ListWorkers handler to extract pagination params from request and pass to registry

## Why?
This allows clients to efficiently retrieve large sets of workers in manageable chunks.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
Minimal. Existing calls that don't specify pagination params will not exercise the new path. We do an early return without applying pagination logic.

### DIFF
--- a/service/matching/workers/registry_test.go
+++ b/service/matching/workers/registry_test.go
@@ -1,6 +1,7 @@
 package workers
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


service/matching/workers/registry.go
```diff
@@ -6,9 +6,22 @@ import (
...
+	// ListWorkersParams contains parameters for listing workers.
+	ListWorkersParams struct {
+		Query         string
+		PageSize      int
+		NextPageToken []byte // Opaque token from a previous response to resume pagination.
+	}
+
+	// ListWorkersResponse contains the result of listing workers.
+	ListWorkersResponse struct {
+		Workers       []*workerpb.WorkerHeartbeat
```
ShahabT, line 18: It's a little odd to me that ListWorkers and DescribeWorker return `WorkerHeartbeat` and not `Worker`.

To me it seems more intuitive and futureproof if we return a `Worker` message that has a `last_heartbeat` field?

Are we certain that later no info other than the last heartbeat is going to be returned in these APIs?

service/matching/workers/registry_impl.go
```diff
@@ -374,21 +384,94 @@ func (m *registryImpl) RecordWorkerHeartbeats(nsID namespace.ID, nsName namespac
 	m.recordPluginMetric(nsName, workerHeartbeat)
 }
 
-func (m *registryImpl) ListWorkers(nsID namespace.ID, query string, _ []byte) ([]*workerpb.WorkerHeartbeat, error) {
-	if query == "" {
-		return m.filterWorkers(nsID, func(_ *workerpb.WorkerHeartbeat) bool { return true }), nil
+func (m *registryImpl) ListWorkers(nsID namespace.ID, params ListWorkersParams) (ListWorkersResponse, error) {
```
ShahabT, line 387: wouldn't all these filtering and pagination logic be removed once we move to CHASM? is there a need to really implement it rn?

rkannan82: You are right. The reason for adding this functionality now is to wrap up the current implementation, and make it usable from the UI (UI team is building this out). I wanted to decouple it from the chasm work as that will land only by end of Q1. This is the only remaining item.

service/matching/workers/registry.go
rkannan82: The matchingservice actually returns a WorkerInfo (heartbeat is wrapped inside it). See handler.go in this PR. The registry is an internal API and that is why it is returning the heartbeat info. We can always evolve this if needed. There are 3 options:

1. Current impl.
2. Wrap the heartbeat inside a WorkerInfo struct. Like you suggested.
3. Repurpose the matchingservice.ListWorkersResponse here.

We can go either way since this is all internal. Let me know what you think.

ShahabT: I think I confused this with service APIs. all good.

service/matching/workers/registry_impl.go
ShahabT: I'm gonna approve this part without reviewing as it's only temporary and for experiment.

ShahabT (approved):

service/matching/workers/registry.go
rkannan82: Discussed offline. Keeping it as is since this is an internal API. Public API is aligned with Shahab's comment.

PR #8851, opened 2025-12-17
Return InvalidArgument error when branch token is invalid

prathyushpv:
## What changed?
History manager returns deserialization error if the branch token in a request is invalid. Change that to return invalid argument error.

## Why?
This error is caused by bad request.

## How did you test it?
- [x] built
- [x] added new unit test(s)


yiminc (commented):
it could be a good to return some reference IDs so people encountering this could troubleshoot it.

prathyushpv:
> it could be a good to return some reference IDs so people encountering this could troubleshoot it.

We only have branch token at this level. We could return base64 encoded branch token. But the client would know the request details when it gets a failure.

PR #8852, opened 2025-12-17
Backlog statistics: Consider default queue when querying current/ramping versions

Shivs11:
## What changed?
- WISOTT

## Why?
 - Correctness.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
- None.

---

> [!NOTE]
> Improves correctness of task queue statistics by accounting for backlog in the default (unversioned) queue when reporting per-version data.
> 
> - Use v32 worker-deployment version strings in stats queries and recognition; include unversioned queue conditionally
> - Attribute default-queue backlog to `current`/`ramping` based on ramp percentage; avoid double counting by subtracting from unversioned when appropriate
> - Compute and adjust stats per priority; add helpers to clone/split/merge/subtract `TaskQueueStats` defensively (handle nil durations)
> - Update DescribeTaskQueue cache key path and propagate `Unversioned` flag only when needed
> - Add extensive unit/functional tests covering current-only, ramping-only, ramp to/from unversioned, inactive versions, unloaded queues, and multi-partition scenarios
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit daad6169137996dcfef3c019601df352bb8db51a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/matching/task_queue_partition_manager.go
+++ b/service/matching/task_queue_partition_manager.go
@@ -769,7 +769,7 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 		//
 		// When the ramping version is "unversioned", isRamping is true which shall make the attribution logic work as expected.
 		currentExists = currentVersion != nil
-		rampingExists = isRamping
+		rampingExists = isRamping && rampPercentage > 0
 
 		// Split the unversioned queue's stats per priority so TaskQueueStatsByPriorityKey can
 		// be adjusted consistently with TaskQueueStats.
@@ -899,7 +899,6 @@ func cloneStatsByPriority(in map[int32]*taskqueuepb.TaskQueueStats) map[int32]*t
 
 // splitTaskQueueStatsByRampPercentage splits a task queue stats record into "current" and "ramping" shares.
 // This split is applied independently per priority bucket (see splitStatsByPriorityByRampPercentage).
-// TODO (Shivam): Handle the special case of ramp to unversioned.
 func splitTaskQueueStatsByRampPercentage(
 	in *taskqueuepb.TaskQueueStats,
 	rampPct float32,
--- a/tests/task_queue_stats_test.go
+++ b/tests/task_queue_stats_test.go
@@ -378,6 +378,89 @@ func (s *TaskQueueStatsSuite) currentAbsorbsUnversionedBacklogWhenRampingToUnver
 	}, 10*time.Second, 200*time.Millisecond)
 }
 
+func (s *TaskQueueStatsSuite) TestRampingAbsorbsUnversionedBacklog_WhenCurrentIsUnversioned_SinglePartition() {
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
+
+	s.rampingAbsorbsUnversionedBacklogWhenCurrentIsUnversioned(1)
+}
+
+func (s *TaskQueueStatsSuite) TestRampingAbsorbsUnversionedBacklog_WhenCurrentIsUnversioned_MultiplePartitions() {
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 4)
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 4)
+
+	s.RunTestWithMatchingBehavior(func() {
+		s.rampingAbsorbsUnversionedBacklogWhenCurrentIsUnversioned(4)
+	})
+}
+
+func (s *TaskQueueStatsSuite) rampingAbsorbsUnversionedBacklogWhenCurrentIsUnversioned(numPartitions int) {
+	s.OverrideDynamicConfig(dynamicconfig.MatchingLongPollExpirationInterval, 10*time.Second)
+	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Millisecond) // zero means no TTL
+
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	tqName := testcore.RandomizeStr("tq-ramping-from-unversioned")
+	deploymentName := tqName + "-deployment"
+	rampingBuildID := "ramping-build-id"
+
+	pollCtx, cancelPoll := context.WithCancel(ctx)
+	s.createVersionsInTaskQueue(pollCtx, tqName, deploymentName, rampingBuildID)
+	cancelPoll() // cancel the pollers so that we can verify the backlog expectations
+
+	// Set current to unversioned (nil current version).
+	s.setCurrentVersion(deploymentName, "")
+
+	// Set ramping to a versioned deployment.
+	rampPercentage := 30
+	s.setRampingVersion(deploymentName, rampingBuildID, rampPercentage)
+
+	// Enqueue unversioned backlog.
+	unversionedWorkflowCount := 10 * numPartitions
+	s.startUnversionedWorkflows(unversionedWorkflowCount, tqName)
+
+	rampingExpectation := TaskQueueExpectations{
+		BacklogCount:     unversionedWorkflowCount * rampPercentage / 100,
+		MaxExtraTasks:    0,
+		ExpectedAddRate:  true,
+		ExpectedDispatch: false,
+	}
+	legacyExpectation := TaskQueueExpectations{
+		BacklogCount:     unversionedWorkflowCount,
+		MaxExtraTasks:    0,
+		ExpectedAddRate:  true,
+		ExpectedDispatch: false,
+	}
+
+	s.EventuallyWithT(func(c *assert.CollectT) {
+		a := require.New(c)
+
+		// We can't query "unversioned" as a WorkerDeploymentVersion, but we can validate that the ramping version
+		// is attributed its ramp share of the unversioned backlog.
+		s.requireWDVTaskQueueStatsStrict(
+			ctx,
+			a,
+			"DescribeWorkerDeploymentVersion[ramping][workflow][current-unversioned]",
+			tqName,
+			enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+			deploymentName,
+			rampingBuildID,
+			rampingExpectation,
+		)
+
+		// Legacy mode should continue to report the total backlog for the task queue.
+		s.requireLegacyTaskQueueStatsStrict(
+			ctx,
+			a,
+			"DescribeTaskQueue[legacy][workflow][current-unversioned]",
+			tqName,
+			enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+			legacyExpectation,
+		)
+	}, 10*time.Second, 200*time.Millisecond)
+}
+
 func (s *TaskQueueStatsSuite) rampingAndCurrentAbsorbsUnversionedBacklog(numPartitions int) {
 	s.OverrideDynamicConfig(dynamicconfig.MatchingLongPollExpirationInterval, 10*time.Second)
 	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Millisecond) // zero means no TTL
--- a/service/matching/physical_task_queue_manager.go
+++ b/service/matching/physical_task_queue_manager.go
@@ -893,6 +893,14 @@ func mergeStats(into, from *taskqueuepb.TaskQueueStats) {
 }
 
 func oldestBacklogAge(left, right *durationpb.Duration) *durationpb.Duration {
+	// Treat nil as zero to keep stats aggregation defensive. It is okay here to reassign the pointer values when
+	// they are nil since a nil Duration proto is equivalent to a zero duration.
+	if left == nil {
+		left = durationpb.New(0)
+	}
+	if right == nil {
+		right = durationpb.New(0)
+	}
 	if left.AsDuration() > right.AsDuration() {
 		return left
 	}


service/matching/matching_engine.go
```diff
@@ -1300,9 +1300,13 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		}
 
 		var buildIds []string
+		var reportUnversioned bool
+
 		if request.Version != nil {
 			// A particular version was requested. This is only available internally; not user-facing.
-			buildIds = []string{worker_versioning.WorkerDeploymentVersionToStringV31(request.Version)}
+			// TODO (Shivam): Confirm with the crew: I think we should be passing in v32 here and then further having checks in the partition manager
```
ShahabT: yeah it makes sense

service/matching/matching_engine.go
```diff
@@ -1358,22 +1369,25 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
...
 				for _, vii := range partitionResp.VersionsInfoInternal {
-					partitionStats := vii.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey
-					for pri, priorityStats := range partitionStats {
+					partitionStatsByPriority := vii.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey
+					for pri, priorityStats := range partitionStatsByPriority {
 						if _, ok := taskQueueStatsByPriority[pri]; !ok {
 							taskQueueStatsByPriority[pri] = &taskqueuepb.TaskQueueStats{}
 						}
-						mergeStats(taskQueueStats, priorityStats)
+						// mergeStats(taskQueueStats, priorityStats)
```
ShahabT: clean up?

service/matching/task_queue_partition_manager.go
```diff
@@ -387,6 +388,11 @@ func (pm *taskQueuePartitionManagerImpl) PollTask(
 
 	task, err := dbq.PollTask(ctx, pollMetadata)
 
+	// TODO (Shivam): Right now, this only considers the backlog of the versioned queue for the poller scaling decision.
```
ShahabT, line 393: this is kinda important and should be done before the next release because it's blocking poller auto-tuning stuff. (but it can come in a separate PR)

service/matching/task_queue_partition_manager.go
```diff
@@ -702,8 +708,10 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 		} else {
 			found := false
 			for k := range pm.versionedQueues {
-				// Storing the versioned queue if the buildID is a v2 based buildID or a versionID representing a worker-deployment version.
-				if k.BuildId() == b || worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(k.Deployment())) == b {
+				// Storing the versioned queue if the buildID is a v2 based buildID or a versionID representing a worker-deployment version (which could be v31 or v32)
+				if k.BuildId() == b ||
+					worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(k.Deployment())) == b ||
+					worker_versioning.ExternalWorkerDeploymentVersionToStringV31(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(k.Deployment())) == b {
```
ShahabT, line 716: I think we're ready to get rid of this representation from CLI commands in the next release. just created a task to remove the version string params from CLI and Go SDK before GA. and one release after than remove it from server as well.

service/matching/task_queue_partition_manager.go
```diff
@@ -727,6 +735,21 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+	if reportStats {
+		// Consider the default/unversioned queue. For current/ramping deployment versions, tasks are backlogged
+		// here, so we include this queue's stats if the version to describe is a current/ramping version.
+		unversionedStatsByPriority = pm.defaultQueue.GetStatsByPriority()
+
+		userData, _, err := pm.GetUserDataManager().GetUserData()
+		if err != nil {
+			return nil, err
+		}
+		perType := userData.GetData().GetPerType()[int32(pm.Partition().TaskType())]
```
ShahabT, line 759: it seems useful to extract this into a method in `UserDataManager.GetDeploymentData(taskType)` since it's repeating in multiple places.

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +772,59 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 			vInfo.PhysicalTaskQueueInfo.Pollers = physicalQueue.GetAllPollerInfo()
 		}
 		if reportStats {
-			vInfo.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey = physicalQueue.GetStatsByPriority()
-			vInfo.PhysicalTaskQueueInfo.TaskQueueStats = aggregateStats(vInfo.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey)
+			physicalStatsByPriority := physicalQueue.GetStatsByPriority()
+			vInfo.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey = physicalStatsByPriority
+
+			// Base stats are from the physical queue. For current/ramping deployment versions, add a proportion
+			// of the *aggregated* default/unversioned queue stats. We intentionally do NOT adjust the per-priority
```
ShahabT: I'm not sure why we should not adjust per-priority the same way as general stats? For one, it'd be weird for user to see total count of say 120 but that is higher than sum of all the counts by priority.

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +772,59 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+			// Attribution model:
+			// - If current and/or ramping deployment versions exist, we first "give away" a portion of the
+			//   unversioned queue's aggregated stats.
+			// - For the unversioned version itself, we subtract the given-away portion (so we don't double count).
+			// - For current/ramping versions, we add their share on top of their physical queue stats.
+
+			currentVersion, _, _, rampingVersion, isRamping, rampPercentage, _, _ := worker_versioning.CalculateTaskQueueVersioningInfo(deploymentData)
+			unversionedAgg := aggregateStats(unversionedStatsByPriority)
+			currentExists := currentVersion != nil
+			rampingExists := rampingVersion != nil && isRamping
```
ShahabT: isRamping seems to be based on status. better to change it to be based on percentage because we don't care if percentage is 0:

```suggestion
			rampingExists := rampingVersion != nil && rampPercentage>0
```

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +772,59 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+			currentShare := &taskqueuepb.TaskQueueStats{ApproximateBacklogAge: durationpb.New(0)}
+			rampShare := &taskqueuepb.TaskQueueStats{ApproximateBacklogAge: durationpb.New(0)}
+			if rampingExists {
+				currentShare, rampShare = splitAggregatedStatsByRampPercentage(unversionedAgg, rampPercentage)
+			} else if currentExists {
+				// If there exist no ramping version, we attribute the entire unversioned backlog to the current version.
+				currentShare = unversionedAgg
+			}
+
+			// fmt.Println("Deployment version for v:", v.Deployment())
```
ShahabT: nit: cleanup

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +772,59 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+
+			// Base stats are from the physical queue. For current/ramping deployment versions, add a proportion
+			// of the *aggregated* default/unversioned queue stats. We intentionally do NOT adjust the per-priority
+			// map in this case.
+			agg := aggregateStats(physicalStatsByPriority)
+
+			// Attribution model:
+			// - If current and/or ramping deployment versions exist, we first "give away" a portion of the
+			//   unversioned queue's aggregated stats.
+			// - For the unversioned version itself, we subtract the given-away portion (so we don't double count).
```
ShahabT: this part seems to be wasted once we move away from the deprecates version selection and only allow "whole TQ" or a single Version. (But it could be useful in the future if we provide a way to specifically ask for unversioned stats)

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +772,59 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+			// - For current/ramping versions, we add their share on top of their physical queue stats.
+
+			currentVersion, _, _, rampingVersion, isRamping, rampPercentage, _, _ := worker_versioning.CalculateTaskQueueVersioningInfo(deploymentData)
```
ShahabT: why do we do these things in the loop? it seems like a waste. should we allow the loop to calculate all the raw stats and then do adjustments just once at the end?

service/matching/task_queue_partition_manager.go
```diff
@@ -773,6 +848,71 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+
+	total := unversionedAgg.GetApproximateBacklogCount()
+	rampCount := int64(float64(total)*float64(rampPct)/100.0 + 0.5)
+	if rampCount < 0 {
+		rampCount = 0
+	} else if rampCount > total {
+		rampCount = total
+	}
+
+	currentCount := total - rampCount
```
ShahabT: I think we should Ceil each count. something like below:
```
rampCount := ceil(float64(total)*float64(rampPct)/100.0)
rampCount := ceil(float64(total)*float64(100-rampPct)/100.0)
```

This is so that an unversioned count of 1 gives you 1 for ramping and 1 for current. We want to be on the safe side.

service/matching/task_queue_partition_manager.go
```diff
@@ -773,6 +848,71 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 	}, nil
 }
 
+// splitAggregatedStatsByRampPercentage splits the unversioned queue's statistics into current and ramping shares
+// based on the ramp percentage.
+func splitAggregatedStatsByRampPercentage(
+	unversionedAgg *taskqueuepb.TaskQueueStats,
+	rampPct float32,
```
ShahabT, line 904: what happens if there is a current version AND ramp to unversioned? (it's fine if we don't add it in this PR but there should be TODO and task for it for later)

service/matching/task_queue_partition_manager.go
Shivs11: Yes, will be coming in a separate PR and have made a ticket for tracking too so I don't forget. Should be coming out soon.

service/matching/task_queue_partition_manager.go
Shivs11: yes. and that was the question that I was alluding to when I had asked this internally (but maybe I did not do a good job of asking it properly). I think my concern was how to split these tasks when these tasks have a priority set for them. We don't know, or at-least I didn't, which tasks of which priority shall be taken up by the current version, no?

Shivs11: Removing them from an arbitrary priority also feels kinda...wrong?

service/matching/task_queue_partition_manager.go
Shivs11: ack - shall move things out.

service/matching/task_queue_partition_manager.go
Shivs11: hmm......but that would be kinda confusing on the user's side?
it's better to show ramping 1, current 0 (or vice versa) than show current 1, ramping 1. Someone could be like one task is being split into two...

Shivs11: We should have a disclaimer saying that these stats show a theoretical prediction of the distribution rather than being exact. If so, then we would not need to ceil these.

service/matching/task_queue_partition_manager.go
Shivs11: may need some refactor in multiple places - i agree btw, shall debate putting it here or in a different PR.

service/matching/task_queue_partition_manager.go
Shivs11: are you talking about the comment here or the general approach of doing the splitting?

IMO, even when conversing about a world where a task queue's deployment Data would only contain the routingConfig, we still would have to fetch the ramping and current versions to get their shares.

Moreover, we do need unversioned stats in the first place to split something in the first place (and yes, if tomorrow we were to provide unversioned stats after the split, this could be nice)

service/matching/task_queue_partition_manager.go
Shivs11: hmmmm, this is a good catch...thinking

service/matching/task_queue_partition_manager.go
carlydf: backlog count is still called `approximateBacklogCount` right? if so, I think that already communicates that this is not 100% exact. If we're allowed to be not 100% exact, then I agree that erring on the side of rounding up is safer.

But I do get where you're coming from of wanting the sum to make sense.

ShahabT: user does not see the unversioned sum anyways. for scaling use cases is better to overcount than undercount. and the count is approximate anyways.

service/matching/task_queue_partition_manager.go
ShahabT: each priority in default queue will break individually the same way that you're doing it for aggregated stats.

In fact you can only split per-priority ones and then aggregate them without having to split the total stats.

Example:
ramp 40%

default queue: P0: 100 tasks, P1: 200 tasks
-> current version P0 gets 60 tasks and current version P1 gets 120 tasks
-> ramping version P0 gets 40 tasks and ramping version P1 gets 80 tasks

service/matching/task_queue_partition_manager.go
ShahabT: I mean this step particularly
>For the unversioned version itself, we subtract the given-away portion...

service/matching/task_queue_partition_manager.go
Shivs11: i'm fine ceiling this as long as users don't have unversioned "version" stats shown to them. The minute they do, I am on the side of things being confusing for a customer. 

Plus, I do agree that rounding up is better for scale up purposes.

service/matching/task_queue_partition_manager.go
Shivs11: ah, I had a wrong assumption while doing the math. sorry.

service/matching/task_queue_partition_manager.go
Shivs11: gonna be doing this later.

service/matching/task_queue_partition_manager.go
Shivs11: I think I am gonna still keep this explanation given that I am splitting stats the way i am right now. I kinda see your point about not having unversioned stats like ever be useful since we don't allow anyone to query them in a standalone manner, but can come back to this sometime later if things change down the line.

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +798,43 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+			// Attribution model (applied per-priority):
+			// - If current and/or ramping deployment versions exist, we first "give away" a portion of the
+			//   unversioned queue's per-priority stats.
+			//
+			// Depending on the version described, we have the following options:
+			// - For the unversioned version itself, subtract the given-away portion (so we don't double count).
+			// - For current/ramping versions, add their share on top of their physical queue stats.
+			deploymentVersion := worker_versioning.DeploymentVersionFromDeployment(v.Deployment())
+
+			isUnversionedDescribe := deploymentVersion == nil
```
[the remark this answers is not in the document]
Shivs11: I don't think we want this to be applied to v2 versioned queues afaik; this is for versioned queues that are either current/ramping.

service/matching/task_queue_partition_manager.go
Shivs11: added a todo and made a ticket - shall be on this next week!

service/matching/task_queue_partition_manager.go
```diff
@@ -727,6 +737,45 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 
 	pm.versionedQueuesLock.RUnlock()
 
+	var unversionedStatsByPriority map[int32]*taskqueuepb.TaskQueueStats
```
Shivs11, line 740: okay this can technically be refactored so that it looks neater given this function is going through a lot - however, have chosen to keep things here so that reviewing code is a little easier (have a different branch with some nice cleaner code)

the only thing that about this is that if someone were to just request stats for a v2 build, which has nothing to do with safe-deploys, we would be going down the path of querying the user data. I don't think these are expensive operations, but just placing this here as a potential for optimization (which I shall anyways be doing)

service/matching/matching_engine.go
```diff
@@ -1301,9 +1301,11 @@ func (e *matchingEngineImpl) DescribeTaskQueue(
 		}
 
 		var buildIds []string
+		var reportUnversioned bool
+
 		if request.Version != nil {
 			// A particular version was requested. This is only available internally; not user-facing.
-			buildIds = []string{worker_versioning.WorkerDeploymentVersionToStringV31(request.Version)}
+			buildIds = []string{worker_versioning.WorkerDeploymentVersionToStringV32(request.Version)}
```
carlydf, line 1308: are we indexing these with the V32 string now? I know at one point we were putting them in with the V31 string

service/matching/task_queue_partition_manager.go
```diff
@@ -747,7 +757,31 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+			worker_versioning.CalculateTaskQueueVersioningInfo(deploymentData)
+
+		// Technically, one could have a current version of "unversioned" which shall make currentExists false according
+		// to the current logic. However, as of now, the user cannot query the stats of the "unversioned" version so this
+		// logic is fine. In other words, this logic is used to only attribute the unversioned backlog to the current version
+		// when current version is NOT "unversioned".
+		//
+		// When the ramping version is "unversioned", isRamping is true which shall make the attribution logic work as expected.
+		currentExists = currentVersion != nil
+		rampingExists = isRamping
```
ShahabT: Nit: is still seems adding a `&& rampPercentage > 0` might slightly optimize this.

service/matching/task_queue_partition_manager.go
```diff
@@ -748,8 +806,50 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
+				// Only subtract the ramping share when ramping is to a versioned deployment. If ramping is to
+				// unversioned, that share should remain part of the unversioned queue stats.
+				if rampingExists && !isRampingToUnversioned {
+					subtractStatsByPriority(adjustedStatsByPriority, unversionedRampingShareByPriority)
+				}
+			} else if isCurrentDescribe && currentExists {
+				mergeStatsByPriority(adjustedStatsByPriority, unversionedCurrentShareByPriority)
+			} else if isRampingDescribe && rampingExists {
+				mergeStatsByPriority(adjustedStatsByPriority, unversionedRampingShareByPriority)
+			}
```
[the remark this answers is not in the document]
Shivs11: you would not have a condition where the current and the ramping would be the exact same though.

service/matching/task_queue_partition_manager.go
```diff
@@ -773,58 +807,49 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
...
 			physicalStatsByPriority := physicalQueue.GetStatsByPriority()
-			vInfo.PhysicalTaskQueueInfo.TaskQueueStatsByPriorityKey = physicalStatsByPriority
 
-			// Base stats are from the physical queue. For current/ramping deployment versions, add a proportion
-			// of the *aggregated* default/unversioned queue stats. We intentionally do NOT adjust the per-priority
-			// map in this case.
-			agg := aggregateStats(physicalStatsByPriority)
+			// Clone the physical queue's stats by priority so we can adjust (either add, subtract) them based on the
+			// attribution model defined below.
+			adjustedStatsByPriority := cloneStatsByPriority(physicalStatsByPriority)
```
ShahabT, line 813: I understand the need for cloning in line 783, but not sure I get it here. it seems all the add and subtract bellow is with the precalculated `unversionedCurrentShareByPriority` and `unversionedRampingShareByPriority` so seems safe to do that on top of physicalStatsByPriority itself without cloning it?

service/matching/task_queue_partition_manager.go
```diff
@@ -901,18 +955,78 @@ func splitAggregatedStatsByRampPercentage(
...
+	if len(from) == 0 {
+		return
+	}
+	for pri, s := range from {
+		if s == nil {
+			continue
+		}
+		ensureStatsWithAge(into, pri)
+		// mergeStats requires non-nil ApproximateBacklogAge on both inputs.
+		s = cloneTaskQueueStats(s)
```
ShahabT, line 993: same, why cloning here?

service/matching/task_queue_partition_manager.go
```diff
@@ -901,18 +955,78 @@ func splitAggregatedStatsByRampPercentage(
...
+func mergeStatsByPriority(into, from map[int32]*taskqueuepb.TaskQueueStats) {
+	if len(from) == 0 {
+		return
+	}
+	for pri, s := range from {
+		if s == nil {
+			continue
+		}
+		ensureStatsWithAge(into, pri)
+		// mergeStats requires non-nil ApproximateBacklogAge on both inputs.
```
ShahabT, line 992: Nit: better to improve mergeStats to treat nil age as 0.

tests/task_queue_stats_test.go
```diff
@@ -168,6 +168,601 @@ func (s *TaskQueueStatsSuite) TestAddMultipleTasks_MultiplePartitions_ValidateSt
...
 }
 
+func (s *TaskQueueStatsSuite) TestCurrentVersionAbsorbsUnversionedBacklog_NoRamping_SinglePartition() {
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
+
+	s.currentVersionAbsorbsUnversionedBacklogNoRamping(1)
+}
+
+func (s *TaskQueueStatsSuite) TestCurrentVersionAbsorbsUnversionedBacklog_NoRamping_MultiplePartitions() {
```
ShahabT, line 178: can we add a test for ramping _from_ unversioned too?

ShahabT (approved):
lgtm otherwise

service/matching/matching_engine.go
Shivs11: yes, just seemed right to do this because that's the latest (and final) version of these string changes i believe - moreover, have taken care inside task queue partition manager by checking if a versioned queue had a v31 format that equated to the requested buildID (which is being passed into the function).

service/matching/task_queue_partition_manager.go
Shivs11: sure; I think this was just done to be conservative on the odd event that there could be a data race that could spur up. I don't have an exact scenario in mind as to how that can happen, but the thought of modifying the underlying physical queue's stat's map directly (by doing arithmetic operations on it) seemed kinda risky....

service/matching/task_queue_partition_manager.go
Shivs11: same reasoning as above

service/matching/task_queue_partition_manager.go
ShahabT: since this logic is run also on the polls for poller autoscaling, we need to be extra careful with cpu and memory usage, so let's keep it efficient.

PR #8857, opened 2025-12-17
Filled remaining ActivityExecutionInfo fields

fretz12:
## What changed?
Filled remaining ActivityExecutionInfo fields. Enabled test that was blocked by chasm bug. Added schedule-to-start-timeout test. Test refactor.

## Why?
We need to return all the fields for GET RPCs. Test cleanup/completion before merge to main.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -686,7 +686,7 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
 		ScheduleToStartTimeout:  a.GetScheduleToStartTimeout(),
 		StartToCloseTimeout:     a.GetStartToCloseTimeout(),
 		StateTransitionCount:    a.Visibility.Get(ctx).Data.TransitionCount,
-		// StateSizeBytes: TODO do we still want this?
+		// TODO(saa-preview): StateSizeBytes?
 		SearchAttributes: sa,
 		Status:           status,
 		TaskQueue:        a.GetTaskQueue().GetName(),
--- a/chasm/lib/activity/activity_tasks.go
+++ b/chasm/lib/activity/activity_tasks.go
@@ -35,7 +35,7 @@ func (e *activityDispatchTaskExecutor) Validate(
 	_ chasm.TaskAttributes,
 	task *activitypb.ActivityDispatchTask,
 ) (bool, error) {
-	// TODO make sure we handle resets when we support them, as they will reset the attempt count
+	// TODO(saa-preview): make sure we handle resets when we support them, as they will reset the attempt count
 	return (TransitionStarted.Possible(activity) &&
 		task.Attempt == activity.LastAttempt.Get(ctx).GetCount()), nil
 }
--- a/chasm/lib/activity/statemachine.go
+++ b/chasm/lib/activity/statemachine.go
@@ -248,7 +248,7 @@ var TransitionTerminated = chasm.NewTransition(
 			}
 			outcome := a.Outcome.Get(ctx)
 			failure := &failurepb.Failure{
-				// TODO if the reason isn't provided, perhaps set a default reason. Also see if we should prefix with "Activity terminated: "
+				// TODO(saa-preview): if the reason isn't provided, perhaps set a default reason. Also see if we should prefix with "Activity terminated: "
 				Message:     req.GetReason(),
 				FailureInfo: &failurepb.Failure_TerminatedFailureInfo{},
 			}
--- a/chasm/lib/activity/validator.go
+++ b/chasm/lib/activity/validator.go
@@ -87,7 +87,7 @@ func validateActivityRetryPolicy(
 	if retryPolicy == nil {
 		return nil
 	}
-	// TODO: this is a namespace setting, not a namespace id setting
+	// TODO(saa-preview): this is a namespace setting, not a namespace id setting
 	defaultActivityRetrySettings := getDefaultActivityRetrySettings(namespaceID.String())
 	retrypolicy.EnsureDefaults(retryPolicy, defaultActivityRetrySettings)
 	return retrypolicy.Validate(retryPolicy)


service/frontend/configs/quotas.go
```diff
@@ -109,6 +109,7 @@ var (
 		"/temporal.api.workflowservice.v1.WorkflowService/PauseActivity":                         2,
 		"/temporal.api.workflowservice.v1.WorkflowService/UnpauseActivity":                       2,
 		"/temporal.api.workflowservice.v1.WorkflowService/ResetActivity":                         2,
+		"/temporal.api.workflowservice.v1.WorkflowService/UpdateWorkflowExecutionOptions":        2,
```
fretz12, line 112: Somehow this was removed in comparison to main and made tests fail. I've added it back.

dandavison: Ah thanks, I saw that in a different PR CI and was wondering.

chasm/lib/activity/activity.go
```diff
@@ -635,25 +634,62 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
 	attempt := a.LastAttempt.Get(ctx)
 	heartbeat, _ := a.LastHeartbeat.TryGet(ctx)
 	key := ctx.ExecutionKey()
+	currentTime := ctx.Now(a)
+
+	executionDuration := durationpb.New(currentTime.Sub(a.GetScheduleTime().AsTime()))
```
dandavison: This looks wrong to me. If the activity has completed already we should be using the time of completion.

chasm/lib/activity/activity.go
```diff
@@ -635,25 +634,62 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+	var nextAttemptScheduleTime *timestamppb.Timestamp
+	interval := attempt.GetCurrentRetryInterval()
+	completeTime := attempt.GetCompleteTime()
+	if interval != nil && interval.AsDuration() > 0 && completeTime != nil {
+		nextAttemptScheduleTime = timestamppb.New(completeTime.AsTime().Add(interval.AsDuration()))
+	}
```
dandavison, line 644: I'm thinking it would be better to store this in the component data when we actually schedule the task, rather than trying to recapitulate the logic here.

The number we compute like this won't exactly match. And in the future we might add the ability to update the retry interval (right?) making it completely wrong.

chasm/lib/activity/activity.go
```diff
@@ -635,25 +634,62 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+
+	var closeTime *timestamppb.Timestamp
+	if a.LifecycleState(ctx) != chasm.LifecycleStateRunning {
+		closeTime = attempt.GetCompleteTime()
+	}
```
dandavison, line 651: nit: could just trust `attempt.GetCompleteTime()` to be `nil` when appropriate?

chasm/lib/activity/activity.go
fretz12: Fixed as we discussed. Follows workflow executionDuration behavior.

chasm/lib/activity/activity.go
fretz12: Passed in currentTime into `recordFailedAttempt` and added TODO to debate if we should persist this field, as we discussed.

chasm/lib/activity/activity.go
fretz12: the attempt complete time could actually be nil on schedule-to-start, schedule-to-close timeouts as those timeouts don't overrwrite attemptState. Also same for cancelled and terminated.

chasm/lib/activity/activity.go
```diff
@@ -636,24 +635,62 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+	var nextAttemptScheduleTime *timestamppb.Timestamp
+	interval := attempt.GetCurrentRetryInterval()
+	completeTime := attempt.GetCompleteTime()
+	if interval != nil && interval.AsDuration() > 0 && completeTime != nil {
+		nextAttemptScheduleTime = timestamppb.New(completeTime.AsTime().Add(interval.AsDuration()))
+	}
+
+	var closeTime *timestamppb.Timestamp
+	var executionDuration = durationpb.New(0)
+	if a.LifecycleState(ctx) != chasm.LifecycleStateRunning && attempt.GetCompleteTime() != nil {
```
dandavison, line 648: not blocking at all but there is an API `a.LifecycleState(ctx).IsClosed()` so I guess we should use that.

tests/standalone_activity_test.go
```diff
@@ -382,63 +457,94 @@ func (s *standaloneActivityTestSuite) TestActivityCancelled() {
 	require.Equal(t, enumspb.ACTIVITY_EXECUTION_STATUS_CANCELED, info.GetStatus(),
 		"expected Canceled but is %s", info.GetStatus())
 	require.Equal(t, "Test Cancellation", info.GetCanceledReason())
+	require.Equal(t, info.GetExecutionDuration().AsDuration(), time.Duration(0)) // Canceled doesn't set attempt completion, thus expect 0 here
```
dandavison, line 460: For Canceled here, and also for Terminated, I'm thinking that we do actually want to return a non-zero execution time. Workflow does I believe. Discussed offline and agreed to track this pending decisions on how to implement it.

tests/standalone_activity_test.go
```diff
@@ -967,6 +1077,15 @@ func (s *standaloneActivityTestSuite) Test_ScheduleToCloseTimeout_WithRetry() {
...
 	require.Equal(t, enumspb.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE, pollActivityResp.GetOutcome().GetFailure().GetTimeoutFailureInfo().GetTimeoutType(),
 		"expected ScheduleToCloseTimeout but is %s", pollActivityResp.GetOutcome().GetFailure().GetTimeoutFailureInfo().GetTimeoutType())
+
+	describeResp, err := s.FrontendClient().DescribeActivityExecution(ctx, &workflowservice.DescribeActivityExecutionRequest{
+		Namespace:  s.Namespace().String(),
+		ActivityId: activityID,
+		RunId:      startResp.RunId,
+	})
+	require.NoError(t, err)
+	require.Greater(t, describeResp.GetInfo().GetExecutionDuration().AsDuration(), time.Duration(0)) // should have non-zero as attempts have been made
```
dandavison, line 1087: Same as comment elswehere -- we've discussed this and are tracking it. Whether or not the execution duration is returned shouldn't depend on whether attempts were made.

dandavison (approved):
LGTM! We've discussed the execution duration calculation and agreed that it's not quite right yet, but are not blocking this PR on it.

tests/standalone_activity_test.go
fretz12: To be discussed offline

tests/standalone_activity_test.go
fretz12: To be discussed offline

chasm/lib/activity/activity.go
```diff
@@ -636,24 +635,62 @@ func (a *Activity) buildActivityExecutionInfo(ctx chasm.Context) (*activity.Acti
...
+		RetryPolicy:             a.GetRetryPolicy(),
 		RunId:                   key.RunID,
 		RunState:                runState,
 		ScheduleTime:            a.GetScheduleTime(),
-		Status:                  status,
-		// TODO(dan): populate remaining fields
+		ScheduleToCloseTimeout:  a.GetScheduleToCloseTimeout(),
+		ScheduleToStartTimeout:  a.GetScheduleToStartTimeout(),
+		StartToCloseTimeout:     a.GetStartToCloseTimeout(),
+		StateTransitionCount:    a.Visibility.Get(ctx).Data.TransitionCount,
```
dandavison, line 688: I'm changing this line in https://github.com/temporalio/temporal/pull/8999

PR #8858, opened 2025-12-17
Authorize cross namespace commands

prathyushpv:
## What changed?
Authorize cross namespace signal, cancel and start child commands.

## Why?
To prevent these operations being executed without authentication.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/tools/elasticsearch/README.md
+++ b/tools/elasticsearch/README.md
@@ -24,15 +24,21 @@ COMMANDS:
    help, h         Shows a list of commands or help for one command
 
 GLOBAL OPTIONS:
-   --endpoint value           hostname or ip address of elasticsearch server (default: "http://127.0.0.1:9200") [$ES_SERVER]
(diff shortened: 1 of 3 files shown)


common/authorization/interceptor.go
```diff
@@ -255,3 +266,70 @@ func (a *Interceptor) getMetricsHandler(nsName string) metrics.Handler {
...
+	claims *Claims,
+	sourceNamespace string,
+	req interface{},
+) error {
+	// Skip if cross-namespace commands are not enabled
+	if !a.enableCrossNamespaceCommands() {
+		return nil
+	}
+
+	wftRequest, ok := req.(*workflowservice.RespondWorkflowTaskCompletedRequest)
```
yiminc, line 288: Add comment that, this is the only way to initiate cross-namespace commands

yiminc (approved):

common/authorization/interceptor.go
prathyushpv: Done

PR #8859, opened 2025-12-17
Alias imports in activity.go

fretz12:
## What changed?
Added alias to imports in activity.go

## Why?
When `make lint-code` is run locally, it is insistent on aliasing the activity API import to `activitypb "go.temporal.io/api/activity/v1"`. This is because the activity model `"go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"`  also has the name. I've renamed both so lint-code won't keep modifying it.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -69,8 +69,8 @@ type Activity struct {
 	// Pointer to an implementation of the "store". For a workflow activity this would be a parent
 	// pointer back to the workflow. For a standalone activity this is nil (Activity itself
 	// implements the ActivityStore interface).
-	// TODO: revisit a standalone activity pointing to itself once we handle storing it more efficiently.
-	// TODO: figure out better naming.
+	// TODO(saa-preview): revisit a standalone activity pointing to itself once we handle storing it more efficiently.
+	// TODO(saa-preview): figure out better naming.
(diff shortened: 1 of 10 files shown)


chasm/lib/activity/activity.go
```diff
@@ -16,7 +16,7 @@ import (
 	"go.temporal.io/server/api/matchingservice/v1"
 	tokenspb "go.temporal.io/server/api/token/v1"
 	"go.temporal.io/server/chasm"
-	"go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"
+	statepb "go.temporal.io/server/chasm/lib/activity/gen/activitypb/v1"
```
fretz12: open to a alternative alias

dandavison: Other chasm libhraries use `schedulerpb` and `callbackpb` (although we alias that to `callbackspb` for some reason!).

I vote we alias the api one to `apiactivitypb` here for consistency with other CHASM libraries.

fretz12: That's what I tried first, but the linter is very insistent on calling the API one activitypb

dandavison (approved):
LGTM!

PR #8861, opened 2025-12-18
Generate Docs fo Dynamic Configs

chaptersix:
## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	stdlog "log"
 	"os"
@@ -103,6 +104,28 @@ func buildCLI() *cli.App {
 					return fmt.Errorf("%d total errors", total)
 				}
 				return nil
+			},
+		},
+		{
+			Name:  "config",
+			Usage: "Configuration related commands",
+			Subcommands: []*cli.Command{
+				{
+					Name:      "dynamic",
+					Usage:     "Show all available dynamic configuration keys",
+					ArgsUsage: " ",
+					Flags: []cli.Flag{
+						&cli.BoolFlag{
+							Name:    "json",
+							Aliases: []string{"j"},
+							Usage:   "Output as JSON",
+						},
+					},
+					Action: func(c *cli.Context) error {
+						jsonOutput := c.Bool("json")
+						return showDynamicConfig(jsonOutput)
+					},
+				},
 			},
 		},
 		{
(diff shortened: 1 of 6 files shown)


chaptersix:
Sample output:
```
go run cmd/server/main.go config dynamic | head -n 10 
KEY                                                                TYPE                          SCOPE        DEFAULT
─────────────────────────────────────────────────────────────────  ────────────────────────────  ───────────  ────────────────────
matching.enableWorkerPluginMetrics                                 bool                          Global       false
history.enableWorkflowExecutionTimeoutTimer                        bool                          Global       true
history.acquireShardConcurrency                                    int                           Global       10
history.timerProcessorPollBackoffInterval                          duration                      Global       5s
history.replicatorProcessorMaxPollInterval                         duration                      Global       1m0s
worker.stickyCacheSize                                             int                           Global       0
worker.protectedNamespaces                                         []string                      Global       []
history.persistenceGlobalNamespaceMaxQPS                           int                           Namespace    0
```

cmd/server/main.go
```diff
@@ -105,6 +106,28 @@ func buildCLI() *cli.App {
 				return nil
 			},
 		},
+		{
+			Name:  "config",
```
dnr, line 110: Integrate this with the validate-dynamic-config command above (probably move that one into here)

common/dynamicconfig/setting.go
```diff
@@ -27,13 +33,23 @@ type (
...
+	}
+
 	// GenericSetting is an interface that all instances of Setting implement (by generated
 	// code in setting_gen.go). It can be used to refer to settings of any type and deal with
 	// them generically..
 	GenericSetting interface {
 		Key() Key
 		Precedence() Precedence
 		Validate(v any) error
+		Documentation() SettingDoc
```
dnr, line 52: There are already getters for Key and Precedence, let's just add new individual getters for description and default value instead of a new struct

cmd/tools/gendynamicconfig/main.go
```diff
@@ -10,6 +10,7 @@ type (
 	settingType struct {
 		Name      string
 		GoType    string
+		TypeName  string // lowercase type name for documentation
```
dnr, line 13: What's the value in having an extra value here over just using the Go type via reflection?

common/dynamicconfig/setting.go
```diff
@@ -48,3 +64,90 @@ type (
...
+	for i, cv := range cdef {
+		result[i] = TypedConstrainedValue[any]{
+			Constraints: cv.Constraints,
+			Value:       formatDefaultValue(cv.Value),
+		}
+	}
+	return result
+}
+
+// getTypeName returns a human-readable type name for documentation.
```
dnr, line 89: is all this complexity worth it instead of just fmt.Sprintf("%T", v)?

cmd/server/main.go
chaptersix: I'll pull in render-config as well

cmd/tools/gendynamicconfig/main.go
chaptersix: forgot why I added it, I'll see if i can remove it

common/dynamicconfig/setting.go
chaptersix: I'll see if I can remove it.

chaptersix:
will open another PR after adding the ability for tdbg to pull these configs.

PR #8862, opened 2025-12-18
CHASM: bug fix: handle empty run ID in PollComponent

dandavison:
## What changed?
In `chasm.PollComponent`, if the passed component ref has empty run ID, resolve this to the current run ID.

## Why?
The previous code state was broken: it was possible to submit `PollActivityExecution` with no run ID, which led to `Subscribe(key)` being called where key lacked a run ID. But the notifications were being sent to a different key (one with the run ID) which led to the long-poll never receiving any notification and hence always timing out.

Existing chasm APIs `ReadComponent` and `UpdateComponent` both accept empty run ID and silently resolve it to current using `getExecutionLease`. It was a bug that `PollComponent` resolved to current using `getExecutionLease`, but then subscribed to notifications using a key that did not contain the resolved run ID.

At the external API level, in several of our non-CHASM user-facing gRPC APIs, empty run ID means "use latest at the time the request is processed" (e.g. `GetWorkflowExecutionHistory`). And we want these semantics for CHASM APIs such as `PollActivityExecution`.  This change gives those semantics.

## How did you test it?
- [x] built
- [x] added new unit test(s)

---

> [!NOTE]
> PollComponent now subscribes using the resolved execution key (including RunID) to avoid missed notifications; adds unit/functional tests including PollActivityExecution with empty RunID.
> 
> - **History/CHASM**:
>   - `PollComponent`: subscribe via notifier using resolved workflow key from `executionLease.GetContext().GetWorkflowKey()` (`NamespaceID`, `BusinessID`, `RunID`) instead of `requestRef.ExecutionKey`.
>   - Minor loop simplification when rechecking predicate after notifications.
> - **Tests**:
>   - `service/history/chasm_engine_test.go`:
>     - Parameterized wait test for `PollComponent` with empty vs non-empty `RunID`; resolves empty `RunID` via `GetCurrentExecution` and verifies notifications use resolved `ExecutionKey`.
>     - Uses context timeouts for long-poll and validates returned ref contains actual `RunID`.
>   - `tests/standalone_activity_test.go`:
>     - Add `TestPollActivityExecution_EmptyRunID` to verify empty `RunId` resolves to current and returns expected outcome.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9cfa1282aa8a7605bc5670cc09fa3c5d0d0404c6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/chasm_engine_test.go
+++ b/service/history/chasm_engine_test.go
@@ -671,132 +671,75 @@ func (s *chasmEngineSuite) TestPollComponent_Success_NoWait() {
 	s.Equal(ref.BusinessID, newRef.BusinessID)
 }
 
-// TestPollComponent_EmptyRunID_Wait tests that PollComponent correctly resolves an empty RunID
-// before subscribing for notifications. This is a regression test: previously, PollComponent
-// would subscribe using the empty RunID from the request, but notifications are sent with the
-// resolved RunID, so the subscription would never receive any notification and the poll would
-// time out.
-func (s *chasmEngineSuite) TestPollComponent_EmptyRunID_Wait() {
-	tv := testvars.New(s.T())
-	actualRunID := tv.Any().RunID()
-	activityID := tv.ActivityID()
-
-	// The poll request uses empty RunID.
-	pollRef := chasm.NewComponentRef[*testComponent](
-		chasm.ExecutionKey{
-			NamespaceID: string(tests.NamespaceID),
-			BusinessID:  tv.WorkflowID(),
-			RunID:       "", // empty - must be resolved before subscribing
-		},
-	)
-
-	// The update request uses the actual RunID (as would happen in practice).
-	resolvedExecutionKey := chasm.ExecutionKey{
-		NamespaceID: string(tests.NamespaceID),
-		BusinessID:  tv.WorkflowID(),
-		RunID:       actualRunID,
-	}
-	updateRef := chasm.NewComponentRef[*testComponent](resolvedExecutionKey)
-
-	// GetCurrentExecution is called to resolve the empty RunID.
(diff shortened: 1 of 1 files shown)


service/history/chasm_notifier.go
```diff
@@ -32,7 +33,10 @@ func NewChasmNotifier() *ChasmNotifier {
 // been reached and resubscribe if necessary, while holding a lock on the execution. The caller must
 // arrange for the unsubscribe function to be called when they have finished monitoring the channel
 // for notifications. It is safe to call the unsubscribe function multiple times and concurrently.
-func (n *ChasmNotifier) Subscribe(key chasm.ExecutionKey) (<-chan struct{}, func()) {
+func (n *ChasmNotifier) Subscribe(key chasm.ExecutionKey) (<-chan struct{}, func(), error) {
+	if key.RunID == "" {
+		return nil, nil, serviceerror.NewInvalidArgument("run id is required")
+	}
```
dandavison: We could use a custom non-serviceerror error here if people prefer. Or we could do a pass on making error strategy consistent later.

service/history/chasm_engine.go
```diff
@@ -321,12 +324,9 @@ func (e *ChasmEngine) PollComponent(
...
 		case <-ch:
-			ref, err := checkPredicateOrSubscribe()
-			if err != nil {
-				return nil, err
-			}
-			if ref != nil {
-				return ref, nil
+			ref, err = checkPredicateOrSubscribe()
+			if err != nil || ref != nil {
+				return ref, err
```
dandavison, line 331: This is just some minor code cleanup, irrelevant to the PR.

fretz12 (approved):

awln-temporal (approved):

PR #8873, opened 2025-12-18
CHASM engine to pass back created flag on NewExecution

fretz12:
## What changed?
Added a return flag on the NewExecution function to indicate if a new execution was actually created. Also added docs to NewExecution. Changed standalone activity on start to pass back started flag if new exec was actually created.

## Why?
On NewExecution today if we use BusinessIDConflictPolicyUseExisting, we need a way to distinguish whether something is newly created or not. Based on that flag, caller code can react differently. 
For standalone activity, we must pass back the started flag according to whether a new exec was actually started, to matching existing workflow behavior.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


service/history/chasm_engine.go
```diff
@@ -141,14 +141,14 @@ func (e *ChasmEngine) NewExecution(
...
 	)
 	if err != nil {
-		return chasm.ExecutionKey{}, nil, err
+		return chasm.ExecutionKey{}, nil, false, err
 	}
 	if !hasCurrentRun {
 		serializedRef, err := newExecutionParams.executionRef.Serialize(e.registry)
 		if err != nil {
-			return chasm.ExecutionKey{}, nil, err
+			return chasm.ExecutionKey{}, nil, true, err
```
yycptt: nit: false

service/history/chasm_engine.go
```diff
@@ -561,11 +561,13 @@ func (e *ChasmEngine) handleExecutionConflict(
 
 	switch currentRunInfo.State {
 	case enumsspb.WORKFLOW_EXECUTION_STATE_CREATED, enumsspb.WORKFLOW_EXECUTION_STATE_RUNNING:
-		return e.handleConflictPolicy(ctx, shardContext, newExecutionParams, currentRunInfo, options.ConflictPolicy)
+		executionKey, serializedRef, err := e.handleConflictPolicy(
+			ctx, shardContext, newExecutionParams, currentRunInfo, options.ConflictPolicy)
+		return executionKey, serializedRef, false, err
```
yycptt: `handleConflictPolicy` may return true when we support ConflictPolicyTerminateExisting later, so let's return the flag from there.

service/history/chasm_engine.go
fretz12: Even though it hits a serialization error, isn't the execution already created?
We can make it false to be consistent for any error nonetheless

yycptt: hmmm yeah good point, can you help add a comment on that. Thanks!

with that said the flag won't be accurate anyway, like on L144 above the execution may actually be created despite an error is returned.

fretz12: True on L144. It gets fairly complicated to actually validate if the execution was actually persisted. Let me add a comment here to say that the false flag in L144 may not always be accurate.

service/history/chasm_engine.go
fretz12: done

chasm/engine.go
```diff
@@ -118,15 +141,40 @@ func WithRequestID(
...
+//   - [NewExecutionResult]: Contains the execution key, serialized ref, and whether a new execution was created
+//   - error: Non-nil if creation failed or policy constraints were violated
 func NewExecution[C Component, I any, O any](
 	ctx context.Context,
 	key ExecutionKey,
 	newFn func(MutableContext, I) (C, O, error),
 	input I,
 	opts ...TransitionOption,
-) (O, ExecutionKey, []byte, error) {
+) (O, NewExecutionResult, error) {
```
yycptt: 🤔 yeah I guess we can justify why O is standalone. 

I was thinking making NewExecutionResult generic and include O as well. The actual engine implementation doesn't need to populate it anyway, so we can just use `struct{}` there. Will need do an extra conversion here though. But i feel it looks nicer.

I don't think this is blocking.

cc @bergundy if you have any opinions here.

fretz12: I was pondering that as well... I do like how it wraps everything, but it also bothered me that `actual engine implementation doesn't need to populate it` and so its misleading for that field to exist from the engine impl perspective

bergundy: Not super opinionated but the generic struct generally makes sense to me if we are already returning a result object.

fretz12: Seems we're leaning towards wrapping the fn output. I've refactored it accordingly; take a look if you'd like

chasm/engine.go
```diff
@@ -75,6 +75,35 @@ type TransitionOptions struct {
...
+type NewExecutionResult[O any] struct {
+	ExecutionKey    ExecutionKey
+	NewExecutionRef []byte
+	Created         bool
+	Output          O
+}
+
+// NewEngineExecutionResult is a type alias for the result type returned by the Engine implementation.
+// This avoids repeating [struct{}] everywhere in the engine implementation.
+type NewEngineExecutionResult = NewExecutionResult[struct{}]
```
yycptt: minor naming suggestion, no strong opinion.
```suggestion
type EngineNewExecutionResult = NewExecutionResult[struct{}]
```

PR #8878, opened 2025-12-19
Make poller autoscaler also consider Unversioned queue stats for promoted versions

Shivs11:
## What changed?
- I have made changes in the function that tells the SDK whether to scale their pollers up or down. The changes now shall consider the unversioned queue's backlog if a task is being dispatched to a poller of the current/ramping version.

## Why?
- The main reason for this is that spooled tasks of the current/ramping version go to the unversioned queue's backlog.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None.

---

> [!NOTE]
> Improves autoscaling accuracy for versioned task queues by attributing unversioned backlog to promoted versions.
> 
> - `physicalTaskQueueManager.MakePollerScalingDecision` now accepts `context.Context` and pulls stats via `partitionMgr.GetPhysicalQueueAdjustedStats(...)`, which derives per-queue `TaskQueueStats` from `Describe(...)` with deployment-aware build IDs
> - New `taskQueuePartitionManagerImpl.GetPhysicalQueueAdjustedStats` adjusts stats for unversioned/current/ramping attribution; `MakePollerScalingDecision` logic updated to use `Get*` accessors and early-exit structure unchanged
> - Interface and mocks updated for the new method signature; call sites updated (`PollTask`) to pass context
> - Added integration tests verifying scale-up decisions for current and 100%-ramping deployment versions that absorb unversioned backlog; minor test update to include expected ramp percentage in worker versioning tests
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit acf401fa931c8f22389ca8fb2625713fb6b6e308. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/matching/physical_task_queue_manager.go
+++ b/service/matching/physical_task_queue_manager.go
@@ -784,16 +784,6 @@ func (c *physicalTaskQueueManagerImpl) counterFactory() counter.Counter {
 	return counter.NewHybridCounter(c.config.FairnessCounter(), src)
 }
 
-func (c *physicalTaskQueueManagerImpl) AllowPollerScalingDecision(now time.Time) bool {
-	// Avoid spiking pollers crazy fast by limiting how frequently change decisions are issued. Be more permissive when
-	// there are more recent pollers.
-	numPollers := c.pollerHistory.history.Size()
-	if numPollers == 0 {
-		numPollers = 1
-	}
-	return c.pollerScalingRateLimiter.AllowN(now, 1e6/numPollers)
-}
-
 func (c *physicalTaskQueueManagerImpl) GetFairnessWeightOverrides() fairnessWeightOverrides {
 	return c.partitionMgr.GetRateLimitManager().GetFairnessWeightOverrides()
 }
@@ -839,10 +829,12 @@ func (c *physicalTaskQueueManagerImpl) makePollerScalingDecisionImpl(
 	} else if !c.queue.Partition().IsRoot() {
 		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
 		return nil
-	} else if (stats.GetTasksAddRate() / stats.GetTasksDispatchRate()) > 1.2 {
-		// Increase if we're adding tasks faster than we're dispatching them. Particularly useful for Nexus tasks,
-		// since those (currently) don't get backlogged.
-		delta = 1
+	} else {
+		if (stats.GetTasksAddRate() / stats.GetTasksDispatchRate()) > 1.2 {
+			// Increase if we're adding tasks faster than we're dispatching them. Particularly useful for Nexus tasks,
+			// since those (currently) don't get backlogged.
+			delta = 1
+		}
 	}
 
 	if delta == 0 {
--- a/common/worker_versioning/worker_versioning_test.go
+++ b/common/worker_versioning/worker_versioning_test.go
@@ -406,7 +406,7 @@ func TestCalculateTaskQueueVersioningInfo(t *testing.T) {
 					},
 				},
 			}},
-		{name: "new format: versioned current with unversioned ramping in same deployment -> current is versioned and ramping is unversioned", wantCurrent: v1, wantRamping: nil,
+		{name: "new format: versioned current with unversioned ramping in same deployment -> current is versioned and ramping is unversioned", wantCurrent: v1, wantRamping: nil, wantRampPercentage: 20,
 			data: &persistencespb.DeploymentData{
 				Versions: []*deploymentspb.DeploymentVersionData{},
 				DeploymentsData: map[string]*persistencespb.WorkerDeploymentData{
--- a/service/matching/physical_task_queue_manager_interface.go
+++ b/service/matching/physical_task_queue_manager_interface.go
@@ -55,9 +55,6 @@ type (
 		// MakePollerScalingDecision makes a decision on whether to scale pollers up or down based on the current state
 		// of the task queue and the task about to be returned.
 		MakePollerScalingDecision(ctx context.Context, pollStartTime time.Time) *taskqueuepb.PollerScalingDecision
-		// AllowPollerScalingDecision returns whether it is allowed to emit a new poller scaling decision at this time.
-		// This keeps rate limiting encapsulated within the physical queue.
-		AllowPollerScalingDecision(now time.Time) bool
 		// GetFairnessWeightOverrides returns current fairness weight overrides for this queue.
 		GetFairnessWeightOverrides() fairnessWeightOverrides
 	}
--- a/service/matching/physical_task_queue_manager_mock.go
+++ b/service/matching/physical_task_queue_manager_mock.go
@@ -71,20 +71,6 @@ func (mr *MockphysicalTaskQueueManagerMockRecorder) AddSpooledTaskToMatcher(task
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "AddSpooledTaskToMatcher", reflect.TypeOf((*MockphysicalTaskQueueManager)(nil).AddSpooledTaskToMatcher), task)
 }
 
-// AllowPollerScalingDecision mocks base method.
-func (m *MockphysicalTaskQueueManager) AllowPollerScalingDecision(now time.Time) bool {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "AllowPollerScalingDecision", now)
-	ret0, _ := ret[0].(bool)
-	return ret0
-}
-
-// AllowPollerScalingDecision indicates an expected call of AllowPollerScalingDecision.
-func (mr *MockphysicalTaskQueueManagerMockRecorder) AllowPollerScalingDecision(now any) *gomock.Call {
-	mr.mock.ctrl.T.Helper()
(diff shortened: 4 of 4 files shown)


service/matching/task_queue_partition_manager.go
```diff
@@ -389,19 +389,76 @@ func (pm *taskQueuePartitionManagerImpl) PollTask(
...
+	stats := info.GetPhysicalTaskQueueInfo().GetTaskQueueStats()
+	if stats.ApproximateBacklogCount > 0 &&
+		stats.ApproximateBacklogAge.AsDuration() > pm.config.PollerScalingBacklogAgeScaleUp() {
+		// Always increase when there is a backlog, even if we're a partition. It's also important to increase for
+		// sticky queues.
+		delta = 1
+	} else if !physicalQueue.QueueKey().Partition().IsRoot() {
+		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
+		return nil, nil
+	} else if (stats.TasksAddRate / stats.TasksDispatchRate) > 1.2 {
```
carlydf: what is this 1.2 constant? if it's coming from somewhere else, can we define it as a constant and refer to it in both places?

Shivs11: this piece of code was directly copied over from the code that was placed in the physical_task_queue_manager file. However, your comment made me realize that I was not removing code in that file (tech debt)

making changes now.....

service/matching/physical_task_queue_manager.go
```diff
@@ -785,62 +785,18 @@ func (c *physicalTaskQueueManagerImpl) counterFactory() counter.Counter {
...
-) *taskqueuepb.PollerScalingDecision {
-	pollWaitTime := c.partitionMgr.engine.timeSource.Since(pollStartTime)
-	// If a poller has waited around a while, we can always suggest a decrease.
-	if pollWaitTime >= c.partitionMgr.config.PollerScalingWaitTime() {
-		// Decrease if any poll matched after sitting idle for some configured period
-		return &taskqueuepb.PollerScalingDecision{
-			PollRequestDeltaSuggestion: -1,
-		}
-	}
-
```
Shivs11, line 810: I have just moved this method from this physical_task_queue_manager to task_queue_partition_manager; the diff looks scary but it's not

service/matching/physical_task_queue_manager_test.go
```diff
@@ -410,90 +407,19 @@ func (s *PhysicalTaskQueueManagerTestSuite) TestTQMInterruptsPollOnClose() {
 	s.Less(time.Since(pollStart), 4*time.Second)
 }
 
-func (s *PhysicalTaskQueueManagerTestSuite) TestPollScalingUpOnBacklog() {
```
Shivs11, line 413: all these unit tests have been copied over and replicated in task_queue_partition_manager_test.go

service/matching/task_queue_partition_manager_test.go
```diff
@@ -584,6 +584,274 @@ func (s *PartitionManagerTestSuite) TestDescribeTaskQueuePartition_UnversionedDo
...
+	s.describeStatsEventually(map[string]bool{"": true}, false, false, false, func(resp *matchingservice.DescribeTaskQueuePartitionResponse) bool {
+		uvStats := resp.VersionsInfoInternal[""].GetPhysicalTaskQueueInfo().GetTaskQueueStats()
+		return uvStats != nil && uvStats.GetApproximateBacklogCount() >= 100
+	})
+
+	// Should emit scaling decision for high backlog on non-root partition
+	decision, err := s.partitionMgr.MakePollerScalingDecision(
+		ctx,
+		time.Now(),
+		s.partitionMgr.defaultQueue,
```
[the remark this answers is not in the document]
Shivs11: a non-root partition also has a default queue so this isn't a valid comment. A non-root partition was created for this test and the default physical queue was created in this partition.

service/matching/task_queue_partition_manager.go
```diff
@@ -389,19 +389,93 @@ func (pm *taskQueuePartitionManagerImpl) PollTask(
...
-		task.pollerScalingDecision = dbq.MakePollerScalingDecision(pollMetadata.localPollStartTime)
+		task.pollerScalingDecision = pm.MakePollerScalingDecision(ctx, pollMetadata.localPollStartTime, dbq)
 	}
 
 	return task, versionSetUsed, err
 }
 
+// getPhysicalQueueStats retrieves task queue stats for the given physical queue.
+// Returns nil if stats cannot be retrieved or are not available.
+func (pm *taskQueuePartitionManagerImpl) getPhysicalQueueStats(
```
ShahabT: maybe call it `getPhysicalQueueAdjustedStats` otherwise one would think this is the raw stats from the name.

service/matching/physical_task_queue_manager.go
```diff
@@ -783,62 +783,18 @@ func (c *physicalTaskQueueManagerImpl) counterFactory() counter.Counter {
 	return counter.NewHybridCounter(c.config.FairnessCounter(), src)
 }
 
-func (c *physicalTaskQueueManagerImpl) MakePollerScalingDecision(
-	pollStartTime time.Time) *taskqueuepb.PollerScalingDecision {
-	return c.makePollerScalingDecisionImpl(pollStartTime, func() *taskqueuepb.TaskQueueStats {
-		return aggregateStats(c.GetStatsByPriority())
```
ShahabT, line 790: hmm... I'm not sold on this refactoring, it doesn't feel right to put `AllowPollerScalingDecision` in one place and `MakePollerScalingDecision`... why not just replacing this line with something like:
```suggestion
    return c.partitionMgr.getPhysicalQueueAdjustedStats(c)
```

Shivs11: why though? I understand the argument about keeping scaling related decisions encapsulated together in the same place, but I think it kinda makes sense for AllowPollerScalingDecision to be in the physical task queue manager. The function returns an answer based on the poller history, which to me is something that is a property of a physical task queue manager.

or am I missing something?

ShahabT: It's not as mush about what AllowPollerScalingDecision uses to make the decision, it is that the decision to send or not send the scaling decision is part of the scaling algorithm itself. to see that more clearly, imagine physical_task_queue_manager decides tomorrow allow less or more decisions by changing the implementation of AllowPollerScalingDecision, that would change the behavior of scaling drastically. Hence, physical_task_queue_manager is not the owner of the logic.

conceptually I'm fine for physical_task_queue_manager to expose a public NumPollers method so the scaling logic uses that when calculating the rate limit. But since the code is already written in physical_task_queue_manager I don't see much value in moving it to partition manager anyways. There could be value in extracting it completely into another struct like poller_scaler or similar, but even that may be premature now.

Shivs11: hmm, I think I see what you mean.

service/matching/physical_task_queue_manager.go
```diff
@@ -783,17 +783,28 @@ func (c *physicalTaskQueueManagerImpl) counterFactory() counter.Counter {
...
 }
 
+func (c *physicalTaskQueueManagerImpl) MakePollerScalingDecision(
+	ctx context.Context,
+	pollStartTime time.Time) *taskqueuepb.PollerScalingDecision {
+	return c.makePollerScalingDecisionImpl(pollStartTime, func() *taskqueuepb.TaskQueueStats {
+		return c.partitionMgr.GetPhysicalQueueAdjustedStats(ctx, c)
+	})
+}
+
```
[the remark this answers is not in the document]
Shivs11: this won't panic with my latest commit which shall ensure that all stats fetches are nil-safe.

service/matching/physical_task_queue_manager.go
```diff
@@ -784,17 +784,28 @@ func (c *physicalTaskQueueManagerImpl) counterFactory() counter.Counter {
...
-		return aggregateStats(c.GetStatsByPriority())
-	})
+func (c *physicalTaskQueueManagerImpl) AllowPollerScalingDecision(now time.Time) bool {
+	// Avoid spiking pollers crazy fast by limiting how frequently change decisions are issued. Be more permissive when
+	// there are more recent pollers.
+	numPollers := c.pollerHistory.history.Size()
+	if numPollers == 0 {
+		numPollers = 1
+	}
+	return c.pollerScalingRateLimiter.AllowN(now, 1e6/numPollers)
```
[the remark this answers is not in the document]
Shivs11: the method does not exist

service/matching/physical_task_queue_manager.go
```diff
@@ -820,18 +821,20 @@ func (c *physicalTaskQueueManagerImpl) makePollerScalingDecisionImpl(
...
 		delta = 1
 	} else if !c.queue.Partition().IsRoot() {
 		// Non-root partitions don't have an appropriate view of the data to make decisions beyond backlog.
 		return nil
-	} else if (stats.TasksAddRate / stats.TasksDispatchRate) > 1.2 {
-		// Increase if we're adding tasks faster than we're dispatching them. Particularly useful for Nexus tasks,
-		// since those (currently) don't get backlogged.
-		delta = 1
+	} else {
+		if (stats.GetTasksAddRate() / stats.GetTasksDispatchRate()) > 1.2 {
```
carlydf: nit: `else if {...}` is cleaner than `else { if {...}}` but not a big deal

ShahabT (approved):

PR #8882, opened 2025-12-19
Add standalone activity validate token tests

fretz12:
## What changed?
- Added token validation tests for activity complete, fail, and cancel. Refactored test layout. 
- Update activity proto docs. Removed TODOs.
- Changed activity closeTime and executionDuration to always be non-nil when the activity is in terminal state
- Add validation on complete/fail/cancel/heartbeat token to ensure chasmRef namespace ID matches that of the request

## Why?
As part of completing TODOs and cleanup of standalone activities. 
Any terminal state, including cancel, termination and timeout, should have a non-nil closeTime and executionDuration. 
We need to prevent any mismatches in request vs Token namespace.

## How did you test it?
- [X] built
- [] added new unit test(s)


bergundy (approved):
Looks great, I skimmed most of the tests but the ones I read LGTM.

PR #8886, opened 2025-12-20
Add TemporalUsedWorkerDeploymentVersions SA

ShahabT:
## What changed?
Add `TemporalUsedWorkerDeploymentVersions` search attribute to contain all the Deployment Versions that completed at least one workflow tasks for the execution.

This Search Attribute replaces, now deprecated, `BuildIds` Search attribute.

## Why?
Needed a replacement for `BuildIds` Search attribute which holds internal values and prefixes.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
None.

### DIFF
--- a/service/history/workflow/mutable_state_impl_test.go
+++ b/service/history/workflow/mutable_state_impl_test.go
@@ -3058,7 +3058,7 @@ func (s *mutableStateSuite) TestUpdateUsedDeploymentVersionsSearchAttribute() {
 	s.NoError(err)
 	// Should have removed oldest entries to fit within size limit
 	versions := s.getUsedDeploymentVersionsFromMutableState()
-	s.Positive(len(versions), "Should have at least one version")
+	s.NotEmpty(versions, "Should have at least one version")
 	// The newest one should be present
 	s.Contains(versions, "deployment-c:build-1")
 
--- a/tests/versioning_3_test.go
+++ b/tests/versioning_3_test.go
@@ -765,7 +765,7 @@ func (s *Versioning3Suite) TestSearchByUsedVersion() {
 				Query:     fmt.Sprintf("TemporalUsedWorkerDeploymentVersions = '%s'", tv.DeploymentVersionStringV32()),
 			})
 			s.NoError(err)
-			s.Equal(0, len(wfs.Executions))
+			s.Empty(wfs.Executions)
 			return respondCompleteWorkflow(tv, vbPinned), nil
 		})
 
@@ -782,7 +782,7 @@ func (s *Versioning3Suite) TestSearchByUsedVersion() {
 			Query:     fmt.Sprintf("TemporalUsedWorkerDeploymentVersions = '%s'", tv.DeploymentVersionStringV32()),
 		})
 		a.NoError(err)
-		a.Equal(1, len(wfs.Executions))
+		a.Len(wfs.Executions, 1)
 	}, 5*time.Second, 200*time.Millisecond)
 }
 


common/searchattribute/sadefs/constants.go
```diff
@@ -153,20 +158,21 @@ var (
...
-		TemporalSchedulePaused:             enumspb.INDEXED_VALUE_TYPE_BOOL,
-		TemporalNamespaceDivision:          enumspb.INDEXED_VALUE_TYPE_KEYWORD,
-		TemporalPauseInfo:                  enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
-		TemporalReportedProblems:           enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
-		TemporalWorkerDeploymentVersion:    enumspb.INDEXED_VALUE_TYPE_KEYWORD,
-		TemporalWorkflowVersioningBehavior: enumspb.INDEXED_VALUE_TYPE_KEYWORD,
-		TemporalWorkerDeployment:           enumspb.INDEXED_VALUE_TYPE_KEYWORD,
+		TemporalChangeVersion:                enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
+		BinaryChecksums:                      enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
+		BuildIds:                             enumspb.INDEXED_VALUE_TYPE_KEYWORD_LIST,
```
carlydf, line 167: Do we want to comment `BuildIds` as deprecated at this time, or are you saving that for later?

service/history/workflow/mutable_state_impl.go
```diff
@@ -2889,7 +2889,8 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
 		ms.GetEffectiveVersioningBehavior() == enumspb.VERSIONING_BEHAVIOR_PINNED {
 		// TODO: [cleanup-old-wv]
 		limit := ms.config.SearchAttributesSizeOfValueLimit(string(ms.namespaceEntry.Name()))
-		if _, err := ms.addBuildIdAndDeploymentInfoToSearchAttributesWithNoVisibilityTask(event.SourceVersionStamp, limit); err != nil {
+		// Passing nil for usedVersion because starting with pinned override does not add the version to used versions SA until the version is actally used.
+		if _, err := ms.addBuildIdAndDeploymentInfoToSearchAttributesWithNoVisibilityTask(event.SourceVersionStamp, nil, limit); err != nil {
```
carlydf: ok, so we update the SAs for version, behavior, and deployment name on task started (and on successful set-override), but only update used-versions on task completed? 

makes sense

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -1265,7 +1265,7 @@ func (m *workflowTaskStateMachine) afterAddWorkflowTaskCompletedEvent(
 	// search attribute here because the reachability deployment may have just been changed by CompleteDeploymentTransition.
 	// This is also useful for unversioned workers.
 	// For v1 and v2 versioned workflows the search attributes should be already up-to-date based on the task started events.
-	if err := m.ms.updateBuildIdsAndDeploymentSearchAttributes(attrs.GetWorkerVersion(), limits.MaxSearchAttributeValueSize); err != nil {
+	if err := m.ms.updateBuildIdsAndDeploymentSearchAttributes(attrs.GetWorkerVersion(), worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(wftDeployment), limits.MaxSearchAttributeValueSize); err != nil {
```
carlydf, line 1315: I wish we could just pass `attrs.GetDeploymentVersion()` instead of `wftDeployment` here, but I suppose `wftDeployment` is correct for older SDKs that are sending deprecated `attrs.GetDeployment()` or deprecated `attrs.GetWorkerDeploymentVersion()` values?

As a follow-up: Should we be passing a form of `wftDeployment` as the first parameter instead of `attrs.GetWorkerVersion()`?

ShahabT: yeah I thought about that too. but I decided against using deprecated types for new args. but I think we're close to getting rid of all the v0.3.0-0.3.1 types. in next Go SDK and CLI we should completely get rid of them (by GA) and then we can remove them from server.

common/searchattribute/sadefs/constants.go
ShahabT: good point. will do. for all intents and purposes BuildIds will be considered deprecated after this change.

service/history/workflow/mutable_state_impl.go
```diff
@@ -2897,16 +2897,30 @@ func (ms *MutableStateImpl) ApplyWorkflowExecutionStartedEvent(
...
+	// Populate the versioningInfo if the inheritedAutoUpgradeInfo is present.
+	if event.GetInheritedAutoUpgradeInfo() != nil {
+		ms.SetVersioningRevisionNumber(event.GetInheritedAutoUpgradeInfo().GetSourceDeploymentRevisionNumber())
+		// TODO (Shivam): Remove this once you make SetDeploymentVersion and SetVersioningBehavior methods with nil checks
+		if ms.executionInfo.VersioningInfo == nil {
+			ms.executionInfo.VersioningInfo = &workflowpb.WorkflowExecutionVersioningInfo{}
+		}
+		ms.executionInfo.VersioningInfo.DeploymentVersion = event.GetInheritedAutoUpgradeInfo().GetSourceDeploymentVersion()
+		// Assume AutoUpgrade behavior for the first workflow task.
+		ms.executionInfo.VersioningInfo.Behavior = enumspb.VERSIONING_BEHAVIOR_AUTO_UPGRADE
```
Shivs11, line 2909: i'm slightly confused - any specific reason of moving this to the top compared to where it was before?

service/history/workflow/mutable_state_impl.go
Shivs11, line 3424: nit: could update comment here because we are now adding a new SA

service/history/workflow/mutable_state_impl.go
```diff
@@ -3572,42 +3628,91 @@ func (ms *MutableStateImpl) saveBuildIds(buildIds []string, maxSearchAttributeVa
...
+	if deployment == "" {
+		saPayloads[sadefs.TemporalWorkerDeployment] = nil
+	} else {
+		deploymentPayload, err := searchattribute.EncodeValue(deployment, enumspb.INDEXED_VALUE_TYPE_KEYWORD)
+		if err != nil {
+			return err
+		}
+		if len(deploymentPayload.GetData()) <= maxSearchAttributeValueSize { // we know the string won't really be over, but still check
+			saPayloads[sadefs.TemporalWorkerDeployment] = deploymentPayload
+		}
```
Shivs11, line 3673: nit: a lot of the code here seems to be following the pattern of setting the SA value to be nil if a passed arg is nil. If it's not, you encode it with the INDEXED_VALUE_TYPE_KEYWORD. 

Could def make a helper function here for clarity

PR #8893, opened 2025-12-22
[Flake] Fix TestCHASMCanListV1Schedules

lina-temporal:
## What changed?
- Schedule is paused in the flaky test before any visibility requests are made.

## Why?
- Otherwise, the schedule may take long enough to have advanced in FutureActionTimes


tests/schedule_test.go
```diff
@@ -1,4 +1,4 @@
-package tests
+ackage tests
```
yycptt: typo

PR #8894, opened 2025-12-22
VersionMembershipCache: Metrics and refactorings!

Shivs11:
## What changed?
- Added metrics, such as cache hits and cache misses, so that we can understand if the currently set TTL for this cache (of 1 second) is too low or too high.
- Also did some re-factorings: While working on this, I realized that it was much simpler to add a new wrapper with a metrics handler, that specifically served the use case of understanding cache hits and missed,  rather than use any of the existing implementations of caches that have a metrics handler attached to them. (See *NewWithMetrics*)
- Thus, I took some inspiration from the way *newEventsCache* was implemented and came up with this.

## Why?
- Explained above.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None.

---

> [!NOTE]
> Introduces a typed, instrumented cache for worker versioning and refactors call sites to use it.
> 
> - Adds `VersionMembershipCache` interface and `NewVersionMembershipCache` wrapper emitting metrics (`VersionMembershipCacheGet`/`Put` with `cache_type=version_membership`)
> - Replaces direct `cache.Cache` usage and ad-hoc keys in `worker_versioning` validation with the new cache API
> - Wires the cache via FX provider in history service, returning the wrapped cache; plumbs through engine, starters, and APIs (`startworkflow`, `signalwithstartworkflow`, `resetworkflow`, `multioperation`, `updateworkflowoptions`)
> - Updates metrics definitions with new cache type tag and operation scopes
> - Test updates: introduce simple/noop implementations and adjust existing tests to the new interface
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 920106f94048a6a7840afe2d09bdb2b02bc60e6e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/chasm/context.go
+++ b/chasm/context.go
@@ -17,7 +17,9 @@ type Context interface {
 	Now(Component) time.Time
 	// ExecutionKey returns the execution key for the execution the context is operating on.
 	ExecutionKey() ExecutionKey
-
+	// ExecutionCloseTime returns the time when the execution was closed. An execution is closed when its root component reaches a terminal
+	// state in its lifecycle. If the component is still running (not yet closed), it returns a zero time.Time value.
+	ExecutionCloseTime() time.Time
 	// Intent() OperationIntent
 	// ComponentOptions(Component) []ComponentOption
 
@@ -109,6 +111,14 @@ func (c *immutableCtx) ExecutionKey() ExecutionKey {
 
 func (c *immutableCtx) getContext() context.Context {
 	return c.ctx
+}
+
+func (c *immutableCtx) ExecutionCloseTime() time.Time {
+	closeTime := c.root.backend.GetExecutionInfo().GetCloseTime()
+	if closeTime == nil {
+		return time.Time{}
+	}
+	return closeTime.AsTime()
 }
 
 // NewMutableContext creates a new MutableContext from an existing Context and root Node.
--- a/chasm/context_mock.go
+++ b/chasm/context_mock.go
@@ -8,9 +8,10 @@ import (
 
 // MockContext is a mock implementation of [Context].
 type MockContext struct {
-	HandleExecutionKey func() ExecutionKey
-	HandleNow          func(component Component) time.Time
-	HandleRef          func(component Component) ([]byte, error)
+	HandleExecutionKey       func() ExecutionKey
+	HandleNow                func(component Component) time.Time
+	HandleRef                func(component Component) ([]byte, error)
+	HandleExecutionCloseTime func() time.Time
 }
 
 func (c *MockContext) getContext() context.Context {
@@ -40,6 +41,13 @@ func (c *MockContext) ExecutionKey() ExecutionKey {
 		return c.HandleExecutionKey()
 	}
 	return ExecutionKey{}
+}
+
+func (c *MockContext) ExecutionCloseTime() time.Time {
+	if c.HandleExecutionCloseTime != nil {
+		return c.HandleExecutionCloseTime()
+	}
+	return time.Time{}
 }
 
 // MockMutableContext is a mock implementation of [MutableContext] that records added tasks for inspection in
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -998,13 +998,6 @@ so forwarding by endpoint ID will not work out of the box.`,
 		`FrontendEnableUpdateWorkflowExecution enables UpdateWorkflowExecution API in the frontend.`,
 	)
 
-	FrontendEnableExecuteMultiOperation = NewNamespaceBoolSetting(
-		"frontend.enableExecuteMultiOperation",
-		true,
-		`FrontendEnableExecuteMultiOperation enables the ExecuteMultiOperation API in the frontend.
-The API is under active development.`,
-	)
-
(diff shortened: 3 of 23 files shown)


common/worker_versioning/worker_versioning_test.go
```diff
@@ -39,6 +39,63 @@ var (
...
+	taskQueue      string
+	taskQueueType  enumspb.TaskQueueType
+	deploymentName string
+	buildID        string
+}
+
+type testVersionMembershipCache struct {
+	mu sync.Mutex
+	m  map[testVersionMembershipCacheKey]bool
+}
```
Shivs11, line 45: I was actually using an instance of cache.Cache in some of the unit tests in this file. I thought I might as well change that by using this newly implemented cache since it shall also then test it's functionality.

service/history/version_membership_cache.go
```diff
@@ -0,0 +1,80 @@
...
+type versionMembershipCacheKey struct {
+	namespaceID    string
+	taskQueue      string
+	taskQueueType  enumspb.TaskQueueType
+	deploymentName string
+	buildID        string
+}
+
+func newVersionMembershipCache(c cache.Cache, metricsHandler metrics.Handler) worker_versioning.VersionMembershipCache {
+	h := metricsHandler.WithTags(metrics.CacheTypeTag("version_membership"))
```
carlydf: I think `"version_membership"` should be a constant in the `metrics` package, as in https://github.com/temporalio/temporal/blob/91893f10641bae49f8181235c5e93f61a3c826a8/service/history/events/cache.go#L87

service/history/version_membership_cache.go
```diff
@@ -0,0 +1,80 @@
...
+	return &versionMembershipCache{
+		cache:   c,
```
carlydf: it's concerning to me that I don't see eviction happening anywhere. But maybe I've just missed it. In the events cache, the underlying cache is the LRU cache, which makes sense to me. That also provides the hit and miss metrics that you need, in a way that is already standardized. Is it possible to just use the existing LRU cache instead of having to reimplement and re-test eviction logic elsewhere?

common/worker_versioning/version_membership_cache.go
```diff
@@ -0,0 +1,28 @@
+//nolint:staticcheck
+package worker_versioning
+
+import enumspb "go.temporal.io/api/enums/v1"
+
+// VersionMembershipCache is used to cache results of Matching's CheckTaskQueueVersionMembership
+// calls (used internally by the worker versioning pinned override validation).
+//
+// Implementations are expected to be safe for concurrent use.
+type VersionMembershipCache interface {
```
carlydf: It feels very clear to me how the cache key, the cache interface, and the cache implementation for events cache are defined right next to each other. Is it possible to do the same for the version membership cache and still achieve the testing that you want? https://github.com/temporalio/temporal/blob/91893f10641bae49f8181235c5e93f61a3c826a8/service/history/events/cache.go#L22-L48

service/history/version_membership_cache.go
Shivs11: This is how the cache here is being initialized (in the fx.go file):
```
func VersionMembershipCacheProvider(
	lc fx.Lifecycle,
	serviceConfig *configs.Config,
	metricsHandler metrics.Handler,
) worker_versioning.VersionMembershipCache {
	c := commoncache.New(serviceConfig.VersionMembershipCacheMaxSize(), &commoncache.Options{
		TTL: max(1*time.Second, serviceConfig.VersionMembershipCacheTTL()),
	})
	lc.Append(fx.Hook{
		OnStop: func(context.Context) error {
			c.Stop()
			return nil
		},
	})
	return newVersionMembershipCache(c, metricsHandler)
}
```

The underlying cache here is a `StoppableCache`(exactly similar to the events cache), which is also an LRU cache. Since the versioning cache is an LRU cache, the eviction logic exists in the `lru.go` file and the tests that I had added in my previous PR validate this.

The underlying  `StoppableCache` does not emit cache hit and cache miss metrics (which is what we are interested in), which is why I had defined this new wrapper on top of it. This also seemed to be one of the main reasons why the events cache was implemented as a layer on top of the `StoppableCache` .

prathyushpv: Actually LRU cache can emit metrics. There is this constructor which will create a cache with a metrics handler: https://github.com/temporalio/temporal/blob/abb63592042a1c347c1314326ea2b4143f5823cf/common/cache/lru.go#L143-L143
Oh, I see that it doesn't emit hit and miss metrics :/

carlydf: ty for showing me where the TTL is set! my main concern is fine then.
I thought that the LRU cache / stoppable would have metrics to provide the hit rate, but honestly I'm not sure what some of the metrics it emits even mean, or how to compute hit rate from them:
- `NewGaugeDef("cache_pinned_usage") // I looked this up and it's the count of elements that are blocked from being evicted even if they are the LRU element`
- `NewTimerDef("cache_entry_age_on_eviction")`
- `NewGaugeDef("cache_usage")`
- `NewTimerDef("cache_entry_age_on_get")`

carlydf (approved):

PR #8896, opened 2025-12-23
Add CHASM context CloseTime API

fretz12:
## What changed?
Added API to chasm ctx to get the close time of a component

## Why?
Chasm library callers need a way to get the time the component transitions to a terminal state. This avoids each component managing this separately.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/chasm/context.go
+++ b/chasm/context.go
@@ -17,16 +17,14 @@ type Context interface {
 	Now(Component) time.Time
 	// ExecutionKey returns the execution key for the execution the context is operating on.
 	ExecutionKey() ExecutionKey
-
+	// ExecutionCloseTime returns the time when the execution was closed. An execution is closed when its root component reaches a terminal
+	// state in its lifecycle. If the component is still running (not yet closed), it returns a zero time.Time value.
+	ExecutionCloseTime() time.Time
 	// Intent() OperationIntent
 	// ComponentOptions(Component) []ComponentOption
 
 	structuredRef(Component) (ComponentRef, error)
 	getContext() context.Context
-
-	// CloseTime returns the time when the execution was closed. An execution is closed when its root component reaches a terminal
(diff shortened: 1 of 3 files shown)


chasm/context_mock.go
```diff
@@ -42,6 +42,10 @@ func (c *MockContext) ExecutionKey() ExecutionKey {
 	return ExecutionKey{}
 }
 
+func (c *MockContext) CloseTime() time.Time {
+	return time.Time{}
```
yycptt, line 50: nit: plz add a HandleCloseTime() mock

chasm/context.go
```diff
@@ -23,6 +23,10 @@ type Context interface {
 
 	structuredRef(Component) (ComponentRef, error)
 	getContext() context.Context
+
+	// CloseTime returns the time when the root component was closed. A component is closed when it reaches a terminal
+	// state in its lifecycle. If the component is still running (not yet closed), it returns a zero time.Time value.
+	CloseTime() time.Time
```
yycptt: nit: group with other exported methods.

chasm/context.go
```diff
@@ -23,6 +23,10 @@ type Context interface {
 
 	structuredRef(Component) (ComponentRef, error)
 	getContext() context.Context
+
+	// CloseTime returns the time when the root component was closed. A component is closed when it reaches a terminal
```
yycptt: nit

```suggestion
	// CloseTime returns the time when the execution was closed. An execution is closed when its root component reaches a terminal
```

chasm/context.go
yycptt: nit: I am thinking if we should call it ExecutionCloseTime()

chasm/context_mock.go
fretz12: done

chasm/context.go
fretz12: Moved, and renamed

PR #8898, opened 2025-12-23
CHASM: Remove component shardingFn option

yycptt:
## What changed?
- Remove component shardingFn option

## Why?
- With separate ID space per archetypes, we don't have an immediate use case right now for customizing shardingFn per archetype. 
- It's also not fully implemented right now and not weird into history client, so dropping to support to avoid confusing.
- If we need the feature later, we can revert this PR.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/chasm_engine.go
+++ b/service/history/chasm_engine.go
@@ -694,8 +694,8 @@ func (e *ChasmEngine) getShardContext(
 ) (historyi.ShardContext, error) {
 	return e.shardController.GetShardByID(
 		common.WorkflowIDToHistoryShard(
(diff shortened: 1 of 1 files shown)


common/util.go
```diff
@@ -383,20 +383,14 @@ func ErrorHash(err error) string {
 }
 
 // WorkflowIDToHistoryShard is used to map namespaceID-workflowID pair to a shardID.
+// TODO: rename to BusinessIDToHistoryShard.
```
yycptt, line 386: I will do the renaming in a separate PR as there are lots of usage on this function.

PR #8899, opened 2025-12-23
CHASM: Do a final staleness check after reload

yycptt:
## What changed?
- Do a final staleness check after reload

## Why?
- Due to data loss or force failover, it's possible that even after reload the mutable state may still be stale.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/chasm_engine.go
+++ b/service/history/chasm_engine.go
@@ -755,7 +755,7 @@ func (e *ChasmEngine) getExecutionLease(
 		return shardContext, executionLease, nil
 	}
 
-	// Mutable state was previously detected as stale and get reloaded,
+	// Mutable state was previously detected as stale and got reloaded,
 	// do a final check to ensure mutable state is not stale after reload.
 	err = executionLease.GetMutableState().ChasmTree().IsStale(ref)
 	if err != nil {


service/history/chasm_engine.go
```diff
@@ -353,16 +353,6 @@ func (e *ChasmEngine) predicateSatisfied(
...
-	// behind the requested reference. However, getExecutionLease does not currently guarantee that
-	// execution VT >= ref VT, therefore we call IsStale() again here and return any error (which at
-	// this point must be ErrStaleState; ErrStaleReference has already been eliminated).
-	err := chasmTree.IsStale(ref)
-	if err != nil {
-		// ErrStaleState
-		// TODO(saa-yichao): this should be retryable if it is the failover version that is stale
-		return nil, err
-	}
 	// We know now that execution VT >= ref VT
```
dandavison: nit: I think this comment is a bit out of place now -- let's get rid of it?

service/history/chasm_engine.go
```diff
@@ -751,10 +741,52 @@ func (e *ChasmEngine) getExecutionLease(
...
+		return nil, nil, err
+	}
+
+	if staleReferenceErr != nil {
+		executionLease.GetReleaseFn()(nil)
+		return nil, nil, staleReferenceErr
+	}
+
+	// Do a final check to ensure mutable state is not stale after reload.
+	err = executionLease.GetMutableState().ChasmTree().IsStale(ref)
```
dandavison, line 760: If there was no reload in `getWorkflowLease` then it is not necessary to do the staleness check a second time. For that reason, should we do this final check after the reload in `getWorkflowLease`?

service/history/chasm_engine.go
```diff
@@ -701,8 +691,6 @@ func (e *ChasmEngine) getShardContext(
 // the state transition specified by the component reference is consistent with mutable state being
 // stale, then mutable state is reloaded from persistence before returning. It does not check that
 // mutable state is non-stale after reload.
```
dandavison: The final sentence of the doc comment was out of date. I reworded the whole thing slightly:


```suggestion
// getExecutionLease returns shard context and mutable state for the chasm execution, with the lock
// held. An error is returned if the state transition specified by the component reference is
// inconsistent with mutable state transition history. If the state transition specified by the
// component reference is consistent with mutable state being stale, then mutable state is reloaded
// from persistence. The returned mutable state is guaranteed not to be stale with respect to the
// supplied component reference.
```

service/history/chasm_engine.go
yycptt: Removed.

service/history/chasm_engine.go
yycptt: yeah make sense. Added a check.

service/history/chasm_engine.go
yycptt: Thanks for catching this!

service/history/chasm_engine.go
```diff
@@ -751,10 +742,57 @@ func (e *ChasmEngine) getExecutionLease(
...
+	if predicateErr != nil {
+		executionLease.GetReleaseFn()(nil)
+		return nil, nil, predicateErr
+	}
+
+	if !needReload {
+		return shardContext, executionLease, nil
+	}
+
+	// Mutable state was previously detected as stale and get reloaded,
```
dandavison:
```suggestion
	// Mutable state was previously detected as stale and got reloaded,
```

PR #8495, opened 2025-10-16
Spell Checker

chaptersix:
## What changed?

* Add the `typos` cli tool to the linters in the Makfile and CI
* Added a config file to exclude generated files and other common false positive scenarios 
## Why?
Reduce the amount of misspellings while not being overly burdensome.

## How did you test it?
- [x] use an AI agent to propose file exclusion rules.
- [x] run locally and tested manually

## Potential risks
Too much time and effort being used on adding exclusion.

PR #8903, opened 2025-12-27
CHASM: Sync structure after every pure task

yycptt:
## What changed?
- Maintain a map from component & data value to their chasm tree Node
- Sync tree structure after processing every pure task

## Why?
- https://github.com/temporalio/temporal/pull/8400 fixed the bug where a pure task deletes another node with pure tasks. However, a node can also have multiple pure tasks and a pure task can delete the node itself via a pointer, in this case remaining pure tasks should not be executed. To fix that, we will need to perform tree structure sync after every pure task processing and be able to tell after syncing if a node/component still exists or not.

- With the value -> Node map we can easily tell if a component still exists or not. It can also simplify some existing implementations where we scan the entire tree to find the node for a Component: e.g. Ref(), component/DataValuePath(). 

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/tree_test.go
+++ b/chasm/tree_test.go
@@ -2073,41 +2073,73 @@ func (s *nodeSuite) TestCloseTransaction_InvalidateComponentTasks() {
 				},
 			},
 		},
+		"SubComponent1": {
+			Metadata: &persistencespb.ChasmNodeMetadata{
+				InitialVersionedTransition:    &persistencespb.VersionedTransition{TransitionCount: 1},
+				LastUpdateVersionedTransition: &persistencespb.VersionedTransition{TransitionCount: 1},
+				Attributes: &persistencespb.ChasmNodeMetadata_ComponentAttributes{
+					ComponentAttributes: &persistencespb.ChasmComponentAttributes{
+						TypeId: testSubComponent1TypeID,
+						SideEffectTasks: []*persistencespb.ChasmComponentAttributes_Task{
+							{
+								TypeId:                    testSideEffectTaskTypeID,
+								VersionedTransition:       &persistencespb.VersionedTransition{TransitionCount: 1},
+								VersionedTransitionOffset: 4,
+								Data:                      taskBlob,
+								PhysicalTaskStatus:        physicalTaskStatusCreated,
+							},
+						},
+						PureTasks: []*persistencespb.ChasmComponentAttributes_Task{
+							{
+								TypeId:                    testPureTaskTypeID,
+								VersionedTransition:       &persistencespb.VersionedTransition{TransitionCount: 1},
+								VersionedTransitionOffset: 5,
+								Data:                      taskBlob,
+								PhysicalTaskStatus:        physicalTaskStatusNone,
+							},
+						},
+					},
+				},
+			},
+		},
 	}
 	root, err := s.newTestTree(persistenceNodes)
 	s.NoError(err)
 
+	nextTransitionCount := int64(2)
+	s.nodeBackend.HandleNextTransitionCount = func() int64 { return nextTransitionCount }
+
 	// The idea is to mark the node as dirty by accessing it with a mutable context.
-	// Otherwise, CloseTransaction logic will skip validating tasks for this node.
-	// This is a no-op change right now, since we are manually setting the LastUpdateVersionedTransition
-	// for the node below.
-	// Once CloseTransaction is fully implemented, this will be mark the node as dirty,
-	// and CloseTransaction logic will take care of updating LastUpdateVersionedTransition.
 	mutableContext := NewMutableContext(context.Background(), root)
 	_, err = root.Component(mutableContext, ComponentRef{})
 	s.NoError(err)
 
-	// TODO: remove this when CloseTransaction is fully implemented.
-	root.serializedNode.Metadata.LastUpdateVersionedTransition = &persistencespb.VersionedTransition{
-		TransitionCount: 2,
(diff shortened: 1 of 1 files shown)


chasm/tree.go
```diff
@@ -1380,57 +1420,43 @@ func (n *Node) CloseTransaction() (NodesMutation, error) {
...
+				if syncStructure {
+					if err := n.syncSubComponents(); err != nil {
+						return err
+					}
+				}
 
-		if syncStructure {
-			if err := n.syncSubComponents(); err != nil {
-				return err
+				// The correspoinding Node may not be found due to several reasons:
```
lina-temporal:
```suggestion
				// The corresponding Node may not be found due to several reasons:
```

lina-temporal: @chaptersix  just FYI - seems like the spellcheck test should have picked this up? Maybe we need to update GHA?

chasm/tree.go
```diff
@@ -2446,112 +2473,96 @@ func isComponentTaskExpired(
...
+		for len(componentAttr.GetPureTasks()) != 0 &&
+			isComponentTaskExpired(referenceTime, componentAttr.GetPureTasks()[0]) {
 
-		executed, err := callback(n, taskAttributes, taskInstance)
-		if err != nil {
-			return false, err
-		}
-		taskExecuted = taskExecuted || executed
+			// Node get deleted when previous pure tasks of the same component are executed.
+			// e.g. via a (parent) pointer.
```
lina-temporal, line 2578: I don't quite follow here how this involves the parent pointer, wouldn't the node only be deleted if some pure task deleted the node's component?

chasm/tree_test.go
```diff
@@ -2767,15 +2780,30 @@ func (s *nodeSuite) TestEachPureTask() {
 			rootComponent, err := root.Component(mutableContext, ComponentRef{})
 			s.NoError(err)
 
-			// delete
-			rootComponent.(*TestComponent).SubComponent2 = Field[*TestSubComponent2]{}
+			rootComponent.(*TestComponent).SubComponent2 = NewEmptyField[*TestSubComponent2]()
+		}
+
+		// When processing task for subcomponent11, delete it's parent SubComponent1 to remaining task is not executed.
```
lina-temporal:
```suggestion
		// When processing task for SubComponent11, delete its parent SubComponent1 so that the remaining task is not executed.
```

chasm/tree.go
yycptt: looks like [the spell checker](https://github.com/temporalio/temporal/pull/8495) is not landed yet.

chasm/tree.go
yycptt: Without a pointer, I don't think a task executor is able to delete the node itself. 

The task executor is given the Component and we don't have any api to delete a component just by it's value (that will be something like `chasmContext.Remove(component)` probably). 
The only way I am aware of for removing a component is `parentComponent.Component = NewEmptyField()` which needs a parent pointer, or a pointer to an ancestor node

chasm/tree.go
```diff
@@ -2468,105 +2525,92 @@ func (n *Node) EachPureTask(
...
+			if executed {
+				if err := n.syncSubComponents(); err != nil {
+					return err
+				}
+			}
+
+			// Processed task should become invalid and will be removed upon CloseTransaction().
+
+			// TODO: Add a validation for that and return an internal error if tasks is still valid after processing.
+			// Alternatively, remove task from PureTasks slice after processing, but that requires persisting the
```
yycptt, line 2608: I realized that I can't simply remove thing from PureTasks here since the change may not be persisted. I've reverted that for now.

chasm/tree.go
```diff
@@ -1625,7 +1649,39 @@ func (n *Node) closeTransactionUpdateComponentTasks(
...
+		// - parent component closing (access rule)
+		// - a pointer field pointing to an updated component
+		// As a result, we need to validate existing tasks for all components if we are in active cluster.
+		if n.isActiveStateDirty {
+			// Ensure this node's component value is hydrated before cleaning up tasks.
+			if err := node.prepareComponentValue(taskValidationContext); err != nil {
+				return err
+			}
+
+			cleanedUp, err := node.closeTransactionCleanupInvalidTasks(taskValidationContext)
```
yycptt, line 1664: Since we are not removing invalid tasks in EachPureTask(), I think this change is necessary (though technically this can be done in a separate PR but included here as I am fixing pure task processing anyway). 

Basically we need to perform task validation for all tasks in the tree to make sure persisted tasks are always valid. 

Here's what can happen if we have a invalid persisted task:
Say at T = 0 we have a component C, which has a pure task T with the smallest scheduled timestamp across the tree. A physical pure task is generated for it, and it's taskStatus is set to created. 
Later at T = 1, some other node get updated (e.g. its parent P), and somehow invalidated the pure task (e.g. due to access rule). Without the change here, since C is not updated, the pure task T won't be deleted when P is updated.
At T = 2, the physical pure task for T is executed, it sees T is invalid, so doesn't do anything. Now as long as C is not updated, the T with taskStatus = create will always be there. Also it's T has the smallest timestamp, it will prevent any new physical pure task from being generated. Refresh task also doesn't help much here as the invalid T is still there.

With this change the pure task will be deleted at T = 1.

yycptt:
I will go ahead and land this first to catch the release. Let me know if there's any comments and I will address them in follow up PRs.

PR #8904, opened 2025-12-27
Fix race in TestUserData_FetchesStickyToNormal

dnr:
## What changed?
- Fix races in TestUserData_FetchesStickyToNormal. Ran 10000 times locally.
- Make a bunch of the userDataManager tests call Wait() to ensure goroutines are stopped before returning from the test, to avoid gomock races.

## Why?
Flaky tests

## How did you test it?
- [x] built

### DIFF
--- a/service/matching/user_data_manager.go
+++ b/service/matching/user_data_manager.go
@@ -151,8 +151,6 @@ func (m *userDataManagerImpl) WaitUntilInitialized(ctx context.Context) error {
 
 func (m *userDataManagerImpl) Stop() {
 	m.goroGroup.Cancel()
-	// ensure fetch/loadUserData are not running so we can safely call setUserDataState
-	m.goroGroup.Wait()
 	// Set user data state on stop to wake up anyone blocked on the user data changed channel.
 	m.setUserDataState(userDataClosed, nil)
 }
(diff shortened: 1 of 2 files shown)


service/matching/user_data_manager.go
```diff
@@ -151,6 +151,8 @@ func (m *userDataManagerImpl) WaitUntilInitialized(ctx context.Context) error {
 
 func (m *userDataManagerImpl) Stop() {
 	m.goroGroup.Cancel()
+	// ensure fetch/loadUserData are not running so we can safely call setUserDataState
+	m.goroGroup.Wait()
```
ShahabT: would this have to wait for the long poll to expire?

dnr: Yes, but it's canceling the context that the long poll is using, so it should return quickly.

The alternative to this is using SetIfNotReady in setUserDataState instead of Ready/Set. Maybe we should do that too. But if we don't do the Wait here, that unit test and maybe others will need to reach in and call `m.goroGroup.Wait()` themselves to avoid issues with stuff still happening after the test completes.

ShahabT: I think my concern is elevated with the context canceling bit.

dnr: Looking again.. 1. there's no reason not to use SetIfNotReady, I just changed that, 2. my comment about the race was not correct, the mutex protects the ready/set thing so we don't really need this here. Maybe it's better to leave it how it was and just fix the test? I'll just do that for now.

PR #8913, opened 2025-12-29
Add minimum-priority poll conditions

dnr:
## What changed?
Allow new option for matching poll request to specify a minimum priority task to match, and also to request not blocking. These are intended to be used together. These options are only on the internal poll rpcs, they can't be set externally.

## Why?
This will be used for priority backlog poll forwarding, to avoid matching with unintended lower-priority tasks. It could also be used eventually as a building block for activity task preemption.

## How did you test it?
- [ ] built
- [x] added new unit test(s)
- [x] added new functional test(s) (in future PRs, not this one)

## Potential risks
Using min priority without nowait could cause excessive cpu usage since the find-match algorithm isn't optimized for it yet. With nowait and backoff on empty response, it shouldn't be a problem.

### DIFF
--- a/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
@@ -30,6 +30,7 @@ message PollWorkflowTaskQueueRequest {
     string poller_id = 2;
     temporal.api.workflowservice.v1.PollWorkflowTaskQueueRequest poll_request = 3;
     string forwarded_source = 4;
+    // Extra conditions on this poll request. Only supported with new matcher.
     PollConditions conditions = 5;
 }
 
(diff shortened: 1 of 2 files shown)


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -646,3 +650,13 @@ message CheckTaskQueueVersionMembershipRequest {
 message CheckTaskQueueVersionMembershipResponse {
     bool is_member = 1;
 }
+
+message PollConditions {
```
ShahabT, line 657: so these are supported only in new matcher?

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -646,3 +650,13 @@ message CheckTaskQueueVersionMembershipRequest {
...
+    // If set (non-zero), this poll will not match a task with lower priority than this value.
+    // Note that "min" priority is "max" numeric value.
```
ShahabT: Maybe add the example because it's not immediately intuitive: 
```
Ex: when `min_priority=3` the returned task can have priority 1,2, or 3.
```

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
dnr: yes. I'll add a comment

PR #8915, opened 2025-12-30
CHASM: SQL separate ID spaces

yycptt:
## What changed?
- Support separate ID spaces for different archetypes in SQL persistence.

## Why?
- Separate ID space workstream

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/common/persistence/sql/sqlplugin/sqlite/conn_pool.go
+++ b/common/persistence/sql/sqlplugin/sqlite/conn_pool.go
@@ -34,7 +34,7 @@ func (cp *connPool) Allocate(
 	cfg *config.SQL,
 	resolver resolver.ServiceResolver,
 	logger log.Logger,
-	create func(cfg *config.SQL, resolver resolver.ServiceResolver, logger log.Logger) (*sqlx.DB, error),
+	create func(*config.SQL, resolver.ServiceResolver, log.Logger) (*sqlx.DB, error),
 ) (db *sqlx.DB, err error) {
 	cp.mu.Lock()
 	defer cp.mu.Unlock()


common/persistence/sql/sqlplugin/mysql/execution.go
```diff
@@ -270,8 +292,15 @@ func (mdb *db) UpdateCurrentExecutions(
 	ctx context.Context,
 	row *sqlplugin.CurrentExecutionsRow,
 ) (sql.Result, error) {
+	if row.ArchetypeID == chasm.WorkflowArchetypeID {
```
prathyushpv, line 304: Don't we also need to check for 0 here for backward compatibility?
like `archetypeID == chasm.WorkflowArchetypeID || archetypeID == chasm.UnspecifiedArchetypeID`

common/persistence/tests/execution_mutable_state.go
```diff
@@ -2371,6 +2372,170 @@ func (s *ExecutionMutableStateSuite) TestDelete_NotExists() {
 	s.AssertMissingFromDB(s.NamespaceID, s.WorkflowID, s.RunID, chasm.WorkflowArchetypeID)
 }
 
+func (s *ExecutionMutableStateSuite) TestArchetypeSeparateIDSpace() {
+	if strings.HasPrefix(s.T().Name(), "TestCassandra") {
+		s.T().Skip("Separate ID space is only implemented by sql persistence")
```
prathyushpv: Please remember to remove this after merging Cassandra changes.

common/persistence/sql/sqlplugin/mysql/execution.go
yycptt: no, not really. Even for existing executions, application layer will pass down `chasm.WorkflowArchetypeID`.

There's also a check in execution manager to guarantee that 0 is never used: https://github.com/temporalio/temporal/blob/main/common/persistence/execution_manager.go#L1186

common/persistence/tests/postgresql_test.go
```diff
@@ -386,7 +388,24 @@ func (p *PostgreSQLSuite) TestPostgreSQLHistoryCurrentExecutionSuite() {
...
+	suite.Run(p.T(), s)
+}
+
+func (p *PostgreSQLSuite) TestPostgreSQLHistoryCurrentChasmExecutionSuite() {
+	cfg := NewPostgreSQLConfig(p.pluginName)
+	SetupPostgreSQLDatabase(p.T(), cfg)
+	SetupPostgreSQLSchema(p.T(), cfg)
+	store, err := sql.NewSQLDB(sqlplugin.DbKindMain, cfg, resolver.NewNoopResolver(), log.NewTestLogger(), metrics.NoopMetricsHandler)
+	if err != nil {
+		p.T().Fatalf("unable to create MySQL DB: %v", err)
```
prathyushpv:
```suggestion
		p.T().Fatalf("unable to create PostgreSQL DB: %v", err)
```

common/persistence/sql/sqlplugin/mysql/execution.go
yycptt: I added an assert to verify that archetypeID is always specified.

PR #8918, opened 2025-12-30
CreateStream and cursor

spkane31:
^

### DIFF
--- a/chasm/lib/stream/proto/v1/stream_state.proto
+++ b/chasm/lib/stream/proto/v1/stream_state.proto
@@ -7,7 +7,4 @@ option go_package = "go.temporal.io/server/chasm/lib/stream/gen/streampb;streamp
 message StreamState {
     int64 head = 1;
     int64 tail = 2;
-    string namespace_id = 3;
-    string stream_id = 4;
-    string run_id = 5;
 }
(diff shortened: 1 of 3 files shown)


chasm/lib/stream/proto/v1/stream_state.proto
```diff
@@ -7,4 +7,7 @@ option go_package = "go.temporal.io/server/chasm/lib/stream/gen/streampb;streamp
 message StreamState {
     int64 head = 1;
     int64 tail = 2;
+    string namespace_id = 3;
+    string stream_id = 4;
+    string run_id = 5;
```
dandavison: I'm thinking we don't want these, right? My understanding is that the CHASM design intends that the execution coordinates shouldn't be stored with the application data.

chasm/lib/stream/stream.go
```diff
@@ -11,20 +11,72 @@ type Stream struct {
...
+		req.FrontendRequest.StreamId,
+		executionKey.RunID,
+		req.FrontendRequest.Messages,
+	)
+
+	return stream, &streampb.CreateStreamResponse{
+		FrontendResponse: &workflowservice.CreateStreamResponse{
+			RunId: stream.RunId,
+		},
+	}, nil
```
dandavison, line 60: Shall we just not pass `RunID` in this fuction; you're setting it afterwards.

PR #8920, opened 2025-12-30
CHASM: add panic wrapper for library logic

yycptt:
## What changed?
- Add panic wrapper for library logic
- Also provides a logger through the Context interface

## Why?
- Capture panic error so that it's easier to tell where the panic actually happens. Otherwise a panic can cause another panic in some unrelated deferred function making debugging harder.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/workflow/mutable_state_impl.go
+++ b/service/history/workflow/mutable_state_impl.go
@@ -287,9 +287,10 @@ func NewMutableState(
 	startTime time.Time,
 ) *MutableStateImpl {
 
+	namespaceName := namespaceEntry.Name().String()
 	logger = log.NewLazyLogger(logger, func() []tag.Tag {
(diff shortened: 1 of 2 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -8072,34 +8083,21 @@ func (ms *MutableStateImpl) createCallerError(
 	return serviceerror.NewInvalidArgument(msg)
 }
 
+// TODO: Deprecate following logging methods and use ms.logger directly.
 func (ms *MutableStateImpl) logInfo(msg string, tags ...tag.Tag) {
```
lina-temporal, line 8276: Nice cleanup on these!

PR #8921, opened 2025-12-31
CHASM: Check access rule for standby tasks

yycptt:
## What changed?
- Check access rule when validating tasks (standby logic)

## Why?
- We already check access rules before running a task in active cluster. Tasks in standby cluster (especially side effect tasks) should do the same, otherwise a physical task may never be invalidated and keep retrying forever (if user provided validator is not checking parent state).

## How did you test it?
- [x] built
- [x] added new unit test(s)


chasm/tree.go
```diff
@@ -1710,6 +1712,7 @@ func (n *Node) deserializeComponentTask(
 }
 
 // validateTask runs taskInstance's registered validation handler.
+// This method assumes component value is already hydrated.
 func (n *Node) validateTask(
```
yycptt, line 1716: I will do some follow up refactoring and always use a progress intent context here.

PR #8922, opened 2025-12-31
Gradual config rollouts

dnr:
## What changed?
Support gradual config rollouts by key (e.g. task queue partition).

## Why?
Rolling out changes slowly, and also mostly-synchronized across processes.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
Time-based changes can be risky if you set something for the future and forget about it. We should have validation that prevents changes from being set too far in the future or outside of business hours.

### DIFF
--- a/common/dynamicconfig/gradual_change.go
+++ b/common/dynamicconfig/gradual_change.go
@@ -3,7 +3,6 @@ package dynamicconfig
 import (
 	"math"
 	"reflect"
-	"strings"
 	"sync"
 	"time"
 
@@ -109,21 +108,8 @@ func ConvertGradualChange[T any](def T) func(v any) (GradualChange[T], error) {
 			return changeConverter(v)
 		}
 	default:
-		valueConverter := ConvertStructure(def)
-		return func(v any) (GradualChange[T], error) {
-			// both valueConverter and changeConverter will probably return success here, so
-			// we can't rely on the error. check if the input has "new" (the only required
-			// key in a GradualChange).
-			if vmap, ok := v.(map[string]any); ok {
-				for k := range vmap {
-					if strings.ToLower(k) == "new" {
-						return changeConverter(v)
-					}
-				}
-			}
-			nv, err := valueConverter(v)
-			return StaticGradualChange(nv), err
-		}
+		// nolint:forbidigo // this will be triggered from a static initializer before it can be triggered from production code
+		panic("ConvertGradualChange can only be used with scalar types for now")
 	}
 }
 
--- a/common/dynamicconfig/gradual_change_test.go
+++ b/common/dynamicconfig/gradual_change_test.go
@@ -215,38 +215,6 @@ func TestConvertGradualChange_DurationNumeric(t *testing.T) {
 	assert.Equal(t, 30*time.Second, gc.New)
 }
 
-func TestConvertGradualChange_StructPlainValue(t *testing.T) {
-	type config struct {
-		Enabled bool
-		Count   int
-	}
-	conv := ConvertGradualChange(config{})
-
-	gc, err := conv(map[string]any{"Enabled": true, "Count": 5})
-	require.NoError(t, err)
-	assert.Equal(t, config{Enabled: true, Count: 5}, gc.New)
-}
-
-func TestConvertGradualChange_StructGradualChange(t *testing.T) {
-	type config struct {
-		Enabled bool
-		Count   int
-	}
-	conv := ConvertGradualChange(config{})
-
-	gc, err := conv(map[string]any{
-		"new":   map[string]any{"enabLed": true, "Count": 10},
-		"old":   map[string]any{"Enabled": false, "counT": 0},
-		"Start": 1704067200,
-		"enD":   1704844800,
-	})
-	require.NoError(t, err)
-	assert.Equal(t, config{Enabled: true, Count: 10}, gc.New)
-	assert.Equal(t, config{Enabled: false, Count: 0}, gc.Old)
-	assert.True(t, gc.Start.Equal(time.Date(2024, 1, 1, 0, 0, 0, 0, time.UTC)))
-	assert.True(t, gc.End.Equal(time.Date(2024, 1, 10, 0, 0, 0, 0, time.UTC)))
-}
-
 func TestSubscribeGradualChange_InitialValues(t *testing.T) {
 	ts := clock.NewEventTimeSource()
 


service/matching/task_queue_partition_manager.go
```diff
@@ -115,18 +117,26 @@ func newTaskQueuePartitionManager(
...
 		pm.unloadFromEngine(unloadCauseConfigChange)
 	}
 
 	var fairness bool
-	fairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(unload)
+	changeKey := []byte(pm.partition.RoutingKey())
+	fairnessChangeChanged := makeChangeChangedFunc(changeKey, &fairness, unload)
+	fairnessChange, cancel1 := tqConfig.EnableFairnessSub(fairnessChangeChanged)
+	fairness = fairnessChange.Value(changeKey, time.Now())
+	pm.cancelFairnessSub = cancel1
```
ShahabT: I think it makes sense to extract this into a helper that can enable you saying the same thing as before, but giving it the changeKey too:

```
fairness, pm.cancelFairnessSub = tqConfig.EnableFairnessSub(changeKey, unload)
```

Doesn't have to be in this PR though.

ShahabT (commented):
nice feature! I did not review the implementation. but the concept lgtm

service/matching/task_queue_partition_manager.go
dnr: Yup, I definitely meant this as a "draft".. the api/helpers can be improved for sure

common/dynamicconfig/gradual_change.go
```diff
@@ -0,0 +1,210 @@
...
+// A setting with type GradualChange can use ConvertGradualChange as a conversion function to
+// handle plain values of the underlying type as well as GradualChange structs.
+type GradualChange[T any] struct {
+	Old, New   T
+	Start, End time.Time
+}
+
+// ConstantGradualChange returns a GradualChange whose Value always returns def and whose When
+// always returns a time in the past.
+func ConstantGradualChange[T any](def T) GradualChange[T] {
```
majiru: Small nit: This took me a second to realize that this was the default "non changing" variant, perhaps something like StaticGradualChange() or FixedGradualChange()?

common/dynamicconfig/gradual_change.go
```diff
@@ -0,0 +1,210 @@
...
+	w.cancelSub()
+	w.setTimerLocked(nil)
+}
+
+func (w *gradualChangeSubscribeWrapper[T]) reevalLocked() (T, bool) {
+	now := w.clock.Now()
+
+	var newTmr clock.Timer
+	if at := w.change.When(w.changeKey); at.After(now) {
+		newTmr = w.clock.AfterFunc(at.Sub(now), w.timerCallback)
```
majiru, line 181: Is it possible to hit some pretty deep recursion here? considering that the timerCallback calls back into reevalLocked. Not too familiar with the clock code here.

common/dynamicconfig/gradual_change.go
dnr: Right, "Constant" was supposed to be that :)  But "constant change" is misleading. Static seems a little better, yeah

common/dynamicconfig/gradual_change.go
dnr: It's just timer.AfterFunc, but fake-able. When the timer callback fires, it should re-evaluate and see that the change is in the past, so it doesn't need another timer. With some very contrived sequence of wall clock jumps it could end up with multiple callbacks in a row, but it should sort itself out quickly as long as time is mostly monotonic.

common/dynamicconfig/collection.go
```diff
@@ -710,6 +712,31 @@ func mapstructureHookDuration(f, t reflect.Type, data any) (any, error) {
...
+		return ts, nil
+	}
+	// treat numeric values as seconds
+	if ival, err := convertInt(data); err == nil {
+		return time.Unix(int64(ival), 0), nil
+	} else if fval, err := convertFloat(data); err == nil {
+		ipart, fpart := math.Modf(fval)
+		return time.Unix(int64(ipart), int64(fpart*float64(time.Second))), nil
+	}
+	return 0, errors.New("value not convertible to Time")
```
prathyushpv: `time.Time{}, errors.New("value not convertible to Time")` ?

common/tqid/task_queue_id.go
```diff
@@ -64,6 +64,8 @@ type (
 		Key() PartitionKey
 		// RoutingKey returns the string that should be used to find the owner of a task queue partition.
 		RoutingKey() string
+		// ChageKey returns an identifier that can be used with gradual changes.
```
prathyushpv: nit: ChangeKey

common/dynamicconfig/gradual_change.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Value returns the value for the given key at the given time.
+func (c *GradualChange[T]) Value(key []byte, now time.Time) T {
+	if !now.Before(c.End) {
+		return c.New
+	} else if !now.After(c.Start) {
+		return c.Old
+	}
+	fraction := float64(now.Sub(c.Start)) / float64(c.End.Sub(c.Start))
```
prathyushpv, line 38: Can someone configure both Start and End to be the same timestamp? That would produce 0 in the denominator.

service/matching/task_queue_partition_manager.go
```diff
@@ -120,13 +121,17 @@ func newTaskQueuePartitionManager(
...
+	fairness, pm.cancelFairnessSub = dynamicconfig.SubscribeGradualChange(
+		tqConfig.EnableFairnessSub, changeKey, unload, e.timeSource)
+
 	// Fairness is disabled for sticky queues for now so that we can still use TTLs.
 	tqConfig.EnableFairness = fairness && partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY
 	if fairness {
 		tqConfig.NewMatcher = true
 	} else {
-		tqConfig.NewMatcher, pm.cancelNewMatcherSub = tqConfig.NewMatcherSub(unload)
+		tqConfig.NewMatcher, pm.cancelNewMatcherSub = dynamicconfig.SubscribeGradualChange(
```
prathyushpv, line 133: Will we need to know if New value was returned for a namespace at a particular time when debugging an issue? Should we consider writing a log line when a value changes for a changeKey?

common/dynamicconfig/collection.go
dnr: oops, yeah, that's better

common/dynamicconfig/gradual_change.go
dnr: it's not possible for Start == End and to pass both conditions above, one or the other will be true. I added a test.

common/dynamicconfig/gradual_change.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	default:
+		valueConverter := ConvertStructure(def)
+		return func(v any) (GradualChange[T], error) {
+			// both valueConverter and changeConverter will probably return success here, so
+			// we can't rely on the error. check if the input has "new" (the only required
+			// key in a GradualChange).
+			if vmap, ok := v.(map[string]any); ok {
+				for k := range vmap {
+					if strings.ToLower(k) == "new" {
```
prathyushpv: Isn't this a bit fragile? What if T is another struct with a field `new`?
We can also use a marker field that is unlikely to collide with another field
```
Gradual    bool `mapstructure:"_gradual_"`
```

dnr: Yeah.. I had a similar idea but I wanted to get this in sooner and then improve it later, also with feedback from other teams. The initial uses are only bools so it doesn't matter. We could even disable its use for structs for now.

service/matching/task_queue_partition_manager.go
dnr: For this particular purpose, we'll see logs for unloading/reloading task queues. So I'm not sure we need another here. Plus getting a logger into the subscription stuff is annoying.

common/dynamicconfig/gradual_change.go
dnr: I just disabled it for structs. We can figure out some special key or something later.

PR #8925, opened 2026-01-02
Forward polls from sticky partitions to high-priority backlog

dnr:
## What changed?
- Add "ephemeral data" propagated alongside user data.
- Include information about which priority levels have significant backlog in ephemeral data (currently only propagated down the tree, not up)
- When sticky partitions see that normal partitions have significant backlog, set up poll forwarders to forward to those partitions.
- The poll forwarders use min priority to ensure they only get higher-priority tasks than available local tasks.

## Why?
Without this functionality, sticky queues interfere with priority dispatch by keeping most of the pollers working on sticky tasks and potentially starving high-priority tasks that appear on the normal queue. With the default values of the new settings, sticky pollers should notice high-priority normal tasks within 10 seconds.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
- Clients have to accept that polls on sticky may return normal tasks. I believe all SDKs are fine with this.
- Increased overhead of user data + ephemeral data.

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1249,6 +1249,11 @@ This can help reduce effects of task queue movement.`,
 		`How often to update ephemeral data (e.g. backlog size for forwarding sticky polls).
 Set to zero to disable ephemeral data updates.`,
 	)
+	MatchingPriorityBacklogForwarding = NewTaskQueueBoolSetting(
+		"matching.priorityBacklogForwarding",
+		true,
+		`Whether to forward polls to partitions with higher-priority backlog.`,
+	)
 	MatchingBacklogNegligibleAge = NewTaskQueueDurationSetting(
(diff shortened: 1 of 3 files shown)


service/matching/matcher_data.go
```diff
@@ -15,8 +15,16 @@ import (
 )
 
 const (
-	invalidHeapIndex      = -13     // use unusual value to stand out in panics
-	pollForwarderPriority = 1000000 // lower than any other priority. must be > maxPriorityLevels.
+	invalidHeapIndex        = -13 // use unusual value to stand out in panics
+	maxPriorityLevels       = 60  // maximum value for priority levels (fits in a bitfield with a few bits reserved)
+	effectivePriorityFactor = 10
```
majiru: I know we talked about this, but it might be nice for a very brief comment here.

PR #8927, opened 2026-01-02
jitter poller timeout

majiru:
Ease issues when we have large amounts of pollers all timing out together.

## What changed?
Add a jitter to our buffer for poller context timeouts

## Why?
we were running in to issues when we had larger numbers of pollers all timing out together, this should spread out that load a bit more.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


service/matching/matching_engine.go
```diff
@@ -2705,7 +2705,16 @@ func (e *matchingEngineImpl) pollTask(
 	// reached, instead of emptyTask, context timeout error is returned to the frontend by the rpc stack,
 	// which counts against our SLO. By shortening the timeout by a very small amount, the emptyTask can be
 	// returned to the handler before a context timeout error is generated.
-	ctx, cancel := contextutil.WithDeadlineBuffer(ctx, pm.LongPollExpirationInterval(), returnEmptyTaskTimeBudget)
+	// We also want to jitter the timeout a bit to ease stampeding hurd issues when the pollers all timeout
+	// together. However take caution to only do this for non forwarded tasks.
```
dnr: do you mean "non-forwarded _polls_"? also is the condition backwards? if `pollMetadata.forwardedFrom == ""` that's a poll from a client, so it should jitter, otherwise it shouldn't?

also I want to go over the logic again and test it.. most/all sdks actually set the poll grpc timeout to 70s, and LongPollExpirationInterval is 60s, so actually I think a "10s" buffer is a no-op, based on the current logic in WithDeadlineBuffer. I think we need to subtract it from LongPollExpirationInterval. see the comments in slack.

service/matching/matching_engine.go
```diff
@@ -2845,7 +2852,13 @@ func (e *matchingEngineImpl) pollTask(
 		return nil, false, errNoTasks
 	}
 
-	ctx, cancel := contextutil.WithDeadlineBuffer(ctx, pm.LongPollExpirationInterval(), returnEmptyTaskTimeBudget)
+	// For non-forwarded polls with a long enough interval, subtract a random jitter to spread
+	// expiration times across pollers and prevent thundering herd reconnects.
+	longPollInterval := pm.LongPollExpirationInterval()
+	if pollMetadata.forwardedFrom == "" && longPollInterval >= forwardedPollMinInterval {
+		longPollInterval -= time.Duration(rand.Int63n(int64(forwardedPollJitterMax)))
```
dnr:
```suggestion
		longPollInterval -= backoff.FullJitter(forwardedPollJitterMax)
```

service/matching/matching_engine_test.go
```diff
@@ -278,6 +278,86 @@ func (s *matchingEngineSuite) newPartitionManager(prtn tqid.Partition, config *C
...
 
+// captureNPollDeadlines injects a mock partition manager, runs n sequential polls, and returns
+// the context deadline observed inside each pm.PollTask call.
+func (s *matchingEngineSuite) captureNPollDeadlines(
+	partition tqid.Partition,
+	longPollInterval time.Duration,
+	meta *pollMetadata,
+	n int,
+) []time.Time {
+	ctrl := gomock.NewController(s.T())
```
dnr: I would create the gomock controller in the Test method and pass it in, in case you want to call this multiple times or something.

or actually wait, this is a method on the Suite? doesn't it have a controller already?

service/matching/matching_engine_test.go
```diff
@@ -278,6 +278,86 @@ func (s *matchingEngineSuite) newPartitionManager(prtn tqid.Partition, config *C
...
+// TestNonForwardedPollsHaveSpreadExpirations verifies that non-forwarded polls receive different
+// expiration times, which is the mechanism that prevents a thundering herd of simultaneous worker
+// reconnects when a wave of long polls all expire at the same instant.
+func (s *matchingEngineSuite) TestNonForwardedPollsHaveSpreadExpirations() {
+	const n = 20
+	partition := newRootPartition(s.ns.ID().String(), "spread-tq", enumspb.TASK_QUEUE_TYPE_WORKFLOW)
+	deadlines := s.captureNPollDeadlines(partition, 60*time.Second, &pollMetadata{}, n)
+
+	spread := deadlineSpread(deadlines)
+	s.Greater(spread, time.Duration(0), "non-forwarded polls must have varied expirations to prevent thundering herd")
```
dnr: I think > 0 will happen just by chance? how about > 5s?

service/matching/matching_engine.go
```diff
@@ -72,6 +72,13 @@ const (
 	// This seems aggressive, but the default sticky schedule_to_start timeout is 5s, so 10s seems reasonable.
 	stickyPollerUnavailableWindow = 10 * time.Second
 
+	// Maximum jitter subtracted from the long poll timeout for non-forwarded polls.
+	// Spreads out expiration times across pollers to prevent thundering herd reconnects.
+	forwardedPollJitterMax = 10 * time.Second
+	// Minimum long poll interval required to apply jitter. Ensures the timeout remains
+	// meaningful even after the maximum jitter is subtracted.
+	forwardedPollMinInterval = 50 * time.Second
```
dnr: should we derive this, like `forwardedPollMinInterval := longPollInterval - forwardedPollJitterMax`? or do we actually want a constant?

thinking: if we adjust LongPollExpirationInterval to 30s for some reason (we've done this before): we would still want to jitter between 20s and 30s. but if we reduce it to 10s, we don't want to jitter down to 0. so okay, either the jitter should be a proportion, or this should be a constant.

dnr (approved):
couple comments but looks good

service/matching/matching_engine.go
majiru: I've done a pass at the proportional logic here as a different commit from the other fixes at the end, maybe take a look and see if you like that better or worse than doing it as we were with a constant.

service/matching/matching_engine.go
```diff
@@ -72,6 +72,15 @@ const (
...
 	stickyPollerUnavailableWindow = 10 * time.Second
 
+	// Fraction of the long poll interval used as the maximum jitter for non-forwarded polls.
+	// Approximately matches the original 10s jitter on the default 60s interval (1/6 ≈ 16.7%).
+	// Spreads out expiration times across pollers to prevent thundering herd reconnects.
+	forwardedPollJitterRatio = 1.0 / 6
+	// Floor for the long poll interval after jitter is applied. Jitter is capped so that
+	// the interval never drops below this value; if the interval is already at or below
+	// this floor, no jitter is applied.
+	forwardedPollMinInterval = 20 * time.Second
```
dnr: maybe could use `common.CriticalLongPollTimeout` if you don't want to have another random constant here, but doesn't really matter

dnr (approved):
nice, I like this version

PR #8929, opened 2026-01-04
Fix Docker platform mismatch warning in multi-arch builds

chaptersix:
Add --platform=$TARGETPLATFORM to FROM instructions in both Dockerfiles
Fixes InvalidBaseImagePlatform warning during multi-arch builds where buildx was pulling linux/386 instead of the expected linux/amd64
Ensures correct platform manifest is pulled for each target architecture


chaptersix:
claude commit seems to be causing the license check to fail

PR #8932, opened 2026-01-05
Fix/scheduler trigger immediately visibility

### LINKED ISSUE #8833: Scheduler does not list workflows runned when using "triggerImmediately"
## Expected Behavior
Temporal's `client.schedule.getHandle(scheduleId).describe()` should return created running workflows in `info.runningActions` even when using `triggerImmediately`

## Actual Behavior
`info.runningActions` is empty when using the option `triggerImmediately`, the workflow is correctly created, but does not seems to be linked to the scheduler.

## Steps to Reproduce the Problem
Here's a reproduction code:
  ```typescript
await client.schedule.create({
        ...,
        state: {
          triggerImmediately: true
        }
});

// Wait for the schedule to be triggered and the workflow to run
await sleep(5000);

// Test the running workflows, runningActions should not be empty, but it is!
console.log((await client.schedule.getHandle(scheduleId).describe()).info.runningActions);
```

By triggering manually, it works fine:
  ```typescript
const handler = await client.schedule.create({
        ...,
});

await handler.trigger();

// Wait for the schedule to be triggered and the workflow to run
await sleep(5000);

// This time, runningActions is populated correctly
console.log((await client.schedule.getHandle(scheduleId).describe()).info.runningActions);
```

## Specifications

  - Version: 1.9.0
  - Platform: Kubernetes

DEVMANISHOFFL:
## What changed?
I updated the `recordAction` method in the scheduler workflow to accept a `manual` boolean flag. This flag is now passed down from `processBuffer`.

I changed the logic inside `recordAction` to guarantee that if a workflow start is `manual` (for example, through `TriggerImmediately` or Backfill), it is always added to the `RunningWorkflows` list. Before, these were often filtered out if they were seen as "overlapping," especially under `ALLOW_ALL` policies.

I also added a reproduction test case, `TestTriggerImmediatelyVisibility`, in `workflow_test`.go.

## Why?
This fixes #8833.

Users reported that workflows triggered with `triggerImmediately`: true did not show up in the scheduler's `info.runningActions` list. This occurred because the scheduler treated them as "overlapping" executions. These are hidden by default to prevent state bloat. This optimization was introduced in version 3.

Manual triggers are clear user actions and should always be visible in the API response and UI, no matter the schedule's overlap policy.

## How did you test it?
- [x] built
- [ ] added new unit test(s) 
- [x] added new functional test(s) (`TestTriggerImmediatelyVisibility`)

### DIFF
--- a/service/worker/scheduler/workflow.go
+++ b/service/worker/scheduler/workflow.go
@@ -1281,7 +1281,7 @@ func (s *scheduler) processBuffer() bool {
 		}
 		metricsWithTag.Counter(metrics.ScheduleActionSuccess.Name()).Inc(1)
 		nonOverlapping := start == action.NonOverlappingStart
-		s.recordAction(result, nonOverlapping)
+		s.recordAction(result, nonOverlapping, start.Manual)
 	}
 
 	// Terminate or cancel if required (terminate overrides cancel if both are present)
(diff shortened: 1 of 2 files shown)


lina-temporal:
Hey Manish,
Thanks for the fix! Because it changes something that makes up the scheduler's info block, it looks like it isn't replay-compatible without a version bump (and similarly, current unit tests are broken). You would have to put this behind a new scheduler version before `TestReplay` will pass, and accordingly, you'll need to update the unit tests around `RecentActions` to account for the fix.

gow:
@DEVMANISHOFFL since there has been no recent activity, we are closing this PR and the associated task. Please feel free to reopen the task and the PR if needed.

PR #8934, opened 2026-01-05
fix(frontend): return error when creating duplicate search attributes

### LINKED ISSUE #8631: Trying to create search attribute that already exists should return error
How issue was found:
Try adding same search attribute multiple times via cli - cli always gives "ok" response, for example try running this command multiple times:

          temporal operator search-attribute create --name MyCustomSA --type Keyword
          Search attributes have been added

When adding SA server has this loop where it creates aliasToFieldMap - https://github.com/temporalio/temporal/blob/4c2403736409e6f38972df3be4de480026a5b6fa/service/frontend/operator_handler.go#L279

In this loop server logs the warn message with errSearchAttributeAlreadyExistsMessage
but because it does not add anything to map it just returns nil - https://github.com/temporalio/temporal/blob/4c2403736409e6f38972df3be4de480026a5b6fa/service/frontend/operator_handler.go#L334

instead of serviceerror
this then in turn return nil to cli as response. cli does check for already exists error, but since server returns nil, error is never shown to user.

Ask is to change this code on server to return serviceError instead that tells cli/sdk/clients that they are trying to create a SA that already exists

DEVMANISHOFFL:
### What's being changed:

This PR changes the behavior of `AddSearchAttributes` to return a clear error when trying to add a search attribute that already exists (Fixes #8631).

**The Issue:**
In the past, the server would log a warning and return `nil` (success) when a duplicate was found. This led clients, such as the CLI, to report "Search attributes have been added" even though no changes were made.

**The Fix:**
I updated `addSearchAttributesElasticsearch` and `addSearchAttributesSQL` to return `serviceerror.NewAlreadyExists` right away when a duplicate alias or field is detected.

**Test Coverage:**
I modified `Test_AddSearchAttributesElasticsearch` and `Test_AddSearchAttributesSQL` in `operator_handler_test.go` to expect this new error instead of the earlier success-with-warning behavior.

### Review checklist

- [ ] Documentation has been updated, if necessary. Link to changed documentation:
- [ ] Chaos pipeline run or not necessary. Link to pipeline:
- [x] All new code is covered by tests where it is reasonable.
- [ ] Performance tests have been run or not necessary.

### DIFF
--- a/service/frontend/operator_handler.go
+++ b/service/frontend/operator_handler.go
@@ -21,7 +21,6 @@ import (
 	"go.temporal.io/server/common"
 	clustermetadata "go.temporal.io/server/common/cluster"
 	"go.temporal.io/server/common/log"
-	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/namespace"
 	"go.temporal.io/server/common/persistence"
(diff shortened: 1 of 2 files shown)


bergundy:
Thanks for the contribution @DEVMANISHOFFL, see my comments in the original issue.

rodrigozhou (changes requested):
@DEVMANISHOFFL Also commented in the issue. This PR not only breaks backwards compatibility, but also blocks dual visibility from working properly.

PR #8936, opened 2026-01-05
Refactor worker-build-id metrics tag.

ShahabT:
## What changed?
- Rename the tag from `worker-build-id` -> `worker_version`.
Use `:` instead of `/` in the tag value.

## Why?
Consistent terminology and delimiter.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Self hosted clusters relying on the old tag name and values would have to adopt new tag. This should be ok because versioning is not GA yet.

### DIFF
--- a/service/matching/matching_engine_test.go
+++ b/service/matching/matching_engine_test.go
@@ -1142,7 +1142,7 @@ func (s *matchingEngineSuite) TestSyncMatchActivities() {
 		assert.EqualValues(collect, 0, s.taskManager.getTaskCount(dbq))
 	}, 2*time.Second, 100*time.Millisecond)
 
-	syncCtr := scope.Snapshot().Counters()["test.sync_throttle_count+namespace="+matchingTestNamespace+",namespace_state=active,operation=TaskQueueMgr,partition=0,service_name=matching,task_type=Activity,taskqueue=makeToast,worker-build-id=__unversioned__"]
+	syncCtr := scope.Snapshot().Counters()["test.sync_throttle_count+namespace="+matchingTestNamespace+",namespace_state=active,operation=TaskQueueMgr,partition=0,service_name=matching,task_type=Activity,taskqueue=makeToast,worker_version=__unversioned__"]
 	s.Equal(1, int(syncCtr.Value())) // Check times zero rps is set = throttle counter
 	expectedRange := int64((taskCount + 1) / 30)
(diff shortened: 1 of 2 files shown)


common/dynamicconfig/constants.go
```diff
@@ -1191,9 +1191,9 @@ observability stack. Disabling this option will disable all the per-Task Queue g
 	MetricsBreakdownByBuildID = NewTaskQueueBoolSetting(
 		"metrics.breakdownByBuildID",
 		true,
-		`MetricsBreakdownByBuildID determines if the 'worker-build-id' tag in Matching metrics should
-contain the actual Build ID or a generic "__versioned__" value. Regardless of this config, the tag value for unversioned
-queues will be "__unversioned__". Disable this option if the Build ID cardinality is too high for your
+		`MetricsBreakdownByBuildID determines if the 'worker_version' tag in Matching metrics should
```
carlydf, line 1194: `worker-version`?
```suggestion
		`MetricsBreakdownByBuildID determines if the 'worker-version' tag in Matching metrics should
```

common/metrics/tags.go
```diff
@@ -34,7 +34,7 @@ const (
 	commandType    = "commandType"
 	serviceName    = "service_name"
 	actionType     = "action_type"
-	workerBuildId  = "worker-build-id"
+	workerVersion  = "worker_version"
```
carlydf, line 37: `worker-version` or `worker_version`?

common/log/tag/tags.go
```diff
@@ -298,12 +298,12 @@ func WorkflowTaskQueueName(taskQueueName string) ZapTag {
...
-func WorkerBuildId(buildId string) ZapTag {
-	if buildId == "" {
-		buildId = "_unversioned_"
+// WorkerVersion returns tag for worker build ID
+func WorkerVersion(version string) ZapTag {
+	if version == "" {
+		version = "_unversioned_"
 	}
-	return NewStringTag("worker-build-id", buildId)
+	return NewStringTag("worker-version", version)
```
carlydf, line 306: maybe refer to the `workerVersion` const here and then it's extra consistent? (unless they should be different)

carlydf: ohh is it ZapTag `worker-version` vs Tag `worker_version`?

carlydf (approved):

PR #8937, opened 2026-01-05
Add watch support to namespace registry

jprieto-temporal:
## What changed?
Add watch support to the namespace registry. When persistence supports namespace watches, the registry receives real-time updates via a channel instead of periodic polling. On startup, the registry attempts to establish a watch. If successful, it processes create/update/delete events as they arrive. If watches aren't supported (ErrWatchNotSupported), it falls back to the existing polling behavior.

  Other changes:
  - Add WatchNamespaces method to the `Persistence` interface used by the registry
  - Change stateChangeCallbacks from map to sync.Map to simplify callback handling from watches
  - Add various namespace registry metrics

## Why?
  Watch-based updates provide lower latency for namespace changes compared to polling, and reduce load on the persistence layer.

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
  - Persistence implementations that don't support watches will continue using polling (no behavior change)
  - Watch reconnection uses unlimited retries after initial startup succeeds, which could cause log spam if persistence becomes permanently unavailable


common/namespace/nsregistry/registry.go
```diff
@@ -85,19 +120,18 @@ type (
...
 
-		// nsMapsLock protects nameToID, idToNamespace and stateChangeCallbacks.
-
+		// nsMapsLock protects nameToID and idToNamespace
 		nsMapsLock    sync.RWMutex
 		nameToID      map[namespace.Name]namespace.ID
 		idToNamespace map[namespace.ID]*namespace.Namespace
 
-		stateChangeCallbacks          map[any]namespace.StateChangeCallbackFn
+		stateChangeCallbacks          sync.Map // map[any]StateChangeCallbackFn
```
collinvandyck, line 130: might be worth dropping a comment why we're using sync.Map instead of using an rw mutex + map. i know the stdlib documents some cases this is handy, like "written once, many reads" which i am guessing is why this changed :+1:

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +402,149 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+			}
+
+			// Create fresh watch context for this attempt
+			watchCtx, watchCancel := context.WithCancel(ctx)
+			currentWatchCancel = watchCancel
+
+			eventCh, err := r.persistence.WatchNamespaces(watchCtx)
+			if err != nil {
+				if !errors.Is(err, persistence.ErrWatchNotSupported) {
+					r.logger.Error("Error starting namespace watch", tag.Error(err))
```
collinvandyck, line 505: Is this going to end up double logging when we return the err afterwards? Might be worth just returning the err w/o the log?

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +402,149 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+// Each retry attempt creates a fresh context to ensure clean resource cleanup.
+func (r *registry) startWatch(ctx context.Context, initialWatch bool) (watchStartResult, error) {
+	// Track the current attempt's cancel function so we can cancel it on retry
+	var currentWatchCancel context.CancelFunc
+
+	result, err := backoff.ThrottleRetryContextWithReturn(
+		ctx,
+		func(ctx context.Context) (watchStartResult, error) {
+			// Cancel previous attempt's context to clean up any partial watch state
+			if currentWatchCancel != nil {
```
collinvandyck: I think it might be more ergonomic to use the pattern I saw elsewhere and defer canceling the context if the return error value is not nil. that way you wouldn't have to retain the currentWatchCancel (along with end of func cleanup) and could just use the result that comes back from backoff.

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +402,149 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+// retries on initial startup, unlimited on reconnection.
+func (r *registry) runWatchLoop(ctx context.Context) (*goro.Handle, bool) {
+	// watchStartedOnce tracks whether the watch has ever started successfully.
+	// Used to determine retry policy and signal to the caller when watch is ready.
+	watchStartedOnce := channel.NewShutdownOnce()
+
+	handle := goro.NewHandle(ctx).Go(
+		func(ctx context.Context) error {
+			// Outer loop handles watch restarts after connection failures.
+			for {
```
collinvandyck, line 456: I think `for ctx.Err() != nil` here is probably more readable. 

Or, in the loop, `if ctx.Err() != nil { return nil }`

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +402,149 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+				select {
+				case <-ctx.Done():
+					return nil
+				default:
+				}
+
+				result, err := r.startWatch(ctx, !watchStartedOnce.IsShutdown())
```
collinvandyck, line 463: I'm not too familiar with some of the APIs or patterns we use in OSS around concurrency, but that being said, I think it might be easier to understand if startWatch and watchLoop were somehow combined, and removing the for loop. You could pass in a `startErr = make(chan error, 1)` to the method to know when it had been started successfully, and then inspect that error to know if you needed to fall back to polling.

common/namespace/nsregistry/registry.go
```diff
@@ -85,19 +120,20 @@ type (
 		logger                  log.Logger
 		refreshInterval         dynamicconfig.DurationPropertyFn
 
-		// nsMapsLock protects nameToID, idToNamespace and stateChangeCallbacks.
-
+		// nsMapsLock protects nameToID and idToNamespace
```
dnr: also stateChangedDuringReadthrough, right?

common/namespace/nsregistry/registry.go
```diff
@@ -85,19 +120,20 @@ type (
...
 		nsMapsLock    sync.RWMutex
 		nameToID      map[namespace.Name]namespace.ID
 		idToNamespace map[namespace.ID]*namespace.Namespace
 
-		stateChangeCallbacks          map[any]namespace.StateChangeCallbackFn
+		// stateChangeCallbacks is a sync.Map so that it can be used without contending for the lock protecting the cache maps; we don't
+		// need to block namespace operations while running or updating callbacks.
+		stateChangeCallbacks          sync.Map // map[any]StateChangeCallbackFn
```
dnr, line 130: I'm slightly nervous about moving the callbacks to a different lock. it means you can't register a callback atomically with getting the current data? I'm sure you've thought about this though...

ok yeah I found your comment on RegisterStateChangeCallback. seems fine.

common/namespace/nsregistry/registry.go
```diff
@@ -154,28 +196,49 @@ func (r *registry) RefreshNamespaceById(id namespace.ID) (*namespace.Namespace,
...
-	if err != nil {
+	watchStarted := false
+	r.refresher, watchStarted = r.runWatchLoop(ctx)
+	if watchStarted {
+		// Watch started successfully
+		return
+	}
+
+	if err := r.refresher.Err(); !errors.Is(err, persistence.ErrWatchNotSupported) {
+		// Watch failed to start for a reason other than ErrWatchNotSupported. Abort.
```
dnr: what does Abort mean here? I mean, both cases fall through...

common/namespace/nsregistry/registry.go
```diff
@@ -154,28 +196,49 @@ func (r *registry) RefreshNamespaceById(id namespace.ID) (*namespace.Namespace,
...
+
+	// Fall back to polling
+	if err := r.refreshNamespaces(ctx); err != nil {
 		r.logger.Fatal("Unable to initialize namespace registry", tag.Error(err))
 	}
-	r.refresher = goro.NewHandle(ctx).Go(r.refreshLoop)
+	r.refresher = goro.NewHandle(ctx).Go(
+		func(ctx context.Context) error {
+			return r.runPollingLoop(ctx)
+		})
```
dnr: you can't write this as

```suggestion
	r.refresher = goro.NewHandle(ctx).Go(r.runPollingLoop)
```

?

common/namespace/nsregistry/registry.go
```diff
@@ -213,21 +279,47 @@ func (r *registry) getAllNamespace() []*namespace.Namespace {
...
+					tag.Key(fmt.Sprintf("%v", key)),
+					tag.NewDurationTag("duration", duration),
+				)
+			}
+		}()
+
+		cb(ns, deletedFromDb)
+	}
+
+	r.stateChangeCallbacks.Store(key, namespace.StateChangeCallbackFn(callbackWithTiming))
```
dnr, line 306: just a thought, maybe StateChangeCallbackFn should be a type alias instead of a new type? the casting just adds noise

common/namespace/nsregistry/registry.go
```diff
@@ -213,21 +279,47 @@ func (r *registry) getAllNamespace() []*namespace.Namespace {
...
+
 	r.nsMapsLock.Lock()
```
dnr: this can be changed to RLock now, right?

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +404,135 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
 	return ns.CustomSearchAttributesMapper(), nil
 }
 
-func (r *registry) refreshLoop(ctx context.Context) error {
+// watchLoop processes namespace watch events and handles restarts on failure.
+// It returns when the context is cancelled or the watch channel is closed.
+func (r *registry) watchLoop(ctx context.Context, watchCh <-chan *persistence.NamespaceWatchEvent) {
+	r.logger.Info("Starting watch loop")
```
dnr:
```suggestion
	r.logger.Info("Starting namespace registry loop")
```

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +404,135 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+		select {
+		case <-ctx.Done():
+			return
+		case event, ok := <-watchCh:
+			var err error
+			if ok {
+				if event.Err == nil {
+					err = r.processWatchEvent(event)
+				}
+				err = errors.Join(err, event.Err)
```
dnr: shouldn't this just be `... } else { err = event.Err }`? you can't have both

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +404,135 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+		}
+	}
+}
+
+// watchStartRetryPolicy returns a retry policy for watch startup attempts.
+// On initial startup (initialWatch=true), retries are limited to avoid blocking server startup indefinitely.
+// On reconnection after a previous success (initialWatch=false), retries continue indefinitely.
+func watchStartRetryPolicy(initialWatch bool) backoff.RetryPolicy {
+	policy := backoff.
+		NewExponentialRetryPolicy(CacheRefreshFailureRetryInterval)
```
dnr: one line?

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +404,135 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+// runWatchLoop starts a goroutine that establishes and maintains a namespace watch. Returns the goroutine handle and a
+// boolean indicating whether the watch started successfully. This method blocks until the watch is established and the first refresh
+// is successful.
+//
+// Uses ShutdownOnce to track whether the watch has ever started successfully, which affects retry behavior: limited
+// retries on initial startup, unlimited on reconnection.
+func (r *registry) runWatchLoop(ctx context.Context) (*goro.Handle, bool) {
+	// watchStartedOnce tracks whether the watch has ever started successfully.
+	// Used to determine retry policy and signal to the caller when watch is ready.
+	watchStartedOnce := channel.NewShutdownOnce()
```
dnr, line 451: hmm, this shutdownonce is just a simplified Future (common/future), right? maybe we should standardize on that (not in this PR, it's fine)

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +404,135 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+func (r *registry) startWatch(ctx context.Context, initialWatch bool) (watchStartResult, error) {
+	return backoff.ThrottleRetryContextWithReturn(
+		ctx,
+		func(ctx context.Context) (startResult watchStartResult, err error) {
+			// Create fresh watch context for this attempt
+			watchCtx, watchCancel := context.WithCancel(ctx)
+			defer func() {
+				if err != nil {
+					// Cancel attempt's watch context to clean up any partial watch state
+					watchCancel()
```
dnr, line 496: I'm feeling a little like the responsibilities here are confused: sometimes this function cancels, sometimes the caller (the func inside runWatchLoop) cancels. I don't have a concrete suggestion, and it doesn't look wrong as-is, but I find it's usually better if the cancels happen in the same place or at least at the same level.

oh heh, I see this was changed due to a comment. well, tradeoffs.

common/namespace/nsregistry/registry.go
```diff
@@ -312,8 +404,135 @@ func (r *registry) GetCustomSearchAttributesMapper(name namespace.Name) (namespa
...
+			return !errors.Is(err, persistence.ErrWatchNotSupported)
+		},
+	)
+}
+
+// runPollingLoop periodically refreshes the namespace cache.
+// Used as fallback when namespace watches are not supported.
+func (r *registry) runPollingLoop(ctx context.Context) error {
 	timer := time.NewTimer(r.refreshInterval())
+	defer timer.Stop()
```
dnr: not really necessary since go 1.23

common/namespace/nsregistry/registry.go
```diff
@@ -397,29 +626,86 @@ func (r *registry) refreshNamespaces(ctx context.Context) error {
...
-	for _, cb := range stateChangeCallbacks {
-		for _, ns := range deletedEntries {
-			cb(ns, true)
-		}
-		for _, ns := range stateChanged {
-			cb(ns, false)
+	r.stateChangeCallbacks.Range(
+		func(_, value any) bool {
+			//revive:disable-next-line:unchecked-type-assertion
+			cb := value.(namespace.StateChangeCallbackFn)
```
dnr, line 635: check for slow callbacks here too?

common/namespace/nsregistry/registry.go
```diff
@@ -519,19 +805,23 @@ func (r *registry) getOrReadthroughNamespaceByID(id namespace.ID) (*namespace.Na
...
+	r.updateSingleNamespace(ns, false)
 
 	return ns, nil
 }
 
-func (r *registry) updateSingleNamespace(ns *namespace.Namespace) {
+// updateSingleNamespace updates the cache with a namespace if it's newer than what we have.
+// Returns true if the namespace state changed.
+// When updatedViaWatch is true, we skip adding to stateChangedDuringReadthrough since watch events
+// trigger callbacks immediately and don't need to be queued for later delivery.
```
dnr, line 812: I feel like we should always "call immediately" and never defer. but don't need to change it now.

common/persistence/data_interfaces.go
```diff
@@ -1075,6 +1082,14 @@ type (
 		LastKnownTableVersion int64
 		ID                    string
 	}
+	NamespaceWatchEventType int
```
dnr, line 1086: blank line before?

common/persistence/data_interfaces.go
```diff
@@ -1075,6 +1082,14 @@ type (
...
+
+	NamespaceWatchEvent struct {
+		Type        NamespaceWatchEventType
+		Response    *GetNamespaceResponse
+		NamespaceID namespace.ID
```
dnr, line 1091: this is inside of Response.. any specific reason to copy it or just convenience?

dnr: oh, it's for deletes, duh

common/namespace/nsregistry/registry_watch_test.go
```diff
@@ -0,0 +1,1523 @@
...
+	s.logger = testlogger.NewTestLogger(s.T(), testlogger.FailOnAnyUnexpectedError)
+	s.captureHandler = metricstest.NewCaptureHandler()
+	s.capture = s.captureHandler.StartCapture()
+	s.registry = s.newRegistry()
+}
+
+func (s *registryWatchSuite) TearDownTest() {
+	s.registry.Stop()
+	s.captureHandler.StopCapture(s.capture)
+	s.controller.Finish()
```
dnr: this isn't required, NewController automatically sets this up

dnr: btw, this isn't required, NewController automatically sets this up

common/persistence/data_interfaces.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: fixed comment ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: yup, cleaned it up ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: agree, will do in followup

common/namespace/nsregistry/registry.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: yeah agreed, i tried to find something more elegant, alas...

common/namespace/nsregistry/registry.go
jprieto-temporal: ✅

common/namespace/nsregistry/registry.go
jprieto-temporal: the stored callbacks are already wrapped with the metric code, so that should be tracked here too.

common/namespace/nsregistry/registry_watch_test.go
jprieto-temporal: ✅

PR #8941, opened 2026-01-06
CHASM: Fix standby retention timer generation

yycptt:
## What changed?
- Do not update CloseTime on standby
- Generate retention timer on standby as long as execution is not previously closed.

## Why?
- Previous logic won't generate retention timer on standby side if there are state transitions after execution close and those transitions get replicated together with the transition that closes the execution.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/history/workflow/task_refresher.go
+++ b/service/history/workflow/task_refresher.go
@@ -77,7 +77,23 @@ func (r *TaskRefresherImpl) Refresh(
 		return err
 	}
 
-	return mutableState.ChasmTree().RefreshTasks()
+	if err := mutableState.ChasmTree().RefreshTasks(); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


yycptt:
I actually have a xdc retention timer functional test [here](https://github.com/temporalio/temporal/pull/8942) passing as well. 

But today we have a hardcoded minimal retention period and I have to overwrite for the tests to work. So that's not ready for merging yet.

service/history/workflow/mutable_state_impl.go
```diff
@@ -7441,22 +7439,26 @@ func (ms *MutableStateImpl) closeTransactionPrepareTasks(
 }
 
 func (ms *MutableStateImpl) closeTransactionGenerateChasmRetentionTask(
-	isStateDirty bool,
+	transactionPolicy historyi.TransactionPolicy,
 ) error {
 
-	if !isStateDirty ||
```
yycptt, line 7447: it's fine to check isStateDirty as well (since state changed to close means the root component state must have changed), but not really necessary, we only need to look at current state and state in db.

PR #8942, opened 2026-01-06
CHASM: XDC retention timer test

yycptt:
## What changed?
- Add xdc functional test for chasm retention timer

## Why?
- Add test coverage.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/xdc/chasm_test.go
+++ b/tests/xdc/chasm_test.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/suite"
-	"go.temporal.io/api/common/v1"
+	commonpb "go.temporal.io/api/common/v1"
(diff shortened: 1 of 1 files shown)


common/namespace/const.go
```diff
@@ -5,7 +5,7 @@ import "time"
 const (
 	// MinRetentionGlobal is a hard limit for the minimun retention duration for global
 	// namespaces (to allow time for replication).
-	MinRetentionGlobal = 1 * 24 * time.Hour
+	MinRetentionGlobal = time.Second // 1 * 24 * time.Hour
```
yycptt: I need to find a way to lower this only in tests.

yycptt: https://github.com/temporalio/temporal/pull/8993

PR #8945, opened 2026-01-06
Use atomic.Int32 in namespace registry

jprieto-temporal:
## What changed?
Replaced the plain int32 status field in the namespace registry with atomic.Int32. Updated Start() and Stop() to use the type's methods instead of package-level atomic functions.

## Why?
Using atomic.Int32 is more type-safe and idiomatic in modern Go.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
None. This is a mechanical refactor with identical runtime behavior.


jprieto-temporal:
@dnr i agree: https://github.com/temporalio/temporal/pull/8948

PR #8947, opened 2026-01-06
Publish blob size error limit exceeded metrics

simvlad:
## What changed?
Added a new counter event_blob_size_error, which is bumped every time the payloads size (BlobSizeLimitError) or  memo size (MemoSizeLimitError) is crossed.

## Why?
Currently we do not have a specific metric monitoring blob size errors and ErrBlobSizeExceedsLimit error that is thrown is serviceerror.NewInvalidArgument, which is aggregated with other errors of that kind. 

## How did you test it?
- [X] built
- [ ] added new unit test(s)

## Potential risks
No

### DIFF
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -648,8 +648,8 @@ var (
 	TlsCertsExpiring                         = NewGaugeDef("certificates_expiring")
 	ServiceAuthorizationLatency              = NewTimerDef("service_authorization_latency")
 	EventBlobSize                            = NewBytesHistogramDef("event_blob_size")
-	EventBlobSizeError                       = NewCounterDef(
-		"event_blob_size_error",
+	BlobSizeError                            = NewCounterDef(
+		"blob_size_error",
 		WithDescription("The number of requests that failed due to blob size exceeding limits configured with BlobSizeLimitError and MemoSizeLimitError."),
 	)
 	HeaderSize        = NewBytesHistogramDef("header_size", WithDescription("The size of the header in bytes passed to the server by the client. This metric is experimental and can be removed in the future."))
--- a/common/util.go
+++ b/common/util.go
@@ -602,7 +602,7 @@ func CheckEventBlobSizeLimit(
 		}
 
 		if actualSize > errorLimit {
-			metrics.EventBlobSizeError.With(metricsHandler).Record(1, metrics.OperationTag(operation))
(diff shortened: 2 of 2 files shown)


common/metrics/metric_defs.go
```diff
@@ -648,13 +648,17 @@ var (
...
-	LockLatency                              = NewTimerDef("lock_latency")
-	SemaphoreRequests                        = NewCounterDef("semaphore_requests")
-	SemaphoreFailures                        = NewCounterDef("semaphore_failures")
-	SemaphoreLatency                         = NewTimerDef("semaphore_latency")
-	ClientRequests                           = NewCounterDef(
+	EventBlobSizeError                       = NewCounterDef(
+		"event_blob_size_error",
+		WithDescription("The number of requests that failed due to blob size exceeding limits configured with BlobSizeLimitError and MemoSizeLimitError."),
+	)
+	HeaderSize        = NewBytesHistogramDef("header_size", WithDescription("The size of the header in bytes passed to the server by the client. This metric is experimental and can be removed in the future."))
```
jmaeagle99, line 655: nit: Should the spacing be restored so it lines up most of the other metric definitions? Or did the formatter remove the extra space?

jmaeagle99 (approved):

common/metrics/metric_defs.go
```diff
@@ -648,13 +648,17 @@ var (
...
 	EventBlobSize                            = NewBytesHistogramDef("event_blob_size")
-	HeaderSize                               = NewBytesHistogramDef("header_size", WithDescription("The size of the header in bytes passed to the server by the client. This metric is experimental and can be removed in the future."))
-	LockRequests                             = NewCounterDef("lock_requests")
-	LockLatency                              = NewTimerDef("lock_latency")
-	SemaphoreRequests                        = NewCounterDef("semaphore_requests")
-	SemaphoreFailures                        = NewCounterDef("semaphore_failures")
-	SemaphoreLatency                         = NewTimerDef("semaphore_latency")
-	ClientRequests                           = NewCounterDef(
+	EventBlobSizeError                       = NewCounterDef(
+		"event_blob_size_error",
```
yycptt: `event` won't be accurate  for chasm payload which could be stored in mutable states. I'd probably not having it from beginning so we don't need to do a migration later. Though yeah, I understand it will be in-consistent with the existing size metric above, but should be ok I feel.

PR #8951, opened 2026-01-07
[Scheduled Actions] Clean up task validation logic

lina-temporal:
## What changed?
- Tasks now also retain a taskVersion, used to invalidate an immediately-scheduled task.
- The inline immediate task execution optimization has been enabled.
- A few bugs found during testing related to improper validation have been fixed as a result of this.

## How did you test it?
- Features, functional, unit tests. This fixes some existing stuff, and I've added a new test for something that was specifically broken without these changes.

### DIFF
--- a/chasm/lib/scheduler/generator.go
+++ b/chasm/lib/scheduler/generator.go
@@ -36,9 +36,7 @@ func NewGenerator(ctx chasm.MutableContext) *Generator {
 // Generate immediately kicks off a new GeneratorTask. Used after updating the
 // schedule specification.
 func (g *Generator) Generate(ctx chasm.MutableContext) {
-	ctx.AddTask(g, chasm.TaskAttributes{
-		ScheduledTime: chasm.TaskScheduledTimeImmediate,
-	}, &schedulerpb.GeneratorTask{})
+	g.scheduleTask(ctx, chasm.TaskScheduledTimeImmediate)
 }
 
 // scheduleTask schedules a GeneratorTask at the given time.


chasm/lib/scheduler/proto/v1/tasks.proto
yycptt, line 1: Are they all pure tasks? If so looks like you can just rely on the immediate pure task optimization which guarantees that it will only be executed once. And if those pure tasks are not immediate, the existing task validation logic based on the timestamp watermark will work?

lina-temporal: > Are they all pure tasks? If so looks like you can just rely on the immediate pure task optimization which guarantees that it will only be executed once. 

Unfortunately, no; after we chatted it occurred to me that `InvokerExecuteTask` would be an immediate side effect task. I figured it'd be a bit simpler to do so for all immediate tasks. 

> And if those pure tasks are not immediate, the existing task validation logic based on the timestamp watermark will work?

Yeah, existing timestamp validation works for any non-immediate time, it's just to handle the "immediate task needs to invalidate itself" case.

yycptt: hmm I see the three tasks changed here (generator, processBuffer, backfiller) are all marked as pure in library.go. `InvokerExecuteTask` is a side effect one yes, but it's not updated. Am I still missing something here?

lina-temporal: 🤔  no, now I'm wondering if I am.. I thought I'd changed all 4. Let me try validating without the counters on this. I'd used them to get the functional test passing before I tackled the unit test updates for the inline immediate execution, so it's possible they aren't necessary at all. Though it still seems like the side effect task should require one and I'm not sure why that doesn't.

yycptt: For side effect task as we discussed yesterday, it's ok if the task is not immediately invalidated after execution (when the processing result is not recorded in the state). As long as eventually (hopefully within several minutes) it's invalidated it's fine. Otherwise the only impact I can think of right now is that it will cause standby task to be retried for a long time on the standby side eventually dropped and trigger alerts.

Actually I do see the execution logic for InvokerExecuteTask will update the state, so the validator for it should be working?

lina-temporal: I was able to remove the counters and keep the optimization, there had actually been a separate bug in `InvokerProcessBufferTask`'s loop that was causing it to continually return valid. As far as `InvokerExecuteTask`, although it's an immediate side effect, however it doesn't validate with the HWM approach, so it doesn't need this separate fix either. Sorry, long time spent in task debugging world yesterday had me a bit turned around.

chasm/lib/scheduler/backfiller.go
```diff
@@ -41,17 +41,22 @@ func addBackfiller(
...
 
-	// Immediately schedule the first backfiller task.
-	ctx.AddTask(backfiller, chasm.TaskAttributes{}, &schedulerpb.BackfillerTask{})
-
 	if scheduler.Backfillers == nil {
 		scheduler.Backfillers = make(chasm.Map[string, *Backfiller])
 	}
 	scheduler.Backfillers[id] = chasm.NewComponentField(ctx, backfiller)
 
+	backfiller.scheduleTask(ctx, chasm.TaskScheduledTimeImmediate)
```
yycptt, line 49: just checking, using `chasm.TaskScheduledTimeImmediate` or not give the same behavior right?

yycptt (approved):
not super familiar with the buffered start logic but the rest LGTM.

chasm/lib/scheduler/backfiller.go
lina-temporal: Yep, same behavior, and backfills reschedule themselves at particular points of time when backing off.

PR #8952, opened 2026-01-07
[Scheduled Actions] Fix a few bugs related to manual starts

lina-temporal:
## What changed?
Fixed a few bugs exposed by the features tests related to manual starts.

- Trigger manual starts should have a workflowID set up front
- startWorkflowDeadline was inadvertently being applied to backfills
- SpecProcessor was inadvertently discarding actions from backfills in the past

### DIFF
--- a/chasm/export_test.go
+++ b/chasm/export_test.go
@@ -29,10 +29,6 @@ func (r *Registry) TaskOf(taskGoType reflect.Type) (*RegistrableTask, bool) {
 	return r.taskOf(taskGoType)
 }
 
-func (r *Registry) TaskByID(id uint32) (*RegistrableTask, bool) {
-	return r.taskByID(id)
-}
-
 func (rc RegistrableComponent) FqType() string {
 	return rc.fqType()
 }
--- a/chasm/registrable_component.go
+++ b/chasm/registrable_component.go
@@ -175,6 +175,11 @@ func (rc *RegistrableComponent) hasBusinessIDAlias() bool {
 	return ok
 }
 
+// GoType returns the reflect.Type of the component's Go struct.
+func (rc *RegistrableComponent) GoType() reflect.Type {
+	return rc.goType
+}
+
 // fqType returns the fully qualified name of the component, which is a combination of
 // the library name and the component type. This is used to uniquely identify
 // the component in the registry.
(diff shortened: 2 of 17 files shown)


chasm/lib/scheduler/backfiller_tasks.go
```diff
@@ -191,6 +191,8 @@ func (b *BackfillerTaskExecutor) processTrigger(
 	nowpb := backfiller.GetLastProcessedTime()
 	now := nowpb.AsTime()
 	requestID := generateRequestID(scheduler, backfiller.GetBackfillId(), now, now)
+	nominalTimeSec := now.Truncate(time.Second)
+	workflowID := fmt.Sprintf("%s-%s", scheduler.WorkflowID(), nominalTimeSec.Format(time.RFC3339))
```
chaptersix: I'd prefer that we have the ID creation logic in a function. I'll likely need to import them into the migration logic.

chaptersix: I can move if in my PR though

chasm/lib/scheduler/spec_processor.go
```diff
@@ -120,9 +120,11 @@ func (s *SpecProcessorImpl) ProcessTimeRange(
 	for next, err = s.NextTime(scheduler, start); err == nil && (!next.Next.IsZero() && !next.Next.After(end)); next, err = s.NextTime(scheduler, next.Next) {
 		lastAction = next.Next
 
-		if scheduler.Info.UpdateTime.AsTime().After(next.Next) {
+		if scheduler.Info.UpdateTime.AsTime().After(next.Next) && !manual {
 			// If we've received an update that took effect after the LastProcessedTime high
 			// water mark, discard actions that were scheduled to kick off before the update.
+			// Skip this check for manual (backfill) actions since they explicitly request
```
chaptersix, line 125: i'm not sure in what context we would receive updates before the high watermark. Is there some way we guarantee that we don't receive "invalid" updates for backfills?

chasm/lib/scheduler/backfiller_tasks.go
lina-temporal: Great point, will fix.

chasm/lib/scheduler/spec_processor.go
lina-temporal: > i'm not sure in what context we would receive updates before the high watermark.

We never would/could (aside from a concurrent write failure). The case here is that the backfiller's high water mark moves through the backfill request's time range, not real time. Backfills are often requested for time stamps in the past, therefore, the *backfill*'s HWM will come before the schedule block's update time.

PR #8956, opened 2026-01-07
External payload metrics

simvlad:
## What changed?
Track metrics for total size and count of uploaded external payloads and the histogram of individual payloads.

## Why?
We only track the payload sizes as they appear in history, and for external payloads, it's a small size the lightweight reference occupies. We'd like to log the additional information about the number and size of uploaded external payloads, as well as the histogram of their sizes. 

## How did you test it?
- [X] built
- [] added new unit test(s)

## Potential risks
No

### DIFF
--- a/common/metrics/metric_defs.go
+++ b/common/metrics/metric_defs.go
@@ -1064,6 +1064,7 @@ var (
 	DynamicWorkerPoolSchedulerDequeuedTasks = NewCounterDef("dynamic_worker_pool_scheduler_dequeued_tasks")
 	DynamicWorkerPoolSchedulerRejectedTasks = NewCounterDef("dynamic_worker_pool_scheduler_rejected_tasks")
 	PausedActivitiesCounter                 = NewCounterDef("paused_activities")
+	ExternalPayloadUploadSize               = NewBytesHistogramDef("external_payload_upload_size", WithDescription("The histogram of sizes in bytes of uploaded external payloads."))
 
 	// Deadlock detector metrics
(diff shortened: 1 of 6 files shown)


jmaeagle99 (approved):

common/metrics/metric_defs.go
```diff
@@ -1060,6 +1060,9 @@ var (
 	DynamicWorkerPoolSchedulerDequeuedTasks = NewCounterDef("dynamic_worker_pool_scheduler_dequeued_tasks")
 	DynamicWorkerPoolSchedulerRejectedTasks = NewCounterDef("dynamic_worker_pool_scheduler_rejected_tasks")
 	PausedActivitiesCounter                 = NewCounterDef("paused_activities")
+	ExternalPayloadUploadCount              = NewCounterDef("external_payload_upload_count", WithDescription("The total number of external payloads uploaded."))
+	ExternalPayloadUploadTotalSize          = NewCounterDef("external_payload_upload_total_size", WithDescription("The total size of all external payloads uploaded."))
```
yycptt: The histogram below automatically gives you the `xxx_count` metric I believe. 

`_sum` is also available but I am not sure if the sum is based on the actual size or bucketed size.

simvlad: You are right. And `_sum` is on the actual size, so only histogram is needed.

yycptt: > _sum is on the actual size

yeah that's my understanding as well. But there was one time I noticed that when emitting 0 value, the `_sum` is not 0 until I explicitly add a 0 bucket 🤷‍♂️

yycptt (approved):

PR #8958, opened 2026-01-07
Add initial stubs and boilerplate for Nexus operation CHASM implementation

pdoerner:
## What changed?
WISOTT
Also added the dynamic config feature flag that will control whether the CHASM-based implementation will be used so that future PRs will not have conflicts with this key.

## Why?
Need to port Nexus operations to CHASM

## How did you test it?
- [x] built
- [ ] added new unit test(s)


chasm/lib/nexusoperation/proto/v1/operation.proto
bergundy, line 1: Might as we call this something other than `message.proto`.

chasm/lib/nexusoperation/operation_statemachine.go
bergundy, line 1: I would consider not exporting the transitions here. There wasn't a need to export them in the original HSM implementation either.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// EventScheduled is triggered when the operation is meant to be scheduled - immediately after initialization.
+type EventScheduled struct {
+}
+
+var TransitionScheduled = chasm.NewTransition(
+	[]nexusoperationpb.OperationStatus{nexusoperationpb.OPERATION_STATUS_UNSPECIFIED},
+	nexusoperationpb.OPERATION_STATUS_SCHEDULED,
+	func(o *Operation, ctx chasm.MutableContext, event EventScheduled) error {
```
bergundy, line 16: nit: I would make these methods on the `Operation` struct. Not sure why we didn't do that for standalone activities.

chasm/lib/nexusoperation/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var OutboundRequestCounter = metrics.NewCounterDef(
+	"nexus_outbound_requests",
+	metrics.WithDescription("The number of Nexus outbound requests made by the history service."),
+)
+var OutboundRequestLatency = metrics.NewTimerDef(
+	"nexus_outbound_latency",
+	metrics.WithDescription("Latency of outbound Nexus requests made by the history service."),
+)
```
bergundy, line 14: You're going to want to reference these in the HSM code because you can't define the same metric twice: https://github.com/temporalio/temporal/blob/7741b1cd035757a5c85969656d41ce638387bf06/components/callbacks/metrics.go#L8

PR #8959, opened 2026-01-07
Release workflow lock earlier in standby transfer task handling

robholland:
## What changed?
Added the ability for transfer task handlers to release the workflow lock early.

## Why?
This lowers lock contention for cases where a GRPC request will be made after mutable state is no longer required.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

---

> [!NOTE]
> Introduces early lock release in standby transfer processing to lower contention and plumbs the release function across standby executors.
> 
> - Extends `standbyActionFn` to include `ReleaseWorkflowContextFunc` and updates timer/transfer standby executors to pass it through
> - In `processCloseExecution` and `processStartChildExecution`, copies required fields from `mutableState` and calls `release(nil)` before `VerifyChildExecutionCompletionRecorded` / `VerifyFirstWorkflowTaskScheduled` RPCs; notes idempotent release
> - Adjusts related action lambdas to accept the new parameter; behavior otherwise unchanged aside from earlier lock release during external verification
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8edc218623de6e67f8e6933bbace801ac9dda23c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


robholland:
I considered the alternative of moving the GRPC requests to `postAction` as the lock is release before that is called, but that felt like it was distorting the action vs postAction just to abuse the lock scope, so preferred this approach.

robholland:
Also the active executor releases the lock in it's actions, it just doesn't have the action vs post split that the standby does.

service/history/timer_queue_standby_task_executor.go
```diff
@@ -175,6 +176,7 @@ func (t *timerQueueStandbyTaskExecutor) executeChasmSideEffectTimerTask(
 		ctx context.Context,
 		wfContext historyi.WorkflowContext,
 		ms historyi.MutableState,
+		release historyi.ReleaseWorkflowContextFunc,
```
yux0: nit: use _ for not used release 
and can you add a comment to mention not to access ms after release?

service/history/transfer_queue_standby_task_executor.go
```diff
@@ -431,6 +443,16 @@ func (t *transferQueueStandbyTaskExecutor) processStartChildExecution(
 		childStarted := childWorkflowInfo.StartedEventId != common.EmptyEventID
 		childAbandon := childWorkflowInfo.ParentClosePolicy == enumspb.PARENT_CLOSE_POLICY_ABANDON
 
+		// Copy needed values from childWorkflowInfo before releasing mutable state
+		targetNamespaceID := childWorkflowInfo.NamespaceId
+		targetNamespaceName := namespace.Name(childWorkflowInfo.Namespace)
+		startedWorkflowID := childWorkflowInfo.StartedWorkflowId
+		startedRunID := childWorkflowInfo.StartedRunId
```
yux0: nit: it would be easier to track with a child prefix in those namings.

service/history/timer_queue_standby_task_executor.go
robholland: Not sure where you mean to leave a comment, both release calls have `// no need for mutable state anymore, release workflow lock`. Do you mean you want that wording changed to say don't touch mutable state?

PR #8960, opened 2026-01-07
Run the features test for CHASM scheduler as part of CI

lina-temporal:
## What changed?
- Runs the Go SDK features test, with CHASM scheduler enabled, as part of CI.
- Relies on https://github.com/temporalio/features/pull/716 being merged
- Relies on pending schedule bugfix PRs #8951 #8952 to pass

## Why?
- These black-box tests have found bugs that the server functional tests don't, they should be part of regular CI until CHASM is flipped on by default.

### DIFF
--- a/.github/workflows/features-integration.yml
+++ b/.github/workflows/features-integration.yml
@@ -103,6 +103,8 @@ jobs:
       version-is-repo-ref: false
       docker-image-artifact-name: temporal-server-docker
       dynamic-config-values: |
+        history.enableTransitionHistory:
+          - value: true
         history.enableChasm:
           - value: true
         history.enableCHASMSchedulerCreation:


.github/workflows/features-integration.yml
```diff
@@ -95,6 +95,19 @@ jobs:
...
+    needs: build-docker-image
+    uses: temporalio/features/.github/workflows/go.yaml@main
+    with:
+      version: __latest_features_docker_image__
+      version-is-repo-ref: false
+      docker-image-artifact-name: temporal-server-docker
+      dynamic-config-values: |
+        history.enableChasm:
+          - value: true
+        history.enableCHASMSchedulerCreation:
```
chaptersix, line 110: callback config?

lina-temporal: I'd probably leave that disabled for now since our first rollout will likely involve non-CHASM callbacks, but I don't feel too strongly.

lina-temporal: Name's a bit confusing in this context, that enables the CHASM implementation of callbacks; both implementations will properly handle delivery to CHASM targets, though.

chaptersix: makes sense to me. Whatever config we are rolling out with next is what we should have.

Issue #5680, opened 2024-04-08
Official support for Opensearch


jsecchiero:
As new versions are released, the features of OpenSearch and Elasticsearch will continue to diverge ([ref1](https://opensearch.org/faq/#q1.14) [ref2](https://nextbrick.com/opensearch-vs-elasticsearch-check-everything-before-you-choose/)). However, i'm wondering if there will be an official statement in the documentation about Temporal's support for OpenSearch in the future. 
Currently, only recent versions [Elasticsearch are supported](https://docs.temporal.io/self-hosted-guide/visibility#elasticsearch). This information will be helpful when deciding on the datastore to use with Temporal

yiminc:
It is not officially claimed as supported, but there is no issue using OpenSearch as visibility store.

mike-wilson-wellhive:
There may be no issue using OpenSearch as the visibility store _today_, but some Temporal consumers need at least some assurance that the integration will continue working in the future, before they decide to start using it.

I would feel comfortable using OpenSearch as my visibility store for production workloads if the Temporal team formally supported it by doing things like:
- Document OpenSearch-specific guidance and compatible versions of OpenSearch and Temporal. A section similar to this one would be great: https://docs.temporal.io/self-hosted-guide/visibility#elasticsearch.
- Regular testing of the integration with the documented compatible version pairs. (Whatever the team does to ensure postrgres, mysql, and elasticseach integrations continue to work)

ninjaMikeG:
Latest versions of opensearch and ES have diverged - the response type has changed so ES client libraries are not recognizing Opensearch responses. With 1.25.1, I am not able to connect to my Opensearch instance after an upgrade from 1.22 (going through intermediate releases)

jc2707:
Can someone from AWS speak to this? Are you working with Temporal folks for support for latest OS versions? Or if temporal has it roadmapped already....

VLZZZ:
We're using Temporal with AWS Opensearch for a while already. I understand it's not supported officially, but I believe the only thing we need to change to support the Opensearch is to enable backward compatibility

```
PUT _cluster/settings
{
    "persistent" : {
        "compatibility.override_main_response_version" : true
    }
}
```

But we're still looking forward to dropping this workaround as soon as it will be supported

samof76:
@VLZZZ which version of Opensearch are you using?

matlegit:
[...]

PR #8964, opened 2026-01-08
Set the "missing" key in ES sorter

rodrigozhou:
## What changed?
Explicitly set the `"missing"` key in ES sorter. This is to address pagination bug in OpenSearch 2.8.0+ (see https://github.com/opensearch-project/OpenSearch/issues/8212#issuecomment-1615325095).

Run CI tests with latest version of OpenSearch 2 and OpenSearch 3.

## Why?
Full support of OpenSearch as visibility store.
Issue: https://github.com/temporalio/temporal/issues/5680

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -427,6 +427,7 @@ jobs:
           - cass_es
           - cass_es8
           - cass_os2
+          - cass_os3
           - sqlite
           - mysql8
           - postgres12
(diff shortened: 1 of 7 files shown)


awln-temporal (approved):

rodrigozhou:
@awln-temporal The bug is when `missing` is not explicitly set, pagination doesn't work correctly when there are null values in the fields in the sorting key.

barryib:
Can we add this support into the doc to clearly highlight that Opensearch is supported ?

rodrigozhou:
@barryib Once v1.30.1 is released, the docs will be updated.

PR #8973, opened 2026-01-08
Review agent guidelines

stephanos:
## What changed?

Added prompt ("skill") for AI review agents.

Note that the source of truth file is in `.github` since it's the only agent that does not support file references.

## Why?

There are common review issues that cannot be linted easily that we want to enforce without human intervention.

## How did you test it?

Added a violation and ran review agents:

### Cursor 

image

### Copilot

image

### Claude Code via `/review`

Screenshot 2026-01-08 at 11 21 15 AM

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -19,13 +19,13 @@ Apply these patterns when reviewing PRs or suggesting code changes.
 - Avoid stuttering: don't use `ActivityStatus` in package `activity`, just `Status`
 - Use `ok` boolean pattern instead of nil checks where idiomatic
 
-## 3. Test Suite Correctness
+## 3. Testify Suite Correctness
 
 - Never use `s.T()` in subtests - use the subtest's `t` parameter
 - Never use suite assertion methods (`s.NoError`, `s.Equal`) from goroutines - causes panics
 - Use `EventuallyWithT` when you need assertions inside eventually blocks, and use that block's `t`
 - Use `require.ErrorAs(t, err, &specificErr)` for specific error type checks
-- Don't use `assert` when you need to stop execution - use `require`
+- Prefer `require` over `assert` - it's rarely useful to continue a test after a failed assertion
 - Add comments explaining why `Eventually` is needed (e.g., eventual consistency)
 
 ## 4. Inline Code / Avoid Abstractions
@@ -42,14 +42,12 @@ Apply these patterns when reviewing PRs or suggesting code changes.
 - Use standard error types (`InvalidArgument`, `NotFound`, `FailedPrecondition`) over custom error types
 - Mark errors as non-retryable when task shouldn't retry in queue
 - Wrap errors with context when there's something interesting or informative to add, e.g. `fmt.Errorf("multi-operation part 2: %w", err)`
-- Use early returns over nested if/else blocks
 - Don't panic in library code - return errors and let caller decide
 - Validate early in handlers, not deep in business logic
 
 ## 6. Consistency with Codebase
 
 - Follow existing patterns: "We have been passing through the frontend request in other libraries. Let's keep the same pattern here"
-- Don't separate temporal/non-temporal imports with blank lines (team decision)
 - Use existing utilities before creating new ones
 - Follow CLI documentation conventions (capitalize proper nouns)
 - Match existing metric tag formats (CONSTANT_CASE for enum values)


stephanos:
cursor review

.cursor/rules/review.mdc
```diff
@@ -0,0 +1,9 @@
+---
+description: Code review guidelines - apply when reviewing code or suggesting changes
+globs: "**/*.go"
```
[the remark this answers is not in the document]
bergundy: Good call, we should consider convention for proto files but for now I think most of the concern is for Go code.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+- Use proper field names: `request_id` not `requestId`, `schedule_time` not `scheduledTime`
+- Don't expose internal concepts in user-facing errors: "LowCardinalityKeyword is not a user facing concept"
+- Accept event attributes structs instead of growing function signatures
+- Use `map[string]Empty` over `[]string` when you might add metadata later
+- Prefer enums over int/string for well-known values
+
+## 8. Concurrency and Safety
+
+- Use `atomic.Value` for concurrent access, not regular variables
+- Don't do IO while holding locks - use immediate tasks instead
```
lina-temporal: I believe this should be, "use side effect tasks"

.github/copilot-instructions.md
bergundy, line 1: I would also add: "Do not export anything that doesn't need to be exported".

.github/copilot-instructions.md
stephanos: Is "immediate task" an HSM, and "side effect task" a CHASM term? I haven't worked with either.

lina-temporal: An immediate task is simply one that should run "ASAP", but it doesn't define when ASAP precisely is, that's up to the executor. Any CHASM pure task will hold a lock for its duration, so an "immediate *pure* task" wouldn't be good for IO, whereas an "immediate *side effect* task" will execute in a separate transaction, without holding a lock.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,73 @@
...
+- Avoid unnecessary wrapper types and generic structs
+- Don't add dependencies for 5 lines of code - "just write 5 lines of code instead of adding more dependency bloat"
+- Don't create testsuite-level helpers that can't be safely used in subtests
+- Prefer explicit code over reflection: "It would be easier to follow without all of the reflection"
+
+## 5. Proper Error Handling
+
+- Use standard error types (`InvalidArgument`, `NotFound`, `FailedPrecondition`) over custom error types
+- Mark errors as non-retryable when task shouldn't retry in queue
+- Always provide context with errors: `fmt.Errorf("context: %w", err)`
```
dnr: small quibble here: errors should be wrapped with context when appropriate (when there's something interesting/informative about the context), not all the time. that produces super long error strings with multiple pieces that just repeat the same thing over and over.

```suggestion
- Wrap errors with context when there's something interesting or informative to add, e.g. `fmt.Errorf("multi-operation part 2: %w", err)`
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,73 @@
...
+- Match existing metric tag formats (CONSTANT_CASE for enum values)
+- Use the same error message style (no punctuation for single sentences)
+
+## 7. API and Proto Design
+
+- Document all proto fields with comments
+- Use proper field names: `request_id` not `requestId`, `schedule_time` not `scheduledTime`
+- Don't expose internal concepts in user-facing errors: "LowCardinalityKeyword is not a user facing concept"
+- Accept event attributes structs instead of growing function signatures
+- Use `map[string]Empty` over `[]string` when you might add metadata later
```
dnr: Are we talking about protos or go code here? It's not clear. And I'm a little skeptical of this as a rule in either case. Proto maps vs repeated fields are pretty different and there's a bunch of factors you'd want to consider, not just adding metadata.

In Go code, it's easy to change later so why make a rule?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,73 @@
...
+- Prefer enums over int/string for well-known values
+
+## 8. Concurrency and Safety
+
+- Use `atomic.Value` for concurrent access, not regular variables
```
dnr: In general, atomics are an advanced concurrency thing. Most code should default to using mutexes, and only reach for atomics when they greatly simplify some particular pattern, or there's a performance concern.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,73 @@
...
+- Don't do IO while holding locks - use side effect tasks
+- Clone data before releasing locks if it might be modified
+- Use `sync.Mutex` over `sync.RWMutex` when you only read once before writing
```
dnr: I would be much stronger here: prefer Mutex over RWMutex almost all the time. Except in two cases:
- reads are _much_ more common than writes (like > 1000×)
- readers need to hold the read lock for a significant amount of time

.github/copilot-instructions.md
```diff
@@ -0,0 +1,73 @@
...
+- Don't expose internal concepts in user-facing errors: "LowCardinalityKeyword is not a user facing concept"
+- Accept event attributes structs instead of growing function signatures
+- Use `map[string]Empty` over `[]string` when you might add metadata later
+- Prefer enums over int/string for well-known values
+
+## 8. Concurrency and Safety
+
+- Use `atomic.Value` for concurrent access, not regular variables
+- Don't do IO while holding locks - use side effect tasks
+- Clone data before releasing locks if it might be modified
```
dnr, line 70: Maybe we should add something like, "Try to use immutable data patterns (for normal structs and especially proto messages) wherever possible to avoid data races and synchronization."

.claude/skills/review/SKILL.md
```diff
@@ -0,0 +1,18 @@
+---
+name: review
+description: Review code for common issues. Use when reviewing PRs, suggesting code changes, or when user asks to review code.
+allowed-tools: Read, Grep, Glob, Bash, Edit
```
Shivs11: I wonder if we need Edit/Bash as tools if this skill is only for "reviewing" code?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
+# Code Review Guidelines
```
chaptersix, line 1: All of this would apply to a human reviewer as well. Should most of this be moved to a file that's more visible?

stephanos: From PR description

> Note that the source of truth file is in .github since it's the only agent that does not support file references.

So it would have to be a copy :-/ I'll say that's outside of the scope for now to not block this PR to serve its original purpose.

.claude/skills/review/SKILL.md
stephanos: Good question; let me change that.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+
+Apply these patterns when reviewing PRs or suggesting code changes.
+
+## 1. Remove Redundant Code (Highest Priority)
+
+- Remove code that doesn't add value to tests or implementation
+- Don't add unnecessary activities/complexity in tests - test only what you need
+- Question randomness in tests - test explicitly what you want
+- Don't add assertions for things you can assume work (e.g., "You are not testing TerminateWorkflowExecution here, you can assume it works")
```
dandavison, line 10: I agree this is usually what one wants. However, it is important to verify that the test is doing what you think it is doing. Therefore personally, I sometimes structure a test so that a routine can be called twice: once to assert that the test routine gives the expected behavior when the feature is not in effect, and a second time with the feature in effect. This style is in violation of this bullet point. Here's an example:

```go
	t.Run("Retry", func(t *testing.T) {
		runTest := func(t *testing.T, requestCancellation bool) {

			if requestCancellation {
				...
			}

			if requestCancellation {
				require.Equal(t, enumspb.ACTIVITY_EXECUTION_STATUS_FAILED, activityResp.GetInfo().GetStatus())
			} else {
				require.Equal(t, enumspb.ACTIVITY_EXECUTION_STATUS_RUNNING, activityResp.GetInfo().GetStatus())
				require.Equal(t, enumspb.PENDING_ACTIVITY_STATE_SCHEDULED, activityResp.GetInfo().GetRunState())
			}
		}
		t.Run("AllowedIfNotCancelRequested", func(t *testing.T) {
			runTest(t, false)  // this is asserting something that's not what we're actually testing
		})
		t.Run("PreventedIfCancelRequested", func(t *testing.T) {
			runTest(t, true)
		})
	})
```

Do you think `t.Run("AllowedIfNotCancelRequested"` should be a panic or something other than a normal test assertion?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+
+## 2. Go Naming Conventions
+
+- Don't use `Get` prefix for getters: `func (a *Activity) Store()` not `GetStore()`
+- Don't use `Impl` suffix for implementations
+- Don't put underscore after `Test` in test names: `TestRetry` not `Test_Retry`
+- Avoid stuttering: don't use `ActivityStatus` in package `activity`, just `Status`
+- Use `ok` boolean pattern instead of nil checks where idiomatic
+
+## 3. Test Suite Correctness
```
bergundy: Maybe clarify that this is for testify suites?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+
+- Never use `s.T()` in subtests - use the subtest's `t` parameter
+- Never use suite assertion methods (`s.NoError`, `s.Equal`) from goroutines - causes panics
+- Use `EventuallyWithT` when you need assertions inside eventually blocks, and use that block's `t`
+- Use `require.ErrorAs(t, err, &specificErr)` for specific error type checks
+- Don't use `assert` when you need to stop execution - use `require`
```
bergundy: I actually don't know when you would use assert, doesn't seem very useful. It is hard to write tests that are expected to continue after a failed assertion.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+- Mark errors as non-retryable when task shouldn't retry in queue
+- Wrap errors with context when there's something interesting or informative to add, e.g. `fmt.Errorf("multi-operation part 2: %w", err)`
+- Use early returns over nested if/else blocks
+- Don't panic in library code - return errors and let caller decide
+- Validate early in handlers, not deep in business logic
+
+## 6. Consistency with Codebase
+
+- Follow existing patterns: "We have been passing through the frontend request in other libraries. Let's keep the same pattern here"
+- Don't separate temporal/non-temporal imports with blank lines (team decision)
```
bergundy: There's a linter for this

```suggestion
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+- Don't add dependencies for 5 lines of code - "just write 5 lines of code instead of adding more dependency bloat"
+- Don't create testsuite-level helpers that can't be safely used in subtests
+- Prefer explicit code over reflection
+
+## 5. Proper Error Handling
+
+- Use standard error types (`InvalidArgument`, `NotFound`, `FailedPrecondition`) over custom error types
+- Mark errors as non-retryable when task shouldn't retry in queue
+- Wrap errors with context when there's something interesting or informative to add, e.g. `fmt.Errorf("multi-operation part 2: %w", err)`
+- Use early returns over nested if/else blocks
```
bergundy: There's a linter for this

```suggestion
```

PR #8974, opened 2026-01-08
Populate additional PollActivityTaskQueueResponse for standalone activities

fretz12:
## What changed?
Populated standalone activity Run ID and scheduled time when returning PollActivityTaskQueueResponse. Updated API deps. Added test specific to validating PollActivityTaskQueueResponse.

## Why?
Standalone activity executions have Run IDs that should be returned to the task poller. The response scheduled time should also be populated as the Temporal SDKs depend on it. 

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/common/namespace/namespace.go
+++ b/common/namespace/namespace.go
@@ -254,6 +254,11 @@ func (ns *Namespace) ReplicationPolicy() ReplicationPolicy {
 	return ReplicationPolicyOneCluster
 }
 
+// GetReplicationResolver return the replication resolover
+func (ns *Namespace) GetReplicationResolver() ReplicationResolver {
+	return ns.replicationResolver
+}
+
 func (ns *Namespace) GetCustomData(key string) string {
 	if ns.info.Data == nil {
 		return ""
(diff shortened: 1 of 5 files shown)


chasm/lib/activity/activity.go
```diff
@@ -220,8 +220,10 @@ func (a *Activity) PopulateRecordStartedResponse(ctx chasm.Context, key chasm.Ex
 	response.Attempt = attempt.GetCount()
 	response.Priority = a.GetPriority()
 	response.RetryPolicy = a.GetRetryPolicy()
+	response.ActivityRunId = key.RunID
 	response.ScheduledEvent = &historypb.HistoryEvent{
 		EventType: enumspb.EVENT_TYPE_ACTIVITY_TASK_SCHEDULED,
+		EventTime: a.GetScheduleTime(),
```
bergundy, line 226: This should be available from CHASM, no point in storing it in the activity state.

fretz12: Sounds like we need to expose to the chasm ctx like we did for CloseTime?
I'll take that on in a separate PR

PR #8976, opened 2026-01-08
TestTooManyCancelRequests improvements

pdoerner:
## What changed?
Improved `TestClientMiscTestSuite/TestTooManyCancelRequests`:
* Reduced the number of target workflows from 50 -> `testcore.ClientSuiteLimit + 1` (11)
* Increased timeout when waiting for failure in `CancelAllWorkflowsAtOnce`
* Removed unnecessary 3s wait

## Why?
Reduce flakiness

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/client_misc_test.go
+++ b/tests/client_misc_test.go
@@ -257,7 +257,7 @@ func (s *ClientMiscTestSuite) TestTooManyCancelRequests() {
 		})
 		s.NoError(err)
 		s.Equal(enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING, workflowExecution.State.ExecutionState.Status)
-		s.Zero(len(workflowExecution.State.RequestCancelInfos))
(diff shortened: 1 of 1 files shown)


tests/client_misc_test.go
```diff
@@ -198,7 +198,7 @@ func (s *ClientMiscTestSuite) TestTooManyCancelRequests() {
 	defer cancel()
 
 	// create a large number of blocked workflows
-	numTargetWorkflows := 50 // should be much greater than s.maxPendingCancelRequests
+	numTargetWorkflows := testcore.ClientSuiteLimit + 1
```
bergundy, line 201: nit: I would just set a fixed number here, not sure what this constant is meant to represent and if we want to tie this test to it.

pdoerner: It represents the maximum limit for pending items for this suite (pending activities, child workflows, etc). The other tests are already tied to this value.

PR #8978, opened 2026-01-08
Update replication task rate limiter priority based on active/standby…

yux0:
## What changed?
Update replication task rate limiter priority based on active/standby state

## Why?
The replication task should have higher priority than standby task 

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
There is a task processing priority change requires review


service/history/queues/scheduler.go
```diff
@@ -218,12 +221,17 @@ func NewRateLimitedScheduler(
...
-		namespaceName, err := namespaceRegistry.GetNamespaceName(namespace.ID(e.GetNamespaceID()))
-		if err != nil {
-			namespaceName = namespace.EmptyName
+		namespaceName := namespace.EmptyName.String()
+		ns, err := namespaceRegistry.GetNamespaceByID(namespace.ID(e.GetNamespaceID()))
+		if err == nil && ns != nil {
+			namespaceName = ns.Name().String()
+		}
+		if ns.ActiveInCluster(clusterMetadata.GetCurrentClusterName()) {
+			return quotas.NewRequest(e.GetType().String(), taskSchedulerToken, namespaceName, e.GetPriority().CallerType(), 0, "")
```
yux0: This is not changed

service/history/queues/scheduler.go
```diff
@@ -218,12 +221,17 @@ func NewRateLimitedScheduler(
...
+		} else {
+			return quotas.NewRequest(e.GetType().String(), taskSchedulerToken, namespaceName, headers.CallerTypePreemptable, 0, "")
```
yux0: Use lowest level for standby tasks

service/history/queues/scheduler.go
```diff
@@ -218,12 +221,16 @@ func NewRateLimitedScheduler(
...
-		if err != nil {
-			namespaceName = namespace.EmptyName
+		namespaceName := namespace.EmptyName.String()
+		ns, err := namespaceRegistry.GetNamespaceByID(namespace.ID(e.GetNamespaceID()))
+		if err == nil && ns != nil {
+			namespaceName = ns.Name().String()
 		}
-
-		return quotas.NewRequest(e.GetType().String(), taskSchedulerToken, namespaceName.String(), e.GetPriority().CallerType(), 0, "")
+		if ns.ActiveInCluster(clusterMetadata.GetCurrentClusterName()) {
```
yycptt: will this cause NPE if `ns` is nil?

service/history/queues/scheduler.go
```diff
@@ -200,6 +202,7 @@ func NewRateLimitedScheduler(
 	rateLimiter SchedulerRateLimiter,
 	timeSource clock.TimeSource,
 	chasmRegistry *chasm.Registry,
+	clusterMetadata cluster.Metadata,
```
yycptt: shall we just pass in current cluster name here and avoid getting it from clusterMetadata for each task?

service/history/queues/scheduler.go
yycptt: hmm guess ideally we should only set priority once on creation and later when failover happens. 
The whole activeness check for task processing probably needs to be refactored.

I think for now, can we move this logic into priority assigner and also update priority [here](https://github.com/yycptt/temporal/blob/447384babeb39138d0eb086c26de5cdee0566047/service/history/queues/executable.go#L360)?

yux0: Updated

service/history/queues/scheduler.go
yux0: Updated

service/history/queues/priority_assigner.go
```diff
@@ -11,18 +12,32 @@ type (
...
+		currentClusterName: currentClusterName,
+	}
 }
 
 func (a *priorityAssignerImpl) Assign(executable Executable) tasks.Priority {
+	ns, err := a.nsRegistry.GetNamespaceByID(namespace.ID(executable.GetNamespaceID()))
+	if ns != nil && err == nil {
+		// Use lowest priority level for standby task processing
+		if !ns.ActiveInCluster(a.currentClusterName) {
+			return tasks.PriorityPreemptable
```
prathyushpv, line 37: 👍

PR #8980, opened 2026-01-08
[Scheduler] Unify RunningWorkflows, RecentActions, BufferedStarts

lina-temporal:
## What changed?
- Unifies CHASM scheduler's Scheduler.Info.{RecentActions, RunningWorkflows} into BufferedStarts.
- Computes the RecentActions and RunningWorkflows views on-demand from BufferedStarts.
- Removes the old requestIDtoWorkflowID map.

## Why?
- Simplifies the data structure; much easier to reason about concurrently, as related records are no longer spread between distinct data structures.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


proto/internal/temporal/server/api/schedule/v1/message.proto
```diff
@@ -41,6 +41,24 @@ message BufferedStart {
...
+    // successfully started. Only used by the CHASM scheduler.
+    google.protobuf.Timestamp start_time = 11;
+    // Populated when the workflow execution completes. Presence indicates the
+    // action is complete and retained for history. Only used by the CHASM scheduler.
+    CompletedResult completed = 12;
+}
+
+// Result when a workflow execution has completed.
+// Only used by the CHASM scheduler.
+message CompletedResult {
```
chaptersix: why do we need the result?

chasm/lib/scheduler/scheduler.go
```diff
@@ -466,75 +435,14 @@ func (s *Scheduler) HandleNexusCompletion(
...
-			ScheduleTime: timestamppb.New(scheduleTime),
-			ActualTime:   timestamppb.New(scheduleTime), // best guess, as we're recording complete before start was recorded
-			StartWorkflowResult: &commonpb.WorkflowExecution{
-				WorkflowId: workflowID,
-			},
-		}
-
-		s.Info.RecentActions = util.SliceTail(
-			append(s.Info.RecentActions, actionResult), recentActionCount)
+	// Record the completed action in the Invoker.
```
chaptersix, line 438:
```
	// Handle pause-on-failure.
	if wfStatus != enumspb.WORKFLOW_EXECUTION_STATUS_COMPLETED &&
		s.Schedule.Policies.PauseOnFailure && !s.Schedule.State.Paused {
		s.Schedule.State.Paused = true
		s.Schedule.State.Notes = fmt.Sprintf(
			"paused, workflow %s: %s",
			strings.ToLower(wfStatus.String()),
			workflowID,
		)
	}
```
why don't we explicitly check for a wf status that is a fail? This seems to work because this handler only gets nexus events for non-running workflows. If that later changes this might break.

chasm/lib/scheduler/invoker.go
```diff
@@ -290,10 +287,90 @@ func (i *Invoker) processingDeadline() time.Time {
...
+// This includes both running workflows (with status RUNNING) and completed
+// workflows (with their final status).
+func (i *Invoker) recentActions() []*schedulepb.ScheduleActionResult {
+	var results []*schedulepb.ScheduleActionResult
+	for _, start := range i.GetBufferedStarts() {
+		// Only include workflows that have been started (have a RunId).
+		if start.GetRunId() == "" {
+			continue
+		}
+		status := enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING
```
chaptersix, line 336: Do we need to consider a paused workflow here?

chasm/lib/scheduler/scheduler.go
lina-temporal: > why don't we explicitly check for a wf status that is a fail?

We're trying to capture any "non-success" reasons that the workflow closed, including canceled/terminated. 

> If that later changes this might break.

For this particular handler, it's only meant for completion events, so it shouldn't break in the future (unless we were to introduce a separate success-completed event 🙂).

chasm/lib/scheduler/invoker.go
lina-temporal: That'd be useful if we could, but I don't think we really can at the moment, because we don't get Nexus callbacks for things other than completions (so we'd never find out if the running action got paused).

PR #8988, opened 2026-01-09
Add metadata context support to mutable state

nikki-dag:
## What changed?
Adds metadata context support to history service gRPC calls while setting information before finalizing any mutable state transaction.

## Why?
We want to pass mutable state metadata, such as workflow type, activity type, workflow task queue, and activity task queue, throughout the gRPC call lifecycle.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

---

> [!NOTE]
> Introduces a thread-safe metadata context and wires it into history workflow transactions and gRPC.
> 
> - Adds `common/contextutil` with `WithMetadataContext`, `ContextMetadataSet/Get/GetAll` for attaching arbitrary metadata to `context.Context` (with comprehensive unit tests)
> - Adds gRPC `MetadataContextInterceptor` to attach a metadata context to all incoming unary requests
> - Updates `MutableStateImpl.closeTransaction` to populate context metadata via `setMetaDataMap` (e.g., `workflow-type`, `workflow-task-queue`) based on workflow archetype
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8120cb9db13a273ec7669882e2470cf74247a8eb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/history/fx.go
```diff
@@ -216,12 +217,18 @@ func HealthCheckInterceptorProvider(
...
+}
+
 func HistoryAdditionalInterceptorsProvider(
 	healthCheckInterceptor *interceptor.HealthCheckInterceptor,
+	metadataContextInterceptor *interceptor.MetadataContextInterceptor,
 	chasmRequestEngineInterceptor *chasm.ChasmEngineInterceptor,
 	chasmRequestVisibilityInterceptor *chasm.ChasmVisibilityInterceptor,
 ) []grpc.UnaryServerInterceptor {
 	return []grpc.UnaryServerInterceptor{
+		metadataContextInterceptor.Intercept,
```
nikki-dag, line 232: We don't need to inject this in OSS yet, let me know if you think I should remove it.

common/contextutil/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	mc, ok := metadataCtx.(*metadataContext)
+	if !ok {
+		return nil
+	}
+	return mc
+}
+
+// AddMetadataContext adds a metadata context to the given context.
+func AddMetadataContext(ctx context.Context) context.Context {
```
bergundy:
```suggestion
func WithMetadataContext(ctx context.Context) context.Context {
```

common/contextutil/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"sync"
+)
+
+type (
+	metadataContextKey struct{}
+
+	// metadataContext is used to store workflow and activity metadata
+	metadataContext struct {
```
bergundy, line 12: I think you're going to want a method to extract the entire map so it can be added to the grpc metadata.

common/contextutil/metadata_test.go
bergundy, line 1: Seem a bit verbose / redundant TBH. I would try to keep tests to the bare minimum required. Concurrency is redundant IMHO.

common/rpc/interceptor/metadata_context.go
bergundy, line 1: I would consider not adding another interceptor and putting this into the ChasmEngineInterceptor to reduce call stack size.

service/history/workflow/mutable_state_impl.go
```diff
@@ -6943,10 +6944,34 @@ type closeTransactionResult struct {
...
+	switch ms.chasmTree.ArchetypeID() {
+	case chasm.WorkflowArchetypeID, chasm.UnspecifiedArchetypeID:
+		// Set workflow type
+		if wfType := ms.GetWorkflowType(); wfType != nil && wfType.GetName() != "" {
+			contextutil.ContextMetadataSet(ctx, "workflow-type", wfType.GetName())
+		}
+
+		// Set workflow task queue
+		if ms.executionInfo != nil && ms.executionInfo.TaskQueue != "" {
+			contextutil.ContextMetadataSet(ctx, "workflow-task-queue", ms.executionInfo.TaskQueue)
```
bergundy, line 6989: Can we use `task-queue` here so we can reuse this value for activity task queues later?

common/rpc/interceptor/metadata_context.go
nikki-dag: We need to inject this interceptor into the Frontend Service, and possibly into Matching as well. Since the `ChasmEngineInterceptor` is only used by the History service, I believe we will require a separate interceptor.

common/contextutil/metadata.go
nikki-dag: Added `ContextMetadataGetAll`

service/history/workflow/mutable_state_impl.go
nikki-dag: I might be wrong, but I was considering having both a `workflow-task-queue` and an `activity-task-queue` within the same context, such as a schedule activity command. It could be beneficial to distinguish between the two. Even for the activity heartbeat, we might find it useful to monitor the `workflow-task-queue` and `activity-task-queue` separately.

common/rpc/interceptor/metadata_context.go
bergundy: Do you need to inject this into those services? I thought we would only need it in history.

service/history/workflow/mutable_state_impl.go
bergundy: That's fine with me, what you have is more flexible.

PR #8989, opened 2026-01-09
Fix the bug fix: Don't put nil elements in the list of Worker Deployment Summaries

Shivs11:
## What changed?
- WISOTT

## Why?
Commit [2072e703ff65059bc7fa8caab1349dba3f910907](https://github.com/temporalio/temporal/commit/2072e703ff65059bc7fa8caab1349dba3f910907) from this morning made it so that nil objects could be in the list of deployment summaries. The consumer of this list expects no nil objects. Changing to append instead of element assignment fixes this issue.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
Risk that this doesn't actually solve the issue, but we believe it does. Concurrently to this merging, @carlydf will write a test so that we don't rely on VCR to generate the racy failure case.

---

> [!NOTE]
> Makes `ListWorkerDeployments` resilient to missing/invalid workflow memos.
> 
> - Switches from pre-sized indexing to `append` when building `workerDeploymentSummaries`, so entries are only added when valid (prevents nil/empty slots on decode errors or race before memo upsert)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ea4887ebfb9cdbb0177bf577c1a6f0b46e6d54a8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -3621,16 +3621,16 @@ func (wh *WorkflowHandler) ListWorkerDeployments(ctx context.Context, request *w
 		return nil, err
 	}
 
-	workerDeployments := make([]*workflowservice.ListWorkerDeploymentsResponse_WorkerDeploymentSummary, 0, len(resp))
-	for _, d := range resp {
-		workerDeployments = append(workerDeployments, &workflowservice.ListWorkerDeploymentsResponse_WorkerDeploymentSummary{
-			Name:                  d.GetName(),
-			CreateTime:            d.GetCreateTime(),
-			RoutingConfig:         d.GetRoutingConfig(),
-			LatestVersionSummary:  d.GetLatestVersionSummary(),
-			RampingVersionSummary: d.GetRampingVersionSummary(),
-			CurrentVersionSummary: d.GetCurrentVersionSummary(),
-		})
+	workerDeployments := make([]*workflowservice.ListWorkerDeploymentsResponse_WorkerDeploymentSummary, len(resp))
+	for i, d := range resp {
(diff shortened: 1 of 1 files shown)


service/worker/workerdeployment/client.go
```diff
@@ -591,8 +591,8 @@ func (d *ClientImpl) ListWorkerDeployments(
 		return nil, nil, err
 	}
 
-	workerDeploymentSummaries := make([]*deploymentspb.WorkerDeploymentSummary, len(persistenceResp.Executions))
-	for i, ex := range persistenceResp.Executions {
+	workerDeploymentSummaries := make([]*deploymentspb.WorkerDeploymentSummary, 0, len(persistenceResp.Executions))
```
spkane31, line 594: If you're doing the append then you shouldn't do the `make` here

spkane31:
```suggestion
	var workerDeploymentSummaries []*deploymentspb.WorkerDeploymentSummary{}
```

carlydf: I think it's fine, because the make statement is creating a slice with length 0 and capacity `len(persistenceResp.Executions)`, so it's an empty slice (but the compiler just pre-allocates `len(persistenceResp.Executions)` of space instead of having to allocate space for a new slice on each append). I had to look this up to confirm, since I haven't used the capacity parameter much, but I think this is correct as is. See: https://go.dev/blog/slices-intro#slices

spkane31: Good catch!

PR #8991, opened 2026-01-09
Ignore unknown fields when decoding Worker Deployment memo for backwards compatibility

carlydf:
## What changed?
Ignore unknown fields when decoding Worker Deployment memo for backwards compatibility 

## Why?
New field was added to memo. We want to be able to roll back server and still list deployments, just without that field

## How did you test it?
Tested it with v148. Also cherry-picked onto v147.6 and ran my unit test. With the new Lenient converter, the unit test passed as expected. With the old strict converter, the unit test failed as expected.

- [x] built
- [x] added new unit test(s)

## Potential risks
Changing the data converter is risky, but in this case, we know we are decoding a proto, and we have tests covering that decoding, so I think it's ok

---

> [!NOTE]
> Makes JSON/protobuf memo decoding lenient to support backward compatibility and adds tests to verify decoding behavior.
> 
> - Update `PreferProtoDataConverter` to use `NewProtoJSONPayloadConverterWithOptions(AllowUnknownFields: true)`, causing proto JSON decoding to ignore unknown fields
> - Add `TestDecodeWorkerDeploymentMemoTolerateUnknownFields` and `TestDecodeWorkerDeploymentMemoTolerateMissingFields` with memo JSON fixtures to validate `DecodeWorkerDeploymentMemo` succeeds and parses known fields
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0d1ea46134d13fd21aaf81da7e97400748fd01a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/worker/workerdeployment/testdata/memo_with_last_current_time.json
```diff
@@ -0,0 +1,18 @@
+{
```
Shivs11, line 1: nice!

do we also want to have a test where the new field is not there and maybe an upgraded server then tries to decode a worker deployment workflow's memo with an outdated memo?

common/sdk/converter.go
```diff
@@ -15,3 +20,22 @@ var (
...
+	if payload == nil {
+		return nil
+	}
+
+	encoding := string(payload.GetMetadata()["encoding"])
+	switch encoding {
+	case "binary/protobuf":
+		return proto.UnmarshalOptions{DiscardUnknown: true}.Unmarshal(payload.GetData(), valuePtr)
+	case "json/protobuf":
+		return temporalproto.CustomJSONUnmarshalOptions{DiscardUnknown: true}.Unmarshal(payload.GetData(), valuePtr)
```
Shivs11: I'm thinking that the DiscardUnknown does the trick here? 
cc - @dnr would love a peek here just to be sure about things

service/worker/workerdeployment/testdata/memo_with_last_current_time.json
carlydf: done!

common/sdk/converter.go
```diff
@@ -15,3 +20,22 @@ var (
...
 	)
 )
+
+// LenientFromPayloadProtoConverter decodes a payload to a proto message, discarding unknown fields.
+// This is useful for forward compatibility when older code needs to decode payloads
+// that may have been written by newer code with additional fields.
+func LenientFromPayloadProtoConverter(payload *commonpb.Payload, valuePtr proto.Message) error {
+	if payload == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
carlydf: removed

common/sdk/converter.go
```diff
@@ -14,4 +14,15 @@ var (
 		converter.NewProtoJSONPayloadConverter(),
 		converter.NewJSONPayloadConverter(),
 	)
+
+	// PreferProtoDataConverterAllowUnknownJSONFields is like the default data converter defined in the SDK, except
+	// that it prefers encoding proto messages with the binary encoding instead of json, and, if decoding with json,
+	// allows unknown fields.
+	PreferProtoDataConverterAllowUnknownJSONFields = converter.NewCompositeDataConverter(
```
dnr: Just change the existing converter, don't make a new thing that'll just add confusion as some thing use it but not others

common/sdk/converter.go
```diff
@@ -14,4 +14,15 @@ var (
...
+		converter.NewNilPayloadConverter(),
+		converter.NewByteSlicePayloadConverter(),
+		converter.NewProtoPayloadConverter(),
+		converter.NewProtoJSONPayloadConverterWithOptions(converter.ProtoJSONPayloadConverterOptions{AllowUnknownFields: true}),
+		converter.NewProtoJSONPayloadConverterWithOptions(converter.ProtoJSONPayloadConverterOptions{AllowUnknownFields: true}),
```
dnr: duplicated?

carlydf: yup, my mistake, I should have reviewed that better

common/sdk/converter.go
carlydf: done

service/worker/workerdeployment/testdata/memo_with_last_current_time.json
dnr, line 1: Call the directory just `testdata`, that's the convention

PR #8993, opened 2026-01-10
Make min namespace retention configurable

yycptt:
## What changed?
- Make min namespace retention configurable

## Why?
- In functional tests and test cluster, we will want to use a short retention period to validate retention timer is working properly.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -187,12 +187,12 @@ config as the other services.`,
 	NamespaceMinRetentionGlobal = NewGlobalDurationSetting(
 		"system.namespaceMinRetentionGlobal",
 		24*time.Hour,
-		`NamespaceMinRetentionGlobal is the minimum retention duration for global namespaces. This value should only be lowered for testing purposes.`,
+		`Minimum retention duration for global namespaces. This value should only be lowered for testing purposes.`,
 	)
 	NamespaceMinRetentionLocal = NewGlobalDurationSetting(
 		"system.namespaceMinRetentionLocal",
 		time.Hour,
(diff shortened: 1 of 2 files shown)


common/dynamicconfig/constants.go
```diff
@@ -184,6 +184,16 @@ config as the other services.`,
 		false,
 		`EnableActivityEagerExecution indicates if activity eager execution is enabled per namespace`,
 	)
+	NamespaceMinRetentionGlobal = NewGlobalDurationSetting(
+		"system.namespaceMinRetentionGlobal",
+		24*time.Hour,
+		`NamespaceMinRetentionGlobal is the minimum retention duration for global namespaces. This value should only be lowered for testing purposes.`,
```
bergundy: No need to document these like go docstrings.

```suggestion
		`Minimum retention duration for global namespaces. This value should only be lowered for testing purposes.`,
```

service/frontend/namespace_handler.go
```diff
@@ -1057,16 +1057,16 @@ func (d *namespaceHandler) maybeUpdateFailoverHistory(
...
 	if err := timestamp.ValidateAndCapProtoDuration(retention); err != nil {
 		return errInvalidRetentionPeriod
 	}
 
-	min := namespace.MinRetentionLocal
+	minRetention := d.config.NamespaceMinRetentionLocal()
 	if isGlobalNamespace {
-		min = namespace.MinRetentionGlobal
+		minRetention = d.config.NamespaceMinRetentionGlobal()
 	}
```
bergundy: nit: better only read the config in the body of an if statement.

bergundy (approved):

PR #8996, opened 2026-01-10
[Scheduler] Compute FutureActionTimes synchronously on Describe, when needed

lina-temporal:
## What changed?
- When `DescribeSchedule` is called, it will compute `FutureActionTimes` synchronously if it hasn't already been computed.

## Why?
- `FutureActionTimes` is populated via the `GeneratorTask`, which means there's a short window of time before the task has run where the field has not been populated. This computes the field synchronously in that case.
- No change is needed for `ListInfo`, since that's updated asynchronously (visibility) regardless.


chaptersix:
Could you add an assertion that check for future actions being populated in describe requests?

PR #9002, opened 2026-01-12
System nexus endpoint

bergundy:
## Overview

This feature introduces a special internal Nexus endpoint called `__temporal_system` that enables adding functionality to workflows **without requiring new workflow commands and events**. Operations on this endpoint are routed internally within Temporal's history service rather than via external HTTP calls.

### 1. System Endpoint Infrastructure

- **Endpoint Name**: `__temporal_system` (constant in `common/nexus/constants.go:6`)
- **Callback URL**: `temporal://system` for internal routing
- **New History Service RPCs** (`proto/internal/temporal/server/api/historyservice/v1/service.proto:433-437`):
  - `StartNexusOperation` - Starts operations on the system endpoint
  - `CancelNexusOperation` - Cancels operations on the system endpoint

### 2. Operation Processor Framework

A new processor pattern (`chasm/nexus_operation_processor.go`) that allows CHASM libraries to:

- **Validate and transform input**: Processors can validate operation inputs and set default values
- **Determine routing**: Each processor returns a routing key that determines which history shard handles the operation
- **Re-serialize input**: Mutated inputs can be re-serialized to persist default values

**Routing strategies**:
- `NexusOperationRoutingKeyExecution` - Routes to the shard owning a specific workflow execution
- `NexusOperationRoutingKeyRandom` - Routes to a random shard

### 3. CHASM Library Integration

CHASM libraries can now provide (`chasm/library.go:16-19`):
- **`NexusServices()`**: Regular Nexus service handlers (implement the actual operation logic)
- **`NexusServiceProcessors()`**: Input processors for validation and routing

Example from test library (`chasm/lib/tests/nexus_service.go`):

```go
// Service handler - implements the actual operation
TestOperation = nexus.NewSyncOperation("TestOperation",
    func(ctx context.Context, input string, options nexus.StartOperationOptions) (string, error) {
        return "Hello, " + input, nil
    })

// Processor - validates input and determines routing
func (o testOperationProcessor) ProcessInput(ctx chasm.NexusOperationProcessorContext, input string)
    (*chasm.NexusOperationProcessorResult, error) {
    return &chasm.NexusOperationProcessorResult{
        RoutingKey: chasm.NexusOperationRoutingKeyExecution{
            NamespaceID: ctx.Namespace.ID().String(),
            BusinessID:  input, // Route based on input
        },
    }, nil
}
```

### 4. Execution Flow

When a workflow schedules a Nexus operation on `__temporal_system` (`components/nexusoperations/executors.go:233-238`):

1. **Input Processing**: The processor validates input and determines routing
2. **Internal RPC**: Instead of HTTP, calls `HistoryClient.StartNexusOperation` with the target shard ID
3. **Handler Execution**: The history service invokes the registered Nexus handler (`service/history/handler.go:2707-2768`)
4. **Result Handling**: Supports both sync (immediate result) and async (operation token) responses
5. **Workflow Completion**: Results flow back through the same completion path as external Nexus operations

### 5. Benefits

✅ **No schema changes**: Add functionality without new commands/events in workflow history
✅ **Consistent API**: Uses existing Nexus operation semantics (sync/async, callbacks, links)
✅ **Proper routing**: Operations are intelligently routed to the correct shard
✅ **Input validation**: Type-safe input validation and default value handling
✅ **Future extensibility**: Foundation for direct client invocation (not yet implemented)

### 6. Technical Details

- **Error handling** (`components/nexusoperations/executors.go:444`): Non-retryable service errors are properly handled and fail operations immediately
- **Metrics**: System operations are tracked separately with `DestinationTag` set to the endpoint name
- **Link conversion**: Helper functions convert between Nexus SDK links and protobuf links (`common/nexus/util.go:17-46`)
- **Operation token handling**: Moved link converters to common package for reuse (`common/nexus/link_converter.go`)

### 7. Current Limitations

- Only accessible from workflows (via `ScheduleNexusOperation` command)
- Direct client invocation not yet implemented
- Headers not supported for system endpoint operations

## Test Coverage

New test (`tests/nexus_workflow_test.go:2763-2843`) demonstrates:
- Scheduling operation on `__temporal_system` endpoint
- Synchronous operation completion
- Result propagation back to workflow

## Architecture

This architecture provides a clean, extensible way to add internal functionality while maintaining compatibility with Temporal's existing workflow execution model. The system endpoint acts as a bridge between workflows and internal CHASM components, enabling:

- **Extensibility**: New operations can be added by implementing CHASM libraries
- **Type safety**: Input validation happens before operations are routed
- **Scalability**: Intelligent routing ensures operations land on the correct shard
- **Consistency**: Same execution model as external Nexus operations

### DIFF
--- a/components/nexusoperations/executors.go
+++ b/components/nexusoperations/executors.go
@@ -278,7 +278,7 @@ func (e taskExecutor) executeInvocationTask(ctx context.Context, env hsm.Environ
 		result, callErr = e.startOnHistoryService(callCtx, ns, args, options)
 	} else {
 		client, err := e.ClientProvider(
-			ctx,
+			callCtx,
 			ns.ID().String(),
 			endpoint,
 			args.service,
--- a/client/history/historytest/clienttest.go
+++ b/client/history/historytest/clienttest.go
@@ -15,6 +15,7 @@ import (
 	"go.opentelemetry.io/otel/trace/embedded"
 	commonspb "go.temporal.io/server/api/common/v1"
 	"go.temporal.io/server/api/historyservice/v1"
+	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/client/history"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/log"
@@ -139,6 +140,7 @@ func createServer(historyTaskQueueManager persistence.HistoryTaskQueueManager) *
 		TaskQueueManager:     historyTaskQueueManager,
 		TracerProvider:       fakeTracerProvider{},
 		TaskCategoryRegistry: tasks.NewDefaultTaskCategoryRegistry(),
+		ChasmRegistry:        chasm.NewRegistry(log.NewNoopLogger()),
 	})
 	if err != nil {
 		panic(err) // nolint:forbidigo // Panic is acceptable in test setup code.
--- a/common/persistence/visibility/chasm_visibility_manager_test.go
+++ b/common/persistence/visibility/chasm_visibility_manager_test.go
@@ -85,6 +85,8 @@ func (s *ChasmVisibilityManagerSuite) SetupTest() {
 		),
 	}).AnyTimes()
 	library.EXPECT().Tasks().Return(nil).AnyTimes()
+	library.EXPECT().NexusServices().Return(nil).AnyTimes()
+	library.EXPECT().NexusServiceProcessors().Return(nil).AnyTimes()
 
 	s.registry = chasm.NewRegistry(log.NewNoopLogger())
 	err := s.registry.Register(library)
@@ -421,6 +423,8 @@ func (s *ChasmVisibilityManagerSuite) TestListExecutions_WithTaskQueueSearchAttr
 		),
 	}).AnyTimes()
 	library.EXPECT().Tasks().Return(nil).AnyTimes()
+	library.EXPECT().NexusServices().Return(nil).AnyTimes()
+	library.EXPECT().NexusServiceProcessors().Return(nil).AnyTimes()
 
 	registry := chasm.NewRegistry(log.NewNoopLogger())
 	err := registry.Register(library)
--- a/common/persistence/visibility/store/elasticsearch/visibility_store_read_test.go
+++ b/common/persistence/visibility/store/elasticsearch/visibility_store_read_test.go
@@ -154,6 +154,8 @@ func (s *ESVisibilitySuite) SetupTest() {
 	)
 	library.EXPECT().Components().Return([]*chasm.RegistrableComponent{rc}).AnyTimes()
 	library.EXPECT().Tasks().Return(nil).AnyTimes()
+	library.EXPECT().NexusServices().Return(nil).AnyTimes()
+	library.EXPECT().NexusServiceProcessors().Return(nil).AnyTimes()
 	s.chasmRegistry = chasm.NewRegistry(log.NewNoopLogger())
 	err := s.chasmRegistry.Register(library)
 	s.NoError(err)
--- a/service/history/fx.go
+++ b/service/history/fx.go
@@ -157,6 +157,9 @@ func HandlerProvider(args NewHandlerArgs) (*Handler, error) {
 
 func buildNexusHandler(chasmRegistry *chasm.Registry) (nexus.Handler, error) {
 	nexusServices := chasmRegistry.NexusServices()
+	if len(nexusServices) == 0 {
+		return nil, nil
+	}
 	serviceRegistry := nexus.NewServiceRegistry()
 	for _, svc := range nexusServices {
 		// No chance of collision here since the registry would have errored out earlier.
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -2504,6 +2504,9 @@ func (h *Handler) StartNexusOperation(
 	})
 
 	// Invoke the operation via the handler
+	if h.nexusHandler == nil {
+		return nil, serviceerror.NewUnimplemented("no nexus services registered")
+	}
 	result, err := h.nexusHandler.StartOperation(ctx, req.GetRequest().GetService(), req.GetRequest().GetOperation(), input, options)
 	if err != nil {
 		var opErr *nexus.OperationError
@@ -2584,6 +2587,9 @@ func (h *Handler) CancelNexusOperation(
 	})
 
 	// Invoke the cancel operation via the handler
+	if h.nexusHandler == nil {
+		return nil, serviceerror.NewUnimplemented("no nexus services registered")
+	}
 	err := h.nexusHandler.CancelOperation(ctx, req.GetRequest().GetService(), req.GetRequest().GetOperation(), req.GetRequest().GetOperationToken(), options)
 	if err != nil {
 		// TODO: redact certain errors


proto/internal/temporal/server/api/historyservice/v1/service.proto
```diff
@@ -429,4 +428,12 @@ service HistoryService {
     // UnpauseWorkflowExecution unpauses the workflow execution specified in the request.
     rpc UnpauseWorkflowExecution (UnpauseWorkflowExecutionRequest) returns (UnpauseWorkflowExecutionResponse) {
     }
+
+    // StartNexusOperation starts a Nexus operation on the __temporal_system endpoint.
+    rpc StartNexusOperation (StartNexusOperationRequest) returns (StartNexusOperationResponse) {
```
pdoerner, line 433: Do we think we will add gRPC support for normal Nexus operations? It makes me nervous to use this name when it is locked to the system endpoint. I would suggest something like `StartSystemNexusOperation`

components/nexusoperations/workflow/commands.go
```diff
@@ -52,23 +55,52 @@ func (ch *commandHandler) HandleScheduleCommand(
 		}
 	}
 
+	requestID := uuid.NewString()
 	var endpointID string
-	endpoint, err := ch.endpointRegistry.GetByName(ctx, ns.ID(), attrs.Endpoint)
-	if err != nil {
-		if errors.As(err, new(*serviceerror.NotFound)) {
+	// Skip endpoint registry lookup for __temporal_system endpoint
+	if attrs.Endpoint == commonnexus.SystemEndpoint {
```
pdoerner, line 61: I think it might be a bit cleaner to have a static `persistence.NexusEndpointEntry` and push the `if endpointName == commonnexus.SystemEndpoint` down into the registry which can return the static entry in that case. That way the system endpoint can be customized by injection and we need less complicated error handling.

components/nexusoperations/executors.go
```diff
@@ -198,93 +212,81 @@ func (e taskExecutor) executeInvocationTask(ctx context.Context, env hsm.Environ
...
-		rawResult, callErr = client.StartOperation(callCtx, args.operation, args.payload, nexus.StartOperationOptions{
-			Header:      header,
-			CallbackURL: callbackURL,
-			RequestID:   args.requestID,
-			CallbackHeader: nexus.Header{
-				commonnexus.CallbackTokenHeader: token,
-			},
-			Links: []nexus.Link{args.nexusLink},
-		})
+		client, err := e.ClientProvider(
```
pdoerner, line 280: nit: I would move this HTTP client / trace logic into `startViaHTTP` (same with cancel)

components/nexusoperations/executors.go
```diff
@@ -688,6 +720,7 @@ func (e taskExecutor) loadArgsForCancelation(ctx context.Context, env hsm.Enviro
 		}
 		if attrs := event.GetNexusOperationScheduledEventAttributes(); attrs != nil {
 			args.headers = attrs.GetNexusHeader()
+			args.payload = attrs.GetInput()
```
pdoerner, line 788: Why accept an input for cancellation? Technically it is a deviation from the Nexus spec

common/nexus/constants.go
```diff
@@ -3,3 +3,7 @@ package nexus
 // SystemCallbackURL is the reserved callback URL used to route Nexus operation callbacks
 // internally within Temporal. It must match the scheme/host used in validation and routing logic.
 const SystemCallbackURL = "temporal://system"
+
+// SystemEndpoint is the reserved endpoint name for Temporal system operations.
+// Operation requests for this endpoint are routed internally within the history service.
+const SystemEndpoint = "__temporal_system"
```
pdoerner, line 9: We need a check to prevent this endpoint name from being registered by users. Probably in `frontend/nexus_endpoint_client`

components/nexusoperations/executors.go
```diff
@@ -198,93 +212,81 @@ func (e taskExecutor) executeInvocationTask(ctx context.Context, env hsm.Environ
...
+	var destTag metrics.Tag
+	if endpoint != nil {
+		destTag = metrics.DestinationTag(endpoint.Endpoint.Spec.GetName())
+	} else {
+		destTag = metrics.DestinationTag(args.endpointName)
+	}
+	outcomeTag := metrics.OutcomeTag(startCallOutcomeTag(callCtx, result, callErr))
 	failureSourceTag := metrics.FailureSourceTag(failureSourceFromContext(callCtx))
 	OutboundRequestCounter.With(e.MetricsHandler).Record(1, namespaceTag, destTag, methodTag, outcomeTag, failureSourceTag)
 	OutboundRequestLatency.With(e.MetricsHandler).Record(time.Since(startTime), namespaceTag, destTag, methodTag, outcomeTag, failureSourceTag)
```
pdoerner: Should system Nexus operations share the same set of metrics as regular Nexus operations?

chasm/lib/tests/library.go
```diff
@@ -14,6 +17,14 @@ func (l *library) Name() string {
 	return "tests"
 }
 
+func (l *library) NexusServices() []*nexus.Service {
```
pdoerner, line 20: Do we expect CHASM components will want to register more than one Nexus service? This feels like something the framework could handle. IMHO it will be easier for CHASM component implementers to think about Operations for each of the APIs they will expose. Component authors could specify a set of Nexus Operations and the framework could handle registering those as a Nexus Service named by the component.

chasm/library.go
```diff
@@ -12,6 +13,11 @@ type (
 		Components() []*RegistrableComponent
 		Tasks() []*RegistrableTask
 		RegisterServices(server *grpc.Server)
+		// NexusServices returns a list of nexus.Service instances to register with the __temporal_system Nexus endpoint.
+		NexusServices() []*nexus.Service
+		// NexusServiceProcessors returns a list of NexusServiceProcessor instances to register with the __temporal_system
+		// Nexus endpoint.
+		NexusServiceProcessors() []*NexusServiceProcessor
```
yycptt, line 20: I feel they should be one method, so that we can guarantee their count matches?

common/nexus/constants.go
bergundy: We already enforce a specific regexp, users can't set the endpoint to anything that starts with an underscore.

components/nexusoperations/workflow/commands.go
bergundy: But we'll have to implement that in every registry implementation so I don't see much of a reason to move the logic to the registry.

proto/internal/temporal/server/api/historyservice/v1/service.proto
bergundy: Yeah, eventually when we add direct Nexus calls but it will probably be the same API to dispatch to a worker or to history.

chasm/lib/tests/library.go
bergundy: It's not components, it's libraries and a library should be able to register as many services or tasks or components as it needs.

components/nexusoperations/executors.go
bergundy: Hmm.. but the start time is recorded after we get the client.

components/nexusoperations/executors.go
bergundy: It's not sent with the request, it's just to resolve the shard.

components/nexusoperations/executors.go
bergundy: I think it's fine if we qualify with the endpoint label.

chasm/library.go
bergundy: No we can't because the processor will need to be instantiated on the frontend when we support direct calls.

proto/internal/temporal/server/api/historyservice/v1/request_response.proto
```diff
@@ -1370,3 +1370,27 @@ message UnpauseWorkflowExecutionRequest {
 }
 
 message UnpauseWorkflowExecutionResponse { }
+
+message StartNexusOperationRequest {
+    option (routing).shard_id = "shard_id";
+
+    string namespace_id = 1;
+    int32 shard_id = 2;
```
chaptersix, line 1409: Would these request always be routed within a cluster? Could the target shard change if the active cluster for a namespace changes?

chaptersix: is the caller namespace always the destination namespace?

bergundy: It's always within the cluster.

chaptersix (approved):

PR #9004, opened 2026-01-12
Fix stale internode gRPC connections after pod termination

prathyushpv:
## What changed?
When Kubernetes pods are terminated, the cached gRPC connections in interNodeGrpcConnections remain stale, causing continuous "dial tcp ... i/o timeout" errors until the service is restarted. 
                                                                            
This fix adds a membership listener to RPCFactory that:                                                                                                                             
- Subscribes to membership ring change events for the history service                                                                                                           
- Evicts cached connections when hosts are removed from the ring                                                                                                                                          
- Properly closes evicted connections via cache RemovedFunc callback

## Why?
To avoid repeated dial attempts to stale pod IPs

## How did you test it?
- [x] built
- [x] added new unit test(s)


alfred-landrum:
👋 The history client has an internal GRPC connection pool, see client/history/connections.go. That predates your work to add a cache inside RPC factory https://github.com/temporalio/temporal/pull/9004 .  Does the history client connection pool need to be removed to work with the changes in this PR?

prathyushpv:
Hi @alfred-landrum, Thanks for pointing this out! I just checked how history cache works. I see that all caches share same *grpc.ClientConn pointer.  When HandleMembershipChange() in RPCFactory calls conn.Close(), it closes the actual gRPC connection object that all caches reference. After this, operations on this connection should return Unavailable error instead of timeout error. Caching redirector should evict the cache entry at this point: https://github.com/temporalio/temporal/blob/b6415c6034b38aef6a204f814cd6290876672e5a/client/history/caching_redirector.go#L108-L108

I see that caching redirector also listens for membership change events here: https://github.com/temporalio/temporal/blob/b6415c6034b38aef6a204f814cd6290876672e5a/client/history/caching_redirector.go#L244-L244

But it waits for staleTTL time duration before removing the cache entry. What is the reason for this wait?

alfred-landrum:
> But it waits for staleTTL time duration before removing the cache entry. What is the reason for this wait?

It's to reduce latency for operations during shard movements. When the history service is told to stop, it leaves membership as a first step, then depending on config, waits a bit for connections & shard ownership to move. Even though it has left membership, the shards it owned may not have been acquired by their new owners. By using the cached entry for a small duration, we allow the new owner time to come up & initialize shards. (And it will then fence out requests from the old owner - so any requests to it will return shard ownership lost.)

wensiet:
Hi! Is there a chance that this PR will be included in the next release?

andyrpi2012:
> Hi @alfred-landrum, Thanks for pointing this out! I just checked how history cache works. I see that all caches share same *grpc.ClientConn pointer. When HandleMembershipChange() in RPCFactory calls conn.Close(), it closes the actual gRPC connection object that all caches reference. After this, operations on this connection should return Unavailable error instead of timeout error. Caching redirector should evict the cache entry at this point:
> 
> https://github.com/temporalio/temporal/blob/b6415c6034b38aef6a204f814cd6290876672e5a/client/history/caching_redirector.go#L108-L108
> 
> I see that caching redirector also listens for membership change events here:
> 
> https://github.com/temporalio/temporal/blob/b6415c6034b38aef6a204f814cd6290876672e5a/client/history/caching_redirector.go#L244-L244
> 
> But it waits for staleTTL time duration before removing the cache entry. What is the reason for this wait?

Hi @prathyushpv, @alfred-landrum your conversation on this staled grpc connection is great. This issue also bothers our team for weeks.

Here is my thought here. It will be great if you can chime in.

1. The Root Cause of the Leak
the issue was introduced by the commit https://github.com/temporalio/temporal/commit/5919da32c, which added the interNodeGrpcConnections cache. It is way before 1.28.0 version. It was on Aug 30, 2024

2. Why it wasn't a visible problem in 1.28?
on Sep 4 2025, the commit https://github.com/temporalio/temporal/commit/b5253b3, added the gRPC dial instrumentation that made the leak visible.

It will be great if you can share estimated timeline for this PR to be in the future release.

yycptt:
> it waits for staleTTL time duration before removing the cache entry.

so we will need the same thing in this PR as well to avoid breaking existing history shard graceful handover story?
cc @prathyushpv @alfred-landrum

a1101992391-ux:
When will this issue be fixed

prathyushpv:
Closing this. I will fix this issue in #9277 and #10250

PR #9006, opened 2026-01-12
fix flaky TaskQueueStatsSuite

stephanos:
WISOTT

### DIFF
--- a/tests/task_queue_stats_test.go
+++ b/tests/task_queue_stats_test.go
@@ -47,11 +47,9 @@ type (
 	}
 
 	TaskQueueExpectations struct {
-		BacklogCount     int
-		MaxExtraTasks    int
-		ExpectedAddRate  bool
-		ExpectedDispatch bool
-		CachedEnabled    bool
+		BacklogCount  int
+		MaxExtraTasks int
+		CachedEnabled bool
 	}
 
 	// TaskQueueExpectationsByType maps task queue types to their expectations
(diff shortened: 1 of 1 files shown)


tests/task_queue_stats_test.go
```diff
@@ -1391,6 +1365,45 @@ func (s *TaskQueueStatsSuite) validateAllTaskQueueStats(
 	}
 }
 
+// validateRates verifies TasksAddRate and/or TasksDispatchRate in a dedicated EventuallyWithT block.
+// This should be called immediately after the relevant operation (enqueue for add rate, poll for dispatch rate)
+// to ensure the rate is checked while still fresh (before the 30-second sliding window decays).
```
stephanos, line 1399: This explains the hypothesis behind this change. The hope is that checking earlier now will reduce flakiness.

tests/task_queue_stats_test.go
```diff
@@ -47,11 +47,9 @@ type (
 	}
```
Shivs11, line 47: nit: I think it would be super nice to know which tests were flaking! was it the whole suite?

Shivs11 (approved):

tests/task_queue_stats_test.go
stephanos: I was focused on `TestTaskQueueStats_Pri_Suite/TestMultipleTasks_MultiplePartitions_WithMatchingBehavior_ValidateStats/ForceTaskForwardForcePollForwardAllowSync` but generally all the tests with several partitions where at risk (as they take much longer than single partitions).

PR #9010, opened 2026-01-13
Nexus caller timeouts

bergundy:
## Overview

This commit implements two new granular timeout types for Nexus operations, allowing callers to have fine-grained control over different phases of operation execution:

- **Schedule-to-Start Timeout**: Maximum time to wait for an operation to be started (or completed if synchronous) by the handler
- **Start-to-Close Timeout**: Maximum time to wait for an asynchronous operation to complete after it has been started

These timeouts complement the existing **Schedule-to-Close Timeout** to provide better control and diagnostics for Nexus operation execution.

See the corresponding API PR: https://github.com/temporalio/api/pull/695.

### DIFF
--- a/components/nexusoperations/executors_test.go
+++ b/components/nexusoperations/executors_test.go
@@ -332,7 +332,7 @@ func TestProcessInvocationTask(t *testing.T) {
 				require.Equal(t, enumsspb.NEXUS_OPERATION_STATE_BACKING_OFF, op.State())
 				require.NotNil(t, op.LastAttemptFailure.GetApplicationFailureInfo())
 				require.Regexp(t, "request timed out", op.LastAttemptFailure.Message)
-				require.Equal(t, 0, len(events))
+				require.Empty(t, events)
 			},
 		},
 		{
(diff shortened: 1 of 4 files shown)


components/nexusoperations/workflow/commands.go
```diff
@@ -152,6 +168,19 @@ func (ch *commandHandler) HandleScheduleCommand(
 		attrs.ScheduleToCloseTimeout = durationpb.New(maxTimeout)
 	}
 
+	// Trim secondary timeouts to the primary timeout.
```
pdoerner, line 171: Should this also go the other way? If only schedule-to-start or start-to-close is set, but not schedule-to-close, should we set schedule-to-close to one of those timeouts?

IIRC the operation deadline we send to the handler is based only on schedule-to-close timeout, so the handler may not be aware that the operation has a deadline in that case.

bergundy: It shouldn't go the other way because the new timeouts are more specific. But you make a good point that the operation timeout header should take the new timeouts into consideration.

bergundy: Done.

PR #9015, opened 2026-01-13
Avoid resetting activity info TimerTaskStatus for each retry

bergundy:
## What changed?

In PR description.

## Why?

Avoid creating extra timers where unnecessary.

## How did you test it?

Should be covered by existing tests.

### DIFF
--- a/service/history/workflow/activity.go
+++ b/service/history/workflow/activity.go
@@ -67,7 +67,7 @@ func UpdateActivityInfoForRetries(
 	failure *failurepb.Failure,
 	nextScheduledTime *timestamppb.Timestamp,
 	isActivityRetryStampIncrementEnabled bool,
-) *persistencespb.ActivityInfo {
+) {
 	previousAttempt := ai.Attempt
 	ai.Attempt = attempt
 	ai.Version = version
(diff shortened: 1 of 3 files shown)


service/history/workflow/activity.go
```diff
@@ -76,7 +76,6 @@ func UpdateActivityInfoForRetries(
 	ai.StartVersion = common.EmptyVersion
 	ai.RequestId = ""
 	ai.StartedTime = nil
-	ai.TimerTaskStatus = TimerTaskStatusNone
```
yycptt, line 79: we do want to reset say StartToClose or heartbeat timer task status right? 
ScheduleToStart is interesting, I think we always extend that to be same as ScheduleToClose?

bergundy: Hmm... schedule to start is set per attempt

yycptt: hmm I somehow thought we have logic for extending ScheduleToStart to be the same as ScheduleToClose if activity has retry policy because retrying scheduleToStart timeout is basically noop. 

but hmm can't really find that logic 🤷‍♂️

PR #9018, opened 2026-01-14
fix race condition of workflow rebuild

feiyang3cat:
## What changed?
Fix race conditions in rebuild workflows:
- clear mutableState before release lock
- query rebuildSpec after lock acquired

## Why?
fix race condition

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


feiyang3cat:
@yycptt please have a look

yycptt (commented):
Synced offline, let's check if we any some functional test coverage for this.

PR #9019, opened 2026-01-14
Improve replication task processing latency log and metrics

xwduan:
## What changed?
Improve replication task processing latency log and metrics:
1. add replication task queue time latency
2. add task attempt
3. add task error metrics

## Why?
To help debug issue when latency is high

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
No risk.

### DIFF
--- a/service/history/replication/executable_task.go
+++ b/service/history/replication/executable_task.go
@@ -336,13 +336,6 @@ func (e *ExecutableTaskImpl) emitFinishMetrics(
 			nsTag,
 		)
 		if processingLatency > 10*time.Second && e.replicationTask != nil && e.replicationTask.RawTaskInfo != nil {
-			shardContext, err := e.ShardController.GetShardByNamespaceWorkflow(
-				namespace.ID(e.replicationTask.RawTaskInfo.NamespaceId),
-				e.replicationTask.RawTaskInfo.WorkflowId,
-			)
-			if err != nil {
-				return
-			}
 			e.Logger.Warn(fmt.Sprintf(
 				"replication task latency is too long: queue=%.2fs processing=%.2fs",
 				queueLatency.Seconds(),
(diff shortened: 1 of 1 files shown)


service/history/replication/executable_task.go
```diff
@@ -301,26 +316,48 @@ func (e *ExecutableTaskImpl) emitFinishMetrics(
...
+
+		// Processing latency: time from execution start to ACK/NACK
+		processingLatency := now.Sub(e.taskExecuteStartTime)
+		metrics.ReplicationTaskProcessingLatency.With(e.MetricsHandler).Record(
+			processingLatency,
+			metrics.OperationTag(e.metricsTag),
+			nsTag,
+		)
+		if processingLatency > 10*time.Second && e.replicationTask != nil && e.replicationTask.RawTaskInfo != nil {
+			shardContext, err := e.ShardController.GetShardByNamespaceWorkflow(
```
yux0: Use e.Config.GetShardID()?

service/history/replication/executable_task.go
```diff
@@ -301,26 +316,48 @@ func (e *ExecutableTaskImpl) emitFinishMetrics(
...
+				namespace.ID(e.replicationTask.RawTaskInfo.NamespaceId),
+				e.replicationTask.RawTaskInfo.WorkflowId,
+			)
+			if err != nil {
+				return
+			}
+			e.Logger.Warn(fmt.Sprintf(
```
yux0, line 339: will this print more logs with threshold reduced to 10s?

yux0: DI the throttled logger?

xwduan: We only print the log when processing time is greater than 10 seconds. This is real processing time i.e. excluding queue time/transmission time. So I would expect less log than before.

I will DI throttled log in a separate pr.

yux0 (approved):

PR #9020, opened 2026-01-14
Improve error when version override does not have the version present in the task queue

Shivs11:
## What changed?
- WISOTT

## Why?
- User experience.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
- None

---

> [!NOTE]
> Enhances error clarity and batch behavior around version overrides.
> 
> - Introduces `ErrPinnedVersionNotInTaskQueueSubstring` and `FormatPinnedVersionNotInTaskQueueError` to produce detailed messages (deployment, build ID, task queue, type)
> - Replaces generic errors in `validatePinnedVersionInTaskQueue` with formatted message; updates related tests to expect new text
> - Adds operation-aware non-retryable detection in `batcher` (`isNonRetryableError`) for `BATCH_OPERATION_TYPE_UPDATE_EXECUTION_OPTIONS` using the new substring; integrates into `handleTaskResult`
> - Removes frontend-added non-retryable string for update-execution-options; responsibility moved to worker
> - Adds unit tests for new formatter and non-retryable detection
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eb8942394b599c09cfa641f1c987aa85c73b9a94. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/common/worker_versioning/worker_versioning.go
+++ b/common/worker_versioning/worker_versioning.go
@@ -40,6 +40,11 @@ const (
 	UnversionedSearchAttribute = buildIdSearchAttributePrefixUnversioned
 	UnversionedVersionId       = "__unversioned__"
 
+	// ErrPinnedVersionNotInTaskQueueSubstring is the key substring used to identify
+	// when a pinned version is not present in a task queue. This is used for error
+	// classification in batch operations.
+	ErrPinnedVersionNotInTaskQueueSubstring = "is not present in task queue"
+
 	// WorkerDeploymentVersionIdDelimiterV31 will be deleted once we stop supporting v31 version string fields
 	// in external and internal APIs. Until then, both delimiters are banned in deployment name. All
 	// deprecated version string fields in APIs keep using the old delimiter. Workflow SA uses new delimiter.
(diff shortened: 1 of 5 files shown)


service/worker/batcher/activities.go
```diff
@@ -608,7 +608,15 @@ func (a *activities) handleTaskResult(
...
-		nonRetryable := slices.Contains(batchOperation.NonRetryableErrors, err.Error())
+		// Check if error is non-retryable by checking if the error message contains
+		// any of the non-retryable error strings
+		nonRetryable := false
+		for _, nonRetryableErr := range batchOperation.NonRetryableErrors {
+			if strings.Contains(err.Error(), nonRetryableErr) {
+				nonRetryable = true
+				break
+			}
+		}
```
Shivs11: may need a review from your side @awln-temporal (since you last touched this bit): TLDR is that I don't want the error message that I am passing from the frontend to be retried over here. However, the error message (with the latest changes) is generated dynamically and needs a task queue name when it is being built. The frontend code, specifically in the batcher, does not have this information. Thus, I am required to pass in only a subset of the string which feels quite...ugly to me.

Having dug through this a bit more, I see that we use NonRetryableErrors, which is a slice of strings. I wonder if we can make this better by changing it to be for specific error types? 

Moreover, if making the change that I described is kinda risky right now, well, is the change that I have made in activities.go safe?

common/worker_versioning/worker_versioning.go
```diff
@@ -546,7 +546,8 @@ func validatePinnedVersionInTaskQueue(ctx context.Context,
 			return nil
 		}
 		return serviceerror.NewFailedPrecondition(
-			"Pinned version is not present in the task queue",
+			fmt.Sprintf("Pinned version '%s:%s' is not present in task queue '%s' of type '%s'",
+				pinnedVersion.GetDeploymentName(), pinnedVersion.GetBuildId(), tq, tqType),
```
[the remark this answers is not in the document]
Shivs11: they shall be formatted in their string counterparts when paired with the %s formatter (tested this in the unit-test)

service/worker/batcher/activities.go
awln-temporal: Yeah, checking for error types seems like a better way to do this. It also seems that the only place that appends to this non-retryable errors slice is for this pinned worker version check. Fmi, why do we need to configure the list of retriable errors at runtime based off the batch operation type?

awln-temporal (approved):
Fine with current changes, but does it make sense to create separate error type for non-retryable, retryable error, create an error subtype for `pinned version not in task queue` that implements non-retryable error, instead of checking for substrings?

PR #9022, opened 2026-01-14
 check if states can be rebuilt(i.e. for now only workflows with events)

feiyang3cat:
## What changed?
check whether the state to be rebuilt is workflow

## Why?
only workflows have event sourcing

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/workflow_rebuilder.go
+++ b/service/history/workflow_rebuilder.go
@@ -111,22 +111,20 @@ func (r *workflowRebuilderImpl) rebuild(
 func (r *workflowRebuilderImpl) rebuildableCheck(
 	mutableState *persistencespb.WorkflowMutableState,
 ) error {
-	nonRebuildableErr := serviceerror.NewInvalidArgument(
-		"RebuildMutableState only supports workflow executions, not other archetype types")
-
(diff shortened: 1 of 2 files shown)


service/history/workflow_rebuilder.go
```diff
@@ -104,6 +104,31 @@ func (r *workflowRebuilderImpl) rebuild(
 	return r.overwriteToDB(ctx, rebuildMutableState)
 }
```
feiyang3cat, line 106: @yycptt

service/history/workflow_rebuilder.go
```diff
@@ -138,6 +161,11 @@ func (r *workflowRebuilderImpl) getRebuildSpecFromMutableState(
 	}
 
 	mutableState := resp.State
+	err = r.rebuildableCheck(mutableState)
```
yycptt, line 164: hmm can you also check if  `currentVersionHistory` is empty or not here? and only allow the rebuild when it's not empty. In case some kind of data loss corruption clears all chasm nodes. You can use `IsCurrentVersionHistoryEmpty`.

yycptt (commented):
as synced offline, plz also add a functional test in tests/chasm_test.go

PR #9024, opened 2026-01-14
Pause replication streams when the scheduler is under pressure.

robholland:
## What changed?
Pause replication streams when the scheduler is under pressure.

## Why?
This allows us to apply backpressure to replication streams when we cannot keep up with the load. The schedulers are shared amongst streams so just keep tracking of tasked tasks is not enough. We use a timer to notify us if a submit is taking too long rather than recording the time after the fact so that the backpressure is more reactive.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

---

> [!NOTE]
> **Backpressure for replication streams**
> 
> - Add slow-submission flow control: `stream_receiver` tracks per-priority slow submission timestamps and `stream_receiver_flow_controller` pauses when within `ReplicationReceiverSlowSubmissionWindow` (in addition to outstanding task count)
> - New dynamic configs: `history.ReplicationReceiverSubmissionLatencyThreshold`, `history.ReplicationReceiverSlowSubmissionWindow`, `history.EnableReplicationReceiverSlowSubmissionFlowControl`, wired through `configs.Config`
> - Refactor `StreamReceiver`: split scheduler selection into `getTaskSchedulerPriority`/`getTaskScheduler`, measure `Submit` latency, and feed `lastSlowSubmission` into flow control signals
> - Tests: expand flow controller tests to cover slow-submission window logic; adjust stream receiver tests and minor robustness checks
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9e6ec6ed812610ac9197fcc52157ff1ffa6d1e3f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/replication/stream_receiver_flow_controller.go
+++ b/service/history/replication/stream_receiver_flow_controller.go
@@ -40,10 +40,12 @@ func (s *streamReceiverFlowControllerImpl) GetFlowControlInfo(priority enumsspb.
 			return enumsspb.REPLICATION_FLOW_CONTROL_COMMAND_PAUSE
 		}
 
-		now := time.Now()
-		slowSubmissionWindow := now.Add(-s.config.ReplicationReceiverSlowSubmissionWindow())
-		if signalData.lastSlowSubmission.After(slowSubmissionWindow) {
-			return enumsspb.REPLICATION_FLOW_CONTROL_COMMAND_PAUSE
+		if s.config.EnableReplicationReceiverSlowSubmissionFlowControl() {
+			now := time.Now()
+			slowSubmissionWindow := now.Add(-s.config.ReplicationReceiverSlowSubmissionWindow())
+			if signalData.lastSlowSubmission.After(slowSubmissionWindow) {
+				return enumsspb.REPLICATION_FLOW_CONTROL_COMMAND_PAUSE
+			}
 		}
 	}
 	return enumsspb.REPLICATION_FLOW_CONTROL_COMMAND_RESUME
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -2576,6 +2576,11 @@ that task will be sent to DLQ.`,
 		10*time.Second,
 		`Time window within which a slow submission will pause replication flow control`,
 	)
+	EnableReplicationReceiverSlowSubmissionFlowControl = NewGlobalBoolSetting(
+		"history.EnableReplicationReceiverSlowSubmissionFlowControl",
+		false,
+		`Enable slow submission flow control check in replication receiver`,
+	)
 	ReplicationResendMaxBatchCount = NewGlobalIntSetting(
 		"history.ReplicationResendMaxBatchCount",
 		10,
--- a/service/history/configs/config.go
+++ b/service/history/configs/config.go
@@ -299,6 +299,7 @@ type Config struct {
 	ReplicationReceiverMaxOutstandingTaskCount          dynamicconfig.IntPropertyFn
 	ReplicationReceiverSlowSubmissionLatencyThreshold   dynamicconfig.DurationPropertyFn
 	ReplicationReceiverSlowSubmissionWindow             dynamicconfig.DurationPropertyFn
+	EnableReplicationReceiverSlowSubmissionFlowControl  dynamicconfig.BoolPropertyFn
 	ReplicationResendMaxBatchCount                      dynamicconfig.IntPropertyFn
 	ReplicationProgressCacheMaxSize                     dynamicconfig.IntPropertyFn
 	ReplicationProgressCacheTTL                         dynamicconfig.DurationPropertyFn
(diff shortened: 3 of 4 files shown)


service/history/replication/stream_receiver.go
```diff
@@ -323,42 +331,73 @@ func (r *StreamReceiverImpl) processMessages(
...
-			scheduler, err := r.getTaskScheduler(streamResp.Resp.GetMessages().Priority, task)
+			schedulerPriority, err := r.getTaskSchedulerPriority(priority, task)
+			if err != nil {
+				return err
+			}
+			scheduler, err := r.getTaskScheduler(schedulerPriority)
 			if err != nil {
 				return err
 			}
+			timer := time.AfterFunc(submissionThreshold, func() {
```
yux0: Could this create too many timers? Can we do it for the batch of tasks (outside the loop)?

service/history/replication/stream_receiver_flow_controller.go
```diff
@@ -31,7 +34,15 @@ func NewReceiverFlowControl(signals map[enumsspb.TaskPriority]FlowControlSignalP
 
 func (s *streamReceiverFlowControllerImpl) GetFlowControlInfo(priority enumsspb.TaskPriority) enumsspb.ReplicationFlowControlCommand {
```
yux0, line 35: Currently, the receiver has a stream liveness check and it will kill the stream if it did not receive messages from sender. I think this is a missing piece that if we pause the stream. we should also pause the stream liveness check. But this doesn't have to be in this PR.

service/history/replication/stream_receiver.go
robholland: I could reduce this to two timers per batch (one per scheduler), but I would have to partition the tasks by scheduler for this to work, otherwise we wouldn't know which scheduler was being slow. What do you think?

robholland: Adjusted to have a timer per priority and reset them.

service/history/replication/stream_receiver_flow_controller.go
robholland: Yeah let's do that in a separate PR as it's not specific to this change.

service/history/replication/stream_receiver.go
```diff
@@ -323,42 +333,82 @@ func (r *StreamReceiverImpl) processMessages(
...
-			scheduler, err := r.getTaskScheduler(streamResp.Resp.GetMessages().Priority, task)
+			schedulerPriority, err := r.getTaskSchedulerPriority(priority, task)
+			if err != nil {
+				return err
+			}
+			scheduler, err := r.getTaskScheduler(schedulerPriority)
 			if err != nil {
 				return err
 			}
+			timer := r.startSlowSubmissionTimer(schedulerPriority, submissionThreshold)
```
xwduan: This a very hot code path, in general, we should avoid using timer here just for the purpose of measuring the submission latency... Can we use 2 timestamps as discussed before?

service/history/replication/stream_receiver_flow_controller.go
```diff
@@ -31,7 +34,15 @@ func NewReceiverFlowControl(signals map[enumsspb.TaskPriority]FlowControlSignalP
...
-		if signal().taskTrackingCount > s.config.ReplicationReceiverMaxOutstandingTaskCount() {
+		signalData := signal()
+
+		if signalData.taskTrackingCount > s.config.ReplicationReceiverMaxOutstandingTaskCount() {
+			return enumsspb.REPLICATION_FLOW_CONTROL_COMMAND_PAUSE
+		}
+
+		now := time.Now()
+		slowSubmissionWindow := now.Add(-s.config.ReplicationReceiverSlowSubmissionWindow())
+		if signalData.lastSlowSubmission.After(slowSubmissionWindow) {
```
yux0: I think there is some implicit condition here. The replication will be resume after ReplicationReceiverSlowSubmissionWindow.

yux0: I don't have better solution here. Can we use a longer time window as default? 10s?

service/history/replication/stream_receiver.go
robholland: Adjusted.

service/history/replication/stream_receiver.go
```diff
@@ -323,42 +331,74 @@ func (r *StreamReceiverImpl) processMessages(
...
+			}
+			scheduler, err := r.getTaskScheduler(schedulerPriority)
 			if err != nil {
 				return err
 			}
+			start := time.Now()
 			scheduler.Submit(task)
+			end := time.Now()
+			if end.Sub(start) > submissionThreshold {
+				r.recordSlowSubmission(schedulerPriority, end)
```
yux0, line 380: LGTM. Can we add a feature flag on this?

yux0 (approved):

PR #9032, opened 2026-01-15
fix race condition AND add type check FOR admin rebuild feature

feiyang3cat:
## What changed?
(1) fix race condition: wrap all read/writes of execution/ctx within ctx lock
(2) only rebuild the workflow archetype + only rebuild when versionHistory not empty

## Why?
(1) it is a bug
(2) it is other archetypes have no events to be rebuilt

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/history/workflow_rebuilder.go
+++ b/service/history/workflow_rebuilder.go
@@ -132,18 +132,15 @@ func (r *workflowRebuilderImpl) rebuildableCheck(
 	}
 
 	// check2: check if the current version history is empty
+	checkErr = serviceerror.NewInvalidArgument("version histories is nil, cannot be rebuilt")
 	if mutableState.ExecutionInfo == nil || mutableState.ExecutionInfo.VersionHistories == nil {
-		return serviceerror.NewInvalidArgument("version histories is nil, cannot be rebuilt")
+		return checkErr
 	}
-	currentVersionHistory, err := versionhistory.GetCurrentVersionHistory(mutableState.ExecutionInfo.VersionHistories)
+	isEmpty, err := versionhistory.IsCurrentVersionHistoryEmpty(mutableState.ExecutionInfo.VersionHistories)
 	if err != nil {
 		return err
 	}
-	if currentVersionHistory == nil {
-		return serviceerror.NewInvalidArgument("current version history is nil, cannot be rebuilt")
-	}
-	checkErr = serviceerror.NewInvalidArgument("current version history is empty, cannot be rebuilt")
-	if versionhistory.IsEmptyVersionHistory(currentVersionHistory) {
+	if isEmpty {
 		return checkErr
 	}
 
--- a/tests/admin_test.go
+++ b/tests/admin_test.go
@@ -15,6 +15,7 @@ import (
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/primitives/timestamp"
+	"go.temporal.io/server/common/testing/testvars"
 	"go.temporal.io/server/tests/testcore"
 )
 
@@ -35,7 +36,7 @@ func (s *AdminTestSuite) SetupSuite() {
 }
 
 func (s *AdminTestSuite) TestAdminRebuildMutableState_ChasmDisabled() {
-	rebuildMutableStateWorkflowHelper(s.testContext, &s.FunctionalTestBase, false)
+	rebuildMutableStateWorkflowHelper(s, false)
 }
 
 func (s *AdminTestSuite) TestAdminRebuildMutableState_ChasmEnabled() {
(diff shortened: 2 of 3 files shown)


service/history/workflow_rebuilder.go
```diff
@@ -104,6 +105,52 @@ func (r *workflowRebuilderImpl) rebuild(
...
+	if err != nil {
+		return err
+	}
+	if currentVersionHistory == nil {
+		return serviceerror.NewInvalidArgument("current version history is nil, cannot be rebuilt")
+	}
+	checkErr = serviceerror.NewInvalidArgument("current version history is empty, cannot be rebuilt")
+	if versionhistory.IsEmptyVersionHistory(currentVersionHistory) {
+		return checkErr
+	}
```
yycptt: nit: there's also a IsCurrentVersionHistoryEmpty function you can use directly

tests/admin_test.go
```diff
@@ -42,13 +65,13 @@ func (s *AdminTestSuite) TestAdminRebuildMutableState() {
 
 	s.Worker().RegisterWorkflow(workflowFn)
 
-	workflowID := "functional-admin-rebuild-mutable-state-test"
+	workflowID := "functional-admin-rebuild-mutable-state-test-" + testcore.RandomizeStr("wf")
```
yycptt: nit: testvar.Any().String() can help with this I think.

tests/admin_test.go
```diff
@@ -42,13 +65,13 @@ func (s *AdminTestSuite) TestAdminRebuildMutableState() {
...
 	workflowOptions := sdkclient.StartWorkflowOptions{
 		ID:                 workflowID,
 		TaskQueue:          s.TaskQueue(),
 		WorkflowRunTimeout: 20 * time.Second,
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
yycptt: nit: I guess this one can just be the default s.testContext.

tests/admin_test.go
```diff
@@ -77,6 +100,14 @@ func (s *AdminTestSuite) TestAdminRebuildMutableState() {
 		})
 		s.NoError(err)
 		if response1.DatabaseMutableState.ExecutionInfo.StateTransitionCount == 3 {
+			// Note: ChasmNodes may be empty even with CHASM enabled, so we only check if the rebuild can be performed,
+			// and not checking whether it is rebuildable because ChasmNodes are present.
+			if testWithChasm {
+				s.T().Logf("CHASM is enabled and workflow has ChasmNodes (new format)")
+				// s.NotEmpty(response1.DatabaseMutableState.ChasmNodes, "CHASM-enabled workflows should have ChasmNodes")
```
yycptt: we can remove those?

tests/chasm_test.go
```diff
@@ -719,4 +717,38 @@ func (s *ChasmTestSuite) TestTaskQueuePreallocatedSearchAttribute() {
...
+	archetype, _ := s.FunctionalTestBase.GetTestCluster().Host().GetCHASMRegistry().ComponentFqnByID(archetypeID)
+	s.NotEqual(archetype, chasm.WorkflowArchetype, "Archetype should not be the workflow archetype")
+
+	_, err = s.AdminClient().RebuildMutableState(testcore.NewContext(), &adminservice.RebuildMutableStateRequest{
+		Namespace: s.Namespace().String(),
+		Execution: &commonpb.WorkflowExecution{
+			WorkflowId: storeID,
+		},
+	})
+	s.Error(err, "RebuildMutableState should fail for non-workflow executions")
```
yycptt: nit: verify the error type is actually InvalidArgument.

yycptt (approved):

tests/chasm_test.go
```diff
@@ -746,9 +771,10 @@ func (s *ChasmTestSuite) TestMutableStateRebuilder() {
 		Namespace: s.Namespace().String(),
 		Execution: &commonpb.WorkflowExecution{
 			WorkflowId: storeID,
+			RunId:      runID,
```
feiyang3cat, line 774: my finding in local test is that if I don't put runID, wtCache returns workflow not found err

maki3cat: @yycptt

yycptt: yeah because there's no workflow with WorkflowID == storeID.

I guess we can argue there's no workflow with runID == storeID, so should also return a NotFound in that case from wfCache / persistency layer, that's a more involved change though, let me track this and we can try fix it later.

The rebuild implementation is specific since it's not using ctx.LoadMutableState() method like all other places. Inside LoadMutableState there's a archetype check and will return NotFound if the loaded ms doesn't have the right archetype. so we should already have consistency in most places.

PR #9034, opened 2026-01-15
Support ExecutionStatus and TaskQueue as chasm search attributes

awln-temporal:
## What changed?
Support `ExecutionStatus` and `TaskQueue` as chasm search attributes. 

`ExecutionStatus` is a registrable chasm search attribute alias which will be used by chasm library authors to map to LowCardinality fields.

`TaskQueue` is a predefined chasm search attribute that will map internally to the TaskQueue system search attribute. TaskQueue values will be retrieved from the list of SearchAttributeKeyValues and will replace the TaskQueue parameter in the visibility request base. On query, the TaskQueue system search attribute value will populate the TaskQueue key in the Chasm search attributes map.

## Why?
Unblock standalone activities.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/chasm/lib/tests/payload.go
+++ b/chasm/lib/tests/payload.go
@@ -16,18 +16,17 @@ const (
 )
 
 const (
-	TestScheduleID                = "TestScheduleID"
-	PayloadTotalCountSAAlias      = "PayloadTotalCount"
-	PayloadTotalSizeSAAlias       = "PayloadTotalSize"
-	PayloadExecutionStatusSAAlias = "PayloadExecutionStatus"
-	ExecutionStatusSAAlias        = "ExecutionStatus"
-	DefaultPayloadStoreTaskQueue  = "payload-store-task-queue"
+	TestScheduleID               = "TestScheduleID"
+	PayloadTotalCountSAAlias     = "PayloadTotalCount"
+	PayloadTotalSizeSAAlias      = "PayloadTotalSize"
+	ExecutionStatusSAAlias       = "ExecutionStatus"
+	DefaultPayloadStoreTaskQueue = "payload-store-task-queue"
 )
 
 var (
 	PayloadTotalCountSearchAttribute = chasm.NewSearchAttributeInt(PayloadTotalCountSAAlias, chasm.SearchAttributeFieldInt01)
 	PayloadTotalSizeSearchAttribute  = chasm.NewSearchAttributeInt(PayloadTotalSizeSAAlias, chasm.SearchAttributeFieldInt02)
-	ExecutionStatusSearchAttribute   = chasm.NewSearchAttributeKeyword(ExecutionStatusSAAlias, chasm.SearchAttributeFieldKeyword03)
+	ExecutionStatusSearchAttribute   = chasm.NewSearchAttributeKeyword(ExecutionStatusSAAlias, chasm.SearchAttributeFieldLowCardinalityKeyword01)
 
 	_ chasm.VisibilitySearchAttributesProvider = (*PayloadStore)(nil)
 	_ chasm.VisibilityMemoProvider             = (*PayloadStore)(nil)
--- a/tests/chasm_test.go
+++ b/tests/chasm_test.go
@@ -611,7 +611,7 @@ func (s *ChasmTestSuite) TestListExecutions_ExecutionStatusAsAlias() {
 
 	// Query using "ExecutionStatus" as a CHASM alias (which maps to TemporalKeyword03)
 	// This tests that CHASM components can use "ExecutionStatus" as an alias for their own search attribute
-	visQuery := fmt.Sprintf("TemporalNamespaceDivision = '%d' AND ExecutionStatus = 'Running'", archetypeID)
+	visQuery := fmt.Sprintf("TemporalNamespaceDivision = '%d' AND ExecutionStatus = 'Running' AND PayloadStoreId = '%s'", archetypeID, storeID)
 
 	var visRecord *chasm.ExecutionInfo[*testspb.TestPayloadStore]
 	s.Eventually(


chasm/lib/tests/payload.go
```diff
@@ -20,12 +20,15 @@ const (
...
 	PayloadExecutionStatusSAAlias = "PayloadExecutionStatus"
+	ExecutionStatusSAAlias        = "ExecutionStatus"
+	DefaultPayloadStoreTaskQueue  = "payload-store-task-queue"
 )
 
 var (
 	PayloadTotalCountSearchAttribute      = chasm.NewSearchAttributeInt(PayloadTotalCountSAAlias, chasm.SearchAttributeFieldInt01)
 	PayloadTotalSizeSearchAttribute       = chasm.NewSearchAttributeInt(PayloadTotalSizeSAAlias, chasm.SearchAttributeFieldInt02)
 	PayloadExecutionStatusSearchAttribute = chasm.NewSearchAttributeKeyword(PayloadExecutionStatusSAAlias, chasm.SearchAttributeFieldLowCardinalityKeyword01)
+	ExecutionStatusSearchAttribute        = chasm.NewSearchAttributeKeyword(ExecutionStatusSAAlias, chasm.SearchAttributeFieldKeyword03)
```
rodrigozhou: Didn't you mean to replace the `PayloadExecutionStatusSearchAttribute` with this new one so it would be mapped to low cardinality field?

common/persistence/visibility/visibility_manager_impl.go
```diff
@@ -200,6 +200,15 @@ func (p *visibilityManagerImpl) convertToChasmExecutionInfo(
 		return nil, err
 	}
 
+	// If TaskQueue is registered in the mapper, populate it from the execution's TaskQueue field.
+	// TaskQueue is a preallocated search attribute that maps to the system column.
+	if chasmAliasedSAs == nil {
+		chasmAliasedSAs = make(map[string]chasm.VisibilityValue)
+	}
```
rodrigozhou: I thought the predefined search attributes in CHASM didn't need to be registered with the component. Do they need?

common/searchattribute/sadefs/constants.go
```diff
@@ -188,6 +188,25 @@ var (
...
+		StartTime:            {},
+		ExecutionTime:        {},
+		CloseTime:            {},
+		HistoryLength:        {},
+		HistorySizeBytes:     {},
+		StateTransitionCount: {},
+		ParentWorkflowID:     {},
+		ParentRunID:          {},
+		RootWorkflowID:       {},
+		RootRunID:            {},
```
rodrigozhou: Parent and root execution is not populated at this moment. Do we want to reserve those as potential CHASM system search attributes?
cc: @bergundy

common/searchattribute/sadefs/constants.go
```diff
@@ -188,6 +188,25 @@ var (
 		VisibilityTaskKey: {},
 	}
 
+	// chasmSystemSearchAttributes are system search attributes used by CHASM internally
+	// and cannot be used as CHASM search attribute aliases. These correspond to fields
+	// in chasm.ExecutionInfo.
+	chasmSystemSearchAttributes = map[string]struct{}{
+		WorkflowID:           {},
+		RunID:                {},
+		WorkflowType:         {},
```
rodrigozhou: I thought workflow type was not gonna be here.
cc: @bergundy

common/persistence/visibility/visibility_manager_impl.go
yycptt: is the purpose for registration to know if TaskQueue needs to be returned? I think we can return it as long as it's not empty. Then we don't need to do the registration and have the same experience as other predefined SA.

common/searchattribute/sadefs/constants.go
yycptt: yeah we can move it later I think given the tight deadline. Add a TODO here though.

common/searchattribute/sadefs/constants.go
yycptt: Those feel like chasm component SA to me as well. I am not sure if we want to make parent/child a general chasm concept. Similar to WorkflowType above, I feel we can move them later.

common/persistence/visibility/visibility_manager_impl.go
awln-temporal: Yeah, maybe we can return it by default then. I wasn't quite sure about all the places that end up setting TaskQueue, but maybe aren't a CHASM execution. And, like for schedules, there may be cases that want to query against chasm and non-chasm executions, but we might not want to return taskQueue as a chasm search attribute for non-chasm executions?

chasm/lib/tests/payload.go
awln-temporal: Left it as a separate search attribute to test both paths as a chasm search attribute alias and as the system attribute alias, didn't really think any harm to test both cases, but I can remove `PayloadExecutionStatusSearchAttribute`

common/searchattribute/sadefs/constants.go
awln-temporal: removed workflowType

chasm/lib/tests/payload.go
```diff
@@ -20,12 +20,14 @@ const (
...
+	DefaultPayloadStoreTaskQueue  = "payload-store-task-queue"
 )
 
 var (
-	PayloadTotalCountSearchAttribute      = chasm.NewSearchAttributeInt(PayloadTotalCountSAAlias, chasm.SearchAttributeFieldInt01)
-	PayloadTotalSizeSearchAttribute       = chasm.NewSearchAttributeInt(PayloadTotalSizeSAAlias, chasm.SearchAttributeFieldInt02)
-	PayloadExecutionStatusSearchAttribute = chasm.NewSearchAttributeKeyword(PayloadExecutionStatusSAAlias, chasm.SearchAttributeFieldLowCardinalityKeyword01)
+	PayloadTotalCountSearchAttribute = chasm.NewSearchAttributeInt(PayloadTotalCountSAAlias, chasm.SearchAttributeFieldInt01)
+	PayloadTotalSizeSearchAttribute  = chasm.NewSearchAttributeInt(PayloadTotalSizeSAAlias, chasm.SearchAttributeFieldInt02)
+	ExecutionStatusSearchAttribute   = chasm.NewSearchAttributeKeyword(ExecutionStatusSAAlias, chasm.SearchAttributeFieldKeyword03)
```
rodrigozhou:
```suggestion
	ExecutionStatusSearchAttribute   = chasm.NewSearchAttributeKeyword(ExecutionStatusSAAlias, chasm.SearchAttributeFieldLowCardinalityKeyword01)
```

chasm/lib/tests/payload.go
```diff
@@ -20,12 +20,14 @@ const (
 	PayloadTotalCountSAAlias      = "PayloadTotalCount"
 	PayloadTotalSizeSAAlias       = "PayloadTotalSize"
 	PayloadExecutionStatusSAAlias = "PayloadExecutionStatus"
```
rodrigozhou: Remove this if unused

PR #9036, opened 2026-01-15
Reduce load on priority activity functional test

majiru:
## What changed?
Reduce the number of tasks to reduce load in CI to hopefully avoid the context timeout.

## Why?
We've observed some flakes in CI for this, this should help.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/priority_fairness_test.go
+++ b/tests/priority_fairness_test.go
@@ -115,6 +115,7 @@ func (s *PrioritySuite) TestActivity_Basic() {
 		s.NoError(err)
 	}
 
+	// wait for activity tasks to appear in the matching backlog (from transfer queue)
 	s.Eventually(func() bool {
 		res, err := s.AdminClient().DescribeTaskQueuePartition(ctx, &adminservice.DescribeTaskQueuePartitionRequest{
 			Namespace: s.Namespace().String(),


tests/priority_fairness_test.go
```diff
@@ -50,12 +49,15 @@ func (s *PrioritySuite) SetupSuite() {
...
-	const N = 100
+func (s *PrioritySuite) TestPriority_Activity_Basic() {
+	const N = 20
 	const Levels = 5
 
 	tv := testvars.New(s.T())
 
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueReadPartitions, 1)
+	s.OverrideDynamicConfig(dynamicconfig.MatchingNumTaskqueueWritePartitions, 1)
+
```
dnr, line 61: should we also wait for the activity tasks to appear in the backlog, to avoid inadvertent sync match that will mess up the order? I guess it's extremely unlikely to mess it up enough to fail the test

tests/priority_fairness_test.go
```diff
@@ -112,6 +114,28 @@ func (s *PrioritySuite) TestActivity_Basic() {
 		s.NoError(err)
 	}
 
+	s.Eventually(func() bool {
+		res, err := s.AdminClient().DescribeTaskQueuePartition(ctx, &adminservice.DescribeTaskQueuePartitionRequest{
```
dnr, line 120: I wanted to say we should factor out all these "wait for various backlog conditions" on the different tests but actually being explicit and duplicating a little is probably better for tests

tests/priority_fairness_test.go
```diff
@@ -112,6 +114,28 @@ func (s *PrioritySuite) TestActivity_Basic() {
 		s.NoError(err)
 	}
 
+	s.Eventually(func() bool {
```
dnr, line 119:
```suggestion
	// wait for activity tasks to appear in the matching backlog (from transfer queue)
	s.Eventually(func() bool {
```

dnr (approved):

PR #9037, opened 2026-01-15
Set external payload stats as pre-defined search attributes in visibility

simvlad:
## What changed?
Set external payload stats as pre-defined search attributes in visibility

## Why?
We'd like to surface external payload size and count in visibility. They will only be set for workflows that use external payloads.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

## Potential risks
No

### DIFF
--- a/service/history/visibility_queue_task_executor.go
+++ b/service/history/visibility_queue_task_executor.go
@@ -38,6 +38,7 @@ type (
 		ensureCloseBeforeDelete       dynamicconfig.BoolPropertyFn
 		enableCloseWorkflowCleanup    dynamicconfig.BoolPropertyFnWithNamespaceFilter
 		relocateAttributesMinBlobSize dynamicconfig.IntPropertyFnWithNamespaceFilter
+		externalPayloadsEnabled       dynamicconfig.BoolPropertyFnWithNamespaceFilter
 	}
 )
 
@@ -52,6 +53,7 @@ func newVisibilityQueueTaskExecutor(
 	ensureCloseBeforeDelete dynamicconfig.BoolPropertyFn,
 	enableCloseWorkflowCleanup dynamicconfig.BoolPropertyFnWithNamespaceFilter,
 	relocateAttributesMinBlobSize dynamicconfig.IntPropertyFnWithNamespaceFilter,
+	externalPayloadsEnabled dynamicconfig.BoolPropertyFnWithNamespaceFilter,
 ) queues.Executor {
 	return &visibilityQueueTaskExecutor{
 		shardContext:   shardContext,
@@ -63,6 +65,7 @@ func newVisibilityQueueTaskExecutor(
 		ensureCloseBeforeDelete:       ensureCloseBeforeDelete,
 		enableCloseWorkflowCleanup:    enableCloseWorkflowCleanup,
 		relocateAttributesMinBlobSize: relocateAttributesMinBlobSize,
+		externalPayloadsEnabled:       externalPayloadsEnabled,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


service/history/visibility_queue_task_executor.go
```diff
@@ -545,6 +545,22 @@ func (t *visibilityQueueTaskExecutor) getClosedVisibilityRequest(
...
+			IndexedFields: make(map[string]*commonpb.Payload),
+		}
+	}
+	if base.SearchAttributes.IndexedFields == nil {
+		base.SearchAttributes.IndexedFields = make(map[string]*commonpb.Payload)
+	}
+	externalPayloadCountPayload, _ := payload.Encode(externalPayloadCount)
+	externalPayloadSizeBytesPayload, _ := payload.Encode(externalPayloadSizeBytes)
+	base.SearchAttributes.IndexedFields[sadefs.TemporalExternalPayloadCount] = externalPayloadCountPayload
+	base.SearchAttributes.IndexedFields[sadefs.TemporalExternalPayloadSizeBytes] = externalPayloadSizeBytesPayload
```
rodrigozhou: These search attribute will only be set for closed workflows?

rodrigozhou: I see that there's a feature flag for external payloads, but here, you're always setting a value, ie., if the feature is disabled, it will write 0 to both search attributes. Is that intended?

service/history/visibility_queue_task_executor.go
```diff
@@ -545,6 +545,22 @@ func (t *visibilityQueueTaskExecutor) getClosedVisibilityRequest(
...
 	historySizeBytes := executionInfo.GetExecutionStats().GetHistorySize()
+
+	externalPayloadCount := executionInfo.GetExecutionStats().GetExternalPayloadCount()
+	externalPayloadSizeBytes := executionInfo.GetExecutionStats().GetExternalPayloadSize()
+	if base.SearchAttributes == nil {
+		base.SearchAttributes = &commonpb.SearchAttributes{
+			IndexedFields: make(map[string]*commonpb.Payload),
+		}
+	}
+	if base.SearchAttributes.IndexedFields == nil {
```
rodrigozhou: This could be `else if`.

service/history/visibility_queue_task_executor.go
simvlad: Yes -- similarly to the HistorySize

simvlad: I'll change to skip if the feature flag is disabled

service/history/visibility_queue_task_executor.go
```diff
@@ -558,6 +558,24 @@ func (t *visibilityQueueTaskExecutor) getClosedVisibilityRequest(
...
+
+	if base.SearchAttributes == nil {
+		base.SearchAttributes = &commonpb.SearchAttributes{
+			IndexedFields: make(map[string]*commonpb.Payload),
+		}
+	} else if base.SearchAttributes.IndexedFields == nil {
+		base.SearchAttributes.IndexedFields = make(map[string]*commonpb.Payload)
+	}
+
+	if t.shardContext.GetConfig().ExternalPayloadsEnabled(mutableState.GetNamespaceEntry().Name().String()) {
```
rodrigozhou: Can you save this dynamicconfig in the task queue object so you won't have to call `t.shardContext.GetConfig()` here every time?

rodrigozhou: I'd add `namespaceEntry` as argument to `getClosedVisibilityRequest` similarly to `getVisibilityRequestBase`.

tests/advanced_visibility_test.go
```diff
@@ -2477,6 +2480,71 @@ func (s *AdvancedVisibilitySuite) TestBuildIdScavenger_DeletesUnusedBuildId() {
...
+		Payloads: []*commonpb.Payload{{
+			Data: []byte("workflow input"),
+			ExternalPayloads: []*commonpb.Payload_ExternalPayloadDetails{
+				{SizeBytes: externalPayloadSize},
+			},
+		}},
+	}
+
+	// Add testSearchAttributeKey to satisfy testHelperForReadOnce validation
+	attrValBytes, _ := payload.Encode(testSearchAttributeVal)
```
awln-temporal: nitpick: I wonder if we should just remove the validation of `testSearchAttributeKey` in testHelperForReadOnce, it looks like it's only used one time to validate the search attribute, which can be done outside of the function like most of the test cases are doing.

PR #9040, opened 2026-01-15
[worker-versioning]: blackhole queries only on root partition

Shivs11:
## What changed?
- We should be checking if a query is blackholed only by checking the appropriate conditions on the root partition.
- Moreover, added the testhook bug about read partitions over here which @dnr spotted!

## Why?
- Correctness. 
- Moreover, the test TestVersioning3FunctionalSuiteV0/TestPinnedQuery_RollbackDrainedVersion was failing after the testhook bug fix. The root partition check also fixed it!

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None.

---

> [!NOTE]
> Ensures correctness of worker-versioned query dispatch and fixes a testhook.
> 
> - Only perform `checkQueryBlackholed` on the root partition in `task_queue_partition_manager.go` to avoid false "no pollers" detections across partitions
> - Correct testhook key in `client/matching/loadbalancer.go` to use `MatchingLBForceReadPartition` when forcing read partitions
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6b182d703b110712abed2eed7529e9e948209a84. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/matching/task_queue_partition_manager.go
```diff
@@ -1722,23 +1722,31 @@ func (pm *taskQueuePartitionManagerImpl) checkQueryBlackholed(
...
-		if versionData.GetVersion() != nil && worker_versioning.DeploymentVersionFromDeployment(deployment).Equal(versionData.GetVersion()) {
-			if versionData.GetStatus() == enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINED && len(pm.GetAllPollerInfo()) == 0 {
-				versionStr := worker_versioning.ExternalWorkerDeploymentVersionToString(worker_versioning.ExternalWorkerDeploymentVersionFromDeployment(deployment))
-				return serviceerror.NewFailedPreconditionf(ErrBlackholedQuery, versionStr, versionStr)
+
+	// Only perform this check on the root partition.
+	// Forwarding (polls/tasks/queries) moves “up” the partition tree, so the root is the convergence point.
+	// Checking non-root partitions can incorrectly conclude “no pollers” simply because the pollers are on
+	// a different partition and have not been forwarded here (yet), leading to false blackhole errors.
+	if pm.partition.IsRoot() {
```
dnr: how about do this as an early return, `if !pm.partition.IsRoot() { return nil }` ? keeps the indentation more manageable

PR #9046, opened 2026-01-15
[Scheduler] CountSchedules implementation for V1 and V2

lina-temporal:
## What changed?
- Implements the new `CountSchedules` API for both CHASM and V1 scheduler.
- Depends on [saas-policy change](https://github.com/temporalio/saas-policy/pull/251), [API change](https://github.com/temporalio/api/pull/696)

## Why?
- Our UI was previously using an undocumented `ListWorkflowExecutions` call to get a total count of schedules. This will allow us to move it to a public API in advance of CHASM release, which will continue to work for both V1 and V2.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- New API is highly symmetric in design to all other Count* APIs, and implementation is highly symmetric to ListSchedules. Seems low.

### DIFF
--- a/tests/schedule_test.go
+++ b/tests/schedule_test.go
@@ -1484,25 +1484,28 @@ func (s *scheduleFunctionalSuiteBase) TestCountSchedules() {
 	}, 15*time.Second, 1*time.Second)
 
 	// Test basic count (all schedules)
-	countResp, err := s.FrontendClient().CountSchedules(s.newContext(), &workflowservice.CountSchedulesRequest{
-		Namespace: s.Namespace().String(),
-	})
-	s.NoError(err)
-	s.GreaterOrEqual(countResp.Count, int64(3), "Expected at least 3 schedules")
+	s.Eventually(func() bool {
+		countResp, err := s.FrontendClient().CountSchedules(s.newContext(), &workflowservice.CountSchedulesRequest{
+			Namespace: s.Namespace().String(),
+		})
+		return err == nil && countResp.Count >= 3
+	}, 15*time.Second, 1*time.Second, "Expected at least 3 schedules")
 
 	// Test count with query filter for paused schedules
-	countResp, err = s.FrontendClient().CountSchedules(s.newContext(), &workflowservice.CountSchedulesRequest{
-		Namespace: s.Namespace().String(),
-		Query:     fmt.Sprintf("%s = true", sadefs.TemporalSchedulePaused),
-	})
-	s.NoError(err)
-	s.GreaterOrEqual(countResp.Count, int64(1), "Expected at least 1 paused schedule")
+	s.Eventually(func() bool {
+		countResp, err := s.FrontendClient().CountSchedules(s.newContext(), &workflowservice.CountSchedulesRequest{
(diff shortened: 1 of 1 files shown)


service/frontend/workflow_handler.go
```diff
@@ -4565,6 +4565,116 @@ func (wh *WorkflowHandler) listSchedulesWorkflow(
 	}, nil
 }
 
+func (wh *WorkflowHandler) CountSchedules(
+	ctx context.Context,
+	request *workflowservice.CountSchedulesRequest,
+) (_ *workflowservice.CountSchedulesResponse, retError error) {
```
chaptersix, line 4587: What is the purpose of including the underscore here?

service/frontend/workflow_handler.go
```diff
@@ -4565,6 +4565,116 @@ func (wh *WorkflowHandler) listSchedulesWorkflow(
...
+	if err != nil {
+		return nil, err
+	}
+
+	if wh.config.DisableListVisibilityByFilter(namespaceName.String()) {
+		return nil, errListNotAllowed
+	}
+
+	// Build query with base filter (same as ListSchedules)
+	query := scheduler.VisibilityBaseListQuery
```
chaptersix: If this is the same as list schedules should we put this in a helper function?

service/frontend/workflow_handler.go
```diff
@@ -4565,6 +4565,116 @@ func (wh *WorkflowHandler) listSchedulesWorkflow(
...
+			return nil, err
+		}
+		query = fmt.Sprintf("%s AND (%s)", scheduler.VisibilityBaseListQuery, request.Query)
+	}
+
+	// Route to CHASM or V1 based on config (same pattern as ListSchedules)
+	if wh.chasmSchedulerEnabled(ctx, namespaceName.String()) {
+		return wh.countSchedulesChasm(ctx, namespaceID, namespaceName, query)
+	}
+	return wh.countSchedulesWorkflow(ctx, namespaceID, namespaceName, query)
```
chaptersix: We are doing a pretty gradual rollout so there might be a significant amount of time where there is schedulers leftover on v1 stack. Once the flag is flipped for the namespace the count drops to 0 and increases as new schedules are created. Some time later, it increases as schedules wake up....

service/frontend/workflow_handler.go
lina-temporal: It's because you need to name both arguments if you name one of them, and yet I never assign to the first return argument, so instead it is bound to "unused"/`_`.

service/frontend/workflow_handler.go
lina-temporal: Yeah, good call-out, will fix.

service/frontend/workflow_handler.go
lina-temporal: The query itself we select with handles this case - `VisibilityListQueryChasm` will return results from both V1 and V2 scheduler, that'll apply to Count, as well.

I'll update the ListV1andV2 functional test to show this.

chaptersix (approved):

PR #9047, opened 2026-01-15
Various SAA fixes

bergundy:
## What changed?

- Replaced custom search attribute aliases with standard ones:
  - ActivityStatus → ExecutionStatus (using the standard execution status field)
  - ActivityTaskQueue → TaskQueue (using the standard task queue field)
  - ActivityType now uses the alias name directly instead of a constant
- Renamed the GetValue() function to SearchAttributeValue() for better clarity

- Changed Activity.Store field type from chasm.Field[ActivityStore] to chasm.ParentPtr[ActivityStore] to avoid the storage overhead of a pointer that is only ever expected to be a parent
- Fixed error types in handler.go: changed from serviceerror.NewFailedPrecondition to serviceerror.NewInvalidArgumentf for invalid policy errors
- Improved ParentPtr.TryGet() to gracefully return false instead of panicking when not initialized

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -31,16 +31,11 @@ const (
 	// WorkflowTypeTag is a required workflow tag for standalone activities to ensure consistent
 	// metric labeling between workflows and activities.
 	WorkflowTypeTag = "__temporal_standalone_activity__"
-
-	TypeSAAlias      = "ActivityType"
-	StatusSAAlias    = "ActivityStatus"
-	TaskQueueSAAlias = "ActivityTaskQueue"
 )
 
 var (
-	TypeSearchAttribute      = chasm.NewSearchAttributeKeyword(TypeSAAlias, chasm.SearchAttributeFieldKeyword01)
-	StatusSearchAttribute    = chasm.NewSearchAttributeKeyword(StatusSAAlias, chasm.SearchAttributeFieldLowCardinalityKeyword01)
-	TaskQueueSearchAttribute = chasm.NewSearchAttributeKeyword(TaskQueueSAAlias, chasm.SearchAttributeFieldKeyword02)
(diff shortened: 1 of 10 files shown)


chasm/lib/activity/activity.go
```diff
@@ -69,9 +64,8 @@ type Activity struct {
 	// Pointer to an implementation of the "store". For a workflow activity this would be a parent
 	// pointer back to the workflow. For a standalone activity this is nil (Activity itself
 	// implements the ActivityStore interface).
-	// TODO(saa-preview): revisit a standalone activity pointing to itself once we handle storing it more efficiently.
 	// TODO(saa-preview): figure out better naming.
-	Store chasm.Field[ActivityStore]
+	Store chasm.ParentPtr[ActivityStore]
```
yycptt, line 68: 👍

chasm/parent_pointer.go
```diff
@@ -50,8 +50,8 @@ func (p ParentPtr[T]) Get(chasmContext Context) T {
 func (p ParentPtr[T]) TryGet(chasmContext Context) (T, bool) {
 	var nilT T
 	if p.Internal.currentNode == nil {
-		// nolint:forbidigo // Panic is intended here for framework error handling.
-		panic(serviceerror.NewInternal("parent pointer not initialized yet"))
+		// ParentPtr not initialized
+		return nilT, false
```
yycptt, line 54: The component logic calling TryGet() on a newly created node?

chasm/parent_pointer.go
dandavison, line 41: (Comment on existing code): would `"parent component not found"` be a better error message here?

chasm/parent_pointer.go
yycptt: ah I see, it happens in a test and framework doesn't have a chance to kick in and initialize the node. ok makes sense.

PR #9050, opened 2026-01-16
Fix forwarding for nexus completion request headers

bergundy:
## What changed?

Properly populate callback headers in completion request.

### DIFF
--- a/components/nexusoperations/frontend/handler.go
+++ b/components/nexusoperations/frontend/handler.go
@@ -306,7 +306,7 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "invalid operation state: %q", r.State)
 	}
 
-	if forwardReq.Header != nil {
+	if forwardReq.Header == nil {
 		forwardReq.Header = make(http.Header)
 	}
 	forwardReq.Header.Set(interceptor.DCRedirectionApiHeaderName, "true")


components/nexusoperations/frontend/handler.go
```diff
@@ -305,8 +306,8 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "invalid operation state: %q", r.State)
 	}
 
-	if r.HTTPRequest.Header != nil {
-		forwardReq.Header = r.HTTPRequest.Header.Clone()
+	if forwardReq.Header != nil {
```
bergundy:
```suggestion
	if forwardReq.Header == nil {
```

components/nexusoperations/frontend/handler.go
```diff
@@ -305,8 +306,8 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "invalid operation state: %q", r.State)
 	}
 
-	if r.HTTPRequest.Header != nil {
```
bergundy, line 308: `http.NewRequestWithContext` populates an empty `Header` object.

pdoerner: I'm not sure what this change is aiming to do? It's checking if the original HTTP request header is not nil, not the newly created `forwardReq` so it shouldn't matter that `httpNewRequestWithContext` creates an empty header

bergundy: See line 277 above, that's the main fix. The code here is just to prevent assigning to a nil map.

bergundy: I see what you're saying. I got confused. I may change this to remove the redundant in that case.

```
&nexusrpc.OperationCompletionUnsuccessful{
			Header:         httpHeaderToNexusHeader(r.HTTPRequest.Header),
```

bergundy:
The fix isn't needed, @pdoerner will follow up.

PR #9053, opened 2026-01-16
Forward original HTTP headers for Nexus CompleteOperation

pdoerner:
## What changed?
Forward originally received HTTP headers for Nexus CompleteOperation requests.

## Why?
This avoids header sanitization that occurs in the interceptors as those headers are still needed on the forwarded request.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/components/nexusoperations/frontend/handler.go
+++ b/components/nexusoperations/frontend/handler.go
@@ -308,6 +308,9 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 	}
 
 	forwardReq.Header = rCtx.originalHeaders
+	if forwardReq.Header == nil {
+		forwardReq.Header = make(http.Header, 1)
+	}
 	forwardReq.Header.Set(interceptor.DCRedirectionApiHeaderName, "true")
 
 	resp, err := client.Do(forwardReq)


components/nexusoperations/frontend/handler.go
```diff
@@ -289,7 +293,7 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 			return nexus.HandlerErrorf(nexus.HandlerErrorTypeInternal, "internal error")
 		}
 		c := &nexusrpc.OperationCompletionUnsuccessful{
-			Header:         httpHeaderToNexusHeader(r.HTTPRequest.Header),
+			Header:         httpHeaderToNexusHeader(rCtx.originalHeaders),
```
bergundy: Instead of populating headers in the separate case statements, assign them below. (line 308). That will avoid the lossy conversion to nexus headers.

components/nexusoperations/frontend/handler.go
```diff
@@ -305,9 +307,7 @@ func (h *completionHandler) forwardCompleteOperation(ctx context.Context, r *nex
 		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "invalid operation state: %q", r.State)
 	}
 
-	if r.HTTPRequest.Header != nil {
-		forwardReq.Header = r.HTTPRequest.Header.Clone()
-	}
+	forwardReq.Header = rCtx.originalHeaders
```
bergundy, line 310:
```suggestion
	forwardReq.Header = rCtx.originalHeaders
	if forwardReq.Header == nil {
		orwardReq.Header = make(http.Header, 1)
	}
```

PR #9054, opened 2026-01-16
add admin batch delete/repliate/rebuild

feiyang3cat:
## What changed?
apply batching feature to admin pre-existing single delete/rebuild/replicate 

## Why?
efficiency

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/proto/internal/temporal/server/api/adminservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/adminservice/v1/request_response.proto
@@ -651,7 +651,7 @@ message BatchOperationDelete {
 }
 
 message BatchOperationReplicate {
-  // target clusters cannot be empty
+  // if target_clusters is empty, replicates to all configured remote clusters.
   repeated string target_clusters = 1;
 }
 
--- a/api/adminservice/v1/request_response.pb.go
+++ b/api/adminservice/v1/request_response.pb.go
@@ -5538,7 +5538,7 @@ func (*BatchOperationDelete) Descriptor() ([]byte, []int) {
 
 type BatchOperationReplicate struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
-	// target clusters cannot be empty
+	// if target_clusters is empty, replicates to all configured remote clusters.
 	TargetClusters []string `protobuf:"bytes,1,rep,name=target_clusters,json=targetClusters,proto3" json:"target_clusters,omitempty"`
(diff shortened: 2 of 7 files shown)


proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -631,6 +631,11 @@ message StartAdminBatchOperationRequest {
   // The admin batch operation to perform.
   oneof operation {
     BatchOperationRefreshTasks refresh_tasks_operation = 10;
+
+    // Add admin batch delete, replicate and rebuild
+    BatchOperationDelete delete_operation = 11;
+    BatchOperationReplicate replicate_operation = 12;
+    BatchOperationRebuild rebuild_operation = 13;
```
dnr, line 638: are these really safe to expose to all users? even replicate?

feiyang3cat: they are actually already exposed in the adminservice though not yet in batching @yycptt

dnr: right, but users don't have access to adminservice. they do have access to the batch activity handler.

dnr: that's not enough to actually prevent it, though, they can send activities to the pernsworker tq from their own workflow

feiyang3cat: hi, I am new to this context. @dnr how could users leverage admin activities?

yycptt: oh that's a good point. ok we need to block that as well (like the activity task queue can not be the pernsworker Q).

dnr: there's a warning on line 609 but I'm worried it's not going to be effective for the future (even if these particular new operations are safe, the next one someone wants to add may not be). I wonder if we should consider building an admin batch feature that's actually properly separated from user batch.

dnr: IMO trying to plug the various holes as we think of them is not a good security strategy. Our api surface is too complicated and has not been designed with task queue separation as a security boundary. We should run these in a different namespace (probably temporal-system)

PR #9055, opened 2026-01-16
Add health check handler for worker service

### LINKED ISSUE #2582: Add health check handler for worker service
Looks as currently worker service https://github.com/temporalio/temporal/tree/master/service/worker
does not expose a gRPC handler and the health check method like matching and history services do. 

Please add this if possible.

yiminc:
## Summary
- Add gRPC health check endpoint to worker service on port 7239
- Worker service now exposes health status like frontend, history, and matching services
- Enables Kubernetes/container orchestration health probes for worker service

## Changes
- Add `server`, `grpcListener`, `healthServer` fields to worker Service struct
- Register health server in `Start()` with `SERVING` status
- Set `NOT_SERVING` status and gracefully stop in `Stop()`
- Add `ServerProvider` in fx.go for dependency injection
- Add unit test for health check functionality

## Test plan
- [x] Unit test `TestHealthCheck` passes
- [x] Linting passes
- [x] Manual test against local server:
$ grpcurl -plaintext -d '{"service": "temporal.api.workflowservice.v1.WorkerService"}' localhost:7239 grpc.health.v1.Health/Check
{
  "status": "SERVING"
}

Fixes #2582

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/service/worker/fx.go
+++ b/service/worker/fx.go
@@ -7,10 +7,12 @@ import (
 	"go.temporal.io/server/api/adminservice/v1"
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/client"
+	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/cluster"
 	"go.temporal.io/server/common/config"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/membership"
 	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/common/namespace"
@@ -176,6 +178,10 @@ func ServiceLifetimeHooks(lc fx.Lifecycle, svc *Service) {
 	lc.Append(fx.StartStopHook(svc.Start, svc.Stop))
 }
 
-func ServerProvider() *grpc.Server {
-	return grpc.NewServer()
+func ServerProvider(rpcFactory common.RPCFactory, logger log.Logger) *grpc.Server {
+	opts, err := rpcFactory.GetInternodeGRPCServerOptions()
+	if err != nil {
+		logger.Fatal("Failed to get gRPC server options", tag.Error(err))
+	}
+	return grpc.NewServer(opts...)
 }


service/worker/service.go
```diff
@@ -25,8 +26,14 @@ import (
 	"go.temporal.io/server/service/worker/parentclosepolicy"
 	"go.temporal.io/server/service/worker/replicator"
 	"go.temporal.io/server/service/worker/scanner"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/reflection"
 )
 
+const serviceName = "temporal.api.workflowservice.v1.WorkerService"
```
bergundy: Doesn't matter because it's not a real service but 🤷 . BTW we have the same made up service problem for matching.

```suggestion
const serviceName = "temporal.server.api.workerservice.v1.WorkerService"
```

service/worker/service_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+)
+
+func TestHealthCheck(t *testing.T) {
```
bergundy: This is just checking grpc heath server, not the worker code...

yiminc: That is fine. The major usage of this health check is for the deployment tool to be happy. We may want to hook up in the future to actually wait until the worker is connected to server, but that will be a future follow up.

dnr: Actually the way Start is written, it will block until the sdk has connected to at least the temporal-system namespace.

I don't think this test has any value here. It's testing someone else's code

service/worker/service.go
```diff
@@ -244,6 +263,18 @@ func (s *Service) Start() {
 		s.workerServiceResolver,
 	)
```
dnr, line 262: I looked through this to see what Start would actually wait for... this function is making calls to clusterMetadata.Start() and namespaceRegistry.Start(), which are handled by fx, so those are redundant. Let's remove them.

I actually just suggested in #8948  that we remove the atomic status thing for namespace registry because I didn't realize there were other calls to namespaceRegistry.Start() 🙁 Can we remove that extra call in this PR and get this in the release?

service/worker/service_test.go
yiminc: My purpose of this test is to verify that heath check service is correctly wired up.

dnr: But the test isn't touching any of our code, it's pure grpc. Am I missing something?

service/worker/service.go
dnr: I fixed this in #9070, so no worries for this pr

service/worker/service_test.go
yiminc: oh my, I didn't read test code carefully enough. Let me update.

service/worker/service.go
yiminc: Hmmm, do you think this is the only other place? Maybe keep that atomic call is better?

tests/testcore/functional_test_base_test.go
```diff
@@ -51,3 +55,23 @@ func (s *FunctionalTestBaseSuite) TestWithFxOptionsForService() {
 	s.Equal(primitives.HistoryService, s.historyServiceName)
 	s.Equal(primitives.WorkerService, s.workerServiceName)
 }
+
+func (s *FunctionalTestBaseSuite) TestWorkerServiceHealthCheck() {
+	// This test verifies that the worker service exposes a working gRPC health check endpoint.
+	// The health check service name matches the one defined in service/worker/service.go.
+	const workerServiceHealthCheckName = "temporal.api.workflowservice.v1.WorkerService"
```
dnr: just make the other const public instead of duplicating it?

tests/testcore/functional_test_base_test.go
```diff
@@ -51,3 +55,23 @@ func (s *FunctionalTestBaseSuite) TestWithFxOptionsForService() {
 	s.Equal(primitives.HistoryService, s.historyServiceName)
 	s.Equal(primitives.WorkerService, s.workerServiceName)
 }
+
+func (s *FunctionalTestBaseSuite) TestWorkerServiceHealthCheck() {
```
dnr, line 61: this is fine, but just something to think about: would this test (and we could add this for matching+history+frontend too) make more sense to do in `temporal/server_test.go` instead of here?

tests/testcore/functional_test_base_test.go
```diff
@@ -51,3 +55,23 @@ func (s *FunctionalTestBaseSuite) TestWithFxOptionsForService() {
...
+
+	conn, err := grpc.NewClient(
+		s.WorkerGRPCAddress(),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	s.NoError(err)
+	defer func() { _ = conn.Close() }()
+
+	healthClient := healthpb.NewHealthClient(conn)
+	resp, err := healthClient.Check(context.Background(), &healthpb.HealthCheckRequest{
```
dnr: is this guaranteed to be up by the time the test func is run, or should we use Eventually here?

service/worker/fx.go
```diff
@@ -173,3 +175,7 @@ func VisibilityManagerProvider(
 func ServiceLifetimeHooks(lc fx.Lifecycle, svc *Service) {
 	lc.Append(fx.StartStopHook(svc.Start, svc.Stop))
 }
+
+func ServerProvider() *grpc.Server {
+	return grpc.NewServer()
```
dnr: sorry I missed this before:

the ServerProvider in matching and history look like this:
```
func ServerProvider(grpcServerOptions []grpc.ServerOption) *grpc.Server {
	return grpc.NewServer(grpcServerOptions...)
}
```
shouldn't this be the same to pick up all our server options?
maybe the interceptors aren't applicable because we don't have a workerservice, only healthservice, but TLS is also in there and should be consistent with the others.

yiminc: I'm actually not sure if it makes sense to install any of the interceptor here because we don't really expect any incoming calls except this health check, and it is only going to be accessible from within the cluster because the port won't be open external.

dnr: If it was just interceptors I might agree, although there's no harm in them. But I don't think the grpc TLS config should be different on history/matching vs worker. It might even be against policy to have a listening port without TLS in many environments, even if there's nothing interesting on it besides a health check. TLS is exactly how a port is made "only accessible from within the cluster" in some environments.

yiminc: Added the TLS config without adding other interceptors.

bergundy:
@yiminc as a follow up we need to update the helm charts to start using this health check.

service/worker/fx.go
```diff
@@ -7,10 +7,12 @@ import (
 	"go.temporal.io/server/api/adminservice/v1"
 	"go.temporal.io/server/chasm"
 	"go.temporal.io/server/client"
+	"go.temporal.io/server/common"
```
dnr: dummy

service/worker/fx.go
dnr: It still feels like making it totally consistent is better, but it doesn't make a difference in practice so I'm fine with this

PR #9058, opened 2026-01-16
add MigrateSchedule proto and conversion functions

chaptersix:
## What changed?

Added proto definitions and conversion functions for V1 to V2 schedule migration.

## Why?

This is the foundation for the V1→V2 schedule migration feature. The `MigrateScheduleRequest` proto captures all scheduler state needed to recreate a schedule in the CHASM architecture.


chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,526 @@
...
+	State *schedulespb.InternalState
+
+	// From workflow execution (search attributes and memo)
+	SearchAttributes map[string]*commonpb.Payload
+	Memo             map[string]*commonpb.Payload
+}
+
+// CHASMState represents the complete state for the CHASM scheduler.
+type CHASMState struct {
+	// Scheduler root component state
```
lina-temporal: I think these comments for the state are a little redundant.

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,526 @@
...
+	for i, wf := range runningWorkflows {
+		// We don't have the original nominal/actual times for running workflows,
+		// so we use migrationTime as a placeholder
+		bufferedStarts[i] = &schedulespb.BufferedStart{
+			NominalTime: timestamppb.New(migrationTime),
+			ActualTime:  timestamppb.New(migrationTime),
+			StartTime:   timestamppb.New(migrationTime),
+			WorkflowId:  wf.WorkflowId,
+			RunId:       wf.RunId,
+			RequestId: schedulescommon.GenerateRequestID(
```
lina-temporal, line 264: Can you add a comment that this request ID will be used as the Nexus callback request ID? (I'm presuming you'll use `AttachRequestID` with this in the follow-up PR)

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,526 @@
...
+				scheduleID,
+				conflictToken,
+				"migrated-completed",
+				action.ScheduleTime.AsTime(),
+				action.ActualTime.AsTime(),
+			),
+			Attempt: 1,
+			Completed: &schedulespb.CompletedResult{
+				Status:    action.StartWorkflowStatus,
+				CloseTime: action.ActualTime, // Use actual time as approximation
```
lina-temporal: The only place CloseTime is used is to drive the `ScheduleActionDelay` metric (time between "could have kicked off start" and "did kick off start"), so I think this would probably be better served by using `migrationTime` as the stand-in.

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,526 @@
...
+				StartWorkflowResult: &commonpb.WorkflowExecution{
+					WorkflowId: start.GetWorkflowId(),
+					RunId:      start.GetRunId(),
+				},
+				StartWorkflowStatus: start.GetCompleted().GetStatus(),
+			})
+		}
+	}
+
+	// Convert pending buffered starts: drop CHASM-specific fields
```
lina-temporal: Ditto for these comments, a bit verbose imo; maybe just a quick pass to prune some of these off in favor of conciseness?

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,526 @@
...
+	// From InternalState
+	State *schedulespb.InternalState
+
+	// From workflow execution (search attributes and memo)
+	SearchAttributes map[string]*commonpb.Payload
+	Memo             map[string]*commonpb.Payload
+}
+
+// CHASMState represents the complete state for the CHASM scheduler.
+type CHASMState struct {
```
lina-temporal: Is the plan to implement converters in their respective packages (e.g., to go from `CHASMState` to an actual scheduler tree)? In the follow-up PRs to wire this up, consider if there's a way to maybe keep these two transient record structs private to the migration package, or potentially remove them in favor of using only the permanent record structs; they definitely simplify testing, so getting rid of them entirely may not be appetizing.

chaptersix: Yes, that's it

proto/internal/temporal/server/api/adminservice/v1/service.proto
```diff
@@ -165,4 +165,8 @@ service AdminService {
     rpc DescribeTaskQueuePartition (DescribeTaskQueuePartitionRequest) returns (DescribeTaskQueuePartitionResponse) {}
 
     rpc ForceUnloadTaskQueuePartition (ForceUnloadTaskQueuePartitionRequest) returns (ForceUnloadTaskQueuePartitionResponse) {}
+
+    // MigrateSchedule migrates a V1 workflow-backed schedule to V2 CHASM.
+    // This signals the V1 workflow to run the migration activity.
+    rpc MigrateSchedule (MigrateScheduleRequest) returns (MigrateScheduleResponse) {}
```
lina-temporal: MigrateSchedule should either be usable bidirectionally, or name the API more specifically (MigrateV1Scheduler).

chasm/lib/scheduler/proto/v1/request_response.proto
```diff
@@ -71,3 +73,27 @@ message ListScheduleMatchingTimesRequest {
...
+
+    // Schedule ID.
+    string schedule_id = 3;
+
+    // Migrated CHASM state components.
+    SchedulerState scheduler_state = 4;
+    GeneratorState generator_state = 5;
+    InvokerState invoker_state = 6;
+    map<string, BackfillerState> backfillers = 7;
+    LastCompletionResult last_completion_result = 8;
```
lina-temporal: Similar comments as with the operation itself. If you go with the approach of using the same RPC operation for both, maybe move the CHASM-specific/Legacy-specific fields into their own message definitions.

chasm/lib/scheduler/migration/migration_test.go
```diff
@@ -0,0 +1,133 @@
...
+	failurepb "go.temporal.io/api/failure/v1"
+	schedulepb "go.temporal.io/api/schedule/v1"
+	workflowpb "go.temporal.io/api/workflow/v1"
+	schedulespb "go.temporal.io/server/api/schedule/v1"
+	schedulerpb "go.temporal.io/server/chasm/lib/scheduler/gen/schedulerpb/v1"
+	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func testSchedule() *schedulepb.Schedule {
```
lina-temporal:
```suggestion
func newTestSchedule() *schedulepb.Schedule {
```

(such a minor nitpick, just differentiating from test names)

chasm/lib/scheduler/migration/migration_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.NotNil(t, req.InvokerState)
+	require.Len(t, req.InvokerState.BufferedStarts, 2) // 1 buffered + 1 running
+	for _, start := range req.InvokerState.BufferedStarts {
+		require.NotEmpty(t, start.RequestId)
+		require.NotEmpty(t, start.WorkflowId)
+	}
+
+	// Backfillers
+	require.Len(t, req.Backfillers, 1)
+	for id, backfiller := range req.Backfillers {
```
lina-temporal: (minor) assertion for startTime?

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -641,3 +641,21 @@ message StartAdminBatchOperationResponse {
 // This regenerates all pending tasks for each execution.
 message BatchOperationRefreshTasks {
 }
+
+// MigrateScheduleRequest triggers migration of a V1 workflow-backed schedule
+// to a V2 CHASM-backed schedule.
```
lina-temporal: This RPC will be bi-directional, right?

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,514 @@
...
+		ConflictToken: state.ConflictToken,
+		Closed:        false,
+	}
+
+	generatorState := &schedulerpb.GeneratorState{
+		LastProcessedTime: common.CloneProto(state.LastProcessedTime),
+		FutureActionTimes: nil, // Regenerated by GeneratorTask
+	}
+
+	chasmBufferedStarts := convertBufferedStartsLegacyToCHASM(
```
lina-temporal:
```suggestion
	pendingBufferedStarts := convertBufferedStartsLegacyToCHASM(
```

nit, aligning with over variable names

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,514 @@
...
+}
+
+// CHASMToSchedulerMigrationState extracts CHASM (V2) scheduler state into a
+// SchedulerMigrationState proto. This creates a deep copy of all state.
+//
+// This function is the inverse of applying a SchedulerMigrationState - it extracts
+// the current CHASM scheduler state back into the request format. Use cases include:
+//   - Re-migration scenarios (e.g., migrating to a different shard)
+//   - State inspection and debugging
+//   - Creating snapshots of scheduler state
```
lina-temporal: I think this is kind of misleading. The primary, and likely only use of this, will be for facilitating rollbacks from V2 -> V1.

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,514 @@
...
+		InvokerState:         common.CloneProto(invoker),
+		Backfillers:          cloneBackfillers(backfillers),
+		LastCompletionResult: common.CloneProto(lastCompletionResult),
+		SearchAttributes:     searchAttributes,
+		Memo:                 memo,
+	}
+}
+
+// SchedulerMigrationStateToLegacyStartScheduleArgs converts CHASM scheduler state
+// to V1 StartScheduleArgs. The migrationTime parameter is used to initialize
```
lina-temporal: If SchedulerMigrationState is meant to be stack-agnostic, I think it'd be more correct to say "converts migration state to V1 StartScheduleArgs."

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,514 @@
...
+	return &schedulespb.StartScheduleArgs{
+		Schedule: schedule,
+		Info:     info,
+		State:    state,
+	}
+}
+
+// -----------------------------------------------------------------------------
+// Helper functions for state conversion
+// -----------------------------------------------------------------------------
```
lina-temporal: style - let's not use banners like this, I don't think it's very idiomatic to Go since they're free-floating, and won't play well with refactoring.

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,514 @@
...
+
+func cloneBackfillers(backfillers map[string]*schedulerpb.BackfillerState) map[string]*schedulerpb.BackfillerState {
```
lina-temporal: Think you can just use `common.CloneProtoMap`.

chasm/lib/scheduler/migration/migration.go
```diff
@@ -0,0 +1,514 @@
...
+				WorkflowId: start.GetWorkflowId(),
+				RunId:      start.GetRunId(),
+			})
+		}
+	}
+
+	return buffered, running, recent
+}
+
+func convertBackfillersToLegacy(
```
lina-temporal: consistency with `convertBackfillsLegacyToCHASM`:

```suggestion
func convertBackfillersCHASMToLegacy(
```

chasm/lib/scheduler/migration/migration_test.go
```diff
@@ -0,0 +1,216 @@
...
+	require.Equal(t, int64(42), migrationState.SchedulerState.ConflictToken)
+	require.False(t, migrationState.SchedulerState.Closed)
+
+	// Generator state
+	require.NotNil(t, migrationState.GeneratorState)
+	require.Equal(t, now, migrationState.GeneratorState.LastProcessedTime.AsTime())
+
+	// Invoker state - buffered starts + running workflows
+	require.NotNil(t, migrationState.InvokerState)
+	require.Len(t, migrationState.InvokerState.BufferedStarts, 2) // 1 buffered + 1 running
```
lina-temporal: let's assert that we see buffered, completed, and running represented.

chasm/lib/scheduler/migration/migration_test.go
```diff
@@ -0,0 +1,216 @@
...
+		},
+	}
+	searchAttrs := map[string]*commonpb.Payload{"Attr": {Data: []byte("value")}}
+	memo := map[string]*commonpb.Payload{"Memo": {Data: []byte("memo")}}
+
+	migrationState := LegacyToSchedulerMigrationState(newTestSchedule(), info, state, searchAttrs, memo, now)
+
+	// Scheduler state
+	require.NotNil(t, migrationState)
+	require.NotNil(t, migrationState.SchedulerState)
```
lina-temporal, line 85: Can we also set the initial scheduler to paused, and assert against that too?

chasm/lib/scheduler/migration/migration_test.go
```diff
@@ -0,0 +1,216 @@
...
+				EndTime:       timestamppb.New(now),
+				OverlapPolicy: enumspb.SCHEDULE_OVERLAP_POLICY_ALLOW_ALL,
+			},
+		},
+		LastCompletionResult: &commonpb.Payloads{
+			Payloads: []*commonpb.Payload{{Data: []byte("result")}},
+		},
+		ContinuedFailure: &failurepb.Failure{Message: "last failure"},
+	}
+	info := &schedulepb.ScheduleInfo{
```
lina-temporal, line 64: Let's add a RecentActions here for completeness.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
chaptersix: yes

chasm/lib/scheduler/migration/migration.go
chaptersix: Ah, I missed that. I'll remove it

chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -89,3 +89,17 @@ message LastCompletionResult {
     temporal.api.common.v1.Payload success = 1;
     temporal.api.failure.v1.Failure failure = 2;
 }
+
+// SchedulerMigrationState is a stack-agnostic interchange format for migrating
+// scheduler state between V1 (workflow-backed) and V2 (CHASM) implementations.
+message SchedulerMigrationState {
```
chaptersix, line 95: is  this the right place to put the proto?

lina-temporal: I think so, since the migration logic lives within the CHASM package too.

chasm/lib/scheduler/migration/migration.go
```diff
@@ -47,8 +47,14 @@ func LegacyToSchedulerMigrationState(
 	infoClone.RunningWorkflows = nil
 	infoClone.RecentActions = nil
 
+	scheduleClone := common.CloneProto(schedule)
+	if scheduleClone.State == nil {
+		scheduleClone.State = &schedulepb.ScheduleState{}
+	}
+	scheduleClone.State.Paused = true
```
lina-temporal: We need to carry over the original paused state.

chasm/lib/scheduler/migration/migration_test.go
```diff
@@ -66,73 +66,78 @@ func TestLegacyToMigrateScheduleRequest(t *testing.T) {
...
+	migrationState := LegacyToSchedulerMigrationState(newTestSchedule(), info, state, searchAttrs, memo, now)
 
 	// Scheduler state
-	require.NotNil(t, req.MigrationState)
-	require.NotNil(t, req.MigrationState.SchedulerState)
-	require.Equal(t, int64(42), req.MigrationState.SchedulerState.ConflictToken)
-	require.False(t, req.MigrationState.SchedulerState.Closed)
+	require.NotNil(t, migrationState)
+	require.NotNil(t, migrationState.SchedulerState)
+	require.True(t, migrationState.SchedulerState.Schedule.State.Paused, "schedule should be paused after migration")
```
lina-temporal: ah, I think I understand the confusion - the pause is applied to the original (source) schedule, not the destination. This is so because the scheduler workflow is roughly structured like:

```
- process signals
- process buffered starts/run loop
- determine whether to exit/continue-as-new/loop
```

.. so if we do our migration in the "process signals" part, and then plan to exit the workflow cleanly, we'll want to pause the original workflow so that its run loop does nothing before it closes out.

lina-temporal (approved):

PR #9059, opened 2026-01-16
Populate namespace in PollActivityTaskQueueResponse

fretz12:
## What changed?
Populate namespace in PollActivityTaskQueueResponse

## Why?
Namespace field required by SDK

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -209,7 +209,6 @@ func (a *Activity) PopulateRecordStartedResponse(
 	namespace string,
 	response *historyservice.RecordActivityTaskStartedResponse,
 ) error {
-	response.WorkflowNamespace = namespace
 	lastHeartbeat, _ := a.LastHeartbeat.TryGet(ctx)
 	if lastHeartbeat != nil {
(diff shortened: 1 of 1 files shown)


chasm/lib/activity/activity.go
```diff
@@ -198,12 +198,17 @@ func (a *Activity) HandleStarted(ctx chasm.MutableContext, request *historyservi
 		return nil, err
 	}
 	response := &historyservice.RecordActivityTaskStartedResponse{}
-	err := a.StoreOrSelf(ctx).PopulateRecordStartedResponse(ctx, ctx.ExecutionKey(), response)
+	err := a.StoreOrSelf(ctx).PopulateRecordStartedResponse(ctx, ctx.ExecutionKey(), request.GetPollRequest().GetNamespace(), response)
```
bergundy, line 201: I think it's fine to use the namespace in the request but there may be edge cases I'm not aware of with namespace renaming that may not be handled here. Maybe it would have been safer to resolve the namespace name from the ID.

Not blocking but worth checking with @yycptt.

PR #9062, opened 2026-01-16
[Flake resolution]: Resolve flakiness on the TestTaskQueueStatsSuite 

Shivs11:
## What changed?
- Tests that are testing the statistics of current/ramping versions were altered to have their checks relaxed. They were relaxed in the sense that we now allow these stats to be over-counted since it's theoretically possible for them to do so.

## Why?
- When calculating per-version stats, each partition has a very narrow view of them. An unequal distribution of tasks in each partition could lead to an aggregated count of the backlog to be more than the total number of scheduled tasks on the workflow. This is possible because each task queue partition calculates a version's stats using the `math.ceil` function (since we are okay with an overcount).

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- None

---

> [!NOTE]
> Adjusts TaskQueueStats tests to tolerate partition-induced overcounting and stabilize flakey assertions.
> 
> - Introduces relaxed helpers `requireWDVTaskQueueStatsRelaxed` and `requireLegacyTaskQueueStatsRelaxed` that allow up to `numPartitions` extra backlog items (due to per-partition ceil) and replaces strict assertions across relevant tests
> - Passes `numPartitions` through call sites; updates test data (task queue names via `common.GenerateRandomString`, simplified build IDs/deployment names)
> - Keeps behavior validation (rates/backlog across current, ramping, inactive, and legacy modes) while allowing slight overcounts, especially in multi-partition scenarios
> - Imports `go.temporal.io/server/common` for random string generation
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 87b889a92636c44a81313212b56066a704ea700e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/task_queue_stats_test.go
+++ b/tests/task_queue_stats_test.go
@@ -19,6 +19,7 @@ import (
 	workflowpb "go.temporal.io/api/workflow/v1"
 	"go.temporal.io/api/workflowservice/v1"
 	deploymentspb "go.temporal.io/server/api/deployment/v1"
+	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/worker_versioning"
 	"go.temporal.io/server/tests/testcore"
@@ -56,19 +57,14 @@ type (
 	TaskQueueExpectationsByType map[enumspb.TaskQueueType]TaskQueueExpectations
 )
 
-func shortRandomizedName(prefix string) string {
-	// Keep names short because they get embedded (and base64-encoded) into internal persisted task_queue_id strings
-	return fmt.Sprintf("%s-%s", prefix, uuid.NewString()[:8])
-}
-
 // TODO(pri): remove once the classic matcher is removed
 func TestTaskQueueStats_Classic_Suite(t *testing.T) {
-	// t.Parallel()
+	t.Parallel()
 	suite.Run(t, &TaskQueueStatsSuite{usePriMatcher: false})
 }
 
 func TestTaskQueueStats_Pri_Suite(t *testing.T) {
-	// t.Parallel()
+	t.Parallel()
 	suite.Run(t, &TaskQueueStatsSuite{usePriMatcher: true})
 }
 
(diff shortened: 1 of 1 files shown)


tests/task_queue_stats_test.go
```diff
@@ -56,6 +56,11 @@ type (
 	TaskQueueExpectationsByType map[enumspb.TaskQueueType]TaskQueueExpectations
 )
 
+func shortRandomizedName(prefix string) string {
+	// Keep names short because they get embedded (and base64-encoded) into internal persisted task_queue_id strings
+	return fmt.Sprintf("%s-%s", prefix, uuid.NewString()[:8])
```
stephanos: Isn't that going to strip away UUID all the time; seeing that most prefixes are longer than 8 chars?

tests/task_queue_stats_test.go
```diff
@@ -645,10 +663,10 @@ func (s *TaskQueueStatsSuite) inactiveVersionDoesNotAbsorbUnversionedBacklog(num
...
 	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Millisecond) // zero means no TTL
 
-	tqName := testcore.RandomizeStr("inactive-version")
-	deploymentName := tqName + "-deployment"
-	currentBuildID := "current-build-id"
-	inactiveBuildID := "inactive-build-id"
+	tqName := shortRandomizedName("inactive-version")
+	deploymentName := shortRandomizedName(tqName + "-deployment")
+	currentBuildID := shortRandomizedName(tqName + "-current-build-id")
+	inactiveBuildID := shortRandomizedName(tqName + "-inactive-build-id")
```
stephanos: If the task queue (and therefore deployment name) are already randomized, why do we need to randomize the build IDs, too?

Shivs11: yes we do; there are two suites that are running in parallel and there could be version conflicts (which actually were one of reason for the flakes)

tests/task_queue_stats_test.go
Shivs11: good point, sorry i missed this; I am going to change this to return something like the first 50 characters which seems fine since no prefixes are that long and we want each task queue name across the two suites to be different.

tests/task_queue_stats_test.go
```diff
@@ -126,7 +127,7 @@ func (s *TaskQueueStatsSuite) TestAddMultipleTasks_MultiplePartitions_ValidateSt
 	s.OverrideDynamicConfig(dynamicconfig.MatchingLongPollExpirationInterval, 10*time.Second)
 	s.OverrideDynamicConfig(dynamicconfig.TaskQueueInfoByBuildIdTTL, 1*time.Hour) // using a long TTL to verify caching
 
-	tqName := testcore.RandomizeStr("backlog-counter-task-queue")
+	tqName := "tq-" + common.GenerateRandomString(5)
```
dnr, line 130: let's standardize on using testvars for these? (I know, the names that testvars returns are too long, but let's fix that in testvars instead?)

PR #9066, opened 2026-01-16
Fix try submit issue in sequential task scheduler

yux0:
## What changed?
Fix try submit issue in sequential task scheduler

## Why?
With concurrent calls to TrySubmit in sequential task scheduler, there is an issue a task can be delay scheduled. This changes limit the concurrency to provide correctness atm.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
No risk as there is no caller atm

### DIFF
--- a/common/tasks/sequential_scheduler_test.go
+++ b/common/tasks/sequential_scheduler_test.go
@@ -219,20 +219,26 @@ func (s *sequentialSchedulerSuite) TestTrySubmitLockLeak() {
 	mockTask1 := NewMockTask(s.controller)
 	mockTask1.EXPECT().RetryPolicy().Return(s.retryPolicy).AnyTimes()
 
+	// Track whether first TrySubmit completed
+	var firstTrySubmitDone atomic.Bool
+	var firstTrySubmitResult atomic.Bool
+
 	// Try to submit - this will timeout after 200ms
 	go func() {
 		result := scheduler.TrySubmit(mockTask1)
-		s.False(result, "First TrySubmit should timeout and return false")
+		firstTrySubmitResult.Store(result)
+		firstTrySubmitDone.Store(true)
 	}()
 
-	// Wait for timeout + a bit more to ensure the spawned goroutine acquires the lock
-	time.Sleep(300 * time.Millisecond)
+	// Wait for the first TrySubmit to timeout and the spawned goroutine to acquire the lock
+	s.Require().Eventually(func() bool {
+		return firstTrySubmitDone.Load()
+	}, 500*time.Millisecond, 10*time.Millisecond, "First TrySubmit should timeout")
 
-	// Release the lock - but the spawned goroutine from TrySubmit has already acquired it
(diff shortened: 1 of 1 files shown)


yycptt:
cc @prathyushpv as you are working on sequential processor as well.

common/tasks/sequential_scheduler.go
```diff
@@ -134,7 +140,22 @@ func (s *SequentialScheduler[T]) Submit(task T) {
...
+	go func() {
+		s.trySubmitLock.Lock()
+		lockCh <- struct{}{}
+	}()
+
+	select {
+	case <-lockCh:
+		defer s.trySubmitLock.Unlock()
+	case <-time.After(trySubmitLockTimeout):
+		return false
```
prathyushpv, line 166: How is `trySubmitLock.Unlock()` called if `case <-time.After(trySubmitLockTimeout):` is executed and returned immediately?

yux0: Yes. The time after will use a new goroutine and a race condition is possible. Let me fix it

common/tasks/sequential_scheduler.go
```diff
@@ -134,7 +140,32 @@ func (s *SequentialScheduler[T]) Submit(task T) {
...
+			// Main goroutine timed out, just unlock and exit
+		}
+	}()
+
+	defer close(unlockCh) // Always signal goroutine when we're done
+
+	select {
+	case <-lockCh:
+		// Lock acquired, proceed with submission
+	case <-time.After(trySubmitLockTimeout):
```
prathyushpv, line 165: looks like this could be memory leak under load. This timer will not be garbage collected until it fires. It maybe better to define time above and stop it in a defer statement.
```
timer := time.NewTimer(trySubmitLockTimeout)
defer timer.Stop()
```

common/tasks/sequential_scheduler.go
```diff
@@ -134,7 +140,32 @@ func (s *SequentialScheduler[T]) Submit(task T) {
 	}
 }
 
+// TrySubmit use mu locking to make it thread safe which has higher latency and not suitable for high throughput
 func (s *SequentialScheduler[T]) TrySubmit(task T) bool {
+	// Try to acquire lock with timeout to prevent concurrent TrySubmit race condition
+	lockCh := make(chan struct{}, 1)
```
prathyushpv, line 146: Why did we decide to go with this approach instead of calling TryLock() function to serialize calls?

common/tasks/sequential_scheduler.go
yux0: As of Go 1.23, the garbage collector can recover unreferenced,
unstopped timers. There is no reason to prefer NewTimer when After will do.

As mentioned in the function doc, This try submit implementation is not suitable for high throughput

common/tasks/sequential_scheduler.go
yux0: The current try lock could reject a request even if the channel is not full. here I try to do is a trylock with a timeout.

prathyushpv (approved):

PR #9075, opened 2026-01-19
Fix flaky worker deployment tests

rkannan82:
## What changed?
Fixes https://github.com/temporalio/temporal/actions/runs/21057850575/job/60557710357?pr=9055

1. TestDeleteWorkerDeployment_ValidDelete: The test started a poller goroutine that kept polling indefinitely, then checked if pollers had disappeared. The poller was still active, so the check failed.

   Fix: Cancel the poller context before checking that pollers are gone.

   Verified by reproducing: Increasing PollerHistoryTTL to 10s and removing the cancel causes consistent failure.

2. TestDescribeWorkerDeployment_SetCurrentVersion: The test had a 10 second timeout to wait for deployment creation, which was insufficient under load.

   Fix: Use ensureCreateVersionInDeployment() which has a 60 second timeout. Also added cancellable contexts for proper cleanup.

## Why?
Fix for flaky test runs.
Fixes https://github.com/temporalio/temporal/actions/runs/21057850575/job/60557710357?pr=9055

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None

### DIFF
--- a/tests/worker_deployment_test.go
+++ b/tests/worker_deployment_test.go
@@ -1516,26 +1516,28 @@ func (s *WorkerDeploymentSuite) TestDescribeWorkerDeployment_SetCurrentVersion()
 	secondVersion := tv.WithBuildIDNumber(2)
 
 	version1CreateTime := timestamppb.Now()
-	go s.pollFromDeployment(ctx, firstVersion)
+	// Start poller with cancellable context for deterministic control
+	pollerCtx, pollerCancel := context.WithCancel(ctx)
+	defer pollerCancel()
+	go s.pollFromDeployment(pollerCtx, firstVersion)
 
-	// No current deployment version set.
-	s.EventuallyWithT(func(t *assert.CollectT) {
-		a := require.New(t)
+	// Wait for version to be created deterministically
+	s.ensureCreateVersionInDeployment(firstVersion)
 
-		resp, err := s.FrontendClient().DescribeWorkerDeployment(ctx, &workflowservice.DescribeWorkerDeploymentRequest{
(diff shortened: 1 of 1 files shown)


tests/worker_deployment_test.go
```diff
@@ -1516,26 +1516,28 @@ func (s *WorkerDeploymentSuite) TestDescribeWorkerDeployment_SetCurrentVersion()
...
+	s.ensureCreateVersionInDeployment(firstVersion)
 
-		resp, err := s.FrontendClient().DescribeWorkerDeployment(ctx, &workflowservice.DescribeWorkerDeploymentRequest{
-			Namespace:      s.Namespace().String(),
-			DeploymentName: tv.DeploymentSeries(),
-		})
-		a.NoError(err)
-		a.Equal(worker_versioning.UnversionedVersionId, resp.GetWorkerDeploymentInfo().GetRoutingConfig().GetCurrentVersion()) //nolint:staticcheck // SA1019: old worker versioning
-		a.Nil(resp.GetWorkerDeploymentInfo().GetRoutingConfig().GetCurrentDeploymentVersion())
-	}, time.Second*10, time.Millisecond*1000)
```
Shivs11, line 1532: I think this test needs more work (in terms of fixing): The changes you have introduced shall prevent failures/flakes on the v0 suite but not on the v2. This is because the v0 suite uses sync workflows and the v2 one using async versioning workflows. 

What needs to change in the v2 suite is that the test, right now, wrongly assumes that each setCurrent call is strictly consistent. This won't be the case in v2 and should be changed. However, this is something that I am tracking and does not block your PR.

PR #9076, opened 2026-01-19
Programmable grpc fault injection

stephanos:
## What changed?

Add gRPC fault injection for functional tests.

## Why?

Today this only addresses a single use case, but going forward it offers the ability to write tests for error cases that were difficult to test before (at least not without brittle and slow `time.Sleep`s).

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -107,6 +107,8 @@ message AddWorkflowTaskRequest {
     // (-- api-linter: core::0140::prepositions=disabled
     //     aip.dev/not-precedent: "to" is used to indicate interval. --)
     google.protobuf.Duration schedule_to_start_timeout = 5;
+    // True if this is a speculative workflow task. Used for testing fault injection.
+    bool speculative = 6;
```
stephanos: This additional makes speculative workflow task-related tests a tiny bit easier but very little extra overhead cost. It also improves observability/debugging for the same reasons.

tests/update_workflow_test.go
```diff
@@ -1397,73 +1419,24 @@ func (s *WorkflowUpdateSuite) TestStickySpeculativeWorkflowTask_AcceptComplete_S
...
-		Logger:                       env.Logger,
-		T:                            s.T(),
-	}
-
-	// Drain existing WT from regular task queue, but respond with sticky enabled response to enable stick task queue.
-	_, err := poller.PollAndProcessWorkflowTask(testcore.WithRespondSticky, testcore.WithoutRetries)
-	s.NoError(err)
-
-	env.Logger.Info("Sleep 10+ seconds to make sure stickyPollerUnavailableWindow time has passed.")
-	time.Sleep(10*time.Second + 100*time.Millisecond) //nolint:forbidigo
```
stephanos, line 1447: Before the change, this had to sleep for 10s.

tests/testcore/onebox.go
```diff
@@ -128,7 +130,9 @@ func newTemporal(t *testing.T, params *temporalParams) *temporalImpl {
 		hostsByProtocolByService:  params.HostsByProtocolByService,
 		workerConfig:              params.WorkerConfig,
 		replicationStreamRecorder: NewReplicationStreamRecorder(),
+		faultInjector:             rpcfaultinjection.NewRPCFaultGenerator(),
 	}
+	testhooks.Set(impl.testHooks, testhooks.RPCFaultGenerator, impl.faultInjector.Generate, testhooks.GlobalScope)
```
stephanos: This is simpler than only setting the testhook when it's used; the overhead is minimal and limited to tests.

service/frontend/fx.go
```diff
@@ -307,6 +309,9 @@ func GrpcServerOptionsProvider(
 		// TODO: Deprecate WithChainedFrontendGrpcInterceptors and provide a inner custom interceptor
 		unaryInterceptors = append(unaryInterceptors, customInterceptors...)
 	}
+	if faultInterceptor := faultinjection.GRPCUnaryServerInterceptor(testHooks); faultInterceptor != nil {
```
stephanos: We don't want any overhead in production.

tests/testcore/fault_injection.go
```diff
@@ -0,0 +1,115 @@
+package testcore
```
stephanos, line 1: This file integrates the fault injection with Go tests.

tests/testcore/fault_injection.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+// InjectRPCFault registers a fault injection that applies to all services
+// (frontend, history, matching). The fault function determines which requests
+// trigger a fault and what error to return.
+//
+// The fault function is called twice per RPC: before the handler (resp=nil, err=nil)
+// and after. Returning an error before handler short-circuits; returning after
+// modifies the response.
```
stephanos: 🤖 "returning after modifies the response" is a bit misleading for this helper: `RPCFault` can only return an error, and the callback below passes `nil` as the replacement response (line 102). So a post-handler match discards the successful response and turns the call into an error — it never *modifies* a response. Suggest wording it as "returning after the handler replaces the response with that error".

tests/testcore/fault_injection.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// Returns a cleanup function that disables the fault injection when called.
+// The test fails if the fault is never injected before the test completes.
+//
+// Example:
+//
+//	testcore.InjectRPCFault(s.T(), s.GetTestCluster(),
+//	    func(req, _ any, _ error) error {
+//	        r, ok := req.(*matchingservice.AddWorkflowTaskRequest)
```
stephanos: 🤖 This example doesn't compile: `r` is declared and never used (`declared and not used: r`). Should be `_, ok := req.(*matchingservice.AddWorkflowTaskRequest)`. Same snippet is duplicated in `docs/development/testing.md`.

docs/development/testing.md
```diff
@@ -197,7 +197,29 @@ or if there's no SDK support for that API available yet.
 
 You'll find a fully initialized task poller in any functional test suite, look for `s.TaskPoller`.
 
-_NOTE: The previous `testcore.TaskPoller` has been deprecated and should not be used in new code._
+_NOTE_: The previous `testcore.TaskPoller` has been deprecated and should not be used in new code._
```
stephanos: 🤖 Unrelated to this PR's feature, and it breaks the markdown: `_NOTE_: ... new code._` leaves a dangling trailing `_`. The original `_NOTE: ... new code._` rendered the whole note in italics. Either revert this line or write `**NOTE**: The previous ... new code.`

docs/development/testing.md
```diff
@@ -197,7 +197,29 @@ or if there's no SDK support for that API available yet.
...
+The `testcore` package injects faults into gRPC calls by intercepting requests and responses.
+The fault function determines which RPCs trigger a fault and returns the error to inject.
+It returns a cleanup function that can be used to remove the fault again.
+
+**Example:**
+
+```go
+testcore.InjectRPCFault(s.T(), s.GetTestCluster(),
+    func(req, _ any, _ error) error {
+        r, ok := req.(*matchingservice.AddWorkflowTaskRequest)
```
stephanos: 🤖 This example doesn't compile: `r` is declared and never used. Use `_, ok := req.(*matchingservice.AddWorkflowTaskRequest)`.

docs/development/testing.md
```diff
@@ -197,7 +197,29 @@ or if there's no SDK support for that API available yet.
...
+### gRPC fault injection
+
+The `testcore` package injects faults into gRPC calls by intercepting requests and responses.
+The fault function determines which RPCs trigger a fault and returns the error to inject.
+It returns a cleanup function that can be used to remove the fault again.
+
+**Example:**
+
+```go
+testcore.InjectRPCFault(s.T(), s.GetTestCluster(),
```
stephanos: 🤖 The documented entry point is the unscoped one, which is the footgun variant: with no `WithNamespaceID`/`WithNamespaceName` option, `matchesNamespace` returns `true` for every request, so on a shared test cluster the fault also fires for RPCs belonging to other tests' namespaces. Per your own comment on `fault_injection.go`, `TestEnv.InjectRPCFault` is what makes this safe.

Suggest making the example `env.InjectRPCFault(func(req, _ any, _ error) error { ... })` and mentioning `testcore.InjectRPCFault` only as the unscoped escape hatch. Worth noting here too that only unary RPCs are intercepted — streaming calls are unaffected.

common/rpc/faultinjection/grpc_test.go
```diff
@@ -0,0 +1,91 @@
+//go:build test_dep
+
+package faultinjection
```
stephanos: 🤖 The file is named after `grpc.go` but only exercises `GRPCUnaryServerInterceptor` from `interceptor_testdep.go`; `RPCFaultGenerator` itself has no unit tests. The behaviors that are easy to get wrong and cheap to cover here: `RegisterCallback`'s returned cleanup actually removing that entry (and being safe to call twice, since `InjectRPCFault` calls it from both the return value and `t.Cleanup`), first-match-wins across multiple registered callbacks, and unregistering during a concurrent `Generate`. Consider moving these tests to `interceptor_testdep_test.go` and adding a `grpc_test.go` for the generator.

common/rpc/faultinjection/grpc.go
```diff
@@ -0,0 +1,93 @@
...
+//   - req: the request proto message
+//   - resp: the response proto message (nil when called before handler)
+//   - err: the error from handler (nil when called before handler)
+//
+// Returns:
+//   - bool: if true, the callback matched and the returned values should be used.
+//     If false, the callback did not match and other callbacks should be checked.
+//   - any: replacement response (only used if matched is true)
+//   - error: replacement error (only used if matched is true)
+type RPCCallback func(ctx context.Context, fullMethod string, req, resp any, err error) (matched bool, newResp any, newErr error)
```
stephanos: 🤖 The pre-handler and post-handler invocations aren't reliably distinguishable by callbacks. The doc says `resp` is nil before the handler, and both `tests/testcore/fault_injection.go` and the example in `docs/development/testing.md` rely on `resp == nil` to mean "pre-handler". But when the handler returns an error, the post-handler call also has `resp == nil` (with `err != nil`), so a callback keyed on `resp == nil` will match a second time and re-inject on a request that already failed.

Consider passing the stage explicitly, e.g. add a `stage` argument (`stagePreHandler`/`stagePostHandler`) or split into two callback slots. That removes the ambiguity and makes the intent readable at the call site instead of inferring it from a nil response.

tests/testcore/fault_injection.go
```diff
@@ -0,0 +1,115 @@
...
+	var fired atomic.Bool
+
+	unregister := generator.RegisterCallback(func(ctx context.Context, fullMethod string, req, resp any, err error) (bool, any, error) {
+		if !options.matchesNamespace(req) {
+			return false, nil, nil
+		}
+
+		if injectedErr := fault(req, resp, err); injectedErr != nil {
+			fired.Store(true)
+			t.Logf("Fault injection fired: %T", req)
```
stephanos: 🤖 `t.Logf` here runs on a server goroutine, and `Generate` snapshots the callback slice before releasing the lock and invoking callbacks. So the `unregister()` in `t.Cleanup` does not fence an already in-flight invocation: a callback can call `t.Logf` after the test function has returned, which panics with `Log in goroutine after TestXxx has completed`.

The window is narrow but real for tests that leave workflows running (e.g. the update test in this PR keeps the workflow open, and history retries `AddWorkflowTask`). Options: drop the log line, or guard it with a flag flipped under a mutex in `t.Cleanup` so post-cleanup invocations are silent (the `fired.Store` itself is safe).

common/rpc/faultinjection/grpc.go
```diff
@@ -0,0 +1,172 @@
+package faultinjection
```
stephanos: This package provides the general fault injection mechanics; not knowing anything about Go tests.

common/rpc/faultinjection/testhooks_testdep.go
```diff
@@ -0,0 +1,52 @@
+//go:build test_dep
```
stephanos: This file is the bridge to the test world, ie providing an API to hook into the faults via testhooks.

tests/testcore/onebox.go
```diff
@@ -129,6 +131,7 @@ func newTemporal(t *testing.T, params *temporalParams) *temporalImpl {
 		workerConfig:              params.WorkerConfig,
 		replicationStreamRecorder: NewReplicationStreamRecorder(),
 	}
+	impl.rpcFaultGenerator = rpcfaultinjection.NewTestRPCFaultGenerator(impl.testHooks)
```
long-nt-tran, line 136: probably need a `//go:build !test_dep` at the top of this file

common/rpc/faultinjection/grpc.go
```diff
@@ -0,0 +1,204 @@
...
+
+type rpcFaultStage int
+
+const (
+	rpcFaultStageRequest rpcFaultStage = iota
+	rpcFaultStageResponse
+)
+
+// RPCRequestCallback is a callback function for pre-handler RPC fault injection.
+type RPCRequestCallback func(ctx context.Context, fullMethod string, req any) (matched bool, newResp any, newErr error)
```
long-nt-tran: should we have the `(matched bool, newResp any, newErr error)` be a struct instead? I think when I see untyped `(bool, any, error)` elsewhere in the PR I forget what each field stands for.

I.e., something like

```
type RPCFault struct {
    Resp any
    Err error
}

type RPCRequestCallback func(ctx context.Context, fullMethod string, req any) *RPCFault
```

then the `nil`-ness of `*RPCFault` can be equated to `matched == false`. I.e., the interceptor then looks something like:

```
func GRPCUnaryServerInterceptor(generator Generator) grpc.UnaryServerInterceptor {
	if generator == nil {
		return nil
	}
	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
		// Check before handler (can short-circuit)
		if fault := generator.GenerateRequest(ctx, info.FullMethod, req); fault != nil {
			return fault.resp, fault.err
		}

		// Call handler
		resp, err := handler(ctx, req)

		// Check after handler (can modify response/error)
		if fault := generator.GenerateResponse(ctx, info.FullMethod, req, resp, err); fault != nil {
			return fault.resp, fault.err
		}
		return resp, err
	}
}
```

WDYT?

common/rpc/faultinjection/grpc.go
```diff
@@ -0,0 +1,204 @@
...
+		return cb(ctx, fullMethod, req, resp, err)
+	})
+}
+
+func (r *RPCFaultGenerator) registerCallback(scope RPCFaultScope, stage rpcFaultStage, cb rpcCallback) func() {
+	if r == nil {
+		return func() {}
+	}
+
+	var scopes []rpcCallbackScope
```
long-nt-tran: nit: `faultScope` and `cbScopes` maybe, since they're not the same "scope"
```suggestion
func (r *RPCFaultGenerator) registerCallback(faultScope RPCFaultScope, stage rpcFaultStage, cb rpcCallback) func() {
	if r == nil {
		return func() {}
	}

	var cbScopes []rpcCallbackScope
```

common/rpc/faultinjection/grpc.go
```diff
@@ -0,0 +1,204 @@
...
+func (r *RPCFaultGenerator) GenerateRequest(ctx context.Context, fullMethod string, req any) (bool, any, error) {
+	return r.generateForRequest(ctx, fullMethod, rpcFaultStageRequest, req, nil, nil)
+}
+
+// GenerateResponse checks registered RPC callbacks after the handler runs.
+func (r *RPCFaultGenerator) GenerateResponse(ctx context.Context, fullMethod string, req, resp any, err error) (bool, any, error) {
+	return r.generateForRequest(ctx, fullMethod, rpcFaultStageResponse, req, resp, err)
+}
+
+func (r *RPCFaultGenerator) generateForRequest(ctx context.Context, fullMethod string, stage rpcFaultStage, req, resp any, err error) (bool, any, error) {
```
long-nt-tran: maybe just `generateFaultForStage` (or `generateFor`?) since it's used by both

tests/testcore/test_env.go
```diff
@@ -458,6 +459,38 @@ func (e *TestEnv) Tv() *testvars.TestVars {
...
+	scope := rpcfaultinjection.RPCFaultScope{
+		NamespaceID:   e.nsID,
+		NamespaceName: e.nsName,
+	}
+	return injectRPCFault(e.t, func(inject func(any, error) (bool, any, error)) func() {
+		return e.GetTestCluster().Host().GetRPCFaultGenerator().RegisterRequestCallback(scope, func(_ context.Context, _ string, req any) (bool, any, error) {
+			return inject(req, fault(req))
+		})
+	})
+}
```
long-nt-tran: WDYT about making the tracking more declarative? I.e.,

```
// fault_injection.go
type faultTracker struct {
    t       testing.TB
    fired   atomic.Bool
    logMu   sync.Mutex
    logging bool
}

func newFaultTracker(t testing.TB) *faultTracker {
    t.Helper()
    return &faultTracker{t: t, logging: true}
}

func (f *faultTracker) markFired(req any) {
    f.fired.Store(true)
    f.logMu.Lock()
    defer f.logMu.Unlock()
    if f.logging {
        f.t.Logf("Fault injection fired: %T", req)
    }
}

func (f *faultTracker) attach(unregister func()) {
    f.t.Cleanup(func() {
        f.logMu.Lock()
        f.logging = false
        f.logMu.Unlock()
        unregister()
        if !f.fired.Load() {
            f.t.Error("fault injection was registered but never fired - the fault was never injected")
        }
    })
}

// test_env.go
func (e *TestEnv) InjectRPCRequestFault(fault RPCRequestFault) func() {
    scope := rpcfaultinjection.RPCFaultScope{NamespaceID: e.nsID, NamespaceName: e.nsName}
    tracker := newFaultTracker(e.t)
    unregister := e.GetTestCluster().Host().GetRPCFaultGenerator().RegisterRequestCallback(scope,
        func(_ context.Context, _ string, req any) (bool, any, error) {
            if err := fault(req); err != nil {
                tracker.markFired(req)
                return true, nil, err
            }
            return false, nil, nil
        })
    return tracker.attach(unregister)
}
```
the closures getting encapsulated was slightly confusing to read for me

tests/update_workflow_test.go
```diff
@@ -1380,15 +1382,35 @@ func (s *WorkflowUpdateSuite) TestStickySpeculativeWorkflowTask_AcceptComplete_S
...
+	// Inject StickyWorkerUnavailable for AddWorkflowTask on the sticky queue.
+	// This causes history to fall back to the normal queue.
+	env.InjectRPCRequestFault(
+		func(req any) error {
+			r, ok := req.(*matchingservice.AddWorkflowTaskRequest)
+			if ok && r.GetTaskQueue().GetKind() == enumspb.TASK_QUEUE_KIND_STICKY {
+				return serviceerrors.NewStickyWorkerUnavailable()
+			}
+			return nil
+		})
```
long-nt-tran, line 1403: nice! It looks very clean in terms of prescribing behavior

stephanos: Agreed! I just also noticed that when we upgrade to 1.27, this might become even slicker as then we can probably put `matchingservice.AddWorkflowTaskRequest` on `InjectRPCRequestFault` as a generic type.

common/rpc/faultinjection/grpc.go
stephanos: Good catch!

common/rpc/faultinjection/grpc.go
stephanos: Agreed!

common/rpc/faultinjection/grpc.go
stephanos: Good idea! Calling it `RPCFaultOutcome`.

stephanos: NOTE: This required a bit of a refactor now to avoid an awkward re-mapping to/from testhooks as `RPCFaultOutcome` otherwise created a cyclic dependency. I've now created a sub-package for the testhooks adapter to break the cycle.

tests/testcore/test_env.go
stephanos: Good call!

tests/testcore/onebox.go
stephanos: Fixed!

long-nt-tran (approved):

yiminc (approved):

PR #9027, opened 2026-01-14
Memory watcher for functional tests

stephanos:
## What changed?

Added script to take memory report snapshots during test execution and print report after test success/failure.

## Why?

Gain insights into memory usage; especially when job is OOM killed.

## How did you test it?

see example: https://github.com/temporalio/temporal/actions/runs/21011468266/job/60407413329?pr=9027#step:9:15

<details><summary>Example</summary>
<p>

```
Memory snapshot at 2026-01-14 22:16:53

Time      Used(%)  Used(MB)  Goroutines  Top Processes
---------------------------------------------------------------
22:06:03    16%      5347         ?     java (2468MB), java (1322MB)
22:06:33    20%      6460         ?     java (2450MB), java (1322MB), compile (376MB)
22:07:03    16%      5418         ?     java (2430MB), java (1323MB)
22:07:33    16%      5401      2781     java (2531MB), java (1333MB)
22:08:06    17%      5702      5861     java (2493MB), java (1339MB), xdc.test (649MB)
22:08:36    18%      6034      5971     java (2528MB), java (1346MB), xdc.test (936MB)
22:09:07    20%      6455      6149     java (2539MB), java (1350MB), xdc.test (1314MB)
22:09:38    26%      8416     45668     xdc.test (3281MB), java (2535MB), java (1375MB)
22:10:10    33%     10596     43034     xdc.test (5657MB), java (2560MB), java (1405MB)
22:10:42    39%     12806     39209     xdc.test (7449MB), java (2551MB), java (1413MB)
22:11:13    42%     13580     33372     xdc.test (8166MB), java (2556MB), java (1413MB)
22:11:44    44%     14196     11067     xdc.test (8168MB), java (2548MB), java (1413MB)
22:12:15    44%     14171     11342     xdc.test (8653MB), java (2555MB), java (1415MB)
22:12:46    44%     14233     11724     xdc.test (8682MB), java (2546MB), java (1421MB)
22:13:17    45%     14536     10186     xdc.test (8858MB), java (2556MB), java (1427MB)
22:13:48    45%     14648     10392     xdc.test (8961MB), java (2555MB), java (1427MB)
22:14:19    45%     14710     10935     xdc.test (9027MB), java (2552MB), java (1428MB)
22:14:49    45%     14716     11060     xdc.test (9020MB), java (2557MB), java (1428MB)
22:15:20    45%     14702     11312     xdc.test (8925MB), java (2557MB), java (1428MB)
[...]

PR #9079, opened 2026-01-19
Capture memory monitor report before OOM crash

stephanos:
## What changed?

Follow-up to https://github.com/temporalio/temporal/pull/9027

1. Print memory stats to both file and stdout (new!)
2. Print Go heap report once a memory threshold is reached (ie close to OOM kill)
3. Add `alloc_objects` to Go heap report

See [example](https://github.com/temporalio/temporal/actions/runs/21153910359/job/60835302610?pr=9079#step:9:89):
Screenshot 2026-01-19 at 4 23 28 PM
(note that Go test output is buffered so they all appear mostly sequentially)

## Why?

Gain ability to capture memory usage and Go heap report **before an OOM kill**.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/develop/github/memory_monitor.sh
+++ b/develop/github/memory_monitor.sh
@@ -78,7 +78,7 @@ print_pprof_analysis() {
   fi
 }
 
-write_snapshot() {
+snapshot() {
   local memtotal_kb memavail_kb memused_kb memused_mb pct
   memtotal_kb="$(awk '/MemTotal/ {print $2}' /proc/meminfo)"
   memavail_kb="$(awk '/MemAvailable/ {print $2}' /proc/meminfo)"
@@ -114,6 +114,7 @@ write_snapshot() {
     HEAP_PRINTED=true
   fi
 
+  # Write snapshot to disk.
   {
     echo "Memory snapshot at $(date '+%Y-%m-%d %H:%M:%S')"
     echo ""
@@ -129,8 +130,8 @@ write_snapshot() {
   } > "$SNAPSHOT_FILE"
 }
 
-# Take snapshots every 30s until killed
+# Take snapshots every 30s until killed.
 while true; do
-  write_snapshot
+  snapshot
   sleep 30
 done


develop/github/memory_monitor.sh
```diff
@@ -17,9 +17,60 @@ fi
 
 SNAPSHOT_FILE="$1"
 HISTORY_FILE="/tmp/memory_history.txt"
+HEAP_PRINTED_FLAG="/tmp/heap_details_printed.flag"
```
dnr: just wondering, why is the snapshot file an argument but the others are fixed paths?

develop/github/memory_monitor.sh
```diff
@@ -17,9 +17,60 @@ fi
...
+  local count="?"
+  if curl -s --max-time 5 "http://${PPROF_HOST}/debug/pprof/goroutine?debug=1" -o /tmp/goroutine.out 2>/dev/null; then
+    count="$(head -1 /tmp/goroutine.out | grep -o '[0-9]*' || echo '?')"
+  fi
+  echo "$count"
+}
+
+# Print pprof heap analysis to stdout
+print_pprof_analysis() {
+  local heap_file="/tmp/heap.out"
```
dnr: should use mktemp and clean up before returning?

develop/github/memory_monitor.sh
```diff
@@ -17,9 +17,60 @@ fi
...
+
+  echo "--- Go Heap Profile ---"
+  if fetch_pprof "heap" "$heap_file"; then
+    echo "=== inuse_space (what's currently held) ==="
+    analyze_heap "$heap_file" "inuse_space" 60
```
dnr: tbh I think 30 lines for these is plenty but 🤷

develop/github/memory_monitor.sh
```diff
@@ -41,8 +90,22 @@ write_snapshot() {
...
   timestamp="$(date '+%H:%M:%S')"
 
-  # Append to history
-  echo "$timestamp $pct $memused_mb $goroutines $top_procs" >> "$HISTORY_FILE"
+  # Append to history.
+  local row="$timestamp $pct $memused_mb $goroutines $top_procs"
+  echo "$row" >> "$HISTORY_FILE"
+
+  # Print to stdout (with descriptors) to preserve the information in case of a crash.
+  echo "$timestamp used=${pct}% mem=${memused_mb}MB goroutines=${goroutines} procs=[${top_procs}]"
```
dnr: so we write to stdout and also $HISTORY_FILE with slightly different formats? where does the stdout end up? and where does the file get displayed?

can we consolidate to one format and use `tee -a`? just to simplify the mental model of what is going on here.. I'm finding the multiple outputs and multiple conditions hard to follow

develop/github/memory_monitor.sh
```diff
@@ -41,8 +90,22 @@ write_snapshot() {
...
+
+  # If memory threshold was reached, print Go heap details. But only once per run.
+  if [[ "$pct" -ge "$HIGH_MEMORY_THRESHOLD" ]] && [[ ! -f "$HEAP_PRINTED_FLAG" ]]; then
```
dnr: why is the flag a file? can't it be a shell variable?

I don't get this, it says we're printing go heap details once per run, but we also call print_pprof_analysis every time below?

develop/github/memory_monitor.sh
dnr: hmm, maybe just write to history_file in the desired format to begin with?

develop/github/memory_monitor.sh
stephanos: You're absolutely right, shell variable is simpler 👍

develop/github/memory_monitor.sh
stephanos: Because the snapshot file is used by the caller (ie the CI job to print it), the other ones are only internal "helper" files.

develop/github/memory_monitor.sh
stephanos: Right; the trick is that the full final report that is printed is a table; but the one-per-iteration echo is standalone (ie it needs labels IMO). But I see your point; I'll remove the table and re-use the same format for both as that's definitely easier 👍

develop/github/memory_monitor.sh
stephanos: The "once per run" was referring to _printing_ the memory snapshot. It it still captured and written to the file on each iteration. I changed it now to re-use the previously printed pprof details in that case to avoid a 2nd call.

PR #9081, opened 2026-01-20
[Nexus-HSM-CHASM] Migrate operation statemachine.

gow:
## What changed?
 - Added `OperationState` proto fields
 - Migrated nexus operation state transitions.

## Why?
Migrating nexus from HSM to CHASM

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
N/A. This code path is currently unreachable.

---

> [!NOTE]
> **Medium Risk**
> Touches persisted proto schemas and core state-transition logic for retries/timeouts; while executors are still unimplemented, any activation of this path could impact task scheduling and retry semantics.
> 
> **Overview**
> Migrates Nexus operation lifecycle handling to CHASM by implementing the operation state machine transitions to emit invocation, backoff-retry, and timeout tasks and to record attempt metadata (last failure/completion time, next retry time, operation token).
> 
> Expands the `OperationState` and task protos to persist endpoint/operation identifiers, scheduling timestamps, retry/attempt fields, and separate timeout task types (`schedule-to-start`, `start-to-close`, `schedule-to-close`), and wires new timeout task executors through Fx and the library task registry. Adds unit tests covering the new transition behavior and task scheduling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4ab5fd08968579143290d9c479919124cc6e33d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/nexusoperations/workflow/commands.go
+++ b/components/nexusoperations/workflow/commands.go
@@ -102,7 +102,7 @@ func (ch *commandHandler) HandleScheduleCommand(
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.ScheduleToStartTimeout); err != nil {
-		return workflow.FailWorkflowTaskError{
+		return chasmcommand.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.ScheduleToStartTimeout is invalid: %v", err),
@@ -110,7 +110,7 @@ func (ch *commandHandler) HandleScheduleCommand(
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.StartToCloseTimeout); err != nil {
-		return workflow.FailWorkflowTaskError{
+		return chasmcommand.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.StartToCloseTimeout is invalid: %v", err),
--- a/components/nexusoperations/workflow/commands_test.go
+++ b/components/nexusoperations/workflow/commands_test.go
@@ -401,7 +401,7 @@ func TestHandleScheduleCommand(t *testing.T) {
 				},
 			},
 		})
-		var failWFTErr workflow.FailWorkflowTaskError
+		var failWFTErr chasmcommand.FailWorkflowTaskError
 		require.ErrorAs(t, err, &failWFTErr)
 		require.False(t, failWFTErr.TerminateWorkflow)
 		require.Equal(t, enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES, failWFTErr.Cause)
@@ -420,7 +420,7 @@ func TestHandleScheduleCommand(t *testing.T) {
 				},
 			},
 		})
-		var failWFTErr workflow.FailWorkflowTaskError
+		var failWFTErr chasmcommand.FailWorkflowTaskError
 		require.ErrorAs(t, err, &failWFTErr)
 		require.False(t, failWFTErr.TerminateWorkflow)
 		require.Equal(t, enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES, failWFTErr.Cause)


chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -14,19 +18,56 @@ var transitionScheduled = chasm.NewTransition(
...
+		// Emit an invocation task to start the operation
+		ctx.AddTask(o, chasm.TaskAttributes{}, &nexusoperationpb.InvocationTask{
+			Attempt: o.Attempt,
+		})
+
+		// Emit a schedule-to-close timeout task if configured
+		if o.ScheduleToCloseTimeout != nil && o.ScheduleToCloseTimeout.AsDuration() != 0 {
+			deadline := o.ScheduledTime.AsTime().Add(o.ScheduleToCloseTimeout.AsDuration())
+			ctx.AddTask(o, chasm.TaskAttributes{
+				ScheduledTime: deadline,
```
pdoerner, line 38: I don't think we should be storing the deadline as the `ScheduledTime`? But maybe I'm misunderstanding how CHASM uses the field.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -14,19 +18,56 @@ var transitionScheduled = chasm.NewTransition(
...
+			}, &nexusoperationpb.InvocationTimeoutTask{
```
bergundy: This is fine but let's call it ScheduleToCloseTimeoutTask since we are adding StartToClose and ScheduleToStart soon.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -14,19 +18,56 @@ var transitionScheduled = chasm.NewTransition(
...
+			})
+		}
+
+		return nil
 	},
 )
 
 // EventAttemptFailed is triggered when an invocation attempt is failed with a retryable error.
 type EventAttemptFailed struct {
+	Time        time.Time
```
bergundy: Do not accept this as an argument, use `ctx.Now()` instead. This will allow us to add time skipping and specific pause capabilities into CHASM directly later.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -39,13 +80,20 @@ var transitionRescheduled = chasm.NewTransition(
 	[]nexusoperationpb.OperationStatus{nexusoperationpb.OPERATION_STATUS_BACKING_OFF},
 	nexusoperationpb.OPERATION_STATUS_SCHEDULED,
 	func(o *Operation, ctx chasm.MutableContext, event EventRescheduled) error {
```
bergundy, line 91: You forgot to reset NextAttemptScheduleTime

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -39,13 +80,20 @@ var transitionRescheduled = chasm.NewTransition(
...
+		})
+
+		return nil
 	},
 )
 
 // EventStarted is triggered when an invocation attempt succeeds and the handler indicates that it started an
 // asynchronous operation.
 type EventStarted struct {
+	Time           time.Time
```
bergundy: Same here, don't inject the time, use `ctx.Now()`.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -55,7 +103,15 @@ var transitionStarted = chasm.NewTransition(
...
-		return serviceerror.NewUnimplemented("unimplemented")
+		// Record the attempt as successful
+		o.Attempt++
+		o.LastAttemptCompleteTime = timestamppb.New(event.Time)
+		o.LastAttemptFailure = nil
+
+		// Store the operation token for async completion
+		o.OperationToken = event.OperationToken
+
+		return nil
```
bergundy, line 148: You will need to kick off cancelation here, add a TODO.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -71,7 +127,9 @@ var transitionSucceeded = chasm.NewTransition(
 	},
 	nexusoperationpb.OPERATION_STATUS_SUCCEEDED,
 	func(o *Operation, ctx chasm.MutableContext, event EventSucceeded) error {
-		return serviceerror.NewUnimplemented("unimplemented")
+		// Terminal state - no tasks to emit
+		// The component will be deleted after this transition
```
bergundy, line 165: This is fine to assume for now, just keep in mind that this will change for standalone nexus ops.

bergundy (commented):
Looks great overall! Nice to see how CHASM makes the code much easier to understand with the removal of the requirement to regenerate tasks and manual conflict resolution logic.

chasm/lib/nexusoperation/operation_statemachine.go
gow: Did you mean to kickoff cancellation here or in start-to-close task handler? Why would we start a cancellation here?
I added a TODO. But I'm not sure if it's correct.

chasm/lib/nexusoperation/operation_statemachine.go
gow: Will update it when we get to standalone nexus ops.

chasm/lib/nexusoperation/operation_statemachine.go
gow: Fixed

chasm/lib/nexusoperation/operation_statemachine.go
gow: Done

chasm/lib/nexusoperation/operation_statemachine_test.go
```diff
@@ -0,0 +1,449 @@
...
+			require.Len(t, ctx.Tasks, len(tc.expectedTasks))
+			for i, expectedTask := range tc.expectedTasks {
+				actualTask := ctx.Tasks[i]
+
+				require.IsType(t, expectedTask.Payload, actualTask.Payload, "expected %T at index %d, got %T",
+					expectedTask.Payload, i, actualTask.Payload)
+
+				switch expectedTask.Payload.(type) {
+				case *nexusoperationpb.InvocationTask:
+					invTask, ok := actualTask.Payload.(*nexusoperationpb.InvocationTask)
```
stephanos: 🤖 `small` **[testing]**: The scheduled-transition test verifies task *types* and *count* but doesn't verify that `o.Attempt` value on the emitted `InvocationTask` matches the operation's attempt. The assertion `require.Equal(t, int32(0), invTask.Attempt)` is hardcoded rather than derived from the operation's starting attempt. If someone changes `newTestOperation()` to start with a non-zero attempt, this test would silently pass with the wrong value. Use `operation.Attempt` instead of a literal `0`.

chasm/lib/nexusoperation/operation_statemachine_test.go
```diff
@@ -0,0 +1,449 @@
...
+					require.Equal(t, int32(0), invTask.Attempt)
+					require.Empty(t, actualTask.Attributes.ScheduledTime)
+				case *nexusoperationpb.ScheduleToCloseTimeoutTask:
+					timeoutTask, ok := actualTask.Payload.(*nexusoperationpb.ScheduleToCloseTimeoutTask)
+					require.True(t, ok, "expected ScheduleToCloseTimeoutTask at index %d", i)
+					require.Equal(t, int32(0), timeoutTask.Attempt)
+					require.Equal(t, defaultTime.Add(tc.scheduleToCloseTimeout), actualTask.Attributes.ScheduledTime)
+				default:
+					t.Fatalf("unexpected task payload type at index %d: %T", i, actualTask.Payload)
+				}
```
stephanos: 🤖 `nit` **[testing]**: `require.Empty(t, actualTask.Attributes.ScheduledTime)` — `ScheduledTime` is `time.Time`, and `require.Empty` checks for the zero value via reflection. `require.True(t, actualTask.Attributes.ScheduledTime.IsZero())` would be more explicit about intent.

chasm/lib/nexusoperation/operation_statemachine_test.go
```diff
@@ -0,0 +1,449 @@
...
+		startingAttemptCount int32
+		expectedAttempt      int32
+		minRetryInterval     time.Duration
+		maxRetryInterval     time.Duration
+		retryPolicy          backoff.RetryPolicy
+	}{
+		{
+			name:                 "first retry",
+			startingAttemptCount: 0,
+			expectedAttempt:      1,
```
stephanos: 🤖 `med` **[testing]**: The retry interval bounds in `TestTransitionAttemptFailed` are fragile because they depend on the jitter implementation details of `backoff.ExponentialRetryPolicy`. If the jitter algorithm changes (e.g., from full jitter to equal jitter, or the coefficient changes), these tests will break even though the state machine logic is correct. Consider either:
1. Using a mock/deterministic retry policy in tests to get predictable delays, or
2. Widening the bounds significantly (e.g., 0 to 10x base interval) so they only catch gross errors.

chasm/lib/nexusoperation/operation_statemachine_test.go
```diff
@@ -0,0 +1,449 @@
...
+func TestTransitionStarted(t *testing.T) {
+	testCases := []struct {
+		name            string
+		startStatus     nexusoperationpb.OperationStatus
+		startingAttempt int32
+		expectedAttempt int32
+		operationToken  string
+	}{
+		{
+			name:            "started from scheduled",
```
stephanos: 🤖 `small` **[testing]**: `TestTransitionStarted` "started from backing off" sets `startingAttempt: 3` and expects `expectedAttempt: 3` (no increment). But it doesn't set `LastAttemptFailure` on the operation before the transition, meaning the test doesn't verify that the transition correctly clears `LastAttemptFailure` when coming from `BACKING_OFF` with a prior failure. Consider setting `LastAttemptFailure` in the setup and verifying it's nil after the transition.

PR #9083, opened 2026-01-20
Implement CHASM UpdateWithStartExecution

awln-temporal:
## What changed?
Implement CHASM UpdateWithStartExecution.

## Why?
Allows callers to update any current execution and also start a new execution if none already exists for a given ExecutionKey.

## How did you test it?
- [X] built
- [X] added new unit test(s)

minor: fixes unit test verification for setting deferred pointers.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -32,9 +32,72 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       job_matrix: ${{ steps.build_matrix.outputs.job_matrix }}
+      full_test_reason: ${{ steps.determine_scope.outputs.full_test_reason }}
       runner_x64: ${{ steps.configure_runners.outputs.runner_x64 }}
       runner_arm: ${{ steps.configure_runners.outputs.runner_arm }}
     steps:
+      - name: Checkout Code
+        uses: actions/checkout@v6
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          ref: ${{ env.COMMIT }}
+          fetch-depth: 0
+
+      - name: Fetch base branch
+        if: ${{ github.event_name == 'pull_request' }}
+        run: git fetch origin ${{ github.event.pull_request.base.ref }}:${{ github.event.pull_request.base.ref }}
+
+      - name: Compute merge base
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          MERGE_BASE="$(git merge-base "${{ env.COMMIT }}" "${{ github.event.pull_request.base.ref }}")"
+          echo "MERGE_BASE=${MERGE_BASE}" >> "$GITHUB_ENV"
+
+      # If possible, run the "abridged" jobs on PRs to save time.
+      #
+      # The goal is to still cover all persistence code paths:
+      #   1. SQL
+      #   2. NoSQL (ie Cassandra)
+      #   3. Elasticsearch/OpenSearch
+      # All other jobs for the remaining databases only run a small set of smoke tests.
+      #
+      # Exceptions to this optimization:
+      #   1. commit is not a PR
+      #   2. `test-all-dbs` label set on PR
+      #   3. changes to persistence packages were made
+      - name: Determine test scope
+        id: determine_scope
+        run: |
+          FULL_TEST_REASON=""
+
+          event_name="${{ github.event_name }}"
+          labels='${{ toJson(github.event.pull_request.labels.*.name) }}'
+          merge_base="${MERGE_BASE:-}"
+
+          echo "event_name=$event_name"
+          echo "labels=$labels"
+          echo "merge_base=$merge_base"
+
+          # Push events (main, release branches) run all tests on all DBs
+          if [[ "$event_name" == "push" ]]; then
+            FULL_TEST_REASON="Running full tests on all DBs (push event)."
+          # Check for test-all-dbs label
+          elif echo "$labels" | jq -e 'any(. == "test-all-dbs")' > /dev/null 2>&1; then
+            FULL_TEST_REASON="Running full tests on all DBs (test-all-dbs label)."
+          # Check for persistence code changes
+          elif [[ -n "$merge_base" ]]; then
+            echo "Changed persistence files:"
+            git diff --name-only "$merge_base" "$COMMIT" | grep -E "^(common/persistence/|schema/)" || echo "(none)"
+            if git diff --name-only "$merge_base" "$COMMIT" | grep -qE "^(common/persistence/|schema/)"; then
+              FULL_TEST_REASON="Running full tests on all DBs (persistence code changes)."
+            fi
+          fi
+
+          echo "full_test_reason=$FULL_TEST_REASON" >> "$GITHUB_OUTPUT"
+
+      - name: ${{ steps.determine_scope.outputs.full_test_reason && 'ℹ️ Full tests' || 'ℹ️ Smoke tests' }}
+        run: echo "::notice::${{ steps.determine_scope.outputs.full_test_reason || 'Running smoke tests on extended DBs. Add the test-all-dbs label to run all tests on all DBs.' }}"
+
       - name: Configure runners
         id: configure_runners
         run: |
@@ -52,14 +115,18 @@ jobs:
             echo "runner_arm=$runner_arm"
           } >> "$GITHUB_OUTPUT"
 
+      # Primary DBs always get full tests, extended DBs get smoke tests (1 job) in abridged PRs.
       - name: Build job matrix
         id: build_matrix
         env:
+          FULL_TEST_REASON: ${{ steps.determine_scope.outputs.full_test_reason }}
           DB_CONFIGS: |
             cass_es:
               persistence_type: nosql
               persistence_driver: cassandra
               containers: [cassandra, elasticsearch]
+              arch: arm
+              required: true
             cass_es8:
               persistence_type: nosql
               persistence_driver: cassandra
@@ -85,6 +152,8 @@ jobs:
               persistence_type: sql
               persistence_driver: postgres12
               containers: [postgresql]
+              arch: arm
+              required: true
             postgres12_pgx:
               persistence_type: sql
               persistence_driver: postgres12_pgx
@@ -95,6 +164,11 @@ jobs:
               test_timeout: 35m
               github_timeout: 40
               sharded: true
+            smoke:
+              cmd: make functional-test-coverage
+              test_timeout: 5m
+              github_timeout: 10
+              test_args: '"-run=TestActivityTestSuite|TestSignalWorkflowTestSuite|TestWorkflowTestSuite"'
             ndc:
               cmd: make functional-test-ndc-coverage
               test_timeout: 10m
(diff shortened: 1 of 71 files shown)


service/history/chasm_engine.go
```diff
@@ -180,8 +180,172 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+		archetypeID,
+		startFn,
+		updateFn,
+		options,
+	)
+	if err != nil {
+		return chasm.ExecutionKey{}, nil, err
+	}
+
+	currentRunInfo, hasCurrentRun, err := e.persistAsBrandNew(
```
yycptt, line 396: if there's an existing execution we can resolve conflict before attempting the creation right?

yycptt: also I don't think persistAsBrandNew support ID reuse case.

service/history/chasm_engine.go
```diff
@@ -488,6 +652,87 @@ func (e *ChasmEngine) createNewExecution(
...
+	chasmContext := chasm.NewMutableContext(ctx, chasmTree)
+
+	// Create the root component using newFn
+	rootComponent, err := newFn(chasmContext)
+	if err != nil {
+		return newExecutionParams{}, err
+	}
+	chasmTree.SetRootComponent(rootComponent)
+
+	if err := updateFn(chasmContext, rootComponent); err != nil {
```
yycptt: hmmm we probably need to better define/implement a "state transition" in chasm tree impl. 

I think for now before calling updateFn(), we need to at least to do a syncStructure on the tree after `newFn()`, we can do it as part of SetRootComponent().

service/history/chasm_engine.go
```diff
@@ -180,8 +180,172 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+
+	lockPriority := locks.PriorityHigh
+	callerType := headers.GetCallerInfo(ctx).CallerType
+	if callerType == headers.CallerTypeBackgroundHigh ||
+		callerType == headers.CallerTypeBackgroundLow ||
+		callerType == headers.CallerTypePreemptable {
+		lockPriority = locks.PriorityLow
+	}
+
+	executionLease, err := consistencyChecker.GetChasmLease(
```
yycptt: it should be possible to reuse the existing `getExecutionLease()` method?

service/history/chasm_engine.go
```diff
@@ -180,8 +180,172 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+		executionLease.GetReleaseFn()(nil)
+		return e.startAndUpdateExecution(ctx, shardContext, executionRef, archetypeID, startFn, updateFn, options)
+	case *serviceerror.NotFound:
+		return e.startAndUpdateExecution(ctx, shardContext, executionRef, archetypeID, startFn, updateFn, options)
+	default:
+		return chasm.ExecutionKey{}, nil, err
+	}
+}
+
+func (e *ChasmEngine) updateExecution(
```
yycptt, line 264: let's see check UpdateComponent and see if we can extract some common logic

service/history/chasm_engine.go
awln-temporal: yeah good point, we should be checking before since we have access to the mutable state.

awln-temporal: actually, is it possible to still run into a conflicting execution before persisting? If so, do we need to do this conflict check twice, before and after?

yycptt: yes it theoretically can run into conflict both before and after. But as long as we do the check before persisting, the chance of getting another conflict after persisting is very low (only during shard movement case I think, because the latest run get locked) and we can just return an error and let the history handler & history client retry the request. I think that's also what signalWithStart is doing today.

service/history/chasm_engine.go
awln-temporal: I see, it may require access to child nodes. I'll add the change in `SetRootComponent()` to sync sub fields always, we can discuss a more formal state transition as a follow up.

service/history/chasm_engine.go
awln-temporal: done

service/history/chasm_engine.go
awln-temporal: yeah fair enough, was wondering if this issue due to "multiple" owners of shard is worth changing between SignalWithStart and UpdateWithStart. Kept the implementation to check before persisting.

service/history/chasm_engine.go
```diff
@@ -180,29 +180,113 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+		return chasm.ExecutionKey{}, nil, err
+	}
+
+	archetypeID, err := executionRef.ArchetypeID(e.registry)
+	if err != nil {
+		return chasm.ExecutionKey{}, nil, err
+	}
+
+	refWithEmptyRunID := executionRef
+	refWithEmptyRunID.RunID = ""
```
yycptt: hmm let's return an Unimplemented error if RunID is specified and let's do the same for StartExecution as well. (and we no longer need to copy)

There are some uses case where supporting a fixed runID will make things a lot easier (e.g. for deduping child workflow creation), but require more changes in persistence layer.

service/history/chasm_engine.go
```diff
@@ -180,29 +180,113 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+	shardContext historyi.ShardContext,
+	executionLease api.WorkflowLease,
+	executionRef chasm.ComponentRef,
 	updateFn func(chasm.MutableContext, chasm.Component) error,
-	opts ...chasm.TransitionOption,
-) (updatedRef []byte, retError error) {
+) (chasm.ExecutionKey, []byte, error) {
+	workflowKey := executionLease.GetContext().GetWorkflowKey()
+	actualRef := executionRef
+	actualRef.RunID = workflowKey.RunID
```
yycptt, line 273: runID shouldn't matter when using the ref inside a chasm tree. or maybe I missed the purpose of this?

service/history/chasm_engine.go
```diff
@@ -456,7 +626,91 @@ func (e *ChasmEngine) createNewExecution(
...
+			shardContext.GetThrottledLogger(),
+			shardContext.GetMetricsHandler(),
+		),
+		mutableState: mutableState,
+		snapshot:     snapshot,
+		events:       events,
+	}, nil
+}
+
+func (e *ChasmEngine) createNewExecutionWithUpdate(
```
yycptt, line 627: looks like the only different with `createNewExecution` is this one has a `updateFn`? Can `createNewExecution` reuse this one with a nil `updateFn`?

service/history/chasm_engine.go
```diff
@@ -232,12 +316,98 @@ func (e *ChasmEngine) UpdateComponent(
...
+	}
+	if !hasCurrentRun {
+		serializedRef, err := newExecutionParams.executionRef.Serialize(e.registry)
+		if err != nil {
+			return newExecutionParams.executionRef.ExecutionKey, nil, err
+		}
+		return newExecutionParams.executionRef.ExecutionKey, serializedRef, nil
+	}
+
+	result, err := e.handleExecutionConflict(
```
yycptt: hmm I think we have to retry the entire operation from beginning in this case apply updateFn to the existing run if possible (i.e. when it's running). 

If conflict is handled here then we, for example, will just return the current run when conflict policy is UseExisting without applying the updateFn.

we can just return the CurrentWorkflowConditionFailedError in currentRunInfo I think.

service/history/chasm_engine.go
```diff
@@ -180,29 +180,113 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+	refWithEmptyRunID.RunID = ""
+
+	_, executionLease, err := e.getExecutionLease(ctx, refWithEmptyRunID)
+	switch err.(type) {
+	case nil:
+		defer func() {
+			executionLease.GetReleaseFn()(retError)
+		}()
+
+		// TODO: Handle BusinessIDConflictPolicyTerminateExisting
```
yycptt: return Unimplemented error

service/history/chasm_engine.go
```diff
@@ -180,29 +180,113 @@ func (e *ChasmEngine) UpdateWithStartExecution(
 	startFn func(chasm.MutableContext) (chasm.Component, error),
 	updateFn func(chasm.MutableContext, chasm.Component) error,
 	opts ...chasm.TransitionOption,
-) (newExecutionKey chasm.ExecutionKey, newExecutionRef []byte, retError error) {
-	return chasm.ExecutionKey{}, nil, serviceerror.NewUnimplemented("UpdateWithStartExecution is not yet supported")
+) (resultKey chasm.ExecutionKey, resultRef []byte, retError error) {
```
yycptt: oh I forgot about this, can you take a look at the return value we defined for StartExecution() and do the same thing for UpdateWithStart as well? Basic idea is that we need to expose to the caller if a new run is actually created or not.

awln-temporal: Yeah will do

service/history/chasm_engine.go
awln-temporal: I shouldn't have been calling handleExecutionConflict to begin with, returning `CurrentWorkflowConditionFailedError` now.

service/history/chasm_engine.go
```diff
@@ -180,29 +180,144 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+	if err != nil {
+		return chasm.EngineUpdateWithStartExecutionResult{}, err
+	}
+
+	archetypeID, err := executionRef.ArchetypeID(e.registry)
+	if err != nil {
+		return chasm.EngineUpdateWithStartExecutionResult{}, err
+	}
+
+	refWithEmptyRunID := executionRef
```
yycptt: now we no longer need refWithEmptyRunID right?

chasm/engine.go
```diff
@@ -102,10 +101,35 @@ type StartExecutionResult[O any] struct {
...
+//     [BusinessIDConflictPolicy] configured via [WithBusinessIDPolicy]), and the
+//     existing execution was returned instead.
+//   - Output1: The output value returned by the first factory function.
+//   - Output2: The output value returned by the second factory function.
+type UpdateWithStartExecutionResult[O1 any, O2 any] struct {
+	ExecutionKey ExecutionKey
+	ExecutionRef []byte
+	Created      bool
+	Output1      O1
+	Output2      O2
```
yycptt: let's give them a more descriptive name. StartOutput and UpdateOutput?

service/history/chasm_engine.go
yycptt: Can you help make the same change in StartExecution as well? Thanks!

service/history/chasm_engine.go
```diff
@@ -180,29 +180,144 @@ func (e *ChasmEngine) UpdateWithStartExecution(
...
+
+		if executionLease.GetMutableState().IsWorkflowExecutionRunning() {
+			executionKey, executionRef, err := e.updateExecution(ctx, shardContext, executionLease, executionRef, updateFn)
+			if err != nil {
+				return chasm.EngineUpdateWithStartExecutionResult{}, err
+			}
+			return chasm.EngineUpdateWithStartExecutionResult{
+				ExecutionKey: executionKey,
+				ExecutionRef: executionRef,
+				Created:      true,
```
yycptt: hmm this is false right? we are not creating a new run.

service/history/chasm_engine.go
awln-temporal: yep removed this

service/history/chasm_engine.go
awln-temporal: done, sorry about that

PR #9085, opened 2026-01-20
Fix migration workflow breaking change for OSS

yycptt:
## What changed?
- Use plain go struct inside migration workflow activity input output
- Customize the json encoding/decoding for migration execution info and make it backward compatible.
- This will temporary break cloud and the fix is in https://github.com/temporalio/temporal/pull/9097
- This PR needs to be part of oss v1.30 release.

## Why?
- Backward compatibility and less confusion over how the encoding will be done.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -2988,11 +2988,6 @@ WorkerActivitiesPerSecond, MaxConcurrentActivityTaskPollers.
 		false,
 		`WorkerEnableHistoryRateLimiter decides whether to generate migration tasks with history length rate limiter.`,
 	)
-	WorkerEnableMigrationExecutionInfoNewJSONEncoding = NewGlobalBoolSetting(
-		"worker.enableMigrationExecutionInfoNewJSONEncoding",
-		false,
-		`If set to true the execution info in the migration workflow will be encoded using "business_id" instead of "workflow_id"`,
-	)
 	MaxUserMetadataSummarySize = NewNamespaceIntSetting(
 		"limit.userMetadataSummarySize",
 		400,
--- a/service/worker/migration/execution_info.go
+++ b/service/worker/migration/execution_info.go
@@ -2,25 +2,7 @@ package migration
 
 import (
 	"encoding/json"
-
-	"go.temporal.io/server/common/dynamicconfig"
-	"go.uber.org/fx"
 )
-
-var ExecutionInfoEncodingModule = fx.Options(
-	fx.Provide(func(
-		dynamicCollection *dynamicconfig.Collection,
-	) enableExecutionInfoNewJSONEncoding {
-		return enableExecutionInfoNewJSONEncoding(dynamicconfig.WorkerEnableMigrationExecutionInfoNewJSONEncoding.Get(dynamicCollection)())
-	}),
-	fx.Populate(&executionInfoNewJSONEncodingEnabled),
-)
-
-type enableExecutionInfoNewJSONEncoding bool
-
-// executionInfoNewJSONEncodingEnabled can be removed in v1.31 when both released OSS and cloud version
-// support decoding "businessID".
-var executionInfoNewJSONEncodingEnabled enableExecutionInfoNewJSONEncoding
 
 type ExecutionInfo struct {
 	executionInfoNewJSON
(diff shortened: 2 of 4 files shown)


common/dynamicconfig/constants.go
```diff
@@ -2993,6 +2988,11 @@ WorkerActivitiesPerSecond, MaxConcurrentActivityTaskPollers.
 		false,
 		`WorkerEnableHistoryRateLimiter decides whether to generate migration tasks with history length rate limiter.`,
 	)
+	WorkerEnableMigrationExecutionInfoNewJSONEncoding = NewGlobalBoolSetting(
+		"worker.enableMigrationExecutionInfoNewJSONEncoding",
+		false,
```
yycptt: will set this to true in cloud.

We have to at least control how the encoding is done to support downgrading in both oss and cloud. Since the current released oss only recognizes "workflow_id" and cloud only recognizes "business_id".

service/worker/migration/execution_info.go
```diff
@@ -0,0 +1,74 @@
...
+		ArchetypeID: e.ArchetypeID,
+	})
+}
+
+func (e *ExecutionInfo) UnmarshalJSON(data []byte) error {
+	if ExecutionInfoNewJSONEncodingEnabled {
+		return json.Unmarshal(data, &e.executionInfoNewJSON)
+	}
+
+	// For forward compatibility, support both workflow_id and business_id here
```
yycptt: This gives us forward compatibility in oss and in next oss release (1.31), we can always encoded to business_id.

service/worker/migration/execution_info.go
```diff
@@ -0,0 +1,74 @@
...
+var ExecutionInfoEncodingModule = fx.Options(
+	fx.Provide(func(
+		dynamicCollection *dynamicconfig.Collection,
+	) enableExecutionInfoNewJSONEncoding {
+		return enableExecutionInfoNewJSONEncoding(dynamicconfig.WorkerEnableMigrationExecutionInfoNewJSONEncoding.Get(dynamicCollection)())
+	}),
+	fx.Populate(&executionInfoNewJSONEncodingEnabled),
+)
+
+type enableExecutionInfoNewJSONEncoding bool
```
yux0: is this intentional to make this config value static?

yycptt: yes, the config should not be changed.

service/worker/migration/execution_info.go
```diff
@@ -0,0 +1,77 @@
...
+	"go.uber.org/fx"
+)
+
+var ExecutionInfoEncodingModule = fx.Options(
+	fx.Provide(func(
+		dynamicCollection *dynamicconfig.Collection,
+	) enableExecutionInfoNewJSONEncoding {
+		return enableExecutionInfoNewJSONEncoding(dynamicconfig.WorkerEnableMigrationExecutionInfoNewJSONEncoding.Get(dynamicCollection)())
+	}),
+	fx.Populate(&executionInfoNewJSONEncodingEnabled),
```
dnr: doesn't this trigger the race detector in functional tests?

yycptt: hmm why will it trigger a race? It will be populated before any fx.Invoke so the service is not started yet. 

For xdc/ndc tests, the clusters are not created concurrently so it shouldn't be a race either.

yycptt: hmm ok I think I see the issue. Invoke needs to depend on the populate or the order is not unspecified I think.

yycptt: hmm let me do this. I am going to remove this config (always false) and populate thing. 
Pick this PR 1.30. This will fix the oss breaking issue.

Then have another PR before next cloud release, to make this work for cloud as well (basically make this config true)

yycptt: Removed.

yycptt:
The change for cloud is in https://github.com/temporalio/temporal/pull/9097

service/worker/migration/execution_info.go
dnr: I think Populate does happen before Invoke, but I meant the Populate from one cluster racing with the reads from another cluster

PR #9086, opened 2026-01-20
Implement detached component

awln-temporal:
## What changed?
Implement detached component as a Field option and Registrable Component option.

Add detached boolean value to ComponentAttributes persistence proto definition.

## Why?
Allow detached components to continue updates and task execution even if parent node lifecycle is closed.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -414,7 +414,7 @@ func (n *Node) Component(
 			fmt.Errorf("%s", reflect.TypeOf(node.value).String()))
 	}
 
-	if err := node.validateParentAccess(validationContext); err != nil {
+	if err := node.validateAccess(validationContext); err != nil {
 		return nil, err
 	}
 
@@ -435,9 +435,9 @@ func (n *Node) Component(
 //
 // When the context's intent is OperationIntentProgress, This check validates that
 // all of a node's ancestors are still in a running state, and can accept writes. In
-// the case of a newly-created node, a detached node, or an OperationIntentObserve
+// the case of a newly created node, a detached node, or an OperationIntentObserve
 // intent, the check is skipped.
-func (n *Node) validateParentAccess(ctx Context) error {
+func (n *Node) validateAccess(ctx Context) error {
 	intent := operationIntentFromContext(ctx.getContext())
 	if intent != OperationIntentProgress {
 		// Read-only operations are always allowed.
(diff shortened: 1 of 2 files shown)


chasm/registrable_component.go
```diff
@@ -54,6 +55,20 @@ func WithSingleCluster() RegistrableComponentOption {
 	}
 }
 
+// WithDetached marks the component type as detached. Detached components ignore
+// parent lifecycle validation, allowing them to continue operating when their
+// parent is closed/terminated.
+func WithDetached() RegistrableComponentOption {
```
yycptt, line 63: is this for marking a component to be detached like by default? and the ComponentFieldOption will take precedence over this if specified?

chasm/tree.go
```diff
@@ -112,6 +112,11 @@ type (
 		// executionState and executionStatus and trigger retention timers.
 		// We could consider extending the force terminate concept to sub-components as well.
 		terminated bool
+
+		// When detached is true, this component ignores parent lifecycle validation.
+		// Detached components can continue operating, accepting writes and executing
+		// tasks, even when their parent is closed/terminated.
+		detached bool
```
yycptt: hmm is it possible to always use the field in componentAttr? Unless there's a strong reason, I think we should avoid having two copies of the same thing.

chasm/tree.go
```diff
@@ -414,11 +419,9 @@ func (n *Node) Component(
 			fmt.Errorf("%s", reflect.TypeOf(node.value).String()))
 	}
 
-	// Access check always begins on the target node's parent, and ignored for nodes
-	// without ancestors.
-	if node.parent != nil {
-		err := node.parent.validateAccess(validationContext)
-		if err != nil {
+	// Access check on ancestors only. Detached nodes skip ancestor validation.
+	if !node.detached && node.parent != nil {
```
yycptt: I think both checks can done inside `validateAccess` and caller can just call `validateAccess` on the node being accessed, instead of its parent.

awln-temporal: i see, if the node is detached, we also ignore the component's lifecycle state in `validateAccess` aka bypass the entire call. Yeah makes more sense.

chasm/registrable_component.go
awln-temporal: Yeah, that's the behavior, i'll add specific docs for the order of precedence.

chasm/tree.go
awln-temporal: edit: serialized node is set during sync of subfields, removing the extra field in Node.

chasm/tree.go
```diff
@@ -739,7 +746,8 @@ func (n *Node) serializeComponentNode() error {
 		}
 
 		n.serializedNode.Data = blob
-		n.serializedNode.GetMetadata().GetComponentAttributes().TypeId = rc.componentID
+		componentAttr := n.serializedNode.GetMetadata().GetComponentAttributes()
```
yycptt: nit: this can be reverted now.

chasm/tree.go
```diff
@@ -442,18 +437,27 @@ func (n *Node) Component(
 // all of a node's ancestors are still in a running state, and can accept writes. In
 // the case of a newly-created node, a detached node, or an OperationIntentObserve
 // intent, the check is skipped.
-func (n *Node) validateAccess(ctx Context) error {
+func (n *Node) validateParentAccess(ctx Context) error {
```
yycptt: nit: I would probably call this one `validateAccess` and the one below `validateAccessHelper`. Also make it clear that only this one should be called.

PR #9091, opened 2026-01-21
Remove nil search attributes from history events / mutable state / ch…

awln-temporal:
## What changed?
Remove nil search attributes from history events / mutable state / chasm node

## Why?
Search attributes can be explicitly set as null on the following events:

* Workflow Start

* Upsert Search Attributes

* Continue-as-new

Starting a workflow or continuing a workflow as new with a nil search attribute records the `{key: nil}` mapping in mutable state and the history event.

Upserting search attributes erases any current search attribute value from persistence (mutable state and visibility record).

Upserting the search attribute as nil in the SDK retains the mapping in the WorkflowInfo as a {key: nil} mapping. On a continue-as-new event, SDK sends this mapping to Server, and the search attribute gets repopulated with the null search attribute in the history event and mutable state.

This last scenario can cause inconsistency between the expected search attributes and what is recorded in history. After upserting a nil search attribute, the mutable state and visibility record no longer shows this search attribute. But, once continuing as new, the search attribute reappears, leading to unexpected NDE if the search attribute is also deregistered.

## CHASM change
This also removes nil search attributes from CHASM nodes on replacement. This mirrors the change to Workflow History Events and Mutable State.

## How did you test it?
- [X] built
- [X] added new unit test(s)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,6 +76,8 @@ Before starting the implementation of any request, you MUST REVIEW the following
 - Run tests after altering code or tests
 - Start with unit tests for fastest feedback
 - Prefer `require` over `assert`, avoid testify suites in unit tests (functional tests require suites for test cluster setup), use `require.Eventually` instead of `time.Sleep` (forbidden by linter)
+- For float comparisons in tests, use `InDelta` or `InEpsilon` instead of `Equal` (enforced by `testifylint`)
+- For error assertions in testify suites, use `s.Require().NoError(err)` instead of `s.NoError(err)` (enforced by `testifylint`)
 
 # Primary Workflows
 ## Software Engineering Tasks
--- a/api/deployment/v1/message.pb.go
+++ b/api/deployment/v1/message.pb.go
@@ -200,6 +200,8 @@ type WorkerDeploymentVersionData struct {
 	// immediately delete the version data from task queues. instead, we mark them as deleted while
 	// keeping the revision number.
 	// Old enough deleted versions are GCed based on update_time.
+	// Deprecated. This mechanism is not safe against reactivation of versions after delete.
+	// Use forget_version flag for synchronous deletion of the version data from TQ.
 	Deleted       bool                             `protobuf:"varint,3,opt,name=deleted,proto3" json:"deleted,omitempty"`
 	Status        v1.WorkerDeploymentVersionStatus `protobuf:"varint,6,opt,name=status,proto3,enum=temporal.api.enums.v1.WorkerDeploymentVersionStatus" json:"status,omitempty"`
 	unknownFields protoimpl.UnknownFields
--- a/api/matchingservice/v1/request_response.pb.go
+++ b/api/matchingservice/v1/request_response.pb.go
@@ -2819,7 +2819,6 @@ type SyncDeploymentUserDataRequest struct {
 	//	aip.dev/not-precedent: Not following Google API format --)
 	UpsertVersionsData map[string]*v110.WorkerDeploymentVersionData `protobuf:"bytes,11,rep,name=upsert_versions_data,json=upsertVersionsData,proto3" json:"upsert_versions_data,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
 	// List of build ids to forget from task queue.
-	// Deprecated. Use upsert_versions_data with deleted=true.
 	ForgetVersions []string `protobuf:"bytes,12,rep,name=forget_versions,json=forgetVersions,proto3" json:"forget_versions,omitempty"`
 	unknownFields  protoimpl.UnknownFields
 	sizeCache      protoimpl.SizeCache
--- a/chasm/lib/scheduler/scheduler.go
+++ b/chasm/lib/scheduler/scheduler.go
@@ -744,6 +744,11 @@ func (s *Scheduler) Update(
 	if s.Sentinel {
 		return nil, ErrSentinel
 	}
+	// UpdateComponent does not reject mutations on completed executions,
+	// so we must check explicitly here.
+	if s.Closed {
+		return nil, ErrClosed
+	}
 	if !s.validateConflictToken(req.FrontendRequest.ConflictToken) {
 		return nil, ErrConflictTokenMismatch
 	}
@@ -795,6 +800,11 @@ func (s *Scheduler) Patch(
 ) (*schedulerpb.PatchScheduleResponse, error) {
 	if s.Sentinel {
 		return nil, ErrSentinel
+	}
+	// UpdateComponent does not reject mutations on completed executions,
+	// so we must check explicitly here.
+	if s.Closed {
+		return nil, ErrClosed
 	}
 	// Handle paused status.
 	if req.FrontendRequest.Patch.Pause != "" {
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -20,24 +20,26 @@ const (
 	buildPlatformTag = "build_platform"
 	goVersionTag     = "go_version"
 
-	instance       = "instance"
-	namespace      = "namespace"
-	namespaceID    = "namespace_id"
-	namespaceState = "namespace_state"
-	sourceCluster  = "source_cluster"
-	targetCluster  = "target_cluster"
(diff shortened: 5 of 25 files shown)


chasm/visibility_test.go
```diff
@@ -176,6 +176,36 @@ func (s *visibilitySuite) TestReplaceCustomSearchAttributes() {
 	_, ok := s.visibility.SA.TryGet(s.mockContext)
 	s.False(ok)
 	s.Nil(s.visibility.CustomSearchAttributes(s.mockContext))
+
+	// Test that nil values are filtered out during replace.
+	s.visibility.ReplaceCustomSearchAttributes(
```
rodrigozhou, line 204: Test `NewVisibilityWithData` with nil values as well please

common/payload/payload.go
```diff
@@ -90,3 +90,19 @@ func isEqual(a, b *commonpb.Payload) bool {
...
 	return bytes.Equal(aEnc, bEnc) && bytes.Equal(a.GetData(), b.GetData())
 }
+
+// FilterNilSearchAttributes returns a new SearchAttributes with nil/empty payload values filtered out.
+// If the input is nil or all values are nil/empty, returns nil.
+// This is used to filter out nil search attributes from workflow start and continue-as-new events.
+// Reuses MergeMapOfPayload which already handles nil payload filtering.
+func FilterNilSearchAttributes(sa *commonpb.SearchAttributes) *commonpb.SearchAttributes {
+	if sa == nil || len(sa.GetIndexedFields()) == 0 {
+		return sa
```
rodrigozhou: Based on your comments above, this should be `return nil`.

service/history/historybuilder/event_factory.go
```diff
@@ -68,7 +69,7 @@ func (b *EventFactory) CreateWorkflowExecutionStartedEvent(
 		FirstExecutionRunId:             firstRunID,
 		OriginalExecutionRunId:          originalRunID,
 		Memo:                            req.Memo,
-		SearchAttributes:                req.SearchAttributes,
+		SearchAttributes:                payload.FilterNilSearchAttributes(req.SearchAttributes),
```
rodrigozhou: I wonder if this is a safe change.
cc: @yycptt

rodrigozhou (changes requested):

chasm/visibility.go
yycptt, line 1: Do we want to do the same thing for memo as well?

service/history/historybuilder/event_factory.go
yycptt: is the concern around replay determinism on the sdk side? I will let sdk folks comment on that.

From server side I can't think of issues on top of my head.

chasm/visibility.go
cretz, line 1: This is great. Can I get confirmation that this PR doesn't affect anything in the past, only newly started/CAN'd workflows?

awln-temporal: I was a bit more unsure about memo fields, since it's actually valid to write nil memo payloads to the visibility record if I recall. If this is the case, I was unsure if this could break any user behavior relying on checking nil memo payloads.

awln-temporal: Yeah that's correct, this would only change any new Workflow runs or new continue-as-new runs. These should all be under separate runIDs and history branches. Let me know if you think otherwise.

rodrigozhou: I think the behavior for memo should be the same.
UpsertMemo has the same behavior as UpsertSearchAttributes.

service/history/historybuilder/event_factory.go
```diff
@@ -480,8 +481,8 @@ func (b EventFactory) CreateContinuedAsNewEvent(
 		Initiator:                    command.Initiator,
 		Failure:                      command.Failure,
 		LastCompletionResult:         command.LastCompletionResult,
-		Memo:                         command.Memo,
-		SearchAttributes:             command.SearchAttributes,
+		Memo:                         payload.FilterNilMemo(command.Memo),
+		SearchAttributes:             payload.FilterNilSearchAttributes(command.SearchAttributes),
```
rodrigozhou: If we ensure that the maps don't have nil values when creating the workflow started event, do we need to call this in continue-as-new command and start child workflow command? They both ended calling `CreateWorkflowExecutionStartedEvent`, so it doesn't seem that it's necessary to call `FilterNil*` here and below.

@yycptt thoughts?

service/history/historybuilder/event_factory.go
```diff
@@ -67,8 +68,8 @@ func (b *EventFactory) CreateWorkflowExecutionStartedEvent(
 		FirstWorkflowTaskBackoff:        request.FirstWorkflowTaskBackoff,
 		FirstExecutionRunId:             firstRunID,
 		OriginalExecutionRunId:          originalRunID,
-		Memo:                            req.Memo,
-		SearchAttributes:                req.SearchAttributes,
+		Memo:                            payload.FilterNilMemo(req.Memo),
+		SearchAttributes:                payload.FilterNilSearchAttributes(req.SearchAttributes),
```
rodrigozhou: Please add some comments explaining why we're doing this here, and not say in the StartWorkflowExecution API.

service/history/historybuilder/event_factory.go
```diff
@@ -727,7 +728,7 @@ func (b *EventFactory) CreateWorkflowPropertiesModifiedEvent(
 	event.Attributes = &historypb.HistoryEvent_WorkflowPropertiesModifiedEventAttributes{
 		WorkflowPropertiesModifiedEventAttributes: &historypb.WorkflowPropertiesModifiedEventAttributes{
 			WorkflowTaskCompletedEventId: workflowTaskCompletedEventID,
-			UpsertedMemo:                 command.UpsertedMemo,
+			UpsertedMemo:                 payload.FilterNilMemo(command.UpsertedMemo),
```
rodrigozhou: This is not needed, is it? Doesn't `UpsertMemo` in the handler already filter nil values?

service/history/historybuilder/event_factory.go
awln-temporal: This is to prevent the continue as new event itself from containing previously deleted search attributes, that may still be present in long running workflows that started with a nil search attribute in the payload (before this change is released). Imo safer to include this.

service/history/historybuilder/event_factory.go
awln-temporal: yeah good point, even more, this prevents users from deleting memo in long running workflows. Removing this part.

service/history/historybuilder/event_factory.go
```diff
@@ -480,9 +483,11 @@ func (b EventFactory) CreateContinuedAsNewEvent(
 		Initiator:                    command.Initiator,
 		Failure:                      command.Failure,
 		LastCompletionResult:         command.LastCompletionResult,
-		Memo:                         command.Memo,
-		SearchAttributes:             command.SearchAttributes,
-		InheritBuildId:               command.InheritBuildId,
+		// Filter nil values here rather than in the API layer because not all
+		// creation paths go through the frontend (continue-as-new, child workflows, replication).
+		Memo:             payload.FilterNilMemo(command.Memo),
+		SearchAttributes: payload.FilterNilSearchAttributes(command.SearchAttributes),
```
rodrigozhou, line 491: This event is added to the source workflow. Please update the comment accordingly.

service/history/historybuilder/event_factory.go
```diff
@@ -847,11 +852,13 @@ func (b *EventFactory) CreateStartChildWorkflowExecutionInitiatedEvent(
...
 			CronSchedule:                 command.CronSchedule,
-			Memo:                         command.Memo,
-			SearchAttributes:             command.SearchAttributes,
-			ParentClosePolicy:            command.GetParentClosePolicy(),
-			InheritBuildId:               command.InheritBuildId,
-			Priority:                     command.Priority,
+			// Filter nil values here rather than in the API layer because not all
+			// creation paths go through the frontend (continue-as-new, child workflows, replication).
+			Memo:              payload.FilterNilMemo(command.Memo),
+			SearchAttributes:  payload.FilterNilSearchAttributes(command.SearchAttributes),
```
rodrigozhou, line 861: ditto

PR #9096, opened 2026-01-21
Add metrics to track invalid search attributes comparisons in ES

rodrigozhou:
## What changed?
Add metrics to track invalid search attributes comparisons in ES

## Why?
Be able to track how often and which namespaces are using invalid comparisons.

## How did you test it?
- [x] built
- [ ] added new unit test(s)

### DIFF
--- a/common/persistence/visibility/store/elasticsearch/query_interceptors.go
+++ b/common/persistence/visibility/store/elasticsearch/query_interceptors.go
@@ -18,7 +18,7 @@ import (
 	"go.temporal.io/server/common/searchattribute/sadefs"
 )
 
-var elasticsearchQueryConverterCmpInvalidTypeCounter = metrics.NewCounterDef("elasticsearch_query_converter_cmp_invalid_type")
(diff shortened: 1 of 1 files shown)


common/persistence/visibility/store/elasticsearch/query_interceptors.go
```diff
@@ -15,6 +18,8 @@ import (
 	"go.temporal.io/server/common/searchattribute/sadefs"
 )
 
+var elasticsearchQueryConverterCmpInvalidTypeCounter = metrics.NewCounterDef("elasticsearch_query_converter_cmp_invalid_type")
```
awln-temporal: imo don't need "type" and "comparison" in the name, invalid comparisons are only possible due to type mismatches. would prefer something like `elasticsearch_query_parameter_type_mismatch`, `elasticsearch_query_invalid_comparison` or something. Likewise, `converter` seems like a niche term for visibility implementation. Just want the metric name to be easier to understand for those discovering metrics.

PR #9107, opened 2026-01-22
Stop using BuildIds SA in Drainage

ShahabT:
## What changed?
The `BuildIds` search attribute is deprecated and we don't want to depend on it for drainage status calculation. updating the internal query to use `TemporalWorklfowVersioningBehavior` and `TemporalWorkerDeploymentVersion` SAs.

## Why?
remove dependency to deprecated SA.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
The new query uses there filters vs two filters in the previous one so it might be slightly less efficient.

### DIFF
--- a/service/worker/workerdeployment/client.go
+++ b/service/worker/workerdeployment/client.go
@@ -1453,10 +1453,14 @@ func (d *ClientImpl) GetVersionDrainageStatus(
 }
 
 func makeDeploymentQuery(version string) string {
-	var statusFilter string
-	deploymentFilter := fmt.Sprintf("= '%s'", worker_versioning.PinnedBuildIdSearchAttribute(version))
-	statusFilter = "= 'Running'"
-	return fmt.Sprintf("%s %s AND %s %s", sadefs.BuildIds, deploymentFilter, sadefs.ExecutionStatus, statusFilter)
+	deploymentFilter := fmt.Sprintf("= '%s'", version)
+	statusFilter := "= 'Running'"
+	behaviorFilter := "= 'Pinned'"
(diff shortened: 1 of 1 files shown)


service/worker/workerdeployment/client.go
```diff
@@ -1453,10 +1453,14 @@ func (d *ClientImpl) GetVersionDrainageStatus(
...
 func makeDeploymentQuery(version string) string {
-	var statusFilter string
-	deploymentFilter := fmt.Sprintf("= '%s'", worker_versioning.PinnedBuildIdSearchAttribute(version))
-	statusFilter = "= 'Running'"
-	return fmt.Sprintf("%s %s AND %s %s", sadefs.BuildIds, deploymentFilter, sadefs.ExecutionStatus, statusFilter)
+	deploymentFilter := fmt.Sprintf("= '%s'", version)
+	statusFilter := "= 'Running'"
+	behaviorFilter := "= 'Pinned'"
+	return fmt.Sprintf("%s %s AND %s %s AND %s %s",
+		sadefs.TemporalWorkerDeploymentVersion, deploymentFilter,
```
Shivs11, line 1460: should this not be TemporalUsedWorkerDeploymentVersions instead of TemporalUsedWorkerDeploymentVersion? 

IMO, we are interested in crafting a query which shall target any workflows which have used the deployment version (which we are sending for the drainage check).

ShahabT: No, if the workflow used to be on this version and still open but now pinned to another version it should not block this version's drainage.

PR #9109, opened 2026-01-22
CHASM workflow command handler for Nexus

stephanos:
## What changed?

- Added workflow command handler registry to CHASM's workflow library.
- Integrated CHASM's workflow library into workflow completion handler.

## Why?

Migrating Nexus from HSM to CHASM.

## How did you test it?

Tests will be ported over once actual command handler implementations are added.

### DIFF
--- a/chasm/lib/nexusoperation/commands.go
+++ b/chasm/lib/nexusoperation/commands.go
@@ -24,8 +24,8 @@ func registerCommandHandlers(registry *command.Registry) error {
 func handleScheduleCommand(
 	chasmCtx chasm.MutableContext,
 	validator command.Validator,
-	workflowTaskCompletedEventID int64,
 	cmd *commandpb.Command,
+	opts command.HandlerOptions,
 ) error {
 	// TODO: Implement CHASM nexus operation scheduling
 	return serviceerror.NewUnimplemented("CHASM nexus operation scheduling not yet implemented")
@@ -34,8 +34,8 @@ func handleScheduleCommand(
 func handleCancelCommand(
 	chasmCtx chasm.MutableContext,
 	validator command.Validator,
-	workflowTaskCompletedEventID int64,
 	cmd *commandpb.Command,
+	opts command.HandlerOptions,
 ) error {
 	// TODO: Implement CHASM nexus operation cancellation
 	return serviceerror.NewUnimplemented("CHASM nexus operation cancellation not yet implemented")
--- a/chasm/lib/workflow/command/handler.go
+++ b/chasm/lib/workflow/command/handler.go
@@ -6,13 +6,17 @@ import (
 	"go.temporal.io/server/chasm"
 )
 
+type HandlerOptions struct {
+	WorkflowTaskCompletedEventID int64
+}
+
 // Handler is a function for handling a workflow command as part of processing a RespondWorkflowTaskCompleted
 // worker request.
 type Handler func(
 	chasmCtx chasm.MutableContext,
 	validator Validator,
-	workflowTaskCompletedEventID int64,
 	command *commandpb.Command,
+	opts HandlerOptions,
 ) error
 
 // Validator is a helper for validating workflow commands.
--- a/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
+++ b/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
@@ -347,14 +347,16 @@ func (handler *workflowTaskCompletedHandler) handleCommand(
 			}
 
 			// Wrap HSM command handler to match CHASM command handler signature.
-			commandHandler = func(_ chasm.MutableContext, v chasmcommand.Validator, id int64, cmd *commandpb.Command) error {
-				return legacyHandler(ctx, handler.mutableState, v, id, cmd)
+			commandHandler = func(_ chasm.MutableContext, v chasmcommand.Validator, cmd *commandpb.Command, opts chasmcommand.HandlerOptions) error {
+				return legacyHandler(ctx, handler.mutableState, v, opts.WorkflowTaskCompletedEventID, cmd)
 			}
 		}
 
 		// Invoke command handler.
 		validator := commandValidator{sizeChecker: handler.sizeLimitChecker, commandType: command.GetCommandType()}
-		err = commandHandler(chasmCtx, validator, handler.workflowTaskCompletedID, command)
+		err = commandHandler(chasmCtx, validator, command, chasmcommand.HandlerOptions{
+			WorkflowTaskCompletedEventID: handler.workflowTaskCompletedID,
+		})
 		var failWFTErr chasmcommand.FailWorkflowTaskError
 		if errors.As(err, &failWFTErr) {
 			if failWFTErr.TerminateWorkflow {
--- a/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler_test.go
+++ b/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler_test.go
@@ -399,7 +399,7 @@ func TestCommandProtocolMessage(t *testing.T) {
 		sentinelErr := errors.New("sentinel: CHASM handler invoked")
 		err := tc.chasmCommandRegistry.Register(
 			enumspb.COMMAND_TYPE_SCHEDULE_NEXUS_OPERATION,
-			func(chasm.MutableContext, chasmcommand.Validator, int64, *commandpb.Command) error {
+			func(chasm.MutableContext, chasmcommand.Validator, *commandpb.Command, chasmcommand.HandlerOptions) error {
 				return sentinelErr
 			},
 		)


chasm/lib/nexusoperation/library.go
```diff
@@ -28,7 +28,7 @@ func (l *Library) Name() string {
 func (l *Library) Components() []*chasm.RegistrableComponent {
 	return []*chasm.RegistrableComponent{
 		chasm.NewRegistrableComponent[*Operation]("operation"),
-		chasm.NewRegistrableComponent[*Operation]("cancellation"),
+		chasm.NewRegistrableComponent[*Cancellation]("cancellation"),
```
stephanos, line 31: Right?

chasm/lib/workflow/workflow.go
```diff
@@ -25,6 +26,9 @@ type Workflow struct {
 
 	// Callbacks map is used to store the callbacks for the workflow.
 	Callbacks chasm.Map[string, *callback.Callback]
+
+	// Operations map is used to store the nexus operations for the workflow.
+	Operations chasm.Map[string, *nexusoperation.Operation]
```
stephanos, line 31: Not used yet, but demonstrates why command registry and types need to live in submodule. If they didn't there'd be an import cycle. Having this already ensures this design will work.

service/history/workflow/command_handler.go
```diff
@@ -7,31 +7,19 @@ import (
...
+	chasmcommand "go.temporal.io/server/chasm/lib/workflow/command"
 	historyi "go.temporal.io/server/service/history/interfaces"
 )
 
 // ErrDuplicateRegistration is returned by a [CommandHandlerRegistry] when it detects duplicate registration.
 var ErrDuplicateRegistration = errors.New("duplicate registration")
 
-// FailWorkflowTaskError is an error that can be returned from a [CommandHandler] to fail the current workflow task and
-// optionally terminate the entire workflow.
-type FailWorkflowTaskError struct {
```
stephanos, line 18: Moving this to CHASM now to break up import cycle.

service/history/workflow/command_handler.go
```diff
@@ -41,12 +29,6 @@ type CommandHandlerRegistry struct {
 	handlers map[enumspb.CommandType]CommandHandler
 }
 
-// CommandValidator is a helper for validating workflow commands.
-type CommandValidator interface {
```
stephanos, line 45: Moved to CHASM, too. More from this package might be moved in follow-up PRs.

chasm/lib/nexusoperation/commands.go
```diff
@@ -0,0 +1,55 @@
...
+		return err
+	}
+	return registry.RegisterHandler(
+		enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION,
+		h.handleCancelCommand,
+	)
+}
+
+func (ch *commandHandler) handleScheduleCommand(
+	chasmCtx chasm.MutableContext,
```
stephanos, line 25: I had briefly prototyped this more (and removed that again), and from that I learned AFACIT that this will need both `MutableContext` as well as `context.Context` (which MutableContext embeds but doesn't yet expose).

chasm/context.go
```diff
@@ -26,6 +26,8 @@ type Context interface {
 	ExecutionCloseTime() time.Time
 	// Logger returns a logger tagged with execution key and other chasm framework internal information.
 	Logger() log.Logger
+	// Library returns a registered library by name.
+	Library(name string) (Library, bool)
```
stephanos: I had to add this to be able to access the Workflow library to get the command registry. Let me know if this is in line with the CHASM design!

chasm/lib/workflow/workflow.go
```diff
@@ -25,15 +29,35 @@ type Workflow struct {
...
+	// This is not persisted.
+	commandRegistry *command.Registry
 }
 
 func NewWorkflow(
-	_ chasm.MutableContext,
+	ctx chasm.MutableContext,
 	msPointer chasm.MSPointer,
 ) *Workflow {
+	lib, ok := ctx.Library(chasm.WorkflowLibraryName)
```
stephanos: I think it's fair to assume that this is always registered.

chasm/lib/nexusoperation/commands.go
```diff
@@ -0,0 +1,55 @@
...
+func (ch *commandHandler) handleCancelCommand(
+	chasmCtx chasm.MutableContext,
+	ms command.Backend,
+	validator command.Validator,
+	workflowTaskCompletedEventID int64,
+	cmd *commandpb.Command,
+) error {
+	// TODO: Implement CHASM nexus operation cancellation
+	return serviceerror.NewUnimplemented("CHASM nexus operation cancellation not yet implemented")
+}
```
stephanos, line 42: My understanding is that we also want to support standalone Nexus operations, and this here might tie this module to workflows too tightly - so we could move all this into `chasm/lib/nexusoperation/nexusworkflow` (now or later) to make that distinction clearer.

chasm/lib/workflow/fx.go
```diff
@@ -2,13 +2,35 @@ package workflow
...
+			// This happens in services that don't have nexus operation support.
+			return nil
+		}
+		return nexusoperation.RegisterCommandHandlers(
+			deps.Library.commandHandlers,
+			deps.EndpointRegistry,
+			deps.Config,
+		)
+	}),
 )
```
stephanos, line 16: Copilot suggested moving the initialization from Nexus' fx to here; which made sense to me because of standalone Nexus operations. Having said that, this is also awkward in scenarios like Matching where there is no Nexus (hence the optional deps). Happy to discuss!

stephanos: (new thought; could move this bit and the command code both into `chasm/lib/nexusoperation/nexusworkflow` together)

chasm/lib/nexusoperation/library.go
gow: Yes

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -317,14 +320,39 @@ func (handler *workflowTaskCompletedHandler) handleCommand(
...
+			commandHandler = func(_ chasm.MutableContext, ms chasmcommand.Backend, v chasmcommand.Validator, id int64, cmd *commandpb.Command) error {
+				return legacyHandler(ctx, ms.(historyi.MutableState), v, id, cmd)
+			}
 		}
+
+		// Invoke command handler.
 		validator := commandValidator{sizeChecker: handler.sizeLimitChecker, commandType: command.GetCommandType()}
-		err := ch(ctx, handler.mutableState, validator, handler.workflowTaskCompletedID, command)
-		var failWFTErr workflow.FailWorkflowTaskError
+		err = commandHandler(chasmCtx, handler.mutableState, validator, handler.workflowTaskCompletedID, command)
```
gow: This looks good. But I have a larger question - Why do we even need this dynamic handler registration (even for hsm)? What was wrong in following the same pattern like in the above `case` statements?
Is it just to break the package dependency?

chasm/lib/nexusoperation/commands.go
gow: Command handlers should be tightly coupled with workflow lib. So this is a good place for it I think.

chasm/context.go
bergundy: I wouldn't do this. Instead put the command registry in the chasm registry.

chasm/lib/nexusoperation/commands.go
bergundy: You don't need context.Context, see here how to deal with that: https://github.com/temporalio/temporal/blob/0ded458f739764eefe2b616285338eead7deb18e/chasm/ms_pointer.go#L24

chasm/lib/nexusoperation/commands.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	return registry.Register(
+		enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION,
+		h.handleCancelCommand,
+	)
+}
+
+func (ch *commandHandler) handleScheduleCommand(
+	chasmCtx chasm.MutableContext,
+	ms command.Backend,
```
bergundy: Let's not call this `Backend`, you want something that the `chasm/lib/workflow` `Workflow` component can implement.

chasm/lib/nexusoperation/commands.go
bergundy: Ideally we would keep everything in a single package IMHO. I would want to register the workflow commands and events along with the rest of CHASM library registration.

chasm/lib/workflow/command/registry.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	enumspb "go.temporal.io/api/enums/v1"
+)
+
+// ErrDuplicateRegistration is returned by a [Registry] when it detects duplicate registration.
+var ErrDuplicateRegistration = errors.New("duplicate registration")
+
+// Registry maintains a mapping of command type to [Handler].
+type Registry struct {
```
bergundy, line 14: Again, can we make this part of the chasm registry?

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
bergundy: I think this is a better pattern because it keeps the entire functionality of the feature contained to a single library.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -317,14 +320,39 @@ func (handler *workflowTaskCompletedHandler) handleCommand(
...
 
 	default:
-		// Nexus command handlers are registered in /components/nexusoperations/workflow/commands.go
-		ch, ok := handler.commandHandlerRegistry.Handler(command.GetCommandType())
-		if !ok {
-			return nil, serviceerror.NewInvalidArgumentf("Unknown command type: %v", command.GetCommandType())
+		var commandHandler chasmcommand.Handler
+		var chasmCtx chasm.MutableContext
+
+		if handler.mutableState.ChasmEnabled() {
```
bergundy, line 330: You will need a way to toggle between HSM and CHASM and you will have to figure out if and how to deal with a mutable state that has both HSM and CHASM nexus operations.

chasm/context.go
pdoerner: Will other CHASM components have commands? IMHO if they are a workflow only concept, I think it is fine to keep them coupled with the `workflow` component

chasm/lib/workflow/fx.go
pdoerner: I prefer keeping the Nexus command handler logic in the `nexusoperation` package, personally

chasm/lib/workflow/workflow.go
pdoerner: Does it still get registered if CHASM is disabled on startup? Might be worth double checking.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
pdoerner: To simplify the migration, we could set a point where we enable CHASM-backed operations. From that time on, new workflows will only use CHASM operations. Then we just have to wait 60 days to know that there are no more valid HSM operations since that is the maximum schedule-to-close timeout.

stephanos: I was under the impression the migration-related work is a separate ticket out of the scope of this PR. Let me know if you think we need to take action on this PR.

chasm/lib/workflow/workflow.go
stephanos: ~in that scenario, why/where would `NewWorkflow` be invoked from? I was making this assumption thinking that it won't/cannot be invoked if the library isn't registered (and CHASM is enabled).~ irrelevant now

stephanos:
I made some changes after a team discussion, most importantly: moved the command handler registry out of the Workflow library; it is passed through the "engine" now just like HSM's. The benefit is that the additional `Library` method I had added to `chasm.Context` isn't needed anymore.

It was also discussed to add it to `chasm.Registry` instead which I prefer not to do as that would make CHASM less workflow-agnostic. In any case, my immediate goal is to unblock future work here and have a separate discussion later on revisiting the best place for it.

chasm/lib/workflow/command/registry.go
stephanos: see https://github.com/temporalio/temporal/pull/9109#issuecomment-3808431814

chasm/context.go
stephanos: see https://github.com/temporalio/temporal/pull/9109#issuecomment-3808431814

chasm/lib/workflow/command/handler.go
```diff
@@ -0,0 +1,32 @@
...
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/server/chasm"
+)
+
+// Handler is a function for handling a workflow command as part of processing a RespondWorkflowTaskCompleted
+// worker request.
+type Handler func(
+	chasmCtx chasm.MutableContext,
+	validator Validator,
+	workflowTaskCompletedEventID int64,
```
bergundy: Maybe for forwards compat we can put this in an options struct?

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -317,14 +322,40 @@ func (handler *workflowTaskCompletedHandler) handleCommand(
...
 
 	default:
-		// Nexus command handlers are registered in /components/nexusoperations/workflow/commands.go
-		ch, ok := handler.commandHandlerRegistry.Handler(command.GetCommandType())
-		if !ok {
-			return nil, serviceerror.NewInvalidArgumentf("Unknown command type: %v", command.GetCommandType())
+		var commandHandler chasmcommand.Handler
+		var chasmCtx chasm.MutableContext
+
+		// TODO: need to handle migration between HSM and CHASM
```
bergundy, line 328: Let's make sure this is tracked.

stephanos: 👍 filed a ticket to make sure.

PR #9111, opened 2026-01-22
Change metric worker_registry_activity_slots_used from histogram to gauge

rkannan82:
## What changed?
Change metric worker_registry_activity_slots_used from histogram to gauge.

Metric was just introduced in https://github.com/temporalio/temporal/pull/9089.

## Why?
Gauge is more appropriate for current slot count as it allows direct queries like max(), avg(), sum() across workers

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
No. The previous change is not released.

### DIFF
--- a/service/matching/workers/registry_impl_test.go
+++ b/service/matching/workers/registry_impl_test.go
@@ -615,8 +615,8 @@ func TestActivitySlotsMetric(t *testing.T) {
 	snapshot := capture.Snapshot()
 	activitySlotsMetrics := snapshot["worker_registry_activity_slots_used"]
 	assert.Len(t, activitySlotsMetrics, 2, "should have activity slots metric for workers with slots info")
-	assert.Equal(t, float64(5), activitySlotsMetrics[0].Value, "should record 5 slots for worker1")
-	assert.Equal(t, float64(10), activitySlotsMetrics[1].Value, "should record 10 slots for worker2")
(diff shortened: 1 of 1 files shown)


service/matching/workers/registry_impl.go
```diff
@@ -396,7 +396,7 @@ func (m *registryImpl) RecordWorkerHeartbeats(nsID namespace.ID, nsName namespac
 func (m *registryImpl) recordActivitySlotsMetric(heartbeats []*workerpb.WorkerHeartbeat) {
 	for _, hb := range heartbeats {
 		if hb.ActivityTaskSlotsInfo != nil {
-			metrics.WorkerRegistryActivitySlotsUsed.With(m.metricsHandler).Record(int64(hb.ActivityTaskSlotsInfo.CurrentUsedSlots))
+			metrics.WorkerRegistryActivitySlotsUsed.With(m.metricsHandler).Record(float64(hb.ActivityTaskSlotsInfo.CurrentUsedSlots))
```
dnr, line 399: what labels is this supposed to have? it looks like there's just a single value that gets overwritten each time?

rkannan82: Yes, no labels for now. Just to get a sense of the overall distribution.

rkannan82: As I think about it .. since the metric is per worker:
1. Either I should export an overall count across all workers
2. or add a label as you suggested. But this is not possible as it will blow up cardinality.
3. Revert and keep histogram.

I should just revert this :)

PR #9114, opened 2026-01-22
[Scheduler] Ensure all errors emitted from scheduler are serviceerrors

lina-temporal:
## What changed?
- All scheduler handler functions now return serviceerrors.
- CHASM's error is wrapped with a conversion helper in the handler file.

## Why?
- Frontend needs to return serviceerrors with status.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/lib/scheduler/handler.go
+++ b/chasm/lib/scheduler/handler.go
@@ -3,7 +3,6 @@ package scheduler
 import (
 	"context"
 	"errors"
-	"fmt"
 
 	"go.temporal.io/api/serviceerror"
 	"go.temporal.io/server/chasm"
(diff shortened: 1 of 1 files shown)


chasm/lib/scheduler/handler.go
```diff
@@ -23,6 +26,27 @@ func newHandler(logger log.Logger, specBuilder *legacyscheduler.SpecBuilder) *ha
...
+	if errors.As(err, &alreadyStartedErr) {
+		return serviceerror.NewAlreadyExists(
+			fmt.Sprintf("schedule %q is already registered", scheduleID),
+		)
+	}
+
+	var notFound *serviceerror.NotFound
+	if errors.As(err, &notFound) {
+		return serviceerror.NewNotFound("schedule not found")
+	}
```
bergundy: Why redact the information here?

chasm/lib/scheduler/handler.go
```diff
@@ -23,6 +26,27 @@ func newHandler(logger log.Logger, specBuilder *legacyscheduler.SpecBuilder) *ha
...
+func convertError(err error, scheduleID string) error {
+	if err == nil {
+		return nil
+	}
+
+	var alreadyStartedErr *chasm.ExecutionAlreadyStartedError
+	if errors.As(err, &alreadyStartedErr) {
+		return serviceerror.NewAlreadyExists(
+			fmt.Sprintf("schedule %q is already registered", scheduleID),
+		)
```
bergundy:
```suggestion
		return serviceerror.NewAlreadyExistsf("schedule %q is already registered", scheduleID)
```

chasm/lib/scheduler/handler.go
```diff
@@ -23,6 +26,27 @@ func newHandler(logger log.Logger, specBuilder *legacyscheduler.SpecBuilder) *ha
 	}
 }
 
+// convertError converts CHASM engine errors to appropriate serviceerrors.
+func convertError(err error, scheduleID string) error {
```
bergundy: You don't need this helper, just wrap the error from `NewExecution`.

lina-temporal: If we're only ever going to have one error that needs wrapping, maybe this should be handled in the CHASM framework, actually.

chasm/lib/scheduler/handler.go
lina-temporal: I hadn't realized I was, only rewriting the error's text to be specific to schedules. I can drop this block and let it through as-is.

PR #9115, opened 2026-01-22
Standalone activities RecordActivityTaskStarted fixes

bergundy:
## What changed?

- Return ObsoleteMatchingTask when standalone activity dispatch task is obsolete.
- Set a stamp on the state and tasks for validation of current attempt.
- Dedupe matching requests by request ID.

## Why?

Avoid retrying these tasks invalid tasks and fix idempotency issues.


fretz12 (approved):

chasm/lib/activity/statemachine.go
dandavison, line 137: I think we need to capture the request ID from the request here (we only have the unit test).

bergundy: Good catch, thanks!

PR #9116, opened 2026-01-22
Return ObsoleteMatchingTask to drop tasks that cannot be retried

spkane31:
## What changed?
`Activity.HandleStarted` now wraps `FailedPrecondition` errors from invalid state transitions into `ObsoleteMatchingTask` errors. 

## Why?
When matching dispatches an activity task that's already in a terminal state (`COMPLETED`, `FAILED`), the state machine returns a `FailedPrecondition` error, which is a retryable error. 

## How did you test it?
- [X] built
- [X] added new unit test(s)

## Potential risks
Minimal, fixes a currently broken path

### DIFF
--- a/chasm/lib/activity/activity.go
+++ b/chasm/lib/activity/activity.go
@@ -1,6 +1,7 @@
 package activity
 
 import (
+	"errors"
 	"fmt"
 	"slices"
 	"time"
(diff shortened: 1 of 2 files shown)


bergundy:
I'm addressing this here already: https://github.com/temporalio/temporal/pull/9115. I will steal your tests though. Thanks.

PR #9117, opened 2026-01-22
Fix emit redirection metric on request on local cluster

yux0:
## What changed?
Fix emit redirection metric on request on local cluster

## Why?
Emit redirection metrics on local requests is confusing

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
No. Just metrics change.

### DIFF
--- a/common/rpc/interceptor/redirection_test.go
+++ b/common/rpc/interceptor/redirection_test.go
@@ -421,6 +421,7 @@ func (s *redirectionInterceptorSuite) TestHandleLocalAPIInvocation_NoRedirection
 
 	redirector := NewRedirection(
 		dynamicconfig.GetBoolPropertyFnFilteredByNamespace(true),
+		dynamicconfig.GetBoolPropertyFnFilteredByNamespace(false),
 		s.namespaceCache,
 		config.DCRedirectionPolicy{Policy: DCRedirectionPolicyAllAPIsForwarding},
 		log.NewNoopLogger(),
@@ -455,6 +456,7 @@ func (s *redirectionInterceptorSuite) TestHandleGlobalAPIInvocation_LocalRouting
 
 	redirector := NewRedirection(
 		dynamicconfig.GetBoolPropertyFnFilteredByNamespace(true),
+		dynamicconfig.GetBoolPropertyFnFilteredByNamespace(false),
 		s.namespaceCache,
 		config.DCRedirectionPolicy{Policy: DCRedirectionPolicyAllAPIsForwarding},
 		log.NewNoopLogger(),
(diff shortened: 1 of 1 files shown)


common/rpc/interceptor/redirection.go
```diff
@@ -293,13 +293,17 @@ func (i *Redirection) AfterCall(
...
 	retError error,
 ) {
-	metricsHandler = metricsHandler.WithTags(metrics.TargetClusterTag(clusterName))
-	metrics.ClientRedirectionLatency.With(metricsHandler).Record(i.timeSource.Now().Sub(startTime))
-	metricsHandler = metricsHandler.WithTags(metrics.NamespaceTag(namespaceName))
-	metrics.ClientRedirectionRequests.With(metricsHandler).Record(1)
-	if retError != nil {
-		metrics.ClientRedirectionFailures.With(metricsHandler).Record(1,
-			metrics.ServiceErrorTypeTag(retError))
+	// Only emit redirection metrics when actual cross-cluster redirection occurred
```
robholland: nit: I think `actual` in the naming and comments is redundant, it's either redirection or it isn't.

common/rpc/interceptor/redirection.go
```diff
@@ -251,7 +251,7 @@ func (i *Redirection) handleRedirectAPIInvocation(
 	namespaceName namespace.Name,
 ) (_ any, retError error) {
 	var resp any
-	var clusterName string
+	var clusterName = i.currentClusterName
```
robholland: Maybe rename `clusterName` to `targetClusterName` within these functions to make it a bit clearer.

common/rpc/interceptor/redirection.go
robholland: If you compare the updated `targetClusterName` (the renamed var) here instead of `targetDC` then it will be the same as the comparison below in `AfterCall` which should help with comprehension.

common/rpc/interceptor/redirection.go
```diff
@@ -293,13 +293,17 @@ func (i *Redirection) AfterCall(
...
 ) {
-	metricsHandler = metricsHandler.WithTags(metrics.TargetClusterTag(clusterName))
-	metrics.ClientRedirectionLatency.With(metricsHandler).Record(i.timeSource.Now().Sub(startTime))
-	metricsHandler = metricsHandler.WithTags(metrics.NamespaceTag(namespaceName))
-	metrics.ClientRedirectionRequests.With(metricsHandler).Record(1)
-	if retError != nil {
-		metrics.ClientRedirectionFailures.With(metricsHandler).Record(1,
-			metrics.ServiceErrorTypeTag(retError))
+	// Only emit redirection metrics when actual cross-cluster redirection occurred
+	isActualRedirection := clusterName != i.currentClusterName
```
robholland: I think with the variable renamed to `targetClusterName` the comparison (which would match above) would be obvious enough that you don't need the variable.

PR #9122, opened 2026-01-23
Add Unwrap to updateError to preserve ResourceExhausted details

spkane31:
## What changed?
Added `Unwrap` function to `updateError` which will preserve the `serviceerrors.ResourceExhausted` error in the `serviceerror.MultiOperationExecutionError`

## Why?
Preserving error details

## How did you test it?
- [X] built
- [ ] added new unit test(s)

## Potential risks
No, giving more information

### DIFF
--- a/chasm/context.go
+++ b/chasm/context.go
@@ -139,7 +139,7 @@ func (c *immutableCtx) getContext() context.Context {
 // NewMutableContext creates a new MutableContext from an existing Context and root Node.
 //
 // NOTE: Library authors should not invoke this constructor directly, and instead use the [UpdateComponent],
-// [UpdateWithNewExecution], or [NewExecution] APIs.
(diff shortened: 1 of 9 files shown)


tests/update_workflow_test.go
```diff
@@ -5738,6 +5738,60 @@ func (s *UpdateWorkflowSuite) TestUpdateWithStart() {
 			s.Fail("timed out waiting for update")
 		}
 	})
+
+	s.Run("return update in-flight limit error", func() {
+		// lower maximum in-flight updates for testing purposes
+		maxInFlight := 1
```
stephanos, line 5744: Is `0` also allowed? Would simplify the test.

spkane31: 0 disables the feature

stephanos: 👍 gotcha

PR #5001, opened 2023-10-19
History Service interceptor

stephanos:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Is hotfix candidate?**

PR #9125, opened 2026-01-23
History Handler cleanup

bergundy:
## What changed?

- Remove defer metrics.CapturePanic and log.CapturePanic calls from all handler methods, that is now handled in the chasm interceptor
- Remove named return parameters (retError, retErr) from handler signatures
- Standardize error variable naming to consistently use 'err'
- Remove isStopped checks as they're redundant with graceful shutdown draining happening before stopping the handler

### DIFF
--- a/chasm/interceptors.go
+++ b/chasm/interceptors.go
@@ -14,16 +14,18 @@ import (
 // ChasmVisibilityInterceptor.
 type ChasmEngineInterceptor struct {
 	engine         Engine
-	logger         log.Logger
+	logger         log.SnTaggedLogger
 	metricsHandler metrics.Handler
 }
 
 func (i *ChasmEngineInterceptor) Intercept(
 	ctx context.Context,
-	req interface{},
+	req any,
 	info *grpc.UnaryServerInfo,
 	handler grpc.UnaryHandler,
-) (resp interface{}, retError error) {
+) (resp any, retError error) {
+	// Capture panics for any handler method, not just CHASM-specific ones. This could have gone into a separate
+	// interceptor, but having it here avoids the overhead of adding another layer to the interceptor chain.
 	defer metrics.CapturePanic(i.logger, i.metricsHandler, &retError)
 
 	ctx = NewEngineContext(ctx, i.engine)
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -1909,7 +1909,9 @@ func (h *Handler) PollWorkflowExecutionUpdate(
 
 func (h *Handler) StreamWorkflowReplicationMessages(
 	server historyservice.HistoryService_StreamWorkflowReplicationMessagesServer,
-) error {
+) (retErr error) {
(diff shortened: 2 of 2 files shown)


chasm/interceptors.go
```diff
@@ -27,9 +24,7 @@ func (i *ChasmEngineInterceptor) Intercept(
 	info *grpc.UnaryServerInfo,
 	handler grpc.UnaryHandler,
 ) (resp interface{}, retError error) {
-	if strings.HasPrefix(info.FullMethod, chasmRequestPrefix) {
-		defer metrics.CapturePanic(i.logger, i.metricsHandler, &retError)
-	}
+	defer metrics.CapturePanic(i.logger, i.metricsHandler, &retError)
```
stephanos, line 29: Why in the CHASM interceptor? Not where I would look for this necessarily. Is a separate, dedicated interceptor unreasonable (e.g. perf reasons)?

service/history/handler.go
```diff
@@ -196,15 +194,10 @@ func (h *Handler) Stop() {
...
 
-func (h *Handler) isStopped() bool {
-	return atomic.LoadInt32(&h.status) == common.DaemonStatusStopped
-}
-
 func (h *Handler) DeepHealthCheck(
 	ctx context.Context,
 	_ *historyservice.DeepHealthCheckRequest,
-) (_ *historyservice.DeepHealthCheckResponse, retError error) {
-	defer log.CapturePanic(h.logger, &retError)
```
stephanos, line 207: I _vaguely_ remember [from my attempt 2+ years ago](https://github.com/temporalio/temporal/pull/5001/changes) that this changed the panic details somehow; but I don't recall the details ... okay, I ran an experiment and there's one meaningful difference: the current version has a `service` log tag that the new one doesn't. Now, is that relevant is the question.

service/history/handler.go
```diff
@@ -2155,13 +1909,7 @@ func (h *Handler) PollWorkflowExecutionUpdate(
 
 func (h *Handler) StreamWorkflowReplicationMessages(
 	server historyservice.HistoryService_StreamWorkflowReplicationMessagesServer,
-) (retError error) {
-	defer metrics.CapturePanic(h.logger, h.metricsHandler, &retError)
```
stephanos, line 2159: Since this is a streaming endpoint; is the CHASM interceptor applied here?

chasm/interceptors.go
bergundy: Perf reasons. Agree it's not the ideal place for it.

service/history/handler.go
bergundy: Hmm... yeah, I can add the tag to the interceptor's logger.

chasm/interceptors.go
stephanos: Could add a comment in the file to redirect readers to the right place.

service/history/handler.go
bergundy: Good catch!

PR #9126, opened 2026-01-23
block per-ns-tq from external users - workflow section

feiyang3cat:
## What changed?
block external users using internal temporal-sys-per-ns-tq workers through triggering workflows

## Why?
enhanced resilience of the cluster

### What is not covered
- activities (need separate pr, impacts current testing)

### What is covered
- **start-workflows (+cron), signal-with-start, multi-option, child-workflows, continue-as-new workflows, schedulers**
- details:

| Workflow Entry Point               | Access Level  | Should Block?                                                     |
|-----------------------------------|---------------|-------------------------------------------------------------------|
| StartWorkflowExecution             | Public API    | Yes — Done                                                         |
| SignalWithStartWorkflowExecution   | Public API    | Yes — Done                                                         |
| ExecuteMultiOperation              | Public API    | Yes — Done                                                         |
| Start Child Workflow               | From Workflow | Yes —> check parent workflow task queue                                                         |
| Continue-As-New                    | From Workflow | No — Inherits parent                                     |
| Schedules                          | Public API    | Yes — Done                                                         |
| Cron                               | Public API    | Yes — Done, covered thru StartWorkflowExecution                                                          |
| Retry                              | Automatic     | No — Inherits original workflow                          |
| ImportWorkflowExecution            | Admin API     | No — Admin-only; used for replication                               |
| StartAdminBatchOperation           | Admin API     | No — Used to start system batch workflows                           |
| Nexus StartOperation/Standalone   Activity          | Public API    | _Yes — not for this pr , syncing related team_                                                       |                                                         |
| Worker Deployment                  | Internal      | No — Allowed        |
| StartBatchOperation                | Public API for dedicated purpose   | No — Allowed        |

## How did you test it?
- [x] built
- [x] added new unit test(s)

## Potential risks
side-effects on legal system workflows/activities -> covered by regression testing (either unit or functional)

### DIFF
--- a/common/primitives/task_queues_test.go
+++ b/common/primitives/task_queues_test.go
@@ -3,66 +3,9 @@ package primitives
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.temporal.io/api/serviceerror"
 )
-
-func TestIsInternalTaskQueueForUserNs(t *testing.T) {
-	tests := []struct {
-		name      string
-		taskQueue string
-		expected  bool
-	}{
-		{
-			name:      "PerNSWorkerTaskQueue is internal",
-			taskQueue: PerNSWorkerTaskQueue,
-			expected:  true,
-		},
-		{
-			name:      "DefaultWorkerTaskQueue is not internal for user namespace",
-			taskQueue: DefaultWorkerTaskQueue,
-			expected:  false,
-		},
-		{
-			name:      "MigrationActivityTQ is not internal for user namespace",
-			taskQueue: MigrationActivityTQ,
-			expected:  false,
-		},
-		{
-			name:      "AddSearchAttributesActivityTQ is not internal for user namespace",
-			taskQueue: AddSearchAttributesActivityTQ,
-			expected:  false,
-		},
-		{
-			name:      "DeleteNamespaceActivityTQ is not internal for user namespace",
-			taskQueue: DeleteNamespaceActivityTQ,
-			expected:  false,
-		},
-		{
-			name:      "DLQActivityTQ is not internal for user namespace",
-			taskQueue: DLQActivityTQ,
-			expected:  false,
-		},
-		{
-			name:      "User defined task queue is not internal",
-			taskQueue: "my-custom-task-queue",
-			expected:  false,
-		},
-		{
-			name:      "Empty string is not internal",
-			taskQueue: "",
-			expected:  false,
-		},
-	}
-
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			result := IsInternalPerNsTaskQueue(tt.taskQueue)
-			assert.Equal(t, tt.expected, result)
-		})
-	}
-}
 
 func TestCheckInternalTaskQueueForUserNsAllowed(t *testing.T) {
 	tests := []struct {
@@ -108,7 +51,7 @@ func TestCheckInternalTaskQueueForUserNsAllowed(t *testing.T) {
 			wantErr:         false,
 		},
 		{
-			name:            "Child is empty, parent is internal - allowed",
+			name:            "Child is empty, parent is internal - not allowed",
 			parentTaskQueue: PerNSWorkerTaskQueue,
 			targetTaskQueue: "",
 			wantErr:         true,
--- a/common/tqid/task_queue_validator_test.go
+++ b/common/tqid/task_queue_validator_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 	enumspb "go.temporal.io/api/enums/v1"
 	taskqueuepb "go.temporal.io/api/taskqueue/v1"
+	"go.temporal.io/server/common/primitives"
 )
 
 func TestNormalizeAndValidate(t *testing.T) {
@@ -147,7 +148,7 @@ func TestNormalizeAndValidate(t *testing.T) {
 }
 
 func TestNormalizeAndValidateUserDefined(t *testing.T) {
-	const perNsTaskQueue = "temporal-sys-per-ns-tq"
+	const perNsTaskQueue = primitives.PerNSWorkerTaskQueue
 
 	tests := []struct {
 		name             string
@@ -230,22 +231,6 @@ func TestNormalizeAndValidateUserDefined(t *testing.T) {
 			expectedError:    "taskQueue length exceeds limit",
 		},
 		{
-			name:             "Sticky queue with internal per-ns task queue without parent",
-			taskQueue:        &taskqueuepb.TaskQueue{Name: perNsTaskQueue, Kind: enumspb.TASK_QUEUE_KIND_STICKY, NormalName: "normal"},
-			parentTaskQueue:  nil,
-			defaultVal:       "",
-			maxIDLengthLimit: 100,
-			expectedError:    "cannot use internal per namespace task queue",
-		},
-		{
-			name:             "Sticky queue with internal per-ns task queue with internal parent",
-			taskQueue:        &taskqueuepb.TaskQueue{Name: perNsTaskQueue, Kind: enumspb.TASK_QUEUE_KIND_STICKY, NormalName: "normal"},
-			parentTaskQueue:  &taskqueuepb.TaskQueue{Name: perNsTaskQueue},
-			defaultVal:       "",
-			maxIDLengthLimit: 100,
-			expectedError:    "",
-		},
-		{
 			name:             "User task queue with empty parent",
 			taskQueue:        &taskqueuepb.TaskQueue{Name: "user-tq"},
 			parentTaskQueue:  &taskqueuepb.TaskQueue{},
@@ -257,7 +242,11 @@ func TestNormalizeAndValidateUserDefined(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err := NormalizeAndValidateUserDefined(tt.taskQueue, tt.defaultVal, tt.parentTaskQueue.GetName(), tt.maxIDLengthLimit)
+			var parentName string
+			if tt.parentTaskQueue != nil {
+				parentName = tt.parentTaskQueue.GetName()
+			}
+			err := NormalizeAndValidateUserDefined(tt.taskQueue, tt.defaultVal, parentName, tt.maxIDLengthLimit)
 
 			if tt.expectedError == "" {
 				require.NoError(t, err)
--- a/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
+++ b/service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
@@ -1447,8 +1447,6 @@ func (handler *workflowTaskCompletedHandler) failWorkflowTaskOnInvalidArgument(
 	err error,
 ) error {
 	var invalidArgument *serviceerror.InvalidArgument
-	// usecases like ValidateContinueAsNewWorkflowExecutionAttributes use fmt.Errorf to
-	// wrap the invalid argument so we need to check the real error type
 	if errors.As(err, &invalidArgument) {
 		return handler.failWorkflowTask(wtFailedCause, err)
 	}
--- a/service/matching/task_queue_partition_manager.go
+++ b/service/matching/task_queue_partition_manager.go
@@ -1121,7 +1121,9 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
 
 func (pm *taskQueuePartitionManagerImpl) ephemeralDataChanged(data *taskqueuespb.EphemeralData) {
 	// for now, only sticky partitions act on ephemeral data, normal partitions ignore it.
-	if pm.partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY || !pm.defaultQueueFuture.Ready() {
+	if pm.partition.Kind() != enumspb.TASK_QUEUE_KIND_STICKY {
+		return
+	} else if !pm.defaultQueueFuture.Ready() {
 		return // not initialized yet
 	}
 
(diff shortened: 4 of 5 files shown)


service/matching/task_queue_partition_manager.go
```diff
@@ -1121,9 +1121,7 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
 
 func (pm *taskQueuePartitionManagerImpl) ephemeralDataChanged(data *taskqueuespb.EphemeralData) {
 	// for now, only sticky partitions act on ephemeral data, normal partitions ignore it.
```
feiyang3cat, line 1123: this change is only for lint purpose

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -452,6 +452,9 @@ func (handler *workflowTaskCompletedHandler) handleCommandScheduleActivity(
 			)
 		},
 	); err != nil || handler.stopProcessing {
+		if err != nil {
```
yycptt: remove

service/history/api/command_attr_validator.go
yycptt, line 1: We should do the same check for continue as new case as well

common/primitives/task_queues.go
```diff
@@ -10,3 +14,35 @@ const (
 	DeleteNamespaceActivityTQ     = "temporal-sys-delete-namespace-activity-tq"
 	DLQActivityTQ                 = "temporal-sys-dlq-activity-tq"
 )
+
+func IsInternalPerNsTaskQueue(taskQueue string) bool {
+	return taskQueue == PerNSWorkerTaskQueue
+}
+
+// CheckNotInternalPerNsTaskQueue is mainly used for workflow task queues.
+func CheckNotInternalPerNsTaskQueue(taskQueue string) error {
```
yycptt: nit: I'd probably move this one and the one below to common/tqid/task_queue_validator.go

common/tqid/task_queue_validator.go
```diff
@@ -59,6 +60,19 @@ func NormalizeAndValidate(
 	return normalizeAndValidate(taskQueue, defaultName, maxIDLengthLimit, true)
 }
 
+// NormalizeAndValidateUserDefined is especially used for user-defined task queues,
+// so that users cannot start any internal workflows.
```
yycptt:
```suggestion
// NormalizeAndValidateUserDefined is similar to NormalizeAndValidate but additionally validate the task queue is not system internal per-namespace task queue.
// This prevents externals callers from operating on any internal namespace level workflows.
```

You may also want to update the comments for NormalizeAndValidate and say something like "considering using NormalizeAndValidateUserDefined wherever possible."

common/primitives/task_queues.go
```diff
@@ -10,3 +14,35 @@ const (
...
+	return nil
+}
+
+// CheckInternalPerNsTaskQueueAllowed is mainly used for activities, child workflows, and other internal components.
+// Only permit use of the internal task queue by activities or child workflows
+// if invoked from a workflow that itself uses the internal task queue.
+func CheckInternalPerNsTaskQueueAllowed(parentTaskQueue, childTaskQueue string) error {
+	if parentTaskQueue == "" && childTaskQueue == "" {
+		return nil // we don't handle both empty string case logic here
+	}
```
yycptt: check out the softassert package and return an internal error here.

common/primitives/task_queues.go
```diff
@@ -10,3 +14,35 @@ const (
...
+	if !IsInternalPerNsTaskQueue(childTaskQueue) {
+		return nil
+	}
+	if !IsInternalPerNsTaskQueue(parentTaskQueue) {
+		return serviceerror.NewInvalidArgumentf(
+			"cannot use internal per-ns-tq, parentTaskQueue=%s, childTaskQueue=%s",
```
yycptt: nit: better not to use abbreviations in errors messages that will be returned to user. 
```suggestion
			"cannot use internal per-namespace task queue, parentTaskQueue=%s, childTaskQueue=%s",
```

common/primitives/task_queues.go
yycptt: nit: now sure how useful this one is, caller can just pass empty string if there's no parent task queue.

common/primitives/task_queues.go
```diff
@@ -10,3 +14,35 @@ const (
...
+func CheckNotInternalPerNsTaskQueue(taskQueue string) error {
+	if IsInternalPerNsTaskQueue(taskQueue) {
+		return serviceerror.NewInvalidArgumentf("cannot use internal per-ns-tq, taskQueue=%s", taskQueue)
+	}
+	return nil
+}
+
+// CheckInternalPerNsTaskQueueAllowed is mainly used for activities, child workflows, and other internal components.
+// Only permit use of the internal task queue by activities or child workflows
+// if invoked from a workflow that itself uses the internal task queue.
```
yycptt:
```suggestion
// CheckInternalPerNsTaskQueueAllowed permits the usage of internal per-namespace task queue when parent component, if exists, is already using the same task queue.
// This is used for ....
```

common/primitives/task_queues.go
feiyang3cat: definitely the code works, but I feel in creases cognitive understanding bar. IMHO, we can keep simple case simple.

common/primitives/task_queues.go
feiyang3cat: I have a different opinion. softassert requires a logger which only held by complex engine/node like objects by design-choice than small util-functions.  I think we either keep it the way it is right now or changing it to a different error message and keeping this a single-purpose simple util function.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -1446,13 +1446,13 @@ func (handler *workflowTaskCompletedHandler) failWorkflowTaskOnInvalidArgument(
 	wtFailedCause enumspb.WorkflowTaskFailedCause,
 	err error,
 ) error {
-
```
feiyang3cat, line 1449: @yycptt request attention

service/history/api/command_attr_validator.go
feiyang3cat: done

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
yycptt: oh wow, so that's an existing bug?

yycptt (approved):

common/primitives/task_queues_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.temporal.io/api/serviceerror"
+)
+
+func TestIsInternalTaskQueueForUserNs(t *testing.T) {
```
dnr: this is really overkill for a function that is a single trivial statement. personally I would just inline the function, but in any case we don't need this unit test.

common/primitives/task_queues_test.go
```diff
@@ -0,0 +1,136 @@
...
+			wantErr:         false,
+		},
+		{
+			name:            "Child is not internal, parent is empty - allowed, this method logic doesn't check for empty string",
+			parentTaskQueue: "",
+			targetTaskQueue: DefaultWorkerTaskQueue,
+			wantErr:         false,
+		},
+		{
+			name:            "Child is empty, parent is internal - allowed",
```
dnr: this one returns an error, not "allowed"

common/tqid/task_queue_validator_test.go
```diff
@@ -144,3 +145,133 @@ func TestNormalizeAndValidate(t *testing.T) {
 		})
 	}
 }
+
+func TestNormalizeAndValidateUserDefined(t *testing.T) {
+	const perNsTaskQueue = "temporal-sys-per-ns-tq"
```
dnr: refer to the constant in primitives, don't duplicate it

common/tqid/task_queue_validator_test.go
```diff
@@ -144,3 +145,133 @@ func TestNormalizeAndValidate(t *testing.T) {
...
+	tests := []struct {
+		name             string
+		taskQueue        *taskqueuepb.TaskQueue
+		parentTaskQueue  *taskqueuepb.TaskQueue
+		defaultVal       string
+		maxIDLengthLimit int
+		expectedError    string
+	}{
+		{
+			name:             "Nil task queue",
```
dnr, line 162: don't duplicate all the tests above here, just include the ones that are relevant to the added logic.

common/tqid/task_queue_validator_test.go
```diff
@@ -144,3 +145,133 @@ func TestNormalizeAndValidate(t *testing.T) {
...
+			name:             "Name exceeds max length",
+			taskQueue:        &taskqueuepb.TaskQueue{Name: strings.Repeat("a", 101)},
+			parentTaskQueue:  nil,
+			defaultVal:       "",
+			maxIDLengthLimit: 100,
+			expectedError:    "taskQueue length exceeds limit",
+		},
+		{
+			name:             "Sticky queue with internal per-ns task queue without parent",
+			taskQueue:        &taskqueuepb.TaskQueue{Name: perNsTaskQueue, Kind: enumspb.TASK_QUEUE_KIND_STICKY, NormalName: "normal"},
```
dnr: this doesn't make any sense as a TaskQueue, that name can't be the name of a sticky queue. why are we using sticky kind here? the 'internal' logic doesn't look at it.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -1446,13 +1446,13 @@ func (handler *workflowTaskCompletedHandler) failWorkflowTaskOnInvalidArgument(
...
-	switch err.(type) {
-	case *serviceerror.InvalidArgument:
+	var invalidArgument *serviceerror.InvalidArgument
+	// usecases like ValidateContinueAsNewWorkflowExecutionAttributes use fmt.Errorf to
+	// wrap the invalid argument so we need to check the real error type
```
dnr: we don't need to write this as a comment, we should assume wrapped errors everywhere and use errors.As consistently (at least across package boundaries)

tests/workflow_test.go
```diff
@@ -1708,6 +1710,103 @@ func (s *WorkflowTestSuite) TestStartWorkflowExecution_Invalid_DeploymentSearchA
 
 }
 
+func (s *WorkflowTestSuite) TestStartWorkflowExecution_InternalTaskQueue() {
```
dnr, line 1710: should include CreateSchedule and UpdateSchedule calls too, with target workflow on the disallowed tq

service/matching/task_queue_partition_manager.go
dnr: lint should not be complaining about files that you're not touching (actually even lines in the same file that you're not touching)

I think this change is misleading since now the "not yet initialized" comment only applies to one of the two clauses. please revert it.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
feiyang3cat: +1

service/matching/task_queue_partition_manager.go
feiyang3cat: I run "make lints" locally it perpetually gets stopped by this one. we can change this to two "if"s

common/primitives/task_queues_test.go
feiyang3cat: emm~ actually at first I included all system related queues as internal task queue and blocked them in the first place. I guess this very much depends on the familiarity of the context

service/matching/task_queue_partition_manager.go
dnr: `make lint-code` is set up to only display the diff from `main`, even locally. See https://github.com/temporalio/temporal/blob/main/Makefile#L381  Is that not working right for you?

common/tqid/task_queue_validator_test.go
feiyang3cat: The two methods have different signature(parameters), and the new method needs to check nil condition.

common/tqid/task_queue_validator_test.go
feiyang3cat: yeah, I checked the sticky queues are named sticky in other tests.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
feiyang3cat: I wrote this only when there was an overlooked bug so as to remind

service/matching/task_queue_partition_manager.go
feiyang3cat: I run `make lint` and that reports this problem. and this is definitely not a good coding style, you are not ok even with `two if` and still want to keep `if and else if` ?

feiyang3cat: or we can drop this comment line ~ I will need to look at this later

tests/workflow_test.go
feiyang3cat: good point~ will add later

service/matching/task_queue_partition_manager.go
dnr: If you remove this file from the diff completely (so it's identical to main), lint still complains? That's not supposed to happen, I'd try to fix that.

I don't mind two if statements, I just mind PRs that aren't well-scoped and just have random other changes

feiyang3cat: no worries. I will revert this change and set another pr for this tmr.

tests/workflow_test.go
feiyang3cat: added

service/matching/task_queue_partition_manager.go
feiyang3cat: reverted. preparing for merge

dnr (approved):

PR #9127, opened 2026-01-23
Convert Unavailable to Aborted when update registry cleared

spkane31:
## What changed?
Changed `AbortedByServerErr` from `serviceerror.NewUnavailable` to `serviceerror.NewABorted`.

## Why?
When the update registry is cleared the server incorrectly returns an Unavailable error which has two issues:
1. Counts against SLA
2. Triggers an internal server retry

Using an Aborted error will cause retries on the client and be consistent with other errors returned from UpdateWithStart

## How did you test it?
- [X] built
- [ ] added new unit test(s)

## Potential risks
Minimal

---

> [!NOTE]
> **Medium Risk**
> Changes retry/error classification and externally observed error codes for workflow update operations, which can affect client retry behavior and operational metrics/SLA accounting.
> 
> **Overview**
> When the workflow Update registry is cleared, the server now returns `AbortedByServerErr` as a gRPC `Aborted` (instead of `Unavailable`), shifting the retry signal to SDK/client retries and avoiding `Unavailable`-based error semantics.
> 
> Server retry classification is updated to treat `serviceerror.Aborted` as handler-retryable (including inside `MultiOperationExecution`), and `UpdateWithStart`/workflow-close paths and docs/tests are adjusted to reflect that `Aborted` may be retried internally first and then by SDKs; the `UpdateWithStart` test expectation changes from “return retryable error” to “succeed after internal retry.”
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3047bc84cf7553095d38fb90081f6c6e1c9739f9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


docs/architecture/workflow-update.md
```diff
@@ -144,21 +144,20 @@ Update Registry intact. But it proved to be too error-prone.
 >
 > There are many more issues that were discovered (and probably more that were not).
 
-Instead, the Workflow Update feature relies on internal retries by the history gRPC handler,
-history gRPC client (on the frontend side), and frontend gRPC handler. So if an Update is removed from
-the Registry due to non-related error, a retryable `WorkflowUpdateAbortedErr` error is returned to
-the `UpdateWorkflowExecution` API caller and subsequent internal retries recreate the Update
+Instead, the Workflow Update feature relies on SDK retries. So if an Update is removed from
```
stephanos: I can see the change addresses the issue; but I don't think the tradeoff (ie to let the issue fall through) to the user is acceptable. One benefit of the current design (using internal retries) is that the latency for users in failure cases stays fairly low. The internal retry - likely - ensures that the internal Update abortion (eg due to a shard ownership lost) is retried internally and never makes its way to the user. I think that's a benefit we can't afford to give up easily given the low latency requirements of the feature.

spkane31: I had two concerns/questions with making Aborted an internal retry:
1. Won't this make the error retried by both the SDK and the Server? So now a failure is going to result in a second request to update which I don't think we want here. I'm looking at how we handle idempotency in this situation.
2. We use `Aborted` in other areas of the code, like in `matchingEngineImpl.ListNexusEndpoints` [here](https://github.com/temporalio/temporal/blob/main/service/matching/matching_engine.go#L2570-L2577) and [here](https://github.com/temporalio/temporal/blob/main/service/matching/matching_engine.go#L2598-L2608). I _think_ that is ok in this scenario, it looks like the error is returned to the user immediately so this is fine

stephanos: re (1): The `Unavailable` (status quo) error is already retried by Server, and then if that wasn't enough, the SDK, isn't it? The idempotency is guaranteed by the fact that the SDK is using the same Update ID (generated if not set); however, this does not account for client restarts. But that's also already the situation.

re (2): There's no issue with returning `Aborted` to the user per se, not at all. But Update Workflow is unique in that it has a stricter low latency requirement than most operations. The idea of the internal retries is that they are faster than a complete roundtrip to the client. I think if we wanted to go in this direction of removing the internal retries, we'd have to demonstrate that there is no negative performance impact.

So I wonder if we can get all 3 constraints satisfied (ie preserve internal retries, have SDK retry when internal retries were not successful, and don't impact SLA).

spkane31: Re-read the implementation and I think that we can get all three of those constraints. The Update ID gives us idempotency and retrying internally first gives us the ability to first retry without waiting for a second retry from the client.

spkane31:
cursor review

stephanos (commented):
🤖 Overall this PR correctly changes `AbortedByServerErr` from `Unavailable` to `Aborted`, adds the right retry classification, updates docs, and adjusts the test. A few items worth addressing.

The existing review thread between @stephanos and @spkane31 covers the key design concern (preserving internal retries + SDK fallback). The latest code appears to satisfy all three constraints mentioned in that discussion: internal retry first, SDK retry as fallback, and Update ID idempotency.

common/util.go
```diff
@@ -329,7 +329,8 @@ func IsServiceHandlerRetryableError(err error) bool {
 
 	switch err := err.(type) {
 	case *serviceerror.Internal,
-		*serviceerror.Unavailable:
+		*serviceerror.Unavailable,
+		*serviceerror.Aborted:
```
stephanos: 🤖 `high` **[design]**: Adding `*serviceerror.Aborted` to `IsServiceHandlerRetryableError` has a broader blast radius than the update use case. This function is used by the `RetryableInterceptor` in *both* the frontend and matching gRPC services (`service/frontend/fx.go:342`, `service/matching/fx.go:67`), as well as the history handler retry utility (`service/history/api/retry_util.go:17`).

This means `Aborted` errors from `matchingEngineImpl.ListNexusEndpoints` (ownership-related failures at `service/matching/matching_engine.go:2624,2628,2654`) will now also be retried internally by the matching service handler interceptor. Previously these were returned immediately to the caller. Whether that's correct or harmful depends on whether those ownership-check failures are transient — but the current semantics of those `Aborted` errors suggest the node is *not* the owner, so retrying on the same node won't help.

Suggestion: Either (1) don't add `Aborted` globally here and instead handle the retry specifically in the update/history path, or (2) verify that the Nexus endpoint `Aborted` errors are safe to retry at the handler level and document that decision.

service/history/workflow/update/errors_failures.go
```diff
@@ -8,8 +8,11 @@ import (
 )
 
 var (
-	registryClearedErr          = errors.New("update registry was cleared")
-	AbortedByServerErr          = serviceerror.NewUnavailable("workflow update was aborted")
+	//nolint:staticcheck
```
stephanos: Could you add a comment at the end mentioning briefly why this was added?

like `//nolint:staticcheck // ...` is often the pattern I see.

service/history/workflow/update/errors_failures.go
```diff
@@ -8,8 +8,11 @@ import (
...
+	registryClearedErr = errors.New("update registry was cleared")
+	// AbortedByServerErr is an SDK-retryable error returned when an update is aborted by the server
+	// (e.g., when the update registry is cleared). SDKs will automatically retry this error.
+	AbortedByServerErr          = serviceerror.NewAborted("workflow update was aborted")
```
stephanos: 🤖 `small` **[naming]**: The variable name `AbortedByServerErr` no longer precisely describes the error type change. Previously it was `Unavailable` (server-side signal), now it's `Aborted` (SDK-retryable signal). The name is still reasonable but the alignment between `registryClearedErr` → `AbortedByServerErr` → `serviceerror.NewAborted(...)` could be clearer. The comment is good though.

docs/architecture/workflow-update.md
```diff
@@ -291,8 +290,8 @@ stateDiagram-v2
     blockCompleted --> wfCompletedFailure: Workflow completed/continuing after Update accepted
     blockCompleted --> updateAbortedError: Registry cleared
 ```
-`ErrWorkflowClosing` and `WorkflowUpdateAbortedErr` are retryable errors. Even they are set on futures,
-they will not be returned to the API caller but will be retried internally.
+`ErrWorkflowClosing` and `AbortedByServerErr` are both internally retryable and SDK-retryable errors.
```
stephanos: 🤖 `med` **[bug]**: The updated documentation says "The server will first attempt internal retries, and only if those fail will the error be returned to the SDK for client-side retry." However, looking at the code in `update.go:WaitLifecycleStage`, the `registryClearedErr` is converted directly to `AbortedByServerErr` and returned — there is no internal retry loop in `WaitLifecycleStage` itself. Internal retries happen at the gRPC interceptor level (because `Aborted` is now in `IsServiceHandlerRetryableError`). The doc should accurately describe the mechanism: the gRPC handler interceptor retries `Aborted`, not a retry loop within the update code.

Suggestion: Clarify that "internal retries" refers to the gRPC handler retry interceptor, not logic within the update package itself.

tests/update_workflow_test.go
```diff
@@ -5675,14 +5675,19 @@ func TestWorkflowUpdateSuite(t *testing.T) {
...
+				// until it succeeds on the new workflow started by the hook.
+				// update-with-start will do an internal retry and succeed
+				_, err = s.TaskPoller().PollAndHandleWorkflowTask(s.Tv(),
+					func(task *workflowservice.PollWorkflowTaskQueueResponse) (*workflowservice.RespondWorkflowTaskCompletedRequest, error) {
+						return &workflowservice.RespondWorkflowTaskCompletedRequest{
+							Messages: s.UpdateAcceptCompleteMessages(s.Tv(), task.Messages[0]),
+						}, nil
+					})
+				s.NoError(err)
+
```
stephanos: 🤖 `med` **[testing]**: The test was renamed from "return retryable error after retry" to "success after internal retry" and the assertions changed from expecting an error to expecting success. This is a significant behavioral change — the test previously verified that after server-side retry exhaustion, the client received a retryable `Aborted` error. Now it verifies that the server keeps retrying until success.

This raises a question: what happens if the internal retry also fails (e.g., the new workflow started by the hook also closes immediately)? There's no test covering the case where internal retries are exhausted and the `Aborted` error is eventually returned to the SDK. That failure path should be tested to ensure the error propagates correctly.

Suggestion: Add a test case for when internal retries are exhausted and the `Aborted` error reaches the client.

docs/architecture/workflow-update.md
```diff
@@ -144,21 +144,20 @@ Update Registry intact. But it proved to be too error-prone.
...
 and might be evicted any time. Therefore, the Workflow Update feature relies on a properly
 configured cache size. If the cache is too small, it will evict Workflow contexts too soon and their
 Update Registry will be cleared. Note that in that case, all in-flight Updates are aborted with a
-retryable error; and the frontend will retry the `UpdateWorkflowExecution` call.
+retryable error; and the SDK will retry the `UpdateWorkflowExecution` call.
 
 ### Aborting an Update
 An Update is aborted when:
-1. The Update Registry is cleared. Then, a retryable `WorkflowUpdateAbortedErr` error is returned
+1. The Update Registry is cleared. Then, a retryable `AbortedByServerErr` error (retried by SDKs) is returned
```
stephanos: Apart from the `AbortedByServerErr` rename, don't we want to revert the other changes in this doc? e.g. "Instead, the Workflow Update feature relies on SDK retries"

service/history/api/multioperation/api.go
```diff
@@ -147,7 +147,7 @@ func Invoke(
 		res, err = uws.Invoke(ctx)
 		if err != nil {
 			// If the Update-with-Start encountered the same error of a closing workflow again, it will convert
-			// the error to Aborted (which is a retryable error) to allow the client to retry the operation.
+			// the error to Aborted (which is an SDK-retryable error) to allow the client to retry the operation.
```
stephanos: Same here and all other comment/doc updates that omit the internal server retries now.

spkane31: Done!

docs/architecture/workflow-update.md
spkane31: Done!

common/serviceerror/update_aborted.go
```diff
@@ -0,0 +1,28 @@
...
+	"go.temporal.io/api/serviceerror"
+	errordetailsspb "go.temporal.io/server/api/errordetails/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// AbortedByServer wraps a serviceerror.Aborted and is returned when a workflow update
+// is aborted by the server (e.g., when the update registry is cleared).
+// This error will be retried internally by the server without forcing all serviceerror.Aborted errors to be retried.
+type AbortedByServer struct {
```
yycptt, line 13: nit: probably put "Update" in the name if this is meant to only be used by WorkflowUpdate. Looks like the errordetails is specific to Update.

yycptt (commented):
> When the update registry is cleared the server incorrectly returns an Unavailable error which has two issues:
>
> 1. Counts against SLA
> 2. Triggers an internal server retry
>
> Using an Aborted error will cause retries on the client and be consistent with other errors returned from UpdateWithStart

Change itself LGTM, but I feel I don't really understand the goal of this PR.

Issue 2 ^ can be removed I think, given we will still retry the new error type on frontend.

Re issue 1, I am under the impression that if it's server clearing mutable state and causing user update to be lost and need to retry, that's a server issue and it should against our SLA. I was not following changes on the Update side so maybe my impression is wrong.

Issue #245, opened 2020-03-23
Cassandra timestamp type


wxing1292:
Try not to use cassandra timestamp, due to loss of precision.

FYI:
https://github.com/temporalio/temporal/blob/master/schema/cassandra/temporal/schema.cql#L71

alexshtin:
Most of them are gone besides these 3:
```
CREATE TABLE executions (
  visibility_ts                  timestamp, -- unique identifier for timer tasks for an execution

CREATE TABLE cluster_membership
(
    session_start        timestamp,
    last_heartbeat       timestamp,

wxing1292:
not sure about the cluster membership, maybe ok?

for execution table, plz change the `timestamp` to bigint or any data type which will not lose precision.

samarabbas:
We may not do this for Temporal.  Millisecond precision should be ok.

wxing1292:
this is not about ok or not ok
this is about persist a record into db, and db modifies it without consent.

consider in the future, using mutable state to calculate the timer to be deleted, say once a workflow is finished, delete all user timer in the future which have not fired yet.

samarabbas:
I'm not sure I fully understand the comment.  Originally the issue was filed to not use Cassandra timestamp due to precision issue.  We decided to not do this for Temporal as we feel timer granularity beyond millisecond is probably not needed at the moment.  I think we have a rounding issue where a timer could be fired earlier than expected due to rounding error.
As for your other comment we have filed an issue to delete unnecessary timers through a DB scan.  #574

wxing1292:
> I'm not sure I fully understand the comment. Originally the issue was filed to not use Cassandra timestamp due to precision issue. We decided to not do this for Temporal as we feel timer granularity beyond millisecond is probably not needed at the moment. I think we have a rounding issue where a timer could be fired earlier than expected due to rounding error.
> As for your other comment we have filed an issue to delete unnecessary timers through a DB scan. #574

short answer: the precision issue will open a can of worms
the main idea here is to have a contract with DB layer: I get the same data that I stored.

If this issue is not addressed, later (now?) the business logic will do all kinds of hacks trying to compensate the missing precision.

yiminc:
Revisiting this issue after several years.

**Current state:**
- The schema still uses Cassandra `timestamp` type for `visibility_ts`, `session_start`, and `last_heartbeat`
- However, the codebase now explicitly handles millisecond precision via `UnixMilliseconds()` in `common/persistence/data_interfaces.go`
[...]

PR #9134, opened 2026-01-25
Add linter rules to enforce UnixMilliseconds() in Cassandra persistence

yiminc:
## Summary
- Add forbidigo linter rules to prevent direct use of `.Unix()`, `.UnixMilli()`, or `.UnixNano()` method calls in the Cassandra persistence package
- Rules enforce using `p.UnixMilliseconds()` for all CQL timestamp values

## Why?
Cassandra TIMESTAMP type has millisecond precision. Using the wrong time conversion can cause:
- **`.Unix()`**: Returns seconds, not milliseconds - timestamps would be off by 1000x
- **`.UnixNano()`**: Returns nanoseconds - precision loss when Cassandra truncates to milliseconds
- **`.UnixMilli()`**: Works but doesn't handle zero-time edge cases like `UnixMilliseconds()` does

The existing `p.UnixMilliseconds()` function (in `common/persistence/data_interfaces.go`) handles these edge cases and should be used consistently.

## What changed?
- Added 3 forbidigo patterns for `Unix`, `UnixMilli`, `UnixNano` method identifiers
- Added exclusion rules so patterns only apply to `common/persistence/cassandra/*.go`
- Tests are excluded from these rules

## Test plan
- [x] Verified linter correctly flags `.Unix()` method calls in cassandra persistence
- [x] Verified `time.Unix()` function calls are NOT flagged (no false positives)
- [x] Verified no existing violations in cassandra persistence code
- [x] Rules scoped to only cassandra persistence package

Addresses #245

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/.golangci.yml
+++ b/.github/.golangci.yml
@@ -38,6 +38,12 @@ linters:
           msg: "Please avoid using panic in application code"
         - pattern: time\.Now
           msg: "Using time.Now is not allowed in chasm/lib package (non-test files), use ctx.Now(component) instead"
+        - pattern: '^Unix$'
+          msg: "Do not use .Unix() for Cassandra timestamps (returns seconds). Use p.UnixMilliseconds() which returns milliseconds."
+        - pattern: '^UnixMilli$'
+          msg: "Do not use .UnixMilli() for Cassandra timestamps. Use p.UnixMilliseconds() for consistency and proper zero-time handling."
+        - pattern: '^UnixNano$'
+          msg: "Do not use .UnixNano() for Cassandra timestamps. Use p.UnixMilliseconds() which returns milliseconds."
     depguard:
       rules:
         main:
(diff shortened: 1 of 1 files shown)


yycptt:
I think the ask of the original ticket is to stop using `timestamp` type in cassandra, at least for new definitions. 
Also sql schema should not use `DATETIME(6)` which is micro-second...

So I guess I am fine with blocking Unix and UnixMilli, but blocking UnixNano will prevent people from doing ^ for new tables.

yiminc:
We can relax the lint rule (to allow UnixNano) when it is needed in future.

.github/.golangci.yml
```diff
@@ -38,6 +38,12 @@ linters:
           msg: "Please avoid using panic in application code"
         - pattern: time\.Now
           msg: "Using time.Now is not allowed in chasm/lib package (non-test files), use ctx.Now(component) instead"
+        - pattern: '^Unix$'
+          msg: "Do not use .Unix() for Cassandra timestamps (returns seconds). Use p.UnixMilliseconds() which returns milliseconds."
+        - pattern: '^UnixMilli$'
+          msg: "Do not use .UnixMilli() for Cassandra timestamps. Use p.UnixMilliseconds() for consistency and proper zero-time handling."
+        - pattern: '^UnixNano$'
+          msg: "Do not use .UnixNano() for Cassandra timestamps. Use p.UnixMilliseconds() which returns milliseconds."
```
yycptt, line 46: nit: probably note it's ok to remove this rule when new time fields are no longer using `timestamp` type

PR #9140, opened 2026-01-26
Deprecate FunctionalTestBase

stephanos:
## What changed?

1. add lint rule to not use FunctionalTestBase for new functional tests.
2. basic documentation for new `TestEnv` alternative approach

## Why?

Enforce policy for new tests while migrating existing ones.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/.github/.golangci.yml
+++ b/.github/.golangci.yml
@@ -187,7 +187,7 @@ linters:
         text: "panic"
         linters:
           - forbidigo
-      - path: tests/testcore/.*\.go$ # still used in tests/testcore/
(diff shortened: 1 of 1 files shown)


.github/.golangci.yml
```diff
@@ -185,6 +187,10 @@ linters:
         text: "panic"
         linters:
           - forbidigo
+      - path-except: tests/[^/]+\.go$ # scoped to files inside tests/ only as it's still used (and wrapped) in tests/testcore/
```
bergundy: Don't you mean `path` instead of `path-except`?

stephanos: Let me remove the double negative and allow it in `tests/testcore/` instead

PR #9141, opened 2026-01-26
Add per-workflow scheduler for history task processing

prathyushpv:
## What changed?
Add a ExecutionQueueScheduler to serialize tasks from a busy workflow execution. Currently, all workflow tasks go through FIFO scheduler. We have a per-workflow-execution lock that must be aquired in each history task. But if a single execution has large number of tasks, each of these tasks compete for this lock and create large number of retries. By Adding this new scheduler, we can serialize task processing from such busy workflows. These tasks are routed to a new WorkflowQueueScheduler when lock contention is detected in a workflow. This will create a new queue for that workflow. Additional tasks for this workflow will be then routed to this new queue. This queue will be cleaned up after a few seconds of inactivity from that workflow execution. We have added a ExecutionAwareScheduler which will manage this routing of workflow tasks to either FIFO scheduler or this new WQ Scheduler.

```
Task → InterleavedWeightedRoundRobinScheduler
            ↓
      ExecutionAwareScheduler
            ↓
      ┌─────┴─────┐
      ↓           ↓
  FIFOScheduler   ExecutionQueueScheduler
  (normal path)   (contended executions)
```

This new scheduler is only enabled when history.taskSchedulerEnableExecutionQueueScheduler is enabled. The number of workflow queues created in this scheduler will be controlled by config history.taskSchedulerWorkflowQueueSchedulerQueueSize.
Tasks are routed to FIFOScheduler(Like the way it was before this change) if number of queues reaches this value.

A new set of goroutines is spawned for each queue in this new scheduler. This is fine here as we don’t expect more than a few hundred hot workflows per history host. This simplifies the design for this scheduler.

## Why?
To reduce workflow lock contention and wasted history CPU when tasks are competing for workflow lock.

## Benchmark Results
Collected by running 2,000 parallel activities from a single workflow with a 5ms lock timeout to trigger contention. EQS queue concurrency = 2, max queues = 500.

### Task Routing & Failures

| Metric | **EQS ENABLED (C=2)** | **EQS DISABLED** | Improvement |
| --- | --- | --- | --- |
| EQS Tasks Submitted | 1,953 | 0 | - |
| EQS Tasks Completed | 1,940 | 0 | - |
| EQS Tasks Failed | 13 | 0 | - |
| EQS Tasks Aborted | 0 | 0 | - |
| EQS Submit Rejected | 0 | 0 | - |
| FIFO Tasks Completed | 65 | 2,051 | - |
| **Total Failures** | **66** | **6,382** | **97x fewer failures** |

### End-to-End Task Latency (task_latency_queue)

| Percentile | **EQS ENABLED (C=2)** | **EQS DISABLED** | Improvement |
| --- | --- | --- | --- |
| Average | 426ms | 908ms | **2.1x faster** |
| P90 | 668ms | **1,599ms** | **2.4x faster** |
| P99 | 729ms | **3,982ms** | **5.5x faster** |
| Max | 736ms | **6,773ms** | **9.2x faster** |

### Runtime

| Metric | **EQS ENABLED (C=2)** | **EQS DISABLED** | Improvement |
| --- | --- | --- | --- |
| Test Time | 4.6s | 8.9s | **1.9x faster** |

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/common/tasks/execution_aware_scheduler.go
+++ b/common/tasks/execution_aware_scheduler.go
@@ -1,6 +1,8 @@
 package tasks
 
 import (
+	"time"
+
 	"go.temporal.io/server/common/clock"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/metrics"
@@ -13,8 +15,14 @@ type (
 	ExecutionAwareSchedulerOptions struct {
 		// Enabled controls whether the executionQueueScheduler is active.
 		Enabled func() bool
-		// ExecutionQueueSchedulerOptions contains configuration for the executionQueueScheduler.
-		ExecutionQueueSchedulerOptions ExecutionQueueSchedulerOptions
+		// MaxQueues is the maximum number of concurrent execution queues.
+		// When this limit is reached, new queues are rejected and tasks fall back to the base scheduler.
+		MaxQueues func() int
+		// QueueTTL is how long an idle queue stays in the map before being swept.
+		QueueTTL func() time.Duration
+		// QueueConcurrency is the max number of worker goroutines per queue.
+		// Values <= 0 are capped to 1 (strictly sequential).
+		QueueConcurrency func() int
 	}
 
 	// ExecutionAwareScheduler is a scheduler that wraps a base scheduler and adds
@@ -45,7 +53,9 @@ func NewExecutionAwareScheduler[T Task](
 	return &ExecutionAwareScheduler[T]{
 		baseScheduler: baseScheduler,
 		executionQueueScheduler: newExecutionQueueScheduler(
-			&options.ExecutionQueueSchedulerOptions,
+			options.MaxQueues,
+			options.QueueTTL,
+			options.QueueConcurrency,
 			queueKeyFn,
 			logger,
 			metricsHandler,
--- a/common/tasks/execution_queue_scheduler.go
+++ b/common/tasks/execution_queue_scheduler.go
@@ -13,17 +13,6 @@ import (
 )
 
 type (
-	ExecutionQueueSchedulerOptions struct {
-		// MaxQueues is the maximum number of concurrent execution queues.
-		// When this limit is reached, TrySubmit will reject new queues.
-		MaxQueues func() int
-		// QueueTTL is how long an idle queue stays in the map before being swept.
-		QueueTTL func() time.Duration
-		// QueueConcurrency is the max number of worker goroutines per queue.
-		// Values <= 0 are capped to 1 (strictly sequential).
-		QueueConcurrency func() int
-	}
-
 	// QueueKeyFn extracts a queue key from a task for queue routing.
 	QueueKeyFn[T Task] func(T) any
 
@@ -56,11 +45,13 @@ type (
 		shutdownChan chan struct{}
 		shutdownWG   sync.WaitGroup
 
-		options        *ExecutionQueueSchedulerOptions
-		queueKeyFn     QueueKeyFn[T]
-		logger         log.Logger
-		metricsHandler metrics.Handler
-		timeSource     clock.TimeSource
+		maxQueues        func() int
+		queueTTL         func() time.Duration
+		queueConcurrency func() int
+		queueKeyFn       QueueKeyFn[T]
+		logger           log.Logger
+		metricsHandler   metrics.Handler
+		timeSource       clock.TimeSource
 
 		// mu protects the queues map, sweeperRunning, and all executionQueue fields.
 		// All task submissions and worker pops are serialized through this lock.
@@ -72,20 +63,24 @@ type (
 
 // newExecutionQueueScheduler creates a new executionQueueScheduler.
 func newExecutionQueueScheduler[T Task](
-	options *ExecutionQueueSchedulerOptions,
+	maxQueues func() int,
+	queueTTL func() time.Duration,
+	queueConcurrency func() int,
 	queueKeyFn QueueKeyFn[T],
 	logger log.Logger,
 	metricsHandler metrics.Handler,
 	timeSource clock.TimeSource,
 ) *executionQueueScheduler[T] {
 	s := &executionQueueScheduler[T]{
-		shutdownChan:   make(chan struct{}),
-		options:        options,
-		queueKeyFn:     queueKeyFn,
-		logger:         logger,
-		metricsHandler: metricsHandler,
-		timeSource:     timeSource,
-		queues:         make(map[any]*executionQueue[T]),
+		shutdownChan:     make(chan struct{}),
+		maxQueues:        maxQueues,
+		queueTTL:         queueTTL,
+		queueConcurrency: queueConcurrency,
+		queueKeyFn:       queueKeyFn,
+		logger:           logger,
+		metricsHandler:   metricsHandler,
+		timeSource:       timeSource,
+		queues:           make(map[any]*executionQueue[T]),
 	}
 	s.status.Store(common.DaemonStatusInitialized)
 	return s
@@ -144,7 +139,7 @@ func (s *executionQueueScheduler[T]) TrySubmit(task T) bool {
 	}
 
 	q, exists := s.queues[key]
-	if !exists && len(s.queues) >= s.options.MaxQueues() {
+	if !exists && len(s.queues) >= s.maxQueues() {
 		s.mu.Unlock()
 		metrics.ExecutionQueueSchedulerSubmitRejected.With(s.metricsHandler).Record(1)
 		return false
@@ -165,7 +160,7 @@ func (s *executionQueueScheduler[T]) appendTaskLocked(key any, q *executionQueue
 		metrics.ExecutionQueueSchedulerQueueCount.With(s.metricsHandler).Record(float64(len(s.queues)))
 	}
 	q.tasks = append(q.tasks, entry)
-	if q.activeWorkers < max(1, s.options.QueueConcurrency()) {
+	if q.activeWorkers < max(1, s.queueConcurrency()) {
 		q.activeWorkers++
 		s.shutdownWG.Add(1)
 		go s.runWorker(q)
@@ -221,7 +216,7 @@ func (s *executionQueueScheduler[T]) runWorker(q *executionQueue[T]) {
 func (s *executionQueueScheduler[T]) runSweeper() {
 	defer s.shutdownWG.Done()
 
-	ch, t := s.timeSource.NewTimer(s.options.QueueTTL())
+	ch, t := s.timeSource.NewTimer(s.queueTTL())
 	defer t.Stop()
 
 	for {
@@ -230,7 +225,7 @@ func (s *executionQueueScheduler[T]) runSweeper() {
 			if s.sweepIdleQueues() {
 				return
 			}
-			t.Reset(s.options.QueueTTL())
+			t.Reset(s.queueTTL())
 		case <-s.shutdownChan:
 			return
 		}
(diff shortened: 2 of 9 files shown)


common/metrics/metric_defs.go
```diff
@@ -1069,6 +1069,21 @@ var (
...
+	ExecutionQueueSchedulerTasksSubmitted    = NewCounterDef("execution_queue_scheduler_tasks_submitted")
+	ExecutionQueueSchedulerTasksCompleted    = NewCounterDef("execution_queue_scheduler_tasks_completed")
+	ExecutionQueueSchedulerTasksFailed       = NewCounterDef("execution_queue_scheduler_tasks_failed")
+	ExecutionQueueSchedulerTasksAborted      = NewCounterDef("execution_queue_scheduler_tasks_aborted")
+	ExecutionQueueSchedulerSubmitRejected    = NewCounterDef("execution_queue_scheduler_submit_rejected")
+	ExecutionQueueSchedulerTaskLatency       = NewTimerDef("execution_queue_scheduler_task_latency")
+	ExecutionQueueSchedulerQueueWaitTime     = NewTimerDef("execution_queue_scheduler_queue_wait_time")
+
+	// FIFOScheduler metrics
+	FIFOSchedulerTasksCompleted = NewCounterDef("fifo_scheduler_tasks_completed")
```
prathyushpv: This is added to compare total task failures when new scheduler is enabled. I can remove this after the change is approved and we don't have to benchmark the new scheduler again.

tests/execution_queue_scheduler_test.go
```diff
@@ -0,0 +1,507 @@
...
+	for _, recording := range snapshot[metricName] {
+		if v, ok := recording.Value.(int64); ok {
+			total += v
+		}
+	}
+	return total
+}
+
+// getTimerStats returns count, total, average, min, max, p50, p90, p99 for a timer metric.
+func getTimerStats(snapshot map[string][]*metricstest.CapturedRecording, metricName string) (count int, avg, min, max, p50, p90, p99 time.Duration) {
```
prathyushpv: This is for benchmarking and comparing new scheduler with existing one. I can remove this before merging.

common/tasks/execution_aware_scheduler.go
```diff
@@ -0,0 +1,117 @@
...
+) *ExecutionAwareScheduler[T] {
+	return &ExecutionAwareScheduler[T]{
+		baseScheduler: baseScheduler,
+		executionQueueScheduler: NewExecutionQueueScheduler(
+			&options.ExecutionQueueSchedulerOptions,
+			queueKeyFn,
+			logger,
+			metricsHandler,
+			timeSource,
+		),
```
yycptt: just for the sake of discussion, I wonder if it make sense to make executionQueueScheduler host level instead of per queue. I don't feel it really matters practically, just logically wanna see if it makes sense to match the scope of the lock which is also host level.

common/tasks/execution_queue_scheduler.go
```diff
@@ -0,0 +1,362 @@
...
+	s.appendTaskLocked(key, entry)
+	s.mu.Unlock()
+	metrics.ExecutionQueueSchedulerTasksSubmitted.With(s.metricsHandler).Record(1)
+	return true
+}
+
+// appendTaskLocked appends a task to the queue for the given key, creating the
+// queue and spawning a worker if needed. Must be called with s.mu held.
+func (s *ExecutionQueueScheduler[T]) appendTaskLocked(key any, entry taskEntry[T]) {
+	q, exists := s.queues[key]
```
yycptt: nit: we should already know if the queue exists or not when this method get called?

common/tasks/fifo_scheduler_test.go
```diff
@@ -199,5 +199,6 @@ func (s *fifoSchedulerSuite) newTestProcessor() *FIFOScheduler[*MockTask] {
 			},
 		},
 		log.NewNoopLogger(),
+		nil, // No metrics in tests
```
yycptt: nit: I'd just use noop metrics handler here in case we started to emit metrics in test in the future.

service/history/queues/scheduler.go
```diff
@@ -182,6 +215,16 @@ func (s *schedulerImpl) TaskChannelKeyFn() TaskChannelKeyFn {
 	return s.taskChannelKeyFn
 }
 
+// HandleBusyWorkflow implements BusyWorkflowHandler by delegating to the
+// underlying ExecutionAwareScheduler. This is called when a task encounters
+// a busy workflow error and needs to be routed to the sequential scheduler.
+func (s *schedulerImpl) HandleBusyWorkflow(executable Executable) bool {
+	if s.executionAwareScheduler == nil {
+		return false
+	}
```
yycptt: when will this be nil?

service/history/memory_scheduled_queue_factory.go
```diff
@@ -64,6 +64,7 @@ func NewMemoryScheduledQueueFactory(
 			WorkerCount: params.Config.MemoryTimerProcessorSchedulerWorkerCount,
 		},
 		logger,
+		nil, // No metrics for memory scheduled queue
```
yycptt: ditto

common/tasks/fifo_scheduler.go
```diff
@@ -212,10 +211,16 @@ func (f *FIFOScheduler[T]) executeTask(
...
+		if f.metricsHandler != nil {
+			metrics.FIFOSchedulerTasksFailed.With(f.metricsHandler).Record(1)
+		}
 		task.Nack(err)
 		return
 	}
 
+	if f.metricsHandler != nil {
+		metrics.FIFOSchedulerTasksCompleted.With(f.metricsHandler).Record(1)
+	}
```
yycptt: those two fifo scheduler metrics will be removed?

common/tasks/execution_queue_scheduler.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	isRetryable := func(err error) bool {
+		return !s.isStopped() && shouldRetry && task.IsRetryableError(err)
+	}
+
+	if err := backoff.ThrottleRetry(operation, task.RetryPolicy(), isRetryable); err != nil {
+		if s.isStopped() {
+			task.Abort()
+			metrics.ExecutionQueueSchedulerTasksAborted.With(s.metricsHandler).Record(1)
```
yycptt: I am not sure how useful the abort metric is. it's only related to the load and when host got shutdown.

common/tasks/execution_queue_scheduler.go
```diff
@@ -0,0 +1,362 @@
...
+
+	operation := func() (retErr error) {
+		var executePanic error
+		defer func() {
+			if executePanic != nil {
+				retErr = executePanic
+				shouldRetry = false
+			}
+		}()
+		defer log.CapturePanic(s.logger, &executePanic)
```
yycptt: I think task.Execute() already captures panic?

common/tasks/execution_queue_scheduler.go
```diff
@@ -0,0 +1,362 @@
...
+	metrics.ExecutionQueueSchedulerQueueCount.With(s.metricsHandler).Record(float64(len(s.queues)))
+	for range sweptCount {
+		s.queueAvailable.Signal()
+	}
+
+	if len(s.queues) == 0 {
+		s.sweeperRunning = false
+		return true
+	}
+	return false
```
yycptt: nit: 
```suggestion
	s.sweeperRunning = !(len(s.queues) == 0)
	return !s.sweeperRunning
```

common/tasks/execution_queue_scheduler.go
```diff
@@ -0,0 +1,362 @@
...
+
+// runWorker is the goroutine that processes tasks for a single execution queue.
+// It pops tasks from the slice under the lock and executes them without the lock.
+// When the queue is empty, the worker marks itself inactive and exits.
+// A new worker is spawned by Submit/TrySubmit when tasks arrive for an inactive queue.
+func (s *ExecutionQueueScheduler[T]) runWorker(q *executionQueue[T]) {
+	defer s.shutdownWG.Done()
+
+	for {
+		s.mu.Lock()
```
yycptt, line 183: shall we have one lock per Q?

common/tasks/execution_queue_scheduler.go
```diff
@@ -0,0 +1,362 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, exists := s.queues[key]
+	return exists
+}
+
+// Submit adds a task to its execution's queue. Creates a new queue if needed.
+// If MaxQueues is reached and the task needs a new queue, Submit blocks until
+// the sweeper frees a slot or the scheduler is stopped.
+func (s *ExecutionQueueScheduler[T]) Submit(task T) {
```
yycptt: wow the conditional variable for supporting this method really makes things a lot more complicated

prathyushpv: yeah I wonder if we really need this complexity since we don't really use Submit().

common/tasks/execution_aware_scheduler.go
prathyushpv: yeah that makes sense. Let me think that through..

common/tasks/execution_queue_scheduler.go
prathyushpv: We will anyways need one global lock. If we add one more per-queue lock, it will add some nested locking complexities. Critical section is very small here.

common/tasks/execution_aware_scheduler.go
prathyushpv: Visibility task queue has considerably high processing time. We will need higher per-queue concurrency to account for that. Still high visibility latency can slow down task processing in other queues. So I think it is better to keep it separate.

common/tasks/execution_queue_scheduler.go
prathyushpv: yes. I have refactored this part to avoid this double check.

common/tasks/fifo_scheduler_test.go
prathyushpv: done

service/history/queues/scheduler.go
prathyushpv: yeah it won't be nil. We initialize it in constructor. I have removed this check

service/history/memory_scheduled_queue_factory.go
prathyushpv: done

common/tasks/fifo_scheduler.go
prathyushpv: yeah. let me remove this.

common/tasks/execution_queue_scheduler.go
prathyushpv: yeah that is correct. I have removed it.

common/tasks/execution_queue_scheduler.go
prathyushpv: Yes! Good catch. Removed this code.

common/tasks/execution_queue_scheduler.go
prathyushpv: Done. Changed it to 
```
	s.sweeperRunning = len(s.queues) != 0
	return !s.sweeperRunning
```

common/tasks/execution_aware_scheduler.go
yycptt: ah true, make sense.

common/metrics/metric_defs.go
yycptt: Approved. let's remove it :)

common/tasks/execution_queue_scheduler.go
yycptt: yeah I am fine with not implementing it to keep things simple.

common/tasks/execution_queue_scheduler.go
yycptt: critical section is small but context switch could be expensive. I guess the use case I have in mind is that those per run tasks will come in as a burst and usually the worker will be the only consumer on it once all those tasks get enqueued.

With that side, I don't have very strong opinion on it, those tasks will be delayed anyway.

yycptt (approved):

common/tasks/execution_queue_scheduler.go
prathyushpv: yeah I removed submit method from this component.

PR #9143, opened 2026-01-26
Reduce TaskQueueStatsSuite partition count

stephanos:
## What changed?

Reduced number of partitions exercised in `TaskQueueStatsSuite` from 4 to 2.

And removed "single partition" tests where there's a multi-partition tests since they are superfluous. 

## Why?

Simplify.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/task_queue_stats_test.go
+++ b/tests/task_queue_stats_test.go
@@ -26,13 +26,6 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
-const (
-	minPriority     = 1
-	maxPriority     = 5
-	defaultPriority = 3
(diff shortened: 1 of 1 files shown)


tests/task_queue_stats_test.go
```diff
@@ -30,6 +30,7 @@ const (
 	minPriority     = 1
 	maxPriority     = 5
 	defaultPriority = 3
+	partitionCount  = 2 // kept low to reduce test time on CI
```
dnr: these names are at package level.. I don't think we want a const with this generic name in the whole tests package. put it in the suite?

stephanos: Fair point; I can do that.

PR #9145, opened 2026-01-27
Skip time in UpdateWorkflowSuite [WiP]

stephanos:
## What changed?

Replaced all "real time" `time` package uses with `clock.TimeSource` across History and Matching.

## Why?

To replace `time.Sleep`s with `s.Advance` in `UpdateWorkflowSuite` to skip time; one for History, one for Matching.

**Before**: TBD
**After**: TBD

## How did you test it?

Tests were run 10x locally without issues.

## Potential risks

### DIFF
--- a/service/history/service.go
+++ b/service/history/service.go
@@ -7,6 +7,7 @@ import (
 
 	"go.temporal.io/server/api/historyservice/v1"
 	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/common/clock"
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
 	"go.temporal.io/server/common/membership"
@@ -35,6 +36,7 @@ type (
 		healthServer      *health.Server
 		readinessCancel   context.CancelFunc
 		chasmRegistry     *chasm.Registry
+		timeSource        clock.TimeSource
 	}
 )
 
@@ -49,6 +51,7 @@ func NewService(
 	metricsHandler metrics.Handler,
 	healthServer *health.Server,
 	chasmRegistry *chasm.Registry,
+	timeSource clock.TimeSource,
 ) *Service {
 	return &Service{
 		server:            server,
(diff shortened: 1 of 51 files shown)


service/history/service.go
```diff
@@ -151,7 +155,7 @@ func (s *Service) Stop() {
 	s.handler.controller.Stop()
 
 	// All grpc handlers should be cancelled now. Give them a little time to return.
-	t := time.AfterFunc(2*time.Second, func() {
+	t := s.timeSource.AfterFunc(2*time.Second, func() {
```
dnr, line 158: This sort of thing is my concern with this idea: I would argue this one really should be 2 seconds of wall-clock time and not virtual time. It's in shutdown so doesn't really matter, but maybe others do and it's hard to audit them all (but we can try?)

service/matching/matching_engine.go
```diff
@@ -332,7 +333,7 @@ func (e *matchingEngineImpl) watchMembership() {
 			// queue bouncing back and forth due to long membership propagation time.
 			batch := partitions[i:min(len(partitions), i+batchSize)]
 			wait := backoff.Jitter(delay, 0.1)
-			time.AfterFunc(wait, func() {
+			e.timeSource.AfterFunc(wait, func() {
```
dnr, line 336: This one also (arguably) requires real time.. virtual time isn't threaded through membership, so ringpop propagation delays will not respect virtual time. Afaik we don't have any functional tests that exercise this functionality though.

In principle we could thread fake time into ringpop, it does support one (different interface though, we'd need an adapter)

service/history/service.go
stephanos: Yeah, I was just curious to see what happens if I try :)

service/matching/matching_engine.go
stephanos: That's interesting about ringpop support a fake time. But generally yes, this shouldn't be converted. My instructions were to replace everything because I preferred converting too many instead of too few to verify my test case.

PR #9147, opened 2026-01-27
Activate `Drained`/`Inactive` version if a workflow is "moved" to it

Shivs11:
## What changed?
- activate drained/inactive versions to draining when they get a workflow started on it 
- also added a history cache, per history node, so that we don't bombard our version workflows with signals that shall change the drainage status of these workflows.

## Why?
- versioning correctness, in the sense that if someone were to move a workflow on to a version that is drained, the drainage status should be updated to draining (since it now has one open workflow working on it)

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- Sure, this change is lowkey risky. Would appreciate a thorough review.

---

> [!NOTE]
> **Medium Risk**
> Touches history start/reset/signal-with-start/update-options code paths and introduces asynchronous signaling + caching; failures won’t block requests but could cause missed or excessive reactivation signals if misconfigured.
> 
> **Overview**
> Ensures worker-versioning *drained/inactive* versions get reactivated: when a workflow is **started, signal-with-started, reset with post-reset option updates, or has execution options updated** with a *pinned* `VersioningOverride`, the history service now fire-and-forgets a `reactivate-version` signal to the corresponding version workflow to move it back to *DRAINING*.
> 
> Adds a per-history-node `version_reactivation_signal` dedup cache (TTL/max-size + metrics) and a global kill-switch (`history.enableVersionReactivationSignals`) to avoid flooding version workflows; wires the cache + `workerdeployment.Client.SignalVersionReactivation` through history engine/fx and adds functional tests covering reactivation and cache dedup behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 06ad026150cd05de067665a2e20e53781fcf3be7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/xdc/chasm_test.go
+++ b/tests/xdc/chasm_test.go
@@ -1,0 +1,157 @@
+package xdc
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/suite"
+	commonpb "go.temporal.io/api/common/v1"
+	namespacepb "go.temporal.io/api/namespace/v1"
+	"go.temporal.io/api/serviceerror"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/server/api/adminservice/v1"
+	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/chasm/lib/tests"
+	"go.temporal.io/server/common/debug"
+	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/testing/testvars"
+	"go.temporal.io/server/tests/testcore"
+	"google.golang.org/protobuf/types/known/durationpb"
+)
+
+const (
+	chasmTestTimeout = 30 * time.Second * debug.TimeoutMultiplier
+)
+
+type ChasmSuite struct {
+	xdcBaseSuite
+
+	chasmContext context.Context
+}
+
+func TestChasmSuite(t *testing.T) {
+	t.Parallel()
+
+	s := &ChasmSuite{}
+	s.enableTransitionHistory = true
+	suite.Run(t, s)
+}
+
+func (s *ChasmSuite) SetupSuite() {
+	s.dynamicConfigOverrides = map[dynamicconfig.Key]any{
+		dynamicconfig.EnableChasm.Key():                 true,
+		dynamicconfig.NamespaceMinRetentionGlobal.Key(): 1 * time.Second,
+	}
+	s.setupSuite()
+}
+
+func (s *ChasmSuite) SetupTest() {
+	s.setupTest()
+
+	chasmEngine, err := s.clusters[0].Host().ChasmEngine()
+	s.Require().NoError(err)
+	s.Require().NotNil(chasmEngine)
+
+	chasmVisibilityMgr := s.clusters[0].Host().ChasmVisibilityManager()
+	s.Require().NotNil(chasmVisibilityMgr)
+
+	s.chasmContext = chasm.NewEngineContext(context.Background(), chasmEngine)
+	s.chasmContext = chasm.NewVisibilityManagerContext(s.chasmContext, chasmVisibilityMgr)
+}
+
+func (s *ChasmSuite) TearDownSuite() {
+	s.tearDownSuite()
+}
+
+func (s *ChasmSuite) TestRetentionTimer() {
+	nsName := s.createGlobalNamespace()
+
+	nsResp, err := s.clusters[0].FrontendClient().DescribeNamespace(testcore.NewContext(), &workflowservice.DescribeNamespaceRequest{
+		Namespace: nsName,
+	})
+	s.NoError(err)
+	nsID := nsResp.NamespaceInfo.GetId()
+
+	tv := testvars.New(s.T())
+	storeID := tv.Any().String()
+
+	ctx, cancel := context.WithTimeout(s.chasmContext, chasmTestTimeout)
+	defer cancel()
+
+	_, err = tests.NewPayloadStoreHandler(
+		ctx,
+		tests.NewPayloadStoreRequest{
+			NamespaceID:      namespace.ID(nsID),
+			StoreID:          storeID,
+			IDReusePolicy:    chasm.BusinessIDReusePolicyRejectDuplicate,
+			IDConflictPolicy: chasm.BusinessIDConflictPolicyFail,
+		},
+	)
+	s.NoError(err)
+
+	chasmRegistry := s.clusters[0].Host().GetCHASMRegistry()
+	archetypeID, ok := chasmRegistry.ComponentIDFor(&tests.PayloadStore{})
+	s.True(ok)
+	archetype, ok := chasmRegistry.ComponentFqnByID(archetypeID)
+	s.True(ok)
+
+	describeExecutionRequest := &adminservice.DescribeMutableStateRequest{
+		Namespace: nsName,
+		Execution: &commonpb.WorkflowExecution{
+			WorkflowId: storeID,
+		},
+		Archetype: archetype,
+	}
+	_, err = s.clusters[0].AdminClient().DescribeMutableState(testcore.NewContext(), describeExecutionRequest)
+	s.NoError(err)
+
+	s.Eventually(func() bool {
+		// Wait for it to be replicated to the standby cluster
+		_, err = s.clusters[1].AdminClient().DescribeMutableState(testcore.NewContext(), describeExecutionRequest)
+		return err == nil
+	}, 10*time.Second, 100*time.Millisecond)
+
+	// Reduce namespace retention to trigger deletion
+	retention := 3 * time.Second
+	_, err = s.clusters[0].FrontendClient().UpdateNamespace(testcore.NewContext(), &workflowservice.UpdateNamespaceRequest{
+		Namespace: nsName,
+		Config: &namespacepb.NamespaceConfig{
+			WorkflowExecutionRetentionTtl: durationpb.New(retention),
+		},
+	})
+	s.NoError(err)
+
+	// Wait for ns update to be replicated
+	s.Eventually(func() bool {
+		// Wait for it to be replicated to the standby cluster
+		resp, err := s.clusters[1].FrontendClient().DescribeNamespace(testcore.NewContext(), &workflowservice.DescribeNamespaceRequest{
+			Namespace: nsName,
+		})
+		s.NoError(err)
+		return resp.GetConfig().GetWorkflowExecutionRetentionTtl().AsDuration() == retention
+	}, 10*time.Second, 100*time.Millisecond)
+
+	// Wait for ns registry refresh
+	time.Sleep(2 * testcore.NamespaceCacheRefreshInterval) //nolint:forbidigo
+
+	// Close the execution and validate it's deleted on both clusters.
+	_, err = tests.ClosePayloadStoreHandler(
+		ctx,
+		tests.ClosePayloadStoreRequest{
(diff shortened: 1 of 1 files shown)


common/worker_versioning/worker_versioning.go
```diff
@@ -60,6 +60,22 @@ const (
 	WorkerDeploymentBuildIDFieldName             = "BuildID"
 )
 
+// GenerateDeploymentWorkflowID is a helper that generates a system accepted
+// workflowID which are used in our Worker Deployment workflows
+func GenerateDeploymentWorkflowID(deploymentName string) string {
```
Shivs11: just minor refactoring

service/history/api/worker_versioning_util.go
```diff
@@ -0,0 +1,93 @@
...
+	nsEntry, err := shardContext.GetNamespaceRegistry().GetNamespaceByID(namespaceID)
+	if err != nil {
+		shardContext.GetLogger().Warn("Failed to get namespace for version workflow signal",
+			tag.WorkflowNamespaceID(namespaceID.String()),
+			tag.Error(err))
+		return
+	}
+
+	// Generate the workflow ID for the version workflow
+	workflowID := worker_versioning.GenerateVersionWorkflowID(pinnedVersion.GetDeploymentName(), pinnedVersion.GetBuildId())
```
ShahabT: the actual signal mechanism should go under the worker deployment client because it's implementation-dependent. History service should not care if the signal in implemented via a workflow signal or say a CHASM rpc later.

service/history/api/worker_versioning_util.go
```diff
@@ -0,0 +1,93 @@
...
+
+// ReactivateVersionSignalName is the signal name for reactivating a version workflow.
+// This duplicates the constant from workerdeployment package to avoid import cycles.
+const ReactivateVersionSignalName = "reactivate-version"
+
+// ReactivateVersionWorkflowIfPinned sends a reactivation signal to the version workflow
+// when workflows are pinned to a potentially DRAINED/INACTIVE version. It also deduplicates
+// signals within the cache TTL window.
+// This is a fire-and-forget operation - errors are logged but don't fail the calling operation.
+func ReactivateVersionWorkflowIfPinned(
```
ShahabT, line 27: for now, can we have a dynamic config to disable sending these signals altogether if we need to please? I think we should have a way to turn it off in prod in case the load to the workflows becomes problematic.

carlydf: was just thinking this!

service/history/api/updateworkflowoptions/api.go
```diff
@@ -94,6 +96,9 @@ func Invoke(
 				}, nil
 			}
 
+			// Notify version workflow if we're pinning to a potentially drained version
+			api.ReactivateVersionWorkflowIfPinned(ctx, shardCtx, ns.ID(), mergedOpts.GetVersioningOverride(), reactivationSignalCache, reactivationSignaler, shardCtx.GetConfig().EnableVersionReactivationSignals())
```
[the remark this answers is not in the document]
Shivs11: oh gg, this is actually valid. making the change.

service/history/api/startworkflow/api.go
```diff
@@ -214,6 +220,9 @@ func (s *Starter) Invoke(
 		return nil, StartErr, err
 	}
 
+	// Notify version workflow if we're pinning to a potentially drained version
+	api.ReactivateVersionWorkflowIfPinned(ctx, s.shardContext, s.namespace.ID(), s.request.StartRequest.GetVersioningOverride(), s.reactivationSignalCache, s.reactivationSignaler, s.shardContext.GetConfig().EnableVersionReactivationSignals())
```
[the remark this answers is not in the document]
Shivs11: yeah, this is a really good point which I missed earlier. sorry.

service/history/api/startworkflow/api.go
```diff
@@ -215,8 +215,10 @@ func (s *Starter) Invoke(
 		if errors.As(err, &currentWorkflowConditionFailedError) && len(currentWorkflowConditionFailedError.RunID) > 0 {
 			// The history and mutable state generated above will be deleted by a background process.
 			resp, outcome, conflictErr := s.handleConflict(ctx, creationParams, currentWorkflowConditionFailedError)
-			if conflictErr == nil {
+			if conflictErr == nil && outcome == StartNew {
```
Shivs11, line 218: the main idea here was to only send the reactivation signal iff we have actually started/created a new workflow execution - all other outcomes, except StartNew, do not do that.

common/worker_versioning/worker_versioning.go
ShahabT: we don't need to move these anymore, right? And I dont think we should because it leaks the workflow implementation.

service/worker/workerdeployment/version_workflow.go
```diff
@@ -132,6 +132,38 @@ func (d *VersionWorkflowRunner) listenToSignals(ctx workflow.Context) {
...
+					LastChangedTime: timestamppb.New(workflow.Now(ctx)),
+					LastCheckedTime: timestamppb.New(workflow.Now(ctx)),
+				}
+
+				d.deleteVersion = false // Clear deletion flag if set
+
+				// Use existing function to update status and sync to task queues
+				d.updateVersionStatusAfterDrainageStatusChange(ctx, enumspb.VERSION_DRAINAGE_STATUS_DRAINING)
+				d.setStateChanged() // Trigger CaN
+			}
```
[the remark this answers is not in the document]
ShahabT: this seems valid, shouldn't we call syncSummary?

service/history/api/worker_versioning_util.go
```diff
@@ -0,0 +1,76 @@
...
+) error
+
+// ReactivateVersionWorkflowIfPinned sends a reactivation signal to the version workflow
+// when workflows are pinned to a potentially DRAINED/INACTIVE version. It also deduplicates
+// signals within the cache TTL window.
+// This is a fire-and-forget operation - errors are logged but don't fail the calling operation.
+func ReactivateVersionWorkflowIfPinned(
+	ctx context.Context,
+	shardContext historyi.ShardContext,
+	namespaceID namespace.ID,
```
ShahabT: I think most places who call this func already have the namespace entry. we should just get that instead of these two args.

service/history/api/worker_versioning_util.go
```diff
@@ -0,0 +1,76 @@
...
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/namespace"
+	"go.temporal.io/server/common/worker_versioning"
+	historyi "go.temporal.io/server/service/history/interfaces"
+)
+
+// VersionReactivationSignalerFn is a function type for sending reactivation signals to version workflows.
+// This abstraction allows the history API layer to use the deployment client without importing it directly,
+// avoiding import cycles between history/api and worker/workerdeployment packages.
+type VersionReactivationSignalerFn func(
```
ShahabT, line 14: any reason for this type instead of passing client itself?

service/history/api/worker_versioning_util.go
```diff
@@ -0,0 +1,76 @@
...
+	nsEntry, err := shardContext.GetNamespaceRegistry().GetNamespaceByID(namespaceID)
+	if err != nil {
+		shardContext.GetLogger().Warn("Failed to get namespace for version workflow signal",
+			tag.WorkflowNamespaceID(namespaceID.String()),
+			tag.Error(err))
+		return
+	}
+
+	// Use signaler function to send the signal
+	err = signaler(ctx, nsEntry, pinnedVersion.GetDeploymentName(), pinnedVersion.GetBuildId())
```
ShahabT: let' make sure this call is async so we don't add latency to user's request.

service/history/api/worker_versioning_util.go
Shivs11: circular dependencies; my first stab at this involved me just directly passing the workerDeployment.Client, but that did not go well as:

service/history/api → service/worker/workerdeployment → service/history/api

service/history/api/worker_versioning_util.go
Shivs11: hmmm, you are right. Noticed that SignalVersionReactivation on the client already logs all errors via convertAndRecordError - so the Warn log in ReactivateVersionWorkflowIfPinned is redundant.

common/worker_versioning/worker_versioning.go
Shivs11: yeah, given that I just removed the tag which was being used in the error, this goes back. sorry and thanks

service/history/api/worker_versioning_util.go
```diff
@@ -49,28 +47,16 @@ func ReactivateVersionWorkflowIfPinned(
...
-		shardContext.GetLogger().Warn("Failed to signal version workflow for reactivation",
-			tag.WorkflowNamespace(nsEntry.Name().String()),
-			tag.WorkflowID(worker_versioning.GenerateVersionWorkflowID(pinnedVersion.GetDeploymentName(), pinnedVersion.GetBuildId())),
-			tag.Error(err))
-	}
+	// Send the signal asynchronously to avoid adding latency to the caller's request.
+	// Errors are logged by the signaler implementation (e.g. via convertAndRecordError).
+	go func() {
+		signaler(context.Background(), namespaceEntry, pinnedVersion.GetDeploymentName(), pinnedVersion.GetBuildId())
+	}()
```
Shivs11, line 64: a thing to remember here is that any error, if present, would be caught in the client's recordAndError method and would be logged out. However, I am not returning the error over here since I don't think we should be stopping existing operations based on this error

common/dynamicconfig/constants.go
```diff
@@ -2789,6 +2789,27 @@ instead of the previous HSM backed implementation.`,
 		`Maximum number of entries in the version membership cache.`,
 	)
 
+	ReactivationSignalCacheTTL = NewGlobalDurationSetting(
+		"history.reactivationSignalCacheTTL",
```
carlydf: I know it's super annoying to wait for CI, but do you think it's worth renaming this to `history.versionReactivationSignalCacheTTL` and the other one to `history.versionReactivationSignalCacheMaxSize`?

I think it may be impossible to change the name of a dynamic config later, and I do feel like using the full "version reactivation cache" name is more clear to people who are not in versioning crew who might need to find and understand this config.

But also open to the idea that it's not worth changing, or even that the shorter name is better. If you're open to the longer name, I'm happy to make the change!

common/metrics/metric_defs.go
```diff
@@ -458,6 +459,8 @@ const (
 	VersionMembershipCacheGetScope = "VersionMembershipCacheGet"
 	// VersionMembershipCachePutScope is the scope used by version membership cache
 	VersionMembershipCachePutScope = "VersionMembershipCachePut"
+	// ReactivationSignalCacheShouldSendScope is the scope used by reactivation signal cache
+	ReactivationSignalCacheShouldSendScope = "ReactivationSignalCacheShouldSend"
```
carlydf: same for these, maybe "version" prefix would make the `reactivation_signal` metric tag more discoverable, but also open to not including it (or I could do a follow up PR right after this goes in, so at least we know that this is in)

service/history/api/updateworkflowoptions/api.go
```diff
@@ -107,6 +115,11 @@ func Invoke(
 	if err != nil {
 		return nil, err
 	}
+
+	// Notify version workflow if we're pinning to a potentially drained version.
+	// This is done after successful persistence to avoid signaling if the update fails.
```
carlydf, line 120: nice :)

PR #9148, opened 2026-01-27
Fix missing RUnlock on RWMutex to prevent lock leak

girichinna27:
Hi Maintainers 👋,

This Pull Request addresses a Semgrep concurrency-related finding caused by a missing unlock operation on a read-write mutex, which could lead to deadlocks or resource contention during runtime.

🔍 Issue Details

Rule ID: missing-runlock-on-rwmutex

Severity: Medium

Rule Message:
Missing RUnlock on an RWMutex (pm.versionedQueuesLock) before returning from a function.

📍 Affected Location

File Path:
/tools/scanResult/unzipped-3492193835/service/matching/task_queue_partition_manager.go

Line: 883

✅ Fix Applied

Ensured that the read lock acquired on pm.versionedQueuesLock is properly released before the function returns, preventing potential lock leaks and improving concurrency safety.

🎯 Impact

This change prevents possible deadlocks and performance degradation by ensuring correct lock lifecycle management when using RWMutex read locks.

The issue was identified and remediated using AI-Guardian, a security analysis tool developed by my company OpsMx.

Thanks for your time and review 🙏

### DIFF
--- a/common/cache/lru.go
+++ b/common/cache/lru.go
@@ -108,12 +108,19 @@ func (it *iteratorImpl) prepareNext() {
 // to the map during iteration can cause a dead lock.
 func (c *lru) Iterator() Iterator {
 	c.mut.Lock()
+	var lockTransferred bool
+	defer func() {
+		if !lockTransferred {
+			c.mut.Unlock()
+		}
+	}()
 	iterator := &iteratorImpl{
 		lru:        c,
 		createTime: c.timeSource.Now().UTC(),
 		nextItem:   c.byAccess.Front(),
 	}
 	iterator.prepareNext()
+	lockTransferred = true
 	return iterator
 }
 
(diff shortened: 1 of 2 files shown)


common/cache/lru.go
```diff
@@ -108,12 +108,19 @@ func (it *iteratorImpl) prepareNext() {
 // to the map during iteration can cause a dead lock.
 func (c *lru) Iterator() Iterator {
 	c.mut.Lock()
+	var lockTransferred bool
+	defer func() {
+		if !lockTransferred {
+			c.mut.Unlock()
+		}
+	}()
```
dnr, line 116: This change isn't described in the PR description. Also it seems useless, the lock is always "transferred". Maybe the function comment should be improved though.

service/matching/task_queue_partition_manager.go
```diff
@@ -880,6 +880,7 @@ func (pm *taskQueuePartitionManagerImpl) Describe(
 		if b == "" {
 			dbq := pm.defaultQueue()
 			if dbq == nil {
+				pm.versionedQueuesLock.RUnlock()
```
dnr, line 883: this should never happen here (it would be a softassert failure), but yeah, it should unlock just for consistency

PR #9162, opened 2026-01-28
Print goroutine profile in CI

stephanos:
## What changed?

1. Extended existing Go test memory monitor to include goroutine profile.
2. Changed snapshot to be of moment with _highest_ memory usage (instead of latest).
3. Unified report into a single one (both printing and disk snapshot).
4. (bonus) added monitor to unit and integration test jobs.

## Why?

Inspect where high goroutine count comes from.

## How did you test it?

Example: https://github.com/temporalio/temporal/actions/runs/21695994550/job/62566409323?pr=9162#step:10:17

### DIFF
--- a/develop/github/memory_monitor.sh
+++ b/develop/github/memory_monitor.sh
@@ -20,6 +20,7 @@ HISTORY_FILE="/tmp/memory_history.txt"
 HIGH_MEMORY_THRESHOLD=95
 PPROF_HOST="${PPROF_HOST:-localhost:7000}"
 HEAP_PRINTED=false
+HIGH_WATER_MARK=0
 
 # Clear history on start
 : > "$HISTORY_FILE"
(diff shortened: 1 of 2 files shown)


develop/github/memory_monitor.sh
```diff
@@ -86,8 +90,11 @@ snapshot() {
   memused_mb=$(( memused_kb / 1024 ))
   pct=$(( memused_kb * 100 / memtotal_kb ))
 
-  local goroutines
-  goroutines="$(print_goroutine_count)"
+  # Collect pprof data once per tick.
+  local goroutine_output goroutine_count pprof_output
+  goroutine_output="$(print_goroutines)"
```
dnr, line 95: do you know the max size of a bash variable? this can be pretty big and I'm wondering if it should be a temp file instead.

develop/github/memory_monitor.sh
```diff
@@ -86,8 +90,11 @@ snapshot() {
...
+  goroutine_count="$(count_goroutines <<< "$goroutine_output")"
```
dnr, line 96: this is the total number of goroutines, or the total created by the top 30 functions?

develop/github/memory_monitor.sh
stephanos: I didn't realize there would be a limit other than the OS' memory.

From the [example output](https://github.com/temporalio/temporal/actions/runs/21695994550/job/62566409323?pr=9162#step:10:176) I see, it's not much longer than the other outputs as it's capped at 30 as well.

develop/github/memory_monitor.sh
stephanos: That would be the total; I'm seeing 
```
Showing nodes accounting for 61014, 100% of 61015 total
```
in the [example output](https://github.com/temporalio/temporal/actions/runs/21695994550/job/62566409323?pr=9162#step:10:176).

develop/github/memory_monitor.sh
dnr: ah okay, I was thinking of the full ?debug=1 output

develop/github/memory_monitor.sh
dnr: cool, I wasn't sure if the top 30 was going to mess it up

PR #9164, opened 2026-01-28
Complete task queue draining

dnr:
## What changed?
When migrating from v1 to v2 task queue persistence, when the "draining" queue is empty, unload it and mark it in the active queue's metadata so that it won't be reloaded in the future.

## Why?
There's no need to reload drained queues anymore.

## How did you test it?
- [x] added new unit test(s)

## Potential risks
Incorrect identification of empty queues

### DIFF
--- a/service/matching/physical_task_queue_manager_test.go
+++ b/service/matching/physical_task_queue_manager_test.go
@@ -573,7 +573,6 @@ func TestDrainCompletionNoReloadDraining(t *testing.T) {
 
 	// unload
 	prtnMgr.Stop(unloadCauseUnspecified)
-	// FIXME tqMgr.Stop(unloadCauseUnspecified)
 
 	// record current counts
 	prevPriStats, prevFairStats := priQueueData.persistenceStats(), fairQueueData.persistenceStats()


service/matching/physical_task_queue_manager.go
```diff
@@ -328,14 +329,21 @@ func (c *physicalTaskQueueManagerImpl) Stop(unloadCause unloadCause) {
 	c.partitionMgr.engine.updatePhysicalTaskQueueGauge(c.partitionMgr.ns, c.partitionMgr.partition, c.queue.version, -1)
 }
 
+// getDrainBacklogMgr returns the draining backlog manager, or nil if none.
+func (c *physicalTaskQueueManagerImpl) getDrainBacklogMgr() backlogManager {
+	c.drainBacklogMgrLock.Lock()
```
majiru, line 334: Only potential concern is that we release the lock but continue to use the reference to the object here. Is there any concern about that reference being used by both the second half of FinishedDraining() and one of the other functions using it simultaneously?

dnr: The mutex is just for the pointer itself. It used to be an atomic.Value, it's just that you can't set an atomic.Value back to nil, so now it has to be a regular value plus mutex.

By definition, if it's a draining backlog manager, there shouldn't be any writes on it. And once it's drained, there shouldn't be any more reads/completions either. And of course the value itself is safe for concurrent use. So I don't see any concerns here.

PR #9166, opened 2026-01-29
Improve multi-cursor reader stuck monitoring

yycptt:
## What changed?
- Only trigger reader stuck action when reader is actually lagging behind
- Add a shadow mode for reader stuck

## Why?
- Previous implementation is purely based on the # of read within one second, however when a shard is actively making progress and generating new tasks, it is normal for it to do hundreds of read per second and that should not trigger reader stuck action.
- The idea in this PR is that only split a second out when we are actually lagging in task loading AND we've already loaded that second too many time.
- Also adding a shadow mode to get confidence before actually enabling it.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
- Reader stuck is currently effectively disable in cloud and shadow mode is enabled by default.

### DIFF
--- a/service/history/queues/monitor.go
+++ b/service/history/queues/monitor.go
@@ -6,6 +6,9 @@ import (
 
 	"go.temporal.io/server/common/clock"
 	"go.temporal.io/server/common/dynamicconfig"
+	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
 	"go.temporal.io/server/service/history/tasks"
 )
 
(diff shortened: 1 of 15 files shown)


service/history/queues/monitor.go
```diff
@@ -176,16 +187,36 @@ func (m *monitorImpl) SetReaderWatermark(readerID int64, watermark tasks.Key) {
...
+	}
+
+	alert := &Alert{
+		AlertType: AlertTypeReaderStuck,
+		AlertAttributesReaderStuck: &AlertAttributesReaderStuck{
+			ReaderID:         readerID,
+			CurrentWatermark: stats.progress.watermark,
+		},
+	}
+	if m.options.ReaderStuckShadowMode() {
```
prathyushpv, line 212: can we add lag duration as well to the log line here?

prathyushpv (approved):

PR #9168, opened 2026-01-29
Add cache for task queue routing info in History

ShahabT:
## What changed?
Cache the result of `GetTaskQueueUserData` that history makes to Matching when an activity wants to start a deployment version transition.

## Why?
This protects the matching root partition from being hammered by requests when a lot of AutoUpgrade workflows want to start activity-initiated transitions. Activity initiated transitions happen in one of the following cases:
1) Target version changed while activity was backlogged.
2) Target version changed while activity was in retry backoff
3) Target version changed in some edge cases involving parallel activities

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
The cache can potentially increase history mem usage but there are knobs to adjust size and ttl.

### DIFF
--- a/common/worker_versioning/routing_info_cache.go
+++ b/common/worker_versioning/routing_info_cache.go
@@ -13,20 +13,21 @@ import (
 //
 // Implementations are expected to be safe for concurrent use.
 type (
+	RoutingInfo struct {
+		Current               *deploymentspb.WorkerDeploymentVersion
+		CurrentRevisionNumber int64
+		Ramping               *deploymentspb.WorkerDeploymentVersion
+		RampPercentage        float32
+		RampingRevisionNumber int64
+	}
+
 	RoutingInfoCache interface {
 		// Get returns the cached routing info. ok=false means there was no cached value.
 		Get(
 			namespaceID string,
 			taskQueue string,
 			taskQueueType enumspb.TaskQueueType,
-		) (
-			current *deploymentspb.WorkerDeploymentVersion,
-			currentRevisionNumber int64,
-			ramping *deploymentspb.WorkerDeploymentVersion,
-			rampPercentage float32,
-			rampingRevisionNumber int64,
-			ok bool,
-		)
+		) (RoutingInfo, bool)
 
 		Put(
 			namespaceID string,
@@ -73,14 +74,7 @@ func (c *RoutingInfoCacheImpl) Get(
 	namespaceID string,
 	taskQueue string,
 	taskQueueType enumspb.TaskQueueType,
-) (
-	current *deploymentspb.WorkerDeploymentVersion,
-	currentRevisionNumber int64,
-	ramping *deploymentspb.WorkerDeploymentVersion,
-	rampPercentage float32,
-	rampingRevisionNumber int64,
-	ok bool,
-) {
+) (RoutingInfo, bool) {
 	handler := c.metricsHandler.WithTags(metrics.OperationTag(metrics.RoutingInfoCacheGetScope), metrics.NamespaceIDTag(namespaceID))
 	metrics.CacheRequests.With(handler).Record(1)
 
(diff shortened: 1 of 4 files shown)


service/history/workflow/mutable_state_impl.go
```diff
@@ -9055,6 +9055,13 @@ func (ms *MutableStateImpl) reschedulePendingActivities() error {
 		// need to update stamp so the passive side regenerate the task
 		err := ms.UpdateActivity(ai.ScheduledEventId, func(info *persistencespb.ActivityInfo, state historyi.MutableState) error {
 			info.Stamp++
+			// Updating scheduled time because we don't want the time since last scheduling until now
+			// be considered in the schedule-to-start latency calculation.
+			// This is specifically needed for the cases that the activity is blocked not because of a
+			// backlog but because of an ongoing transition. See ActivityStartDuringTransition error usage.
+			// TODO (shahab): can we limit this adjustment to apply only for activities who actually faced the
+			// ActivityStartDuringTransition error, and not all others?
+			info.ScheduledTime = timestamppb.New(ms.timeSource.Now())
```
ShahabT: @yycptt could you please review this line? what to ensure it does not breaks some assumptions in the system.

common/worker_versioning/routing_info_cache.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	enumspb "go.temporal.io/api/enums/v1"
+	deploymentspb "go.temporal.io/server/api/deployment/v1"
+	"go.temporal.io/server/common/cache"
+	"go.temporal.io/server/common/metrics"
+)
+
+// RoutingInfoCache is used to cache results of GetTaskQueueUserData
+// calls followed by CalculateTaskQueueVersioningInfo computation.
```
Shivs11, line 12: nit: we can just say that this cache is used to cache results of GetTaskQueueUserData operations since that is the main RPC that is actually intensive.

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -277,32 +280,43 @@ func recordActivityTaskStarted(
...
-			NamespaceId:   namespaceID,
-			TaskQueue:     taskQueueName,
-			TaskQueueType: taskQueueType,
-		})
-	if err != nil {
-		return nil, 0, err
-	}
-	tqData, ok := resp.GetUserData().GetData().GetPerType()[int32(taskQueueType)]
+	// Check cache first for task queue routing info (independent of workflow ID)
+	current, currentRevisionNumber, ramping, rampingPercentage, rampingRevisionNumber, ok := routingInfoCache.Get(namespaceID, taskQueueName, taskQueueType)
```
Shivs11: something to remember about this change - I know that the cache TTL is only one second which is quite small, but during that one second, there could in theory be stale task queue user data reads which could make AU workflow/activity tasks bounce back and forth.

I think this is fine, just wanted to paste this here.

service/history/workflow/mutable_state_impl.go
Shivs11: sorry to be pendantic but I wonder if this should come in a different PR which has some tests that test that there the time delay has been present

tests/versioning_3_test.go
```diff
@@ -5305,6 +5305,147 @@ func (s *Versioning3Suite) TestMaxVersionsInTaskQueue() {
 		"Error message should include the limit value")
 }
 
+func (s *Versioning3Suite) TestActivityRetryAutoUpgradeDuringBackoff() {
```
Shivs11, line 5308: should we also have some functional tests that test the actual cache implementation that is now present?
I had placed some tests in this file where I increased the TTL and noticed if the calls are going through or not.

Shivs11: sure, I see you have placed unit tests for the cache and since we are in a time crunch, you can ignore my top comment.

tests/versioning_3_test.go
```diff
@@ -5305,6 +5305,147 @@ func (s *Versioning3Suite) TestMaxVersionsInTaskQueue() {
...
+		return "v2", nil
+	}
+
+	// Workflow that executes an activity with retry policy
+	wf := func(ctx workflow.Context) (string, error) {
+		var result string
+		err := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+			StartToCloseTimeout: 10 * time.Second,
+			RetryPolicy: &temporal.RetryPolicy{
+				InitialInterval:    3 * time.Second, // Give us time to change deployment
```
Shivs11: could this be flaky? 3 seconds seems less....like what happens if we change the deployment before that?

ShahabT: changed it to 5s

service/history/workflow/mutable_state_impl.go
ShahabT: It's a fair comment. I did try to spend some time writing the test but this part is soooo internal that I could not come up with a good way to write a non-flaky test. I did though test it manually and verified the delay existed before and my change is canceling it.

PR #9172, opened 2026-01-29
Fix ExceededPageSizeIsCapped to await results indexed

fretz12:
## What changed?
Fix ExceededPageSizeIsCapped to await results indexed

## Why?
Previously, we were not correctly awaiting all expected results to be index. By using the count API in the eventually loop instead, we should be guaranteed the right number of results before proceeding

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/tests/standalone_activity_test.go
+++ b/tests/standalone_activity_test.go
@@ -853,7 +853,7 @@ func (s *standaloneActivityTestSuite) TestFailed() {
 	})
 }
 
-func (s *standaloneActivityTestSuite) TestCancelled() {
+func (s *standaloneActivityTestSuite) TestCancellation() {
 	t := s.T()
 	ctx, cancel := context.WithTimeout(t.Context(), 10*time.Second)
 	defer cancel()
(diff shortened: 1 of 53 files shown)


tests/standalone_activity_test.go
```diff
@@ -2504,25 +2504,32 @@ func (s *standaloneActivityTestSuite) TestListActivityExecutions() {
...
+		})
+		require.NoError(t, err)
 		require.Len(t, resp.GetExecutions(), 1)
 
-		// Get next page. Use pageSize > FrontendVisibilityMaxPageSize
-		resp, err := s.FrontendClient().ListActivityExecutions(ctx, &workflowservice.ListActivityExecutionsRequest{
+		// Get next page. Use pageSize > FrontendVisibilityMaxPageSize to test it is capped by the server
+		resp, err = s.FrontendClient().ListActivityExecutions(ctx, &workflowservice.ListActivityExecutionsRequest{
 			Namespace:     s.Namespace().String(),
 			PageSize:      2,
```
dandavison: These comments say

> Use pageSize > FrontendVisibilityMaxPageSize to test it is capped by the server

But then the code uses the value 2. 2 seems unlikely to be greater than any max page size. But even if it is, can we compute the value from FrontendVisibilityMaxPageSize?

fretz12: Max page size has been explicitly set to 1 in this test: 
```
		s.OverrideDynamicConfig(
			dynamicconfig.FrontendVisibilityMaxPageSize,
			1,
		)

dandavison: OK cool. Could we compute the new page size value from the dc value rather than hard-coding 2? But, not blocking.

fretz12: Changed to FrontendVisibilityMaxPageSize + 1 to make it more readable as we discussed

PR #9175, opened 2026-01-30
Add readiness checks to matching handler

yaninyzwitty:
## What changed?
- Replaced `sync.WaitGroup`–based startup coordination in the matching handler with an atomic readiness flag.
- Added a `checkReady` guard to prevent request-handling APIs from executing before the handler is fully started.
- Updated handler lifecycle (`Start` / `Stop`) to explicitly manage readiness state.
- Added unit tests to validate handler behavior before and after startup.

## Why?
The matching handler previously relied on a `WaitGroup` to coordinate startup, but exposing or extending this mechanism (for example by adding a `Wait()` method) risks blocking gRPC request threads and causing startup deadlocks.

Using a non-blocking atomic readiness flag makes handler state explicit, avoids halting request processing, and ensures requests fail fast with a clear `Unavailable` error until the handler is ready.

## How did you test it?
- [x] built
- [x ] covered by existing tests
- [ ] added new unit test(s)

## Potential risks
- Requests may now return `Unavailable` during a brief startup window where they were previously blocked; this behavior is intentional and safer than blocking gRPC threads.
- Only selected request-handling APIs are gated by readiness; describe and metadata APIs remain callable during startup.


dnr (changes requested):
It looks like startWG was never waited on, so the PR description is incorrect, there is no waiting and no deadlocks. Have you observed any actual problems?

startWG can definitely be removed, but we don't need any additional ready logic either: we don't register in membership until handler.Start returns, so we can't get any rpcs. And engine.Start is instantaneous anyway, so there's nothing to wait for.

yaninyzwitty:
@dnr you are right and so my original comment might be off somehow
startWG was not being waited on anywhere, so my original PR description overstated the risk of blocking or deadlocks. Thanks for pointing that out.
The primary motivation for this change for me was cleanup and clarity rather than fixing an observed issue.
startWG did not actually gate request handling, and its presence suggested startup coordination that wasn’t real. Removing it avoids that confusion.
While membership registration does occur after handler.Start returns today, the readiness check was added defensively to make the handler’s assumptions explicit and to guard against future changes where request routing might precede full initialization. That said, I agree that with the current lifecycle and instantaneous engine.Start, the additional readiness logic may be redundant.
If you prefer, I’m happy to drop the readiness checks and keep this PR focused solely on removing the unused startWG, or update the PR description to reflect that this is a cleanup with no functional change in behavior.

dnr:
Just removing startWG is fine if you want to do that.

PR #9176, opened 2026-01-30
Attempt to reduce flakeyness of multi-heartbeat test.

robholland:
## What changed?
Reduce the heartbeat timeout.

## Why?
It has been hard to replicate failures locally (and CI logs don't show the issue due to filtering) but my theory is that the test is timing out because the heartbeat timeout is too long, the sleep intended to cause a heartbeat timeout isn't long enough to ensure that reliably happens (although the failover delays sometimes also cause it to happen). The test expects a second activity attempt due to heartbeat failure on the first.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

---

> [!NOTE]
> **Low Risk**
> Test-only timing/logic adjustments; risk is limited to altering expectations and timing sensitivity of a single functional test.
> 
> **Overview**
> Reduces flakiness in `TestActivityMultipleHeartbeatsAcrossFailover` by shortening the activity `HeartbeatTimeout` (10s → 3s) and explicitly sleeping after failover to reliably force the first activity attempt to heartbeat-timeout.
> 
> Tightens the activity logic to abort on cancellation by checking `ctx.Err()` after post-failover heartbeats, removing the prior no-op heartbeat-details read.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 740974cd413bc26f88879cd0d2537661f9794d45. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/xdc/failover_test.go
```diff
@@ -2790,10 +2792,12 @@ func (s *FunctionalClustersWithRedirectionTestSuite) TestActivityMultipleHeartbe
 	}, 10*time.Second, 200*time.Millisecond)
 
 	s.failover(namespace, 0, s.clusters[1].ClusterName(), 2)
+	// sleep to trigger heartbeat timeout for first attempt
 	// nolint:forbidigo
 	time.Sleep(time.Second * 4)
```
yux0, line 2797: This should be less than heartbeat timeout?

tests/xdc/failover_test.go
```diff
@@ -2726,20 +2726,22 @@ func (s *FunctionalClustersWithRedirectionTestSuite) TestActivityMultipleHeartbe
...
 		activity.RecordHeartbeat(ctx, hb2Val)
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
 		select {
 		case hb2Ch <- struct{}{}:
 		default:
 		}
 		activity.RecordHeartbeat(ctx, hb3Val)
+		if ctx.Err() != nil {
```
yux0, line 2742: can it be:
`select {
case <-allowComplete:
return nil
case <-ctx.Done():
return ctx.Err()
}`

robholland: No, `allowComplete` would never get closed because it's only closed once `hb3ch` has been written to, and we must only write to that channel if the heartbeat has successfully been recorded.

tests/xdc/failover_test.go
robholland: The idea is to ensure a heatbeat timeout, and confirm that new attempt(s) will eventually make progress.

PR #9177, opened 2026-01-30
chore: update code to use new log aliases where applicable

tdeebswihart:
## What changed?

This updates the server code to use the shorthand log tag constructors introduced in #9174.

As part of this it _does_ make a breaking change to the `Bool` constructor: it now takes in the key as a string to be consistent.

The only inconsistent one is now `Error`, but that's used so heavily that changing it is likely not worth the time.

## Why?

Consistency!

## How did you test it?
Existing tests

## Potential risks
The only risk is that I _have_ introduced a breaking change to the Bool constructor. I'm happy to undo that if my reviewers desire: my goal is minimal breaking changes.

I'd prefer none, but I made this change to stir up discussion

### DIFF
--- a/client/matching/metric_client.go
+++ b/client/matching/metric_client.go
@@ -162,7 +162,7 @@ func (c *metricClient) emitForwardedSourceStats(
 		// it means some mangled name come here; need to check why
 		_, err := tqid.NewTaskQueueFamily("", taskQueue.GetName())
 		if err != nil {
-			c.logger.Info("invalid tq name", tag.Error(err), tag.Strings("proto", []string{taskQueue.GetName()}))
+			c.logger.Info("invalid tq name", tag.Error(err), tag.String("proto", taskQueue.GetName()))
 			metrics.MatchingClientInvalidTaskQueueName.With(metricsHandler).Record(1)
 		}
 	}


client/matching/metric_client.go
```diff
@@ -162,7 +162,7 @@ func (c *metricClient) emitForwardedSourceStats(
 		// it means some mangled name come here; need to check why
 		_, err := tqid.NewTaskQueueFamily("", taskQueue.GetName())
 		if err != nil {
-			c.logger.Info("invalid tq name", tag.Error(err), tag.NewStringsTag("proto", []string{taskQueue.GetName()}))
+			c.logger.Info("invalid tq name", tag.Error(err), tag.Strings("proto", []string{taskQueue.GetName()}))
```
dnr: drive-by cleanup:
```suggestion
			c.logger.Info("invalid tq name", tag.Error(err), tag.String("proto", taskQueue.GetName()))
```

service/history/workflow/mutable_state_impl.go
```diff
@@ -4166,7 +4166,7 @@ func (ms *MutableStateImpl) AddActivityTaskCompletedEvent(
 		ms.logger.Warn(mutableStateInvalidHistoryActionMsg, opTag,
 			tag.WorkflowEventID(ms.GetNextEventID()),
 			tag.ErrorTypeInvalidHistoryAction,
-			tag.Bool(ok),
+			tag.Bool("hasActivityInfo", ok),
```
dnr, line 4170: this is an improvement anyway!

tdeebswihart:
> Thanks!
> 
> One more idea that occurred to me, before it goes in:
> 
> Did we consider having a `tag.Int` that's generic for all integral types? The issue there is that we'd need to split them out again for zap, which is annoying, but it could be nice to have the nice interface on our side.

How about `tag.Number` instead? I'll throw floats in too.

It has the additional benefit that it won't affect the behavior of any existing callsites

EDIT: Go won't let you do that; it won't compile the type switch on the generic value inside the generic function.

I'm going to set this to merge without, and we can keep discussing further improvements as we go

PR #9180, opened 2026-01-30
WorkerDeployment Read API's to have their own Rate Limiters!

Shivs11:
## What changed?
- WISOTT, mainly
- Also changed the grouping of the ListWorkerDeployments API to the Visibility API's, which seem to have a limit of 10 RPS.

## Why?
- Moreover, we noticed that a lot of folks were calling read WorkerDeployment API's (like DescribeDeploymentVersion) at 200 RPS. This is.....quite a lot.
- Right now, I noticed that there was no real way to change the rate limits for these specific API's unless I added a new RL for these. So I went ahead and did that. 

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- This is my first time working on RL stuff; would love a thorough review given that I did use a lot of existing code as my template.

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -737,6 +737,39 @@ If this is set, it overwrites the per instance limit configured with
 "frontend.namespaceRPS.namespaceReplicationInducingAPIs".
 This config is EXPERIMENTAL and may be changed or removed in a later release.`,
 	)
+	FrontendWorkerDeploymentReadAPIsRPS = NewGlobalIntSetting(
+		"frontend.rps.workerDeploymentReadAPIs",
+		10,
+		`FrontendWorkerDeploymentReadAPIsRPS limits the per second request rate for worker deployment read APIs
+(e.g. DescribeWorkerDeploymentVersion, DescribeWorkerDeployment).
+This config is EXPERIMENTAL and may be changed or removed in a later release.`,
+	)
+	FrontendMaxNamespaceWorkerDeploymentReadAPIsRPSPerInstance = NewNamespaceIntSetting(
+		"frontend.namespaceRPS.workerDeploymentReadAPIs",
+		10,
+		`FrontendMaxNamespaceWorkerDeploymentReadAPIsRPSPerInstance is a per host/per namespace RPS limit for
+worker deployment read APIs (e.g. DescribeWorkerDeploymentVersion, DescribeWorkerDeployment).
+This config is EXPERIMENTAL and may be changed or removed in a later release.`,
+	)
+	FrontendMaxNamespaceWorkerDeploymentReadAPIsBurstRatioPerInstance = NewNamespaceFloatSetting(
+		"frontend.namespaceBurstRatio.workerDeploymentReadAPIs",
+		10,
+		`FrontendMaxNamespaceWorkerDeploymentReadAPIsBurstRatioPerInstance is a per host/per namespace burst limit for
+worker deployment read APIs (e.g. DescribeWorkerDeploymentVersion, DescribeWorkerDeployment)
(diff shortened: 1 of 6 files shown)


common/dynamicconfig/constants.go
```diff
@@ -735,6 +735,39 @@ namespace replication inducing APIs (e.g. RegisterNamespace, UpdateNamespace, Up
...
+	FrontendMaxNamespaceWorkerDeploymentReadAPIsBurstRatioPerInstance = NewNamespaceFloatSetting(
+		"frontend.namespaceBurstRatio.workerDeploymentReadAPIs",
+		10,
+		`FrontendMaxNamespaceWorkerDeploymentReadAPIsBurstRatioPerInstance is a per host/per namespace burst limit for
+worker deployment read APIs (e.g. DescribeWorkerDeploymentVersion, DescribeWorkerDeployment)
+as a ratio of namespace WorkerDeploymentReadAPIs RPS. The RPS used here will be the effective RPS from global and
+per-instance limits. This config is EXPERIMENTAL and may be changed or removed in a later release. The value must
+be 1 or higher.`,
+	)
+	FrontendGlobalNamespaceWorkerDeploymentReadAPIsRPS = NewNamespaceIntSetting(
```
dnr, line 763: can we have just one of the per-instance or global limits? which one do we plan to use? why not have just that one?

common/dynamicconfig/constants.go
```diff
@@ -735,6 +735,39 @@ namespace replication inducing APIs (e.g. RegisterNamespace, UpdateNamespace, Up
...
+This config is EXPERIMENTAL and may be changed or removed in a later release.`,
+	)
+	FrontendMaxNamespaceWorkerDeploymentReadAPIsRPSPerInstance = NewNamespaceIntSetting(
+		"frontend.namespaceRPS.workerDeploymentReadAPIs",
+		10,
+		`FrontendMaxNamespaceWorkerDeploymentReadAPIsRPSPerInstance is a per host/per namespace RPS limit for
+worker deployment read APIs (e.g. DescribeWorkerDeploymentVersion, DescribeWorkerDeployment).
+This config is EXPERIMENTAL and may be changed or removed in a later release.`,
+	)
+	FrontendMaxNamespaceWorkerDeploymentReadAPIsBurstRatioPerInstance = NewNamespaceFloatSetting(
```
dnr, line 754: do we really need this? let's just use a default burst ratio of 1?

dnr (commented):
I don't have an opinion on if this is needed at all (I wish it wasn't but maybe it is), but if it is, can we make it as simple as possible?

common/dynamicconfig/constants.go
Shivs11: what happens if you assign a request to an instance but the instance is currently overloaded?

ShahabT: I think we want some limit that users understand. like "100 rps per namespace". per instance does not makes sense for users because it depends on how many frontend nodes exist in the cluster.

+1 to davids point about keeping it simple. is just one global limit not enough? 

> what happens if you assign a request to an instance but the instance is currently overloaded?

note that this limit is not to protect frontend nodes, is to protect the workers. so this scenario is not a problem (the limit will not be high enough to overwhelm a single frontend node)

dnr: by "assign a request to an instance" you mean a request gets load-balanced to an instance? if it hits a rate limit then it's rejected and the client should retry

Shivs11: I think my biggest concern was that the global RL does not have a view into how overloaded an instance is. I think im okay having a global and an instance one and removing the per-ns if that simply gets overriden all the time (which it seems like it does) when the global one is set

Shivs11:
gonna close this one out since I have decided to embed this in our frontend directly.

PR #9190, opened 2026-02-01
Scopable InjectHook

stephanos:
## What changed?

Make test hooks scopable to namespace.

## Why?

Right now using `InjectHook` requires a dedicated cluster, but most hooks can be scoped to a namespace and allow using a shared cluster. This reduces resource consumption and increases test concurrency.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


client/matching/loadbalancer.go
```diff
@@ -76,7 +76,7 @@ func NewLoadBalancer(
 func (lb *defaultLoadBalancer) PickWritePartition(
 	taskQueue *tqid.TaskQueue,
 ) *tqid.NormalPartition {
-	if n, ok := testhooks.Get[int](lb.testHooks, testhooks.MatchingLBForceWritePartition); ok {
+	if n, ok := testhooks.Get(lb.testHooks, testhooks.MatchingLBForceWritePartition, testhooks.NamespaceID(taskQueue.NamespaceId())); ok {
```
stephanos: Using explicit type `testhooks.NamespaceID` to communicate type requirement clearly.

tests/testcore/functional_test_base.go
```diff
@@ -602,11 +602,23 @@ func (s *FunctionalTestBase) OverrideDynamicConfig(setting dynamicconfig.Generic
 	return s.testCluster.host.overrideDynamicConfig(s.T(), setting.Key(), value)
 }
 
-func (s *FunctionalTestBase) InjectHook(key testhooks.Key, value any) (cleanup func()) {
-	if s.isShared {
-		s.T().Fatalf("InjectHook cannot be called on a shared cluster; use testcore.WithDedicatedCluster()")
+// InjectHook sets a test hook. It auto-detects the scope from the hook:
+// - For namespace-scoped hooks: automatically scopes to the test's namespace (safe for shared clusters)
+// - For global hooks: requires a dedicated cluster
+func (s *FunctionalTestBase) InjectHook(hook testhooks.Hookable) (cleanup func()) {
```
stephanos: This won't scale to support scoping to task queues (as it wouldn't know which one); but that's fine. Maybe we'll never need it. And if we do, this could maybe take an optional scope parameter.

tests/update_workflow_test.go
```diff
@@ -5553,18 +5553,17 @@ func TestWorkflowUpdateSuite(t *testing.T) {
...
-				s := testcore.NewEnv(t, testcore.WithDedicatedCluster())
+				s := testcore.NewEnv(t)
 				startReq := startWorkflowReq(s, s.Tv())
 				startReq.WorkflowIdConflictPolicy = enumspb.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING
 				updateReq := updateWorkflowRequest(s, s.Tv(),
 					&updatepb.WaitPolicy{LifecycleStage: enumspb.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED})
 
 				// simulate a race condition
-				s.InjectHook(testhooks.UpdateWithStartInBetweenLockAndStart, func() {
+				s.InjectHook(testhooks.NewHook(testhooks.UpdateWithStartInBetweenLockAndStart, func() {
```
stephanos, line 5566: Using `testhooks.NewHook` since `InjectHook` is method and Go doesn't allow type parameters there. This way it's a bit more verbose but preserves type checking.

common/testing/testhooks/test_impl.go
```diff
@@ -19,40 +19,68 @@ var Module = fx.Options(
...
 // Get gets the value of a test hook from the registry.
 //
 // TestHooks should be used sparingly, see comment on TestHooks.
-func Get[T any](th TestHooks, key Key) (T, bool) {
+func Get[T any, S any](th TestHooks, key Key[T, S], scope S) (T, bool) {
 	var zero T
-	if th.data == nil {
+	if th.lock == nil {
+		// This means TestHooks wasn't created via NewTestHooks. Ignore.
+		return zero, false
```
stephanos, line 43: This happened in `service/worker/workerdeployment/workflow_test.go`.

common/testing/testhooks/hooks.go
```diff
@@ -1,14 +1,86 @@
...
+	"go.temporal.io/server/common/namespace"
+)
+
+// keyCounter provides unique IDs for keys.
+var keyCounter atomic.Int64
+
+// Test hook keys with their return type and scope.
+// Try to avoid global scope as it requires a dedicated test cluster.
+var (
+	MatchingDisableSyncMatch                 = newKey[bool, namespace.ID]()
```
dnr, line 12: It's unfortunate to do a bunch of stuff at static init time just for tests (of course much better than doing it at runtime). What if we did something like define newKey differently in test/prod and have the prod version just return a struct{}{}?

common/testing/testhooks/key.go
```diff
@@ -1,14 +1,86 @@
...
+type hook[T any, S any] struct {
+	key   Key[T, S]
+	value T
+}
+
+func (h hook[T, S]) Scope() Scope {
+	return h.key.scope
+}
+
+func (h hook[T, S]) Apply(th TestHooks, scope any) func() {
```
dnr: why the new interface here? we can't roll this check into Set?

common/testing/testhooks/test_impl.go
```diff
@@ -19,40 +19,68 @@ var Module = fx.Options(
 // concerns into production code. In general you should prefer other ways of writing tests
 // wherever possible, and only use TestHooks sparingly, as a last resort.
 type TestHooks struct {
-	data *sync.Map
+	lock  *sync.Mutex
+	hooks map[keyID]map[any]any // id -> scope -> value
```
dnr: did you consider a single level map with a struct key? (then it could still be a sync.Map which might simplify the code?)

common/testing/testhooks/test_impl.go
```diff
@@ -19,40 +19,68 @@ var Module = fx.Options(
...
-	return func() { th.data.Delete(key) }
+func Set[T any, S any](th TestHooks, key Key[T, S], val T, scope S) func() {
+	th.lock.Lock()
+	defer th.lock.Unlock()
+
+	scopes := th.hooks[key.id]
+	if scopes == nil {
+		scopes = make(map[any]any)
+		th.hooks[key.id] = scopes
+	}
```
dnr:
```suggestion
	scopes := util.GetOrSetMap(th.hooks, key.id)
```

tests/update_workflow_test.go
dnr: oh, hmm. they are relaxing this constraint, but I guess it'll be in go 1.27

stephanos: oh that'd be really neat

stephanos:
👍 Thank you for the feedback; it's shorter and cleaner now I think.

common/testing/testhooks/hooks.go
```diff
@@ -0,0 +1,63 @@
...
+		apply: func(th TestHooks, scope any) func() {
+			if _, ok := scope.(S); !ok {
+				panic("testhooks: scope type mismatch")
+			}
+			return Set(th, key, value, scope)
+		},
+	}
+}
+
+type Key[T any, S any] struct {
```
dnr, line 39: can you move this into test_impl/noop_impl and have it be defined as `struct{}` in noop_impl?

yeah, it works fine. I had to move Hook+NewHook also, and defined it as `struct{}` also. and defined Hook.Scope/Apply to do the same panic as Set()

common/testing/testhooks/hooks.go
```diff
@@ -0,0 +1,63 @@
...
+	id    keyID
+	scope Scope
```
dnr: I got a little confused at first between the difference between Key and hookKey because they both have a "scope" but the meaning of the word is different. maybe:

```suggestion
	scopeType ScopeType
```

dnr (approved):

PR #9195, opened 2026-02-02
Add handler-level rate limiting for worker deployment read APIs + move ListWorkerDeployments to Visibility group 

Shivs11:
## What changed?
Adds per-namespace rate limiting directly in WorkflowHandler for DescribeWorkerDeployment and DescribeWorkerDeploymentVersion APIs.

This is an alternative approach to interceptor-level rate limiting, implementing the check at the handler level after the request passes through the existing interceptor chain. The reason for adding this rate limiter at the handler level is purely because adding it on the frontend is a lot of code. 

Dynamic config keys:
- frontend.workerDeploymentReadRPS (default: 50)
- frontend.workerDeploymentReadBurstRatio (default: 2.0)

Additionally, move the ListWorkerDeployments API to the visibility grouping since it relies on visibility. This was wrongly *not present* in this group prior to this PR.

## Why?
- Rate limiting.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- idts

---

> [!NOTE]
> **Medium Risk**
> Touches frontend request throttling and API prioritization; misconfiguration or limiter logic errors could cause unexpected `ResourceExhausted` responses or alter quota distribution under load.
> 
> **Overview**
> Adds a new dynamic config `frontend.globalNamespaceWorkerDeploymentReadRPS` (default 50) and wires a **cluster-global, per-namespace** rate limiter into `WorkflowHandler` to throttle `DescribeWorkerDeployment` and `DescribeWorkerDeploymentVersion`, returning `ResourceExhausted` when exceeded.
> 
> Reclassifies `ListWorkerDeployments` into the *visibility* priority group (and updates tests) so it shares visibility quotas/prioritization rather than execution P3 handling, and introduces a reusable `NewGlobalNamespaceRateLimiter` helper to build global-only namespace limiters.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c8870ae2158fe8a2263194a3d0497682a213d1d9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>c8870ae</u></sup>


service/frontend/workflow_handler.go
```diff
@@ -233,6 +238,23 @@ func NewWorkflowHandler(
 		registry:            registry,
 	}
 
+	handler.workerDeploymentReadRateLimiter = quotas.NewNamespaceRequestRateLimiter(
```
Shivs11: I chose not to add tests (unit-tests, functional) over here only cause the primitives that this rate limiter is using is well tested. Moreover, looked around the codebase and did not find other areas of code where the users of these well-tested primitives had tests. (unless I just missed something)

service/frontend/workflow_handler.go
```diff
@@ -5688,6 +5718,26 @@ func (wh *WorkflowHandler) validateOnConflictOptions(opts *workflowpb.OnConflict
...
+		namespaceName,
+		callerInfo.CallerType,
+		-1,
+		callerInfo.CallOrigin,
+	)
+
+	if !wh.workerDeploymentReadRateLimiter.Allow(time.Now().UTC(), req) {
+		return serviceerror.NewResourceExhausted(
+			enumspb.RESOURCE_EXHAUSTED_CAUSE_RPS_LIMIT,
+			fmt.Sprintf("%s rate limit exceeded for namespace %q", apiName, namespaceName),
```
ShahabT: Nit: putting only one of the API names here could give the impression to the user that the limit is separately applied on individual APIs, while it's actually a combined limit.

common/dynamicconfig/constants.go
```diff
@@ -657,6 +657,16 @@ This config is EXPERIMENTAL and may be changed or removed in a later release.`,
 		`FrontendMaxNamespaceBurstRatioPerInstance is workflow namespace burst limit as a ratio of namespace RPS. The RPS
 used here will be the effective RPS from global and per-instance limits. The value must be 1 or higher.`,
 	)
+	FrontendWorkerDeploymentReadRPS = NewNamespaceIntSetting(
+		"frontend.workerDeploymentReadRPS",
```
dnr: put Namespace in the key? I hate long verbose names too but all the other per-namespace ones have it clearly in the key, so this should too

common/dynamicconfig/constants.go
```diff
@@ -657,6 +657,16 @@ This config is EXPERIMENTAL and may be changed or removed in a later release.`,
...
+		50,
+		`FrontendWorkerDeploymentReadRPS is the per-namespace rate limit for Worker Deployment Read APIs (DescribeWorkerDeployment, DescribeWorkerDeploymentVersion)`,
+	)
+	FrontendWorkerDeploymentReadBurstRatio = NewNamespaceFloatSetting(
+		"frontend.workerDeploymentReadBurstRatio",
+		2.0,
+		`FrontendWorkerDeploymentReadBurstRatio is the burst ratio for Worker Deployment Read APIs (DescribeWorkerDeployment, DescribeWorkerDeploymentVersion)`,
+	)
```
dnr: Is there a clear need for this? Can we just set it to 1 as a constant?

service/frontend/workflow_handler.go
```diff
@@ -104,6 +106,8 @@ const (
 	maxReasonLength              = 1000 // Maximum length for the reason field in RateLimitUpdate configurations.
 	defaultUserTerminateReason   = "terminated by user via frontend"
 	defaultUserTerminateIdentity = "frontend-service"
+
+	workerDeploymentReadRateLimitDefaultToken = 1
```
dnr: I don't think we need a constant for this, just use "1"

service/frontend/workflow_handler.go
```diff
@@ -233,6 +238,23 @@ func NewWorkflowHandler(
...
 
+	handler.workerDeploymentReadRateLimiter = quotas.NewNamespaceRequestRateLimiter(
+		func(req quotas.Request) quotas.RequestRateLimiter {
+			return quotas.NewRequestRateLimiterAdapter(
+				quotas.NewDynamicRateLimiter(
+					configs.NewNamespaceRateBurst(
+						req.Caller,
+						func(ns string) float64 {
+							return float64(config.WorkerDeploymentReadRPS(ns))
+						},
```
dnr: why not
```suggestion
						config.WorkerDeploymentReadRPS(ns),
```
?
cast it if the type doesn't match exactly

service/frontend/workflow_handler.go
```diff
@@ -3668,6 +3694,18 @@ func (wh *WorkflowHandler) DescribeWorkerDeployment(ctx context.Context, request
 		return nil, errRequestNotSet
 	}
 
+	if len(request.Namespace) == 0 {
```
Shivs11, line 3680: i think we forgot to include this so have done that in this PR - following the same pattern that is present in DescribeWorkerDeploymentVersion

common/dynamicconfig/constants.go
dnr: oh hmm, this is per-instance, not global? can we make it global? that seems better for this situation

service/frontend/workflow_handler.go
```diff
@@ -3668,6 +3694,10 @@ func (wh *WorkflowHandler) DescribeWorkerDeployment(ctx context.Context, request
 		return nil, errRequestNotSet
 	}
 
+	if err := wh.checkWorkerDeploymentReadRateLimit(ctx, request.GetNamespace(), "DescribeWorkerDeployment"); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
dnr: they should be as consistent as possible here, but we have interceptors to check namespace validity, it's not really a big deal

common/dynamicconfig/constants.go
ShahabT: oh yeah it should be global

common/dynamicconfig/constants.go
ShahabT: I'm fine with hardcoded 1 ratio

Shivs11: With a RPS of 50 and a burst ratio of 2.0, I think the idea was to have an allowance (some elastic room) for a short outburst - with a ratio like 1, I think at an instant the maximum # of requests that shall go through will only be 50?

I looked around elsewhere (such as at FrontendMaxNamespaceBurstRatioPerInstance) which also has a value of 2.0. Is there a strong reason to change this?

Shivs11: Sure, one could argue with a configuration of 50 RPS and 2 burst, one could send 100 requests every 2 seconds or so (since the bucket would be at capacity after every 2 seconds) but I think our system is also okay handling those 100 requests? which is why I had this in the first place

common/dynamicconfig/constants.go
Shivs11: oh....yes, you are right. I agree we should make it global.

common/dynamicconfig/constants.go
dnr: Two separate points:
1. The exact number doesn't really matter, I was mostly questioning why it has to be configurable. I think we use that knob very rarely if ever, since we fixed the burst size logic to use a minimum 1. We probably don't need it for new rate limiters.
2. A burst of 50 already seems excessive to me.. how many deployments do we expect in one namespace and how often might a client need to describe them? But sure, ratio of 1.0 or 2.0 doesn't really matter.

Shivs11: > how many deployments do we expect in one namespace and how often might a client need to describe them
the max # of deployments in a namespace is at a 100 right now

We do have clients that are sending Describe requests at 200 RPS already since they use the information the API returns to understand if they should be decommissioning their workers or not.

To your first point: Sure, yeah, maybe this does not need to be configurable. I shall change that and just make it 2 to keep things simple if you are okay with the number

ShahabT (approved):

common/dynamicconfig/constants.go
dnr: But is the 200 rps actually plausibly useful, or is that an abuse and 1 rps would be totally fine?

service/frontend/configs/quotas.go
```diff
@@ -384,6 +386,34 @@ func NewNamespaceReplicationInducingAPIPriorityRateLimiter(
...
+// only a global quota with no per-instance fallback.
+func NewGlobalNamespaceRateLimiter(
+	memberCounter calculator.MemberCounter,
+	globalQuota dynamicconfig.IntPropertyFnWithNamespaceFilter,
+	logger log.Logger,
+) quotas.RequestRateLimiter {
+	rateFn := calculator.NewLoggedNamespaceCalculator(
+		calculator.ClusterAwareNamespaceQuotaCalculator{
+			MemberCounter:    memberCounter,
+			PerInstanceQuota: func(ns string) int { return 0 },
```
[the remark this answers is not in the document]
dnr: If there are membership issues these calls won't work anyway since they do workflow queries

common/dynamicconfig/constants.go
```diff
@@ -657,15 +657,10 @@ This config is EXPERIMENTAL and may be changed or removed in a later release.`,
 		`FrontendMaxNamespaceBurstRatioPerInstance is workflow namespace burst limit as a ratio of namespace RPS. The RPS
 used here will be the effective RPS from global and per-instance limits. The value must be 1 or higher.`,
 	)
-	FrontendWorkerDeploymentReadRPS = NewNamespaceIntSetting(
-		"frontend.workerDeploymentReadRPS",
+	FrontendGlobalWorkerDeploymentReadRPS = NewNamespaceIntSetting(
```
dnr, line 660: nit: could put Namespace in this name too. I guess it's in the type though so unlikely to be confused

dnr (approved):

PR #9199, opened 2026-02-02
[Scheduler] Refactor CHASM scheduler tests to remove testify suite

lina-temporal:
## What changed?
- Refactored CHASM scheduler tests to remove the testify suite pattern
- Removed the mockSpecProcessor in favor of using the real one
- Added new helpers, as needed, to replace suite functionality

## Why?
- Simpler tests, more consistent with other CHASM tests.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


chasm/lib/scheduler/generator_tasks_test.go
```diff
@@ -16,129 +17,139 @@ import (
...
 		SpecBuilder:    legacyscheduler.NewSpecBuilder(),
 	})
 }
 
-func (s *generatorTasksSuite) TestExecute_ProcessTimeRangeFails() {
-	sched := s.scheduler
-	ctx := s.newMutableContext()
+func TestGeneratorTask_Execute_ProcessTimeRangeFails(t *testing.T) {
+	// Create a custom mock spec processor that fails on ProcessTimeRange.
+	ctrl := gomock.NewController(t)
```
chaptersix, line 32: IIRC we don't want to use mocks anymore. Is this difficult to remove?

chasm/lib/scheduler/backfiller_tasks_test.go
```diff
@@ -157,25 +125,26 @@ func (s *backfillerTasksSuite) TestBackfillTask_TriggerImmediateFullBuffer() {
...
-	s.runTestCase(&backfillTestCase{
+	runBackfillTestCase(t, env, &backfillTestCase{
 		InitialBackfillRequest: request,
 		ExpectedBufferedStarts: 5,
 		ExpectedComplete:       true,
-		ValidateInvoker: func(invoker *scheduler.Invoker) {
+		ValidateInvoker: func(t *testing.T, invoker *scheduler.Invoker) {
 			for _, start := range invoker.GetBufferedStarts() {
-				s.Equal(request.OverlapPolicy, start.OverlapPolicy)
+				require.Equal(t, request.OverlapPolicy, start.OverlapPolicy)
```
chaptersix, line 143: are the stack traces easy enough to follow when this assertion fails?

chasm/lib/scheduler/generator_tasks_test.go
lina-temporal: Mocks for validating codepaths we wouldn't normally expect to hit are still fine (as are mocks for service dependencies).

chasm/lib/scheduler/backfiller_tasks_test.go
lina-temporal: in terms of them being in a for loop or because it's in an anonymous function? Either way, I find the traces readable enough, I really just need to see what test is failing and on what assertion to start debugging.

chasm/lib/scheduler/invoker_tasks.go
```diff
@@ -79,10 +79,6 @@ const (
 	// Lower bound for the deadline in which buffered actions are dropped.
 	startWorkflowMinDeadline = 5 * time.Second
 
-	// Because the catchup window doesn't apply to a manual start, pick a custom
-	// execution deadline before timing out a start.
-	manualStartExecutionDeadline = 1 * time.Hour
```
chaptersix, line 84: why can this be removed? Was it only used in tests?

chasm/lib/scheduler/backfiller_tasks_test.go
chaptersix: yes, anonymous function and nested in `runBackfillTestCase` . Good with me

chasm/lib/scheduler/invoker_tasks.go
chaptersix: odd, I don't see a ref in main. I thought go would catch that.

chaptersix (approved):

chasm/lib/scheduler/invoker_tasks.go
chaptersix: ah, const/var at the package level are the exception.

PR #9201, opened 2026-02-02
Add a dummy workflow type to worker

lina-temporal:
## What changed?
- Adds a new dummy workflow type to the system worker.

## Why?
- See the comments on the workflow. This will be used to block a key in the
workflow ID space from being written to during the process of creating a CHASM
schedule, preventing a logical duplicate across separate ID spaces.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


service/worker/dummy/workflow.go
```diff
@@ -0,0 +1,28 @@
...
+// DummyWorkflow is a sentinel workflow that hangs open for a fixed duration.
+// It is used to reserve a particular key in the traditional workflow ID-space,
+// as opposed to the CHASM archetype-specific ID spaces.
+//
+// CHASM Scheduler creates a DummyWorkflow to block a V1 schedule of the same
+// logical schedule ID from being created in the workflow ID-space. The key stays
+// reserved for a period to account for a fleet's potentially inconsistent view of
+// the dynamic config feature flags controlling which stack schedulers are being
+// created on.
+func DummyWorkflow(ctx workflow.Context) error {
```
chaptersix, line 26: should we clear this workflow v2 -> v1 migration?

lina-temporal: Yes, we'll want to attempt to clear this on V2->V1 migrate (granted, this wf will also automatically clear itself after the fudge factor time I set).

PR #9202, opened 2026-02-02
Add CancelOutstandingWorkerPolls RPC to matching service

rkannan82:
## What changed?
Adds infrastructure to cancel all outstanding polls for a worker instance during shutdown.

Key Changes:

1. request_response.proto / service.proto
- New CancelOutstandingWorkerPolls RPC that cancels polls by worker_instance_key instead of individual poller_id

2. matching_engine.go:
- Added workerInstancePollers map to track pollers by worker instance key
- pollTask() now registers pollers in both outstandingPollers (by pollerID) and workerInstancePollers (by worker key)
- CancelOutstandingWorkerPolls() cancels all pollers for a worker instance and returns count

## Why?
To support eager cancellation of outstanding polls when worker call ShutdownWorker.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None. No changes to existing functionality.

### DIFF
--- a/api/matchingservice/v1/request_response.pb.go
+++ b/api/matchingservice/v1/request_response.pb.go
@@ -1529,12 +1529,11 @@ func (*CancelOutstandingPollResponse) Descriptor() ([]byte, []int) {
 
 // CancelOutstandingWorkerPollsRequest cancels all outstanding polls for a given worker instance key.
 type CancelOutstandingWorkerPollsRequest struct {
-	state       protoimpl.MessageState `protogen:"open.v1"`
-	NamespaceId string                 `protobuf:"bytes,1,opt,name=namespace_id,json=namespaceId,proto3" json:"namespace_id,omitempty"`
-	TaskQueue   *v14.TaskQueue         `protobuf:"bytes,2,opt,name=task_queue,json=taskQueue,proto3" json:"task_queue,omitempty"`
-	// If unset, both workflow and activity task queues will be cancelled.
-	TaskQueueType     v19.TaskQueueType `protobuf:"varint,3,opt,name=task_queue_type,json=taskQueueType,proto3,enum=temporal.api.enums.v1.TaskQueueType" json:"task_queue_type,omitempty"`
-	WorkerInstanceKey string            `protobuf:"bytes,4,opt,name=worker_instance_key,json=workerInstanceKey,proto3" json:"worker_instance_key,omitempty"`
+	state             protoimpl.MessageState `protogen:"open.v1"`
+	NamespaceId       string                 `protobuf:"bytes,1,opt,name=namespace_id,json=namespaceId,proto3" json:"namespace_id,omitempty"`
+	TaskQueue         *v14.TaskQueue         `protobuf:"bytes,2,opt,name=task_queue,json=taskQueue,proto3" json:"task_queue,omitempty"`
+	TaskQueueType     v19.TaskQueueType      `protobuf:"varint,3,opt,name=task_queue_type,json=taskQueueType,proto3,enum=temporal.api.enums.v1.TaskQueueType" json:"task_queue_type,omitempty"`
+	WorkerInstanceKey string                 `protobuf:"bytes,4,opt,name=worker_instance_key,json=workerInstanceKey,proto3" json:"worker_instance_key,omitempty"`
 	unknownFields     protoimpl.UnknownFields
 	sizeCache         protoimpl.SizeCache
 }


proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -240,6 +240,20 @@ message CancelOutstandingPollRequest {
 message CancelOutstandingPollResponse {
 }
 
+// CancelOutstandingWorkerPollsRequest identifies the worker that could be having outstanding polls.
+// These polls could be waiting on different partitions of the task queue.
```
dnr: We could be even more explicit in this comment that all polls from this worker are cancelled, even if they don't match the partition in `task_queue`

```suggestion
// CancelOutstandingWorkerPollsRequest cancels all outstanding polls in this process from the worker
// identified by worker_instance_key (regardless of which partitions or even task queues they are on).
```

service/matching/matching_engine.go
```diff
@@ -154,6 +155,9 @@ type (
 		// the frontend detects client connection is closed to prevent tasks being dispatched
 		// to zombie pollers.
 		outstandingPollers collection.SyncMap[string, context.CancelFunc]
+		// workerInstancePollers tracks pollerIDs by worker instance key for bulk cancellation
+		// during worker shutdown. Maps workerInstanceKey -> (pollerID -> struct{}).
+		workerInstancePollers collection.SyncMap[string, *collection.SyncMap[string, struct{}]]
```
dnr: In the draft that I sketched out, I had this map to the context.CancelFuncs directly instead of requiring the extra map lookup. What do you think about that?

service/matching/matching_engine.go
```diff
@@ -175,8 +179,9 @@ var (
 
 	errNoTasks = errors.New("no tasks")
 
-	pollerIDKey pollerIDCtxKey = "pollerID"
-	identityKey identityCtxKey = "identity"
+	pollerIDKey           pollerIDCtxKey       = "pollerID"
+	identityKey           identityCtxKey       = "identity"
+	workerInstanceKeyName workerInstanceCtxKey = "workerInstanceKey"
```
dnr: Will have to sync with #9221. I would say that these ctx key types should just be new types with concrete type `struct{}`. But if we're going to get rid of them then it doesn't matter.

service/matching/matching_engine.go
```diff
@@ -1166,6 +1174,32 @@ func (e *matchingEngineImpl) CancelOutstandingPoll(
 	return nil
 }
 
+func (e *matchingEngineImpl) CancelOutstandingWorkerPolls(
+	_ context.Context,
+	request *matchingservice.CancelOutstandingWorkerPollsRequest,
+) (*matchingservice.CancelOutstandingWorkerPollsResponse, error) {
+	workerInstanceKey := request.WorkerInstanceKey
+
+	pollerSet, ok := e.workerInstancePollers.Get(workerInstanceKey)
```
dnr: use `Pop` instead of `Get` here and then you don't need Delete below?

service/matching/matching_engine.go
dnr: collection.SyncMap is written to be a copyable value so that you can do:
```suggestion
		workerInstancePollers collection.SyncMap[string, collection.SyncMap[string, struct{}]]
```
we should probably do that to avoid the extra indirection?

service/matching/matching_engine.go
```diff
@@ -2767,7 +2802,26 @@ func (e *matchingEngineImpl) pollTask(
...
+				pollerSet, _ = e.workerInstancePollers.GetOrSet(workerInstanceKey, &newPollerSet)
+			}
+			pollerSet.Set(pollerID, struct{}{})
+		}
+
+		defer func() {
+			e.outstandingPollers.Delete(pollerID)
+			if hasWorkerKey && workerInstanceKey != "" {
+				if pollerSet, ok := e.workerInstancePollers.Get(workerInstanceKey); ok {
+					pollerSet.Delete(pollerID)
```
dnr: delete from parent map if it drops to zero size? if the ShutdownWorker call gets lost then we'd never clean up the top-level entry, right?

kind of tricky to do that without a race though. I wonder if it should be a plain map and mutex

service/matching/matching_engine.go
```diff
@@ -1166,6 +1174,32 @@ func (e *matchingEngineImpl) CancelOutstandingPoll(
...
+	for pollerID := range pollerSet.PopAll() {
+		if cancel, ok := e.outstandingPollers.Pop(pollerID); ok {
+			cancel()
+			cancelledCount++
+		}
+	}
+
+	// Clean up the worker instance entry
+	e.workerInstancePollers.Delete(workerInstanceKey)
+
```
dnr: just so we don't lose track of it: let's also remove from pollerhistory at this point (can be in a followup PR)

service/matching/matching_engine.go
rkannan82: Oh yea, this double lookup is unnecessary. The cancelFunction is a ptr anyway. Thanks for pointing this out.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
rkannan82: Not sure what you meant by "even if they don't match the partition".
I just changed it to: cancels all outstanding polls for a given worker instance key.

service/matching/matching_engine.go
rkannan82: That was an oversight. Fixed.

service/matching/matching_engine.go
rkannan82: True. I was thinking of Option-2.
1. [Adds contention] Add a mutex like you said. But this means all call are now serialized. Is the contention acceptable?
2. [Simple] Make SDK retry failed ShutdownWorker calls (its idempotent anyway).

service/matching/matching_engine.go
rkannan82: This requires a bit more work. So will address separately.
One thing is if its ok to remove the entry even though there can be more pollers with that identity. I understand that when those polls happen, the cache will be automatically populated again.

service/matching/matching_engine.go
rkannan82: Let me rework with Option-1. Didn't realize syncMap is our implementation and it already uses regular map + mutex. Contention is not that much of a problem given the operations are fairly quick.
Besides, ungraceful worker shutdown will result in memory leak. So we don't want that.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
```diff
@@ -240,6 +240,20 @@ message CancelOutstandingPollRequest {
 message CancelOutstandingPollResponse {
 }
 
+// CancelOutstandingWorkerPollsRequest cancels all outstanding polls for a given worker instance key.
+message CancelOutstandingWorkerPollsRequest {
+    string namespace_id = 1;
+    temporal.api.taskqueue.v1.TaskQueue task_queue = 2;
+    // If unset, both workflow and activity task queues will be cancelled.
```
dnr: this can't be unset, this is needed for routing, right?

service/matching/matching_engine.go
```diff
@@ -168,6 +180,42 @@ type (
...
+// Add registers a poller for a worker instance. Thread-safe.
+func (t *workerPollerTracker) Add(workerKey, pollerID string, cancel context.CancelFunc) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+	pollerCancels := t.pollers[workerKey]
+	if pollerCancels == nil {
+		pollerCancels = make(map[string]context.CancelFunc)
+		t.pollers[workerKey] = pollerCancels
+	}
+	pollerCancels[pollerID] = cancel
```
dnr: I made a little utility for this. maybe we should do it for remove too? there are a bunch of instances of this pattern.

```suggestion
	util.GetOrSetMap(t.pollers, workerKey)[pollerID] = cancel
```

service/matching/matching_engine.go
```diff
@@ -2774,7 +2831,23 @@ func (e *matchingEngineImpl) pollTask(
...
 	if pollerID, ok := ctx.Value(pollerIDKey).(string); ok && pollerID != "" {
 		e.outstandingPollers.Set(pollerID, cancel)
-		defer e.outstandingPollers.Delete(pollerID)
+
+		// Also track by worker instance key for bulk cancellation during shutdown.
+		// We use partition:pollerID as the key because when a poll is forwarded to a parent
+		// partition, the same pollerID is reused. Without the partition prefix, the parent's
+		// cancel func would overwrite the child's entry.
+		workerInstanceKey := pollMetadata.workerInstanceKey
+		partitionPollerKey := partition.RpcName() + ":" + pollerID
```
dnr: the key doesn't matter here, why not just use a random uuid? less explanation required then.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
dnr: The "rpc name" in `task_queue` specifies a partition, the rpc is directed at a specific partition of a specific task queue. I mean to say explicitly that when the rpc arrives, the partition that it's directed to is ignored and it applies to all polls on the whole instance. You already said it, I just thought it could be even more explicit, because it's kind of an unusual thing.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
rkannan82: If unset, the implementation does it for both workflow and activity types. wdyt?
In Pr #9203, service/frontend/workflow_handler.go
cancelOutstandingWorkerPolls
```
// Use task queue types from request, or default to both workflow and activity
	taskTypes := request.GetTaskQueueTypes()
	if len(taskTypes) == 0 {
		taskTypes = []enumspb.TaskQueueType{
			enumspb.TASK_QUEUE_TYPE_WORKFLOW,
			enumspb.TASK_QUEUE_TYPE_ACTIVITY,
		}
	}
```

service/matching/matching_engine.go
rkannan82: Done. Added util.DeleteFromMap

service/matching/matching_engine.go
rkannan82: Ack. Simplified the comment as well.

service/matching/matching_engine.go
```diff
@@ -168,6 +178,32 @@ type (
...
+	defer t.lock.Unlock()
+	util.DeleteFromMap(t.pollers, workerKey, pollerID)
+}
+
+// CancelAll cancels all pollers for a worker and removes the worker entry. Returns cancelled count. Thread-safe.
+func (t *workerPollerTracker) CancelAll(workerKey string) int32 {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+	pollerCancels := t.pollers[workerKey]
+	delete(t.pollers, workerKey)
```
dnr, line 199: nit: could unlock after here before doing the cancels

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
dnr: that's the workflowservice.ShutdownWorkerRequest, this is the matchingservice.CancelOutstandingWorkerPollsRequest

service/matching/matching_engine.go
dnr: sure. but I mean, there's no reason to talk about pollerid or forwarding at all: what we want is conceptually a map from worker instance key to set of CancelFuncs. so the natural way to express it is `map[string]map[CancelFunc]struct{}`. but that doesn't work in go, you can't use CancelFunc as a map key. so we introduce arbitrary unique keys so that the CancelFuncs can be values instead.

proto/internal/temporal/server/api/matchingservice/v1/request_response.proto
rkannan82: Oh yea! Got it. Removed that comment.

service/matching/matching_engine.go
rkannan82: You have a keen eye for this 😄

dnr (approved):

PR #9203, opened 2026-02-02
Implement eager poll cancellation on worker shutdown

rkannan82:
## What changed?

Problem: When a worker sends ShutdownWorker, it may still have outstanding poll requests waiting in matching. If a task is dispatched to one of these polls, we risk losing the task if the worker were to restart, or the poll times out. Since the worker is anyway wanting to shutdown, might as well not deliver any more tasks to it.

Solution: ShutdownWorker now fans out CancelOutstandingWorkerPolls to all task queue partitions, causing outstanding polls to return empty before tasks can be dispatched to them.

Key Changes:

1. workflow_handler.go:
- cancelOutstandingWorkerPolls() fans out cancellation to all partitions in parallel

2. service.go / constants.go: EnableCancelWorkerPollsOnShutdown dynamic config flag (default: false)

Flow:
- Worker calls ShutdownWorker with worker_instance_key and task_queue
- Frontend fans out CancelOutstandingWorkerPolls to all partitions (workflow + activity)
- Matching engine cancels all polls for that worker, returning empty responses
- Worker receives empty responses and completes shutdown cleanly

## Why?
Avoid dispatching tasks to workers that are shutting down by proactively cancelling their outstanding polls. This way we reduce the chances of losing a task due to worker restart/poll timeout.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
None. New feature disabled by default.

### DIFF
--- a/service/frontend/workflow_handler.go
+++ b/service/frontend/workflow_handler.go
@@ -2846,18 +2846,14 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
 
 	// Cancel outstanding polls (best-effort)
 	if wh.config.EnableCancelWorkerPollsOnShutdown(request.GetNamespace()) {
-		waitGroup.Add(1)
-		go func() {
-			defer waitGroup.Done()
+		waitGroup.Go(func() {
 			wh.cancelOutstandingWorkerPolls(ctx, namespaceID.String(), request)
-		}()
+		})
 	}
 
 	// Record final heartbeat (best-effort)
 	if request.WorkerHeartbeat != nil && wh.config.WorkerHeartbeatsEnabled(request.GetNamespace()) {
-		waitGroup.Add(1)
-		go func() {
-			defer waitGroup.Done()
+		waitGroup.Go(func() {
 			_, err := wh.matchingClient.RecordWorkerHeartbeat(ctx, &matchingservice.RecordWorkerHeartbeatRequest{
 				NamespaceId: namespaceID.String(),
 				HeartbeartRequest: &workflowservice.RecordWorkerHeartbeatRequest{
@@ -2871,7 +2867,7 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
 					tag.WorkflowTaskQueueName(request.WorkerHeartbeat.GetTaskQueue()),
 					tag.Error(err))
 			}
-		}()
+		})
 	}
 
 	// Unload sticky task queue (required - error fails shutdown)
@@ -2939,9 +2935,7 @@ func (wh *WorkflowHandler) cancelOutstandingWorkerPolls(
 		tq := tqFamily.TaskQueue(taskType)
 		for partitionID := range numPartitions {
 			partition := tq.NormalPartition(partitionID)
-			waitGroup.Add(1)
-			go func() {
-				defer waitGroup.Done()
+			waitGroup.Go(func() {
 				resp, err := wh.matchingClient.CancelOutstandingWorkerPolls(ctx, &matchingservice.CancelOutstandingWorkerPollsRequest{
 					NamespaceId: namespaceID,
 					TaskQueue: &taskqueuepb.TaskQueue{
@@ -2960,7 +2954,7 @@ func (wh *WorkflowHandler) cancelOutstandingWorkerPolls(
 				} else {
 					totalCancelled.Add(resp.CancelledCount)
 				}
-			}()
+			})
 		}
 	}
 	waitGroup.Wait()


common/dynamicconfig/constants.go
```diff
@@ -3071,6 +3071,14 @@ WorkerActivitiesPerSecond, MaxConcurrentActivityTaskPollers.
 		`WorkerHeartbeatsEnabled is a "feature enable" flag. It allows workers to send periodic heartbeats to the server.`,
 	)
 
+	EnableCancelWorkerPollsOnShutdown = NewNamespaceBoolSetting(
+		"frontend.enableCancelWorkerPollsOnShutdown",
+		false,
```
dnr, line 3091: nice to have a flag to turn it off, but any reason not to make it true by default?

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+	// Use task queue types from request, or default to both workflow and activity
+	taskTypes := request.GetTaskQueueTypes()
+	if len(taskTypes) == 0 {
+		taskTypes = []enumspb.TaskQueueType{
+			enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+			enumspb.TASK_QUEUE_TYPE_ACTIVITY,
+		}
+	}
+
+	// TODO: Optimize by grouping partitions by host and making one RPC per host instead of per partition.
```
dnr: I think that would be nice, although not a big deal for now. I think our rpc stack makes it kind of hard to do, unfortunately

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+	// The partition is only used for routing; the matching engine cancels all pollers for the workerInstanceKey.
+	// TODO: Consider retrying on transient failures.
```
dnr: I wouldn't bother with this, it's just an optimization

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+		if numPartitions < 1 {
+			numPartitions = 1
+		}
+
+		tq := tqFamily.TaskQueue(taskType)
+		for partitionId := 0; partitionId < numPartitions; partitionId++ {
+			partition := tq.NormalPartition(partitionId)
+			waitGroup.Add(1)
+			go func(p *tqid.NormalPartition, tt enumspb.TaskQueueType) {
+				defer waitGroup.Done()
```
dnr: can't wait for go 1.26 and waitGroup.Go()

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+		taskTypes = []enumspb.TaskQueueType{
+			enumspb.TASK_QUEUE_TYPE_WORKFLOW,
+			enumspb.TASK_QUEUE_TYPE_ACTIVITY,
+		}
+	}
+
+	// TODO: Optimize by grouping partitions by host and making one RPC per host instead of per partition.
+	// The partition is only used for routing; the matching engine cancels all pollers for the workerInstanceKey.
+	// TODO: Consider retrying on transient failures.
+	tqFamily := tqid.UnsafeTaskQueueFamily(namespaceId, taskQueueName)
```
dnr: don't use the Unsafe version outside of tests, just use the safe version and abort on error

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+		numPartitions := wh.config.NumTaskQueueReadPartitions(namespaceName, taskQueueName, taskType)
+		if numPartitions < 1 {
+			numPartitions = 1
+		}
+
+		tq := tqFamily.TaskQueue(taskType)
+		for partitionId := 0; partitionId < numPartitions; partitionId++ {
+			partition := tq.NormalPartition(partitionId)
+			waitGroup.Add(1)
+			go func(p *tqid.NormalPartition, tt enumspb.TaskQueueType) {
```
dnr: just use partition (or partitionId) and taskType in the function, no need to do this since go 1.22

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+	var failedPartitions atomic.Int32
+
+	for _, taskType := range taskTypes {
+		numPartitions := wh.config.NumTaskQueueReadPartitions(namespaceName, taskQueueName, taskType)
+		if numPartitions < 1 {
+			numPartitions = 1
+		}
+
+		tq := tqFamily.TaskQueue(taskType)
+		for partitionId := 0; partitionId < numPartitions; partitionId++ {
```
dnr: we settled on "ID" as the recommended style, except for proto-generated code that we can't control. also use range:

```suggestion
		for partitionID := range numPartitions {
```

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+						Kind: enumspb.TASK_QUEUE_KIND_NORMAL,
+					},
+					TaskQueueType:     tt,
+					WorkerInstanceKey: workerInstanceKey,
+				})
+				if err != nil {
+					failedPartitions.Add(1)
+					wh.logger.Warn("Failed to cancel outstanding polls for worker.",
+						tag.WorkflowTaskQueueName(p.RpcName()),
+						tag.NewStringTag("worker-instance-key", workerInstanceKey),
```
dnr: use the new aliases?
```suggestion
						tag.String("worker-instance-key", workerInstanceKey),
```

service/frontend/workflow_handler.go
```diff
@@ -2869,6 +2876,82 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+						tag.Error(err))
+				} else {
+					totalCancelled.Add(resp.CancelledCount)
+				}
+			}(partition, taskType)
+		}
+	}
+	waitGroup.Wait()
```
dnr, line 2960: this will block the other stuff that ShutdownWorker does from happening while we're waiting for the fan-out rpcs (which could be many)... what do you think about trying to do all three things (cancel polls, final heartbeat, unload sticky queue) concurrently and do a Wait on all of them at once?

common/dynamicconfig/constants.go
rkannan82: We can explicitly enable this for specific customers first before turning it on everywhere.

service/frontend/workflow_handler.go
rkannan82: Thanks for pointing this out.

service/frontend/workflow_handler.go
rkannan82: Logging an error and returning early.

service/frontend/workflow_handler.go
rkannan82: Actually the auto generatd retryable_client will handle retries. Removing this.

service/frontend/workflow_handler.go
```diff
@@ -2835,42 +2836,149 @@ func (wh *WorkflowHandler) ShutdownWorker(ctx context.Context, request *workflow
...
+				wh.logger.Error("Failed to record worker heartbeat during shutdown.",
+					tag.WorkflowTaskQueueName(request.WorkerHeartbeat.GetTaskQueue()),
+					tag.Error(err))
+			}
+		}()
+	}
+
+	// Unload sticky task queue (required - error fails shutdown)
+	// TODO: update poller info to indicate poller was shut down (pass identity/reason along)
+	var unloadErr atomic.Pointer[error]
```
dnr: there's no reason to make this atomic, access is synchronized by the waitgroup.

actually, why not just do this one inline since it's always done?

tests/task_queue_test.go
```diff
@@ -943,3 +943,78 @@ func (s *TaskQueueSuite) runActivitiesWithPriorities(
...
+			})
+			pollResults <- struct {
+				resp *workflowservice.PollWorkflowTaskQueueResponse
+				err  error
+			}{resp, err}
+		}()
+	}
+
+	// Give polls time to register with matching as there is no deterministic signal when pollers are registered.
+	time.Sleep(10 * time.Second)
```
dnr: the comment above says the test should complete in 1s, but this is a 10s sleep? 1s is probably enough here, right?

we could also do it with Eventually and DescribeTaskQueue to get the poller list, but personally I'm fine with a sleep.

tests/task_queue_test.go
```diff
@@ -943,3 +943,78 @@ func (s *TaskQueueSuite) runActivitiesWithPriorities(
...
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// All polls completed
+	case <-time.After(10 * time.Second):
+		s.Fail("polls did not complete within expected time after shutdown")
+	}
```
```suggestion
	s.True(common.AwaitWaitGroup(&wg, 10*time.Second), "polls did not complete within expected time after shutdown")
```

service/frontend/workflow_handler.go
rkannan82: Duh!

tests/task_queue_test.go
rkannan82: Nice!

tests/task_queue_test.go
rkannan82: I used a larger sleep to significantly lower the probability of a flake. DescribeTaskQueue is not helpful because the pollers are registered in history before the poll is actually recorded in matcher.
To remove timing assumptions, I have reworked the test to just keep issuing shutdown until we get the desired response.

dnr (approved):

PR #9204, opened 2026-02-02
[Scheduler] add sentinel support on the CHASM side of scheduler

lina-temporal:
## What changed?
- We reserve keys on the opposite stack to account for a fleet's inconsistent view of dynamic config.
- This is similar to the [dummy workflow PR](https://github.com/temporalio/temporal/pull/9201) for V1 stack.

## How did you test it?
- [ ] built
- [x] added new unit test(s)


chasm/lib/scheduler/proto/v1/message.proto
```diff
@@ -29,6 +29,10 @@ message SchedulerState {
     // The closed flag is set true after a schedule completes, and the idle timer
     // expires.
     bool closed = 9;
+
+    // When true, this scheduler is a sentinel that exists only to reserve the
+    // schedule ID. All API operations return FailedPrecondition.
+    bool sentinel = 10;
```
chaptersix, line 35: how should this  be handled in the functions that construct the migration request?
Do we still close the v1 workflow when after migrating it to v2?

chasm/lib/scheduler/handler_test.go
```diff
@@ -0,0 +1,97 @@
...
+	t *testing.T,
+	callFn func(sentinel *scheduler.Scheduler, ctx chasm.MutableContext, specBuilder *legacyscheduler.SpecBuilder) error,
+) {
+	sentinel, ctx, _ := setupSentinelForTest(t)
+	specBuilder := legacyscheduler.NewSpecBuilder()
+
+	err := callFn(sentinel, ctx, specBuilder)
+
+	require.Error(t, err)
+	var failedPreconditionErr *serviceerror.FailedPrecondition
```
chaptersix: should this also return and error specific to this being a sentinel workflow? There are other reasons we return this error code.

chasm/lib/scheduler/proto/v1/message.proto
lina-temporal: > how should this be handled in the functions that construct the migration request?

It should always be false, since we want to never attempt to migrate a sentinel. Migration cases don't need to create sentinels themselves, however, since they necessarily already occupy the source ID-space.

>  Do we still close the v1 workflow when after migrating it to v2?

Yes. The concern that sentinels address specifically is on the `CreateSchedule` case, where neither ID-space is occupied yet, during the period at which the `enableChasmSchedulerCreation: true` DC is taking effect (mixed-brain on dynamic config). Since frontend can route the `CreateSchedule` to any pod, including on retry, the sentinel strategy is used to occupy both ID-spaces. 

In the case of V1->V2, we don't have this issue. The source workflow already exists, and is assigned to a particular shard (and through shard ownership, a specific pod), which means we have a single writer. There's no confusion as to which ID space we're creating the next key in - we know we just want to create it in the CHASM space. There's also no confusion for frontend calls coming in during this migration, because our routing to CHASM vs WF isn't based on dynamic config: we simply always check both ID-spaces in all scheduler handlers.

chasm/lib/scheduler/handler_test.go
lina-temporal: Yeah, I think you're right, lack of differentiation makes the checks in workflow_handler too broad. Will fix.

PR #9206, opened 2026-02-03
Reduce history suggest CaN size for flaky test

spkane31:
## What changed?
Reduce `dynamicconfig.HistorySizeSuggestContinueAsNew` from 2 MB -> 10KB for a flaky test.

## Why?
Hopefully reduce flakiness of test

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
NA, test only


tests/transient_task_test.go
yycptt, line 137: we can revert this back to 2s now I think.

Issue #6459, opened 2024-08-28
Excessive new connections to Postgres 


DonForbes:
## Expected Behavior
Under load the number of connections to the Postgres database backend remains fairly consistent over time.

## Actual Behavior
When under load it has been observed that there are many new connections being made to the Postgres database being used as the DB backing for the history service.  (200+ new connections per second during load test). The expectation is that the number of connections may rise to handle the load but it should achieve a steady state and relatively few connections killed and re-established.

One suggestion is that the method get may be being called frequently and for some reason the refcount is not incremented so remains at 0 and many new connections returned.  To look into the problem further to see if this theory is valid.

https://github.com/temporalio/temporal/blob/b383ffffcbbeacdfce2fe021c30f093bab64b5d9/common/persistence/sql/factory.go#L195

## Steps to Reproduce the Problem

  1. Install self-hosted platform using Postgres as the DB
  2. Setup montoring for the number of connections/new connections being made to DB
  3. Run load test

## Specifications

  - Version: 1.24
  - Platform: Kubernetes via Helm charts

raymondregrello:
Out of curiosity, in the helm chart `values.yaml` under `server.config.persistence.default.sql`, have you set `maxIdleConns`, `maxConns` and `maxConnLifetime` to sufficiently high values?

I believe the defaults for `maxIdleConns` are pretty low; we initially saw a lot of connection thrashing when load testing until we set it to a higher value.

prathyushpv:
Hi @DonForbes, Since you're running a load test, it's possible the database is returning errors that are forcing connection refreshes. If that's the case, your total connection count might still be under the configured limit.

To confirm, could you check the `persistence_session_refresh_attempts` metric? If this metric is incrementing, that would indicate connection refreshes are happening and likely causing the issue you're seeing.

Setting a higher value for maxIdleConns might also fix this as @raymondregrello suggested.

prathyushpv:
I am closing this issue for now. Please feel free to reopen this if you still have this issue after setting maxIdleConns.

Aegdesil:
We are having the same issue under load.

The `persistence_session_refresh_attempts` stays constant and low (around 2).

I tried setting maxConns = 50 and maxIdleConns = 50, but I still see around 80 connections / s on the database, the total number of connections on PostgreSQL is actually around 30.

Using Temporal version 1.30.4.

PR #9211, opened 2026-02-04
Issue with spike of connection to database. Potential solution

oherych:
During load testing we found that we have too many short-lived (< 1 sec) connections to database. We tried to play with the configuration of the connection pool, but we found that Temporal just ignored them. It was the main surprise. 

Then I found that we have `type DbConn` that has an inside counter. And when this counter is zero Temporal force to close all connections to the database. So, if nobody uses the database right now, we just close all connections to it. 

For me, is not clear we do this. During the investigation, I found that those changes came from Cadence in 2019. Zero commends why we need them and what problem we try to resolve. Also, I found a similar issue https://github.com/temporalio/temporal/issues/6459.

So I see two potential way how to resolve it:
1) Just remove this counter. It is more cheap to keep all connections
2) Add some delay that should reduce spikes
 
I prefer the first option, but I implemented the second. This is not production ready and well-tested code. I just want to hear opinions of maintainers. 

## What changed?

## Why?

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

### DIFF
--- a/common/persistence/sql/factory.go
+++ b/common/persistence/sql/factory.go
@@ -3,6 +3,7 @@ package sql
 import (
 	"fmt"
 	"sync"
+	"time"
 
 	"go.temporal.io/server/common/config"
 	"go.temporal.io/server/common/log"
@@ -36,7 +37,8 @@ type (
 		sqlplugin.DB
 
 		sync.Mutex
-		refCnt int
+		refCnt     int
+		closeTimer *time.Timer
 	}
 )
 
@@ -189,6 +191,11 @@ func (c *DbConn) Get() (sqlplugin.DB, error) {
 		c.DB = conn
 	}
 	c.refCnt++
+	if c.closeTimer != nil {
+		c.closeTimer.Stop()
+		c.closeTimer = nil
+	}
+
 	return c, nil
 }
 
@@ -197,6 +204,11 @@ func (c *DbConn) ForceClose() {
 	c.Lock()
 	defer c.Unlock()
 	if c.DB != nil {
+		if c.closeTimer != nil {
+			c.closeTimer.Stop()
+			c.closeTimer = nil
+		}
+
 		err := c.DB.Close()
 		if err != nil {
 			fmt.Println("failed to close database connection, may leak some connection", err)
@@ -209,9 +221,19 @@ func (c *DbConn) ForceClose() {
 func (c *DbConn) Close() error {
 	c.Lock()
 	defer c.Unlock()
+
 	c.refCnt--
 	if c.refCnt == 0 {
-		return c.DB.Close()
+		c.closeTimer = time.AfterFunc(time.Minute, func() {
+			c.Lock()
+			defer c.Unlock()
+
+			if c.refCnt == 0 {
+				c.closeTimer = nil
+
+				_ = c.DB.Close()
+			}
+		})
 	}
 	return nil
 }


prathyushpv:
Hi @oherych, thanks for flagging this! Since you're running a load test, it's possible the database is returning errors that are forcing connection refreshes. If that's the case, your total connection count might still be under the configured limit.

To confirm, could you check the persistence_session_refresh_attempts metric? If this metric is incrementing, that would indicate connection refreshes are happening and likely causing the issue you're seeing.

Another thing to try out is setting maxIdleConns to a higher value. If temporal is deployed through helm chart, it sets the same value for maxIdleConns and maxConns: https://github.com/temporalio/helm-charts/blob/e4a3894a653a5ee8b5841dfb64f14804737531e6/charts/temporal/values/values.postgresql.yaml#L25
You can check if this is true in your deployment and set both maxIdleConns and maxConns to same value.
If maxIdleConns is much lower than maxConns, it could create this churn.

oherych:
@prathyushpv, thank for your answer. 

Do you have information what this logic exists? What sense to have `refCnt`? What sense to close all connections?

prathyushpv:
Hi @oherych,
The refCnt in DbConn is not related to managing idle connections. It's a resource lifecycle mechanism for the shared sql.DB connection pool as you noted in the description. I also don't have background about this mechanism. This is the PR that added this logic in cadence: https://github.com/cadence-workflow/cadence/pull/1808
We could have followed the approach of keeping all connections until persistence factory is closed(suggestion 1). But I wonder if this could leak connections in some tests where we create and delete persistence stores(I have to check and confirm this).

The problem you have noticed — closing excess connections down to maxIdleConns — is handled entirely by Go's standard database/sql package. When a connection is returned to the pool and the number of idle connections already exceeds maxIdleConns, Go automatically closes the excess one. So I don't think the problem you are facing is caused by refCnt logic. So this change may not fix that.

oherych:
@prathyushpv, thanks for your time. I found mistake in my process of investigation this issue. We randomly took few stuck traces from log and they all pointed to `Close()` method.  So, when I repeated the same issue in my local env I found that we have closes only in connection pool. We dont close connection in  DbConn so often. 

```
@@@ [0x14000520540].NewRefCountedDBConn()
@@@ [0x14000dd4bd8].Get() 0 
@@@ [0x14000dd4bd8].Close() 0 
@@@ [0x14000dd4bd8].ForceClose() 0 
@@@ [0x14001200720].NewRefCountedDBConn()
@@@ [0x14000ae6cf8].Get() 0 
@@@ [0x14001201920].NewRefCountedDBConn()
@@@ [0x140012018c0].Get() 0 
@@@ [0x1400101a420].NewRefCountedDBConn()
@@@ [0x14000dd5cb8].Get() 0 
@@@ [0x140012833e0].NewRefCountedDBConn()
@@@ [0x14001283380].Get() 0 
@@@ [0x14000e6fc80].NewRefCountedDBConn()
@@@ [0x1400105c638].Get() 0 
@@@ [0x14001200180].NewRefCountedDBConn()
@@@ [0x14001200120].Get() 0 
@@@ [0x14001035620].NewRefCountedDBConn()
@@@ [0x14000dd5838].Get() 0 
@@@ [0x14000520c00].NewRefCountedDBConn()
@@@ [0x14000520ba0].Get() 0 
@@@ [0x14001230c00].NewRefCountedDBConn()
@@@ [0x14001450cf8].Get() 0 
@@@ [0x14001450cf8].Close() 0 
@@@ [0x14001450cf8].ForceClose() 0 
```

And you're absolutely right `maxConns == maxIdleConns` fixed our problem. We tried this solution before but previous investigation confused me. I wanted to be sure I clear understood why we need this counter

PR #9214, opened 2026-02-04
Fix WorkerInstanceKey not forwarded between task queue partitions

rkannan82:
## What changed?
- When workflow or activity poll requests are forwarded from one partition to another, the WorkerInstanceKey field was not being copied to the forwarded request. This caused the worker's instance key to be lost.
- WorkerInstanceKey is a new field added to all poll requests. See [PR](https://github.com/temporalio/api/pull/686).

## Why?
We will be recording the WorkerInstanceKey in the ActivityInfo as part of task matching. This will be used later to send activity cancellation tasks to that worker.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks

### DIFF
--- a/service/matching/forwarder_test.go
+++ b/service/matching/forwarder_test.go
@@ -362,7 +362,7 @@ func (t *ForwarderTestSuite) TestForwardPollForActivityPreservesWorkerInstanceKe
 	task, err := t.fwdr.ForwardPoll(ctx, &pollMetadata{
 		workerInstanceKey: workerInstanceKey,
 	})
-	t.NoError(err)
+	t.Require().NoError(err)
 	t.NotNil(task)
 	t.NotNil(request)
 	t.Equal(workerInstanceKey, request.GetPollRequest().GetWorkerInstanceKey(),


service/matching/pri_forwarder.go
```diff
@@ -241,6 +242,7 @@ func ForwardPollWithTarget(
 				TaskQueueMetadata:         pollMetadata.taskQueueMetadata,
 				WorkerVersionCapabilities: pollMetadata.workerVersionCapabilities,
 				DeploymentOptions:         pollMetadata.deploymentOptions,
+				WorkerInstanceKey:         pollMetadata.workerInstanceKey,
```
dnr, line 245: why not nexus?

service/matching/matching_engine.go
```diff
@@ -638,6 +639,7 @@ pollLoop:
 			deploymentOptions:         request.DeploymentOptions,
 			forwardedFrom:             req.ForwardedSource,
 			conditions:                req.Conditions,
+			workerInstanceKey:         request.WorkerInstanceKey,
```
dnr, line 642: not blocking, but some cleanup if you want to do it:
I think pollerID and identity should also be in pollMetadata instead of context

service/matching/pri_forwarder.go
rkannan82: Sorry, I didn't understand this comment.

service/matching/matching_engine.go
rkannan82: Separate PR for this refactor.
https://github.com/temporalio/temporal/pull/9221

service/matching/pri_forwarder.go
rkannan82: Done. Updated the code path for PollNexusTaskQueueRequest as well.

service/matching/forwarder_test.go
```diff
@@ -312,6 +342,64 @@ func (t *ForwarderTestSuite) TestForwardPollForActivity() {
...
+	t.client.EXPECT().PollActivityTaskQueue(gomock.Any(), gomock.Any(), gomock.Any()).Do(
+		func(arg0 context.Context, arg1 *matchingservice.PollActivityTaskQueueRequest, arg2 ...interface{}) {
+			request = arg1
+		},
+	).Return(resp, nil)
+
+	task, err := t.fwdr.ForwardPoll(ctx, &pollMetadata{
+		workerInstanceKey: workerInstanceKey,
+	})
+	require.NoError(t.T(), err)
```
dnr: suite should set up a require instance automatically:
```suggestion
	t.Require().NoError(err)
```

PR #9218, opened 2026-02-04
Add a timeout to functional tests

spkane31:
## What changed?
Add a default 90s timeout to individual tests using the `testcore.NewEnv` for setup. Tests should use the `context.Context` returned from `testcore.NewEnv(...)` to fail early when a test exceeds the timeout.

## Why?
Go only allows to set a timeout for an entire test run via `go test`, but a single test case can make the entire run be stuck for the full test timeout time. This change will make those tests exit early speeding up the dev process.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Test change only, could introduce flakes if tests did not have previous context timeouts.

---

> [!NOTE]
> **Medium Risk**
> Touches shared test infrastructure and changes how contexts/timeouts are created and propagated across functional tests, which could introduce new test flakes or premature cancellations if timeouts are too aggressive.
> 
> **Overview**
> Functional tests created via `testcore.NewEnv` now get a *test-scoped* timeout context (default **90s**, configurable via `TEMPORAL_TEST_TIMEOUT` or per-test `testcore.WithTimeout`) that is canceled on timeout and fails the test during cleanup.
> 
> `testcore.NewContext` is updated to optionally derive from a parent context, and key functional tests are migrated to pass `s.Context()`/`env.Context()` into RPC calls and `context.WithTimeout(...)` to respect cancellations. Linting/docs are updated to discourage `context.Background()` in tests and document the new timeout knob.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 98852a5e43d62aec769af4caae7de9783f314065. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -194,17 +194,6 @@ func buildCLI() *cli.App {
 					tag.Bool("debug-mode", debug.Enabled),
 				)
 
-				var dynamicConfigClient dynamicconfig.Client
-				if cfg.DynamicConfigClient != nil {
-					dynamicConfigClient, err = dynamicconfig.NewFileBasedClient(cfg.DynamicConfigClient, logger, temporal.InterruptCh())
-					if err != nil {
-						return cli.Exit(fmt.Sprintf("Unable to create dynamic config client. Error: %v", err), 1)
-					}
-				} else {
-					dynamicConfigClient = dynamicconfig.NewNoopClient()
-					logger.Info("Dynamic config client is not configured. Using noop client.")
-				}
-
 				authorizer, err := authorization.GetAuthorizerFromConfig(
 					&cfg.Global.Authorization,
 				)
@@ -233,7 +222,6 @@ func buildCLI() *cli.App {
 				s, err := temporal.NewServer(
 					temporal.ForServices(services),
 					temporal.WithConfig(cfg),
-					temporal.WithDynamicConfigClient(dynamicConfigClient),
 					temporal.WithLogger(logger),
 					temporal.InterruptOn(temporal.InterruptCh()),
 					temporal.WithAuthorizer(authorizer),
--- a/common/dynamicconfig/file_based_client.go
+++ b/common/dynamicconfig/file_based_client.go
@@ -10,6 +10,7 @@ import (
 
 	"go.temporal.io/server/common/log"
 	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/metrics"
 )
 
 var _ Client = (*fileBasedClient)(nil)
@@ -37,9 +38,10 @@ type (
 		values          atomic.Value // ConfigValueMap
 		logger          log.Logger
 		reader          FileReader
-		lastUpdatedTime time.Time
+		lastCheckedTime time.Time
 		config          *FileBasedClientConfig
 		doneCh          <-chan any
+		metricsHandler  metrics.Handler
 
 		NotifyingClientImpl
 	}
@@ -50,20 +52,21 @@ type (
 )
 
 // NewFileBasedClient creates a file based client.
-func NewFileBasedClient(config *FileBasedClientConfig, logger log.Logger, doneCh <-chan any) (*fileBasedClient, error) {
+func NewFileBasedClient(config *FileBasedClientConfig, logger log.Logger, doneCh <-chan any, metricsHandler metrics.Handler) (*fileBasedClient, error) {
 	if config == nil {
 		return nil, errors.New("configuration for dynamic config client is nil")
 	}
 	reader := &osReader{path: config.Filepath}
-	return NewFileBasedClientWithReader(reader, config, logger, doneCh)
+	return NewFileBasedClientWithReader(reader, config, logger, doneCh, metricsHandler)
 }
 
-func NewFileBasedClientWithReader(reader FileReader, config *FileBasedClientConfig, logger log.Logger, doneCh <-chan any) (*fileBasedClient, error) {
+func NewFileBasedClientWithReader(reader FileReader, config *FileBasedClientConfig, logger log.Logger, doneCh <-chan any, metricsHandler metrics.Handler) (*fileBasedClient, error) {
 	client := &fileBasedClient{
 		logger:              logger,
 		reader:              reader,
 		config:              config,
 		doneCh:              doneCh,
+		metricsHandler:      metricsHandler,
 		NotifyingClientImpl: NewNotifyingClientImpl(),
 	}
 
@@ -110,15 +113,25 @@ func (fc *fileBasedClient) init() error {
 
 // This is public mainly for testing. The update loop will call this periodically, you don't
 // have to call it explicitly.
-func (fc *fileBasedClient) Update() error {
-	modtime, err := fc.reader.GetModTime()
-	if err != nil {
-		return fmt.Errorf("dynamic config file: %s: %w", fc.config.Filepath, err)
+func (fc *fileBasedClient) Update() (updateErr error) {
+	modtime, updateErr := fc.reader.GetModTime()
+	retryOnErr := true
+	defer func() {
+		if updateErr != nil {
+			metrics.DynamicConfigUpdateFailure.With(fc.metricsHandler).Record(1)
+		} else {
+			metrics.DynamicConfigUpdateFailure.With(fc.metricsHandler).Record(0)
+		}
+		if updateErr == nil || !retryOnErr {
+			fc.lastCheckedTime = modtime
+		}
+	}()
+	if updateErr != nil {
+		return fmt.Errorf("dynamic config file: %s: %w", fc.config.Filepath, updateErr)
 	}
-	if !modtime.After(fc.lastUpdatedTime) {
+	if !modtime.After(fc.lastCheckedTime) {
 		return nil
 	}
-	fc.lastUpdatedTime = modtime
 
 	contents, err := fc.reader.ReadFile()
 	if err != nil {
@@ -127,12 +140,14 @@ func (fc *fileBasedClient) Update() error {
 
 	lr := LoadYamlFile(contents)
 	for _, e := range lr.Errors {
-		fc.logger.Warn("dynamic config error", tag.Error(e))
+		fc.logger.Error("dynamic config error", tag.Error(e))
 	}
 	for _, w := range lr.Warnings {
 		fc.logger.Warn("dynamic config warning", tag.Error(w))
 	}
 	if len(lr.Errors) > 0 {
+		// we don't retry on parsing errors which will fail deterministically until the file is fixed
+		retryOnErr = false
 		return fmt.Errorf("loading dynamic config failed: %d errors, %d warnings",
 			len(lr.Errors), len(lr.Warnings))
 	}
--- a/common/dynamicconfig/file_based_client_test.go
+++ b/common/dynamicconfig/file_based_client_test.go
@@ -1,6 +1,7 @@
 package dynamicconfig_test
 
 import (
+	"errors"
 	"strings"
 	"testing"
 	"time"
@@ -11,6 +12,8 @@ import (
 	enumsspb "go.temporal.io/server/api/enums/v1"
 	"go.temporal.io/server/common/dynamicconfig"
 	"go.temporal.io/server/common/log"
+	"go.temporal.io/server/common/metrics"
+	"go.temporal.io/server/common/metrics/metricstest"
 	"go.temporal.io/server/common/retrypolicy"
 	"go.uber.org/mock/gomock"
 )
@@ -37,7 +40,7 @@ func (s *fileBasedClientSuite) SetupSuite() {
 	s.client, err = dynamicconfig.NewFileBasedClient(&dynamicconfig.FileBasedClientConfig{
 		Filepath:     "config/testConfig.yaml",
 		PollInterval: time.Second * 5,
-	}, logger, s.doneCh)
+	}, logger, s.doneCh, metrics.NoopMetricsHandler)
 	s.Require().NoError(err)
 	s.collection = dynamicconfig.NewCollection(s.client, logger)
 	s.collection.Start()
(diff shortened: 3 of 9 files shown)


tests/testcore/test_env.go
```diff
@@ -80,6 +91,21 @@ func WithDynamicConfig(setting dynamicconfig.GenericSetting, value any) TestOpti
...
+		o.timeout = duration
+	}
+}
+
+// WithoutTimeout disables the default test timeout.
+func WithoutTimeout() TestOption {
+	return func(o *testOptions) {
+		o.disableTimeout = true
+	}
+}
```
stephanos: I think it'd be better to not provide this; everything should have a timeout. In a pinch, someone could use `WithTimeout` to provide a very high timeout, right?

tests/reset_workflow_test.go
```diff
@@ -1152,3 +1152,7 @@ func (s *ResetWorkflowTestSuite) TestResetWorkflowWithExternalPayloads() {
 	s.Equal(int64(1), descResp.WorkflowExecutionInfo.ExternalPayloadCount)
 	s.Equal(workflowExternalPayloadSize, descResp.WorkflowExecutionInfo.ExternalPayloadSizeBytes)
 }
+
+func (s *ResetWorkflowTestSuite) Context() context.Context {
+	return context.Background()
+}
```
stephanos, line 1158: I can't quite connect the dots here, what's this for?

tests/testcore/test_env.go
```diff
@@ -212,3 +258,72 @@ func canBeNamespaceScoped(p dynamicconfig.Precedence) bool {
...
+	t.Cleanup(func() {
+		cancel()
+
+		if timedOut.Load() {
+			t.Errorf("Test exceeded timeout of %v", timeout)
+		}
+	})
+
+	return ctx
+}
```
stephanos, line 343: Since there's `context.go`, should we move the new functions there?

tests/testcore/test_env.go
```diff
@@ -135,6 +161,12 @@ func NewEnv(t *testing.T, opts ...TestOption) *testEnv {
 		t.Fatalf("Failed to register namespace: %v", err)
 	}
 
+	// Setup test timeout monitoring with context
+	ctx := context.Background()
```
stephanos: Should we use `t.Context()` as the parent? Devs might use it already in some code and also use it in the future, that way we can cancel that and make sure there's no dangling context.

tests/testcore/context.go
```diff
@@ -2,14 +2,27 @@ package testcore
...
-// NewContext create new context with default timeout 90 seconds.
-func NewContext() context.Context {
-	ctx, _ := rpc.NewContextWithTimeoutAndVersionHeaders(90 * time.Second * debug.TimeoutMultiplier)
+// NewContext creates a context with default 90-second timeout and RPC headers.
+// If a parent context is provided, the returned context will be canceled when
+// either the timeout expires OR the parent is canceled (whichever comes first).
+// Usage:
+//	ctx := testcore.NewContext()                    // Standalone RPC context
+//	ctx := testcore.NewContext(env.Context())       // Derived from test context
+func NewContext(parent ...context.Context) context.Context {
```
stephanos, line 22: What about only exposing `TestEnv`'s `Context` and make this an internal detail? I think that simplifies things. And we could always init the env's context with `NewContextFromParentWithTimeoutAndVersionHeaders`, no harm there AFAICT.

tests/testcore/test_env.go
```diff
@@ -212,3 +258,72 @@ func canBeNamespaceScoped(p dynamicconfig.Precedence) bool {
...
+		// Leave 30 second buffer before hard deadline so cleanup can run
+		timeout := time.Until(deadline) - (30 * time.Second)
+		if timeout <= 0 {
+			// Already close to deadline, skip timeout monitoring
+			return 0
+		}
+		return timeout
+	}
+
+	return defaultTestTimeout
```
stephanos: Should we move the `* debug.TimeoutMultiplier`  part from `defaultTestTimeout` here to have them all in one place?

tests/testcore/test_env.go
```diff
@@ -212,3 +258,72 @@ func canBeNamespaceScoped(p dynamicconfig.Precedence) bool {
...
+
+	deadline, hasDeadline := t.Deadline()
+	if hasDeadline {
+		// Leave 30 second buffer before hard deadline so cleanup can run
+		timeout := time.Until(deadline) - (30 * time.Second)
+		if timeout <= 0 {
+			// Already close to deadline, skip timeout monitoring
+			return 0
+		}
+		return timeout
```
stephanos: I could be wrong, but doesn't this return mean that whenever there's a `-timeout` set, we always default to that instead of `defaultTestTimeout`?

tests/testcore/test_env.go
```diff
@@ -212,3 +258,72 @@ func canBeNamespaceScoped(p dynamicconfig.Precedence) bool {
...
+//  3. Test deadline (via -timeout flag)
+//  4. Default 90 seconds
+func calculateTimeout(t *testing.T, customTimeout time.Duration) time.Duration {
+	if customTimeout > 0 {
+		// Use custom timeout if provided
+		return customTimeout * debug.TimeoutMultiplier
+	}
+
+	// Check for environment variable
+	if envTimeout := os.Getenv("TEMPORAL_TEST_TIMEOUT"); envTimeout != "" {
```
stephanos: This should be documented in `testing.md` as well.

tests/testcore/context.go
stephanos: And we could also require a parent context (since every test has one).

spkane31:
cursor review

tests/testcore/test_env.go
```diff
@@ -212,3 +258,72 @@ func canBeNamespaceScoped(p dynamicconfig.Precedence) bool {
...
+			timedOut.Store(true)
+		}
+	}()
+
+	// Register cleanup to cancel context and check timeout
+	// t.Cleanup() functions run in LIFO order, so this runs after test code
+	t.Cleanup(func() {
+		cancel()
+
+		if timedOut.Load() {
```
stephanos: There's a race here, found by Claude:
> t.Cleanup calls cancel(), which sets the error
   to context.Canceled. If the goroutine hasn't yet read ctx.Err() when cancel() fires, you might miss reporting a genuine timeout. Or conversely, if the deadline fires at the exact moment
  cleanup runs, the check may be unreliable.

tests/testcore/test_env.go
```diff
@@ -174,6 +207,21 @@ func (e *testEnv) Tv() *testvars.TestVars {
...
+//
+// For RPC operations that need headers, use:
+//
+//	ctx, _ := rpc.NewContextFromParentWithTimeoutAndVersionHeaders(env.Context(), 90*time.Second)
+//
+// For custom timeouts, use:
+//
+//	ctx, cancel := context.WithTimeout(env.Context(), 10*time.Second)
+//	defer cancel()
+func (e *testEnv) Context() context.Context {
```
stephanos, line 277: One more thought; we should add a lint rule that flags the user of `context.Background()` inside `tests/`; asking the user to use `t.Context()`.

stephanos: In a separate PR we could find-replace all `context.Background`s.

tests/reset_workflow_test.go
spkane31: This and the same in versioning_3_test.go is to have those suites meet the interface for `testcore.Env`.

tests/testcore/test_env.go
```diff
@@ -80,6 +91,21 @@ func WithDynamicConfig(setting dynamicconfig.GenericSetting, value any) TestOpti
 	}
 }
 
+// WithTimeout sets a custom timeout for the test. The test will fail if it runs longer
+// than this duration. The timeout is multiplied by debug.TimeoutMultiplier when debugging.
+func WithTimeout(duration time.Duration) TestOption {
+	return func(o *testOptions) {
+		o.timeout = duration
+	}
```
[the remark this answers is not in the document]
spkane31: Removing WithoutTime

tests/testcore/test_env.go
spkane31: Yes, I'm going to remove this. If we use the full test timeout for each test then this has no real change

tests/reset_workflow_test.go
stephanos: Oh I see. Until we migrate, couldn't they use `s.T().Context()` instead?

tests/testcore/context.go
spkane31: I think this should be done in a separate PR because it will touch a ton of files.

spkane31:
cursor review

spkane31:
cursor review

docs/development/testing.md
```diff
@@ -16,6 +16,7 @@ This document describes the project's testing setup, utilities and best practice
 - `TEMPORAL_TEST_OTEL_OUTPUT`: Enables OpenTelemetry (OTEL) trace output for failed tests to the provided file path.
 - `TEMPORAL_TEST_SHARED_CLUSTERS`: Number of shared clusters in the pool. Each can be used by multiple tests simultaneously.
 - `TEMPORAL_TEST_DEDICATED_CLUSTERS`: Number of dedicated clusters in the pool. Each can be used by one test only at a time.
+- `TEMPORAL_TEST_TIMEOUT`: Sets the default timeout for tests in seconds (e.g., `90` for 90 seconds). This can be overridden per-test using `testcore.WithTimeout()`. The timeout is multiplied by `debug.TimeoutMultiplier` when debugging.
```
stephanos: Maybe clarify a little more that it's a per-test limit (vs all tests)

stephanos: PS: why not allow time duration like `90s` actually?

tests/testcore/context.go
```diff
@@ -2,14 +2,97 @@ package testcore
...
+		return ctx
+	}
+
+	// Create standalone RPC context
+	ctx, _ := rpc.NewContextWithTimeoutAndVersionHeaders(defaultTestTimeout)
+	return ctx
+}
+
+// calculateTimeout determines the appropriate timeout duration based on custom timeout,
+// environment variable, test deadline, and default values. Returns 0 if timeout should be skipped.
```
stephanos: nit: it's not using the `test deadline`, is it?

nit: it never `returns 0`, does it? I'd actually say it also shouldn't ever since that means unbounded (undesirable)

tests/testcore/context.go
```diff
@@ -2,14 +2,97 @@ package testcore
...
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		<-ctx.Done()
+		// Store the error immediately to capture whether it was DeadlineExceeded or Canceled
+		contextErr.Store(ctx.Err())
+	}()
+
+	// Register cleanup to cancel context and check timeout
+	// t.Cleanup() functions run in LIFO order, so this runs after test code
```
stephanos:
```suggestion
	// Register cleanup to cancel context and check timeout.
	// t.Cleanup() functions run in LIFO order, so this runs after test code.
```

(easer to read)

tests/testcore/context.go
```diff
@@ -2,14 +2,97 @@ package testcore
...
+}
+
+// setupTestTimeoutWithContext creates a context that will be canceled on timeout,
+// and reports the timeout error during cleanup. Returns a context that tests can
+// use to be interrupted when timeout occurs. The context includes RPC version headers.
+func setupTestTimeoutWithContext(t *testing.T, customTimeout time.Duration) context.Context {
+	t.Helper()
+
+	timeout := calculateTimeout(customTimeout)
+	if timeout <= 0 {
```
stephanos: As mentioned above; is this something we actually want to allow?

tests/testcore/context.go
```diff
@@ -2,14 +2,97 @@ package testcore
...
+	// t.Cleanup() functions run in LIFO order, so this runs after test code
+	t.Cleanup(func() {
+		cancel()
+		<-done
+
+		// Check the captured error, not the current ctx.Err()
+		if err, ok := contextErr.Load().(error); ok && err == context.DeadlineExceeded {
+			t.Errorf("Test exceeded timeout of %v", timeout)
+		}
+	})
```
stephanos, line 75: I'm not sure yet I understand why this can't be

```
    t.Cleanup(func() {
          cancel()
          if ctx.Err() == context.DeadlineExceeded {
              t.Errorf("Test exceeded timeout of %v", timeout)
          }
      })
```

If the deadline fires first, `ctx.Err()` is DeadlineExceeded.
If cancel() fires first, `ctx.Err()` is Canceled.

spkane31:
cursor review

tests/testcore/test_env.go
```diff
@@ -92,6 +99,15 @@ func WithDynamicConfig(setting dynamicconfig.GenericSetting, value any) TestOpti
...
 }
 
+// WithTimeout sets a custom timeout for the test. The test will fail if it runs longer
+// than this duration. The timeout is multiplied by debug.TimeoutMultiplier when debugging.
+// The TEMPORAL_TEST_TIMEOUT environment variable can also set the default timeout in seconds.
+func WithTimeout(duration time.Duration) TestOption {
+	return func(o *testOptions) {
+		o.timeout = duration
+	}
+}
```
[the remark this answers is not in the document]
spkane31: THis will be called later, ok for now

tests/testcore/context.go
spkane31: It will not return 0

tests/testcore/context.go
spkane31: No, timeout will never be <=0 so removing this

tests/testcore/context.go
```diff
@@ -2,14 +2,78 @@ package testcore
...
+//  1. Custom timeout (via WithTimeout option)
+//  2. TEMPORAL_TEST_TIMEOUT environment variable (in seconds)
+//  3. Default 90 seconds
+func calculateTimeout(customTimeout time.Duration) time.Duration {
+	if customTimeout > 0 {
+		return customTimeout * debug.TimeoutMultiplier
+	}
+
+	if envTimeout := os.Getenv("TEMPORAL_TEST_TIMEOUT"); envTimeout != "" {
+		if seconds, err := strconv.Atoi(envTimeout); err == nil && seconds > 0 {
```
stephanos: Should this be parsing a `time.Duration` now instead of an integer?

PR #9221, opened 2026-02-05
Refactor: Move pollerID and identity from context to pollMetadata struct

rkannan82:
## What changed?
Consolidates poll-related data by moving pollerID and identity from context keys to the pollMetadata struct
Removes pollerIDKey and identityKey context key types.

## Why?
Consistent with how workerInstanceKey and other poll fields are already passed.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


dnr (approved):
Thanks! This will have some conflicts with the other ones, but approved for when you resolve all those

PR #9229, opened 2026-02-05
Patch for standalone activity rollback

fretz12:
## What changed?
Drop outstanding standalone activity chasm tasks. Return unimplemented error if standalone activity is disabled. Disabled standalone activity integration tests.

## Why?
In the event of a rollback for a server version where standalone activity was enabled to v1.30.0-148, we need to gracefully handle any outstanding tasks and ensure standalone APIs return unimplemented. Integration test also must be disabled since standalone activity will not function in this version.

## How did you test it?
- [X] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/transfer_queue_task_executor_base.go
+++ b/service/history/transfer_queue_task_executor_base.go
@@ -11,6 +11,7 @@ import (
 	"go.temporal.io/server/api/matchingservice/v1"
 	taskqueuespb "go.temporal.io/server/api/taskqueue/v1"
 	"go.temporal.io/server/chasm"
+	"go.temporal.io/server/chasm/lib/activity"
 	"go.temporal.io/server/common"
 	"go.temporal.io/server/common/debug"
 	"go.temporal.io/server/common/locks"
@@ -203,6 +204,10 @@ func (t *transferQueueTaskExecutorBase) processDeleteExecutionTask(
 	task *tasks.DeleteExecutionTask,
 	ensureNoPendingCloseTask bool,
 ) error {
+	if activity.ShouldDropStandaloneActivityTask(task.ArchetypeID) {
+		return nil
+	}
+
 	if task.ArchetypeID == chasm.UnspecifiedArchetypeID {
 		task.ArchetypeID = chasm.WorkflowArchetypeID
 	}


chasm/lib/activity/frontend.go
```diff
@@ -83,7 +83,7 @@ func (h *frontendHandler) IsStandaloneActivityEnabled(namespaceName string) bool
 // 3. Sends the request to the history activity service.
 func (h *frontendHandler) StartActivityExecution(ctx context.Context, req *workflowservice.StartActivityExecutionRequest) (*workflowservice.StartActivityExecutionResponse, error) {
 	if !h.config.Enabled(req.GetNamespace()) {
-		return nil, serviceerror.NewUnavailable(StandaloneActivityDisabledError)
+		return nil, ErrStandaloneActivityDisabled
```
bergundy: Put this outside of the if statement. When we roll back, we will not want to roll back the dynamic config.

service/history/timer_queue_task_executor_base.go
```diff
@@ -341,3 +341,11 @@ func (t *timerQueueTaskExecutorBase) executeStateMachineTimers(
...
 	return processedTimers, nil
 }
+
+// shouldDropStandaloneActivityTask indicates whether a task should be dropped based on whether it is a standalone
+// activity task. This is used in the event of server downgrade from a version where standalone activities was supported,
+// so that such tasks won't block queue processing.
+func (t *timerQueueTaskExecutorBase) shouldDropStandaloneActivityTask(archetypeID uint32) bool {
+	fqn, found := t.shardContext.ChasmRegistry().ComponentFqnByID(archetypeID)
+	return found && fqn == "activity.activity"
+}
```
yycptt: I think there's a ActivityArchetypeID constant?

yycptt (commented):
There a chasm task for visibility queue as well. (also for outbound queue but not used by SAA) 

hmm and what about retention timer task? That one is ok to run? I guess I am a bit confused on the expected behavior here.

If the idea is just to make all component tasks no-op, can change the chasm task validator and executor logic to be noop?

chasm/lib/activity/frontend.go
fretz12: Hardcoded to disable all SAA endpoints

service/history/timer_queue_task_executor_base.go
fretz12: We've not declared it, but I can add it so we don't have to do the lookup

fretz12: Changed

fretz12:
> There a chasm task for visibility queue as well. (also for outbound queue but not used by SAA)
> 
> hmm and what about retention timer task? That one is ok to run? I guess I am a bit confused on the expected behavior here.
> 
> If the idea is just to make all component tasks no-op, can change the chasm task validator and executor logic to be noop?

@yycptt - Seems the visibility queue and retention timer should function normally without SAA? Not too sure about the implications here. The action here is to clean up anything SAA related that can block queues or impact existing code.
@bergundy add your thoughts

chasm/lib/activity/library.go
```diff
@@ -6,6 +6,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+const archetypeID = 3332429922
```
yycptt, line 9: oh I was referring to [this](https://github.com/temporalio/temporal/blob/main/chasm/lib/activity/library.go#L20), but looks like that's not in 148 :(, sorry 🙏 

but that works for me as well. just one time thing for this release.

yycptt:
> > There a chasm task for visibility queue as well. (also for outbound queue but not used by SAA)
> > hmm and what about retention timer task? That one is ok to run? I guess I am a bit confused on the expected behavior here.
> > If the idea is just to make all component tasks no-op, can change the chasm task validator and executor logic to be noop?
> 
> @yycptt - Seems the visibility queue and retention timer should function normally without SAA? Not too sure about the implications here. The action here is to clean up anything SAA related that can block queues or impact existing code. @bergundy add your thoughts

I see. yeah makes sense. What will happen to executions that have some tasks skipped? they may be in a stuck state (e.g. a timeout task get skipped). How will we eventually find those executions and delete it from DB?

> that can block queues

I am not really concerned about the queues themselves TBH, if a task can't be processed, it will eventually be DLQed. but yeah it will trigger alerts which will be annoying.

fretz12:
> > @yycptt - Seems the visibility queue and retention timer should function normally without SAA? Not too sure about the implications here. The action here is to clean up anything SAA related that can block queues or impact existing code. @bergundy add your thoughts
> 
> I see. yeah makes sense. What will happen to executions that have some tasks skipped? they may be in a stuck state (e.g. a timeout task get skipped). How will we eventually find those executions and delete it from DB?

@yycptt - we had discussed it's not necessary salvage those executions that are happening midway since the rollback can potentially happen only in private preview IIRC. 

Valid point on cleaning up orphans in DB. I'm not sure how the schema is laid out for chasm executions. Can we put together a cleanup script to query activites that re in running state for X amount of time? 
And maybe a product question- we'll know which namespaces have SAA enabled. Maybe just nuke it after private preview?

bergundy:
Agree that we should avoid having orphans in both the executions DB and in the visibility DB.

I think a better approach here would be to invalidate all of the tasks and prevent them from running and avoid implementing the search attributes provider API. The visibility component should be allowed to delete the execution from visibility in that case. There may be other visibility related changes we need though, I will need to see the diff between 148 and 149 to tell.

bergundy:
@yycptt and I just synced. Let's drop the all of tasks that are relevant to the activity.activity archetype, including retention and visibility. This will create garbage but we can manually clean the bulk of it up if needed.

service/history/transfer_queue_task_executor_base.go
```diff
@@ -203,6 +204,10 @@ func (t *transferQueueTaskExecutorBase) processDeleteExecutionTask(
 	task *tasks.DeleteExecutionTask,
 	ensureNoPendingCloseTask bool,
 ) error {
+	if activity.ShouldDropStandaloneActivityTask(task.ArchetypeID) {
```
yycptt: hmm we can have this one but this is not the retention timer. Retention timer is the `DeleteHistoryEventTask` timer task. (sorry for the historically bad name...).

fretz12: Ah ok, added check at DeleteHistoryEventTask as well

yycptt (approved):

PR #9231, opened 2026-02-05
Store worker attributes needed by server to propagate nexus tasks to worker

rkannan82:
## What changed?
As part of RecordActivityTaskStarted flow, store worker_control_task_queue for an activity in the mutable state (ActivityInfo).

Main changes:
- executions.proto: Added the new worker_control_task_queue field.
- mutable_state_impl.go: Update mutable state.
- matching/forwarder.go: Propagate worker_control_task_queue when polls get forwarded. Otherwise, RecordActivityTaskStarted request will not have it set when invoked from a forwarded poll.

## Why?
To support activity cancellation without activity heartbeat.

Overall flow:
- [This PR] Store worker attributes in ActivityInfo as part of RecordActivityTaskStarted call.
- [#9232] When user cancels a workflow, create 1 or more tasks. Group all activities belonging to a worker into the task (for efficiency).
- [#9233] Lookup the Nexus task queue for each worker, and send a Nexus operation for each transfer task.
- [SDK] Worker will receive this cancel task and cancel the running activities.

## How did you test it?
- [ ] built
- [x] added new unit test(s)


service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -556,6 +556,7 @@ func (handler *workflowTaskCompletedHandler) handlePostCommandEagerExecuteActivi
 		stamp,
 		nil,
 		nil,
+		"", // workerInstanceKey not available for eager dispatch
```
yycptt: do we know the workerInstanceKey for the worker that's completing the workflow task?

service/history/workflow/mutable_state_impl.go
```diff
@@ -4096,6 +4099,7 @@ func (ms *MutableStateImpl) AddActivityTaskStartedEvent(
 		activityInfo.RequestId = requestID
 		activityInfo.StartedTime = timestamppb.New(ms.timeSource.Now())
 		activityInfo.StartedIdentity = identity
+		activityInfo.WorkerInstanceKey = workerInstanceKey
```
yycptt: L4069 already updated the field?

rkannan82: Removed.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: Ah, I didn't realize the eager activity execution path.
Sent out API PR to add this field. https://github.com/temporalio/api/pull/711.

rkannan82: Done.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -630,6 +630,12 @@ message ActivityInfo {
     bool reset_heartbeats = 48;
 
     int64 start_version = 50;
+
+    // Unique identifier of the worker that is this activity.
+    string worker_instance_key = 51;
```
yycptt: synced offline, this field is not really required and can be removed.

yycptt (approved):
LGTM overall. Please land after removing the unnecessary field.

PR #9232, opened 2026-02-05
Add WorkerCommandsTask outbound task to dispatch worker commands via Nexus

rkannan82:
## What changed?

New outbound task type (`WorkerCommandsTask`) that carries worker commands to be dispatched to workers via Nexus. Uses the generic `WorkerCommand` proto (not cancel-activity-specific), so this task type can carry any future command types.

Suggested review order: proto changes → `worker_commands_task.go` → `task_generator.go` → `workflow_task_completed_handler.go`

Key pieces:
- **Proto**: `TASK_TYPE_WORKER_COMMANDS` enum, `WorkerCommandsTask` in `OutboundTaskInfo` with `repeated WorkerCommand`.
- **Task definition**: `worker_commands_task.go` — implements outbound `Task` and `HasDestination` interfaces.
- **Task creation** (`workflow_task_completed_handler.go`, `task_generator.go`): When `RequestCancelActivityTask` is processed for a started activity whose worker has a control queue, collects a `CancelActivityCommand` with the activity's task token. Commands are batched by destination control queue and flushed as one `WorkerCommandsTask` per queue at the end of WFT processing.
- **Serialization**: `task_serializers.go` for persistence round-tripping.

Dispatch is a no-op here — handled in #9233. Gated by dynamic config `EnableCancelActivityWorkerCommand` (default: off).

## Why?

To support proactive activity cancellation without waiting for heartbeat. This is the task creation leg of the flow.

1. [#9231] Store `worker_control_task_queue` in `ActivityInfo` at activity start.
2. **[This PR]** On `RequestCancelActivityTask`, batch commands by control queue into `WorkerCommandsTask` outbound tasks.
3. [#9233] Dispatch each task as a Nexus `ExecuteCommands` operation to the worker, with a 3-attempt retry cap.
4. [SDK] Worker receives the cancel command and cancels the running activity.

Gated by dynamic config `EnableCancelActivityWorkerCommand` (default: off).

## How did you test it?

**Unit tests** cover task generation, command batching (including multi-queue batching), task serialization round-tripping, and the feature-flag-off path.

### DIFF
--- a/service/history/workflow/task_generator_test.go
+++ b/service/history/workflow/task_generator_test.go
@@ -1164,9 +1164,9 @@ func TestGenerateWorkerCommandsTasks(t *testing.T) {
 				require.Len(t, capturedTasks, 1)
 				commandTask, ok := capturedTasks[0].(*tasks.WorkerCommandsTask)
 				require.True(t, ok)
-				assert.Equal(t, tc.commands, commandTask.Commands)
-				assert.Equal(t, tc.controlQueue, commandTask.Destination)
-				assert.Equal(t, tests.NamespaceID.String(), commandTask.NamespaceID)
+				require.Equal(t, tc.commands, commandTask.Commands)
+				require.Equal(t, tc.controlQueue, commandTask.Destination)
+				require.Equal(t, tests.NamespaceID.String(), commandTask.NamespaceID)
 			}
 		})
 	}


proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -349,11 +349,21 @@ message TransferTaskInfo {
         // by some other task, so this task doesn't need to worry about it.
         bool can_skip_visibility_archival = 1;
     }
+
+    // Details for a Nexus task that cancels activities belonging to a specific worker.
+    message CancelActivityNexusTaskDetails {
+        // Scheduled event IDs of activities to cancel.
+        repeated int64 scheduled_event_ids = 1;
+        string worker_instance_key = 2;
```
yycptt: should be `worker_control_task_queue` now?

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -668,6 +669,13 @@ func (handler *workflowTaskCompletedHandler) handleCommandRequestCancelActivity(
 				return nil, err
 			}
 			handler.activityNotStartedCancelled = true
+		} else if ai.WorkerInstanceKey != "" && handler.config.EnableActivityCancellationNexusTask() {
+			// Activity has started and worker supports Nexus tasks - create cancel task.
+			handler.mutableState.AddTasks(&tasks.CancelActivityNexusTask{
+				WorkflowKey:       handler.mutableState.GetWorkflowKey(),
+				ScheduledEventIDs: []int64{ai.ScheduledEventId},
+				WorkerInstanceKey: ai.WorkerInstanceKey,
+			})
```
yycptt: I'd move the logic into ApplyActivityTaskCanceledEvent in mutable state impl, and do AddTasks in task_generator.go, to be consistent with rest of the codebase for task generation.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -668,6 +669,13 @@ func (handler *workflowTaskCompletedHandler) handleCommandRequestCancelActivity(
 				return nil, err
 			}
 			handler.activityNotStartedCancelled = true
+		} else if ai.WorkerInstanceKey != "" && handler.config.EnableActivityCancellationNexusTask() {
+			// Activity has started and worker supports Nexus tasks - create cancel task.
+			handler.mutableState.AddTasks(&tasks.CancelActivityNexusTask{
```
yycptt: Version is not specified?

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
yycptt, line 1: what about RespondActivityTaskCanceled api?

service/history/transfer_queue_active_task_executor.go
yycptt, line 105: plz update this (and related) methods as well for metric tags.

service/history/transfer_queue_active_task_executor.go
```diff
@@ -145,6 +145,9 @@ func (t *transferQueueActiveTaskExecutor) Execute(
 		err = t.processDeleteExecutionTask(ctx, task)
 	case *tasks.ChasmTask:
 		err = t.executeChasmSideEffectTransferTask(ctx, task)
+	case *tasks.CancelActivityNexusTask:
```
yycptt: if it's ok for this task to be delayed when namespace is throttled (I think so since delaying this task won't impact user workflows from making progress), mark it as low priority in priority_assigner.go

proto/internal/temporal/server/api/persistence/v1/executions.proto
rkannan82: Yea, Done.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: I kept the top level check in handler because we need to generate the nexus tasks only if the activity is started. Moved the AddTasks code inside the generator as you suggested. Please see if this makes sense.

Just fyi: I have a separate PR to batch activities by the queue. Right now we end up creating 1 task per activity.

service/history/transfer_queue_active_task_executor.go
rkannan82: Done.

service/history/transfer_queue_active_task_executor.go
rkannan82: Done.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: RespondActivityTaskCanceled is what the SDK sends after it has cancelled. Here, we are initiating that cancel (after Step-5). From my understanding, this is how it works e2e.

1. User calls RequestCancelWorkflowExecution API
2. Server creates a new workflow task for the workflow
3. SDK picks up the workflow task
4. Workflow code sees cancellation, decides to cancel activities
5. SDK sends RespondWorkflowTaskCompleted with RequestCancelActivityTask commands

I think standalone activities needs to be handled separately.

service/history/transfer_queue_standby_task_executor.go
```diff
@@ -108,6 +108,9 @@ func (t *transferQueueStandbyTaskExecutor) Execute(
 		err = t.processDeleteExecutionTask(ctx, task, false)
 	case *tasks.ChasmTask:
 		err = t.executeChasmSideEffectTransferTask(ctx, task)
+	case *tasks.CancelActivityNexusTask:
```
rkannan82: @yycptt Need help figuring out if it is OK to skip in standby.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: Done.

proto/internal/temporal/server/api/enums/v1/task.proto
```diff
@@ -59,6 +59,15 @@ enum TaskType {
 
     // A task with side effects generated by a CHASM component.
     TASK_TYPE_CHASM = 33;
+
+    // A task to notify activities via Nexus. Also see ActivityNotificationType.
+    TASK_TYPE_NOTIFY_ACTIVITY = 34;
+}
+
+// ActivityNotificationType specifies the type of notification to send to activities.
+enum ActivityNotificationType {
```
yycptt: nit: probably should have its own file

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
yycptt: > RespondActivityTaskCanceled is what the SDK sends after it has cancelled.

ah ok make sense. I was thinking about the ByID version of RespondActivityTaskCanceled api, but yeah you are right, they only get called when the activity is actually cancelled.

What about other cases like workflow termination/timeout?

service/history/tasks/notify_activity_task.go
```diff
@@ -0,0 +1,72 @@
...
+	// NotifyActivityTask is an task that notifies activities via Nexus.
+	NotifyActivityTask struct {
+		definition.WorkflowKey
+		VisibilityTimestamp time.Time
+		TaskID              int64
+
+		// NotificationType specifies the type of notification.
+		NotificationType enumsspb.ActivityNotificationType
+		// ScheduledEventIDs of activities to notify (batched by worker).
+		ScheduledEventIDs []int64
```
yycptt: I think we still need failover version here, scheduledEventID is not unique for global namespace in the case of conflict resolution.

service/history/transfer_queue_standby_task_executor.go
yycptt: yeah I think it's best effort anyway so ok to skip and won't cause workflow to stuck. Worker also won't poll standby matching I think.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: I will work on those cases along with standalone activity once we have this explicit workflow cancel completed e2e.

service/history/tasks/notify_activity_task.go
rkannan82: The code now uses task tokens instead of scheduled event IDs, and task tokens embed version info. I suppose that implicitly addresses the concern?

proto/internal/temporal/server/api/enums/v1/task.proto
rkannan82: This code is obsolete now.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -664,11 +670,56 @@ func (handler *workflowTaskCompletedHandler) handleCommandRequestCancelActivity(
...
+				handler.mutableState.GetWorkflowKey().WorkflowID,
+				handler.mutableState.GetWorkflowKey().RunID,
+				ai.ScheduledEventId,
+				ai.ActivityId,
+				ai.ActivityType.GetName(),
+				ai.Attempt,
+				nil, // Clock not needed for cancel
+				ai.Version,
+				ai.StartVersion,
+				nil,
```
yycptt, line 713: this works but I feel we are storing a lot of redundant information in the tasks (nsID, workflowID, runID, activity type etc.). Theoretically we only need ScheduledEventID + ai.Version I think.

I don't feel this is blocking but something to optimize as a follow up.

service/history/queues/metrics.go
```diff
@@ -196,6 +196,8 @@ func GetOutboundTaskTypeTagValue(
 		return prefix + "." + task.StateMachineTaskType()
 	case *tasks.ChasmTask:
 		return prefix + "." + getCHASMTaskTypeTagValue(task, chasmRegistry)
+	case *tasks.WorkerCommandsTask:
+		return prefix + ".WorkerCommands"
```
yycptt, line 200: nit: I think we are only using `.` for chasm/hsm framework tasks. maybe just prefix+task.GetType().String().

service/history/tasks/worker_commands_task.go
yycptt, line 1: Plz update priority_assigner.go and mark the task as either Priority Low or Preemptable.

service/history/queues/metrics.go
rkannan82: prefix + task.GetType().String() would give "OutboundActiveTASK_TYPE_WORKER_COMMANDS" — that doesn't seem right. Is that what you meant, or something else? \I feel the prefix OutboundActive is meaningful for all commands?

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: The dispatcher doesn't have access to mutable state — it just sends the task's commands via Nexus. The token needs to be self-contained at creation time so the dispatcher can pass it through without loading the workflow.

service/history/tasks/worker_commands_task.go
rkannan82: Done. Added to PriorityLow.

service/history/queues/metrics.go
yycptt: oh my bad, I was thinking OutboundActiveWorkerCommandsTask basically. not something critical.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
yycptt: maybe I missed something. I am thinking those worker commands can be constructed at outbound task processing time by accessing mutable state.

service/history/tasks/notify_activity_task.go
yycptt: we can discuss this in the other [thread](https://github.com/temporalio/temporal/pull/9232#discussion_r3034653340)

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
rkannan82: We explicitly chose to use task tokens here because they generalize to standalone activities (which don't have scheduled event IDs). Scheduled event IDs would require a proto change later to support that.

rkannan82: Discussed offline. Keeping it as-is and erring on the side of simplicity, while acknowledging we can optimize the key for the workflow case.

PR #9233, opened 2026-02-05
Dispatch activity cancellation to worker using Nexus

rkannan82:
## What

Dispatches worker commands (starting with activity cancellation) to workers via their Nexus control queue. When the outbound queue processes a `WorkerCommandsTask`, the dispatcher sends an `ExecuteCommands` Nexus operation to the worker's control queue via `DispatchNexusTask`. 

- Retries are capped at 3 attempts since these commands are best-effort (the activity will eventually time out anyway). Uses the retryable matching client so transient RPC errors (ShardOwnershipLost, Unavailable) are retried at the RPC layer without consuming queue-level attempts.
- Stores the clock used to generate the task token in the ActivityInfo. This is needed to reconstruct the same task token when dispatching to the worker. Otherwise, it will not match the token expected by the sdk.

Suggested review order: `worker_commands_task_dispatcher.go` → `nexus_dispatch_response.go` → `recordactivitytaskstarted/api.go`.

## Why

To support activity cancellation without activity heartbeat. This is the dispatch leg of the flow:

1. [#9231] Store `worker_control_task_queue` in `ActivityInfo` at activity start.
2. [#9232] On `RequestCancelActivityTask`, batch commands by control queue into `WorkerCommandsTask` outbound tasks.
3. **[This PR]** Dispatch each task as a Nexus `ExecuteCommands` operation to the worker, with a 3-attempt retry cap.
4. [SDK] Worker receives the cancel command and cancels the running activity.

Gated by dynamic config `EnableCancelActivityWorkerCommand` (default: off).

## How did you test it?

- **Unit tests** cover all dispatch outcomes (success, RPC error, timeout, worker error, feature-flag-off, max-attempts-exceeded) and response-to-error conversion paths.
- **Functional test** verifies end-to-end: cancel request → Nexus dispatch → correct payload arrives on the control queue, and asserts that the cancel command's task token matches the one from the original activity poll response.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -697,6 +697,10 @@ start-xdc-cluster-b: temporal-server
 start-xdc-cluster-c: temporal-server
 	./temporal-server --config-file config/development-cluster-c.yaml --allow-no-auth start
 
+start-jwt: temporal-server
+	@./config/jwt/setup-keys.sh
+	./temporal-server --config-file config/development-jwt.yaml start --service frontend --service internal-frontend --service history --service matching --service worker
+
 ##### Grafana #####
 update-dashboards:
 	@printf $(COLOR) "Update dashboards submodule from remote..."
--- a/config/development-jwt.yaml
+++ b/config/development-jwt.yaml
@@ -1,0 +1,150 @@
+log:
+  stdout: true
+  level: info
+
+persistence:
+  defaultStore: sqlite-default
+  visibilityStore: sqlite-visibility
+  numHistoryShards: 1
+  datastores:
+    sqlite-default:
+      sql:
+        user: ""
+        password: ""
+        pluginName: "sqlite"
+        databaseName: "default"
+        connectAddr: "localhost"
+        connectProtocol: "tcp"
+        connectAttributes:
+          mode: "memory"
+          cache: "private"
+        maxConns: 1
+        maxIdleConns: 1
+        maxConnLifetime: "1h"
+        tls:
+          enabled: false
+          caFile: ""
+          certFile: ""
+          keyFile: ""
+          enableHostVerification: false
+          serverName: ""
+
+    sqlite-visibility:
+      sql:
+        user: ""
+        password: ""
+        pluginName: "sqlite"
+        databaseName: "default"
+        connectAddr: "localhost"
+        connectProtocol: "tcp"
+        connectAttributes:
+          mode: "memory"
+          cache: "private"
+        maxConns: 1
+        maxIdleConns: 1
+        maxConnLifetime: "1h"
+        tls:
+          enabled: false
+          caFile: ""
+          certFile: ""
+          keyFile: ""
+          enableHostVerification: false
+          serverName: ""
+global:
+  membership:
+    maxJoinDuration: 30s
+    broadcastAddress: "127.0.0.1"
+  pprof:
+    port: 7936
+  metrics:
+    prometheus:
+      #      # specify framework to use new approach for initializing metrics and/or use opentelemetry
+      #      framework: "opentelemetry"
+      framework: "tally"
+      timerType: "histogram"
+      listenAddress: "127.0.0.1:8000"
+  authorization:
+    authorizer: "default"
+    claimMapper: "default"
+    jwtKeyProvider:
+      keySourceURIs:
+        - "file:///tmp/temporal-jwt-test/.well-known/jwks.json"
+
+services:
+  frontend:
+    rpc:
+      grpcPort: 7233
+      membershipPort: 6933
+      bindOnLocalHost: true
+      httpPort: 7243
+
+  internal-frontend:
+    rpc:
+      grpcPort: 7236
+      membershipPort: 6936
+      bindOnLocalHost: true
+
+  matching:
+    rpc:
+      grpcPort: 7235
+      membershipPort: 6935
+      bindOnLocalHost: true
+
+  history:
+    rpc:
+      grpcPort: 7234
+      membershipPort: 6934
+      bindOnLocalHost: true
+
+  worker:
+    rpc:
+      grpcPort: 7239
+      membershipPort: 6939
+      bindOnLocalHost: true
+
+clusterMetadata:
+  enableGlobalNamespace: false
+  failoverVersionIncrement: 10
+  masterClusterName: "active"
+  currentClusterName: "active"
+  clusterInformation:
+    active:
+      enabled: true
+      initialFailoverVersion: 1
+      rpcName: "frontend"
+      rpcAddress: "localhost:7233"
+      httpAddress: "localhost:7243"
+
+dcRedirectionPolicy:
+  policy: "noop"
+
+archival:
+  history:
+    state: "enabled"
+    enableRead: true
+    provider:
+      filestore:
+        fileMode: "0666"
+        dirMode: "0766"
+      gstorage:
+        credentialsPath: "/tmp/gcloud/keyfile.json"
+  visibility:
+    state: "enabled"
+    enableRead: true
+    provider:
+      filestore:
+        fileMode: "0666"
+        dirMode: "0766"
+
+namespaceDefaults:
+  archival:
+    history:
+      state: "disabled"
+      URI: "file:///tmp/temporal_archival/development"
+    visibility:
+      state: "disabled"
+      URI: "file:///tmp/temporal_vis_archival/development"
+
+dynamicConfigClient:
+  filepath: "config/dynamicconfig/development-sql.yaml"
+  pollInterval: "10s"
--- a/config/jwt/README.md
+++ b/config/jwt/README.md
@@ -1,0 +1,59 @@
+# Local JWT Development Setup
+
+This document describes how to run Temporal server locally with JWT authentication enabled for test and development purposes.
+
+## Overview
+
+The setup uses:
+- RSA key pair for signing/verifying JWTs
+- JWKS file loaded directly from disk via `file://` URI
+- Custom config file `development-jwt.yaml` with authorization enabled
+
+## Files
+
+- `config/development-jwt.yaml` - Server config with JWT auth enabled
+- `config/jwt/setup-keys.sh` - Generates RSA key pair and JWKS
+- `config/jwt/generate-token.sh` - Helper script to generate test JWTs
+
+**Generated files** (created by `setup-keys.sh` in `/tmp/temporal-jwt-test/`):
+
+- `/tmp/temporal-jwt-test/private-key.pem` - RSA private key for signing test JWTs
+- `/tmp/temporal-jwt-test/.well-known/jwks.json` - JWKS file with RSA public key
+
+## Makefile Targets
+
+- `make start-jwt` - Start Temporal with JWT auth (no `--allow-no-auth` flag)
+
+## Usage
+
+### 1. Start Temporal with JWT Auth
+
+```bash
+make start-jwt
+```
+
+Note: Unlike other `start-*` targets, this does NOT use the `--allow-no-auth` flag, so authentication is enforced.
+
+### 2. Generate Test JWTs
+
+```bash
+# Default: test-user@example.com with system:admin
+./config/jwt/generate-token.sh
+
+# Custom subject with system:admin
+./config/jwt/generate-token.sh alice@company.com
+
+# Custom subject with namespace permission
+./config/jwt/generate-token.sh alice@company.com default:admin
+
+# Multiple permissions
+./config/jwt/generate-token.sh alice@company.com system:admin default:writer
+```
+
+### 3. Use the Token
+
+```bash
+TOKEN=$(./config/jwt/generate-token.sh alice@example.com default:admin)
+temporal --tls=false --api-key "$TOKEN" workflow list
+```
+
--- a/config/jwt/generate-token.sh
+++ b/config/jwt/generate-token.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+# Generate a test JWT signed with the local private key
+# Usage: ./generate-token.sh [subject] [permissions...]
+#
+# Examples:
+#   ./generate-token.sh                                    # Default: test-user@example.com with system:admin
+#   ./generate-token.sh alice@company.com                  # Custom subject with system:admin
+#   ./generate-token.sh alice@company.com default:admin    # Custom subject with namespace permission
+#   ./generate-token.sh alice@company.com system:admin default:writer  # Multiple permissions
+#
+# The generated JWT has the following structure:
+#   Header: {"alg": "RS256", "typ": "JWT", "kid": "test-key-1"}
+#   Payload: {"sub": "<subject>", "permissions": ["<perm1>", ...], "iat": <now>, "exp": <now+1h>}
+
+set -euo pipefail
+
+base64url() { base64 | tr -d '\n' | tr '+/' '-_' | tr -d '='; }
+
+KEY_DIR="/tmp/temporal-jwt-test"
+PRIVATE_KEY="$KEY_DIR/private-key.pem"
+
+if [ ! -f "$PRIVATE_KEY" ]; then
+    echo "Private key not found at $PRIVATE_KEY, running setup-keys.sh..." >&2
+    "$(dirname "$0")/setup-keys.sh"
+fi
+
+if [ ! -f "$PRIVATE_KEY" ]; then
+    echo "Error: Private key still not found at $PRIVATE_KEY" >&2
+    exit 1
+fi
+
+SUBJECT="${1:-test-user@example.com}"
+shift 2>/dev/null || true
+
+# Collect permissions (default to system:admin if none specified)
+if [ $# -eq 0 ]; then
+    PERMISSIONS='["system:admin"]'
+else
+    PERMISSIONS=$(printf '%s\n' "$@" | jq -R . | jq -s .)
+fi
+
+NOW=$(date +%s)
+EXP=$((NOW + 3600))
+
+HEADER='{"alg":"RS256","typ":"JWT","kid":"test-key-1"}'
+HEADER_B64=$(echo -n "$HEADER" | base64url)
+
+PAYLOAD=$(jq -n -c --arg sub "$SUBJECT" --argjson perms "$PERMISSIONS" \
+    --argjson iat "$NOW" --argjson exp "$EXP" \
+    '{sub: $sub, permissions: $perms, iat: $iat, exp: $exp}')
+PAYLOAD_B64=$(echo -n "$PAYLOAD" | base64url)
+
+SIGNATURE=$(echo -n "${HEADER_B64}.${PAYLOAD_B64}" | \
+    openssl dgst -sha256 -sign "$PRIVATE_KEY" | \
+    base64url)
+
+# Output the complete JWT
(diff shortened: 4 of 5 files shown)


service/history/cancel_activity.go
```diff
@@ -0,0 +1,157 @@
...
+
+	taskTokens, err := buildActivityTaskTokens(mutableState, task)
+	if err != nil {
+		return err
+	}
+	if len(taskTokens) == 0 {
+		return nil
+	}
+
+	return t.dispatchCancelTaskToWorker(ctx, task.NamespaceID, task.WorkerControlTaskQueue, taskTokens)
```
bergundy: You definitely want to release the lock before making this RPC.

service/history/cancel_activity.go
```diff
@@ -0,0 +1,157 @@
...
+
+	nexusRequest := &nexuspb.Request{
+		Header: map[string]string{},
+		Variant: &nexuspb.Request_StartOperation{
+			StartOperation: &nexuspb.StartOperationRequest{
+				Service:   nexus.WorkerControlService,
+				Operation: nexus.CancelActivitiesOperation,
+				Payload: &commonpb.Payload{
+					Metadata: map[string][]byte{
+						"encoding": []byte("proto"),
```
bergundy: Use the SDK's data converter here instead of inventing your own proto encoding scheme please.

common/nexus/constants.go
```diff
@@ -3,3 +3,11 @@ package nexus
...
 // internally within Temporal. It must match the scheme/host used in validation and routing logic.
 const SystemCallbackURL = "temporal://system"
+
+// Worker control Nexus service and operations for server-to-worker communication.
+const (
+	// WorkerControlService is the Nexus service name for worker control operations.
+	WorkerControlService = "temporal.api.worker.v1.WorkerService"
+	// CancelActivitiesOperation is the operation to cancel running activities.
+	CancelActivitiesOperation = "cancel-activities"
+)
```
bergundy: These aren't nexus properties IMHO. It's part of the worker service, not generic for Nexus.

service/history/cancel_activity.go
```diff
@@ -0,0 +1,157 @@
...
+	if len(taskTokens) == 0 {
+		return nil
+	}
+
+	return t.dispatchCancelTaskToWorker(ctx, task.NamespaceID, task.WorkerControlTaskQueue, taskTokens)
+}
+
+// buildActivityTaskTokens builds task tokens for activities that need cancellation.
+// Returns the serialized task tokens and any error.
+func buildActivityTaskTokens(
```
bergundy: Share this function everywhere where you construct task tokens instead of reimplementing this.

service/history/cancel_activity.go
```diff
@@ -0,0 +1,157 @@
...
+		Request: nexusRequest,
+	})
+	if err != nil {
+		t.logger.Warn("Failed to dispatch activity cancel to worker",
+			tag.NewStringTag("control_queue", controlQueueName),
+			tag.Error(err))
+		return err
+	}
+
+	if resp.GetRequestTimeout() != nil {
```
bergundy: What about async outcomes and operation failures?

service/history/cancel_activity.go
bergundy: Why store the worker control task queue on the task? That seems redundant if you're already reading from mutable state, which IMHO should serve as the source of truth.

service/history/cancel_activity.go
```diff
@@ -0,0 +1,157 @@
...
+	tokenspb "go.temporal.io/server/api/token/v1"
+	"go.temporal.io/server/common"
+	"go.temporal.io/server/common/log/tag"
+	"go.temporal.io/server/common/nexus"
+	historyi "go.temporal.io/server/service/history/interfaces"
+	"go.temporal.io/server/service/history/tasks"
+	"google.golang.org/protobuf/proto"
+)
+
+func (t *transferQueueActiveTaskExecutor) processCancelActivityTask(
```
bergundy: I would put this on the outbound queue since it sync matches into workers and would have different latency characteristics than typical internal tasks.

service/history/cancel_activity.go
```diff
@@ -0,0 +1,157 @@
...
+) (retError error) {
+	if !t.config.EnableActivityCancellationNexusTask() {
+		return nil
+	}
+
+	if len(task.ScheduledEventIDs) == 0 {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, taskTimeout)
```
bergundy: What's taskTimeout? You should be deliberate about setting a reasonable timeout for the whole dispatch to the worker. That's different from dispatching tasks internally within the cluster.

service/history/cancel_activity.go
yycptt: That will an exception re. how we divided queues IMO. 

Not sure I follow the concern re. fail to sync match, if it fails the task can retry using a user error, which can be handled in the same way as namespace rate limit error today. I think this task is best effort and will be discarded after 5mins.

Maybe I missed something on the nexus sync match dispatch flow?

bergundy: IMHO this is still outbound because it blocks on a response from a worker. We would need to allow multiple seconds of timeout for this round trip.

yycptt: ah...we wait for the response! that's the part I missed. I thought DispatchNexusTask only do the dispatch part. 

yeah agree if we need to allow multiple seconds for a task, we can't use transfer queue.  cc @rkannan82 sorry I missed this when reviewing the design and pointed you in the wrong direction.

service/history/cancel_activity.go
rkannan82: Addressed — dispatch now happens in the outbound queue executor, which doesn't hold mutable state locks. `cancel_activity.go` has been removed.

service/history/cancel_activity.go
rkannan82: Addressed — now using `payload.Encode` / `payload.Decode` (Temporal's standard data converter).

common/nexus/constants.go
rkannan82: Addressed — constants removed from `common/nexus/constants.go`. Service and operation names now come from `workerservicepb.WorkerService` where they belong.

service/history/cancel_activity.go
rkannan82: Addressed in PR #9232 (task creation side) — task token construction is shared there.

service/history/cancel_activity.go
rkannan82: Addressed — `nexus_dispatch_response.go` now handles all variants: SyncSuccess, AsyncSuccess, operation Failure (via `OperationErrorFromTemporalFailure`), plus timeout and failure at the dispatch envelope level.

service/history/cancel_activity.go
rkannan82: Addressed in PR #9232 — the destination is set at task creation time from mutable state. The outbound task is self-contained so we don't need to re-read mutable state at dispatch time.

service/history/cancel_activity.go
rkannan82: Addressed — dispatch now runs on the outbound queue, not the transfer queue.

service/history/cancel_activity.go
rkannan82: Addressed — explicit `workerCommandsTaskTimeout = 10s * debug.TimeoutMultiplier` in `worker_commands_task_dispatcher.go`.

service/history/outbound_queue_active_task_executor.go
```diff
@@ -35,17 +38,26 @@ func newOutboundQueueActiveTaskExecutor(
 	logger log.Logger,
 	metricsHandler metrics.Handler,
 	chasmEngine chasm.Engine,
+	matchingRawClient resource.MatchingRawClient,
+	config *configs.Config,
```
yycptt: nit: already have access to this with `shardCtx.GetConfig()`

service/history/api/recordactivitytaskstarted/api.go
```diff
@@ -152,6 +152,7 @@ func recordActivityTaskStarted(
 		if ai.RequestId == requestID {
 			response.StartedTime = ai.StartedTime
 			response.Attempt = ai.Attempt
+			response.Clock = ai.StartedClock
```
yycptt: I guess we need to handle backward compatibility case here as well.

service/history/queues/executable.go
```diff
@@ -714,6 +715,10 @@ func (e *executableImpl) State() ctasks.State {
 	return e.state
 }
 
+func (e *executableImpl) Attempt() int {
+	return e.attempt
+}
```
yycptt, line 721: This will be very error prone because if it's called outside the task executor (e.g. in multi-cursor code or task scheduler code) it's a race condition. 

If you have to have it you'll essentially need to revert what I did in https://github.com/temporalio/temporal/pull/8732 and protect the field with a lock.

I would avoid ^ if possible and make the discard logic based on duration (with task.GetVisibilityTime()).

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -644,6 +644,18 @@ message ActivityInfo {
...
+    // Matching uses this clock to build the task token sent to the worker. Stored here so
+    // that history can later reconstruct the same task token (e.g. for cancel worker commands).
+    //
+    // IMPORTANT: The clock approach requires history to reconstruct the token using
+    // the same fields and logic as matching — if NewActivityTaskToken changes, both
+    // call sites must stay in sync or the tokens will silently diverge. An alternative
+    // is to store the full serialized task token (~150-300 bytes), which avoids
+    // reconstruction entirely and is immune to token format changes. We chose the
+    // clock approach to keep the per-activity memory footprint minimal (~24 bytes).
+    temporal.server.api.clock.v1.VectorClock started_clock = 52;
```
yycptt: so this clock value needs to be replicated as well I assume? if so please add a comment on it as normally shard clock is cluster specific and should not be replicated.

state-based replication will automatically replicate this so you don't need to do anything special here I think.  

cc @yux0 fyi

service/history/outbound_queue_active_task_executor.go
yycptt, line 1: I understand the standby task executor can simply drop the task but I don't see a change for that? Is that in a different PR?

service/history/api/recordactivitytaskstarted/api.go
rkannan82: The early-return path returns nil when StartedClock hasn't been stored yet — same behavior as before this PR. On the cancel side (handler), we added a nil check: if StartedClock is nil, we skip the cancel command and let the activity time out normally.

service/history/outbound_queue_active_task_executor.go
rkannan82: Added in this PR — the standby executor now drops WorkerCommandsTask (returns nil).

service/history/outbound_queue_active_task_executor.go
rkannan82: Done. Using shardCtx.GetConfig() instead.

proto/internal/temporal/server/api/persistence/v1/executions.proto
rkannan82: Added a comment. State-based replication covers this automatically since it's part of ActivityInfo.

service/history/queues/executable.go
rkannan82: Good point. Added a lock around the Attempt() read for safety.
The reason why I want to use attempts over duration is:
- Any slowness will eat into the budget. 
-  Also, from what i see there is no meaningful performance penalty from using this lock.

common/metrics/metric_defs.go
```diff
@@ -862,7 +841,9 @@ var (
 		WithDescription("The amount of time it took to successfully send a task to the DLQ. This only records the"+
 			" latency of the final attempt to send the task to the DLQ, not the cumulative latency of all attempts."),
 	)
-	TaskDiscarded                   = NewCounterDef("task_errors_discarded")
+	TaskDiscarded = NewCounterDef("task_errors_discarded")
+
+	WorkerCommandsSent              = NewCounterDef("worker_commands_sent")
```
bergundy: I would put this somewhere that's not in the middle of all of the task metrics.

proto/internal/temporal/server/api/persistence/v1/executions.proto
```diff
@@ -522,394 +496,410 @@ message WorkerCommandsTask {
...
+
+    // set to true if reset heartbeat flag was set with an activity reset
+    bool reset_heartbeats = 48;
+
+    int64 start_version = 50;
+
+    // The task queue on which the server will send control tasks to the worker running this activity.
+    string worker_control_task_queue = 51;
+
+    // The shard clock at the time this activity was started (RecordActivityTaskStarted).
```
bergundy: Do we still need the vector clock now that we are using a stamp for validation?

service/history/queues/executable.go
rkannan82: Decided to add the thread safe API in separate PR: https://github.com/temporalio/temporal/pull/9924

rkannan82:
> Can you confirm whether you're covering cancel requests and pause requests? Do we also care about canceling activities that we know timed out or are we letting the worker take care of that? When will you be adding support for standalone activities too?

1. Cancel only, no pause yet. The WorkerCommand proto only has CancelActivity today — pause will be a separate oneof variant added later.

2. Timeouts/termination: Separate PR once we get the base case working e2e .
Context:For scenarios where the server proactively cancels activities (termination, timeout), we will cancel all in-flight activities by default. To let users opt out per-activity, we will introduce an ActivityClosePolicy (analogous to ParentClosePolicy for child workflows), defined in ScheduleActivityTaskCommandAttributes and persisted in ActivityInfo, with values REQUEST_CANCEL (default) and ABANDON.

service/history/api/recordactivitytaskstarted/api.go
yycptt: Not about the cancellation side, but for normal respondActivityCompleted path.

Existing logic will have clock always populated (in history/handler.go) in the task token, based on which, respondActivityCompleted will do a staleness check on the shard. With the change, this early return path will no longer have the clock in response.

service/history/nexus_dispatch_response.go
```diff
@@ -0,0 +1,84 @@
...
+	nexuspb "go.temporal.io/api/nexus/v1"
+	"go.temporal.io/server/api/matchingservice/v1"
+	commonnexus "go.temporal.io/server/common/nexus"
+	"go.temporal.io/server/common/nexus/nexusrpc"
+)
+
+// convertTemporalFailure converts a Temporal API Failure proto into a Go error
+// via the Nexus SDK failure converter.
+func convertTemporalFailure(failure *failurepb.Failure) (nexusErr error, err error) {
+	nexusFailure, err := commonnexus.TemporalFailureToNexusFailure(failure)
```
bergundy: I'm not sure this is required. You can use the Temporal Go SDK's failure converter to construct handler errors directly. Not much of a point to translate to a Nexus failure.

service/history/nexus_dispatch_response.go
```diff
@@ -0,0 +1,84 @@
...
+	// Operation processed but failed — the worker returned an explicit failure.
+	case *nexuspb.StartOperationResponse_Failure:
+		return operationErrorFromFailure(t.Failure)
+	default:
+		return nexus.NewHandlerErrorf(nexus.HandlerErrorTypeInternal, "empty or unknown start operation response variant")
+	}
+}
+
+// operationErrorFromFailure converts a Temporal API Failure into a Nexus SDK operation error.
+func operationErrorFromFailure(failure *failurepb.Failure) error {
```
bergundy: Same here, the SDK already sends back a temporal Failure object that you can construct an error from using the Temporal Go SDK's default failure converter.

service/history/outbound_queue_factory.go
```diff
@@ -229,6 +231,7 @@ func (f *outboundQueueFactory) CreateQueue(
 		logger,
 		metricsHandler,
 		f.ChasmEngine,
+		f.MatchingRawClient,
```
bergundy: Is the "raw" client the right client to use here? I think you want the other client with retries and all.

service/history/nexus_dispatch_response.go
bergundy: You just won't get back an operation error, you'll get an ApplicationError or a CanceledError. That's fine IMHO.

service/history/worker_commands_task_dispatcher.go
```diff
@@ -0,0 +1,205 @@
...
+	nexusErr := dispatchResponseToError(resp)
+	if nexusErr == nil {
+		metrics.WorkerCommandsSent.With(d.metricsHandler).Record(1, metrics.OutcomeTag("success"))
+		return nil
+	}
+
+	return d.handleError(nexusErr, task)
+}
+
+func (d *workerCommandsTaskDispatcher) handleError(nexusErr error, task *tasks.WorkerCommandsTask) error {
```
bergundy, line 168: Check `handlerErr.Retryable()` to decide whether to retry the error.

service/history/outbound_queue_factory.go
yycptt: All other tasks executors also use raw clients and let task processing side controling the retry/throttling etc.

service/history/api/recordactivitytaskstarted/api.go
rkannan82: Ah very good catch! I have fixed it and also added a test. I am assuming it is ok to create this vector clock while holding the workflow lock.

proto/internal/temporal/server/api/persistence/v1/executions.proto
yycptt: hmm I don't see stamp being used in e.g. respondActivityTaskCompleted. 

with that said I **_think_** shard clock is mostly an optimization for activity/workflow task related apis at this point. It will unload entire shard when check fails so other requests doesn't need to do unnecessary mutable state reload. Correctness wise I **_feel_** nothing will break even if we don't have it. but will need to do a more thorough review. 

At least I won't touch it in this PR. and I guess sdk side needs task token to match exactly for the cancellation?

service/history/outbound_queue_factory.go
rkannan82: Switched to retryable_client. Talked to Yichao and we agree this is better.

service/history/nexus_dispatch_response.go
rkannan82: Please see if the updated code is what you intended.

service/history/nexus_dispatch_response.go
rkannan82: Please see if the updated code is what you intended.

service/history/outbound_queue_factory.go
yycptt: fwiw, I think nothing will break given outbound queue has different threshold for task processing latency and separate goroutines for each destination. (actually now that I think about it may impact OPS throttling a bit because ops is done on task level and has no idea about that retry, but probably fine)

I am not sure about the "better" part.

rkannan82:
@copilot resolve the merge conflicts in this pull request

rkannan82:
Dummy comment to trigger license check.

PR #9238, opened 2026-02-06
Fix unload versioned queue with redirected backlog

dnr:
## What changed?
If we want to unload just one versioned queue, but it has backlog tasks in its matcher from another queue (e.g. the default), we should send those tasks back to be reprocessed.

## Why?
Tasks would get temporarily dropped until the default queue was reloaded.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)
- [ ] added new functional test(s) - will add functional test

### DIFF
--- a/service/matching/fair_task_reader.go
+++ b/service/matching/fair_task_reader.go
@@ -303,7 +303,9 @@ func (tr *fairTaskReader) addErrorBehavior(err error) (drop, retry bool) {
 	// - versioning wants to re-spool the task on a different queue and that failed
 	// - versioning says StickyWorkerUnavailable
 	if errors.Is(err, errTaskQueueClosed) || common.IsContextCanceledErr(err) {
-		return false, false
+		// maybe we tried to add a task to a versioned queue as it was unloading, and have to
+		// retry here. if tqCtx is closing, addTaskToMatcher will give up.
+		return false, true
 	}
 	var stickyUnavailable *serviceerrors.StickyWorkerUnavailable
 	if errors.As(err, &stickyUnavailable) {
--- a/service/matching/pri_task_reader.go
+++ b/service/matching/pri_task_reader.go
@@ -299,7 +299,7 @@ func (tr *priTaskReader) addTaskToMatcher(task *internalTask) {
 
 	if drop, retry := tr.addErrorBehavior(err); drop {
 		task.finish(nil, false)
-	} else if retry && tr.backlogMgr.tqCtx.Err() == nil {
+	} else if retry {
 		// This should only be due to persistence problems. Retry in a new goroutine
 		// to not block other tasks, up to some concurrency limit.
 		if tr.addRetries.Acquire(tr.backlogMgr.tqCtx, 1) != nil {
@@ -317,8 +317,8 @@ func (tr *priTaskReader) addErrorBehavior(err error) (drop, retry bool) {
 	// - versioning wants to re-spool the task on a different queue and that failed
 	// - versioning says StickyWorkerUnavailable
 	if errors.Is(err, errTaskQueueClosed) || common.IsContextCanceledErr(err) {
-		// if tqCtx is closing, addTaskToMatcher will give up, otherwise we can retry and we'll
-		// reload a versioned queue
+		// maybe we tried to add a task to a versioned queue as it was unloading, and have to
+		// retry here. if tqCtx is closing, addTaskToMatcher will give up.
 		return false, true
 	}
 	var stickyUnavailable *serviceerrors.StickyWorkerUnavailable


service/matching/task_queue_partition_manager.go
```diff
@@ -652,6 +652,9 @@ func (pm *taskQueuePartitionManagerImpl) AddSpooledTask(
 		// make sure to reset redirectInfo in case it was set in a previous loop cycle
 		task.redirectInfo = nil
 	}
+	// mark if task is being redirected from default queue to a deployment queue (V3 versioning)
+	task.redirectedToDeployment = backlogQueue.version.Deployment() == nil &&
+		syncMatchQueue.QueueKey().Version().Deployment() != nil
```
ShahabT: Would something like this not work? (supports V2 and V3)

```suggestion
	task.redirected = backlogQueue != syncMatchQueue
```

service/matching/matcher_data.go
```diff
@@ -388,6 +388,19 @@ func (d *matcherData) ReprocessTasks(pred func(*internalTask) bool) []*internalT
...
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	for _, task := range d.tasks.heap {
+		// V2 uses redirectInfo, V3 uses redirectedToDeployment
+		if task.redirectInfo != nil || task.redirectedToDeployment {
+			return true
+		}
+	}
+	return false
```
ShahabT: we should protect against the race that the task comes after this decision, before the actual unload

service/matching/task_queue_partition_manager.go
ShahabT: maybe it's best to even put the backlogQueue in the internal task (as opposed to a bool `redirected` field) and do the equals check in physical queue itself.

dnr: I wrote in the TODO above that I want to consolidate this.. several fields are just references to things in either the source or dest queue and we can just store them directly. but it's a larger refactoring and I'd rather keep this minimal

service/matching/pri_matcher.go
```diff
@@ -153,6 +153,13 @@ func (tm *priTaskMatcher) Start() {
 
 func (tm *priTaskMatcher) Stop() {
 	tm.priorityBacklogForwarders.Sync(nil, nil)
+
+	// When we're stopping, sync tasks and pollers will be cancelled by tqCtx being canceled.
+	// Backlog tasks held in this matcher will be dropped. That's okay if we're stopping the
+	// whole partition, or for tasks that came from this partition's readers. The exception is
+	// backlog tasks that were redirected from another versioned queue (or the default). To
+	// handle those, the caller of Stop should also call ReprocessRedirectedTasksAfterStop
+	// when applicable.
```
yiminc, line 165: why not just call ReprocessRedirectedTasksAfterStop() here?

dnr: 1. Because I want it to happen after the whole physicalTaskQueueManagerImpl.Stop() returns, not just priTaskMatcher.Stop().
2. Because I don't want it to happen at all when we're unloading the whole partition at once, only when we're unloading a single versioned queue.

Actually I do think that eventually we should do "it" here, where "it" is a more general version of ReprocessRedirectedTasksAfterStop that kicks _all_ backlog tasks back to their readers. To do that, we need to change the order of the matcher.Stop and tqCtxCancel in ptqm.Stop so the context is canceled. I wanted to make a more selective change first.

service/matching/matcher_data.go
```diff
@@ -235,13 +237,25 @@ func newMatcherData(config *taskQueueConfig, logger log.Logger, timeSource clock
...
 	defer d.lock.Unlock()
 
+	d.stopped = true
+}
+
+func (d *matcherData) EnqueueTaskNoWait(task *internalTask) error {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	if d.stopped {
```
rkannan82, line 251: If we didn't add this code, will the following code still return some error -- since the queue is unloaded?
If so, is this new error only for clarity

service/matching/fair_task_reader.go
```diff
@@ -316,7 +303,9 @@ func (tr *fairTaskReader) addErrorBehavior(err error) (drop, retry bool) {
 	// - versioning wants to re-spool the task on a different queue and that failed
 	// - versioning says StickyWorkerUnavailable
 	if errors.Is(err, errTaskQueueClosed) || common.IsContextCanceledErr(err) {
-		return false, false
+		// maybe we tried to add a task to a versioned queue as it was unloading, and have to
+		// retry here. if tqCtx is closing, addTaskToMatcher will give up.
+		return false, true
```
ShahabT, line 308: is it ok to retry if the task queue as a whole is closed?

service/matching/matcher_data.go
dnr: matcherData is a low-level data structure. It doesn't know anything about "the queue" or "unloaded". Well, now it does. But besides this line, no.

Unfortunately I think we need to push this one bit down to this level so that "Stop" and "Pull out redirected tasks so we can reprocess them" can be atomic w.r.t. each other. Otherwise we could still have a race between someone adding a (redirected) task to the matcher and someone stopping the matcher.

service/matching/fair_task_reader.go
dnr: I agree this is the riskiest part of the change, but note that if tqCtx (i.e. this versioned queue) is canceled, then the Acquire on line 291 will fail and it'll drop the task, which is what we want (what would have happened if we still returned false)

service/matching/matcher_data.go
rkannan82: After re-reading, I think this is critical. We want to stop adding tasks if the queue is already stopped. Otherwise, we may end up with a lost task if this call happens after ReprocessRedirectedTasksAfterStop.

service/matching/pri_task_reader.go
```diff
@@ -317,7 +317,9 @@ func (tr *priTaskReader) addErrorBehavior(err error) (drop, retry bool) {
 	// - versioning wants to re-spool the task on a different queue and that failed
 	// - versioning says StickyWorkerUnavailable
 	if errors.Is(err, errTaskQueueClosed) || common.IsContextCanceledErr(err) {
-		return false, false
+		// if tqCtx is closing, addTaskToMatcher will give up, otherwise we can retry and we'll
```
rkannan82, line 320: I didn't understand what this change is about.

service/matching/matcher_data.go
dnr: Well, the main fix is calling ReprocessRedirectedTasksAfterStop. This part closes a tiny tiny race window surrounding that call.

service/matching/pri_task_reader.go
rkannan82: As in, why was it non retryable earlier?

dnr: This is the other half of `if d.stopped { return errMatcherStopped }`: this causes priTaskReader to handle that errMatcherStopped (which is a Canceled) by sleeping for 1s and reprocessing the task.

rkannan82 (approved):

service/matching/pri_task_reader.go
dnr: > As in, why was it non retryable earlier?

I think I just assumed that any Canceled indicated that the whole partition was unloading, but that's not necessarily true, I didn't think through the partial unload case carefully (hence this whole PR)

PR #9239, opened 2026-02-06
Set TargetVersionChanged instead of SuggestCaN when TargetVersionChanged

carlydf:
## What changed?
Set TargetVersionChanged instead of SuggestCaN when TargetVersionChanged
https://github.com/temporalio/api/pull/709

## Why?
Setting SuggestContinueAsNew=true for Pinned workflows whenever their is a new Target Version available for that workflow causes Pinned workflows to hit that condition much more frequently than they expect. Users who are currently doing: if workflow_info.suggestContinueAsNew{ do continue-as-new } in their Pinned workflow code would need to change that code to protect themselves from running into an infinite-CaN-loop, because the default CaN behavior for a Pinned workflow is to stay Pinned.

We should not force users to protect themselves from such a situation.

Because upgrading on continue-as-new is opt-in, receiving the suggestion to continue-as-new-onto-new-target-version should be opt-in as well. If people are forced to check the new suggest-continue-as-new-reasons field to "opt out," that is unsafe, because inevitably some people will forget to do so or misunderstand, and then get hit by this unexpected footgun.

Much safer and still ergonomical to let upgrade-on-can be opt-in on both fronts, as proposed here. With this change, the people who are currently doing if workflow_info.suggestContinueAsNew{ do continue-as-new } won't see any change in semantics, regardless of their versioning behavior.

People who consciously know that they want to do upgrade-on-can / Trampolining will have to change their CaN options anyway, so it's easy enough to teach them to pay attention to this new TargetWorkerDeploymentVersionChanged flag.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
None

---

> [!NOTE]
> **Medium Risk**
> Touches workflow task started event generation/persistence and versioning-related signaling, which can affect worker behavior and history compatibility; changes are gated by dynamic config and covered by tests.
> 
> **Overview**
> Stops using `SuggestContinueAsNew` (and its reason tags) to signal pinned workflows that a newer target worker deployment version exists, and instead introduces an explicit `TargetWorkerDeploymentVersionChanged` boolean on `WorkflowTaskStarted` events and persisted `WorkflowExecutionInfo`.
> 
> Adds namespace dynamic config `EnableSendTargetVersionChanged` (default on) and a new metric `workflow_target_version_changed_count` emitted when this flag is set; updates the workflow task state machine, mutable state plumbing/mocks, proto/pb persistence, and functional tests accordingly. Also bumps `go.temporal.io/api` to pick up the new event attribute.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3886491826188826a2d261c7d7bf6b6a7a3250bb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/history/workflow/workflow_task_state_machine.go
```diff
@@ -491,17 +494,22 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskStartedEvent(
 	// checking whether targetDeploymentVersion == nil means that we won't send the CaN recommendation to workflows
 	// that are about to transition to the target version. This is good, because if their transition succeeds, they
 	// don't need to CaN to start using the new version.
-	if m.ms.config.EnableSuggestCaNOnNewTargetVersion(m.ms.namespaceEntry.Name().String()) &&
```
Shivs11, line 494: I think we should still keep the dynamic config present - would be great to just flick the switch if something were to go wrong somewhere down the line.

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -491,17 +494,22 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskStartedEvent(
...
-		m.ms.GetEffectiveVersioningBehavior() != enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED &&
+	var targetDeploymentVersionChanged bool
+	if m.ms.GetEffectiveVersioningBehavior() != enumspb.VERSIONING_BEHAVIOR_UNSPECIFIED &&
 		targetDeploymentVersion != nil {
 		if currentDeploymentVersion := m.ms.GetEffectiveDeployment(); currentDeploymentVersion != nil &&
 			(currentDeploymentVersion.BuildId != targetDeploymentVersion.BuildId ||
 				currentDeploymentVersion.SeriesName != targetDeploymentVersion.DeploymentName) {
```
Shivs11, line 503: I think we should use the .getter methods here and make things nil safe

service/history/workflow/workflow_task_state_machine.go
```diff
@@ -491,17 +494,22 @@ func (m *workflowTaskStateMachine) AddWorkflowTaskStartedEvent(
...
-			suggestContinueAsNew = cmp.Or(suggestContinueAsNew, true)
-			suggestContinueAsNewReasons = append(suggestContinueAsNewReasons, enumspb.SUGGEST_CONTINUE_AS_NEW_REASON_TARGET_WORKER_DEPLOYMENT_VERSION_CHANGED)
+			targetDeploymentVersionChanged = true
```
Shivs11, line 504: should we not be checking if the workflow's effective behaviour is pinned or not?

Shivs11: actually, thought about it a bit more and realized that while there is no immediate harm in having this be set to true for workflows of any versioning behaviour type. However, gonna keep this comment here just to think about if we want this flag to show up and only be used by folks who are authoring pinned workflows.

service/history/workflow/workflow_task_state_machine.go
carlydf: made a new config!

service/history/workflow/mutable_state_impl.go
```diff
@@ -3251,10 +3251,11 @@ func (ms *MutableStateImpl) ApplyWorkflowTaskStartedEvent(
 	versioningStamp *commonpb.WorkerVersionStamp,
 	redirectCounter int64,
 	suggestContinueAsNewReasons []enumspb.SuggestContinueAsNewReason,
+	targetWorkerDeploymentVersionChanged bool,
 ) (*historyi.WorkflowTaskInfo, error) {
 	return ms.workflowTaskManager.ApplyWorkflowTaskStartedEvent(workflowTask, version, scheduledEventID,
 		startedEventID, requestID, timestamp, suggestContinueAsNew, historySizeBytes, versioningStamp, redirectCounter,
-		suggestContinueAsNewReasons)
+		suggestContinueAsNewReasons, targetWorkerDeploymentVersionChanged)
```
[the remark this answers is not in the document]
carlydf: resolved

PR #9240, opened 2026-02-06
History: Expose and call SetContextMetadata in readonly operations

nikki-dag:
## What changed?
1. Exposed metadata functionality:
    - Made `SetContextMetadata()` public in MutableState interface
    - Added constants for metadata keys (`MetadataKeyWorkflowType`, `MetadataKeyWorkflowTaskQueue`)
    - Added `ContextHasMetadata()` helper function to check if context has metadata support
  2. Fixed readonly operations:
    - `QueryWorkflow`: Now explicitly calls `SetContextMetadata()` since queries never close transactions (readonly operation)
    - `RespondWorkflowTaskCompleted`: Calls `SetContextMetadata()` when there are only messages (e.g., update rejections) and no commands, since these also don't close transactions
  3. Added tests for the new `ContextHasMetadata()` helper and `SetContextMetadata()` functionality

## Why?
Context metadata (workflow type, task queue) is automatically populated when mutable state transactions close. However, readonly operations never close transactions, so the metadata was missing from their contexts.
This change ensures that successful readonly operations (queries, update rejections) also have workflow metadata populated in their contexts

## How did you test it?
- [ ] built
- [x] added new unit test(s)

---

> [!NOTE]
> **Medium Risk**
> Touches core history API request paths and mutable state interfaces; while behavior is additive (context tagging), incorrect invocation or key usage could affect observability/metrics and tests across multiple workflows.
> 
> **Overview**
> Ensures workflow context metadata (workflow type/task queue) is consistently populated even for readonly paths that don’t close mutable-state transactions.
> 
> This exposes `MutableState.SetContextMetadata(ctx)` (renaming the internal helper), standardizes metadata keys via `contextutil.MetadataKeyWorkflowType`/`MetadataKeyWorkflowTaskQueue`, adds `ContextHasMetadata()` for debugging, and explicitly calls `SetContextMetadata` in `QueryWorkflow` and in `RespondWorkflowTaskCompleted` when `Commands` is empty (e.g., update rejections / heartbeats). Tests are expanded to assert metadata presence across success and several error/edge cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 69c3140233241c2072e3bacff7580debe91ed884. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/api/respondworkflowtaskcompleted/api.go
+++ b/service/history/api/respondworkflowtaskcompleted/api.go
@@ -188,7 +188,6 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 			}
 		}
 
-
 		workflowLease.GetReleaseFn()(errForRelease)
 	}()
 
--- a/service/history/history_engine_test.go
+++ b/service/history/history_engine_test.go
@@ -672,6 +672,10 @@ func (s *engineSuite) TestQueryWorkflow_WorkflowTaskDispatch_Timeout() {
 
 	wg := &sync.WaitGroup{}
 	wg.Add(1)
+	var capturedWorkflowType interface{}
+	var capturedWorkflowTypeOk bool
+	var capturedTaskQueue interface{}
+	var capturedTaskQueueOk bool
 	go func() {
 		metadataCtx := contextutil.WithMetadataContext(context.Background())
 		ctx, cancel := context.WithTimeout(metadataCtx, time.Second*2)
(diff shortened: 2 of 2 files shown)


service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -180,6 +180,13 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 			}
 		}
 
+		if len(request.Commands) == 0 && len(request.Messages) > 0 {
```
yycptt: hmm... do we have any action metric to emit if both are empty? Is workflow task heartbeat an action?

nikki-dag: Thanks for catching this!

service/history/api/queryworkflow/api_test.go
yycptt, line 1: Is it possible to change the existing tests in history_engine_tests.go?

service/history/api/respondworkflowtaskcompleted/api.go
```diff
@@ -180,6 +180,13 @@ func (handler *WorkflowTaskCompletedHandler) Invoke(
 			}
 		}
 
+		if len(request.Commands) == 0 {
+			// Context metadata is automatically set during mutable state transaction close. For RespondWorkflowTaskCompleted
+			// with no commands (e.g., workflow task heartbeat or only readonly messages like `update.Rejection`), the transaction
+			// is never closed. We explicitly call SetContextMetadata here to ensure workflow metadata is populated in the context.
+			workflowLease.GetMutableState().SetContextMetadata(ctx)
```
yycptt: fwiw, I don't think the mutable state you get from `workflowLease.GetMutableState()` will always be the latest mutable state because it's a separate copy in this workflowLease object, while the real one is in workflowLease.Context. Doesn't really matter for your case the can be confusing? 

Why not use the `ms` variable instead and do it before the defer() method? similar to what you have in the Query api

yycptt (approved):

PR #9241, opened 2026-02-06
History: Expose and call SetContextMetadata through GetReleaseFn

nikki-dag:
## What changed?
1. Exposed metadata functionality:
    - Made `SetContextMetadata()` public in MutableState interface
    - Added constants for metadata keys (`MetadataKeyWorkflowType`, `MetadataKeyWorkflowTaskQueue`)
    - Added `ContextHasMetadata()` helper function to check if context has metadata support
2. Modified `GetReleaseFn` signature: Changed from `GetReleaseFn()` to `GetReleaseFn(ctx context.Context)` to accept a context parameter
3.  Metadata setting moved: The workflow context metadata (workflow type and task queue) is now set when `GetReleaseFn` is called, rather than during `closeTransaction`
4. Added tests for the new `ContextHasMetadata()` helper and `SetContextMetadata()` functionality

## Why?
This change ensures that workflow metadata (workflow type, task queue) is available in the context for observability purposes. By setting metadata when the release function is obtained rather than during transaction closure, we guarantee that metadata is set for both readonly and write operations.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks

---

> [!NOTE]
> **Medium Risk**
> Touches a widely-used locking/release API across history code paths; incorrect context wiring could lead to missing metadata or unexpected side effects during lease release, though functional behavior should otherwise be unchanged.
> 
> **Overview**
> Ensures workflow metadata (workflow type and task queue) is available for observability by introducing context metadata key constants and a `ContextHasMetadata` helper, plus a new `MutableState.SetContextMetadata(ctx)` API.
> 
> Refactors workflow lease/workflow interfaces so `GetReleaseFn` now requires a `context.Context`; `GetReleaseFn(ctx)` populates context metadata via `SetContextMetadata` before returning the release closure, and all history API/NDC/replication call sites and mocks are updated accordingly. Metadata population is removed from `MutableStateImpl.closeTransaction` so read-only code paths also get metadata.
> 
> Adds/updates unit tests for `ContextHasMetadata`, `SetContextMetadata`, and adjusts mocks to expect the new `GetReleaseFn(ctx)` and `SetContextMetadata` interactions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de30d9b2960a2915e026d958c067e2dfff9f451c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/service/history/api/recordchildworkflowcompleted/api_test.go
+++ b/service/history/api/recordchildworkflowcompleted/api_test.go
@@ -88,10 +88,10 @@ func Test_Recordchildworkflowcompleted_WithForwards(t *testing.T) {
 
 	oldParentWFLease := ndc.NewMockWorkflow(ctrl)
 	oldParentWFLease.EXPECT().GetMutableState().Return(oldParentMutableState) // old parent's mutable state is accesses just once.
-	oldParentWFLease.EXPECT().GetReleaseFn().Return(func(_ error) {})
+	oldParentWFLease.EXPECT().GetReleaseFn(gomock.Any()).Return(func(_ error) {})
 	newParentWFLease := ndc.NewMockWorkflow(ctrl)
 	newParentWFLease.EXPECT().GetMutableState().Return(newParentMutableState).AnyTimes() // new parent's mutable state would be accessed many times.
-	newParentWFLease.EXPECT().GetReleaseFn().Return(func(_ error) {})
+	newParentWFLease.EXPECT().GetReleaseFn(gomock.Any()).Return(func(_ error) {})
 	newParentWFLease.EXPECT().GetContext().Return(mockWFContext)
 
 	consistencyChecker := api.NewMockWorkflowConsistencyChecker(ctrl)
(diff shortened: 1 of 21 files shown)


service/history/api/workflow_lease.go
```diff
@@ -60,6 +62,9 @@ func (w *workflowLease) GetMutableState() historyi.MutableState {
 	return w.mutableState
 }
 
-func (w *workflowLease) GetReleaseFn() historyi.ReleaseWorkflowContextFunc {
+func (w *workflowLease) GetReleaseFn(ctx context.Context) historyi.ReleaseWorkflowContextFunc {
+	if w.mutableState != nil {
+		w.mutableState.SetContextMetadata(ctx)
+	}
```
nikki-dag, line 68: This introduces a side effect in a getter function. I am happy to rename the function to something like `BuildReleaseFn` to reflect this.

yycptt: This feels very error prone and easy to break to me as there's no guarantee that GetReleaseFn() is called after all the mutations at the very end of a transaction. 

The api logic can call GetReleaseFn() at the beginning and directly use the function in defer(). Though i understand today almost all api handler call GetReleaseFn() inside the defer. 

Ideally I think it should be part of the actual releaseFn itself (I am going to add it anyway in the future for other projects), then you have guarantee that it's only called at the end of the transaction. But this approach has it's own issues. The MutableContext field inside workflow context (not this workflowLease wrapper) is not populated in StartWorkflow case I believe.

nikki-dag: Makes sense to me. Maybe we can revisit this approach (moving `SetContextMetadata` from closeTransaction to the actual releaseFn) later and clean up the read-only–request–specific logic in the other PR (if we decide to go that way).

PR #9242, opened 2026-02-06
Consolidate task dispatch latency metric

dnr:
## What changed?
Move task dispatch latency metric recording to matchingEngine.

## Why?
1. Fixes a bug where dispatch latency wasn't emitted for backlog tasks from pri matcher.
2. Fixes a bug where RecordTaskStarted duration wasn't included in dispatch latency for backlog tasks in old matcher.
3. Makes bugs like that less likely to happen since there's fewer and more obvious places to emit it.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)


service/matching/task.go
```diff
@@ -347,3 +351,22 @@ func (task *internalTask) finishInternal(res taskResponse, wasValid bool) {
...
+func (task *internalTask) emitDispatchLatency() {
+	if task.dispatchMetricsHandler == nil {
+		return // not set (test or special case)
+	}
+	// Only workflow/activity tasks have event.Data.CreateTime.
+	// Query and Nexus tasks are sync-match only and don't track dispatch latency.
+	if task.event == nil || task.event.Data.GetCreateTime() == nil {
+		return
+	}
+	metrics.TaskDispatchLatencyPerTaskQueue.With(task.dispatchMetricsHandler).Record(
```
ShahabT, line 366: so the `worker_version` tag will still show matcher's version which is poller's version, right?

PR #9243, opened 2026-02-06
Enable Eager Activities by default

Shivs11:
## What changed?
- WISOTT

## Why?
- functionality

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
- none, afaik


bergundy (commented):
Missing also the override in `service/history/tests/vars.go`. I am imagining that the versioning tests that mention eager should be renamed since this is the default and I am not sure whether y'all want coverage for non eager in those tests.

PR #9246, opened 2026-02-06
Fix HistoryScannerVerifyRetention config description

prathyushpv:
## What changed?
Fix HistoryScannerVerifyRetention config description.

## Why?
Old description was no longer true since archival now happens at workflow close time.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -35,23 +35,41 @@ jobs:
       github_timeout: ${{ steps.generate_output.outputs.github_timeout }}
       test_timeout: ${{ steps.generate_output.outputs.test_timeout }}
       runs_on: ${{ steps.generate_output.outputs.runs_on }}
+      runner_x64: ${{ steps.generate_output.outputs.runner_x64 }}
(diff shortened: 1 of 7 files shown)


common/dynamicconfig/constants.go
```diff
@@ -2951,8 +2951,8 @@ because executions scanner support for SQL is not yet implemented.`,
 	HistoryScannerVerifyRetention = NewGlobalBoolSetting(
 		"worker.historyScannerVerifyRetention",
 		true,
-		`HistoryScannerVerifyRetention indicates the history scanner verify data retention.
-If the service configures with archival feature enabled, update worker.historyScannerVerifyRetention to be double of the data retention.`,
+		`HistoryScannerVerifyRetention indicates if the history scavenger should verify data retention.
+When enabled, the scavenger will delete completed workflow execution data that has aged past the namespace retention period plus worker.executionDataDurationBuffer.`,
```
stephanos: "has aged past" sound a bit clunky to me; "older than"?

prathyushpv: Yeah! Will change..

PR #9247, opened 2026-02-06
Nexus endpoint rate limiter

prathyushpv:
## What changed?
Added a per-caller per-endpoint rate limiter for Nexus dispatch requests. A new dynamic config frontend.nexusEndpointRPS (type DestinationIntSetting, default 0 = unlimited) controls the limit.

## Why?
To control the total traffic processed by an endpoint from a caller. Existing rate limiters operate at namespace or global level and don't account for the caller level traffic on endpoints.

## How did you test it?
- [x] built
- [x] added new unit test(s)

### DIFF
--- a/service/frontend/nexus_handler.go
+++ b/service/frontend/nexus_handler.go
@@ -234,11 +234,6 @@ func (c *operationContext) interceptRequest(
 		return commonnexus.ConvertGRPCError(err, false)
 	}
 
-	if err := c.namespaceRateLimitInterceptor.Allow(c.namespace.Name(), c.apiName, header); err != nil {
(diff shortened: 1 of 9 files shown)


service/frontend/nexus_handler.go
```diff
@@ -226,6 +234,18 @@ func (c *operationContext) interceptRequest(
 		return commonnexus.ConvertGRPCError(err, false)
 	}
 
+	if c.endpointName != "" && c.endpointRateLimiter != nil {
+		if !c.endpointRateLimiter.Allow(time.Now().UTC(), quotas.Request{
+			API:         c.apiName,
+			Token:       1,
+			Caller:      c.namespaceName,
```
prathyushpv, line 241: I just realized this is not the caller namespace. Will have to fix this.

PR #9249, opened 2026-02-06
Add a success report to track improvements to our CI

spkane31:
## What changed?
WISOTT

## Why?
Help reduce the friction in our CI

## How did you test it?
- [ ] built
- [X] added new unit test(s)

## Potential risks
No

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -5,7 +5,7 @@ permissions:
   contents: read
 jobs:
   lint-actions:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -24,7 +24,7 @@ jobs:
         shell: bash
 
   lint-protos:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -44,7 +44,7 @@ jobs:
           make lint-protos
 
   lint-api:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -64,7 +64,7 @@ jobs:
           make lint-api
 
   lint-workflows:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -83,7 +83,7 @@ jobs:
         run: make workflowcheck
 
   fmt-imports:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -108,7 +108,7 @@ jobs:
           fi
 
   lint-yaml:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -123,7 +123,7 @@ jobs:
         run: make lint-yaml
 
   golangci:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@v6
         with:
@@ -156,7 +156,7 @@ jobs:
       - fmt-imports
       - lint-yaml
       - golangci
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     if: always()
     env:
       RESULTS: ${{ toJSON(needs.*.result) }}
--- a/.github/workflows/optimize-test-sharding.yml
+++ b/.github/workflows/optimize-test-sharding.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch: # Allow manual trigger
 
 permissions:
-  actions: read
+  actions: write
 
 jobs:
   update-salts:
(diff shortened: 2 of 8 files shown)


tools/ci-notify/app.go
```diff
@@ -14,43 +14,86 @@ const (
...
-			Required: false, // Not required for dry-run
-			EnvVars:  []string{"SLACK_WEBHOOK"},
+		{
+			Name:  "report",
+			Usage: "Generate report for CI runs",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     FlagBranch,
+					Usage:    "Branch to generate report for",
+					Required: true,
```
stephanos: Should this default to `main` instead?

stephanos: same for `failure` command

tools/ci-notify/app.go
```diff
@@ -60,11 +103,13 @@ func run(c *cli.Context) error {
 	defer func() { _ = logger.Sync() }()
 
 	runID := c.String(FlagRunID)
+	branch := c.String(FlagBranch)
 	slackWebhook := c.String(FlagSlackWebhook)
 	dryRun := c.Bool(FlagDryRun)
 
-	logger.Info("Starting ci-notify",
+	logger.Info("Starting ci-notify failure",
```
stephanos: Reads a little funny; how about sth like `starting ci-notify:failure` or some other formatting?

tools/ci-notify/app.go
```diff
@@ -60,11 +103,13 @@ func run(c *cli.Context) error {
 	defer func() { _ = logger.Sync() }()
 
 	runID := c.String(FlagRunID)
+	branch := c.String(FlagBranch)
```
stephanos: This is only used for logging?

tools/ci-notify/app.go
```diff
@@ -14,43 +14,86 @@ const (
...
+			},
+			Action: runFailureCommand,
 		},
-		&cli.StringFlag{
-			Name:     FlagSlackWebhook,
-			Usage:    "Slack webhook URL",
-			Required: false, // Not required for dry-run
-			EnvVars:  []string{"SLACK_WEBHOOK"},
+		{
+			Name:  "report",
```
stephanos: I don't know about the command names; both of them do "report" and both are about "failures".

How about calling the commands `alert` and `digest` instead?

stephanos: If you like `digest`, `BuildSuccessReport` could become `BuildDigest`

tools/ci-notify/slack.go
```diff
@@ -149,3 +149,146 @@ func SendSlackMessage(webhookURL string, message *SlackMessage) error {
...
+	sb.WriteString(fmt.Sprintf("  Failed Runs: %d\n", report.FailedRuns))
+	sb.WriteString(fmt.Sprintf("  Average Duration: %s\n", formatDuration(report.AverageDuration)))
+	sb.WriteString(fmt.Sprintf("  Median Duration: %s\n", formatDuration(report.MedianDuration)))
+
+	sb.WriteString("\nRun Duration Distribution:\n")
+	sb.WriteString(fmt.Sprintf("  Under 20 minutes: %.1f%%\n", report.Under20MinutesPercent))
+	sb.WriteString(fmt.Sprintf("  Under 25 minutes: %.1f%%\n", report.Under25MinutesPercent))
+	sb.WriteString(fmt.Sprintf("  Under 30 minutes: %.1f%%\n", report.Under30MinutesPercent))
+
+	if report.FailedRuns > 0 {
```
stephanos: I'm not sure about this part; it's pretty easy to find failed runs by going to https://github.com/temporalio/temporal/commits/main/ - I'd be in favor of removing it

tools/ci-notify/github.go
```diff
@@ -90,3 +91,170 @@ func BuildFailureReport(runID string) (*FailureReport, error) {
...
+// GetWorkflowRuns fetches workflow runs for a branch within a time range
+func GetWorkflowRuns(branch, workflowName string, since time.Time) ([]WorkflowRunSummary, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	// Format the date as YYYY-MM-DD for gh CLI
+	sinceDate := since.Format("2006-01-02")
+
+	cmd := exec.CommandContext(ctx, "gh", "run", "list",
+		"--branch", branch,
```
stephanos, line 172: In the ticket I had mentioned it'd also be interesting to see the duration times for non-main as well (for passing PRs only). If that's too much effort right now, we can make that a follow-up. Generally I think it's important to do as well since we want to look at that metric separately since we have a stricter threshold for that one in mind.

tools/ci-notify/github.go
```diff
@@ -90,3 +91,170 @@ func BuildFailureReport(runID string) (*FailureReport, error) {
...
+	if len(durations) == 0 {
+		return 0
+	}
+
+	// Make a copy to avoid modifying original
+	sorted := make([]time.Duration, len(durations))
+	copy(sorted, durations)
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i] < sorted[j]
+	})
```
stephanos, line 131: nit: there's `slices.SortFunc` now

tools/ci-notify/app.go
spkane31: Removing

tools/ci-notify/app.go
spkane31: Changing to `"Starting 'ci-notify alert'"`

stephanos (approved):

Issue #7637, opened 2025-04-20
Add a dedicated metrics for corrupted workflows


phuongdnguyen:
**Is your feature request related to a problem? Please describe.**

Currently, corrupted workflows are either:
- Logged & counted in namespace-level metrics (like ConditionFailedError or CurrentWorkflowConditionFailedError)
- Logged only (DataLoss error)

In our environment when there millions of workflow are running, it takes more effort to determine which workflows are corrupted by looking at the log or metrics (`persistence_error_with_type`).

**Describe the solution you'd like**
Adding a new metrics (eg: `workflow_corrupted`) which label of workflowID and runID. This is more efficient to detect problematic workflows rather than running a full db scan using `tdbg`.

I can help to come up with a proposal (deployed to production env in our fork) if you guys think it's useful.

**Describe alternatives you've considered**

**Additional context**

yycptt:
Looks like there's some misunderstanding re. the error type. `ConditionFailedError`, `CurrentWorkflowConditionFailedError`, `WorkflowConditionFailedError` are expected errors and don't mean workflows are corrupted. They can happen during history shard movement or trying to start an existing workflow. 

DataLoss errors yes, signals one corruption type that we are able to detect & well known, in this case the corruption of history event batches. Other types of corruption, for example resurrected workflows/activity/child workflow info, missing current workflow record etc. can't even be detected without introducing a lot more operations to the normal hot code path and that's not something we can afford.
Putting it another way, if we can detect the corrupted workflows (especially on the workflow state side), we can go ahead and fix those corruptions (by replaying the history events) and not emit the corruption metric.

Due to cardinality reasons, we also can't emit metric with workflowID & runID tags, they are too expensive for metric backends.
For now, error log is the probably the best way to figure out workflowID & runIDs.

phuongdnguyen:
Hi @yycptt , thanks for the response.
It's been a while and we've go ahead to add metrics in our internal fork.
So we found several combinations indicator of corruption:
- `ConditionFailedError` when Temporal do a workflow execution update (the api in persistence layer)
- DataLoss Error as you've mentioned.
About `CurrentWorkflowConditionFailedError` and `WorkflowConditionFailedError`, yes we see it's a transient error that usually related to new workflow start or when a lot of workflows doing continueasnew

Moving forward, do you think we need a system workflow (like the existing ones registered for internal worker) to do scanning for corruptions? Something like what `tdbg admin db scan` do but in a more controlled way

[...]

PR #9256, opened 2026-02-08
Add workflow_current_record_missing_errors metric for corruption detection

harsh543:
## What changed

Add a new persistence metric `workflow_current_record_missing_errors` that fires when a `CurrentWorkflowConditionFailedError` is returned with an **empty RunID** — indicating the `current_executions` row is missing from the database (Cassandra/SQL). This is a strong, actionable corruption signal.

A plain `ConditionFailedError` does **not** imply corruption and is intentionally **not** instrumented. The only case we target is:
- `CurrentWorkflowConditionFailedError` with `RunID == ""` → current execution record is gone

The metric is gated under the existing `system.enableDataLossMetrics` flag (default `false`). A missing current record is semantically a data-loss/corruption event, so sharing this flag is correct and avoids giving operators two separate toggles for the same class of corruption signal.

### Changes
| File | Change |
|------|--------|
| `common/metrics/metric_defs.go` | New counter `workflow_current_record_missing_errors` |
| `common/persistence/metrics.go` | Helper `IsMissingCurrentRecordError()` + `EmitCurrentRecordMissingMetric()`; renamed `source` → `operation` param for consistency |
| `common/persistence/persistence_metric_clients.go` | Emit in `UpdateWorkflowExecution` and `ConflictResolveWorkflowExecution`; gated under `EnableDataLossMetrics`; also fixed `ConflictResolveWorkflowExecution` to pass real `workflowID`/`runID` to `recordDataLossMetrics` (was passing empty strings) |
| `common/persistence/client/{factory,fx}.go` | No new flag needed — metric is gated by the existing `EnableDataLossMetrics` |
| `common/persistence/persistence_metrics_clients_test.go` | 7 new tests: positive/negative paths for both update and conflict-resolve operations, gating behavior, and dedicated helper tests |
| `common/persistence/client/factory_test.go`, `persistence_rate_limited_clients_test.go`, `persistence-tests/persistence_test_base.go` | Updated `NewFactory` call sites (no new param) |

## Why

Temporal emits `DataLoss` metrics for certain corruption scenarios, but has no signal when the **current execution record itself is missing**. This gap means operators cannot detect or alert on a specific, well-understood corruption case. This metric fills that gap with minimal noise (opt-in via `system.enableDataLossMetrics`, high-cardinality tags gated behind the same flag that protects existing data-loss metrics).

## How did you test it
- [x] Built
- [x] All new and existing persistence metric client tests pass
- [x] Positive path: metric emitted for `CurrentWorkflowConditionFailedError` with empty RunID on both `UpdateWorkflowExecution` and `ConflictResolveWorkflowExecution`
- [x] Negative paths: metric not emitted when flag is disabled, when RunID is non-empty (normal conflict), or for unrelated errors


harsh543:
## Summary for reviewers

This PR closes a real observability gap that has bitten operators in production (see #7637). Today, when the `current_executions` row silently disappears from Cassandra or SQL, there is **zero signal** — no metric, no alert, nothing. The workflow just stops making progress and the only way to find it is a full DB scan or log parsing. That's not acceptable at scale.

### Why this specific detection is high-signal and low-noise

Not all `ConditionFailedError` variants indicate corruption — most are benign (shard movement, duplicate workflow starts, etc.). We deliberately avoid blanket instrumentation. The **only** case we target is:

> `CurrentWorkflowConditionFailedError` with `RunID == ""`

This means the database has no `current_executions` row for the workflow at all. It's not a conflict with another run — there is simply no current record. This is an unambiguous corruption signal that already exists in the persistence layer; we're just surfacing it as a metric.

### Design choices

| Decision | Rationale |
|----------|-----------|
| **Opt-in via dynamic config** (`system.enableCurrentRecordMissingMetric`, default `false`) | High-cardinality labels (`workflow_id`, `run_id`) — same pattern as existing `DataLoss` metrics |
| **Only two emission paths** (`UpdateWorkflowExecution`, `ConflictResolveWorkflowExecution`) | These are the only places where this error surfaces today |
| **No new DB calls, no proto changes, no API changes** | Zero overhead when disabled; negligible when enabled |
| **Helper function `IsMissingCurrentRecordError`** | Exported, tested independently, reusable for future scanner integration |

### Bonus fix

While wiring the new metric into `ConflictResolveWorkflowExecution`, we noticed the existing `recordDataLossMetrics` call was passing empty strings for `workflowID`/`runID` even though the request has that data. Fixed that too — existing `DataLoss` metrics for conflict-resolve now carry proper workflow identifiers.

### Test coverage

- Metric fires on empty-RunID corruption (via `UpdateWorkflowExecution`)
- Metric fires via `ConflictResolveWorkflowExecution` path
- Metric does **not** fire for normal conflicts (non-empty RunID)
- Metric does **not** fire when dynamic config is disabled
- Table-driven unit test for `IsMissingCurrentRecordError` covering nil, unrelated errors, `ConditionFailedError`, non-empty RunID, empty RunID, and wrapped errors

This is a minimal, surgical change that gives operators a way to detect a specific, well-understood corruption class without any risk to existing behavior.

common/dynamicconfig/constants.go
```diff
@@ -270,6 +270,14 @@ operator API calls (highest priority). Should be >0.0 and <= 1.0 (defaults to 20
 		`EnableDataLossMetrics determines whether dataloss metrics are emitted when dataloss errors are encountered`,
 	)
 
+	EnableCurrentRecordMissingMetric = NewGlobalBoolSetting(
```
prathyushpv: Thanks @harsh543 for the change! Overall it looks good. But I think we can use `EnableDataLossMetrics` flag itself to enable or disable emitting this metric. We don't really need a new flag.

harsh543: Good call, @prathyushpv — completely agree. A missing `current_executions` record is a data-loss/corruption event, so gating it under `EnableDataLossMetrics` is semantically correct and avoids giving operators two overlapping toggles for the same category of corruption signal.

Done in the latest commit:
- Removed `system.enableCurrentRecordMissingMetric` entirely.
- `workflow_current_record_missing_errors` is now gated by `EnableDataLossMetrics`, the same flag that controls `DataLossCounter`.
- Also took the opportunity to rename the `source` param to `operation` in both `EmitDataLossMetric` and `EmitCurrentRecordMissingMetric` for consistency with the rest of the metrics API, and added two conflict-resolve negative tests (flag disabled + non-empty RunID) to match the update-path coverage.

PR #9191, opened 2026-02-01
Migrate TaskQueueStatsSuite to TestEnv

stephanos:
## What changed?

WISOTT

## Why?

TestEnv allows running tests in parallel; speeding up local execution of the suite.

TBD

## How did you test it?
- [ ] built
- [ ] run locally and tested manually
- [x] covered by existing tests
- [ ] added new unit test(s)
- [ ] added new functional test(s)

PR #9260, opened 2026-02-09
Improve TestEnv

stephanos:
## What changed?

Improved TestEnv:

1. When using `MustRunSequential`, ensures the cluster is only used by _that_ suite
2. Makes namespace ID available, too (not just name)
3. Applies test sharding

## Why?

Make TestEnv behave like testify suites while we migrate suites over.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

This PR was extracted from a suite migration where this is used: https://github.com/temporalio/temporal/pull/9191

### DIFF
--- a/tests/testcore/test_env.go
+++ b/tests/testcore/test_env.go
@@ -2,10 +2,13 @@ package testcore
 
 import (
 	"fmt"
+	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
 
+	"github.com/dgryski/go-farm"
 	"github.com/stretchr/testify/require"
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/workflowservice/v1"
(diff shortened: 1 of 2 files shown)


tests/testcore/test_env.go
```diff
@@ -80,28 +85,54 @@ func WithDynamicConfig(setting dynamicconfig.GenericSetting, value any) TestOpti
...
 }
 
 // NewEnv creates a new test environment with access to a Temporal cluster.
-// Tests are run in parallel - use MustRunSequential to run suite sequentially.
+//
+// By default, tests are marked as parallel. Use MustRunSequential on the
+// test's parent `testing.T` to run them sequentially instead.
 func NewEnv(t *testing.T, opts ...TestOption) *testEnv {
+	// Check test sharding early, before any expensive operations.
+	checkTestShard(t)
```
stephanos, line 127: Ensures this only runs when test is on that shard.

PR #9261, opened 2026-02-09
V1 to CHASM scheduler migration RPC

chaptersix:
## What changed

Adds an admin RPC (`MigrateSchedule`) that migrates workflow-backed (V1) schedulers to CHASM (V2) schedulers.

The flow:
1. Admin API signals the V1 scheduler workflow with `migrate-to-chasm`
2. V1 workflow runs a local activity that snapshots state and calls `CreateFromMigrationState` on the CHASM scheduler service
3. On success, the V1 workflow closes. On failure, a `PendingMigration` flag persists so the next run loop iteration retries automatically

Key details:
- `CreateSchedulerFromMigration` initializes the full CHASM scheduler tree (generator, invoker, backfillers, visibility) from V1 state
- Running/completed workflows are converted to buffered starts; ongoing backfills are preserved
- Migration is idempotent -- if the CHASM schedule already exists, it's treated as success
- Metrics: `schedule_migration_started`, `schedule_migration_completed`, `schedule_migration_failed` (with direction tag)
- Local activity uses 10m schedule-to-close timeout; retries happen at the workflow level via the persistent flag

## Why

Needed for migrating existing V1 schedulers to CHASM without downtime or data loss.

## Follow-up items

- V2 to V1 migration (rollback path)
- Sentinel key handling when `EnableCHASMSchedulerCreation` was previously enabled
- Attach completion callbacks to running workflows after migration


service/frontend/admin_handler.go
```diff
@@ -2399,3 +2400,47 @@ func convertFailoverHistoryToReplicationProto(
...
+	if request == nil {
+		return nil, errRequestNotSet
+	}
+	if request.GetNamespace() == "" {
+		return nil, errNamespaceNotSet
+	}
+	if request.GetScheduleId() == "" {
+		return nil, errScheduleIDNotSet
+	}
+	if request.GetTarget() == adminservice.MigrateScheduleRequest_SCHEDULER_TARGET_UNSPECIFIED {
```
chaptersix, line 2413: can target be empty?

chasm/lib/scheduler/scheduler_test.go
```diff
@@ -27,3 +36,209 @@ func TestListInfo(t *testing.T) {
...
+			},
+		},
+	}
+}
+
+func TestCreateSchedulerFromMigration_InitializesComponents(t *testing.T) {
+	_, ctx, node := setupSchedulerForTest(t)
+	req := testMigrateScheduleRequest()
+
+	// Create a new context for the migration (we need a fresh transaction)
```
chaptersix: will remove these comments and revise these tests

service/frontend/admin_handler.go
```diff
@@ -2399,3 +2400,47 @@ func convertFailoverHistoryToReplicationProto(
...
+		return nil, errMigrationTargetNotSet
+	}
+	if request.GetTarget() != adminservice.MigrateScheduleRequest_SCHEDULER_TARGET_CHASM {
+		return nil, serviceerror.NewUnimplemented("Only migration to CHASM is currently supported.")
+	}
+	// TODO: do we need to validate identity?
```
chaptersix: ?

chasm/lib/scheduler/handler.go
```diff
@@ -140,3 +140,36 @@ func (h *handler) ListScheduleMatchingTimes(ctx context.Context, req *schedulerp
...
+			chasm.BusinessIDReusePolicyRejectDuplicate,
+			chasm.BusinessIDConflictPolicyFail,
+		),
+	)
+	if err != nil {
+		var alreadyStartedErr *chasm.ExecutionAlreadyStartedError
+		if errors.As(err, &alreadyStartedErr) {
+			return nil, serviceerror.NewWorkflowExecutionAlreadyStarted(
+				"CHASM schedule already exists",
+				"",
```
chaptersix: should we include a this info n the request?

service/frontend/admin_handler.go
lina-temporal: Only for invalid requests

service/frontend/admin_handler.go
lina-temporal: I don't think so, since this is an admin-only API.

service/worker/scheduler/workflow.go
```diff
@@ -957,6 +981,40 @@ func (s *scheduler) handleForceCANSignal(ch workflow.ReceiveChannel, _ bool) {
...
+		s.Info,
+		s.State,
+		workflowInfo.SearchAttributes,
+		workflowInfo.Memo,
+		s.now(),
+	)
+	// inc the sequence number to prevent to invalidate signals to
+	// this workflow after the migration has started.
+	// they should target the chasm scheduler after this point
+	s.incSeqNo()
```
chaptersix: the conflict token is not checked until a signal is processed. So this likely has no benefit.

service/worker/scheduler/workflow.go
```diff
@@ -310,6 +313,24 @@ func (s *scheduler) run() error {
...
 		}
+
+		if s.pendingMigration {
+			err := s.executeMigration()
+			if err == nil {
+				s.logger.Info("Migration to CHASM succeeded, terminating V1 workflow",
+					"namespace", s.State.Namespace,
+					"schedule-id", s.State.ScheduleId,
+				)
+				return nil
```
chaptersix, line 329: anything else that should be done before closing the workflow?

chaptersix: we will drop any signals received during the migration.

chasm/lib/scheduler/handler.go
chaptersix: will circle back to this.

service/worker/scheduler/workflow.go
lina-temporal: > we will drop any signals received during the migration.

yep, regrettably true, though updates are already accepted (or rejected) asynchronously, so it's not new behavior.

service/worker/scheduler/activities.go
```diff
@@ -372,3 +374,27 @@ func (r responseBuilder) makeResponse(result *commonpb.Payloads, failure *failur
...
+			return nil
+		}
+		return translateError(err, "MigrateSchedule")
+	}
+
+	a.Logger.Info("MigrateSchedule: successfully created CHASM schedule",
+		tag.WorkflowNamespace(req.GetState().GetSchedulerState().GetNamespace()),
+		tag.ScheduleID(req.GetState().GetSchedulerState().GetScheduleId()),
+	)
+	a.MetricsHandler.Counter(metrics.ScheduleMigrationCompleted.Name()).Record(1)
```
chaptersix: this should happen at the same level as the start

service/worker/scheduler/activities.go
```diff
@@ -372,3 +374,27 @@ func (r responseBuilder) makeResponse(result *commonpb.Payloads, failure *failur
...
+	_, err := a.SchedulerClient.MigrateSchedule(ctx, req)
+	if err != nil {
+		// Treat "already exists" as success (idempotency)
+		var alreadyExists *serviceerror.WorkflowExecutionAlreadyStarted
+		if errors.As(err, &alreadyExists) {
+			a.Logger.Warn("MigrateSchedule: CHASM schedule already exists, treating as success",
+				tag.WorkflowNamespace(req.GetState().GetSchedulerState().GetNamespace()),
+				tag.ScheduleID(req.GetState().GetSchedulerState().GetScheduleId()),
+			)
+			a.MetricsHandler.Counter(metrics.ScheduleMigrationCompleted.Name()).Record(1)
```
chaptersix: same here.

service/worker/scheduler/activities.go
```diff
@@ -372,3 +374,27 @@ func (r responseBuilder) makeResponse(result *commonpb.Payloads, failure *failur
 	}
 	return res
 }
+
+func (a *activities) MigrateSchedule(ctx context.Context, req *schedulerpb.MigrateScheduleRequest) error {
+	_, err := a.SchedulerClient.MigrateSchedule(ctx, req)
```
chaptersix: I'm still apprehensive to resume (unpause) schedule if we get an error here.  
Let's say on the 10th attempt to the chasm schedule is able to be created but the success response does not make it. TCP connection drops or something to that effect. 

No retries are left 

This activity would be marked as failed, v1 scheduler would the resume running as normal. 

I'd prefer if we left the v1 schedule paused in the event of failure.

service/worker/scheduler/activities.go
chaptersix: also need to rename this to indicate the direction.

chaptersix:
Attaching callbacks to workflows that have already been started is not yet supported so I'll do that in another PR

service/worker/scheduler/activities.go
chaptersix: regardless we should have an alert for "migration failure (after all attempts) failed" > 0

chaptersix: after what ever RCA / fix. we can call call the migrate schedule again. 

It would also be useful to have a tdbg command that allows us to delete a schedule on a give stack.

service/frontend/admin_handler.go
```diff
@@ -2399,3 +2400,49 @@ func convertFailoverHistoryToReplicationProto(
...
+	}
+	if request.GetScheduleId() == "" {
+		return nil, errScheduleIDNotSet
+	}
+	if request.GetTarget() == adminservice.MigrateScheduleRequest_SCHEDULER_TARGET_UNSPECIFIED {
+		return nil, errMigrationTargetNotSet
+	}
+	if request.GetTarget() != adminservice.MigrateScheduleRequest_SCHEDULER_TARGET_CHASM {
+		return nil, serviceerror.NewUnimplemented("Only migration to CHASM is currently supported.")
+	}
```
lina-temporal, line 2419: Can you add a TODO here, since AdminService#MigrateSchedule needs to work in both directions?

service/worker/scheduler/workflow.go
```diff
@@ -310,6 +313,34 @@ func (s *scheduler) run() error {
...
+			// Pause the V1 scheduler to prevent it from starting new workflows
+			// while migration is in progress.
+			wasPaused := s.Schedule.State.Paused
+			previousNotes := s.Schedule.State.Notes
+			s.Schedule.State.Paused = true
+			s.Schedule.State.Notes = "paused for migration to CHASM"
+
+			err := s.executeMigration()
+			if err == nil {
+				s.logger.Info("Migration to CHASM succeeded, terminating V1 workflow",
```
lina-temporal: "terminating" has pretty specific meaning for Temporal, I'd update this to "closing V1 workflow"

service/worker/scheduler/workflow.go
```diff
@@ -957,6 +991,49 @@ func (s *scheduler) handleForceCANSignal(ch workflow.ReceiveChannel, _ bool) {
...
+	)
+	migrateOptions := workflow.LocalActivityOptions{
+		ScheduleToCloseTimeout: 1 * time.Hour,
+		StartToCloseTimeout:    5 * time.Second,
+		RetryPolicy: &temporal.RetryPolicy{
+			InitialInterval: 1 * time.Second,
+			MaximumInterval: 60 * time.Second,
+			MaximumAttempts: 10,
+		},
+	}
```
lina-temporal, line 1026: 1. In this case, I think it's okay to have both start-to-close and schedule-to-close as identical low timeouts. The intention is to be as non-disruptive as possible during migration, so we wouldn't want the possibility of waiting a long time for the task to get picked up; I'd put both at something like 10s or 5s. 
2. Since we're running migration as part of the scheduler's regular run loop, I don't think it makes sense to retry in a loop here. We can simply let the next scheduler run re-attempt.

service/worker/scheduler/activities.go
lina-temporal: > Let's say on the 10th attempt to the chasm schedule is able to be created but the success response does not make it. TCP connection drops or something to that effect. No retries are left

See my comment on the retries (we shouldn't retry inline here IMO). 

> This activity would be marked as failed, v1 scheduler would the resume running as normal.

That's not possible with this sequencing.

```
1. Scheduler timer fires (or signal is received) in history service
2. A workflow task is scheduled to be picked up by the worker
3. The workflow task is started by the worker, which triggers the `run` loop
4. The run loop pauses the schedule
5. The local activity runs, creating the CHASM schedule
6. Local activity finishes, and the run loop continues (success case)
```

In the success case, the local activity completes, the worker adds a `MarkerRecorded` event, which will be sent as part of the next workflow task completion. Our logic then unpauses the scheduler - this also gets written (as the memo block) when we complete that workflow task. The key bit is that there's *no* delineation that the history service is aware of between our run loop progressing, and our local activity running; you only get the marker recorded in the event history. While the local activity is running, the SDK may refresh the workflow task (to prevent temporal from thinking it's timed out), but otherwise, history has no idea that the worker's running a local activity.

Now, think about the failure case. Let's say that we pause (locally updating our WF memo block), then create the CHASM schedule in that local activity, then our thread immediately dies, as in your example. At this point, history service doesn't even know that we'd paused the schedule (since that's sent as part of the WFT completion)! Temporal doesn't (can't) just resume in execution from where it last left off; all history can do is detect that the workflow task timed out (thread went away), and reschedule it.  That rescheduled WFT will cause the scheduler to re-enter the run loop, *including* any signals that haven't yet been marked as processed (again. by virtue of being included from the SDK as part of a workflow task completion response). 

Since the first thing we check in `run` is always for a pending migration, the rescheduled WFT will go through the exact same migration logic - pause, then attempt migration. Only this time, we'll fast-succeed migration, since we'll detect the CHASM scheduler was already created, and exit. Remember that the V1 scheduler must go sequentially through that run loop, every time, so it's not possible for a V1 scheduler to concurrently be crashed/timed out on migration, and yet still executing the local WFT/progressing further into the `run` loop. 

> regardless we should have an alert for "migration failure (after all attempts) failed" > 0

For sure, we'll probably want to get a (low priority) alert for *any* migration failures. 

> It would also be useful to have a tdbg command that allows us to delete a schedule on a give stack.

Do you mean one intelligent enough to figure out the right stack to delete from? We can delete either from tdbg as-is, you'd just have to specify the right archetype (and prefix with `temporal-sys-scheduler:` for V1, if you want to target it with `tdbg workflow` commands).

chasm/lib/scheduler/migration/migration.go
lina-temporal, line 115: ('m guessing you'll update this when working on the CHASM->V1 path? feel free to just resolve if so)

chasm/lib/scheduler/proto/v1/service.proto
```diff
@@ -31,4 +31,7 @@ service SchedulerService {
     rpc ListScheduleMatchingTimes(ListScheduleMatchingTimesRequest) returns (ListScheduleMatchingTimesResponse) {
         option (temporal.server.api.routing.v1.routing).business_id = "frontend_request.schedule_id";
     }
+    rpc MigrateSchedule(MigrateScheduleRequest) returns (MigrateScheduleResponse) {
```
lina-temporal: Since we need to handle both V1->V2 and V2->V1, this might make more sense as something like `MigrateFromWorkflow` (to set up for `MigrateToWorkflow`). Or even call the outer API `CreateFromMigrationState` as well. 

Similar comment on the scheduler workflow; if we name our signal `migrate`, that may make it a bit more confusing when we want to add a signal for incoming migrations (e.g., where we're passing the `SchedulerMigrationState` record via signals onto the new V1 sched workflow). 

The only name that covers both directions would be the AdminService's `MigrateSchedule`, since it should be capable of both V1->V2, and V2->V1.

chasm/lib/scheduler/scheduler.go
```diff
@@ -218,6 +218,74 @@ func CreateScheduler(
...
+) (*Scheduler, error) {
+	state := req.GetState()
+	if state == nil {
+		state = &schedulerpb.SchedulerMigrationState{}
+	}
+
+	schedulerState := state.GetSchedulerState()
+	if schedulerState == nil {
+		schedulerState = &schedulerpb.SchedulerState{}
+	}
```
lina-temporal: simplify to `state := req.GetState().GetSchedulerState()`, protobuf handles the nil-chaining. If there's no state, just error out.

chasm/lib/scheduler/scheduler.go
```diff
@@ -218,6 +218,74 @@ func CreateScheduler(
...
+
+	invokerState := state.GetInvokerState()
+	if invokerState == nil {
+		invokerState = &schedulerpb.InvokerState{
+			BufferedStarts: []*schedulespb.BufferedStart{},
+		}
+	}
+	invoker := &Invoker{
+		InvokerState: invokerState,
+	}
```
lina-temporal: I think constructing these components (Invoker, Generator, Backfiller) outside of their normal path will be a little brittle. Instead, can you refactor the constructors to something like:

```
func NewInvoker(...) *Invoker {
  return &Invoker{...}
}

// to

func NewInvoker(ctx, ...) *Invoker {
  return newInvokerWithState(&schedulerpb.InvokerState{
   ... // defaults here
  })
}

// private constructor here
func newInvokerWithState(ctx, state) *Invoker {
  inv := &Invoker{state}
  inv.addTasks(ctx) // adds own tasks as needed
  return inv
}
```

chasm/lib/scheduler/scheduler.go
```diff
@@ -218,6 +218,74 @@ func CreateScheduler(
...
+	sched.Invoker = chasm.NewComponentField(ctx, invoker)
+	if len(invoker.BufferedStarts) > 0 {
+		invoker.addTasks(ctx)
+	}
```
lina-temporal: If we go through the regular constructor, then we should also make sure the constructor is responsible for creating initial tasks (see Generator).

chasm/lib/scheduler/scheduler.go
```diff
@@ -218,6 +218,74 @@ func CreateScheduler(
...
+		LastCompletionResult: chasm.NewDataField(ctx, state.GetLastCompletionResult()),
+	}
+	sched.setNullableFields()
+
+	invokerState := state.GetInvokerState()
+	if invokerState == nil {
+		invokerState = &schedulerpb.InvokerState{
+			BufferedStarts: []*schedulespb.BufferedStart{},
+		}
+	}
```
lina-temporal: Aren't we always at least setting the InvokerState in the protobuf (even if we don't have BufferedStarts)? IMO, reducing some of these nil checks would be good for readability, if we don't need them.

service/worker/scheduler/workflow_test.go
```diff
@@ -2269,3 +2269,107 @@ func (s *workflowSuite) TestCANBySignal() {
...
+			}},
+		},
+	}, 3)
+
+	// Verify the workflow continued running after the migration failure
+	// rather than terminating (returning nil) like it does on success.
+	// The test harness forces CAN after 3 iterations, so a CAN error
+	// here proves the workflow kept going.
+	s.True(s.env.IsWorkflowCompleted())
+	s.True(workflow.IsContinueAsNewError(s.env.GetWorkflowError()))
```
lina-temporal: couldn't we set the `IterationsBeforeContinueAsNew` higher and assert on `!IsWorkflowCompleted`? If not, I think this is okay, just a bit of a roundabout way to assert.

service/worker/scheduler/activities.go
lina-temporal: Thinking more about this - I don't think we actually need to adjust the pause state for a V1 workflow *at all*, since:

- on success, we immediately return from the loop/WFT, so there's no duplicate WF
- on explicit migration failure, we want to keep running actions
- in the case that our local activity succeeds CHASM create, then crashes, we restart the whole WFT from `run`; pause never gets written anyways (nor does it need to be, since we'll still try to process the signal). 

That said, you may want to add a flag in the memo block to indicate that the schedule is flagged for migration, so that we can retry in subsequent loops (after we've already received the first migration signal). Like moving `pendingMigrate` to the `State` protobuf. 

In V2->V1 case, we need to do the pause when rolling back, because CHASM tasks can execute concurrently (so even while you're migrating, the Generator might have a task fire and start trying to queue things up).

lina-temporal: quick summary: I don't think we have any issue for the "crash" case, which means we also don't need to worry about pausing in V1 (V2 still needs to). We may want to use scheduler's protobuf `State` for the `pendingMigrate` flag, so that when we process signals successfully (but fail to migrate, no crash), we can retry in the next `run` iteration without an additional signal.

service/worker/scheduler/workflow_test.go
```diff
@@ -2269,3 +2269,107 @@ func (s *workflowSuite) TestCANBySignal() {
...
+}
+
+func (s *workflowSuite) TestMigrateFailureThenSignal() {
```
lina-temporal, line 2404: See comments above re: paused flag, I'm not sure we'll need end up needing this test in its current form (but may want something to assert that the 'pending migration' flag sticks).

chasm/lib/scheduler/migration/migration.go
chaptersix: yes

chaptersix:
revalidate activity policy

chasm/lib/scheduler/handler.go
```diff
@@ -163,3 +163,40 @@ func (h *handler) ListScheduleMatchingTimes(ctx context.Context, req *schedulerp
...
 	return resp, err
 }
+
+// CreateFromMigrationState creates a CHASM schedule from migrated V1 state.
+// Used during migration from workflow-backed schedules to CHASM schedules.
+//
+// TODO: attach completion callbacks to running workflows migrated from V1.
+// TODO: handle sentinel key that may exist if EnableCHASMSchedulerCreation was
+// enabled when the schedule was originally created. The existing CHASM component
+// must be replaced with the migrated state.
```
chaptersix: task to eval next run?

common/metrics/metric_defs.go
```diff
@@ -1341,6 +1341,18 @@ var (
 		"schedule_payload_size",
 		WithDescription("The size in bytes of a customer payload (including action results and update signals)"),
 	)
+	ScheduleMigrationStarted = NewCounterDef(
```
chaptersix, line 1349: indicate direction

proto/internal/temporal/server/api/schedule/v1/message.proto
```diff
@@ -78,6 +78,8 @@ message InternalState {
     int64 conflict_token = 7;
 
     bool need_refresh = 9;
+
+    bool pending_migration = 11;
```
chaptersix, line 82: might need a "paused before migration bool"

chaptersix: will address in next PR if needed.

chasm/lib/scheduler/invoker.go
```diff
@@ -27,14 +27,22 @@ func (i *Invoker) LifecycleState(ctx chasm.Context) chasm.LifecycleState {
...
+	})
+}
+
+func newInvokerWithState(ctx chasm.MutableContext, state *schedulerpb.InvokerState) *Invoker {
+	i := &Invoker{
+		InvokerState: state,
+	}
+	if len(i.BufferedStarts) > 0 {
+		i.addTasks(ctx)
 	}
```
lina-temporal: `addTasks` already has logic to ensure exactly the right tasks are created, I'd elide this length check here. I don't think it could ever happen in reality, but technically, if the state had cancel/terminate workflows loaded, we'd miss a case where we'd add a task:

```
	if len(i.GetCancelWorkflows()) > 0 || len(i.GetTerminateWorkflows()) > 0 || eligibleStarts > 0 {
		ctx.AddTask(i, chasm.TaskAttributes{}, &schedulerpb.InvokerExecuteTask{})
...

File: common/metrics/metric_defs.go [hunk 24]
```diff
@@ -1342,6 +1346,18 @@ var (
...
+		"schedule_migration_started",
+		WithDescription("The number of times a V1 schedule starts migration to CHASM"),
```
lina-temporal: Since you have the metric tag for direction, isn't this for any sort of migration, not just V1->CHASM?

service/worker/scheduler/workflow.go
```diff
@@ -270,7 +272,10 @@ func (s *scheduler) run() error {
 		info := workflow.GetInfo(s.ctx)
 		suggestContinueAsNew := info.GetCurrentHistoryLength() >= impossibleHistorySize
 		if s.tweakables.IterationsBeforeContinueAsNew > 0 {
-			suggestContinueAsNew = suggestContinueAsNew || iters <= 0
+			// forceCAN must be checked here too, not just in the else branch,
```
lina-temporal, line 275: great catch 👍

PR #9262, opened 2026-02-09
Add a fallback behaviour if the matching RPC: CheckTaskQueueVersionMembership does not exist in matching

Shivs11:
## What changed?
- WISOTT

## Why?
- During the a rolling upgrade, there could be a world where history (the caller of the matching RPC) is on a higher version than matching (the service that has the RPC implemented)
- In this world, we want to fallback on an already implemented matching RPC (GetTaskQueueUserData) so that the functionality does not differ.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
- None, cause the fallback RPC has been in matching since the start of time.

---

> [!NOTE]
> **Medium Risk**
> Changes behavior in task-queue version membership checks and pinned override validation during RPC errors; incorrect fallback parsing or user-data assumptions could allow/deny pinned versions incorrectly during upgrades.
> 
> **Overview**
> Adds a rolling-upgrade-safe fallback when `matching.CheckTaskQueueVersionMembership` is unavailable: on `serviceerror.Unimplemented`, the code now calls `GetTaskQueueUserData` and locally checks whether the deployment version exists.
> 
> This fallback is applied both to workflow task-queue membership detection (`GetIsWFTaskQueueInVersionDetector`) and pinned-version validation (`validatePinnedVersionInTaskQueue`), including caching the fallback result and preserving the existing failed-precondition error when the version is not a member. New unit tests cover the `Unimplemented` fallback paths for both override validation and the detector.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0c26b0cc662ab70bd724c03debebf3cde6642e9b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/common/worker_versioning/worker_versioning.go
+++ b/common/worker_versioning/worker_versioning.go
@@ -2,6 +2,7 @@ package worker_versioning
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


common/worker_versioning/worker_versioning.go
```diff
@@ -296,12 +297,45 @@ func GetIsWFTaskQueueInVersionDetector(matchingClient resource.MatchingClient) I
...
+	resp, err := matchingClient.GetTaskQueueUserData(ctx,
+		&matchingservice.GetTaskQueueUserDataRequest{
+			NamespaceId:   namespaceID,
+			TaskQueue:     tq,
+			TaskQueueType: tqType,
+		})
+	if err != nil {
+		return false, err
+	}
+	tqData, ok := resp.GetUserData().GetData().GetPerType()[int32(tqType)]
```
yiminc, line 332: any nil pointer panic here?

Shivs11: nope; the usage of the .Get methods make this nil safe 
moreover, the function which we are calling on Line 336 (HasDeploymentVersion) also returns early if the value being passed is nil or not)

PR #9267, opened 2026-02-09
Fix CancelOutstandingPoll for forwarded polls

rkannan82:
## What changed?
Fixes a bug where CancelOutstandingPoll fails to cancel polls that have been forwarded between task queue partitions.

**Problem**
When a poll is forwarded from a child partition to a parent partition, the same pollerID is registered in outstandingPollers on both partitions. This causes two issues:

- Overwrite bug: If both partitions are on the same matching node, the parent's cancel function overwrites the child's in the outstandingPollers map. When cancelled, only the parent poll is cancelled; the child poll remains active.

- Routing bug: Frontend calls CancelOutstandingPoll with the original task queue name, which always routes to partition 0 (root). If the poll was routed to a different partition (e.g., partition 2) or forwarded, the cancel request goes to the wrong partition.

**Solution**
- Matching: Use partition:pollerID as the key in outstandingPollers instead of just pollerID. This prevents overwrites when the same poller is registered on multiple partitions.
- Frontend: Fan out CancelOutstandingPoll to all read partitions, similar to how CancelOutstandingWorkerPolls works. This ensures the cancel reaches whichever partition(s) the poll is registered on.

## Why?
Fixes a bug where CancelOutstandingPoll fails to cancel polls that have been forwarded between task queue partitions.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
No. The fix is backward compatible - upgrade order doesn't matter, and behavior during transition is no worse than today.

Key points:
- Old FE + New Matching: Root partition works (key matches). Non-root fails (same bug as today - cancel routes to wrong partition).
- New FE + Old Matching: Old matching does e.outstandingPollers.Pop(request.PollerId) - it ignores TaskQueue entirely and uses just pollerID. So it still finds the poll. ✓
- New FE + New Matching: Both use partition:pollerID format. ✓

### DIFF
--- a/api/matchingservice/v1/request_response.go-helpers.pb.go
+++ b/api/matchingservice/v1/request_response.go-helpers.pb.go
@@ -560,6 +560,80 @@ func (this *CancelOutstandingPollResponse) Equal(that interface{}) bool {
 	return proto.Equal(this, that1)
 }
 
+// Marshal an object of type CancelOutstandingWorkerPollsRequest to the protobuf v3 wire format
+func (val *CancelOutstandingWorkerPollsRequest) Marshal() ([]byte, error) {
+	return proto.Marshal(val)
+}
+
+// Unmarshal an object of type CancelOutstandingWorkerPollsRequest from the protobuf v3 wire format
+func (val *CancelOutstandingWorkerPollsRequest) Unmarshal(buf []byte) error {
+	return proto.Unmarshal(buf, val)
+}
+
+// Size returns the size of the object, in bytes, once serialized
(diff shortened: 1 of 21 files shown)


dnr (commented):
The routing bug used to not be there since the routing client would overwrite the name in the request pb 😱  We changed that because it's awful, but I guess it broke this.

As discussed, this whole thing is obsolete. Let's remove CancelOutstandingPoll entirely and have a test that cancellation is propagated. (Maybe the test in this PR works)

PR #7092, opened 2025-01-16
Update Registry size limit warning

stephanos:
## What changed?

Adding log for when update registry size limit is reached.

## Why?

To identify a update registry size limit; and then enforce it in the next change.

## How did you test it?

Added test. And ran functional test with limit.

PR #9269, opened 2026-02-10
Prune two workflow update related metrics in favor of logs w/ namespace

spkane31:
## What changed?
Remove `invalid_state_transition_workflow_update_message` and `workflow_update_registry_size_limited` metrics. Add the `namespace` tag to the logs/softasserts in the `instrumentation` methods the metrics are used for.

## Why?
The logs will give us the same information and the namespace without cardinality concerns. These metrics are not expected to fire frequently.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks
Minimal, metrics changes only.


service/history/workflow/update/util.go
```diff
@@ -78,15 +78,15 @@ func (i *instrumentation) countSentAgain() {
...
-	i.oneOf(metrics.InvalidStateTransitionWorkflowExecutionUpdateCounter.Name())
+func (i *instrumentation) invalidStateTransition(updateID string, msg proto.Message, state state, namespace string) {
 	softassert.Fail(
 		i.log,
 		"invalid state transition attempted",
 		tag.ComponentWorkflowUpdate,
 		tag.String("update-id", updateID),
 		tag.String("message", fmt.Sprintf("%T", msg)),
 		tag.Stringer("state", state),
+		tag.String("namespace", namespace),
```
stephanos: 👍

service/history/workflow/update/util.go
```diff
@@ -48,13 +48,13 @@ func (i *instrumentation) countRateLimited() {
...
 }
 
-func (i *instrumentation) countRegistrySizeLimited(updateCount, registrySize, payloadSize int) {
-	i.oneOf(metrics.WorkflowExecutionUpdateRegistrySizeLimited.Name())
-	// TODO: remove log once limit is enforced everywhere
+func (i *instrumentation) countRegistrySizeLimited(updateCount, registrySize, payloadSize int, namespace string) {
 	i.log.Warn("update registry size limit reached",
 		tag.Int("registry-size", registrySize),
 		tag.Int("payload-size", payloadSize),
-		tag.Int("update-count", updateCount))
```
stephanos, line 57: Actually, looking back at the [original PR](https://github.com/temporalio/temporal/pull/7092), the reason for this log line was to get better data on how/when the registry size limit was hit (avoiding some of the issues with a metrics-based solution). Since this is a rate limit, I think metrics might be a better way to capture this actually, no?

spkane31: I think getting the namespace is important here and using logs instead of metrics removes the cardinality issue. I'd rather use a log here and add a metric if we hit this often. We also have `workflow_update_registry_size` to get similar information.

stephanos: > I think getting the namespace is important here and using logs instead of metrics removes the cardinality issue. I'd rather use a log here and add a metric if we hit this often.

I might need more clarity on how the cost works out; if (1) we'll need namespace tags for other metrics anyway and (2) the volume is low; what's the issue?

Apart from that, rate limits are typically tracked as metrics across the codebase AFAIK; logs are not as useful as they are much more limiting to query. Our log queries puts a cap on how much data it can ingest. On a big cluster that limits how far back in time you can go (I've had it unable to process more than 1h, for example). Metrics don't have that issue.

> We also have workflow_update_registry_size to get similar information.

It's not quite true; as you cannot make a leap from that to whether a limit was hit. If the size is at 99%, you cannot assume it hit the limit. Or if it's at 10%, it can still happen that an Update hits the limit.

spkane31: Thoughts on leaving the metric and adding namespace to the log for now? Once we eventually add the namespace to the metric we can remove the log entirely but keep the metric for now

stephanos: 👍 I'm good with that

spkane31: Restored

service/history/workflow/update/util.go
```diff
@@ -45,16 +45,18 @@ func (i *instrumentation) countResponseMsg() {
 }
 
 func (i *instrumentation) countRateLimited() {
+	i.metrics.Counter(metrics.WorkflowExecutionUpdateRequestRateLimited.Name()).Record(1, metrics.NamespaceTag("some-namespace"))
```
stephanos: this one has a hard-coded namespace tag

service/history/workflow/update/store.go
```diff
@@ -16,6 +17,7 @@ type (
 		GetUpdateOutcome(ctx context.Context, updateID string) (*updatepb.Outcome, error)
 		GetCurrentVersion() int64
 		IsWorkflowExecutionRunning() bool
+		GetNamespaceEntry() *namespace.Namespace
```
stephanos: What if we pass the namespace into the registry at construction time (and maybe instrumentation, too)? if that works, we don't need this new interface method and can prob remove many of the new parameters, too.

service/history/workflow/update/util.go
stephanos: @claude explain more

stephanos: clarify this

spkane31: I talked with devprod about adding namespace to some metrics, going to remove this until they have time to look into it more.

service/history/workflow/update/store.go
spkane31: Passed into instrumentation, removing decent amount of code

PR #9270, opened 2026-02-10
Remove branch clean up logic on state based replication

xwduan:
## What changed?
Remove branch clean up logic on state based replication

## Why?
Create/UpdateWorkflow API may return error(i.e. context deadline exceeded) when workflow actually successfully created. So we should not just delete the branch 
## How did you test it?
- [x] built
- [ ] added new unit test(s)

## Potential risks
maybe some garbage event/branch left when retry, but it is safer then deletion.

### DIFF
--- a/service/history/ndc/workflow_state_replicator.go
+++ b/service/history/ndc/workflow_state_replicator.go
@@ -453,7 +453,7 @@ func (r *WorkflowStateReplicatorImpl) handleFirstReplicationTaskWithoutNewRun(
 	}
 
 	defer func() {
-		r.deleteNewBranchWhenError(ctx, newBranchToken, retErr)
+		r.deleteNewBranchWhenError(ctx, namespace.ID(executionInfo.NamespaceId), executionInfo.WorkflowId, executionState.RunId, archetypeID, newBranchToken, retErr)
 	}()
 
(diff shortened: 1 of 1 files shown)


service/history/ndc/workflow_state_replicator.go
```diff
@@ -646,12 +646,11 @@ func (r *WorkflowStateReplicatorImpl) deleteNewBranchWhenError(
...
-			ShardID:     r.shardContext.GetShardID(),
-			BranchToken: newBranchToken,
-		}); err != nil {
-			r.logger.Error("failed to clean up workflow execution", tag.Error(err))
-		}
+		// Don't delete the branch - Workflow may created successfully even CreateWorkflow returning error
+		metrics.ReplicationOrphanedHistoryBranch.With(r.shardContext.GetMetricsHandler()).Record(1)
+		r.logger.Warn("skipping history branch cleanup on error - branch may be orphaned",
+			tag.Error(err),
+			tag.ShardID(r.shardContext.GetShardID()))
```
yycptt, line 661: can we log the execution info here as well? namespaceID, workflowID, runID.

yycptt (approved):

PR #9272, opened 2026-02-10
Refresh tasks after reviving a zombie workflow run.

robholland:
## What changed?
Refresh tasks after reviving a zombie workflow run.

## Why?
As the run was suppressed it's tasks may be stale.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

---

> [!NOTE]
> **Medium Risk**
> Touches replication/conflict-resolution transaction paths and task generation behavior; failures could surface as missing/duplicate tasks after revive, though the change is narrow and covered by existing unit tests.
> 
> **Overview**
> When NDC processing revives a workflow run from *zombie* state, it now refreshes the run’s tasks to avoid leaving stale task state after suppression.
> 
> This threads a `workflow.TaskRefresher` through NDC transaction managers and changes `Workflow.Revive` to accept `ctx` + refresher and call `TaskRefresher.Refresh` after updating the workflow state; tests/mocks are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4151c4243aa9dbd8473b166e84031fae3c29adc4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/history/ndc/transaction_manager_existing_workflow.go
```diff
@@ -333,6 +337,15 @@ func (r *nDCTransactionMgrForExistingWorkflowImpl) suppressCurrentAndUpdateAsCur
 		newWorkflowPolicy = historyi.TransactionPolicyPassive.Ptr()
 	}
 
+	if err := r.taskRefresher.Refresh(ctx, targetWorkflow.GetMutableState(), false); err != nil {
```
yux0: Is it possible to have this in Revive()?

service/history/ndc/workflow.go
```diff
@@ -132,7 +131,10 @@ func (r *WorkflowImpl) Revive() error {
 		state,
 		enumspb.WORKFLOW_EXECUTION_STATUS_RUNNING,
 	)
-	return err
+	if err != nil {
+		return err
+	}
+	return taskRefresher.Refresh(ctx, r.mutableState, false)
```
yux0, line 137: can we just do wfContext.RefreshTasks()?

yux0: You may have to update the NewWorkflow and DI the shard context.

robholland: We should not use `wfContext.RefreshTasks()`, it re-loads mutable state and also has persistence calls which we don't want in the task manager use case.

PR #9275, opened 2026-02-10
Change duplicate task error type to already exists

yux0:
## What changed?
Change duplicate task error type to already exists

## Why?
This should not be a unknown error and log it in every request

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/service/history/api/verifychildworkflowcompletionrecorded/api.go
+++ b/service/history/api/verifychildworkflowcompletionrecorded/api.go
@@ -160,7 +160,9 @@ func Invoke(
 	}
 	err = engine.ReplicateVersionedTransition(ctx, chasm.WorkflowArchetypeID, resp.VersionedTransitionArtifact, activeClusterName)
 	if err != nil {
-		return nil, err
+		if !errors.Is(err, consts.ErrDuplicate) {
+			return nil, err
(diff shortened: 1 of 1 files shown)


xwduan:
I think we also need to handle the error [here](https://github.com/temporalio/temporal/blob/main/service/history/api/verifychildworkflowcompletionrecorded/api.go). And try to verify again. Duplicate error does not guarantee parent has the child completion, it just indicating the workflow parent workflow is there.

Besides, we also need to make sure the error is properly handled at replication task, otherwise we may see DLQ.

yux0:
> I think we also need to handle the error [here](https://github.com/temporalio/temporal/blob/main/service/history/api/verifychildworkflowcompletionrecorded/api.go). And try to verify again. Duplicate error does not guarantee parent has the child completion, it just indicating the workflow parent workflow is there.
> 
> Besides, we also need to make sure the error is properly handled at replication task, otherwise we may see DLQ.

I am not sure about the replication task. We have handled it in https://github.com/temporalio/temporal/blob/172726236b6f891d6214ae967dbc8b274544c926/service/history/replication/executable_sync_versioned_transition_task.go#L128?

PR #9277, opened 2026-02-10
Remove client cache in RPCFactory

prathyushpv:
## What changed?
Remove client cache from RPCFactory.

## Why?
This cache was added to avoid creating multiple connections to the same host for different services. But this cache does not have a way to identify when to close a connection and remove it. So it accumulates connections. This defeats the purpose of having this cache to reduce the number of connections. It just keep accumulating stale connections instead. All clients that uses RPCFactory have their own cache. They can listen to membership changes and cleanup stale connections.

## How did you test it?
- [x] built
- [ ] added new unit test(s)


dnr (commented):
Can we use a refcount somehow to know when to remove entries?

prathyushpv:
Yes. We can return a releaseFn like workflow cache which clients can call 🤔

prathyushpv:
Yes, refcount would work, but I think removing this cache is simpler. Clients are already cached at the service level (history, matching), so this RPCFactory cache only helps when multiple services share a process — a case where the connection and memory savings are small. Adding refcount logic to RPCFactory introduces non-trivial bookkeeping for that small gain, so removing the cache seems like the better tradeoff.

simvlad (approved):
Accepting, but I have no idea what this cache can do. On the source `RPCFactory` is constructed once per service, so history, matching will get a different cache. The entries use hostName, but I think hostName is really host:port. I believe history/matching would use different ports on destination, so this is also not helping.

dnr (approved):

PR #9278, opened 2026-02-10
Nexus ScheduleCommand in CHASM

stephanos:
## What changed?

Ported command handler for Nexus "schedule" command from HSM to CHASM.

## Why?

CHASM migration.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/service/history/interfaces/shard_context_mock.go
+++ b/service/history/interfaces/shard_context_mock.go
@@ -135,20 +135,6 @@ func (mr *MockShardContextMockRecorder) ChasmRegistry() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ChasmRegistry", reflect.TypeOf((*MockShardContext)(nil).ChasmRegistry))
 }
 
-// EndpointRegistry mocks base method.
-func (m *MockShardContext) EndpointRegistry() chasm.EndpointRegistry {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "EndpointRegistry")
-	ret0, _ := ret[0].(chasm.EndpointRegistry)
-	return ret0
-}
-
-// EndpointRegistry indicates an expected call of EndpointRegistry.
-func (mr *MockShardContextMockRecorder) EndpointRegistry() *gomock.Call {
-	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "EndpointRegistry", reflect.TypeOf((*MockShardContext)(nil).EndpointRegistry))
-}
-
 // ConflictResolveWorkflowExecution mocks base method.
 func (m *MockShardContext) ConflictResolveWorkflowExecution(ctx context.Context, request *persistence0.ConflictResolveWorkflowExecutionRequest) (*persistence0.ConflictResolveWorkflowExecutionResponse, error) {
 	m.ctrl.T.Helper()
@@ -205,6 +191,20 @@ func (m *MockShardContext) DeleteWorkflowExecution(ctx context.Context, workflow
 func (mr *MockShardContextMockRecorder) DeleteWorkflowExecution(ctx, workflowKey, archetypeID, branchToken, closeExecutionVisibilityTaskID, workflowCloseTime, stage any) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DeleteWorkflowExecution", reflect.TypeOf((*MockShardContext)(nil).DeleteWorkflowExecution), ctx, workflowKey, archetypeID, branchToken, closeExecutionVisibilityTaskID, workflowCloseTime, stage)
+}
+
+// EndpointRegistry mocks base method.
+func (m *MockShardContext) EndpointRegistry() chasm.EndpointRegistry {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "EndpointRegistry")
+	ret0, _ := ret[0].(chasm.EndpointRegistry)
+	return ret0
+}
+
+// EndpointRegistry indicates an expected call of EndpointRegistry.
+func (mr *MockShardContextMockRecorder) EndpointRegistry() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "EndpointRegistry", reflect.TypeOf((*MockShardContext)(nil).EndpointRegistry))
 }
 
 // GenerateTaskID mocks base method.
@@ -926,20 +926,6 @@ func (mr *MockControllableContextMockRecorder) ChasmRegistry() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ChasmRegistry", reflect.TypeOf((*MockControllableContext)(nil).ChasmRegistry))
 }
 
-// EndpointRegistry mocks base method.
-func (m *MockControllableContext) EndpointRegistry() chasm.EndpointRegistry {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "EndpointRegistry")
-	ret0, _ := ret[0].(chasm.EndpointRegistry)
-	return ret0
-}
-
-// EndpointRegistry indicates an expected call of EndpointRegistry.
-func (mr *MockControllableContextMockRecorder) EndpointRegistry() *gomock.Call {
-	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "EndpointRegistry", reflect.TypeOf((*MockControllableContext)(nil).EndpointRegistry))
-}
-
 // ConflictResolveWorkflowExecution mocks base method.
 func (m *MockControllableContext) ConflictResolveWorkflowExecution(ctx context.Context, request *persistence0.ConflictResolveWorkflowExecutionRequest) (*persistence0.ConflictResolveWorkflowExecutionResponse, error) {
 	m.ctrl.T.Helper()
@@ -996,6 +982,20 @@ func (m *MockControllableContext) DeleteWorkflowExecution(ctx context.Context, w
 func (mr *MockControllableContextMockRecorder) DeleteWorkflowExecution(ctx, workflowKey, archetypeID, branchToken, closeExecutionVisibilityTaskID, workflowCloseTime, stage any) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DeleteWorkflowExecution", reflect.TypeOf((*MockControllableContext)(nil).DeleteWorkflowExecution), ctx, workflowKey, archetypeID, branchToken, closeExecutionVisibilityTaskID, workflowCloseTime, stage)
+}
+
+// EndpointRegistry mocks base method.
+func (m *MockControllableContext) EndpointRegistry() chasm.EndpointRegistry {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "EndpointRegistry")
+	ret0, _ := ret[0].(chasm.EndpointRegistry)
+	return ret0
+}
+
+// EndpointRegistry indicates an expected call of EndpointRegistry.
+func (mr *MockControllableContextMockRecorder) EndpointRegistry() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "EndpointRegistry", reflect.TypeOf((*MockControllableContext)(nil).EndpointRegistry))
 }
 
 // FinishStop mocks base method.


chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	return registry.Register(
+		enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION,
+		handleCancelCommand,
+	)
+}
+
+func (h *commandHandler) handleScheduleCommand(
+	chasmCtx chasm.MutableContext,
+	wf *workflow.Workflow,
```
stephanos: Had to change the `Handler` interface to gain access to workflow's `Operations`. I've considered a few options here and this one felt the most natural. Happy to discuss.

chasm/context.go
```diff
@@ -26,12 +30,18 @@ type Context interface {
...
+	// GetNamespaceEntry returns the namespace entry for the execution.
+	GetNamespaceEntry() *namespace.Namespace
+	// GetExecutionInfo returns the execution info for the execution.
+	GetExecutionInfo() *persistencespb.WorkflowExecutionInfo
 
 	// Intent() OperationIntent
 	// ComponentOptions(Component) []ComponentOption
 
+	// GetContext returns the underlying context.Context for use in I/O operations (e.g., RPC calls).
+	GetContext() context.Context
```
stephanos: This had to become exported because of the `h.endpointRegistry.GetByName` call in the command handler. I've considered a few alternative options (e.g. adding `ctx` to the handler interface, adding `GetNexusEndpoint` to `MutableContext` ...); but this seemed like the best in the end. Happy to discuss!

components/nexusoperations/workflow/commands.go
```diff
@@ -102,15 +102,15 @@ func (ch *commandHandler) HandleScheduleCommand(
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.ScheduleToStartTimeout); err != nil {
-		return workflow.FailWorkflowTaskError{
+		return chasmcommand.FailWorkflowTaskError{
```
stephanos, line 138: This had moved in the last PR.

chasm/lib/nexusoperation/operation_statemachine.go
```diff
@@ -10,7 +10,7 @@ import (
 type EventScheduled struct {
 }
 
-var transitionScheduled = chasm.NewTransition(
+var TransitionScheduled = chasm.NewTransition(
```
stephanos: Required to be accessible from `chasm/lib/nexusoperation/workflow/commands.go`

chasm/lib/nexusoperation/workflow/fx.go
```diff
@@ -0,0 +1,27 @@
...
+	commonnexus "go.temporal.io/server/common/nexus"
+	"go.uber.org/fx"
+)
+
+type registerParams struct {
+	fx.In
+
+	Registry         *command.Registry
+	Config           *nexusoperation.Config
+	EndpointRegistry commonnexus.EndpointRegistry `optional:"true"`
```
stephanos: Without this, `matching` doesn't start.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -0,0 +1,256 @@
...
+	); err != nil {
+		return err
+	}
+	return registry.Register(
+		enumspb.COMMAND_TYPE_REQUEST_CANCEL_NEXUS_OPERATION,
+		handleCancelCommand,
+	)
+}
+
+func (ch *commandHandler) handleScheduleCommand(
```
stephanos, line 54: migrated from `components/nexusoperations/workflow/commands.go`

stephanos: <details>
<summary>diff</summary>

```
--- components/nexusoperations/workflow/commands.go
+++ chasm/lib/nexusoperation/workflow/commands.go
@@ -1,38 +1,38 @@
-func (ch *commandHandler) HandleScheduleCommand(
-	ctx context.Context,
-	ms historyi.MutableState,
-	validator chasmcommand.Validator,
-	workflowTaskCompletedEventID int64,
-	command *commandpb.Command,
+func (ch *commandHandler) handleScheduleCommand(
+	chasmCtx chasm.MutableContext,
+	wf *workflow.Workflow,
+	validator command.Validator,
+	cmd *commandpb.Command,
+	opts command.HandlerOptions,
 ) error {
-	ns := ms.GetNamespaceEntry()
-	nsName := ms.GetNamespaceEntry().Name().String()
+	ns := chasmCtx.GetNamespaceEntry()
+	nsName := ns.Name().String()
 
 	if !ch.config.Enabled() {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_FEATURE_DISABLED,
 			Message: "Nexus operations disabled",
 		}
 	}
 
-	attrs := command.GetScheduleNexusOperationCommandAttributes()
+	attrs := cmd.GetScheduleNexusOperationCommandAttributes()
 	if attrs == nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: "empty ScheduleNexusOperationCommandAttributes",
 		}
 	}
 
 	var endpointID string
-	endpoint, err := ch.endpointRegistry.GetByName(ctx, ns.ID(), attrs.Endpoint)
+	endpoint, err := ch.endpointRegistry.GetByName(chasmCtx.GetContext(), ns.ID(), attrs.Endpoint)
 	if err != nil {
 		if errors.As(err, new(*serviceerror.NotFound)) {
-			return chasmcommand.FailWorkflowTaskError{
+			return command.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 				Message: fmt.Sprintf("endpoint %q not found", attrs.Endpoint),
 			}
 		} else if errors.As(err, new(*serviceerror.PermissionDenied)) {
-			return chasmcommand.FailWorkflowTaskError{
+			return command.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 				Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.Name(), attrs.Endpoint),
 			}
@@ -44,7 +44,7 @@
 	}
 
 	if len(attrs.Service) > ch.config.MaxServiceNameLength(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.Service exceeds length limit of %d",
@@ -54,7 +54,7 @@
 	}
 
 	if len(attrs.Operation) > ch.config.MaxOperationNameLength(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.Operation exceeds length limit of %d",
@@ -64,7 +64,7 @@
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.ScheduleToCloseTimeout); err != nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.ScheduleToCloseTimeout is invalid: %v", err),
@@ -72,7 +72,7 @@
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.ScheduleToStartTimeout); err != nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.ScheduleToStartTimeout is invalid: %v", err),
@@ -80,7 +80,7 @@
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.StartToCloseTimeout); err != nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.StartToCloseTimeout is invalid: %v", err),
@@ -88,7 +88,7 @@
 	}
 
 	if !validator.IsValidPayloadSize(attrs.Input.Size()) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:             enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message:           "ScheduleNexusOperationCommandAttributes.Input exceeds size limit",
 			TerminateWorkflow: true,
@@ -102,7 +102,7 @@
 		lowerCaseHeader[lowerK] = v
 		headerLength += len(lowerK) + len(v)
 		if slices.Contains(ch.config.DisallowedOperationHeaders(), lowerK) {
-			return chasmcommand.FailWorkflowTaskError{
+			return command.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 				Message: fmt.Sprintf("ScheduleNexusOperationCommandAttributes.NexusHeader contains a disallowed header key: %q", k),
 			}
@@ -110,28 +110,26 @@
 	}
 
 	if headerLength > ch.config.MaxOperationHeaderSize(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: "ScheduleNexusOperationCommandAttributes.NexusHeader exceeds size limit",
 		}
 	}
 
-	root := ms.HSM()
-	coll := nexusoperations.MachineCollection(root)
 	maxPendingOperations := ch.config.MaxConcurrentOperations(nsName)
-	if coll.Size() >= ch.config.MaxConcurrentOperations(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+	if wf.PendingNexusOperationCount() >= maxPendingOperations {
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_PENDING_NEXUS_OPERATIONS_LIMIT_EXCEEDED,
 			Message: fmt.Sprintf("workflow has reached the pending nexus operation limit of %d for this namespace", maxPendingOperations),
 		}
 	}
 
 	// Trim timeout to workflow run timeout.
-	runTimeout := ms.GetExecutionInfo().WorkflowRunTimeout.AsDuration()
+	runTimeout := wf.WorkflowRunTimeout()
 	opTimeout := attrs.ScheduleToCloseTimeout.AsDuration()
 	if runTimeout > 0 && (opTimeout == 0 || opTimeout > runTimeout) {
-		attrs.ScheduleToCloseTimeout = ms.GetExecutionInfo().WorkflowRunTimeout
-		opTimeout = attrs.ScheduleToCloseTimeout.AsDuration()
+		attrs.ScheduleToCloseTimeout = durationpb.New(runTimeout)
+		opTimeout = runTimeout
 	}
 
 	// Trim timeout to max allowed timeout.
@@ -153,7 +151,8 @@
 		}
 	}
 
-	event := ms.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED, func(he *historypb.HistoryEvent) {
+	requestID := uuid.NewString()
+	event := chasmCtx.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED, func(he *historypb.HistoryEvent) {
 		he.Attributes = &historypb.HistoryEvent_NexusOperationScheduledEventAttributes{
 			NexusOperationScheduledEventAttributes: &historypb.NexusOperationScheduledEventAttributes{
 				Endpoint:                     attrs.Endpoint,
@@ -165,12 +164,35 @@
 				ScheduleToStartTimeout:       attrs.ScheduleToStartTimeout,
 				StartToCloseTimeout:          attrs.StartToCloseTimeout,
 				NexusHeader:                  lowerCaseHeader,
-				RequestId:                    uuid.NewString(),
-				WorkflowTaskCompletedEventId: workflowTaskCompletedEventID,
+				RequestId:                    requestID,
+				WorkflowTaskCompletedEventId: opts.WorkflowTaskCompletedEventID,
 			},
 		}
-		he.UserMetadata = command.UserMetadata
+		he.UserMetadata = cmd.UserMetadata
 	})
 
-	return nexusoperations.ScheduledEventDefinition{}.Apply(root, event)
+	scheduledTime := event.GetEventTime()
+	if scheduledTime == nil {
+		scheduledTime = timestamppb.Now()
+	}
+
+	op := nexusoperation.NewOperation(&nexusoperationpb.OperationState{
+		EndpointId:             endpointID,
+		Endpoint:               attrs.Endpoint,
+		Service:                attrs.Service,
+		Operation:              attrs.Operation,
+		ScheduledTime:          scheduledTime,
+		ScheduleToCloseTimeout: attrs.ScheduleToCloseTimeout,
+		RequestId:              requestID,
+		Attempt:                1,
+	})
+
+	if err := nexusoperation.TransitionScheduled.Apply(op, chasmCtx, nexusoperation.EventScheduled{}); err != nil {
+		return err
+	}
+
+	key := strconv.FormatInt(event.GetEventId(), 10)
+	wf.AddNexusOperation(chasmCtx, key, op)
+
+	return nil
 }

```
</details>

chasm/lib/nexusoperation/proto/v1/operation.proto
```diff
@@ -2,13 +2,56 @@ syntax = "proto3";
 
 package temporal.server.chasm.lib.nexusoperation.proto.v1;
 
+import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 import "temporal/api/failure/v1/message.proto";
 
 option go_package = "go.temporal.io/server/chasm/lib/nexusoperation/gen/nexusoperationpb;nexusoperationpb";
 
 message OperationState {
```
stephanos, line 11: Made this the exact same as in Chetan's PR https://github.com/temporalio/temporal/pull/9081/changes#diff-acb4219132cc3df0b2bedbcfe29621174e7fa552f2dd1c02834ce096f1be5f8f

chasm/lib/workflow/workflow.go
```diff
@@ -118,6 +118,17 @@ func (w *Workflow) AddCompletionCallbacks(
...
+	if w.Operations == nil {
+		w.Operations = make(chasm.Map[string, *nexusoperation.Operation])
+	}
+	w.Operations[key] = chasm.NewComponentField(ctx, op)
+}
+
+func (w *Workflow) PendingNexusOperationCount() int {
+	return len(w.Operations)
+}
+
```
stephanos, line 141: ^ Technically these are not needed as `Operations` is exported; but it's my understanding that that is for technical reasons only and this makes for better encapsulation.

chasm/lib/nexusoperation/workflow/commands_test.go
```diff
@@ -0,0 +1,591 @@
...
+//   - operation not found
+//   - operation already completed
+//   - operation already completed - completion buffered (requires buffered event support)
+//   - sets event attributes with UserMetadata and spawns cancelation child machine
+
+// TODO: TestOperationNodeDeletionOnTerminalEvents - port when CHASM operation lifecycle is implemented:
+//   - completed event deletes node
+//   - failed event deletes node
+//   - canceled event deletes node
+//   - timed out event deletes node
```
stephanos: ^ ported from `components/nexusoperations/workflow/commands_test.go`

chasm/lib/workflow/command/handler.go
```diff
@@ -14,6 +15,7 @@ type HandlerOptions struct {
 // worker request.
 type Handler func(
 	chasmCtx chasm.MutableContext,
+	wf *chasmworkflow.Workflow,
```
stephanos, line 18: Had to change the `Handler` interface to gain access to workflow's `Operations`. I've considered a few options here and this one felt the most natural. Happy to discuss.

chasm/lib/nexusoperation/workflow/commands_test.go
```diff
@@ -0,0 +1,880 @@
+package workflow
```
stephanos, line 1: Ported from `components/nexusoperations/workflow/commands_test.go`

stephanos: <details>
<summary>diff</summary>

```
--- components/nexusoperations/workflow/commands.go
+++ chasm/lib/nexusoperation/workflow/commands.go
@@ -1,38 +1,38 @@
-func (ch *commandHandler) HandleScheduleCommand(
-	ctx context.Context,
-	ms historyi.MutableState,
-	validator chasmcommand.Validator,
-	workflowTaskCompletedEventID int64,
-	command *commandpb.Command,
+func (ch *commandHandler) handleScheduleCommand(
+	chasmCtx chasm.MutableContext,
+	wf *workflow.Workflow,
+	validator command.Validator,
+	cmd *commandpb.Command,
+	opts command.HandlerOptions,
 ) error {
-	ns := ms.GetNamespaceEntry()
-	nsName := ms.GetNamespaceEntry().Name().String()
+	ns := chasmCtx.GetNamespaceEntry()
+	nsName := ns.Name().String()
 
 	if !ch.config.Enabled() {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_FEATURE_DISABLED,
 			Message: "Nexus operations disabled",
 		}
 	}
 
-	attrs := command.GetScheduleNexusOperationCommandAttributes()
+	attrs := cmd.GetScheduleNexusOperationCommandAttributes()
 	if attrs == nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: "empty ScheduleNexusOperationCommandAttributes",
 		}
 	}
 
 	var endpointID string
-	endpoint, err := ch.endpointRegistry.GetByName(ctx, ns.ID(), attrs.Endpoint)
+	endpoint, err := ch.endpointRegistry.GetByName(chasmCtx.GetContext(), ns.ID(), attrs.Endpoint)
 	if err != nil {
 		if errors.As(err, new(*serviceerror.NotFound)) {
-			return chasmcommand.FailWorkflowTaskError{
+			return command.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 				Message: fmt.Sprintf("endpoint %q not found", attrs.Endpoint),
 			}
 		} else if errors.As(err, new(*serviceerror.PermissionDenied)) {
-			return chasmcommand.FailWorkflowTaskError{
+			return command.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 				Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.Name(), attrs.Endpoint),
 			}
@@ -44,7 +44,7 @@
 	}
 
 	if len(attrs.Service) > ch.config.MaxServiceNameLength(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.Service exceeds length limit of %d",
@@ -54,7 +54,7 @@
 	}
 
 	if len(attrs.Operation) > ch.config.MaxOperationNameLength(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.Operation exceeds length limit of %d",
@@ -64,7 +64,7 @@
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.ScheduleToCloseTimeout); err != nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.ScheduleToCloseTimeout is invalid: %v", err),
@@ -72,7 +72,7 @@
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.ScheduleToStartTimeout); err != nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.ScheduleToStartTimeout is invalid: %v", err),
@@ -80,7 +80,7 @@
 	}
 
 	if err := timestamp.ValidateAndCapProtoDuration(attrs.StartToCloseTimeout); err != nil {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause: enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: fmt.Sprintf(
 				"ScheduleNexusOperationCommandAttributes.StartToCloseTimeout is invalid: %v", err),
@@ -88,7 +88,7 @@
 	}
 
 	if !validator.IsValidPayloadSize(attrs.Input.Size()) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:             enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message:           "ScheduleNexusOperationCommandAttributes.Input exceeds size limit",
 			TerminateWorkflow: true,
@@ -102,7 +102,7 @@
 		lowerCaseHeader[lowerK] = v
 		headerLength += len(lowerK) + len(v)
 		if slices.Contains(ch.config.DisallowedOperationHeaders(), lowerK) {
-			return chasmcommand.FailWorkflowTaskError{
+			return command.FailWorkflowTaskError{
 				Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 				Message: fmt.Sprintf("ScheduleNexusOperationCommandAttributes.NexusHeader contains a disallowed header key: %q", k),
 			}
@@ -110,28 +110,26 @@
 	}
 
 	if headerLength > ch.config.MaxOperationHeaderSize(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
 			Message: "ScheduleNexusOperationCommandAttributes.NexusHeader exceeds size limit",
 		}
 	}
 
-	root := ms.HSM()
-	coll := nexusoperations.MachineCollection(root)
 	maxPendingOperations := ch.config.MaxConcurrentOperations(nsName)
-	if coll.Size() >= ch.config.MaxConcurrentOperations(nsName) {
-		return chasmcommand.FailWorkflowTaskError{
+	if len(wf.Operations) >= maxPendingOperations {
+		return command.FailWorkflowTaskError{
 			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_PENDING_NEXUS_OPERATIONS_LIMIT_EXCEEDED,
 			Message: fmt.Sprintf("workflow has reached the pending nexus operation limit of %d for this namespace", maxPendingOperations),
 		}
 	}
 
 	// Trim timeout to workflow run timeout.
-	runTimeout := ms.GetExecutionInfo().WorkflowRunTimeout.AsDuration()
+	runTimeout := wf.WorkflowRunTimeout()
 	opTimeout := attrs.ScheduleToCloseTimeout.AsDuration()
 	if runTimeout > 0 && (opTimeout == 0 || opTimeout > runTimeout) {
-		attrs.ScheduleToCloseTimeout = ms.GetExecutionInfo().WorkflowRunTimeout
-		opTimeout = attrs.ScheduleToCloseTimeout.AsDuration()
+		attrs.ScheduleToCloseTimeout = durationpb.New(runTimeout)
+		opTimeout = runTimeout
 	}
 
 	// Trim timeout to max allowed timeout.
@@ -153,7 +151,7 @@
 		}
 	}
 
-	event := ms.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED, func(he *historypb.HistoryEvent) {
+	event := chasmCtx.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_SCHEDULED, func(he *historypb.HistoryEvent) {
 		he.Attributes = &historypb.HistoryEvent_NexusOperationScheduledEventAttributes{
 			NexusOperationScheduledEventAttributes: &historypb.NexusOperationScheduledEventAttributes{
 				Endpoint:                     attrs.Endpoint,
@@ -166,11 +164,33 @@
 				StartToCloseTimeout:          attrs.StartToCloseTimeout,
 				NexusHeader:                  lowerCaseHeader,
 				RequestId:                    uuid.NewString(),
-				WorkflowTaskCompletedEventId: workflowTaskCompletedEventID,
+				WorkflowTaskCompletedEventId: opts.WorkflowTaskCompletedEventID,
 			},
 		}
-		he.UserMetadata = command.UserMetadata
+		he.UserMetadata = cmd.UserMetadata
 	})
 
-	return nexusoperations.ScheduledEventDefinition{}.Apply(root, event)
+	scheduledEventID := event.GetEventId()
+	scheduledTime := event.GetEventTime()
+	if scheduledTime == nil {
+		scheduledTime = timestamppb.Now()
+	}
+
+	op := nexusoperation.NewOperation(&nexusoperationpb.OperationState{
+		ScheduledTime:    scheduledTime,
+		ScheduledEventId: scheduledEventID,
+		Attempt:          1,
+	})
+
+	if err := nexusoperation.TransitionScheduled.Apply(op, chasmCtx, nexusoperation.EventScheduled{}); err != nil {
+		return err
+	}
+
+	key := strconv.FormatInt(scheduledEventID, 10)
+	if wf.Operations == nil {
+		wf.Operations = make(chasm.Map[string, *nexusoperation.Operation])
+	}
+	wf.Operations[key] = chasm.NewComponentField(chasmCtx, op)
+
+	return nil
 }

```
</details>

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -0,0 +1,269 @@
+package workflow
```
stephanos, line 1: This had to be moved here from `chasm/lib/nexusoperation/commands.go` to break the cyclical import dependency: CHASM `workflow` needs `nexusoperation`'s `Operation` - but `nexusoperation` (now) required CHASM `workflow`'s `Workflow` to access the workflow's operations.

chasm/lib/nexusoperation/proto/v1/operation.proto
```diff
@@ -2,13 +2,56 @@ syntax = "proto3";
...
+
+  // Operation name.
+  string operation = 5;
+
+  // The time when the operation was scheduled.
+  google.protobuf.Timestamp scheduled_time = 6;
+
+  // Schedule-to-close timeout for this operation.
+  // This is the only timeout settable by a workflow.
+  google.protobuf.Duration schedule_to_close_timeout = 7;
```
bergundy: You're going to want to add schedule_to_start and start_to_close at this point.

chasm/lib/nexusoperation/proto/v1/operation.proto
```diff
@@ -2,13 +2,56 @@ syntax = "proto3";
...
+  string service = 4;
+
+  // Operation name.
+  string operation = 5;
+
+  // The time when the operation was scheduled.
+  google.protobuf.Timestamp scheduled_time = 6;
+
+  // Schedule-to-close timeout for this operation.
+  // This is the only timeout settable by a workflow.
```
bergundy:
```suggestion
```

chasm/lib/nexusoperation/proto/v1/operation.proto
```diff
@@ -2,13 +2,56 @@ syntax = "proto3";
...
+  google.protobuf.Duration schedule_to_close_timeout = 7;
+
+  // Unique request ID allocated for all retry attempts of the StartOperation request.
+  string request_id = 8;
+
+  // Token for fetching the scheduled event from history.
+  bytes scheduled_event_token = 9;
```
bergundy: We need to finalize this concept in CHASM. Let's sync in our crew channel.

chasm/lib/nexusoperation/workflow/commands.go
bergundy: Makes sense. IMHO it's worth keeping the implementation inside the lib the way you have it. The other alternative is to have a big switch case statement in the workflow component, which would remove the need for this command registry but I would rather keep this structure. It allows workflow to potentially be extended by arbitrary components.

chasm/context.go
bergundy: We don't want this exposed in other cases. The command handler is an API handler that should have access to a Go context. I would rather expose the endpoint registry via the context instead.

chasm/context.go
```diff
@@ -26,12 +29,16 @@ type Context interface {
 	ExecutionCloseTime() time.Time
 	// Logger returns a logger tagged with execution key and other chasm framework internal information.
 	Logger() log.Logger
+	// GetNamespaceEntry returns the namespace entry for the execution.
+	GetNamespaceEntry() *namespace.Namespace
```
bergundy:
```suggestion
	// NamespaceEntry returns the namespace entry for the execution.
	NamespaceEntry() *namespace.Namespace
```

chasm/context.go
```diff
@@ -42,8 +49,8 @@ type MutableContext interface {
 	// referencing the component.
 	AddTask(Component, TaskAttributes, any)
 
-	// Add more methods here for other storage commands/primitives.
-	// e.g. HistoryEvent
+	// AddHistoryEvent adds a history event for the execution.
+	AddHistoryEvent(t enumspb.EventType, setAttributes func(*historypb.HistoryEvent)) *historypb.HistoryEvent
```
bergundy: This should be a method of the `Workflow` component. You can expose it via the `MSPointer` struct.

chasm/lib/nexusoperation/workflow/commands.go
bergundy: You can create an interface that exposes just the methods you need from `Workflow` IMHO and keep everything in the same package.

chasm/lib/nexusoperation/workflow/fx.go
bergundy: This is because we are registering the workflow lib in matching unnecessarily. Let's avoid that.

chasm/lib/workflow/command/handler.go
bergundy: I would define this as an interface to prevent the circular dependency issue.

chasm/lib/nexusoperation/workflow/commands.go
bergundy: But I'm on the fence about this. I want to meet and consider the alternatives.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -0,0 +1,269 @@
...
+		Service:                attrs.Service,
+		Operation:              attrs.Operation,
+		ScheduledTime:          scheduledTime,
+		ScheduleToCloseTimeout: attrs.ScheduleToCloseTimeout,
+		RequestId:              requestID,
+		ScheduledEventToken:    eventToken,
+		Attempt:                1,
+	})
+
+	if err := nexusoperation.TransitionScheduled.Apply(op, chasmCtx, nexusoperation.EventScheduled{}); err != nil {
```
bergundy: It would be good to have this logic be part of history event application so it's not duplicated. In HSM we made some compromises here but we don't need to carry those to CHASM.

chasm/lib/workflow/command/handler.go
bergundy: There are other approaches too. Because you already have the workflow in a parent pointer, you could potentially use that.

bergundy (commented):
I added some comments but I have a more meta comment.
I am now wondering if making everything so modular is worth it. I really like making event and command definitions something that is registrable and and adhere to a specific interface. We can extend these definitions later and use them for action attribution and other use cases.

chasm/context.go
stephanos: @bergundy to clarify, this would be in temporal/service/history/workflow/mutable_state_impl.go then, right?

```
func (ms *MutableStateImpl) GetNamespaceEntry() *namespace.Namespace {
	return ms.namespaceEntry
}

func (ms *MutableStateImpl) NamespaceEntry() *namespace.Namespace {
	return ms.namespaceEntry
}
```

bergundy: It doesn't have to propagate to mutable state.

stephanos: 👍 I have it convert it now in `immutableCtx` from one to the other

chasm/lib/nexusoperation/workflow/fx.go
stephanos: @bergundy So I pushed on this for a good while and I'm not sure there's a simple change here. The `ChasmLibraryOptions` is used across prod and test code; and taking it apart requires some nuance. For example, Frontend seems to require parts of it (to my surprise). I'd like to only use what's needed in each service; but then I'd also have to duplicate it (onebox 🙄 ) and it'll become a larger change. Follow up PR?

chasm/context.go
stephanos: renamed in https://github.com/temporalio/temporal/pull/9278/changes/14cb55f0d1c2682dbe9e5b941da628d9cde888ac

chasm/context.go
stephanos: addressed in https://github.com/temporalio/temporal/pull/9278/changes/09c3ed4624db1067a17dc7a6636fefefb53422df

chasm/context.go
stephanos: @bergundy Right, that's what I had considered - but are we okay with coupling CHASM's `Context` to Nexus that way? That didn't seem right to me.

Another option I considered was adding `ctx` to `Handler` func type - but then you have both `chasmCtx` and `ctx`; seemed odd.

Having said that, adding Nexus to CHASM's Context might be the lesser evil.

chasm/lib/nexusoperation/workflow/commands.go
stephanos: Dropped for now in https://github.com/temporalio/temporal/pull/9278/commits/569a64e840786b09a9860f386414ec4a2abd3a59

chasm/lib/workflow/command/handler.go
stephanos: @bergundy would you prefer that over the separate package? Interfaces ofc add more indirection and CHASM already has its fair share of interfaces making it harder to navigate. I like the explicitness of this; with a minor tradeoff of a sub package. But that's a preference, no strong objection.

PS: I don't think the Nexus schedule command handler has a parent pointer since there's no `Operation` yet?

chasm/context.go
stephanos: I tried it out now in https://github.com/temporalio/temporal/pull/9278/changes/b4e79f54aa27e8c5f909b1af55a37f712b1fd4de

chasm/lib/nexusoperation/workflow/fx.go
bergundy: Do you even need the endpoint registry anymore? Seems like you can just get it from the CHASM context now.

chasm/lib/workflow/command/handler.go
bergundy: True. Let's keep as is. That's also what we tended to converge on over zoom.

chasm/lib/workflow/workflow.go
bergundy: Here my take:
Seems redundant since it's likely going to be only called in the `nexusoperation/workflow` package which is essentially the integration point between both components. I don't like adding bloat to this shared class unnecessarily.

chasm/context.go
```diff
@@ -26,12 +28,22 @@ type Context interface {
...
 	// ComponentOptions(Component) []ComponentOption
 
-	structuredRef(Component) (ComponentRef, error)
+	// getContext returns the underlying context.Context for use in I/O operations (e.g., RPC calls).
 	getContext() context.Context
+
+	structuredRef(Component) (ComponentRef, error)
+}
+
+type EndpointRegistry interface {
```
bergundy, line 56: Do you need this separate interface? It's already defined in `common/nexus`.

chasm/lib/workflow/workflow.go
bergundy: That being said we may need to count both HSM and CHASM operations here to enforce the max which would require extracting more information from the underlying mutable state.

chasm/context.go
stephanos: So there's an import cycle: chasm → common/nexus → common/persistence → chasm :-/

chasm/lib/nexusoperation/workflow/fx.go
stephanos: The CHASM context now has the `EndpointByName` method (taking in a `EndpointRegistry`). If it's not injected here it would have to come from somewhere else, like the CHASM context itself. But that would mean the CHASM Context constructor (`NewContext`) now needs it; which I think it the wrong direction.

chasm/lib/workflow/workflow.go
stephanos: I'm happy to leave it here and inline it later once we find it's not needed after the migration.

chasm/lib/nexusoperation/workflow/fx.go
bergundy: Doesn't seem true to me.

stephanos: It would be helpful if you clarified which part you are referring to.

So you were saying, "you can just get it from the CHASM context now". `Context` is an interface; it's created via `NewContext`/`newContext`. Right now it only has `ctx context.Context` and `node *Node`. `Node` also has `NodeBackend` but `MutableState` doesn't have access to `EndpointRegistry` either. 

So I'm not really sure what you're suggesting; where do you think `EndpointRegistry` should be coming from?

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -75,24 +77,55 @@ func (ch *commandHandler) handleScheduleCommand(
...
+			if errors.As(err, new(*serviceerror.PermissionDenied)) {
+				return command.FailWorkflowTaskError{
+					Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES,
+					Message: fmt.Sprintf("caller namespace %q unauthorized for %q", ns.Name(), attrs.Endpoint),
+				}
+			}
+			return err
+		}
+		endpointID = endpoint.Id
 	}
```
stephanos, line 127: ^ ported this change over from HSM in `main`

chasm/lib/nexusoperation/workflow/commands_test.go
```diff
@@ -586,6 +587,44 @@ func TestHandleScheduleCommand(t *testing.T) {
...
+				},
+			},
+		}, command.HandlerOptions{WorkflowTaskCompletedEventID: 1, ChasmWorkflow: tcx.wf})
+		var failWFTErr command.FailWorkflowTaskError
+		require.ErrorAs(t, err, &failWFTErr)
+		require.False(t, failWFTErr.TerminateWorkflow)
+		require.Equal(t, enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_SCHEDULE_NEXUS_OPERATION_ATTRIBUTES, failWFTErr.Cause)
+		require.Contains(t, failWFTErr.Message, "not found")
+		require.Empty(t, tcx.history.Events)
+	})
```
stephanos, line 630: ^ ported over from `main` to match new system-endpoint behavior

chasm/lib/nexusoperation/workflow/fx.go
stephanos: I can put it into the `chasm.Registry`; since `NexusEndpointProcessor` is already there - but that has the same issue wrt `optional` and is even worse IMO since it'll be harder to remove in the matching context where it's not there.

stephanos:
There are a lot of threads; I think I've addressed or commented on all.
I've tried some refactorings around the interface vs package threads; but they were all worse IMO.

chasm/lib/nexusoperation/workflow/fx.go
bergundy: 1. We should not be injecting any CHASM libraries into matching or worker services, we should fix this.
2. You added `EndpointByName` to `chasm.Context`, why can't you use that in the command handler?

stephanos: re 1: discussed offline, will file a ticket to follow-up since there's already a work item for CHASM crew to do that.

re 2: discussed offline, will make a change to pipe the EndpointRegistry through mutable state.

stephanos: (2) addressed in https://github.com/temporalio/temporal/pull/9278/changes/ae1cdc25a183038718fc9413e9a65f9b3b384ff7

PR #9285, opened 2026-02-11
Add additional standalone activity request validations

fretz12:
## What changed?
Add additional standalone activity request validations. Renamed blob error tags to be in line with existing workflow tags. Code cleanup. Refactor tests into grouped subtests.

## Why?
After an audit of the standalone activity frontend validations, this PR provides additional coverage of missing corners.

## How did you test it?
- [X] built
- [ ] added new unit test(s)


chasm/lib/activity/validator.go
```diff
@@ -276,18 +273,100 @@ func ValidatePollActivityExecutionRequest(
 	return nil
 }
 
-// ValidateActivityTaskToken validates a task token against the current activity state.
-func ValidateActivityTaskToken(
-	ctx chasm.Context,
-	a *Activity,
-	token *tokenspb.Task,
+// validateRequestCancelActivityExecutionRequest validates RequestCancelActivityExecutionRequest.
```
dandavison: nit: no docstring required since it's not exported. I have no problem with docstrings on non-exported things but in this case it's not adding anything.

```suggestion
```

chasm/lib/activity/validator.go
```diff
@@ -276,18 +273,100 @@ func ValidatePollActivityExecutionRequest(
...
+		logger,
+		req.GetNamespace())
+	if err != nil {
+		return serviceerror.NewInvalidArgument("reason exceeds length limit")
+	}
+
+	return nil
+}
+
+// validateTerminateActivityExecutionRequest validates TerminateActivityExecutionRequest.
```
dandavison:
```suggestion
```

tests/standalone_activity_test.go
```diff
@@ -35,6 +35,7 @@ import (
 const (
 	defaultStartToCloseTimeout = 1 * time.Minute
 	defaultIdentity            = "test-worker"
+	defaultMaxIDLengthLimit    = 1000
```
dandavison: nit: it would be nice to take this from the dynamic config that is in effect; these tests will break when someone changes that.

dandavison (approved):
LGTM, made a few suggestions

chasm/lib/activity/validator.go
fretz12: removed

chasm/lib/activity/validator.go
fretz12: removed

tests/standalone_activity_test.go
fretz12: extracted from DC default directly

PR #9286, opened 2026-02-11
Improve priority semaphore

yycptt:
## What changed?
- When checking waiters, only check waiter with higher or equal priority. 
- Allow specifying priority in TryAcquire()

## Why?
- If there's all waiters have lower priority, we don't really need to create a new waiter and later unblock it.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/common/locks/priority_semaphore_test.go
+++ b/common/locks/priority_semaphore_test.go
@@ -85,6 +85,8 @@ func (s *prioritySemaphoreSuite) TestTryAcquire_HighAfterWaiting() {
 	cLock := make(chan struct{})
 	go func() {
 		// Acquire the function to make the next call blocking.
+		// nolint:testifylint
+		// Must use assert.Assertions instead of require.Assertions here because this is running in a separate goroutine.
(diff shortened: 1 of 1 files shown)


common/locks/priority_semaphore_impl.go
```diff
@@ -156,10 +156,10 @@ func (s *PrioritySemaphoreImpl) Acquire(ctx context.Context, priority Priority,
 
 // TryAcquire acquires the semaphore with a weight of n without blocking.
 // On success, returns true. On failure, returns false and leaves the semaphore unchanged.
-func (s *PrioritySemaphoreImpl) TryAcquire(n int) bool {
+func (s *PrioritySemaphoreImpl) TryAcquire(priority Priority, n int) bool {
 	s.mu.Lock()
```
prathyushpv, line 166: It is good to validate the value of priority like we do in Acquire() here.

yycptt: ah makes sense. will do!

yycptt: Done. added sanity to make sure priority < NumPriority

PR #9287, opened 2026-02-11
Time skip to StickyWorkerUnavailable 

stephanos:
## What changed?

1. Replaced (most) use of `time` package in Matching with virtual `systemClock`.
2. Added separate `taskClock` to Matching to be able to force `StickyWorkerUnavailable` to trigger asap.
3. Replaced `time.Sleep(10*time.Second)` in `TestWorkflowUpdateSuite` with time skipping.

Note that (1) can also be a separate PR; if that helps.

## Why?

Test runtime before: **10.21s**
Test runtime after: **1.3s**

There's some extra overhead for starting the cluster; but as we can see here, it's a worthwhile tradeoff.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

## Potential risks

1. `taskClock` isn't used everywhere it could yet. This might cause confusion since certain things you might expect to be time-skippable won't work. But you have to start somewhere; blocking this effort on identifying every single location first will be difficult.

2. Advancing `taskClock` in tests requires a dedicated cluster. This makes its use expensive. Either we accept that it's a rare use case; or if not, could consider a namespace-aware timesource (provider).


tests/update_workflow_test.go
```diff
@@ -1391,105 +1392,88 @@ func TestWorkflowUpdateSuite(t *testing.T) {
...
-			Logger:                       s.Logger,
-			T:                            s.T(),
-		}
-
-		// Drain existing WT from regular task queue, but respond with sticky enabled response to enable stick task queue.
-		_, err := poller.PollAndProcessWorkflowTask(testcore.WithRespondSticky, testcore.WithoutRetries)
-		s.NoError(err)
-
-		s.Logger.Info("Sleep 10+ seconds to make sure stickyPollerUnavailableWindow time has passed.")
-		time.Sleep(10*time.Second + 100*time.Millisecond) //nolint:forbidigo
```
stephanos, line 1461: So funny thing; this doesn't even work. The 10s are being waited on for nothing as the sticky task queue is never actually created since there's never a poller.

PR #9288, opened 2026-02-11
Nexus CancelCommand in CHASM

stephanos:
## What changed?

Ported command handler for Nexus "cancel" command from HSM to CHASM.

## Why?

CHASM migration.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/lib/nexusoperation/workflow/commands_test.go
+++ b/chasm/lib/nexusoperation/workflow/commands_test.go
@@ -153,6 +153,16 @@ func TestHandleScheduleCommand(t *testing.T) {
 		require.Empty(t, tcx.history.Events)
 	})
 
+	t.Run("chasm nexus not enabled", func(t *testing.T) {
+		tcx := newTestContext(t, &nexusoperation.Config{
+			Enabled:           dynamicconfig.GetBoolPropertyFn(true),
+			ChasmNexusEnabled: dynamicconfig.GetBoolPropertyFnFilteredByNamespace(false),
+		})
+		err := tcx.scheduleHandler(tcx.chasmCtx, tcx.wf, commandValidator{maxPayloadSize: 1}, &commandpb.Command{}, command.HandlerOptions{WorkflowTaskCompletedEventID: 1})
+		require.ErrorIs(t, err, command.ErrNotSupported)
+		require.Empty(t, tcx.history.Events)
+	})
+
 	t.Run("empty attributes", func(t *testing.T) {
 		tcx := newTestContext(t, defaultConfig)
 		err := tcx.scheduleHandler(tcx.chasmCtx, tcx.wf, commandValidator{maxPayloadSize: 1}, &commandpb.Command{}, command.HandlerOptions{WorkflowTaskCompletedEventID: 1})
@@ -650,6 +660,16 @@ func TestHandleCancelCommand(t *testing.T) {
 		require.ErrorAs(t, err, &failWFTErr)
 		require.False(t, failWFTErr.TerminateWorkflow)
 		require.Equal(t, enumspb.WORKFLOW_TASK_FAILED_CAUSE_FEATURE_DISABLED, failWFTErr.Cause)
+		require.Empty(t, tcx.history.Events)
+	})
+
+	t.Run("chasm nexus not enabled", func(t *testing.T) {
+		tcx := newTestContext(t, &nexusoperation.Config{
+			Enabled:           dynamicconfig.GetBoolPropertyFn(true),
+			ChasmNexusEnabled: dynamicconfig.GetBoolPropertyFnFilteredByNamespace(false),
+		})
+		err := tcx.cancelHandler(tcx.chasmCtx, tcx.wf, commandValidator{maxPayloadSize: 1}, &commandpb.Command{}, command.HandlerOptions{WorkflowTaskCompletedEventID: 1})
+		require.ErrorIs(t, err, command.ErrNotSupported)
 		require.Empty(t, tcx.history.Events)
 	})
 


chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -286,13 +286,98 @@ func (ch *commandHandler) handleScheduleCommand(
...
+	if !operationFound && !hasBufferedEvent {
+		return command.FailWorkflowTaskError{
+			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_REQUEST_CANCEL_NEXUS_OPERATION_ATTRIBUTES,
+			Message: fmt.Sprintf("requested cancelation for a non-existing or already completed operation with scheduled event ID of %d", attrs.ScheduledEventId),
+		}
+	}
+
+	var op *nexusoperation.Operation
+	if operationFound {
+		op = operationField.Get(chasmCtx)
```
stephanos: The original in HSM had a TODO here
> // TODO(bergundy): Remove this when operation auto-deletes itself on terminal state.

but it seems to me that this is not applicable in CHASM.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -286,13 +286,98 @@ func (ch *commandHandler) handleScheduleCommand(
...
+	// batch, after command events from the workflow task.
+	event := wf.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED, func(he *historypb.HistoryEvent) {
+		he.Attributes = &historypb.HistoryEvent_NexusOperationCancelRequestedEventAttributes{
+			NexusOperationCancelRequestedEventAttributes: &historypb.NexusOperationCancelRequestedEventAttributes{
+				ScheduledEventId:             attrs.ScheduledEventId,
+				WorkflowTaskCompletedEventId: opts.WorkflowTaskCompletedEventID,
+			},
+		}
+		he.UserMetadata = cmd.UserMetadata
+	})
```
stephanos, line 344: Just as in the schedule command handler for CHASM, `Apply` is not invoked here anymore.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -286,13 +286,98 @@ func (ch *commandHandler) handleScheduleCommand(
...
+
+	if !operationFound {
+		// Operation was deleted on terminal state but the terminal event is buffered. Ignore.
+		return nil
+	}
+
+	err := op.Cancel(chasmCtx, event.GetEventId())
```
stephanos: 🤖 `small` **[duplication]**: AddHistoryEvent block for NEXUS_OPERATION_CANCEL_REQUESTED is duplicated.

**Issue:** The identical `AddHistoryEvent` call with the same attributes appears in two branches (here and in the `!ok` branch above at ~line 327). Both construct the same `NexusOperationCancelRequestedEventAttributes` struct with `ScheduledEventId`, `WorkflowTaskCompletedEventId`, and `UserMetadata`.

**Suggestion:** Consolidate by restructuring the flow to perform error checks first, then create the event in one place:

```go
if !ok && !hasBufferedEvent {
    return command.FailWorkflowTaskError{...not found...}
}

var op *nexusoperation.Operation
if ok {
    op = opField.Get(chasmCtx)
    if !nexusoperation.TransitionCanceled.Possible(op) && !hasBufferedEvent {
        return command.FailWorkflowTaskError{...already complete...}
    }
}

event := wf.AddHistoryEvent(enumspb.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED, ...)

if !ok {
    return nil
}

err := op.Cancel(chasmCtx, event.GetEventId())
...
```

service/history/historybuilder/history_builder.go
```diff
@@ -53,7 +53,7 @@ type (
 
 	TaskIDGenerator func(number int) ([]int64, error)
 
-	BufferedEventFilter func(*historypb.HistoryEvent) bool
+	BufferedEventFilter = func(*historypb.HistoryEvent) bool
```
stephanos, line 56: `MutableStateImpl.HasAnyBufferedEvent` uses `BufferedEventFilter` and since it needs to satisfy the `NodeBackend` interface, we need to make it an alias as we cannot import it (cycle).

chasm/lib/nexusoperation/config.go
```diff
@@ -10,7 +10,7 @@ import (
 	"go.temporal.io/server/common/rpc/interceptor"
 )
 
-var ChasmNexusEnabled = dynamicconfig.NewGlobalBoolSetting(
+var ChasmNexusEnabled = dynamicconfig.NewNamespaceBoolSetting(
```
stephanos, line 13: Changed to namespace scope to match `ChasmEnabled`.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -335,7 +335,7 @@ func (handler *workflowTaskCompletedHandler) handleCommand(
 		handlerOpts := chasmcommand.HandlerOptions{
 			WorkflowTaskCompletedEventID: handler.workflowTaskCompletedID,
 		}
-		if handler.mutableState.ChasmEnabled() {
+		if handler.config.NexusChasmEnabled(handler.mutableState.GetNamespaceEntry().Name().String()) {
```
stephanos: changed to `NexusChasmEnabled` otherwise MixedBrain fails.

chasm/lib/workflow/workflow.go
```diff
@@ -135,6 +136,14 @@ func (w *Workflow) AddHistoryEvent(t enumspb.EventType, setAttributes func(*hist
 	return w.MSPointer.AddHistoryEvent(t, setAttributes)
 }
 
+func (w *Workflow) HasAnyBufferedEvent(filter historybuilder.BufferedEventFilter) bool {
+	return w.MSPointer.HasAnyBufferedEvent(filter)
+}
+
+func (w *Workflow) RemoveNexusOperation(key string) {
+	delete(w.Operations, key)
+}
```
stephanos, line 149: Will be used by Chetan's workstream, too.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -286,13 +286,98 @@ func (ch *commandHandler) handleScheduleCommand(
...
+			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_REQUEST_CANCEL_NEXUS_OPERATION_ATTRIBUTES,
+			Message: "empty CancelNexusOperationCommandAttributes",
+		}
+	}
+
+	key := strconv.FormatInt(attrs.ScheduledEventId, 10)
+	operationField, operationFound := wf.Operations[key]
+	hasBufferedEvent := func() bool {
+		return wf.HasAnyBufferedEvent(makeNexusOperationTerminalEventFilter(attrs.ScheduledEventId))
+	}
```
bergundy: Why does this need to be a function?
```suggestion
	hasBufferedEvent := wf.HasAnyBufferedEvent(makeNexusOperationTerminalEventFilter(attrs.ScheduledEventId))
```

chasm/lib/nexusoperation/workflow/commands.go
bergundy: Not applicable, the operation should be deleted after completion in the CHASM implementation.

chasm/lib/nexusoperation/workflow/commands.go
bergundy: @gow is fixing this. We discussed this and said that `workflow.AddHistoryEvent` is going to apply the event. All of the logic below will move to the event definition (not blocking this PR).

chasm/lib/nexusoperation/workflow/commands.go
bergundy: You should be able to remove this entire if statement. The operation should either be deleted or the transition to canceled is possible.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -286,13 +286,98 @@ func (ch *commandHandler) handleScheduleCommand(
...
+
+	err := op.Cancel(chasmCtx, event.GetEventId())
+	if errors.Is(err, nexusoperation.ErrCancellationAlreadyRequested) {
+		return command.FailWorkflowTaskError{
+			Cause:   enumspb.WORKFLOW_TASK_FAILED_CAUSE_BAD_REQUEST_CANCEL_NEXUS_OPERATION_ATTRIBUTES,
+			Message: fmt.Sprintf("cancelation was already requested for an operation with scheduled event ID %d", attrs.ScheduledEventId),
+		}
+	}
+	if errors.Is(err, nexusoperation.ErrOperationAlreadyCompleted) {
```
bergundy: This should not be possible in a workflow.

chasm/lib/nexusoperation/workflow/commands.go
```diff
@@ -286,13 +286,98 @@ func (ch *commandHandler) handleScheduleCommand(
...
+		he.UserMetadata = cmd.UserMetadata
+	})
+
+	if !operationFound {
+		// Operation not found but there's a buffered terminal event. The workflow couldn't know
+		// the operation completed while its task was in flight. Ignore.
+		return nil
+	}
+
+	err := op.Cancel(chasmCtx, event.GetEventId())
```
bergundy, line 353: Note to self. Ideally all of the event ID tracking should be done in a location that is agnostic to where the component is embedded. Figure out a way to keep track of this in a way that the operation logic isn't directly affected.

chasm/lib/nexusoperation/cancellation.go
```diff
@@ -15,8 +15,8 @@ type Cancellation struct {
 	*nexusoperationpb.CancellationState
 }
 
-func NewCancellation() *Cancellation {
-	return &Cancellation{}
+func NewCancellation(state *nexusoperationpb.CancellationState) *Cancellation {
```
bergundy: I wonder if we want to export a constructor given that the cancelation should only be instantiated via `operation.Cancel`.

And if we decide to keep it, please add a short docstring.

chasm/lib/nexusoperation/config.go
bergundy: Good call!

chasm/lib/nexusoperation/operation.go
```diff
@@ -1,12 +1,22 @@
...
 import (
+	"errors"
+
 	"go.temporal.io/server/chasm"
 	nexusoperationpb "go.temporal.io/server/chasm/lib/nexusoperation/gen/nexusoperationpb/v1"
 )
 
-var _ chasm.Component = (*Operation)(nil)
-var _ chasm.StateMachine[nexusoperationpb.OperationStatus] = (*Operation)(nil)
+var (
```
bergundy: nit: There's no value in grouping these, it's just instructing Go documentation viewers to bundle all of these variables together instead of defaulting to displaying them in alphabetical order.

chasm/lib/nexusoperation/operation.go
```diff
@@ -1,12 +1,22 @@
...
+	_ chasm.Component                                      = (*Operation)(nil)
+	_ chasm.StateMachine[nexusoperationpb.OperationStatus] = (*Operation)(nil)
+
+	// ErrCancellationAlreadyRequested is returned when a cancellation has already been requested for an operation.
+	ErrCancellationAlreadyRequested = errors.New("cancellation already requested")
+
+	// ErrOperationAlreadyCompleted is returned when trying to cancel an operation that has already completed.
+	ErrOperationAlreadyCompleted = errors.New("operation already completed")
```
bergundy: Make these `serviceerror.NewFailedPrecondition`. That will take care of error handling for standalone ops.

chasm/lib/nexusoperation/operation.go
```diff
@@ -45,3 +57,23 @@ func (o *Operation) StateMachineState() nexusoperationpb.OperationStatus {
 func (o *Operation) SetStateMachineState(status nexusoperationpb.OperationStatus) {
 	o.Status = status
 }
+
+func (o *Operation) Cancel(ctx chasm.MutableContext, requestedEventID int64) error {
```
bergundy, line 66: Please document all exported members.

chasm/lib/nexusoperation/operation.go
```diff
@@ -19,6 +29,8 @@ type Operation struct {
 	// Pointer to an implementation of the "store". For a workflow-based Nexus operation
 	// this is a parent pointer back to the workflow. For a standalone Nexus operation this is nil.
 	Store chasm.ParentPtr[OperationStore]
+
+	Cancellation chasm.Field[*Cancellation]
```
bergundy, line 32: Missing doc string.

chasm/lib/nexusoperation/operation.go
```diff
@@ -45,3 +57,23 @@ func (o *Operation) StateMachineState() nexusoperationpb.OperationStatus {
...
+	if _, ok := o.Cancellation.TryGet(ctx); ok {
+		return ErrCancellationAlreadyRequested
+	}
+
+	cancellation := NewCancellation(&nexusoperationpb.CancellationState{
+		RequestedEventId: requestedEventID,
+	})
+	o.Cancellation = chasm.NewComponentField(ctx, cancellation)
+
+	if o.Status == nexusoperationpb.OPERATION_STATUS_STARTED {
```
bergundy, line 81:
```suggestion
	// Once started, the handler returns a token that can be used in the cancelation request.
	// Until then, no need to schedule the cancelation.
	if o.Status == nexusoperationpb.OPERATION_STATUS_STARTED {
```

chasm/lib/workflow/workflow.go
```diff
@@ -135,6 +136,14 @@ func (w *Workflow) AddHistoryEvent(t enumspb.EventType, setAttributes func(*hist
 	return w.MSPointer.AddHistoryEvent(t, setAttributes)
 }
 
+func (w *Workflow) HasAnyBufferedEvent(filter historybuilder.BufferedEventFilter) bool {
+	return w.MSPointer.HasAnyBufferedEvent(filter)
+}
+
+func (w *Workflow) RemoveNexusOperation(key string) {
```
bergundy, line 147: All exported members must be documented.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
bergundy: IMHO this should be ChasmEnabled and NexusChasmEnabled should be checked in the command handler. That's a cleaner way and will allow us to handle other commands in the future without changing this.

stephanos:
I believe I addressed all comments.

service/history/api/respondworkflowtaskcompleted/workflow_task_completed_handler.go
```diff
@@ -326,43 +324,33 @@ func (handler *workflowTaskCompletedHandler) handleCommand(
...
+		// Try CHASM command handler first, fall back to HSM if not supported.
+		handledByCHASM := false
+		if handler.config.ChasmEnabled(handler.mutableState.GetNamespaceEntry().Name().String()) {
+			if chasmHandler, ok := handler.chasmCommandRegistry.Handler(command.GetCommandType()); ok {
+				chasmWorkflow, chasmCtx, chasmErr := handler.mutableState.ChasmWorkflowComponent(ctx)
+				if chasmErr != nil {
+					return nil, chasmErr
+				}
+				err = chasmHandler(chasmCtx, chasmWorkflow, validator, command, handlerOpts)
+				handledByCHASM = !errors.Is(err, chasmcommand.ErrNotSupported)
```
bergundy, line 343: Side note: You'll still need to handle the case where the command is `RequestCancelNexusOperation` and a component (nexus operation) is in the HSM tree. I see that the TODO is still in line 327.

stephanos: I'll file a reminder for this; it was originally meant to be covered by the "Migration" ticket but we deprioritized that.

PR #9290, opened 2026-02-11
Overhaul Nexus error model

bergundy:
## What changed?

Replace the limited Nexus HandlerError proto with full Temporal Failure objects for transmitting errors across Nexus operation boundaries. This preserves failure chains, stack traces, and typed failure info (ApplicationFailure, CanceledFailure, etc.) end-to-end.

Care was taken to ensure Temporal->Temporal communication works as expected as well as support for non-Temporal Nexus implementations (to maintain support for external endpoints).
This ended up being more refactoring that I had originally anticipated but I think the unification of error serialization across the codebase and the more straightforward conversion logic was worth it.
The code was validated across variations of new and old caller and handler workers and servers using a [harness](https://github.com/bergundy/nexus-error-compat-tests) that was built specifically to validate the behavior. The harness was also run with an SDK setup that encoded failure attributes to confirm failures are properly transmitted over the different boundaries.

Proto changes:
- Add `failure` field to DispatchNexusTaskResponse carrying a Failure with NexusHandlerFailureInfo, deprecate `handler_error`
- Add `capabilities` field to nexus Request for feature negotiation

Failure conversion (`common/nexus/failure.go`):
- Rewrite TemporalFailureToNexusFailure and NexusFailureToTemporalFailure with support for recursive cause chains and stack traces
- Add special-case handling for NexusHandlerFailureInfo (serialized as nexus.HandlerError type) and OperationError (mapped to CanceledFailure/ApplicationFailure based on state)

Nexus SDK layer (`common/nexus/nexusrpc/`):
- Inline the failure converter from the upstream SDK with a custom FailureConverter interface that round-trips Temporal failure metadata through Nexus Failure objects
- Add a completion client to abstract away HTTP and provide consistent failure conversion across all APIs
- Support `unwrap-error` metadata marker so Temporal-to-Temporal calls can unwrap the OperationError envelope and recover the original cause

Frontend and backend handlers:
- Route responses through new failure path when caller sends `temporal-nexus-failure-support` header
- Handle both new `failure` and deprecated `handler_error` response types in nexus_handler.go and nexus_http_handler.go
- Update completion handling in history handler and mutable state to produce Temporal Failures instead of Nexus HandlerErrors
- Update matching engine to forward the new failure field

Dependencies:
- Bump `nexus-rpc/sdk-go` to pre-release with StackTrace, Cause, and OriginalFailure fields on Failure and OperationError

## Why?

Part of getting Nexus to GA in all SDKs, this change fixes a couple of notable issues with errors in Nexus+Temporal applications:
- Nexus SDK errors did not have a way to set an error message, which made them difficult to use in various languages
- The protocol used a custom failure format that is diverges from proxy expectations and may result in failures not being encrypted

## Potential risks
- Metric label values for callback outcomes now use Nexus handler error values instead of HTTP response codes. Alerts that check these outcomes will need adjustment.

### DIFF
--- a/common/nexus/nexusrpc/completion_test.go
+++ b/common/nexus/nexusrpc/completion_test.go
@@ -60,7 +60,7 @@ func (h *successfulCompletionHandler) CompleteOperation(ctx context.Context, com
 		return err
 	}
 	if result != 666 {
-		return nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "invalid result: %q", result)
+		return nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "invalid result: %d", result)
 	}
 	return nil
 }


api/matchingservice/v1/request_response.pb.go
```diff
@@ -3945,12 +3948,23 @@ func (x *DispatchNexusTaskResponse) GetRequestTimeout() *DispatchNexusTaskRespon
...
+	return nil
+}
+
 type isDispatchNexusTaskResponse_Outcome interface {
 	isDispatchNexusTaskResponse_Outcome()
 }
 
 type DispatchNexusTaskResponse_HandlerError struct {
-	// Set if the worker's handler failed the nexus task.
+	// Deprecated. Use failure field instead.
```
KeithB-Temporal, line 3965: Do we need two deprecated comments?

chasm/lib/callback/chasm_invocation.go
```diff
@@ -135,47 +134,44 @@ func (c chasmInvocation) getHistoryRequest(
...
 		}
-	case *nexusrpc.OperationCompletionUnsuccessful:
-		apiFailure, err := commonnexus.NexusFailureToAPIFailure(op.Failure, true)
+	} else {
+		failure, err := nexusrpc.DefaultFailureConverter().ErrorToFailure(c.completion.Error)
 		if err != nil {
-			return nil, fmt.Errorf("failed to convert failure type: %v", err)
+			return nil, fmt.Errorf("failed to convert error to failure: %w", err)
+		}
+		// Unwrap the operation error since it's not meant to be sent for Temporal->Temporal completions.
```
KeithB-Temporal: What does this part of the comment mean: "since it's not meant to be sent for Temporal-> Temporal completions" 

Consider instead, something like because Temporal completions should know about the error but not pass around the operation error object or something like that... (The meant to is confusing, but then I am thinking what is meant to be sent)

chasm/lib/callback/nexus_invocation.go
```diff
@@ -80,107 +61,54 @@ func (n nexusInvocation) Invoke(
...
-		return invocationResultRetry{err: err}
-	}
-	return invocationResultFail{err}
+	return invocationResultOK{}
 }
 
-// Reads and replaces the http response body and attempts to deserialize it into a Nexus failure. If successful,
-// returns a nexus.HandlerError with the deserialized failure as the Cause. If there is an error reading the body or
-// during deserialization, returns a nexus.HandlerError with a generic Cause based on response status.
-// TODO: This logic is duplicated in the frontend handler for forwarded requests. Eventually it should live in the Nexus SDK.
```
KeithB-Temporal, line 140: Does deleting this code mean that it now needs to be added in all of the SDKs, or is it already added to those SDKs?

common/nexus/nexusrpc/cancel_test.go
```diff
@@ -23,19 +23,19 @@ func (h *asyncWithCancelHandler) StartOperation(ctx context.Context, service, op
 
 func (h *asyncWithCancelHandler) CancelOperation(ctx context.Context, service, operation, token string, options nexus.CancelOperationOptions) error {
 	if service != testService {
-		return nexus.HandlerErrorf(nexus.HandlerErrorTypeBadRequest, "unexpected service: %s", service)
+		return nexus.NewHandlerErrorf(nexus.HandlerErrorTypeBadRequest, "unexpected service: %s", service)
```
KeithB-Temporal, line 26: Nit: I don't like this naming NewHandlerErrorf for 2 reasons:
1) New prefix almost makes it seem like it is is creating a new error object
2) I'm pretty sure the f represents format (why not just type out format to make it clear?) as is, if you miss the f at the end the code almost looks like a factory pattern to generate a new HandlerError.

Suggested Naming FormatHandlerErrorV2

common/nexus/nexusrpc/client.go
```diff
@@ -74,6 +74,66 @@ func newUnexpectedResponseError(message string, response *http.Response, body []
...
+		return newUnexpectedResponseError(err.Error(), response, body)
+	}
+	handlerErr := &nexus.HandlerError{
+		Type: errorType,
+		// For compatibility with older servers.
+		RetryBehavior: retryBehaviorFromHeader(response.Header),
+		Cause:         cause,
+	}
+
+	// Ensure we the original failure is available, the calling code expects it.
```
KeithB-Temporal: Nit: Remove "we" this looks like a vestige of a previous version of the comment.

common/nexus/nexusrpc/client.go
```diff
@@ -74,6 +74,66 @@ func newUnexpectedResponseError(message string, response *http.Response, body []
...
+	originalFailure, err := c.failureConverter.ErrorToFailure(handlerErr)
+	if err != nil {
+		return newUnexpectedResponseError("failed to construct handler error from response: "+err.Error(), response, body)
+	}
+	handlerErr.OriginalFailure = &originalFailure
+	return handlerErr
+}
+
+// BestEffortHandlerErrorFromResponse attempts to read a handler error from the response, but falls back to a default error.
+// This method is exposed as a workaround because the functionality is required for completion.
```
KeithB-Temporal: This method is a workaround to what problem. Or is the problem that that the functionality is required for completion. If so, I would just focus on that part. 
Suggestion: 
This method ensures that there is a valid error to return for completion.

common/nexus/nexusrpc/client.go
```diff
@@ -74,6 +74,66 @@ func newUnexpectedResponseError(message string, response *http.Response, body []
...
+func (c *baseHTTPClient) bestEffortHandlerErrorFromResponse(response *http.Response, body []byte) error {
+	failure, err := c.failureFromResponse(response, body)
+	if err != nil {
```
KeithB-Temporal, line 123: Suggest Add comment:
Could not parse a failure from response, return default error

common/nexus/nexusrpc/client.go
```diff
@@ -74,6 +74,66 @@ func newUnexpectedResponseError(message string, response *http.Response, body []
...
+		return c.defaultErrorFromResponse(response, body, nil)
+	}
+	convErr, err := c.failureConverter.FailureToError(failure)
```
KeithB-Temporal, line 126: Suggest add comment:
Failure from response could not be convert to a recognized failure type, return unexpectedResponseError

common/nexus/nexusrpc/client.go
```diff
@@ -74,6 +74,66 @@ func newUnexpectedResponseError(message string, response *http.Response, body []
...
+	if err != nil {
+		return newUnexpectedResponseError(fmt.Sprintf("failed to convert Failure to error: %s", err.Error()), response, body)
+	}
+	if _, ok := convErr.(*nexus.HandlerError); !ok {
```
KeithB-Temporal, line 130: Suggest Add comment:
Successfully parsed and convert the failure to a known failure type

common/nexus/nexusrpc/completion.go
```diff
@@ -179,36 +101,68 @@ type OperationCompletionUnsuccessfulOptions struct {
...
-func (c *OperationCompletionUnsuccessful) applyToHTTPRequest(request *http.Request) error {
-	if request.Header == nil {
-		request.Header = make(http.Header, len(c.Header)+2) // +2 for headerOperationState and content-type
+	// Set the body and operation state based on whether the completion is successful or not.
+	if c.Error != nil {
+		failure, err := cc.failureConverter.ErrorToFailure(c.Error)
+		if err != nil {
+			return err
+		}
+		// Backwards compatibility: if the failure has a cause, unwrap it to maintain the behavior as older servers.
```
KeithB-Temporal: Nit: Suggest improved wording:
Backwards compatibility: if the failure has a cause, unwrap it to ensure behavior remains compatible with older servers.

common/nexus/nexusrpc/failure_converter.go
```diff
@@ -0,0 +1,211 @@
...
+// FailureConverter is used by the framework to transform [error] instances to and from [Failure] instances.
+// To customize conversion logic, implement this interface and provide your implementation to framework methods such as
+// [NewHTTPClient] and [NewHTTPHandler].
+// By default the SDK translates [HandlerError], [OperationError] and [FailureError] to and from [Failure]
+// objects maintaining their cause chain.
+// Arbitrary errors are translated to a [Failure] object with its Message set to the Error() string, losing the cause
+// chain.
+type FailureConverter interface {
+	// ErrorToFailure converts an [error] to a [Failure].
+	// Note that the provided error may be nil.
```
KeithB-Temporal, line 19: Why is it important to Note that the provided error may be nil? Is this special case handled by the ErrorToFailure constructor, or is this something that the user of the Failure Converter needs to handle for...

common/nexus/nexusrpc/failure_converter.go
```diff
@@ -0,0 +1,211 @@
...
+// nolint:revive // Keeping all of the logic together for readability, even if it means the function is long.
+func (e knownErrorFailureConverter) ErrorToFailure(err error) (nexus.Failure, error) {
+	if err == nil {
+		return nexus.Failure{}, nil
+	}
+	// NOTE: not using errors.Unwrap here we are intentionally only supporting unwrapping known errors.
+	switch typedErr := err.(type) {
+	case *nexus.FailureError:
+		f := typedErr.Failure
+		// Convert the erorr cause if set and failure cause is not already set.
```
KeithB-Temporal: Two thoughts about this comment:
1) "erorr" is spelled wrong.
2) I don't understand the need for this comment, the logic below is pretty straightforward explaining the condition. Though after reading the comment, my thought is that entering this clause is the normal happy path....shouldn't f.Cause usually be nill at this point and under what conditions might that not be true...I think that might be the interesting thing to explain in a comment here.

common/nexus/nexusrpc/failure_converter.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	return nexus.HandlerErrorRetryBehaviorNonRetryable
+}
+
+type serializedOperationError struct {
+	State string `json:"state,omitempty"`
+}
+
+// ErrorToFailure implements FailureConverter.
+// nolint:revive // Keeping all of the logic together for readability, even if it means the function is long.
```
KeithB-Temporal, line 47: No sure how I feel about keeping all of this logic together for readability. There is significant overlap on the HandlerError and OperatorError cases, and I think it would improve maintainability to put those in a method. 

But if not putting those in a method I think each case should describe what handling it is doing for the errors.

common/nexus/nexusrpc/failure_converter.go
```diff
@@ -0,0 +1,211 @@
...
+		}, nil
+	}
+}
+
+// FailureToError implements FailureConverter.
+// nolint:revive // Keeping all of the logic together for readability, even if it means the function is long.
+func (e knownErrorFailureConverter) FailureToError(f nexus.Failure) (error, error) {
+	if f.Metadata != nil {
+		switch f.Metadata["type"] {
+		case "nexus.HandlerError":
```
KeithB-Temporal, line 137: Nit: There is a lot of overlap on these cases switches.  The only differences are name for error, and different parameters for the constructor. Could we make this code more compact with more re-use?

api/matchingservice/v1/request_response.pb.go
bergundy: This is generated code. Don't review it.

chasm/lib/callback/chasm_invocation.go
bergundy: How about this?


```suggestion
		// Unwrap the operation error, the handler on the other side is expecting to receive the underlying cause.
```

chasm/lib/callback/nexus_invocation.go
bergundy: The comment was outdated, we already moved the HTTP transport from the Nexus SDK into this repo under `common/nexus/nexusrpc`. I moved all error serialization logic into that package now and resolved this TODO.

common/nexus/nexusrpc/cancel_test.go
bergundy: `Errorf` is standard for Go. I would have preferred not to add the `New` prefix but we had to change the behavior from the old constructor with the addition of the `Message` field. To avoid breaking users, we ended up deprecating the old constructor.

common/nexus/nexusrpc/client.go
bergundy:
```suggestion
	// Ensure the original failure is available, the calling code expects it.
```

common/nexus/nexusrpc/failure_converter.go
```diff
@@ -0,0 +1,211 @@
...
+			return nil, err
+		}
+		fe.Cause = c
+	}
+	return fe, nil
+}
+
+var defaultFailureConverter FailureConverter = knownErrorFailureConverter{}
+
+// DefaultFailureConverter returns the SDK's default [FailureConverter] implementation. Translates [HandlerError],
```
Quinn-With-Two-Ns:
```suggestion
// DefaultFailureConverter returns the Servers default [FailureConverter] implementation. Translates [HandlerError],
```

The failure converter is no longer part of the Nexus SDK.

Quinn-With-Two-Ns:
Do we have any tests in the server repo that simulate an old SDK replaying to the Server with the old failure format? I couldn't find any, but maybe I just missed them.

service/frontend/nexus_http_handler.go
```diff
@@ -248,14 +237,15 @@ func (h *NexusHTTPHandler) baseNexusContext(apiName string) *nexusContext {
 		apiName:                              apiName,
 		requestStartTime:                     time.Now(),
 		responseHeaders:                      make(map[string]string),
+		callerFailureSupport:                 header.Get("temporal-nexus-failure-support") == "true",
```
Quinn-With-Two-Ns: nit: we repeat this same string 5 times by value, maybe we should make a constant?

common/nexus/nexusrpc/client.go
```diff
@@ -219,8 +284,10 @@ func (c *HTTPClient) StartOperation(
 	}
 	addContextTimeoutToHTTPHeader(ctx, request.Header)
 	addNexusHeaderToHTTPHeader(options.Header, request.Header)
+	// If this request is handled by a newer server that supports Nexus failure serialization, trigger that behavior.
+	request.Header.Set("temporal-nexus-failure-support", "true")
```
Quinn-With-Two-Ns: Do we want a dynamic config to disable this? I am thinking if we hit some bug in the new behaviour on the server or SDK or a user complains about the visible changes in behaviour?

common/nexus/nexusrpc/client.go
bergundy: Ah, the workaround comment is no longer relevant, I ended up not needing to expose this.

bergundy:
```suggestion
// bestEffortHandlerErrorFromResponse attempts to read a handler error from the response, but falls back to an unexpected response error.
```

common/nexus/nexusrpc/client.go
bergundy: Seems redundant TBH.

common/nexus/nexusrpc/client.go
bergundy: Same here. I don't typically like to add comments that describe what the code does, if it's unclear, the code should be rewritten. It's better to document why than what as a rule of thumb.

common/nexus/nexusrpc/completion.go
bergundy:
```suggestion
		// Backwards compatibility: if the failure has a cause, unwrap it to ensure behavior remains compatible with older servers.
```

common/nexus/nexusrpc/failure_converter.go
bergundy: It's a callout to implementors of this interface that they will need to handle this case.

common/nexus/nexusrpc/failure_converter.go
bergundy: Can you explain what you would want to see in the explanation apart from writing in English what the code does?

common/nexus/nexusrpc/failure_converter.go
```diff
@@ -0,0 +1,211 @@
...
+// Arbitrary errors are translated to a [Failure] object with its Message set to the Error() string, losing the cause
+// chain.
+// [Failure] instances are converted to [FailureError] to allow access to the full failure metadata and details if
+// available.
+func DefaultFailureConverter() FailureConverter {
+	return defaultFailureConverter
+}
+
+func retryBehaviorAsOptionalBool(e *nexus.HandlerError) *bool {
+	// nolint:exhaustive // this is a simple optional boolean.
```
KeithB-Temporal, line 203: This comment is redundant. The method signature is OptionalBool, I know we need to put in something to appease the linter, but not sure this comment does much.

common/nexus/nexusrpc/failure_converter.go
bergundy:
```suggestion
		// FailureError is has both a Cause error and an underlying Failure Cause.
		// When instantiated directly, there are cases where only the Go error cause is set.
		// Preserve the embedded failure's cause, as the embedded failure is treated similarly to the OriginalFailure field for well-known other error types.
```

common/nexus/nexusrpc/failure_converter.go
bergundy: Not really... most of the code here is for instantiating a specific Go struct.

common/nexus/failure.go
```diff
@@ -52,135 +54,254 @@ var failureTypeString = string((&failurepb.Failure{}).ProtoReflect().Descriptor(
...
+			Message:    failure.GetMessage(),
+			StackTrace: failure.GetStackTrace(),
+			Metadata: map[string]string{
+				"type": "nexus.HandlerError",
+			},
+			Details: details,
+			Cause:   causep,
+		}, nil
+	}
+	// Unset message and stack trace so it's not serialized in the details.
```
KeithB-Temporal, line 151: I really like this comment, to explain why this is necessary to do

components/nexusoperations/completion.go
```diff
@@ -70,12 +86,23 @@ func handleOperationError(
 
 		return FailedEventDefinition{}.Apply(node.Parent, event)
 	case nexus.OperationStateCanceled:
+		if originalCause.GetCanceledFailureInfo() == nil {
+			// Old SDKs may send an ApplicationFailure for canceled operation causes.
```
KeithB-Temporal, line 90: So in case nexus.OperationStateCanceled, if the .GetCanceledFailureInfo() == nil, then we know it is from an old SDK and we need to check for Application failure to get the cause.

Are there any other conditions that could trigger this block? It's also kindof inconclusive that we don't really know if we are in that state, we just assume that we are and read that info from the originalCause...

service/frontend/nexus_handler.go
```diff
@@ -46,6 +46,9 @@ const (
 // Generic Nexus context that is not bound to a specific operation.
 // Includes fields extracted from an incoming Nexus request before being handled by the Nexus HTTP handler.
 type nexusContext struct {
+	// Whether to use the new Temporal failure responses path.
+	// Set from the incoming nexus request's "temporal-nexus-failure-support" header.
```
KeithB-Temporal, line 50: So all new versions of SDKs will pass this header to opt into the new error types, right?

common/nexus/nexusrpc/failure_converter.go
bergundy: I typically like to explain why a linter exception is put in place next to the exception itself.

components/nexusoperations/completion.go
bergundy: The only place I saw this was with old SDKs in the synchronous path.

service/frontend/nexus_handler.go
bergundy: The caller server header this flag, not the SDK.

common/nexus/nexusrpc/failure_converter.go
bergundy: That's true...
```suggestion
// DefaultFailureConverter returns the package's default [FailureConverter] implementation. Translates [HandlerError],
```

service/frontend/nexus_http_handler.go
bergundy: This is a matter of taste IMHO. I've seen arguments in either direction. I can change if it's deemed important enough.

service/history/workflow/workflow_test/mutable_state_impl_test.go
```diff
@@ -379,15 +378,13 @@ func TestGetNexusCompletion(t *testing.T) {
...
-				require.Equal(t, "1", success.Reader.Header.Get("length"))
-				buf, err := io.ReadAll(success.Reader)
-				require.NoError(t, err)
-				require.Equal(t, []byte("3"), buf)
-				require.Equal(t, event.GetEventTime().AsTime(), success.CloseTime)
+			verifyCompletion: func(t *testing.T, event *historypb.HistoryEvent, completion nexusrpc.CompleteOperationOptions) {
+				require.Nil(t, completion.Error)
+				require.Equal(t, &commonpb.Payload{
+					Metadata: map[string][]byte{"encoding": []byte("json/plain")},
+					Data:     []byte("3"),
```
KeithB-Temporal, line 385: Why "3"?

bergundy:
> Do we have any tests in the server repo that simulate an old SDK replaying to the Server with the old failure format? I couldn't find any, but maybe I just missed them.

We actually only have those tests (`nexus_api_test.go`, and also some SDK usage in `nexus_workflow_test.go` which currently uses the old SDK). Let me add functional test coverage for the new response formats.

service/history/workflow/workflow_test/mutable_state_impl_test.go
bergundy: I just wanted a valid json string, not sure why I ended up with 3 :)

common/nexus/nexusrpc/client.go
bergundy: I think it would be best to move this header injection to the call site in this case. I'm _mostly_ trying to keep this package Temporal agnostic, plumbing in dynamic configuration is a step too far.

service/frontend/nexus_http_handler.go
KeithB-Temporal: +1 for a constant. I do agree that seeing it inline makes it obvious and discoverable what the string is, but if it is being used in multiple places I like a constant so if it ever needs to evolve, you can update it once or discover all places used very easily.

common/nexus/nexusrpc/cancel_test.go
KeithB-Temporal: Ugg... I don't love it, but understand.

common/nexus/nexusrpc/client.go
KeithB-Temporal: Maybe, but with a quick comment, you can scan this code in 5 seconds and understand each condition, rather than take a minute to figure out the differences between the different blocks.

common/nexus/nexusrpc/failure_converter.go
KeithB-Temporal: Could you update the comment to clarify that? I miss in .NET where. you can define in the header for the method all of the inputs, outputs and exceptions...

common/nexus/nexusrpc/failure_converter.go
KeithB-Temporal: The code in *nexus.HandlerError and *nexus.OperationError: is 90% the same. I would move that to a sub-method that takes a parameter and just put if blocks around the two places it is different. Sad that Go doesn't have a version of the ternary  operator...
https://stackoverflow.com/questions/19979178/what-is-the-idiomatic-go-equivalent-of-cs-ternary-operator

bergundy:
> > Do we have any tests in the server repo that simulate an old SDK replaying to the Server with the old failure format? I couldn't find any, but maybe I just missed them.
> 
> We actually only have those tests (`nexus_api_test.go`, and also some SDK usage in `nexus_workflow_test.go` which currently uses the old SDK). Let me add functional test coverage for the new response formats.

[Refactor API tests to include new failure paths](https://github.com/temporalio/temporal/pull/9290/commits/ae9814e4dae173209abb21748913e713732ce7cb)

service/frontend/nexus_http_handler.go
bergundy: Done.

common/nexus/nexusrpc/client.go
bergundy: Done.

PR #9296, opened 2026-02-11
CHASM: fix force terminate checks

yycptt:
## What changed?
- Prevent terminated executions from updating execution state which will result in an error
- Handle terminated executions when checking for access rules
- Wherever the existing terminated flag is checked, also check the persisted execution state in mutable state.

## Why?
- Current checks uses the in-memory terminated field on a node which will be lost after reload or replication.
- No real impact today as terminated execution don't really need any updates today, but that won't be true in the future when chasm executions can have callbacks.

## How did you test it?
- [ ] built
- [x] added new unit test(s)

### DIFF
--- a/chasm/tree.go
+++ b/chasm/tree.go
@@ -599,9 +599,11 @@ func (n *Node) isDetached() bool {
 	if componentAttr == nil {
 		return false
 	}
-	if componentAttr.GetTypeId() == CallbackComponentID {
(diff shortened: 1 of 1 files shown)


chasm/callback.go
yycptt, line 1: have to define it in the framework to avoid cycle dependency...

PR #9300, opened 2026-02-12
Backlog metrics (count, age) to now have versioning based attribution

Shivs11:
## What changed?
- WISOTT
- In simpler terms, the backlog count metric that is emitted today has a discrepancy with the value that is returned from the DescribeVersionAPI. The main discrepancy is that the backlog count *metric*  of an un-versioned queue does not take into consideration the current/ramping version, which means that it over-counts and the ramping/current version emit under-counts. This PR changes that by resolving the discrepancy.
- Additionally, I have added new metrics called *physical_backlog_count/age*. These metrics are basically going to emit the values that the current backlog metrics (before this PR) were emitting. The main purpose of these newly added metrics is to add operational value.
- Moreover, I moved the metric emission to the task queue partition manager since it felt more appropriate.

## Why?
- Correctness!

## How did you test it?
- [ ] built
- [x] added new unit test(s)

## Potential risks
- yes, in the sense that I am altering values that a metric that is in prod and customer facing. 

---

> [!NOTE]
> **Medium Risk**
> Changes customer-facing matching backlog gauges and adds periodic emission logic, which can affect dashboards/alerts and introduces timing/cardinality edge cases (especially around version tags and unload zeroing).
> 
> **Overview**
> Aligns backlog `approximate_backlog_count`/`approximate_backlog_age_seconds` gauges with `DescribeTaskQueue` by **periodically emitting version-attributed (logical) backlog metrics** from the partition manager, including splitting the default queue’s backlog across current/ramping worker versions.
> 
> Introduces new **physical** backlog gauges (`physical_approximate_backlog_count`/`physical_approximate_backlog_age_seconds`) to preserve the prior, un-attributed behavior, adds `matching.backlogMetricsEmitInterval` to control emission (0 disables), and updates unload/stop paths to zero out both logical and physical gauges to avoid stale series.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2ae0a9c86975cfda9e4e1063dc910b77cd38cfbf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


service/matching/task_queue_partition_manager.go
```diff
@@ -1320,6 +1321,95 @@ func subtractStatsByPriority(into, sub map[int32]*taskqueuepb.TaskQueueStats) {
 	}
 }
 
+// updateBacklogMetricsAttribution sets adjustment closures on each relevant physical
+// queue so that backlog metrics reflect versioning attribution. Called on initialize()
+// and userDataChanged().
+func (pm *taskQueuePartitionManagerImpl) updateBacklogMetricsAttribution() {
```
ShahabT: I'm not sure about all these repetitive logic. I think we should just have a separate goroutine in partition manager level that calls `Describe` on the partition once a minute and emit the age and count for each version+priority.

Previously, we were logging "physical" backlog, now we want to log "logical" backlog, once you see it that way, it'd be clear that the db/backlogmanager/physicalQueue should not bear the responsibility.

Separately, I think physical backlog for default queue is important for investigation still. Let's keep emitting that as a separate metric such as `default_queue_backlog_count` or something. (That can be done in the old place)

Shivs11: > Previously, we were logging "physical" backlog, now we want to log "logical" backlog, once you see it that way

I have a feeling that would be quite a big refactoring; not saying no, just putting it out there.

Shivs11: I quite like your comment about thinking of this from a "logical" pov than a "physical" pov; however, if we were to that, we would have to make our customers aware about:

- unversioned: logical <= physical 
- versioned queues (any arbitrary version): logical >= physical

I think this is important to surface this to our customers *if* we have both these metrics (although I do wonder if we will be leaking any implementation details if we do so)

maybe we should not have the physical counts external anymore and only use it for operational issues.
as for the naming, honestly, something such as having "physical" in it would make a lot of sense. I think folks reading our code for the first time would quickly understand the difference between "physical" and "logical"

service/matching/task_queue_partition_manager.go
```diff
@@ -1120,6 +1121,65 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
...
+		interval := pm.config.BacklogMetricsEmitInterval()
+		if interval == 0 { // disabled
+			_ = util.InterruptibleSleep(ctx, time.Minute)
+			continue
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(backoff.Jitter(interval, 0.05)):
```
ShahabT: I'm not sure if a jitter will help here. it's gonna mess with metrics unnecessarily.

service/matching/task_queue_partition_manager.go
```diff
@@ -1120,6 +1121,65 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
 	return backlogPriority
 }
 
+func (pm *taskQueuePartitionManagerImpl) emitLogicalBacklogMetrics(ctx context.Context) error {
+	for {
+		interval := pm.config.BacklogMetricsEmitInterval()
+		if interval == 0 { // disabled
+			_ = util.InterruptibleSleep(ctx, time.Minute)
```
ShahabT, line 1129: @dnr did a similar thing [here](https://github.com/temporalio/temporal/blob/b85b3bc35c51f57b09501b8479dc97d640d3517e/service/matching/fair_backlog_manager.go#L214) in a different way, not sure if any is better than the other. just wanted to mention

service/matching/task_queue_partition_manager.go
```diff
@@ -1120,6 +1121,78 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
...
+			continue
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(interval):
+			pm.fetchAndEmitLogicalBacklogMetrics(ctx)
+		}
+	}
```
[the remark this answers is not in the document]
Shivs11: oh wow, good catch.
we should prolly have this in the ephemeral data change function too!

service/matching/task_queue_partition_manager.go
```diff
@@ -249,6 +250,7 @@ func (pm *taskQueuePartitionManagerImpl) Stop(unloadCause unloadCause) {
 	pm.engine.updateTaskQueuePartitionGauge(pm.Namespace(), pm.partition, -1)
 
 	pm.goroGroup.Cancel()
+	pm.emitZeroLogicalBacklogMetrics()
```
Shivs11: there is one tiny race condition here which I would like to point out - from the documentation, it seems like pm.goroGroup.Cancel() does not actually ensure that all the spawned go-routines (in our case, there are 2) complete. It just cancels their context. Thus, there could be a world where we go ahead and emit zero valued metrics but then immediately have non-zero backlog values enabled (flip-flapping)

the way one could solve this is to use something like goroGroup.Wait(); However, I wonder if using something like that is fine since if a go-routine does hang, then well, we would be stuck on the wait call (reading this from the documentation)

My hypothesis is that we should keep things as is implemented in this PR as the race condition's probability is small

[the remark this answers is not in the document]
Shivs11: i have placed a comment above. think having .wait is not safe in prod too

Shivs11: i just moved the zeroing out logic to the physical task queue manager - i think the risk still exists, but it is a low one given that the .Stop() method do cancel contexts and a racy Describe call would just have stale data

service/matching/physical_task_queue_manager.go
```diff
@@ -323,12 +323,23 @@ func (c *physicalTaskQueueManagerImpl) Stop(unloadCause unloadCause) {
...
 
+func (c *physicalTaskQueueManagerImpl) emitZeroLogicalBacklogMetrics() {
+	if !c.config.BreakdownMetricsByTaskQueue() || !c.config.BreakdownMetricsByPartition() {
+		return
+	}
+	for pri := range c.GetStatsByPriority(false) {
+		metrics.ApproximateBacklogCount.With(c.metricsHandler).Record(0, metrics.MatchingTaskPriorityTag(pri))
+	}
+	metrics.ApproximateBacklogAgeSeconds.With(c.metricsHandler).Record(0)
+}
```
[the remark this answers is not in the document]
ShahabT: this is important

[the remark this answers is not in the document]
ShahabT: also we should emit zero for each version and pri, not only pri.

service/matching/task_queue_partition_manager.go
ShahabT: I think we should just return from the whole thing in case of config being disabled. it's totally fine for this config to require restart to take effect.

common/dynamicconfig/constants.go
```diff
@@ -1279,6 +1274,11 @@ This can help reduce effects of task queue movement.`,
 		`How often to update ephemeral data (e.g. backlog size for forwarding sticky polls).
 Set to zero to disable ephemeral data updates.`,
 	)
+	MatchingBacklogMetricsEmitInterval = NewTaskQueueDurationSetting(
+		"matching.backlogMetricsEmitInterval",
+		time.Minute,
+		`How often to emit attributed backlog metrics.`,
```
carlydf: "attributed backlog metrics" with no context is difficult to understand. will read more before proposing a better description

service/matching/db.go
```diff
@@ -733,13 +733,15 @@ func (db *taskQueueDB) cachedQueueInfo() *persistencespb.TaskQueueInfo {
...
 
-// emitBacklogGaugesLocked emits the approximate_backlog_count, approximate_backlog_age_seconds, and the legacy
-// task_lag_per_tl gauges. For these gauges to be emitted, BreakdownMetricsByTaskQueue and BreakdownMetricsByPartition
-// should be enabled. Additionally, for versioned queues, BreakdownMetricsByBuildID should also be enabled.
-func (db *taskQueueDB) emitBacklogGaugesLocked() {
+// emitPhysicalBacklogGaugesLocked emits the physical_approximate_backlog_count,
+// physical_approximate_backlog_age_seconds, and the legacy task_lag_per_tl gauges.
+// Physical backlog metrics are only emitted for the default (unversioned) queue.
+// Attributed backlog metrics for all versions (including current/ramping attribution)
+// are emitted separately by the partition manager via fetchAndEmitLogicalBacklogMetrics.
```
carlydf:
```suggestion
// emitPhysicalBacklogGaugesLocked emits the physical_approximate_backlog_count,
// physical_approximate_backlog_age_seconds, and the legacy task_lag_per_tl gauges
// tagged by priority key.
// Physical backlog metrics are only emitted for the default (unversioned) queue.
// Version-attributed backlog metrics for all worker versions (including appropriate current / 
// ramping attribution of tasks in the default queue) are emitted separately by the partition 
// manager via fetchAndEmitLogicalBacklogMetrics.
```

service/matching/db.go
```diff
@@ -733,13 +733,15 @@ func (db *taskQueueDB) cachedQueueInfo() *persistencespb.TaskQueueInfo {
...
+func (db *taskQueueDB) emitPhysicalBacklogGaugesLocked() {
 	if !db.config.BreakdownMetricsByTaskQueue() ||
 		!db.config.BreakdownMetricsByPartition() ||
-		(db.queue.IsVersioned() && !db.config.BreakdownMetricsByBuildID()) {
+		db.queue.IsVersioned() {
```
carlydf, line 746: confirming my understanding: we are removing the build id breakdown check here because physical backlog count is no longer broken down by build id?

common/dynamicconfig/constants.go
carlydf:
```suggestion
		`How often to emit version-attributed backlog metrics. Done on an interval because accurate attribution requires checking the routing config of a task queue to correctly count the tasks in the default queue to the appropriate current or ramping versions.`,
```

service/matching/task_queue_partition_manager.go
carlydf: just confirming my understanding: if we just returned if disabled, enabling it would require matching restart. I feel like that's fine too, especially to reduce code complexity / risk

carlydf: that being said, it seems to me like this for loop correctly:
- if disabled, checks dynamic config every minute to refresh
- if not disabled, returns error if `Done` or calls `pm.fetchAndEmitLogicalBacklogMetrics(ctx)` once per interval

Shivs11: sorry, why is it fine for a matching restart to happen if we want to toggle this dynamic config? I think the whole point of a dynamic config is to be dynamic right? 

+, the cost of having things as-is right now is just an additional go-routine with a pattern that has already been implemented by david while playing with the user data

common/dynamicconfig/constants.go
```diff
@@ -1279,6 +1274,11 @@ This can help reduce effects of task queue movement.`,
 		`How often to update ephemeral data (e.g. backlog size for forwarding sticky polls).
 Set to zero to disable ephemeral data updates.`,
 	)
+	MatchingBacklogMetricsEmitInterval = NewTaskQueueDurationSetting(
+		"matching.backlogMetricsEmitInterval",
+		time.Minute,
+		`How often to emit version-attributed backlog metrics. Done on an interval because accurate attribution requires checking the routing config of a task queue to correctly attribute the default queue's tasks to the appropriate current or ramping versions.`,
```
carlydf:
```suggestion
		`How often to emit version-attributed backlog metrics. Done on an interval because accurate attribution requires checking the routing config of a task queue to correctly attribute the default queue's tasks to the appropriate current or ramping versions. Set to zero to disable versioned backlog data updates.`,
```

carlydf: should we say this too?

service/matching/task_queue_partition_manager.go
carlydf: I just looked at the David code that Shahab linked (https://github.com/temporalio/temporal/blob/b85b3bc35c51f57b09501b8479dc97d640d3517e/service/matching/fair_backlog_manager.go#L214), and the only difference is that `UpdateAckInterval` does not have the "set to 0 to disable" feature.

on the other hand, Shivam's code here exactly mirrors the pattern of `func (pm *taskQueuePartitionManagerImpl) updateEphemeralData(ctx context.Context)` which has the same behavior of "set to 0 to disable" so that's why it is different `UpdateAckInterval`.

I think it's safe to leave this as is.

service/matching/task_queue_partition_manager.go
```diff
@@ -1120,6 +1122,96 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
...
+		return
+	}
+	handler := pm.metricsHandler.WithTags(
+		metrics.WorkerVersionTag(version.MetricsTagValue(), pm.config.BreakdownMetricsByBuildID()),
+	)
+	for pri := range pq.GetStatsByPriority(false) {
+		metrics.ApproximateBacklogCount.With(handler).Record(0, metrics.MatchingTaskPriorityTag(pri))
+	}
+	metrics.ApproximateBacklogAgeSeconds.With(handler).Record(0)
+}
```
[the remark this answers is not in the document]
Shivs11: yeah, great catch honestly.

service/matching/task_queue_partition_manager.go
ShahabT: I'm fine if you guys are comfortable with it.

> whole point of a dynamic config is to be dynamic right

There are two types of dynamic config: hot and cold. cold requires restart and is ok for things that do not change often and need additional logic for being hot. They're still super valuable because they don't need code change and deployment.

service/matching/task_queue_partition_manager.go
```diff
@@ -1120,6 +1122,99 @@ func (pm *taskQueuePartitionManagerImpl) updateEphemeralDataIteration(prevBacklo
...
+// reflect versioning attribution: for current/ramping versions, a proportional share of the
+// unversioned queue's backlog is added to their count, and the unversioned queue's count is
+// reduced accordingly. This ensures metrics match what DescribeTaskQueue returns.
+func (pm *taskQueuePartitionManagerImpl) fetchAndEmitLogicalBacklogMetrics(ctx context.Context) {
+	if !pm.config.BreakdownMetricsByTaskQueue() || !pm.config.BreakdownMetricsByPartition() {
+		return
+	}
+
+	buildIds := map[string]bool{"": true} // include unversioned
+	resp, err := pm.Describe(ctx, buildIds, true, true, false, false)
```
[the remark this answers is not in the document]
carlydf: Fixed in https://github.com/temporalio/temporal/pull/9316/

PR #9304, opened 2026-02-12
Last part of making Nexus work OOTB

bergundy:
## What changed?
Set `component.nexusoperations.useSystemCallbackURL` to true by default.

## Why?
It's now safe to turn this on since server v.130 supports this format and is a valid rollback target.


KeithB-Temporal (approved):
Is there any field comms or enablement required for making this change? If so, I assume you are already coordinating this with Phil.

PR #9305, opened 2026-02-12
Batch activity cancel tasks by control queue

rkannan82:
Collect cancel requests during workflow task completion processing and create one CancelActivityNexusTask per control queue at the end. This minimizes the number of Nexus dispatch operations when multiple activities on the same worker are cancelled together.

## Changes
- Updated AddCancelActivityNexusTasks signature to accept multiple scheduled event IDs and control queue
- Handler collects cancellations by control queue during command processing
- Creates batched tasks via mutable state API at end of workflow task completion


rkannan82:
Superseded by the generalized WorkerCommandsTask approach in task-definition and dispatch-logic branches.

PR #9306, opened 2026-02-12
CHASM: Avoid double map lookup in registry

yycptt:
## What changed?
- Avoid double map look up when getting registableComponent/Task by ID

## Why?
- Minor performance improvement.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/chasm/registry.go
+++ b/chasm/registry.go
@@ -18,14 +18,17 @@ var (
 
 type (
 	Registry struct {
-		libraries         map[string]Library                     // library name -> library
-		componentByType   map[string]*RegistrableComponent       // fully qualified type name -> component
-		componentFqnByID  map[uint32]string                      // component ID -> fully qualified type name
-		componentByGoType map[reflect.Type]*RegistrableComponent // component go type -> component
+		libraries map[string]Library // library name -> library
 
-		taskByType   map[string]*RegistrableTask       // fully qualified type name -> task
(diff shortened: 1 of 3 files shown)


chasm/registry.go
```diff
@@ -77,28 +80,30 @@ func (r *Registry) RegisterServices(server *grpc.Server) {
...
 
 // ComponentByID returns the registrable component for a given archetype ID.
+// This method should only be used by CHASM framework internal code,
+// NOT CHASM library developers.
 func (r *Registry) ComponentByID(id uint32) (*RegistrableComponent, bool) {
-	fqn, ok := r.componentFqnByID[id]
-	if !ok {
-		return nil, false
-	}
-	return r.component(fqn)
```
yycptt, line 101: This is the main change. It used to be two lookups, from ID -> FQN -> Component.

chasm/registry.go
```diff
@@ -113,16 +118,22 @@ func (r *Registry) ComponentIDFor(componentInstance any) (uint32, bool) {
 }
 
 // TaskByID returns the registrable task for a given task type ID.
+// This method should only be used by CHASM framework internal code,
+// NOT CHASM library developers.
 func (r *Registry) TaskByID(id uint32) (*RegistrableTask, bool) {
-	return r.taskByID(id)
```
yycptt, line 117: same here for Task.

PR #9309, opened 2026-02-12
CHASM: support archetypeID in admin handler

yycptt:
## What changed?
- Accept archetypeID in admin API requests

## Why?
- With this we no longer need to register chasm components to worker service, which currently performs an archetypeID to name conversion before calling admin apis.
- For backward compatibility, we can only stop registering chasm components to worker service starting from next cloud release.

## How did you test it?
- [ ] built
- [ ] added new unit test(s)

### DIFF
--- a/proto/internal/temporal/server/api/adminservice/v1/request_response.proto
+++ b/proto/internal/temporal/server/api/adminservice/v1/request_response.proto
@@ -57,6 +57,8 @@ message DescribeMutableStateRequest {
   temporal.api.common.v1.WorkflowExecution execution = 2;
   bool skip_force_reload = 3;
   string archetype = 4;
+  // (-- api-linter: core::0141::forbidden-types=disabled --)
+  uint32 archetype_id = 5;
 }
 
 message DescribeMutableStateResponse {
@@ -363,6 +365,8 @@ message RefreshWorkflowTasksRequest {
   string namespace_id = 3;
   temporal.api.common.v1.WorkflowExecution execution = 2;
   string archetype = 4;
+  // (-- api-linter: core::0141::forbidden-types=disabled --)
+  uint32 archetype_id = 5;
 }
 
 message RefreshWorkflowTasksResponse {
@@ -403,6 +407,8 @@ message DeleteWorkflowExecutionRequest {
   string namespace = 1;
   temporal.api.common.v1.WorkflowExecution execution = 2;
   string archetype = 3;
+  // (-- api-linter: core::0141::forbidden-types=disabled --)
+  uint32 archetype_id = 4;
 }
 
 message DeleteWorkflowExecutionResponse {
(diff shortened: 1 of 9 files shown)


tools/tdbg/tdbg_commands.go
```diff
@@ -155,6 +155,10 @@ func newAdminExecutionCommands(clientFactory ClientFactory, prompterFactory Prom
 					Usage:       "Fully qualified archetype name of the execution",
 					DefaultText: chasm.WorkflowArchetype,
 				},
+				&cli.UintFlag{
```
bergundy, line 169: I wonder if anyone is going to use the numeric ID. What's the use case?

bergundy (commented):
Could you also remove the registration of the workflow and scheduler libraries from the other services?

bergundy:
> Could you also remove the registration of the workflow and scheduler libraries from the other services?

I see that we will need it for a while longer but let's make sure we do that soon.
We will want to break up all of the libraries into the component only parts and other parts with more dependencies for registration in the frontend to handle visibility APIs.

yycptt:
> > Could you also remove the registration of the workflow and scheduler libraries from the other services?
> 
> I see that we will need it for a while longer but let's make sure we do that soon. We will want to break up all of the libraries into the component only parts and other parts with more dependencies for registration in the frontend to handle visibility APIs.

yeah that will have to wait until next cloud release. I will have the work tracked.

tools/tdbg/tdbg_commands.go
yycptt: I am thinking If any log message logs the ID or if we are looking at raw persistence records that would be useful.
Update: this happened recently when our persistence team is looking to internal persistence records which only has the ID

I guess I was also thinking: why not supporting given It's optional, and not complicating the logic a lot. The help message is also clear re. the behavior.

proto/internal/temporal/server/api/adminservice/v1/request_response.proto
```diff
@@ -57,6 +57,8 @@ message DescribeMutableStateRequest {
   temporal.api.common.v1.WorkflowExecution execution = 2;
   bool skip_force_reload = 3;
   string archetype = 4;
+  // (-- api-linter: core::0141::forbidden-types=disabled --)
+  uint32 archetype_id = 5;
```
yux0, line 61: So if this is not provided, it will be WorkflowArchetypeID correct?

yycptt: it will first fallback to the archetype (name) field, if that's also empty, then assume it to be a Workflow.

Fallback logic can be found here.
https://github.com/temporalio/temporal/pull/9309/changes#diff-ba52f3d4055774030a3132e40f78f29de18ed6ea1aa407f22bd109e47e453cfeR2381

